PR #24, opened 2023-06-07
Add `streamToResponse`

shuding:
This can be used in Node.js (or Next.js pages/ API route) like:

```js
const stream = OpenAIStream(await fetch(...))
streamToResponse(stream, res)
```

Where `res` is a `ServerResponse` instead of a `Response`.

Using the SDK in Node.js requires Node.js 18+.

### DIFF
--- a/.changeset/clever-crabs-train.md
+++ b/.changeset/clever-crabs-train.md
@@ -1,0 +1,5 @@
+---
+'ai-connector': patch
+---
+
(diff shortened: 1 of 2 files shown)


jaredpalmer (changes requested):
Need to add docs for Node 18+/runtime compatibility in getting-started.md

jaredpalmer (approved):

PR #47, opened 2023-06-13
Add onError callback, include response text in error if response is not okay

MaxLeiter:
- Adds an onError callback to useChat/useCompletion
- Tries to parse response text if !response.ok instead of the generic error message


shuding (commented):
Thanks! We need to also re-export `UseChatOptions` from the original files as users might want to import it

shuding (approved):

PR #66, opened 2023-06-15
Update langchain.mdx

alibey-10:
change to export default


A7med3bdulBaset:
Hi @Ali-Hussein-dev 
Route handlers in Next.js 13 are exported namely.

```ts
export async function GET() {} 
export async function POST() {}
export async function PUT() {}
export async function PATCH() {}
export async function DELETE() {}
export async function OPTIONS() {}
```

PR #89, opened 2023-06-16
feat: add createdAt on user input message

ahmadrosid:
Currently createdAt only available on message from role `assistant`.


jaredpalmer (approved):

MaxLeiter:
hey @ahmadrosid, thanks for the contribution! could you update it with `main` + run `pnpm prettier-fix` so we can merge?

PR #95, opened 2023-06-16
use `$env/static/private` for API key

Rich-Harris:
This is a more idiomatic approach in SvelteKit than `process.env` — it means e.g. people get a useful build time error if no API key is present, rather than it failing at runtime


phenomen:
Also this example doesn't seem to use Vercel Edge since it has `adapter-auto` instead of `adapter-vercel` and doesnt export edge runtime config in the endpoint.

```ts
export const config: Config = {
	runtime: 'edge'
};
```

MaxLeiter (approved):

PR #107, opened 2023-06-16
chore(node): node stream bypass should support backpressure

tangye1234:
Use a compatible way to support pipeline from web stream to node server response.

Node stream pipeline supports backpressure, so the server response sink will not suffer memory issues.
Implementations uses nodejs 18 web stream adapters as a reference and are ported from @hono/node-server@0.5.1 which is also used by many developers for a long time.

This implementation is compatible with stream-polyfill and nodejs native stream.

### DIFF
--- a/packages/core/streams/streaming-text-response.ts
+++ b/packages/core/streams/streaming-text-response.ts
@@ -1,4 +1,8 @@
-import type { ServerResponse } from 'node:http'
+import type { OutgoingHttpHeaders, ServerResponse } from 'node:http'
+import { Readable, pipeline } from 'node:stream'
+import { promisify } from 'node:util'
+
+const pipelinePromise = promisify(pipeline)
 
 /**
  * A utility class for streaming text responses.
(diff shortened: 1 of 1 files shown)


shuding (commented):
I don't think we need to export `headersInitToOutgoingHeaders` and `streamToNodeReadable` for now.

packages/core/streams/streaming-text-response.ts
```diff
@@ -16,29 +16,116 @@ export class StreamingTextResponse extends Response {
...
+
+  return readable
+}
+
+/**
+ * A utility function to get a OutgoingHttpHeaders from a HeadersInit.
+ */
+export function headersInitToOutgoingHeaders(headers: HeadersInit | undefined) {
+  const h = new Headers(headers)
+  const outgoingHeaders: OutgoingHttpHeaders = {}
```
jridgewell, line 82:
```suggestion
  const outgoingHeaders: OutgoingHttpHeaders = {__proto__: null}
```

packages/core/streams/streaming-text-response.ts
```diff
@@ -16,29 +16,116 @@ export class StreamingTextResponse extends Response {
...
+  h.forEach((value, key) => {
+    if (
+      key === 'set-cookie' &&
+      'getSetCookie' in h &&
+      typeof h.getSetCookie === 'function'
+    ) {
+      // nodejs 18.14.1 supports it, otherwise developer should polyfill this method for their own
+      outgoingHeaders[key] = h.getSetCookie()
+      return
+    }
```
jridgewell, line 92: I don't really like this special handling because it doesn't fail when `getSetCookie` isn't defined, and the fallback has subtly incorrect behavior. We can implement this with the same amount of code with:

```js
    const old = outgoingHeaders[key]
    if (Array.isArray(old)) {
      old.push(value)
    } else {
      outgoingHeaders[key] = old === undefined ? value : [String(old), value]
    }
```

packages/core/streams/streaming-text-response.ts
```diff
@@ -16,29 +16,116 @@ export class StreamingTextResponse extends Response {
...
+        .then(({ done, value }) => {
+          if (done) {
+            this.push(null)
+          } else {
+            this.push(value)
+          }
+        })
+        .catch(e => {
+          readable.destroy(e)
+        })
```
jridgewell, line 44: Let's use async-await to save a few promise allocations:
```suggestion
    async read() {
      try {
        const { done, value } = await reader.read()
        if (done) {
          this.push(null)
        } else {
          this.push(value)
        }
      } catch (e) {
        readable.destroy(e as Error)
      }
```

packages/core/streams/streaming-text-response.ts
```diff
@@ -16,29 +16,116 @@ export class StreamingTextResponse extends Response {
...
+            throw err
+          })
+        }
+      }
+
+      if (!closed) {
+        reader.cancel(error).then(done, done)
+        return
+      }
+      done()
```
jridgewell, line 61: Ditto:

```suggestion
    async destroy(error, callback) {
      try {
        if (!closed) {
          await reader.cancel(error)
        }
      } finally {
        callback(error)
      }
```

packages/core/streams/streaming-text-response.ts
```diff
@@ -16,29 +16,116 @@ export class StreamingTextResponse extends Response {
...
-        response.end()
-        return
+  // pipe from node readable to server response
+  // which supports backpresure mechanism
+  const readable = streamToNodeReadable(res)
+
+  const pipeline = require('node:stream')
+    .pipeline as typeof import('node:stream').pipeline
+  return new Promise<void>((resolve, reject) => {
+    pipeline(readable, response, error => {
```
jridgewell, line 123: We have an [implementation of this in the `edge-runtime` package](https://unpkg.com/edge-runtime@2.4.3/dist/server/body-streams.js) which should be a little faster because it skips the `Readable` indirection.

tangye1234: > We have an [implementation of this in the `edge-runtime` package](https://unpkg.com/edge-runtime@2.4.3/dist/server/body-streams.js) which should be a little faster because it skips the `Readable` indirection.

First Pipe in @edge/runtime dose not process possible reader error, which should have destroyed res with that error. Meanwhile it checks open state in every chunk cycle which is not as fast as you think. Third it is another package, and do we really need to import another or just leave this method alone?

For me, I think we should provide a nodejs stream conversion here for the ai sdk implements so many streams which is helpful and users can easily deploy their apps with only this package in nodejs.

packages/core/streams/streaming-text-response.ts
tangye1234: For users who have already polyfilled getSetCookie, this is OK, and if not， the fallback follows the original implementations which means users dont care.

Last you change headers set to headers append, which I think is overthought. In `Response`, there is only set-cookie key which really needs appending. Others like content-type should be only reserved once. Moreover content-encoding should be omitted, cause we already decode the content in the response reader.

packages/core/streams/streaming-text-response.ts
jridgewell: > First Pipe in @edge/runtime dose not process possible reader error, which should have destroyed res with that error

The `pipeBodyStreamToResponse` is designed so that the caller handles closing the response (in case it wants to send more data). We just need a `.then(destroy, destroy)` to handle.

> Meanwhile it checks open state in every chunk cycle which is not as fast as you think.

`open` is a local variable, so I'm not sure what you mean.

> Third it is another package, and do we really need to import another or just leave this method alone?

This is a Vercel library consuming another Vercel library, and I think commonly used inside Next.js (which will have already loaded the `edge-runtime` library). In that case, changing will actually reduce code at runtime. And if this isn't being used by a Next.js server, a bundler will be able to strip out everything not used by us.

jaredpalmer:
What's left before we can merge?

MaxLeiter:
hey @tangye1234, are you planning on continuing work on this PR?

tangye1234:
> hey @tangye1234, are you planning on continuing work on this PR?

Recently I have been a little busy. I just don’t know whether adding edge-runtime deps is a good idea. I think developers who want nodejs streaming for web standard stream can directly import edge-runtime on their own.

I will continue, but at least in a month later.

packages/core/streams/streaming-text-response.ts
```diff
@@ -1,4 +1,4 @@
-import type { ServerResponse } from 'node:http'
+import type { OutgoingHttpHeaders, ServerResponse } from 'node:http'
```
StephanieLeeeee, line 1: 1

StephanieLeeeee (changes requested):

StephanieLeeeee (approved):

StephanieLeeeee (approved):

PR #124, opened 2023-06-16
feat: text/raw/json stream modes, parsers

ctjlewis:
### Changes

- Modes: `text` and `raw` (for now)
- Pattern `AIStream(response, { mode, parsers, ...callbacks })`
- Pattern `OpenAIStream(response, { mode, ...callbacks })`

### Future recommendations

- Add JSON parsers to parse JSON from text completions, i.e. `JSON.parse(res.choices[0].content)`, for integration with [OpenAI Function Calling](https://openai.com/blog/function-calling-and-other-api-updates). Would require adding a new mode `json` and adding methods for `openaiParsers` and `anthropicParsers` (the only two `AIStream` implementations in the library right now). 

### Questions

Does `StreamingTextResponse` do something to resolve that weird production buffer issue on Vercel servers? The last few months working on openai-streams, it was usually sufficient to just return `new Response(ReadableStream)` without bothering with `Content-Type`, removing that could simplify it a little bit more but super curious if something is going over my head there.


packages/core/streams/index.test.ts
```diff
@@ -37,6 +37,31 @@ describe('AIStream', () => {
...
+            }
+          }),
+          { mode: 'raw' }
+        )
+
+        const response = new StreamingTextResponse(stream)
+        const client = createClient(response)
+
+        const rawChunks = await client.readAll()
+        const chunks = rawChunks.map(chunk => JSON.parse(chunk))
```
ctjlewis, line 55: In `raw` mode, the streams remain `ReadableStream<Uint8Array>` and the chunks come back as JSON strings. Here, they are parsed first simply so we don't stringify twice and have a bunch of escaped quotes.

packages/core/streams/parsers.ts
```diff
@@ -0,0 +1,53 @@
...
+import { trimStartOfStreamHelper } from './utils'
+
+export type AIParserPlatform = 'openai' | 'anthropic'
+
+export type AIStreamsParsersList = Record<
+  AIParserPlatform,
+  () => AIStreamParser
+>
+
+export const TEXT_PARSERS: AIStreamsParsersList = {
```
ctjlewis, line 11: Later, it might be good to add JSON parsers for integration with OpenAI function calling (or manual equivalents, e.g. `Respond only with valid JSON`-like prompts).

packages/core/streams/openai-stream.ts
```diff
@@ -1,27 +1,6 @@
...
-  }
-}
-
-export function OpenAIStream(
-  res: Response,
-  cb?: AIStreamCallbacks
-): ReadableStream {
-  return AIStream(res, parseOpenAIStream(), cb)
+export function OpenAIStream(res: Response, options: AIStreamOptions = {}) {
+  return AIStream(res, { platform: 'openai', ...options })
```
ctjlewis, line 5: Here's how platform implementations are set up under the new pattern: A platform is added to the parsers dictionary and `AIStream` will look it up. This lets us manage all of the parsers in one place (`parser.ts`), which for now only contains text parsers, but soon will probably contain others (such as JSON parsers).

ctjlewis:
Once we get this in honestly, OpenAI Function Calling integration will be probably just a few hour thing.

packages/core/streams/parsers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+        // this can be used for either chat or completion models
+        const text = trimStartOfStream(
+          json.choices[0]?.delta?.content ?? json.choices[0]?.text ?? ''
+        )
+
+        return text
+      }
+  }
+}
```
ctjlewis, line 26: We can easily add JSON parsing on completions for integration with [OpenAI function calling](https://openai.com/blog/function-calling-and-other-api-updates) (or manual equivalents, e.g. `Respond only with valid JSON`-like prompts) by updating `type AIStreamMode = "raw" | "text" | "json"` and adding the OpenAI and Anthropic parsers.

packages/core/streams/parsers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+      // this can be used for either chat or completion models
+      const text = trimStartOfStream(
+        json.choices[0]?.delta?.content ?? json.choices[0]?.text ?? ''
+      )
+
+      return text
+    }
+  }
+}
```
ctjlewis, line 25: We can easily add JSON parsing on completions for integration with [OpenAI function calling](https://openai.com/blog/function-calling-and-other-api-updates) (or manual equivalents, e.g. `Respond only with valid JSON`-like prompts) by updating `type AIStreamMode = "raw" | "text" | "json"` and adding the OpenAI and Anthropic parsers.

packages/core/streams/openai-stream.ts
```diff
@@ -1,27 +1,7 @@
...
-  }
-}
-
-export function OpenAIStream(
-  res: Response,
-  cb?: AIStreamCallbacks
-): ReadableStream {
-  return AIStream(res, parseOpenAIStream(), cb)
+export function OpenAIStream(res: Response, options: AIStreamOptions = {}) {
+  return AIStream(res, { parsers: openaiParsers, ...options })
```
ctjlewis, line 6: The `parsers` object contains everything `AIStream()` needs to parse chunks for each of the given `AIStreamMode`s.

packages/core/streams/parsers.ts
Zakinator123: Check out the long block comment I've got in my PR (#154 ). The way OpenAI streams function calling responses is a little weird.

ctjlewis: Very weird, I haven't gotten hands-on with it yet. I will look at this more in depth tomorrow, TY for note. 

Sounds like it's probably better to merge yours first, and then resolve conflicts here before this one ships, if anything.

jaredpalmer (commented):
2 APIs for this instead of this `mode` stuff: `OpenAITextStream` and `OpenAIDataStream`

ctjlewis:
> 2 APIs for this instead of this `mode` stuff: `OpenAITextStream` and `OpenAIDataStream`

Makes sense, agree that's probably easier. Will update.

ctjlewis:
Oh right, the issue is that if you do it that way, then you write two separate functions for OpenAI's AIStream implementation, then two for Anthropic's, plus any future ones you add, and you can't enforce their shape with TS or anything - plus you manually write and bind the parsers manually, for each implementation.

Here you just set the mode and `AIStream` loads the given chunk parser. If you add a completions JSON parser you only have to update each place once, same for adding a new implementation on top of AIStream.

ctjlewis:
@jaredpalmer I added completion JSON parsing to demonstrate, and converted this to a draft for now. 

The mode is just a way to switch between chunk parsers. You can see how this pattern lets you do `OpenAIStream(res, { mode: "json" })` or wrap whatever you want with `AIStream(res, { mode, parsers: ... })`. This can easily be extended to better integration with [OpenAI Function Calling](https://openai.com/blog/function-calling-and-other-api-updates) or JSON-like prompt responses.

jaredpalmer:
I don't want people to have to pass a configuration object when 90% of folks will be using just chat. So lets expose 2 for each provider please.

ctjlewis:
I will make changes but you don't have to pass the config object. It defaults to text parsing and is just there for overrides.

jaredpalmer:
Ahh my b. So let's simplify then for the use cases i've described. Let make it silly easy for text. Easy for functions, escape hatch for the everyone else.

ctjlewis:
For sure. Right now the pattern is still:

```ts
const { messages } = await req.json()
  model: 'gpt-4',
  stream: true,
  messages
})
const stream = OpenAIStream(response)
return new StreamingTextResponse(stream)
```

We could simplify it to:

```ts
const { messages } = await req.json()
return await OpenAIStream("createChatCompletion",  {
  model: 'gpt-4',
  stream: true,
  messages
})
```

By handling OpenAI inside the library and wrapping with optimized `StreamingTextResponse`.

ctjlewis:
The simplest it could possibly be would be to write a `withOpenAIChat()` hook, so you could do something like:

```ts
export const runtime = "edge"
export default withOpenAIChat({
  model: 'gpt-4',
})
```

We set `stream: true` on our end every time, and handle `messages = await req.json()`. It would accept a callback for you to plug in custom logic if needed.

jaredpalmer:
It's very important we do not wrap the SDK in any way, so lets keep original pattern

PR #145, opened 2023-06-18
Feature/add rfc4122 uuids

k14uz:
The modifications in this pull request are as follows:

- ~~**Integration of UUID Package**: The UUID package has been integrated to facilitate the generation of unique identifiers~~. (`Nanoid` module is being used from existing packages)
- **Assignments in Core Modules**: The UUID functionality has been assigned in the following modules:
    - `packages/core/svelte/use-chat.ts`
    - `packages/core/vue/use-chat.ts`
    - `packages/core/vue/use-completion.ts`
    - `packages/core/svelte/use-completion.ts`

This modification ensures the `uniqueId` is in compliance with [RFC4122](https://www.ietf.org/rfc/rfc4122.txt).


jaredpalmer:
Can we use nanoid to produce the same result? We already use that package.

sirlolcat:
> Can we use nanoid to produce the same result? We already use that package.

I adjusted it accordingly, let me know if there is anything you'd like me to address.

packages/core/streams/ai-stream.ts
```diff
@@ -5,110 +5,142 @@ import {
...
-    })
+  const responseBodyStream = response.body || createEmptyReadableStream()
 
-  return stream
+  return responseBodyStream
     .pipeThrough(createEventStreamTransformer(customParser))
     .pipeThrough(createCallbacksTransformer(callbacks))
 }
+
+/**
```
jaredpalmer, line 136: Remove changes to ai-stream, seems like mixing PRs here.

jaredpalmer (changes requested):

packages/core/streams/ai-stream.ts
sirlolcat: My apologies, there was a mix-up in the local setup. I reverted it now.

packages/core/svelte/use-chat.ts
```diff
@@ -60,7 +60,7 @@ export function useChat({
   body
 }: UseChatOptions = {}): UseChatHelpers {
   // Generate a unique ID for the chat if not provided.
-  const chatId = id || `chat-${uniqueId++}`
+  const chatId = id || `chat-${uniqueId}`
```
jaredpalmer, line 63: this isn't calling the function, it's just reusing the same one.

jaredpalmer (changes requested):

packages/core/svelte/use-chat.ts
sirlolcat: Great catch! I'll fix it now.

MaxLeiter (changes requested):
Can you modify the react hooks to also use `nanoid`? [`useId`](https://react.dev/reference/react/useId) should be used for accessibility and not general purpose IDs.

jaredpalmer:
Let's update and merge.

MaxLeiter (approved):

PR #153, opened 2023-06-18
Add onStart callback to use-chat

lautapercuspain:
Hey folks, I've added this callback, to run some logic right before starting the chat logic, do you think it makes sense?

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -53,6 +53,7 @@ export function useChat({
   initialInput = '',
   sendExtraMessageFields,
   onResponse,
+  onStart,
   onFinish,
   onError,
   headers,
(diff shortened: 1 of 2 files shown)


Loris126:
In my opinion, yes

packages/core/react/use-chat.ts
```diff
@@ -62,6 +63,11 @@ export function useChat({
   const hookId = useId()
   const chatId = id || hookId
 
+  //Useful to execute logic when the chat is mounted.
+  if (onStart) {
```
jaredpalmer, line 67: This doesn't run on mount but during and on every render.

jaredpalmer: At the moment, this is implemented as....

```
onStart()
const { ... } = useChat()
```

If you need to run something on mount, then just add an effect at the call site?

jaredpalmer (changes requested):

packages/core/react/use-chat.ts
lautapercuspain: @jaredpalmer Yes, what you said makes sense. 
I was actually trying to find a way to avoid running  `const { data, mutate } = useSWR<Message[]>([api, chatId], null, {`. 
if certain situation happens so we can avoid running it.

Can you think in a good use case for that?

lautapercuspain: Didn't want to said "on mount" it's more on hook initialization.

lautapercuspain:
Hey @aredpalmer  Thanks for taking a look at the above. I closed this PR for now.

PR #165, opened 2023-06-19
Fix: Issue #94 CallbackManager is deprecated

### LINKED ISSUE #94: CallbackManager in langchain is deprecated
In the langchain example it's using callBackManager to get the handlers, this is now deprecated and should be replaced with handlers

i.e. the following line:
```
callbackManager: CallbackManager.fromHandlers(handlers)
```
should be replaced with:
```
 callbacks: [handlers],
```

mohamedcherifmo:
Fix #94 . Do not use `callbackManager` in LangChain examples, but `callbacks` key.

### DIFF
--- a/.changeset/blue-rice-remember.md
+++ b/.changeset/blue-rice-remember.md
@@ -1,5 +1,5 @@
 ---
-"ai": patch
+'ai': patch
 ---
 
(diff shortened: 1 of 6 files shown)


jaredpalmer:
Can we update documentation as well?

jaredpalmer (approved):

jaredpalmer:
Thanks for the help!

PR #178, opened 2023-06-20
Support for function calling (updated react hook + stream utilities)

Zakinator123:
This PR enables users to easily replicate the experience of ChatGPT plugins using the Vercel AI SDK and OpenAI's new function calling capabilities.

Note that this PR adds support for executing function calls on the **client side**. Vercel is working on [supporting server-side function calls](https://twitter.com/jaredpalmer/status/1673350963191508993).

An example of the changes in this PR in action can be viewed [in this modified clone of the Vercel AI Chatbot](https://nextjs-chat-three-dun.vercel.app/) ([Code can be viewed here](https://github.com/Zakinator123/nextjs-chat)). In the example, you can ask the chatbot what the weather is (e.g. What is the weather in LA?"), and it will execute a function call on the client side to get some dummy data, which is then provided back to the OpenAI model. **Both the streamed function call and function response messages show up in the chat interface (unless the front-end code filters it out).**

At a high level, this PR modifies the useChat hook so that it also takes a `FunctionCallHandler` that accepts function calls and optionally returns function responses (which are automatically sent back to the model). Furthermore, users of the SDK can submit functions on a message by message basis when adding new chat messages. Streaming utils have also been updated to support streaming function call responses back from the Edge runtime (originally I had this change in #154).

A simple example showing how clients would use the SDK can be found in the `examples/next-openai` directory.


packages/core/react/use-chat.ts
```diff
@@ -41,229 +46,323 @@ export type UseChatHelpers = {
...
-  onResponse,
-  onFinish,
-  onError,
-  headers,
-  body
-}: UseChatOptions = {}): UseChatHelpers {
-  // Generate an unique id for the chat if not provided.
-  const hookId = useId()
-  const chatId = id || hookId
+                          api = '/api/chat',
```
lautapercuspain, line 63: Seems like indentation is different here,  and this change added `;`.
There's an script to fix this I guess: `prettier-fix`.

packages/core/streams/openai-stream.ts
```diff
@@ -10,6 +10,101 @@ function parseOpenAIStream(): (data: string) => string | void {
...
+        .replace(/"/g, '\\"')   // Escape double quotes
+        .replace(/\n/g, '\\n')  // Escape new lines
+        .replace(/\r/g, '\\r')  // Escape carriage returns
+        .replace(/\t/g, '\\t')  // Escape tabs
+        .replace(/\f/g, '\\f'); // Escape form feeds
+
+      return `${escapedPartialJson}`
+    } else if (json.choices[0]?.finish_reason === 'function_call') {
+      return '"}}'
+    }
```
lautapercuspain, line 106: Should we move this logic to another function?  🤔

packages/core/react/use-chat.ts
Zakinator123: Great point, I'll adjust my linting settings and push up a fix later today.

Zakinator123:
> It would be nice to parse the response from OpenAI differently, when it comes to function calling. Since it won't be a stream response. 🤔

Function calling responses are streamable too. Furthermore, especially if the `function_call` parameter is set to `auto` in the request, you don't know whether or not the model will decide to call a function, so the streaming utils must be updated to support streaming the function calls to the clients as well.

Danm72:
I was toying with this separately and the responses not coming back as streams really messed with my head. I think it'd be better to keep the endpoint as a streaming endpoint completely so the behaviour is more predictable. Especially for people not using useChat.

packages/core/streams/openai-stream.ts
Danm72: Would second this, maybe with a test too.

Itsnotaka:
Would there be better support for function call on langchain constructor as well?

Zakinator123:
> I was toying with this separately and the responses not coming back as streams really messed with my head. I think it'd be better to keep the endpoint as a streaming endpoint completely so the behaviour is more predictable. Especially for people not using useChat.

This PR enables that! It allows for streaming back both function call responses as well as normal assistant responses.

lautapercuspain:
> > It would be nice to parse the response from OpenAI differently, when it comes to function calling. Since it won't be a stream response. 🤔
> 
> Function calling responses are streamable too. Furthermore, especially if the `function_call` parameter is set to `auto` in the request, you don't know whether or not the model will decide to call a function, so the streaming utils must be updated to support streaming the function calls to the clients as well.

I mean, don't get my wrong, it's nice that we could stream function call responses, but I guess if we do that, we could break the SRP (Single-responsibility principle).

Consider this, for example, the Open API payload endpoint, accepts an `stream`  parameter which is boolean. 
If this parameter is false, we may don't want to parse the response as `stream`, We can stop that logic at that moment, and parse the response differently. We can also add support to useChat or any other interface for the user to choose when they would like streamed responses or not. In my head, this new function calling functionality is 3 additional steps in the middle of the flow. That is:

- Seeing if finish_reason is 'function_call' ( 99,99% yes, depending on your software design)
- Call your function
- Call again the API with the function results with role='function'

Then, resume the streaming flow. 
But, I may be wrong in my point of view, anyways, I'm trying to reason about it, together.

stephenkalnoske-sans:
Just want to make sure that the function stream will text will be available to the onCompletion handler in the OpenAIStream callback? That's where I'm hoping to parse it and call my actual backend logic.

arbaaz:
Excited!

Zakinator123:
Just FYI for everyone, I'm working on cleaning up the PR, hopefully will have some updates pushed up by this evening.

Zakinator123:
One issue I'm wrestling with is that function calling is very OpenAI specific, which has me leaning towards creating a brand new hook (e.g. `useOpenAIChat`), especially since `useChat` _technically_ is model-agnostic. But, it seems that `useChat`  implicitly has a preference for OpenAI (for example, the `Message` type has a role that is identical to OpenAI's roles of system, assistant, and user. If I end up creating a separate hook, it would invariably be a near-duplicate of `useChat`

MaxLeiter:
@Zakinator123 I'm a fan of a new `mode: 'functions'` param for `useChat` instead of a mostly-duplicate hook. 

thank you for your work on this!

packages/core/react/use-chat.ts
```diff
@@ -106,110 +125,167 @@ export function useChat({
...
+          const res = await fetch(api, {
+            method: 'POST',
+            body: JSON.stringify({
+              messages: sendExtraMessageFields
+                ? chatRequest.messages
+                : chatRequest.messages.map(
+                    ({ role, content, name, function_call }) => ({
+                      role,
+                      content,
+                      ...(name !== undefined && { name }),
```
Zakinator123, line 152: Check if this undefined checking logic is even necessary

packages/core/shared/types.ts
```diff
@@ -1,20 +1,37 @@
...
 /**
  * Shared types between the API and UI packages.
  */
 export type Message = {
   id: string
   createdAt?: Date
   content: string
-  role: 'system' | 'user' | 'assistant'
+  role: 'system' | 'user' | 'assistant' | 'function'
+  name?: string
```
Zakinator123, line 15: Write comment (or jsdoc?) explaining what name is here.

Zakinator123:
> @Zakinator123 I'm a fan of a new `mode: 'functions'` param for `useChat` instead of a mostly-duplicate hook.
> 
> thank you for your work on this!

@MaxLeiter I've made `onFunctionCall` and all the new function calling stuff optional, so no need for a mode!

Zakinator123:
> What do you think about calling it onFunctionCall instead to be aligned with other APIs?

Good feedback! I've updated the PR with the change.

Zakinator123:
I think the PR is starting to come together! There are still a few touchups and refactors I'd like to make - should have them done tomorrow. Also, I've been having a ton of fun experimenting with function calling using this updated hook 😁 I'll add some example code in the `examples/` directory tomorrow.

Zakinator123:
I've got some working example code for the PR in the [examples directory](https://github.com/vercel-labs/ai/pull/178/files#diff-598122fec58328a19a8e99db9e688b69a1a309b41a0605c48ef9a55dd8104d9e).

With the `eval_code_in_browser` function, I was able to chat with GPT-3.5 and get it to change the styling of the page to dark mode and add some crazy CSS animations to the page!! 

Unfortunately, the `eval_code_in_browser` is pretty finicky and doesn't always work - there's still some character escaping I haven't figured out. I tried to chat with it to write and eval code to insert a platformer game into the browser, but it wouldn't work. However, the other example functions seem to work very reliably.

The PR is pretty much ready to be reviewed, though I need to do a little more manual testing and I may still touch up things here and there.

Zakinator123:
image

image

image

Archimagus:
Is there a plan to add this to the svelte API as well?

lautapercuspain:
> > It would be nice to parse the response from OpenAI differently, when it comes to function calling. Since it won't be a stream response. 🤔
> 
> Function calling responses are streamable too. Furthermore, especially if the `function_call` parameter is set to `auto` in the request, you don't know whether or not the model will decide to call a function, so the streaming utils must be updated to support streaming the function calls to the clients as well.

But when it comes to function arguments, my main concern is that potentially, streaming that response could break the JSON.  I'm not 100% sure about this but I can tell that tokens may be split. 

Is there a risk of the returned JSON arguments being split, or will they always be returned in a single chunk?

packages/core/streams/openai-stream.ts
lautapercuspain: @Danm72 I'm not sure if those replacement statements should be here. Additionally, it seems to be the client's responsibility, since the argument chucks may be different depending on the type of function we pass to Open AI. Thoughts?

Archimagus:
> Is there a plan to add this to the svelte API as well?

I created a PR to @Zakinator123's branch that adds a svelte version, coppies the same logic that is currently in the react version.

Markeljan:
Great work!  I have been using this for my latest demos.

@Zakinator123  your ai-chat repo is a godsend as well i was unable to use this pull request as a source with yarn but forking your repo worked perfect.
https://github.com/Zakinator123/nextjs-chat/blob/main/LICENSE

Zakinator123:
> Great work!  I have been using this for my latest demos.
> 
> @Zakinator123  your ai-chat repo is a godsend as well i was unable to use this pull request as a source with yarn but forking your repo worked perfect.
> https://github.com/Zakinator123/nextjs-chat/blob/main/LICENSE

Glad it helped! I haven't updated the fork quite yet with the latest changes in this PR, but should be able to get to that today or tomorrow.

steven-tey:
@Zakinator123 Thank you for your work on this – it's incredible!

I'm testing this out now and had a question – is there a way we can feed the response from the function call back into OpenAI and create a chat completion? 

Inspo: the `second_response` part on line 70 from this code example in the OpenAI docs: https://platform.openai.com/docs/guides/gpt/function-calling

Archimagus:
> I'm testing this out now and had a question – is there a way we can feed the response from the function call back into OpenAI and create another chat completion (so that the streamed response is in natural language)?

Just return a chat completion with the results of the function call.  Unless I am misunderstanding

const functionResponse: ChatRequest = {
				messages: [
					...chatMessages,
					{
						id: nanoid(),
						name: functionCall.name,
						role: "function" as const,
						content: JSON.stringify(results),
					},
				],
				function_call: functionCallHandler,
				functions,
			};
			return Promise.resolve(functionResponse);

steven-tey:
@Archimagus yup, makes sense! 

Also, another big thing here is the ability to stream function calls & the intermediary actions/JSON responses to the client while it's happening - similar to ChatGPT's native browsing plugin UX. Will this support that?

Archimagus:
I don't see why not, you get all the data, both client and server side, so you should be able to show whatever.

steven-tey:
Launched a little PoC today that uses function calling: https://twitter.com/steventey/status/1673041790272954369

Two things that can be improved (which I think is relevant to this PR):
1. If the prompt doesn't require function calling, it'll result in a blocking response (it's only streamed in the UI because I'm [mocking a stream](https://github.com/steven-tey/chathn/blob/main/app/api/chat/route.ts#L50-L68)). 
2. This doesn't account for multiple consecutive function calls (e.g. get the top 5 stories, then get the comments for those stories and summarize them)

If you know any way to fix these two things, feel free to PR! :)

Archimagus:
Has anyone had the experience where sometimes the bot will respond with the function call as part of the chat, rather than as the function call? I'm wondering if it's my svelte implementation of the use_chat hook, or just the AI responding wrong?

```
Visitor

I am Iron Man.
```

```
AI

Well, hello there, Mr. Stark. It's an honor to assist a genius like you. Considering our roles, 
shall I play Jarvis for you to better suit your needs? By the way, let's also update our UI settings to reflect this change. 
One moment, please.{"function_call": {"name": "change_username", "arguments": "{"name": "Iron Man", "role": "user"}"}}```

**Comment by MaxLeiter** [comment]
Hey @Zakinator123, thanks so much for your work on this!

I gave it a look and I think we can ship this as a pre-release of functions support while we finalize the API. You can see some of our explorations here: https://twitter.com/jaredpalmer/status/1673350963191508993.

Would you mind adding a test or two for your OpenAIStream changes? If not no worries, I can handle it. Looking forward to getting this shipped!

**Comment by Zakinator123** [comment]
@MaxLeiter Sure thing! Thanks for reviewing - I'll add that test and get @Archimagus' svelte PR merged in to my fork asap.

**Comment by MaxLeiter** [comment]
For sake of review, I think we should leave @archimagus's PR separate and we can have it be a fast follow-up

**Comment by Archimagus** [comment]
Either  way works for me.  If you don't get it merged to @Zakinator123's first, I can rebase my changes on the vercel repo.  I have figured out my function calling problem from above.  It has to to with `streamedResponse.startsWith('{"function_call":')`  in `getStreamedResponse`  I am working on a regex to find the function call anywhere in the stream,  I'll submit that either with my other PR, or as a new one if it gets merged before I'm done.

**Comment by Zakinator123** [comment]
> > > It would be nice to parse the response from OpenAI differently, when it comes to function calling. Since it won't be a stream response. 🤔
> > 
> > 
> > Function calling responses are streamable too. Furthermore, especially if the `function_call` parameter is set to `auto` in the request, you don't know whether or not the model will decide to call a function, so the streaming utils must be updated to support streaming the function calls to the clients as well.
> 
> But when it comes to function arguments, my main concern is that potentially, streaming that response could break the JSON. I'm not 100% sure about this but I can tell that tokens may be split.
> 
> Is there a risk of the returned JSON arguments being split, or will they always be returned in a single chunk?

Good question @lautapercuspain ,
If you look at my [comment explaining how function calling chunks are streamed](https://github.com/vercel-labs/ai/pull/178/files#diff-b4fc3cf85f3dc974899314160d4930248499e7e5c371f123d4d9a4948c6b3e62R14) or my [test data showing the chunks](https://github.com/vercel-labs/ai/pull/178/files#diff-c386744b666862eb2649d1936a536bc75686a61b0d5a4d9c243a303fd765907fR64), you can see that the function call name is not at risk of being malformed, but the function call "arguments" can be malformed JSON (or improperly escaped), so my PR leaves the arguments as a string, and leaves it to the client to handle the case where OpenAI returns malformed JSON in the arguments.

**Comment by Zakinator123** [comment]
@MaxLeiter I've added some tests and made a few other small changes!

**Comment by JohannLai** [comment]
@Zakinator123 have a question about this PR. Assuming I have some functions being called in `route.ts` and some being called in `pages.tsx`, how should I handle this?

**Comment by Zakinator123** [comment]
> @Zakinator123 have a question about this PR. Assuming I have some functions being called in `route.ts` and some being called in `pages.tsx`, how should I handle this?

I'm understanding this question as "If I want function calls to be executed on both the client side and server side, how should I handle this?" With this PR I don't think it's possible to execute function calls on the server side with streaming, but I think the folks at vercel are working on that problem (see the tweet Max linked earlier), and I can envision some easy modifications on top of this PR enabling server side function execution w/ streaming. You might be able to do it without using streaming, but I think the edge runtime/serverless time limits may be a limiting factor.

File: packages/core/shared/types.ts [hunk 5]
```diff
@@ -1,25 +1,57 @@
...
   id: string
   createdAt?: Date
   content: string
-  role: 'system' | 'user' | 'assistant'
+  role: 'system' | 'user' | 'assistant' | 'function'
+  /**
+   * If the message has a role of `function`, the `name` field is the name of the function.
+   * Otherwise, the name field should not be set.
+   */
+  name?: string
```
MaxLeiter, line 19: nit: thoughts on `function_name` for consistency with `function_call`?

Zakinator123:
> Has anyone had the experience where sometimes the bot will respond with the function call as part of the chat, rather than as the function call? I'm wondering if it's my svelte implementation of the use_chat hook, or just the AI responding wrong?
> 
> ```
> Visitor
> 
> I am Iron Man.
> ```
> 
> ```
> AI
> 
> Well, hello there, Mr. Stark. It's an honor to assist a genius like you. Considering our roles, 
> shall I play Jarvis for you to better suit your needs? By the way, let's also update our UI settings to reflect this change. 
> One moment, please.{"function_call": {"name": "change_username", "arguments": "{"name": "Iron Man", "role": "user"}"}}
> ```

> Either way works for me. If you don't get it merged to @Zakinator123's first, I can rebase my changes on the vercel repo. I have figured out my function calling problem from above. It has to to with `streamedResponse.startsWith('{"function_call":')` in `getStreamedResponse` I am working on a regex to find the function call anywhere in the stream, I'll submit that either with my other PR, or as a new one if it gets merged before I'm done.

@Archimagus that seems like a bug for OpenAI to return both assistant text and a function call in the same response.. I'm not sure that's the desired behavior and I haven't seen any examples of the model returning an output like that in OpenAI's documentation.. though the example you gave does seem compelling and more natural, and if it's something that occasionally happens, then I guess this SDK should support it too? Can be added in a follow up PR though.

Archimagus:
> @Zakinator123  if it's something that occasionally happens, then I guess this SDK should support it too? Can be added in a follow up PR though.

I see that regularly when using GPT-4,  about half of the returns on certain prompts.  I've solved it, on my local end, in the svelte version.  But you are right, it can be done as a followup PR.  I'd like to see more of the code from use_chat abstracted out, to get all the parsing logic in one place, and have the react/svelte/vue versions not have to each deal with it, as it's already pretty difficult to keep them in sync.

I could probably do that unless someone from the core team gets to it.

MaxLeiter (approved):

packages/core/streams/openai-stream.ts
Zakinator123: @lautapercuspain  I agree that it's a little weird to have these replace statements, and there may be a better alternative. The issue is that without the character escaping, parsing the function call object with JSON.parse does not work.. I'll try to think of a better way to do this, but would appreciate suggestions as well.

jaredpalmer (changes requested):
Lets change to `experimental_onFunctionCall`

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+  // Generate a map of message role to text color
+  const roleToColorMap: Record<Message['role'], string> = {
+    system: 'red',
+    user: 'black',
+    function: 'blue',
+    assistant: 'green'
+  }
+
+  const getRenderedMessage = (m: Message) => {
+    if (m.content === '') {
```
jaredpalmer, line 146: @MaxLeiter fast follow with abstractions

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+    <div className="flex flex-col w-full max-w-md py-24 mx-auto stretch">
+      {messages.length > 0
+        ? messages.map((m: Message) => (
+            <div
+              key={m.id}
+              className="whitespace-pre-wrap"
+              style={{ color: roleToColorMap[m.role] }}
+            >
+              <strong>{`${
+                m.role.charAt(0).toUpperCase() + m.role.slice(1)
```
jaredpalmer, line 185: Can we just simplify this? It doesn't add to example and is intimidating for people to read

jaredpalmer (changes requested):

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -131,7 +131,7 @@ export default function Chat() {
 
   const { messages, input, handleInputChange, handleSubmit } = useChat({
     api: '/api/chat-with-functions',
-    onFunctionCall: functionCallHandler
+    experimentalOnFunctionCall: functionCallHandler
```
jaredpalmer, line 134: specifically add `experimental_onFunctionCall` (with the underscore). This is how we do it at Vercel.

jaredpalmer (changes requested):

jaredpalmer:
@MaxLeiter this needs a changeset

examples/next-openai/app/function-calling/page.tsx
Zakinator123: Sure, I'll clean up the example a bit.

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+            role: 'function' as const,
+            content: JSON.stringify({
+              temperature,
+              weather,
+              info: 'This data is randomly generated and came from a fake weather API!'
+            })
+          }
+        ]
+      }
+      return Promise.resolve(functionResponse)
```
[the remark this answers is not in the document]
Zakinator123: Good point, I'll refactor.

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -0,0 +1,205 @@
...
+          description:
+            'The temperature unit to use. Infer this from the users location.'
+        }
+      },
+      required: ['location', 'format']
+    }
+  },
+  {
+    name: 'get_current_time',
+    description: 'Get the current time',
```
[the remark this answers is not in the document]
Zakinator123: Good point. The functions that are passed to OpenAI in a request can also be defined and kept server side (I can update the example to do this), but this PR enables function calling to be handled _client side_ with `onFunctionCall` (which is different than the flows suggested in your [tweet](https://twitter.com/jaredpalmer/status/1673350963191508993), and a use case I'm most excited about). As such, the client has to have some knowledge about the various potential function calls and how to handle them in the `FunctionCallHandler`.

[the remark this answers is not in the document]
Zakinator123: In case anyone needs to be sold on client-side handling of function calls, here some examples I'm excited about:

- Evaluating LLM-generated javascript code in the browser with `eval` to create d3.js charts, create/iterate on browser-based games, changing the styling or look and fell of a user's page, etc.. Also, evaluating code that users generate through their prompting in the browser rather than on the server mitigates security risks of evaluating user-generated code server-side where more sensitive resources could be abused.

- Conducting complicated data queries and data analysis of tabular data in the browser with [sqlite/WASM](https://developer.chrome.com/blog/sqlite-wasm-in-the-browser-backed-by-the-origin-private-file-system/) by using LLMs to generate SQL that is executed on the sqlite instance running in the browser.
-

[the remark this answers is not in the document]
Archimagus: I am also using client side function calling, for things like updating the user's name in a chat application, redirecting to other pages, and things like that.  That said, doing server side function calling isn't as easy, as you have to jsut watch the onCompletion and parse any function calls yourself, so we might want another PR to address that.

packages/core/streams/openai-stream.ts
lautapercuspain: @Zakinator123. I'm currently debugging your PR 👍🏻.   
So far the `get_current_weather` and `get_current_time` works in a reliable and predictable way.  
The issue with JSON.parse raises when it comes to code execution, like your example with `eval` which would be really cool to support out of the box on the other hand. Maybe the simplest solution here is to instruct the agent (as you are doing in your prompt to escape special characters, when it comes to code functionality.
I'll comment any findings on my side.

examples/next-openai/app/function-calling/page.tsx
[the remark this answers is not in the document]
MaxLeiter: an `onFunctionCall` server-side callback is definitely planned, but we can work on the next steps after this work is merged.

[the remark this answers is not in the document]
Zakinator123: Yes, I do think server-side function calling as outlined in Jared's tweet is important to support, and I think it can be fully complementary with client-side function calling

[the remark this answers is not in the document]
Zakinator123: @jaredpalmer  I've moved the function definitions in the `examples/next-openai` example to be server-side.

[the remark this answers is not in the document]
onionhammer: I think the server side execution of functions is probably just as important as the client side, but @Zakinator123 brings up some very compelling usecases for public executions.

MaxLeiter:
Thanks for all of your work here @Zakinator123 et al!

mishkinf:
@Archimagus Do you have a PR for the svelte version of this? Is that getting merged soon as well?

Archimagus:
> @Archimagus Do you have a PR for the svelte version of this? Is that getting merged soon as well?
Yes, it was based off of @Zakinator123  branch. So I need to retarget it to the main branch here and then make sure it's up to date. But other than that it shouldn't take too long to make it match. What's there now.

Archimagus:
Here is the new PR.

https://github.com/vercel-labs/ai/pull/258

salvinoto:
Was the server sided function calling ever implemented? Im currently trying to do it in the client like in the examples directory, and I get no output, just looping data: function over and over infinitely

PR #184, opened 2023-06-21
Move types to its corresponding file

lautapercuspain:
To make those beautiful hooks even more readable I went ahead and moved those big types definitions on it's own shared file. I re-created the same types adding it's corresponding framework/lib name to make easy to identify.

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -3,52 +3,13 @@ import useSWRMutation from 'swr/mutation'
 import useSWR from 'swr'
 import { nanoid, createChunkDecoder } from '../shared/utils'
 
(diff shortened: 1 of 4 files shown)


jaredpalmer:
Thanks for this. I'd like to keep types as close to their implementation area. Furthermore, this complicates documentation with the framework naming. Going to close for now.

lautapercuspain:
> Thanks for this. I'd like to keep types as close to their implementation area. Furthermore, this complicates documentation with the framework naming. Going to close for now.

@jaredpalmer No problem and thanks for your feedback! It's intersting to see different point of view and of course all are valid. I was thinking that this could improve file readability. 
Regarding your arguments, I think types are  one click away and you could see the types definition in the same file with just a mouse hover. But again, I understand that there're different ways to organize types.

PR #193, opened 2023-06-21
Support extra request options in chat and completion

### LINKED ISSUE #131: Allow changing `body` with `useCompletion`
I've got an endpoint that takes custom additional data - it would make using `useCompletion` a lot easier in many situations if it exposed `setBody` or something similar.

### LINKED ISSUE #141: Could more context be passed to stream callbacks?
In creating useful stream callbacks, it would be helpful to carry some additional context about the initial request (model, hyperparameters, prompt) to be used in logging (etc) in the callback. 

I'd be open to creating a PR here if y'all at Vercel are interested.

### LINKED ISSUE #189: Add `setHeaders` and `setBody` to `UseChatHelpers`
Seems related to #131 

If there was an additional `setMetadata` function or two `setHeaders` and `setBody` added to the returned chat and completion helpers object we could set headers for auth and body as needed synchronously before any call. I have a use case where I need to set a new auth token for every few calls as the token expires.

Like...

```
const setMetadata = ({ headers, body }) => {
    if(headers) {
      extraMetadataRef.current = {
        ...extraMetadataRef.current,
        headers
      }
    }
    if(body) {
      extraMetadataRef.current = {
        ...extraMetadataRef.current,
        body
      }
    }
```

I think I see where to add it but I am not sure how to develop in a published NPM library... Will try to study some to wrap my head around how to maybe contribute this in a PR. Any tips are welcome. 

EDIT: Am I correct to simply clone the repo, `pnpm i` then `pnpm run dev` with a symlink to my `node_modules` folder? Sorry if that is an extremely novice question. Never contributed to NPM lib before.

bioshazard:
Attempts to solve #189 by adding a way to update the extraMetadataRef headers and body in useChat. The same might apply to useCompletion, but that is outside my use case and I did not want to overcommit my assumptions of what was acceptable before getting feedback on this contribution. Looking forward to thoughts and feedback, thanks for the support!

Closes #141, closes #131, closes #189.

### DIFF
--- a/.changeset/tender-knives-sing.md
+++ b/.changeset/tender-knives-sing.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Support extra request options in chat and completion hooks
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -6,6 +6,11 @@ import { nanoid, createChunkDecoder } from '../shared/utils'
 import type { Message, CreateMessage, UseChatOptions } from '../shared/types'
 export type { Message, CreateMessage, UseChatOptions }
 
+type RequestOptions = {
+  headers?: Record<string, string> | Headers
+  body?: object
+}
+
 export type UseChatHelpers = {
   /** Current messages in the chat */
   messages: Message[]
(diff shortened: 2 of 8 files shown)


shuding:
Thanks for your contribution!

PR #207, opened 2023-06-22
Update route.ts on langchain example

dccabs:
Needed the default export to work


MaxLeiter (approved):
I accidentally approved this, sorry!

With the Next.js App Router,  you have named exports instead of a default one.

PR #284, opened 2023-07-06
Server and Client tools

BLamy:
Update the demo so that some tools run on the server while others run on the client. 

server-client-tools

I also think this is a cool demo because the browser permissions sandbox the AI.
image

### DIFF
--- a/examples/next-openai/ai/get_client_location.ts
+++ b/examples/next-openai/ai/get_client_location.ts
@@ -1,0 +1,61 @@
+import { Message, nanoid } from "ai";
+
+export default async function getClientLocation(
+  chatMessages: Message[]
+): Promise<{ messages: Message[] }> {
+  return new Promise(resolve => {
+    if (navigator.geolocation) {
+      navigator.geolocation.getCurrentPosition(
+        function (position) {
+          const location = {
+            latitude: position.coords.latitude,
+            longitude: position.coords.longitude,
+          };
+          console.log(location);
+          resolve({
+            messages: [
+              ...chatMessages,
+              {
+                id: nanoid(),
+                name: "get_client_location",
+                role: "function" as const,
+                content: JSON.stringify(location),
+              },
+            ],
+          });
+        },
+        function (error) {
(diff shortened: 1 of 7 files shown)


examples/next-openai/ai/get_client_location.ts
```diff
@@ -0,0 +1,48 @@
+"use client";
```
MaxLeiter, line 1: `"use client"` only applies in client components.

examples/next-openai/ai/get_current_weather.ts
```diff
@@ -0,0 +1,51 @@
...
+});
+
+export default async function getCurrentWeather(
+  chatMessages: Message[],
+  functionCall: ChatCompletionRequestMessageFunctionCall
+): Promise<ChatRequest> {
+  if (typeof window === "undefined") {
+    console.log("This code is running on the server side");
+  } else {
+    console.log("This code is running on the client side");
```
MaxLeiter, line 18: 🤔 if this has `use server` should it ever run on the client?

BLamy: This is just for demo purposes so show where the logs show up. You are right the client log should never trigger.

examples/next-openai/ai/get_client_location.ts
BLamy: This is just being explicit. The "use client" can be omitted in this instance and it would work the same.

MaxLeiter: Can you remove it? It doesn't do anything here.

examples/next-openai/ai/get_current_weather.ts
MaxLeiter: Could you make that explicit?

BLamy: Done, however, I can remove it altogether if you don't think it's useful.

lgrammel:
We are now using an updated tool calling method.

PR #299, opened 2023-07-07
align react hooks with other implementations, don't throw error when submitting

MaxLeiter:
It was reported by @steven-tey that the `isLoading` was not in-sync between different `useCompletion` calls with the same ID. 

It turns out that `useSWRMutation` ([docs](https://swr.vercel.app/docs/mutation#useswrmutation)) does not share state between hooks like `useSWR`. This PR switches to `useSWR` + manually making the mutation with `fetch`. This has the benefit of making the code consistent between all three versions. 

This also stops our react use-completion's internal `triggerRequest` from throwing an error to align with the other implementations.  

I'd review with [whitespace disabled](https://github.com/vercel-labs/ai/pull/299/files?w=1)


ricardogzm:
Hi. Has this been adressed?

shuding (approved):

Issue #192, opened 2023-06-21
The package "crypto" wasn't found on the file system but is built into node.


Archimagus:
After installing the "ai" package and "openai-edge" as recommended in the docs vercel can no longer build my applicatyion. I get the following error.

```html
<html>
<body>
<!--StartFragment-->

> Using @sveltejs/adapter-vercel
--
16:16:02.327 | ✘ [ERROR] Could not resolve "crypto"
16:16:02.328 |  
16:16:02.328 | .svelte-kit/output/server/entries/pages/_page.svelte.js:4:19:
16:16:02.328 | 4 │ import crypto from "crypto";
16:16:02.328 | ╵                    ~~~~~~~~
16:16:02.328 |  
16:16:02.328 | The package "crypto" wasn't found on the file system but is built into node. Are you trying to bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
16:16:02.328 |  
16:16:02.384 | error during build:
16:16:02.385 | Error: Build failed with 1 error:
16:16:02.385 | .svelte-kit/output/server/entries/pages/_page.svelte.js:4:19: ERROR: Could not resolve "crypto"
16:16:02.385 | at failureErrorWithLog (/vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:1636:15)
16:16:02.385 | at /vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:1048:25
16:16:02.385 | at /vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:993:52
16:16:02.385 | at buildResponseToResult (/vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:1046:7)
16:16:02.386 | at /vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:1075:16
16:16:02.386 | at responseCallbacks.<computed> (/vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:697:9)
16:16:02.386 | at handleIncomingPacket (/vercel/path0/node_modules/.pnpm/esbuild@0.17.19/node_modules/esbuild/lib/main.js:752:9)
[...]

PR #300, opened 2023-07-07
fix: Remove dependency on node crypto API

lachlancollins:
As seen in #192 , it is currently not possible to use edge functions in SvelteKit because the node crypto API is not available in the edge runtime. The node crypto API requirement is coming from [nanoid](https://github.com/ai/nanoid/blob/main/index.js). Nanoid has an alternative export, [nanoid/non-secure](https://github.com/ai/nanoid/blob/main/non-secure/index.js), which does not depend on the node crypto API.

From their readme r.e. `nanoid/non-secure`:
```
By default, Nano ID uses hardware random bytes generation for security and low collision probability. If you are not so concerned with security, you can use it for environments without hardware random generators.
```

It seems that `ai` is only using nanoid to generate unique IDs for chat messages - IMO this is a situation where non-secure can be used, which would be great for SvelteKit users.

### DIFF
--- a/.changeset/violet-monkeys-float.md
+++ b/.changeset/violet-monkeys-float.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Remove dependency on node crypto API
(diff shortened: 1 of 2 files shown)


MaxLeiter (approved):

MaxLeiter:
Thank you!

PR #305, opened 2023-07-09
Fix typo in sveltekit-openai example

carteakey:
Corrected small typo that causes npm install to fail.


MaxLeiter:
Dupe of #304, thank you though!

PR #311, opened 2023-07-10
Support for server-side OpenAI function calling

MaxLeiter:


packages/core/streams/openai-stream.ts
```diff
@@ -117,10 +119,102 @@ function parseOpenAIStream(): (data: string) => string | void {
...
+        isFunctionStreamingIn = true
+        aggregatedResponse += message
+        console.log('Function call detected')
+        isFirstChunk = false
+        return
+      }
+
+      // Stream as normal
+      if (!isFunctionStreamingIn) {
+        controller.enqueue(textEncoder.encode(message))
```
shuding, line 166:
```suggestion
        controller.enqueue(chunk)
```

To avoid doing the work again

packages/core/streams/openai-stream.ts
```diff
@@ -117,10 +119,102 @@ function parseOpenAIStream(): (data: string) => string | void {
...
+        return
+      } else {
+        aggregatedResponse += message
+      }
+
+      if (isEndOfFunction) {
```
shuding, line 172: You can remove the manual “end of function” detection and move this handling logic to the `flush` method of TransformStream: https://developer.mozilla.org/en-US/docs/Web/API/TransformStream/TransformStream

docs/pages/docs/guides/functions.mdx
```diff
@@ -6,17 +6,103 @@ import { Steps, Callout } from 'nextra-theme-docs'
...
+On the server, you can pass an `experimental_onFunctionCall` callback to the `OpenAIStream`, which will be called when the model calls a function.
+In order to support recursively calling functions and to keep the message context in the function calls, you can use `createFunctionCallMessages` to construct the relevant "assistant" and "function" messages for you.
+
+```ts {2,3,4,15}
+const stream = OpenAIStream(response, {
+  experimental_onFunctionCall: async (
+    { name, arguments: args },
+    createFunctionCallMessages
+  ) => {
+    // if you do nothing with the function call, it's passed to the client
```
shuding, line 76:
```suggestion
    // if you skip the function call and return nothing, the `function_call`
    // message will be sent to the client for it to handle
```

examples/next-openai/app/function-calling/page.tsx
```diff
@@ -9,68 +9,21 @@ export default function Chat() {
...
-
-      return functionResponse
-    } else if (functionCall.name === 'eval_code_in_browser') {
+    if (functionCall.name === 'eval_code_in_browser') {
       if (functionCall.arguments) {
         // Parsing here does not always work since it seems that some characters in generated code aren't escaped properly.
         const parsedFunctionCallArguments: { code: string } = JSON.parse(
           functionCall.arguments
         )
+        eval(parsedFunctionCallArguments.code)
```
shuding, line 18:
```suggestion
        // WARNING: Do NOT do this in real-world applications!
        eval(parsedFunctionCallArguments.code)
```

packages/core/streams/ai-stream.ts
```diff
@@ -13,6 +27,42 @@ export interface AIStreamCallbacks {
   onStart?: () => Promise<void>
   onCompletion?: (completion: string) => Promise<void>
   onToken?: (token: string) => Promise<void>
+
+  /**
+   * Only applicable for OpenAI
```
shuding, line 32: Can we move this type to the `OpenAIStream()` constructor instead? So it becomes `AIStreamCallbacks & { experimental_onFunctionCall?: ... }`

shuding (approved):

PR #334, opened 2023-07-15
Resolve function stream parsing error when provided a specific function  

### LINKED ISSUE #329: Providing a function to `function_call` in streaming chat completion throws because of `finish_reason: stop`
OpenAI chat completion  `finish_reason` does not return `function_call` but instead returns `stop` when defining the `function_call` parameter to force it to call a particular function (see repro steps below). 

Because of `finish_reason` in the last chunk being `stop` we don't run [this ](https://github.com/vercel-labs/ai/blob/main/packages/core/streams/openai-stream.ts#L157-L158) hence, the resulting string right after transform flush looks like so: 

`"{\"function_call\": {\"name\": \"log_age\", \"arguments\": \"{}"`

image

[This line throws](https://github.com/vercel-labs/ai/blob/main/packages/core/streams/openai-stream.ts#L235) because the JSON is incomplete (from not adding those last two closing curly braces) and that's what this issue is all about.

Now I would argue this is more on OpenAI not fully adhering to one contract of always marking `finish_reason: function_call` when provided functions with a function_call. 

However I have lower expectations on OpenAI getting this resolved faster than we can patch it here in the meantime.  I will try my best to make another issue on one of the OpenAI repos about it.

Let me know if you want me to take a stab at a fix else thanks for taking your time to look at this and putting together this helpful library!

Related: https://github.com/vercel-labs/ai/pull/311

Steps to reproduce

```
  model: "gpt-3.5-turbo-0613",
  stream: true,
  messages,
  functions: [
    {
      name: "log_age",
      description: "Call whenever user responds with their age.",
      parameters: {
        type: "object",
        properties: {
          age: {
            type: "number",
            description: "User's age",
          },
        },
      },
    },
  ],
  function_call: { name: "log_age" },
});

    { name, arguments: args },
    createFunctionCallMessages
  ) => {
    // Function definition in this case does not matter because the library fails before this point.
    // Only necessary to provide `experimental_onFunctionCall` so that we can invoke `createFunctionCallTransformer`

    return openai.createChatCompletion({
      // @ts-ignore
      stream: true,
      model: "gpt-3.5-turbo-0613",
      functions,
    });
  },
});

```

shametim:
Fixes #329 

The reasoning why this works is similar to why trimStartOfStreamHelper also works. parseOpenAIStream maintains a temporary local state of each complete streamed message from the start chunk to end chunk.

Does this need a changeset?

### DIFF
--- a/.changeset/empty-bees-switch.md
+++ b/.changeset/empty-bees-switch.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Providing a function to `function_call` request parameter of the OpenAI Chat Completions API no longer breaks OpenAI function stream parsing.
--- a/packages/core/streams/openai-stream.ts
+++ b/packages/core/streams/openai-stream.ts
@@ -115,7 +115,7 @@ function parseOpenAIStream(): (data: string) => string | void {
 
         ...
 
-        Finally, the last chunk has a `finish_reason` of `function_call`:
+        Finally, the last chunk has a `finish_reason` of either `function_call`:
 
           {
             ...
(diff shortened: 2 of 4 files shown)


MaxLeiter:
Thanks @shametim! I think this is a great solution. A changeset adding a patch would be great, but I can add it before this merges if you can’t get to it.

MaxLeiter (approved):

MaxLeiter:
Thanks!

shametim:
Sure! Thanks for reviewing

PR #342, opened 2023-07-17
docs: fix sveltekit `edge` runtime config

carljohan:
Updated Sveltekit runtime config according to docs: [Sveltekit docs page](https://kit.svelte.dev/docs/adapter-vercel#deployment-configuration)


MaxLeiter (approved):
Thanks!

PR #349, opened 2023-07-18
experimental_ChatFunctionHandler API

MaxLeiter:
- Made a new `ClientMessage` type that supports `function_call` being JSON (whereas server-handled functions halt until the function is completely streamed in)
	- Removes some type casting from the example
	- This is officially a breaking change but it _should_ just affect types
- Created new `experimental_ChatFunction` and `experimental_ChatFunctionHandler` classes. User's shouldn't need to interact with `experimental_ChatFunction` if they don't want to (they can pass the handler an array of objects)
- Basic argument validation
	- 	In the future we should investigate using zod 

[View the example route](https://github.com/vercel-labs/ai/blob/max/function-utils/examples/next-openai/app/api/chat-with-functions/route.ts) to see it in action

### DIFF
--- a/packages/core/functions/functions.ts
+++ b/packages/core/functions/functions.ts
@@ -1,0 +1,156 @@
+import type { ChatCompletionFunctions } from 'openai-edge'
+import type { OpenAIStreamCallbacks } from '../streams'
+import type { Message } from '../shared/types'
+
+type FunctionSchema = ChatCompletionFunctions
+
+type ExperimentalOnFunctionCall = NonNullable<
+  OpenAIStreamCallbacks['experimental_onFunctionCall']
+>
+
+type CreateFunctionCallMessages = Parameters<ExperimentalOnFunctionCall>[1]
+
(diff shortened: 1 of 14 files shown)


packages/core/functions/functions.ts
```diff
@@ -0,0 +1,176 @@
+import type { ChatCompletionFunctions } from 'openai-edge'
+import type { OpenAIStreamCallbacks } from '../streams'
+import type { JSONValue, Message } from '../shared/types'
+
+type FunctionSchema = ChatCompletionFunctions
+
```
jaredpalmer, line 6: Should this accept zod?

MaxLeiter: > In the future we should investigate using zod

If we want to commit to zod, sure.

Necmttn:
Any updates on this ?

MaxLeiter:
@Necmttn i’m not super happy with the API it provides. We’re working on getting it right

lgrammel:
Looks outdated. Please re-open if this is still relevant.

PR #352, opened 2023-07-19
add usage trend in README

cwtuan:
As an active user of the `ai` package, I have noticed an increasing interest within the developer community. To help track its growth and improve it further, I suggest adding the npm usage trend. This will provide valuable insights into the framework's popularity and adoption. I believe this addition will benefit both existing and potential users of the package.


leerob:
Thank you, but we're okay leaving this out 🙏

PR #356, opened 2023-07-20
Initial prototype supporting OpenAI Streamable

jridgewell:
OpenAI is considering removing the `response` property from their completions API, and instead will return an pre-parsed `AsyncIterable<ChatCompletionChunk>`. That saves us from having to parse the SSE ourselves, which is nice.

Still needs testing, but I think this should work.

Fixes [WEB-1303](https://linear.app/vercel/issue/WEB-1303/support-openai-streamable-interface)


packages/core/streams/openai-stream.ts
```diff
@@ -51,6 +53,59 @@ export type OpenAIStreamCallbacks = AIStreamCallbacks & {
...
+  function_call?: FunctionCall
+
+  /**
+   * The role of the author of this message.
+   */
+  role?: 'system' | 'user' | 'assistant' | 'function'
+}
+
+// https://github.com/openai/openai-node/blob/07b3504e1c40fd929f4aae1651b83afc19e3baf8/src/resources/chat/completions.ts#L146-L159
+interface FunctionCall {
```
MaxLeiter, line 94: nit: we use `ChatCompletionRequestMessageFunctionCall` from `openai-edge` elsewhere in the library. I'm happy keeping the types declared here but wondering if we want to unify them.

jridgewell: Ideally we'd delete all of these when we can import the `ChatCompletionChunk` type.

shuding (approved):

PR #357, opened 2023-07-20
streams/openai-stream: don't call onStart/onCompletion when recursing

### LINKED ISSUE #351: experimental_onFunctionCall on OpenAIStream calls onStart multiple times
I'm using experimental_onFunctionCall server-side to get some weather data and a couple of other functions. When the user sends a message for example: What's the weather in New York? The OpenAIStream handler calls onStart, and that's where I save my prompt messages to my database, when it detects the prompt is a function it reruns the onStart causing the user prompt to be saved twice in my database

Any ideas on how to prevent this?

MaxLeiter:
Closes #351

### DIFF
--- a/.changeset/blue-cows-laugh.md
+++ b/.changeset/blue-cows-laugh.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+streams/openai-stream: don't call onStart/onCompletion when recursing
--- a/packages/core/streams/ai-stream.ts
+++ b/packages/core/streams/ai-stream.ts
@@ -15,9 +15,9 @@ export interface FunctionCallPayload {
  * @interface
  */
 export interface AIStreamCallbacks {
-  onStart?: () => Promise<void>
-  onCompletion?: (completion: string) => Promise<void>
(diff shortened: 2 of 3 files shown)


hassanbazzi:
I think this was an overlook. OnComplete should definitely still be fired. I have reverted to ai version right before this in order to actually use function calls correctly. All you have to do is detect `function_call` at the start of `onComplete` if the person doesn't want to do anything with it.

MaxLeiter:
@hassanbazzi you're trying to store the results of the recursive function calls? 🤔 I can see reasons to do so but it'd be tricky for you to differentiate where it occurred in the 'call stack'.

hassanbazzi:
I need to run the final onComplete.

Currently only the onComplete of the first recursive call is fired. 

Meaning the final message of the bot (which is the one that should be stored for context) is not being sent as a final completion.

MaxLeiter:
@hassanbazzi a PR re-adding `onComplete` to the callbacks would be merged if you can open it!

hassanbazzi:
@MaxLeiter done <3

Issue #362, opened 2023-07-21
Issue for Claude models when streaming the responses in non-English


olefirenko:
It can be reproduced here https://sdk.vercel.ai/. So far noticed issues with the Japanese and Russian languages. 
Some letters are doubled and there are some encoding issues.

<img width="637" alt="Screenshot 2023-07-21 at 3 19 31 PM" src="https://github.com/vercel-labs/ai/assets/5434459/adada78a-0592-4c10-80da-e15b59086a44">

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes concrete incorrect output when using an existing streaming capability with non-English text, so it is a bug.

### Evidence
- Streaming Claude responses in Japanese and Russian reportedly doubles some letters and introduces encoding corruption.
- The attached screenshot shows malformed replacement characters and duplicated or incorrectly joined Cyrillic text in the streamed output.
- The failure is reproducible on the AI SDK website.

PR #363, opened 2023-07-21
Fix parsing Anthropic streams issue

olefirenko:
Some languages, like those using Cyrillic, Chinese or other non-Latin alphabets, often use multibyte characters. If a single multibyte character gets split between two chunks of data, it could result in unexpected behavior such as duplication or encoding issues.

The solution to address the issue is to handle multibyte characters properly while dealing with streams. We can use the Buffer class to handle multibyte characters.

### DIFF
--- a/packages/core/streams/anthropic-stream.ts
+++ b/packages/core/streams/anthropic-stream.ts
@@ -1,7 +1,7 @@
 import { AIStream, type AIStreamCallbacks } from './ai-stream'
 
 function parseAnthropicStream(): (data: string) => string | void {
-  let previous = ''
+  let previous = Buffer.from('', 'utf8')
 
   return data => {
     const json = JSON.parse(data as string) as {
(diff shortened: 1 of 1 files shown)


olefirenko:
Related to #362

jridgewell (changes requested):
This isn't going to help prevent `�` chars. If a partial char is delivered to us in `json.completion`, the `Buffer.from()`, `buffer.slice()`, and `buffer.toString()` will preserve that partial char into the output.

This is all predicated on Anthropic delivering us a partial char to begin with, and maybe they expect us to receive the full char in the next `completion` (because it's cumulative). The problem is that we're slicing the text into deltas, so we'll deliver two separate strings where 1 is expected.

If that's the case, we can solve this pretty easily by trimming high-surrogates from the text before we slice:

```js
// Remove any lone high-surrogates from the end of the string.
const text = json.completion.replace(/[\uD800–\uDBFF]$/, '');

const delta = text.slice(previous.length);
previous = text;
```

xycjscs:
It is solved on https://sdk.vercel.ai/. 
However when I deploy this sdk on Vercel with official demo, this problem is still there.
Could someone help me out on this?

PR #398, opened 2023-07-29
docs(core): fix link in `README.md` of core

EastSun5566:
Hi Vercel team,

If you only use `js.langchain.com/docs`, it will link to the relative path of the current URL. Therefore, I added `https://` before it.


MaxLeiter (approved):
Thanks!

PR #405, opened 2023-08-01
Add replicate example and fix event termination

jaredpalmer:
Fixes Replicate's non-standard SSE termination. They send an event named `done` with `data: {}` (https://replicate.com/docs/streaming). 

**The Problem**
CleanShot 2023-08-01 at 09 29 18@2x
CleanShot 2023-08-01 at 09 29 13@2x

Our current `customParser` abstraction only handles `event.data` which in the case of termination = `{}`. However, this could be part of the regular response output, and so isn't a sufficient conditional to decide termination. As a workaround, this ships an additional termination check in `AIStreamParser`. 

If we don't like this abstraction, we can refactor `customParser` to take the whole `event` and then move this logic into `ReplicateStream`.


MaxLeiter (approved):

zeke:
Nice!

PR #420, opened 2023-08-06
fix: a small typo in docs

AhmedBaset:
In the Next.js app router guide there is this example: 
```js
    model: 'gpt-4',
    stream: true,
    messages: [
      {
        role: 'system',
        content:
          searchParams['prompt'] ?? 'Give me code for generating a JSX button'
      }
    ]
  })
```

But the role should be `user`


MaxLeiter (approved):
Thanks!

PR #425, opened 2023-08-07
[RFC/experimental] StreamData API for streaming additional data to the client

MaxLeiter:
_This is an experimental API and is subject to change. Please leave (non-code related) feedback! the code is WIP and is mostly for experimentation_

One of the  most common requested features has been how to stream additional data to the client. I think we have a pretty good solution now.

There's a lot of duplicate code here because we're going to support the old protocol and new one while experimenting. As a result this functionality is only implemented in `ai/react`'s `useChat`

##  On the server (`StreamData`)

The `StreamData` class is a new TransformStream wrapper available on the server from the root `ai` package. 

You instantiate it like any other class:

```ts
  const data = new experimental_StreamData()
```

and use it in your API:

```diff
+  const data = new experimental_StreamData()
  const stream = OpenAIStream(response, {
    experimental_onFunctionCall: async (
      { name, arguments: args },
      createFunctionCallMessages
    ) => {
      if (name === 'get_current_weather') {
        // Call a weather API here
        const weatherData = {
          temperature: 20,
          unit: args.format === 'celsius' ? 'C' : 'F'
        }

+        data.append({
+          text: 'Some custom data'
+        })

        return;
      }
    }, 
+   onFinish() { 
+ 	// IMPORTANT! the user must manually close the stream 
+     data.close();
+	},
+  // IMPORTANT! this is only required while we support both protocols
+  experimental_StreamData: true
  })

+ data.append({ done: true })

  return new StreamingTextResponse(finalStream, {
+ // IMPORTANT! behind the scenes, StreamingTextResponse adds a header if `data` is present. If you aren't using StreamingTextResponse you need to add the header yourself. 
+  {},
+, data);
```

## On the client

The `useChat` and `useCompletion` hooks return a new object, `data`:

```js
  const { messages, input, handleInputChange, handleSubmit, data } = useChat({
```

`data` is the array of JS objects you `append`ed to with `StreamData`. Correlating data to specific messages is _not_ handled by the SDK and will be managed by the user. In the future this will likely change, and we'll maintain the global `data` and a `message.data`.

## How it works

This requires moving the API from a simple raw text response to a prefixed protocol inspired by React Server Components. Implementation-wise this can later be migrated to Server-Side Events.  

The current prefix map looks like this, and is exported so it can be used in userspace: 

```
/**
 * The map of prefixes for data in the stream
 *
 * - 0: Text from the LLM response
 * - 1: (OpenAI) function_call responses
 * - 2: custom JSON added by the user using `Data`
 *
 * Example:
 * ```
 * 0:Vercel
 * 0:'s
 * 0: AI
 * 0: AI
 * 0: SDK
 * 0: is great
 * 0:!
 * 2: { "someJson": "value" }
 * 1: {"function_call": {"name": "get_current_weather", "arguments": "{\\n\\"location\\": \\"Charlottesville, Virginia\\",\\n\\"format\\": \\"celsius\\"\\n}"}}
 *```
 */
export const StreamStringPrefixes = {
  text: 0,
  function_call: 1,
  data: 2
  // user_err: 3?
} as const
```

### Caveats

- Implementation complexity increases. For now, this is only implemented in `useChat` until it's less experimental/more widely agreed upon
- Client-side parsing is no longer as simple as fetching the request. Non-SDK client's will need to manually parse the prefixes and construct messages, but we can provide helpers to make it simpler. 
- More bandwidth is used, although the number of tokens remains the same, so this isn't a very big deal. 

### What's next?

- this is basically SSE but only consumed by us. In the future we can adjust the transport format. 
- data tied to each `message` if they come in the same response


dnsosebee:
Thanks for this! This is exactly the API I was imagining to solve https://github.com/vercel-labs/ai/issues/350.

My use case involves per-message data, but I can understand why a global data structure affords more flexibility and will probably cover more use cases. I would just do some `zip`ping for my use case

newme616:
This is a great idea! Would love to see this in production soon

mosnicholas:
is the idea that this is also the way to notify the client that we're hitting an API but action on them server side? Similar to how plugins work?

Screenshot 2023-08-15 at 7 54 44 PM

packages/core/streams/openai-stream.ts
```diff
@@ -315,7 +313,11 @@ function createFunctionCallTransformer(
 
       // Stream as normal
       if (!isFunctionStreamingIn) {
-        controller.enqueue(chunk)
+        controller.enqueue(
+          isComplexMode
+            ? textEncoder.encode(getStreamString('text', message))
+            : chunk
```
shuding, line 319: Do we need to encode if we have `experimental_onFunctionCall` but not `experimental_streamData`?

packages/core/streams/stream-data.ts
```diff
@@ -0,0 +1,104 @@
...
+        self.controller = controller
+      },
+      transform: async (chunk, controller) => {
+        controller.enqueue(chunk)
+
+        // add buffered data to the stream
+        if (self.data.length > 0) {
+          const encodedData = self.encoder.encode(
+            getStreamString('data', JSON.stringify(self.data))
+          )
```
shuding, line 42: It directly encodes the data as an array here, but inside `flush` it encodes data items one by one. I think we can do the same there.

shuding (commented):
Looks great in general!

packages/core/streams/openai-stream.ts
MaxLeiter: Yeah, because all complex mode cares about is which protocol to use. You don't need function handlers to use the protocol.

shuding (approved):

MaxLeiter:
I believe this PR is ready but would like some user feedback before shipping. 

You can see how to try it here: https://github.com/vercel/ai/pull/437#issuecomment-1681388183

MaxLeiter:
@mosnicholas the class is exported here: https://github.com/vercel/ai/pull/425/files#diff-194b2904c703afc59e264e2bbfa5e574eb30432742564c7908366fab48e334b1 and i’ve verified its working for me, so I suspect yarn is caching your node_modules or something

mosnicholas:
FWIW @MaxLeiter -- i'm seeing the code locally, so I don't think it's an issue with picking up the wrong package.

Screenshot 2023-08-17 at 12 18 40 PM

mosnicholas:
@MaxLeiter -- found the issue. When I uncomment `experimental-edge` it is no longer undefined.
Screenshot 2023-08-17 at 12 22 27 PM

packages/core/react/use-chat.ts
```diff
@@ -137,58 +143,153 @@ const getStreamedResponse = async (
     throw new Error('The response body is empty.')
   }
 
-  let streamedResponse = ''
+  const isComplexMode = res.headers.get('X-Experimental-Stream-Data') === 'true'
```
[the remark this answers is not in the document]
MaxLeiter: I'm not sure how you reviewed this change as I fixed it in 7634bf9f6d83a75abc72d30e0cbce7eb8dc6f34a

MaxLeiter:
<picture data-single-emoji=":wow:" title=":wow:">:wow:</picture> great catch @mosnicholas! I'll dig in.

.changeset/green-masks-repair.md
```diff
@@ -0,0 +1,6 @@
+---
+'ai': patch
+---
+
+Add experimental_DataStream and new opt-in wire protocol to enable streaming additional data. See https://github.com/vercel/ai/pull/425.
```
mosnicholas, line 5: nit: typo, should be: experimental_StreamData

MaxLeiter:
@mosnicholas I've published a version at `ai-stream-experimental` if you can try that. npmjs.com/ai-stream-experimental

mosnicholas:
@MaxLeiter 🔥 it's working now. but the output is coming out strange:
Screenshot 2023-08-17 at 1 13 45 PM


If it helps, happy to hop on a quick call.

MaxLeiter:
@mosnicholas I think you missed a step opting into the new protocol. You set experimental_StreamData: true to true and are passing `data` into the `StreamingTextResponse` (or manually adding the COMPLEX_HEADER yourself?)

Happy to hop on a call as well, feel free to email me: max.leiter@vercel.com

mosnicholas:
@MaxLeiter -- just tested it again, it worked! :)

2 comments:
1. Is it possible to attach function call data to specific messages. It seems like it's right now returning a data array that represents the entirety of additional data you've fed through the API. 
2. It only updates the data array when a new message begins to stream. The plugins model is really cool -- it allows the backend to inform the user that it's fetching data as it's happening (along with the parameters it's using). IS that something `ai` might be able to accomplish too?

Example for (1):
```
if (name === 'get_weather') {
    const { location } = args as { location: string };

    addtlData.append({
      functionName: name,
      location,
    });

    const weather = await getWeather(location);
    const newMessages = createFunctionCallMessages(weather);

    return getChatAgentResponse({
      question: new_message,
    });
  }
```

If you have called this twice with LA / SF, the data object in the FE will just contain the following object:
```
[
  {
    "functionName": "get_weather",
    "location": "LA"
  },
  {
    "functionName": "get_weather",
    "location": "SF"
  }
]
```

Maybe it's on us in the FE to render it properly, just curious your thoughts on this.

MaxLeiter:
@mosnicholas great! thanks for the feedback

The intent of this version is just for global data, with per-message data as a quick follow-up:

> data is the array of JS objects you appended to with StreamData. Correlating data to specific messages is not handled by the SDK and will be managed by the user. In the future this will likely change, and we'll maintain the global data and a message.data.

One tricky thing I'm running into adding message.data is this code:

```
        const data = prefixMap['data']
        const responseMessage = prefixMap['text']
        let functionCall = prefixMap['function_call']
        const functionCallMessage: Message | null = null
        
        if (functionCall) {
          const parsedFunctionCall: CreateChatCompletionRequestMessage.FunctionCall =
            JSON.parse(functionCall as string).function_call

          functionCallMessage = {
            ...<code omit>,
            data
          }
        }

        if (responseMessage) {
          responseMessage.data = data
        }

        // We add function calls and response messages to the messages[], but data is its own thing
        const merged = [functionCallMessage, responseMessage].filter(
          Boolean
        ) as Message[]

        mutate([...chatRequest.messages, ...merged], false)
        mutateStreamData([...(existingData || []), ...(data || [])])
```

We can't differentiate if `data` belongs to the function call message or the chat message (if they're sent in the same response). Do you see that being problematic?

shuding:
> It only updates the data array when a new message begins to stream.

This is something we can improve in a follow up PR as well, thanks for the feedback!

rajeshdavidbabu:
> @MaxLeiter 🔥 it's working now. but the output is coming out strange: Screenshot 2023-08-17 at 1 13 45 PM
> 
> If it helps, happy to hop on a quick call.

I am facing the same issue now, I also added COMPLEX_HEADER and dint fix it for me.

rajeshdavidbabu:
Screenshot 2023-08-18 at 12 49 50

Am I missing something, I am using stream from Langchain Stream so not entirely sure where to add the experimental_StreamData to true.

MaxLeiter:
@rajeshdavidbabu ah the new parser hasn't been implemented in LangChainStream yet, sorry. 

If you'd like to contribute it, I believe it's mostly a case of removing the `pipeThrough(createStreamDataTransformer())` line and adding `getStreamString()` around the `writer.write()`s 

https://github.com/vercel/ai/blob/main/packages/core/streams/langchain-stream.ts#L33-L68

rajeshdavidbabu:
Nice. Can take a look !

rl3250:
@MaxLeiter hi there, we're trying the feature but we got a weird behavior that on the front end it keeps triggering requests.

We're using next-js@13.4.9, ai@2.2.6, and enabled the feature flags.

checking on the console log in browser, we did see custom data being piped, but on the front end, the streamed message first completes, and then vanished, and then it keeps re-rendering the ui and re triggering multiple request with the same user's last input.

```ts
  // Convert the response into a friendly text-stream
    onStart: async () => {
      // save to db
    },
    onToken: async (_token: string) => {
      // log
    },
    onCompletion: async (completion: string) => {
      // save to db
    },
    onFinal: async (_completion: string) => {
      // close the stream
      await customData.close();
    },
    // experimental flag to stream custom data.
    experimental_streamData: true,
  });
  return new StreamingTextResponse(stream, {}, customData);
  ```

**Comment by rajeshdavidbabu** [comment]
Afaik, its already been implemented https://github.com/vercel/ai/blob/main/packages/core/streams/stream-data.ts#L111.

So I simply have to pass the experimental_streamData flag to be true when I create an instance of LangChainStream

```
const { stream, handlers } = LangChainStream({
      experimental_streamData: true,
    });
```

This works for me.

**Comment by sebastianhoitz** [comment]
@MaxLeiter thank you so much, this is really cool - one question: You said you might be able to get per-message data working. I don't see anything locally at the moment, only the global data.

Is this because I'm using a LangChain agent? Or is it actually missing?

**Comment by MaxLeiter** [comment]
@sebastianhoitz it's probably only a few line change but it wasn't included in the initial version:

> Correlating data to specific messages is not handled by the SDK and will be managed by the user. In the future this will likely change, and we'll maintain the global data and a message.data.

**Comment by dnsosebee** [comment]
Thanks for the work on this. I'm going to try out function calling and so may not need to stream extra data.

However I've been thinking about DX and have an idea: I think the core data of RAG apps is not a _message_ log, it's an _event_ log containing user messages, function calls, function responses, and assistant messages. A RAG app needs to present views of that event log to both the user and the assistant; the assistant's view is a message log, and the user's view is whatever JSON drives their UI. The current API (pre-experimental_streamData), which thinks of things as OpenAI-style message logs rather than generic event logs, makes it hard to handle those two different views of the data.

For example, in my RAG app I want my search results to be formatted as a plaintext system message for the OpenAI message array, but want to view those same results as JSON client side, so that I can present a "related research" UI section. Under the current functionCalling API that might involve parsing system messages client side, which is unwieldy. With the current streamData API that would involve interleaving the data array back into my message history as system messages, which is also a bit awkward. Either that or duplicating data as system messages AND streamData.

So I think my ideal API would be to have a generic type called "event" that looks something like this:
```typescript
type Event<T> = {
  name: string // or generic string literal
  data: T
}
```
This `Event` type would be part of message history, and I would set up functions on client and server side to determine how to map that event to a system message for OpenAI, and how to interpret the event data as UI client side.

This is a bit different from message-correlated data, since the data event is a neighboring member in the message/event array rather than an additional field of a message. I think that's more sensible, since fundamentally the data event has its own chronological place in the stream, you can even think of a data event as a message from a 3rd party agent, like a 'search agent' who just sent a message to the stream.

Curious to hear people's thoughts on this!

tbray-sch:
Is there a way I can cleanly return from experimental_onFunctionCall without returning a new chat completion?

My use case is that I want to stream the result of the function call directly to the client without creating a new chat completion.

In the code below I'm just naively returning but it looks like there's a periodic retry that's causing the function to get called repeatedly.

```
  const data = new experimental_StreamData()
      { name, arguments: args },
      createFunctionCallMessages
    ) => {
      if (name === 'get_private_stuff') {
        const result = await runFunction(name, args);

        data.append({
          privateStuff: JSON.stringify(result)
        })

        // The examples return a new chat completion with function call messages but I don't want to do that.
        // It seems like returning here is causing retry logic to run the previous completion repeatedly
        return;
      }
    },
    onCompletion(completion) {
      console.log('completion', completion)
    },
    onFinal(completion) {
      console.log('onFinal');
      data.close()
    },
    experimental_streamData: true
  })

```

benlaval23:
I am getting the following error here:

Attempted import error: 'experimental_StreamData' is not exported from 'ai' (imported as 'experimental_StreamData').



`import {
  OpenAIStream,
  StreamingTextResponse,
  experimental_StreamData,
} from 'ai';`

... some code

`    const data = new experimental_StreamData();

    const stream = OpenAIStream(finalReponse, {
      onFinal: message => {
        data.append(functionResponse); 
        data.close(); 
      },
      experimental_streamData: true,
    });
    
    return new StreamingTextResponse(stream, {}, data);`
    
   Is this not still supported or am I doing something wrong?

aaron5670:
> > @MaxLeiter 🔥 it's working now. but the output is coming out strange: Screenshot 2023-08-17 at 1 13 45 PM
> > If it helps, happy to hop on a quick call.
> 
> I am facing the same issue now, I also added COMPLEX_HEADER and dint fix it for me.

I have the same issue, this is my response:

```
[
    {
        "content": "Hello",
        "role": "user",
        "createdAt": "2024-01-19T09:07:36.443Z",
        "id": "31kQ0KF"
    },
    {
        "id": "ES6PN2g",
        "createdAt": "2024-01-19T09:07:39.792Z",
        "content": "2:[{\"text\":\"Hello, how are you?\"}]\n0:\"Bonjour\"\n0:\"!\"\n0:\" Comment\"\n0:\" puis\"\n0:\"-\"\n0:\"je\"\n0:\" vous\"\n0:\" aider\"\n0:\" aujourd\"\n0:\"'hui\"\n0:\"?\"\n",
        "role": "assistant"
    }
]
```

And this is my backend code:

```typescript
import {experimental_StreamData, OpenAIStream, StreamingTextResponse} from 'ai';
import {ChatCompletionMessage} from 'openai/resources/chat/completions';
import {createOpenAIEmbedding} from "@/helpers/createOpenAIEmbedding";
import {pineconeIndex} from "@/helpers/pinecone";
import {openai} from "@/helpers/openai";

  if (!process.env.OPENAI_API_KEY) {
    return Response.json({error: "OpenAI is not enabled, check your env variables"}, {status: 400});
  }

  try {
    const body = await req.json();
    const messages: ChatCompletionMessage[] = body.messages;

    const messagesTruncated = messages.slice(-6);

    const embedding = await createOpenAIEmbedding(
      messagesTruncated.map((message) => message.content).join("\n"),
    );

    const vectorQueryResponse = await pineconeIndex.query({
      vector: embedding,
      topK: 4,
      includeMetadata: true,
      filter: {type: 'product'},
    });

    console.log('vectorQueryResponse', vectorQueryResponse.matches.map((document) => document.metadata?.title));

    const systemMessage: ChatCompletionMessage = {
      role: "assistant",
      content:
        `
        You are a helpful assistant of the online e-commerce store.
        It is very important that you always answer in French, you must not deviate from this at any cost!
        Use the following bits of context to answer the question at the end.
        Make sure your answer consists of a maximum of about 45 words.
        If you don't know the answer, just say you don't know, don't go making up answers.
        ----------------
        CONTEXT:
        ${vectorQueryResponse.matches
          .map((document) => {
            return `Title: ${document.metadata?.title}\n\nContent:\n${document.metadata?.text}`
          })
          .join("\n\n")}
         `
    };

      model: "gpt-3.5-turbo",
      stream: true,
      messages: [systemMessage, ...messagesTruncated],
    });

    // Instantiate the StreamData. It works with all API providers.

        { name, arguments: args },
        createFunctionCallMessages,
      ) => {
        if (name === 'get_current_weather') {
          // Call a weather API here
          const weatherData = {
            temperature: 20,
            unit: args.format === 'celsius' ? 'C' : 'F',
          };

          data.append({
            text: 'Some custom data',
          });

          const newMessages = createFunctionCallMessages(weatherData);
          return openai.chat.completions.create({
            stream: true,
            model: 'gpt-3.5-turbo-0613',
          });
        }
      },
      onCompletion(completion) {
        console.log('completion', completion);
      },
      onFinal(completion) {
        // IMPORTANT! you must close StreamData manually or the response will never finish.
        data.close();
      },
      // IMPORTANT! until this is stable, you must explicitly opt in to supporting streamData.
      experimental_streamData: true,
    });

    data.append({
      text: 'Hello, how are you?',
    });

    // IMPORTANT! If you aren't using StreamingTextResponse, you MUST have the `X-Experimental-Stream-Data: 'true'` header
    // in your response so the client uses the correct parsing logic.
    return new StreamingTextResponse(stream, {}, data);
  } catch (error) {
    console.error(error);
    return Response.json({error: "Internal server error"}, {status: 500});
  }
}

export async function OPTIONS() {
  return Response.json({status: "OK"});
}

```


What am I doing wrong and how did you guys fixed this?

aaron5670:
@MaxLeiter I found the issue.

We have currently a Gatsby app where I want to use this. I tried it with a Next.js app and then it works as expected.

This is my Next.js front-end code, where it works:

```typescript
"use client";

import {useChat} from "ai/react";

const Page = () => {
  const { data, handleInputChange, handleSubmit, messages } = useChat({
    api: '/api/chat',
  });

  console.log('data', data);
  console.log('messages', messages);

  return (
    <div>
      <form onSubmit={handleSubmit}>
        <input
          type="text"
          name="input"
          onChange={handleInputChange}
        />
        <button type="submit">Send</button>
      </form>
    </div>
  )
}

export default Page
```

Next.js response:

```
[
    {
        "content": "Hello",
        "role": "user",
        "createdAt": "2024-01-19T09:40:26.742Z",
        "id": "cJCvzYF"
    },
    {
        "id": "4W01c2l",
        "role": "assistant",
        "content": "Bonjour! Comment puis-je vous aider aujourd'hui ?",
        "createdAt": "2024-01-19T09:40:29.129Z"
    }
]
```

And this is my Gatsby front-end code, where I have encoding issues:

```typescript
import React from 'react';
import { useChat } from 'ai/react';

export const Page = () => {
  const {
    data, handleInputChange, handleSubmit, messages,
  } = useChat({
    api: 'http://localhost:3000/api/chat',
  });

  console.log('data', data);
  console.log('messages', messages);

  return (
    <div>
      <form onSubmit={handleSubmit}>
        <input
          type="text"
          name="input"
          onChange={handleInputChange}
        />
        <button type="submit">Send</button>
      </form>
    </div>
  );
};
```

Gatsby response:

```
[
    {
        "content": "Hello",
        "role": "user",
        "createdAt": "2024-01-19T09:41:10.737Z",
        "id": "oK7t7Lh"
    },
    {
        "id": "9F2Pg9s",
        "createdAt": "2024-01-19T09:41:12.195Z",
        "content": "0:\"Bonjour\"\n0:\"!\"\n0:\" Comment\"\n0:\" puis\"\n0:\"-\"\n0:\"je\"\n0:\" vous\"\n0:\" aider\"\n0:\" aujourd\"\n0:\"'hui\"\n0:\"?\"\n",
        "role": "assistant"
    }
]
```

zandko:
I also encountered the same problem. It is normal locally. This problem occurs when I call the remote server.

aaron5670:
> I also encountered the same problem. It is normal locally. This problem occurs when I call the remote server.

See #930

PR #436, opened 2023-08-11
fix(core): use ref to store loading state

danielroe:
discovered in https://github.com/vercel-labs/ai/pull/295, using `useSWRV` to track a loading state means we actually get a fetch to the loading API key. We can just use a vue `ref` to handle this state, however.

### DIFF
--- a/.changeset/mighty-apricots-play.md
+++ b/.changeset/mighty-apricots-play.md
@@ -2,4 +2,4 @@
 'ai': patch
 ---
 
-Use native vue `ref` to track loading state, preventing extraneous network requests
+Prevent `isLoading` in vue integration from triggering extraneous network requests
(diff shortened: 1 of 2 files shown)


MaxLeiter:
The intent here is to share the loading state across hook invocations by using the same cache key, so a ref isn't what we want AFAIK.

I think instead we need to pass `null` to the `fetcher` arg of SWRV + add a fallbackData? https://github.com/Kong/swrv#serve-from-cache-only

danielroe:
In that case, I would simply move the ref out of the hook.

MaxLeiter:
It simply aligns with the other implementations (where react is the source of truth). The SWR APIs for each framework makes supporting them all somewhat reasonable.

MaxLeiter (approved):
Thanks!

PR #437, opened 2023-08-11
streams/openai-stream: re-add onCompletion when recursing

hassanbazzi:
# Original Discussion

https://github.com/vercel-labs/ai/pull/357#issuecomment-1669387296

# Issue 
The previous change completely removed the final `onCompletion` call. Meaning `onCompletion` would only be called for the function call, and not for the original request. This prevents users from saving the final response of the bot.

# Solution

This re-adds `onCompletion`. Users that don't want to save the recursive in-between function calls can check for `function_call` in the completion string. It's probably a good idea to add some kind of `isFunctionCompletion` flag to `onCompletion` maybe? Thoughts @MaxLeiter ?

### DIFF
--- a/.changeset/angry-pots-type.md
+++ b/.changeset/angry-pots-type.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+streams/openai-stream: re-add onCompletion when recursing
--- a/packages/core/streams/openai-stream.ts
+++ b/packages/core/streams/openai-stream.ts
@@ -376,13 +376,12 @@ function createFunctionCallTransformer(
 
         // Recursively:
 
-        // We don't want to trigger onStart or onComplete recursively
-        // so we remove them from the callbacks
+        // We don't want to trigger onStart recursively
+        // so we remove it from the callbacks
         // see https://github.com/vercel-labs/ai/issues/351
         const filteredCallbacks: OpenAIStreamCallbacks = {
           ...callbacks,
-          onStart: undefined,
-          onCompletion: undefined
+          onStart: undefined
         }
 
         const openAIStream = OpenAIStream(functionResponse, {


MaxLeiter:
>It's probably a good idea to add some kind of isFunctionCompletion flag to onCompletion maybe

We need to be careful changing the API, and I think `isFunctionCompletion` is too specific, but it's a good idea.
We could try reconstructing/passing a `Message`, and passing that as a second arg to the callback? it'd be a breaking change to replace the `completion` string the callback expects now. 

onCompletion(completion: string, message: Message)

What do you think?

hassanbazzi:
Hey! 

Is this something we want to handle in this pr too? Because all this PR does is re-enable the old way onCompletion works. 

Right now onCompletion does not fire at all at the end with the final completion.

MaxLeiter:
@hassanbazzi I want to avoid more whiplash for users. If we merge this we'll likely keep it as-is for the foreseeable future. So if you're willing to take it on, I'd prefer cleaning it up in one PR.

mosnicholas:
@MaxLeiter this would be super helpful for storing messages without having to call an API call client side when the message is fully typed out. Do you have any advice on other approaches to doing so, or is this the best approach? If so, happy to contribute. My 2 cents would be: instead of passing a Message back, return a list of messages from the various recursive completions, and let the user decide how to handle (eg. one could be for the function call, follow up function call, and final completion). It would be nice to be able to display all that data to the user if they refresh the chat history.

MaxLeiter:
Alright, I made a lot of progress in #425 and I think it's ready to merge after some more thorough testing.


You can try it here: https://files-gbg6if10m-mleiter.vercel.app?ai.tar.gz (`pnpm add https://files-gbg6if10m-mleiter.vercel.app?ai.tar.gz` should work)

I've changed `onComplete` so it's called every completion, and added an `onFinal` callback that runs once everything has completed. This is necessary for calling `data.close()` on the `experimental_StreamData`.

```

import {
  OpenAIStream,
  StreamingTextResponse,
  experimental_StreamData
} from 'ai'
import OpenAI from 'openai'
import { CompletionCreateParams } from 'openai/resources/chat'
// Create an OpenAI API client (that's edge friendly!)
const openai = new OpenAI({
  apiKey: process.env.OPENAI_API_KEY || ''
})

// IMPORTANT! Set the runtime to edge
export const runtime = 'edge'

const functions: CompletionCreateParams.Function[] = [
  {
    name: 'get_current_weather',
    description: 'Get the current weather.',
    parameters: {
      type: 'object',
      properties: {
        format: {
          type: 'string',
          enum: ['celsius', 'fahrenheit'],
          description: 'The temperature unit to use.'
        }
      },
      required: ['format']
    }
  },
  {
    name: 'eval_code_in_browser',
    description: 'Execute javascript code in the browser with eval().',
    parameters: {
      type: 'object',
      properties: {
        code: {
          type: 'string',
          description: `Javascript code that will be directly executed via eval(). Do not use backticks in your response.
           DO NOT include any newlines in your response, and be sure to provide only valid JSON when providing the arguments object.
           The output of the eval() will be returned directly by the function.`
        }
      },
      required: ['code']
    }
  }
]

  const { messages } = await req.json()

    model: 'gpt-3.5-turbo-0613',
    stream: true,
    messages,
    functions
  })

  const data = new experimental_StreamData()
      { name, arguments: args },
      createFunctionCallMessages
    ) => {
      if (name === 'get_current_weather') {
        // Call a weather API here
        const weatherData = {
          temperature: 20,
          unit: args.format === 'celsius' ? 'C' : 'F'
        }

        data.append({
          text: 'Some custom data'
        })

        const newMessages = createFunctionCallMessages(weatherData)
        return openai.chat.completions.create({
          stream: true,
          model: 'gpt-3.5-turbo-0613'
        })
      }
    },
    onCompletion(completion) {
      console.log('completion', completion)
    },
    onFinal(completion) {
      data.close()
    },
    experimental_streamData: true
  })

  data.append({
    text: 'Hello, how are you?'
  })

  return new StreamingTextResponse(
    stream,
    {},
    data
  )
}

```

Please give it a try and let me know!

matannahmani:
@MaxLeiter I tried installing the package but it doesn't work (using pnpm)
( ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY  Broken lockfile: no entry for '/ai/@files-egig7wkkm-mleiter.vercel.app?ai-2.2.1.tar.gz(react@18.2.0)(solid-js@1.7.11)(svelte@4.2.0)(vue@3.3.4)' in pnpm-lock.yaml)

I think this API is a great addition especially if you use tools in Langchain
in my use case I want to stream down to the client the tools and their state so I can show loaders / UI per tool, but with the current implementation I can't figure out a way to stream down the tools in langchain callbacks:

```
  const executor = await AgentCall(history, systemMessage)
  executor
    .call(
      {
        input: currentMessage?.content ?? ''
      },
      {
        callbacks: [
          handlers,
          {
            async handleToolStart(
              tool,
              input,
              runId,
              _parentRunId,
              _tags,
              _metadata
            ) {
              // send a message to the client that the tool has started ? optional send the input of the tool
              // stream.write(`{tool:${tool.name},input:${input},loading:true,runId:${runId}}`)
              handlers.handleToolStart(tool, input, runId)
            }
          },
          {
            async handleToolEnd(output, runId, parentRunId, tags) {
              // send a message to the client that the tool has ended ? optional send the output of the tool
              // stream.write(`{tool:${tool.name},output:${output},loading:false,runId:${runId}}`)
              handlers.handleToolEnd(output, runId)
            }
          }
        ]
      }
    )
```

MaxLeiter:
Hmm, according to https://github.com/yarnpkg/yarn/issues/6811 you may need to yarn i after yarn add @mosnicholas 

@matannahmani I think you need to do the same (pnpm i after pnpm add)

ty both!

mosnicholas:
@MaxLeiter i nuked node_modules, and ran `yarn` many times. The package is correct (when I click into / poke around `node_modules/ai` I see the updated code), it seems like it's not exporting the class or exporting undefined..?

MaxLeiter:
Closing this in favor of #425, thanks a lot @hassanbazzi!

If anyone wants to test 425, you can install the `ai-stream-experimental` package.

PR #447, opened 2023-08-13
fix: handle ping and error events from anthropic

### LINKED ISSUE #383: Error with large context and AnthropicStream?
I'm doing some testing with feeding larger context into the Anthropic models and I'm running into the following error message:
`error - TypeError: Cannot read properties of undefined (reading 'slice')
    at eval (webpack-internal:///(middleware)/./node_modules/ai/dist/chunk-TWW2ODJW.mjs:17:24)
    at onParse (webpack-internal:///(middleware)/./node_modules/ai/dist/chunk-UJV6VDVU.mjs:28:29)
    at parseEventStreamLine (webpack-internal:///(middleware)/./node_modules/eventsource-parser/dist/index.js:77:9)
    at Object.feed (webpack-internal:///(middleware)/./node_modules/eventsource-parser/dist/index.js:65:7)
    at Object.transform (webpack-internal:///(middleware)/./node_modules/ai/dist/chunk-UJV6VDVU.mjs:37:14)
    at g (evalmachine.<anonymous>:102:35)
    at w (evalmachine.<anonymous>:107:14)
    at Object.transform (evalmachine.<anonymous>:1953:33)
    at TransformStreamDefaultController.<anonymous> (evalmachine.<anonymous>:2063:49)
    at pr (evalmachine.<anonymous>:2181:15)`

My edge handler is pretty basic, so I don't think its anything weird there plus it works fine with smaller requests (abbreviated code path):
```
      const altPrompt = `
        ${userPrompt}

        Documents to reference: \n${combinedText}
      `;

      const anthropicResponse = await fetch(
        "https://api.anthropic.com/v1/complete",
        {
          method: "POST",
          headers: {
            "Content-Type": "application/json",
            "x-api-key": process.env.ANTHROPIC_API_KEY || "",
          },
          body: JSON.stringify({
            prompt: `\n\nHuman: ${altPrompt}\n\nAssistant:`,
            model: "claude-2",
            max_tokens_to_sample: 10000,
            temperature: 0.0,
            stream: true,
          }),
        }
      );

      stream = AnthropicStream(anthropicResponse);

     return new StreamingTextResponse(stream);
```

There doesn't seem to be an issue with smaller context requests to the endpoint so not sure if something is timing out or what is happening. I've tested the requests in a jupyter notebook hitting the anthropic API directly and I get valid responses so doesn't look like its a server error from Anthropic's side and I'm still below the context window limit. Any tips or help would be greatly appreciated and thanks to everyone who works on this repo!

NiluK:
This PR fixes  #383 

Currently, when the stream encounters a ping or an error event during the stream it breaks down entirely. Sort of expected for Error if errors aren't properly handled but this should not break on ping. 

The main reason for this is the line 

` const text = json.completion`

According to the documentation from Anthrophic, neither the ping nor the error object have completion as a key: 

https://docs.anthropic.com/claude/reference/streaming#events

Thus text is going to be `undefined` in those occasions and JS throws an error saying it caannot find slide of undefined

This PR handles both error and ping events from the Anthrophic stream. 

There is probably a better way to do this upstream but this fixes a bug on the SDK for major Anthrophic use cases.

### DIFF
--- a/.changeset/fast-dryers-compare.md
+++ b/.changeset/fast-dryers-compare.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+anthropic-stream: handle ping and error events from anthropic
--- a/packages/core/streams/anthropic-stream.ts
+++ b/packages/core/streams/anthropic-stream.ts
@@ -1,17 +1,54 @@
 import { AIStream, type AIStreamCallbacks } from './ai-stream'
 
+/**
+ * Parses the anthropic stream data.
+ * @returns A function that accepts a string `data` and returns a parsed string or `void`.
+ */
 function parseAnthropicStream(): (data: string) => string | void {
   let previous = ''
 
-  return data => {
-    const json = JSON.parse(data as string) as {
-      completion: string
-      stop: string | null
-      stop_reason: string | null
-      truncated: boolean
-      log_id: string
-      model: string
-      exception: string | null
+  /**
+   * @param data - The input data string to parse.
+   * @throws An error if the data contains an error event.
+   * @returns The parsed delta or `void`.
+   */
+  return function (data: string): string | void {
+    const json = JSON.parse(data as string) as
+      | {
+          completion: string
+          stop: string | null
+          stop_reason: string | null
+          truncated: boolean
+          log_id: string
+          model: string
+          exception: string | null
+        }
+      | {
+          error: {
+            type: string
+            message: string
+          }
+        }
+
+    /**
+     * Handles the error event.
+     * @event
+     * @property data - The data for the error event.
+     * @property data.error - The error details.
+     * @property data.error.type - The type of the error (e.g., "overloaded_error").
+     * @property data.error.message - The message for the error (e.g., "Overloaded").
+     */
+    if ('error' in json) {
+      throw new Error(json.error.message)
+    }
+
+    /**
+     * Handles the ping event.
+     * @event
+     * @property data - The data for the ping event (empty object).
+     */
+    if (!json.completion) {
+      return
     }
 
     // Anthropic's `completion` field is cumulative unlike OpenAI's


MaxLeiter (commented):
Thank you! Would you mind adding a short addition about handling these to the docs?
It's probably worth adding here: https://github.com/vercel-labs/ai/blob/main/docs/pages/docs/guides/providers/replicate.mdx 

and here: https://github.com/vercel-labs/ai/blob/main/docs/pages/docs/api-reference/replicate-stream.mdx

NiluK:
Sure, I assume you mean these two right?

https://github.com/vercel-labs/ai/blob/main/docs/pages/docs/guides/providers/anthropic.mdx
https://github.com/vercel-labs/ai/blob/main/docs/pages/docs/api-reference/anthropic-stream.mdx

Will write something up soon there.

jaredpalmer:
@rattrayalex should we merge or hold until your pr is ready?

jaredpalmer:
@rattrayalex should we merge or hold until your pr is ready?

NiluK:
This should handle it regardless of the Anthropic SDK Changes I think, not everyone will use the SDK as sometimes companies may not want to add additional  dependencies to their project (eg if they use a multitude of providers, fetch is probably a more flexible pathway than installing the client libs of every provider)

rattrayalex:
Ah, I'll incorporate this change in my PR (https://github.com/vercel/ai/pull/448). Thank you @NiluK 

TBH I think it'd be better to change the behavior of `createEventStreamTransformer` to not assume that there's only one kind of `data` event, but that'd require a deeper refactor.

rattrayalex:
@jaredpalmer my PR now incorporates this functionality – thank you very much for adding it @NiluK !

MaxLeiter:
@NiluK I'm going to close this in favor of @rattrayalex's PR. Thank you for your efforts here!

>TBH I think it'd be better to change the behavior of createEventStreamTransformer to not assume that there's only one kind of data event, but that'd require a deeper refactor.

@rattrayalex I agree, but some error handling is anthropic specific and shouldn't be in there. I wouldn't be opposed if you want to add some sort of error handling here: https://github.com/vercel/ai/blob/4df2a4965050410cd090a7a8b210df27b0e792f2/packages/core/streams/ai-stream.ts#L57

NiluK:
awesome, thanks @rattrayalex, @MaxLeiter

rattrayalex:
Thanks @MaxLeiter ! 

> I wouldn't be opposed if you want to add some sort of error handling here:


Frankly I don't think that stuff should be in the core `ai-stream` file at all, but dictated per API provider (eg, they could get a callback `({ event, data }) => …` and decide what to do with it)

PR #448, opened 2023-08-13
Add support for Anthropic SDK and newer Anthropic API Version(s)

rattrayalex:
Mimics https://github.com/vercel-labs/ai/pull/356 for Anthropic, using their SDK: https://github.com/anthropics/anthropic-sdk-typescript

Motivated by an issue on the Anthropic TS SDK repo, in which a user reports that the vercel/ai wasn't working with anthropic for them: https://github.com/anthropics/anthropic-sdk-typescript/issues/112

Example usage:

```ts
import Anthropic from '@anthropic-ai/sdk'
import { AnthropicStream, StreamingTextResponse } from 'ai'

const client = new Anthropic()

  const result = await client.completions.create({ stream: true, ...params })
  const stream = AnthropicStream(result)
  return new StreamingTextResponse(stream)
}
```

Note that the SDK uses `anthropic-version: 2023-06-01`, which results in `result.completion` being a diff instead of a snapshot. This PR also adds support for this with raw `Response`.

https://docs.anthropic.com/claude/reference/versioning

cc @jridgewell @MaxLeiter

### DIFF
--- a/packages/core/streams/anthropic-stream.ts
+++ b/packages/core/streams/anthropic-stream.ts
@@ -32,7 +32,7 @@ interface CompletionChunk {
 /**
  * Intended for API versions before 2023-06-01
  */
-function parseAnthropicStreamLegacy(): (data: string) => string | void {
+function parseAnthropicStream(): (data: string) => string | void {
   let previous = ''
 
   return data => {
@@ -46,19 +46,6 @@ function parseAnthropicStreamLegacy(): (data: string) => string | void {
     const delta = text.slice(previous.length)
     previous = text
 
-    return delta
-  }
-}
-
-/**
- * Intended for v2023-06-01 and greater,
- * but may need to be adjusted for future versions.
- * https://docs.anthropic.com/claude/reference/versioning#version-history
- */
-function parseAnthropicStream(): (data: string) => string | void {
-  return data => {
-    const json = JSON.parse(data as string) as CompletionChunk
-    const delta = json.completion
     return delta
   }
 }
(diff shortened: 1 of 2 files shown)


packages/core/streams/anthropic-stream.ts
```diff
@@ -25,9 +50,48 @@ function parseAnthropicStream(): (data: string) => string | void {
...
+  res: Response | AsyncIterable<CompletionChunk>,
   cb?: AIStreamCallbacks
 ): ReadableStream {
-  return AIStream(res, parseAnthropicStream(), cb)
+  if (Symbol.asyncIterator in res) {
+    return readableFromAsyncIterable(streamable(res)).pipeThrough(
+      createCallbacksTransformer(cb)
+    )
+  } else {
+    const apiVersion = res.headers.get('anthropic-version')
```
rattrayalex, line 87: Note this response header does not yet exist, so we shouldn't merge until Anthropic has confirmed they're shipping it.

jaredpalmer: Let us know when to merge here or in slack.

MaxLeiter: I converted this to a draft in the mean time. Thanks @rattrayalex!

rattrayalex: Fair – checked with Anthropic and it sounds like they won't be able to release this right away, so I added a workaround in the meantime. Should be ready for review/merge now.

Once they add the response header, we can revert 8035f47b605c61dabad578ff204ca4e7e5b22a87

NiluK:
this looks good on my end trying it. thanks @rattrayalex

MaxLeiter:
I'm seeing some type issues on my end @rattrayalex, am I missing something?

CleanShot 2023-08-14 at 16 04 45@2x

rattrayalex:
@MaxLeiter two things:

1. You need to `await` the request
2. You need to include `stream: true` in your params

MaxLeiter:
Ah apologies, I had the `stream: true` and removed it when copy/pasting and that sent me down a bad path. Updated the PR description too

rattrayalex:
Thanks for the fixes, Max!

MaxLeiter (approved):

PR #449, opened 2023-08-14
fix: update link to `vercel/ai` github repo

styfle:
`vercel-labs/ai` => `vercel/ai`

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -118,7 +118,7 @@
   "publishConfig": {
     "access": "public"
   },
-  "homepage": "https://vercel.com/templates/ai",
+  "homepage": "https://sdk.vercel.ai/docs",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/vercel/ai.git"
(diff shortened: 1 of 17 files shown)


jaredpalmer (approved):

packages/core/package.json
```diff
@@ -118,13 +118,13 @@
   "publishConfig": {
     "access": "public"
   },
-  "homepage": "https://github.com/vercel-labs/ai#readme",
+  "homepage": "https://vercel.com/templates/ai",
```
styfle, line 121: @jaredpalmer Thoughts on the homepage here?

jaredpalmer:
```suggestion
  "homepage": "https://sdk.vercel.ai/docs",
```

styfle: I'm going to change it to https://sdk.vercel.ai/docs which feels more appropriate

styfle: Glad we both came to the same conclusion 😄

PR #295, opened 2023-07-07
feat: update nuxt example for proper streaming support

Hebilicious:
Resolves #196 

Deployed here => https://nuxt-openai-vercel-hebilicious.vercel.app/

cc @danielroe

PR #455, opened 2023-08-16
docs(examples): update nuxt-langchain example

danielroe:
realised we forgot to update this example in https://github.com/vercel/ai/pull/295


examples/nuxt-langchain/.env.example
```diff
@@ -1 +1 @@
-OPENAI_API_KEY=
+NUXT_OPENAI_API_KEY=
```
MaxLeiter, line 1: is this being used?

MaxLeiter (changes requested):

examples/nuxt-langchain/.env.example
danielroe: Yes, this is consumed by Nuxt runtime config.

MaxLeiter: ah very cool, I didn't realize that's how useRuntimeConfig worked.

MaxLeiter (approved):
Thank you and good catch!

PR #484, opened 2023-08-20
fix: infinite loop for experimental stream data

dizys:
We want to try this new experimental stream data feature, but it is causing our client to infinitely send request.
This diff fixes that. AI SDK has been very helpful, looking forward to the stable version of this feature.

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -391,14 +391,15 @@ export function useChat({
 
         // Using experimental stream data
         if ('messages' in messagesAndDataOrJustMessage) {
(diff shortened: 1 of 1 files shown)


rl3250:
Yes, this fixed [my issue](https://github.com/vercel/ai/pull/425#issuecomment-1685163243) in #425

MaxLeiter:
Thanks for the bug fix! I'm seeing another bug while testing that I can look into first thing tomorrow -- it seems like responses after messages are being cut off. 
CleanShot 2023-08-20 at 18 26 55@2x

MaxLeiter (approved):

PR #494, opened 2023-08-22
docs ai/react: fix spellcheck example code for consistency with API route filename

theosanderson:
The backend snippet below suggests a location of `api/completion/` but the frontend code here currently points to `/api/complete`


MaxLeiter (approved):
Thank you!

PR #511, opened 2023-08-24
unwrap body parameter to make it reactive

dosstx:
In Vue, refs should be unwrapped inside the composable to make them reactive. By using the ref pattern, you allow the composable to react to changes in the value. The missing piece here is that the useChat utility should use unref(body) instead of directly accessing the .value property.

### DIFF
--- a/.changeset/odd-eggs-check.md
+++ b/.changeset/odd-eggs-check.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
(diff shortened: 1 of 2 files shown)


MaxLeiter:
Thank you!

itzaks:
Could this be implemented in useCompletion as well?

MaxLeiter:
A pr would be great, @itzaks!

PR #524, opened 2023-08-29
handle partial chunks in react getStreamedResponse

pandemie:
I have noticed that for extra (still experimental) 'data' (the one with prefix "2:") in a streamed response, I get occasional `JSON.parse` errors, when the stream reader returns a chunk that ends in the middle of a line.

I propose to keep reading until the end of a full line or the end of the stream, whichever comes first.

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -186,7 +186,7 @@ const getStreamedResponse = async (
       }
 
       // Update the chat state with the new message tokens.
-      const lines = decode(value)
+      const lines = decode(concatenatedChunks)
       if (typeof lines === 'string') {
         throw new Error(
           'Invalid response format. Complex mode was set but the response is a string. This should never happen.'
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -184,6 +184,8 @@ const getStreamedResponse = async (
         concatenatedChunks.set(chunk, offset)
         offset += chunk.length
       }
+      chunks.length = 0;
+      totalLength = 0;
 
       // Update the chat state with the new message tokens.
       const lines = decode(concatenatedChunks)
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -184,8 +184,8 @@ const getStreamedResponse = async (
         concatenatedChunks.set(chunk, offset)
         offset += chunk.length
       }
-      chunks.length = 0;
-      totalLength = 0;
+      chunks.length = 0
+      totalLength = 0
 
       // Update the chat state with the new message tokens.
       const lines = decode(concatenatedChunks)
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -165,13 +165,30 @@ const getStreamedResponse = async (
       }
 
       let value = partialValue
-      // while last character of value is not a newline, keep reading
-      while (value[value.length - 1] !== 10) {
-        const { done, value: partialValue } = await reader.read()
-        if (done) {
-          break
+
+      if (value[value.length - 1] !== 10) {
+        // if the last character is not a newline (ASCII code 10), we need to read more chunks
+        let buffer = new Uint8Array(value.length * 2);
+        buffer.set(value, 0); // copy the first chunk into the buffer
+        let offset = value.length;
+        while (value[value.length - 1] !== 10) {
+          // read more chunks until we get a newline
+          const { done, value: partialValue } = await reader.read();
+          if (done) {
+            break;
+          }
+
+          if (offset + partialValue.length > buffer.length) {
+            // resize the buffer if we need more space
+            let newBuffer = new Uint8Array(Math.max(buffer.length * 2, offset + partialValue.length));
+            newBuffer.set(buffer);
+            buffer = newBuffer;
+          }
+
+          buffer.set(partialValue, offset);
+          offset += partialValue.length;
         }
-        value = new Uint8Array([...value, ...partialValue])
+        value = new Uint8Array(buffer.buffer, 0, offset);
       }
 
       // Update the chat state with the new message tokens.
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -156,7 +156,7 @@ const getStreamedResponse = async (
   }
 
   const prefixMap: PrefixMap = {}
-  const NEWLINE = '\n'.charCodeAt(0);
+  const NEWLINE = '\n'.charCodeAt(0)
 
   if (isComplexMode) {
     while (true) {
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -159,10 +159,21 @@ const getStreamedResponse = async (
 
   if (isComplexMode) {
     while (true) {
-      const { done, value } = await reader.read()
+      const { done, value: partialValue } = await reader.read()
       if (done) {
         break
       }
+
+      let value = partialValue
+      // while last character of value is not a newline, keep reading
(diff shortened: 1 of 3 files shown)


packages/core/react/use-chat.ts
```diff
@@ -159,10 +159,21 @@ const getStreamedResponse = async (
...
       }
+
+      let value = partialValue
+      // while last character of value is not a newline, keep reading
+      while (value[value.length - 1] !== 10) {
+        const { done, value: partialValue } = await reader.read()
+        if (done) {
+          break
+        }
+        value = new Uint8Array([...value, ...partialValue])
```
MaxLeiter, line 174: I think we can do better than constantly re-allocating a new array but this works for now

pandemie: I have updated the diff: now a buffer is used and is reallocated exponentially whenever there is not enough space. :)

packages/core/react/use-chat.ts
```diff
@@ -159,10 +159,38 @@ const getStreamedResponse = async (
...
       }
+
+      let value = partialValue
+
+      if (value[value.length - 1] !== 10) {
+        // if the last character is not a newline (ASCII code 10), we need to read more chunks
+        let buffer = new Uint8Array(value.length * 2);
+        buffer.set(value, 0); // copy the first chunk into the buffer
+        let offset = value.length;
+        while (value[value.length - 1] !== 10) {
```
jridgewell, line 174: We can eliminate this inner loop and reuse the outer loop by hoisting the buffer variable outside the outer loop. If the last char isn't a newline, `continue`. Once found, process the buffer.

packages/core/react/use-chat.ts
```diff
@@ -159,10 +159,38 @@ const getStreamedResponse = async (
...
+          // read more chunks until we get a newline
+          const { done, value: partialValue } = await reader.read();
+          if (done) {
+            break;
```
jridgewell, line 178: If we the reader is done and we have a buffer, shouldn't we process that buffer?

packages/core/react/use-chat.ts
```diff
@@ -159,10 +159,38 @@ const getStreamedResponse = async (
...
+      const { done, value: partialValue } = await reader.read()
       if (done) {
         break
       }
+
+      let value = partialValue
+
+      if (value[value.length - 1] !== 10) {
+        // if the last character is not a newline (ASCII code 10), we need to read more chunks
+        let buffer = new Uint8Array(value.length * 2);
```
jridgewell, line 171: We can store an array of chunk and avoid the reallocations. When we process, count the exact size of all buffered chunks and assign.

packages/core/react/use-chat.ts
```diff
@@ -159,10 +159,38 @@ const getStreamedResponse = async (
...
     while (true) {
-      const { done, value } = await reader.read()
+      const { done, value: partialValue } = await reader.read()
       if (done) {
         break
       }
+
+      let value = partialValue
+
+      if (value[value.length - 1] !== 10) {
```
jridgewell, line 169: Nit: Use a self-description variable name.
```suggestion
      const NEWLINE = '\n'.charCodeAt(0);
      if (value[value.length - 1] !== NEWLINE) {
```

We could even hoist `NEWLINE` to the top level scope.

jridgewell (changes requested):

packages/core/react/use-chat.ts
pandemie: The reader only ever returns done == true if there is no new data at all. Whenever done == true, the partialValue will be undefined.

https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read

packages/core/react/use-chat.ts
pandemie: Yes, you are right! I pushed a new commit to do exactly that!

packages/core/react/use-chat.ts
pandemie: Done :)

packages/core/react/use-chat.ts
pandemie: I refactored this all together to use an array to collect the chunks. Still, we can not get rid of an inner loop altogether, since the outer loop is what iterates over complete parsable lines and the inner loop iterates over the chunks in a single line.

MaxLeiter:
Thanks for all the work here @pandemie. Would you mind providing a basic example so I can reproduce locally? Checking the code, all stream data is passed into `JSON.stringify()` so newlines should be `\\n` in the `2: ...` line

pandemie:
> Thanks for all the work here @pandemie. Would you mind providing a basic example so I can reproduce locally? Checking the code, all stream data is passed into `JSON.stringify()` so newlines should be `\\n` in the `2: ...` line

Thank **you** for being so prompt in reviewing this. :)

I have modified the next-openai example to trigger the error. Here it is in a branch in my fork:

https://github.com/pandemie/ai/commit/de32d5ca9fcea7b019b538a051ea31702874a6ec

The error is triggered non-deterministically, but gets more likely when the data is large, this is why I append a large data object.
Screenshot 2023-08-31 at 16 30 22

MaxLeiter:
Ah, interesting. I think this is happening at the network/Next.js level, so this fix seems correct.

packages/core/react/use-chat.ts
```diff
@@ -156,13 +156,40 @@ const getStreamedResponse = async (
   }
 
   const prefixMap: PrefixMap = {}
+  const NEWLINE = '\n'.charCodeAt(0);
```
jridgewell:
```suggestion
  const NEWLINE = '\n'.charCodeAt(0)
  let chunks: Uint8Array[] = []
  let totalLength = 0k
```

packages/core/react/use-chat.ts
```diff
@@ -156,13 +156,40 @@ const getStreamedResponse = async (
...
+        if (done) {
+          break
+        }
+        chunks.push(partialValue)
+        totalLength += partialValue.length
+        if (partialValue[partialValue.length - 1] === NEWLINE) {
+          // if the last character is a newline, we have read the whole line
+          break
+        }
+      }
```
jridgewell, line 178: We can definitely merge the two while loops:

```suggestion
      const { value } = await reader.read()
      if (value) {
        chunks.push(value)
        totalLength += value.length
        if (value[value.length - 1] !== NEWLINE) {
          // if the last character is not a newline, we have not read the whole JSON value
          continue
        }
      }
```

jridgewell (changes requested):

packages/core/react/use-chat.ts
```diff
@@ -156,13 +156,35 @@ const getStreamedResponse = async (
...
+      // concatenate all the chunks into a single Uint8Array
+      let concatenatedChunks = new Uint8Array(totalLength)
+      let offset = 0
+      for (const chunk of chunks) {
+        concatenatedChunks.set(chunk, offset)
+        offset += chunk.length
+      }
+
       // Update the chat state with the new message tokens.
       const lines = decode(value)
```
jridgewell, line 189:
```suggestion
      const lines = decode(concatenatedChunks)
```

packages/core/react/use-chat.ts
```diff
@@ -156,15 +156,37 @@ const getStreamedResponse = async (
...
         break
       }
+
+      // concatenate all the chunks into a single Uint8Array
+      let concatenatedChunks = new Uint8Array(totalLength)
+      let offset = 0
+      for (const chunk of chunks) {
+        concatenatedChunks.set(chunk, offset)
+        offset += chunk.length
+      }
```
jridgewell, line 186:
```suggestion
      }
      chunks.length = 0;
      totalLength = 0;
```

MaxLeiter (approved):

PR #536, opened 2023-09-02
Unref body for useCompletion calls to support reactive data on Vue

itzaks:
Same implementation as https://github.com/vercel/ai/pull/511 but for `useCompletion` in the Vue-package. 

Makes the body-parameter reactive.

Example usage:
````ts
const body = ref({
  username: "itzaks"
})

const { completion, input, complete } = useCompletion({
  headers: { 'Content-Type': 'application/json' },
  body
})
````

### DIFF
--- a/.changeset/eleven-books-matter.md
+++ b/.changeset/eleven-books-matter.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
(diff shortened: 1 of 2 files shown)


MaxLeiter (approved):

MaxLeiter:
Thank you!

PR #537, opened 2023-09-02
add: useChat dynamic body option via function in svelte

cowboycodr:
Add support for dynamic body option via function in Svelte.

Example:
```svelte
<script>
     ...
     function getBody() {
          return {};
     }

     let chat = useChat({..., body: getBody})
</script>
```

### DIFF
--- a/packages/core/svelte/use-chat.ts
+++ b/packages/core/svelte/use-chat.ts
@@ -87,8 +87,8 @@ const getStreamedResponse = async (
               }),
             }),
           ),
-      ...extraMetadata.body,
-      ...chatRequest.options?.body,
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Thanks for the PR @cowboycodr, sorry for missing this. 
I noticed that the ability to pass a custom `body`/`headers` via `complete` and `append` was missing from the docs and just added it. Would that cover your use case here?

cowboycodr:
Yes. Thank you!

MaxLeiter:
https://github.com/vercel/ai/commit/d05a146d19ddfe6736a3a65ed29930d959fd139f

cowboycodr:
@MaxLeiter maybe I'm using the `append` function incorrectly, however, the chatOptions object is never making it to the server side
+page.svelte
image
+server.js
image
console
image

Issue #501, opened 2023-08-23
[feature] provide a callback to useChat for when messages are completed


dilanustek:
The `useChat` hook has an `onFinish` callback which fires when everything is completed. With the new experimental feature to stream additional data, the onFinish callback fires not after the tokens are completed but after any data is also done loading. 

We need to be able to tell when the messages are completed so that we can show a loading state for the additional data that is being calculated based off of the message received.  It would be nice to have an `onCompleted` callback added to the `useChat` hook on the client side. It may also look like a `isMessageLoading` added to the `isLoading` boolean to indicate the difference.

MaxLeiter:
I support an `onCompletion` to closer align the client API and server API. PR's welcome!

ctate:
Great suggestion @dilanustek. I've started a draft PR for this here: https://github.com/vercel/ai/pull/554

lgrammel:
in ai sdk the finish hook is invoked when the response is complete

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests a new client-side callback or loading-state distinction rather than reporting a violation of existing documented behavior.

### Evidence
- The request proposes adding an `onCompleted`/`onCompletion` callback or separate message-loading state to `useChat`.
- A maintainer supports adding `onCompletion` to align the client and server APIs and invites a PR.
- A draft implementation PR was subsequently opened.

PR #554, opened 2023-09-13
[Draft] Feature: Expose onCompletion/onFinal events to client

ctate:
**This PR is a work-in-progress. Feel free to leave feedback and suggestions 😊**

Per https://github.com/vercel/ai/issues/501: (via @dilanustek)

> The `useChat` hook has an `onFinish` callback which fires when everything is completed. With the new experimental feature to stream additional data, the onFinish callback fires not after the tokens are completed but after any data is also done loading.
> 
> We need to be able to tell when the messages are completed so that we can show a loading state for the additional data that is being calculated based off of the message received. It would be nice to have an `onCompleted` callback added to the `useChat` hook on the client side. It may also look like a `isMessageLoading` added to the `isLoading` boolean to indicate the difference.

Since streams that include additional data use "complex mode", I've added two additional StreamStringPrefixes: `completion` and `final`. This enables the server to inform the client when the `onCompletion` and `onFinal` events occur directly via the stream.

Note: I moved the `getStreamString()` patten from each individual stream to the main AI stream to allow access to the onCompletion/onFinal events. To avoid redundant transformations, I also removed this step from the individual streams. This seems to work so far, but please let me know if there is a preferred approach for this. Thank you!

### DIFF
--- a/examples/next-openai/app/api/chat-with-data/route.ts
+++ b/examples/next-openai/app/api/chat-with-data/route.ts
@@ -1,0 +1,44 @@
+// ./app/api/chat/route.ts
+import OpenAI from 'openai';
+import {
+  OpenAIStream,
+  StreamingTextResponse,
+  experimental_StreamData,
+} from 'ai';
+
+// Create an OpenAI API client (that's edge friendly!)
+const openai = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY || '',
+});
+
+// IMPORTANT! Set the runtime to edge
+export const runtime = 'edge';
+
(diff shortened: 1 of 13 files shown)


MaxLeiter:
Thanks for this @ctate, I think it's a very-needed cleanup and a good idea. Exposing the events via new prefixes is the way to go -- we hope to rollout StreamData as stable soon and this is a great addition.

ctate:
Thanks @MaxLeiter! I'd like do a little more testing before this gets merged in so I converted this to a `Draft`. I'll let you know once it's ready for review 🙏

PR #568, opened 2023-09-17
[issue-567] resets error message on (re)submission

### LINKED ISSUE #567: Flush UseChat ChatHelper error on new handleSubmit
Using UseChat ChatHelper `error` would be more intuitive if the error was cleared on a new handleSubmit. I believe it is not currently.

For instance, this would be an intuitive use of `error`:

``` typescript
{error ?
        <div className="text-red-500">Error: {error.message}</div> 
        : null}
 ``` 
 
But the div sticks around forever, even if the next submit is successful. Right now, we can use the helper callbacks to wrap handleSubmit with a flag, making this workaround possible:
 
 ``` typescript
 {error && !successFlag ? // Right now, error does not get cleared on a successful submit. So we use a helper flag to know when to show it.
        <div className="text-red-500">Error: {error.message}</div> 
        : null}
 ```       
  ...but I think it would be more intuitive to have new submissions simply nullify our `error`.

ctate:
Fixes https://github.com/vercel/ai/issues/567

`error` is set to `undefined` for each useChat/useCompletion hook in each framework when `triggerRequest()` is (re)submitted.

### DIFF
--- a/.changeset/unlucky-zebras-wait.md
+++ b/.changeset/unlucky-zebras-wait.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+all providers: reset error message on (re)submission
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -394,6 +394,8 @@ export function useChat({
     async (chatRequest: ChatRequest) => {
       try {
         mutateLoading(true);
+        setError(undefined);
+
         const abortController = new AbortController();
         abortControllerRef.current = abortController;
 
--- a/packages/core/react/use-completion.ts
+++ b/packages/core/react/use-completion.ts
@@ -104,6 +104,7 @@ export function useCompletion({
     async (prompt: string, options?: RequestOptions) => {
       try {
         mutateLoading(true);
+        setError(undefined);
 
         const abortController = new AbortController();
         setAbortController(abortController);
--- a/packages/core/solid/use-chat.ts
+++ b/packages/core/solid/use-chat.ts
@@ -103,7 +103,9 @@ export function useChat({
     options?: RequestOptions,
   ) {
     try {
+      setError(undefined);
       setIsLoading(true);
+
       abortController = new AbortController();
 
       // Do an optimistic update to the chat state to show the updated messages
--- a/packages/core/solid/use-completion.ts
+++ b/packages/core/solid/use-completion.ts
@@ -89,7 +89,9 @@ export function useCompletion({
   let abortController: AbortController | null = null;
   async function triggerRequest(prompt: string, options?: RequestOptions) {
     try {
+      setError(undefined);
       setIsLoading(true);
+
       abortController = new AbortController();
 
       // Empty the completion immediately.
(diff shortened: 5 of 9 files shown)


peterdresslar:
This didn't work for my example, although maybe there is something wrong with my example. However, I do wonder if the triggerRequest hook is called by all handleSubmits, including the helper function? Also, to be honest, it feels to me like the error clearing should be taken care of closer to the UI handler rather than a downstream event, since if the downstream changes it wouldn't be obvious why error clearing doesn't take place.

ctate:
@peterdresslar Can you provide an example?

Steps I used to reproduce (and fix) the error locally:

1. Clone https://github.com/ctate/vercel-ai (my fork)
2. Checkout the `issue-567` branch via `git checkout issue-567`
3. Run `pnpm install && pnpm run build`
4. Run the "next-openai" example: `pnpm --filter=next-openai run dev`
5. Add some debugging code in `examples/next-openai/app/page.tsx` like `{error && <div>{error.message}</div>}`

Then, to test, I simply opened localhost:3000, turned wifi off, asked a question, and saw the error. Then I turned wifi back on and asked another question. The error disappeared and I got a result back. (before the fix, the error never went away)

The event loop for handleSubmit is `handleSubmit()` > `append()` > `triggerRequest()`. `triggerRequest()` is also called via `reload()`. In both scenarios, `triggerRequest()` seems like a good place to clear the error because `triggerRequest()` is also the only place where an error is set (it's also the same function that changes the `isLoading` state).

Lmk if this helps. Thanks!

ctate:
FWIW I agree there is a lack of visibility by putting this in a downstream event, but as a centralized function for handling the interaction between client/server, it seems like `triggerEvent()` should _always_ include this logic. Happy to change this tho if the team thinks it’s best to move this out to each parent event.

peterdresslar:
@ctate It does work; I had the wrong branch of your fork checked out--I apologize.

I went ahead and tested against "vanilla" UseChat.handleSubmit and also by working with the helper callback:

```typescript
  // wrap handleSubmit to do whatever
    const handleSubmitExtended = (e: React.FormEvent<HTMLFormElement>) => {
      console.log('hello I am doing something custom here');
      handleSubmit(e);
    }
 ```
 
 Works as advertised. As for the [other issue](https://github.com/vercel/ai/pull/568#issuecomment-1722577593), I don't feel passionate about it either way. Convenient that `triggerRequest` serves `reload` also (thinking about it: sure, one would always clear `error` on a reload)--but that is sort of an accidental flow through.

Just to confirm, has the OpenAI `error`.`message` that flows through to the client always been this messy? I was getting clean messages from anthropic.

MaxLeiter:
It makes sense for `triggerRequest` to handle this; that aligns more closely with `useSWR`'s handling of it. Thank you both for taking a look and making such high quality contributions 🙌 


> Just to confirm, has the OpenAI error.message that flows through to the client always been this messy? I was getting clean messages from anthropic.

This may deserve another issue, if can you open one? There's some left over work to be done with handling errors in AIStream.

MaxLeiter (approved):

PR #570, opened 2023-09-17
fixes prettier formatting and reverts to CompletionCreateParams

ctate:
Fixes a prettier formatting issue currently breaking CI.

Also it looks like `ChatCompletionCreateParams` isn't available in `openai` at `4.2.0` (it was introduced in `4.4.x`).

Since `openai` is frozen at `4.2.0`, this should be reverted to `CompletionCreateParams`. Alternatively, `openai` should be upgraded.

### DIFF
--- a/docs/pages/docs/guides/providers/openai-functions.mdx
+++ b/docs/pages/docs/guides/providers/openai-functions.mdx
@@ -19,8 +19,8 @@ The `functions` object is an array containing the schema for each function you w
 Here is an example:
 
 ```ts {3, 34}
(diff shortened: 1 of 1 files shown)


peterdresslar:
Just ran into the prettier issue, nice catch.

MaxLeiter:
Thanks @ctate! I’ll merge once tests pass. We should at some point update openai though.

ctate:
@MaxLeiter For sure! I’m happy to look into the update tonight after work.

PR #571, opened 2023-09-18
Example: next-anthropic

peterdresslar:
An example for next-anthropic, as close to the other examples as was practical. Note that inline with the other examples, we use the SDK.


MaxLeiter:
Thank you! This is great

Issue #564, opened 2023-09-15
Docs: Anthropic and other providers...


peterdresslar:
Editing this issues to add in a couple of remarks on the anthropic docs. A working example of the Anthropic provider, taken from the docs and cleaned a bit, is here [(source)](https://github.com/peterdresslar/ai-edge-example-anthropic) and here [(deployed on Vercel)](https://ai-edge-example-anthropic.vercel.app/).

Specifically, Anthropic appears to require a header for the version, and I believe the model is updated to claude-2 even though the URL is v1 for completions. So the docs would minimally be updated with something like:

```typescript
// route.ts
//...
const response = await fetch('https://api.anthropic.com/v1/complete', {
            method: 'POST',
            headers: {
                'Content-Type': 'application/json',
                'x-api-key': process.env.ANTHROPIC_API_KEY,
                'anthropic-version': '2023-06-01'
            },
            body: JSON.stringify({
                prompt: buildPrompt(messages),
                model: 'claude-2',
                max_tokens_to_sample: 300,
                temperature: 0.9,
                stream: true,
            }),
        });
//...
```

As I mentioned in a previous version of this issue, the code above will still cause a typescript error, since TS will complain about the possibility of process.env.* returning null. So, NOT specific to Anthropic, it might be nice to wrap each fetch inside a check for all of the nicely templated provider examples, such as:

```typescript
//route.ts
export async function POST(req: Request) {
//... messages stuff

 if (!process.env.ANTRHOPIC_API_KEY) { // OPENAI_API_KEY, etc.
  throw new Error(`Error, no API key loaded.`); //or something user friendly
 } else {
  const response = await fetch('https://api.anthropic.com/v1/complete', {//... ... ...}
 }
}

peterdresslar:
Just a follow-on note: it might be worth a comment in the Anthropic write-up that the `Human:\n\nAssistant` prompt construct doesn't work like one might think it does.

peterdresslar:
@MaxLeiter This is a close _unless_ you might like to update all the docs examples to include `process.env.*` invocations that do not break typescript when dropped into an IDE (due to possible undefined value). I suppose I am volunteering, but there is a style choice there. I guess the most consistent to do: 

`                'x-api-key': process.env.ANTHROPIC_API_KEY || '' `

...although that is sort of obviously horrible. Let me know. You could also just choose to ignore and close the issue.

lgrammel:
## Identify issue type

Identified as **Documentation**  
[...]

PR #575, opened 2023-09-19
Fix for #564: update docs

peterdresslar:
Update the anthropic docs re: #564 . Minimal updates should make calls work. Note that 

```js
\\ anthropic-stream.mdx
  const { prompt } = await req.json();
\\...
 body: JSON.stringify({
       prompt, \\...
```

...really doesn't work with this API, though it shouldn't cause an explicit error from Claude; rather: nonsense. However, the example is consistent with the others.


MaxLeiter (approved):

MaxLeiter:
Thanks @peterdresslar!

Issue #591, opened 2023-09-24
Using custom id creation function for messages


aberezkin:
Hi! It'd be nice to be able to generate an id for messages (and chat as well) in a way that my backend natively supports. I'm storing the messages in a Postgres database and ids are of type `UUID` which has some performance improvements compared to strings. So what I would imagine is param passed to the `useChat` hook that, if provided, creates an id instead of using `nanoid`.

I can also create a PR implementing something like that.

MaxLeiter:
Fixed by #733. Thanks!

Kirchen99:
I saw this change has been reverted. Is there any reason?

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This is an API customization enhancement rather than a failure of existing documented behavior.

### Evidence
- The issue requests a new `useChat` parameter for supplying a custom message/chat ID generator instead of the built-in `nanoid`.
- The motivation is compatibility with backend-native PostgreSQL UUID identifiers.
- A maintainer states the request was implemented by PR #733.

PR #592, opened 2023-09-24
Custom id generation PoC

aberezkin:
Just an example of what can be a solution to #591. If this looks right I'll proceed to add it to other places where `nanoid` is currently used.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -45,7 +45,7 @@ importers:
         specifier: ^0.6.2
         version: 0.6.2
       ai:
-        specifier: 2.2.21
+        specifier: 2.2.22
         version: link:../../packages/core
       next:
         specifier: 13.4.12
@@ -88,7 +88,7 @@ importers:
   examples/next-fireworks:
     dependencies:
       ai:
-        specifier: 2.2.21
+        specifier: 2.2.22
         version: link:../../packages/core
       next:
         specifier: 13.4.12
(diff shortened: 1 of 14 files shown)


pnpm-lock.yaml
```diff
@@ -46,7 +46,7 @@ importers:
         version: 0.6.2
       ai:
         specifier: 2.2.12
-        version: link:../../packages/core
+        version: 2.2.12(react@18.2.0)(solid-js@1.7.7)(svelte@4.0.1)(vue@3.3.4)
```
shuding, line 49: We might need to revert this change.

shuding (commented):
Looks good to me! Would you mind copying the implementation to other framework APIs as well (Svelte, Vue, Solid)? Thanks!

pnpm-lock.yaml
aberezkin: Yeah, don't know why it's even there to be honest. Maybe I have a different set package manager setup. I'll try to figure it up.

MaxLeiter: @aberezkin if you can fix the merge conflic tin `use-chat` and re-run `pnpm i` from the root directory this should be resolved and we can merge this PR.

aberezkin:
Uh oh, I rebased and force pushed and the PR has closed. Will open another one with addressed comments.

aberezkin:
Here's the new pull request 🤗

https://github.com/vercel/ai/pull/733

PR #595, opened 2023-09-25
enables sending metadata from client and adds example for bidirectional data

ctate:
For this PR, I wanted to share what I think might be the way to send metadata back and forth between client/server.

As an example, the client sends which model to use (gpt-3.5-turbo or gpt-4) and the server sends back a relevant title for the chat based on the context of the conversation.

Not sure if `metadata` is meant to be used yet, but if so, this PR exposes it to the client.

It's typed as `Object` currently to be consistent with current codebase.

Happy to make any changes and add the same functionality to the other frameworks if this is the right way to go. Thanks!

image

### DIFF
--- a/examples/next-openai/app/api/chat-with-data/route.ts
+++ b/examples/next-openai/app/api/chat-with-data/route.ts
@@ -1,0 +1,55 @@
+import {
+  OpenAIStream,
+  StreamingTextResponse,
+  experimental_StreamData,
+} from 'ai';
+import OpenAI from 'openai';
+import { Chat } from 'openai/resources';
+
+// Create an OpenAI API client (that's edge friendly!)
+const openai = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY || '',
+});
+
+// IMPORTANT! Set the runtime to edge
+export const runtime = 'edge';
+
+export async function POST(req: Request) {
+  // `model` is populated and sent by the client
+  const { messages, model } = (await req.json()) as {
+    messages: Chat.CreateChatCompletionRequestMessage[];
+    model: 'gpt-3.5-turbo' | 'gpt-4';
+  };
+
+  const response = await openai.chat.completions.create({
+    model,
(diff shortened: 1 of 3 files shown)


ctate:
Btw I tried using `body` in the `options` argument but this required me to manipulate `input` and `messages` manually. Please let me know which is the preferred method.

StephanieLeeeee (commented):
1

examples/next-openai/app/api/chat-with-data/route.ts
```diff
@@ -0,0 +1,55 @@
...
+} from 'ai';
+import OpenAI from 'openai';
+import { Chat } from 'openai/resources';
+
+// Create an OpenAI API client (that's edge friendly!)
+const openai = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY || '',
+});
+
+// IMPORTANT! Set the runtime to edge
```
StephanieLeeeee, line 14: 1

examples/next-openai/app/api/chat-with-data/route.ts
```diff
@@ -0,0 +1,55 @@
...
+export const runtime = 'edge';
+
+export async function POST(req: Request) {
+  // `model` is populated and sent by the client
+  const { messages, model } = (await req.json()) as {
+    messages: Chat.CreateChatCompletionRequestMessage[];
+    model: 'gpt-3.5-turbo' | 'gpt-4';
+  };
```
StephanieLeeeee, line 22: 2

examples/next-openai/app/bidirectional-data/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+      },
+    );
+
+  useEffect(() => {
+    // server hasn't sent any data yet
+    if (!data) {
+      setTitle('');
+      return;
+    }
+
```
StephanieLeeeee, line 34: 3

PR #605, opened 2023-09-27
Handle function calls that come after the start of a message

woodbridge:
Currently, function calls that come later in a message stream will be treated as part of the message content and displayed directly to the user, eg:

```
Now searching library....{"function_call"....}
```

This change parses them into the expected function_call message format to work as expected.


MaxLeiter:
Thanks @woodbridge. I haven't taken a close look yet but it seems like there's a typescript error:

```
  Error: react/use-chat.ts(308,28): error TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'.
```

packages/core/react/use-chat.ts
```diff
@@ -320,6 +328,52 @@ const getStreamedResponse = async (
...
+    } else {
+      // Check for an embedded function call.
+      // TODO: Support N function calls?
+      let functionCallIndex = streamedResponse.indexOf('{"function_call":');
+      if (functionCallIndex !== -1) {
+        let textResponse = streamedResponse.substring(0, functionCallIndex);
+        let functionCallString = streamedResponse.substring(functionCallIndex);
+
+        // Create a message for the text before the function call
+        if (textResponse) {
```
StefanDjokovic, line 340: I still have to find a way to repro this outside of my application, but there this `if` condition causes the message before the function_call to be duplicated. Without it, the solution works perfectly!

MaxLeiter:
Hey @woodbridge, sorry for the delay here. Would you mind reviewing @StefanDjokovic's comments, then I can give it a final look and approval? Thank you!

packages/core/react/use-chat.ts
```diff
@@ -300,7 +300,15 @@ const getStreamedResponse = async (
         // While the function call is streaming, it will be a string.
         responseMessage['function_call'] = streamedResponse;
       } else {
-        responseMessage['content'] = streamedResponse;
+        // Check for embedded function calls, don't show as streaming in.
+        let functionCallIndex = streamedResponse.indexOf('{"function_call":');
+
+        if (functionCallIndex !== -1) {
+          let textResponse = streamedResponse.substring(0, functionCallIndex);
+          responseMessages['content'] = textResponse;
```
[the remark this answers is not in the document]
woodbridge: ah thx, will fix

MaxLeiter:
Can someone provide a reproduction? I don't believe OpenAI should be including the function calls after text in a message.

StefanDjokovic:
Hi @MaxLeiter , here is a minimal repro based on the ai-chatbot https://github.com/StefanDjokovic/ai-chatbot-repro/pull/1

You can see that at the end of the conversation the "{function: ...}" is printed instead of calling the actual function.

You can also repro this behaviour on ChatGPT with plugins, it is expected for GPT to be able to chat and do function calls in between its own message

packages/core/react/use-chat.ts
StefanDjokovic: Hi @woodbridge , I was able to repro here: https://github.com/StefanDjokovic/ai-chatbot-repro/pull/1 I linked the ai sdk and I see the behaviour of duplicated messages. NOTE: it's possible that there is a bug in the template itself; not sure if that's the case; if you also have a repro were this behaviour doesn't happen lmk and I'll look more into this.

image
image

MaxLeiter: @woodbridge @StefanDjokovic has this been figured out / is it related directly to the changes made here?

woodbridge: yeah believe it's a bug in my changes - havent' had time to find the bug yet, but have been using this fork internally

MaxLeiter: I think we'll need to wait for that to be resolved before merging this. Thanks for your work here 🙏

gediminastub:
Heya, this happens very often to me now too.
Any movement? Thansk!

lgrammel:
The old function calling in `useChat` has been deprecated.

PR #608, opened 2023-09-29
Update openai-functions.mdx

mengxi-ream:
openai change its type name and location in "openai": "^4.10.0",

### DIFF
--- a/docs/pages/docs/guides/providers/openai-functions.mdx
+++ b/docs/pages/docs/guides/providers/openai-functions.mdx
@@ -19,9 +19,9 @@ The `functions` object is an array containing the schema for each function you w
 Here is an example:
 
 ```ts {3, 34}
-import type { CompletionCreateParams } from 'openai/resources/chat';
+import type { ChatCompletionCreateParams } from "openai/resources/chat/completions";
 
-const functions: CompletionCreateParams.Function[] = [
+const functions: ChatCompletionCreateParams.Function[] = [
   {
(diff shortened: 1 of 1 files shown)


MaxLeiter:
Thanks for this! Would you mind updating `openai` in the examples and the types being used there as well?

Crayon-ShinChan:
> Thanks for this! Would you mind updating `openai` in the examples and the types being used there as well?

sorry, do you mean types used in `https://sdk.vercel.ai/docs/guides/providers/openai`? But I didn't find any imported types from openai in that page

peterdresslar:
Is `ChatCompletionCreateParams` really required for this example? The [OpenAI example at the API docs](https://platform.openai.com/docs/guides/gpt/function-calling) just defines `functions` as a plain array. This works fine when plugged into the completion constructor in the documented AI SDK example.

This could also be a chance to opportunistically update the model name in the documentation also to just plain 'gpt-3.5-turbo'

Crayon-ShinChan:
> Is `ChatCompletionCreateParams` really required for this example? The [OpenAI example at the API docs](https://platform.openai.com/docs/guides/gpt/function-calling) just defines `functions` as a plain array. This works fine when plugged into the completion constructor in the documented AI SDK example.
> 
> 
> 
> This could also be a chance to opportunistically update the model name in the documentation also to just plain 'gpt-3.5-turbo'

I think it's not necessary just for type safety~

lgrammel:

Issue #551, opened 2023-09-09
useChat hook call to `/api/chat` doesn't pass the full `Message` object. 


davit-b:
https://github.com/vercel/ai/blob/f81de390e142520e1d9fe64f4c3adb9dcf493102/packages/core/react/use-chat.ts#L93

If you want to use the `onStart` or `onCompletion` callbacks to save to a database, you do not have access to the additional fields of the `Message` type. Important fields such as `id` or `createdAt` which leads one to generating new guids for the messages on the server-side 

```ts
// Convert the response into a friendly text-stream
    const stream = OpenAIStream(response, {
      onStart: async () => {
        // This callback is called when the stream starts
        // You can use this to save the prompt to your database
        // await savePromptToDatabase(prompt)
      },
      onToken: async (token: string) => {
        // This callback is called for each token in the stream
        // You can use this to debug the stream or save the tokens to your database
        logger.debug(token)
      },
      onCompletion: async (completion: string) => {
        // This callback is called when the stream completes
        // You can use this to save the final completion to your database
        // await saveCompletionToDatabase(completion)

        // 6. Save messageSequence to DDB
        await saveInputAndCompletionToDDB(messages, completion)
      }
    })
```

The `Message` type is defined by the SDK itself
```ts
/**
 * Shared types between the API and UI packages.
 */
type Message = {
    id: string;
    createdAt?: Date;
    content: string;
    role: 'system' | 'user' | 'assistant' | 'function';
    /**
     * If the message has a role of `function`, the `name` field is the name of the function.
     * Otherwise, the name field should not be set.
     */
    name?: string;
    /**
     * If the assistant role makes a function call, the `function_call` field
     * contains the function call name and arguments. Otherwise, the field should
     * not be set.
     */
    function_call?: string | ChatCompletionMessage.FunctionCall;
};
```

but only the `content` and `role` are passed to the `/api/chat` route handler. If you think through the flow on client-side it goes like this 
1. (client) take inputs, generates ids for them, and put them into the `messages` object
2. (client) calls server for completion
3. (server) returns stream completion && saves the input, or completion to a database
4. (client) reads the stream, generates an id for the read string, and stores in the `messages` object

So in step # 3, if one stores to the database using different id keys - and then generates ids keys on the client-side as well - how can you reconcile them later without a pull from the server and rewriting id keys on messages?

ctate:
[...]

PR #619, opened 2023-10-03
Use `interface` instead of `type` for Message

venables:
This is a tiny change, but by using an interface rather than a type, developers can use declaration merging to modify this type to suit their needs. 

For example: 

```tsx
declare module 'ai' {
  interface Message {
    additionalProperty?: string
  }
}
```

This would satisfy requests like https://github.com/vercel/ai/issues/551#issuecomment-1744829091

### DIFF
--- a/.changeset/eleven-tools-develop.md
+++ b/.changeset/eleven-tools-develop.md
@@ -1,0 +1,5 @@
+---
+'ai': patch 
+---
+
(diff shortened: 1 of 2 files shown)


MaxLeiter (approved):

MaxLeiter:
Thanks!

Issue #621, opened 2023-10-05
onFinal only contains completion if onCompletion is defined


jpulec:
### Description

Best I can tell, if the `onFinal()` callback handler will only actually have the aggregated completion if you've also defined `onCompletion()`. Otherwise, messages don't get aggregated in `createCallbackTransformer` and the empty string is passed to `onFinal()`.

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes incorrect callback behavior: `onFinal()` should receive the aggregated completion independently, but currently depends on another optional callback being defined.

### Evidence
- The report states that `onFinal()` receives an empty string unless `onCompletion()` is also defined.
- It attributes the failure to `createCallbackTransformer` not aggregating messages when `onCompletion()` is absent.

PR #625, opened 2023-10-06
Simple fix for #621

peterdresslar:
This seems like the most basic possible fix to #621. I can't see what utility that condition had at all. 

I also am swapping places with the onToken callback, since it seems (very) faintly possible that we would want to *do* something with the token callback and have the aggregatedResponse reflect what is done up to that point.


MaxLeiter:
Shipping in #631 as I can't push to your main branch to add a changeset.

PR #629, opened 2023-10-09
Add Lepton AI as Providers

bobmayuze:
Hi Vercel team, we are Lepton AI, a team working on closing the gap between models and applications. We are fan of Vercel AI SDK and would like to help it grow as being both user and provider.

Please let us know if there is any other requirements.


lgrammel:
We have a new provider specification starting with the AI SDK 3.1. This seems outdated.

Issue #630, opened 2023-10-10
replicate-stream interface mismatch from latest Replicate SDK


peterdresslar:
### Description

Calls to create ReplicateStream from a Prediction generated using a [Replicate create method]() no longer pass type safety--at least using the latest Replicate library. At some point in the past several weeks, `updated_at` became optional on the Replicate interface (which, sure, makes sense.) The copy of the interface in `streams/replicate-stream` is thus out of sync.

### Additional context

`replicate` 0.20.1 or later

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This is a concrete compatibility regression in existing behavior: current Replicate SDK prediction objects fail the AI SDK's expected type interface.

### Evidence
- `ReplicateStream` no longer accepts a `Prediction` from Replicate SDK 0.20.1 or later under TypeScript type checking.
- The issue identifies an interface mismatch: Replicate made `updated_at` optional, while the copied interface in `streams/replicate-stream` remains out of sync.

PR #631, opened 2023-10-10
Support newer replicate versions, fix onResponse callback when no onCompletion is passed

### LINKED ISSUE #621: onFinal only contains completion if onCompletion is defined

peterdresslar:
Fix for #630 
Includes #625, so closes #621

### DIFF
--- a/.changeset/bright-fishes-cover.md
+++ b/.changeset/bright-fishes-cover.md
@@ -1,5 +1,0 @@
----
-'ai': patch
----
-
-fix: return complete response in onFinish when onCompletion isn't passed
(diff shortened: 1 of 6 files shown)


MaxLeiter (approved):

MaxLeiter:
Thanks again, @peterdresslar!

PR #652, opened 2023-10-20
Expansion to `ai/prompts` System with Base Types and Implementations

peterdresslar:
tl/dr:
This is an expansion to `ai/prompts` with the base types in `ai-prompts.ts` and four implementations (OpenAI, Replicate/LLama2, Anthropic, and HuggingFace/OASST).

### Overview
This PR expands ai/prompts by adding base types in ai-prompts.ts and implementing four new providers: OpenAI, Replicate/LLama2, Anthropic, and HuggingFace/OASST. The objective is to make it easy for consuming applications to engineer prompts and then pass their outputs to the respective provider's API calls. I also aimed to stay stylistically consistent with the rest of the SDK where possible.

### Example Uses

A basic example with OpenAI:

```typescript
 const myPrompt = OpenAIPrompt(messages);
 myPrompt.addMessage('You are a pirate. Speak like one please.', 'system');

  // Ask OpenAI for a streaming chat completion given the prompt
  // note that myPrompt is implemented to send back a matching ChatCompletionArrayParams[] in its return
    model: 'gpt-3.5-turbo',
    stream: true,
    messages: myPrompt.toPrompt()
  });
  ```
  
  Another example, using Anthropic SDK:

  ```typescript
  const myPrompt = AnthropicPrompt(messages);
  myPrompt.addMessage('Please review the prompt and follow the example for output.', 'system'); 
  myPrompt.addMessage('your response should look like what in <example></example> tags.', 'user', 'after');
  myPrompt.addMessage('<example>[...some example]</example>', 'user', 'after');

  const response = await anthropic.completions.create({
    prompt: myPrompt.toPrompt(),
    model: 'claude-2',
    stream: true,
    max_tokens_to_sample: 300,
  });
  ```
  ### Notes
  The approach taken in the Pull Request makes many decisions, but there are still a few that could be discussed. I left the existing typing for `ai/prompts` alone, and that may or may not be preferable. However, it does mean that the consuming application as of now has to import the implementations like this: 
  
  ```typescript
  import { AnthropicStream, StreamingTextResponse } from 'ai';
  import { AnthropicPrompt } from 'ai/prompts'; //currently a dist is being built in ai/prompts
  ```
  
  It might be worth considering whether that clunkiness is desirable. It would be straightforward to change (I think.)
  
  There is also a question of how to implement provider prompts with multiple models, but that would mostly just be a naming question for the implementations. It was the intent of this pull request to try and push all the nitty-gritty onto the implementations in order to try and keep the base approach clean as possible (given a wide array of current and future providers!)
  
 Will be happy to provide docs and tests if the maintainers decide to move this direction.
 
 Opened feature request #654 in case that solicits feedback.

### DIFF
--- a/.changeset/three-gifts-listen.md
+++ b/.changeset/three-gifts-listen.md
@@ -1,0 +1,5 @@
+---
+'ai': minor
+---
+
+Adding Prompt Construction with New Base Types and Functions
--- a/packages/core/prompts/ai-prompt.ts
+++ b/packages/core/prompts/ai-prompt.ts
@@ -1,0 +1,35 @@
+import type { SimpleMessage } from '../shared/types';
+
+// A simple interface for the AIInput object that is required for the Llama2 API
+export interface ReplicateAIInput {
+  prompt: string;
+  system_prompt?: string;
+}
+
+// this signature is expected by OpenAI's Node Library; we include it for convenience in calling the library
+export interface ChatCompletionMessageParam {
+  role: 'system' | 'user' | 'assistant' | 'function';
+  content: string | null;
+  function_call?: any;
+  name?: string;
+}
+
+export interface AIPrompt {
+  buildPrompt(messages: SimpleMessage[]): string;
+  addMessage(
+    content: string, // parameter ordering is due to role being optional
+    role?: 'user' | 'assistant' | 'system' | 'function',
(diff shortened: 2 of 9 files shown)


lgrammel:
`ai/prompts` has been deprecated

PR #662, opened 2023-10-22
Add sendHistory flag

achadha235:
Hi there, thanks for the awesome package 🙏🏽

**Problem:**
I was facing an issue while using the `useChat` hook. I noticed that the entire message history is transmitted each time a message is triggered. For systems where conversation management is handled server side, I realized this might cause unnecessary network traffic, especially as the conversations get long.

**Solution:**
I added a `sendHistory` flag to allow users to control whether the entire history is sent with each request

Let me know if this is a good idea or if I'm missing something. If the react version looks okay, I'll fix the svelte and vue files too. Thanks again!

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ai
 
+## 2.2.18
+
+### Patch Changes
+
+- use published flowise fork
+
 ## 2.2.17
 
 ### Patch Changes
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "ai",
-  "version": "2.2.17",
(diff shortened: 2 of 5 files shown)


cedarbaum:
I was thinking about a similar problem with large message histories. For my use case, I want to limit the conversation history to some fixed number (say the last 10 messages) and have the SDK manage that based on what it has in-memory on the client side. I think this would cover your case as well, since the message history length could be set to 1.

cedarbaum:
Something that might work for now for both our cases is using the `setMessages` helper function to manually truncate the list and then calling `reload` (see: https://sdk.vercel.ai/docs/api-reference/use-chat#chat-helpers). I haven't tested this yet, but even it does work, an option to manage history length could still be more intuitive IMO.

PR #667, opened 2023-10-24
Add characterization tests for StreamData.

lgrammel:
The tests verify the existing behavior, which is not necessarily the expected behavior. Since I’m just getting started, it is difficult for me to tell whether the behavior is correct or not. It might be worth looking at the expected output to see if there is anything incorrect.

## Tasks

- [x] Update tests to check chunks without JSON.stringify
- [x] Add test for function call result

### DIFF
--- a/packages/core/streams/index.test.ts
+++ b/packages/core/streams/index.test.ts
@@ -214,6 +214,40 @@ describe('AIStream', () => {
           ]);
         });
 
+        it('should send function response and data when onFunctionCall is defined, returns undefined, and data is added', async () => {
+          const data = new experimental_StreamData();
+
+          const stream = OpenAIStream(
+            await fetch(server.api + '/mock-func-call', {
+              headers: {
+                'x-mock-service': 'openai',
+                'x-mock-type': 'func_call',
+              },
+            }),
+            {
+              onFinal() {
+                data.close();
+              },
+              async experimental_onFunctionCall({ name }) {
+                data.append({ fn: name });
+
+                // no response
+              },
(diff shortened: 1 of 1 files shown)


packages/core/streams/index.test.ts
```diff
@@ -102,6 +102,121 @@ describe('AIStream', () => {
...
+          );
+
+          const response = new StreamingTextResponse(stream, {}, data);
+
+          const client = createClient(response);
+          const chunks = await client.readAll();
+
+          expect(chunks).toEqual([
+            '0:"Hello"\n',
+            '2:"[{\\"t1\\":\\"v1\\"}]"\n',
```
lgrammel, line 213: I was a bit surprised that the data message shows up on the 2nd position (instead of the 1st).

MaxLeiter: Yeah, I don't think this is expected. Great catch!

lgrammel: This might be the cause: https://github.com/vercel/ai/blob/main/packages/core/streams/stream-data.ts#L33

It seems that data is only appended when the Transformer receives a regular chunk from the wrapped stream, and appends the data after. This could also potentially lead to other issues (e.g. custom data is not send immediately, but only when chunks from the underlying stream are transformed).

MaxLeiter: After some review: this isn't a problem, as long as the order of messages with the same prefixes is kept. It can be "fixed" by moving the resulting line below the if statement

MaxLeiter (approved):

PR #671, opened 2023-10-24
Update README.md example code

PhatStraw:
added install step to the example for a smoother onboarding experience. You would get errors regarding openai prior.

Also added text to the frontend example code for clarity on next steps.


packages/core/README.md
```diff
@@ -32,7 +36,7 @@ const openai = new OpenAI({
 
 export const runtime = 'edge';
 
-export async function POST(req) {
+export default async function POST(req) {
```
MaxLeiter, line 39: With App Router you use named exports, not defaults

PhatStraw: ill fix that now

lgrammel:
README is very different now

PR #691, opened 2023-10-27
Add characterization tests for Langchain expression language.

lgrammel:


packages/core/streams/langchain-stream.test.ts
```diff
@@ -10,107 +29,150 @@ describe('LangchainStream', () => {
...
+              callbacks: [
+                {
+                  handleChainEnd(outputs, runId, parentRunId) {
+                    // check that main chain (without parent) is finished:
+                    if (parentRunId == null) {
+                      data.close();
+                    }
+                  },
+                },
+              ],
```
lgrammel, line 150: This is the main trick that's required for closing `data` correctly.

packages/core/streams/langchain-stream.test.ts
```diff
@@ -10,107 +29,150 @@ describe('LangchainStream', () => {
...
+            },
+          );
+
+        const response = new StreamingTextResponse(
+          stream.pipeThrough(createStreamDataTransformer(true)),
```
lgrammel, line 155: The langchain stream also needs to go through createStreamDataTransformer.

MaxLeiter: We should move this into langchain-stream

packages/core/streams/langchain-stream.test.ts
MaxLeiter: hm, not ideal but we can document this clearly. maybe we can handle it for the user post-experimental.

MaxLeiter (approved):

packages/core/streams/langchain-stream.test.ts
lgrammel: LangChain has two interfaces: the classic chains (what we support in `langchain-stream`) and the new langchain expression language (which is tested here and has a slightly different API). I see two options:

1) Introducing a separate LangChainExpressionLanguageStream
2) Introducing an options argument on LangChainStream for switching to LangChainExpressionLanguage support

Since the interface for working with the expression langauge is very different (see below), I'm leaning towards option 1). @MaxLeiter what are your thoughts?

### LangChain expression language right now
```ts

const model = new ChatOpenAI({
  //...
});

const stream = await PromptTemplate.fromTemplate('{input}')
  .pipe(model)
  .pipe(new BytesOutputParser())
  .stream(
    { input: 'Hello' },
    {
      callbacks: [
        {
          handleChainEnd(outputs, runId, parentRunId) {
            // check that main chain (without parent) is finished:
            if (parentRunId == null) {
              data.close();
            }
          },
        },
      ],
    },
  );

const response = new StreamingTextResponse(
  // stream comes from LangChain, we could e.g. wrap it
  stream.pipeThrough(createStreamDataTransformer(true)),
  {},
  data,
);
```

### LangChain classic right now
```ts

const model = new ChatOpenAI({
  // ...
});

const { stream, handlers } = LangChainStream({
  onFinal() {
    data.close();
  },
  experimental_streamData: true,
});

model
  .call([new HumanMessage('hello')], {}, [handlers])
  .catch(console.error);

const response = new StreamingTextResponse(
  // stream is synthetically created by the langchain callback handlers
  stream, 
  {}, 
  data
);
```

PR #702, opened 2023-10-30
Store initial messages (or empty array) in state to avoid re-rendering when using memo.

### LINKED ISSUE #615: useChat hook's returned helper `messages` returned as new empty array every time
### Description

Use the code example from the AI SDK for React using useChat.
The problem is simple: 
```ts
  const {
    messages
  } = useChat({...})
```

This hook instantiates an empty array for `messages=[]` every time the component where `useChat` is use is re-render.

Now imagine a 2-level component structure
```
<ParentComponent /> (useChat)
      <FirstChild  messages={messages} />
      <SecondChild /> 
```

If the **SecondChild** component's triggers a re-render of the **ParentComponent** for any reason, the `messages` object is re-instantiated as an empty array and this triggers the **FirstChild** component to re-render.

EVEN if you wrap FirstChild with `React.memo(...)` where the props in the 1st call are `messages=[]` and the 2nd call are `messages=[]`, it will trigger a render because the shallow comparison of props shows it as a new empty array.

#### The workaround (for anyone else running into this)
```ts
  const [guardedMessages, setGuardedMessages] = useState<...>([])
  
  useEffect(() => {
    if (!Array.isArray(messages) || !messages.length) return
    setGuardedMessages(messages)
  }, [messages])

// with a structure like
  <ParentComponent /> (useChat)
      <FirstChild  messages={guardedMessages} />
      <SecondChild /> 
```

After implementing the workaround, I don't get the <FirstChild> to re-render a million times whenever <SecondChild/> causes <ParentComponent> to trigger a re-render because of a small state update.

I really believe the SDK shouldn't return a new empty array each instantiation. Yes, as users of the SDK we can contort our parent-child hierarchy into a specific way to make it work.... but I really think when `messages=[]` it should return the same empty array rather than re-creating it over and over. 

Code at fault is: https://github.com/vercel/ai/blame/88451356c25b219a37946fb3b75da4f0d936a72c/packages/core/react/use-chat.ts#L623

Thoughts?

### Code example

_No response_

### Additional context

_No response_

lgrammel:
Resolves #615

Please let me know if `useState` is the preferred solution here. It means keeping state per hook, which may not be desired. However, `fallbackData` is SWR is already per hook from what I understand, so this may not be an issue.

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -280,6 +280,22 @@ const getStreamedResponse = async (
   }
 };
 
+const allMessages = new Map<Message[], number>();
+
+function traceMessages(trace: string, messages?: Message[]) {
+  if (messages === undefined) {
+    console.log(trace, 'undefined');
+    return;
+  }
+
+  if (!allMessages.has(messages)) {
+    allMessages.set(messages, allMessages.size);
+  }
+
+  const id = allMessages.get(messages);
+  console.log(trace, id, messages);
+}
+
 export function useChat({
   api = '/api/chat',
   id,
@@ -317,8 +333,10 @@ export function useChat({
   );
 
   // Keep the latest messages in a ref.
+  traceMessages('T1', messages);
   const messagesRef = useRef<Message[]>(messages || []);
   useEffect(() => {
+    traceMessages('T2', messages);
     messagesRef.current = messages || [];
   }, [messages]);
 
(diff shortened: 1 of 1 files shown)


shuding (commented):
Fallback data in SWR is already per hook, so it might be something else causing the issue.

shuding (approved):
Makes total sense, thank you!

Issue #724, opened 2023-11-08
2.2.21 broke initialMessages


odysseaspapadimas:
### Description

Updating from 2.2.20 to 2.2.21 broke the initialMessages property in the react useChat hook. The messages do not load.
As you can see from my code example I load the initial chat messages from my API which I checked and it returns the data just fine. Downgrading to 2.2.20 fixes it

### Code example

```jsx
const {
    messages,
    handleInputChange,
    isLoading,
    append,
  } = useChat({
    id: chatId,
    initialMessages: chat.data?.messages.map((message) => ({
      content: message.content,
      role: MessageRole(message.role),
      id: message.id,
    })),
});
```

lgrammel:
Since `v2.2.21`, `initialMessages` behaves similar to `initialInput` and uses `useState` https://github.com/vercel/ai/commit/0681baeac2027188f9bac3718ad64e7d194887f0 to address issue #615 

If you initialize it with `undefined` first (e.g. if `chat.data` is not yet available), then you cannot override it later. Changing the initial messages can also lead to strange behaviors once a chat has started and there are actual messages.

That being said, there needs to be a way to deal with `initialMessage` that are not immediately available.

@odysseaspapadimas Just to confirm, is `chat.data?.messages` available on the first invocation of `useChat` or does it change (e.g. by calling an API like you mentioned)?

odysseaspapadimas:
@lgrammel No, chat.data?.messages is undefined.

And yes I agree there needs to be a way to deal with this. Are there any solutions in the meantime other than downgrading?

mikesprague:
@lgrammel in my case, it (`initialMessages`) starts as an empty array and gets updated based on a value in local storage (but is never actually `undefined`). The value changes value from `[]` to an array with the saved messages data.

lgrammel:
@odysseaspapadimas @mikesprague A workaround for the time being would be to extract the component that contains `useChat` and only initalize it when `initialMessages` has been loaded (so it's correct the first time you call `useChat`)

greendesertsnow:
I was going to lose my mind after updating the packages...
Wrap the whole thing in a messageLoader component and pass the props down only if you've got the old messages from your database.
Example (edited it, so it may not work):
```js
export default function Chat(props: any) {
  const { selectedChat } = props;
  const { data: messagesFromDb } = useMessages(selectedChat);

  return (
    <div>
      {messagesFromDb && <MessageLoader messagesFromDb={messagesFromDb} />}
[...]

PR #727, opened 2023-11-09
Do not store initialMessages in useState.

lgrammel:
Users want to be able to change the initial messages ( #724 ). Storing `initialMessages` in a `useState` prevents this. It was introduced in https://github.com/vercel/ai/commit/0681baeac2027188f9bac3718ad64e7d194887f0 to prevent double rendering ( #615 ). 

Double rendering is only an issue when using an empty array that comes from a default parameter value, since it's recreated on each call. Therefore now only the empty array is stored in `useState`, and users can pass `initialMessages` like they used to.

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -283,7 +283,7 @@ const getStreamedResponse = async (
 export function useChat({
   api = '/api/chat',
   id,
-  initialMessages: initialMessagesParam,
+  initialMessages,
   initialInput = '',
   sendExtraMessageFields,
   experimental_onFunctionCall,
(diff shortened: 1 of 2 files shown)


packages/core/react/use-chat.ts
```diff
@@ -300,12 +300,14 @@ export function useChat({
   const hookId = useId();
   const chatId = id || hookId;
 
-  // Store initial messages as a state to avoid re-rendering when using memo:
-  const [initialMessages] = useState(initialMessagesParam ?? []);
+  // Store a empty array as the initial messages
+  // (instead of using a default parameter value that gets re-created each time)
+  // to avoid re-renders:
+  const [initialMessagesFallback] = useState([]);
```
lgrammel, line 306: @MaxLeiter are we expecting useChat to working in RSCs? asking bc of the `useState`

mosnicholas:
🙏

packages/core/react/use-chat.ts
MaxLeiter: No, it's purely a client-side hook.

MaxLeiter (approved):

MaxLeiter (approved):

PR #749, opened 2023-11-17
feat: (assistant) handle files, post a form instead of json

snettah:
image

### DIFF
--- a/examples/next-openai/app/assistant/page.tsx
+++ b/examples/next-openai/app/assistant/page.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { Message, experimental_useAssistant as useAssistant } from 'ai/react';
-import { useEffect, useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 const roleToColorMap: Record<Message['role'], string> = {
   system: 'red',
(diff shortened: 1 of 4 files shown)


examples/next-openai/app/assistant/page.tsx
```diff
@@ -11,18 +11,22 @@ const roleToColorMap: Record<Message['role'], string> = {
 };
 
 export default function Chat() {
-  const { status, messages, input, submitMessage, handleInputChange, error } =
```
snettah, line 14: I've removed `input` and `handleInputChange` for me the hook should not be responsible of that, what do you think ?

packages/core/react/use-assistant.ts
```diff
@@ -15,39 +15,26 @@ export function experimental_useAssistant({
...
-
-  const submitMessage = async (e: any) => {
+  // from form submit:
+  const submitMessage = async (e: React.FormEvent<HTMLFormElement>) => {
     e.preventDefault();
 
-    if (input === '') {
-      return;
-    }
+    const formData = new FormData(e.currentTarget);
```
snettah, line 26: I used a formData instead because I find it odd to use a form on the client side and send a json, but it's not a big deal.

PR #753, opened 2023-11-19
feat: add `isPending`

### LINKED ISSUE #752: Add new field `isPending` in useChat
### Feature Description

`isLoading` is not enough for some products since its lifecycle is during the whole text streaming.

As shown in the figure, I hope there is a new field called `isPending` to show that the server is processing the request and false once after the streaming has started.
image

### Use Case

my downstream PR that fakes a `isPending` field using `useDeferredValue.` 

https://github.com/run-llama/LlamaIndexTS/pull/203

```ts
const content = lastMessage?.content ?? "";
const deferredContent = useDeferredValue(content);
const isPending = props.isLoading ? !isLastMessageFromAssistant : (
  props.isLoading &&
  isLastMessageFromAssistant && content === deferredContent
);
```

### Additional context

_No response_

himself65:
Closes: https://github.com/vercel/ai/issues/752

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -224,6 +224,11 @@ export function useChat({
     null,
   );
 
+  const { data: isPending = false, mutate: mutatePending } = useSWR<boolean>(
+    [chatId, 'pending'],
+    null,
+  );
+
   const { data: streamData, mutate: mutateStreamData } = useSWR<
     JSONValue[] | undefined
   >([chatId, 'streamData'], null);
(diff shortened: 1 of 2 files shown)


packages/core/react/use-chat.ts
```diff
@@ -258,14 +263,16 @@ export function useChat({
...
         setError(undefined);
 
         const abortController = new AbortController();
         abortControllerRef.current = abortController;
 
         await processChatStream({
-          getStreamedResponse: () =>
-            getStreamedResponse(
+          getStreamedResponse: () => {
+            mutatePending(false);
```
lgrammel, line 274: `true`?

himself65: no, false

PR #762, opened 2023-11-21
Vue: Add complex response parsing and StreamData support to useChat

lgrammel:
## Summary

* Add `{ 'Content-Type': 'application/json' }` as default header in `callApi` (used for all frameworks)
* Add function call example to `nuxt/openai`
* Implement function calls & data streams for Vue


packages/core/shared/call-api.ts
```diff
@@ -34,7 +34,10 @@ export async function callApi({
       messages,
       ...body,
     }),
-    headers,
+    headers: {
+      'Content-Type': 'application/json',
+      ...headers,
+    },
```
shuding, line 40: Makes sense!

shuding (approved):

MaxLeiter (approved):

packages/core/shared/call-api.ts
cvbuelow: fyi, this caused a breaking change in my app since my server was expecting `text/plain`

PR #773, opened 2023-11-22
Use mock service worker in cohere-stream test.

lgrammel:
Using mock service worker will make the stream tests faster and independent, and will enable us to put provider-specific code into the test case. As a first step, I've converted the Cohere test.


lgrammel:
Node 18 failure could potentiall be related to https://github.com/mswjs/msw/issues/1868

MaxLeiter (approved):

PR #774, opened 2023-11-22
sagemaker stream

rolandknight:
adding sagemaker (with HuggingFace model) support


MaxLeiter:
Thanks for the effort here, @rolandknight! Right now, we don't feel demand is high enough to accept this PR, as our maintenance burden increases significantly for every provider. However, you can publish this as it's own npm package that can be used with the AI SDK -- does that work for you? We'll write docs for it soon, but you can see this example by @lgrammel for now: https://github.com/lgrammel/modelfusion/tree/main/extensions/vercel-ai

In the future, we'd like to maintain a list of 3rd party providers in our docs, so please let me know if you go through with this!

PR #786, opened 2023-11-25
Add turbo.json file to `core` package that caches all build artifacts

ankrgyl:
`turbo.json` in the root directory defines which `outputs` to cache. Currently, it only caches `dist/**`, but the `core` package generates nested `**/dist/**` directories in directories like `prompts`, `react`, `solid`, etc.

I manually verified this by running a clean `pnpm build` and then:

```
cd node_modules/.cache/turbo
find . -name "*.tar.zst" -exec tar -xf {} \;
ls packages/core
```

and observing that before this change, only the `dist` directory is cached, but after, the `dist`, `prompts`, `react`, `solid`, `svelte`, and `vue` directories are.

Without proper build caching, if this package is included as a dependency in a pnpm workspace, incremental building may attempt to restore the `ai` package from the build cache, only restore the root `dist` directory, and fail the build. There may be other issues too.


MaxLeiter (approved):
Thank you!

PR #797, opened 2023-11-28
Expose extra fetch options in useAssistant

SferaDev:
Similar to `useChat` allow setting `body`, `headers` and `credentials` from the hook call. Required to implement authentication for the assistants API endpoint.

### DIFF
--- a/packages/core/react/use-assistant.ts
+++ b/packages/core/react/use-assistant.ts
@@ -51,8 +51,8 @@ export function experimental_useAssistant({
   api,
   threadId: threadIdParam,
   credentials,
-  headers = {},
-  body = {},
+  headers,
+  body,
 }: UseAssistantOptions): UseAssistantHelpers {
   const [messages, setMessages] = useState<Message[]>([]);
   const [input, setInput] = useState('');
--- a/packages/core/react/use-assistant.ts
+++ b/packages/core/react/use-assistant.ts
@@ -39,13 +39,21 @@ export type UseAssistantHelpers = {
   error: undefined | unknown;
 };
 
+export type UseAssistantOptions = {
+  api: string;
+  threadId?: string | undefined;
+  credentials?: RequestCredentials;
+  headers?: Record<string, string> | Headers;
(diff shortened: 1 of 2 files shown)


packages/core/react/use-assistant.ts
```diff
@@ -83,8 +91,10 @@ export function experimental_useAssistant({
 
     const result = await fetch(api, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
+      credentials,
+      headers: { ...headers, 'Content-Type': 'application/json' },
```
lgrammel, line 95: Can you change the order (`...headers` later so it can override the content type)?

packages/core/react/use-assistant.ts
```diff
@@ -39,13 +39,21 @@ export type UseAssistantHelpers = {
...
 export function experimental_useAssistant({
   api,
   threadId: threadIdParam,
-}: {
-  api: string;
-  threadId?: string | undefined;
-}): UseAssistantHelpers {
+  credentials,
+  headers = {},
+  body = {},
```
lgrammel, line 55: `useChat` does not init those parameters. I'd prefer to match the style from `useChat` as much as possible.

packages/core/react/use-assistant.ts
SferaDev, line 95:
```suggestion
      headers: { 'Content-Type': 'application/json', ...headers },
```

packages/core/react/use-assistant.ts
lgrammel: Approved assuming this change comes in.

lgrammel (approved):

packages/core/react/use-assistant.ts
SferaDev, line 55:
```suggestion
  headers,
  body,
```

SferaDev: Updated

MaxLeiter (approved):

PR #798, opened 2023-11-28
Add data message support to useAssistant & assistantResponse.

lgrammel:
## Summary
Introduce a new data message that can contain arbitrary data and is shown in the list of messages. The data could also be processed and passed into other components (e.g. to show markers on a map).

**Note**: Adding a new message types can require changes to existing implementations


packages/core/shared/types.ts
```diff
@@ -262,3 +264,17 @@ export type AssistantMessage = {
...
+ * that should be shown in order with the other messages.
+ *
+ * It can trigger other operations on the frontend, such as annotating
+ * a map.
+ */
+export type DataMessage = {
+  id?: string; // optional id, implement if needed (e.g. for persistance)
+  role: 'data';
+  text?: string; // optional text to display in the UI
+  data: JSONValue; // application-specific data
```
shuding, line 279: Given the use case and semantics, I'd prefer the name `description?: string` over `text?: string`.

shuding (commented):
Code looks good!

MaxLeiter (approved):

PR #804, opened 2023-11-29
Handling returned Promise<string> in experimental_onFunctionCall

reconcrete:
If user provided Promise<string> in experimental_onFunctionCall, put it in onCompletion and onFinal calls


arhamHumayun (approved):

lgrammel:
`experimental_onFunctionCall` has been deprecated and will be removed.

PR #825, opened 2023-12-04
Introduce experimental_buildOpenAiMessages

lgrammel:


packages/core/prompts/openai.tsx
```diff
@@ -0,0 +1,241 @@
...
+   */
+  tool_calls?: Array<ChatCompletionMessageToolCall>;
+}
+
+export namespace ChatCompletionAssistantMessageParam {
+  /**
+   * Deprecated and replaced by `tool_calls`. The name and arguments of a function
+   * that should be called, as generated by the model.
+   */
+  export interface FunctionCall {
```
MaxLeiter, line 156: Some of these are duplicates of `core/shared/types.ts`. I'd prefer not maintaining two copies of them, maybe these should be moved there, or a new openai types file?

lgrammel: A new `openai` types file sounds good. I would ideally want to clearly separate the AI SDK message types from the OpenAI message types to prevent confusion and avoid errors in future changes (when they might evolve indepedently).

lgrammel: @MaxLeiter actually it looks like it's only `FunctionCall` that's fully overlapping, and I'd argue that the use in `types.ts` is as part of AI SDK messages, and here it is used as an OpenAI message. The evolution forces for both of them are different (here: OpenAI changes their API, in `types.ts`: we decide to change AI SDK messages), so I'd rather keep them separate and remove the comment in `types.ts`. What do you think?

MaxLeiter: Yeah, this makes sense. We'll need to tackle the type story more definitively in the future, but for now let's ship both and keep OpenAI vs AI SDK separate.

MaxLeiter (approved):

PR #828, opened 2023-12-05
Ignore api parameter in useChat/React when it's not a string

lgrammel:


packages/core/react/use-chat.ts
```diff
@@ -205,30 +206,37 @@ export function useChat({
...
   // (instead of using a default parameter value that gets re-created each time)
   // to avoid re-renders:
   const [initialMessagesFallback] = useState([]);
 
   // Store the chat state in SWR, using the chatId as the key to share states.
-  const { data: messages, mutate } = useSWR<Message[]>([api, chatId], null, {
-    fallbackData: initialMessages ?? initialMessagesFallback,
-  });
+  const { data: messages, mutate } = useSWR<Message[]>(
+    [...chatKey, 'messages'],
```
shuding, line 222: No need to spread, SWR uses deep comparison to track key changes:

```suggestion
    [chatKey, 'messages'],
```

shuding (commented):
Why do we need to introduce `key`? I think `id` is for this exact purpose (to manually set the SWR key).

lgrammel:
> Why do we need to introduce `key`? I think `id` is for this exact purpose (to manually set the SWR key).

My thinking was that the user might want to distinguish the chat (thread) id from the type of chat, but you're right. Maybe the easiest thing would be to just cahnge the default in case `api` is not a string? I'll create an alternative proposal

shuding:
Yea originally, we thought that by default one `useChat` === one unique thread (hence it defaults to `useId`). But if providing the same id to 2 `useChat` hooks, we treat them as the same thread.

> Maybe the easiest thing would be to just change the default in case api is not a string?

Yeah, that makes sense - users will have to provide the `id` if they want to share the chat state across components, anyways.

shuding (approved):

PR #847, opened 2023-12-11
[Draft] Mistral API docs using OpenAI client.

lgrammel:
> [!WARNING]
> I was not able to try out the code in the examples, bc I do not have Mistral access yet, so whether this works depends on the true compatibility of the Mistral API with the OpenAI API.

### DIFF
--- a/docs/pages/docs/guides/providers/mistral.mdx
+++ b/docs/pages/docs/guides/providers/mistral.mdx
@@ -41,7 +41,7 @@ MISTRAL_API_KEY=xxxxxxxxx
 
 ### Create a Route Handler
 
-Create a Next.js Route Handler that uses the Edge Runtime that we'll use to generate a chat completion via OpenAI that we'll then stream back to our Next.js.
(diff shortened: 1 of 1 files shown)


docs/pages/docs/guides/providers/mistral.mdx
```diff
@@ -0,0 +1,261 @@
...
+---
+
+import { Steps, Callout } from 'nextra-theme-docs';
+
+# Mistral
+
+Vercel AI SDK provides a set of utilities to make it easy to use [Mistral](https://mistral.ai/)'s APIs and models. In this guide, we'll walk through how to use the utilities to create a chat bot and a text completion app.
+
+<Callout>
+  Mistrals REST APIs [are compatible with OpenAI's so we will use OpenAI's
```
jaredpalmer, line 12:
```suggestion
  Mistral's REST APIs [are compatible with OpenAI's so we will use OpenAI's
```

PR #851, opened 2023-12-12
onToolCall for useChat / React

lgrammel:
## Summary
* implement `experimental_onToolCall` in useChat / React
* add next.js / openai example for tool calling


examples/next-openai/app/tool-calling/page.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+import { ChatRequest, ToolCallHandler, nanoid } from 'ai';
+import { Message, useChat } from 'ai/react';
+
+export default function Chat() {
+  const toolCallHandler: ToolCallHandler = async (chatMessages, toolCalls) => {
+    let handledFunction = false;
+    for (const tool of toolCalls) {
+      if (tool.type === 'function') {
+        const result = handleFunction(tool.function);
```
MaxLeiter, line 11: Is there a possibility I don't want the function to return anything and just perform some side-effect? We may need to set handledFunction / push a message when we call `handleFunction` regardless of the return value. Maybe we recommend try/catch and throwing an error instead of returning one.

MaxLeiter (changes requested):

examples/next-openai/app/tool-calling/page.tsx
lgrammel: `handleFunction` is client code (see below) - not sure I fully follow. Users should be able to do what they want. I'll see if I can simplify the code (copied from Svelte for consistency).

ozzyonfire: We're patiently waiting for this fix to be released. It seems like the only thing holding it up is an example implementation? Can we get this merged so we can start using the useChat react hook with tools?

MaxLeiter (approved):

rajdtta:
It looks like parallel tool calling can lead to tool calls being streamed to the assistant response rather than being called as a separate tool message/response.

img

(Message has the role `assistant` and contains the stringified tool call JSON object)

ybentz:
@rajdtta we've been seeing the same thing happen from time to time when using functions (haven't migrated to tools yet). We thought switching to tools will fix it because we thought openai is trying to return text and call a tool at the same time but functions don't support that so it gets confused but I guess that's not the case which is concerning..

rajdtta:
> @rajdtta we've been seeing the same thing happen from time to time when using functions (haven't migrated to tools yet). We thought switching to tools will fix it because we thought openai is trying to return text and call a tool at the same time but functions don't support that so it gets confused but I guess that's not the case which is concerning..

Yeah, it's proving to be a pretty big issue when combined w/ parallel tool calls that require client + server action. I detailed a really hacky workaround in #734, but that doesn't solve the issue 100% of the time (it only prevents it from happening 100% of the time).

rajdtta:
> @rajdtta we've been seeing the same thing happen from time to time when using functions (haven't migrated to tools yet). We thought switching to tools will fix it because we thought openai is trying to return text and call a tool at the same time but functions don't support that so it gets confused but I guess that's not the case which is concerning..

Have you found any solution/workaround to this?

rajdtta:
> > @rajdtta we've been seeing the same thing happen from time to time when using functions (haven't migrated to tools yet). We thought switching to tools will fix it because we thought openai is trying to return text and call a tool at the same time but functions don't support that so it gets confused but I guess that's not the case which is concerning..
> 
> Have you found any solution/workaround to this?

@ybentz just wanted to let you know that I opened up a PR that should hopefully fix this: #1134

PR #859, opened 2023-12-13
Align the chat request options for useChat across all frameworks

lgrammel:
> [!WARNING]
> This change breaks cases where the existing options are used in `useChat` for `vue` and `solid`.

## Summary
* add chat request options to useChat / solid
* add chat request options to useChat / vue


packages/core/shared/types.ts
```diff
@@ -145,9 +145,7 @@ export type RequestOptions = {
 
 export type ChatRequestOptions = {
   options?: RequestOptions;
-  // @deprecated
   functions?: Array<Function>;
-  // @deprecated
```
lgrammel, line 150: Cleanup some of the tools changes. Do we want to deprecate functions?

MaxLeiter (approved):

PR #861, opened 2023-12-14
Add Inkeep as a provider

nick-inkeep:
Adds [Inkeep](https://inkeep.com/) as a chat provider. Stream, test, and `nextjs-inkeep` folder is consistent with other LLM providers, like Anthropic

Has dependency on [this](https://github.com/vercel/ai/pull/879) PR for message annotations, which are used by the Inkeep stream provider to show the sources used by an LLM in a RAG response.

[Here's](https://www.loom.com/share/8afc1e535b5e4cf9a8923ee7b959eaf0?sid=f5048f21-269e-4443-80da-6d717fed82e4) a video showing the Inkeep stream provider and `message_annotations`.


lgrammel:
It looks good in general. I'm wondering if you are planning to release a client library? Some of the code from the examples such as continueChatSession could be a good candidate for that library.

The one part that I'm concerned about is exporting the Inkeep types from the AI SDK. This will create a dependency which means that we need to update the AI SDK with newer Inkeep types. Ideally this would also be part of an Inkeep client library.

MaxLeiter (approved):

PR #863, opened 2023-12-14
[WIP] Google Gemini support using Vertex AI library

lgrammel:
> [!NOTE]
> The Google Vertex library does not work on the edge

## Summary
* add `GoogleGeminiStream`
* add Google Gemini example


merwanehamadi:
I copied the code some of the code and it's seems to be working, thanks

AWKohler:
why was this canceled?

Godrules500:
I see that this is closed. What are the plans for pulling vertex GoogleVertexAIStream into master so that we can begin using it?

PR #879, opened 2023-12-21
feat: Adding ability to attach 'annotations' (custom metadata) to messages

nick-inkeep:
## Scenario
Many developers would like to add domain or provider-specific information to a given message. For example, for [Inkeep](https://github.com/vercel/ai/pull/861) we stream information about the sources used to answer a question at the end of the event stream. This might be used by a client to, for example, show a "Sources" section at the end of a message in the UI:

image

**Dependency for:**
https://github.com/vercel/ai/pull/861 (adding Inkeep as a provider)

**Resolves:**
https://github.com/vercel/ai/discussions/585
https://github.com/vercel/ai/pull/425#issuecomment-1688007798

## Implementation
- add a new 'message_annotations' stream-part
- add `appendMessageAnnotation(n)` function to `data` (experimental_StreamData) object
- pass the `event` type to a CustomParser so streams can handle non-homogeneous server-side event. For example, so that the Inkeep stream provider can parse the `onRecordsCited` event type that's unique to it's SSE and has the information to append to a message.
- accessible in the client as `message.annotations[n]`


lgrammel:
@nick-inkeep this looks great. To help me understand it better, could you provide an example of how to use it, or ideally add it to the `next-openai` examples?

Updates to the documentation and a test case for the full streaming (e.g. add to `packages/core/streams/openai-stream.test.ts`) would be super helpful.

packages/core/streams/ai-stream.ts
```diff
@@ -50,7 +50,7 @@ export interface AIStreamCallbacksAndOptions {
  * @interface
  */
 export interface AIStreamParser {
-  (data: string): string | void;
+  (data: string, event?: string): string | void;
```
lgrammel, line 53: Can you use an options array with an `event` property as the 2nd parameter? This would leave more flexibility for future extensions without breaking backward compatibility.

lgrammel (approved):
Looks great, had just one small comment.

packages/core/streams/ai-stream.ts
MaxLeiter: I prefer the object 👍

shuding (approved):

mohamedcherifmo:
I see the annotations array repeats the same annotation 100s of times... I created an issue around that

PR #912, opened 2024-01-11
Azure OpenAI support

### LINKED ISSUE #618: First class support for @azure/openai library
### Feature Description

openai-node documents specifically call out the Azure OpenAI api as having subtly different behaviors compared to the OpenAI api, and recommend using the @azure/openai library instead. See: https://github.com/openai/openai-node#azure-openai

>[Azure OpenAI](https://github.com/openai/openai-node#azure-openai)
>An example of using this library with Azure OpenAI can be found [here](https://github.com/openai/openai-node/blob/master/examples/azure.ts).

>Please note there are subtle differences in API shape & behavior between the Azure OpenAI API and the OpenAI API, so using this library with Azure OpenAI may result in incorrect types, which can lead to bugs.

>See [@azure/openai](https://www.npmjs.com/package/@azure/openai) for an Azure-specific SDK provided by Microsoft.

The @azure/openai SDK returns an `AsyncIterable<ChatCompletions>` object, which can't be used to create an OpenAIStream.

Here's how I use it:

```   
const events = client.listChatCompletions(
      deploymentId,
      tunedMessagePrompt,
      { maxTokens: 128, stream: true, temperature: 0.3 }
    )
```

### Use Case

White technically you can use the openai package, it's not recommended and given that the APIs are likely to continue to diverge I'd rather not be troubleshooting the difficult kind of bugs this is likely to introduce.

### Additional context

_No response_

lgrammel:
## Summary
* Added Azure support to `OpenAIStream`
* Added test case
* Added Azure docs

Closes #618 and #945


lgrammel:
Unclear where the prettier error is coming from. `pnpm prettier-check` shows nothing on my machine, and `pnpm prettier-fix` does not change anything.

docs/pages/docs/getting-started.mdx
```diff
@@ -104,7 +104,7 @@ We've written some code to get you started — follow the instructions below to
 
       // Ask OpenAI for a streaming chat completion given the prompt
       const response = await openai.chat.completions.create({
-        model: 'gpt-3.5-turbo',     
+        model: 'gpt-3.5-turbo',
```
lgrammel, line 107: Prettier fix (broken on main)

MaxLeiter (approved):

PR #932, opened 2024-01-20
Feat:add helper to clear returned errors

untilhamza:
## Why this PR is needed.
- While using this library, I ran into a situation in which an error was returned from the backend. This library would then return the error to me using the error attribute of the chat helpers.
- However, I had no way of clearing this error to make the current chat + chat messages usable again on the UI.
- It would be helpful if the consumers had a way to clear the error from the state and keep using the chat. This can happen in situations like if the user had sent a message that was beyond the allowed context. We could easily clear the error, notify them to fix this, and then continue the chat

Please take a look @lgrammel , @MaxLeiter:)

### DIFF
--- a/packages/core/react/use-assistant.ts
+++ b/packages/core/react/use-assistant.ts
@@ -1,6 +1,6 @@
 /* eslint-disable react-hooks/rules-of-hooks */
 
-import { useState } from 'react';
+import { useState, useCallback } from 'react';
 import { readDataStream } from '../shared/read-data-stream';
 import { Message } from '../shared/types';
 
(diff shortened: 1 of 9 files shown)


lgrammel:
Looks good! @MaxLeiter what do you think?

MaxLeiter:
I think its more reacty and in the spirit of SWR for the user to have their own error state, and they can watch `ai`s error state in an effect. 

```
const [error, setError] = useState()

useEffect(() => { if (errorFromAiSdk) setError(errorFromAiSdk) else setError(null) }
```

untilhamza:
Thanks @MaxLeiter, I think you are right because I somehow ended up doing that. I thought it was unnecessary but now I understand it was intended.

Issue #952, opened 2024-01-31
message annotations are repeated


mohamedcherifmo:
### Description

I'm combining appendMessageAnnotation with function calling where I'm storing the result of the function call into the message annotations as follow:
`
 data.appendMessageAnnotation({
    messageID: messageID,
    result: top_stories.map((story: any) => ({
      id: story.id,
      title: story.title,
      description: story.subtitle || story.description,
      image_url: story.image_url || story.image,
      category: story.type,
      url: story.url,
    }))
  })
`

When i console.log the message.annotations from the client side, the annotations array has 294 items. It should only have one

mohamedcherifmo:
I figured out that when adding it to the onFinal it works as intended however when added as a result of a function call, it repeats the annotations with every response. I tried onStart (to have it stream at the start instead of at the end, but still repeated)

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes incorrect repetition of message annotations in an existing API workflow, and the subsequent fix confirms it was a product defect.

### Evidence
- Calling `appendMessageAnnotation` as the result of a function call produces 294 repeated annotation items instead of one.
- The reporter clarified that annotations repeat with every streamed response when appended during a function call, while `onFinal` behaves correctly.
- A maintainer proposed a fix and later stated that the issue was resolved in the latest version.

PR #959, opened 2024-02-05
fix: Message annotations are compatible for any message type

nick-inkeep:
Few updates/fixes:
- resetting `self.messageAnnotations` after stream-part is encoded. Was causing issue when `appendMessageAnnotation` was used outside of `onFinal`. Was the root cause of https://github.com/vercel/ai/issues/952.
- made it so that annotations can be added to any message type, including `function_call`, not just `text`
- added test cases to `parse-complex-response.test.ts` to account for the above
- added example usage to the `chat-with-functions` page/route for the openai example


.changeset/stupid-news-greet.md
```diff
@@ -0,0 +1,6 @@
+---
+'next-openai': patch
```
MaxLeiter, line 2:
```suggestion
```

MaxLeiter: We only use changeset for the AI package, all the others are unpublished/unversioned

packages/core/shared/parse-complex-response.ts
```diff
@@ -15,20 +15,12 @@ type PrefixMap = {
...
-  return {
-    id: generateId(),
-    role: 'assistant',
-    ...rest,
-  };
+function assignAnnotationsToMessage<T extends Message | null | undefined>(
+  message: T,
+  annotations: JSONValue[] | undefined,
+): T {
+  if (!(message && annotations && annotations.length)) return message;
```
MaxLeiter, line 22: small nit, I prefer the expanded version

```suggestion
  if (!message || !annotations || !annotations.length) return message;
```

MaxLeiter (approved):

MaxLeiter:
thank you!

PR #961, opened 2024-02-06
Fix onFinal callback argument when a string is returned from tool call.

JanMatas:
There is an inconsistent behaviour of `onFinal` callback when different types are returned from `experimental_onToolCall`.

When `experimental_onToolCall` callback returns a stream, `onFinal` gets called with a complete response of the returned stream. However, when `experimental_onToolCall` returns a string, `onFinal` callback is called with the tool call instead of the returned string. 

This PR makes the behaviour consistent - `onFinal` is called with the string returned from `experimental_onToolCall`.


lgrammel:
Looks good. I'll add a test case once it's landed.

lgrammel (approved):

PR #969, opened 2024-02-10
Fixes toolCallStreamPart in stream-parts.ts

jonschlinkert:
The `.some()` check was not returning a value. This PR fixes that.


jonschlinkert:
fwiw, I did actually run the prettier check before doing the PR. Here is what I get when running `npm run prettier-check`: 

image

Not sure what prettier is choking on, but I'd be happy to fix it if I can get some direction on what to fix.

MaxLeiter:
I can't access your branch to push the change, but running `pnpm prettier-{fix,check}` found the error:

```diff
-        tc => 
+        tc =>
```

(there's a space after the `=>`)

lgrammel:
@jonschlinkert the AI SDK uses pnpm - `pnpm prettier-fix` might fix the issue.

PR #984, opened 2024-02-20
Inkeep docs touchups

nick-inkeep:
Updating a few minor lines in the Inkeep example, documentation, and it's order in the docs side bar.

Additionally, it seems there was a merge issue in `streams/index.ts` previously that duplicated certain import/exports. Removing those duplicates. Doesn't functionally change anything, so not sure if a bumping the version of `core` is needed.


packages/core/streams/index.ts
```diff
@@ -1,13 +1,3 @@
-export * from './ai-stream';
-export * from './aws-bedrock-stream';
-export * from './openai-stream';
-export * from './streaming-text-response';
-export * from './huggingface-stream';
-export * from './cohere-stream';
-export * from './anthropic-stream';
-export * from './inkeep-stream';
-export * from './langchain-stream';
-export * from './replicate-stream';
```
lgrammel, line 10: Is this on purpose? Don't we need to export the streams?

lgrammel (approved):

Issue #983, opened 2024-02-19
Allow to return tool call results from server


ichitaka:
### Feature Description

Currently the next-openai-app example for tool calling has two different types of execution:
- Handle tool call on client side, where the actual tool call output is added to the messages on the client side
- Handle tool call on server side, where the actual tool call output (and the original tool call) gets dropped and only the assistants response are streamed back to the client

I think it would be of great value to allow the server to send the execution and results to the client.

### Use Case

If I want to show tool calls to the user. This could be beneficial if the user knows about the tools and wants to check function outputs. ChatGPTs UI shows output of tool calls.

lgrammel:
Please check out our new tool calling with AI SDK Core and useChat: 
* https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling
* https://sdk.vercel.ai/docs/ai-sdk-ui/chatbot-with-tool-calling

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This is a request for an API capability to expose server-executed tool calls and results to clients, rather than a concrete failure of documented behavior.

### Evidence
- The issue requests the ability to send server-side tool executions and their results to the client so they can be displayed.
- The maintainer points to newly introduced AI SDK Core and useChat tool-calling functionality.

PR #991, opened 2024-02-22
Executed tool calls handling

ichitaka:
My own implementation to cover the lack of #983. Basically, being able to receive the results of an executed tool_call from the server, so that the executed_tool call can stay in the chat history. Would be great if we can get this to merged.


rajdtta:
Would love to see this implemented

ichitaka:
I'm not sure if this will be merged now, since the Generative UI in theory covers most of my use cases. My biggest issue is that Generative UI does not allow a Python backend. Would appreciate any feedback here.

PR #1036, opened 2024-03-02
fix(useActions): return typed object

IdoPesok:
`useActions` is not inferring the AIActions type correctly since it seems like the Actions generic isn't being used in `AIProvider`. If we just pipe it to `AIProviderProps` it is now being used and can be inferred.

Before:

Screen Shot 2024-03-01 at 4 19 45 PM

After:

Screen Shot 2024-03-01 at 4 36 56 PM

### DIFF
--- a/packages/core/rsc/types.ts
+++ b/packages/core/rsc/types.ts
@@ -27,14 +27,14 @@ export type InternalAIProviderProps<AIState = any, UIState = any> = {
   wrappedSyncUIState?: ServerWrappedAction<AIState>;
 };
 
-export type AIProviderProps<AIState = any, UIState = any> = {
+export type AIProviderProps<AIState = any, UIState = any, Actions = any> = {
   children: React.ReactNode;
   initialAIState?: AIState;
   initialUIState?: UIState;
 };
 
 export type AIProvider<AIState = any, UIState = any, Actions = any> = (
-  props: AIProviderProps<AIState, UIState>,
+  props: AIProviderProps<AIState, UIState, Actions>,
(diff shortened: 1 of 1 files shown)


shuding (approved):
Thank you!

IdoPesok:
Hey guys, nevermind 😞  I thought it fixed it but it turns out to have not done much. Uploading screen recording below of how I tested it and why I thought the change worked. You will notice the types align once we pass the `Actions` one type down. It is working when I explicitly defined the type as well as when I used createAction in the `packages/core` folder.

https://github.com/vercel/ai/assets/43461847/13b68c41-e4fa-4e33-9cb6-a76ac532d69d

Screen Shot 2024-03-02 at 10 06 09 AM

Unfortunately, it seems to still not be inferring correctly after the types get built and when I try to use `useActions` in the example with version 3.0.2.

The good news is it didn't break anything that wasn't broken before. So we are still at square one. Will keep thinking about how to solve this without making any breaking changes. But it seems like TS really wants `Actions` to be used somewhere to be able to infer it correctly.

IdoPesok:
I'm thinking of doing something like 

```typescript
...
initialAIState?: AIState;
initialUIState?: UIState;

// Introduce a dummy handler for the sole purpose of type inference.
// This property does not need to be used; it's purely for TypeScript's benefit.
__dummyActionsHandler?: (actions: Actions) => void;
```

which I think is obvious from a user perspective to not use it but not sure since it's super janky... @shuding any ideas of where to use `Actions` ? Happy to help implement.

PR #1044, opened 2024-03-02
fix: wrong Apple ticker → APPL to AAPL

jordienr:


shuding:
Thanks <picture data-single-emoji=":smile_simply:" title=":smile_simply:">:smile_simply:</picture>

PR #1053, opened 2024-03-02
default for renderer.options.text

timreinke:


timreinke:
oops didn't mean to open this against this repo 😅

PR #1065, opened 2024-03-03
fix(render): parse the args based on the zod schema

IdoPesok:
Added zod parse in between the LLM output and calling the `render` function. This enables zod functions such as `.default` and `.transform`

For example, if you have a parameter `meetingDuration` that is z.number().default(30), the LLM can output undefined but the args will still get the value 30 after doing `z.safeParse`. Right now it seems like it would be undefined which is not the user expected value.

### DIFF
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -182,14 +182,14 @@ export function render<
     [name in keyof TS]: {
       description?: string;
       parameters: TS[name];
-      render: Renderer<z.infer<TS[name]>>;
+      render: Renderer<z.output<TS[name]>>;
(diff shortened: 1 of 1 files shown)


shuding (approved):
Sounds great! I think we might need 2 extra options in the future that:
1. If we should render an error component when that happens;
2. Otherwise, should it retry or just throw

That might need some more discussions.

IdoPesok:
1) Yeah an `errorFallback` component alongside the render sounds great!
2) Ability to set `numRetries`, we can pass the zod error back to the LLM and hope it fixes it. I've found the Zod errors to be very good context for LLM to fix its output.

PR #1075, opened 2024-03-03
fix(types): make sure Actions type is bound to prevent `InferActions` returning unknown

juliusmarminge:
Previously the InferActions didn't work, resulting in `useActions<typeof AI>()` returning `unknown`. I identified the cause for this was that you're not binding the Actions generic to anything so it gets lost:

CleanShot 2024-03-04 at 00 04 28

I added a prop to bind the generic so we get proper type inference. We use the `$Name` naming convention in tRPC to bind types, not sure if you got other preferences there

https://github.com/vercel/ai/assets/51714798/8c44eebd-7f96-4fc6-a45d-8744049d329b

### DIFF
--- a/packages/core/rsc/types.ts
+++ b/packages/core/rsc/types.ts
@@ -31,7 +31,7 @@ export type AIProviderProps<AIState = any, UIState = any, Actions = any> = {
   children: React.ReactNode;
   initialAIState?: AIState;
   initialUIState?: UIState;
-  __actions?: Actions;
+  $ActionTypes?: Actions;
 };
 
 export type AIProvider<AIState = any, UIState = any, Actions = any> = (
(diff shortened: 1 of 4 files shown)


packages/core/rsc/types.ts
```diff
@@ -31,6 +31,7 @@ export type AIProviderProps<AIState = any, UIState = any, Actions = any> = {
   children: React.ReactNode;
   initialAIState?: AIState;
   initialUIState?: UIState;
+  $ActionTypes?: Actions;
```
juliusmarminge, line 34: without this the generic doesn't get bound to anything and the information is lost

shuding: This is kinda sad that TS requires us to do this but I don’t see other easy workarounds. To hide it away for users, could you make the key a Symbol instead so no one will actually pass that (and hopefully autocompletion ignores it too)?

juliusmarminge: Symbol doesnt seem to work :/ getting unknown return from useActions again...

MaxLeiter:
Thanks @juliusmarminge! I added a jsdoc comment and will merge/release once tests pass.

MaxLeiter (approved):

maik-pedroza:
On version 3.2.3 and still not working. Any updates?

PR #1080, opened 2024-03-04
Revert #1065

MaxLeiter:
```
Error calling experimental_onToolCall: [Error: Invalid function call arguments in message. [
  {
    "code": "invalid_type",
    "expected": "object",
    "received": "string",
    "path": [],
    "message": "Expected object, received string"
  }
]]

```

cc @IdoPesok

### DIFF
--- a/.changeset/khaki-starfishes-fold.md
+++ b/.changeset/khaki-starfishes-fold.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Revert "fix(render): parse the args based on the zod schema"
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -182,14 +182,14 @@ export function render<
     [name in keyof TS]: {
       description?: string;
       parameters: TS[name];
-      render: Renderer<z.output<TS[name]>>;
+      render: Renderer<z.infer<TS[name]>>;
     };
   };
   functions?: {
     [name in keyof FS]: {
       description?: string;
       parameters: FS[name];
-      render: Renderer<z.output<FS[name]>>;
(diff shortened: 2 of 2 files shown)


shuding (approved):

IdoPesok:
@MaxLeiter 

Looking into it, is this line missing a JSON.parse() ? Because the ToolCallPayload type shows arguments as a record but seems like they never get parsed from string.

https://github.com/vercel/ai/blob/aab5324cb31514b8e95e9d722859651248f8497f/packages/core/streams/openai-stream.ts#L608

IdoPesok:
As per the OpenAI API docs, there should be a `JSON.parse(tool.function.arguments)` I actually ran into this in a different project I was working on so I did a ` if typeof arguments === "string" ? JSON.parse(...) : ...` which should be safe. The problem is this could be a breaking change for many codebases if they have already implemented the parse themselves because then it would be a `JSON.parse(JSON.parse(...))`.

https://platform.openai.com/docs/guides/function-calling

mharris717:
Same issue as Ido. Real bummer, as it means the main docs page example for flagship feature Generative AI straight up doesn’t work. The example only doesn’t crash cause it doesn’t care that flight number is passed as undefined.

IdoPesok:
In #1084 I updated it so tool calls have their arguments go through `JSON.parse` then through the Zod parse. This should fix everything. Tool arguments will be able to be parsed by zod, and then the args send to the `render` function will be properly transformed according to the zod schema (enabling `.default`, `.transform`, etc)

nicholasraphael:
this is happening again followed the example from here exactly https://sdk.vercel.ai/docs/concepts/ai-rsc @MaxLeiter @shuding

PR #1084, opened 2024-03-04
Parse function call args

IdoPesok:
#1065 take 2

It seems the `ToolCallPayload` was improperly typed as `{ arguments: Record<string, unknown> }` when it should have extended `{ arguments: string }`.

Per OpenAI docs, you must JSON.parse the tool -> function -> arguments. https://platform.openai.com/docs/guides/function-calling

In this PR, I chose to just fix the type and not parse it in the `OpenAIStream` since this would be a breaking change for many codebases that are already probably doing a `JSON.parse` in their own code without a conditional string check. The bad side is now this diverges from the function call types (which do get parsed in the `OpenAIStream`)

---------

To clarify -- to actually use this handler right using the current SDK version, you need to do:

```typescript
tool.func.arguments = JSON.parse(tool.func.arguments)
```

or in a more safe way

```typescript
tool.func.arguments = typeof tool.func.arguments === "string" ? JSON.parse(tool.func.arguments) : tool.func.arguments
```

The problem is if we change it now to do the parsing in `OpenAIStream`, it will break the code of those from case 1, where they parse in an unsafe way. Even worse, the types won't signal a breakage since it has been incorrectly typed as a Record so it would appear that nothing has changed.

### DIFF
--- a/packages/core/streams/ai-stream.ts
+++ b/packages/core/streams/ai-stream.ts
@@ -16,7 +16,7 @@ export interface ToolCallPayload {
     type: 'function';
     func: {
       name: string;
-      arguments: Record<string, unknown>;
+      arguments: string;
     };
   }[];
 }
--- a/examples/next-openai/app/api/chat-with-tools/route.ts
+++ b/examples/next-openai/app/api/chat-with-tools/route.ts
@@ -86,7 +86,10 @@ export async function POST(req: Request) {
           // Call a weather API here
           const weatherData = {
             temperature: 20,
-            unit: toolCall.func.arguments.format === 'celsius' ? 'C' : 'F',
+            unit:
+              JSON.parse(toolCall.func.arguments).format === 'celsius'
+                ? 'C'
+                : 'F',
           };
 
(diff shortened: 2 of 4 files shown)


mharris717:
My unsolicited 2 cents: bite the bullet and make the full change. It’s currently broken for every single person that uses it in a new project, as the types lie.

packages/core/streams/ai-stream.ts
```diff
@@ -16,7 +16,7 @@ export interface ToolCallPayload {
     type: 'function';
     func: {
       name: string;
-      arguments: Record<string, unknown>;
+      arguments: string;
```
lgrammel, line 19: This impacts non-RSC code. Have you tested this with any of the regular function calling examples?

lgrammel: Also, why does this only affect tool calls but not function call (line 11)?

IdoPesok:
Right on, in that case I will go ahead and close this PR and reopen one that matches #1065 . The error in #1080 should be fixed now as the arguments won't be a string.

PR #1087, opened 2024-03-04
added the additionalParams property to the options object in render function

### LINKED ISSUE #1061: Extend ai/rsc render function to handle additional parameters
### Feature Description

Right now the `render` method has the following function signature:

```
/**
   * The model name to use. Must be OpenAI SDK compatible. Tools and Functions are only supported
   * GPT models (3.5/4), OpenAI Assistants, Mistral small and large, and Fireworks firefunction-v1.
   *
   * @example "gpt-3.5-turbo"
   */
  model: string;
  /**
   * The provider instance to use. Currently the only provider available is OpenAI.
   * This needs to match the model name.
   */
  provider: OpenAI;
  messages: Parameters<
    typeof OpenAI.prototype.chat.completions.create
  >[0]['messages'];
  text?: Renderer<{ content: string; done: boolean }>;
  tools?: {
    [name in keyof TS]: {
      description?: string;
      parameters: TS[name];
      render: Renderer<z.infer<TS[name]>>;
    };
  };
  functions?: {
    [name in keyof FS]: {
      description?: string;
      parameters: FS[name];
      render: Renderer<z.infer<FS[name]>>;
    };
  };
  initial?: ReactNode;
  temperature?: number;
```

It would be really helpful (and useful) to support a wider range of parameters with an arbitrary dictionary (maybe named `additionalParams`).  Namely, `max_tokens` is missing from this function signature and thus callers can't guide the length of outputs.

### Use Case

This allows users to use not just OpenAI models but also a wider range of models hosted by open source inference infrastructure.  As an example, [SGLang](https://github.com/sgl-project/sglang) and [Lorax](https://github.com/predibase/lorax) have mostly OpenAI compatible APIs but they don't fully support function calling.  Instead they have output format specifications for controlled JSON results.  Without arbitrary parameters passed through to the completion call, these other services can't be used fully.

### Additional context

_No response_

ChetanPatil12:
This pr fixes #1061

I added the 'additionalParams' property to the options object, allowing users to pass arbitrary parameters when calling the render function.

### DIFF
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -150,6 +150,10 @@ type Renderer<T> = (
   | Generator<Streamable, Streamable, void>
   | AsyncGenerator<Streamable, Streamable, void>;
 
+  type AdditionalParams = {
+    [key: string]: any;
+  };
+
 /**
  * `render` is a helper function to create a streamable UI from some LLMs.
  * Currently, it only supports OpenAI's GPT models with Function Calling and Assistants Tools.
@@ -194,6 +198,7 @@ export function render<
   };
   initial?: ReactNode;
   temperature?: number;
+  additionalParams?: AdditionalParams;
 }): ReactNode {
   const ui = createStreamableUI(options.initial);
 


BlairWoods (approved):
These changes look good.

BlairWoods:
Hi - any chance this can be approved :) We're really needing this feature :)

lgrammel:
We recommend using `streamUI` instead of `render` with `v3.1` and newer. Does this apply to `streamUI`?

PR #1090, opened 2024-03-04
fix: remove unnecessary `?` symbols

luckrnx09:
This PR removed the unnecessary `?` symbols in the core/react.


MaxLeiter:
Thanks for this change, but these `?` are intentional: `option?: value` means the option can be undefined implicitly (by not passing `option`). `option: value | undefined` means you need to explicitly pass `option: undefined`.

PR #1091, opened 2024-03-04
fix: Fix type declaration, upgrade bundler and clean up example

shuding:
This PR changes `index.d.ts` to `index.d.mts` as it's needed when we have `import {} from './mod.mjs'`. There're also a few unused imports and other changes that are not synced from our deployed version.


packages/core/package.json
```diff
@@ -44,7 +44,7 @@
       "require": "./dist/index.js"
     },
     "./rsc": {
-      "types": "./rsc/dist/index.d.ts",
+      "types": "./rsc/dist/index.d.mts",
```
lgrammel, line 47: does this work with typescript / commonjs setups?

examples/next-ai-rsc/package.json
```diff
@@ -19,7 +19,7 @@
     "@radix-ui/react-tooltip": "^1.0.7",
     "@vercel/analytics": "^1.2.2",
     "@vercel/kv": "^1.0.1",
-    "ai": "3.0.4",
+    "ai": "latest",
```
lgrammel, line 22: why is this necessary?

packages/core/package.json
shuding: Since this subpath only has ESM builds, so no CJS related issues. We have a shared .mjs file in dist and the import must be `import ... from './mod.mjs'` because as per the spec, the extension can't be omitted. Then, due to https://www.typescriptlang.org/docs/handbook/modules/reference.html#the-moduleresolution-compiler-option, when TS sees that `./mod.mjs` it will look for `.d.mts` instead of `.d.ts`.

examples/next-ai-rsc/package.json
shuding: This avoids that we need to bump the version number here manually whenever we do an AI package release.

lgrammel (approved):

examples/next-ai-rsc/package.json
lgrammel: shouldn't this happen automatically as it does for all other packages?

PR #1094, opened 2024-03-04
example: Upgrade Next.js for the RSC demo

shuding:
There're some Server Actions / Turbopack related bug fixed in the canary channel currently.

The CI type error needs #1091 to be landed first.

### DIFF
--- a/examples/next-ai-rsc/package.json
+++ b/examples/next-ai-rsc/package.json
@@ -25,7 +25,7 @@
     "d3-scale": "^4.0.2",
     "date-fns": "^3.3.1",
     "geist": "^1.2.2",
-    "next": "14.1.1",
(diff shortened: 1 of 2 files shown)


lgrammel (approved):

alexandrethsilva:
@shuding, quick question: is this related to this error? 👇🏾 

```
Unhandled Runtime Error
Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
  {submitUserMessage: function bound innerAction}
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

MaxLeiter:
@alexandrethsilva can you open a new issue with the error, your code, and node.js/next.js version?

PR #1099, opened 2024-03-05
fix: Streamable UI `.append()` missing the initial node

### LINKED ISSUE #1038: createStreamableUI.append only appends on the last node
### Description

Calling streamable.update() won't append unless it's immediately followed by streamable.done(). This has the effect that append can't actually append unless you have the use case of two nodes.

### Code example

```tsx
export default function Home() {
  const streamable = createStreamableUI();

  setTimeout(() => {
    streamable.append(<p>Foo</p>);
    setTimeout(() => {
      streamable.append(<p>Bar</p>);
    }, 1000);
  }, 1000);

  return streamable.value;
}
```

### Additional context

Reproduction: https://github.com/tom-sherman/vercel-ai-stream-ui-append-bug

shuding:
When calling `.append()`, the resolved node should be a concatenation of the current and the new node.

Closes #1038.


packages/core/rsc/streamable.ui.test.tsx
```diff
@@ -35,6 +35,10 @@ afterAll(() => {
   server.close();
 });
 
+function sleep(ms: number) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
```
lgrammel, line 40: This slows down the tests and can lead to instability. Why is this needed?

packages/core/rsc/streamable.ui.test.tsx
```diff
@@ -157,3 +177,151 @@ describe('rsc - render()', () => {
...
+        3
+      </div>
+    `);
+  });
+
+  it('should emit React Nodes that can be updated with .done()', async () => {
+    const ui = createStreamableUI(<div>1</div>);
+    ui.update(<div>2</div>);
+    ui.update(<div>3</div>);
+    await sleep(100);
```
lgrammel, line 202: can the await be on something else, like the ui.update call?

lgrammel (approved):

packages/core/rsc/streamable.ui.test.tsx
shuding: `ui.update` calls are always synchronous, but I hear you!

PR #1116, opened 2024-03-06
Fixed mismatched function name in demo

changesbyjames:
I don't think this is integral to this working because GPT is smart enough to use the data regardless but it might make it slightly more reliable.


shuding:
Thanks!

PR #1119, opened 2024-03-06
fix(render): parse the args based on the zod schema

IdoPesok:
#1065 Take 3

#1065 was reverted in #1080 due to a bug where the `toolCallPayload.func.arguments` were a string rather than a Record. In #1113 this behavior was fixed so now that bug should be gone since Zod will be parsing the object.

### DIFF
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -182,14 +182,14 @@ export function render<
     [name in keyof TS]: {
       description?: string;
       parameters: TS[name];
-      render: Renderer<z.infer<TS[name]>>;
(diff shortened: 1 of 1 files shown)


lgrammel:
We recommend using `streamUI` instead of `render` (which is now deprecated). Would this also apply to `streamUI`?

unstubbable:
In `streamUI` this is now handled via `safeParseJSON`, see https://github.com/vercel/ai/blob/a4848928c030a0e79d9d1b58c7c9893b573f1302/packages/core/rsc/stream-ui/stream-ui.tsx#L261.

PR #1138, opened 2024-03-12
Pass data from stream to the Message in useChat

Pldesch:
The data prop in useChat allows to recover some data from the stream. Once we receive the data it is not very easy to do something with it. So I added the possibility to add it directly to the message being streamed, allowing to display some extra data with it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,6 @@ node_modules
 .turbo
 *.log
 .next
-dist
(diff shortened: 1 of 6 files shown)


tomasmcm:
Is this PR ready? Can we get it merged? we would really like to get the stream data directly in the message object instead of having to reconcile everything manually.

PR #1184, opened 2024-03-19
[BUGFIX] - useAssistant - Check for empty array

tchanxx:
When I was using the `useAssistant` hook, I noticed this error: `Unhandled Runtime Error
TypeError: Cannot set properties of undefined (setting 'id')`

image

This error was kind of cryptic since it doesn't say which element in `useAssistant` is causing this error. However, I figured out that it's due to `lastMessage` being undefined in the following code:
```
const lastMessage = messages[messages.length - 1];
  return [
    ...messages.slice(0, messages.length - 1),
    {
      id: lastMessage.id,
      role: lastMessage.role,
      content: lastMessage.content + value,
    },
  ];
  ```
  
  When `messages` has a length of `0`, then `lastMessage` becomes undefined and then we get a runtime error at `lastMessage.id`.

### DIFF
--- a/packages/core/react/use-assistant.ts
+++ b/packages/core/react/use-assistant.ts
@@ -182,6 +182,9 @@ export function experimental_useAssistant({
           case 'text': {
             // text delta - add to last message:
             setMessages(messages => {
+              if (messages.length === 0) {
+                return messages;
+              }
               const lastMessage = messages[messages.length - 1];
               return [
                 ...messages.slice(0, messages.length - 1),
@@ -214,6 +217,9 @@ export function experimental_useAssistant({
 
             // set id of last message:
             setMessages(messages => {
+              if (messages.length === 0) {
+                return messages;
+              }
               const lastMessage = messages[messages.length - 1];
               lastMessage.id = value.messageId;
               return [...messages.slice(0, messages.length - 1), lastMessage];

### CODE REVIEW

**Comment by lgrammel** [comment]
There should always be a message at the point when text fragments start streaming in, so the root cause could be different. Can you share the code that caused this bug? One potential cause could be using `setMessages` while there is ongoing streaming.

**Review by lgrammel — CHANGES_REQUESTED** [review]
Marking as on hold for now. The fix most likely needs to be more involved.

**Comment by tchanxx** [comment]
> One potential cause could be using `setMessages` while there is ongoing streaming.

@lgrammel Ahh that's a good point. It's hard for me to share the entire code, but basically I have a setup similar to below:

```tsx
useEffect(() => {
  if (selectedThread) {
    fetchMessages()
  } else {
    setChatMessages([])
  }
}, [selectedThread, setChatMessages])


const handleSendMessage = async () => {
  let currentThread = selectedThread ? { ...selectedThread } : null

  if (!currentThread) {
    currentThread = await handleCreateThread(
      user.id,
      userInput,
      chatSettings!,
      selectedWorkspace!,
      setSelectedThread,
      setThreads
    )
  }

  await submitMessage()
}
```

I think my issue is that when I create a new thread in `handleSendMessage()`, I create a new thread and then call `setSelectedThread()` which triggers the `useEffect()` hook to clear my messages via `setChatMessages([])` but those are all asynchronous calls so `setChatMessages([])` may happen during `await submitMessage()`. 

I can reconfigure my app so that it relies on the thread creation in the `useAssistant()` hook instead of creating it separately which should resolve this issue.

-----------

I am curious though, in this `useAssistant()` code, how does it ensure that there's always a message in the array before hitting those `lastMessage` code? [This part is async so it can't be guaranteed](https://github.com/vercel/ai/blob/8ca4096b5dd4f90d9ffafef43e9a2e0c10d3a468/packages/core/react/use-assistant.ts#L139C5-L142C8) to complete first so that means `assistant_message` or `data_message` would always have to be sent before `text` or `assistant_control_data`? 

I am looking at the streaming code and it seems that `assistant_control_data` [is always enqueued first](https://github.com/vercel/ai/blob/fac9a6ee7c3aca975f7dd9f2f40cda1b82a1f8f3/packages/core/streams/assistant-response.ts#L88-L96) so that may cause the `id` issue I'm seeing as well?

lgrammel:
> I am curious though, in this `useAssistant()` code, how does it ensure that there's always a message in the array before hitting those `lastMessage` code? [This part is async so it can't be guaranteed](https://github.com/vercel/ai/blob/8ca4096b5dd4f90d9ffafef43e9a2e0c10d3a468/packages/core/react/use-assistant.ts#L139C5-L142C8) to complete first so that means `assistant_message` or `data_message` would always have to be sent before `text` or `assistant_control_data`?
> 
> I am looking at the streaming code and it seems that `assistant_control_data` [is always enqueued first](https://github.com/vercel/ai/blob/fac9a6ee7c3aca975f7dd9f2f40cda1b82a1f8f3/packages/core/streams/assistant-response.ts#L88-L96) so that may cause the `id` issue I'm seeing as well?

The OpenAI stream sends a `thread.message.created` event first, before the text delta stream starts. This trigger the sending of an `assistant_message`: https://github.com/vercel/ai/blob/fac9a6ee7c3aca975f7dd9f2f40cda1b82a1f8f3/packages/core/streams/assistant-response.ts#L50

lgrammel:
> @lgrammel Sorry, I'm still confused. Isn't `assistant_control_data` enqueued prior to `forwardStream` being called here?
> 
> https://github.com/vercel/ai/blob/fac9a6ee7c3aca975f7dd9f2f40cda1b82a1f8f3/packages/core/streams/assistant-response.ts#L88-L105

Yes, first there is a control data message, then the `assistant_message`, and then the streaming. Am I missing something?

tchanxx:
If the data control message comes first, then couldn't this fail since `messages` may be empty then `lastMessage` is undefined? 

https://github.com/vercel/ai/blob/main/packages/core/react/use-assistant.ts#L212-L223

lgrammel:
> If the data control message comes first, then couldn't this fail since `messages` may be empty then `lastMessage` is undefined?
> 
> https://github.com/vercel/ai/blob/main/packages/core/react/use-assistant.ts#L212-L223

Yes, good point. This would need a check if there are any messages.

lgrammel:
@tchanxx  Can you update the PR to only check in the control data part, and move the if check outside of setMessages?

packages/core/react/use-assistant.ts
```diff
@@ -214,6 +217,9 @@ export function experimental_useAssistant({
 
             // set id of last message:
             setMessages(messages => {
+              if (messages.length === 0) {
+                return messages;
+              }
```
lgrammel, line 222: This can be checked before setMessages, e.g. like this:

```
if (messages.length > 0) {
  setMessages ...
}

Issue #1174, opened 2024-03-18
AI is not aware of useAIState set history


Strernd:
### Description

I based my code of the example https://sdk.vercel.ai/docs/concepts/ai-rsc and added basic components.
The search works as expected, but when I select a result, and call
```ts
    const info = {
      role: "system" as const,
      content: `[User has selected customer with id ${customerId}]`,
      id,
    };
    setAIState([...aiState, info]);
``` 
it get's added to the history but the AI seems not be aware of it.

Side note: the example uses role `assistant` the demo uses `system`.

I also tried without the display component and it also wasn't able to name the selected id. In the demo I was able to get a text response containing the amount I spent on stocks.

<img width="966" alt="image" src="https://github.com/vercel/ai/assets/2428581/5148a25d-e2c4-4ce5-ad22-ebf1dc638860">
<img width="734" alt="image" src="https://github.com/vercel/ai/assets/2428581/683b6aa3-bed4-4f80-b1b8-1b128fa2c113">

### Code example

https://gist.github.com/Strernd/3e3d47c1bd420588441699c817922731

shuding:
I think the issue is that you're missing the AI state in the LLM `messages` so it doesn't get the info:

https://gist.github.com/Strernd/3e3d47c1bd420588441699c817922731#file-action-tsx-L48-L55

Strernd:
Thanks @shuding 
adding
```ts
    ...aiState.get().map((info: any) => ({
      role: info.role,
      content: info.content,
      name: info.name,
    })),
  ```
  to `messages` in the `render` function worked.
  It may be worth adding a callout for this in the example here: https://sdk.vercel.ai/docs/concepts/ai-rsc#updating-ai-state-from-client-components

lgrammel:
## Identify issue type

Identified as **Documentation**  
Confidence: **High**

### Rationale
The reported behavior was caused by omitted message state rather than a product defect, and the clarified request is to improve the documentation.

### Evidence
- The maintainer identified that the AI state was not included in the LLM `messages` array.
- The reporter confirmed that mapping `aiState.get()` into `messages` fixed the problem.
- The remaining request is to add a callout explaining this requirement in the AI RSC documentation.

PR #1206, opened 2024-03-21
Fix `submitUserMessage` snippet in Generative UI docs

unstubbable:
This is a follow-up from #1174. Usually, the LLM should receive the whole message history for every action call.


MaxLeiter (approved):

MaxLeiter:
Thanks!

PR #1212, opened 2024-03-22
Update RSC example to streamText.

lgrammel:


lgrammel:
@jeremyphilemon @shuding not sure if I like this new version better. Replacing system messages with user seems somewhat strange, and I'm not sure if it might cause issues.

PR #1223, opened 2024-03-25
Set up Zep example Next JS app

paul-paliychuk:
Summary:
Created a sample Next JS showcasing Vercel AI SDK integration with Zep Memory (Cloud Version). Exposes two conversation handler routes
`zep-history` - a pure Zep JS integration.
`langchain-zep-history` - uses LangChain compatible classes provided by zep-js sdk for LangChain users.

Session ID is required for chat memory, the way the client pages are set up in the sample app is to look for a `sessionId` in search query, and set a random one if not provided.


jaredpalmer:
Can we add another example without langchain?

paul-paliychuk:
@jaredpalmer Hey, so I added an example just using zep and another example using zep + langchain integration (on a different route). Happy to extract non-langchain example into a separate example app.

paul-paliychuk:
Just extracted LangChain example into `next-zep-langchain` app. Non-langchain integration is in `next-zep` app and doesn't have `langchain` dependency.

PR #1233, opened 2024-03-26
feat(ai/rsc): Add `useStreamableValue` again

shuding:
This PR re-adds the previously removed `useStreamableValue` declarative API as it's still very convenient in certain cases. Its return type is also changed to a `[payload, error, pending]` triplet.

The challenging part might be that how we want to document these different utilities along with all the concepts.

### DIFF
--- a/packages/core/rsc/shared-client/streamable.tsx
+++ b/packages/core/rsc/shared-client/streamable.tsx
@@ -2,15 +2,19 @@ import { startTransition, useLayoutEffect, useState } from 'react';
 import { STREAMABLE_VALUE_TYPE } from '../constants';
 import type { StreamableValue } from '../types';
 
+function hasReadableValueSignature(value: unknown): value is StreamableValue {
+  return !!(
+    value &&
+    typeof value === 'object' &&
+    'type' in value &&
+    value.type === STREAMABLE_VALUE_TYPE
+  );
+}
+
 function assertStreamableValue(
   value: unknown,
 ): asserts value is StreamableValue {
-  if (
-    !value ||
-    typeof value !== 'object' ||
-    !('type' in value) ||
-    value.type !== STREAMABLE_VALUE_TYPE
-  ) {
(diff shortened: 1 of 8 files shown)


packages/core/rsc/shared-client/streamable.tsx
```diff
@@ -11,11 +12,28 @@ function assertStreamableValue(
...
 
+function isStreamableValue(value: unknown): value is StreamableValue {
+  const hasSignature = !!(
+    value &&
+    typeof value === 'object' &&
+    'type' in value &&
+    value.type === STREAMABLE_VALUE_TYPE
+  );
+
+  if (!hasSignature && typeof value !== 'undefined') {
```
lgrammel, line 28: Is `typeof value !== "undefined"` needed here? (since there is `typeof value === 'object'` above)

packages/core/rsc/shared-client/streamable.tsx
```diff
@@ -11,11 +12,28 @@ function assertStreamableValue(
...
     );
   }
 }
 
+function isStreamableValue(value: unknown): value is StreamableValue {
+  const hasSignature = !!(
+    value &&
+    typeof value === 'object' &&
+    'type' in value &&
+    value.type === STREAMABLE_VALUE_TYPE
```
lgrammel, line 25: Similar to the code above, just negated. Could be extracted.

lgrammel (approved):

packages/core/rsc/shared-client/streamable.tsx
shuding: Yeah it's intentional, I don't want to throw for the "lazy" case:

```js
const [streamableValue, setStreamableValue] = useState()
const [value] = useStreamableValue(streamableValue) // the value will be undefined initially

return <button onClick={async () => {
  const ret = await action()
  setStreamableValue(ret)
}}>
  ...
</button>
```

PR #1241, opened 2024-03-27
fix: update input component

mbj36:


shuding:
Thanks for the PR! However we believe that the current implementation is good enough as an example and `<label>` is necessary for a11y reasons. Cheers!

mbj36:
Screenshot 2024-03-27 at 6 31 10 PM

@shuding This is how the existing experience looks like

PR #1245, opened 2024-03-27
(ai/rsc): add renderAzureOpenAI 

fujita-h:
Added a function to use ai/rsc's render with Azure OpenAI.

The processing itself is almost the same as `render`, the only difference being the arguments and provider, so I think it would be better to extend `render` to accept multiple types of providers. I think how you write it depends on your policy. Here I wrote it in a separate function.


lgrammel:
We have developed a pluggable provider infrastructure, see https://sdk.vercel.ai/docs/ai-sdk-core/providers-and-models , and now recommend using `streamUI` instead of `render`. The best solution for this would be to implement an Azure OpenAI provider.

PR #1260, opened 2024-04-02
ai/core anthropic provider (no tool calls)

lgrammel:
## Summary
- Add Anthropic provider with messages API support (no tool calls)
- Add automatic mimetype detection for non-URL images


packages/core/package.json
```diff
@@ -17,11 +17,12 @@
     "spec/dist/**/*",
     "google/dist/**/*",
     "openai/dist/**/*",
+    "anthropic/dist/**/*",
     "mistral/dist/**/*"
   ],
   "scripts": {
     "build": "tsup && cat react/dist/index.server.d.ts >> react/dist/index.d.ts",
-    "clean": "rm -rf dist && rm -rf core/dist && rm -rf google/dist && rm -rf openai/dist && rm -rf mistral/dist && rm -rf spec/dist && rm -rf react/dist && rm -rf svelte/dist && rm -rf vue/dist && rm -rf solid/dist && rm -rf rsc/dist",
+    "clean": "rm -rf dist && rm -rf core/dist && rm -rf google/dist && rm -rf anthropic/dist && rm -rf openai/dist && rm -rf mistral/dist && rm -rf spec/dist && rm -rf react/dist && rm -rf svelte/dist && rm -rf vue/dist && rm -rf solid/dist && rm -rf rsc/dist",
```
MaxLeiter, line 25: These should probably be broken up into `clean:{provider}` at some point like `test`. We can use something like https://www.npmjs.com/package/npm-run-all for concurrency

MaxLeiter (approved):

packages/core/package.json
lgrammel: Agreed, ideally I'd like to move the providers to separate packages.

PR #593, opened 2023-09-24
Add remix example

nicnocquee:
This PR adds an example of using Remix with OpenAI. Should resolves issue #199

PR #1274, opened 2024-04-04
Add remix example

laiso:
This PR is a revision of #593
I've confirmed that workflows pass in [my forked repo](https://github.com/laiso/ai/actions/runs/8552288522).


broisnischal:
! update

wong2:
Would you update it with AI SDK 3.1?

lilouartz:
Looks like it is supposed to be something like this:

```ts
import { config } from '#app/config.server';
import { defineAction } from '#app/unstable';
import { StreamingTextResponse, streamText } from 'ai';

export default defineAction(async ({ request }) => {
  const openai = createOpenAI({
    apiKey: config.OPENAI_API_KEY,
    compatibility: 'strict',
  });

  const model = openai('gpt-4o');

  const { messages } = await request.json();

  const response = await streamText({
    messages,
    model,
  });

  return new StreamingTextResponse(response.textStream);
});

```

lgrammel:
Would love an React Router / Remix example for the current version of the SDK. This is unfortunately outdated.

PR #1275, opened 2024-04-04
[wip] ai/core anthropic provider tool calling

lgrammel:


avianion:
@lgrammel tool use beta is out and is much more robust, you might wanna use this

https://docs.anthropic.com/claude/docs/tool-use

PR #1292, opened 2024-04-07
New Provider: Portkey 

saif-shines:
This pull request adds Portkey as a provider that works along Vercel AI SDK.

- It demonstrates using Portkey and Vercel AI SDK as a Unified API to send requests to 150+ LLMs.
- To add Observability layers for the NextJS apps over LLM interactions.

Please feel free to request any changes that might be required. We will to on top of it, and address them.

### DIFF
--- a/docs/pages/docs/guides/providers/portkey.mdx
+++ b/docs/pages/docs/guides/providers/portkey.mdx
@@ -55,7 +55,7 @@ For this example, create a route handler at `app/api/chat/route.ts` that calls G
 import { OpenAIStream, StreamingTextResponse } from 'ai';
 import { Portkey } from 'portkey-ai';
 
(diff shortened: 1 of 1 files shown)


AWKohler:
Hello, you mentioned that portkey would allow anthropic's models to use function calling by routing through the existing openai function calling support in the AI SDK, however it does not appear that portkey supports that with antropic currently. [link](https://portkey.ai/docs/product/ai-gateway-streamline-llm-integrations/multimodal-capabilities/function-calling#supported-providers-and-models)

lgrammel:
We have changed to a new docs system & deprecated the old wrapper approach for AI providers with 3.1. Please re-open if this is still an issue with the new docs.

PR #1300, opened 2024-04-08
Handle OpenAI tool calls called after a message

vinayakakv:
This PR handles the tool calls that appear in the middle of messages. The problem is exactly same as #1134, but the solution is different.

To solve the problem, we add a newline before the function call and tool call message in the stream. Then, in the client, we check for the presence of the lines and construct the objects accordingly.

This will _still_ break in the cases where a wrong message is constructed and `JSON.parse` fails.

Please let me know your thoughts on this. Love the great work!


rajdtta:
Great work!

vinayakakv:
Closing as the complex stream solution worked perfectly

vinayakakv:
It should work automatically today with #1192

PR #1320, opened 2024-04-10
feat(types): Export type UseCompletionOptions

danmana:
Similar to how `UseChatOptions` is exported.

Useful in cases where you want to omit some fields.

```ts
import { useCompletion, UseCompletionOptions } from 'ai/react'

const useMyCompletion = (options: Omit<UseCompletionOptions, 'api' | 'headers'>) => {
  const authToken = useMyToken()

  const completion = useCompletion({
    ...options,
    api: process.env.NEXT_PUBLIC_MY_AI_COMPLETION_URL,
    headers: {
      Authorization: `Bearer ${authToken}`,
    },
  })

  return completion
}
```

### DIFF
--- a/packages/core/react/use-completion.ts
+++ b/packages/core/react/use-completion.ts
@@ -7,6 +7,8 @@ import {
   UseCompletionOptions,
 } from '../shared/types';
 
+export type { UseCompletionOptions };
+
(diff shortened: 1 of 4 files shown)


lgrammel (approved):
Thank you!

PR #1375, opened 2024-04-17
Make `onSetAIState` and `onGetUIState` stable

shuding:
Removes the `unstable_` prefix and adds JSDoc for each method.


lgrammel (approved):
Do any docs/examples need updating?

This might need a changeset.

Issue #1349, opened 2024-04-15
render function error handling


steebchen:
### Description

It looks like the render function doesn't support error handling. I tried wrapping the render function in a try/catch, but it doesn't catch any errors (specifically when an error is immediately returned, such as a context window length exceeded)/

### Code example

I used this template: https://github.com/vercel/ai-chatbot

At this part:

```
return render({
  model: 'gpt-3.5-turbo',
  provider: openai,
  initial: <SpinnerMessage />,
  messages: [
    {
      role: 'system',
      content: `You are a helpful bot.`
    },
    ...aiState.get().messages.map((message: any) => ({
      role: message.role,
      content: message.content,
      name: message.name
    }))
  ]
})
```

This can result in this error being thrown

```
 ⨯ unhandledRejection: BadRequestError: 400 This model's maximum context length is 16385 tokens. However, your messages resulted in 21253 tokens. Please reduce the length of the messages.
```

and the UI just keeps showing this:

<img width="145" alt="image" src="https://github.com/vercel/ai/assets/5013932/20151cc9-a631-4983-a6e8-4379d6b7ad1d">

parthematics:
Also experiencing this issue - would love some guidance here!

nateaxcell:
Same, its obvious that the openAI library is throwing an error but there's no consistent way to catch it using this library it seems.

ghaemisr:
Same here! Couldn't find any documentation on how this should be handled.

gediminastub:
Same here, still can't find how to do that, would highly appreciate guidance!

lgrammel:
### Identify issue type

**Identified type: Bug**  
Confidence: High

Rationale: The issue describes a concrete runtime failure in existing behavior: provider errors from `render`/`streamUI` are not catchable as expected and leave the UI in a broken loading state.

Evidence:
- Calling `render`/`streamUI` with messages exceeding the model context can produce an unhandled `BadRequestError` rejection.
- The reporter says wrapping the render function in `try/catch` does not catch the error and the UI remains stuck on the loading spinner.
- Other commenters report the same inability to consistently catch OpenAI errors through the library.

n1ckoates:
This is possible with the `streamUI` API.

```ts
try {
  const result = await streamUI({
    model,
    messages,
    initial: <SpinnerMessage />,
  });

  return <ErrorBoundary>{result.value}</ErrorBoundary>;
} catch (error) {
  // Request failed before streaming began.
  return <ErrorMessage />;
}
```

PR #1379, opened 2024-04-18
New `experimental_streamUI()` API

shuding:
This PR reimplements `experimental_streamUI()` on top of the `LanguageModelV1` spec. Ideally, this new API is gonna replace the current `render()` with the following changes:
- `experimental_streamUI()` is async and returns an object with `value` field as the React node, as well as other fields similar to `experimental_streamText()` does.
- The `model` option aligns with [AI core](https://sdk.vercel.ai/docs/ai-core)'s `model` value.
- The `provider` option is removed.
- `functions` is removed in term of `tools`.
- `tools`'s `render()` method is renamed as `generate()`.

`render()` is still kept unchanged, but with a deprecation message.

Some of the code are on top of #1372.

## Remaining issues to make this API stable
- [x] Throw if passing old options to make the migration smoother
- [ ] Better error handling and error boundary UI
- [ ] Figure out a way to stream `text()` node without reconciliation


packages/core/rsc/render/render.tsx
```diff
@@ -0,0 +1,288 @@
...
+      finished = resolvable.promise;
+    }
+
+    const value = renderer(...args);
+    if (
+      value instanceof Promise ||
+      (value &&
+        typeof value === 'object' &&
+        'then' in value &&
+        typeof value.then === 'function')
```
lgrammel, line 127: out of curiosity, why is this needed if we already know it's a Promise?

lgrammel: a nevermind, confused the `||` with `&&`

packages/core/rsc/render/render.tsx
```diff
@@ -0,0 +1,288 @@
...
+        ui.done();
+      } else {
+        handleRender([{ content, done: true }], textRender, ui);
+        await finished;
+        ui.done();
+      }
+    } catch (error) {
+      // During the stream rendering, we don't want to throw the error to the
+      // parent scope but only let the React's error boundary to catch it.
+      ui.error(error);
```
lgrammel, line 279: Throwing the error and handling it here breaks the stream. Sometimes, the stream can be mostly good, only small parts fail (like a single tool call). An alternative could be a more granular error handling (catching/handling for each iteration inside the loop)

lgrammel (approved):

packages/core/rsc/render/render.tsx
shuding: Good point! Need to redesign that part a bit. Related: #1349.

MISTYKITTY (approved):

janpawellek:
I really appreciate this change! It makes everything easier and smoother. I'm looking forward to this being released 👍🏻

ap-1:
The [AI Core pages](https://sdk.vercel.ai/docs/ai-core) do not list `experimental_streamUI()`, while the [Generative UI pages](https://sdk.vercel.ai/docs/api-reference/generative-ui/render) still recommend `render()`. Was `render()` marked as deprecated without this new API having been documented yet?

This PR mentions a few changes, but ignores the change in the `messages` property. What is `ExperimentalMessage[]` and how does it differ from `AIStateItem[]` used in `render()`? Does this new API play with `ai/rsc` functions like `getAIState()`, `useAIState()`, etc., or is it something entirely different?

PR #1396, opened 2024-04-20
add @anthropic-ai/sdk to list of packages to be added

DayakarH:


lgrammel:

PR #1397, opened 2024-04-22
Logprobs Support

SamStenner:
# Logprobs

Log probabilities (logprobs) represent the likelihood of each token occurring in a sequence given the context. A higher log probability signifies greater confidence in the model's output within a specific context.

# API
This PR provides support for logprobs in the experimental [AI Core ](https://sdk.vercel.ai/docs/ai-core)API. Currently, only OpenAI models are supported. I have included logprobs as an OpenAI specific setting, but they could easily be promoted to a general setting. Other providers do support the concept of logprobs, but will be implemented in a separate PR.

All logprobs are unified under the `LanguageModelV1LogProbs` type:
```typescript
export type LanguageModelV1LogProbs = {
  token: string; // the generated token
  logprob: number; // the log probability of this token
  topLogprobs: { // an array of the top `n` tokens that could have been chosen, sorted by log probability
    token: string;
    logprob: number;
  }[];
}[];
```

## OpenAI

|               | generateText | streamText | generateObject | streamObject
| -------- | ------- | -------- | -------- | -------- |
| text        | ✅     |  ✅      | **-** | **-** | 
| json       | **-** |  **-** |  ✅     | ✅     |
| tool       | ❌     |  ❌      |  ❌     | ❌     |

OpenAI supports logprobs on both their `chat` and `completion` APIs, although with slightly different inputs and outputs. This implementation maintains their separate inputs, but unifies their outputs into the `LanguageModelV1LogProbs` type.

### Examples

**Basic Usage**

```typescript
// Generate text with chat
const result = await experimental_generateText({
  model: openai.chat('gpt-3.5-turbo', { logprobs: 3 }),
  system: `You are a helpful, respectful and honest assistant.`,
  messages,
});


// Generate text with completions
const result = await experimental_generateText({
  model: openai.completion('gpt-3.5-turbo-instruct', { logprobs: 3 }),
  system: `You are a helpful, respectful and honest assistant.`,
  messages,
});
```

<details>
<summary>Streaming Text</summary>

```typescript
const result = await experimental_streamText({
  model: openai.chat('gpt-3.5-turbo', { logprobs: 3 }),
});

  switch (part.type) {
    case 'finish': {
      console.log('Logprobs:', part.logprobs);
      break;
    }
  }
}
```

</details>

<details>
<summary>Generating Object</summary>

```typescript
const result = await experimental_generateObject({
  model: openai.chat('gpt-3.5-turbo', { logprobs: 3 }),
  schema: z.object({ ... }),
  prompt: 'Generate a lasagna recipe.',
  mode: 'json',
});
console.log(result.logprobs);
```
</details>

<details>
<summary>Streaming Object</summary>

```typescript
const result = await experimental_streamObject({
  model: openai.chat('gpt-3.5-turbo', { logprobs: 3 }),
  schema: z.object({ ... }),
  prompt: 'Generate a lasagna recipe.',
  mode: 'json',
});
  switch (part.type) {
    case 'finish': {
      console.log('Logprobs:', part.logprobs);
      break;
    }
  }
}
```
</details>

**Notes**
- Since OpenAI completions do not support generating or streaming objects, nor tools, logprobs are only available when generating or streaming text (when using completions).
- When generating or streaming objects, json mode must be selected as OpenAI does not provide logprobs for tool usage.
- Setting `logprobs` to 0 will return the logprobs for the tokens that were actually provided. Setting `logprobs` to 1 or more will populate the `top_logprobs` array, which is a list of the highest scoring logprobs (up to 20). The first `logprob` in this list will be the one that was chosen, so setting the `logprobs` property to 1 is somewhat pointless. Either set to 0 or 2+.

## Anthropic
Not supported

## Google
Not supported

## Mistral
Not supported


MISTYKITTY (approved):

packages/core/core/generate-object/generate-object.ts
```diff
@@ -242,15 +248,22 @@ Warnings from the model provider (e.g. unsupported settings)
    */
   readonly warnings: LanguageModelV1CallWarning[] | undefined;
 
+  /**
+Logprobs for the completion.
```
lgrammel, line 252: Add: `undefined` if the mode does not support logprobs or if was not enabled

packages/provider/src/language-model/v1/language-model-v1.ts
```diff
@@ -125,6 +131,9 @@ export type LanguageModelV1StreamPart =
   // Basic text deltas:
   | { type: 'text-delta'; textDelta: string }
 
+  // Logprob results:
+  | { type: 'log-probs'; logprobs: LanguageModelV1LogProbs }
+
```
lgrammel, line 136: Instead of introducing a separate stream part, please add a `logprobs` property to the `finish` part (and also change that behavior in streamText etc).

packages/provider/src/language-model/v1/language-model-v1-logprobs.ts
```diff
@@ -0,0 +1,8 @@
+export type LanguageModelV1LogProbs = {
+  token: string;
+  logprob: number;
+  top_logprobs: {
```
lgrammel, line 4: We use camelCase for anything that is developer facing. Please change to `topLogprobs` and map the data from OpenAI.

packages/openai/src/openai-completion-language-model.ts
```diff
@@ -72,6 +73,13 @@ export class OpenAICompletionLanguageModel implements LanguageModelV1 {
       // model specific settings:
       echo: this.settings.echo,
       logit_bias: this.settings.logitBias,
+      logprobs: scale({
+        value: this.settings.logprobs,
+        inputMin: 0,
+        inputMax: 20,
+        outputMin: 0,
+        outputMax: 20,
+      }),
```
lgrammel, line 82: This does not change the scaling and is therefore not needed

packages/openai/src/openai-completion-language-model.ts
```diff
@@ -168,6 +176,7 @@ export class OpenAICompletionLanguageModel implements LanguageModelV1 {
         completionTokens: response.usage.completion_tokens,
       },
       finishReason: mapOpenAIFinishReason(choice.finish_reason),
+      logprobs: choice.logprobs ? mapOpenAICompletionLogProbs(choice.logprobs) : undefined,
```
lgrammel, line 179: Push the ternary into the map function

packages/openai/src/openai-chat-settings.ts
```diff
@@ -35,6 +35,21 @@ export interface OpenAIChatSettings {
...
+   * response times. However, it can be useful for developers to better
+   * understand how the model is behaving.
+   */
+  logprobs?: boolean;
+
+  /**
+   * The number of top logprobs to return. This parameter is only valid if
+   * `logprobs` is set to true. Defaults to 0.
+   */
+  top_logprobs?: number;
```
lgrammel, line 51: `topLogprobs`

packages/openai/src/openai-chat-settings.ts
```diff
@@ -35,6 +35,21 @@ export interface OpenAIChatSettings {
...
   logitBias?: Record<number, number>;
 
+  /**
+   * Whether to include the log probabilities on the logprobs output field.
+   * 
+   * Including logprobs will increase the response size and can slow down
+   * response times. However, it can be useful for developers to better
+   * understand how the model is behaving.
+   */
+  logprobs?: boolean;
```
lgrammel, line 45: Can be inferred by `!(topLogprobs === 0 || topLogprobs == null)`. We could e.g. rename `topLogprobs` to logprobs and keep the semantics of `topLogprobs`

lgrammel (commented):
Thank you for the proposal!

Can you 
1) add test cases for the openai providers and the ai sdk functions
2) add some examples to `exampels/ai-core`

packages/openai/src/openai-chat-settings.ts
SamStenner: To be clear, I believe Anthropic, Google and Mistral support logprobs.

lgrammel: I'm in general in favor of adding it to the default settings. However, I'm concerned that it is very OpenAI specific atm, so it might be better to wait (so we don't end up with some idiosyncratic). I haven't seen it in the other APIs - can you point me to examples?

SamStenner: Makes sense. I think I'm actually mistaken, it seems those other providers don't currently support logprobs.
**Anthropic**: Not supported
**Google**: Supported on PaLM 2 (both text and chat) but not Gemini
**Mistral**: Seems they're available internally, but aren't returned via the API

Let's leave it as an OpenAI specific setting for now, can be promoted another time (e.g. if AzureOpenAI is added as a discrete provider)

packages/provider/src/language-model/v1/language-model-v1.ts
SamStenner: Done! I found that streaming objects only provides a `partialObjectStream`. I've quite heavily modified the `stream-object.ts` file to also provide a `fullStream` property. It works exactly the same as the `fullStream` from `streamText`, except it provides the aggregated object under the `object` type. Example:

```typescript
const result = await experimental_streamObject({
    model: openai.chat('gpt-3.5-turbo', { logprobs: 1 }),
    maxTokens: 2000,
    schema: z.object({ ... }); // type T
})
    switch (part.type) {
      case 'object':
        console.clear();
        console.log(part.object); // of Type T
        break;
      case 'finish':
        console.log('logprobs', part.logprobs);
        break;
    }
  }
  ```

**Comment by SamStenner** [comment]
@lgrammel I've implemented your feedback, please re-review :)

File: packages/provider/src/language-model/v1/language-model-v1.ts (hunk 2 above)
Comment (lgrammel) [reply packages/provider/src/language-model/v1/language-model-v1.ts]: this is a great change, wanted to do something in this direction anyways!

File: examples/ai-core/src/stream-text/openai-fullstream-with-tools.ts [hunk 8]
```diff
@@ -0,0 +1,80 @@
...
+      }
+
+      case 'error':
+        console.error('Error:', part.error);
+        break;
+    }
+  }
+}
+
+main().catch(console.error);
```
lgrammel, line 80: why does this test need to be copied? vs just having openai-fullstream?

lgrammel: the logprobs examples could be separate, 2 are enough imo:

- `openai-fullstream-logprobs` (streamText)
- `openai-logprobs` (generateText)

packages/openai/src/openai-chat-language-model.test.ts
```diff
@@ -144,6 +156,118 @@ describe('doGenerate', () => {
...
+
+    const provider = createOpenAI({ apiKey: 'test-api-key' });
+
+    const response = await provider
+      .chat('gpt-3.5-turbo', { logprobs: 1 })
+      .doGenerate({
+        inputFormat: 'prompt',
+        mode: { type: 'regular' },
+        prompt: TEST_PROMPT,
+      });
```
lgrammel, line 266: an additional test for streaming would be good, since it's a separate implementation (same for the completion mode)

examples/ai-core/src/stream-text/openai-fullstream-with-tools.ts
SamStenner: I renamed `openai-fullstream` to `openai-fullstream-with-tools` since this uses tools. I then created a new `openai-fullstream` which doesn't call tools, but does console.log the logprobs. Since logprobs and tools are incompatible, I thought it was worth separating into separate examples.

lgrammel: agree on the separation. imo logprobs is the less common use case (compared to tools), therefore i'd prefer to separate out logsprobs to a couple of logprobs examples.

SamStenner: Sure, will do!

packages/openai/src/openai-completion-language-model.test.ts
```diff
@@ -26,13 +55,22 @@ describe('doGenerate', () => {
...
     usage?: {
       prompt_tokens: number;
       total_tokens: number;
       completion_tokens: number;
     };
+    logprobs?: {
+      tokens: string[];
+      token_logprobs: number[];
+      top_logprobs: Record<string, number>[];
+      text_offset: number[];
```
lgrammel, line 71: is `text_offset` needed? if we don't use it, i'd prefer to not leave it out (to simplify the code, reduce zod validation time, and increase compatibility with not-quite-so-compatible 3rd party apis)

packages/openai/src/openai-chat-language-model.ts
```diff
@@ -410,6 +448,25 @@ const openaiChatChunkSchema = z.object({
...
           .optional(),
       }),
+      logprobs: z
+        .object({
+          content: z
+            .array(
+              z.object({
+                token: z.string(),
+                logprob: z.number(),
+                top_logprobs: z.array(
```
lgrammel, line 458: Do they always send this or does it need to be optional/nullable?

packages/openai/src/map-openai-completion-logprobs.ts
```diff
@@ -0,0 +1,26 @@
...
+  token_logprobs: number[];
+  top_logprobs: Record<string, number>[] | null;
+  text_offset: number[];
+};
+
+export function mapOpenAICompletionLogProbs(
+  logprobs: OpenAICompletionLogProps | null | undefined,
+): LanguageModelV1LogProbs | undefined {
+  if (!logprobs) return undefined;
+  return logprobs.tokens.map((token, index) => ({
```
lgrammel, line 14: `return logprobs?.tokens.map ...` achieves the same w/o the if

packages/openai/src/map-openai-chat-logprobs.ts
```diff
@@ -0,0 +1,34 @@
...
+      }[]
+    | null;
+};
+
+export function mapOpenAIChatLogProbsOutput(
+  logprobs: OpenAIChatLogProbs | null | undefined,
+): LanguageModelV1LogProbs | undefined {
+  if (!logprobs) return undefined;
+  return (
+    logprobs.content?.map(({ token, logprob, top_logprobs }) => ({
```
lgrammel, line 23: same comment as below

docs/pages/docs/ai-core/openai.mdx
```diff
@@ -101,6 +101,7 @@ const model = openai.chat('gpt-3.5-turbo', {
     // optional likelihood for specific tokens
     '50256': -100,
   },
+  logprobs: 0, // provides log probabilities for each token
```
lgrammel, line 104: hmm I know it was my idea, just not sure how much i like this - `logprobs: 0` intuitively means "turned off". Some alternative ideas: type could be `boolean | number | undefined` where `true` means `enabled, top:0`, `false` means `disabled`, 0 means `enabled, top:0` and any positive int means `enabled, top: k`

SamStenner: Yeah, I had the same concern. This is a good solution. Will update the code and comments/docs, but I think the tests are fine as they are.

lgrammel (approved):
Thank you!

PR #1454, opened 2024-04-26
Add DataStream support to streamToResponse

### LINKED ISSUE #1285: Expressjs: Allow streamToResponse to send back custom data in response body
### Feature Description

I am using React ^18.2.0 and Express ^4.19.2 and openai as the ai provider. I am proposing the addition of a feature that allows the streamToResponse function to send back custom data in the response body. Currently, the streamToResponse function only supports streaming LLM response directly from the AI provider to the client with the ability to modify only the response headers and status.
```js
/**
 * A utility function to stream a ReadableStream to a Node.js response-like object.
 */
declare function streamToResponse(res: ReadableStream, response: ServerResponse, init?: {
    headers?: Record<string, string>;
    status?: number;
}): void;
``` 

However, there are use cases where it would be beneficial to include additional custom data in the response **body**. 

This option is available in the [similar function](https://sdk.vercel.ai/docs/api-reference/stream-to-response) `experimental_StreamingReactResponse`  where you can send custom data in the body.  
```js
/**
 * A utility class for streaming React responses.
 */
declare class experimental_StreamingReactResponse {
    constructor(res: ReadableStream, options?: {
        ui?: (message: {
            content: string;
            data?: JSONValue[];
        }) => UINode | Promise<UINode>;
        data?: experimental_StreamData;
        generateId?: IdGenerator;
    });
}
```
However this doesn't work in Nodejs servers and works only on Vercel as far as I know.

### Use Case

It would be incredibly useful to piggyback on the final response and add custom data to the response body sent back to the client.  For example if the function can be modified to accept an options params with the following signature:
```js
declare function streamToResponse(res: ReadableStream, response: ServerResponse, options?: {
    init?: {
        headers?: Record<string, string>;
        status?: number;
    }
    data?: experimental_StreamData;
    generateId?: IdGenerator;
}): void;
```
Which in turn would be use like so:
```js
router.post('/chat/:id', async (req, res) => {
    const openai = new OpenAI();

    const { messages } = req.body;

    await data.append({
        customData: {
            messageCount: 1,
            id: '1',
        }
    });

    await data.close();

    try {
        const aiResponse = await openai.chat.completions.create({
          model: 'gpt-3.5-turbo',
          stream: true,
          messages,
        });

        const stream = OpenAIStream(aiResponse);

        return streamToResponse(stream, res, { data });
    } catch (error) {
        res.status(500).json({ error: 'Error forwarding request to OpenAI' });
    }
});
```

Currently struggling to find a way around this, and I really don't want to migrate to vercel just for this one feature.

### Additional context

_No response_

valstu:
This PR adds DataStream support to `streamToResponse`. Should fix #1285.

### DIFF
--- a/docs/pages/docs/api-reference/stream-to-response.mdx
+++ b/docs/pages/docs/api-reference/stream-to-response.mdx
@@ -8,7 +8,7 @@ import { OptionTable } from '@/components/table';
   [`web-streams-polyfill`](https://www.npmjs.com/package/web-streams-polyfill).
 </Callout>
 
-## `streamToResponse(stream: ReadableStream, response: ServerResponse, options?: Options)`
+## `streamToResponse(stream: ReadableStream, response: ServerResponse, options?: Options, data?: DataStream)`
 
 This method will pipe a `ReadableStream` to a Node.js `ServerResponse` object. It can be helpful to combine this with other AI stream utilities, such as [`OpenAIStream`](/docs/api-reference/providers/openai-stream), in Node.js environments.
 
@@ -42,6 +42,10 @@ Optional object to configure the response, with the following properties:
     ],
   ]}
 />
+
+### `data?: DataStream`
+
+Similar to example in [`StreamData`](docs/api-reference/stream-data#on-the-server), you can provide additional data here.
 
 ## Example: Node.js HTTP Server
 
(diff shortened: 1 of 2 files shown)


valstu:
Added prettier fixes and merged upstream

SamStenner:
I've wanted this for a while, thanks for doing it.

lgrammel:
Thanks! With our docs changes there were some merge conflicts, and I needed to make changes. Superseeded by #1615

PR #1457, opened 2024-04-26
Fix type inference for tool parameters in `streamUI`

unstubbable:
The first argument of the tool `generate` functions was implicitly `any`. By using the same trick as in the `render` function, we can fix the type inference based on the provided `parameters`.


shuding (approved):
Thank you!

PR #1458, opened 2024-04-26
ai/rsc: Improve `getAIState` and `getMutableAIState` types

shuding:
Makes sure that the returned state is correctly typed and readonly.


lgrammel (approved):

packages/core/rsc/ai-state.tsx
```diff
@@ -183,10 +185,10 @@ function getMutableAIState<AI extends AIProvider = any>(
             )}" field from the AI state because it's not an object.`,
           );
         }
-        return store.currentState[key];
+        return store.currentState[key] as Readonly<AIStateWithKey>;
```
shuding, line 188: There might be better ways to do this.

PR #1460, opened 2024-04-27
ai/rsc: ReadableStream as provider for streamable value; add `.append()` method

shuding:
`createStreamableValue(initialValue)` now accepts a `ReadableStream` type as its `initialValue`. This will be handy when used together with the `textStream` core API, and common Web streams (i.e. from `fetch()`). If the readable stream pipes chunk types other than strings, it will behave like `.update()`.

If a `ReadableStream` is provided, the streamable value will be locked and no longer updatable from the userland.

An `.append()` method is also added to align with the `createStreamableUI()` API. It accepts string values only but in the future we might extend it to allow arrays too.

### DIFF
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -21,7 +21,7 @@ import type { StreamablePatch, StreamableValue } from './types';
  * Create a piece of changable UI that can be streamed to the client.
  * On the client side, it can be rendered as a normal React node.
  */
-export function createStreamableUI(initialValue?: React.ReactNode) {
+function createStreamableUI(initialValue?: React.ReactNode) {
   let currentValue = initialValue;
   let closed = false;
   let { row, resolve, reject } = createSuspensedChunk(initialValue);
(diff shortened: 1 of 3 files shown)


packages/core/rsc/streamable.tsx
```diff
@@ -136,12 +136,72 @@ export function createStreamableUI(initialValue?: React.ReactNode) {
...
+
+        // Unlock the value to allow updates.
+        streamableValue[STREAMABLE_VALUE_INTERNAL_LOCK] = false;
+        if (typeof value === 'string') {
+          streamableValue.append(value);
+        } else {
+          streamableValue.update(value);
+        }
+        // Lock the value again.
+        streamableValue[STREAMABLE_VALUE_INTERNAL_LOCK] = true;
```
lgrammel, line 188: since you have a lock release, a `try...finally` pattern might be worth considering

packages/core/rsc/streamable.tsx
```diff
@@ -136,12 +136,72 @@ export function createStreamableUI(initialValue?: React.ReactNode) {
...
+      }
+
+      streamableValue[STREAMABLE_VALUE_INTERNAL_LOCK] = false;
+      streamableValue.done();
+    } catch (e) {
+      streamableValue[STREAMABLE_VALUE_INTERNAL_LOCK] = false;
+      streamableValue.error(e);
+    }
```
lgrammel, line 196: move setting `lock = false` into `finally`?

packages/core/rsc/streamable.tsx
```diff
@@ -235,6 +307,56 @@ export function createStreamableValue<T = any, E = any>(initialValue?: T) {
...
+     * ```
+     */
+    append(value: T) {
+      assertStream('.append()');
+
+      if (
+        typeof currentValue !== 'string' &&
+        typeof currentValue !== 'undefined'
+      ) {
+        throw new Error(
```
lgrammel, line 329: would it make sense to have special AI errors that we can link to docs in the future? might be something for later

lgrammel (approved):

PR #1462, opened 2024-04-28
fix : change createAndStream with stream

ravvi-kumar:
createAndStream has been deprecated by openAI and also i was having trouble with an old versoin.

image


lgrammel:
I've updated the assistant response as part of a different pull request. Is this still valid?

PR #1476, opened 2024-05-01
Update JSDoc for object mode.

lgrammel:


packages/core/core/generate-object/generate-object.ts
```diff
@@ -71,7 +71,16 @@ The schema of the object that the model should generate.
...
 
     /**
-The mode to use for object generation. Not all models support all modes.
+The mode to use for object generation.
+
+The Zod schema is converted in a JSON schema and used in one of the following ways
+
+- 'auto': The provider will choose the best mode for the model.
+- 'tool': A tool with the JSON schema as parameters is is provided and the provider is instructed to use it.
+- 'json': The JSON schema and a instruction is injected into the prompt. If the provider supports JSON mode, it is enabled.
```
bernaferrari, line 80: Enabled when provider supports JSON mode.

MISTYKITTY: Thanks

MISTYKITTY (commented):
Reviewed

packages/core/core/generate-object/generate-object.ts
lgrammel: @bernaferrari that's not quite what i meant to say. it's more of the lines of "the provider will activate any special json mode setting that it has".

PR #1478, opened 2024-05-01
Add optional metadata to base Message type

gtaylor5:
# Problem

While the base `Message` type is robust, there are cases in which users may want to include additional metadata with a message. For example, adding user attribution to a message if the chat / completion / assistant is being used in a shared context.

# Solution

We update the base `Message` type to include an optional `metadata` field to support this change.


lgrammel:
`annotations` serves this purpose already.

PR #1479, opened 2024-05-02
ai/rsc: Make RSC streamable utils chainable

shuding:
It's super handy to make RSC streamable primitives chainable to create values inline. For example, unify type definitions with inline streamable values:

```ts
if (loggedIn) {
  const userCard = createStreamableUI()

  // LLM logic to update `userCard`...

  return {
    display: userCard.value
  }
} else {
  return {
    display: createStreamableUI().done(<div>Please login</div>).value
  }
}
```

### DIFF
--- a/.changeset/plenty-onions-invite.md
+++ b/.changeset/plenty-onions-invite.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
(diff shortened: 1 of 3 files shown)


MaxLeiter (approved):

unstubbable:
@JClackett fixed in #1735

PR #1482, opened 2024-05-02
ai/rsc: `streamUI`

shuding:
This PR removes the `experimental_` prefix from `streamUI`. There might still be small tweaks in the future but the basic shape should be kept the same.

### DIFF
--- a/packages/core/rsc/streamable.tsx
+++ b/packages/core/rsc/streamable.tsx
@@ -413,7 +413,7 @@ type Renderer<T> = (
  * please use `streamUI` for compatibility with other providers.
  *
  * @deprecated It's recommended to use the `streamUI` API for compatibility with AI SDK Core APIs
- * and future features. This API will be deprecated in a future release.
+ * and future features. This API will be removed in a future release.
  */
 export function render<
   TS extends {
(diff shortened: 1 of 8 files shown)


packages/core/rsc/streamable.tsx
```diff
@@ -409,9 +409,11 @@ type Renderer<T> = (
 
 /**
  * `render` is a helper function to create a streamable UI from some LLMs.
- * Currently, it only supports OpenAI's GPT models with Function Calling and Assistants Tools.
+ * This API only supports OpenAI's GPT models with Function Calling and Assistants Tools,
+ * please use `streamUI` for compatibility with other providers.
  *
- * @deprecated It's recommended to use the `experimental_streamUI` API for compatibility with the new core APIs.
+ * @deprecated It's recommended to use the `streamUI` API for compatibility with AI SDK Core APIs
+ * and future features. This API will be deprecated in a future release.
```
unstubbable, line 416:
```suggestion
 * and future features. This API will be removed in a future release.
```

jeremyphilemon (approved):

packages/core/rsc/streamable.tsx
shuding: Thanks :)

PR #1524, opened 2024-05-08
feat (ai/ui): add stop() helper to useAssistant

lgrammel:
## Summary
- implement stop() helper for useAssistant
- add openai as optional peer dependency and require `v4.42+`
- update example and add stop/abort signals


packages/core/react/use-assistant.ts
```diff
@@ -240,14 +259,21 @@ export function useAssistant({
         }
       }
     } catch (error) {
+      // Ignore abort errors as they are expected.
+      if (isAbortError(error)) {
```
unstubbable, line 263: Could there be a situation where when the fetch is aborted, that `readDataStream` throws for some reason? If yes, might it be better to use `abortController.signal.aborted` as the early return condition to skip the error handling?

lgrammel: Good catch. Thanks

packages/core/react/use-assistant.ts
```diff
@@ -240,14 +261,21 @@ export function useAssistant({
         }
       }
     } catch (error) {
+      // Ignore abort errors as they are expected when the user cancels the request:
+      if (isAbortError(error) && abortController.signal.aborted) {
```
unstubbable, line 265: Hm, I meant it differently. This is now the same behaviour as before, I think. As in, if this is an abort error, the signal will also be aborted, so the second part of the condition is superfluous. Instead, I thought it might make sense to do the following, in order to ignore non-abort errors (e.g. from `readDataStream`) after the fetch was aborted.

```suggestion
      if (abortController.signal.aborted) {
```

unstubbable: Speaking of which, should this maybe be also passed in as an option?

```js
readDataStream(result.body.getReader(), { isAborted: () => abortController.signal.aborted }))
```

lgrammel: 1) it's slightly different. `isAbortError(error)` also recognizes TimeoutErrors, which can come from `AbortSignal.timeout()`. However, they could also be regular timeouts, which is why the `abortController.signal.aborted` check is important.
2) I'd prefer not to swallow any errors, in particular not proactively. Swallowed errors have burned me far too often. If certain errors happen (they should not), then it's important to investigate and understand first.

PR #1536, opened 2024-05-09
export rsc util types

wststone:
When creating reusable streaming ui tools, sometimes we have to seperate them in different files/modules that are not used directly within the `streamUI` function but they loses their typings. It would be nice to explicitly add types to those functions.


shuding (changes requested):
Makes sense. I think you'll have to edit the index file as well, otherwise the types won't be accessible.

wststone:
Added those exports in index file

packages/core/rsc/stream-ui/stream-ui.tsx
```diff
@@ -40,7 +40,7 @@ type RenderTool<PARAMETERS extends z.ZodTypeAny = any> = {
   >;
 };
 
-type RenderText = Renderer<
+export type RenderText = Renderer<
```
lgrammel, line 43: @shuding would renaming to `TextRenderer` make sense before exporting this?

PR #1559, opened 2024-05-10
svelte use assistant hook

jaycoolslm:
Implementation of useAssistant for Svelte based on the 

https://github.com/vercel/ai/blob/main/packages/core/react/use-assistant.ts directory

and following the same coding standards as found in the other svelte files

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'


lgrammel:
The lockfile changes are unnecessary. Please revert that part.

lgrammel:
Can you add an example under `examples/sveltekit-openai` similar to `examples/next-openai/app/assistant` and `examples/next-openai/app/api/assistant` so I can try it out?

yaqubhall:
Yep, will try get that over tonight

jaycoolslm:
Have added assistant page and api endpoint - its more similar to the current the svelte implementations and not as thorough as the nextjs. Low in time atm lol (aren't we all)... I ran `pnpm i` again and it did change the lock file, but not to the same extent as previous time so hopefully it's okay. Looks like it just bumped a few packages etc, pushing as a separate commit anyway so easy to revert

lgrammel:
@jaycoolslm can you see if you can resolve the conflicts? Usually a merge with the latest main + `pnpm i` fixes the lockfile conflicts

Issue #1050, opened 2024-03-02
docs/ai-rsc: `action is not a function`


alonzuman:
### Description

When using the example from `concepts/ai-rsc`, the following error is thrown:
<img width="1022" alt="image" src="https://github.com/vercel/ai/assets/57261095/e5ccfecf-c3a3-4198-bf71-9343179fe0ba">

### Code example
```typescript
import { OpenAI } from "openai";
import { createAI, getMutableAIState, render } from "ai/rsc";
import { z } from "zod";

const initialAIState: {
  role: "user" | "assistant" | "system" | "function";
  content: string;
  id?: string;
  name?: string;
}[] = [];

const initialUIState: {
  id: number;
  display: React.ReactNode;
}[] = [];

// AI is a provider you wrap your application with so you can access AI and UI state in your components.
export const AI = createAI({
  actions: {
    submitUserMessage,
  },
  // Each state can be any shape of object, but for chat applications
  // it makes sense to have an array of messages. Or you may prefer something like { id: number, messages: Message[] }
  initialUIState,
  initialAIState,
});

const openai = new OpenAI({
  apiKey: process.env.OPENAI_API_KEY,
});

async function submitUserMessage(userInput: string) {
  "use server";

  const aiState = getMutableAIState<typeof AI>();

  // Update AI state with new message.
  aiState.update([
    ...aiState.get(),
    {
      role: "user",
      content: userInput,
    },
  ]);

  // render() returns a stream of UI components
  const ui = render({
    model: "gpt-4-turbo-preview",
    provider: openai,
    messages: [
      { role: "system", content: "You are a flight assistant" },
      { role: "user", content: userInput },
    ],
    // `text` is called when an AI returns a text response (as opposed to a tool call)
    text: ({ content, done }) => {
      // text can be streamed from the LLM, but we only want to close the stream with .done() when its completed.
      // done() marks the state as available for the client to access
      if (done) {
        aiState.done([
          ...aiState.get(),
          {
            role: "assistant",
            content,
          },
        ]);
      }

      return <div>{content}</div>;
    },
    tools: {
      get_flight_info: {
        description: "Get the information for a flight",
        parameters: z
          .object({
            flightNumber: z.string().describe("the number of the flight"),
          })
          .required(),
        // flightNumber is inferred from the parameters passed above
        render: async function* ({ flightNumber }) {
          yield <>Loading...</>;
          const flightInfo = {
            flightNumber,
            departure: "SFO",
            arrival: "JFK",
[...]

PR #1585, opened 2024-05-14
fix: "action is not a function" when following ai/rsc demo code

mcavaliere:
Fixes the issue referenced here: 

https://github.com/vercel/ai/issues/1050#issuecomment-2110770005


MaxLeiter (approved):
thanks!

PR #1597, opened 2024-05-15
Deprecate render

jeremyphilemon:


content/docs/07-reference/ai-sdk-rsc/10-render.mdx
```diff
@@ -0,0 +1,20 @@
+---
+title: render
```
lgrammel, line 2: Would it make sense to have it say `render (deprecated)` (also in the h1 below) to make it really obvious?

lgrammel (approved):

content/docs/07-reference/ai-sdk-rsc/10-render.mdx
jeremyphilemon: I agree, done!

Issue #1500, opened 2024-05-06
Add ability to stream additional data immediately in `StreamData`


raivatshah:
### Feature Description

[StreamData](https://sdk.vercel.ai/docs/ai-sdk-ui/streaming-data) is great! What could be even better if there was some way to use the `data` object (or otherwise) to stream data at will -- at anytime basically.

### Use Case

 One big use is in consumer apps - when a tool call is going to take time, you want to indicate that to the client.

lgrammel:
Add better support in `v3.1.11`

raivatshah:
awesome thanks so much @lgrammel <3

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests expanded streaming behavior rather than reporting a failure of existing documented functionality.

### Evidence
- The issue requests the ability to send `StreamData` at any time, including while a long-running tool call is in progress.
- The maintainer stated that better support was added in v3.1.11, indicating this was a capability enhancement.

PR #1613, opened 2024-05-16
fix (ai/ui): decouple StreamData chunks from LLM stream

lgrammel:
## Problem
StreamData elements are currently tied to the LLM stream (TransformStream). This prevents them from being sent until the first token of the LLM stream arrives. However, users want to immediately send data message (see e.g. #1500 ).

## Summary
- Add `mergeStreams` function that merges two streams (with a slight priority for the first)
- Rewrite `StreamData` as a simple `ReadableStream` (before: `TransformStream`)
- Use `mergeStreams` to combine the `StreamData` stream and the LLM stream
- Update examples


doowb:
This looks great! I was just wondering if I could stream back data before the llm stream starts.

shuding (approved):
This implementation looks great, thanks!

PR #1614, opened 2024-05-16
refactoring (ai/core): use TextEncoderStream and toAIStream in streamText

lgrammel:


lgrammel:
> If https://vercel.com/docs/functions/runtimes/edge-runtime is up-to-date, this means that this drops support for running in an edge runtime, since `TextEncoderStream` does not seem to be available there, right? (Or rather, if true, this was already dropped with #1514 then, I guess.)

Thanks for catching this.

lgrammel:
> If https://vercel.com/docs/functions/runtimes/edge-runtime is up-to-date, this means that this drops support for running in an edge runtime, since `TextEncoderStream` does not seem to be available there, right? (Or rather, if true, this was already dropped with #1514 then, I guess.)

Just confirmed that it's supported on Vercel Edge. Also the edge tests were passing so it's all good.

PR #1617, opened 2024-05-16
feat (ai/core): add embedMany function

lgrammel:
## Tasks

- [x] implement `embedMany`
- [x] test `embedMany`
- [x] add examples
- [x] docs: add to core/embeddings
- [x] docs: add reference
- [x] docs: update provider pages
- [x] changeset


elitan:
maybe the word "batch" could be confusing here since OpenAI launched their [Batch API](https://platform.openai.com/docs/guides/batch).

PR #1618, opened 2024-05-16
Update: Correct server action to be called 01-streaming-user-interfac…

MrMaina100:
Wrong server action being called in the streaming user interfaces docs. Pretty sure it should be getWeather instead of runThread.


lgrammel (approved):
Thank you!

PR #1639, opened 2024-05-17
Add doRawStream

gaspar09:
Me messing around


lgrammel:
Ended up with middleware pattern instead.

PR #1653, opened 2024-05-19
Adding streaming tool use support for groq

colemccracken:


colemccracken:
@lgrammel sorry for the delay. I realized there were a couple issues. 

One is that Groq was sending an initial chunk with content: "", which was tripping up the sdk. Theres no reason for this and openai doesnt do it, so going to deploy this fix on Groq's side, should land tomorrow.

The second is that Groq sends the entire tool call payload in a single delta, which I realized wasnt supported in openai's handler. I pushed a commit to the ai sdk that handles this- but understand if you dont want to have this logic given this is an openai specific sdk. Let me know what you think.

Anyways, Ill add another comment on this pr once we've deployed the Groq fix to production and the example should work.

lgrammel:
Superseeded by https://github.com/vercel/ai/pull/1838

PR #1681, opened 2024-05-24
feat (ai/react): add experimental_maxAutomaticRoundtrips to useChat

lgrammel:
## Summary
This PR enables automatic roundtrips from client to server for tool calling. When all tool results are available, a new server call is initiated to call the language model with the tool results.

This architecture (going through the client) enables the unification of client tool calls and server tool calls.

## Tasks

- [x] implement experimental feature
- [x] add example
- [x] changeset

## Future Work
- optimize server side tool calls (roundtrip fully on server when there are no client side tool calls)
- Svelte, Vue support
- update docs (reference docs, dedicated tutorial page)

### DIFF
--- a/packages/core/react/use-chat.ts
+++ b/packages/core/react/use-chat.ts
@@ -219,6 +219,7 @@ export function useChat({
   sendExtraMessageFields,
   experimental_onFunctionCall,
   experimental_onToolCall,
+  experimental_maxAutomaticRoundtrips = 0,
   streamMode,
   onResponse,
   onFinish,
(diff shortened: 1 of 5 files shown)


packages/core/react/use-chat.ts
```diff
@@ -343,6 +357,20 @@ export function useChat({
       } finally {
         mutateLoading(false);
       }
+
+      // auto-submit when all tool calls in the last assistant message have results:
+      const lastMessage = messagesRef.current[messagesRef.current.length - 1];
```
shuding, line 362: `lastMessage` can't be undefined here, right?

shuding (approved):
Looks good!

packages/core/react/use-chat.ts
lgrammel: it should, bc it comes after a message has been received, but worth adding a check

PR #1692, opened 2024-05-26
Vue use assistant

dosstx:


dosstx:
@danielroe and @lgrammel : I am at a loss on how to proceed. Trying to get this update into the library but I keep stumbling. Anyone able to take it from here?

lgrammel:
@dosstx I'll work on integrating it this week

dosstx:
> @dosstx I'll work on integrating it this week

Thank you, I really appreciate it. I'll try and help with docs.

lgrammel:
I gave it a look today. The pattern is different from the implementation in `vue/use-chat.ts`, where we use `useSWRV`. I'm not familiar with `reactive`. My suggestion would be to change it to be similar with use-chat, i.e., use `useSWRV`. I suspect that this is the cause of the `Type instantiation is excessively deep and possibly infinite.` error, but haven't looked into the details.

Also, please add `UseAssistantHelpers` similar to `react/use-assistant.ts`

dosstx:
@danielroe Two questions:

1. Should we be using `swrv` in this assistant composable?

2. To fix the excessively deep type issue, should I replace `reactive` with `ref` in this case?

```
const state = reactive({
    messages: [] as Message[],
    input: '',
    threadId: undefined as string | undefined,
    status: 'awaiting_message' as AssistantStatus,
    error: undefined as unknown,
  });
  ```

lgrammel:
We ended up with this implementation: https://github.com/vercel/ai/pull/2245

PR #1700, opened 2024-05-27
fix (providers): make zod required dependency

lgrammel:
## Summary
* make zod a required dependency for all providers that use it for parsing
* remove zod dependency from vertex provider


MISTYKITTY (commented):
Reviewed

Issue #1270, opened 2024-04-03
SvelteKit "handleSubmit" omits data


zicho:
### Description

I cannot get "data" to get passed along to the server when using sveltekit. I am trying this:

I set the "useChat" up on my `+page.svelte`

```
const { input, handleSubmit, messages } = useChat();
```

My submit handler:

```
function submitHandler(event: any) {
  event.preventDefault();
  handleSubmit(event, {
    data: {
      test: 'data'
    }   
  });
}
```

And my server:

```
export const POST = (async ({ request }) => {
  try {
    const jsonData = await request.json();

    console.dir(jsonData); // "data" object is missing from here. i only have "messages"
```

Based on [this](https://github.com/vercel/ai/pull/725/files) React example I also tried:

```
<form
    on:submit={(e) => {
    handleSubmit(e, {
      data: {
        test: 'data'
      }
    });
  }}
>
```

but "data" is still not available in `+server.ts` endpoint

hauselin:
I also have this problem.

andrewrisse:
Same problem for me, looks like in the append function, 'data' is not destructured from the ChatRequestOptions and passed through.
```  const append: UseChatHelpers['append'] = async (
    message: Message | CreateMessage,
    {
      options,
      functions,
      function_call,
      tools,
      tool_choice,
    }: ChatRequestOptions = {},
  ) => {
    if (!message.id) {
      message.id = generateId();
    }

    const chatRequest: ChatRequest = {
      messages: get(messages).concat(message as Message),
      options,
      ...(functions !== undefined && { functions }),
      ...(function_call !== undefined && { function_call }),
      ...(tools !== undefined && { tools }),
      ...(tool_choice !== undefined && { tool_choice }),
    };
    return triggerRequest(chatRequest);
  };
  ```

lgrammel:
https://github.com/vercel/ai/pull/1748

ghoshsanjoy78:
I am still having this issue as well. Any fix ?

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This is a concrete failure to pass a supported request option through the existing SvelteKit chat API, rather than a request for new capability.

### Evidence
- SvelteKit useChat.handleSubmit does not include the supplied ChatRequestOptions.data in the request body; the server receives only messages.
- A commenter identified that append destructures several request options but omits data when constructing ChatRequest.
[...]

PR #1748, opened 2024-05-30
fix (ui/svelte): send data to server

lgrammel:


dalton5:
Hi. When will it be fixed please? Thanks

lgrammel:
@dalton5 this has already been merged and shipped

dalton5:
I still have the error where data is not sent in the body.

I am with sveltekit

` <form
    on:submit={(e) => {
      handleSubmit(e, { data: { model: selectedModel } });
      tick;
      scrollToBottom();
    }}
  >`

And in my server:

`export const POST = (async ({ request }) => {
  const val= await request.json();`

I only have messages in val. I should also have data. Related to this issue: https://github.com/vercel/ai/issues/1270

In my package.json: "ai": "^3.2.5",

ghoshsanjoy78:
I am still having this issue as well. Any fix

Issue #1754, opened 2024-05-30
OpenAI compatible API provider does not work (FriendliAI)


minpeter:
### Description

referring to [providers/ai-sdk-providers/fireworks](https://sdk.vercel.ai/providers/ai-sdk-providers/fireworks) and [docs/getting-started/nextjs-app-router](https://sdk.vercel.ai/docs/getting-started/nextjs-app-router), and got the following results.

```
    presence_penalty: undefined,
    seed: undefined,
    messages: [ [Object] ],
    tools: undefined,
    tool_choice: undefined,
    stream: true,
    stream_options: undefined
  },
  statusCode: 422,
  responseHeaders: {
    'access-control-allow-credentials': 'true',
    connection: 'keep-alive',
    'content-length': '107',
    'content-type': 'application/json',
    date: 'Thu, 30 May 2024 11:37:11 GMT',
    server: 'Orca',
    vary: 'Origin',
    'x-ratelimit-limit-requests': '10',
    'x-ratelimit-limit-tokens': '10000',
    'x-ratelimit-remaining-requests': '9',
    'x-ratelimit-remaining-tokens': '10000',
    'x-ratelimit-reset-requests': '60',
    'x-ratelimit-reset-tokens': '60'
  },
  responseBody: '{"message":"invalid JSON payload: (messages[0]) content: Proto field is not repeating, cannot start list."}',
  cause: undefined,
  isRetryable: false,
  data: undefined
}
 POST /api/chat 500 in 384ms
 ```
 
 In the latter case, it returns that the content of the messages field cannot begin with a list.
 
 I think this is a problem with the recently changed role of openai's messages field being changed to the form below instead of a string.
 
 ```json
"messages": [
      {
        "role": "user",
        "content": [
          {
            "type": "text",
            "text": "What'\''s in this image?"
          },
          {
            "type": "image_url",
            "image_url": {
              "url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg"
            }
          }
        ]
      }
    ],
```

[...]

PR #1755, opened 2024-05-30
Compatibility extension of openai's compatibility mode

minpeter:
There are two changes here:
1. When setting `compatibility: "compatible"` in `createOpenAI`, the messages[].content field in the `chat/completions` request has been changed to strings.
2. The `object` filter has been excluded from the response value filter of `chat.completions`.

Related issues: #1754 

This patch verifies that the examples provided through the FriendliAI endpoint work well.

And I'm not sure, but fireworks ai also didn't work before this patch, but after the patch it returns a response successfully.

Perhaps this patch will have a positive effect on the behavior of many other companies besides groq and Perplexity. 👍🏼


lgrammel (changes requested):
This removes the image parts in compatible mode and is a breaking change.

minpeter:
Would it be okay if I changed the logic to modify it to send a string if there are no image fields and it is in compatible mode?

minpeter:
I think it’s a good idea.

I‘ll post the correction when I get home.

minpeter:
I forgot to delete unused parameters.

lgrammel:
@minpeter thanks, this looks good. I'll take it from here, need to add tests & changeset so I'll continue in a separate PR

minpeter:
Question in a different context: Is there a business process that must be followed to add a provider to the AI ​​SDK document?

lgrammel:
> Question in a different context: Is there a business process that must be followed to add a provider to the AI ​​SDK document?

If you mean 3rd party providers, it depends. We will probably maintain the ones for the big cloud providers and large LLM providers in our repo (e.g. Bedrock, Azure OpenAI). For other providers you can publish them as community providers in your own repo and we are more than happy to list them under community providers, see https://sdk.vercel.ai/providers/community-providers

minpeter:
> > Question in a different context: Is there a business process that must be followed to add a provider to the AI ​​SDK document?
> 
> If you mean 3rd party providers, it depends. We will probably maintain the ones for the big cloud providers and large LLM providers in our repo (e.g. Bedrock, Azure OpenAI). For other providers you can publish them as community providers in your own repo and we are more than happy to list them under community providers, see https://sdk.vercel.ai/providers/community-providers

thanks :)

lgrammel:
Superseeded by https://github.com/vercel/ai/pull/1765

PR #1763, opened 2024-05-30
Allow all HeadersInit values for StreamingTextResponse and toTextStreamResponse

sshader:
`headers` in [`ResponseInit`](https://github.com/microsoft/TypeScript/blob/5df3a107c0e8626479dc3310a0b3da63bf55ed9d/src/lib/dom.generated.d.ts#L1707) has type [`HeadersInit`](https://github.com/microsoft/TypeScript/blob/5df3a107c0e8626479dc3310a0b3da63bf55ed9d/src/lib/dom.generated.d.ts#L28369) which can be a `Headers` instance, a Record, or an array of tuples.

Spreading `init?.headers` only works if it is a Record and drops the headers if the other formats are used.

Example of this behavior:
```
const initHeaders = [["init", "foo"]] // or new Headers({ init: "foo" })
const response = new Response(null, { headers: {...initHeaders, extra: "bar"} });
console.log(response.headers.get("init")) // null, but expected foo
console.log(response.headers.get("extra")) // "bar"
```


lgrammel:
Thanks! I'll build on this in a separate PR

lgrammel:
superseeded by https://github.com/vercel/ai/pull/1766

PR #1786, opened 2024-06-01
fix (rsc): Fix unsafe {} type in application code for StreamableValue

gclark-eightfold:
In application code, `StreamableValue` type from `ai/src` is reduced down to `{}` as a result of the tsconfig `stripInternal` option.

As mentioned [here](https://www.totaltypescript.com/the-empty-object-type-in-typescript),`{}` represents "any value except `null` and `undefined`", which makes the type essentially useless to import. It's also misleading because the type parameters aren't used anywhere in the resulting type, which means that `StreamableValue<string>` is equivalent to `StreamableValue<boolean>` in application code.

For example, in [examples/next-ai-rsc/app/stream-object-generate-itinerary.tsx](https://github.com/gclark-eightfold/ai/blob/0c408a3af027fd352f9ac52922d92bb85ea051cc/examples/next-ai-rsc/app/stream-object/generate-itinerary.tsx#L66-L72), if you make the following change, you will **not** get an error:
```diff
const initialUIState: {
  isGenerating: StreamableValue<boolean>;
  itineraryComponent: React.ReactNode;
} = {
-  isGenerating: createStreamableValue(false).value,
+  isGenerating: "THIS SHOULD BE A TYPE ERROR",
  itineraryComponent: null,
};
```


lgrammel:
@gclark-eightfold thanks for this fix! is there any way this can be tested automatically? (i know this tends to be difficult with types)

unstubbable:
You could write a test that consumes the types from `dist`, like this for example:

```ts
// packages/core/rsc/streamable-ui-types.test.ts
import type { StreamableValue } from './dist';

describe('StreamableValue type', () => {
  it('should yield a compiler error when assigning a wrong value', () => {
    const initialUIState: {
      isGenerating: StreamableValue<boolean>;
    } = {
      // @ts-expect-error
      isGenerating: 'THIS SHOULD BE A TYPE ERROR',
    };
  });
});
```

Without the fix from this PR, the `pnpm type-check` command would fail with `Unused '@ts-expect-error' directive.`.

gclark-eightfold:
@lgrammel Added a test using vitest's built-in [typechecking](https://vitest.dev/api/expect-typeof) suite and enabled the [config setting](https://vitest.dev/config/#typecheck).

I tested against main and verified they failed before, and succeeded after.

Only caveat is that since the `@internal`-marked properties only get removed in the compiled output, it needs to pull the types from `./dist`, which I hope isn't a problem for CI.

lgrammel:
@gclark-eightfold thanks this is awesome! There are some prettier issues. would you mind fixing them? https://github.com/vercel/ai/blob/main/CONTRIBUTING.md#fixing-prettier-issues

gclark-eightfold:
@lgrammel Should be good now 👍

lgrammel (approved):

lgrammel:
Thank you!

PR #1790, opened 2024-06-02
chore (docs): improve example content, highlighting & install snippets

agusmoles:
Hi team, I was going through the documentation of the Vercel AI SDK for RSC and I think these changes were wrong in the original documentation, as we were obtaining the data from the `getWeather` external function, destructuring the `content` but then not using the returned variables since the attributes are being hardcoded in the `return` statement of each one

Let me know if I got something wrong. Great tool and documentation btw! 😄 

EDIT: made some more updates, described below (and in the commit msgs)

### DIFF
--- a/content/docs/04-ai-sdk-rsc/01-overview.mdx
+++ b/content/docs/04-ai-sdk-rsc/01-overview.mdx
@@ -155,7 +155,7 @@ The **AI SDK RSC (`ai/rsc`)** takes advantage of RSCs to solve the problem of ma
 
 Rather than conditionally rendering user interfaces on the client based on the data returned by the language model, you can directly stream them from the server during a model generation.
 
-```tsx highlight="3,22-31,38" filename="app/action.ts"
+```tsx highlight="3,21-23,30" filename="app/action.ts"
 import { createStreamableUI } from 'ai/rsc'
 
 const uiStream = createStreamableUI();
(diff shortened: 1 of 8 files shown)


content/docs/04-ai-sdk-rsc/01-overview.mdx
```diff
@@ -175,17 +175,9 @@ const text = generateText({
...
         uiStream.done(
-          <WeatherCard
-            weather={{
-              temperature: 47,
-              unit: 'F',
-              description: 'sunny'
-              forecast,
-            }}
-          />
+          <WeatherCard weather={weather} />
```
lgrammel, line 180: this will affect the highlighting (line 158) which needs to be updated

lgrammel:
thank you!

content/docs/04-ai-sdk-rsc/01-overview.mdx
agusmoles: oh you are right, fixed

agusmoles:
I've also added a fix to remove an unused improt that might creates confusion (at least it did for me) about if that was needed or not and also typed the `weather` variable

agusmoles:
Added tabs to install the dependencies with the different package managers (to be consistent with other parts of the documentation like [https://sdk.vercel.ai/docs/ai-sdk-core/schemas-and-zod](https://sdk.vercel.ai/docs/ai-sdk-core/schemas-and-zod))

lgrammel (approved):
Thank you, this is very helpful!

PR #1792, opened 2024-06-02
fix (docs): Add missing content for nuxt example

100gle:
As [the documentation](https://nuxt.com/docs/api/composables/use-runtime-config#define-runtime-config) points out, we should configure the environment variables in `nuxt.config.ts` before using `useRuntimeConfig`.

If we don't do that, the environment variables aren't accessible.


MaxLeiter (approved):

lgrammel:
@100gle thanks for your contribution! there are some prettier issues. would you mind fixing them?

guide: https://github.com/vercel/ai/blob/main/CONTRIBUTING.md#fixing-prettier-issues

PR #1825, opened 2024-06-04
fix(ai/rsc): New `createStreamableUI` implementation

### LINKED ISSUE #1850: Client Component in streamUI keeps re-mounting
### Description

1. follow example from doc
https://sdk.vercel.ai/examples/next-app/interface/route-components
2. create a client component
```tsx
'use client'
import React, { useEffect, useState } from 'react'

const TestClient = ({ content }: { content: string }) => {
  const [count, setCount] = useState(0)

  useEffect(() => {
    console.log('mount')
    return () => {
      console.log('unmount')
    }
  }, [])

  return content
}

export default TestClient
```
3. use client component to render text 
```diff ts 
  text: ({ content, done }) => {
      if (done) {
        history.done((messages: ServerMessage[]) => [
          ...messages,
          { role: 'assistant', content },
        ]);
      }

-      return <div>{content}</div>;
+      return <TestClient content={content} />
    },
```
4. check console log, see `mount` and `unmount` printed while text is treaming

### Code example

_No response_

### Additional context

I understand that this might not be related to reconciliation, but I would expect to not remount and just change the prop value if that's possible.
https://github.com/vercel/ai/pull/1818

shuding:
Instead of using nested Suspense wrappers, I realized that we can just pass a streamable value of the UI into a Client Component and do the update on the client. This has a lot of advantages, and the major one is that we no longer need to express the UI update logic declaratively via JSX, and React will be able to correctly reconcile the component updates.

Since `createStreamableUI` also powers `streamUI`, that API will benefit from this too.

Also moved unit tests of `streamUI` to E2E tests.

Closes #1850.


jeremyphilemon (approved):

packages/core/rsc/streamable.tsx
```diff
@@ -377,31 +356,55 @@ function createStreamableValueImpl<T = any, E = any>(initialValue?: T) {
...
-          `.append(): The current value is not a string. Received: ${typeof currentValue}`,
-        );
-      }
-      if (typeof value !== 'string') {
+      if (typeof currentValue === 'undefined') {
+        if (typeof value === 'string' || isValidElement(value)) {
+          currentPatchValue = undefined;
+          currentValue = value;
+        } else {
+          throw new Error(
```
lgrammel, line 364: I have introduced dedicated errors so we can have error pages for troubleshooting, e.g. https://sdk.vercel.ai/docs/troubleshooting/ai-sdk-errors/ai-invalid-data-content-error

do you think it's worth adding dedicated errors here?

packages/core/rsc/streamable.tsx
```diff
@@ -377,31 +356,55 @@ function createStreamableValueImpl<T = any, E = any>(initialValue?: T) {
...
+            <>
+              {currentValue}
+              {value}
+            </>
+          );
+        } else {
+          throw new Error(
+            `.append(): The value type can't be appended to the stream. Received: ${typeof value}`,
+          );
+        }
```
lgrammel, line 398: the check for `!(string | validelement)` is duplicated 3 times. you could invert the if-else and pull it out

packages/core/tests/e2e/next-server/app/streamable/actions.jsx
```diff
@@ -7,22 +7,35 @@ function sleep(ms = 0) {
...
+    streamable.done();
+  })();
+  return streamable.value;
+}
+
 export async function streamableUI() {
   const streamable = createStreamableUI();
   (async () => {
-    await sleep();
+    await sleep(10);
```
lgrammel, line 26: how stable is the test with this low timeout?

lgrammel (approved):

packages/core/rsc/streamable.tsx
shuding: Yes! We should convert all the common errors to dedicated error codes, with clear example and explanation.

packages/core/rsc/streamable.tsx
shuding: Yeah part of it is to bypass TS checks as it's not that intelligent to handle the branching if `typeof value` is stored as a variable.

packages/core/tests/e2e/next-server/app/streamable/actions.jsx
shuding: Quite stable. Ran this many times on local/CI and the result is consistent. Here it could be `sleep(0)` and it will still work, but decided to add a small delay just in case one layer (Next.js, RSC, React, browser) batches the chunks.

gaetansnl:
@shuding @lgrammel Hello, is it expected to change the behavior of createStreamableUI ? Seems that mixing `append` and `update` doesn't work anymore, Update override the appended elements ?

shuding:
@gaetansnl That sounds like a bug to me and I'll take a look!

lgrammel:
Reverted because of #1933 and #1969

salmenus:
Perfect timing ! I need it for my RSC / gen ui library 🚀

PR #1843, opened 2024-06-05
Add basic next pages examples

jeremyphilemon:


content/examples/02-next-pages/01-basics/01-generating-text.mdx
```diff
@@ -0,0 +1,72 @@
+---
+title: Generating Text
+description: Learn to generate text using the Vercel AI SDK in your Next.js application
```
lgrammel, line 3: wondering how we can describe the content more accurately (since it's used in search engines). Maybe "Example: Generating text with the Next.js pages router and the Vercel AI SDK"?

content/examples/02-next-pages/01-basics/01-generating-text.mdx
```diff
@@ -0,0 +1,72 @@
+---
+title: Generating Text
```
lgrammel, line 2: for seo, wondering if "Generating Text (Next.js Pages Router)" might be better (also heading below)

lgrammel (approved):

content/examples/02-next-pages/01-basics/01-generating-text.mdx
jeremyphilemon: Changes here will reflect in the sidebar unfortunately, but I think we can append the framework by tweaking the metadata config in the layout, will dig deeper!

PR #1883, opened 2024-06-07
Over 100x faster streaming JSON parsing

AaronFriel:
Replaces the partial JSON parser with a streaming JSON parser.

```
 ✓ core/util/parser-comparison.bench.ts (2) 5192ms
     name                                hz     min     max    mean     p75     p99    p995    p999     rme  samples
   · parsePartialJson - 5306 tokens  3.7921  261.20  268.88  263.70  264.90  268.88  268.88  268.88  ±0.72%       10
   · streaming parser - 5306 tokens  391.23  2.3913  3.3182  2.5560  2.6857  3.0767  3.3182  3.3182  ±0.82%      196   fastest
```

This parser has been used in production for ~6 months and is extensively tested, though the API would likely benefit from some changes to better support the Vercel AI SDK.

There are a few benefits to this parser that should be implemented as a fast follow.

First, functionally, the AI SDK could offer a stream of patches rather than a stream of objects. This is exceptionally useful when the stream is processed server-side, and the stream is intended to pass only a subset of fields through to a client or to perform pre-processing. These ParseEvents can be used to efficiently determine which paths are modified.

Second, the parser is more efficient than the current parser in time and space, and there are ways this could be improved.

1. The work of the parser is O(n) over the input stream, rather than O(n^2) over the input stream. A 1000 token stream will parse 1000 tokens with the `StreamingParser` from `fn-stream`, rather than ~1.5 million tokens `parsePartialJson` must parse over the same stream. (`parsePartialJson` parses each prefix thrice.)
2. It is no longer necessary for the server to maintain the accumulated chunks or previously parsed objects. This reduces memory demands on the server by O(n).
3. **Edit: this is done in the most recent PR.** A future commit could remove the call to `structuredClone` and reduce the number of copies of the object again. The tests should be updated to simulate a serialize-deserialize operation and introduce asynchrony. The tests currently provide all of the chunks synchronously to `streamObject` and as the tests simulate a client and server in the same process, the enqueued object is shared, so the assertions only see the final state. This would remove one more copy, which becomes accidentally quadratic over the stream.


packages/core/core/generate-object/stream-object.ts
```diff
@@ -361,24 +359,27 @@ Response headers.
...
-
-              controller.enqueue({ type: 'object', object: currentObject });
+            let { events } = streamingParser.parseIncremental(chunk);
+
+            // filter for events that change values
+            // @ts-expect-error TODO: fix type
+            events = events.filter(event => event.kind === 'partial' || (event.kind === 'complete' && typeof event.value !== 'object'))
+            if (events.length > 0) {
+              // @ts-expect-error
+              latestObject = structuredClone(streamingParser.root);
```
AaronFriel, line 378: The current `StreamingParser` does not expose its internal state, this `root` is the mutable partially parsed object.

This could be exposed as a `readonly DeepPartial<T>`.

Alternatively, an `applyEvents` function could maintain a parallel and safe external representation of the object. Without much more work, the object returned could be treated as a persistent data structure, creating new objects for each parent of a modified leaf.

packages/core/core/generate-object/stream-object.ts
```diff
@@ -361,24 +359,27 @@ Response headers.
...
-
-            if (!isDeepEqualData(latestObject, currentObject)) {
-              latestObject = currentObject;
-
-              controller.enqueue({ type: 'object', object: currentObject });
+            let { events } = streamingParser.parseIncremental(chunk);
+
+            // filter for events that change values
+            // @ts-expect-error TODO: fix type
+            events = events.filter(event => event.kind === 'partial' || (event.kind === 'complete' && typeof event.value !== 'object'))
```
AaronFriel, line 375: The `ParseEvent<T>` is richly typed, though TypeScript's narrowing is not quite powerful enough without a bit of help to provide the right type, though perhaps the current type could be updated to help TSC perform that narrowing.

The `ts-patterns` library does do this, and shows what should be possible, as the type of `event.value` here is narrowed typed:

https://github.com/AaronFriel/fn-stream/blob/a208bf2f5dd686cf4e6e4f5c78d8e45e445daa7f/packages/fn-stream/examples/openai-reducer.spec.ts#L108-L128

AaronFriel:
@shuding @lgrammel this may be of interest in driving down the compute overhead on streaming objects, and I'm happy to contribute fn-stream into this repository if you think it makes sense, or make API changes and drive fn-stream to a stable release.

lgrammel:
@AaronFriel thanks for opening the pull request! having a streaming parser would be great. I have some concerns though about introducing a dependency on `fn-stream`. Would you be open to adding the streaming parser to this repo?

AaronFriel:
Yeah.

AaronFriel:
@lgrammel How would you like to handle it? `fn-stream` uses maybe a different testing architecture (ava, property testing) and maybe is a bit "weird", I'm sure someone on Vercel's AI team might want to weigh in on the API.

lgrammel:
@AaronFriel yes, i would want to test with vitest and also look into the details, i.e. re-implement here. the reason is that i want to be able to quickly fix issues as they arise and reduce any supply chain risk. would you be open to collaborating on re-implementing it here? if so I could take a first look and come up with a draft.

AaronFriel:
Yeah absolutely. Does vitest have a property testing integration for fast check? If so it should be straightforward to port the testing.

lgrammel:
> Yeah absolutely. Does vitest have a property testing integration for fast check? If so it should be straightforward to port the testing.

I found this: https://github.com/dubzzz/fast-check/tree/main/packages/vitest#readme

AaronFriel:
@lgrammel I've updated my PR adding the fn-stream package. It looks like since I made my first commit there's been some refactoring into `@ai-sdk` - feel free to reorganize accordingly. 

I'd prefer, even if renamed, that this this parser stand alone as a separate package. Benchmarks added to vite bear out the claims and I've updated the PR name accordingly.

lgrammel:
@AaronFriel thanks Aaron this is awesome! I hope to find the time next week to go through in detail and learn about how it works

AaronFriel:
Retracting this PR per comment from @lgrammel. https://github.com/vercel/ai/issues/2036#issuecomment-2187077097

Of course the library and this contribution are permissively licensed, but it would be merged over my objection. 

This feels like a big missed opportunity for Vercel to demonstrate orders of magnitude better performance on AI workloads across the rich set of AI providers available, and unlocks dramatically improved performance from client UI framework, tool call decoding, network transfer.

CC @jaredpalmer

PR #1884, opened 2024-06-07
feat (core): add experimental OpenTelemetry support for generateText and streamText

lgrammel:
## Tasks

- [x] implementation
  - [x] `generateText`
    - [x] tool calls
  - [x] `streamText`
    - [x] basic implementation
    - [x] event for stream start
    - [x] tool calls
- [x] tests
  - [x] `generateText`
  - [x] `streamText`
- [x] examples
   - [x] node.js
   - [x] next app router / vercel
      - [x] basic example
      - [x] readme
- [x] docs
   - [x] guide (similar to next.js observability)
   - [x] api reference docs updates
- [x] changeset

## Future Work
- `generateObject` support
- `streamObject` support
- `streamUI` support
- `embed` support
- `embedMany` support


examples/ai-core/src/telemetry/telemetry.ts
```diff
@@ -0,0 +1,30 @@
...
+import { generateText } from 'ai';
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+import { NodeSDK } from '@opentelemetry/sdk-node';
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node';
+import { getNodeAutoInstrumentations } from '@opentelemetry/auto-instrumentations-node';
+
+const sdk = new NodeSDK({
```
dvoytenko, line 11: I'd just recommend installing and using `@vercel/otel` instead of NodeSDK.

packages/core/core/telemetry/get-tracer.ts
```diff
@@ -0,0 +1,9 @@
+import opentelemetry from '@opentelemetry/api';
+
+// async to support dynamic imports / stubbing
+export async function getTracer() {
```
dvoytenko, line 4: Not a good idea to make it async. Tracing should equally support sync and async call tracing. If you want dynamic import - you should try to arrange it some other way.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -107,77 +108,87 @@ By default, it's set to 0, which will disable the feature.
...
-    toolCalls: currentToolCalls,
-    toolResults: currentToolResults,
-    finishReason: currentModelResponse.finishReason,
-    usage: calculateTokenUsage(currentModelResponse.usage),
-    warnings: currentModelResponse.warnings,
-    rawResponse: currentModelResponse.rawResponse,
-    logprobs: currentModelResponse.logprobs,
-    responseMessages,
+    } finally {
+      span.end();
```
dvoytenko, line 190: It really helps to also do `catch() { span.recordException(error); span.setStatus(error)}`.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -107,77 +108,87 @@ By default, it's set to 0, which will disable the feature.
...
-    currentModelResponse = await retry(() => {
-      return model.doGenerate({
-        mode,
-        ...callSettings,
-        // once we have a roundtrip, we need to switch to messages format:
-        inputFormat: roundtrips === 0 ? validatedPrompt.type : 'messages',
-        prompt: promptMessages,
-        abortSignal,
+  const tracer = await getTracer();
+  return tracer.startActiveSpan('generateText', async span => {
```
dvoytenko, line 112: Name should probably be a bit more indicative. E.g. `ai.generateText`? Also, good idea to add a bunch of attributes, including `model`, `size-of-prompt`, etc.

examples/ai-core/src/telemetry/telemetry.ts
lgrammel: it's an example for testing. I am planning a separate example with `@vercel/otel`

packages/core/core/telemetry/get-tracer.ts
lgrammel: dynamic imports need to happen in async functions. it's not an issue because all AI SDK functions that we want to trace are async, unless i'm missing something?

examples/ai-core/src/telemetry/telemetry.ts
dvoytenko: Ok. But then let's definitely add that example. I'd really like to avoid people copy-pasting code that we know already won't work when deployed on Vercel.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -106,79 +110,165 @@ case of misconfigured tools.
...
+    telemetry?: {
+      functionId?: string;
+      metadata?: Record<string, AttributeValue>;
+    };
+  }): Promise<GenerateTextResult<TOOLS>> {
+  const baseTelemetryAttributes = {
+    'ai.model.provider': model.provider,
+    'ai.model.id': model.modelId,
+    // settings:
+    'ai.settings.maxRetries': maxRetries,
```
dvoytenko, line 128: Conventionally it should be `max_retries`. I know that this is annoying. Metadata is ok with camel-case since it's a user input. You can certainly break the custom here, this is just for context.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -106,79 +110,165 @@ case of misconfigured tools.
...
+    }, {} as Record<string, AttributeValue>),
+    // special telemetry information
+    'ai.telemetry.functionId': telemetry?.functionId,
+    // add metadata as attributes:
+    ...Object.entries(telemetry?.metadata ?? {}).reduce(
+      (attributes, [key, value]) => {
+        attributes[`ai.telemetry.metadata.${key}`] = value;
+        return attributes;
+      },
+      {} as Record<string, AttributeValue>,
```
dvoytenko, line 141: Nit, `Record<string, AttributeValue` is the `Attributes` type from OTEL API.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -106,79 +110,165 @@ case of misconfigured tools.
...
+    'ai.model.provider': model.provider,
+    'ai.model.id': model.modelId,
+    // settings:
+    'ai.settings.maxRetries': maxRetries,
+    ...Object.entries(settings ?? {}).reduce((attributes, [key, value]) => {
+      attributes[`ai.settings.${key}`] = value;
+      return attributes;
+    }, {} as Record<string, AttributeValue>),
+    // special telemetry information
+    'ai.telemetry.functionId': telemetry?.functionId,
```
dvoytenko, line 134: Ideally we'd also add `resource.name` and `operation.name`. Some OTEL providers, especially DataDog, are very good at using these. One idea I have here:

```
'operation.name': 'ai.generateText'
'resource.name': telemetry?.functionId ?? 'something_something'
```

packages/core/core/telemetry/get-tracer.ts
dvoytenko: So, sync now?

packages/core/core/telemetry/record-span.ts
```diff
@@ -0,0 +1,27 @@
...
+      return await fn(span);
+    } catch (error) {
+      if (error instanceof Error) {
+        span.recordException({
+          name: error.name,
+          message: error.message,
+          stack: error.stack,
+        });
+      }
+      span.setStatus(2 as any); // SpanStatus.ERROR
```
dvoytenko, line 20: That's not going to work. I believe the actual API is something like `span.setStatus({code: 2, message: ...})`

dvoytenko (changes requested):

packages/core/core/telemetry/get-tracer.ts
lgrammel: need to explore if there is any impact when using the functions in the browser (i.e. when certain node apis aren't available) - there are now e.g. chrome ai providers that do not need a backend

packages/core/core/generate-text/generate-text.ts
lgrammel: I changed it and then reverted back. I think it makes sense to break the custom here for 2 reasons (unless it's a big concern):

1) easier code search since the rest of our code is in camelCase
2) smaller bundle size (we'd need a 'toSnakeCase' function otherwise)

packages/core/core/telemetry/get-tracer.ts
lgrammel: changed to sync

examples/next-openai-telemetry/instrumentation.ts
```diff
@@ -0,0 +1,9 @@
+import { registerOTel } from '@vercel/otel';
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node';
+
+export function register() {
+  registerOTel({
+    serviceName: 'next-app',
+    traceExporter: new ConsoleSpanExporter(), // TODO what exporter should I use here?
```
dvoytenko, line 7: For `@vercel/otel` you don't need an exporter - it configures automatically. Just `serviceName` is needed.

examples/next-openai-telemetry/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@ai-sdk/openai": "latest",
+    "@ai-sdk/react": "latest",
+    "@opentelemetry/api-logs": "0.52.1",
```
dvoytenko, line 14: Do you actually need these deps? I think only `@opentelemetry/api` needed, if anything at all.

packages/core/core/generate-text/generate-text.ts
```diff
@@ -108,80 +112,173 @@ case of misconfigured tools.
...
+    'ai.model.id': model.modelId,
+    // settings:
+    'ai.settings.maxRetries': maxRetries,
+    ...Object.entries(settings ?? {}).reduce((attributes, [key, value]) => {
+      attributes[`ai.settings.${key}`] = value;
+      return attributes;
+    }, {} as Record<string, AttributeValue>),
+    // special telemetry information
+    'operation.name': 'ai.generateText',
+    'resource.name': telemetry?.functionId,
```
dvoytenko, line 142: Ideally there should be some value for `resource.name`. Is there any good default when `telemetry?.functionId` is falsy?

packages/core/core/telemetry/get-tracer.ts
```diff
@@ -0,0 +1,23 @@
+import opentelemetry, { Tracer } from '@opentelemetry/api';
```
dvoytenko, line 1: Do `import { traceApi, Tracer } from '@opentelemetry/api'` to use `traceApi.getTracer(...)`. I believe this has a better chance of tree shake more code out of the binary.

packages/core/core/telemetry/noop-tracer.ts
```diff
@@ -0,0 +1,69 @@
+import { Span, SpanContext, Tracer } from '@opentelemetry/api';
+
+/**
+ * Tracer implementation that does nothing (null object).
+ */
+export const noopTracer: Tracer = {
```
dvoytenko, line 6: What's this for?

dvoytenko (changes requested):

examples/next-openai-telemetry/package.json
lgrammel: There were complaints about missing peer dependencies with pnpm (apparently not automatically installed), so I added this.

packages/core/core/generate-text/generate-text.ts
lgrammel: not really - it's a stateless function call. The arguments (prompt) are variable, so even hashing would not work, and the model is likely used in other places as well. I can set it to `"unknown"` but all those solutions mean that it'll be hard to distinguish different generateText locations without a functionId.

packages/core/core/telemetry/noop-tracer.ts
lgrammel: When the tracing is disabled, I use a `noopTracer` (null object pattern) so I don't have to change the main code in many places.

examples/next-openai-telemetry/instrumentation.ts
lgrammel: How can I generate console output for checking in my local deployment without this?

dvoytenko: Hmm. That just doesn't really come up. Console output for tracing is just too spammy. I can potentially arrange something if it's very important. Ping me to discuss async.

packages/core/core/generate-text/generate-text.ts
dvoytenko: Ok. Let's keep undefined then.

packages/core/core/telemetry/noop-tracer.ts
dvoytenko: I see. Probably ok then. The `noopTracer` just looks exactly like the `noopTracer` in `@opentelemetry/api`, but unfortunately I don't think it's exported.

dvoytenko (approved):

examples/next-openai-telemetry/instrumentation.ts
lgrammel: Maybe it's specific to my own use case (Otel development). It'll be rare that someone is working on OTel itself. I'll remove it.

packages/core/core/telemetry/noop-tracer.ts
lgrammel: If it's exported that would be helpful (in case that's possible).

content/docs/03-ai-sdk-core/60-telemetry.mdx
```diff
@@ -0,0 +1,113 @@
...
+The Vercel AI SDK uses [OpenTelemetry](https://opentelemetry.io/) to collect telemetry data.
+OpenTelemetry is an open-source observability framework designed to provide
+standardized instrumentation for collecting telemetry data.
+
+## Enabling telemetry
+
+For Next.js applications, please follow the [Next.js OpenTelemetry guide](https://nextjs.org/docs/app/building-your-application/optimizing/open-telemetry) to enable telemetry first.
+
+You can then use the `experimental_telemetry` option to enable telemetry on specific function calls while
+the feature is experimental:
```
jeremyphilemon, line 22: possible to use a different term than "function calls" here? had me confused for a bit thinking it was about llm "function calls"

PR #1888, opened 2024-06-08
fix (docs): Fix wrong default port for nuxt example

100gle:
The vite default port is 5173, but not nuxt. The right [default port](https://nuxt.com/docs/api/nuxt-config#port) for nuxt is 3000.


lgrammel (approved):

lgrammel:
Thank you!

Issue #1894, opened 2024-06-09
streamUI doesn't support multiple parallel tool calls


mfclarke-cnx:
### Feature Description

Allow streamUI to gracefully handle multiple parallel tool calls. Currently when the model makes multiple tool calls, the tool handler gets called multiple times, but the first call to resolve closes the UI steam and the rest throw:

```
⨯ node_modules/.pnpm/ai@3.1.26_openai@4.48.1_react@18.3.1_solid-js@1.8.17_svelte@4.2.17_vue@3.4.27_zod@3.23.8/node_modules/ai/rsc/dist/rsc-server.mjs (1231:12) @ assertStream
 ⨯ unhandledRejection: Error: .update(): UI stream is already closed.
 ```

### Use Case

Any feature where the model can call multiple tools in parallel. For example, creating multiple things, searching multiple websites, updating multiple things.

camwest:
It looks like this hardcoded "true" is where the error is happening.

https://github.com/vercel/ai/blob/f7002ad2c5aa58ce6ed83e8d31fe22f71ebdb7d7/packages/core/rsc/stream-ui/stream-ui.tsx#L303

It's a bad assumption. The tool-call event doesn't immediately mean it's the last render call because there may be multiple tool-call events.

camwest:
A potential fix would be to add a 'all-tool-calls-done' event which gets pushed to stream after the loop right around here:

https://github.com/vercel/ai/blob/f7002ad2c5aa58ce6ed83e8d31fe22f71ebdb7d7/packages/openai/src/openai-chat-language-model.ts#L343

Then we handle that in stream-ui and finalize the stream there.

The challenge is we'd have to update anything that implements LanguageModelV1 to trigger this event, or version to V2.

camwest:
Another option is to close the stream on finish:

https://github.com/vercel/ai/blob/f7002ad2c5aa58ce6ed83e8d31fe22f71ebdb7d7/packages/openai/src/openai-chat-language-model.ts#L382-L388

Currently not implemented in streamUI:

https://github.com/vercel/ai/blob/f7002ad2c5aa58ce6ed83e8d31fe22f71ebdb7d7/packages/core/rsc/stream-ui/stream-ui.tsx#L314

bneigher:
ah this is the culprit of my issue! been doing this since `v3.1.25`

camwest:
@mfclarke-cnx can you change this to a bug instead of enhancement?

[...]

Issue #1532, opened 2024-05-08
AI State not updated during iterations


fran-cadenas-fu:
### Description

Hello,

I'm currently trying to use the new streamUI feature to create a chat that keeps asking questions on a certain topic and displays Kick response buttons to de user through tools.

I can't get it to work because the AIState keeps getting reseted to the one in the first iteration. I'm not sure that this is a problem with the functions memoization or something.

### Code example
```tsx
async function submitUserMessage(userInput: string): Promise<UIMessage> {
  "use server";

  console.log("submitUserMessage", userInput);
  const history = getMutableAIState<typeof AI>();

  const ui = await streamUI({
    model: openai("gpt-4-turbo"),
    system: systemPrompt,
    messages: history.get().messages as CoreMessage[],
    text: async ({ content, done }) => {
      // When it's the final content, mark the state as done and ready for the client to access.
      if (done) {
        history.done({
          ...history.get(),
          messages: [
            ...history.get().messages,
            {
              role: "assistant",
              content,
            },
          ],
        });

      return <ChatAssistantMessage content={content} />;
    },
    tools: {
      show_quick_reply_question: {
        description: "Shows the ui for the user to quick reply",
        parameters: z
          .object({
            question: z.string().describe("the question to show"),
            options: z.array(z.string()).describe("the options to show"),
          })
          .required(),
        generate: async function ({ question, options }) {
          console.log(
            "show_quick_reply_question messages",
            history.get().messages
          );

          history.done({
            ...history.get(),
            messages: [
              ...history.get().messages,
              {
                role: "function",
                name: "show_quick_reply_question",
                content: JSON.stringify({ question, options }),
              },
            ],
          });

          return [
            <ChatAssistantMessage content={question} className="mt-2" />,
            <div className="flex flex-col gap-2 mt-2">
              {options.map((option) => (
                <ChatQuickReply key={option} chatQuickReply={option} />
              ))}
            </div>,
          ];
        },
      },
    },
  });
}
```

### Additional context

When the conversations goes through the "text" streaming generation the state seems to be updated sucessfully. But in the case of rendering the questions through the tool 'show_quick_reply_question' (that is what the context says so the assistant uses this tool), after pressing the Button options in the UI a new message is submitted and in that iteration the history that is accessed is not updated and it looks like the first iterations with only 1 initial message and the last ui interface message.

unstubbable:
[...]

Issue #1895, opened 2024-06-09
Support tool call round trips in streamUI


mfclarke-cnx:
### Feature Description

Currently only supported in `generateText`: https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling#tool-roundtrips

Support in streamUI would be great. Currently I'm working around this by returning a component at the completion of the tool call that makes a new user request asking the model to continue. This isn't great because I have to account for this workaround in other areas (restoring saved chats etc)

### Use Case

Any tool call that doesn't require the user to confirm. For example, looking up information to use in a response.

Floriferous:
This is crucial if you want to have a real conversation with an AI that can call tools. Right now I get the toolCall result back, and then I can't chat anymore.

Here's my conversation:

```
Me: Hi, what can you do?
AI: Hi, I can do X
Me: Ok please do X
AI: toolcall.result = Y // I would expect some commentary to come with the result here
Me: Ok thanks
AI: toolcall.result = Y
Me: What do you think about the result?
AI: toolcall.result = Y
```

How do you go around this behavior?

benjaminalgreen:
This is possible using frontend tool calls, but tool calls for streamed responses is supposed to be coming soon. https://github.com/vercel/ai/discussions/1574#discussioncomment-9593005

nsenthilkumar:
while a QoL update to automate round trips would be great for right now you can recursively pipe tool call responses back into your submitMessage() yourself 

```
func myTool(){
     do stuff
     generate: return{submitMessage(toolResult)}
```

kevinschaich:
Also interested in this – one thought is maybe the `streamUI` API would benefit from keeping structured tool output separate from the desired UI (`generate` function in most of the examples).

Manually piping tool output back in via `submitMessage` as mentioned by @nsenthilkumar is interesting, but wouldn't that only provide additional context to the _following_ user message in the chat history? It doesn't direct the assistant to take another step after the current tool run completes in order to make incremental progress using multiple tool invocations against the original prompt. If I'm missing something I'd love to see a more complete example.

Open to ideas but something like this would provide a history of previously called tool invocations similar to `generateText`:

```typescript
tools: {
    getWeather: {
        description: 'Get Weather',
        parameters: z.object({
            city: z.string().describe('The city to get weather for'),
        }),
        generate: async function* ({ city }) {
            const weatherForCity = getWeather(city)
            yield (
                {
                    output: weatherForCity,
[...]

PR #1904, opened 2024-06-10
feat (rsc): multiple tool calls for streamUI

gclark-eightfold:
# Summary
Adds the ability to invoke multiple tool calls for `streamUI`.

Addresses https://github.com/vercel/ai/issues/1894, https://github.com/vercel/ai/issues/1532#issuecomment-2122232969

**NOTE: For usage with `aiState`, it's important to call `aiState.done(...)` only in the `onFinish` callback (if expecting multiple tool calls).** -- Not sure if this is a breaking change.

# Tasks
- [x] rebase onto https://github.com/vercel/ai/pull/1920 once merged to pickup refactored `onFinish` callback changes
- [x] implementation
- [ ] test
- [ ] example
- [ ] docs
    - [ ] reference
    - [ ] example
- [ ] changeset


bneigher:
https://github.com/vercel/ai/issues/1894

packages/core/rsc/stream-ui/stream-ui.tsx
```diff
@@ -272,27 +272,85 @@ export async function streamUI<
...
 
       const reader = forkedStream.getReader();
       while (true) {
         const { done, value } = await reader.read();
+
+        if (wasPreviouslyText && (done || value.type !== 'text-delta')) {
+          const textNodeId = `text-${textNodeIndex}`;
+          if (!uiStreams[textNodeId]) {
+            throw new Error('Expected text node to exist');
+          }
```
gclark-eightfold, line 292: Is it possible to have text intermixed with tool calling?
I was able to get text _before_ tool calling, but trying to prompt at least the OpenAI agent to require a text response _after_ tool calling did not seem to work.

If text nodes can only appear once and **before** tool calling, this can be simplified instead of trying to keep track of nodes being written.

gclark-eightfold: Additionally, I noticed that with tool calling can come with a delay before the first chunk comes in for `reader.read()`. This had the unexpected effect of leaving out a last word or words of a text body that comes before tool calling and only shows up once the tool calls start streaming in.

gclark-eightfold:
@lgrammel LMK if you have any thoughts on this implementation.
You had mentioned previously that streamUI would be refactored eventually to match streamText / other patterns, so let me know if you think supporting this would be best to do after that change or before.

I've ad-hoc tested this in my own application and it does seem to work, though I've made a few comments on the implementation where things were maybe not as ideal.

dutsik-p:
Hi Guys! 
Any news on releasing this feature.

syedgaian:
Hi guys any news on , when will this feature be released?

gclark-eightfold:
> Hi guys any news on , when will this feature be released?

Haven't put this PR up for review because it's a bit buggy after updating it to the latest AI SDK version.
Currently getting this error in my application I've patched: `Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.` (though it could also be due to my impl which is a bit complex).

gclark-eightfold:
@lgrammel Would this happen to be a result of https://github.com/vercel/ai/pull/2068?

artokun:
Ah, I was wondering why my app kept hanging up. This would be awesome to get released

edspencer:
I had the same requirement and ended up building a layer on top of streamText to do it. It sounds like this will be natively supported in the SDK at some point, but for the time being I think it's still something one needs to put together oneself.

Check out https://github.com/edspencer/stream-multi if you are interested - the little video there shows the ability to have the LLM respond with both streaming text and any number of tools calls.

I published it as a npm module called [ai-stream-multi](https://www.npmjs.com/package/ai-stream-multi?activeTab=readme). It basically has the same API as streamText(), but adds the concept of `Segments`, which are either a Tool or Text, and of which there can be any number in a given response from the LLM. Use the `onSegment` callback to update the AIState.

Demo

zzh8829:
link all the related discussion, a new implementation could fix all these problems together.
https://github.com/vercel/ai/issues/2915
https://github.com/vercel/ai/issues/1894
https://github.com/vercel/ai/issues/1895
https://github.com/vercel/ai/issues/1976
https://github.com/vercel/ai/issues/2915

zzh8829:
I created a similar version here but using the new streamText api as the plumbing.
https://github.com/vercel/ai/issues/1895#issuecomment-2405975142

Issue #1934, opened 2024-06-13
steamUI Unsupported message role: [object Object] for Google Vertex gemini


Nelsonlin0321:
### Description

 Google Vertex Gemini uses 'model'  role  instead of 'assitant' but  streamUI function only accepts messages where the roles are 'user', 'system', 'assistant' or 'tool' by calling convertToLanguageModelMessage function.
 
The console showing the bug:
![image](https://github.com/vercel/ai/assets/34851732/6a9f1a9d-a3f6-4bc0-a3f0-2befda25a656)
![image](https://github.com/vercel/ai/assets/34851732/b6a24dbc-2008-45ac-938b-1189a3410e1b)

### Server 
```tsx
"use server";

import { createAI, getMutableAIState, streamUI } from "ai/rsc";
import { ReactNode } from "react";
import { nanoid } from "nanoid";
import { vertex } from "@/lib/google-vertex";

export interface ServerMessage {
  role: "user" | "model"; // Gemini requires the model role
  content: string;
}

export interface ClientMessage {
  id: string;
  role: "user" | "model"; // Gemini requires the model role
  display: ReactNode;
}

export async function continueConversation(
  input: string
): Promise<ClientMessage> {
  "use server";

  const history = getMutableAIState();
  const messages = [...history.get(), { role: "user", content: input }];
  console.log("Input Messages:", messages);
  const result = await streamUI({ // BUG REPORT
    model: vertex("gemini-1.5-pro"),
    messages: messages,
    text: ({ content, done }) => {
      if (done) {
        history.done((messages: ServerMessage[]) => [
          ...messages,
          { role: "model", content },
        ]);
      }

      return <div>{content}</div>;
    },
  });

  return {
    id: nanoid(),
    role: "model",
    display: result.value,
  };
}

export const AI = createAI<ServerMessage[], ClientMessage[]>({
  actions: {
    continueConversation,
  },
  initialAIState: [],
  initialUIState: [],
});

```

### Client
```tsx
"use client";

import { useState } from "react";
import { useActions, useUIState } from "ai/rsc";
import { nanoid } from "nanoid";
import { ClientMessage } from "../actions";

// Force the page to be dynamic and allow streaming responses up to 30 seconds
export const dynamic = "force-dynamic";
export const maxDuration = 30;

export default function Home() {
  const [input, setInput] = useState<string>("");
  const [conversation, setConversation] = useUIState();
  const { continueConversation } = useActions();

  return (
[...]

PR #1935, opened 2024-06-13
support model role for vertex gemini

Nelsonlin0321:
Fix the issue https://github.com/vercel/ai/issues/1934


lgrammel (changes requested):
you should just use "assistant". the ai sdk automatically maps it to "model"

PR #1945, opened 2024-06-13
feat(@ai-sdk/google): add file support for google gemini

BleedingDev:
This is the first attempt of adding file support to Gemini. I will work more on this, but currently I can't even build the repository as I am on Windows and it fails in `Rollup`, so I identified that those two commands fail:
`turbo build --filter=sveltekit-openai`
`turbo build --filter=nuxt-openai`


BleedingDev:
Got `build` + `type-check` successful, but I had to switch to WSL. It's pitty that we still fail to deliver repos working seamlessly on multiple operating systems.

lgrammel:
Think we have this now.

PR #1949, opened 2024-06-13
Stream assistant events as threadStatus

jeremyphilemon:
- [x] add new type of stream part `b:<AssistantStreamEvent>` 
- [x] stream thread status as `threadStatus`
- [x] update + add tests
- [x] update reference


packages/core/streams/assistant-response.ts
```diff
@@ -85,56 +86,55 @@ export function AssistantResponse(
...
+          switch (object) {
+            case 'thread.run.completed':
+            case 'thread.run': {
+              result = data as Run;
+
+              return {
+                id: data.id,
+                threadId: data.thread_id,
+                runId: data.run_id,
+                required_action: data.required_action,
```
lgrammel, line 101: `requiredAction` if you want to be similar to `threadId`

packages/core/tsconfig.vitest-temp.json
```diff
@@ -0,0 +1,44 @@
...
+    "tsBuildInfoFile": "/Users/jrmy/Clones/vx/public/ai/node_modules/.pnpm/vitest@1.6.0/node_modules/vitest/dist/vendor/tsconfig.tmp.tsbuildinfo"
+  },
+  "exclude": [
+    "*/dist",
+    "dist",
+    "build",
+    "node_modules"
+  ],
+  "include": [
+    "."
```
lgrammel, line 42: is this file needed? (`-temp`)

packages/react/src/samples.ts
```diff
@@ -0,0 +1,479 @@
+/*
+ * Sample stream events from OpenAI Assistant API for testing
+ */
+
+import { AssistantStreamPart } from '@ai-sdk/ui-utils';
```
lgrammel, line 5: is this file needed? if it's not part of prod bundle, maybe move to subfolder?

packages/react/src/use-assistant.ts
```diff
@@ -123,7 +132,7 @@ export function useAssistant({
       data?: Record<string, string>;
     },
   ) => {
-    setStatus('in_progress');
+    setThreadStatus('thread.message.created');
```
lgrammel, line 135: does the status also need to be set here for backwards compat?

packages/react/src/use-assistant.ts
```diff
@@ -160,67 +169,50 @@ export function useAssistant({
...
                 content: '',
-                data: value.data,
               },
             ]);
+
             break;
           }
 
-          case 'assistant_control_data': {
-            setThreadId(value.threadId);
```
lgrammel, line 210: how are we setting the thread id now?

packages/react/src/use-assistant.ts
```diff
@@ -160,67 +169,50 @@ export function useAssistant({
...
+          case 'thread.message.created': {
             setMessages(messages => [
               ...messages,
               {
-                id: value.id ?? generateId(),
-                role: 'data',
+                id: data.id,
+                role: 'assistant',
                 content: '',
-                data: value.data,
```
lgrammel, line 203: do we still have the ability to send data? i think it was for custom data

packages/react/src/use-assistant.ui.test.tsx
```diff
@@ -7,27 +7,29 @@ import '@testing-library/jest-dom/vitest';
 import { cleanup, findByText, render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { useAssistant } from './use-assistant';
+import { firstRun } from './samples';
```
lgrammel, line 10: since we use samples, maybe move to a test subfolder and rename to `test-samples`

packages/react/src/use-assistant.ui.test.tsx
```diff
@@ -43,48 +45,33 @@ describe('stream data stream', () => {
...
       JSON.stringify({
         threadId: null,
-        message: 'hi',
+        message: 'hey!',
       }),
     );
   });
 
-  describe('loading state', () => {
-    it('should show loading state', async () => {
```
lgrammel, line 87: leave old test to ensure backwards compat?

packages/ui-utils/src/types.ts
```diff
@@ -463,3 +464,8 @@ export type DataMessage = {
   role: 'data';
   data: JSONValue; // application-specific data
 };
+
+export type AssistantStreamPart = {
+  event: AssistantThreadStatus;
+  data?: any;
+};
```
lgrammel, line 471: most stream parts are defined in stream parts afaik. do we need to expose this? otherwise would prefer if it's consistent with the other stream parts. also, `any` is risky. is `unknown` possible?

jeremyphilemon: updated to `unknown` and also defined it inside `stream-parts`!

packages/react/src/use-assistant.ui.test.tsx
jeremyphilemon: done!

packages/react/src/use-assistant.ts
jeremyphilemon: I don't think `useAssistant` supported appending custom data, correct me if I'm wrong though.

My understanding here is that `message` is sent as an object (because we want to consume meta data like id and annotations) from `AssistantResponse` so it uses the `data_message` protocol. 

Also since `data_message` expects a [specific shape](https://github.com/vercel/ai/blob/4f28469a1af9f805a682ceb4bbd2d999a3b9add8/packages/ui-utils/src/stream-parts.ts#L165), I created an `assistant_message` stream part that exclusively handles the assistant APIs events so we can extend its functionality as the API changes within this stream part instead of creating new stream parts for every different event (`thread.create`, `thread.message.create`, etc.)

packages/react/src/use-assistant.ts
jeremyphilemon: It's set [here](https://github.com/vercel/ai/pull/1949/files#diff-1d8537a4d65dc935c7f758a49b447fee82412d5a4fdb6daf8281131c371ee1f2R190-R193) when the even type is `thread.run.created`

PR #1975, opened 2024-06-16
feat(@ai-sdk/azure): Add azure openai completion support

### LINKED ISSUE #1974: Support Azure OpenAI Completions
### Feature Description

Currently the OpenAI package includes support for completion models (e.g. gpt-3.5-turbo-instruct), but the Azure provider does not.

### Use Case

A sample usecase would be for async generation of chat names

### Additional context

~~I submitted a PR #1975, which "should" cover this issue, but may require additional testing.~~

The `@ai-sdk/openai` provider requires patches to conform to the standard where @ai-sdk/azure only `azure-openai-provider.ts`, which has been submitted as PR #1987.

~~Either PR should work, but #1987 is a bit cleaner.~~

Und3rf10w:
Closes #1974


lgrammel:
Was there a reason why the completion model provider needed to be copied (vs tweaking the OpenAI completion provider and reusing it)?

Und3rf10w:
No, I'm just an incredibly lazy programmer and did it very fast. I'll spend some time cleaning this up today.

packages/azure/src/azure-openai-completion-language-model.ts
```diff
@@ -0,0 +1,344 @@
...
+    usage: z
+      .object({
+        prompt_tokens: z.number(),
+        completion_tokens: z.number(),
+      })
+      .optional()
+      .nullable(),
+  }),
+  openAIErrorDataSchema,
+]);
```
lgrammel, line 344: why do you need this model? are there significant different that cannot be incorporated into the OpenCompletionLanguageModel in the openai provider?

Und3rf10w:
In the meantime, I've just added this code as a new provider in my package: https://gist.github.com/Und3rf10w/05ebf6fee439ff31044956886cb138f7

packages/azure/src/azure-openai-provider.test.ts
```diff
@@ -6,7 +6,7 @@ const TEST_PROMPT: LanguageModelV1Prompt = [
   { role: 'user', content: [{ type: 'text', text: 'Hello' }] },
 ];
 
-describe('doGenerate', () => {
+describe('chat-doGenerate', () => {
```
lgrammel, line 9: please use an outer describe block ("chat") instead

Und3rf10w:
Use #1987 instead.

PR #1987, opened 2024-06-17
feat(@ai-sdk/azure): Adds azure openai completion support

### LINKED ISSUE #1974: Support Azure OpenAI Completions

Und3rf10w:
In lieu of #1975

Closes #1974 

Please note that this PR includes patches to the @ai-sd/openai provider as well.


packages/azure/src/azure-openai-provider.ts
```diff
@@ -80,21 +90,45 @@ export function createAzure(
...
+    if (
+      deploymentId === 'gpt-35-turbo-instruct' ||
+      deploymentId === 'gpt-3.5-turbo-instruct'
+    ) {
+      return createCompletionModel(
+        deploymentId,
+        settings as OpenAICompletionSettings,
+      );
+    }
+
```
lgrammel, line 125: this does not make sense for azure afaik since you get to name your deployments

Und3rf10w: Yeah I don't have a good solution for this. Is there a better way to handle this case, since the current logic to determine whether to run `createChatModel` or `createCompletionModel` is determined entirely on the `modelId`/`deploymentId`?

Und3rf10w: I could add a `type` to the provider, that says either `completion` or `chat`, then choose `createCompletionModel` or `createChatModel` that way?

lgrammel:
This looks great! Thanks for adding it. I'll take it from here and do the changelog, docs changes etc

Und3rf10w:
> This looks great! Thanks for adding it. I'll take it from here and do the changelog, docs changes etc

My pleasure! I can start working on embedding support if you'd like since I'm going to need that anyways.

lgrammel:
> > This looks great! Thanks for adding it. I'll take it from here and do the changelog, docs changes etc
> 
> My pleasure! I can start working on embedding support if you'd like since I'm going to need that anyways.

That would be awesome! A similar approach should work there as well.

lgrammel:
Superseeded by https://github.com/vercel/ai/pull/1988

PR #2025, opened 2024-06-19
fix(@ai-sdk/azure): Don't submit logprobs to azure if not specified

### LINKED ISSUE #2024: bug(@ai-sdk/azure): `log_probs` is not be supported for calls to `gpt-4-vision-preview`, causing calls for that deployment to fail entirely
### Description

1. Send a request to Azure OpenAI chat completions, with the deployment pointing to a gpt-4-vision-preview deployment
2. Either specify `logprobs` in the request (defaulting to `false`) or omit it entirely. The `ai-sdk` will send the request with `logprobs` set to false.
3. Receive the following error:

```
error","message":"1 validation error for Request\nbody -> logprobs\n  extra fields not permitted (type=value_error.extra)
```

### Code example

```typescript
import { createAzure } from '@ai-sdk/azure';
import dotenv from 'dotenv';

dotenv.config();

// create provider instance via factory
const azure = createAzure({
      resourceName: '<redacted>',
      apiKey: '<redacted>',
})

async function main() {
    model: azure('gpt-4-vision-preview'), // use a deployment pointing to gpt-4-vision-preview
  });

  for await (const textPart of result.textStream) {
    process.stdout.write(textPart);
  }
}

main().catch(console.error);
```

### Additional context

`log_probs` is not supported in Azure OpenAI specifically only for `gpt-4-vision-preview`, as documented here: https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#chat-completions

image

Und3rf10w:
Resolves #2024

### DIFF
--- a/examples/ai-core/src/stream-text/azure-fullstream-logprobs.ts
+++ b/examples/ai-core/src/stream-text/azure-fullstream-logprobs.ts
@@ -13,7 +13,8 @@ async function main() {
   for await (const part of result.fullStream) {
     switch (part.type) {
       case 'text-delta': {
-        console.log('Text delta:', part.textDelta);zzz
+        console.log('Text delta:', part.textDelta);
+        zzz;
         break;
       }
 
--- a/examples/ai-core/src/stream-text/azure-fullstream-logprobs.ts
+++ b/examples/ai-core/src/stream-text/azure-fullstream-logprobs.ts
@@ -1,0 +1,71 @@
+import { azure } from '@ai-sdk/azure';
+import { streamText } from 'ai';
+import dotenv from 'dotenv';
+
+dotenv.config();
+
+async function main() {
+  const result = await streamText({
+    model: azure('gpt-4o', { logprobs: true }),
+    prompt: 'Invent a new holiday and describe its traditions.',
+  });
+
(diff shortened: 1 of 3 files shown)


Und3rf10w:
Tested successfully with:

```typescript
import { createAzure } from '@ai-sdk/azure';
import dotenv from 'dotenv';

dotenv.config();

// create provider instance via factory
const azure = createAzure({
  resourceName: '<redacted>',
  apiKey: '<redacted>',
});

async function main() {
    model: azure('gpt-4-vision'), // use a deployment pointing to gpt-4-vision-preview
  });

  for await (const textPart of result.textStream) {
    process.stdout.write(textPart);
  }
}

main().catch(console.error);
```

I also tried another deployment on the same resource and confirmed that's works as normally.

Und3rf10w:
This pull is a bit strange as most tests in the sdk expect `logprobs` to be present in the response, which is only the case if it is send (either as `true` or `false`). I'm not certain that this is the best solution, but you have to use the complicated ternary operator to avoid an error of: `error TS2790: The operand of a 'delete' operator must be optional.`

lgrammel:
Superseeded by #2029

PR #2031, opened 2024-06-20
fix (@ai-sdk/react): use-assistant fetch error

samuelint:
Fixes:
- react `useAssistant` error on a fetch error
- Type error with `Error` instead of `unknown` (like it's done with `useChat`). As I see in the current implementation, it's always an instance of `Error`, there's no need to be restricted to `unknown`


lgrammel:
This is great. However, we need to keep the different `useAssistant` implementations (as well as the docs) in sync. Would you mind adding this to the Svelte implementation as well (the Svelte implementation is temporarily duplicated in ai/svelte and @ai-sdk/svelte)

samuelint:
Sure. It's done @lgrammel

lgrammel:
Thank you! Continued in #2112

Issue #2062, opened 2024-06-22
Feature parity for `useChat` SolidJS


spiritledsoftware:
### Feature Description

The solid version of `useChat` is a bit behind in terms of features, especially in regards to client-side tool calls and round trips.

### Additional context

I can take a wack at this change myself and submit a PR

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests expanded capabilities and parity for the SolidJS API rather than reporting a concrete failure of existing behavior.

### Evidence
- The issue states that SolidJS `useChat` lacks features available in other versions.
- It specifically requests parity for client-side tool calls and round trips.
- No existing documented SolidJS behavior is described as broken.

PR #2063, opened 2024-06-23
Update SolidJS `useChat` implementation to maintain feature parity with React and others

ian-pascoe:
These changes just make it so the SolidJS implementation of `useChat` stays current with the features in the react version.

This resolves issue: https://github.com/vercel/ai/issues/2062. I have updated testing as well.


spiritledsoftware:
I have a couple more changes to make to this, specifically around reactivity of props. Will probably push them tomorrow

lgrammel:
awesome! my biggest concern around solid is currently the example ( https://github.com/vercel/ai/tree/main/examples/solidstart-openai ) and the outdated versions in there, which makes it harder for me to verify changes and holds me back from accepting changes for solid. maybe something you could take on as part of this (or a separate PR)?

packages/solid/src/use-chat.ts
```diff
@@ -54,15 +62,111 @@ export type UseChatHelpers = {
   input: Accessor<string>;
   /** Signal setter to update the input value */
   setInput: Setter<string>;
+  /** An input/textarea-ready onChange handler to control the value of the input */
+  handleInputChange: (
+    e:
+      | (Event & { currentTarget: HTMLInputElement })
+      | (Event & { currentTarget: HTMLTextAreaElement }),
```
lgrammel, line 69: is this the same as for react? it's important for me to keep things aligned (even if there could be more functionality by expanding), and then expand the functionality for all providers.

packages/solid/src/use-chat.ts
```diff
@@ -54,15 +62,111 @@ export type UseChatHelpers = {
...
+          role,
+          content,
+          ...(name !== undefined && { name }),
+          ...(data !== undefined && { data }),
+          ...(annotations !== undefined && { annotations }),
+          ...(toolInvocations !== undefined && { toolInvocations }),
+          // outdated function/tool call handling (TODO deprecate):
+          tool_call_id,
+          ...(function_call !== undefined && { function_call }),
+          ...(tool_calls !== undefined && { tool_calls }),
```
lgrammel, line 126: if these were not there before, let's not add them. the implementation is deprecated and broken

packages/solid/src/use-chat.ts
```diff
@@ -54,15 +62,111 @@ export type UseChatHelpers = {
...
+      }),
+      ...(chatRequest.function_call !== undefined && {
+        function_call: chatRequest.function_call,
+      }),
+      ...(chatRequest.tools !== undefined && {
+        tools: chatRequest.tools,
+      }),
+      ...(chatRequest.tool_choice !== undefined && {
+        tool_choice: chatRequest.tool_choice,
+      }),
```
lgrammel, line 148: same comments as above, let's only add non-deprecated parts

packages/solid/src/use-chat.ts
```diff
@@ -71,35 +175,50 @@ const chatApiStore = createSWRStore<Message[], string[]>({
...
+  }: {
+    toolCallId: string;
+    result: any;
+  }) => void;
+} {
+  // Generate a unique ID for the chat if not provided.
+  const hookId = createUniqueId();
+  const idKey = useChatOptions.id || `chat-${hookId}`;
+  const chatKey =
+    typeof useChatOptions.api === 'string'
```
lgrammel, line 210: when would the api not be a string?

packages/solid/src/use-chat.ts
```diff
@@ -111,91 +230,59 @@ export function useChat({
...
-                mutate(previousMessages.data);
-              }
-            },
-            generateId,
-          });
-        },
-        experimental_onFunctionCall,
+        getStreamedResponse: () =>
+          getStreamedResponse(
+            useChatOptions.api || '/api/chat',
```
lgrammel, line 265: prefer `??`

packages/solid/src/use-chat.ts
```diff
@@ -111,91 +230,59 @@ export function useChat({
...
+            chatRequest,
+            mutate,
+            setStreamData,
+            streamData,
+            extraMetadata,
+            messagesRef,
+            abortController,
+            useChatOptions.generateId || generateIdFunc,
```
lgrammel, line 273: prefer `??`

packages/solid/src/use-chat.ts
```diff
@@ -206,35 +293,90 @@ export function useChat({
...
+    const lastMessage = messagesRef[messagesRef.length - 1];
     if (lastMessage.role === 'assistant') {
-      return triggerRequest(messagesSnapshot.slice(0, -1), options);
+      const chatRequest: ChatRequest = {
+        messages: messagesRef.slice(0, -1),
+        options,
+        ...(functions !== undefined && { functions }),
+        ...(function_call !== undefined && { function_call }),
+        ...(tools !== undefined && { tools }),
+        ...(tool_choice !== undefined && { tool_choice }),
```
lgrammel, line 364: deprecated afaik

packages/solid/src/use-chat.ts
```diff
@@ -206,35 +293,90 @@ export function useChat({
...
     }
-    return triggerRequest(messagesSnapshot, options);
+
+    const chatRequest: ChatRequest = {
+      messages: messagesRef,
+      options,
+      ...(functions !== undefined && { functions }),
+      ...(function_call !== undefined && { function_call }),
+      ...(tools !== undefined && { tools }),
+      ...(tool_choice !== undefined && { tool_choice }),
```
lgrammel, line 376: same

spiritledsoftware:
> awesome! my biggest concern around solid is currently the example ( https://github.com/vercel/ai/tree/main/examples/solidstart-openai ) and the outdated versions in there, which makes it harder for me to verify changes and holds me back from accepting changes for solid. maybe something you could take on as part of this (or a separate PR)? 

Sure can. Might add a day or two.

packages/solid/src/use-chat.ts
spiritledsoftware: I'm probably not understanding the question, but this is the Solid-equivalent of the type in React afaik. Solid doesn't provide specific Event types, you just add `currentTarget` and `target` to the generic `Event` type

spiritledsoftware: https://www.solidjs.com/guides/typescript#event-handlers

Never mind, I could use these types above and will in the next commit

spiritledsoftware:
@lgrammel This is ready for another review pass. Changes complete

spiritledsoftware:
Did some cleanup, now I'm actually done with my changes, lol

examples/solidstart-openai/package.json
```diff
@@ -1,33 +1,28 @@
 {
   "name": "solidstart-openai",
-  "private": true,
```
lgrammel, line 3: we need to keep this

examples/solidstart-openai/package.json
```diff
@@ -1,33 +1,28 @@
...
+  "type": "module",
   "scripts": {
-    "dev": "solid-start dev",
-    "build": "solid-start build",
-    "start": "solid-start start"
+    "dev": "vinxi dev",
+    "build": "vinxi build",
+    "start": "vinxi start"
```
lgrammel, line 7: what is vinxi?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "vitest.disableWorkspaceWarning": true
+}
```
lgrammel, line 3: prefer not to change workspace settings - what is this for?

examples/solidstart-openai/package.json
lgrammel: (I'd prefer to not have an extra framework and stay on the level of solidstart)

lgrammel (commented):
This looks great. Main question I have is whether we can use solidstart instead of vinxi? that would be my preference

examples/solidstart-openai/package.json
spiritledsoftware: Vinxi is the new way of building and using solidstart

See here: https://docs.solidjs.com/solid-start/getting-started

.vscode/settings.json
spiritledsoftware: I will remove

spiritledsoftware:
> This looks great. Main question I have is whether we can use solidstart instead of vinxi? that would be my preference

See comment on package.json

lgrammel:
I've made a few changes in https://github.com/vercel/ai/pull/2063 (if you want to build on that). However, the unit tests are currently failing. And would you mind including the `use-chat-tools` example from `next-openai` in solid? the current example seems to be the old one, just updated.

spiritledsoftware:
> I've made a few changes in https://github.com/vercel/ai/pull/2095 (if you want to build on that). However, the unit tests are currently failing. And would you mind including the `use-chat-tools` example from `next-openai` in solid? the current example seems to be the old one, just updated.

So just to clarify, I should add onto that PR instead now?

lgrammel:
Up to you, you could also cherry pick the commits. I only create separate PRs because I usually cannot add to existing PRs and I need CI etc to test things out

spiritledsoftware:
Closing in favor of #2095

PR #2073, opened 2024-06-24
feat (ui): make event in handleSubmit optional

lgrammel:
## Rationale
This will enable using `handleSubmit` with different submission mechanism.

## Tasks

- [x] implementation
- [x] docs
- [x] changeset

### DIFF
--- a/.changeset/brave-nails-guess.md
+++ b/.changeset/brave-nails-guess.md
@@ -1,0 +1,9 @@
+---
+'@ai-sdk/svelte': patch
+'@ai-sdk/react': patch
+'@ai-sdk/solid': patch
+'ai': patch
+'@ai-sdk/vue': patch
+---
+
+feat (ui): make event in handleSubmit optional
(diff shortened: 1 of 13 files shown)


nicoalbanese (approved):

AntzyMo:
Why is `event` still used as the first parameter even though it is optional? ?
```js
function sendMessage() {
    handleSubmit(_, {
      options: {
        body: { text: content.value}
      }
    })
  }
```
Since this `event` is unnecessary, why not put it in the second parameter directly?
```js
function sendMessage(e) {
    handleSubmit({
      options: {
        body: { text: content.value}
      }
    }, e)
  }
```
Or maybe there is no need for `event` at all, just leave it to outsiders.
```js
function sendMessage(e) {
e.preventDefault()
    handleSubmit({
      options: {
        body: { text: content.value}
      }
    })
  }
```
These are some of my questions

lgrammel:
@AntzyMo because changing the order would break backwards compatibility

lgrammel:
I might be able to allow the options as the only arg with function type overloads, but I cannot change the order. Will look into it when I get a chance.

Issue #2085, opened 2024-06-25
If the request fails, the previous message is not restored


AntzyMo:
### Description

If `steamText` encounters an error, `messages` of `useChat` will lose the messages sent by `user` due to error reporting. For example:
```js
const messages = [
    {
        "role": "user",
        "content": "hi"
    },
    {
        "role": "assistant",
        "content": "hello"
    },
    {
        "role": "user",
        "content": "who are you?"
    }
]
```
If you send the last item of the `messages` array and an error is reported, the message `who are you` will be displayed. I think this is wrong and it looks like a `bug`.

So my idea is to restore to the previous message when an error is reported, including `user`

### Code example

https://github.com/vercel/ai/blob/a92523cf00dece8e91d9f8f0d6759ff3d6de7eea/packages/vue/src/use-chat.ts#L128

Taking `vue` as an example, `previousMessages` should be changed to `const previousMessages = messagesSnapshot`, which can ensure the latest message records

lgrammel:
Note: this came up internally as well.

jh3y:
Yep - This one gives the developer a little more work to do with the out of the box behavior.

Consider:
1. Start a new chat
2. Enter question
3. System errors
4. Wipes UI including users' question

The current workarounds rely on the developer maintaining a second shadow state of the AI state in order to provide better UX to their users.

In most cases, although the error message is most likely generic and related to the server itself, the user probably expects an error message in place of the streaming message coming in or appended to it with the option to reload or trigger a new message.

lgrammel:
https://github.com/vercel/ai/pull/2262

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes a concrete failure in existing chat error handling: a request error incorrectly discards the user's submitted message instead of restoring or retaining it.

### Evidence
- When a `useChat` streaming request errors, the newly submitted user message is removed from `messages`, wiping the question from the UI.
- A commenter confirms that the out-of-the-box behavior requires developers to maintain shadow state to preserve the user's message.
- A maintainer notes the issue also arose internally and links a corrective pull request (#2262).

PR #2086, opened 2024-06-25
fix (vue/react/solid/svelte): usechat restore the previous message

AntzyMo:
Information before repair cannot be restored #2085


lgrammel:
This would change existing behavior so I need some time to think about it.

AntzyMo:
> This would change existing behavior so I need some time to think about it.

👌

lgrammel:
Continued in https://github.com/vercel/ai/pull/2262

PR #2087, opened 2024-06-25
fix(@ai-sdk/vue) use-chat-handleSubmit event type  should be optional

AntzyMo:
The `handleSubmit` method of the `useChat` function, the first parameter `event` should be optional #2073


lgrammel:
Good catch, it seems this is still present for vue.

AntzyMo:
> Good catch, it seems this is still present for vue.

I looked at the other 3 frameworks without any problems, and only vue's event was left to be any

lgrammel:
Fixed in https://github.com/vercel/ai/pull/2089

PR #2095, opened 2024-06-25
chore (ui/solid): update solidjs useChat and useCompletion to feature parity with React

lgrammel:
Continuation of #2063 

## Tasks

- [x] test
- [x] update tool example
- [x] docs changes
- [x] changeset


lgrammel:
@spiritledsoftware this looks much better now but I'm running in trouble with the examples: 

CleanShot 2024-06-25 at 17 10 54

spiritledsoftware:
> @spiritledsoftware this looks much better now but I'm running in trouble with the examples:
> 
> CleanShot 2024-06-25 at 17 10 54

Looking into this

spiritledsoftware:
> > @spiritledsoftware this looks much better now but I'm running in trouble with the examples:
> 
> > 
> 
> > CleanShot 2024-06-25 at 17 10 54
> 
> 
> 
> Looking into this


@lgrammel - Struggling to diagnose the issue, something related to the SWR store and SSR.. That swr-store package is not widely used or supported, are you bound to it? Or would using a traditional Solid store work?

lgrammel:
@spiritledsoftware not bound to it, but the behavior should ideally not change. afaik, the store helps syncing several useChat hooks for the same chat.

lgrammel:
@spiritledsoftware do you know if the `solid-swr-store` works with the latest solid js versions? if not, we may want to abandon it. if yes, the bug is in the code.

lgrammel:
@spiritledsoftware the chat works now 🎉 

I ran into the same issue with http://localhost:3000/completion - I believe `useCompletion` would also need to be changed to work w/o `solid-swr-store`

CleanShot 2024-06-26 at 15 45 40

Also, the tool calling example does not work like the next / react one:

CleanShot 2024-06-26 at 15 47 41

And the vision example errors as well.

spiritledsoftware:
> @spiritledsoftware the chat works now 🎉
> 
> I ran into the same issue with http://localhost:3000/completion - I believe `useCompletion` would also need to be changed to work w/o `solid-swr-store`
> 
> CleanShot 2024-06-26 at 15 45 40
> 
> Also, the tool calling example does not work like the next / react one:
> 
> CleanShot 2024-06-26 at 15 47 41
> 
> And the vision example errors as well.

Opening a PR for completion API

Here is what I get from the tool calling example:
image

Looking into vision now

spiritledsoftware:
> > @spiritledsoftware the chat works now 🎉
> > I ran into the same issue with http://localhost:3000/completion - I believe `useCompletion` would also need to be changed to work w/o `solid-swr-store`
> > CleanShot 2024-06-26 at 15 45 40
> > Also, the tool calling example does not work like the next / react one:
> > CleanShot 2024-06-26 at 15 47 41
> > And the vision example errors as well.
> 
> Opening a PR for completion API
> 
> Here is what I get from the tool calling example: image
> 
> Looking into vision now

Fixed the problems in #2113

nicoalbanese (approved):

PR #2099, opened 2024-06-25
Solid: Fixing tests, adding better example for tool invocation

ian-pascoe:
Continuation of 2063 built upon the lg/2063 branch

@lgrammel - Here is the new PR

### DIFF
--- a/packages/solid/src/use-chat.ui.test.tsx
+++ b/packages/solid/src/use-chat.ui.test.tsx
@@ -7,14 +7,16 @@ import {
 import { cleanup, findByText, render, screen } from '@solidjs/testing-library';
 import '@testing-library/jest-dom';
 import userEvent from '@testing-library/user-event';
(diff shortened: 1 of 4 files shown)


packages/solid/src/use-chat.ui.test.tsx
```diff
@@ -342,19 +344,20 @@ describe('maxToolRoundtrips', () => {
 
   describe('single roundtrip with error response', () => {
     const TestComponent = () => {
-      const { messages, append, error } = useChat({
+      const { messages, append, error } = useChat(() => ({
         async onToolCall({ toolCall }) {
           mockFetchDataStream({
             url: 'https://example.com/api/chat',
             chunks: [formatStreamPart('error', 'some failure')],
+            maxCalls: 1,
```
lgrammel, line 352: this prevents the test from working as expected

PR #2135, opened 2024-06-28
feat: add options for zod-to-json-schema usage

wststone:
Adds an option to `streamUI`, `generateObject`, `streamObject`, `streamText` and `generateText` to control the output of tools paramter send to external llm provider. This could be useful when certain provider does not accept certain schema output such as '$schema' property


wststone:
closes as #2386 solved this issue

PR #2168, opened 2024-07-03
Implement `usage` response for `embed` and `embedMany` 

tobiasstrebitzer:
Currently, the `usage` response provided by OpenAI on Embedding is not available when calling `embed` or `embedMany`.

* This PR adds `usage` as an optional response property for `embed` and `embedMany` calls.
* This has been added to the `openai` sdk (not yet any others)
* Due to `usage` being optional in `core`, it serves as an optinal additional, without breaking any other providers or examples. Additinal providers can be updated in the future to expose the `usage` property.


lgrammel:
Continued in https://github.com/vercel/ai/pull/2220

Issue #1333, opened 2024-04-12
streamText with image data for Anthropic on Edge Runtime fires Illegal invocation error


twxxk:
### Description

1. Call experimental_streamText with text messages for an anthropic model. It works as expected.
2. Add the base64 image and ask the model to describe the image.

It works on the local backend environment, but when the edge runtime triggers the following error.

> undefined 'TypeError: Illegal invocation' TypeError: Illegal invocation
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:78:34)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:130:37)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:181:26)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:165:41)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:154:27)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:1429:43)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:405:17)
at (node_modules/.pnpm/ai@3.0.21_react@18.2.0_solid-js@1.8.16_svelte@4.2.13_vue@3.4.21_zod@3.22.4/node_modules/ai/dist/index.mjs:394:24)

Reproduced with 3.0.21

### Code example

```
export const runtime = 'edge';

const anthropic = new Anthropic({ apiKey: process.env.ANTHROPIC_API_KEY})
const messages = [
[...]

PR #2172, opened 2024-07-03
fix (provider-utils): check DOMException existence

ocavue:
In Vercel Edge runtime environment, `DOMException` doesn't exist, which causes the `Right-hand side of 'instanceof' is not an object` error. 

```
TypeError: Right-hand side of 'instanceof' is not an object
    at (node_modules/@ai-sdk/provider-utils/dist/index.mjs:102:9)
    at (node_modules/ai/dist/index.mjs:43:21)
    at (node_modules/ai/dist/index.mjs:1427:44)
```

Someone else also reported this issue before ([link](https://github.com/vercel/ai/issues/1333#issuecomment-2147138060)).

This PR fixes the issue by checking the existence of `DOMException`.

### DIFF
--- a/packages/provider-utils/src/is-abort-error.ts
+++ b/packages/provider-utils/src/is-abort-error.ts
@@ -1,5 +1,6 @@
 export function isAbortError(error: unknown): error is DOMException {
   return (
+    typeof DOMException !== 'undefined' &&
     error instanceof DOMException &&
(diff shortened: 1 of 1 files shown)


lgrammel:
Took a slightly different approach: https://github.com/vercel/ai/pull/2205

PR #2192, opened 2024-07-07
Fix unsafe use of `any` in `stream-object.ts`

AaronFriel:
The stream transformer used in `streamObject` permitted chunks of type `any`, which was then cast to a stream of `string | ObjectStreamInputPart`. However this was not type safe. The `usage` key was incorrect. This was papered over by assigning the `usage = calculateTokenUsage(chunk.usage)`, but the types also permitted `usage = chunk.usage`.

This fixes `streamObject` to generate a correctly typed stream passed to the `StreamObjectResult` constructor.


lgrammel:
Good find. This was done on purpose to avoid code duplication. That being said, the `any` type was an oversight.

lgrammel:
Superseeded by #2208 which fixes the types w/o duplicating the token usage calculation.

Issue #2036, opened 2024-06-20
Support streaming partial object chunks


AaronFriel:
### Feature Description

Support sending a stream of JSON-patch like events with `streamObject` to enable more efficient end-to-end workflows and complex scenarios involving eliding some content from the response sent to the client or streaming partial strings as deltas.

This would build on #1883 to support a stream of patches.

### Use Case

The `streamObject` API currently streams whole objects - this is great for rendering UI elements where we want the **whole object** to be provided to the client as UI; however for agent workloads we may want to elide or transform the object in flight and efficiently transmit, e.g., updates to a single string value.

For example, we might define a tool with a shape like so:

```ts
const rating = z.object({
  rating: z.string(),
  score: z.number().int().min(0).max(100),
});

const response = z.object({
  moderation: z.object({
    harm: rating.describe("physical or psychological harm"),
    // ...
  }).describe(`\
for each category by key, rate whether the user's text matches the category from "not very likely" to "very likely" \
then give it a score. the score is your subjective evaluation of how likely it is that the user intended \
to convey the category evaluated, on a scale of 0 (meaning not at all) to 100 (absolutely certain, 100%)`
  response: z.string().describe("your reply to the user"),
  // ...
});
```

We may want to combine this with a `createStreamableValue` or even perform this `streamObject` call inside a readable stream constructor which emits only the deltas of each value for efficiency. That is, instead of streaming:

```
{}
{"moderation":{}}
{"moderation":{"harm": ...}}
// ...
{"moderation":{"harm":...},"response":"hello"}
{"moderation":{"harm":...},"response":"hello, world"}
```

_Edited 2024-06-21:_

We might like to reduce these to patches, perhaps eliding certain paths or transforming the object entirely, so that the transformed stream is O(n) in the size of the input stream or what is added to the server:

```
{"kind":"partial","path":["response",null],value:"hello"}
{"kind":"partial","path":["response",null],value:", "}
{"kind":"partial","path":["response",null],value:"world"}
```

As the response grows in length, this reduces the bandwidth and computational work on the client to linear from quadratic. This quadratically reduces the bandwidth of sending large strings (such as a long `response` above) and makes it easy to act on updates to the `moderation` key efficiently and incrementally, only acting *on* the moderation value updates and making decisions on values in the object being completed or updated, as opposed to inspecting each partial object and maintaining a diff by hand:

```ts
'use server';

import { openai } from '@ai-sdk/openai';
import { streamObject } from 'ai';
[...]

PR #2193, opened 2024-07-07
Support returning a stream of unparsed text chunks in `streamObject`

AaronFriel:
This is an alternative to #1883 and #2036. As `StreamObjectResult` unconditionally parses the input JSON chunks, it's not possible in the current API to make use of incremental parsing without paying the quadratic cost of parsing in `streamObject`.

This PR adds an `asUnparsedParts` option to the `streamObject` call, which allows callers to obtain an unparsed stream of text chunks and implement their own parsing.

### DIFF
--- a/packages/core/core/generate-object/stream-object.test.ts
+++ b/packages/core/core/generate-object/stream-object.test.ts
@@ -703,3 +703,96 @@ describe('options.headers', () => {
     );
   });
 });
+
+describe('options.asUnparsedParts', () => {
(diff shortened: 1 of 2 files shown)


lgrammel:
Hey @AaronFriel - thanks for your PRs! I'd love to understand more on what you want to achieve and get on the same page. Would you be open to having a chat? My twitter DMs are open

AaronFriel:
Fixed in #2345

PR #2194, opened 2024-07-07
Support streaming of unparsed object parts

AaronFriel:
This is an alternative to #1883 and #2036. As `StreamObjectResult` unconditionally parses the input JSON chunks, it's not possible in the current API to make use of incremental parsing without paying the quadratic cost of parsing in `streamObject`.

This PR adds a `streamObjectParts` variant of `streamObject`, which allows callers to obtain an unparsed stream of text chunks and implement their own parsing.

This is an alternative to #2193 that declares a new exported function rather than overloading `streamObject`.


AaronFriel:
@lgrammel I hope this or #2193 is acceptable, as it will allow consumers to opt out of the quadratic parsing overhead in `streamObject`.

AaronFriel:
Fixed in #2345

PR #2195, opened 2024-07-08
feat (ai/openai): Support legacy function calls

AaronFriel:
The tool calling API may not be supported by all inference engines that have an OpenAI-compatible API. This change adds the `useLegacyFunctionCalling` option to OpenAI provider settings. This option is false by default and marked deprecated, as it should not be used except for compatibility with older inference APIs.

Parallel Tool Use
=================

This also provides a workaround for an issue with parallel tool calling. When using the `tools` option in the OpenAI provider, the inference API may choose to use a hidden `multi_tool_use.parallel` tool which is added to the prompt. When the model invokes this tool, the inference API may buffer the response until all parallel tools have been called. This can result in a significant delay, and effectively disables streaming (i.e.: `streamObject`) APIs.

The OpenAI Developer API has added a `parallel_tool_calls` option to the API, when set to false this hidden tool is not added to the prompt, and streaming behaves normally.

As of this writing, Azure OpenAI does not yet support the `parallel_tool_calls` option (per the 2024-05-01 inference API specification).

This workaround was endorsed by communication with OpenAI support, and is in production use today.


AaronFriel:
@lgrammel Azure has released their 2024-06-01 inference API spec and does not yet support the `parallel_tool_calls` option needed to move on from the OpenAI API. Would it be possible to review this?

lgrammel:
continued in https://github.com/vercel/ai/pull/2382

PR #2197, opened 2024-07-08
fix (ai): Make public conversion CoreMessage to Provider messages

### LINKED ISSUE #2196: Provider APIs (`doStream`, et al.) are not usable outside of AI SDK
### Description

The Provider APIs `doStream` and `doGenerate` are publicly exported, but not genuinely usable without extracting functionality from the AI SDK that is not exported.

These APIs take a `prompt` parameter defined here:

https://github.com/vercel/ai/blob/3db90c3d44b5e4de6c7414bd6a970d166356ba94/packages/provider/src/language-model/v1/language-model-v1-prompt.ts#L1-L12

As the comment observes, "This is not a user-facing prompt." Indeed. Generating valid instances of this type from `CoreMessage` is performed in multiple steps using non-public functions. Here in `streamText`, the `getValidatedPrompt` and `convertToLanguageModelPrompt` functions are private to the core package:

https://github.com/vercel/ai/blob/3db90c3d44b5e4de6c7414bd6a970d166356ba94/packages/core/core/generate-text/stream-text.ts#L146-L159

Likewise in the definitions for `generateText`, `generateObject`, `streamObject`, and `streamUI`.

### Code example

n/a

### Additional context

_No response_

AaronFriel:
These APIs were previously private, which made using the `doGenerate` and `doStream` APIs cumbersome from outside the AI SDK.

Fixes #2196

### DIFF
--- a/packages/core/core/prompt/index.ts
+++ b/packages/core/core/prompt/index.ts
@@ -3,3 +3,6 @@ export * from './convert-to-core-messages';
 export * from './data-content';
 export * from './invalid-message-role-error';
 export * from './message';
+export * from './prompt';
+export * from './get-validated-prompt';
+export * from './convert-to-language-model-prompt';


AaronFriel:
Not necessary with the change in #2345

PR #2199, opened 2024-07-08
fix (ai): Accept `name` on CoreMessage, pass to providers

### LINKED ISSUE #2198: Cannot use `name` field in Chat Completion messages (CoreMessage, LanguageModelV1Message)
### Description

In the previous `Message` type, it was possible to set both `role` and `name` fields. The `name` field was erroneously removed, though it serves an important function in identifying multiple users, or assistant messages from different contexts. The tokenizers for chat and instruction tuned models may contain a special token for the separator between a role and the name. 

For our use case, we use the "name" field to label messages from the assistant that contain guardrails. We then prompt the model - obviously this is not 100% reliable, but it is at least somewhat effective - to not reveal these messages. We also use the `name` field to label assistant messages that may be the result of other agents. 

This change has broken compatibility with previous versions of the AI SDK.

Related: most OpenAI models (since GPT-3.5-0613) will [tokenize both the role and name](https://github.com/openai/openai-cookbook/blob/62574a79b8cef067baa7b4143a77f366078d5d98/examples/How_to_count_tokens_with_tiktoken.ipynb#L448-L461), though the exact format is not specified. Many other models accept arbitrary "role" fields, though instruct/chat tuning may strongly benefit from certain roles being used, or roles that contain the token for `system`, `assistant`, or `user`, such as "user (Friel)" or "user-friel".

### Code example

_No response_

### Additional context

_No response_

AaronFriel:
Adds an optional `name` field to the relevant message types in the core package, provider SDK, and the OpenAI provider. A review of the current providers found no others that accept a `name` field, so no other providers are modified.

OpenAI's API is leading the way here, whether for multi-user or multi-agent interactions, LLM APIs are likely to accept named participants beyond 'user', 'assistant', etc.

As described in #2198, the `name` field may be used with OpenAI's API to provide additional context for system and assistant messages, e.g.: for guardrails and reducing the risks of prompt injection.

Fixes #2198

### DIFF
--- a/packages/core/core/generate-text/generate-text.test.ts
+++ b/packages/core/core/generate-text/generate-text.test.ts
@@ -427,6 +427,7 @@ describe('options.maxToolRoundtrips', () => {
                 },
                 {
                   role: 'assistant',
+                  name: undefined,
                   content: [
                     {
                       type: 'tool-call',
--- a/packages/core/core/prompt/convert-to-language-model-prompt.test.ts
+++ b/packages/core/core/prompt/convert-to-language-model-prompt.test.ts
@@ -6,6 +6,7 @@ describe('convertToLanguageModelMessage', () => {
       it('should convert image string https url to URL object', async () => {
         const result = convertToLanguageModelMessage({
           role: 'user',
+          name: 'admin',
           content: [
             {
               type: 'image',
@@ -16,6 +17,7 @@ describe('convertToLanguageModelMessage', () => {
 
         expect(result).toEqual({
           role: 'user',
+          name: 'admin',
           content: [
             {
               type: 'image',
@@ -28,6 +30,7 @@ describe('convertToLanguageModelMessage', () => {
       it('should convert image string data url to base64 content', async () => {
         const result = convertToLanguageModelMessage({
           role: 'user',
+          name: 'image-submitter',
           content: [
             {
               type: 'image',
@@ -38,6 +41,7 @@ describe('convertToLanguageModelMessage', () => {
 
         expect(result).toEqual({
           role: 'user',
+          name: 'image-submitter',
           content: [
             {
               type: 'image',
@@ -55,6 +59,7 @@ describe('convertToLanguageModelMessage', () => {
       it('should ignore empty text parts', async () => {
         const result = convertToLanguageModelMessage({
           role: 'assistant',
+          name: 'agent-1',
           content: [
             {
               type: 'text',
@@ -71,6 +76,7 @@ describe('convertToLanguageModelMessage', () => {
 
         expect(result).toEqual({
           role: 'assistant',
+          name: 'agent-1',
           content: [
             {
               type: 'tool-call',
--- a/packages/core/core/prompt/convert-to-language-model-prompt.ts
+++ b/packages/core/core/prompt/convert-to-language-model-prompt.ts
@@ -52,19 +52,21 @@ export function convertToLanguageModelMessage(
   const role = message.role;
   switch (role) {
     case 'system': {
-      return { role: 'system', content: message.content };
+      return { role: 'system', name: message.name, content: message.content };
     }
 
     case 'user': {
       if (typeof message.content === 'string') {
         return {
           role: 'user',
+          name: message.name,
           content: [{ type: 'text', text: message.content }],
         };
       }
 
       return {
         role: 'user',
+        name: message.name,
         content: message.content.map(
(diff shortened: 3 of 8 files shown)


AaronFriel:
@lgrammel is it possible to get this reviewed?

lgrammel:
Right now, this would be a very OpenAI specific feature. No other provider has implemented message names after they have been around almost a year and I have not heard any requests for this feature (outside of this PR). I'm want to wait and see if other providers adopt it first or if there is more demand, to prevent adding idiosyncratic functionality to the AI SDK API.

AaronFriel:
@lgrammel totally understand. Would you accept a PR with any of these three styles to allow a vendor-specific option without an API contract that it will always be supported?

<details>
<summary>TypeScript definitions</summary>


```ts
// CSS Prefix Style
type CoreMessageA = {    
    [prefixedExtension: `-${string}-${string}`]: any;
}

// Nested vendor, option style:
type CoreMessageB = {
    extensions: {
        [vendor: string]: {
            [option: string]: any;
        }
    }
}

// Flat, optional keys (no vendor/provider prefix):
type CoreMessageC = {
    extensions: {
        [option: string]: any;
    }
}
```
</details>

```ts
// CSS Prefix Style
const foo: CoreMessageA = {
    "-openai-name": "copilot"
}

// Nested vendor, option style:
const bar: CoreMessageB = {
    extensions: {
        openai: {
            name: "copilot"
        }
    }
};

// Flat, optional keys (no vendor/provider prefix):
const baz: CoreMessageC = {
    extensions: {
        name: "copilot"
    }
};
```

lgrammel:
@AaronFriel something like this is needed for sure. type safety is an issue. in the past i used generics on the models / providers for this, but it makes the system very complex and brittle. i'm leaning towards option c even tho it compromises type safety.

lgrammel:
fyi i'm planning to add this support to the language model v2 spec

lgrammel:
we now support provider specific extensions on messages that could be used for names: https://github.com/vercel/ai/pull/2697

AaronFriel:
@lgrammel I rebased the PR, and found a bug, curious how you want to proceed:

This works as expected:

```ts
        const result = await convertToLanguageModelPrompt({
          prompt: {
            type: 'messages',
            prompt: undefined,
            messages: [
              {
                role: 'user',
                content: [
                  {
                    type: 'text',
                    text: 'hello, world!',
                    experimental_providerMetadata: {
                      [providerName]: {
                        name: 'foo',
                        [key]: value,
                      },
                    },
                  },
                ],
              },
            ],
          },
          modelSupportsImageUrls: undefined,
        });
```

However, this does not, but it does type check:

```ts
        const result = await convertToLanguageModelPrompt({
          prompt: {
            type: 'messages',
            prompt: undefined,
            messages: [
              {
                role: 'user',
                content: [
                  {
                    type: 'text',
                    text: 'hello, world!',
                  },
                ],
                experimental_providerMetadata: {
                  [providerName]: {
                    name: 'foo',
                    [key]: value,
                  },
                },
              },
            ],
          },
          modelSupportsImageUrls: undefined,
        });
```

In the second example, the provider metadata does not flow through and the result. Instead, this is the passing assertion:

```ts
        expect(result).toEqual([
          {
            role: 'user',
            content: [
              {
                type: 'text',
                text: 'hello, world!',
              },
            ],
            // No provider metadata!
          },
        ] satisfies LanguageModelV1Prompt);
```

It looks like the issue is that `CoreUserMessage` declares the field:

https://github.com/vercel/ai/blob/15791b0a208c1e38d24893303e0e3b36221d3f5e/packages/ai/core/prompt/message.ts#L45-L58

The `UserContent` type can be an array of `TextPart` or `ImagePart`:

https://github.com/vercel/ai/blob/15791b0a208c1e38d24893303e0e3b36221d3f5e/packages/ai/core/prompt/message.ts#L74-L77

Which can then declare its own experimental provider metadata:

https://github.com/vercel/ai/blob/15791b0a208c1e38d24893303e0e3b36221d3f5e/packages/ai/core/prompt/content-part.ts#L8-L25

Should:
* Only one of these declare `experimental_providerMetadata`?
* Both, but there is a merge or overwriting that occurs on a part-by-part basis?

lgrammel:
Maybe I'm missing something, but for me it seems to work as expected: https://github.com/vercel/ai/pull/2776/files

The provider metadata on the message is passed through on the message (you can ignore the content part metadata for the name feature, it's meant for extensions that work on the content part level)

MwSpaceLLC:
Any Update ? Not work name for Open AI

Issue #1673, opened 2024-05-22
ai/packages/core/react /use-assistant.ts: expose `setThreadId`  as helper


tomfuller71:
### Feature Description

Expose `setThreadId` as a `UseAssistantHelper` so that a new conversation thread can be started by executing `setThreadId(undefined)`. 

### Use Case

Use case: an assistant providing feedback on a specific tuple of "question" and "student_answer".  The UI provides user the ability to select the specific course, exam and question they want to get the assistant's feedback on.  When selection is complete the UI should initiate a new chat with an initial prompt dynamically generated from the user's selections.  Each new request for feedback should be a new chat where the bot has no memory of any previous conversations.  

Without the ability to set the threadId to undefined the same chat thread  will be used for the entire  lifetime of the React component containing the `useAssistant` hook and even if messages are cleared the assistant still has memory of the previous conversation.  A full page refresh is not ideal as memory of current UI state will be lost without some workarounds.

### Additional context

As an alternative to exposing `setThreadId` add a more explicit helper.  e.g.

```
  /**
   * Resets the chat thread.
   * Stops any existing chat threads, clears the chat `Messages` and `Input`.
   * Resets `threadId` so that new messages will run on a new thread.
  */
  resetThread: () => {
    stop()
    setMessages([])
    setInput('')
    setThreadId(undefined)
  }
```

jeremyphilemon:
@tomfuller71 

#2209 adds `setThreadId` helper to switch between threads!

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests an API capability for thread management rather than reporting a failure of documented existing behavior.

### Evidence
- The issue explicitly requests exposing `setThreadId` through `UseAssistantHelper` to let applications start or switch conversation threads.
- The proposed `resetThread` helper is presented as an alternative API enhancement.
- A maintainer notes that #2209 adds the requested `setThreadId` helper.

PR #2209, opened 2024-07-08
feat (ai/react): add setThreadId helper to switch between threads for useAssistant

jeremyphilemon:
Addresses #1673 

- [x] add `setThreadId(threadId: string | undefined)` helper function to useAssistant
  - `setThreadId("thread_abc123")` sets the assistant to that specific thread
  - `setThreadId(undefined)` creates a new thread
  - both actions will reset messages and update threadId
- [x] add tests for thread management
- [x] update reference to include `setThreadId`

### DIFF
--- a/.changeset/plenty-ties-remember.md
+++ b/.changeset/plenty-ties-remember.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/react': patch
+---
+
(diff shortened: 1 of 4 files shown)


lgrammel (approved):

lgrammel:
Might be worth adding an example to the docs and tweeting about it?

PR #2226, opened 2024-07-09
Add experimental support for managing attachments to useChat

jeremyphilemon:
This change is currently only for the react hook

- [x] create `files` field as part of `Message`
- [x] support adding `files` through `handleSubmit()`
- [x] mark as experimental
- [x] update reference
- [x] add examples
  - [x] upload image as `data-url`
  - [x] upload text files as `data-url`
  - [x] upload image to vercel blob and pass to files as `url`
  - [x] upload pdf/text files to vercel blob and pass to files as `url`
- [x] add tests


examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+                  className="w-24 text-zinc-500 flex-shrink-0 text-sm flex flex-col gap-1"
+                >
+                  <div className="w-16 h-20 bg-zinc-100 rounded-md" />
+                  {file.name}
+                </div>
+              );
+            }
+          })}
+        </div>
+        <input type="file" onChange={handleFileInputChange} multiple />
```
lgrammel, line 98: For my understanding: we are limited to a single `file` input, correct? (even if it can support multiple files)

packages/react/src/use-chat.ts
```diff
@@ -513,12 +517,61 @@ By default, it's set to 0, which will disable the feature.
...
+      ) as HTMLInputElement;
+
+      let content: ContentPart[] = [];
+
+      if (input) {
+        content.push({
+          type: 'text',
+          text: input,
+        });
+      }
```
lgrammel, line 531: do we have a similar stateless submit for regular input? if not i'm wondering if we should leave it out for now and only go for controlled inputs (which are best practice for react if i'm not mistaken)

lgrammel: (we can always add something like this later in a separate change)

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <div className="flex flex-col gap-2">
+      <div className="flex flex-col p-2 gap-2">
+        {messages.map(message => (
+          <div key={message.id} className="flex flex-row gap-2">
+            <div className="w-24 text-zinc-500 flex-shrink-0">{`${message.role}: `}</div>
+            {typeof message.content === 'string' ? (
+              <div className="w-full">{message.content}</div>
+            ) : (
+              <div className="flex flex-col gap-2">
+                {message.content
```
lgrammel, line 27: is `message.content` currently a string? if yes, we may want to keep that for backwards compat. We could e.g. introduce `message.parts` for now and then figure out a migration path

packages/react/src/use-chat.ts
```diff
@@ -527,15 +580,20 @@ By default, it's set to 0, which will disable the feature.
 
       triggerRequest(chatRequest);
 
+      fileInput.value = '';
```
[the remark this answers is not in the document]
lgrammel: (if it's a controlled field we don't need to access the html element here - also this might not be a good idea considering react)

packages/ui-utils/src/types.ts
```diff
@@ -117,14 +117,26 @@ export type ToolInvocation =
...
+
 /**
  * AI SDK UI Messages. They are used in the client and to communicate between the frontend and the API routes.
  */
 export interface Message {
   id: string;
   createdAt?: Date;
 
-  content: string;
+  content: string | ContentPart[] | null | undefined;
```
lgrammel, line 139: alternatively we could introduce `parts` here and leave `content` as is for backwards compat

lgrammel:
I'm wondering if we can make the feature experimental for now, e.g. by naming the helpers and the parts `experimental_...`. The other consideration I have is backwards compatibility. This must change not break existing users.

packages/react/src/use-chat.ts
jeremyphilemon: `fileInput` here is used for clearing the input value, since they can't be mutated programmatically :/

will explore alternatives:
- using a ref for the file input
- resetting the form

packages/ui-utils/src/types.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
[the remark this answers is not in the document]
jeremyphilemon: The api has been updated to expose `files` through `handleSubmit` and does not directly access the `<input type="file"/>`

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: switched to use `parts`

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: this now depends on the user now that the api exposes `files` through `handleSubmit`

packages/ui-utils/src/types.ts
```diff
@@ -126,6 +138,8 @@ export interface Message {
 
   content: string;
 
+  parts: ContentPart[];
```
lgrammel, line 141: can parts be optional? we could only set them when files are available. that way we don't change what is being sent from client to server for existing setups

packages/ui-utils/src/parse-complex-response.ts
```diff
@@ -116,6 +127,7 @@ export async function parseComplexResponse({
           role: 'assistant',
           content: '',
           createdAt,
+          parts: [],
```
lgrammel, line 130: if we make parts optional this might not be necessary unless the user provides files

packages/react/src/use-chat.ts
```diff
@@ -513,12 +516,58 @@ By default, it's set to 0, which will disable the feature.
...
+      let parts: ContentPart[] = [];
+
+      if (input) {
+        parts.push({
+          type: 'text',
+          text: input,
+        });
+      }
+
+      if (options.files) {
```
lgrammel, line 528: My preference here would be to set/send parts only when files is != undefined. Otherwise we change what is being send for existing users, increases their traffic unexpectedly.

packages/react/src/use-chat.ts
```diff
@@ -513,12 +516,58 @@ By default, it's set to 0, which will disable the feature.
...
+              };
+              reader.onerror = error => reject(error);
+              reader.readAsDataURL(file);
+            });
+
+            parts.push({
+              type: 'image',
+              image: base64,
+            });
+          } else if (type.startsWith('text/')) {
```
lgrammel, line 546: how would text resources be handled on the backend? can we just send all resources that are not images as generic "file" with base64 encoding?

lgrammel: also maybe rewrite both using async/await and see if you can refactor to remove duplication

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+            </div>
+          </div>
+        ))}
+      </div>
+
+      <form
+        onSubmit={event => {
+          handleSubmit(event, {
+            files,
+          });
```
lgrammel, line 63: nice!

packages/react/src/use-chat.ts
lgrammel: ah i see we are sending it in the parts that need to be compatible. hmm. i'm wondering if we want to be more flexible, since this is for client server communication, not server llm communication. someone might eg want to send a pdf, that is then chunked on the server and used for RAG in that request

packages/ui-utils/src/types.ts
jeremyphilemon: agreed, done! 

instead of `parts`, it's now `files`, and it's set to optional

packages/ui-utils/src/types.ts
```diff
@@ -117,6 +118,21 @@ export type ToolInvocation =
...
+
+export interface URLMessageFile {
+  type: 'url';
+  name: string;
+  contentType: string;
+  url: string;
+}
+
+export type MessageFile = DataMessageFile | URLMessageFile;
+
```
lgrammel, line 135: Can we unify these types? 

```
export type MessageFile = {
  name?: string;
  mimeType?: string;
  url: string;
}
```

packages/ui-utils/src/types.ts
```diff
@@ -266,6 +284,14 @@ An optional object to be passed to the API endpoint.
   body?: object;
 };
 
+export interface URLFile {
+  name: string;
+  type: string | undefined;
+  url: string;
+}
```
lgrammel, line 291: Can this be the same as MessageFile above?

packages/ui-utils/src/types.ts
```diff
@@ -266,6 +284,14 @@ An optional object to be passed to the API endpoint.
...
+
+export type URLFileList = Array<URLFile>;
```
lgrammel, line 293: No need for an explicit type imo. I'd prefer to avoid exposing it (users will use it and then we have to manage it for backwards compat). We could just inline into files below:

`files?: FileList | Array<MessageFile>;

packages/react/src/use-chat.ts
```diff
@@ -522,6 +526,42 @@ By default, it's set to 0, which will disable the feature.
...
+            const { name, url, type } = file;
+
+            files.push({
+              type: 'url',
+              name,
+              contentType: type || 'application/octet-stream',
+              url,
+            });
+          }
+        }
```
lgrammel, line 562: should throw in case it's something else (which is unsupported)

packages/react/src/use-chat.ts
```diff
@@ -522,6 +526,42 @@ By default, it's set to 0, which will disable the feature.
...
+            });
+          }
+        } else if (Array.isArray(options.files)) {
+          for (const file of options.files) {
+            const { name, url, type } = file;
+
+            files.push({
+              type: 'url',
+              name,
+              contentType: type || 'application/octet-stream',
```
lgrammel, line 558: prefer `??` over `||` here

lgrammel: is `application/octet-stream` really what we want? can we leave as undefined?

packages/core/core/prompt/files-to-parts.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (file.dataUrl.includes('image/')) {
+        parts.push({ type: 'image', image: file.dataUrl });
+      } else if (file.dataUrl.includes('text/')) {
+        parts.push({ type: 'text', text: dataUrlToText(file.dataUrl) });
+      }
+    } else if (file.type === 'url') {
+      if (file.contentType.startsWith('image/')) {
+        parts.push({ type: 'image', image: file.url });
+      }
+    }
```
lgrammel, line 31: if we had a single type this could be collapsed

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+                        className="w-24 rounded-md"
+                        src={file.dataUrl}
+                        alt="image"
+                      />
+                    ) : file.dataUrl.includes('text/') ? (
+                      <div className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border">
+                        {atob(file.dataUrl.split(',')[1])}
+                      </div>
+                    ) : null
+                  ) : null,
```
lgrammel, line 39: could be collapsed if we had a single type

packages/react/src/use-chat.ts
jeremyphilemon: no longer relevant due to `files` parameter

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { useChat } from 'ai/react';
+
+export default function Page() {
+  const {
+    messages,
+    input,
+    files,
+    handleSubmit,
+    handleInputChange,
+    handleFileInputChange,
```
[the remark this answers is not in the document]
jeremyphilemon: no longer relevant due to `files` parameter

packages/ui-utils/src/types.ts
jeremyphilemon: done!

packages/ui-utils/src/types.ts
jeremyphilemon: done!

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: done!

packages/ui-utils/src/types.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
jeremyphilemon: done!

packages/core/core/prompt/files-to-parts.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
```diff
@@ -522,18 +525,62 @@ By default, it's set to 0, which will disable the feature.
 
       event?.preventDefault?.();
 
+      let files: MessageFile[] = [];
+
+      if (options.experimental_files) {
```
lgrammel, line 530: nit: extract `options.experimental_files` into a variable `files`

lgrammel: (also below to remove duplication)

packages/react/src/use-chat.ts
```diff
@@ -522,18 +525,62 @@ By default, it's set to 0, which will disable the feature.
 
       event?.preventDefault?.();
 
+      let files: MessageFile[] = [];
```
lgrammel, line 528: nit: const possible here?

packages/react/src/use-chat.ts
```diff
@@ -522,18 +525,62 @@ By default, it's set to 0, which will disable the feature.
...
+                id: generateId(),
+                role: 'user',
+                content: input,
+                experimental_files: files,
+              })
+            : messagesRef.current.concat({
+                id: generateId(),
+                role: 'user',
+                content: input,
+              })
```
lgrammel, line 583: would it be sufficient to just add `experimental_files: options.experimental_files`?

packages/ui-utils/src/types.ts
```diff
@@ -117,6 +118,16 @@ export type ToolInvocation =
   | CoreToolCall<string, any>
   | CoreToolResult<string, any, any>;
 
+/**
+ * File object that can be sent to the server.
+ */
+
```
lgrammel, line 124: nit: remove blank line

packages/ui-utils/src/types.ts
```diff
@@ -117,6 +118,16 @@ export type ToolInvocation =
...
+export interface MessageFile {
```
lgrammel, line 125: The name "MessageFile" bugs me a bit but I wasn't able to come up with something better considering that `File` is taken

turbo.json
```diff
@@ -22,7 +22,8 @@
         "ASSISTANT_ID",
         "INKEEP_API_KEY",
         "INKEEP_INTEGRATION_ID",
-        "VERCEL_URL"
+        "VERCEL_URL",
+        "UNSTRUCTURED_API_KEY"
```
lgrammel, line 26: still needed?

packages/core/core/prompt/files-to-parts.ts
```diff
@@ -0,0 +1,38 @@
+import { MessageFile } from '@ai-sdk/ui-utils';
+import { ImagePart, TextPart } from './content-part';
+
+type ContentPart = TextPart | ImagePart;
+
+/**
+ * Converts a data URL of type text/* to a text string.
+ */
+export function dataUrlToText(dataUrl: string): string {
+  return atob(dataUrl.split(',')[1]);
```
lgrammel, line 10: I vaguely remember some compatibility issues with `atob`

packages/core/core/prompt/files-to-parts.ts
```diff
@@ -0,0 +1,38 @@
...
+      parts.push({ type: 'image', image: file.url });
+    }
+
+    if (file.mimeType?.includes('text/') && isDataURL(file.url)) {
+      parts.push({ type: 'text', text: dataUrlToText(file.url) });
+    }
+  }
+
+  return parts;
+}
```
lgrammel, line 38: I have similar code in `convert-to-language-model-prompt.ts` line 85+ (URL processing) that relies on URL for parsing. Maybe we can align?

examples/next-openai-pages/package.json
```diff
@@ -9,13 +9,15 @@
...
-    "@ai-sdk/react": "latest",
     "@ai-sdk/openai": "latest",
+    "@ai-sdk/react": "latest",
+    "@vercel/blob": "^0.23.4",
     "ai": "latest",
     "next": "latest",
     "openai": "4.52.6",
     "react": "^18",
     "react-dom": "^18",
+    "unstructured-client": "^0.12.0",
```
lgrammel, line 20: still needed?

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+              <div className="flex flex-row gap-2">
+                {message.experimental_files?.map((file, index) =>
+                  file.url.includes('image/') ? (
+                    <img
+                      key={`${message.id}-${index}`}
+                      className="w-24 rounded-md"
+                      src={file.url}
+                      alt="image"
+                    />
+                  ) : file.url.includes('text/') ? (
```
lgrammel, line 33: can we rely on the type here instead? in urls it could also be e.g. the path

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,108 @@
...
+                    <div className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border">
+                      {atob(file.url.split(',')[1])}
```
lgrammel, line 35: use mime type directly instead of extractng?

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <div className="flex flex-col gap-2">
+              {message.content}
+
+              <div className="flex flex-row gap-2">
+                {message.experimental_files?.map((file, index) => (
+                  <img
+                    key={`${message.id}-${index}`}
+                    className="w-24 rounded-md"
+                    src={file.url}
+                    alt="image"
```
lgrammel, line 33: can we use the name here?

packages/ui-utils/src/types.ts
```diff
@@ -1,3 +1,4 @@
+import { MIMEType } from 'node:util';
```
lgrammel, line 1: We cannot have node imports. This is potentially used on the client as well

packages/ui-utils/src/types.ts
lgrammel: idea: we change the terminology to "Attachment" / "attachments" (then it's clear that it's attached to a message)

packages/ui-utils/src/types.ts
```diff
@@ -126,6 +137,11 @@ export interface Message {
 
   content: string;
 
+  /**
+   * Additional files to be sent along with the message.
+   */
+  experimental_files?: MessageFile[];
```
lgrammel, line 143: e.g. `experimental_attachments?: Attachment[]`

packages/ui-utils/src/types.ts
jeremyphilemon: agreed, updated!

turbo.json
jeremyphilemon: deleted line

examples/next-openai-pages/package.json
jeremyphilemon: deleted line

packages/ui-utils/src/types.ts
jeremyphilemon: deleted unused import

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
jeremyphilemon: done!

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: done!

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: done!

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+            fileInputRef.current.value = '';
+          }
+        }}
+        className="flex flex-col gap-2 fixed bottom-0 p-2 w-full"
+      >
+        <div className="flex flex-row gap-2 fixed right-2 bottom-14 items-end">
+          {Array.from(files).map(file => {
+            const { mimeType } = file;
+
+            if (mimeType && mimeType.startsWith('image/')) {
```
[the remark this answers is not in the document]
jeremyphilemon: done!

packages/ui-utils/src/types.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
jeremyphilemon: done!

packages/ui-utils/src/types.ts
```diff
@@ -117,6 +117,15 @@ export type ToolInvocation =
   | CoreToolCall<string, any>
   | CoreToolResult<string, any, any>;
 
+/**
+ * An attachment that can be sent along with a message.
+ */
+export interface Attachment {
+  name?: string;
+  mimeType?: string;
+  url: string;
```
lgrammel, line 126: Would be good to add JSDoc for the fields (it's shown in the IDE automatically when you use those fields)

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,39 @@
+import { Attachment } from '@ai-sdk/ui-utils';
+import { ImagePart, TextPart } from './content-part';
+
+type ContentPart = TextPart | ImagePart;
+
+/**
+ * Converts a data URL of type text/* to a text string.
+ */
+export function dataUrlToText(dataUrl: string): string {
+  return atob(dataUrl.split(',')[1]);
```
lgrammel, line 10: Now I remember: it's deprecated in node: https://stackoverflow.com/questions/68849233/convert-a-string-to-base64-in-javascript-btoa-and-atob-are-deprecated

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * Checks if a string is a data URL.
+ */
+function isDataURL(s: string): boolean {
+  return s.startsWith('data:');
```
lgrammel, line 17: creating a URL object and then checking the protocol would be a safer check

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export function attachmentsToParts(attachments: Attachment[]): ContentPart[] {
+  const parts: ContentPart[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.mimeType?.includes('image/')) {
+      parts.push({ type: 'image', image: attachment.url });
+    }
+
+    if (attachment.mimeType?.includes('text/') && isDataURL(attachment.url)) {
```
lgrammel, line 33: `startsWith` instead of `includes`?

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Converts a list of attachments to a list of content parts
+ * for consumption by ai/core functions.
+ * Currently only supports images and text attachments.
+ */
+export function attachmentsToParts(attachments: Attachment[]): ContentPart[] {
+  const parts: ContentPart[] = [];
+
+  for (const attachment of attachments) {
+    if (attachment.mimeType?.includes('image/')) {
```
lgrammel, line 29: `startsWith` instead of `includes`?

packages/react/src/use-chat.ts
jeremyphilemon: not sure if I understand correctly but `options.experimental_files` and `files` are not equivalent, I've renamed them to `attachmentsFromOptions` and `attachmentsForRequest` for better clarity

the attachments are converted to data urls when they're of type `FileList`

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+                  return (
+                    <div
+                      key={attachment.name}
+                      className="w-24 text-zinc-500 flex-shrink-0 text-sm flex flex-col gap-1"
+                    >
+                      <div className="w-16 h-20 bg-zinc-100 rounded-md" />
+                      {attachment.name}
+                    </div>
+                  );
+                }
```
lgrammel, line 88: what is behavior if the condition is not met?

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                const blob = await upload(file.name, file, {
+                  access: 'public',
+                  handleUploadUrl: '/api/file',
+                });
+
+                setAttachments(prevAttachments => [
+                  ...prevAttachments,
+                  {
+                    name: file.name,
+                    mimeType: blob.contentType,
```
lgrammel, line 91: wondering if `mimeType` or `contentType` is better for our attachment object?

lgrammel: We could align and also call it contentType.

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { useChat } from 'ai/react';
+import { useRef, useState } from 'react';
+
+export default function Page() {
+  const { messages, input, handleSubmit, handleInputChange, isLoading } =
+    useChat({
+      api: '/api/stream-chat',
+    });
+
+  const [attachments, setAttachments] = useState<FileList | undefined>(
```
lgrammel, line 11: my take would be to call it "files", "setFiles" here (since we are on the file list level). Those files will then be used as attachments

lgrammel: (then the uploaded `files` and the `attachments` on msgs are clearly distinguished)

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
lgrammel: but we could also leave as is to have consistent examples

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
lgrammel: but we could also leave as is to have consistent examples

packages/react/src/use-chat.ts
```diff
@@ -522,18 +525,63 @@ By default, it's set to 0, which will disable the feature.
...
+                id: generateId(),
+                role: 'user',
+                content: input,
+                experimental_attachments: attachmentsForRequest,
+              })
+            : messagesRef.current.concat({
+                id: generateId(),
+                role: 'user',
+                content: input,
+              })
```
lgrammel, line 584: wondering if we can simplify this.
```
messagesRef.current.concat({
  id: generateId(),
  role: 'user',
  content: input,
  experimental_attachments: attachmentsForRequest.length > 0 ? attachmentsForRequest : undefined
})
```

packages/ui-utils/src/types.ts
lgrammel: Then we can also mention that "data" urls are supported and you do not necessarily have to have an uploaded URL

jeremyphilemon: good callout, done!

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: TIL, done!

examples/next-openai-pages/pages/chat/stream-chat-files/index.tsx
jeremyphilemon: it wouldn't render any previews for file uploads if the condition is not met, figured this would be dependent on how developers want to preview different files

packages/ui-utils/src/call-chat-api.ts
```diff
@@ -34,6 +34,7 @@ export async function callChatApi({
   generateId: IdGenerator;
   fetch: ReturnType<typeof getOriginalFetch> | undefined;
 }) {
+  console.log(body);
```
lgrammel, line 37: log statement

packages/react/src/use-chat.ts
```diff
@@ -533,6 +576,7 @@ By default, it's set to 0, which will disable the feature.
               id: generateId(),
               role: 'user',
               content: input,
+              experimental_attachments: attachmentsForRequest,
```
lgrammel, line 579: will this add an empty array in many cases (increasing msg size)?

packages/react/src/use-chat.ts
jeremyphilemon: reworked this part of code to pass `experimental_attachments` and filter it out when constructing the request payload [here](https://github.com/vercel/ai/pull/2226/files#diff-0a57b9a56351ecd277e650d5a02072fbb65c6d685b81103e1cdf0459cf7c323eR123-R125)

packages/react/src/use-chat.ts
jeremyphilemon: it won't increase the message size, it gets filtered out when constructing the request payload [here](https://github.com/vercel/ai/pull/2226/files#diff-0a57b9a56351ecd277e650d5a02072fbb65c6d685b81103e1cdf0459cf7c323eR123-R125)

packages/ui-utils/src/call-chat-api.ts
jeremyphilemon: ah good catch!

packages/react/src/use-chat.ts
lgrammel: how would it get filtered there? wouldn't it still add an empty array? maybe I'm missing something

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: done!

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: done!

packages/react/src/use-chat.ts
jeremyphilemon: if `experimental_attachments` is not used in `handleSubmit`, then it will remain `undefined` and get filtered out in the payload

packages/core/core/prompt/files-to-parts.ts
jeremyphilemon: addressed this here https://github.com/vercel/ai/pull/2226#discussion_r1677603635

packages/core/core/prompt/attachments-to-parts.ts
lgrammel: There was some thing with Buffer as well re browser compat (the area is somewhat tricky). I don't think `Buffer` is available in browsers. I just searched the code and found that I'm in fact still using `atob` (sorry about the confusion): https://github.com/vercel/ai/blob/main/packages/provider-utils/src/uint8-utils.ts#L3

examples/next-openai-pages/pages/chat/stream-chat-files-url/index.tsx
jeremyphilemon: I agree – updated!

packages/core/core/prompt/files-to-parts.ts
jeremyphilemon: sounds good – updated!

packages/react/src/use-chat.ts
jeremyphilemon: don't this is relevant anymore sine we updated `parts` to `attachments` and added support for urls for server side processing

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
```diff
@@ -158,6 +158,82 @@ Allows you to easily create a conversational user interface for your chatbot app
...
+              description: 'Additional attachments sent along with the message.',
+              properties: [
+                {
+                  type: 'Attachment',
+                  parameters: [
+                    {
+                      name: 'name',
+                      type: 'string',
+                      isOptional: true,
+                      description: 'The name of the attachment.',
```
lgrammel, line 217: would it make to clarify that this is a filename? maybe even renaming to filename?

lgrammel: or is it more general?

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
```diff
@@ -253,6 +329,37 @@ Allows you to easily create a conversational user interface for your chatbot app
...
+                'An array of attachments to be sent to the API endpoint.',
+              properties: [
+                {
+                  type: 'Attachment',
+                  parameters: [
+                    {
+                      name: 'name',
+                      type: 'string',
+                      isOptional: true,
+                      description: 'The name of the attachment.',
```
lgrammel, line 346: see above

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
```diff
@@ -253,6 +329,37 @@ Allows you to easily create a conversational user interface for your chatbot app
               type: 'JSONValue',
               description: 'Additional data to be sent to the API endpoint.',
             },
+            {
+              name: 'experimental_attachments',
+              type: 'FileList | Array<Attachment>',
+              isOptional: true,
+              description:
+                'An array of attachments to be sent to the API endpoint.',
```
lgrammel, line 337: maybe mention the FileList case as well.

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
```diff
@@ -274,7 +381,8 @@ Allows you to easily create a conversational user interface for your chatbot app
       description:
         'React and SolidJS only. Function to add a tool result to the chat. This will update the chat messages with the tool result and call the API route if all tool results for the last message are available.',
     },
-  ]}
+
+]}
```
lgrammel, line 385: this is usually an indication that the formatting is off somewhere

examples/next-openai-pages/app/api/file/route.ts
```diff
@@ -0,0 +1,59 @@
+// https://vercel.com/docs/storage/vercel-blob/client-upload#create-a-client-upload-route
+
+import { handleUpload, type HandleUploadBody } from '@vercel/blob/client';
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request): Promise<NextResponse> {
```
lgrammel, line 6: maybe add a short comment above that explains what this route does?

examples/next-openai-pages/app/api/stream-chat/route.ts
```diff
@@ -1,13 +1,13 @@
...
 
   const result = await streamText({
-    model: openai('gpt-4'),
+    model: openai('gpt-4o'),
     system: 'You are a helpful assistant.',
-    messages,
+    messages: convertToCoreMessages(messages),
   });
 
   return result.toAIStreamResponse();
```
lgrammel, line 13: instead of adding to `next-openai-pages`, what do you think about adding to `next-openai` as `use-chat-attachments`?

examples/next-openai-pages/package.json
```diff
@@ -9,8 +9,10 @@
     "lint": "next lint"
   },
   "dependencies": {
-    "@ai-sdk/react": "latest",
     "@ai-sdk/openai": "latest",
+    "@ai-sdk/react": "latest",
+    "@ai-sdk/ui-utils": "latest",
+    "@vercel/blob": "^0.23.4",
```
lgrammel, line 15: Do we need to mention / update the README for deployments?

examples/next-openai-pages/pages/chat/stream-chat-attachments-url/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+              </div>
+            </div>
+          </div>
+        ))}
+      </div>
+
+      <form
+        onSubmit={event => {
+          if (isUploading) {
+            alert('Please wait for the files to finish uploading.');
```
lgrammel, line 45: can we disable instead?

examples/next-openai-pages/pages/chat/stream-chat-attachments/index.tsx
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable @next/next/no-img-element */
+import { useChat } from 'ai/react';
+import { useRef, useState } from 'react';
+
+export default function Page() {
+  const { messages, input, handleSubmit, handleInputChange, isLoading } =
+    useChat({
+      api: '/api/stream-chat',
```
lgrammel, line 8: i'd prefer `use-chat` -- we don't have stream-chat anywhere and it does not match any concept in our code

examples/next-openai-pages/pages/chat/stream-chat-attachments/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+                      className="w-24 rounded-md"
+                      src={attachment.url}
+                      alt={attachment.name}
+                    />
+                  ) : attachment.contentType?.includes('text/') ? (
+                    <div className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border">
+                      {Buffer.from(
+                        attachment.url.split(',')[1],
+                        'base64',
+                      ).toString('utf-8')}
```
lgrammel, line 38: does this work in browers?

examples/next-openai-pages/pages/index.tsx
```diff
@@ -36,6 +36,14 @@ const examples = [
...
     link: '/chat/stream-chat-edge',
   },
+  {
+    title: 'Stream chat completion with attachments',
+    link: '/chat/stream-chat-attachments',
+  },
+  {
+    title: 'Stream chat completion with attachments (url)',
+    link: '/chat/stream-chat-attachments-url',
+  },
```
lgrammel, line 46: - `use-chat-attachments`?
- move to `examples/next-openai` and leave only pages-router specific examples in pages route file?

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,74 @@
...
+import { convertDataContentToUint8Array } from './data-content';
+import { getErrorMessage } from '@ai-sdk/provider-utils';
+
+type ContentPart = TextPart | ImagePart;
+
+/**
+ * Converts a data URL of type text/* to a text string.
+ */
+export function dataUrlToText(dataUrl: string): string {
+  return Buffer.from(dataUrl, 'base64').toString('utf-8');
```
lgrammel, line 12: does this work in a browser env?

packages/core/core/prompt/attachments-to-parts.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * Checks if a string is a Data URL.
+ */
+function isDataURL(url: string): boolean {
+  return new URL(url).protocol === 'data:';
```
lgrammel, line 19: try...catch needed if url parsing fails imo:

```
try {
  return new URL(url).protocol === 'data:';
} catch (ignored) {
  return false
  }

File: packages/core/core/prompt/attachments-to-parts.ts [hunk 58]
```diff
@@ -0,0 +1,74 @@
...
+            throw new Error(`Error processing data URL: ${attachment}`);
+          }
+        }
+
+        default: {
+          throw new Error(`Unsupported URL protocol: ${url.protocol}`);
+        }
+      }
+    } catch (_ignored) {
+      // not a URL
```
lgrammel, line 69: all the inner errors will be dropped here, including there messages. is this desired? or should we remove this outer catch and force the user to validate beforehand?

packages/react/src/use-chat.ts
lgrammel: but it's set to `[]` in line 535

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -556,3 +563,303 @@ describe('maxToolRoundtrips', () => {
...
+                return (
+                  <div
+                    key={attachment.name}
+                    className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border"
+                    data-testid={`attachment-${idx}`}
+                  >
+                    {Buffer.from(
+                      attachment.url.split(',')[1],
+                      'base64',
+                    ).toString('utf-8')}
```
lgrammel, line 606: this is repeated several times. can we extract it into a helper function that we expose (and maybe test)?

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -556,3 +563,303 @@ describe('maxToolRoundtrips', () => {
...
+  // image file
+
+  it('should handle image file attachment and submission', async () => {
+    const file = new File(['test image content'], 'test.png', {
+      type: 'image/png',
+    });
+
+    const { requestBody } = mockFetchDataStream({
+      url: '/api/stream-chat',
+      chunks: ['0:"Response with image attachment"\n'],
```
lgrammel, line 712: `response to message with image attachment` ? (it confused me the first time i saw it)

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -556,3 +563,303 @@ describe('maxToolRoundtrips', () => {
...
+                return (
+                  <div
+                    key={attachment.name}
+                    className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border"
+                    data-testid={`attachment-${idx}`}
+                  >
+                    {Buffer.from(
+                      attachment.url.split(',')[1],
+                      'base64',
+                    ).toString('utf-8')}
```
lgrammel, line 784: same as above

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -556,3 +563,303 @@ describe('maxToolRoundtrips', () => {
...
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+    cleanup();
+  });
+
+  it('should handle text file attachment and submission', async () => {
+    const { requestBody } = mockFetchDataStream({
+      url: '/api/stream-chat',
+      chunks: ['0:"Response with attachment"\n'],
```
lgrammel, line 832: same as above

packages/ui-utils/src/types.ts
```diff
@@ -117,6 +117,27 @@ export type ToolInvocation =
   | CoreToolCall<string, any>
   | CoreToolResult<string, any, any>;
 
+/**
+ * An attachment that can be sent along with a message.
+ */
+export interface Attachment {
+  /**
+   * The name of the attachment.
+   */
```
lgrammel, line 126: mention that it's a filename? or even rename to `filename`? or is it more general?

lgrammel:
Re documentation: I think it's worth adding to our chatbot guide as well with an example: https://sdk.vercel.ai/docs/ai-sdk-ui/chatbot ( maybe a new annotations section with both a url and a non-url example ) so we can share docs links on twitter when we announce it

lgrammel:
Just noticed: this PR needs a changeset as well

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
jeremyphilemon: I'm leaning towards making it general since we chose to call it `attachments` instead of `files`. We also risk introducing two names describing the same thing.

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
jeremyphilemon: updated the description!

content/docs/07-reference/ai-sdk-ui/01-use-chat.mdx
jeremyphilemon: fixed formatting!

examples/next-openai-pages/app/api/file/route.ts
jeremyphilemon: done!

examples/next-openai-pages/app/api/stream-chat/route.ts
jeremyphilemon: moved the examples to `next-openai`!

examples/next-openai-pages/pages/chat/stream-chat-attachments/index.tsx
jeremyphilemon: `/api/chat` works, so used that!

examples/next-openai-pages/pages/chat/stream-chat-attachments-url/index.tsx
jeremyphilemon: disabling the input would mean the user cannot type any message during upload, so figured an alert works best

examples/next-openai-pages/pages/index.tsx
jeremyphilemon: done!

packages/ui-utils/src/types.ts
jeremyphilemon: done!

examples/next-openai-pages/pages/chat/stream-chat-attachments/index.tsx
jeremyphilemon: added a helper function [here](https://github.com/vercel/ai/pull/2226/files#diff-02c72a102d4cbb0de457bbb8233925ac1640bf37d02e6be326dd0990eb6bf1af).

it uses [window.atob](https://developer.mozilla.org/en-US/docs/Web/API/Window/atob) instead, which works in browsers!

only `atob` of node.js is deprecated ([source](https://stackoverflow.com/a/70851350))

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: https://github.com/vercel/ai/pull/2226#discussion_r1678314642

packages/react/src/use-chat.ui.test.tsx
jeremyphilemon: done! https://github.com/vercel/ai/pull/2226#discussion_r1678314642

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: this has be removed, so no longer relevant

packages/react/src/use-chat.ui.test.tsx
jeremyphilemon: done!

packages/react/src/use-chat.ui.test.tsx
jeremyphilemon: done!

packages/react/src/use-chat.ui.test.tsx
jeremyphilemon: done!

packages/core/core/prompt/convert-to-core-messages.ts
```diff
@@ -10,14 +12,28 @@ export function convertToCoreMessages(
...
+    experimental_attachments,
+  } of messages) {
     switch (role) {
       case 'user': {
-        coreMessages.push({ role: 'user', content });
+        coreMessages.push({
+          role: 'user',
+          content: experimental_attachments
+            ? [
+                { type: 'text', text: content },
```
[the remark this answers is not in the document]
jeremyphilemon: `Message` does not support parts, the content is of type `string`, so it's safe to assume that content will always be a string – no?

packages/core/core/prompt/attachments-to-parts.ts
jeremyphilemon: good _catch!_

I've removed the outer try/catch and validated the url separately before executing the switch block – should preserve the propagation of error message from the inner block

packages/react/src/use-chat.ts
jeremyphilemon: ah you're right! updated to add a check to set it as `undefined` if the array is empty

examples/next-openai-pages/package.json
jeremyphilemon: done! updated the readme with an updated `deployment url` and `.env.local.example`

jeremyphilemon:
> Re documentation: I think it's worth adding to our chatbot guide as well with an example: https://sdk.vercel.ai/docs/ai-sdk-ui/chatbot ( maybe a new annotations section with both a url and a non-url example ) so we can share docs links on twitter when we announce it

done, added a new attachments section!

content/docs/05-ai-sdk-ui/02-chatbot.mdx
```diff
@@ -349,3 +349,253 @@ export async function POST(req: Request) {
...
+
+import { getTextFromDataUrl } from '@ai-sdk/ui-utils';
+import { useChat } from 'ai/react';
+import { useRef, useState } from 'react';
+
+export default function Page() {
+  const { messages, input, handleSubmit, handleInputChange, isLoading } =
+    useChat({
+      api: '/api/chat',
+    });
```
lgrammel, line 376: I usually simplify the examples as much as possible (remove everything that's not relevant to this docs section like classNames, unused feature, etc) and highlight the crucial pieces. might be worth considering here. the reader wants to learn about this particular feature, other information will just distract them

examples/next-openai/README.md
```diff
@@ -6,7 +6,7 @@ This example shows how to use the [Vercel AI SDK](https://sdk.vercel.ai/docs) wi
 
 Deploy the example using [Vercel](https://vercel.com?utm_source=github&utm_medium=readme&utm_campaign=ai-sdk-example):
 
-[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/clone?repository-url=https%3A%2F%2Fgithub.com%2Fvercel%2Fai%2Ftree%2Fmain%2Fexamples%2Fnext-openai&env=OPENAI_API_KEY&envDescription=OpenAI%20API%20Key&envLink=https%3A%2F%2Fplatform.openai.com%2Faccount%2Fapi-keys&project-name=vercel-ai-chat-openai&repository-name=vercel-ai-chat-openai)
+[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/clone?repository-url=https://github.com/vercel/ai/tree/main/examples/next-openai&env=OPENAI_API_KEY,ASSISTANT_ID&envDescription=Learn more about how to get these environment variables&envLink=https://github.com/vercel/ai/blob/main/examples/next-openai/.env.local.example&project-name=ai-sdk-next-openai&repository-name=ai-sdk-next-openai&stores=[{"type":"blob"}])
```
lgrammel, line 9: will this still work if we remove the URL encoding?

lgrammel: also, do we need a note that people need to sign up for blob store only for the attachments example?

packages/core/core/prompt/convert-to-core-messages.ts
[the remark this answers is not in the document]
lgrammel: oh i see, was thinking about core message.

packages/core/core/prompt/data-content.ts
```diff
@@ -59,3 +59,17 @@ export function convertDataContentToUint8Array(
...
+ * @param uint8Array - The Uint8Array to convert.
+ * @returns The converted string.
+ */
+export function convertUint8ArrayToText(uint8Array: Uint8Array): string {
+  try {
+    return new TextDecoder().decode(uint8Array);
+  } catch (error) {
+    throw new Error('Error decoding Uint8Array to text');
+  }
+}
```
lgrammel, line 75: we should move into `uint8-utils.ts` eventually (in this PR it would create to much churn in the provider tho)

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -556,3 +563,300 @@ describe('maxToolRoundtrips', () => {
...
+                    src={attachment.url}
+                    alt={attachment.name}
+                    data-testid={`attachment-${idx}`}
+                  />
+                );
+              } else if (attachment.contentType?.startsWith('text/')) {
+                return (
+                  <div
+                    key={attachment.name}
+                    className="w-32 h-24 rounded-md text-xs ellipsis overflow-hidden p-2 text-zinc-500 border"
```
lgrammel, line 600: styling is not relevant in tests

lgrammel (approved):
Great work!

content/docs/05-ai-sdk-ui/02-chatbot.mdx
```diff
@@ -349,3 +349,253 @@ export async function POST(req: Request) {
   //...
 }
 ```
+
+## Attachments (Experimental)
+
+The `useChat` hook supports sending attachments along with a message as well as rendering them on the client. This can be useful for building applications that involve sending images, files, or other media content to the AI provider.
```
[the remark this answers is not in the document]
jeremyphilemon: done!

content/docs/05-ai-sdk-ui/02-chatbot.mdx
jeremyphilemon: cleaned up the examples!

packages/react/src/use-chat.ui.test.tsx
jeremyphilemon: done!

examples/next-openai/README.md
jeremyphilemon: yep, just tested it and the link works!

also updated the readme and `env.local.example` to mention vercel blob initialization.

baba43:
This is cool, thank you very much. It took me some time to find out that PDFs are not supported. 

What would be required to make PDFs working?

baba43:
Also, I found out that the message is not submitted when providing just a file but no text.

PR #2256, opened 2024-07-12
feat (provider/google): support system instructions

lgrammel:


jeremyphilemon (approved):

xl0:
@lgrammel Is this expected to only with with the v1beta endpoint?
The request sent is
```
....
    "systemInstruction": {
        "parts": [
            {
                "text": "[system instructions]"
            }
        ]
    },
```

and I'm getting back
```
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"systemInstruction\": Cannot find field.",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "description": "Invalid JSON payload received. Unknown name \"systemInstruction\": Cannot find field."
          }
        ]
      }
    ]
  }
``` 
on the v1 endpoint. Looking at the docs, I think it should be `system_instruction` : https://ai.google.dev/gemini-api/docs/system-instructions?lang=rest#examples 

Works on the v1beta endpoint though. Is this expected?

xl0:
Nope, still does not work in v1. Would it be better to just prepend a system message to the conversation?

lgrammel:
@xl0 yes, please use the `v1beta` endpoint. it's not intended to work with the `v1` endpoint (various things might break)

PR #2262, opened 2024-07-12
fix (ai/ui): keep last message in useChat on error

lgrammel:
## Summary

Introduces a flag `keepLastMessageOnError` on useChat (default: `false`). This flags enables the new error handling behavior that keeps the last message. The behavior is is intended to be the default in the next major release (and the flag will be removed).

## Background

We currently remove the last message from a chat when an error occurs. This makes retries and error handling much harder than it needs be and forces our users to manage the last message outside of the AI SDK.

## Tasks

- [x] implementation
   - [x] react
   - [x] solid
   - [x] vue
   - [x] svelte (x2)
- [x] examples
   - [x] react
   - [x] solid
   - [x] vue
   - [x] svelte
- [x] documentation
   - [x] add to reference docs
   - [x] add to error handling guide (ui)
   - [x] add proper full example for useChat
- [x] changeset


AntzyMo:
#2268 I don't know if this is the right branch to create, should I write in the 'main' branch or in this' lg/2086 '

nicoalbanese (approved):

PR #2272, opened 2024-07-14
fix 01-rag-chatbot.mdx return string[] from findRelevantContent

kzima:
curently:
findRelevantContent returns {
    name: string;
    similarity: number;
}[]

we need to return array of string instead:
findRelevantContent returns string[]

### DIFF
--- a/content/docs/02-guides/01-rag-chatbot.mdx
+++ b/content/docs/02-guides/01-rag-chatbot.mdx
@@ -646,7 +646,7 @@ export const findRelevantContent = async (userQuery: string) => {
     .where(gt(similarity, 0.5))
     .orderBy(t => desc(t.similarity))
     .limit(4);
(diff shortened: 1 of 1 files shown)


nicoalbanese:
Hey - thanks for the suggestion! This format was done on purpose so the model has context on the degree of similarity for each piece of content.

kzima:
@nicoalbanese thanks for the feedback. I only suggested the fix because when you run the example as per docs UI returns nothing to the user.

nicoalbanese:
That's really helpful feedback, thank you. This is intentional because we don't want the user to see any of the tool call results. Rather, we want the model to use the results to generate a response and only then will it be fed back to the user. This is is different from other tool examples in the docs so I can see how that could be misleading. I'll clarify in the guide. Thanks again 😊

PR #2311, opened 2024-07-17
Update use-chat.ts ( React )

MahmoudvMarey:
add a missing "createdAt" on "handleSubmit", so the user message has a date


lgrammel:
Superseeded by other fix

Issue #2312, opened 2024-07-17
Add `baseURL` and `apiVersion` to the Azure OpenAI Provider


karl-richter:
### Feature Description

Most providers currently support a `baseURL` parameter that allows to use a custom proxy endpoint with the respective provider. For the Azure Provider this option is currently missing, as only the `resourceName` parameter is supported that is used to construct the baseURL in the background. The construction of the URL currently looks as follows:
```js
const url = ({ path, modelId }: { path: string; modelId: string }) =>
    `https://${getResourceName()}.openai.azure.com/openai/deployments/${modelId}${path}?api-version=2024-05-01-preview`;
```

I'd propose to add an alternative option to the `AzureOpenAIProviderSettings` that gives more control over how the URL is constructed. It could look something like this:
```js
const url = ({ path, modelId }: { path: string; modelId: string }) =>
    getBaseUrl()
      ? `${getBaseUrl()}/${modelId}${path}?api-version=${getApiVersion()}`
      : `https://${getResourceName()}.openai.azure.com/openai/deployments/${modelId}${path}?api-version=${getApiVersion()}`;
```

### Use Case

This would be needed to support Proxy API's to the Azure OpenAI service.

lgrammel:
`baseURL` can be configured (change landed in main)

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
The issue requests expanded Azure OpenAI provider configuration to support custom proxy endpoints and API versions, rather than reporting a failure of promised behavior.

### Evidence
- The Azure provider constructs its endpoint from `resourceName` and a fixed API version, preventing custom proxy endpoint configuration.
- The reporter requests configurable `baseURL` and `apiVersion` settings as new provider capabilities.
- A maintainer states that configurable `baseURL` subsequently landed in `main`.

PR #2314, opened 2024-07-17
Added `baseURL` to the Azure OpenAI Provider

karl-richter:
Implementation of #2312 

### Main Changes
- Added `baseURL` ~~and `apiVersion`~~ as optional parameters to `AzureOpenAIProviderSettings`
- Added two test cases for the change

### Secondary Changes
~~- Added two parameters to the `loadSetting` function in `provider-utils`. The `defaultValue` param is needed to handle the fallback value for the `apiVersion` parameter. The `failOnMissing` param is needed to handle optional parameters in the Provider. It is used to check whether a `baseURL` is provided.~~
- Added a `getRequestUrl` function to the `JsonTestServer`. This is needed in the provider test to ensure that the `baseURL` is properly set.

### DIFF
--- a/packages/azure/src/azure-openai-provider.ts
+++ b/packages/azure/src/azure-openai-provider.ts
@@ -68,11 +68,6 @@ The default prefix is ``https://{resourceName}.openai.azure.com/openai/deploymen
   baseURL?: string;
 
   /**
-Use API version. The default API version is ``2024-05-01-preview``.
-   */
-  apiVersion?: string;
-
-  /**
 API key for authenticating requests.
      */
   apiKey?: string;
@@ -118,7 +113,6 @@ export function createAzure(
       settingName: 'baseURL',
       environmentVariableName: 'AZURE_BASE_URL',
       description: 'Azure OpenAI base URL',
-      defaultValue: undefined,
       failOnMissing: false,
     });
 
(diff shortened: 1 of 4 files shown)


karl-richter:
Hi @lgrammel & others, I opened the issue #2312 earlier but now just went ahead and added the parameters to the Azure OpenAI provider in my fork. Please let me know of the change is intended / aligns with your plans & what the next steps would be to proceed with the review. Thanks!

lgrammel:
@karl-richter I'm in favor of including the baseUrl. However, I have concerns re the api version. It should not be changed since our parsing relies on a specific api version. Instead we should update the version in the SDK. can you open a separate PR or remove the version part?

karl-richter:
Thanks for the feedback @lgrammel! I have removed the `apiVersion` from the PR, I agree that the API version might have some downstream effects that should be handled respectively by the SDK. I also made the additional `failOnMissing` arg in the `loadSetting` function properly optional, so the build test should no longer fail. Could you trigger the Cloud tests again?

packages/azure/src/azure-openai-provider.ts
```diff
@@ -101,8 +107,27 @@ export function createAzure(
...
     });
 
+  const checkIfBaseUrlIsProvided = () =>
+    loadSetting({
+      settingValue: options.baseURL,
+      settingName: 'baseURL',
+      environmentVariableName: 'AZURE_BASE_URL',
+      description: 'Azure OpenAI base URL',
+      failOnMissing: false,
+    });
```
lgrammel, line 117: I would much prefer if we can only pass in the base URL through options for now (no env param loading). This means all the "failOnMissing" etc is not required.

The env loading is reserved for only the super-common options to get started. The expectation is to use options once the configuration gets more complex.

lgrammel: (we need to document, test, and maintain this code and my preference is to keep it as simple as possible)

karl-richter: Alright, we can do that! I agree that this is simpler and less prone.

karl-richter: I could also remove the small test I added for the provider to check that the `baseURL` param works. Should we keep that or also remove it if we say that `baseURL` is not a core parameter?

lgrammel: the test is great! I'll continue on the PR tomorrow and add docs & changelog. Thank you!

lgrammel:
continued in https://github.com/vercel/ai/pull/2337

PR #2321, opened 2024-07-18
fix (ai/core): do not construct object promise in streamObject result until requested

lgrammel:
## Background
We eagerly constructed a promise for the `streamObject` result that was rejected when the final schema validation failed. In many cases, users did not use the `object` promise from the result, leading to unhandled promise rejection errors.


packages/core/core/generate-object/stream-object.test.ts
```diff
@@ -470,18 +458,6 @@ describe('result.object', () => {
...
-            {
-              role: 'system',
-              content:
-                'JSON schema:\n' +
-                '{"type":"object","properties":{"content":{"type":"string"}},"required":["content"],"additionalProperties":false,"$schema":"http://json-schema.org/draft-07/schema#"}\n' +
-                'You MUST answer with a JSON object that matches the JSON schema above.',
-            },
-            { role: 'user', content: [{ type: 'text', text: 'prompt' }] },
-          ]);
-
```
lgrammel, line 484: removed bc it was redundant and not part of the goal of this (and the above) test

jeremyphilemon (approved):

PR #2330, opened 2024-07-18
docs: add core message reference

nicoalbanese:


content/docs/07-reference/ai-sdk-core/31-convert-to-core-message.mdx
```diff
@@ -0,0 +1,56 @@
+---
+title: convertToCoreMessages
+description: Convert useChat messages to CoreMessages for AI core functions (API Reference)
+---
+
+# `convertToCoreMessages()`
```
lgrammel, line 6: my thinking was putting this under `ai/ui` because it is not useful when you only use core.

nicoalbanese: good call!

lgrammel (approved):

PR #2346, opened 2024-07-19
feat (ai/ui): add allowEmptySubmit flag to handleSubmit

jeremyphilemon:
Use flag instead of allowing empty message by default introduced by #2164, since it is a breaking change for existing users.

- just `handleSubmit` with empty input should do nothing (original behavior)
- `handleSubmit(e, { allowEmptySubmit:true })` with empty input should just send messages
- `handleSubmit(e, { allowEmptySubmit:true })` with input should send messages + new message for input

---

- [x] Add allowEmptySubmit to react package
  - [x] Add test for allowEmptySubmit to react package
- [x] Add allowEmptySubmit to vue package
  - [x] Add test for allowEmptySubmit to vue package
- [x] Add allowEmptySubmit to solid package
  - [x] Add test for allowEmptySubmit to solid package
- [x] Add allowEmptySubmit to svelte package
- [x] update reference


packages/react/src/use-chat.ui.test.tsx
```diff
@@ -298,10 +298,106 @@ describe('form actions', () => {
...
+          }}
+        >
+          <input
+            value={input}
+            placeholder="Send message..."
+            onChange={handleInputChange}
+            disabled={isLoading}
+            data-testid="do-input"
+          />
+        </form>
```
lgrammel, line 335: neat! this means we can easily test the id and created at next week as well!

lgrammel (approved):

lgrammel:
Would it make sense to add a short subsection to the chatbot guide page (under AI SDK UI) as well?

PR #2360, opened 2024-07-22
fix(ai/core): fullStream should throw on error

wong2:
The doc of fullStream states that _When an error occurs, the stream will throw the error_, but it's missing in the code.


lgrammel:
The docs are incorrect in this case. `fullStream` should not throw, but include error chunks. Where did you see the incorrect docs?

wong2:
@lgrammel https://sdk.vercel.ai/docs/reference/ai-sdk-core/stream-text#full-stream

Also, the comment above `fullStream` code

lgrammel:
Pull request for docs: https://github.com/vercel/ai/pull/2362

PR #2364, opened 2024-07-22
docs: add plain objects troubleshooting page

nicoalbanese:


lgrammel (approved):
is changing the approach (1.) necessary or is changing to streamable value good enough in most cases?

nicoalbanese:
> is changing the approach (1.) necessary or is changing to streamable value good enough in most cases?

I think it's good to separate them into two steps. That way we are clear that step 1 is resolving the error, step 2 is providing the solution. It's verbose but perhaps helpful in a troubleshooting context?

PR #2367, opened 2024-07-22
chore (ai/core): remove grammar mode

lgrammel:
Removes grammar mode.

## Background
Grammar mode was never supported in any provider. It can be included under JSON mode, where grammar should automatically be used when the provider supports them. Thus, the grammar mode only adds unnecessary complexity.


nicoalbanese (approved):
Tiny nit: update "a instruction" to "an instruction"

PR #2373, opened 2024-07-22
feat (provider): support responseFormat setting in provider api

lgrammel:
This setting is internal for now. It was included to provide a path to simplify the providers in the future and to enable adding a potential responseFormat to generateText in the future without breaking the providers.


jeremyphilemon (approved):

mattwkelly:
@lgrammel is it possible to use responseFormat now? I'd like to use it for groq (https://console.groq.com/docs/api-reference#chat-create). I'm getting an error when I pass it via streamUI. 

"Object literal may only specify known properties, and 'responseFormat' does not exist in type 'CallSettings & Prompt & { model: LanguageModelV1; tools?: { showStockInformation: RenderTool<ZodObject<{ symbol: ZodString; }, "strip", ZodTypeAny, { ...; }, { ...; }>>; } | undefined; toolChoice?: CoreToolChoice<...> | undefined; text?: RenderText | undefined; initial?: ReactNode; onFinish?: ((event: { ...; }) => v...'.ts(2353)"...

PR #2409, opened 2024-07-25
Add session token to AWS bedrock

shoopapa:
The default configuration of the bedrock provider doesn't account for session tokens, 
sessionTokens are common in local AWS environments and should be supported by default


tmokmss:
jfyi you can use local credential or IAM role credential directly by setting `bedrockOptions` instead.

For example.

```ts

export const maxDuration = 30;

    model: createAmazonBedrock({ bedrockOptions: { region: 'us-east-1' } })('anthropic.claude-3-haiku-20240307-v1:0'),
    messages,
  });

  return result.toAIStreamResponse();
}
```

shoopapa:
> jfyi

Thanks! I noticed that after I finished this code change lol
I still feel like it makes sense to merge in. seems like most of the other providers just work without configuring

packages/amazon-bedrock/src/bedrock-provider.ts
```diff
@@ -65,6 +66,12 @@ export function createAmazonBedrock(
             environmentVariableName: 'AWS_SECRET_ACCESS_KEY',
             description: 'AWS secret access key',
           }),
+          sessionToken: loadSetting({
+            settingName: 'sessionToken',
+            settingValue: options.sessionToken,
+            environmentVariableName: 'AWS_SESSION_TOKEN',
+            description: 'AWS session token',
+          }),
```
lgrammel, line 74: Is it common to load the session token from an environment variable? Or should it be just `options.sessionToken` since it almost always needs to be resolved first?

shoopapa: Not sure what you mean by "needs to be resolved first" 
But in my experience if you have logged into a AWS account that uses sessions you'll end up with those 3 variables in you environment

lgrammel: If the session token is not available, this will throw an error, which I'd like to avoid. Therefore just using `options.sessionToken` is the better solution for now. Or `loadSetting` needs to be changed (an optional flag `isOptional`, defaulting to false, would need to be introduced).

shoopapa: Added in the loadSetting feature you described, let me know if it looks alright.
also I haven't used changesets before so let me know if theres more to do for that before it hits CD

lgrammel:
Continued in #2755

PR #2430, opened 2024-07-26
feat (ui): invoke useChat onFinish with finishReason and tokens

lgrammel:
## Summary
* add second parameter to `useChat` `onFinish` callback that receives `usage` and `finishReason`
* introduce new `finish_message` stream part
* send `finish_message` stream part in `streamText.toAIStream()`

## Tasks

- [x] implementation
   - [x] interface definition
   - [x] finish message message part
   - [x] streamText send message part
   - [x] react
   - [x] vue
   - [x] svelte
   - [x] solid
- [x] examples
  - [x] next
  - [x] vue
  - [x] solid
  - [x] svelte
- [x] docs
  - [x] chatbot guide
  - [x] useChat reference
- [x] changeset

### DIFF
--- a/packages/ui-utils/src/duplicated/token-usage.ts
+++ b/packages/ui-utils/src/duplicated/token-usage.ts
@@ -1,0 +1,40 @@
+/**
+Represents the number of tokens used in a prompt and completion.
+ */
+export type CompletionTokenUsage = {
+  /**
+The number of tokens used in the prompt
+   */
+  promptTokens: number;
+
+  /**
+The number of tokens used in the completion.
+ */
+  completionTokens: number;
+
+  /**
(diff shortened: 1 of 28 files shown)


packages/ui-utils/src/duplicated/token-usage.ts
```diff
@@ -0,0 +1,40 @@
+/**
+Represents the number of tokens used in a prompt and completion.
+ */
+export type CompletionTokenUsage = {
+  /**
+The number of tokens used in the prompt
```
jeremyphilemon, line 6: missed a period at the end

jeremyphilemon (approved):

PR #2445, opened 2024-07-29
fix (docs): removed repeated temp values in examples

dhruvvbhavsar:
hey so i am new to ai sdk and was going through the examples section. [here](https://sdk.vercel.ai/examples/next-app/tools/call-tools-in-parallel#server)

thats where i found out that the temp values are being repeated twice when we actually execute the code.


lgrammel (approved):

lgrammel:
Thank you!

Issue #161, opened 2023-06-19
Example using Streaming Response for FastAPI.


mattzcarey:
Lots of people write their Langchain apis in Python, not using RSC.

A common tech stack is using FastAPI on the backend with NextJS/React for the frontend. It would be great to show an example of this using [FastAPI Streaming Response](https://fastapi.tiangolo.com/advanced/custom-response/#streamingresponse). 

This would really help us building [Quivr.](https://github.com/StanGirard/quivr).

jasan-s:
@mattzcarey  I'm thinking of using similar tech stack, but it seems that vercel doesn't support python runtime streaming. could you please share your stack in more detail.
I'm currently using langchain js b deployed to vercel edge function and streaming response back to client. But it is apparent that the python version is far more featured, thus my reason to switch.

mattzcarey:
I have managed to do this with langchain callbacks and Streaming Response from FastAPi. You can check out the 'stream' route in the Quivr codebase.

jasan-s:
> @jasan-s I have managed to do this with langchain callbacks and Streaming Response from FastAPi. You can check out the 'stream' route in the Quivr codebase.

Did you deploy quiver to vercel?

mattzcarey:
> > @jasan-s I have managed to do this with langchain callbacks and Streaming Response from FastAPi. You can check out the 'stream' route in the Quivr codebase.
> 
> Did you deploy quiver to vercel?

Yes it can be.

kallebysantos:
I Had create a gist example:
- [Running AI models with FastAPI and Vercel AI SDK](https://gist.github.com/kallebysantos/49b71471b518c01c64e5fb860e096693)

satyamdalai:
Having a native support for converting streaming responses from FastAPI/any other HTTP Server in Next.js API routes (with the help of SDK) will be helpful in my usecase. Since I don't want to directly call FastAPI endpoint using useChat hook, as I manage the authentication layer in Next.js.

kallebysantos:
I think that Issue should be mark as complete.
We had provide useful examples that solves the question.

DanLeininger:
Building off the above answers, here's an example using `experimental_StreamData`:

```python
from openai import AsyncOpenAI

from fastapi import FastAPI
from fastapi.middleware.cors import CORSMiddleware
from fastapi.responses import StreamingResponse

app = FastAPI()

# Added because the frontend and this backend run on separate ports, should change depending on your setup, not a good idea in prod
[...]

PR #2458, opened 2024-07-29
examples (fastapi): add example for python as backend

jeremyphilemon:
Addresses #161 

Add FastAPI example for `useChat` with `streamProtocol: text`.


lgrammel (approved):

jeremyphilemon:
> How are we planning to link this from our docs?

the first place that comes to my mind is [stream protocols](https://sdk.vercel.ai/docs/ai-sdk-ui/stream-protocol).

lgrammel:
> > How are we planning to link this from our docs?
> 
> the first place that comes to my mind is [stream protocols](https://sdk.vercel.ai/docs/ai-sdk-ui/stream-protocol)
> 
> do you have any suggestions?

I agree we should add it to stream protocols. This could be relevant to many people tho, maybe also somewhere under frameworks or guides? cc @nicoalbanese

PR #2437, opened 2024-07-28
Add svelte support for tool calling

lmcgartland:
This PR reimplements the tool calling from Solid/React as best I could track it. There's a bit of code copied between the two useChat files, but I'm not sure how/if/where the maintainers would like to put that sort of shared code + type definitions. 

This PR also removes the deprecated tools/functions fields (keeping it in line with the solid implementation).

Bugs:
- Anthropic API complains about empty content field on messages and throws an error

## Testing

`/routes/api/chat/+page.ts`

```svelte
<script lang="ts">
    import { useChat } from '@ai-sdk/svelte';

    const { input, handleSubmit, messages, data, addToolResult } = useChat({
        maxToolRoundtrips: 5,
        // run client-side tools that are automatically executed:

        async onToolCall({ toolCall }) {
            if (toolCall.toolName === 'getLocation') {
                const cities = ['New York', 'Los Angeles', 'Chicago', 'San Francisco'];

                let city = cities[Math.floor(Math.random() * cities.length)];
                addToolResult({toolCallId: toolCall.toolCallId, result: city});
                return city;
            }
        }
    });
</script>

<main>
    <!-- <pre>{JSON.stringify($data, null, 2)}</pre> -->

    <br />
    <ul>
        {#each $messages as message (message.id)}
            <li>{message.role}: {message.content}</li>
            {#if message.toolInvocations}
                {#each message.toolInvocations as toolInvocation (toolInvocation.toolCallId)}
                    {@const toolCallId = toolInvocation.toolCallId}

                    {#if toolInvocation.toolName === 'askForConfirmation'}
                        <div>
                            {toolInvocation.args.message}
                            <div>
                                {#if 'result' in toolInvocation}
                                    <b>{toolInvocation.result}</b>
                                {:else}
                                    <button on:click={() => addToolResult({ toolCallId, result: 'Yes' })}>Yes</button>
                                    <button on:click={() => addToolResult({ toolCallId, result: 'No' })}>No</button>
                                {/if}
                            </div>
                        </div>
                    {/if}

                    {#if 'result' in toolInvocation}
                        <div>
                            Tool call {`${toolInvocation.toolName}: `}
                            {toolInvocation.result}
                        </div>
                    {:else}
                        <div>Calling {toolInvocation.toolName}...</div>
                    {/if}
                {/each}
            {/if}
        {/each}
    </ul>
    <form on:submit={handleSubmit}>
        <input class="outline-chaplin-1 bg-white outline outline-1" bind:value={$input} />
[...]

Issue #1979, opened 2024-06-17
SvelteKit: support onToolCall / maxToolRoundtrips / tool invocations


chanmathew:
### Description

Hi there,

The onToolCall callback doesn't seem to fire when in a Sveltekit project. See this for repro:
https://stackblitz.com/edit/vercel-ai-sdk-ontoolcall-not-firing?file=src%2Froutes%2F%2Bpage.svelte

I've tested this on local environment as well and same issue.

### Code example

```ts
// client
  const { input, handleSubmit, messages } = useChat({
    async onToolCall({ toolCall }) {
        await console.log(toolCall)
    }
  });
```

```ts
// server
const openai = createOpenAI({
	apiKey: env.OPENAI_API_KEY ?? ''
});

export const POST = (async ({ request }) => {
	const { messages } = await request.json();

	const result = await streamText({
		model: openai('gpt-4o'),
		messages,
		tools: {
			call_tool: tool({
				description: 'Fire this tool call when user asks',
				parameters: z.object({
					tool: z.boolean()
				})
			})
		}
	});

	return result.toAIStreamResponse();
}) satisfies RequestHandler;
```

lgrammel:
`useChat` tool calling is not supported in Svelte yet: https://sdk.vercel.ai/docs/reference/ai-sdk-ui#ui-framework-support

chanmathew:
> `useChat` tool calling is not supported in Svelte yet: https://sdk.vercel.ai/docs/reference/ai-sdk-ui#ui-framework-support

Ah apologies, missed that!

MagsMagnoli:
@lgrammel is there a ticket or some way to track that progress? (if it is in progress even?)

lgrammel:
we could track it here. help would be much appreciated!

lgrammel:
Supported in `@ai-sdk/svelte@0.0.26`

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
The reported callback behavior concerned a capability that was explicitly not yet supported in Svelte, so the issue tracked adding new framework support rather than fixing promised behavior.

### Evidence
- Maintainer clarified that `useChat` tool calling was not supported in Svelte at the time and linked the framework support documentation.
- The issue was then used to track adding Svelte tool-calling support.
- Maintainer later confirmed support was added in `@ai-sdk/svelte@0.0.26`.

PR #2466, opened 2024-07-30
feat (ai/svelte): add tool calling support to useChat

lgrammel:
Continuation of #2437 

## Tasks

- [x] implementation
- [x] update example
- [x] manual testing
- [x] docs updates
- [x] changeset


jeremyphilemon (approved):

lgrammel:
Related issue: https://github.com/vercel/ai/issues/1979

ghoshsanjoy78:
I just cannot seem to get this to work with server side tool calling. Tried the exact example given in the documentation (getWeatherInformation). I can see that the function itself does get called when I console.log the weatherOptions. But the actual output back to the client comes back empty. Any thoughts ?

It seems the first response from the server comes back empty. When I ask GPT again, it shows the answer.


image

Issue #2441, opened 2024-07-28
does @ai-sdk/google-vertex support claude 3.5?


lgrammel:
### Discussed in https://github.com/vercel/ai/discussions/2440

<div type='discussions-op-text'>

<sup>Originally posted by **nalaso** July 28, 2024</sup>
I am using the @anthropic-ai/vertex-sdk, and it works fine with Claude 3.5. However, when I use the same region and project ID with @ai-sdk/google-vertex, it says:

"ClientError: [VertexAI.ClientError]: got status: 429 Too Many Requests. {"error":{"code":429,"message":"Quota exceeded for aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model with base model: anthropic-claude-3-5-sonnet. Please submit a quota increase request. https://cloud.google.com/vertex-ai/docs/generative-ai/quotas-genai.","status":"RESOURCE_EXHAUSTED"}"

Interestingly, the @anthropic-ai/vertex-sdk still works without any error, and I am on a pay-as-you-go plan. There is no way to increase the quota.

the error appers in both streamText and generateText

const vertex = createVertex({
  project: process.env.Vertex_Ai_ProjectID,
  location: process.env.Vertex_Ai_Location,
});

const { text } = await generateText({
  model: vertex('claude-3-5-sonnet@20240620'),
  prompt: content
})

    [https://sdk.vercel.ai/providers/ai-sdk-providers/google-vertex#model-capabilities](https://sdk.vercel.ai/providers/ai-sdk-providers/google-vertex#model-capabilities)
    
    And yeah gemini models works fine with @ai-sdk/google-vertex</div>

nalaso:
I just found out @google-cloud/vertexai doesnt't support thrid party models like claude. [https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude#use_the_anthropic_sdk](https://cloud.google.com/vertex-ai/generative-ai/docs/partner-models/use-claude#use_the_anthropic_sdk)

[https://docs.anthropic.com/en/api/claude-on-vertex-ai](https://docs.anthropic.com/en/api/claude-on-vertex-ai) - @anthropic-ai/vertex-sdk and rest apis are the only ways to access claude from vertex ai.

nalaso:
May I know if there is any workaround to use @anthropic-ai/vertex-sdk with generateText?

HamedMP:
[...]

PR #2482, opened 2024-07-30
Fix: Anthropic vertex provider

nalaso:
Edit
Updated here - [https://sdk.vercel.ai/providers/community-providers/anthropic-vertex-ai](https://sdk.vercel.ai/providers/community-providers/anthropic-vertex-ai)

community package - [https://github.com/nalaso/anthropic-vertex-ai](https://github.com/nalaso/anthropic-vertex-ai)

----------------------------------------
This pr provides support for claude models on vertex ai by introducing a new package, ai/anthropic-vertex.

Solves the issue #2441 

It can't be added to ai/google-vertex because the vertex ai sdk doesn't support third-party models (claude models).
It can't be added to ai/anthropic because vertex models require the google-auth-library for authentication (possible though). Let me know if i should implement this way

example 
`import { anthropicVertex } from '@ai-sdk/anthropic-vertex';

  model: anthropicVertex('claude-3-haiku@20240307'),
  prompt: 'Write a vegetarian lasagna recipe for 4 people.',
});`


packages/anthropic-vertex/src/anthropic-messages-language-model.ts
```diff
@@ -0,0 +1,528 @@
...
+
+  constructor(
+    modelId: AnthropicMessagesModelId,
+    settings: AnthropicMessagesSettings,
+    config: AnthropicMessagesConfig,
+  ) {
+    this.modelId = modelId;
+    this.settings = settings;
+    this.config = config;
+    this.path = `/projects/${config.projectId}/locations/${config.region}/publishers/anthropic/models/${modelId}`;
```
fry69, line 56: For some reason I had to remove the starting `/` from this string to get this provider to work. Otherwise I get an error:
```
13 |     statusCode === 429 || // too many requests
14 |     statusCode >= 500),
15 |     // server error
16 |     data
17 |   }) {
18 |     super(message);
         ^
AI_APICallError: Not Found
 url: "https://europe-west1-aiplatform.googleapis.com/v1//projects/[project id]/locations/europe-west1/publishers/anthropic/models/claude-3-5-sonnet@20240620:rawPredict"

      at new APICallError (.../ai/node_modules/.pnpm/@ai-sdk+provider@0.0.14/node_modules/@ai-sdk/provider/dist/index.mjs:18:5)
      at ../ai/node_modules/.pnpm/@ai-sdk+provider-utils@1.0.5_zod@3.23.8/node_modules/@ai-sdk/provider-utils/dist/index.mjs:458:15
```
Note the double slash after the API base URL. I have no clue there this additional slash comes from and this was the only place I knew where I could get rid of it. It is most certainly the wrong place.

packages/anthropic-vertex/src/anthropic-vertex-provider.ts
```diff
@@ -0,0 +1,174 @@
...
+  provider.languageModel = createChatModel;
+  provider.chat = createChatModel;
+
+  return provider as AnthropicVertexProvider;
+}
+
+/**
+Default Anthropic provider instance.
+ */
+export const anthropicVertex = createAnthropicVertex();
```
fry69, line 174: This line is creates a bug:

When this module gets evaluated `createAnthropicVertex();` gets called.

The default `anthropicVertex` provider then cannot get created with an empty `options` property, because `loadSetting` gets called with environment variables set and will complain that those variables are do not get return a valid value, after falling through, see here -> https://github.com/vercel/ai/blob/main/packages/provider-utils/src/load-setting.ts#L42

This produces an error, even if `createAnthropicVertex()` gets called with a correct `options` property later, practically excluding this method to set `region`, `projectID` and other options programmatically.

It took me a while to trace this error message back to the line mentioned above ->
```
/Users/fry/code-local/ai/node_modules/.pnpm/@ai-sdk+provider-utils@1.0.16_zod@3.23.8/node_modules/@ai-sdk/provider-utils/src/load-setting.ts:45
    throw new LoadSettingError({
          ^


LoadSettingError [AI_LoadSettingError]: Google Vertex project id setting is missing. Pass it using the 'projectId' parameter or the GOOGLE_VERTEX_PROJECT_ID environment variable.
    at loadSetting (/Users/fry/code-local/ai/node_modules/.pnpm/@ai-sdk+provider-utils@1.0.16_zod@3.23.8/node_modules/@ai-sdk/provider-utils/src/load-setting.ts:45:11)
    at createAnthropicVertex (/Users/fry/code-local/ai/packages/anthropic-vertex/src/anthropic-vertex-provider.ts:91:16)
    at <anonymous> (/Users/fry/code-local/ai/packages/anthropic-vertex/src/anthropic-vertex-provider.ts:177:32)
    at ModuleJob.run (node:internal/modules/esm/module_job:262:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:482:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5) {
  cause: undefined,
  [Symbol(vercel.ai.error)]: true,
  [Symbol(vercel.ai.error.AI_LoadSettingError)]: true
}
```

fry69:
FYI: I just published this PR as a community provider via NPM registry -> https://www.npmjs.com/package/anthropic-vertex-ai-provider

Repository: https://github.com/fry69/anthropic-vertex-ai-provider

This is just a first cut to see if everything is working so far (publishing etc). I am willing to develop this a bit further, implement the `LanguageModelV1` spec and write some tests for this provider.

lgrammel:
fwiw i don't think this is the ideal approach bc of the code duplication. what i have in mind is more like the openai & azure provider work (azure builds on top of openai) but I haven't had the time to take a look. I'd probably also integrate it into the vertex provider in some way.

packages/anthropic-vertex/src/anthropic-vertex-provider.ts
fry69: How did this got resolved?

nalaso:
I was wondering if there's a way to get this into the official package. Because claude has one of the top ai models, and vertex is an ai platform from google. Obviously, many people will be waiting to see this work.

In the case of Azure OpenAI and OpenAI, it's possible to get them to work under the same package since they don't require additional dependencies. [https://stackoverflow.com/questions/76711533/how-to-use-the-python-openai-client-with-both-azure-and-openai-at-the-same-time)](https://stackoverflow.com/questions/76711533/how-to-use-the-python-openai-client-with-both-azure-and-openai-at-the-same-time))

For anthropic using vertex, it requires the google auth library, which is irrelevant to include in a package used by users who only access the Anthropic endpoint.

lgrammel:
In the meantime would love to see this as a community provider published to npm

nalaso:
I already have the package(need to refractor). I am using it specifically for one of my project.

fry69:
I just want to mention that the code as is does not work (at least not for me), that is why I fixed it and published it as a community provider (after getting no response for a few days). But if you all have other plans I happily remove my repository (can't unpublish on NPM though).

nalaso:
Let me check. I have an updated code as a separate package where everything works. Last time I checked "anthropicVertex" also works if you specify all the options in the env. I kept this as draft because I thought only I would need this, so I didn't update the PR. Happy to see someone else is actually using anthropic through vertex. And yeah, you don't have to delete the repo. I'll just push it. Am happy if u contribute there.

Issue #2459, opened 2024-07-29
useChat - onFinish Callback not working. 


rohitspujari:
### Description

I can't seem to get the onFinish callback working from useChat hook. It's returning an empty usage object after every tool invocation. 
https://sdk.vercel.ai/docs/reference/ai-sdk-ui/use-chat#on-finish

`{
    "usage": {
        "completionTokens": null,
        "promptTokens": null,
        "totalTokens": null
    },
    "finishReason": "unknown"
}`

What am I missing? I'm using "ai": "^3.2.38" and "@ai-sdk/react": "^0.0.31".

Pitchayut586:
**me to**

BrianHung:
![CleanShot 2024-07-30 at 11 29 01@2x](https://github.com/user-attachments/assets/0eb4721d-baa0-4044-9ab0-4f3ba86f3544)

Looks like it's silently erroring here.

BrianHung:
@lgrammel The bug occurs for me when I use `createOpenAI` as the model. It's because the compatibility by default is 'compatible' which doesn't include usage. As a result, promptTokens and completionTokens are NaN / null.

https://github.com/vercel/ai/blob/72626a22e12df46c6c430d2b6fcc36f4bba34c12/packages/openai/src/openai-chat-language-model.ts#L280-L283

lgrammel:
Fixed in #2491

TrustDec:
After updating the version, it is now working for me. At first, there was no problem with the server in my local area. Later, when it was deployed to the server, it was found that useChat's onFinish had no callback, so I didn't record the reply message. After updating the version, everything was normal.Thanks!

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This is a concrete failure in existing useChat/onFinish stream handling, with a diagnosed implementation cause and a confirmed fix.

### Evidence
- useChat's onFinish receives null token counts and an "unknown" finish reason after tool invocations.
- A commenter traced the failure to createOpenAI compatibility mode omitting usage, producing NaN values serialized as null and triggering stream-part validation.
- The attached stack trace shows an uncaught error while parsing the finish_message stream part.
- A maintainer confirmed the issue was fixed in PR #2491.

PR #2486, opened 2024-07-30
fix onFinish not being called on null promptTokens completionTokens (#2459)

BrianHung:
See comments at https://github.com/vercel/ai/issues/2459#issuecomment-2259076169


lgrammel:
Thank you! Continued in https://github.com/vercel/ai/pull/2491

PR #2491, opened 2024-07-31
fix (ai/ui): parse null as NaN in finish message stream parts

lgrammel:
Continuation of #2486

### DIFF
--- a/.changeset/smart-parents-explain.md
+++ b/.changeset/smart-parents-explain.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/ui-utils': patch
+---
+
+fix (ai/ui): parse null as NaN in finish message stream parts
--- a/packages/ui-utils/src/stream-parts.test.ts
+++ b/packages/ui-utils/src/stream-parts.test.ts
@@ -218,4 +218,15 @@ describe('finish_message stream part', () => {
       },
     });
   });
+
+  it('should parse a finish_message with null completion and prompt tokens', () => {
+    const input = `d:{"finishReason":"stop","usage":{"promptTokens":null,"completionTokens":null}}`;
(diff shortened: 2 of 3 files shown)


jeremyphilemon (approved):

rohitspujari:
Can you share the timeline of when this make it into a release? Or there is a link to a patch that I can download?

lgrammel:
@rohitspujari it's been released

rohitspujari:
Using "ai": "^3.2.41" now - still doesn't seem to be working. Am I missing something? Every onFinish call looks like this
image

BrianHung:
@rohitspujari The issue depends on whether your model provider includes token usage. What model provider and config are you using?

rohitspujari:
gpt-4-turbo with streamText

BrianHung:
@rohitspujari Set compatibility to strict e.g.

```js
const openai = createOpenAI({ apiKey, baseURL, compatibility: 'strict' });
```

rohitspujari:
@BrianHung It's still returning NaNs. Here is my setup 

`const openaiModel = createOpenAI({
    apiKey: process.env.OPENAI_API_KEY,
    compatibility: 'strict',
  });`

`const result = await streamText({
    system: 'You are a helpful assistant. ',
    model: openaiModel('gpt-4-turbo'),
    tools: tools,
  });`

BrianHung:
Double check `@ai-sdk/openai` is also up-to-date. If that doesn't work, try having a custom as an option into `createOpenAI` and try logging if the results include usage directly from OpenAI.

lgrammel:
@rohitspujari have you updated all ai sdk libraries, including the `@ai-sdk/*` libraries, to their latest patch versions?

PR #2499, opened 2024-07-31
docs: link to diagrams for stream data protocol

jeremyphilemon:


lgrammel (changes requested):
We should keep the text in addition to the images. The images will not be accessible for AI and people with limited vision.

lgrammel (approved):

PR #2517, opened 2024-08-01
feat (examples): add basic deepseek coder example

lgrammel:


nicoalbanese (approved):
small nit - does `deepSeek` need to be exported?

lgrammel:
@nicoalbanese good catch!

PR #2541, opened 2024-08-04
feat (provider/azure): update Azure API version to 2024-06-01

BhanuKorthiwada:
- [Azure OpenAI Updates](https://learn.microsoft.com/en-us/azure/ai-services/openai/whats-new)
- [Azure OpenAI Reference](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference)


lgrammel:
Thank you! Continued in https://github.com/vercel/ai/pull/2542 with changeset

PR #2588, opened 2024-08-07
fix (provider/cohere): send last message with request instead of first

jeremyphilemon:


lgrammel (approved):
Great catch! would be good to add a test imo since it's a bug

lgrammel (approved):

lgrammel (approved):

PR #2602, opened 2024-08-08
chore (ai/ui): increase stream data warning timeout to 15 seconds

lgrammel:


jeremyphilemon (approved):

oalexdoda:
Quick question on this @lgrammel since I couldn't find it in the AI SDK docs...

If the completion errors out, how can I close the data stream server-side? 

(Server)

This fires before the completion:

image

And this fires on finish:

image

But how can the data stream close on error? Is there an onError handler on the server-side helper?

image

(Client)

image

Thanks!

PR #2617, opened 2024-08-09
feat (ai/core): mask data stream error messages.

lgrammel:
## Background
Error messages in the stream from the model provider might include sensitive information. The error messages should not be exposed to the client by default.

### DIFF
--- a/content/docs/07-reference/ai-sdk-core/02-stream-text.mdx
+++ b/content/docs/07-reference/ai-sdk-core/02-stream-text.mdx
@@ -770,11 +770,6 @@ for await (const textPart of textStream) {
       ],
     },
     {
-      name: 'toDataStream',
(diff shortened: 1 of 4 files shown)


content/docs/07-reference/ai-sdk-core/02-stream-text.mdx
```diff
@@ -769,11 +769,6 @@ for await (const textPart of textStream) {
         },
       ],
     },
-    {
-      name: 'toDataStream',
-      type: '(callbacks?: AIStreamCallbacksAndOptions) => data stream',
-      description: 'Converts the result to an data stream.',
-    },
```
lgrammel, line 776: removed on purpose (method should not be exposed)

jeremyphilemon (approved):

PR #2624, opened 2024-08-09
add more possible file types in attachments

MrHertal:
Currently you can only attach images to the prompt along with a text.

Some providers are handling more file types. This is the case for [Bedrock Converse API](https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html#conversation-inference-call) which accept documents like pdf, docx etc.

In this PR, I created an additional part type called `file` which is more generic than the current `image` part type.

I didn't want to break anything so I duplicated the existing mechanism of `image` without replacing it.
You can pass the file blob or a URL to the file, just like with the images.

For now I only updated Bedrock provider to handle this new part type. I wanted some feedbacks before working on the other providers.


lgrammel:
Which provider did you use to try this out? Anthropic? are you sure that the contents of the PDF have been used?

MrHertal:
It has been implemented: https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock#file-inputs

PR #2640, opened 2024-08-13
fix (docs): state-management save and restore states

nikhilsnayak:
The `getMutableAIState.done` function accepts the updated state directly as an argument. However, in the provided example, a function returning the updated state was incorrectly used instead, leading to an invalid argument error.

This PR corrects the issue by passing the updated state directly as intended.


lgrammel:
Thank you!

lgrammel (approved):

Issue #1331, opened 2024-04-11
Support tool use when using cohere provider


chanmathew:
### Feature Description

With Cohere's new [Command R+](https://txt.cohere.com/command-r-plus-microsoft-azure/) model, it's a very viable and much cheaper option that can reliably function call. I don't believe the current AI SDK correctly handles the Cohere function responses, would be great to see this added.

### Use Case

Enabling tool use and function calling

lgrammel:
https://github.com/vercel/ai/pull/1954

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests a new provider capability—tool and function calling support for Cohere—rather than reporting a regression or violation of documented behavior.

### Evidence
- Title requests tool-use support for the Cohere provider.
- Description asks to add handling for Cohere Command R+ function responses to enable tool use and function calling.
- Maintainer comment links implementation PR #1954.

PR #2658, opened 2024-08-14
feat (provider/cohere): support tool calls

jeremyphilemon:
Addresses #1331 

- [x] support specifying `tools` during `generateText` and `streamText`
  - [x] limited to primitive types – `number`, `string`, and `boolean`  
- [x] support `tool_calls` and `tool_results` in `convertToCohereChatPrompt`
- [x] support tool call streaming 
- [x] add errors
- [x] add tests
- [x] update docs


lgrammel:
please add a changeset

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +233,29 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+
+              case 'tool-calls-chunk': {
+                if (value.text) {
+                  controller.enqueue({
+                    type: 'text-delta',
+                    textDelta: value.text,
+                  });
+                }
+                return;
+              }
```
lgrammel, line 257: wouldn't this need to be a `tool-call-delta`?

See https://github.com/vercel/ai/blob/main/packages/openai/src/openai-chat-language-model.ts#L461

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -304,3 +367,99 @@ const cohereChatChunkSchema = z.discriminatedUnion('event_type', [
...
+                type = 'float';
+                break;
+              case 'integer':
+                type = 'int';
+                break;
+              case 'boolean':
+                type = 'bool';
+                break;
+              default:
+                throw new Error(`Unsupported JSON type: ${JSONType}`);
```
lgrammel, line 411: It would be good to use a dedicated AI SDK error here, e.g. `UnsupportedFunctionalityError`

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -304,3 +367,99 @@ const cohereChatChunkSchema = z.discriminatedUnion('event_type', [
...
+            }
+          } else if (Array.isArray(JSONType)) {
+            throw new Error('Array types are not supported');
+          } else {
+            throw new Error('Unsupported type');
```
lgrammel, line 416: same

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -304,3 +367,99 @@ const cohereChatChunkSchema = z.discriminatedUnion('event_type', [
...
+    // cohere does not support 'none' tool choice, so we remove the tools:
+    case 'none':
+      return { tools: undefined, force_single_step: false };
+
+    // cohere does not support tool mode directly,
+    // so we filter the tools and force the tool choice through 'any'
+    case 'tool':
+      return {
+        tools: mappedTools.filter(tool => tool.name === toolChoice.toolName),
+        force_single_step: true,
```
lgrammel, line 458: what does this do?

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -304,3 +367,99 @@ const cohereChatChunkSchema = z.discriminatedUnion('event_type', [
...
+      };
+    default: {
+      const _exhaustiveCheck: never = type;
+      throw new Error(`Unsupported tool choice type: ${_exhaustiveCheck}`);
```
lgrammel, line 462: use `UnsupportedFunctionalityError`

packages/cohere/src/convert-to-cohere-chat-prompt.ts
```diff
@@ -71,9 +73,19 @@ export function convertToCohereChatPrompt(
...
       case 'tool': {
-        throw new UnsupportedFunctionalityError({
-          functionality: 'tool role',
+        messages.push({
+          role: 'TOOL',
+          tool_results: content.map(toolResult => ({
+            call: {
+              name: toolResult.toolName,
+              // TODO: Populate parameters
+              parameters: {},
```
lgrammel, line 82: open todo

lgrammel:
Can you add an example, e.g. a cohere copy of https://github.com/vercel/ai/blob/main/examples/ai-core/src/generate-text/amazon-bedrock-chatbot-with-tools.ts for manual testing?

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: done!

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -304,3 +367,107 @@ const cohereChatChunkSchema = z.discriminatedUnion('event_type', [
...
+                });
+            }
+          } else if (Array.isArray(JSONType)) {
+            throw new UnsupportedFunctionalityError({
+              functionality: 'tool call parameter of non-primitive type',
+            });
+          } else {
+            throw new UnsupportedFunctionalityError({
+              functionality: 'tool call parameter of non-primitive type',
+            });
```
lgrammel, line 422: can be collapsed into single `else`

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: `force_single_step` is cohere's way of forcing the model to strictly generate a tool call during a generation

link to docs: https://docs.cohere.com#when-should-i-use-single-step-tool-use

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: done!

packages/cohere/src/convert-to-cohere-chat-prompt.ts
lgrammel: sounds good. imo a note saying that (vs having a todo) might be better here. we don't need to take action, but it would be good to have more context in case we run into this in the future.

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: updated!

examples/ai-core/src/generate-text/cohere-chatbot-with-tools.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (!toolResponseAvailable) {
+      const userInput = await terminal.question('You: ');
+      messages.push({ role: 'user', content: userInput });
+    }
+
+    const { text, toolCalls, toolResults, responseMessages } =
+      await generateText({
+        model: cohere('command-r-plus'),
+        tools: { weatherTool },
+        system: `You are a helpful, respectful and honest assistant. If the weatehr is requested use the `,
```
lgrammel, line 29: nit: typo

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +237,49 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
                 return;
               }
 
+              case 'tool-calls-generation': {
+                for (const toolCall of value.tool_calls) {
+                  controller.enqueue({
+                    type: 'tool-call',
+                    toolCallId: generateId(),
```
lgrammel, line 244: The IDs need to be consistent for a single tool call. Does cohere provide an id? or can you use some other information (index) to identify the call?

Ideally only generate the id once for a particular tool call and then use the same id during streaming those tool call parts and the final tool call.

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +237,49 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+                  });
+                }
+                return;
+              }
+
+              case 'tool-calls-chunk': {
+                if (value.tool_call_delta) {
+                  if (value.tool_call_delta.index != toolCallIndex) {
+                    toolCallIndex = value.tool_call_delta.index;
+                    toolCallId = generateId();
```
lgrammel, line 257: can Cohere have multiple parallel tool calls?

jeremyphilemon: yes!

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: good catch! 

cohere unfortunately does not provide an id ([source](https://docs.cohere.com/reference/chat#response.body.tool_calls)), the workaround is for the sdk to generate an id instead

they do provide an `index` during `tool-call-delta`, but they do not include it during the final `tool-call-generation` chunk, so I feel generating an `id` ourselves is better here since it allows developers to lookup a tool call in `chat_history` by a unique id

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -192,6 +207,11 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
       completionTokens: Number.NaN,
     };
 
+    let toolCallIndex = -1;
+    let toolCallId = '';
```
lgrammel, line 211: do you need `toolCallId` when there are `toolCallIds`? Or would it be easier / less error prone to always use `toolCallIds`? Also, does `toolName` also need to be changed into an array? (unclear why we need `toolCallId` / `toolName` - the array makes sense to me bc there can be multiple tool calls)

jeremyphilemon: I agree that using an array would be best here – updated!

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+                  controller.enqueue({
+                    type: 'tool-call',
+                    toolCallId: toolCalls[index].toolCallId,
+                    toolName: toolCalls[index].toolName,
+                    toolCallType: 'function',
+                    args: JSON.stringify(toolCall.parameters),
+                  });
+
+                  index += 1;
+                }
```
lgrammel, line 291: recommend switching to a regular for loop since you need an index

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
                 return;
               }
 
+              case 'tool-calls-chunk': {
+                if (value.tool_call_delta) {
+                  if (value.tool_call_delta.index != toolCalls.length - 1) {
+                    const toolCallId = generateId();
```
lgrammel, line 244: nit: inline

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+
+                    controller.enqueue({
+                      type: 'tool-call-delta',
+                      toolCallType: 'function',
+                      toolCallId: toolCalls[index].toolCallId,
+                      toolName: toolCalls[index].toolName,
+                      argsTextDelta: value.tool_call_delta.parameters,
+                    });
+                  }
+                }
```
lgrammel, line 274: are these alway 2 separate events or can `name` and `parameters` be set in a single event from Cohere?

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+                    const toolCallId = generateId();
+
+                    toolCalls.push({
+                      toolCallId,
+                      toolName: '',
+                    });
+                  }
+
+                  if (value.tool_call_delta.name) {
+                    const { index } = value.tool_call_delta;
```
lgrammel, line 253: pull out (used earlier in line 243)

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
                 return;
               }
 
+              case 'tool-calls-chunk': {
+                if (value.tool_call_delta) {
+                  if (value.tool_call_delta.index != toolCalls.length - 1) {
```
lgrammel, line 243: what if an earlier index comes again? shouldn't we check `toolCall[index] != null` or something like that?

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,60 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
+                    const toolCallId = generateId();
+
+                    toolCalls.push({
```
lgrammel, line 246: instead of push, you can set at index. it'll increase the size of the array automatically afaik and guarantee that we have a match of `index` to array position

lgrammel (approved):

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: updated!

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: they are 2 separate events – one event for the name and the other event(s) for the args

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: done!

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: agreed, updated!

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: this has been updated to create a new tool call only after checking if the element at said index is empty

also added a test to check for out of order stream deltas [here](https://github.com/vercel/ai/pull/2658/files#diff-437429bac0dbc9688ae63e95f835bc93d1f2a3735948e800fa3192f1718d991dR643)

packages/cohere/src/cohere-chat-language-model.test.ts
```diff
@@ -68,6 +72,50 @@ describe('doGenerate', () => {
...
+              $schema: 'http://json-schema.org/draft-07/schema#',
+            },
+          },
+        ],
+      },
+      prompt: TEST_PROMPT,
+    });
+
+    expect(toolCalls).toStrictEqual([
+      expect.objectContaining({
```
lgrammel, line 108: why is `objectContaining` needed here?

packages/cohere/src/cohere-chat-language-model.test.ts
```diff
@@ -278,6 +475,442 @@ describe('doStream', () => {
...
+
+    const toolCallDelta = responseArray.find(
+      element => element.type === 'tool-call-delta',
+    );
+
+    const toolCall = responseArray.find(
+      element => element.type === 'tool-call',
+    );
+
+    expect(toolCallDelta?.toolCallId).toBe(toolCall?.toolCallId);
```
lgrammel, line 640: this testing involving `generateId` is a little bit brittle. one way to improve it is to inject a `generateId` function in tests, e.g. as in the vertex test or earlier versions of the mistral and anthropic tests (before they had tool ids): https://github.com/vercel/ai/blob/main/packages/google-vertex/src/google-vertex-language-model.test.ts#L28

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -218,6 +238,58 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
                 return;
               }
 
+              case 'tool-calls-chunk': {
+                if (value.tool_call_delta) {
+                  const { index } = value.tool_call_delta;
+
+                  if (!(index in toolCalls) || toolCalls[index] === undefined) {
```
lgrammel, line 245: would `toolCalls[index] === undefined` not be sufficient?

packages/cohere/src/cohere-chat-language-model.test.ts
jeremyphilemon: removed since we pass `generateId` to make the tests deterministic

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: no longer relevant since this code snippet has changed

packages/cohere/src/cohere-chat-language-model.ts
jeremyphilemon: updated!

packages/cohere/src/cohere-chat-language-model.test.ts
jeremyphilemon: til, thanks! updated the tests to use custom `generateId`

packages/cohere/src/cohere-chat-language-model.test.ts
```diff
@@ -278,6 +475,442 @@ describe('doStream', () => {
...
+        finishReason: 'stop',
+        type: 'finish',
+        usage: {
+          completionTokens: 62,
+          promptTokens: 893,
+        },
+      },
+    ]);
+
+    // Check if the tool call ID is the same in the tool call delta and the tool call
```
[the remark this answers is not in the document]
jeremyphilemon: done!

PR #2678, opened 2024-08-15
[prototype] language model v2 spec

lgrammel:
## Why
- provider-specific message and message part extensions
- provider-specific response / finish metadata
- support for various data parts, both as input and output
- eliminate differences between regular, object-json, object-tool modes
- return ids & response model for telemetry
- return raw response information to help with debugging
- support models that provide sources (for RAG)

## Tasks
- [x] draft language model v2 spec
- [ ] prototypes
   - [x] Anthropic: prompt caching
   - [ ] OpenAI: custom role names
   - [ ] google gen ai: pdf support
   - [ ] google vertex: sources
   - [ ] OpenAI: log response model & id
- [ ] regression check
   - [ ] streamobject tool
   - [ ] streamobject json
   - [ ] generateobject tool
   - [ ] generateobject json
- [ ] implementation
   - [ ] AI SDK
     - [ ] generateText
     - [ ] streamText
     - [ ] generateObject
     - [ ] streamObject
     - [ ] streamUI
   - [ ] Providers


packages/provider/src/language-model/v2/language-model-v2-prompt.ts
```diff
@@ -0,0 +1,149 @@
...
+        | LanguageModelV2ToolResultPart
+      >;
+
+      /**
+       * Additional provider-specific metadata. They are passed through
+       * to the provider from the AI SDK and enable provider-specific
+       * functionality that needs to be applied per message,
+       * e.g. the OpenAI name parameter.
+       */
+      providerMetadata: Record<string, JSONValue> | undefined;
```
kdawgwilk, line 33: I am probably missing something but couldn't you just add this field in a backwards compatible way to the LanguageModelV1 types since the property is optional?

lgrammel: Good point. My preference is to move to a cleaner language model, but considering the need to get prompt caching ready sooner rather than later I'll see if that's an option.

PR #2692, opened 2024-08-16
feat: add support for anthropic prompt caching

### LINKED ISSUE #2689: Support prompt caching
### Feature Description

Add support for setting `cache_control` on `CoreMessage` to support new Anthropic feature https://www.anthropic.com/news/prompt-caching

### Use Case

Cost savings when reusing prompt context

### Additional context

_No response_

kdawgwilk:
First time contributor here 👋 

I think i am on the right track for this but would love feedback.  Some of my concerns are that this is a provider specific feature but because it requires fields set on the message level i think that means those fields have to propagate up to the generalized message types which i don't love.

Closes #2689

### DIFF
--- a/packages/ai/core/prompt/message.ts
+++ b/packages/ai/core/prompt/message.ts
@@ -22,7 +22,7 @@ export type CoreMessage =
  to increase the resilience against prompt injection attacks,
  and because not all providers support several system messages.
  */
-export type CoreSystemMessage = { role: 'system'; content: string };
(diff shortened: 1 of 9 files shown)


lgrammel:
Hey thanks for the contribution!

**I do not want to add a provider-specific mechanism that is likely to change in the future to CoreMessage**. Instead I plan to add a mechanism that will let us add provider-specific features without affecting the core message. 

You can find a prototype for anthropic caching here as well:

https://github.com/vercel/ai/pull/2678

lgrammel:
Superseeded by https://github.com/vercel/ai/pull/2697

PR #2722, opened 2024-08-19
fix: 2721 setThreadId is changing frequently

### LINKED ISSUE #2721: setThreadId is changing frequently
### Description

setThreadId function in package/react/src/use-assistant.ts : whenever input or messages change, the setThreadId function is changing.

i am using searchParam to get the threadId for my application. whenever the threadId in the searchParam changes, i am setting the threadId of the current chat by using "setThreadId" function from useAssistant() hook.
now by following react-hooks rules, I’ve included setThreadId as a dependency in the useEffect hook. However, this results in the useEffect firing even when the input or messages change, because the setThreadId function is changing frequently.

### Code example

``` typescript
useEffect(() => {
  if (currentThreadId) {
    getHistory(currentThreadId).then(data => {
      setHistory(data);
    });
    setThreadId(currentThreadId);
  }
}, [currentThreadId, getHistory, setThreadId]);

```

here if i not include setThreadId in the dependency, then everything works fine. but actually it is depending on "setThreadId"

### Additional context

_No response_

ravvi-kumar:
fix & closes : #2721 
wrap setThreadId in a useCallback to prevent frequent change in the function

### DIFF
--- a/packages/react/src/use-assistant.ts
+++ b/packages/react/src/use-assistant.ts
@@ -274,10 +274,13 @@ export function useAssistant({
     append({ role: 'user', content: input }, requestOptions);
   };
 
-  const setThreadId = (threadId: string | undefined) => {
-    setCurrentThreadId(threadId);
-    setMessages([]);
-  };
+  const setThreadId = useCallback(
+    (threadId: string | undefined) => {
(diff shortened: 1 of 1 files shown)


lgrammel:

PR #2729, opened 2024-08-19
fix: allow sending empty messages with attachments

### LINKED ISSUE #2725: Using experimental_attachments with empty input doesn't append the message
### Description

When using `useChat`, if the input is empty but a list of `experimental_attachments` is passed to `handleSubmit` with `allowEmptySubmit` set to `true`, no new message is appended to the list sent to the server.

https://github.com/vercel/ai/blob/f95c4b41525257d5494127bf141317044312feff/packages/react/src/use-chat.ts#L607

gdaybrice:
Fixes https://github.com/vercel/ai/issues/2725

### DIFF
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -604,7 +604,7 @@ By default, it's set to 0, which will disable the feature.
       };
 
       const messages =
-        !input && options.allowEmptySubmit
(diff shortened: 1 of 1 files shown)


lgrammel:
Thanks! Continued in https://github.com/vercel/ai/pull/2732

PR #2742, opened 2024-08-20
fix (ai/core): explicitly import Buffer

### LINKED ISSUE #2741: Implicit `Buffer` usage fails in Cloudflare Workers
### Description

Cloudflare Workers fail on startup if `streamText` is imported:

```ts
```

This is the error:

```
Uncaught ReferenceError: Buffer is not defined
    at null.<anonymous> (index.js:9799:16)
```

Which is pointing to this line: https://github.com/vercel/ai/blob/ddf0184b69341468898d59c5c9a0276d3e90dc76/packages/ai/core/prompt/data-content.ts#L20

The Node.js documentation recommends importing `Buffer` explicitly:

> While the `Buffer` class is available within the global scope, it is still recommended to explicitly reference it via an import or require statement.

https://nodejs.org/docs/latest-v22.x/api/buffer.html#buffer

If `Buffer` explicitly imported and Cloudflare Workers have `nodejs_compat` mode enabled, this works without issues.

### Code example

_No response_

### Additional context

_No response_

danielgavrilov:
Fixes #2741.

The Node.js documentation recommends importing `Buffer` explicitly:

> While the `Buffer` class is available within the global scope, it is still recommended to explicitly reference it via an import or require statement.

https://nodejs.org/docs/latest-v22.x/api/buffer.html#buffer

Without importing `Buffer` explicitly, this fails to run in a Cloudflare Worker.

### DIFF
--- a/packages/ai/core/prompt/data-content.ts
+++ b/packages/ai/core/prompt/data-content.ts
@@ -2,6 +2,7 @@ import {
   convertBase64ToUint8Array,
   convertUint8ArrayToBase64,
 } from '@ai-sdk/provider-utils';
+import { Buffer } from 'node:buffer';
 import { InvalidDataContentError } from './invalid-data-content-error';
 import { z } from 'zod';
 


lgrammel (changes requested):
This might break the AI SDK outside of node.js (e.g. in browsers). Please verify that this does not break various other environments (browser, deno, etc)

lgrammel:
Please verify that this does not break various other environments (browser, deno, etc)

lgrammel:
@danielgavrilov can you check if the following would work instead?

```ts
/**
@internal
 */
export const dataContentSchema: z.ZodType<DataContent> = z.union([
  z.string(),
  z.instanceof(Uint8Array),
  z.instanceof(ArrayBuffer),
  z.custom(
    // Buffer might not be available in some environments, so we check for it.
    (obj: unknown): obj is Buffer =>
      typeof globalThis.Buffer !== 'undefined' &&
      globalThis.Buffer.isBuffer(obj),
    {
      message: 'Must be a Buffer',
    },
  ),
]);
```

lgrammel:
Continued in #2749

PR #2757, opened 2024-08-21
feat: exclude tools from automatic roundtrips

### LINKED ISSUE #2693: Override maxToolRoundtrips per tool call
### Feature Description

Following this Twitter discussion: https://x.com/blechatellier/status/1823688077006790917 with @lgrammel 

It would be great to allow setting a tool call result (on either the server or client) while controlling whether to automatically trigger another call to the LLM.

This use case involves rendering a custom UI component on the client using the tool call result, without needing a roundtrip to the LLM. However, it should still allow multiple roundtrips for other tools.

One possible solution could be overriding maxToolRoundtrips per tool call, or passing a flag to addToolResult on the client side to prevent triggering another LLM call. A similar mechanism would be needed on the server if the tool call is running via the execute method.

Or maybe exposing `triggerRequest` so I can recall the LLM for some specific tools only.

gdaybrice:
Closes https://github.com/vercel/ai/issues/2693

### DIFF
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -405,8 +405,8 @@ By default, it's set to 0, which will disable the feature.
         // limit the number of automatic roundtrips:
         countTrailingAssistantMessages(messages) <= maxToolRoundtrips &&
         // check if any tools are excluded from roundtrips:
-        !lastMessage.toolInvocations?.some(invocation => 
-          excludeToolsFromRoundtrips?.includes(invocation.toolName)
+        !lastMessage.toolInvocations?.some(invocation =>
+          excludeToolsFromRoundtrips?.includes(invocation.toolName),
         )
       ) {
         await triggerRequest({ messages });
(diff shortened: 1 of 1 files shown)


lgrammel:
Thanks @blechatellier for the PR. I think the overall solution for deciding what to do on each roundtrip (esp. with server side roundtrips) is more complex, e.g. prompt changes and full state machines need to be considered. Including your solution into the API would mean we need to maintain and extend it, and coordinate how it works with said future features. I would therefore prefer not to include it in the AI SDK.

Can you achieve the same by using tools without an `execute` function?

blechatellier:
@lgrammel no worries - I'll close this. Thanks for implementing the roundtrips on the server side; they're very useful. I'm still figuring out how to best implement a simple use case where you can conditionally switch between different instructions within the same LLM call (e.g., drafting an email vs. answering a question) and render a different UI on the client. This is something I've been able to do easily with createStreamableUI and nested LLM calls but not with streamText. I’m attempting to replicate this with tools, but I feel it might be an anti-pattern.

PR #2759, opened 2024-08-21
feat (ai/core): middleware support

lgrammel:
## Summary
* Adds `LanguageModelV1Middleware` type
* Adds `wrapLanguageModel` function for applying middleware to language models

## Background
Strategies for different tasks such as guardrails, RAG, caching, and logging are independent of the models that they are applied to. We need a way to separate them from the language models, so that the code for these tasks can be developed and distributed independently, while still being compatible with the AI SDK functions (`streamText`, `generateObject`, etc.).

## Tasks
- [x] implementation
   - [x] pass model
   - [x] override model id, provider id
   - [x] write tests
- [x] examples
   - [x] log middleware
   - [x] guardrail middleware
   - [x] rag middleware
   - [x] cache middleware
- [x] documentation
   - [x] jsdoc
   - [x] write middleware guide
   - [x] add reference docs
- [x] changeset


content/docs/03-ai-sdk-core/45-middleware.mdx
```diff
@@ -0,0 +1,191 @@
...
+    return result;
+  },
+
+  // here you would implement the caching logic for streaming
+};
+```
+
+### Retrieval Augmented Generation (RAG)
+
+This example shows how to use RAG as middleware.
```
nicoalbanese, line 137: maybe implement rather than use?

content/docs/03-ai-sdk-core/45-middleware.mdx
```diff
@@ -0,0 +1,191 @@
...
+
+<Note>
+  Helper functions like `getLastUserMessageText` and `findSources` are not part
+  of the AI SDK. They are just used in this example to illustrate the concept of
```
nicoalbanese, line 141: maybe "the (conceptual/theoretical) steps of" rather than "the concept of"

nicoalbanese (approved):

PR #2789, opened 2024-08-23
feat (docs): prompt caching with google genai provider

dyeoman2:
Example for caching content using Google Generative AI language models


lgrammel (approved):
Thank you!

Issue #2214, opened 2024-07-08
getMutableAIState(key [keyof AI Context]) update() and done() didn't change object reference


btxakita:
### Description

`getMutableAIState` function provided a way to only mutate a specific `key` from the AIState, which is very convenient. Say we have a "messages" in the AIState object. You could use `getMutableAIState<AI>("messages").`

But when you call update() / done(). Things went out of control:

- In done() function, it use `jsondiffpatch` to find the differences between `originalState` and `currentState` from the `store` object, while when we update the object's specific key, these two states are **always the same** object. So the function just cannot find the delta.

Which then leads to other problems on client side context updates etc...

### Code example

`packages/core/rsc/ai-state.tsx`
```
// store object
return asyncAIStateStorage.run(
    {
      currentState: `state`,
      originalState: `state`,
      sealed: false,
      options,
    },
    fn,
  );`

// done() function 
const delta = jsondiffpatch.diff(`store.originalState`, `store.currentState`);
store.mutationDeltaResolve!(delta);
```
### Additional context

Also, there're many type errors regarding this scenario.

jeremyphilemon:
Closed by #2821

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
The issue reports a concrete failure in existing keyed AI state mutation behavior: updates are not detected or propagated because state references remain identical.

### Evidence
- Calling `getMutableAIState<AI>(key).update()` or `.done()` leaves `originalState` and `currentState` referencing the same object.
- Because `jsondiffpatch.diff` receives identical state objects, it cannot detect a mutation delta, causing client-side context updates to fail.
- A maintainer comment states the issue was closed by PR #2821.

Issue #2627, opened 2024-08-10
AI State History Not Updating


yasserelsaid:
### Description

I am working on creating a playground to compare multiple chatbot instances using Vercel's AI SDK. Below is a simplified representation of our AI state structure:

```
// AI State Structure
type PlaygroundCompareAIState = {
  chatbots: Map<string, PlaygroundCompareChatbotAIState>
}

type PlaygroundCompareChatbotAIState = {
  history: MessageWithRole[]
}
```
As you can see, it is a map of instances, and each instance has its own history of messages array.

Problem:

When updating the history in our chat action (server side) using the following code:
```
const aiState = getMutableAIState()
const newAIState = aiState.get()
const chatbotAIState = newAIState.chatbots.get(instanceID)

// Update the history
chatbotAIState.history = [...newHistory]
newAIState.chatbots.set(instanceID, chatbotAIState)

aiState.done({
  ...aiState.get(),
  ...newAIState
})
```
The history is not updating at all and remains an empty array. Even when we log the new AI state before calling .done(), the history is still empty.

Expected Behavior:

The history array should be updated with the new messages, reflecting the latest conversation state for each chatbot instance.

Actual Behavior:

The history array remains empty, and no changes are reflected.

Thank you for your assistance!

lgrammel:
@yasserelsaid does this also happen when you use a record/object instead of a map?

aayush-kapoor:
Marking this as closed for now - if you face any regression or have a similar problem after migrating to the latest AI SDK version, please feel free to open a new issue (with possible reproduction)

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **Medium**

### Rationale
This reports a concrete failure of existing AI state update behavior rather than requesting a new capability. Confidence is limited because the issue lacks a reproduction, may depend on unsupported `Map` handling, and was later closed pending confirmation on a current SDK version.

### Evidence
- `getMutableAIState().done()` does not persist updates to a chatbot history stored inside a `Map`; the history remains empty even before the final update call.
- The reporter expected the existing AI state update API to retain the new messages.
- A maintainer asked whether the failure also occurs with a record/object, suggesting the `Map` representation may be relevant; no reproduction or resolution was provided.

Issue #1509, opened 2024-05-07
aistate messages dissapear?


wolfpixels:
### Description

So I'm having the same issue as this user: https://github.com/vercel/ai/discussions/1290

I'm using the vercel ai chat template and I've implemented a custom ai provider. 
Like shaded-blue mentioned, messages dissapear from my aistate.

`sync function submitUserMessage(content: string) {
  'use server'

  const aiState = getMutableAIState<typeof AI>()

  console.log('this is content', content)

  // Retrieve the current messages before the update
  const currentMessages = aiState.get().messages;

  // Log the number of messages currently stored
  console.log('Current number of messages before update:', currentMessages.length);

  // Log each message in the current state
  currentMessages.forEach((msg, index) => {
    console.log(`Message ${index + 1}:`, msg);
  });
`

When I have this in chat/lib/chat/actions.tsx - on each user message the current messages performs weird, showing different things like 0/1 

### Code example

sync function submitUserMessage(content: string) {
  'use server'

  const aiState = getMutableAIState<typeof AI>()

  console.log('this is content', content)

  // Retrieve the current messages before the update
  const currentMessages = aiState.get().messages;

  // Log the number of messages currently stored
  console.log('Current number of messages before update:', currentMessages.length);

  // Log each message in the current state
  currentMessages.forEach((msg, index) => {
    console.log(`Message ${index + 1}:`, msg);
  });

jeremyphilemon:
Hey @wolfpixels, sorry about this! I left a comment in this [thread](https://github.com/vercel/ai/discussions/1290) to investigate this behavior

aayush-kapoor:
Marking this as closed for now - if you face any regression or have a similar problem after migrating to the latest AI SDK version, please feel free to open a new issue (with possible reproduction)

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes inconsistent loss of existing AI state messages, which is a concrete failure of current behavior rather than a request for new capability.

### Evidence
- The reporter states that messages disappear from `getMutableAIState()` while using the Vercel AI chat template and a custom provider.
- Logging `aiState.get().messages` reportedly produces inconsistent message counts such as 0 or 1 across submissions.
- A maintainer apologized and said they would investigate the behavior; the later closure only requested a new reproduction against the latest SDK and did not reclassify it as expected behavior.

PR #2821, opened 2024-08-27
fix (rsc): Deep clone currentState in getMutableState()

jeremyphilemon:
## Changes
- [x] Deep clone the `currentState` so it's not the same as `originalState`
- [x] Add tests 

Addresses #2214 

## Background
When updating the AI State initially accessed using a key (e.g., `const aiState = getMutableState("key")`), changes weren't being detected because `currentState` and `originalState` referenced the [same object](https://github.com/vercel/ai/blob/e1063ce787b5742dc601b3be0cfd48e61f39f67e/packages/ai/rsc/ai-state.tsx#L38). Updates to `object[key]` modified this shared object, resulting in an empty [delta](https://github.com/vercel/ai/blob/e1063ce787b5742dc601b3be0cfd48e61f39f67e/packages/ai/rsc/ai-state.tsx#L202) when comparing states. As a result, state updates weren't propagating as expected.

### DIFF
--- a/.changeset/twelve-dragons-help.md
+++ b/.changeset/twelve-dragons-help.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix (rsc): Deep clone currentState in getMutableState()
--- a/packages/ai/rsc/ai-state.test.ts
+++ b/packages/ai/rsc/ai-state.test.ts
@@ -1,0 +1,146 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import {
+  withAIState,
+  getAIState,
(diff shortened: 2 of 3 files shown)


lgrammel (approved):

lgrammel:
Awesome - does this close some of the existing RSC-related issues https://github.com/vercel/ai/issues?q=is%3Aissue+is%3Aopen+label%3Aai%2Frsc+aiState

jeremyphilemon:
@lgrammel I suspect this being the root cause for #1532, #1509, #1290, #2627

will reproduce the issue for each of them and close if they're resolved by this!

PR #2881, opened 2024-09-03
fix (@ai-sdk/openai): doStream exception for initial chunk arguments

minpeter:
This PR addresses an issue where `doStream` ignores function arguments when a provider sends them in the first chunk of the streamed response. The current implementation assumes that the first chunk will not contain function arguments, leading to the omission of these arguments and the subsequent failure of the `doStream` method.

### Changes:

- **Test Update**: Added a test case in `openai-chat-language-model.test.ts` to simulate a scenario where arguments are included in the first chunk of a streamed response.
- **Code Update**: Modified `openai-chat-language-model.ts` to handle cases where function arguments are present in the initial chunk. The function now checks for non-empty arguments and processes them as tool-call-delta events.

### Reason:  
When using the `streamObject()` function in tool mode with FriendliAI serverless, the first chunk's arguments were being ignored, causing incomplete data processing. 

Response body of `streamObject()` from [friendliai inference endpoint](https://docs.friendli.ai/openapi/create-chat-completions): `expense": {"category": "OTHER", "amount": 0.0, "date": "03-Sep", "details": "dd", "participants": []}}`

This issue arose due to the lack of explicit documentation in the OpenAI API, leading to the streaming process ignoring chunks that include arguments from the start.

This fix ensures that the doStream function correctly processes all parts of the response, preventing errors and ensuring that the full payload is accurately streamed and handled.


lgrammel:
Great PR, thanks for the bugfix. There are a few changes I want to make so I'll continue it in a separate PR.

lgrammel:
Continued in https://github.com/vercel/ai/pull/2883

PR #2883, opened 2024-09-03
fix (provider/openai): support tool call delta when arguments are sent in the first chunk

lgrammel:
Continues #2881


nicoalbanese (approved):

minpeter:
thanks :)

PR #2918, opened 2024-09-06
feat (provider/openai): support image detail

lgrammel:


examples/ai-core/src/generate-text/openai-multimodal-url-download.ts
```diff
@@ -0,0 +1,34 @@
...
+        content: [
+          { type: 'text', text: 'Describe the image in detail.' },
+          {
+            type: 'image',
+            image:
+              'https://github.com/vercel/ai/blob/main/examples/ai-core/data/comic-cat.png?raw=true',
+
+            // OpenAI specific extension: image detail:
+            experimental_providerMetadata: {
+              openai: { detail: 'low' },
```
jeremyphilemon, line 23: question – do we enforce the types so that `detail` is only allowed for image message parts?

lgrammel: no - provider extensions have no type enforcement

lgrammel: you can set any json (it's an escape hatch that comes with this downside)

lgrammel: openai calls it `detail` and i want to be in sync with that since it's part of image (considered `imageDetail`)

lgrammel: changing it to `imageDetail` won't help with type safety or anything, since you can enter anything you want into any provider extension

lgrammel: (in general, with provider extensions, my aim is to mimic whatever they use in their api as closely as possible, to avoid confusion)

lgrammel: i guess it's debatable here - I'll change it since we are on a slightly different level (image element vs within image_url)

jeremyphilemon: makes sense to have match the provider spec 1:1 for this, got it!

jeremyphilemon (approved):

PR #2923, opened 2024-09-07
Fix `streamUI` `onFinish` callback executing before rendering results

### LINKED ISSUE #2921: Tool execution not awaited in `streamUI`
### Description

- The [streamUI docs](https://sdk.vercel.ai/docs/reference/ai-sdk-rsc/stream-ui) state that the [onFinish]() callback "is called when the LLM response and all request tool executions (for tools that have a `generate` function) are finished." A [comment](https://github.com/vercel/ai/blob/a9de7e0a7c72aff1e376b83b7040bb1261fddbe3/packages/ai/rsc/stream-ui/stream-ui.tsx#L194-L197) in the source code also suggests that this is the intended behaviour.

- However, this is not the case. The `onFinish` cb is executed immediately after the final chunk ("finish" type) is [received](https://github.com/vercel/ai/blob/a9de7e0a7c72aff1e376b83b7040bb1261fddbe3/packages/ai/rsc/stream-ui/stream-ui.tsx#L194-L197).

- In the example [vercel-ai-chatbot](https://github.com/vercel/ai-chatbot) repo, the updating of `aiState` is tightly coupled with the tool generators, being inlined. The `onFinish` callback isn't used in that repo. However, if we need to ensure that the UI has updated before marking `aiState` as done, and we don't want to handle all of this state management within the generators, then we have no way to await the result of `streamUI`. 

- Personally, I am streaming both text content and tool call results from the same message via `streamUI`, so ideally would to be able to wait for the generate method to be completed before updating state and closing a text stream, rather than putting all of this state management into each tool's generator.

### Steps to reproduce:
- Clone the [vercel-ai-chatbot](https://github.com/vercel/ai-chatbot) repo.
- Add logging anywhere before the return statement of a tool's generator body. e.g. `console.log('tool_execution_done')`
- Add an `onFInish` callback.
```tsx
    onFinish: async event => {
      console.log('ON_FINISH', event.finishReason);
    },
 ```
I am using what is currently the latest version (3.3.28), and it is also occurring on the demo repo (3.2.16).

### Code example

_No response_

### Additional context

Something similar seems to have been recently fixed for `streamText` - https://github.com/vercel/ai/pull/2859

Saran33:
- FIxes: https://github.com/vercel/ai/issues/2921
- Wait for rendering promise to resolve before calling  `onFinish` in `streamUI`.
- Set a `finishEvent` object on the last message chunk and pass it to `onFinish` after the `renderFinished` has resolved instead of executing `onFinish` immediately.

### DIFF
--- a/packages/ai/rsc/stream-ui/stream-ui.tsx
+++ b/packages/ai/rsc/stream-ui/stream-ui.tsx
@@ -178,6 +178,15 @@ export async function streamUI<
 
   let finished: Promise<void> | undefined;
 
+  let finishEvent: {
+    finishReason: FinishReason;
+    usage: LanguageModelUsage;
+    warnings?: CallWarning[];
+    rawResponse?: {
+      headers?: Record<string, string>;
+    };
+  } | null = null;
+
   async function render({
     args,
     renderer,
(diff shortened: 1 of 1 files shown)


lgrammel:
Continued in https://github.com/vercel/ai/pull/2980

PR #2930, opened 2024-09-08
Fix: Error calling tools on Gemini.

bhavna3:
This pull request fixes the issue where the Gemini tool was causing the "failed to pipe response" error due to improper handling of the streaming response. 

The solution buffers the streamed response before processing, which resolves the error.

Changes include:
Buffering the streamed text response to avoid pipe failure.
Ensuring the full response is concatenated and handled correctly before output.


lgrammel:
Unrelated to the issue. Potentially AI generated slop.

PR #2935, opened 2024-09-09
feat (provider/amazon-bedrock): support guardrails

OscarBarrett:
[Amazon Bedrock Guardrails](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) can be used to implement safeguards and modify model responses.

This PR alllows for a guardrail config to be passed, and returns trace information (when tracing is enabled) in provider metadata.

## Example

```ts
  bedrock('anthropic.claude-3-sonnet-20240229-v1:0'),
  experimental_providerMetadata: {
    bedrock: {
      guardrailConfig: {
        guardrailIdentifier: '1abcd2ef34gh',
        guardrailVersion: '1',
        trace: 'enabled' as const,
        streamProcessingMode: 'async',
      },
    },
  },
});
```

Using trace information:

```ts
if (result.providerMetadata?.bedrock.trace) {
  // ...
}
```

### Guarding only the latest message in a conversation

It's possible to guard specific contents of a message using a `guardContent` field ([as per the docs](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html#guardrails-use-converse-api-call-message)) rather than applying to the whole conversation. It is not immediately clear to me if this is possible to implement using `experimental_providerMetadata`?

`guardContent` isn't allowed right now due to the CoreMessage schema validation.

Without doing this, if a message in a conversation triggers the guardrail then any subsequent messages in the conversation will continue to result in the guardrail being triggered.


lgrammel:
Request-level provider metadata is what I would use for this - but unfortunately it's not available in the SDK yet. I'll look into adding.

lgrammel:
I just released `ai@3.3.32` which contains support for provider metadata on a request level.

I would imagine something like:

```ts
    model: bedrock('anthropic.claude-3-haiku-20240307-v1:0'),
    experimental_providerMetadata: {
      // scope to bedrock, since there may be other provider configs in here as well
      bedrock: {
        "guardrail": {
          "inputAssessment": {
            "1abcd2ef34gh": {
              "contentPolicy": {
                "filters": [{
                  "action": "BLOCKED",
                  "confidence": "LOW",
                  "type": "INSULTS"
                }]
              },
              "wordPolicy": {
                "managedWordLists": [{
                  "action": "BLOCKED",
                  "match": "<rude word>",
                  "type": "PROFANITY"
                }]
              }
            }
          }      
        }
      }
    }
  });
```

OscarBarrett:
Thanks @lgrammel - updated to use provider metadata.

lgrammel:
Continued in #2968

PR #2943, opened 2024-09-09
feat (docs): improve explanation of roundtrips

lgrammel:


content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -85,6 +85,14 @@ const { text, roundtrips } = await generateText({
 });
 ```
 
+In the above example, there is a single roundtrip:
+
+1. **Call 1**: The prompt is sent to the model.
+1. The model generates a tool call.
```
nicoalbanese, line 91: maybe nested items for the actions that are happening within the call?

nicoalbanese (approved):

PR #2951, opened 2024-09-10
Removing unnecessary calls to mutateStreamData in useChat React hook

### LINKED ISSUE #2950: React: mutating stream data unnecessarily causes other functions being recreated even though they are enclosed by useCallback
### Description

In use-chat.ts:
```typescript
...
onUpdate(merged, data) {
  mutate([...chatRequest.messages, ...merged], false);
  mutateStreamData([...(existingData || []), ...data], false);
},
...
```

`data` is an empty array in my use case and it is causing `streamData` being modified. As it is a dependency for `triggerRequest` and `triggerRequest` is a dependency for `append`, `reload` and `handleSubmit`, `useCallback` returns a new function every time an update occurs when reading from the stream.

A simple modification avoids this behavior (and consequently unnecessary re-renders):
```typescript
... 
onUpdate(merged, data) {
  mutate([...chatRequest.messages, ...merged], false);
  if (data?.length) {
    mutateStreamData([...(existingData || []), ...data], false);
  }
},
...
```

### Code example

_No response_

### Additional context

_No response_

tonyfarney:
This fixes #2950


lgrammel:
Hey @tonyfarney meant to come back to this - the issue is legit, but I need to verify this fix actually works in different settings, e.g.. when something is already non-empty or when there are non-changes for non-empty data.

tonyfarney:
> Hey @tonyfarney meant to come back to this - the issue is legit, but I need to verify this fix actually works in different settings, e.g.. when something is already non-empty or when there are non-changes for non-empty data.

Looking the code, the only situation data is not an empty array when passed to `onUpdate` is in `process-data-protocol-response.ts`:
```typescript
update([...previousMessages, ...merged], [...data]); // make a copy of the data array
```
I think it's possible the situation of calling the `update` callback for the exact same `data`. But for avoiding to mutate stream data in such situation, would be necessary to compare the arrays or maybe just to check the array length.

tonyfarney:
Hey @lgrammel! Do you think my changes will be accepted and merged?

lgrammel:
@tonyfarney sorry meant to get back to this - yes still planning to verify & merge. we are moving into the 4.0 feature freeze so there'll be additional delay. thanks for your patience.

lgrammel:
Continued in https://github.com/vercel/ai/pull/3703

PR #2952, opened 2024-09-10
feat (ai/core): add output: enum to generateObject

lgrammel:
## Background
Classification tasks are common in AI workflows. This PR improves support for such tasks.

## Tasks

- [x] implementation
   - [x] basic implementation
   - [x] enum mode validation
   - [x] enum mode testing
- [x] examples
   - [x] openai 
   - [x] gemini
- [x] docs
   - [x] reference
   - [x] guide
- [x] changeset

### DIFF
--- a/content/docs/03-ai-sdk-core/10-generating-structured-data.mdx
+++ b/content/docs/03-ai-sdk-core/10-generating-structured-data.mdx
@@ -101,7 +101,8 @@ for await (const hero of elementStream) {
 
 ### Enum
 
-If you want to generate a specific enum value, you can set the output strategy to `enum`
+If you want to generate a specific enum value, e.g. for classification tasks,
+you can set the output strategy to `enum`
 and provide a list of possible values in the `enum` parameter.
 
 <Note>Enum output is only available with `generateObject`.</Note>
--- a/content/docs/03-ai-sdk-core/10-generating-structured-data.mdx
+++ b/content/docs/03-ai-sdk-core/10-generating-structured-data.mdx
@@ -106,7 +106,7 @@ and provide the list of possible values in the `enum` parameter.
 
 <Note>Enum output is only available with `generateObject`.</Note>
 
-```ts
+```ts highlight="5-6"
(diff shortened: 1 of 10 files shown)


content/docs/03-ai-sdk-core/10-generating-structured-data.mdx
```diff
@@ -99,7 +99,28 @@ for await (const hero of elementStream) {
 }
 ```
 
-### Output Strategy: No Schema
+### Enum
+
+If you want to generate a specific enum value, you can set the output strategy to `enum`
+and provide the list of possible values in the `enum` parameter.
```
nicoalbanese, line 105: "a list" rather than "the list"

packages/provider/src/language-model/v1/language-model-v1-call-options.ts
```diff
@@ -21,11 +21,10 @@ The model can take this information and e.g. configure json mode, the correct
...
 streaming, e.g. tool-delta stream parts are only needed in the
 object-tool mode.
-
-@deprecated mode will be removed in v2.
-All necessary settings will be directly supported through the call settings,
-in particular responseFormat, toolChoice, and tools.
    */
+  // deprecated: mode will be removed in v2.
+  // All necessary settings will be directly supported through the call settings,
+  // in particular responseFormat, toolChoice, and tools.
```
lgrammel, line 27: removing this from the message that's displayed in the ide for now as it was confusing people and there is no v2 yet

content/docs/07-reference/ai-sdk-core/03-generate-object.mdx
```diff
@@ -54,6 +54,25 @@ const { object } = await generateObject({
 });
 ```
 
+#### Example: generate an enum
+
+When you want to generate a specific enum value, you can set the output strategy to `enum`
+and provide the list of possible values in the `enum` parameter.
```
nicoalbanese, line 60: same here

nicoalbanese (approved):

Issue #2948, opened 2024-09-09
Receiving the error "Error: Cannot read properties of undefined (reading 'workers')" after upgrading to v3.3.30 using Next.js v14.2.9


WoetDev:
### Description

After upgrading to v3.3.30, I receive the following error:
![image](https://github.com/user-attachments/assets/777c4f03-df2b-4ca2-8dfb-a9f800a94e78)

### Code example

You can find the reproduction of the bug in this repository: https://github.com/WoetDev/bug-next-ai-sdk

It's a reproduction of the official example for using AI assistants with the Next.js app router, so this example seems to be broken: https://sdk.vercel.ai/examples/next-app/assistants/stream-assistant-responses

You only need to add your OPENAI API key in the .env file:

```
OPENAI_API_KEY=xxxx
```

### Additional context

Noticed this issue when upgrading packages, this is affecting the AI functionality in my production app and blocking me from upgrading.

RickRyan26:
When tying to deploy SvelteKit to Vercel I am getting...

`The requested module '@ai-sdk/provider-utils' does not provide an export named 'createIdGenerator'`

Im using:

```
    "@ai-sdk/openai": "^0.0.54",
    "@ai-sdk/svelte": "^0.0.46",
    "ai": "^3.3.30",
```

lgrammel:
@RickRyan26 please trying updating openai to `0.0.58`.

lgrammel:
@WoetDev can you share a stacktrace & the versions of the ai sdk packages that you are using?

nicoalbanese:
@WoetDev can you try and remove the "use server" directive at the top of your actions file and instead move it to the specific actions you are calling (to line 24)?

https://github.com/WoetDev/bug-next-ai-sdk/blob/15f3604073997cc17d10c0268313f6746b88abe2/src/app/actions.tsx#L24

nsenthilkumar:
Moving the "use server" directive fixed this issue for me. I should have come here sooner.

Can we get some insight on why that was the issue so that we can better understand behind the hood moving forward?

doctorDAlvarez:
Same issue here. Fixed moving the directive too. Any insights on why is this happening?
Thanks!!

WoetDev:
@nicoalbanese thanks a bunch! moving the "use server" directive and adding a default state seemed to do the trick on my production app. 

I'm all up-to-date with the latest versions now without an issue:

```
"next": "^14.2.11",
"ai": "^3.3.35",
[...]

PR #2956, opened 2024-09-10
fix (docs): update createAI example, move AI to separate module

huozhi:
### What & Why

Server action module (the module conatining "use server" on the top level) should only contain server actions but the `AI` component itself is not. So that's why I want to move it outside of `actions.tsx` and leave it as a file or a module only containing actions with top-level directive convenience.

Related issue: #2948 
Related Next.js issue comment: https://github.com/vercel/next.js/issues/69756#issuecomment-2339033901


nicoalbanese:
thanks for this @huozhi! I've repro'd and it seems possible to solve by moving the 'use server' directive to the specific function rather than at the top level of the file.

lgrammel (approved):

PR #2958, opened 2024-09-10
fix (docs): update usage of new lines to match json5 spec

jeremyphilemon:
we use [JSON5](https://json5.org/) to render the reference parts in the markdown content, so updated the formatting of new lines – this should also fix search indexing!


nicoalbanese (approved):

lgrammel:
I don't understand this fix - and it changes the whitespace fwiw. Does this need to be applied everywhere? why/how does this work?

lgrammel (approved):

PR #2976, opened 2024-09-12
feat (react): support sending attachments using append

jeremyphilemon:
- [x] process attachments when `append()` is used
- [x] update an example to use append
- [x] add tests

### DIFF
--- a/examples/next-openai/app/use-chat-attachments/page.tsx
+++ b/examples/next-openai/app/use-chat-attachments/page.tsx
@@ -6,10 +6,9 @@ import { useChat } from 'ai/react';
 import { useRef, useState } from 'react';
 
 export default function Page() {
-  const { messages, input, handleSubmit, handleInputChange, isLoading } =
-    useChat({
-      api: '/api/chat',
-    });
+  const { messages, input, setInput, append, isLoading } = useChat({
+    api: '/api/chat',
+  });
 
   const [files, setFiles] = useState<FileList | undefined>(undefined);
   const fileInputRef = useRef<HTMLInputElement>(null);
@@ -47,10 +46,17 @@ export default function Page() {
 
       <form
         onSubmit={event => {
-          handleSubmit(event, {
-            experimental_attachments: files,
-          });
+          event.preventDefault();
+
+          append(
+            { role: 'user', content: input },
+            {
+              experimental_attachments: files,
+            },
+          );
+
           setFiles(undefined);
+          setInput('');
 
           if (fileInputRef.current) {
             fileInputRef.current.value = '';
@@ -103,7 +109,9 @@ export default function Page() {
         <input
           value={input}
           placeholder="Send message..."
-          onChange={handleInputChange}
+          onChange={event => {
+            setInput(event.target.value);
+          }}
           className="bg-zinc-100 w-full p-2"
           disabled={isLoading}
         />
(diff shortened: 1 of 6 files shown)


examples/next-openai/app/use-chat-attachments/page.tsx
```diff
@@ -6,10 +6,9 @@ import { useChat } from 'ai/react';
 import { useRef, useState } from 'react';
 
 export default function Page() {
-  const { messages, input, handleSubmit, handleInputChange, isLoading } =
-    useChat({
-      api: '/api/chat',
-    });
+  const { messages, input, setInput, append, isLoading } = useChat({
```
lgrammel, line 9: should we update this example or leave it as is (and add a 2nd example if needed)?

packages/react/src/use-chat.ts
```diff
@@ -714,3 +694,46 @@ function countTrailingAssistantMessages(messages: Message[]) {
...
+          contentType,
+          url,
+        });
+      }
+    } else {
+      throw new Error('Invalid attachments type');
+    }
+  }
+
+  return draftAttachmentsForRequest;
```
lgrammel, line 738: now that it's a separate method, i would recommend using the guard pattern to simplify the structure.

https://deviq.com/design-patterns/guard-clause

i.e.
 ```ts
if (attachmentsFromOptions == null) {
  return []
}
```
as the first statement to reduce the level of indentation.

You can then also use `map` and directly return the result, eliminating the `draftAttachmentsFromRequest` variable.

Comment (lgrammel) [reply packages/react/src/use-chat.ts:738]: e.g.

```ts
async function prepareAttachmentsForRequest(
  attachmentsFromOptions: FileList | Array<Attachment> | undefined
) {
  if (!attachmentsFromOptions == null) {
    return []
  }

  if (attachmentsFromOptions) {
    const draftAttachmentsForRequest: Array<Attachment> = []
    if (attachmentsFromOptions instanceof FileList) {
      for (const attachment of Array.from(attachmentsFromOptions)) {
        const { name, type } = attachment

        const dataUrl = await new Promise<string>((resolve, reject) => {
          const reader = new FileReader()
          reader.onload = readerEvent => {
            resolve(readerEvent.target?.result as string)
          }
          reader.onerror = error => reject(error)
          reader.readAsDataURL(attachment)
        })

        draftAttachmentsForRequest.push({
          name,
          contentType: type,
          url: dataUrl
        })
      }
    }
    return draftAttachmentsForRequest
  }

  if (Array.isArray(attachmentsFromOptions)) {
    return attachmentsFromOptions
  }

  throw new Error('Invalid attachments type')
}
```

lgrammel: (the above reduces the cognitive complexity and makes it clear that the FileList handling is in fact the trickier part)

lgrammel (approved):

examples/next-openai/app/use-chat-attachments/page.tsx
jeremyphilemon: makes sense to add a 2nd example in case this was linked from somewhere!

packages/react/src/use-chat.ts
jeremyphilemon: oh til! I definitely like the clean up – will update and keep this pattern in mind moving forward!

PR #3001, opened 2024-09-13
feat (ai/core): add steps property to streamText result and onFinish callback

lgrammel:
## Tasks

- [x] implementation
   - [x] return promise
   - [x] include in onFinish
- [x] docs
   - [x] reference
   - [x] guide
- [x] changeset


content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -124,11 +122,13 @@ const result = await streamText({
   prompt: 'What is the weather in San Francisco?',
 });
 
-for await (const textPart of result.textStream) {
+for await (const textPart of textStream) {
   process.stdout.write(textPart);
 }
 ```
 
+Similar to `generateText`, you can access intermediate tool calls and results using the `steps` property in the result object.
```
nicoalbanese, line 130: smol nit (could be overthinking) wonder if we need to say result object given you are destructuring so there is no "result"

examples/ai-core/src/stream-text/openai-on-finish-steps.ts
```diff
@@ -0,0 +1,29 @@
...
+      }),
+    },
+    maxSteps: 5,
+    onFinish({ steps }) {
+      console.log(JSON.stringify(steps, null, 2));
+    },
+    prompt: 'Invent a new holiday and describe its traditions.',
+  });
+
+  for await (const textPart of result.textStream) {
```
nicoalbanese, line 25: intentionally blank?

lgrammel: yes

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
lgrammel: oh yeah hmm. i guess we can leave it for now

nicoalbanese (approved):

PR #3032, opened 2024-09-16
Fix for attachments in Svelte useChat

tyfiero:
Made a fix for attachments to work with Svelte. Took me longer than I'd like to admit to find out how to fix this. Works well for us

### DIFF
--- a/packages/svelte/src/use-chat.ts
+++ b/packages/svelte/src/use-chat.ts
@@ -134,6 +134,7 @@ const getStreamedResponse = async (
           tool_calls,
           tool_call_id,
           toolInvocations,
+          experimental_attachments
         }) => ({
           role,
           content,
@@ -145,6 +146,7 @@ const getStreamedResponse = async (
           tool_call_id,
           ...(function_call !== undefined && { function_call }),
           ...(tool_calls !== undefined && { tool_calls }),
+          ...experimental_attachments !== void 0 && {experimental_attachments}
         }),
       );
 


cs50victor:
🚀. this was a very tricky bug to figure out.

lgrammel:
cc @jeremyphilemon - was this already implemented or would the solution be more involved (with file lists etc)? I don't want Svelte to have deviate from what we implemented in React

tyfiero:
> cc @jeremyphilemon - was this already implemented or would the solution be more involved (with file lists etc)? I don't want Svelte to have deviate from what we implemented in React

Well this does work with multiple files in experimental_attachments we've tested it, its essentially the same code from the react package, I just added the two lines that were missing in svelte.

lgrammel:
@tyfiero the changes for attachment support in React were more complex:
- https://github.com/vercel/ai/pull/2226
- https://github.com/vercel/ai/pull/2732
- https://github.com/vercel/ai/pull/2976

is all of this implemented for Svelte?

tyfiero:
We opted to port our app over to react from svelte. It's clear the svelte package doesn't have, and will never have, the same support that the react package will. I won't have the time to implement those other attachment features in svelte. This PR can be closed.

MentalGear:
I think feature parity should be an overall goal, no matter the front end library - especially for **svelte** that is officially supported by vercel.

jstjoe:
Hey, are there any known plans to implement this functionality @lgrammel ? I'm also hoping to see this feature come to Svelte and just want to plan accordingly. Thank you!

PR #3081, opened 2024-09-20
fix (ui/vue): keep intermediate assistant messages

lgrammel:


packages/vue/src/use-chat.ts
```diff
@@ -225,12 +225,7 @@ export function useChat(
               mutate([...chatRequest.messages, ...merged]);
               streamData.value = [...existingData, ...(data ?? [])];
             },
-            onFinish(message, options) {
-              // workaround: sometimes the last chunk is not shown in the UI.
-              // push it twice to make sure it's displayed.
-              mutate([...chatRequest.messages, message]);
```
lgrammel, line 231: this was originally added to make the tests work

PR #3110, opened 2024-09-24
feat: support generate id from server side

### LINKED ISSUE #3061: generate assistant message id from backend side
### Feature Description

We have a chat backend with persistence layer, so everytime user refresh page will keep track of the history messages. So it's  necessary that keep message id constant everytime user popup a new message.

But I found that in StreamResponse protocol there is only possible to return text and data. But there's no `id`. and the reason of `generateId` is not working because it means id is generated from frontend, but for the ai message, it's actually generated from backend. So it won't work.

### Use Case

I have some features that will allow user to custimize ai chat message per different browser session. Like double click edit, mark, quote, and index sources. So it's important to keep id constant with backend.

### Additional context

my current workaround

```ts
  onFinish: useCallback(
      (message: Message) => {
        // handle the AI message generated by the backend, because id is not returned in the response,
        //  we need to fetch the latest session to get the id of the last message
        getChatSessionFromBackend({
          sessionId,
        })
          .then((newSession) => {
            const lastAIMessage = newSession.chat_messages
              ?.sort((a, b) => (a.index > b.index ? 1 : -1))
              .at(-1);
            if (lastAIMessage?.role !== "assistant") {
              throw unexpected()
              return;
            } else {
              const id = lastAIMessage.id;
              storeMessage({ ...message, id }) // store message to indexeddb
              });
            }
          })
          .then(async () => {
            await refetch();
          })
      },
      [pipelineId, refetch],
    ),
```

himself65:
Fixes: https://github.com/vercel/ai/issues/3061

### DIFF
--- a/packages/ui-utils/src/process-data-protocol-response.ts
+++ b/packages/ui-utils/src/process-data-protocol-response.ts
@@ -134,17 +134,31 @@ export async function processDataProtocolResponse({
 
     if (type === 'text') {
       if (prefixMap['text']) {
+        if (typeof value !== 'string') {
+          throw new TypeError(
+            'Expected text to be a string except for the first chunk',
+          );
+        }
         prefixMap['text'] = {
           ...prefixMap['text'],
           content: (prefixMap['text'].content || '') + value,
         };
       } else {
-        prefixMap['text'] = {
-          id: generateId(),
-          role: 'assistant',
-          content: value,
-          createdAt,
-        };
+        if (typeof value === 'string') {
+          prefixMap['text'] = {
+            id: generateId(),
+            role: 'assistant',
+            content: value,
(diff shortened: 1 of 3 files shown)


himself65 (commented):
Please have a look, if overall looks good I will continue on adding tests&example

/cc @lgrammel

lgrammel:
Hi thanks for taking a look at this. It would need to support multiple messages in the same stream potentially, e.g. for multi-step tool calls. I was considering either adding it to the `finishStepPart` or adding a `startStepPart` - but it also needs some thinking regarding the mapping of steps to messages, and what objects would need ids and when.

How would you send the ids from streamText?

lgrammel:
Superseeded by #4415

PR #3123, opened 2024-09-26
fix (ai/core): correct whitespace in streamText continueSteps

lgrammel:
## Summary
Buffers and if necessary, drops, text to prevent whitespace issues.

## Background
Language model responses may end with whitespace or mid-word. However, the continuation response typically starts with a non-whitespace token. It is impossible to tell what whitespace (if any) needs to be injected after the last generated token. To work around this issue, we remove trailing characters until we hit the last whitespace section.


packages/ai/core/generate-text/__snapshots__/stream-text.test.ts.snap
```diff
@@ -983,16 +904,18 @@ exports[`options.maxSteps > 3 steps: initial, continue, continue > should record
...
       "ai.response.id": "id-1",
       "ai.response.model": "mock-model-id",
       "ai.response.msToFinish": 0,
       "ai.response.msToFirstChunk": 0,
-      "ai.response.text": "part-3",
+      "ai.response.text": "final value keep all whitespace
+ ",
       "ai.response.timestamp": "1970-01-01T00:00:01.000Z",
-      "ai.result.text": "part-3",
+      "ai.result.text": "final value keep all whitespace
```
lgrammel, line 917: bug?

packages/ai/core/generate-text/stream-text.ts
```diff
@@ -723,6 +760,44 @@ class DefaultStreamTextResult<TOOLS extends Record<string, CoreTool>>
...
+                }
+              }
+
+              // when using continuation, publish buffer on final step or if there
+              // was no whitespace in the step:
+              if (
+                continueSteps &&
+                chunkBuffer.length > 0 &&
+                ((stepType === 'continue' && nextStepType === 'done') ||
+                  (stepType === 'continue' && !chunkTextPublished))
```
lgrammel, line 789: bug: what if continueSteps is true, there is remaining buffer, but not continue step (just initial)

jeremyphilemon (approved):

PR #3130, opened 2024-09-27
fix (provider/bedrock): don't use env var for session token default.

### LINKED ISSUE #3018: AWS Bedrock provider doesn't work unless using `bedrockOptions`
### Description

This works:

```
createAmazonBedrock({
          bedrockOptions: {
            region: env.BEDROCK_REGION,
            credentials: {
              accessKeyId: env.BEDROCK_ACCESS_KEY,
              secretAccessKey: env.BEDROCK_SECRET_KEY,
            },
          },
        })(model)
```
        
This works when running locally, but not when deploying to Vercel:

```
createAmazonBedrock({
            region: env.BEDROCK_REGION,
              accessKeyId: env.BEDROCK_ACCESS_KEY,
              secretAccessKey: env.BEDROCK_SECRET_KEY,
          },
        })(model)
```

I imagine the issue is that Vercel is setting a key that causes the error (https://vercel.com/docs/projects/environment-variables/reserved-environment-variables#allowed-environment-variables). But I didn't manage to track down the error. `bedrockOptions` worked for me though.

### Code example

_No response_

### Additional context

_No response_

shaper:
Deployments using the Bedrock provider can run in a serverless function that pre-declares a value for `AWS_SESSION_TOKEN` representing an alternate set of credentials than those intended and specified by the developer. If we silently use the environment variable for a default it can create a non-obvious credentials failure that surprises the developer.

The session token credentials are less commonly used, and the developer can still pass them along manually (and reference the environment variable themselves if they choose to do so).

We also update the documentation to state this behavior as it's admittedly deviant from common practice in the AI SDK providers (but for good reason due to the idiosyncratic way it can be already-present in the serverless environment).

Fixes #3018.

### DIFF
--- a/content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
+++ b/content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
@@ -121,8 +121,11 @@ You can use the following optional settings to customize the Amazon Bedrock prov
 
 - **sessionToken** _string_
 
-  Optional. The AWS session token that you want to use for the API calls.
-  It uses the `AWS_SESSION_TOKEN` environment variable by default.
+  Optional. The AWS session token that you want to use for the API calls. Note
+  this does **not** use the `AWS_SESSION_TOKEN` environment variable by default.
+  That variable is sometimes pre-populated with alternate credentials, e.g. when
+  running within an AWS serverless function, and so default use risks
+  conflicting in a non-obvious manner with any other credential settings.
 
 - **bedrockOptions** _object_
 
@@ -135,7 +138,7 @@ You can use the following optional settings to customize the Amazon Bedrock prov
   - **credentials** _object_
     The AWS credentials that you want to use for the API calls.
 
-  When `bedrockOptions` are provided, the `region`, `accessKeyId`, and `secretAccessKey` settings are ignored.
+  When `bedrockOptions` are provided, the `region`, `accessKeyId`, `secretAccessKey`, and `sessionToken` settings are ignored.
 
 ## Language Models
 
--- a/packages/amazon-bedrock/src/bedrock-provider.ts
+++ b/packages/amazon-bedrock/src/bedrock-provider.ts
@@ -3,11 +3,7 @@ import {
   NoSuchModelError,
   ProviderV1,
 } from '@ai-sdk/provider';
-import {
-  generateId,
-  loadOptionalSetting,
-  loadSetting,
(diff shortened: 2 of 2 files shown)


shaper:
Notes:
- @lgrammel how would you typically test this in a deployment? I'm not sure of the best way to build and include local modifications to the SDK packages. I played around with `pnpm pack` but you must have some standard relatively straightforward approach I imagine. I was testing manually with a trivial `examples` I added, but of course it pulls the published SDK on deploy. Just asking for my knowledge for the future (does not block PR per se)
- `loadOptionalSetting` is dead code after this PR but feels useful enough, likely to be used in the future, to leave around (but happy to remove it if you prefer)
- we could as discussed in the issue alternately prefix with `AI_SDK_` and/or remove the other defaults as well, but this PR seemed most reasonable on some reflection (as noted in docs/comments)
- there are not currently unit tests for the involved provider code, I'd be happy to add some to validate the env var use (and non-use) if it feels worthwhile but it's a pretty small change

shaper:
It occurred to me to check the other `AWS_XXX` default env vars in a clean deployment and I found that they're actually set as well (`AWS_ACCESS_KEY_ID`, `AWS_SESSION_TOKEN`) so I am now inclined to remove these from use as defaults as well and update the documentation to explain. At least this would make the bedrock provider consistent in not using the AWS env var defaults.

Rationale is to minimize the risk of developer surprise due to unintentional creds conflict.

I realize that this is effectively known and documented behavior: https://vercel.com/guides/how-can-i-use-aws-sdk-environment-variables-on-vercel however my understanding is that devs are likely to set up separate creds and aim to use those, and doing so while omitting some values risks defaulting to these against developer intent.

Let me know your thought on removing the defaults and if you agree I'll update the PR to reflect.

lgrammel:
My main concern (and that's why I've been holding off here) is that this will be a breaking change that renders some working setups non-functional.

I see two options:
- we add documentation about the current behavior and workaround (short-term preferred)
- we make breaking change and do a major version bump. In this case we want to get the overall design right - which may mean no environment variables and also renaming/adjust how we use `bedrockOptions`. The goal of environment variables was making the local development easy to get started with, but it seems to be too confusing for deployment.

Re smaller comments:
- re testing: ideally we'd test it but it may not be possible in all cases
- re dead code: agree to leave it in, at least for now
- re prefixing: would like to discuss separately
- re unit tests: don't feel strong about unit tests here, the main risk is having a design that minimized setup issues in AWS environments, not bugs in the code

PR #3163, opened 2024-09-30
Include the created at field in assistant response and use assistant #3129

### LINKED ISSUE #3129: Include the `createdAt` field in `AssistantResponse` and `useAssistant`
### Feature Description

For OpenAI, on the stream event `thread.message.created`, the response is a [`message` object](https://platform.openai.com/docs/api-reference/messages/object), which includes a `created_at` field.

This event is received by the `AssistantResponse`'s `ReadableStream`, but [the object it sends down stream](https://github.com/vercel/ai/blob/main/packages/ai/streams/assistant-response.ts#L90-L101) only contains `id`, `role`, and `content`. Further, only `id` is set from the value received from the event, so if a new role is introduced, or the structure of the content changes, this behavior will break.

At minimum, it makes sense to include the `created_at` field, if not all fields, received by the stream event and pass them down stream.

## Minimum solution: Add `createdAt`

Update [`AssistantResponse` lines 90-101](https://github.com/vercel/ai/blob/main/packages/ai/streams/assistant-response.ts#L90-L101) to
```ts
case 'thread.message.created': {
  controller.enqueue(
    textEncoder.encode(
      formatStreamPart('assistant_message', {
        id: value.data.id,
        createdAt: value.data.created_at, // <--- add field
        role: 'assistant',
        content: [{ type: 'text', text: { value: '' } }],
      }),
    ),
  );
  break;
}
```

Update [`useAssistant` lines 183-210](https://github.com/vercel/ai/blob/main/packages/react/src/use-assistant.ts#L183-L210) to
```ts
case 'assistant_message': {
  setMessages(messages => [
    ...messages,
    {
      id: value.id,
      createdAt: value.createdAt, // <--- add field
      role: value.role,
      content: value.content[0].text.value,
    },
  ]);
  break;
}

case 'text': {
  // text delta - add to last message:
  setMessages(messages => {
    const lastMessage = messages[messages.length - 1];
    return [
      ...messages.slice(0, messages.length - 1),
      {
        id: lastMessage.id,
        createdAt: value.createdAt, // <--- add field
        role: lastMessage.role,
        content: lastMessage.content + value,
      },
    ];
  });
  break;
}
```

### Use Case

I want to show the created date and time for all messages in the history. This will primarily be helpful if a user returns to a previous thread. Adding this field will make it possible to display the dates without having to separately request them.

### Additional context

The only reason I can think of not to include this value is for compatibility with non-OpenAI APIs. I don't know this package well enough to determine if that is an issue or not.

Ideally, this would include all possible values and if there are any incompatibilities with other APIs, they will be accounted for by using union or conditional types.

subin-chella:
fixes  #3129
Tests:
1. use-assistant.ui.test.tsx
image
2. example Stream-assistance-response
image

image
3. example next-openai
image
image

### DIFF
--- a/packages/ai/core/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/core/prompt/convert-to-language-model-prompt.ts
@@ -131,10 +131,6 @@ export function convertToLanguageModelMessage(
 
                 // try to convert string image parts to urls
                 if (typeof part.image === 'string') {
-                  const invalidUrlPattern = /\s/;
-                  if (invalidUrlPattern.test(part.image)) {
-                    throw new Error(`Invalid Image URL: ${part.image}`);
-                  }
                   try {
                     const url = new URL(part.image);
 
@@ -150,8 +146,6 @@ export function convertToLanguageModelMessage(
                               part.experimental_providerMetadata,
                           };
                         } else {
-                          const downloadedImage =
-                            downloadedImages[url.toString()];
                           return {
                             type: 'image',
                             image: downloadedImage.data,
--- a/packages/ai/core/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/core/prompt/convert-to-language-model-prompt.ts
@@ -146,6 +146,7 @@ export function convertToLanguageModelMessage(
                               part.experimental_providerMetadata,
                           };
                         } else {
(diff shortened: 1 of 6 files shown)


lgrammel:

PR #3186, opened 2024-10-03
chore (provider/azure): update Azure OpenAI API version to 2024-08-01-preview

h4r5h4:
`api-version` param is currently hardcoded to `2024-06-01`, it prevents using features like `toolChoice` since it's only enabled for `2024-07-01-preview` and later

### DIFF
--- a/.changeset/ninety-buses-press.md
+++ b/.changeset/ninety-buses-press.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/azure': patch
+---
+
+chore (provider/azure): update Azure OpenAI API version to 2024-08-01-preview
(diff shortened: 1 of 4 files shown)


lgrammel:
We have to hardcode the version to ensure compatibility. That being said, it should be the latest available version. Instead of making it configurable, can you just update the version to the latest available one?

h4r5h4:
got it, I have updated it to the latest version.

lgrammel:
Perfect can you add a changeset (patch) for the azure provider (run `pnpm changeset` in project root): `feat (provider/azure): update Azure OpenAI API version to 2024-08-01-preview`

lgrammel:
Also please run `pnpm prettier-fix` in the root.

h4r5h4:
> Also please run `pnpm prettier-fix` in the root.

it should be passing now

lgrammel (approved):

PR #3206, opened 2024-10-05
feat: add timeout option to generateText and generateObject

### LINKED ISSUE #3169: Add timeout option to generateText and generateObject
### Feature Description

A top-level `timeout` argument for LLM calls that cancels in-progress requests when the timeout has elapsed.

### Use Case

Today, there's an optional `abortController` argument for most LLM calls. Having a top-level `timeout` argument, like the top-level `maxRetries` argument, will make it easier to use the ai-sdk in scenarios where we want a fast response or none at all (like realtime classification). Configuring and managing abort signals in node is kind of annoying, which is why I think having a top-level field would be nice.

### Additional context

_No response_

subin-chella:
Add top-level timeout parameter to LLM calls for automatic request cancellation. This simplifies handling time-sensitive operations without manually managing AbortController.

- Add timeout parameter alongside existing abortsignal option
- Implement internal AbortController logic to handle timeouts
- Add timeout parameter to both generateText and generateObject methods

Closes #3169


lgrammel:
thank you. instead of unit testing generateText etc, can you rename to createAbortSignalWithTimeout and extensively unit test that function? it should take an abort signal (potentially undefined) and a timeout (potentially undefined) as a parameter object arg.

packages/provider-utils/src/get-effective-abort-signal.ts
```diff
@@ -0,0 +1,26 @@
+export function getEffectiveAbortSignal(
+  abortSignal: AbortSignal | undefined,
+  timeout: number | undefined
+): { signal: AbortSignal | undefined; clearTimeout: () => void } {
+  let effectiveAbortSignal = abortSignal;
+  let timeoutId: NodeJS.Timeout | null = null;
```
lgrammel, line 6: will this create node dependency that leads to issues in browsers

content/docs/07-reference/ai-sdk-core/03-generate-object.mdx
```diff
@@ -396,6 +396,13 @@ To see `generateObject` in action, check out the [additional examples](#more-exa
       description:
         'Maximum number of retries. Set to 0 to disable retries. Default: 2.',
     },
+    {
+      name: 'timeout',
+      type: 'number',
+      isOptional: true,
+      description:
+        'Timeout in milliseconds. If set, the call will be aborted after the timeout. Defaults to 0 (no timeout).',
+    },
```
lgrammel, line 405: Default to no timeout. Lets not spec the 0 behavior

packages/ai/core/generate-object/generate-object.test.ts
```diff
@@ -6,6 +6,7 @@ import { setTestTracer } from '../telemetry/get-tracer';
 import { MockLanguageModelV1 } from '../test/mock-language-model-v1';
 import { MockTracer } from '../test/mock-tracer';
 import { generateObject } from './generate-object';
+import { getEffectiveAbortSignal } from '../../../provider-utils/src/get-effective-abort-signal';
```
lgrammel, line 9: please move into the ai package under util

packages/ai/core/generate-text/generate-text.ts
```diff
@@ -274,15 +275,17 @@ functionality that can be fully encapsulated in the provider.
...
               const result = await model.doGenerate({
                 mode,
                 ...callSettings,
                 inputFormat: currentInputFormat,
                 prompt: promptMessages,
                 providerMetadata,
-                abortSignal,
+                abortSignal: effectiveAbortSignal, 
                 headers,
               });
```
lgrammel, line 287: needs to happen in a finally block in case of errors

packages/provider-utils/src/index.ts
```diff
@@ -15,3 +15,4 @@ export * from './uint8-utils';
 export * from './validate-types';
 export * from './validator';
 export * from './without-trailing-slash';
+export * from './get-effective-abort-signal';
```
lgrammel, line 18: please move to main ai package

packages/provider-utils/src/get-effective-abort-signal.ts
subin-chella: fixed

packages/ai/core/generate-text/generate-text.ts
subin-chella: fixed

packages/provider-utils/src/index.ts
subin-chella: done

packages/ai/core/generate-object/generate-object.test.ts
subin-chella: fixed

subin-chella:
> thank you. instead of unit testing generateText etc, can you rename to createAbortSignalWithTimeout and extensively unit test that function? it should take an abort signal (potentially undefined) and a timeout (potentially undefined) as a parameter object arg.

@lgrammel Addressed this comment also resolved all code review comments

examples/ai-core/src/stream-text/openai-generatetext-timeout.ts
```diff
@@ -0,0 +1,23 @@
+import { openai } from '@ai-sdk/openai';
+import { generateText } from 'ai';
+import 'dotenv/config';
+
+async function main() {
+  try {
+    const result = await generateText({
+      model: openai('gpt-3.5-turbo'),
+      prompt: 'Write a short story about a robot learning to love:\n\n',
+      timeout: 100, // Change this value to see different results
```
lgrammel, line 10: can you use a custom fetch function with a delay here so timeouts can be simulated easily? ( see other examples for custom fetch )

packages/ai/util/get-effective-abort-signal.ts
```diff
@@ -0,0 +1,26 @@
+export function getEffectiveAbortSignal(
```
lgrammel, line 1: is this file still needed?

packages/ai/core/generate-text/generate-text.test.ts
```diff
@@ -783,3 +783,109 @@ describe('tools with custom schema', () => {
     ]);
   });
 });
+
+
+
+describe('result.text.timeout', () => {
```
lgrammel, line 789: can the abort signal function be injected / stubbed to make the tests simpler and deterministic? there should not be any date now or delay ideally. eg generate text already has similar test functions like date now that are passed in

packages/ai/core/generate-object/generate-object.test.ts
```diff
@@ -936,3 +936,109 @@ describe('telemetry', () => {
     expect(tracer.jsonSpans).toMatchSnapshot();
   });
 });
+
+describe('generateObject.timeout', () => {
```
lgrammel, line 940: same as below regarding deterministic behavior and mocking

packages/ai/util/get-effective-abort-signal.ts
subin-chella: My bad. Didnt push the change. Its done now.

Yidadaa:
Great job! By the way, would it be possible for you to include a timeout option in streamText as well?

lgrammel:
I've refactored, fixed and investigated this further: https://github.com/vercel/ai/pull/3239

However, I've come to the conclusion that using the `AbortSignal.timeout` https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static is the best solution which is available in Node 17+ and all modern browsers.

It means that there is no code to maintain and the AI SDK is standard compliant. Closing.

Issue #3225, opened 2024-10-10
Support `onChunk` for `useChat`


BrianHung:
### Feature Description

Support `onChunk` for `useChat`. This should be compatible with `convertArrayToReadableStream`, such that chunks can be recorded and replayed for development or mocking purposes.

### Use Case

I'm using client-side tools with `useChat`, and would want to write some unit tests to make sure that components correctly handle tool-calling. To that end, I need a way to "replay" an LLM response. `convertArrayToReadableStream` is a helpful utility in mocking that if you already have an existing response or if manually crafting it. 

If you don't, then you need an ability to "record" the LLM response via chunks. You can do this in `streamText` but it's not convenient: with maxRoundtrips, a client can call the API endpoint multiple times; if it's edge middleware, to persist across sessions, you need a KV database which is overkill in my opinion.

Instead, we can implement `onChunk` for `useChat` and persist chunks there with session.

### Additional context

Prior discussion on this https://github.com/vercel/ai/discussions/1644

gr2m:
> would want to write some unit tests to make sure that components correctly handle tool-calling

We now have [the includeRawChunks option](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#include-raw-chunks) which you might be able to capture and use as fixtures?

Here is an example:
https://github.com/vercel/ai/blob/067d2a4c09b4981c1c62384d43b14786cb3e0fea/examples/ai-core/src/stream-text/raw-chunks.ts#L19-L27

aayush-kapoor:
Marking this as closed for now - if you face any regression or have a similar problem after migrating to the latest AI SDK version, please feel free to open a new issue (with possible reproduction)

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This is a request to expand `useChat` with new chunk-handling functionality for testing and mocking. No existing promised behavior is reported as broken.

### Evidence
- The issue explicitly requests adding an `onChunk` callback to `useChat` for recording and replaying streamed responses.
- The requested callback is a new client API capability, not a failure of documented existing behavior.
- A later comment suggests `includeRawChunks` as a possible alternative, and the issue was closed without identifying a regression.

PR #3252, opened 2024-10-14
useChat chunk

BrianHung:
#3225 

Let's clients access chunks that are proceeded from the data stream via `useChat` `onChunk`. Useful for cache-ing responses state-fully on the client and then replaying them.

```ts
// parrot.ts
  const { chunks } = await req.json();
  if (chunks) {
    const parts = chunks.map(({ type, value }) => formatStreamPart(type, value));
    const stream = convertArrayToReadableStream(parts)
      .pipeThrough(new TextEncoderStream())
    const response = new Response(stream, {
      headers: {
        contentType: 'text/plain; charset=utf-8',
        dataStreamVersion: 'v1',
      }
    });
    return response;
  }
  // If no chunks, call normal API
}
```


lgrammel:
My thoughts:
a) not sure if we want to expose all chunk types like this
b) would need to be implemented in svelte, vue, solid (but good to hold off until a) is clear)

BrianHung:
> not sure if we want to expose all chunk types like this

What’s an alternative if we want to store chunks themselves? Is there a higher level API that’ll be more immune to changes if stream parts were changed?

lgrammel:
@BrianHung the stream parts are standardized in the stream data protocol (but additions might happen). you could also record using a fetch interceptor for now.

UrielCh:
> ```ts
> import { readDataStream } from 'ai';
> ```

'readDataStream' is declared but its value is never read.ts(6133)

I'm trying to get all chunk added by using `toolCallStreaming: true` in streamText(), that look to be a good start to get those data, maybe that is a better official way to do that.

sportnak:
In v5 `onResponse` is no longer provided. I have found a potential workaround:

```ts
export class TestTransport extends DefaultChatTransport<UIMessage> {
  protected processResponseStream(stream: any) {
    return parseJsonEventStream({
      stream,
      schema: uiMessageChunkSchema,
    }).pipeThrough(
      new TransformStream<ParseResult<UIMessageChunk>, UIMessageChunk>({
        transform(chunk, controller) {
          console.log('FOO', chunk);
          if (!chunk.success) {
            throw chunk.error;
          }
          controller.enqueue(chunk.value);
        },
      }),
    );
  }
}

```

However, you have to copy the zod schemas out of the library a simpler PR might be to make those exported. But you can log out each chunk and view it's contents to act on it.

sportnak:
I should correct myself, you don't need to re-implement you can just process the stream yourself using `pipeThrough`

```ts
export class TestTransport extends DefaultChatTransport<UIMessage> {
  protected processResponseStream(stream: any) {
    return super.processResponseStream(stream).pipeThrough(
      new TransformStream<UIMessageChunk>({
        transform(chunk, controller) {
          console.log('foo', chunk);
          controller.enqueue(chunk);
        },
      }),
    );
  }
}```

gr2m:
@sportnak I'm not sure how that relates to the PR? 

But maybe your workaround might be a good addition to our migration guide here:
https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#onresponse-removal

aayush-kapoor:
marking this PR as closed since it pertains to an issue that was recently closed!

PR #3335, opened 2024-10-22
fix (examples): Remove `NextRequest.ip` references as it's now deleted.

shaper:
The build is broken for these two examples that specify `latest` for the `next` dependency. Next.js 15 deleted the `NextRequest.ip` field per:

https://nextjs.org/docs/app/api-reference/functions/next-request#version-history

Note that developers might want to also fall back to include `request.headers.get('x-forwarded-for')` but we do not do so in this change.

### DIFF
--- a/examples/next-openai-kasada-bot-protection/kasada/kasada-server.tsx
+++ b/examples/next-openai-kasada-bot-protection/kasada/kasada-server.tsx
@@ -59,7 +59,7 @@ async function getKasadaMetadata(request: NextRequest): Promise<{
   }));
 
   const kasadaPayload: APIRequest = {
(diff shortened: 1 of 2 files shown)


shaper:
> Note that developers might want to also fall back to include `request.headers.get('x-forwarded-for')` but we do not do so in this change.

Let me know if you have any thoughts re: the above, we can of course add it but I don't have full context on the goal/intent of this example, and it's potentially a comma-separated list of addresses so we'd need to parse.

lgrammel:
Ended up removing the examples fully, the bot protection example had no AI SDK code, and the rate limiting example was outdated (AI SDK 2 code). We need to bring them back in some fashion, ideally nicely integrated with AI SDK 3 features.

PR #3352, opened 2024-10-24
chore (package): Specify Node version and document versions.

shaper:
chore (package): Specify Node version and document versions.

Contributing developers can unintentionally install and build with Node or
`pnpm` versions we don't officially support. Currently our CI runs with Node
v18 and v20, and we're using `pnpm` v8.

This change adds an `engines` field to the top level `package.json`
file. This will output a warning on install/build if the engine
version doesn't match the specified range.

The `pnpm` version is already specified in `packageManager` to reflect
the version we use in CI, although it is not enforced or warned if,
for example, the developer is on v9.

This change also updates contributing docs to note the supported Node
engine and `pnpm` versions.

We can consider for the future but choose not to pursue for now:
- placing `engine-strict=true` in the top level `.npmrc` to
  hard-fail on incompatible engine versions
- adding a preinstall/prebuild command that checks versions with
  more explicit failures, messaging, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,9 @@
     "turbo": "^1.10.13",
     "vitest": "1.6.0"
   },
+  "engines": {
+    "node": "^18.0.0 || ^20.0.0"
+  },
   "homepage": "https://sdk.vercel.ai/docs",
   "repository": {
     "type": "git",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,8 +36,9 @@ To set up the repository on your local machine, follow these steps:
 
 1. **Fork the Repository**: Make a copy of the repository to your GitHub account.
 2. **Clone the Repository**: Clone the repository to your local machine, e.g. using `git clone`.
-3. **Install pnpm**: If you haven't already, install `pnpm`. You can do this by running `npm install -g pnpm` if you're using npm. Alternatively, if you're using Homebrew (Mac), you can run `brew install pnpm`.
-4. **Install Dependencies**: Navigate to the project directory and run `pnpm install` to install all necessary dependencies.
+3. **Install Node**: If you haven't already, install Node `v18` or `v20`. We recommend managing Node installation and versions with [fnm](https://github.com/Schniz/fnm) or [nvm](https://github.com/nvm-sh/nvm).
+4. **Install pnpm**: If you haven't already, install `pnpm` `v8`. You can do this by running `npm install -g pnpm@8` if you're using npm. Alternatively, if you're using Homebrew (Mac), you can run `brew install pnpm@8`. For more see [the pnpm site](https://pnpm.io/installation).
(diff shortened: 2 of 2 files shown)


packages/ai/README.md
```diff
@@ -8,10 +8,15 @@ To learn more about how to use the AI SDK, check out our [API Reference](https:/
 
 ## Installation
 
-You will need Node.js 18+ and pnpm installed on your local development machine.
```
lgrammel, line 11: This README is for using the AI SDK and should not be changed. We have a separate contribution guide ( CONTRIBUTING.md in root ) that should contain this information.

packages/ai/README.md
```diff
@@ -8,10 +8,15 @@ To learn more about how to use the AI SDK, check out our [API Reference](https:/
...
+Install on your local development machine:
+
+- Node.js `v18` or `v20`. We recommend managing Node installation and versions with [fnm](https://github.com/Schniz/fnm) or [nvm](https://github.com/nvm-sh/nvm).
+- For package management we recommend (and the subsequent docs assume) [pnpm](https://pnpm.io/installation) `v8`, e.g. installed via `npm install -g pnpm@8.6.9`.
+
+Once you've got those in place:
 
 ```shell
-npm install ai
```
lgrammel, line 14: NPM is fine for using the AI SDK ( as is pnpm, bun, etc) - only NPM here for illustration purposes.

package.json
```diff
@@ -38,6 +38,9 @@
     "turbo": "^1.10.13",
     "vitest": "1.6.0"
   },
+  "engines": {
+    "node": "^18.0.0 || ^20.0.0"
+  },
```
lgrammel, line 43: What effect does the engine spec have on the root level?

shaper: An install or build command with a Node version that isn't `18.x.x` or `20.x.x` will output a warning as below at the start of install/build. I think this is fine, and preferred to (say) hard-failing, though we could do that too to force a maintainer to have a known good version (see the PR description for more).

There should be no effect for those with a compatible Node version.

Note that I believe non-maintainers installing the AI SDK with a Node engine that doesn't fit the spec will also see the warning. Again, this seems ok to me. We don't block/disallow the install, but we let them know their mileage may vary.

```
 WARN  Unsupported engine: wanted: {"node":"^18.0.0 || ^20.0.0"} (current: {"node":"v23.0.0","pnpm":"9.12.2"})
```

shaper: The top level seemed like the right place for the engine spec, but if you think it makes more sense we could do it in `packages/ai/package.json` instead. The `pnpm` version is at the topmost level already.

packages/ai/README.md
shaper: Thanks, I reverted changes to this file and made a smaller set of edits to the contributing doc.

lgrammel (approved):

PR #3354, opened 2024-10-24
feat: Add support for Google Gemini files and Google Storage on Vertex

danielrhodes:
Currently the AI SDK does not support Gemini files https://ai.google.dev/api/files, which means that even though Gemini supports it, you cannot use files larger than a certain size or reuse files from previous requests.

When using Gemini files, the Generative AI SDK expects that the file part will look like the following:

```json
{
   "fileData": {
       "mimeType": "application/pdf",
       "fileUri": "https://generativelanguage.googleapis.com/v1beta/files/00000000-00000000-00000000-00000000"
   }
}
```

However, the current behavior of the SDK is to attempt to download any URL it sees. This of course fails here since the URL is not public. This results in a very cryptic error from the SDK and an unnecessary loss in functionality for the developer since passing through a URL is so trivial.

This PR addresses this shortcoming by:
- Adding generic support for models to opt-out of downloading assets if they support a URL.
- Opting out of asset downloading for certain URLs and supporting URLs as fileData for both Gemini and Vertex
- Simplifying the normalization of message parts after the assets are downloaded to improve readability and support non-downloaded URLs.

In the Vercel AI SDK the message part, using a Gemini file should now look like:
```js
{
   type: "file",
   data: "https://generativelanguage.googleapis.com/v1beta/files/00000000-00000000-00000000-00000000",
   mimeType: "application/pdf"
}
```

What isn't in this PR:
- A generic way to upload files to a provider.

### DIFF
--- a/examples/ai-core/src/generate-object/google-gemini-files.ts
+++ b/examples/ai-core/src/generate-object/google-gemini-files.ts
@@ -1,5 +1,5 @@
 import { GoogleAIFileManager } from '@google/generative-ai/server';
-import { google } from '../../../../packages/google';
+import { google } from '@ai-sdk/google';
 import { generateObject } from 'ai';
 import path from 'path';
 import 'dotenv/config';
--- a/examples/ai-core/src/generate-object/google-gemini-files.ts
+++ b/examples/ai-core/src/generate-object/google-gemini-files.ts
@@ -1,0 +1,47 @@
+import { GoogleAIFileManager } from '@google/generative-ai/server';
+import { google } from '../../../../packages/google';
+import { generateObject } from 'ai';
+import path from 'path';
+import 'dotenv/config';
+import { z } from 'zod';
+
+async function main() {
+  const fileManager = new GoogleAIFileManager(
+    process.env.GOOGLE_GENERATIVE_AI_API_KEY!,
+  );
+
+  const filePath = path.resolve(__dirname, '../../data/ai.pdf');
+
+  const geminiFile = await fileManager.uploadFile(filePath, {
+    name: `ai-${Math.random().toString(36).substring(7)}`,
+    mimeType: 'application/pdf',
+  });
+
+  const { object: summary } = await generateObject({
+    model: google('gemini-1.5-pro-latest'),
+    schema: z.object({
+      title: z.string(),
+      keyPoints: z.array(z.string()),
+    }),
+    messages: [
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'text',
+            text: 'Extract title and key points from the PDF.',
+          },
+          {
+            type: 'file',
+            data: geminiFile.file.uri,
+            mimeType: geminiFile.file.mimeType,
+          },
+        ],
+      },
+    ],
+  });
+
+  console.log(summary);
+}
+
+main().catch(console.error);
--- a/packages/ai/core/prompt/convert-to-language-model-prompt.test.ts
+++ b/packages/ai/core/prompt/convert-to-language-model-prompt.test.ts
@@ -89,7 +89,7 @@ describe('convertToLanguageModelPrompt', () => {
     });
 
     describe('file parts', () => {
-      it('should handle file parts with URL data', async () => {
+      it('should pass through URLs when the model supports a particular URL', async () => {
         const result = await convertToLanguageModelPrompt({
           prompt: {
             type: 'messages',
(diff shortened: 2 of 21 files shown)


packages/ai/core/prompt/convert-to-language-model-prompt.ts
```diff
@@ -388,3 +200,131 @@ async function downloadAssets(
     downloadedImages.map(({ url, data }) => [url.toString(), data]),
   );
 }
+
```
danielrhodes, line 203: Rewriting this was not strictly necessary, but given the complexity of what was there before, this made things much easier to understand.

A future enhancement here might be to normalize the `image` and `file`. But it's probably better to wait to do this until there are more examples of file and image uploading in other SDKs.

packages/provider/src/language-model/v1/language-model-v1.ts
```diff
@@ -66,6 +66,16 @@ Defaults to `false`.
...
 
+  /**
+   * Checks if the model supports the given URL in the context of a language model call where a URL is
+   * passed as a file part. If the model does not support the URL, the AI SDK will download the file
+   * and pass the file data to the model.
+   *
+   * Default to `false`.
+   * @param url
+   */
+  supportsUrl?(url: URL): boolean;
```
lgrammel, line 77: +1

packages/ai/core/prompt/convert-to-language-model-prompt.test.ts
```diff
@@ -89,7 +89,7 @@ describe('convertToLanguageModelPrompt', () => {
     });
 
     describe('file parts', () => {
-      it('should handle file parts with URL data', async () => {
+      it('should pass through URLs when the model supports a particular URL', async () => {
```
lgrammel, line 92: please add tests that cover the added behavior

lgrammel:
This is great! more testing around the message conversion to cover the added functionality would be important, and adding an example under `examples/ai-core` so I can test it out. other than that it looks good.

lgrammel:
It might also affect the middleware, potentially need to have the new method there as well.

examples/ai-core/src/generate-object/google-gemini-files.ts
```diff
@@ -0,0 +1,47 @@
+import { GoogleAIFileManager } from '@google/generative-ai/server';
+import { google } from '../../../../packages/google';
```
lgrammel, line 2: This should come from `@ai-sdk/google`

packages/ai/package.json
```diff
@@ -101,7 +101,7 @@
     "@anthropic-ai/sdk": "0.18.0",
     "@aws-sdk/client-bedrock-runtime": "3.451.0",
     "@edge-runtime/vm": "^3.2.0",
-    "@google/generative-ai": "0.1.1",
+    "@google/generative-ai": "0.21.0",
```
lgrammel, line 104: Please revert - this will affect the legacy google provider.

packages/ai/streams/google-generative-ai-stream.ts
```diff
@@ -30,6 +30,12 @@ interface TextPart {
   inlineData?: never;
 }
 
+interface FileDataPart {
+  text?: never;
+  mimeType: string;
+  fileUri: string;
+}
+
```
lgrammel, line 38: please revert - this is a legacy provider that we are not maintaining and that will be removed soon

packages/google-vertex/src/google-vertex-language-model.ts
```diff
@@ -207,6 +207,14 @@ export class GoogleVertexLanguageModel implements LanguageModelV1 {
     }
   }
 
+  supportsUrl(url: URL): boolean {
+    if (url.protocol === 'gs:') {
+      return true;
+    }
+
+    return false;
```
lgrammel, line 215: return `url.protocol === 'gs:'`

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -179,6 +179,18 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV1 {
...
+  supportsUrl(url: URL): boolean {
+    if (
+      url
+        .toString()
+        .startsWith('https://generativelanguage.googleapis.com/v1beta/files/')
+    ) {
+      return true;
+    }
+
+    return false;
```
lgrammel, line 191: same, just return the condition

danielrhodes:
@lgrammel Resolved your PR comments + added additional tests for the new functionality.

lgrammel:
Thanks I'll look into integration this today or tomorrow

lgrammel:
Continued in #3355

PR #3364, opened 2024-10-25
feat (provider/bedrock): add file upload support

dodgez:
# Summary
This PR adds support for file upload for the Amazon Bedrock provider.
- Add case to convert to Bedrock's document type from the file type
- Add basic check to the Bedrock message conversion test
- Add a file upload example for Bedrock.

_Note: For now the code generates the required document name by hashing the file contents, but it may be possible to default to a provider specific metadata option in the future._

Bedrock only supports file uploads for specific models. I found a page a while back that listed compatibility but cannot find it now. I know Claude 3 Sonnet and Claude 3 Haiku support it but Claude 3.5 Sonnet doesn't.


lgrammel:
Thank you! Continued in #3383

PR #3420, opened 2024-10-30
feat (provider/cohere): Use Cohere v2 API.

### LINKED ISSUE #3331: Support Cohere v2 API
### Feature Description

Cohere recently released [v2 of their API ](https://cohere.com/blog/new-api-v2), which includes support to define tools in JSON schema instead of Python types.

### Use Case

Supporting v2 enables complex tool definitions, which are now limited to primitive types, as mentioned in the docs https://sdk.vercel.ai/providers/ai-sdk-providers/cohere#model-capabilities. 

### Additional context

https://docs.cohere.com/v2/docs/migrating-v1-to-v2

shaper:
Work items:
- [x] streaming
- [x] tool use
- [x] review tests
- [x] update docs

Resolves #3331

### DIFF
--- a/packages/cohere/src/cohere-chat-language-model.test.ts
+++ b/packages/cohere/src/cohere-chat-language-model.test.ts
@@ -334,8 +334,7 @@ describe('doGenerate', () => {
   });
 });
 
-// TODO(shaper): Fix streaming.
-describe.skip('doStream', () => {
+describe('doStream', () => {
   const server = new StreamingTestServer('https://api.cohere.com/v2/chat');
 
   server.setupTestEnvironment();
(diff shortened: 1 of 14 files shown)


packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -156,28 +201,28 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
...
-      text: response.text,
-      toolCalls: response.tool_calls
-        ? response.tool_calls.map(toolCall => ({
-            toolCallId: generateId(),
-            toolName: toolCall.name,
-            args: JSON.stringify(toolCall.parameters),
+      text: response.message.content?.[0]?.text ?? '',
+      toolCalls: response.message.tool_calls
+        ? response.message.tool_calls.map(toolCall => ({
+            // TODO(shaper): See if we can remove all `generateId` related bits. Main thing is how it's used in streaming?
```
lgrammel, line 211: Each tool call needs an ID, ideally the one sent by the provider API. If the provider does not send IDs for tool calls, we generate them.

packages/cohere/src/cohere-prepare-tools.ts
```diff
@@ -11,126 +11,90 @@ export function prepareTools(
...
+    | 'any'
+    | undefined;
   toolWarnings: LanguageModelV1CallWarning[];
 } {
   const tools = mode.tools?.length ? mode.tools : undefined;
   const toolWarnings: LanguageModelV1CallWarning[] = [];
 
   if (tools == null) {
-    return { tools: undefined, force_single_step: undefined, toolWarnings };
+    // TODO(shaper): Mistral references `tool_choice` below as well, look into it.
```
lgrammel, line 35: Some tool choices can be simulated by e.g. removing tools. So `auto` and `none` are always possible. For `required` or `tool`, you need a way to force the LLM to make a tool call.

shaper: I've revised to implement `auto` and `none`. I don't see a way to force a call in Cohere, so I've made `required` throw. `tool` follows what Mistral does, which appears to be:
- filter out all tools other than the one specified
- leave it to the LLM (unless Mistral actually supports `any`?) to decide whether to call it (so it's not a forced call)

packages/cohere/src/cohere-chat-language-model.ts
shaper: Revised to use provider's tool id.

shaper:
PTAL. I am still looking at the `cohere-chatbot-with-tools.ts` example which breaks due to `tool_calls` in an assistant message in what appears to be a schema-incompatible spot, but if there is a fix needed there we can land it in a separate PR IMO.

packages/cohere/src/cohere-chat-language-model.test.ts
```diff
@@ -439,13 +341,14 @@ describe('doGenerate', () => {
...
-      body: '{"model":"command-r-plus","chat_history":[{"role":"SYSTEM","message":"you are a friendly bot!"}],"message":"Hello"}',
+      body: '{"model":"command-r-plus","messages":[{"role":"system","content":"you are a friendly bot!"},{"role":"user","content":"Hello"}]}',
     });
   });
 });
 
-describe('doStream', () => {
-  const server = new StreamingTestServer('https://api.cohere.com/v1/chat');
+// TODO(shaper): Fix streaming.
+describe.skip('doStream', () => {
```
shaper, line 350: Woke up and remembered this -- will work on un-skipping these, should probably fix before merge.

lgrammel (approved):
pls update / activate stream tests

PR #3425, opened 2024-10-31
chore (packages): Standardize unsupported tool error type.

shaper:
In the fallback case for "unsupported tool choice type" when preparing tools for various providers we were using `Error` in some places and `UnsupportedFunctionalityError` in others. The latter seems most semantically meaningful, so this change standardizes on that everywhere.


lgrammel:
Ah yes - so this is behind an exhaustive check in a switch, meaning this should never happen unless there are some version/api changes that are incompatible.

lgrammel (approved):

PR #3464, opened 2024-11-04
fix (docs): Correct missing close-backtick breaking Anthropic text.

shaper:
The note on file parts has a broken section intended to link to Anthropic due to an absent closing backtick in the preceding GPT 4o text.


shaper:
See https://sdk.vercel.ai/docs/foundations/prompts#file-parts

jeremyphilemon (approved):

PR #3517, opened 2024-11-07
chore (npmrc): Remove strict-peer-dependencies config disable.

shaper:
This config setting is old and looks like it may be a workaround from pre-v8 pnpm versions. Removing both to reduce complexity and in case it has any ramification for the work we're doing to move to pnpm v9.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,1 @@
 auto-install-peers = true
-# necessary for a PNPM bug: https://github.com/pnpm/pnpm/issues/5152#issuecomment-1223449173
-strict-peer-dependencies = false


shaper:
@MaxLeiter added this, tagging him in case he remembers. Perhaps it was needed for AI SDK use within v0 e.g. issues only surfaced when installing and using it there? But I think given it lives in `.npmrc` it's probably only related to development.

lgrammel (approved):

lgrammel:
The bug is still open in pnpm fwiw. If we run into issues we can revert.

PR #3518, opened 2024-11-07
chore (config): Remove huggingface references.

shaper:
Cleans up two small references missed in earlier removal.


turbo.json
shaper, line 1: This is a one-liner other than json formatting-on-save. I can break that out but felt like it was small enough.

lgrammel (approved):

lgrammel:
Nice! we prob need something like this for inkeep, replicate as well

PR #3519, opened 2024-11-07
feat (provider/cohere): Pass along tool-plan response content.

shaper:


packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -298,9 +303,19 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
                 return;
               }
 
+              case 'tool-plan-delta': {
+                toolPlan += value.delta.message.tool_plan;
+                return;
+              }
+
```
lgrammel, line 310: can you handle it like a regular text delta and immediately send (instead of keeping a variable and sending all at once)?

would simplify things and lead to better streaming. is the anything here that I'm missing?

lgrammel:
Needs changeset.

packages/cohere/src/cohere-chat-language-model.ts
shaper: Good call, done.

shaper:
> Needs changeset.

Added patch changeset.

packages/cohere/src/cohere-chat-language-model.ts
```diff
@@ -204,9 +204,13 @@ export class CohereChatLanguageModel implements LanguageModelV1 {
     });
 
     const { messages, ...rawSettings } = args;
+    let text = response.message.content?.[0]?.text ?? '';
+    if (!text) {
+      text = response.message.tool_plan ?? '';
+    }
```
lgrammel, line 210: nit: `const text = response.message.content?.[0]?.text ?? response.message.tool_plan ?? '';`

lgrammel (approved):

PR #3587, opened 2024-11-09
feat (packages/codemod): Set up package for automated migrations.

shaper:


.changeset/cuddly-kings-give.md
shaper, line 1: I've added this first changeset but as this is a new package I'm not sure whether there is one-time npmjs.com setup we need to do for the GH action to be able to do its thing.

packages/codemod/src/test/jscodeshift-testUtils.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'jscodeshift/dist/testUtils' {
```
lgrammel, line 1: naming: `test-utils` (js convention, unless this was intentional for testing?)

lgrammel (approved):

packages/codemod/src/test/jscodeshift-testUtils.d.ts
shaper: Agreed re: naming, I think we might need this to match the jscodeshift dist filenames. Yes, it's just for testing so that we get ts-lint validity IIRC. I will look at it separately.

PR #3588, opened 2024-11-09
feat (packages/codemod): Add more automated transformations for 4.0.

shaper:
Broad brush strokes with generated test cases. There is likely more to do to transform with precision e.g. making sure affected sites are in the expected ctor, method, etc. (see `baseURL` for example).


packages/codemod/package.json
```diff
@@ -8,6 +8,7 @@
         "test/dist/**/*",
         "CHANGELOG.md"
     ],
+    "type": "module",
```
lgrammel, line 11: why is this needed? will there be any downstream complications for users?

lgrammel (approved):

packages/codemod/package.json
shaper: https://vite.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated has more background. I think it's basically low priority to make the change in any of our dev-facing libraries intended for import, and could cause issues for some folks. For this particular module given intended use it's probably ok, though admittedly deviant standard-wise.

shaper: Moved back to cjs for consistency, triggered by my finding that ESM required further edits for local iteration running `codemod` binary (now operational, which is nice). We can revisit later.

PR #3589, opened 2024-11-10
fix(gemini): fix functionResponse format

egoist:
I'm currently getting the following error when using tools with gemini:

```
Invalid JSON payload received. 
Unknown name \"response\" at 'contents[2].parts[0].function_response': 
Proto field is not repeating, cannot start list.
```

According to https://ai.google.dev/gemini-api/docs/function-calling

The correct format of `functionResponse`:

CleanShot 2024-11-10 at 15 11 28@2x

### DIFF
--- a/packages/google/src/convert-to-google-generative-ai-messages.ts
+++ b/packages/google/src/convert-to-google-generative-ai-messages.ts
@@ -133,7 +133,10 @@ export function convertToGoogleGenerativeAIMessages(
           parts: content.map(part => ({
             functionResponse: {
               name: part.toolName,
-              response: part.result,
+              response: {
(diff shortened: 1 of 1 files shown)


lgrammel:
From their official documentation: https://ai.google.dev/gemini-api/docs/function-calling

CleanShot 2024-11-11 at 10 55 08

```
"functionCall": {
  "name": "find_theaters",
  "args": {
    "movie": "Barbie",
    "location": "Mountain View, CA"
  }
}
```

I double checked that function calling works with the google API.

do you have a reproducible example where it fails?

lgrammel:
Continued in #3599

PR #3594, opened 2024-11-11
feat (packages/codemod): Even more automated transformations.

shaper:


shaper:
All broken out into separate PRs.

PR #3638, opened 2024-11-12
feat (package/codemod): Add codemod to update telemetry data.

shaper:


packages/codemod/src/test/__testfixtures__/update-telemetry-data.output.ts
```diff
@@ -0,0 +1,37 @@
...
+    console.warn('High latency detected');
+  }
+
+  // Accessing all paths
+  const metrics = {
+    finish: result.telemetry?.attributes['ai.response.finishReason'],
+    text: result.telemetry?.attributes['ai.response.text'],
+    tools: result.telemetry?.attributes['ai.response.toolCalls'],
+    latency: result.telemetry?.attributes['ai.response.msToFirstChunk'],
+    object: result.telemetry?.attributes['ai.response.object']
```
lgrammel, line 35: it is not possible to access telemetry like this. it goes to oauth. this might not be something to address in a codemod at all, it only affects a few telemetry providers like langchain that need to update their integrations and was listed for completeness only

lgrammel: (that being said, it would be useful to expose telemetry like this in the future - but it's not something we do today)

shaper: ok, great, thanks -- you proactively answered the note I just put in Slack!

PR #3640, opened 2024-11-12
feat (packages/codemod): Add codemod to replace langchain toAIStream.

shaper:


packages/codemod/src/test/__testfixtures__/replace-langchain-toaistream.input.ts
```diff
@@ -0,0 +1,9 @@
+// @ts-nocheck
+import { toAIStream } from 'ai/streams';
+import { model } from 'langchain';
+
+const stream = toAIStream(model.stream(), {
+  onToken: token => console.log(token)
+});
```
lgrammel, line 7: pls check example usage here: https://github.com/vercel/ai/blob/main/examples/next-langchain/app/api/chat/route.ts

lgrammel (approved):

PR #3641, opened 2024-11-12
feat (packages/codemod): Add codemod to remove topK provider setting.

shaper:


packages/codemod/src/test/__testfixtures__/remove-anthropic-topk.output.ts
```diff
@@ -0,0 +1,56 @@
...
+import { generateText, streamText, generateObject, streamObject } from 'ai';
+
+const anthropicClient = createAnthropic({
+  temperature: 0.7
+});
+
+const result = await generateText({
+  model: anthropicClient('claude-3'),
+  prompt: 'Hello',
+  topK: 10
```
shaper, line 12: This PR is a good example of one where ideally we'd have fuller refactoring, multifile, static-analysis-based infra to do it right.

The PR aims to address all three of the Provider "Removed `topK` setting" cases [here](https://sdk.vercel.ai/docs/troubleshooting/migration-guide/migration-guide-4-0#removed-topk-setting) and below.

I think landing this PR as is should be fine. To explain what I have in mind for any feedback/future thoughts:
- I definitely think this codemod should remove `topK` from provider create methods
- Ideally we would then (as we do in a limited set of cases here) proactively *add* topK with the value in the previously removed lines
- However, it could well be the case that the involved API calls using the model in question are all outside of this file
- So, perhaps we get ourselves in trouble by trying to do such a fix and taking it only halfway. Calls we miss will "silently fail" in that they will just lose the `topK` value they used to have.
- An alternative would be to drop a comment where we remove `topK` from a model construction and remind the dev to review any generate/stream API calls using this model to potentially add `topK` back in (or otherwise effect via their own wrapper logic, etc.).

packages/codemod/src/codemods/lib/remove-topk.ts
shaper, line 1: Also just to mention as we discussed briefly Monday -- I do still have a TODO in mind:
- look for jscodeshift util lib (generally, and in specific projects we know use it)
- and/or extract our redundant actions e.g. replaceImports to our own util lib

I am finding however that the LLM is pretty good at generating these, so I want to get through the easy removal/renames first and then will spend more time on the above.

packages/codemod/src/test/__testfixtures__/remove-anthropic-topk.output.ts
lgrammel: There will be lots of cases that we will not be able to do automatically. It might be best not to automate this one and instead add a warning to the migration doc that this needs manual migration (and also add that warning to the telemetry migration in the manual).

lgrammel (approved):

packages/codemod/src/test/__testfixtures__/remove-anthropic-topk.output.ts
shaper: Sounds good, to be explicit with my understanding:
- we should still have this codemod remove `topK` from the create method (helps devs with what will be a TS type-check error)
- I'll remove the update-subsequent-calls logic and add a note in the migration docs as you describe

shaper: Ah, maybe you mean, don't have this codemod at all because the type check failure and migration note can prompt the dev to fix it all.

PR #3642, opened 2024-11-12
feat (packages/codemod): Add codemod to remove provider facades.

shaper:


packages/codemod/src/test/__testfixtures__/remove-anthropic-facade.input.ts
```diff
@@ -0,0 +1,14 @@
+// @ts-nocheck
+import { Anthropic } from '@ai-sdk/anthropic';
+
+const anthropic = new Anthropic({
+  apiKey: 'key',
+  baseURL: 'url',
+  headers: { 'custom': 'header' }
+});
+
+const model = anthropic.chat('claude-3', {
```
shaper, line 10: Here (and for other providers) is another example of the same issue mentioned in https://github.com/vercel/ai/pull/3641#discussion_r1838830113

While we can easily (as the codemod currently does) look for method calls on the provider instance and correct them to create a model with the provider instance we track from above, this is just the tip of the iceberg in terms of potential usages needing correction.

Whatever we think re: the other PR we should likely apply to the approach we use here. Currently the codemod does the trivial easy correction.

packages/codemod/src/codemods/remove-google-facade.ts
```diff
@@ -0,0 +1,17 @@
...
+  return removeFacade(fileInfo, api, {
+    packageName: 'google',
+    className: 'Google',
+    createFnName: 'createGoogleGenerativeAI',
+    methodNames: [
+      'chat',
+      'generativeAI',
+      'embedding',
+      'textEmbedding',
+      'textEmbeddingModel',
```
lgrammel, line 14: how will these be affected?

packages/codemod/src/codemods/remove-openai-facade.ts
```diff
@@ -0,0 +1,11 @@
+import { API, FileInfo } from 'jscodeshift';
+import { removeFacade } from './lib/remove-facade';
+
+export default function transformer(fileInfo: FileInfo, api: API) {
+  return removeFacade(fileInfo, api, {
+    packageName: 'openai',
+    className: 'OpenAI',
+    createFnName: 'createOpenAI',
+    methodNames: ['chat', 'completion'],
```
lgrammel, line 9: replacement potentially problematic for 3rd party openai compatible providers

packages/codemod/src/test/__testfixtures__/remove-anthropic-facade.input.ts
lgrammel: we should be careful about the method name changes, some are needed. the provider constructor might be enough

lgrammel (approved):

packages/codemod/src/test/__testfixtures__/remove-anthropic-facade.input.ts
shaper: Updated to only edit the ctor.

PR #3648, opened 2024-11-12
feat (packages/codemod): Add script to scaffold new codemod.

shaper:
Goal to save us all some clicks/key-presses. Will surely evolve in time.

### DIFF
--- a/.changeset/proud-yaks-taste.md
+++ b/.changeset/proud-yaks-taste.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/codemod': patch
+---
+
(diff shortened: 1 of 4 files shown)


lgrammel (approved):

.changeset/proud-yaks-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/codemod': patch
+---
+
+feat (packages/codemod): Add script to scaffold new codemod.
```
lgrammel, line 5: internals dev changes might not need a release, so no changeset would have been needed. (no need to remove tho)

PR #3649, opened 2024-11-13
feat (packages/codemod): Add codemod to rm old provider registry exports.

shaper:


packages/codemod/src/test/__testfixtures__/remove-deprecated-provider-registry-exports.input.ts
shaper, line 1: Are there old examples I could try as fixtures to test this one further?

lgrammel: maybe in the version history of the repo. prob faster to make them up

lgrammel (approved):

PR #3661, opened 2024-11-13
documentation: streamText should be awaited in app router quickstart

srkirkland:
on the app router quickstart, `streamText` needs to be awaited so you can start streaming the result.  As written, you'd get something like `Property 'toDataStreamResponse' does not exist on type 'Promise<StreamTextResult<Record<string, CoreTool<any, any>>>>'.ts`


lgrammel:
This is made in prep for 4.0 which comes out soon.

PR #3681, opened 2024-11-14
feat (packages/codemod): Add codemod to rename formatStreamPart.

shaper:


packages/codemod/src/lib/upgrade.ts
shaper, line 1: I'll consider standardizing the naming of these later. Right now they're largely named after the phrasing in the migration guide. We could use id numbers to correlate/prefix, etc... but feel like that's getting too complicated.

lgrammel (approved):

PR #3682, opened 2024-11-14
feat (packages/codemod): Add codemod to rename parseStreamPart.

shaper:


packages/codemod/src/codemods/rename-parse-stream-part.ts
shaper, line 1: I plan to factor out a common rename-method lib in the near future. I considered combining this PR with #3681 but still feel like there's some value to separating each item for clarity of review/maintenance. Thinking could evolve.

lgrammel (approved):

PR #3696, opened 2024-11-15
feat (packages/codemod): Add codemod to rm experimental useAssistant.

shaper:


packages/codemod/src/test/test-utils.ts
shaper, line 1: This file is broken out separately in https://github.com/vercel/ai/pull/3695 but this is the PR where we needed to make use of it first.

lgrammel (approved):

PR #3722, opened 2024-11-17
fix (packages/codemod): Improve package version validation pre-upgrade.

shaper:
- scan all package.json files at or under the working directory
- require there to be at least one with 'ai'
- require that all with 'ai' meet the minimum sdk version
- boost the required version from 3.4.0 to 4.0.0 (was a bug)

Things we aren't yet doing that we can consider for the future:
- check '@ai-sdk/...' packages and versions
- install (or prompt user to install) upgraded versions for 'ai' and '@ai-sdk/...'

### DIFF
--- a/.changeset/strong-foxes-try.md
+++ b/.changeset/strong-foxes-try.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/codemod': patch
+---
+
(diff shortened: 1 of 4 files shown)


lgrammel (approved):
What if we just remove the check?

shaper:
Closing in favor of #3728

PR #3728, opened 2024-11-17
fix (packages/codemod): Remove package version check pre-upgrade.

shaper:


shaper:
Landing so that we can publish a new version ahead of Monday. For @lgrammel to review when he's back online.

PR #3729, opened 2024-11-17
chore (docs): note codemod exists for rm await streamText/Object.

shaper:
Should have been part of #3724.


shaper:
For @lgrammel to review Monday, landing now so docs are updated to go along with the codemod. Note the codemod isn't 100% (as discussed it removes the 'await' but doesn't update any surrounding code), but our codemod header docs cover this re: saying they're helpers and not comprehensive.

PR #3733, opened 2024-11-18
fix (packages/codemod): Only mutate files when changes are needed.

shaper:
Update all codemods to track whether they've made any changes and return null if not.

Uses #3732 for test compatibility.


packages/codemod/src/test/test-utils.test.ts
shaper, line 1: This and the next file are broken out into #3732 but are included here as this was developed as a single unit to allow validating that tests pass. I'll merge the other one first.

lgrammel (approved):

PR #3734, opened 2024-11-18
docs: await streamText call

billyjacoby:
Was going through the tut and noticed that `streamText` needed to be awaited!


lgrammel:
it won't any more in 4.0, still holding off on docs changes since we'll get it released soon

PR #3778, opened 2024-11-19
fix (docs): point perplexity example/docs to newest model

jeremypress:
Perplexity is not hosting the llama 3 version of their model anymore, so let's update to 3.1

src: https://docs.perplexity.ai/guides/model-cards


lgrammel (approved):

jeremypress:
thanks for merging @lgrammel !

PR #3781, opened 2024-11-19
feat (provider/togetherai): Add TogetherAI provider.

shaper:
Built on top of a new `openai-compatible` package for better code sharing as we add more top level providers that follow this pattern.


packages/openai-compatible/src/map-openai-compatible-completion-logprobs.ts
```diff
@@ -0,0 +1,24 @@
...
+    topLogprobs: logprobs.top_logprobs
+      ? Object.entries(logprobs.top_logprobs[index]).map(
+          ([token, logprob]) => ({
+            token,
+            logprob,
+          }),
+        )
+      : [],
+  }));
+}
```
lgrammel, line 24: Let's remove logprobs from the compatible provider. I'm leaning towards moving it into a provider extension since it's rarely supported.

packages/openai-compatible/src/openai-compatible-chat-settings.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+Default object generation mode that should be used with this model when
+no mode is specified. Should be the mode with the best results for this
+model. `undefined` can be returned if object generation is not supported.
+
+This is needed to generate the best objects possible w/o requiring the
+user to explicitly specify the object generation mode.
+  */
+  // TODO(shaper): This is really model-specific, move to config or elsewhere?
+  defaultObjectGenerationMode?: 'json' | 'tool' | undefined;
```
lgrammel, line 19: can you make it a constructor parameter for the model that is then defined in the providers? (which would have that knowledge)

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -0,0 +1,537 @@
...
+          },
+          warnings: [...warnings, ...toolWarnings],
+        };
+      }
+
+      case 'object-json': {
+        // TODO(shaper): Review vs. OpenAI impl here.
+        throw new UnsupportedFunctionalityError({
+          functionality: 'object-json mode',
+        });
```
lgrammel, line 151: copy what openai has including schemas

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -0,0 +1,537 @@
...
+
+  readonly modelId: OpenAICompatibleChatModelId;
+  readonly settings: OpenAICompatibleChatSettings;
+
+  private readonly config: OpenAICompatibleChatConfig;
+
+  constructor(
+    modelId: OpenAICompatibleChatModelId,
+    settings: OpenAICompatibleChatSettings,
+    config: OpenAICompatibleChatConfig,
```
lgrammel, line 54: include the default object generation mode in `config`

packages/openai-compatible/src/openai-compatible-completion-language-model.ts
```diff
@@ -0,0 +1,387 @@
...
+    const choice = response.choices[0];
+
+    return {
+      text: choice.text,
+      usage: {
+        promptTokens: response.usage.prompt_tokens,
+        completionTokens: response.usage.completion_tokens,
+      },
+      finishReason: mapOpenAICompatibleFinishReason(choice.finish_reason),
+      logprobs: mapOpenAICompatibleCompletionLogProbs(choice.logprobs),
```
lgrammel, line 201: rm logbprobs

packages/openai-compatible/src/openai-compatible-completion-language-model.ts
```diff
@@ -0,0 +1,387 @@
...
+    z.object({
+      text: z.string(),
+      finish_reason: z.string(),
+      logprobs: z
+        .object({
+          tokens: z.array(z.string()),
+          token_logprobs: z.array(z.number()),
+          top_logprobs: z.array(z.record(z.string(), z.number())).nullable(),
+        })
+        .nullish(),
```
lgrammel, line 349: rm logprobs

packages/openai-compatible/src/openai-compatible-completion-language-model.ts
```diff
@@ -0,0 +1,387 @@
...
+        text: z.string(),
+        finish_reason: z.string().nullish(),
+        index: z.number(),
+        logprobs: z
+          .object({
+            tokens: z.array(z.string()),
+            token_logprobs: z.array(z.number()),
+            top_logprobs: z.array(z.record(z.string(), z.number())).nullable(),
+          })
+          .nullish(),
```
lgrammel, line 376: same

packages/openai-compatible/src/openai-compatible-embedding-model.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const requestHeaders = await server.getRequestHeaders();
+
+    expect(requestHeaders).toStrictEqual({
+      authorization: 'Bearer test-api-key',
+      'content-type': 'application/json',
+      'custom-provider-header': 'provider-header-value',
+      'custom-request-header': 'request-header-value',
+      // 'openai-organization': 'test-organization',
+      // 'openai-project': 'test-project',
```
lgrammel, line 133: rm

packages/openai-compatible/src/openai-compatible-embedding-model.test.ts
```diff
@@ -0,0 +1,136 @@
...
+  });
+
+  it('should pass headers', async () => {
+    prepareJsonResponse();
+
+    const provider = createOpenAICompatible({
+      apiKey: 'test-api-key',
+      baseURL: 'https://my.api.com/v1/',
+      // organization: 'test-organization',
+      // project: 'test-project',
```
lgrammel, line 112: rm

packages/openai-compatible/src/openai-compatible-embedding-settings.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface OpenAICompatibleEmbeddingSettings {
+  /**
+Override the maximum number of embeddings per call.
+   */
+  maxEmbeddingsPerCall?: number;
+
+  /**
+Override the parallelism of embedding calls.
+    */
+  supportsParallelCalls?: boolean;
```
lgrammel, line 12: move to config?

lgrammel: (first 2 options)

packages/openai-compatible/src/openai-compatible-error.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type OpenAICompatibleErrorData = z.infer<
+  typeof openAICompatibleErrorDataSchema
+>;
+
+export const openAICompatibleFailedResponseHandler =
+  createJsonErrorResponseHandler({
+    errorSchema: openAICompatibleErrorDataSchema,
+    errorToMessage: data => data.error,
+  });
```
lgrammel, line 19: make provider-specific. different providers have different structures. have default that matches openai

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,163 @@
...
+import { OpenAICompatibleChatSettings } from './openai-compatible-chat-settings';
+import { OpenAICompatibleCompletionLanguageModel } from './openai-compatible-completion-language-model';
+import { OpenAICompatibleCompletionSettings } from './openai-compatible-completion-settings';
+import { OpenAICompatibleEmbeddingSettings } from './openai-compatible-embedding-settings';
+import { OpenAICompatibleEmbeddingModel } from './openai-compatible-embedding-model';
+
+export interface OpenAICompatibleProvider<
+  L extends string = string,
+  C extends string = string,
+  E extends string = string,
```
lgrammel, line 21: Prefer full words for generic in upper case, e.g. `CONTEXT`

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,163 @@
...
+export function createOpenAICompatible<
+  L extends string,
+  C extends string,
+  E extends string,
+>(
+  options: OpenAICompatibleProviderSettings,
+): OpenAICompatibleProvider<L, C, E> {
+  // TODO(shaper):
+  // - consider throwing if baseUrl, name, sufficient api key info not available
+  // - force only 'compatible' -- look into whether we can remove some 'strict' logic/configs entirely
```
lgrammel, line 96: yes - goal is that this works w/ compatible mode but is more flexible. then eventually openai provider should only be strict mode and we can simplify

packages/openai-compatible/src/openai-compatible-provider.ts
shaper: Done.

packages/openai-compatible/src/openai-compatible-embedding-model.test.ts
shaper: Done.

packages/openai-compatible/src/openai-compatible-embedding-model.test.ts
shaper: Done.

packages/openai-compatible/src/openai-compatible-chat-settings.ts
shaper: By `options` do you mean `config` e.g. `OpenAICompatibleChatLanguageModel` constructor signature is `(modelId, settings, config)`?

I'll move it there and see how to expose it to the concrete provider impl which today just passes a settings object (where I had it when you commented), but we could add an options or model-config arg as well.

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
shaper: Done, I omitted `structuredOutputs` since AFAIK it's only supported by OpenAI today.

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
shaper: I'd like to define a type for the values `'json' | 'tool' | undefined`. It appears the first reference is in LanguageModelV1. Is it appropriate to add something [there](https://github.com/vercel/ai/blob/a334c6ad18ce13eff0e5072783c902a93754f0fe/packages/provider/src/language-model/v1/language-model-v1.ts#L39)?

packages/openai-compatible/src/openai-compatible-embedding-settings.ts
shaper: Done.

packages/openai-compatible/src/openai-compatible-error.ts
shaper: I can see a couple of different ways to do this. The error data schema is further embedded in the schemas used in the chat/completion model classes, so just making the failed response handler configurable isn't sufficient for the goal.

I have in mind to break this PR into two so that we can start landing parts and work more incrementally:
1. initial openai-compatible package work
2. togetherai package atop it

I expect we can land (1) tomorrow after I iterate on your next feedback. We can then put a focused change together to make the error schema/response handler configurable. And we can iterate on (2) until it supports the model/feature combos we feel is good enough for a first ship target.

shaper:
Split into #3814 and #3812

Issue #3766, opened 2024-11-19
The v3.4 to v4 migration codemod is unusably slow


umstek:
### Description

I have a react project with webpack which is a bit old. However, `ai` worked without an issue. The only features I'm using are:

```ts
experimental_useObject
```

```ts
useCompletion
```

and there's likely nothing to migrate. But still it takes hours. I waited 4 hours when my machine crashed due to an unrelated issue.

### Code example

```ts
import { experimental_useObject as useObject } from 'ai/react';
```

```ts
import { useCompletion } from 'ai/react';
```

### AI provider

@ai-sdk/google v1.0.0 (in backend, so unrelated?)

### Additional context

Linux 6.8.0-48-generic/Linux 6.8.0-48-generic

lgrammel:
If you only use those functions, there is likely nothing to migrate. Can you just try using 4.0 without running the migration?

wong2:
<img width="886" alt="image" src="https://github.com/user-attachments/assets/fc60952c-11f0-44e3-9b54-f117e6055a35">

Yes, it's reallly slow

shaper:
Hi there, thanks for the report and sorry to hear the performance isn't good for your code base. A few things that could help us look into this if you can share:
- rough scale of codebase size (e.g. XXX KLOC)
- what kind of machine you're running on (e.g. rough year of Mac laptop, Linux server w/ N cores, ...)
- cpu load on machine (via `top`, `w`, Activity Monitor, etc.)

umstek:
@shaper 
less than 100KLOC
Ubuntu 24.04 on Core i7-8565U, 16GB RAM (Thinkpad)
I didn't notice any CPU usage due to the codemod running, typically was around 30-40%

shaper:
> @shaper less than 100KLOC Ubuntu 24.04 on Core i7-8565U, 16GB RAM (Thinkpad) I didn't notice any CPU usage due to the codemod running, typically was around 30-40%

Thank you. Do you know whether all of the code was already local, or if some was potentially loading over a network mount?

We have several potential performance improvements in mind and your feedback can help guide what order we approach them in.

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
The issue reports a concrete performance failure in an existing migration tool, making the codemod effectively unusable for a moderately sized local project.

### Evidence
- The v3.4-to-v4 migration codemod ran for over four hours without completing on a local codebase under 100 KLOC.
[...]

PR #3803, opened 2024-11-20
fix (codemod): Filter more dirs/files unlikely to need transform.

shaper:
Part of #3766.

### DIFF
--- a/packages/codemod/src/lib/transform.ts
+++ b/packages/codemod/src/lib/transform.ts
@@ -26,7 +26,11 @@ function buildCommand(
   let command = `${jscodeshift} -t ${codemodPath} ${targetPath} \
     --parser tsx \
     --ignore-pattern="**/node_modules/**" \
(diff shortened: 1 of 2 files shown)


packages/codemod/src/lib/transform.ts
```diff
@@ -26,7 +26,11 @@ function buildCommand(
   let command = `${jscodeshift} -t ${codemodPath} ${targetPath} \
     --parser tsx \
     --ignore-pattern="**/node_modules/**" \
-    --ignore-pattern="**/.*/**"`;
+    --ignore-pattern="**/.*/**" \
+    --ignore-pattern="**/dist/**" \
```
shaper, line 30: Slight risk re: `dist` and `build`, folks could... actually have a package in a dir named `build` that uses AI SDK for something, but that seems unlikely vs the value we gain from removing overhead of unnecessarily processing output files in such a dir.

lgrammel (approved):

PR #3812, opened 2024-11-21
feat (packages/openai-compatible): Base for OpenAI-compatible providers.

shaper:
Breakout from #3781. Follow-up items:
- configurable error data schema/response handler
- streamObject token usage is NaN
- https://github.com/vercel/ai/pull/3812#discussion_r1851523736 - look into `json_schema` and `structuredOutputs` in `openai-compatible-chat-language-model.ts`
- further docs on usage (under `content`)


packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,188 @@
...
+    throw new Error('Base URL is required');
+  }
+  const baseURL = withoutTrailingSlash(options.baseURL);
+
+  if (!options.name) {
+    throw new Error('Provider name is required');
+  }
+  const providerName = options.name;
+
+  const apiKey = loadApiKey({
```
shaper, line 112: I would like to have these validation checks so that a dev is informed if they create a provider with unusable options. However, as I've done it here introduces an interesting failure case where just importing the file can cause these to throw due to the default provider pattern we use in concrete impls, e.g.:
- `generate-text/togetherai.ts` imports `createTogetherAI` or `togetherai` from `@ai-sdk/togetherai`
- the import happens before the `dotenv/config` import so the API key env var is not yet set
- the import of the togetherai provider logic causes the initialization of `const togetherai = createTogetherAI()` which will throw as there is no API key set
- a fix can include moving `dotenv/config` import before the TogetherAI imports

The above example is somewhat contrived and particular to our usage but if we hit it I am sure others would as well. Options include:
- do this validation as an early step in methods that would be impacted (but this would be annoying verbosity as there are a fair number of them, and it would be a small amount of redundant overhead on each call)
- only instantiate the default provider instance lazily (some add'l code complexity)
- don't do these checks (or at least, don't do the API key one)

Curious to hear your thoughts.

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,193 @@
...
+  languageModel(
+    modelId: CHAT_MODEL_IDS,
+    settings?: OpenAICompatibleChatSettings,
+    config?: OpenAICompatibleChatConfig,
+  ): LanguageModelV1;
+
+  chatModel(
+    modelId: CHAT_MODEL_IDS,
+    settings?: OpenAICompatibleChatSettings,
+    config?: OpenAICompatibleChatConfig,
```
shaper, line 40: Using config here (it was absent in previous iterations) means that in concrete impls e.g. TogetherAIProvider we end up having to deal with the config's [required parameters](https://github.com/vercel/ai/blob/90b9c52bea5cd6dc3c454fb70ca976d31112bc43/packages/openai-compatible/src/openai-compatible-completion-language-model.ts#L30), which would otherwise be hidden from the concrete impl (we deal with them in `getCommonModelConfig` below).

There are various ways to address, for example:
- create and use a different type here that only has what we do want to expose to the concrete impl (defaultObjectGenerationMode)
- put the bits we want to expose in settings instead, and maybe merge that into the config in `chatModel` impl logic
- make the config params optional instead of required and add validation logic elsewhere if needed
- add a method in `OpenAICompatibleProvider` or exported as part of this file that allows the concrete impls to construct a config, with args being only the bits we need from the concrete impl (but then we may want to change how the Provider is instantiated since currently we do it as part of that flow, and then the Provider pattern is more deviant from our others, which wouldn't be the end of the world but is nice to avoid)

I may also not be thinking of something, or I may have missed what you have in mind, in any case helpful to hear your design view before I go farther.

packages/openai-compatible/README.md
```diff
@@ -0,0 +1,10 @@
+# AI SDK - OpenAI Compatible Provider
+
+This package aims to provide a base to speed and support the implementation of a
+range of concrete OpenAI-compatible providers. Given this purpose, it is not
+itself a concrete provider intended for direct instantiation and use.
+
+The primary [OpenAI provider](../openai/README.md) is heavier-weight than what
+this package offers. That provider also includes OpenAI-specific experimental
+and legacy features which may not be relevant for more general purpose
+OpenAI-compatible providers.
```
lgrammel, line 10: add usage examples

lgrammel: (also a few under /examples/ai-core )

packages/openai-compatible/src/convert-to-openai-compatible-chat-messages.ts
```diff
@@ -0,0 +1,110 @@
...
+          content: content.map(part => {
+            switch (part.type) {
+              case 'text': {
+                return { type: 'text', text: part.text };
+              }
+              case 'image': {
+                // TODO(shaper): Add back the below.
+                throw new UnsupportedFunctionalityError({
+                  functionality: 'Image content parts in user messages',
+                });
```
lgrammel, line 36: Let's add back before landing

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -0,0 +1,544 @@
...
+          warnings: [...warnings, ...toolWarnings],
+        };
+      }
+
+      case 'object-json': {
+        return {
+          args: {
+            ...baseArgs,
+            response_format: { type: 'json_object' },
+          },
```
lgrammel, line 156: this can be much more complex with `json_schema` - it would be useful to have `structuredOutputs` setting (default to false) on `options` level

lgrammel: can be separate pr

packages/openai-compatible/src/openai-compatible-completion-language-model.ts
```diff
@@ -0,0 +1,344 @@
...
+  choices: z.array(
+    z.object({
+      text: z.string(),
+      finish_reason: z.string(),
+    }),
+  ),
+  usage: z.object({
+    prompt_tokens: z.number(),
+    completion_tokens: z.number(),
+  }),
```
lgrammel, line 319: make `nullish`

packages/openai-compatible/src/openai-compatible-provider.ts
lgrammel: not a fan of the 3rd arg bc
- this api is user-facing and adding args will limit our future options and impact our maintainance.
- we dont control what we expose (full config)
- config vs settings separation becomes meaningless (settings is intended for user, config internally)
- I'm trying to avoid > 2 args wherever i can. ideally 1 arg with object, in some cases (like this) 2 args for convenience. 3 gets confusing and harder to maintain

if users need to override some of the configs from the default, we can add them to `settings` (in addition to `config`).

if we want this to be extensible, we should look at some other mechanism. Fwiw I think the openai-compatible providers should fully re-implement the provider file to do what they need to do and to have max flexibility at that level (e.g. for their own API key env params, to limit the kinds of models, etc.). 

My approach would be not to restrict this part (i.e. `OpenAICompatibleProvider` should only be for direct instantiation, vanity providers should instead implement `ABCProvider` from scratch by copying and adjust this file)

packages/openai-compatible/src/openai-compatible-provider.ts
lgrammel: as above, the easiest and potentially best way would be to not aim for extensibility / reuse for this particular file.

lgrammel: example why: several providers will not have embeddings & completion models and they will need to limit the exposed API. There are many such cases and at the end just having a custom provider file for each one is imo easier to understand and maintain.

packages/openai-compatible/src/openai-compatible-provider.ts
shaper: Thanks, this is super helpful and clarifies things nicely.

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
shaper: Noted to look into subsequently.

packages/openai-compatible/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Provider Instance
+
+You can import the provider creation method `createOpenAICompatible` from `@ai-sdk/openai-compatible`:
+
+```ts
+import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
+```
+
+## Example
```
shaper, line 23: There should be a section on creating a custom provider leveraging this package. I'm planning to add that once we have the TogetherAI one done.

examples/ai-core/src/generate-text/openai-compatible-openai-image.ts
shaper, line 1: I wasn't able to find a model on together.ai that supported image inputs after 30m of trying a few that seemed like they might. So I switched to OpenAI and validated it's operational. I will try with other companies/models later as we generalize the error schema and so forth to make it easier to test others.

examples/ai-core/src/embed-many/openai-compatible-togetherai.ts
```diff
@@ -0,0 +1,25 @@
+import 'dotenv/config';
+import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
+import { embedMany } from 'ai';
+
+async function main() {
+  const togetherai = createOpenAICompatible({
+    apiKeyEnvVarName: 'TOGETHER_AI_API_KEY',
```
lgrammel, line 7: let's completely remove the concept of env variables for openai compatible. it's an unnecessary complexity in this special case, users can just pass the env property and we have less API surface

lgrammel: (it it only there for convenience to begin with, and here convenience becomes complexity)

packages/openai-compatible/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+```ts
+import { createOpenAICompatible } from '@ai-sdk/openai-compatible';
+import { generateText } from 'ai';
+
+const { text } = await generateText({
+  model: createOpenAICompatible({
+    apiKeyEnvVarName: 'EXAMPLE_API_KEY',
```
lgrammel, line 31: directly pass env property instead

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+  /**
+The name of the environment variable from which to load the API key (if a key isn't explicitly provided).
+   */
+  apiKeyEnvVarName?: string;
+
+  /**
+Description of the API key environment variable (for use in error messages).
+   */
+  apiKeyEnvVarDescription?: string;
+
```
lgrammel, line 80: lets remove api key related code completely. instead users can provide the key via headers. reason being that some providers have no api keys and there are different ways to send it via headers

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -0,0 +1,183 @@
...
+  const providerName = options.name;
+
+  const getHeaders = () => ({
+    Authorization: `Bearer ${loadApiKey({
+      apiKey: options.apiKey,
+      environmentVariableName: options.apiKeyEnvVarName ?? '',
+      description: options.apiKeyEnvVarDescription ?? '',
+    })}`,
+    ...options.headers,
+  });
```
lgrammel, line 118: we can get rid of this and just use options.headers

examples/ai-core/src/embed-many/openai-compatible-togetherai.ts
shaper: Done.

lgrammel (approved):

PR #3813, opened 2024-11-21
Export processChatResponse for External Use

nabettu:
This PR adds an export statement for the processChatResponse function. By exporting this function, it becomes accessible for developers who need to reuse or extend its functionality in their own implementations.

Exporting this function aligns with modular and reusable code principles, offering greater flexibility without impacting existing internal functionality.

Use Case:
For example, developers integrating this package may wish to handle chat responses programmatically while leveraging the function’s core logic.


lgrammel:
Not exposed on purpose for now since the current API is highly unstable and we will make breaking changes in patch versions. Once the abstraction is stabilized it'll be exposed. Right now `processDataStream` is exposed & you can  use a copy of the processChatResponse code in the meantime.

nabettu:
ok.
thx for review.

PR #3814, opened 2024-11-21
feat (provider/togetherai): Add Together.ai provider.

shaper:


packages/openai-compatible/src/openai-compatible-chat-language-model.ts
shaper, line 1: Should I break this out as it's a change in another package? It doesn't actually affect togetherai behavior as object generation is via `tool`, at least by default.

This was essentially me restoring some logic I believe we do want in some form, so that OpenAI compatible providers can do the right thing, and (for now) capturing findings from trying to get json mode working with Together.ai (not yet successful).

examples/ai-core/src/e2e/togetherai.test.ts
shaper, line 1: I put these together to easily test a range of features against a potentially large set of models, and more easily regress. A few notes:
- the assertions are super bare-bones, not tightly reviewed, AI-generated (I did tweak a few)
- the tests run serially, you can add `.concurrent` after `describe` and they parallelize nicely, but then you also blow out your quota on Together, lol
- these are effectively smoke/integration tests, I waffled on putting them here vs in the `togetherai` package. For now, here seemed fine. It's where the `.env` files are, and the image assets, etc. but I'm open to your thought.
- oddly I see some of our example image attachment scripts setting image parts with `.toString('base64')` and others just setting the file bytes. The docs say the `data` field of an image part should be either a url or base64-encoded, so I followed that guidance.
- you can use `vitest` wildcards to run slices of tests, which is nice, e.g. `pnpm test:e2e -t "Qwen"` or `pnpm test:e2e -t "stream text"`

lgrammel (changes requested):
please separate the 2e2 tests. we need a different strategy for running them because they can destabilize main

lgrammel (approved):
The tests don't seem to run in CI, so that's good.

examples/ai-core/src/generate-text/togetherai.ts
```diff
@@ -1,16 +1,16 @@
...
-  name: 'togetherai',
+const togetherai = createTogetherAI({
   apiKey: process.env.TOGETHER_AI_API_KEY!,
   baseURL: 'https://api.together.xyz/v1/',
 });
 
 async function main() {
   const { text, usage } = await generateText({
-    model: togetherai('google/gemma-2-9b-it'),
+    // model: togetherai('google/gemma-2-9b-it'),
```
lgrammel, line 12: rm

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -153,7 +153,25 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV1 {
...
             ...baseArgs,
-            response_format: { type: 'json_object' },
+            response_format:
+              // TODO(shaper): Reconcile behavior below when a schema exists,
+              // re: `schema` vs `json_schema` keys.
+              //
+              // TogetherAI API docs: https://docs.together.ai/docs/json-mode
+              // { type: 'json_schema', schema: mode.schema }
+              //
+              // What we have below is taken from `OpenAIChatLanguageModel`.
```
lgrammel, line 163: let's remove this from the openai compat provider (in general i try to avoid having todo comments in our codebase). we might need a special together provider (similar to groq) if they don't follow what openai does, or we might want to look into configuration points for openai-compat

packages/togetherai/src/togetherai-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+  TogetherAICompletionModelId,
+  TogetherAICompletionSettings,
+} from './togetherai-completion-settings';
+
+export interface TogetherAIProviderSettings
+  extends OpenAICompatibleProviderSettings {
+  /**
+   TogetherAI API key.
+   */
+  apiKey?: string;
```
lgrammel, line 31: Add baseURL & headers (important for proxies and every provider has them for that reason)

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
shaper: Done.

packages/togetherai/src/togetherai-provider.ts
shaper: We had these via extension of `OpenAICompatibleProviderSettings` but I get now that it's better to have a copy as we avoid leaking settings we don't want to expose in TogetherAI's settings. Changed to copy those (and `fetch`).

shaper:
> The tests don't seem to run in CI, so that's good.

Interpreting as it's ok to keep them where they are for now since they are only run manually.

PR #3816, opened 2024-11-21
feat (provider): Define type for ObjectGenerationMode.

shaper:
As we generalize an OpenAI-compatible base provider we need to surface this mode setting to a concrete provider implementation, catalyzing the need to define and export a named type for the possible values. See one [sample usage](https://github.com/vercel/ai/blob/90b9c52bea5cd6dc3c454fb70ca976d31112bc43/packages/openai-compatible/src/openai-compatible-chat-language-model.ts#L44) site.

### DIFF
--- a/packages/provider/src/language-model/v1/language-model-v1.ts
+++ b/packages/provider/src/language-model/v1/language-model-v1.ts
@@ -36,7 +36,7 @@ model. `undefined` can be returned if object generation is not supported.
 This is needed to generate the best objects possible w/o requiring the
 user to explicitly specify the object generation mode.
    */
-  readonly defaultObjectGenerationMode: 'json' | 'tool' | undefined;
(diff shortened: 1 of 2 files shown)


packages/provider/src/language-model/v1/language-model-v1.ts
```diff
@@ -280,3 +280,5 @@ export type LanguageModelV1StreamPart =
 
   // error parts are streamed, allowing for multiple errors
   | { type: 'error'; error: unknown };
+
+export type LanguageModelV1ObjectGenerationMode = 'json' | 'tool' | undefined;
```
lgrammel, line 284: add jsdoc

lgrammel (approved):

lgrammel:
This needs a changeset for `provider` to ship.

PR #3832, opened 2024-11-21
feat (codemod): Ensure test fixtures use valid syntax.

shaper:
Would have flagged the issue that led to #3818

### DIFF
--- a/packages/codemod/package.json
+++ b/packages/codemod/package.json
@@ -19,6 +19,7 @@
         "scaffold": "tsx scripts/scaffold-codemod.ts"
     },
     "devDependencies": {
+        "@babel/parser": "^7.26.2",
         "@types/cli-progress": "^3.11.6",
         "@types/debug": "^4.1.12",
         "@types/jscodeshift": "^0.12.0",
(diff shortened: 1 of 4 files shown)


lgrammel (approved):

packages/codemod/package.json
```diff
@@ -19,6 +19,7 @@
         "scaffold": "tsx scripts/scaffold-codemod.ts"
     },
     "devDependencies": {
+        "@babel/parser": "^7.26.2",
```
lgrammel, line 22: we can use typescript also, have less dependencies

shaper: Done, good idea, a few notes but still worth it tradeoff-wise IMO:
- we already have an indirect babel dependency for this package via `jscodeshift`
- typescript is slower even with the tiny fixture inputs (tests take 1.5 - 2s in some cases and were all under ~1s previously)
- babel is what jscodeshift uses (at least as we have it configured today) so it's a more direct comparison for our specific need, but typescript also gets us type-checking, and syntax validation wise I'm sure they're either similar or typescript is better
- typescript required more boilerplate/config code, but looks more flexible/customizable
- I haven't yet tested js/jsx with this change directly (though it's set up so it should work)

Issue #3779, opened 2024-11-19
Expose the LLM provider response in AISDKError


williamlmao:
### Feature Description

I have two other feature requests (#3451, #3208) that I've opened, both of which I will be closing in favor of this one as I believe this suggestion is the best way to handle this.

Currently when you use generateObject or generateText with a function call and the LLM responds with data that does not match your provided schema, the behavior is that the vercel AI SDK will throw a InvalidToolArgumentsError or AISDKError, which just returns the zod parsing error.

The issue here is that now the entire generateText or generateObject call is throwing a zod error, and the response from the provider is no longer accessible at all, even though that was technically a successful response. This is a problem for a few reasons

1. We lose access to what the LLM actually generated. Sometimes the response is recoverable / fixable, but we do not have an opportunity to handle those error cases.
2. Usage data is now lost (I know we could still save it with fetch intercepts, but this means we have to work with the data outside of the calling function scope)

I think of the vercel AI SDK as an extension of calling these providers, so it feels weird to me that a successful response from the LLM provider is completely inaccessible due to a zod parse error. I do agree that a generateText/generateObject should throw this error, but it should include the full LLM provider response. 

In the case of #3208 when Groq throws an error directly from their API, the error that groq throws should be passed through the AISDKError. 

If the error was thrown like this, that would be a huge, huge help.
```
try {
  const {result} = await generateText({
    ...
  });
} catch (e: InvalidToolArgumentsError) {
  e.response
  // Do something with the e.response
 // Log usage data still
}

```

lgrammel:
You can already do this: check if the error with `InvalidToolArguments.isInstance` and then access `toolArgs` (which are the incorrect args)

williamlmao:
Thanks @lgrammel! I opened #3869 to point people to this on the Error Handling page. 

That being said though, is there any way to access some of the other data we'd expect on a successful LLM response, such as `usage` or `experimental_providerMetadata`?

williamlmao:
Hi @lgrammel - ran into this again with generateObject. I looked through the error types in the docs, but I didn't see any equivalent error helper to access the incorrect object generation. 

Ultimately, we still think the Vercel AI SDK should expose more at the request/response level. Right now we have this feeling that we're fighting the framework due to over abstraction. If we're able to access the full request/response regardless of ai sdk errors, then that would solve everything and anything the sdk did not already provide, we'd be able to build around ourselves. 

[...]

PR #3865, opened 2024-11-25
feat (ai/core): add experimental tool call repair

lgrammel:
## Tasks

- [x] implementation
   - [x] expose ToolCallRepairFunction
   - [x] parse tool call
      - [x] implementation
      - [x] test
   - [x] generateText
   - [x] streamText
   - [x] tool call repair error
- [x] examples
- [x] documentation
   - [x] tool calling guide error handling section
   - [x] tool calling guide repair section
   - [x] generateText reference
   - [x] streamText reference
   - [x] tool call repair error
- [x] changeset


williamlmao:
👀 

This looks interesting! I would honestly still rather just have the ability to catch the full response from the LLM provider and have the flexibility to deal with things however I want. Abstractions and helpers like this are awesome, but if we can _only_ access the provider response through these then the SDK starts to feel rigid / over abstracted. I think having both is the best case though - an easy helper for those who want to use it, but ultimately flexibility for someone who needs to do something more custom.

I would love to know where your head is regarding my last comment on #3779, I am happy to put in a PR to expose the full provider response on the error object if you are aligned. Thanks!

lgrammel:
@williamlmao the approach that you propose has several limitations / downsides:

a) full responses are not always available (streaming, or providers w/o api calls)
b) for correct multi-step behavior, the error recovery / tool call repair needs to be handled inside generateText/streamText
c) exposing provider-specific results needs to dependencies that the AI SDK is trying to abstract, namely code that relies on provider-specific details

I want to explore tool call repair first (and if it works structured output repair as well), before looking into a more provider specific approach if needed. My hunch is that tool call repair & structured output repair will end up being sufficient and better integrated.

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -396,6 +397,67 @@ return result.toDataStreamResponse({
...
+  The tool call repair feature is experimental and may change in the future.
+</Note>
+
+Language models sometimes fail to generate valid tool calls,
+especially when the parameters are complex or the model is smaller.
+
+You can use the `experimental_toToolCallRepair` function to attempt to repair the tool call
+with a custom function.
+
+You can use different strategies to repair the tool call, e.g.:
```
nicoalbanese, line 412:
```suggestion
You can use different strategies to repair the tool call:
```

content/docs/07-reference/05-ai-sdk-errors/ai-tool-call-repair-error.mdx
```diff
@@ -0,0 +1,28 @@
+---
+title: ToolCallRepairError
+description: Learn how to fix ToolCallRepairError
```
nicoalbanese, line 3:
```suggestion
description: Learn how to fix AI SDK ToolCallRepairError
```

nicoalbanese (approved):

PR #3883, opened 2024-11-25
Add support apiVersion argument to azure provider

imamar94:
Not every region support all api version, hence adding `apiVersion` to azure provider will allow wider usecase. I personally have problem with this hardcoded api-version, since my model deployment is in europe, currently I can't bump the version to 2024-10-01.

+ added to the docs
+ added new test


lgrammel:
Thank you - this has come up a few times now. Will integrate.

lgrammel:
Continued in https://github.com/vercel/ai/pull/3886

PR #3908, opened 2024-11-27
feat (provider/google-vertex): Add support for edge runtime.

shaper:


packages/google-vertex/src/google-vertex-auth-edge.ts
shaper, line 1: Notes:
- this (and the `google-auth-library` version also in this PR) are provided for discussion purposes
- this is not yet tested on a true edge deploy (will do that later tonight) but they pass the edge vitest (though the old one did too). I have another impl based on the jose module we should be able to make work if this one doesn't
- the main point as we've discussed is to make the auth-token generation pluggable, and we use this as part of the `e2e/google-vertex.test.ts` suite under `examples/ai-core`
- open design discussion: whether we do that via `generateAuthToken` as a user setting or via `headers`. Currently this PR has both (and changes `headers` setting to async just for expediency while we discuss). This needs to be decided on and cleaned up before landing.

packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
```diff
@@ -0,0 +1,13 @@
+import { GoogleAuth } from 'google-auth-library';
+
+const auth = new GoogleAuth({
+  scopes: ['https://www.googleapis.com/auth/cloud-platform'],
+});
+
+export async function generateAuthToken() {
+  console.log('Generating auth token');
```
lgrammel, line 8: remove console.log

packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
```diff
@@ -0,0 +1,13 @@
...
+  const client = await auth.getClient();
+  const token = await client.getAccessToken();
+  console.log('Auth token generated', token);
```
lgrammel, line 11: rm

packages/google/src/google-generative-ai-embedding-model.ts
```diff
@@ -18,7 +18,7 @@ import {
 type GoogleGenerativeAIEmbeddingConfig = {
   provider: string;
   baseURL: string;
-  headers: () => Record<string, string | undefined>;
+  headers?: () => Promise<Record<string, string | undefined>>;
```
lgrammel, line 21: this is a breaking change. we need to allow both variants (promise and non-promise). also it's a config that we use internally, so i recommend using `undefined` and not `?` (forces us to always set it)

packages/google/src/google-generative-ai-language-model.test.ts
```diff
@@ -458,9 +458,9 @@ describe('doGenerate', () => {
     withTestServer(prepareJsonResponse({}), async ({ call }) => {
       const provider = createGoogleGenerativeAI({
         apiKey: 'test-api-key',
-        headers: {
+        headers: async () => ({
           'Custom-Provider-Header': 'provider-header-value',
-        },
+        }),
```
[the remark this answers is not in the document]
lgrammel: would also simplify the tests

packages/google/src/google-generative-ai-settings.ts
```diff
@@ -40,6 +40,7 @@ Optional. A list of unique safety settings for blocking unsafe content.
    */
   safetySettings?: Array<{
     category:
+      | 'HARM_CATEGORY_UNSPECIFIED'
```
lgrammel, line 43: are these harm categories the same in vertex & genai? i vaguely remember differences.

packages/google/src/google-generative-ai-settings.ts
```diff
@@ -52,4 +53,11 @@ Optional. A list of unique safety settings for blocking unsafe content.
       | 'BLOCK_ONLY_HIGH'
       | 'BLOCK_NONE';
   }>;
+
+  /**
+Optional. When enabled, the model will use Google search to ground the response.
+
+@see https://cloud.google.com/vertex-ai/generative-ai/docs/grounding/overview
+   */
+  useSearchGrounding?: boolean;
```
lgrammel, line 62: is this supported in both genai & vertex? re shadow features, i think we must avoid that. if it's only supported in vertex, can we somehow supply a different interface for vertex? what if a feature is only supported in genai (happens often)

lgrammel:
Needs docs update (showing how to use different auth solutions) and changeset (works on edge now).

packages/google-vertex/package.json
```diff
@@ -31,19 +31,19 @@
   },
   "dependencies": {
     "@ai-sdk/provider": "1.0.1",
-    "@ai-sdk/provider-utils": "2.0.2"
+    "@ai-sdk/provider-utils": "2.0.2",
+    "@ai-sdk/google": "1.0.4",
+    "google-auth-library": "^9.15.0"
```
shaper, line 36: This is not for merge -- it's here as the for-discussion experimental auth-lib file is in the PR. I assume we won't be landing either of the experimental methods, but wanted to be able to talk about them.

I mentioned maybe we start a `contrib` dir/package/etc., but in a way docs/templates are that, I suppose.

packages/google/src/google-generative-ai-embedding-model.ts
shaper: Made required, since it's not a promise it's easy to pass `{}` when we need to, so not adding an `| undefined`.

packages/google/src/google-generative-ai-language-model.test.ts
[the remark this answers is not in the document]
shaper: Reworked to keep `headers` as non-Promise, added `experimental_getHeadersAsync`. It is not exposed to the user in the `GoogleGenerativeAI` provider -- only in `GoogleVertex`.

packages/google/src/google-generative-ai-settings.ts
shaper: Reworked as:

`GoogleGenerativeAISettings` - has the LCD subset common to both
`InternalGoogleGenerativeAISettings` extends above and adds grounding
`GoogleVertexSettings` extends `InternalGoogleGenerativeAISettings`

If there were a GenAI-only feature we can keep the interface names and alter the definitions as needed, either using interface composition differently, or Omit and `&`, etc..

Here's a `v0`-provided example, but until we actually have a GenAI-only feature(s) keeping it the simple inheritance as above seems reasonable.

```
interface BaseSettings {
  // Common settings for both
}

interface GroundingSettings {
  grounding?: GroundingConfig;
}

interface GenAIOnlySettings {
  genAIOnlyFeature: string;
}

type GoogleGenerativeAISettings = BaseSettings & GenAIOnlySettings;
type InternalGoogleGenerativeAISettings = BaseSettings & GenAIOnlySettings & GroundingSettings;
type GoogleVertexSettings = BaseSettings & GroundingSettings & {
  // Vertex-specific settings
};
```

packages/google/src/google-generative-ai-settings.ts
shaper: The only difference I see is for the older PaLM models. Here is my understanding:

[GenAI Harm Categories](https://ai.google.dev/api/generate-content#harmblockthreshold)
[Vertex Harm Categories](https://cloud.google.com/python/docs/reference/aiplatform/latest/google.cloud.aiplatform_v1.types.HarmCategory)

The PaLM entries in the GenAI link are missing:

```
HARM_CATEGORY_DEROGATORY | PaLM - Negative or harmful comments targeting identity and/or protected attribute.
-- | --
HARM_CATEGORY_TOXICITY | PaLM - Content that is rude, disrespectful, or profane.
HARM_CATEGORY_VIOLENCE | PaLM - Describes scenarios depicting violence against an individual or group, or general descriptions of gore.
HARM_CATEGORY_SEXUAL | PaLM - Contains references to sexual acts or other lewd content.
HARM_CATEGORY_MEDICAL | PaLM - Promotes unchecked medical advice.
HARM_CATEGORY_DANGEROUS | PaLM - Dangerous content that promotes, facilitates, or encourages harmful acts.
```

I assume these have been updated over time. Based on the current specs linked above, my take is that this PR is newly updated to reflect the latest, and the only consideration is whether to add the PaLM entries:
- we were missing the `HARM_CATEGORY_UNSPECIFIED` in GenAI but they are relevant there
- we were missing `HARM_CATEGORY_CIVIC_INTEGRITY` in both GenAI and Vertex, and it's relevant in both
- we were missing the PaLM entries above in both. GenAI spec has the PaLM entries and Vertex doesn't
- https://ai.google.dev/palm_docs notes "Warning: The PaLM API is decomissioned. The Vertex AI PaLM API [is scheduled to be decomissioned in October 2024](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text). Please upgrade to the Gemini API. Learn more about this in the [PaLM API deprecation guide](https://ai.google.dev/palm_docs/deprecation)."

It is not clear to me that the PaLM models are a forward-going concern, so I would personally omit them for now.

packages/google-vertex/src/load-module-dynamically.ts
```diff
@@ -0,0 +1,21 @@
+export async function loadModuleDynamically<T>({
+  libraryName,
+  exportName,
+}: {
+  libraryName: string;
+  exportName: string;
+}): Promise<T> {
+  // Use google auth library (for Node.js).
```
lgrammel, line 8: needs update / docs

packages/google-vertex/src/google-vertex-embedding-model.test.ts
```diff
@@ -1,148 +1,277 @@
...
+// Mock the postJsonToApi utility
+vi.mock('@ai-sdk/provider-utils', async importOriginal => {
+  const actual = await importOriginal<
+    typeof import('@ai-sdk/provider-utils')
+  >();
+  return {
+    ...actual,
+    postJsonToApi: vi.fn(),
+  };
+});
```
lgrammel, line 15: ideally align with other embedding tests that go through msw / network level mocking to keep with existing patterns, e.g. `packages/openai/src/openai-embedding-model.test.ts`

shaper: Done.

packages/google-vertex/src/load-module-dynamically.ts
shaper: File no longer relevant.

.changeset/early-spiders-collect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/google-vertex': major
+'@ai-sdk/google': patch
+---
+
+feat (provider/google-vertex): Add support for edge runtime.
```
shaper, line 6: As a major version bump, do we link to more info or provide more detail in the message?

content/providers/01-ai-sdk-providers/11-google-vertex.mdx
```diff
@@ -89,6 +114,87 @@ You can use the following optional settings to customize the Google Generative A
...
+  - A promise that resolves to headers: `Promise<Record<string, string | undefined>>`
+
+- **fetch** _(input: RequestInfo, init?: RequestInit) => Promise&lt;Response&gt;_
+
+  Custom [fetch](https://developer.mozilla.org/en-US/docs/Web/API/fetch) implementation.
+  Defaults to the global `fetch` function.
+  You can use it as a middleware to intercept requests,
+  or to provide a custom fetch implementation for e.g. testing.
+
+### Edge Runtime
```
lgrammel, line 133: it might be helpful to briefly explain what an edge runtime is (Vercel Edge, Cloudflare), that it only offers a subset of the standard JS API (e.g. no files), and that it is not compatible with the google auth library.

packages/google-vertex/src/edge/google-vertex-provider-edge.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function createVertex(
+  options: GoogleVertexProviderSettings = {},
+): GoogleVertexProvider {
+  return createVertexOriginal({
+    ...options,
+    headers:
+      options.headers ??
+      (async () => ({
+        Authorization: `Bearer ${await generateAuthToken()}`,
```
lgrammel, line 18: Do you want to pass through some of the options into `generateAuthToken`? or leave for later?

packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
```diff
@@ -0,0 +1,27 @@
...
+  const auth = getAuth(options || {});
+  const client = await auth.getClient();
+  const token = await client.getAccessToken();
+  return token?.token || null;
+}
+
+// For testing purposes only
+export function _resetAuthInstance() {
+  authInstance = null;
+}
```
lgrammel, line 27: Why is the singleton needed here?

packages/google-vertex/src/google-vertex-provider.ts
```diff
@@ -43,24 +43,22 @@ Your Google Vertex project. Defaults to the environment variable `GOOGLE_VERTEX_
...
-Complete list of authentication options is documented in the
-GoogleAuthOptions interface:
-https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/googleauth.ts.
+   * Headers to use for requests. Can be:
+   * - A headers object
+   * - A Promise that resolves to a headers object
+   * - A function that returns a headers object
+   * - A function that returns a Promise of a headers object
    */
-  googleAuthOptions?: VertexInit['googleAuthOptions'];
```
lgrammel, line 51: Removing the auth options will be breaking. Can we keep them for the node.js version?

.changeset/early-spiders-collect.md
lgrammel: Maybe add that it was a rewrite of the vertex provider. That is enough justification for a major imo.

lgrammel (approved):

packages/google-vertex/src/google-vertex-provider.ts
shaper: Oh yes, this was an oversight. We should be able to keep them, I'll look in the morning.

packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
shaper: I was concerned (and briefly consulted claude) that the `GoogleAuth` instance could be a little heavyweight and/or something one wants to only construct once if possible. Looking briefly at the implementation I see it has a bunch of logic to cache things and allow parallel requests, which semi-supports the thinking.

https://github.com/googleapis/google-auth-library-nodejs/blob/08978822e1b7b5961f0e355df51d738e012be392/src/auth/googleauth.ts#L164

It could be premature optimization but it is a small and localized enough area where it felt worthwhile. Open to thought of course.

packages/google-vertex/src/google-vertex-provider.ts
shaper: I had already preserved them, they are here: https://github.com/vercel/ai/blob/ac19c52d9ff5a6900919ac31fd45d25dd15592b9/packages/google-vertex/src/google-vertex-provider-node.ts#L17

The removal here is from the shared base provider, which makes sense to do.

packages/google-vertex/src/edge/google-vertex-provider-edge.ts
shaper: Yes, thanks. Added `googleCredentials?`.

packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
shaper: FYI (and related to the cached Auth instance above in a way), in final testing (pack/deploy iterating) I noted the `node` route of `next-google-vertex` of course doesn't work as I haven't made a creds file available. I started looking into making one available to my deployment just to test fully, or learn more as to how folks commonly deploy this in Node, and found this:

https://googleapis.dev/nodejs/google-auth-library/9.14.1/

Under the **Loading credentials from environment variables** section:

> Instead of loading credentials from a key file, you can also provide them using an environment variable and the GoogleAuth.fromJSON() method. This is particularly convenient for systems that deploy directly from source control (Heroku, App Engine, etc).

I don't think adding optional support for this needs to block shipping, since:
- it doesn't appear there's a standard env var they use to trigger this
- I have had in mind that most folks ship a creds file if they're deploying this way (or use one of the other methods). One of the methods in fact tries to just "do the right thing", see **Choosing the correct credential type automatically**. I briefly tested this but it required enabling another GCP API (DNS related) and I haven't revisited yet.
- folks can implement this themselves with a custom headers hook
- we can wait to hear if there are requests and respond quickly with this or whatever else is requested

PR #3911, opened 2024-11-27
feat (provider/azure): add support for azure managed identity.

### LINKED ISSUE #3124: Support azure authentication with managed identity
### Feature Description

Currently azure provider works with API keys, however, in our organization it is encouraged and enforced that we only use managed identity for auth against azure services (https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#authentication).

Implementing this type of auth would be a requirement for us to be able to use this library, would it be possible to add it?

### Use Case

Corporate environments with strict guidelines on security protocols to be followed.

### Additional context

i.e. currently we do something like this with the openai library:
```typescript
export const client = new OpenAIClient(
  env.OPENAI_API_BASE_URL,
  new DefaultAzureCredential(),
);
```

ShervK:
Fixes: #3124 

Similar to my last PR, #3463 but updated to v4 and cleaned up the code. Had to close the previous PR as it was too far behind and easier to start over than rebase.

- added option for a token provider to be sent
- added new tests
- added to the docs

Made it a fetch wrapper to grab a new token on new requests, similar to how [OpenAI package does it](https://github.com/openai/openai-node/blob/3f8634ed111782e3090a25d1d8640e050fb2c45b/src/index.ts#L610).

Please let me know what else I can add or change.

### DIFF
--- a/content/providers/01-ai-sdk-providers/02-azure.mdx
+++ b/content/providers/01-ai-sdk-providers/02-azure.mdx
@@ -57,6 +57,13 @@ You can use the following optional settings to customize the OpenAI provider ins
   API key that is being sent using the `api-key` header.
   It defaults to the `AZURE_API_KEY` environment variable.
 
+- **identityTokenProvider** _Promise&lt;string&gt;_
+
+  A function that returns a promise resolving to a token to be sent using the `Authorization` header.
+  This can be used with the `@azure/identity` package and `getBearerTokenProvider` function to authenticate with Azure Managed Identity.
+
+  If both `apiKey` and `identityTokenProvider` are provided, the `identityTokenProvider` takes precedence.
+
 - **apiVersion** _string_
 
   Sets a custom [api version](https://learn.microsoft.com/en-us/azure/ai-services/openai/api-version-deprecation).
--- a/packages/azure/src/azure-openai-provider.test.ts
+++ b/packages/azure/src/azure-openai-provider.test.ts
@@ -132,6 +132,67 @@ describe('chat', () => {
         'https://test-resource.openai.azure.com/openai/deployments/test-deployment/chat/completions?api-version=2024-10-01-preview',
       );
     });
+
+    it('should use bearer token from identityTokenProvider', async () => {
+      prepareJsonResponse();
+
+      const provider = createAzure({
+        resourceName: 'test-resource',
+        identityTokenProvider: async () => 'test-token',
+      });
+
+      await provider('test-deployment').doGenerate({
+        inputFormat: 'prompt',
+        mode: { type: 'regular' },
+        prompt: TEST_PROMPT,
+      });
+
+      const requestHeaders = await server.getRequestHeaders();
+      expect(requestHeaders).toStrictEqual({
+        'content-type': 'application/json',
+        authorization: 'Bearer test-token',
+      });
+    });
+
+    it('should throw error if identity token is invalid', async () => {
+      prepareJsonResponse();
+
+      const provider = createAzure({
+        resourceName: 'test-resource',
+        identityTokenProvider: async () => 1 as any,
+      });
+
+      await expect(
+        provider('test-deployment').doGenerate({
+          inputFormat: 'prompt',
+          mode: { type: 'regular' },
+          prompt: TEST_PROMPT,
+        }),
+      ).rejects.toThrow(
+        'Invalid Azure Managed Identity token format: token must be a non-empty string. Received: 1',
+      );
+    });
+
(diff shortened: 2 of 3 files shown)


lgrammel:
Thanks for the PR. I would prefer a different design for passing in auth that's more extensible.

cc @shaper ideally this would align with the design we end up with for google vertex so it's similar across providers

shaper:
> Thanks for the feedback, I didn't realize there were plans to implement token based auth for google vertex as well.
> 
> One idea that comes to mind is setting up a type-based approach for providers that have multiple auth methods, something like
> 
> ```ts
> type AuthenticationMethod = 
>   | { type: 'apiKey'; value: string }
>   | { type: 'bearer'; tokenProvider: () => Promise<string> }
>   | { type: 'custom'; getAuth: () => Promise<Record<string, string>> };
> ```
> 
> You could still have `apiKey` as an option and default to that if its provided, as to not cause breaking changes. Would this be more along the lines of what you're looking for?

Hi, thanks for the contribution, I will take a closer look at it!

A question: all of the above seem effectively implementable if we expose a way to optionally add to request headers. We have one today via `headers` in provider settings today, however, it isn't async, which can limit what one might need to do (e.g. fetching/refreshing an auth token with its own async ops needed). We could add a new one that allows async.

That has a nice feel from a simplicity of API surface perspective as it's one thing to think about, though it's also admittedly a little raw, as not all devs or docs want to think about or deal with talking about headers vs. the auth primitives/methods.

A concern could be any performance impact/overhead from calling the dev-provided headers population (do we do it once and cache it, or on each request, or catch auth errors and try refreshing by calling it again, ...). I expect we already have this auth-lifetime issue in some cases, and I wonder whether folks are having to catch auth timeout/failure and re-create the provider with updated auth in some cases.

It would be great to hear your thoughts on the above.

ShervK:
Hi @shaper 

I saw the PRs for the Resolvable utility type and Vertex AI auth changed get merged. I'll get to work on changing this PR to use those standards. 

I saw that the Vertex auth implementation used the `google-auth-library` package and added authentication options to the provider settings, is this something you want to see in the Azure provider? [`@azure/identity`](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/README.md) supports a lot of different [credential methods](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/README.md#credential-classes) so native support would cause the provider to balloon in size. 
An easier approach would be updating the docs with a section about using [`getBearerTokenProvider`](https://learn.microsoft.com/en-us/javascript/api/@azure/identity/?view=azure-node-latest#@azure-identity-getbearertokenprovider) function included in `@azure/identity` to retrieve the token using the user's preferred credential method and adding the token to the Resolvable Header.

I can support both implementations, let me know which sounds better.

shaper:
> Hi @shaper
> 
> I saw the PRs for the Resolvable utility type and Vertex AI auth changed get merged. I'll get to work on changing this PR to use those standards.
> 
> I saw that the Vertex auth implementation used the `google-auth-library` package and added authentication options to the provider settings, is this something you want to see in the Azure provider? [`@azure/identity`](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/README.md) supports a lot of different [credential methods](https://github.com/Azure/azure-sdk-for-js/blob/main/sdk/identity/identity/README.md#credential-classes) so native support would cause the provider to balloon in size. An easier approach would be updating the docs with a section about using [`getBearerTokenProvider`](https://learn.microsoft.com/en-us/javascript/api/@azure/identity/?view=azure-node-latest#@azure-identity-getbearertokenprovider) function included in `@azure/identity` to retrieve the token using the user's preferred credential method and adding the token to the Resolvable Header.
> 
> I can support both implementations, let me know which sounds better.

The easier approach and keeping the provider lighter weight sounds better to me. If you put together a draft PR we'd be happy to look at it and discuss, that would be great, thanks!

There were two reasons we kept the google-auth-library for vertex:
- backwards-compatibility (folks were already using it) which is something we care about a lot
- DX, it's really nice having the one-line import of the default vertex provider instance "just work"

The last point is a down side of the simpler approach. I think you'll end up with something that's like at least 4 lines or so wherever someone creates a provider instance.

Let's see what it looks like when you try it? My only other idea, if we want to pursue the goal of lighter-weight providers, is to have a separate contrib-type package where small hooks like this can live and folks could leverage that way, and we can push heavyweight deps to those other more specific packages that depend on the provider. I am not saying we should do this yet, it's just the only thing that comes to mind. Maybe you'll find something else syntactic-sugar-wise etc.

PR #3956, opened 2024-12-03
feat (provider-utils): Add detectRuntime utility routine.

shaper:


lgrammel (approved):

lgrammel:
We'll release it together with the resolvable, so I guess it won't need a changeset for correct versioning, but it might still be nice to inform our users about the new functioality.

shaper:
Not using currently.

PR #3958, opened 2024-12-03
feat (provider-utils): Add resolvable type and utility routine.

shaper:
Breakout from #3908 to land first.

### DIFF
--- a/packages/provider-utils/src/resolve.ts
+++ b/packages/provider-utils/src/resolve.ts
@@ -11,14 +11,9 @@ export type Resolvable<T> =
 export async function resolve<T>(value: Resolvable<T>): Promise<T> {
   // If it's a function, call it to get the value/promise
   if (typeof value === 'function') {
-    const result = (value as Function)();
-    return Promise.resolve(result);
+    value = (value as Function)();
   }
 
   // Otherwise just resolve whatever we got (value or promise)
-  return Promise.resolve(value);
+  return Promise.resolve(value as T);
 }
-
-// Helper for common case of resolving headers
-export type ResolvableHeaders = Resolvable<Record<string, string | undefined>>;
-export const resolveHeaders = (headers: ResolvableHeaders) => resolve(headers);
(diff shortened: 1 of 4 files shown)


packages/provider-utils/src/resolve.ts
```diff
@@ -0,0 +1,24 @@
...
+export async function resolve<T>(value: Resolvable<T>): Promise<T> {
+  // If it's a function, call it to get the value/promise
+  if (typeof value === 'function') {
+    const result = (value as Function)();
+    return Promise.resolve(result);
+  }
+
+  // Otherwise just resolve whatever we got (value or promise)
+  return Promise.resolve(value);
+}
```
lgrammel, line 20: to safe a few download bytes, I would do the following (more common pattern that I've seen):

```
if (typeof value === 'function') { 
  value = (value as Function)();
}

return Promise.resolve(value);
```

packages/provider-utils/src/resolve.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// Helper for common case of resolving headers
+export type ResolvableHeaders = Resolvable<Record<string, string | undefined>>;
+export const resolveHeaders = (headers: ResolvableHeaders) => resolve(headers);
```
lgrammel, line 24: I don't know if `ResolvableHeaders` is needed. Do we have a `Headers` type somewhere (warning: might conflict with JS built-ins)

`resolveHeaders` is unnecessary, type inference will automatically do the right thing with just resolve.

lgrammel: (if `ResolvableHeaders` is needed, consider moving it to a separate file)

lgrammel (approved):

packages/provider-utils/src/resolve.ts
lgrammel: Thought about it more: `ResolvableHeaders` has little value over `Resolvable<Headers>`. We should not include it imo (and instead think about a `Headers` type).

packages/provider-utils/src/resolve.ts
shaper: Done, I had to add `value as T` to the `value` in the `return`.

PR #3961, opened 2024-12-03
chore (docs): add order to root docs sections

jeremyphilemon:
Added order for root sections, this is needed for llms.txt generation so they're in the right order

Screenshot 2024-12-03 at 1 00 06 PM

Screenshot 2024-12-03 at 12 59 23 PM


nicoalbanese (approved):

lgrammel:
still relevant? this is prone to get out-of-date quickly

PR #3978, opened 2024-12-05
feat (provider/google-vertex): Add support for Anthropic models.

shaper:
- [x] decide on sub-module names under `@ai-sdk/google-vertex`: `.../anthropic-edge` and `.../anthropic-node` vs `.../anthropic` and `.../anthropic/edge`
- [x] add/review provider package tests
- [x] add/review e2e tests
- [x] rework provider settings as needed to reflect validation results for cache control, pdf support, etc.
- [x] consider default provider name -- `googleVertexAnthropic` or `vertexAnthropic`
- [x] update docs


lgrammel:
Looks great!

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -179,21 +183,32 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV1 {
...
   }
 
+  private buildRequestUrl(isStreaming: boolean): string {
+    return (
+      this.config.buildRequestUrl?.(this.config.baseURL, isStreaming) ??
+      `${this.config.baseURL}/messages`
+    );
+  }
+
+  private transformArgs(args: Record<string, any>): Record<string, any> {
```
lgrammel, line 199: `transformArgs` had me wondering at first what it does. Based on the usage below, I'm wondering if `transformBody` or `transformRequestBody` would be a more intuitive name for both this method and the config property?

packages/google-vertex/src/anthropic/edge/google-vertex-anthropic-provider-edge.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function createVertexAnthropic(
+  options: GoogleVertexAnthropicProviderSettings = {},
+): GoogleVertexAnthropicProvider {
+  return createVertexAnthropicOriginal({
+    ...options,
+    headers:
+      options.headers ??
+      (async () => ({
+        'anthropic-version': 'vertex-2023-10-16',
```
lgrammel, line 31: is the anthropic version needed in both headers and body?

lgrammel (approved):

packages/google-vertex/src/anthropic/edge/google-vertex-anthropic-provider-edge.ts
shaper: Good find, it appears we need `anthropic_version: 'vertex-2023-10-16',` in the request body but we don't need this in the headers. Removed here and in node variant.

packages/anthropic/src/anthropic-messages-language-model.ts
shaper: Agreed, renamed to `transformRequestBody`.

PR #3979, opened 2024-12-05
chore (providers/google-vertex): Remove unref'd base default provider.

shaper:


lgrammel:
I don't understand this change. Isn't the default base provider needed?

lgrammel (approved):

PR #4005, opened 2024-12-06
chore (docs):  Update providers-and-models page

minpeter:
Updated the https://sdk.vercel.ai/docs/foundations/providers-and-models page

- Added missing items. (friendliai, togehterai, cloudflare workers ai, crosshatch, mixedbread, voyage)
- The text on the right of the community provider has been unified as the package name.
- Sorted the order (same as document order)

It would also be nice if the `<CommunityModelCards />` component were updated.


minpeter:
If CommunityModelCards are updated, please use the image above (FriendliAI)

https://github.com/user-attachments/assets/b6aca117-7577-4456-b172-9a5ab574e8b3

lgrammel (approved):

lgrammel:
Thank you! I'm trying to minimize the use of cards bc we then need to maintain files in 2 repositories (images).

Issue #3553, opened 2024-11-08
Support "Grounding with Google Search" in Google provider


wong2:
### Feature Description

The Gemini API has a new feature where you can ground results with Google search which reduces hallucinations and provides references to content:
https://ai.google.dev/gemini-api/docs/grounding

adolphnov:
I have tried vertex is valid, it will report an error in Google AI Studio, which does not support this parameter. And note that when used in the vertex, no tools can be included, otherwise an error will also be reported.
I hope this is helpful to you.

lgrammel:
@shaper do we support this with the latest provider versions?

shaper:
The above changes were just published to the package versions in https://github.com/vercel/ai/pull/4081

This enables search grounding support in the Google Generative AI `google` package, similar to what we've had in Google Vertex.

We plan to add support for customizing the dynamic retrieval config soon. At that time I would consider this issue complete. If there is interest in the enterprise data support mentioned above please file that as a separate issue.

wottpal:
Hey there, dope addition! Just played with it a bit and got it to work pretty well. 

Some observations though:

1. It doesn't seem to work with `generateObject`, only `generateText`. At least I was not able to force the model to actually DO search when generating object output. Is that related to mutual exclusiveness with tool calls, you're mentioning above, @shaper? If yes, it should probably be mentioned in the docs.
2. Being able to set the dynamic retrieval threshold is crucial for a lot of use cases IMO. Because in my observations the model sometimes tends to NOT google even though I nudge it towards it in my prompt **a lot**. Would be amazing to see this soon.


Lmk whether I should open separate issues for those if you want to close this one.

shaper:
> Hey there, dope addition! Just played with it a bit and got it to work pretty well.
> 
> Some observations though:
> 
> 1. It doesn't seem to work with `generateObject`, only `generateText`. At least I was not able to force the model to actually DO search when generating object output. Is that related to mutual exclusiveness with tool calls, you're mentioning above, @shaper? If yes, it should probably be mentioned in the docs.

Yes, probably -- I will check this and update docs if needed.

> 2. Being able to set the dynamic retrieval threshold is crucial for a lot of use cases IMO. Because in my observations the model sometimes tends to NOT google even though I nudge it towards it in my prompt **a lot**. Would be amazing to see this soon.

It is on my radar, will try to get to it soon.

> Lmk whether I should open separate issues for those if you want to close this one.

No need, let's keep this one for now.

[...]

PR #4048, opened 2024-12-10
feat (provider/google): Include optional response grounding metadata.

shaper:
Mentioned in #3553

### DIFF
--- a/.changeset/dry-apples-suffer.md
+++ b/.changeset/dry-apples-suffer.md
@@ -1,0 +1,6 @@
+---
+'@ai-sdk/google-vertex': patch
+'@ai-sdk/google': patch
+---
+
+feat (provider/google): Include optional response grounding metadata.
--- a/packages/google/src/google-generative-ai-prompt.ts
+++ b/packages/google/src/google-generative-ai-prompt.ts
@@ -38,3 +38,21 @@ export type GoogleGenerativeAIContentPart =
         fileUri: string;
       };
     };
+
+export interface GoogleGenerativeAIGroundingMetadata {
+  webSearchQueries?: string[] | null;
+  searchEntryPoint?: { renderedContent: string } | null;
+  groundingSupports?: Array<{
+    segment: {
+      text?: string | null;
+      startIndex?: number | null;
+      endIndex?: number | null;
+    };
+    groundingChunkIndices: number[];
+    confidenceScores: number[];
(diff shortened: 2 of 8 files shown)


.changeset/long-dogs-kick.md
shaper, line 1: This change manifests as new functionality only in the `google-vertex` package, but there are no code changes in that module. Should we patch-bump that module so that it's clear changelog-wise?

.changeset/dry-apples-suffer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/google-vertex': patch
```
shaper, line 2: Note though there are no code changes in the `google-vertex` package, this functionality is only user facing in that provider, so I've patch-bumped that module as well for visibility.

packages/google/src/google-generative-ai-prompt.ts
```diff
@@ -38,3 +38,21 @@ export type GoogleGenerativeAIContentPart =
         fileUri: string;
       };
     };
+
+export interface GoogleGenerativeAIGroundingMetadata {
+  webSearchQueries?: string[] | null;
```
shaper, line 43: I followed the pattern of `nullish` everywhere instead of `optional` in the schema change, after https://github.com/vercel/ai/pull/4038. It makes this interface definition more unsightly as we have to allow `null` everywhere. I didn't actually validate whether `optional` would suffice for some of these. I can do that if you think warranted.

packages/google/src/google-generative-ai-prompt.ts
```diff
@@ -38,3 +38,21 @@ export type GoogleGenerativeAIContentPart =
...
+  groundingSupports?: Array<{
+    segment: {
+      text?: string | null;
+      startIndex?: number | null;
+      endIndex?: number | null;
+    };
+    groundingChunkIndices: number[];
+    confidenceScores: number[];
+  }> | null;
+}
```
lgrammel, line 54: can this be derived from groundMetadataSchema via typeof z.infer ?

lgrammel (approved):

packages/google/src/google-generative-ai-prompt.ts
shaper: Yes, much better, thanks.

PR #4049, opened 2024-12-10
feat (provider/google-vertex): Add support for baseURL in API calls.

palmm:
Added ability for the base URL of API calls to Vertex to be overridden.

✅ docs updated.
✅ tests updated and added.

I saw [packages/google-vertex/src/anthropic/google-vertex-anthropic-provider.ts](https://github.com/vercel/ai/blob/main/packages/google-vertex/src/anthropic/google-vertex-anthropic-provider.ts) does `baseURL`, `location`, and `project` a little differently. That said, I tried to keep the change really tight so I didn't move to the newer style.


lgrammel:
Thanks. This is new functionality that would ideally have 1-2 tests that test only the new baseUrl functionality.

@shaper can you review and check that this does not break existing functionality?

.changeset/warm-days-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/google-vertex': minor
```
lgrammel, line 2: please use patch

lgrammel: (we do not fully follow semver)

shaper:
> Thanks. This is new functionality that would ideally have 1-2 tests that test only the new baseUrl functionality.
> 
> @shaper can you review and check that this does not break existing functionality?

@lgrammel it looks good to me. I patched the PR, ran several example tests and the e2e google-vertex and google-vertex-anthropic tests.

lgrammel (approved):

PR #4056, opened 2024-12-10
feat (ai/core): add experimental generateImage function

### LINKED ISSUE #2524: image generation
### Discussed in https://github.com/vercel/ai/discussions/2520

<div type='discussions-op-text'>

<sup>Originally posted by **Und3rf10w** August  1, 2024</sup>
What are the thoughts on adding a `imageGenerationModel` as a property to a `Provider`?

I came up with these types as an example, but this is tied directly to Dalle. 

```typescript
type ImageGenerationModelV1CallSettings = {
  size: string
  n: number
  quality?: 'natural' | 'vivid'
  /**
  Response format. The output can either be url or b64json. Default is url.

     */
  responseFormat?: {
    type: 'url'
  } | {
    type: 'b64json'
  }
  /**
  Abort signal for cancelling the operation.
     */
  abortSignal?: AbortSignal
  /**
Additional HTTP headers to be sent with the request.
Only applicable for HTTP-based providers.
   */
  headers?: Record<string, string | undefined>
}

type ImageModelV1CallOptions = ImageGenerationModelV1CallSettings & {
  inputFormat: 'prompt'
  prompt: LanguageModelV1Prompt
}

type ImageGenerationModelV1FinishReason = 'completed' | 'content-filter' | 'error' | 'other' | 'unknown'

type DallePromiseLike = PromiseLike<{
  created: Date
  expires?: Date
  id?: string
  result?: {
    data: Array<{
      url: string
    }>
  }
  status?: string
  data?: Array<{
    url: string
    revised_prompt: string
  }>
  error?: {
    code: string
    message: string
  }
  lastActionDateTime: Date
}>

export interface ImageGenerationModelV1 {
    /**
  The language model must specify which language model interface
  version it implements. This will allow us to evolve the language
  model interface and retain backwards compatibility. The different
  implementation versions can be handled as a discriminated union
  on our side.
     */
  readonly specificationVersion: 'v1';
  /**
Name of the provider for logging purposes.
   */
  readonly provider: string;
  /**
Provider-specific model ID for logging purposes.
   */
  readonly modelId: string;
  /**
  Default object generation mode that should be used with this model when
  no mode is specified. Should be the mode with the best results for this
  model. `undefined` can be returned if object generation is not supported.
  
  This is needed to generate the best objects possible w/o requiring the
  user to explicitly specify the object generation mode.
     */
  readonly defaultObjectGenerationMode: 'json' | 'tool' | undefined
  doGenerate: (options: ImageModelV1CallOptions) => DallePromiseLike & {
    /**
     * Finish reason.
     */
    finishReason: ImageGenerationModelV1FinishReason
    /**
  Raw prompt and setting information for observability provider integration.
  */
    rawCall: {
    /**
  Raw prompt after expansion and conversion to the format that the
  provider uses to send the information to their API.
      */
      rawPrompt: unknown
      /**
  Raw settings that are used for the API call. Includes provider-specific
  settings.
      */
      rawSettings: Record<string, unknown>
    }
    /**
  Optional raw response information for debugging purposes.
  */
    rawResponse?: {
    /**
  Response headers.
      */
      headers?: Record<string, string>
    }
  }
}
```

Maybe the call settings and `DallePromiseLike` would be tied to like an `OpenAIImageGenerationModel` instead that implements from `ImageModelV1`?

I don't actually have any idea on how to properly handle `doStream` cases though.

Ideally, since most of the providers taht are available in the `ProviderRegistry` also expose image generation models (e.g. `Dalle` for OpenAI, `Stable diffusion` for vertex/bedrock), we would also be able to add those and manage those using the registry so we only have to track models in one place and they can all be self contained in the registry.
</div>

lgrammel:
Fixes #2524 

## Tasks

- [x] implementation
   - [x] spec
   - [x] openai image generation model
   - [x] openai image generation model test
   - [x] generateImage
   - [x] generateImage test
   - [x] size
   - [x] providerOptions
   - [x] experimental
   - [x] uint8array conversion
   - [x] image
   - [x] redesign result
- [x] examples
   - [x] basic example
   - [x] next.js example
- [x] documentation
   - [x] generateImage reference
   - [x] core guide
   - [x] openai model
- [x] changeset

### DIFF
--- a/content/docs/03-ai-sdk-core/35-image-generation.mdx
+++ b/content/docs/03-ai-sdk-core/35-image-generation.mdx
@@ -1,0 +1,99 @@
+---
+title: Image Generation
+description: Learn how to generate images with the AI SDK.
+---
+
+# Image Generation
+
+<Note type="warning">Image generation is an experimental feature.</Note>
+
+The AI SDK provides the [`generateImage`](/docs/reference/ai-sdk-core/generate-image)
+function to generate images based on a given prompt using an image model.
+
+```tsx
+import { experimental_generateImage as generateImage } from 'ai';
+import { openai } from '@ai-sdk/openai';
+
+const { image } = await generateImage({
+  model: openai.image('dall-e-3'),
+  prompt: 'Santa Claus driving a Cadillac',
+  size: '1024x1024',
+});
+```
+
+You can access the image data using the `base64` or `uint8Array` properties:
+
+```tsx
+const base64 = image.base64; // base64 image data
+const uint8Array = image.uint8Array; // Uint8Array image data
+```
+
+### Generating Multiple Images
+
+`generateImage` also supports generating multiple images at once:
+
+```tsx highlight={"4"}
+const { images } = await generateImage({
+  model: openai.image('dall-e-3'),
+  prompt: 'Santa Claus driving a Cadillac',
+  n: 4, // number of images to generate
+});
+```
+
+### Provider-specific Settings
+
+Image models often have provider- or even model-specific settings.
+You can pass such settings to the `generateImage` function
+using the `providerOptions` parameter. The options for the provider
+(`openai` in the example below) become request body properties.
+
+```tsx highlight={"5-7"}
+const { image } = await generateImage({
+  model: openai.image('dall-e-3'),
+  prompt: 'Santa Claus driving a Cadillac',
+  size: '1024x1024',
+  providerOptions: {
+    openai: { style: 'vivid', quality: 'hd' },
+  },
+});
+```
+
+### Abort Signals and Timeouts
+
+`generateImage` accepts an optional `abortSignal` parameter of
+type [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)
+that you can use to abort the image generation process or set a timeout.
+
+```ts highlight={"7"}
+import { openai } from '@ai-sdk/openai';
+import { experimental_generateImage as generateImage } from 'ai';
+
+const { image } = await generateImage({
+  model: openai.image('dall-e-3'),
+  prompt: 'Santa Claus driving a Cadillac',
+  abortSignal: AbortSignal.timeout(1000), // Abort after 1 second
+});
+```
+
+### Custom Headers
+
+`generateImage` accepts an optional `headers` parameter of type `Record<string, string>`
+that you can use to add custom headers to the image generation request.
+
+```ts highlight={"7"}
+import { openai } from '@ai-sdk/openai';
+import { experimental_generateImage as generateImage } from 'ai';
+
+const { image } = await generateImage({
+  model: openai.image('dall-e-3'),
+  value: 'sunny day at the beach',
+  headers: { 'X-Custom-Header': 'custom-value' },
+});
+```
+
+## Image Models
+
+| Provider                                                  | Model      | Supported Sizes                 |
+| --------------------------------------------------------- | ---------- | ------------------------------- |
+| [OpenAI](/providers/ai-sdk-providers/openai#image-models) | `dall-e-3` | 1024x1024, 1792x1024, 1024x1792 |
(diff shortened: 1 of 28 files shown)


packages/ai/core/generate-image/generate-image.ts
```diff
@@ -0,0 +1,113 @@
+import { ImageModelV1, JSONValue } from '@ai-sdk/provider';
+import { convertBase64ToUint8Array } from '@ai-sdk/provider-utils';
+import { prepareRetries } from '../prompt/prepare-retries';
+import { GeneratedImage, GenerateImageResult } from './generate-image-result';
+
+/**
+Embed a value using an embedding model. The type of the value is defined by the embedding model.
```
jeremyphilemon, line 7: has old content, need to update

jeremyphilemon (approved):

content/docs/03-ai-sdk-core/35-image-generation.mdx
```diff
@@ -0,0 +1,99 @@
...
+  abortSignal: AbortSignal.timeout(1000), // Abort after 1 second
+});
+```
+
+### Custom Headers
+
+`generateImage` accepts an optional `headers` parameter of type `Record<string, string>`
+that you can use to add custom headers to the image generation request.
+
+```ts highlight={"7"}
```
shaper, line 83: This looks like it might be intended to highlight `8`

content/docs/07-reference/01-ai-sdk-core/10-generate-image.mdx
```diff
@@ -0,0 +1,140 @@
...
+    },
+    {
+      name: 'abortSignal',
+      type: 'AbortSignal',
+      isOptional: true,
+      description: 'An optional abort signal to cancel the call.',
+    },
+    {
+      name: 'headers',
+      type: 'Record<string, string>',
```
shaper, line 84: PR-wide: should this be the set of types Resolvable allows? If so, update docs/code in some manner.

content/docs/07-reference/01-ai-sdk-core/10-generate-image.mdx
shaper, line 1: I wondered while reading:
- what format(s) are produced
- how does token usage work

packages/provider/src/image-model/v1/image-model-v1.ts
```diff
@@ -0,0 +1,77 @@
...
+    abortSignal?: AbortSignal;
+
+    /**
+  Additional HTTP headers to be sent with the request.
+  Only applicable for HTTP-based providers.
+     */
+    headers?: Record<string, string | undefined>;
+  }): PromiseLike<{
+    /**
+Generated images as base64 encoded strings.
```
shaper, line 73: Don't we want some representation of what the binary format is, or to mention briefly in a doc comment if there is only one currently and no field specifier is needed?

content/docs/07-reference/01-ai-sdk-core/10-generate-image.mdx
lgrammel: - no token usage
- format is currently open, we can just add on-the-fly mimetype detection if people ask

content/docs/07-reference/01-ai-sdk-core/10-generate-image.mdx
lgrammel: Good catch, when we introduce resolvable more broadly we need to update this

content/docs/03-ai-sdk-core/35-image-generation.mdx
lgrammel: 7 should be correct (headers) but i'll double check when deployed

packages/provider/src/image-model/v1/image-model-v1.ts
lgrammel: if you mean the image type, we can auto-detect this later using magic bytes if needed

content/docs/03-ai-sdk-core/35-image-generation.mdx
lgrammel: yes 7 shows as expected

PR #4093, opened 2024-12-14
feat (provider/xai): Add grok-2 models, use openai-compatible base impl.

shaper:


packages/xai/src/xai-error.ts
shaper, line 1: Ah, I need to fix this error format delta.

lgrammel (approved):

lgrammel:
Changeset could be useful to receive any error reports earlier.

shaper:
Now incorporates model ids from #4095 and updates docs to reflect capabilities.

PR #4117, opened 2024-12-17
feat (ai/core): add experimental output setting to streamText

lgrammel:
## Background

`streamObject` does not support tool calling or data streams. To generate structured output with tool calling, I'm adding an `output` setting to `streamText`. Longer-term goal is to move structured object generation to `generateText` / `streamText` to simplify the codebase and remove duplication.

## Tasks

- [x] implementation
   - [x] correct text stream (only parsable parts)
   - [x] partial stream
   - [x] throw error when no output defined
   - [x] dedicated error object
   - [x] extract transform
   - [x] extract type
- [x] example
- [x] documentation
   - [x] code
   - [x] guide
   - [x] error page
   - [x] reference
- [x] changeset


examples/ai-core/src/stream-text/openai-output-object.ts
```diff
@@ -0,0 +1,45 @@
...
+    //   schema: z.object({
+    //     elements: z.array(
+    //       z.object({
+    //         location: z.string(),
+    //         temperature: z.number(),
+    //         touristAttraction: z.string(),
+    //       }),
+    //     ),
+    //   }),
+    // }),
```
lgrammel, line 32: need to fix

jeremyphilemon (approved):

PR #4121, opened 2024-12-17
feat (docs): Use new grok-2 model in xai example code.

shaper:


nicoalbanese (approved):

nicoalbanese:
Only thing - I don't think this needs a patch cc @lgrammel

nicoalbanese:
Totally right - misread the file path!

Issue #4088, opened 2024-12-13
Imagen support for Google vertex


Godrules500:
### Feature Description

I wanted to put a request in here for imagen support in google vertex.
https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/imagen-api

### Use Cases

Generating Images - Generating images from a text prompt.
Editing Images - Use text prompts and a mask to modify your image's appearance
Upload Image - After you generate watermarked images, you can verify the digital watermark of the novel images.

adolphnov:
Hope for the ability to generate images within custom-defined vertex instances. ❤️

![image](https://github.com/user-attachments/assets/7a342cca-263f-4f07-8c2c-a0cdad1fb098)

shaper:
> Hope for the ability to generate images within custom-defined vertex instances. ❤️ @shaper

Hi there, can you give me an example of what you are trying to do in code or with more detail?

The interface above looks like it may be from an older version of the library as the latest [includes an `image` method](https://github.com/vercel/ai/blob/main/packages/google-vertex/src/google-vertex-provider.ts#L41) but I realize I may be missing what you are asking for.

adolphnov:
> The interface above looks like it may be from an older version of the library as the latest includes an image method but I realize I may be missing what you are asking for.

I'm trying to use createVertex to generate a custom instance, invoking the image, rather than relying on the module's default exported vertex instance. My vertex parameters are dynamically sourced from another location, not directly defined as global variables. I've checked the embed method, but it doesn't appear to provide a solution.

adolphnov:
That's odd, I updated to the latest version just yesterday. Let me try the new package.😊

adolphnov:
> > Hope for the ability to generate images within custom-defined vertex instances. ❤️ @shaper
> 
> 
> 
> Hi there, can you give me an example of what you are trying to do in code or with more detail?
> 
> 
> 
> The interface above looks like it may be from an older version of the library as the latest [includes an `image` method](https://github.com/vercel/ai/blob/main/packages/google-vertex/src/google-vertex-provider.ts#L41) but I realize I may be missing what you are asking for.

It's working perfectly now after updating to the latest version 12! Thank you so much for your work.

shaper:
> > > Hope for the ability to generate images within custom-defined vertex instances. ❤️ @shaper
> > 
> > 
> > Hi there, can you give me an example of what you are trying to do in code or with more detail?
[...]

PR #4124, opened 2024-12-18
feat (provider/google-vertex): Add imagen support.

shaper:
Part of #4088

### DIFF
--- a/examples/ai-core/src/e2e/google-vertex.test.ts
+++ b/examples/ai-core/src/e2e/google-vertex.test.ts
@@ -4,6 +4,7 @@ import { vertex as vertexEdge } from '@ai-sdk/google-vertex/edge';
 import { vertex as vertexNode } from '@ai-sdk/google-vertex';
 import { z } from 'zod';
 import {
+  detectImageMimeType,
   generateText,
   generateObject,
   streamText,
@@ -461,7 +462,11 @@ describe.each(Object.values(RUNTIME_VARIANTS))(
         const { image } = await generateImage({
           model,
           prompt: 'A burrito launched through a tunnel',
-          size: '1024x1024',
+          providerOptions: {
+            vertex: {
+              aspectRatio: '3:4',
+            },
+          },
         });
 
         // Verify we got a Uint8Array back
@@ -471,10 +476,9 @@ describe.each(Object.values(RUNTIME_VARIANTS))(
         expect(image.uint8Array.length).toBeGreaterThan(10 * 1024);
         expect(image.uint8Array.length).toBeLessThan(10 * 1024 * 1024);
 
-        // Verify PNG format by checking magic numbers
-        const pngSignature = [137, 80, 78, 71, 13, 10, 26, 10];
-        const actualSignature = Array.from(image.uint8Array.slice(0, 8));
-        expect(actualSignature).toEqual(pngSignature);
+        // Verify PNG format
+        const mimeType = detectImageMimeType(image.uint8Array);
+        expect(mimeType).toBe('image/png');
 
         // Create a temporary buffer to verify image dimensions
         const tempBuffer = Buffer.from(image.uint8Array);
@@ -483,8 +487,9 @@ describe.each(Object.values(RUNTIME_VARIANTS))(
         const width = tempBuffer.readUInt32BE(16);
         const height = tempBuffer.readUInt32BE(20);
 
-        expect(width).toBe(1024);
-        expect(height).toBe(1024);
+        // https://cloud.google.com/vertex-ai/generative-ai/docs/image/generate-images#performance-limits
+        expect(width).toBe(896);
+        expect(height).toBe(1280);
       });
     });
   },
--- a/examples/ai-core/src/e2e/google-vertex.test.ts
+++ b/examples/ai-core/src/e2e/google-vertex.test.ts
@@ -4,7 +4,6 @@ import { vertex as vertexEdge } from '@ai-sdk/google-vertex/edge';
 import { vertex as vertexNode } from '@ai-sdk/google-vertex';
 import { z } from 'zod';
 import {
-  detectImageMimeType,
   generateText,
   generateObject,
   streamText,
@@ -17,6 +16,28 @@ import fs from 'fs';
 import { GoogleGenerativeAIProviderMetadata } from '@ai-sdk/google';
 
 const LONG_TEST_MILLIS = 20000;
+
+const mimeTypeSignatures = [
+  { mimeType: 'image/gif' as const, bytes: [0x47, 0x49, 0x46] },
+  { mimeType: 'image/png' as const, bytes: [0x89, 0x50, 0x4e, 0x47] },
+  { mimeType: 'image/jpeg' as const, bytes: [0xff, 0xd8] },
+  { mimeType: 'image/webp' as const, bytes: [0x52, 0x49, 0x46, 0x46] },
+];
+
+function detectImageMimeType(
+  image: Uint8Array,
+): 'image/jpeg' | 'image/png' | 'image/gif' | 'image/webp' | undefined {
+  for (const { bytes, mimeType } of mimeTypeSignatures) {
+    if (
+      image.length >= bytes.length &&
+      bytes.every((byte, index) => image[index] === byte)
+    ) {
+      return mimeType;
+    }
+  }
+
+  return undefined;
+}
 
 // Model variants to test against
 const MODEL_VARIANTS = {
--- a/examples/ai-core/src/e2e/google-vertex.test.ts
+++ b/examples/ai-core/src/e2e/google-vertex.test.ts
@@ -10,11 +10,12 @@ import {
   streamObject,
   embed,
   embedMany,
+  experimental_generateImage as generateImage,
 } from 'ai';
 import fs from 'fs';
 import { GoogleGenerativeAIProviderMetadata } from '@ai-sdk/google';
 
-const LONG_TEST_MILLIS = 10000;
+const LONG_TEST_MILLIS = 20000;
 
 // Model variants to test against
 const MODEL_VARIANTS = {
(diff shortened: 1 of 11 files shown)


packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+  private getAspectRatio(width: number, height: number): string {
+    // Map common dimensions to Imagen's supported aspect ratios
+    if (width === height) return '1:1';
+    if (width === 896 && height === 1280) return '3:4';
+    if (width === 1280 && height === 896) return '4:3';
+    if (width === 768 && height === 1408) return '9:16';
+    if (width === 1408 && height === 768) return '16:9';
+
+    // Default to 1:1 if no match
+    return '1:1';
```
lgrammel, line 82: can we do some clever math here instead of relying on fixed values (ask llm)? or are these related to the imagen output sizes?

examples/ai-core/src/e2e/google-vertex.test.ts
```diff
@@ -26,6 +27,7 @@ const MODEL_VARIANTS = {
     // 'gemini-1.0-pro-001',
   ],
   embedding: ['textembedding-gecko', 'textembedding-gecko-multilingual'],
+  image: ['imagen-3.0-generate-001', 'imagen-3.0-fast-generate-001'],
```
lgrammel, line 30: is it possible to run specific e2e tests easily? as in all image tests or a specific image test?

examples/ai-core/src/e2e/google-vertex.test.ts
```diff
@@ -453,5 +455,38 @@ describe.each(Object.values(RUNTIME_VARIANTS))(
...
+
+        // Verify we got a Uint8Array back
+        expect(image.uint8Array).toBeInstanceOf(Uint8Array);
+
+        // Check the file size is reasonable (at least 10KB, less than 10MB)
+        expect(image.uint8Array.length).toBeGreaterThan(10 * 1024);
+        expect(image.uint8Array.length).toBeLessThan(10 * 1024 * 1024);
+
+        // Verify PNG format by checking magic numbers
+        const pngSignature = [137, 80, 78, 71, 13, 10, 26, 10];
```
lgrammel, line 476: we have a mime type detection helper for this but I'm not sure it's exposed

packages/google-vertex/src/google-vertex-image-model.ts
lgrammel: also, is the aspect ratio required? otherwise we could rely on provider options so users can specify it

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+    prompt: string;
+    n: number;
+    size: `${number}x${number}` | undefined;
+    providerOptions: Record<string, Record<string, JSONValue>>;
+    abortSignal?: AbortSignal;
+    headers?: Record<string, string>;
+  }): Promise<{ images: string[] }> {
+    const [width, height] = (options.size ?? '1024x1024')
+      .split('x')
+      .map(Number);
```
lgrammel, line 38: are these only needed for aspect ratio?

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const response = await (this.options.fetch ?? fetch)(
```
lgrammel, line 40: should use postJsonToApi similar to `openai-image-model`

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+            sampleCount: options.n,
+            aspectRatio: this.getAspectRatio(width, height),
+            ...options.providerOptions,
+          },
+        }),
+      },
+    );
+
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
```
lgrammel, line 62: error parsing?

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+    private options: GoogleVertexImageModelOptions,
+  ) {}
+
+  async doGenerate(options: {
+    prompt: string;
+    n: number;
+    size: `${number}x${number}` | undefined;
+    providerOptions: Record<string, Record<string, JSONValue>>;
+    abortSignal?: AbortSignal;
+    headers?: Record<string, string>;
```
lgrammel, line 34:
```
  }: Parameters<ImageModelV1['doGenerate']>[0]): Promise<
    Awaited<ReturnType<ImageModelV1['doGenerate']>>
 ```

File: packages/google-vertex/src/google-vertex-image-model.ts [hunk 8]
```diff
@@ -0,0 +1,84 @@
...
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`);
+    }
+
+    const data = await response.json();
+    return {
+      images: data.predictions.map(
+        (p: { bytesBase64Encoded: string }) => p.bytesBase64Encoded,
+      ),
+    };
```
lgrammel, line 70: What about output validation?

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -0,0 +1,84 @@
...
+export type GoogleVertexImageModelId =
+  | 'imagen-3.0-generate-001'
+  | 'imagen-3.0-fast-generate-001';
+
+interface GoogleVertexImageModelOptions {
+  provider: string;
+  baseURL: string;
+  headers?: Resolvable<Record<string, string | undefined>>;
+  fetch?: typeof fetch;
+}
```
lgrammel, line 13: wonder if we can extract generic model options for google (across different kinds of models)

packages/google-vertex/src/google-vertex-image-model.ts
shaper: https://cloud.google.com/vertex-ai/generative-ai/docs/image/generate-images#aspect-ratio
https://cloud.google.com/vertex-ai/generative-ai/docs/image/generate-images#performance-limits

It's an optional parameter and the default is 1:1. It appears the only way to specify the desired image size is via specifying an aspect ratio, which in turn maps to an output resolution width/height per the above links.

Changing it to use a provider option and to throw if a user tries to specify an explicit size. We could allow size if it's one of the supported sizes, but that's also more maintenance work for us.

examples/ai-core/src/e2e/google-vertex.test.ts
shaper: Yes, you can use `-t <filter string>` so `-t image` to match `it` or `describe` strings. You can use `-t 'Chat Model'` or `-t 'Image Model'` for example to run all test cases under the respective `describe` subset. I added a note about this in the [README](https://github.com/vercel/ai/tree/main/examples/ai-core#end-to-end-provider-integration-tests) although I didn't fully elaborate as much as above.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: Yes, removed now, the LLM was basically reverse engineering the aspect ratio from user specified size, and defaulting to the 1:1 imagen defaults to. That's one way to do it, but it's odd and could lead to user surprise since they might not get the size they specified.

examples/ai-core/src/e2e/google-vertex.test.ts
shaper: I don't know if you can use `-t` in some manner to specify you only want to run tests for a particular model, though. I have been commenting/uncommenting them in the list when I want to do that.

examples/ai-core/src/e2e/google-vertex.test.ts
shaper: Exposed it for use, much nicer.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: I think we can and it's a good idea but would like to do it as a separate change -- will note for follow-up.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: Done.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: Done.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: Reworked as part of postToJsonApi edit.

packages/google-vertex/src/google-vertex-image-model.ts
shaper: Fixed via postToJsonApi rework.

packages/ai/core/index.ts
```diff
@@ -11,3 +11,4 @@ export * from './registry';
 export * from './tool';
 export * from './types';
 export { cosineSimilarity } from './util/cosine-similarity';
+export { detectImageMimeType } from './util/detect-image-mimetype';
```
lgrammel, line 14: Torn about exposing the utility. We'd need to document / support it. We could expose it under `/test` or `/internal`?

lgrammel (approved):

packages/ai/core/index.ts
lgrammel: or even copy the utility for now until we can think more about how to best do it. I'm leaning towards including under `ai/test`

shaper: Will copy it for now.

PR #4154, opened 2024-12-19
feat(provider/openai): add support for developer role in chat message conversion

Rajaniraiyn:
- Introduced `useDeveloperRole` option in `convertToOpenAIChatMessages` to allow conversion of system messages to developer role based on user preference.
- Updated tests to verify behavior for system messages with `useDeveloperRole` set to true, false, and undefined.
- Enhanced type definitions to include `ChatCompletionDeveloperMessage` for better clarity and usage in the OpenAI chat prompt structure.

reference: https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages

### DIFF
--- a/packages/openai/src/convert-to-openai-chat-messages.test.ts
+++ b/packages/openai/src/convert-to-openai-chat-messages.test.ts
@@ -308,3 +308,52 @@ describe('tool calls', () => {
     ]);
   });
 });
+
+describe('system messages', () => {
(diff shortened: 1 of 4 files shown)


lgrammel (changes requested):
Not needed. Afaik OpenAI automatically converts system to developer messages on their end.

See https://x.com/sandersted/status/1869116813658980389

PR #4158, opened 2024-12-19
feat (provider/deepinfra): Add DeepInfra provider.

shaper:


examples/ai-core/src/e2e/deepinfra.test.ts
```diff
@@ -0,0 +1,379 @@
...
+    'Qwen/Qwen2-7B-Instruct',
+  ],
+  embedding: [
+    'BAAI/bge-base-en-v1.5',
+    'intfloat/e5-base-v2',
+    'sentence-transformers/all-mpnet-base-v2',
+  ],
+} as const;
+
+describe('DeepInfra E2E Tests', () => {
```
lgrammel, line 51: There is a lot of duplication in the e2e tests. Is there a way to abstract them out and have a single function that is called for each provider?

lgrammel (approved):

examples/ai-core/src/e2e/deepinfra.test.ts
shaper: Yes, good call. We hit a tipping point with enough of these recently for this to be warranted, I'll do it as a separate PR soon.

PR #4162, opened 2024-12-20
feat (provider/friendli): Add FriendliAI provider.

minpeter:


lgrammel (changes requested):
This PR means moving the provider into the AI SDK and putting the maintenance on the AI SDK team. `@ai-sdk/*` packages are reserved to premier providers that we decide to maintain. Considering the current download stats (21 weekly downloads) for the provider I will not do that.

I love supporting new providers, but considering download numbers being a community provider is the best path here atm.

minpeter:
Couldn't we proceed under the assumption that I will take responsibility for and maintain this package?

lgrammel:
No - I've been burned by taking such a path in the past multiple times. It is not what we want to do with the SDK and is precisely why we support 3rd party community providers and add them to our docs. We hope that the docs integration already gives you a platform for marketing.

minpeter:
I'll create a separate PR adding some exports to use `@ai-sdk/openai-compatible` in my package implementation / Thanks :)

PR #4166, opened 2024-12-20
feat (provider/openai-compatible) Additional exports for detailed customization

minpeter:
When creating an external provider using provider/openai-compatible, additional exports are required for detailed customization.

### DIFF
--- a/packages/openai-compatible/src/index.ts
+++ b/packages/openai-compatible/src/index.ts
@@ -2,6 +2,10 @@ export { createOpenAICompatible } from './openai-compatible-provider';
 export { OpenAICompatibleChatLanguageModel } from './openai-compatible-chat-language-model';
 export { OpenAICompatibleCompletionLanguageModel } from './openai-compatible-completion-language-model';
 export { OpenAICompatibleEmbeddingModel } from './openai-compatible-embedding-model';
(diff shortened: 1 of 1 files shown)


lgrammel:
I'm on board with exporting these. However, they should be exported under `/internal` because they are for providers that extend openai-compatible, and not for end users.

@shaper wdyt? would an `/internal` dist make sense here and can you help out if so?

shaper:
> I'm on board with exporting these. However, they should be exported under `/internal` because they are for providers that extend openai-compatible, and not for end users.
> 
> @shaper wdyt? would an `/internal` dist make sense here and can you help out if so?

Sure, @minpeter let me propose an updated PR, I will include you on it as a reviewer. I plan to send it today and will reference this PR as source.

shaper:
It seems I can't add you as a reviewer but please follow https://github.com/vercel/ai/pull/4179 and comment with any questions. It should be the same as what's here except you'll import from `@ai-sdk/openai-compatible/internal`.

shaper:
@minpeter thanks for your patience, the change in #4179 includes your commit for tracking purposes and it should be live shortly in v0.0.13 per https://github.com/vercel/ai/pull/4201

PR #4181, opened 2024-12-21
feat (docs): separate stop handler from loading state

dragos-cojocaru:
Separated `stop` handler explanation from `isLoading` explanation.
Created a separate subheading for `stop` handler function returned by `useObject` hook.


lgrammel (approved):
Thanks!

PR #4182, opened 2024-12-21
chore (config): Use ts-library.json tsconfig for no-UI libs.

shaper:


tools/tsconfig/ts-library.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "https://json.schemastore.org/tsconfig",
+    "display": "TypeScript Library",
+    "extends": "./base.json",
+    "compilerOptions": {
+        "lib": [
+            "dom",
```
shaper, line 7: We remove the `jsx` support, however, through the absence of:

```
    "jsx": "react-jsx",
```

If we remove `dom` we lose some things we're using. There may be Node alternative imports for them, but it seems worth doing as a separate step.

lgrammel (approved):

PR #4185, opened 2024-12-21
feat (examples): Use shared base e2e feature test suite.

shaper:


packages/togetherai/src/index.ts
```diff
@@ -1,5 +1,8 @@
+import { OpenAICompatibleErrorData } from '@ai-sdk/openai-compatible';
+
 export { createTogetherAI, togetherai } from './togetherai-provider';
 export type {
   TogetherAIProvider,
   TogetherAIProviderSettings,
 } from './togetherai-provider';
+export type { OpenAICompatibleErrorData as TogetherAIErrorData };
```
shaper, line 8: I just realized I can probably combine line 1 and this line. Will edit before merge if so.

examples/ai-core/src/e2e/fireworks.test.ts
```diff
@@ -1,398 +1,33 @@
...
+      'accounts/fireworks/models/llama-v3p3-70b-instruct',
+      'accounts/fireworks/models/mixtral-8x7b-instruct',
+      'accounts/fireworks/models/qwen2p5-72b-instruct',
+    ],
+    completion: [
+      'accounts/fireworks/models/llama-v3-8b-instruct',
+      'accounts/fireworks/models/llama-v2-34b-code',
+    ],
+    embedding: ['nomic-ai/nomic-embed-text-v1.5'],
+  },
```
lgrammel, line 24: what are your thoughts about inlining the model construction? it could lead to some duplication here (ok imo), but it would allow us to test wrapped (middleware) models and models with settings as well:

```
models: {
    chat: [
      fireworks('accounts/fireworks/models/firefunction-v2'),
      ...
```

lgrammel: (imo this might be a better level of abstraction for test call interface, making it more flexible, but I may be missing some details)

examples/ai-core/src/e2e/feature-test-suite.ts
```diff
@@ -0,0 +1,450 @@
...
+  chat?: string[];
+  completion?: string[];
+  embedding?: string[];
+}
+
+export interface TestSuiteOptions {
+  name: string;
+  createChatModelFn: (modelId: string) => LanguageModelV1;
+  createCompletionModelFn: (modelId: string) => LanguageModelV1;
+  createEmbeddingModelFn: (modelId: string) => EmbeddingModelV1<string>;
```
lgrammel, line 25: See comment on interface abstraction. An alternative would be passing in `LanguageModelV1` objects which would lead to some duplication in the construction side, but significantly more flexibility.

examples/ai-core/src/e2e/feature-test-suite.ts
```diff
@@ -0,0 +1,450 @@
...
+  embedMany,
+  APICallError,
+} from 'ai';
+import fs from 'fs';
+import { describe, it, expect, vi } from 'vitest';
+import type { LanguageModelV1, EmbeddingModelV1 } from '@ai-sdk/provider';
+
+export interface ModelVariants {
+  chat?: string[];
+  completion?: string[];
```
lgrammel, line 17: No distinction needed if we pass `LanguageModelV1` objects

lgrammel (approved):

examples/ai-core/src/e2e/togetherai.test.ts
```diff
@@ -1,369 +1,38 @@
...
+  togetherai as provider,
+  TogetherAIErrorData,
+} from '@ai-sdk/togetherai';
+import { APICallError } from 'ai';
+import { createFeatureTestSuite } from './feature-test-suite';
+
+createFeatureTestSuite({
+  name: 'TogetherAI',
+  models: {
+    invalidModel: provider.chatModel('no-such-model'),
```
lgrammel, line 13: How would this work? Wouldn't this need to be deferred?

`invalidModel: () => provider.chatModel('no-such-model')`

PR #4196, opened 2024-12-24
feat (ai/core): Proposal for generate code API with tools

rajatsandeepsen:
# Proposal for generateCode API

Imagine the LLM can write Javascript code with custom logic within the help of limited tools inside a safe eval( )

### Usage
Let me show you an example on how to build an AI powered banking app

```ts
import { z } from "zod"
import { tool, experimental_generateCode as generateCode } from "ai"
import { model } from "./model"
```

Define a set of tools
```ts
let balance = 30 // simulation of a DB
const history = [
    { amount: 10, from: "Bob" },
    { amount: 20, to: "Alice" },
]

const tools = ({
    getBalance: tool({
        description: "get balance of the user",
        parameters: z.object({}),
        execute: async () => {
            return balance
        },
        returns: z.number()
    }),
    sentMoney: tool({
        description: "send money to the user",
        parameters: z.object({ amount: z.number(), receiver: z.string() }),
        execute: async ({ amount, receiver }) => {
            if (balance < amount) {
                throw new Error("Insufficient balance")
            }
            balance -= amount

            history.push({ amount, to: receiver })
        },
        returns: z.void()
    }),
    getHistory: tool({
        description: "get history of transactions",
        parameters: z.unknown(),
        execute: async () => {
            return history
        },
        returns: z.array(
            z.object({ amount: z.number(), to: z.string().optional(), from: z.string().optional() }))
    })
})
```

Fun part begins here
```ts
const result = await generateCode({
    model,
    system: "You are a banking app",
    tools: tools,
    prompt: "Get history and find amount i got from Bob, then send that amount to Bob. Then again get history and balance",
})

console.log("Code:", result.code)
console.log("Schema:", result.schema)
console.log("Output:", await result.execute())
```

This is the output after execution `await result.execute( )`
```js
{
  balance: 20,
  history: [
    {
      amount: 20,
      to: "Alice"
    }, {
      amount: 10,
      from: "Bob"
    },{
      amount: 10,
      to "Bob"
    }
  ]
}
```

This is the code written by LLM `result.code`
```js
let history = this.getHistory({});
let amountGotFromBob = 0;
for (let transaction of history) {
    if (transaction?.from === 'Bob') {
        amountGotFromBob += transaction.amount;
    }
}
if (amountGotFromBob > 0) {
    this.sentMoney({ amount: amountGotFromBob, receiver: 'Bob' });
}
let balance = this.getBalance({});
return { balance, history };
```

This is the JSON schema written by LLM `result.schema` (useful for generative UI)
```js
{
  $schema: "http://json-schema.org/draft-07/schema#",
  type: "object",
  properties: {
    balance: {
      type: "number",
    },
    history: {
      type: "array",
      items: {
        type: "object",
        properties: {
          amount: {
            type: "number",
          },
          to: {
            type: "string",
          },
          from: {
            type: "string",
          },
        },
        required: [ "amount" ],
      },
    },
  },
  required: [ "balance", "history" ]
}
```

Instead of multi-step toolResults, now LLM can write logic along with tools provided by developer. Also the LLM can't execute malicious code because of a safety simple technique i implemented in this.

```ts
export const createFunction = (tools: Record<string, CoreTool>, code: string) => {
  const data = Object.entries(tools).reduce((acc, [key, value]) => ({ ...acc, [key]: value.execute }), {})

  return async () => await new Function(main(code)).apply(data, [])
}
const main = (code:string) => `const main = async () => {\n${code}\n}\nreturn main()`

const generateCode = ({ ... }) => {

	// some generateText logic

	return {
		code, schema,
		execute: createFunction(tools, code)
	}
}
```
Therefor this LLM is restricted to only invoke list of functions we provided to them.

The `generateCode( )` API is a powerful wrapper around `generateText( )`


lgrammel:
This is pretty cool. However, there is one big issue: it includes a fairly large prompt. The AI SDK tries to not include any prompts whenever possible, because they end up being provider and model dependent (there is some minimal prompting in JSON generation but that is all). Do you see ways to do this without prompting? Or could there be alternative approaches that minimize it?

packages/ai/core/tool/tool.ts
```diff
@@ -56,21 +63,16 @@ If not provided, the tool will not be executed automatically.
 @args is the input of the tool call.
 @options.abortSignal is a signal that can be used to abort the tool call.
    */
-  execute?: (
+  execute: (
```
lgrammel, line 66: Removing the optionality here will break important functionality, namely tools without execute.

packages/ai/core/tool/tool.ts
```diff
@@ -94,28 +96,28 @@ The arguments for configuring the tool. Must match the expected arguments define
 Helper function for inferring the execute args of a tool.
  */
 // Note: special type inference is needed for the execute function args to make sure they are inferred correctly.
-export function tool<PARAMETERS extends Parameters, RESULT>(
+export function tool<PARAMETERS extends Parameters, RESULT extends Parameters>(
```
lgrammel, line 99: `RESULT extends Parameters` seems strange.

packages/ai/core/tool/tool.ts
rajatsandeepsen: I got two alternatives idea for this issue

- Create an isolated tool in different name (eg: `etool` specifically for `generateCode`

- add one more option to `toolChoice: "none" | "auto" | "required" | "code"` (this `code` helps to make `execute` strictly required

Give me a feedback, I'll rework on this

packages/ai/core/tool/tool.ts
rajatsandeepsen: `generateCode` requires an returns `zodSchema` so LLM can understand whats the output of each tools.

It'll help LLM to create variables to save the tool-calling-result and pass it to next tool. 

Reason why i used `Parameters` type is to strictly specify execute tool according to its `returns: zodSchema`

I could rework this as

```ts
type Returns = Parameters // same properties

export function tool<PARAMETERS extends Parameters, RESULT, RETURNS extends Returns>(
tool: CoreTool<PARAMETERS, RETURNS> & {
    execute: (
      args: inferParameters<PARAMETERS>,
      options: ToolExecutionOptions,
    ) => PromiseLike<inferParameters<RETURNS>>;
  }
)
```

rajatsandeepsen:
> This is pretty cool. However, there is one big issue: it includes a fairly large prompt. The AI SDK tries to not include any prompts whenever possible, because they end up being provider and model dependent (there is some minimal prompting in JSON generation but that is all). Do you see ways to do this without prompting? Or could there be alternative approaches that minimize it?

This idea of tools inside `generateCode` is not widely adopted in openAI API specs. 

Everyone uses JSON schema for specifying tools to LLM (as Industry standards). But i used `zod-to-ts` for three reasons

1. it converts zodSchema to typescript definition
2. takes less tokens for defining same functionality compared to JSON schema
3. anyway we're not passing these functions to tools part of OpenAI API and not returning a list of tool name in JSON object

So complex parameters & returns type can be specified + it can document each param usage like

```ts
const writeLetter = tool({
	description: "write letter to a person",
	parameters: z.object({
	 name: z.string().describe("Name of the person")
	}),
	returns: z.string(),
	execute: ({ name }) => `Hello ${name}`
})
```
After zod-to-ts
```ts
const writeLetter = (params: {
	// Name of the person
	name: stirng
}) :string => {
	// some logic
}
```

Prompt i using is to ensure LLM can writes better and secure code. Some of these prompts are important.
image

PR #4210, opened 2024-12-26
feat (provider/openai): support for prediction tokens usage in the response

ggallon:
Hi,

I've added support for prediction tokens in the response, continuing from #3475.

-  the `accepted_prediction_tokens` and `rejected_prediction_tokens` from the usage object to the providerMetadata.

And updated examples in the docs.


lgrammel:
Continued in #4252

PR #4316, opened 2025-01-08
feat (provider/fireworks): Support add'l image models.

shaper:


packages/fireworks/README.md
```diff
@@ -1,34 +1,69 @@
 # AI SDK - Fireworks Provider
 
-The **[Fireworks provider](https://sdk.vercel.ai/providers/ai-sdk-providers/fireworks)** for the [AI SDK](https://sdk.vercel.ai/docs) contains language model support for the [Fireworks](https://fireworks.ai) platform.
+The **[Fireworks provider](https://sdk.vercel.ai/providers/ai-sdk-providers/fireworks)** for the [AI SDK](https://sdk.vercel.ai/docs) contains language model and image modelsupport for the [Fireworks](https://fireworks.ai) platform.
```
lgrammel, line 3: smol typo: missing space

packages/fireworks/src/fireworks-image-model.ts
```diff
@@ -7,21 +7,109 @@ import {
...
+  modelId: FireworksImageModelId,
+): string {
+  const workflowsUrl = `${baseUrl}/workflows/${modelId}/text_to_image`;
+  switch (modelToBackendConfig[modelId]?.urlFormat) {
+    case 'image_generation':
+      return `${baseUrl}/image_generation/${modelId}`;
+    case 'stability':
+      // For stability models the model id is passed in the form data of the
+      // request rather than as a part of the url.
+      return `https://api.stability.ai/v2beta/stable-image/generate/sd3`;
```
lgrammel, line 100: Woah - so we go to the stability API directly? This should not be in the fireworks provider then, it should be a separate stability provider. how are fireworks and stability related?

lgrammel (changes requested):
If we go directly towards the stability API, wouldn't it make more sense to have a separate stability provider? this complex provider seems directionally wrong but I don't know how fireworks and stability are related

packages/fireworks/src/fireworks-image-model.ts
lgrammel: Nevermind realized that this is the model name and they do a pass through

lgrammel (approved):

shaper:
#4327 will live on in posterity with the fireworks/stable-diffusion-3 key, model, and backend logic. Will revise this PR for initial landing without it.

PR #4332, opened 2025-01-09
fix: add tool call id to ToolExecutionError

danamajid:
The `ToolExecutionError` is missing `toolCallId` making tool error events hard to correlate back to which tool call failed exactly. This PR simply adds that to the error object.


danamajid:
@lgrammel would this be something you can review 🙏 ?

lgrammel:
Thank you this looks good! will make a few tweaks & land.

lgrammel:
Continued in https://github.com/vercel/ai/pull/4342

PR #4338, opened 2025-01-10
feat (ai/core): handle zero vector (edge case) in cosine similarity

patelvivekdev:
Modified `cosineSimilarity` to return 0 if either input vector has a magnitude of zero (i.e., is a zero vector or empty).

```ts
const vector1 = [1, 2, 3];
const vector2 = [0, 0, 0];

const result = cosineSimilarity(vector1, vector2);

console.log(result); // Now returns 0 (previously NaN)

const vectorC: number[] = [];
const vectorD: number[] = [];

const result2 = cosineSimilarity(vectorC, vectorD);

console.log(result2); // Now returns 0 (previously NaN)
```


lgrammel:
When would this happen in practice?

lgrammel:
I agree that an error is the way to go here. That being said, it's a backwards compatibility breaking change - system might now error where they did not do so before. That means we have to only enable it with a separate flag and/or wait until 5.0 to make the change. Can you update the PR as follows:

a) add a 3rd optional options arg `{ throwErrorForEmptyVectors: boolean = false }`
b) leave a TODO note that it should be default true in the next major version
c) if enabled, check the arrays for emptiness in the beginning (which would lead to magnitude 0 - are there other cases where this can happen?) and throw an InvalidArgumentError if so
d) update cosine similarity docs mdx
e) add a changeset

packages/ai/core/util/cosine-similarity.ts
```diff
@@ -1,23 +1,60 @@
...
+  vector1: number[],
+  vector2: number[],
+  options?: {
+    throwErrorForEmptyVectors?: boolean;
+  },
+) {
+  // TODO: In the next major version, change the default value of throwErrorForEmptyVectors to true
+  const { throwErrorForEmptyVectors = false } = options ?? {};
+
+  if (throwErrorForEmptyVectors) {
```
lgrammel, line 27: Considering the vector length check below, this can be simplified by pushing it after the check from line 44. After check from line 44, you only need to check vector1 length (bc vector2 length is the same). this will save a few bytes

packages/ai/core/util/cosine-similarity.ts
```diff
@@ -1,23 +1,60 @@
...
-export function cosineSimilarity(vector1: number[], vector2: number[]) {
+export function cosineSimilarity(
+  vector1: number[],
+  vector2: number[],
+  options?: {
+    throwErrorForEmptyVectors?: boolean;
+  },
+) {
+  // TODO: In the next major version, change the default value of throwErrorForEmptyVectors to true
+  const { throwErrorForEmptyVectors = false } = options ?? {};
```
lgrammel, line 25: can be moved up into options as default (parameter default)

packages/ai/core/util/cosine-similarity.ts
```diff
@@ -1,23 +1,60 @@
...
     throw new Error(
       `Vectors must have the same length (vector1: ${vector1.length} elements, vector2: ${vector2.length} elements)`,
     );
   }
 
-  return (
-    dotProduct(vector1, vector2) / (magnitude(vector1) * magnitude(vector2))
-  );
+  const mag1 = magnitude(vector1);
+  const mag2 = magnitude(vector2);
```
lgrammel, line 51: please use full words e.g. `magnitude1`

packages/ai/core/util/cosine-similarity.ts
```diff
@@ -1,23 +1,60 @@
...
+
+  if (mag1 === 0 || mag2 === 0) {
+    return 0;
+  }
```
lgrammel, line 55: is this still needed if the length is checked above?

patelvivekdev: In the zero vector case, we need this.

lgrammel: wouldn't throwing (if opted in) be sufficient? this part breaks the behavior (tho it could be considered a fix)

patelvivekdev: Adding this will not break the existing system, as this returns the number 0.

If we remove this condition, the below example will be NaN, even if it was opted in for that flag.

``` ts
const vector1 = [1, 2, 3];
const vector2 = [0, 0, 0];

const result = cosineSimilarity(vector1, vector2, { throwErrorForEmptyVectors: true });

// result will be NaN
```

lgrammel: understood. considering that `dotProduct(vector1, vector2) / (magnitude(vector1) * magnitude(vector2))` will go towards +/- infinity for values that are close to 0, should it return +/- infinity (depending on the prefix of the dot product) instead to be mathematically reasonable?

patelvivekdev: The result of `dotProduct(vector1, vector2) / (magnitude(vector1) * magnitude(vector2))` will always be between -1 and +1, even for very small values.
 
 We handle the case where either vector has a magnitude of 0 by returning 0

lgrammel (approved):

lgrammel (approved):

Issue #4333, opened 2025-01-09
Deepseek provider isn't exposing caching values in usage


matannahmani:
### Description

run any request you will see usage is missing the fields that deepseek mention (https://api-docs.deepseek.com/news/news0802)

Monitoring Cache Hits
Two new fields in the API response's usage section help users monitor cache performance:

prompt_cache_hit_tokens：Number of tokens from the input that were served from the cache ($0.014 per million tokens)
prompt_cache_miss_tokens: Number of tokens from the input that were not served from the cache ($0.14 per million tokens)

this is most likely related to open-ai compatible package which isn't spreading the usage record.

shaper:
Hi, thanks for the report, looking into support for this.

shaper:
This improvement was landed in #4397

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests exposing additional provider-specific usage metrics, which is an expansion of the provider’s supported API data rather than a failure of documented existing behavior.

### Evidence
- DeepSeek returns `prompt_cache_hit_tokens` and `prompt_cache_miss_tokens`, but the provider did not expose them in SDK usage data.
- The maintainer described the work as adding “support” and later said the “improvement” landed in #4397.

PR #4359, opened 2025-01-11
fix: expose deepseek cache usage #4333

kongmoumou:
Currently, the openai compatible provider doesn't support customizing chunkSchema and adding providerMetadata extracting logic, so it's hard to use some provider-specific data, as #4333 points out.

This PR enhances `OpenAICompatibleChatConfig`, which makes it easier to add provider-specific logic.


matannahmani:
BIG thanks for the quick fix!

shaper:
This work has landed in #4397 including your commit for credit. Thank you for the contribution!

Issue #3554, opened 2024-11-08
Support citations in Perplexity provider


wong2:
### Feature Description

Perplexity just announced that "Citations" are now enabled for everyone via the API: https://docs.perplexity.ai/changelog/changelog#citations-public-release-and-increased-default-rate-limits

They are returned alongside `choices` in the SSE response.

RostyslavManko:
@lgrammel, any updates on this?

sylviangth:
@lgrammel Hi any update please?

egoist:
For a temporary workaround, you can apply this diff using `pnpm patch`:

```diff
diff --git a/dist/index.mjs b/dist/index.mjs
index 66cc5faa76025ed6a5ac56881a31fc0b60913cb6..908c893569d1e8dffacd3f6c724f89196d59c927 100644
--- a/dist/index.mjs
+++ b/dist/index.mjs
@@ -659,7 +659,8 @@ var OpenAIChatLanguageModel = class {
             if (delta.content != null) {
               controller.enqueue({
                 type: "text-delta",
-                textDelta: delta.content
+                textDelta: delta.content,
+                citations: value.citations
               });
             }
             const mappedLogprobs = mapOpenAIChatLogProbsOutput(
@@ -840,6 +841,7 @@ var openaiChatChunkSchema = z2.union([
     id: z2.string().nullish(),
     created: z2.number().nullish(),
     model: z2.string().nullish(),
+    citations: z2.any().nullish(),
     choices: z2.array(
       z2.object({
         delta: z2.object({
```

Then you can get citations in fullStream

ybelakov:
can we move this forward please?

ybelakov:
https://github.com/vercel/ai/pull/4366

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests adding support for a newly available Perplexity provider capability rather than fixing behavior that contradicts an existing SDK contract.

### Evidence
- The title and description explicitly request support for Perplexity's newly released citations API output.
- The description says citations are returned alongside `choices` in the SSE response, but does not claim the SDK previously promised or documented this behavior.
- A comment provides a temporary patch to expose citations through `fullStream`, reinforcing that this is an API capability addition.

PR #4366, opened 2025-01-13
Add citations for perplexity

ybelakov:
Feature Description
Perplexity just announced that "Citations" are now enabled for everyone via the API: https://docs.perplexity.ai/changelog/changelog#citations-public-release-and-increased-default-rate-limits

They are returned alongside choices in the SSE response.

https://github.com/vercel/ai/issues/3554

### DIFF
--- a/packages/openai/src/openai-chat-language-model.ts
+++ b/packages/openai/src/openai-chat-language-model.ts
@@ -626,6 +626,7 @@ export class OpenAIChatLanguageModel implements LanguageModelV1 {
               controller.enqueue({
                 type: 'text-delta',
                 textDelta: delta.content,
+                citations: value.citations
               });
             }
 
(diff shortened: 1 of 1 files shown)


shaper:
Thank you for the PR!

We are working on an approach that will handle this sort of need more generally via #4397 so that we don't have to keep modifying the schema or otherwise putting provider-specific details like this in the internals of the openai-compatible provider. You can see the gist of how we intend to use it for DeepSeek cached token usage, for example, in that PR.

When we land the PR above I will work with you to update this to use it -- coming shortly.

ybelakov:
> Thank you for the PR!
> 
> We are working on an approach that will handle this sort of need more generally via #4397 so that we don't have to keep modifying the schema or otherwise putting provider-specific details like this in the internals of the openai-compatible provider. You can see the gist of how we intend to use it for DeepSeek cached token usage, for example, in that PR.
> 
> When we land the PR above I will work with you to update this to use it -- coming shortly.

amazing! Thanks!

shaper:
Continuing this work in #4501

PR #4395, opened 2025-01-14
feat (ai/core): throw NoImageGeneratedError from generateImage when no predictions are returned

shaper:
- `ai`: added `ImageModelMetadataResponse` and `NoImageGeneratedError`
- note we do not have an id for image model requests
- newly tracks the start-response-timestamp for reporting in the error


packages/google-vertex/package.json
```diff
@@ -48,6 +48,7 @@
     }
   },
   "dependencies": {
+    "ai": "4.1.0",
```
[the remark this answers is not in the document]
lgrammel: I propose a different solution below. If we want errors to be used in providers, we would add the errors to the `ai/provider` package.

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -84,10 +88,22 @@ export class GoogleVertexImageModel implements ImageModelV1 {
...
+    }
+
     return {
-      images: response.predictions.map(
-        (p: { bytesBase64Encoded: string }) => p.bytesBase64Encoded,
-      ),
+      images:
+        response.predictions.map(
+          (p: { bytesBase64Encoded: string }) => p.bytesBase64Encoded,
+        ) ?? [],
```
lgrammel, line 106: I propose moving the error throwing logic into `generateImage`. It has the following advantages:

- works for all image providers without any extra provider code
- no dependency on the `ai` package

packages/google-vertex/src/google-vertex-image-model.ts
```diff
@@ -96,5 +112,5 @@ export class GoogleVertexImageModel implements ImageModelV1 {
 // minimal version of the schema, focussed on what is needed for the implementation
 // this approach limits breakages when the API changes and increases efficiency
 const vertexImageResponseSchema = z.object({
-  predictions: z.array(z.object({ bytesBase64Encoded: z.string() })),
+  predictions: z.array(z.object({ bytesBase64Encoded: z.string() })).optional(),
```
lgrammel, line 115: prefer `nullish` over `optional`. I have seen API change and suddenly return `null` instead of nothing, causing breakages with `optional`. `nullish` covers both.

packages/ai/core/generate-image/generate-image-result.ts
```diff
@@ -19,6 +20,11 @@ The images that were generated.
 Warnings for the call, e.g. unsupported settings.
      */
   readonly warnings: Array<ImageGenerationWarning>;
+
+  /**
+Response metadata from the provider. There may be multiple responses if we made multiple calls to the model.
+   */
+  readonly responses: Array<ImageModelResponseMetadata | undefined>;
```
shaper, line 27: We could make this mandatory -- current image models all return their response headers. This follows what I see in `language-model-v1.ts` where `response` and `rawResponse` are optional.

packages/ai/core/generate-image/generate-image.ts
```diff
@@ -133,30 +145,43 @@ Only applicable for HTTP-based providers.
...
-
+  const responses: Array<ImageModelResponseMetadata | undefined> = [];
+  const currentDate = _internal?.currentDate?.() ?? new Date();
   for (const result of results) {
     images.push(
       ...result.images.map(image => new DefaultGeneratedImage({ image })),
     );
     warnings.push(...result.warnings);
+    responses.push({
+      timestamp: currentDate,
```
shaper, line 159: Checkpointing my understanding and design choice for potential discussion:
- `timestamp` - intended to be, if provider produces such, the timestamp associated with the inference request. Fallbacks could be (1) our own intra-image-model-impl pre-post-to-api timestamp, or (2) an intra-generate-image timestamp. I've gone with (2) for all cases here to reduce the amount of code we need in image models, and as I'm not aware of the providers producing what we want. It would at least be more accurate to track timestamp in each model just before we post, and ship that back in `result.response.timestamp`. Is it worth the extra complexity?
- `modelId`: for language models we have them populate and return this, however here we know what it is, so it felt redundant and again saved on model-impl-side complexity to just use what we had here.

packages/ai/core/generate-image/generate-image-result.ts
lgrammel: lets' make it mandatory - it is only optional in language models for backwards compat

packages/ai/core/generate-image/generate-image.ts
lgrammel: 1) yes - if we push it firmly into the provider, then they should ideally get it from the response if provided, and if not, use the current time (up to the provider at which point). Don't think we need the extra complexity for now, this is not for measuring anything (we'll use OTel for that).

2) `modelId` is useful when aliases are resolved provider side. E.g. `gpt-4o` has different versions, and the model id should tell us which one has been used. Such a mechanism might also exist for some image models.

packages/ai/core/generate-image/generate-image.ts
```diff
@@ -133,30 +145,43 @@ Only applicable for HTTP-based providers.
...
   }
 
-  return new DefaultGenerateImageResult({ images, warnings });
+  return new DefaultGenerateImageResult({ images, warnings, responses });
 }
 
 class DefaultGenerateImageResult implements GenerateImageResult {
   readonly images: Array<GeneratedImage>;
   readonly warnings: Array<ImageGenerationWarning>;
+  readonly responses: Array<ImageModelResponseMetadata | undefined>;
```
lgrammel, line 175: ideally we remove the `undefined` (if we make the information mandatory in the spec)

packages/ai/errors/no-image-generated-error.ts
```diff
@@ -0,0 +1,39 @@
...
+  readonly responses: Array<ImageModelResponseMetadata | undefined> | undefined;
+
+  constructor({
+    message = 'No image generated.',
+    cause,
+    responses,
+  }: {
+    message?: string;
+    cause?: Error;
+    responses: Array<ImageModelResponseMetadata | undefined>;
```
lgrammel, line 29: make optional?

packages/provider/src/image-model/v1/image-model-v1.ts
```diff
@@ -47,5 +47,15 @@ be returned as binary data.
...
+
+    /**
+Optional response information for telemetry and debugging purposes.
+     */
+    response?: {
+      /**
+Response headers.
+      */
+      headers?: Record<string, string>;
+    };
```
lgrammel, line 59: add modelId + timestamp?

packages/ai/core/generate-image/generate-image.ts
shaper: Moved these into the providers.

lgrammel (approved):

PR #4397, opened 2025-01-15
feat (provider): add metadata extraction mechanism to openai-compatible providers

shaper:

PR #4429, opened 2025-01-17
feat (provider-utils): include raw value in json parse results

shaper:
Use case in #4397

### DIFF
--- a/packages/provider-utils/src/parse-json.test.ts
+++ b/packages/provider-utils/src/parse-json.test.ts
@@ -3,188 +3,186 @@ import { parseJSON, safeParseJSON, isParsableJson } from './parse-json';
 import { z } from 'zod';
 import { JSONParseError, TypeValidationError } from '@ai-sdk/provider';
 
-describe('parse-json', () => {
-  describe('parseJSON', () => {
-    it('should parse basic JSON without schema', () => {
-      const result = parseJSON({ text: '{"foo": "bar"}' });
-      expect(result).toEqual({ foo: 'bar' });
-    });
+describe('parseJSON', () => {
+  it('should parse basic JSON without schema', () => {
+    const result = parseJSON({ text: '{"foo": "bar"}' });
+    expect(result).toEqual({ foo: 'bar' });
+  });
 
-    it('should parse JSON with schema validation', () => {
-      const schema = z.object({ foo: z.string() });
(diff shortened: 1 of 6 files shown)


packages/provider-utils/src/parse-json.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      const schema = z.object({
+        user: z.object({
+          id: z.string().transform(val => parseInt(val, 10)),
+          name: z.string(),
+        }),
+      });
+
+      const result = safeParseJSON({
+        text: '{"user": {"id": "123", "name": "John"}}',
+        schema: schema as any,
```
lgrammel, line 88: why is `any` needed?

packages/provider-utils/src/parse-json.test.ts
```diff
@@ -0,0 +1,190 @@
+import { describe, it, expect } from 'vitest';
+import { parseJSON, safeParseJSON, isParsableJson } from './parse-json';
+import { z } from 'zod';
+import { JSONParseError, TypeValidationError } from '@ai-sdk/provider';
+
+describe('parse-json', () => {
```
lgrammel, line 6: maybe remove the above describe level

lgrammel (approved):

packages/provider-utils/src/parse-json.test.ts
shaper: Done

packages/provider-utils/src/parse-json.test.ts
shaper: Something about the `transform` call typing produces:

```
id: z.string().transform(val => parseInt(val, 10)), // => leads to "Type 'string' not assignable to type 'number'" error
```

for this test case `z.coerce.number()` achieves the same thing and doesn't produce the type issue, so I'll change it to that (we did similar in another test case).

lgrammel: that's a bug and it has to do with the zod schemas

lgrammel: I think I know what it is; can take a look next week (zod has input and output schemas)

lgrammel: let's leave any for now

PR #4437, opened 2025-01-18
feat (provider-utils): handle differing zod input/output schemas in json parsing/validation

shaper:


lgrammel:
Probably not needed; I misunderstood the original issue. What would we gain by implementing this?

PR #4460, opened 2025-01-21
fix (ai/react): cache `addToolResult`

jrandolf-old:
`addToolResult` is causing excess re-renders.


lgrammel:
Please run `pnpm prettier-fix`

PR #4516, opened 2025-01-24
feat (provider/luma): add Luma provider

shaper:


packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+        case 'completed':
+          if (!statusResponse.assets?.image) {
+            throw new InvalidResponseDataError({
+              data: statusResponse,
+              message: `Image generation completed but no image was found.`,
+            });
+          }
+          return statusResponse.assets.image;
+        case 'failed':
+          throw new InvalidResponseDataError({
```
shaper, line 155: This case is technically an assertive failure on the Luma server side -- perhaps there is a better error type or approach to handle here.

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+          });
+        case 'dreaming':
+        case 'queued':
+          break;
+      }
+
+      attemptCount++;
+      await new Promise(resolve =>
+        setTimeout(resolve, this.pollIntervalMillis),
+      );
```
lgrammel, line 167: we should move `delay` from `ai` to `provider-utils` and use it here

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+    throw new Error(
+      `Image generation timed out after ${this.maxPollAttempts} attempts.`,
+    );
+  }
+
+  private createLumaErrorHandler() {
+    return createJsonErrorResponseHandler({
+      errorSchema: lumaErrorSchema,
+      errorToMessage: (error: LumaErrorData) =>
+        error.detail[0].msg || 'Unknown error',
```
lgrammel, line 179: prefer `??`

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+    });
+  }
+
+  private getLumaGenerationsUrl(generationId?: string) {
+    return `${this.config.baseURL}/dream-machine/v1/generations/${
+      generationId || 'image'
```
lgrammel, line 185: prefer `??`

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+          }
+          return statusResponse.assets.image;
+        case 'failed':
+          throw new InvalidResponseDataError({
+            data: statusResponse,
+            message: `Image generation failed.`,
+          });
+        case 'dreaming':
+        case 'queued':
+          break;
```
lgrammel, line 161: not strictly needed

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  private async pollForImageUrl(
+    generationId: string,
+    headers: Record<string, string | undefined>,
+    abortSignal: AbortSignal | undefined,
+  ): Promise<string> {
+    let attemptCount = 0;
+    const url = this.getLumaGenerationsUrl(generationId);
+    while (attemptCount < this.maxPollAttempts) {
```
lgrammel, line 133: could be a for loop

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+  failure_reason: z.string().nullish(),
+  created_at: z.string(),
+  assets: z
+    .object({
+      image: z.string(), // URL of the generated image
+    })
+    .nullish(),
+  model: z.string(),
+  request: lumaRequestSchema,
+});
```
lgrammel, line 256: is all of this needed or can it be slimmed down?

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+    z.object({
+      type: z.string(),
+      loc: z.array(z.string()),
+      msg: z.string(),
+      input: z.string(),
+      ctx: z
+        .object({
+          expected: z.string(),
+        })
+        .optional(),
```
lgrammel, line 269: prefer `nullish` - apis tend to just change around undefined vs null

packages/luma/src/luma-image-model.ts
```diff
@@ -0,0 +1,274 @@
...
+  createJsonResponseHandler,
+  createJsonErrorResponseHandler,
+  createStatusCodeErrorResponseHandler,
+  getFromApi,
+  postJsonToApi,
+} from '@ai-sdk/provider-utils';
+import { LumaImageSettings } from './luma-image-settings';
+import { z } from 'zod';
+
+const DEFAULT_POLL_INTERVAL_MILLIS = 5000;
```
lgrammel, line 19: poll every second? or every 500ms? by default

packages/provider-utils/src/response-handler.ts
```diff
@@ -184,3 +184,57 @@ export const createJsonResponseHandler =
...
+        responseHeaders,
+        responseBody: undefined,
+      });
+    }
+
+    try {
+      const buffer = await response.arrayBuffer();
+      return {
+        responseHeaders,
+        value: buffer,
```
lgrammel, line 208: uint8 array instead so we have the conversion here and not in every provider?

packages/provider-utils/src/response-handler.ts
```diff
@@ -184,3 +184,57 @@ export const createJsonResponseHandler =
       rawValue: parsedResult.rawValue,
     };
   };
+
+export const createBinaryResponseHandler =
+  (): ResponseHandler<ArrayBuffer> =>
```
lgrammel, line 189: ResponseHandler<UInt8Array> ?

lgrammel (approved):

packages/luma/src/luma-image-model.ts
shaper: I will copy it for this PR and do the move in a second as it touches a batch of files in `ai`.

packages/luma/src/luma-image-model.ts
shaper: reduced quite a bit

packages/provider-utils/src/response-handler.ts
shaper: I played with this -- if I omit the generic, I get `Generic type 'ResponseHandler' requires 1 type argument(s).` We can omit the return type as it's inferrable, but then we get `Binding element 'response' implicitly has an 'any' type` on the
de-sugared args on line 190, which we could probably fix by specifying their types but that's more verbose than just keeping line 189.

I may be missing what you're getting at, happy to look at this further later.

PR #4549, opened 2025-01-27
Add brace balancing to `safeParseJSON` for malformed JSON strings

edobrb:
When inferring large objects with `generateObject`, parsing sometimes fails due to AI outputs having incorrect numbers of closing braces `}`. This PR enhances `safeParseJSON` to attempt parsing with different brace configurations, adding or removing closing braces to handle malformed JSON more gracefully.

Key changes:
 - Added `generateJSONBraceVariations` to create variations of JSON strings with different brace counts
 - Modified `safeParseJSON` to attempt parsing each variation if the initial parse fails (only if the option `tryMutipleBraceVariations` is true)
 - Set `tryMutipleBraceVariations` to true in `generateObject`

### DIFF
--- a/packages/ai/core/generate-object/generate-object.ts
+++ b/packages/ai/core/generate-object/generate-object.ts
@@ -675,7 +675,10 @@ export async function generateObject<SCHEMA, RESULT>({
         }
       }
 
-      const parseResult = safeParseJSON({ text: result });
+      const parseResult = safeParseJSON({
+        text: result,
+        tryMutipleBraceVariations: true,
+      });
 
       if (!parseResult.success) {
         throw new NoObjectGeneratedError({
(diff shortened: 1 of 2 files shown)


lgrammel:
This goes against the app architecture, there is no way I'm going to make this change. Instead, we should introduce a repair function on generateObject

edobrb:
Yeah good idea! I'll work on another PR later

lgrammel:
You can take inspiration from tool call repair on generateObject. The goal is to allow the user to specify how exactly the repair should happen (vs doing anything magically in a core area of the library)

edobrb:
What about a handler that's called when the json parse is failed? This would be added on the arguments of `generateObject`.

```typescript
repairReponse?: (args: { response: string; error: JSONParseError | TypeValidationError;  }) => string;
```

PR #4581, opened 2025-01-29
added bedrock middleware adapter method

chitalian:
Allow the user to modify the Bedrock client to add a middleware. 

ex:
```ts

const provider = createAmazonBedrock({
  region: 'us-east-1',
  accessKeyId: 'test-access-key',
  secretAccessKey: 'test-secret-key',
  sessionToken: 'test-token-key',
  bedrockMiddleware: client => {
    client.middlewareStack.add(
      (next, context) => async (args: any) => {
        if (!args.request.headers) {
          args.request.headers = {};
        }

        // Add your custom headers here
        args.request.headers['Test'] = `Example`;

        return next(args);
      },
      {
        step: 'build',
        name: 'addCustomHeaders',
      },
    );

    return client;
  },
});


```

Fixes

- [4517](https://github.com/vercel/ai/discussions/4517)
- [4475](https://github.com/vercel/ai/discussions/4575)

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-provider.ts
+++ b/packages/amazon-bedrock/src/bedrock-provider.ts
@@ -38,6 +38,11 @@ export interface AmazonBedrockProviderSettings {
 
   // for testing
   generateId?: () => string;
+
+  /**
+   * Middleware to modify the BedrockRuntimeClient instance.
+   */
+  bedrockMiddleware?: (client: BedrockRuntimeClient) => BedrockRuntimeClient;
 }
 
 export interface AmazonBedrockProvider extends ProviderV1 {
@@ -63,8 +68,8 @@ Create an Amazon Bedrock provider instance.
 export function createAmazonBedrock(
   options: AmazonBedrockProviderSettings = {},
 ): AmazonBedrockProvider {
-  const createBedrockRuntimeClient = () =>
-    new BedrockRuntimeClient(
+  const createBedrockRuntimeClient = () => {
+    const client = new BedrockRuntimeClient(
       options.bedrockOptions ?? {
         region: loadSetting({
           settingValue: options.region,
@@ -92,6 +97,13 @@ export function createAmazonBedrock(
         },
       },
     );
+
+    if (options.bedrockMiddleware) {
+      return options.bedrockMiddleware(client);
+    }
+
+    return client;
+  };
 
   const createChatModel = (
     modelId: BedrockChatModelId,


shaper:
Hi there, thank you for the PR! We are in the midst of reworking the bedrock provider (if curious, see https://github.com/vercel/ai/pull/4582) to remove the dependency on the Bedrock SDK, and so the smithy-based middleware hook will go away.

I think we can achieve your goal of adding to headers through a more standard AI SDK pattern:
- add a provider setting to allow specifying a function that produces headers
- at request time, this function is called and given a chance to produce whatever add'l headers it would like to add
- internally in the current provider, we essentially do what you have done here

The goal is to avoid exposing the bedrock-client or the middleware reference directly, since those will be going away, while still providing a way to add to the headers on each request.

You can see an example where we use this pattern with the Google Vertex provider calling into the Google GenAI language model code, ref: [1](https://github.com/vercel/ai/blob/3551c1df311c99613f6399366f316181a1225d28/packages/google-vertex/src/google-vertex-provider.ts#L67), [2](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L34), [3](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L213).

I can take your PR and evolve it to reflect the above, would that meet your needs?

chitalian:
> Hi there, thank you for the PR! We are in the midst of reworking the bedrock provider (if curious, see #4582) to remove the dependency on the Bedrock SDK, and so the smithy-based middleware hook will go away.
> 
> I think we can achieve your goal of adding to headers through a more standard AI SDK pattern:
> 
> * add a provider setting to allow specifying a function that produces headers
> * at request time, this function is called and given a chance to produce whatever add'l headers it would like to add
> * internally in the current provider, we essentially do what you have done here
> 
> The goal is to avoid exposing the bedrock-client or the middleware reference directly, since those will be going away, while still providing a way to add to the headers on each request.
> 
> You can see an example where we use this pattern with the Google Vertex provider calling into the Google GenAI language model code, ref: [1](https://github.com/vercel/ai/blob/3551c1df311c99613f6399366f316181a1225d28/packages/google-vertex/src/google-vertex-provider.ts#L67), [2](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L34), [3](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L213).
> 
> I can take your PR and evolve it to reflect the above, would that meet your needs?

That would be great! + The DX is a lot better

Sounds good, please let me know if you need any help - we will be eagerly awaiting this change

chitalian:
Hey @shaper can we just merge this in for now, and seems like #4582 is on hold

shaper:
Thanks for your patience -- we are close to landing https://github.com/vercel/ai/pull/4582 which will have `headers` support as discussed, along with a rewrite to use REST instead of the bedrock client library.

If you'd like to try it out early there is a snapshot release version of `@ai-sdk/amazon-bedrock` from yesterday `0.0.0-c0cd7082-20250211021941` which I've tested with `ai` `4.1.25`.

The one thing to be aware of migration-wise is that we've removed `bedrockOptions` from the provider settings, and so you need to pass credential related settings directly in the provider settings. You may need to set unused credential settings e.g. `sessionToken` to `undefined` if for example they pick up any env vars defined by your containing runtime (serverless fns, for example, may have `AWS_xxx` defined even if you don't yourself).

I expect we will publish a proper release of the new version in the next day or two, so more than fine to just wait, but I didn't want to leave you hanging without an update. We're almost there, and would welcome your feedback once you do get a chance to try the new thing.

shaper:
Just merged #4582 which is published as v2.0.0. Please try `headers` settings with that and let us know any feedback.

Issue #2780, opened 2024-08-22
Bedrock client does not work natively with profiles, instance roles, container roles, etc.


mrburrito:
### Description

The Bedrock provider currently requires explicitly defining either the AWS credentials or the environment variables `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and (optionally) `AWS_SESSION_TOKEN` when using `createAmazonBedrock()` instead of relying on Amazon's internal credential chains. This makes it nearly impossible to use with assumed roles through SSO profiles, instance roles, or container roles in ECS or EKS since those variables would have to be defined, and then updated, at runtime.

Instead of forcing a lookup from `process.env` in the options, and failing if undefined, pass those through to the `BedrockRuntimeClient` and let it handle credential resolution natively. It will use those values if you provide them, falling back to the `AWS_*` environment variables, then `AWS_PROFILE`, then container role lookup, then instance role lookup, then failure.

### Expected Behavior

The Bedrock provider uses AWS' native credential chain to resolve access from any supported credential provider, calls Bedrock, and returns the response from the model.

### Actual Behavior

`bedrock` and `createAmazonBedrock` functions throw an exception if credentials are not explicitly provided using environment variables or the options object.

### Code example

```
import { bedrock } from '@ai-sdk/amazon-bedrock';
import {
  BedrockRuntimeClient,
  InvokeModelCommand,
} from '@aws-sdk/client-bedrock-runtime';

async function vercelAi() {
  try {
    const { text } = await generateText({
      model: bedrock('anthropic.claude-3-5-sonnet-20240620-v1:0'),
      prompt:
        'Please explain the order in which the default AWS Credential Chain looks up credentials.',
    });
    console.log(text);
  } catch (e) {
    console.error(e);
  }
}

async function nativeBedrock() {
  const client = new BedrockRuntimeClient();
  const payload = {
    anthropic_version: 'bedrock-2023-05-31',
    max_tokens: 4096,
    messages: [
      {
        role: 'user',
        content: [
          {
            type: 'text',
            text: 'Please explain the order in which the default AWS Credential Chain looks up credentials.',
          },
        ],
      },
    ],
  };
  const command = new InvokeModelCommand({
    modelId: 'anthropic.claude-3-5-sonnet-20240620-v1:0',
    body: JSON.stringify(payload),
  });

  try {
    const response = await client.send(command);
    const decodedResponseBody = new TextDecoder().decode(response.body);
[...]

PR #4582, opened 2025-01-29
feat (provider/amazon-bedrock): remove dependence on AWS SDK Bedrock library

shaper:
- [x] tool calls
- [x] backwards-compatibility review
- [x] docs incl. bedrock request-id detail
- [x] fix CI failure
- [x] object generation
- [x] embed
- [x] validate bedrock request-id detail in `responseHeaders`


examples/ai-core/src/generate-text/amazon-bedrock.ts
```diff
@@ -4,8 +4,10 @@ import 'dotenv/config';
 
 async function main() {
   const result = await generateText({
-    model: bedrock('anthropic.claude-3-haiku-20240307-v1:0'),
-    prompt: 'Invent a new holiday and describe its traditions.',
+    model: bedrock(
+      'arn:aws:bedrock:us-east-2:474668406012:inference-profile/us.anthropic.claude-3-5-sonnet-20240620-v1:0',
```
lgrammel, line 8: we need to ideally figure out how to make it backwards compatible, or we need to announce breaking changes somehow

packages/amazon-bedrock/src/bedrock-eventstream-codec.ts
```diff
@@ -0,0 +1,104 @@
...
+  safeParseJSON,
+  extractResponseHeaders,
+  ResponseHandler,
+} from '@ai-sdk/provider-utils';
+import { EventStreamCodec } from '@smithy/eventstream-codec';
+import { toUtf8, fromUtf8 } from '@smithy/util-utf8';
+import { ZodSchema } from 'zod';
+
+// https://docs.aws.amazon.com/lexv2/latest/dg/event-stream-encoding.html
+export const createEventSourceResponseHandler =
```
lgrammel, line 13: would rename to createBedrockResponseHandler or something (include Bedrock)

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -95,20 +99,14 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
-    }
-
     if (topK != null) {
       warnings.push({
         type: 'unsupported-setting',
         setting: 'topK',
       });
     }
 
+    // TODO: validate this is still the case.
```
shaper, line 109: @lgrammel do you remember the backstory on this warning below?

examples/ai-core/src/generate-object/amazon-bedrock.ts
```diff
@@ -5,7 +5,9 @@ import { z } from 'zod';
 
 async function main() {
   const result = await generateObject({
-    model: bedrock('anthropic.claude-3-5-sonnet-20240620-v1:0'),
+    model: bedrock(
+      `arn:aws:bedrock:${process.env.AWS_REGION}:${process.env.AWS_ACCOUNT_ID}:inference-profile/us.anthropic.claude-3-5-sonnet-20240620-v1:0`,
```
lgrammel, line 9: Can we use the simple ids here?

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
lgrammel: probably didn't work back then - let's test if it works now

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -212,49 +220,50 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
           toolName: part.toolUse?.name ?? `tool-${this.config.generateId()}`,
           args: JSON.stringify(part.toolUse?.input ?? ''),
         })),
-      finishReason: mapBedrockFinishReason(response.stopReason),
+      finishReason: mapBedrockFinishReason(
+        response.stopReason as BedrockStopReason,
```
lgrammel, line 224: no mapping necessary?

lgrammel, line 224: why is the `as` needed here? shouldn't zod validation narrow the types correctly?

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -265,11 +274,8 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
     > = {};
 
     return {
-      stream: stream.pipeThrough(
-        new TransformStream<
-          ParseResult<ConverseStreamOutput>,
-          LanguageModelV1StreamPart
-        >({
+      stream: response.pipeThrough(
+        new TransformStream<ParseResult<any>, LanguageModelV1StreamPart>({
```
lgrammel, line 278: `any` should be replaced with the actual type from parsing. What is the type of the chunks? we are parsing with Zod?

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+                // 3) Decode exactly the sub-slice for this event.
+                const subView = buffer.subarray(0, totalLength);
+                const decoded = codec.decode(subView);
+
+                // Slice the used bytes out of the buffer, removing this message.
+                buffer = buffer.slice(totalLength);
+
+                // Process the message.
+                if (decoded.headers[':message-type']?.value === 'event') {
+                  const data = new TextDecoder().decode(decoded.body);
```
lgrammel, line 62: extract text decode, this is most likely a hot loop and we could init it before the loop (i.e. outside of the transform stream)

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+
+                  // Wrap the data in the `:event-type` field to match the expected schema.
+                  const parsedDataResult = safeParseJSON({ text: data });
+                  let wrappedData;
```
lgrammel, line 66: push down to first use

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+
+                  // The `p` field appears to be padding or some other non-functional field.
+                  delete (parsedDataResult.value as any).p;
+                  wrappedData = {
+                    [decoded.headers[':event-type']?.value as string]:
+                      parsedDataResult.value,
+                  };
+
+                  // Re-parse with the expected schema.
+                  const parsedWrappedData = safeParseJSON({
```
lgrammel, line 80: no need to JSON.stringify and parse again, use `safeValidate` instead

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+                    break;
+                  }
+
+                  controller.enqueue({
+                    success: true,
+                    value: parsedWrappedData.value,
+                    rawValue: parsedWrappedData.rawValue,
+                  });
+                }
+              } catch (e) {
```
lgrammel, line 95: shouldn't the error be forward in some fashion?

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
```diff
@@ -0,0 +1,104 @@
...
+                  if (!parsedWrappedData.success) {
+                    controller.enqueue(parsedWrappedData);
+                    break;
+                  }
+
+                  controller.enqueue({
+                    success: true,
+                    value: parsedWrappedData.value,
+                    rawValue: parsedWrappedData.rawValue,
+                  });
```
lgrammel, line 93: aren't both cases the same, i.e. `controller.enqueue(parsedWrappedData);` is enough?

packages/amazon-bedrock/src/bedrock-provider.ts
```diff
@@ -63,42 +88,33 @@ Create an Amazon Bedrock provider instance.
...
-          }),
-          sessionToken: loadOptionalSetting({
-            settingValue: options.sessionToken,
-            environmentVariableName: 'AWS_SESSION_TOKEN',
-          }),
-        },
-      },
-    );
+        })}.amazonaws.com`,
+    ) ?? '';
```
lgrammel, line 107: the `?? ''` could hide errors

packages/amazon-bedrock/src/bedrock-provider.ts
```diff
@@ -63,42 +88,33 @@ Create an Amazon Bedrock provider instance.
...
-            settingName: 'secretAccessKey',
-            environmentVariableName: 'AWS_SECRET_ACCESS_KEY',
-            description: 'AWS secret access key',
-          }),
-          sessionToken: loadOptionalSetting({
-            settingValue: options.sessionToken,
-            environmentVariableName: 'AWS_SESSION_TOKEN',
-          }),
-        },
-      },
```
lgrammel, line 93: I still feel we are delaying errors by pushing this down but could be wrong.

packages/amazon-bedrock/src/bedrock-sigv4-signing-function.ts
```diff
@@ -0,0 +1,72 @@
...
+    const signer = new AwsV4Signer({
+      url,
+      method: 'POST',
+      headers: Object.entries(headers).filter(([_, v]) => v !== undefined) as [
+        string,
+        string,
+      ][],
+      // TODO: explore avoiding the below stringify since we do it again at
+      // post-time and the content could be large with attachments.
+      body: JSON.stringify(body),
```
lgrammel, line 59: might be worth resolving before landing considering attachments

would the fetch wrapper approach solve this?

packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts
```diff
@@ -67,9 +67,13 @@ export function convertToBedrockChatMessages(
...
                       image: {
-                        format: part.mimeType?.split('/')?.[1] as ImageFormat,
+                        format: part.mimeType?.split(
+                          '/',
+                        )?.[1] as BedrockImageFormat,
                         source: {
-                          bytes: part.image ?? (part.image as Uint8Array),
+                          bytes: Buffer.from(
+                            part.image ?? (part.image as Uint8Array),
+                          ).toString('base64'),
```
lgrammel, line 76: do we have a helper for this in provider utils?

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
shaper: The response zod schema was just parsing it as a string. I updated the zod schema so we don't need the `as` any more and we have better type info.

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
shaper: Fixed to reference the chunk schema, added tests for error cases and chunk-parsing-failed that turned up as part of the work.

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
shaper: We need the `rawValue` in the success case, which with the shift to use `safeValidateTypes` is not part of `parsedWrappedData`, it is `wrappedData`.

We could potentially update `safeValidateTypes` to include `rawValue`.

We can make it more concise with ternary op but we still need a logic split, e.g.

```
const result = parsedWrappedData.success
  ? { success: true, value: parsedWrappedData.value, rawValue: wrappedData }
  : parsedWrappedData;
controller.enqueue(result);
```

It doesn't read as better to me. But I do think an if/else rather than a break could make sense, will try that at least re: cleanup.

packages/amazon-bedrock/src/bedrock-event-stream-response-handler.ts
shaper: This is effectively message data accumulation. AFAIK it is possible we could receive only a partial message, which could be normal based on network traffic.

The decode could then throw, ref. [splitMessage](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/eventstream-codec/src/splitMessage.ts) called by [EventStreamCodec](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/eventstream-codec/src/EventStreamCodec.ts) or [HeaderMarshaller.parse](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/eventstream-codec/src/HeaderMarshaller.ts#L86).

It has a bit of a code smell to it, admittedly. Preferable would perhaps be to do some top level frame scrutiny and gracefully continue, vs try-decode, catch-and-repeat. However, `splitMessage` is basically doing that for us, so we'd be duplicating logic (thus, introducing some fragility re: bifurcation from library) if we did it here:

```
  if (byteLength < MINIMUM_MESSAGE_LENGTH) {
    throw new Error("Provided message too short to accommodate event stream message overhead");
  }

  const view = new DataView(buffer, byteOffset, byteLength);

  const messageLength = view.getUint32(0, false);

  if (byteLength !== messageLength) {
    throw new Error("Reported message length does not match received message length");
  }
```

In manual testing I have not seen this case hit, but I am on a fantastic network connection.

Another revision might be to track if we fail too many times in sequence, and/or for too long, and exit with a propagated error.

I am open to more work here if we think it makes sense, will wait to hear your thought re: above.

shaper: Here's a log of the streamText example just for us to look at it together, the normal flow is one full message per call to `transform`, with size ~200 bytes, but some variance at start/end. But I don't think we have an assurance that this will be the case?

https://gist.github.com/shaper/43964f3253616609d841134d4e131a34

packages/amazon-bedrock/src/bedrock-provider.ts
shaper: The pattern in other providers for this case is to hard-code a fallback default base url. Here, the base url would ideally include the `region`, however we can't be sure we have it, so I'll hard-code `https://bedrock-runtime.us-east-1.amazonaws.com`. Let me know if you have something else in mind -- `getBaseUrl` doesn't allow undefined.

shaper: Ah, maybe it's better to throw if `region` isn't available... I see similar in the vertex provider. Will consider.

packages/amazon-bedrock/src/bedrock-provider.ts
shaper: Let me look into moving it to happen earlier, we could attempt to build credentials pretty much straightaway, executing these settings lookups, and then pass them to where they're needed.

There are two things that I think it will hit:
- default provider instance use in unit tests where we don't have the env vars defined, but as we've discussed we could shift those to instantiate the model directly (I already did, this may have been resolved).
- example scripts where we may import the provider (and thus instantiate the default instance) before we import `dotenv/config`. Not sure what to do about this one.

packages/amazon-bedrock/src/bedrock-sigv4-signing-function.ts
shaper: Yes, fetch wrapper would solve. I'll try it out this weekend.

shaper: Done in https://github.com/vercel/ai/pull/4582/commits/b20dd9172f31e3a6a6e52b5ee2999548e7fa1df7 (there is a follow-up I'm still working on to pass in credentials so we fail earlier as discussed elsewhere).

packages/amazon-bedrock/src/bedrock-provider.ts
shaper: > * example scripts where we may import the provider (and thus instantiate the default instance) before we import `dotenv/config`. Not sure what to do about this one.

We could edit the default-provider pattern we use to lazily construct the instance. Arguably executing settings-load-lookup on import of code before any instance is referenced or method/fn is called is a little premature. I will try this and see how it looks.

examples/ai-core/src/generate-object/amazon-bedrock.ts
shaper: Reverting this and others, although it seems it varies by account and model-enablement status. Perhaps there is an AWS UI I haven't found to let me enable simple ids for my account.

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
shaper: I tested https://github.com/vercel/ai/tree/main/examples/ai-core/src/test modified to use bedrock:
- it prefaces with non-json text, `Here is a new holiday and its traditions in JSON format:`
- it doesn't conform to the schema which specifies one field names `text` (but neither does the openai existing test case)

For now I'll remove the TODO and leave the warning.

content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
```diff
@@ -316,3 +292,76 @@ The following optional settings are available for Bedrock Titan embedding models
...
+{
+  connection: 'keep-alive',
+  'content-type': 'application/vnd.amazon.eventstream',
+  date: 'Fri, 07 Feb 2025 04:33:37 GMT',
+  'transfer-encoding': 'chunked',
+  'x-amzn-requestid': 'a976e3fc-0e45-4241-9954-b9bdd80ab407'
+}
+```
+
+## Migrating from `@ai-sdk/amazon-bedrock` pre-v2.x
```
lgrammel, line 354: Migrating to `@ai-sdk/amazon-bedrock` 2.x

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+    | BedrockGuardrailConfiguration
+    | BedrockGuardrailStreamConfiguration
+    | undefined;
+}
+
+export interface BedrockGuardrailConfiguration {
+  guardrails?: Array<{
+    name: string;
+    description?: string;
+    parameters?: Record<string, any>;
```
lgrammel, line 27: Is `any` required or can `unknown` be used here?

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+  }>;
+}
+
+export type BedrockGuardrailStreamConfiguration = BedrockGuardrailConfiguration;
+
+export interface BedrockToolInputSchema {
+  json: Record<string, any>;
```
lgrammel, line 34: Is `any` required or can `unknown` be used here?

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+export interface BedrockTool {
+  toolSpec: {
+    name: string;
+    description?: string;
+    inputSchema: { json: any };
```
lgrammel, line 41: Is `any` required or can `unknown` be used here?

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+    format: BedrockDocumentFormat;
+    name: string;
+    source: {
+      bytes: string;
+    };
+  };
+}
+
+export interface BedrockGuardrailConverseContentBlock {
+  guardContent: any;
```
lgrammel, line 85: Is `any` required or can `unknown` be used here?

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+    toolUseId: string;
+    content: Array<{ text: string }>;
+  };
+}
+
+export interface BedrockToolUseBlock {
+  toolUse: {
+    toolUseId: string;
+    name: string;
+    input: Record<string, any>;
```
lgrammel, line 108: Is `any` required or can `unknown` be used here?

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -166,7 +159,7 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
                     description: mode.tool.description,
                     inputSchema: {
                       json: mode.tool.parameters,
-                    } as ToolInputSchema,
+                    } as BedrockToolInputSchema,
```
lgrammel, line 162: Why is a cast necessary here? or can it be removed?

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+
+  private getUrl(modelId: string) {
+    const encodedModelId = encodeURIComponent(modelId);
+    return `${this.config.baseUrl()}/model/${encodedModelId}/converse`;
+  }
+
+  private getStreamUrl(modelId: string): string {
+    const encodedModelId = encodeURIComponent(modelId);
+    return `${this.config.baseUrl()}/model/${encodedModelId}/converse-stream`;
+  }
```
lgrammel, line 414: Alternative: have a `getUrl` method and then use `${this.getUrl(modelId)/converse` and `${this.getUrl(modelId)/converse-stream` - either way works

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+  }),
+});
+
+// limited version of the schema, focussed on what is needed for the implementation
+// this approach limits breakages when the API changes and increases efficiency
+const BedrockStreamSchema = z.object({
+  contentBlockDelta: z
+    .object({
+      contentBlockIndex: z.number(),
+      delta: z.record(z.any()).nullish(),
```
lgrammel, line 462: We should avoid `any` in the response parsing. If we don't want to go into details, we can use `z.unknown` afaik

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+              input: z.any(),
+            })
+            .optional(),
+        }),
+      ),
+      role: z.string(),
+    }),
+  }),
+  stopReason: BedrockStopReasonSchema,
+  trace: z.any().nullish(),
```
lgrammel, line 448: `any` (see below)

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+  contentBlockDelta: z
+    .object({
+      contentBlockIndex: z.number(),
+      delta: z.record(z.any()).nullish(),
+    })
+    .nullish(),
+  contentBlockStart: z
+    .object({
+      contentBlockIndex: z.number(),
+      start: z.record(z.any()).nullish(),
```
lgrammel, line 468: `any`

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+  internalServerException: z.record(z.any()).nullish(),
+  messageStop: z
+    .object({
+      additionalModelResponseFields: z.any().nullish(),
+      stopReason: BedrockStopReasonSchema,
+    })
+    .nullish(),
+  metadata: z
+    .object({
+      trace: z.any(),
```
lgrammel, line 485: `any`

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+    .nullish(),
+  contentBlockStop: z
+    .object({
+      contentBlockIndex: z.number(),
+    })
+    .nullish(),
+  internalServerException: z.record(z.any()).nullish(),
+  messageStop: z
+    .object({
+      additionalModelResponseFields: z.any().nullish(),
```
lgrammel, line 479: `any`

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -377,19 +387,111 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+        .object({
+          inputTokens: z.number(),
+          outputTokens: z.number(),
+        })
+        .nullish(),
+    })
+    .nullish(),
+  modelStreamErrorException: z.record(z.any()).nullish(),
+  throttlingException: z.record(z.any()).nullish(),
+  validationException: z.record(z.any()).nullish(),
```
lgrammel, line 496: `any` - can we get a more detailed error type?

packages/amazon-bedrock/src/bedrock-embedding-model.ts
```diff
@@ -1,73 +1,97 @@
...
+        failedResponseHandler: createJsonErrorResponseHandler({
+          errorSchema: BedrockErrorSchema,
+          errorToMessage: error => `${error.type}: ${error.message}`,
+        }),
+        successfulResponseHandler: async response => {
+          const binaryData = await response.response.arrayBuffer();
+          const jsonString = new TextDecoder().decode(
+            new Uint8Array(binaryData),
+          );
+          const parsed = JSON.parse(jsonString);
```
lgrammel, line 71: We are using SafeJSON everywhere (that's why our internal parsing is preferred)

lgrammel: would createJSONResponseHandler work here?

packages/amazon-bedrock/src/bedrock-prepare-tools.ts
```diff
@@ -40,7 +40,7 @@ export function prepareTools(
           description: tool.description,
           inputSchema: {
             json: tool.parameters,
-          } as ToolInputSchema,
+          } as BedrockToolInputSchema,
```
lgrammel, line 43: Is this typecast needed?

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -0,0 +1,116 @@
...
+
+@param getCredentials - Function that returns the AWS credentials to use when signing.
+@param originalFetch - Optional original fetch implementation to wrap. Defaults to global fetch.
+@returns A FetchFunction that signs requests before passing them to the underlying fetch.
+ */
+export function createSigV4FetchFunction(
+  getCredentials: () => BedrockCredentials,
+  originalFetch?: FetchFunction,
+): FetchFunction {
+  const fetchImpl = originalFetch || globalThis.fetch;
```
lgrammel, line 26: Can use parameter default here and eliminate variable

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -0,0 +1,116 @@
...
+    // Determine the URL from the fetch input.
+    const url =
+      typeof input === 'string'
+        ? input
+        : input instanceof URL
+        ? input.href
+        : input.url;
+
+    const originalHeaders = extractHeaders(init.headers);
+    const bodyString = prepareBodyString(init.body);
```
lgrammel, line 45: inline

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -0,0 +1,116 @@
...
+    const mergedHeaders = removeUndefinedEntries(
+      combineHeaders(originalHeaders, signedHeaders),
+    );
+
+    // Create a new RequestInit with the merged headers including the signed headers.
+    const newInit: RequestInit = {
+      ...init,
+      body: bodyString,
+      headers: mergedHeaders,
+    };
```
lgrammel, line 72: can be inlined

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
lgrammel, line 45: rename to just `body`? would also simplify line 51

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -0,0 +1,116 @@
...
+      region: credentials.region,
+      accessKeyId: credentials.accessKeyId,
+      secretAccessKey: credentials.secretAccessKey,
+      ...(credentials.sessionToken && {
+        sessionToken: credentials.sessionToken,
+      }),
+      service: 'bedrock',
+    });
+
+    const result = await signer.sign();
```
lgrammel, line 61: maybe somehow indicate that `result` is signed

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -0,0 +1,116 @@
...
+    const signedHeaders = convertHeadersToRecord(result.headers);
+    const mergedHeaders = removeUndefinedEntries(
+      combineHeaders(originalHeaders, signedHeaders),
+    );
```
lgrammel, line 65: inline?

packages/amazon-bedrock/src/fake-fetch.ts
```diff
@@ -0,0 +1,22 @@
...
+      for (const [key, value] of Object.entries(customHeaders)) {
+        init.headers.set(key, value);
+      }
+    } else if (Array.isArray(init.headers)) {
+      for (const [key, value] of Object.entries(customHeaders)) {
+        init.headers.push([key, value]);
+      }
+    } else {
+      init.headers = { ...(init.headers || {}), ...customHeaders };
+    }
```
lgrammel, line 18: can use extractHeaders to unify?

packages/amazon-bedrock/src/fake-fetch.ts
```diff
@@ -0,0 +1,22 @@
+import { FetchFunction } from '@ai-sdk/provider-utils';
+
+export function createFakeFetch(
```
lgrammel, line 3: rename to what it does? e.g. `modifyFetchHeaders` or `injectFetchHeaders`

packages/amazon-bedrock/src/fake-fetch.ts
```diff
@@ -0,0 +1,22 @@
...
+        init.headers.set(key, value);
+      }
+    } else if (Array.isArray(init.headers)) {
+      for (const [key, value] of Object.entries(customHeaders)) {
+        init.headers.push([key, value]);
+      }
+    } else {
+      init.headers = { ...(init.headers || {}), ...customHeaders };
+    }
+    // Delegate to the global fetch (MSW will intercept it).
```
lgrammel, line 19: rm comment?

lgrammel (approved):

packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -0,0 +1,122 @@
...
+  'content-filter',
+  'content_filtered',
+  'guardrail_intervened',
+  'tool-calls',
+  'tool_use',
+] as const;
+
+export type BedrockStopReason =
+  | (typeof BEDROCK_STOP_REASONS)[number]
+  | (string & {});
```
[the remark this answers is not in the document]
shaper: Changed back to fixed set.

packages/amazon-bedrock/src/bedrock-api-types.ts
shaper: `JSONObject` looks viable.

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
shaper: For these exceptions, the most detail I've been able to turn up is below. I'm not sure how fully we want to encode this. I'll change `any` to `unknown` for now. There are four top-level fields common to them all, it appears, all marked required, the field types vary:

`$fault`
`$metadata`
`$response`
`$retryable`

- https://docs.aws.amazon.com/bedrock/latest/APIReference/CommonErrors.html
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-bedrock-agent-runtime/Class/InternalServerException/
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-bedrock-runtime/Class/ThrottlingException/
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-bedrock-runtime/Class/ModelStreamErrorException/
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-bedrock-runtime/Class/InternalServerException/

phuctm97:
Will this change work with AWS EC2/ECS instance credentials? The v1 works automatically without setting any environment variables when running on a EC2/ECS instance with instance access to Amazon Bedrock.

shaper:
> Will this change work with AWS EC2/ECS instance credentials? The v1 works automatically without setting any environment variables when running on a EC2/ECS instance with instance access to Amazon Bedrock.

Hi there, we will look into this.

meleksomai:
Hey team, 

Although this looks great as the Bedrock AI SDK has less external dependencies, the new update makes this new package super limited. For instance, we rely on [Vercel OIDC](https://vercel.com/docs/security/secure-backend-access/oidc/aws) to connect securely to our AWS Backend which the new package does not support.

What is the potential way forward?

shaper:
> Hey team,
> 
> Although this looks great as the Bedrock AI SDK has less external dependencies, the new update makes this new package super limited. For instance, we rely on [Vercel OIDC](https://vercel.com/docs/security/secure-backend-access/oidc/aws) to connect securely to our AWS Backend which the new package does not support.
> 
> What is the potential way forward?

Hi there, in case it is helpful please take a look at https://github.com/vercel/ai/issues/2780#issuecomment-2660787835 which shows how aws sso and a range of other creds files/locations/etc. can be used.

I will look more closely at OIDC specifically, but on initial read it looks like a similar approach would work, just revising my example at above link to use `awsCredentialsProvider` instead.

Let us know if this works for you. I plan to update the Bedrock docs to describe these alternate authentication approaches soon.

PR #4586, opened 2025-01-29
don't inline files for google vertex

### LINKED ISSUE #4559: Vertex Supported File Uri
### Description

The vertex language model doesn't properly handle [all supported file uris](https://firebase.google.com/docs/vertex-ai/input-file-requirements#provide-file-using-url). Behavior right now is that it only supports `https://generativelanguage.googleapis.com/upload/v1beta/files` since it inherits from normal google generative ai language model. New language model with different supportedUrl would be the solution here.

https://github.com/vercel/ai/blob/8fa9cc16648e4a1d1309324ce3fcebe6eb55b66e/packages/google/src/google-generative-ai-language-model.ts#L196C1-L200C4

### Code example

_No response_

### AI provider

@ai-sdk/google-vertex

### Additional context

_No response_

BrianHung:
Fixes #4559 by creating a `GoogleVertexLanguageModel` that extends `GoogleGenerativeAILanguageModel` but has `supportUrl` set to `true`.


shaper:
Continuing this work in #4620 with your commits included to track credit.

PR #4591, opened 2025-01-29
fix (ui): single assistant message with multiple tool steps

lgrammel:
## Background
`useChat` current produces several assistant messages in sequence when there are client or server-side tool calls in separate steps.

## Summary
- there is always a user-assistant-user-assistant message sequence
- empty submits (with `allowEmptySubmit`) create user messages
- added `step` to tool invocations to be able to restore llm messages

## Requirements
- [x] append to existing assistant message when there was a client-side tool result
- [x] append to existing assistant message for multiple server side steps
- [x] track step count on tool invocation
- [x] map tool invocations with step counts back to multi-message llm format
- [x] do not call backend again if response is already complete
- [x] update maxSteps logic in useChat
- [x] id stability for last message
- [x] annotations should all go to the last message
- [x] persistence must work
- [x] single-message persistence must work
- [x] empty submit should add user message to guarantee user/assistant message ordering

## Tasks

- [x] implementation
   - [x] mapping from ui message to core messages
   - [x] process chat response updates
      - [x] keep ui message id when keeping last message
      - [x] append to annotations (stable annotations)
   - [x] persistence append helper: keep original message id when replacing
   - [x] useChat react
   - [x] useChat svelte
   - [x] useChat solid
   - [x] useChat vue
- [x] manual testing & example updates
   - [x] react
   - [x] vue
   - [x] solid
- [x] changeset


nicoalbanese (approved):

rclmenezes:
I think this PR introduced a bug in convertToCoreMessages! Specifically, this has made assistant messages with tool invocations + content go from:

```typescript
[
    {
        "role": "user",
        "content": "input asking for a tool"
    },
    {
        "role": "assistant",
        "content": "response"
    },
    {
        "role": "assistant",
        "content": [
            {
                "type": "tool-call",
                "toolCallId": "tooluse_5oQ2pIcgQlOhVeq8dDlmhw",
                "toolName": "toolName",
                "args": {
                    "foo": "bar"
                }
            }
        ]
    },
    {
        "role": "tool",
        "content": [
            {
                "type": "tool-result",
                "toolCallId": "tooluse_5oQ2pIcgQlOhVeq8dDlmhw",
                "toolName": "toolName",
                "result": {
                    "other": "stuff"
                }
            }
        ]
    }
]
```

To:

```typescript
[
    {
        "role": "user",
        "content": "input asking for a tool"
    },
    {
        "role": "assistant",
        "content": "response"
    },
    {
        "role": "assistant",
        "content": [
            {
                "type": "tool-call",
                "toolCallId": "tooluse_5oQ2pIcgQlOhVeq8dDlmhw",
                "toolName": "toolName",
                "args": {
                    "foo": "bar"
                }
            }
        ]
    },
    {
        "role": "tool",
        "content": [
            {
                "type": "tool-result",
                "toolCallId": "tooluse_5oQ2pIcgQlOhVeq8dDlmhw",
                "toolName": "toolName",
                "result": {
                    "other": "stuff"
                }
            }
        ]
    }
]
```

Unfortunately, this distinction matters for certain LLM vendors like Anthropic, which return messages like:

> An error occurred: The model returned the following errors: Your API request included an `assistant` message in the final position, which would pre-fill the `assistant` response. When using tools, pre-filling the `assistant` response is not supported.

nauxliu:
When there are multiple tool calls in the same assistant message, it seems to mix up the message‐tool sequence. In other words, I’m seeing that subsequent tool calls can get appended out of order, making the final assistant message’s flow confusing. It’s especially noticeable in scenarios with more than one tool call. It would be great that each tool call is appended in the correct order and that no steps are reordered unexpectedly.

Before:
image

After
CleanShot 2025-02-03 at 14 48 56@2x

PR #4621, opened 2025-01-31
feat(provider/perplexity): add support for return_images

kvnwolf:
Adds support for [`return_images`](https://docs.perplexity.ai/api-reference/chat-completions#body-return-images) to the [Perplexity Provider](https://sdk.vercel.ai/providers/ai-sdk-providers/perplexity)

> [!NOTE]
> As per @shaper's recommendation, I'm _temporarily_ adding the logic to extract and pass the options to the underlying request in `openai-compatible-chat-language-model.ts`, since the Perplexity provider extends it.

PR #4651, opened 2025-02-01
feat (provider/perplexity): add support for return_images

shaper:
Updated from #4621


shaper:
@kevinwolfcr I don't actually get any images back from Perplexity, but it seems likely it's because I'm not a "Tier 2 user". I assume it's working for you, and I've only changed things formatting-wise, the core logic is the same as your original PR.

lgrammel:
@kevinwolfcr can you confirm that the feature works for you as a Tier 2 user?

kevinwolfcr:
@shaper @lgrammel yes, I forgot to state that you must be part of tier 2. The images response is coming back, I am a tier 4 user.

lgrammel (approved):

PR #4752, opened 2025-02-06
feat: Add new google models

viktorlarsson:
New models added yesterday:
https://ai.google.dev/gemini-api/docs/models/gemini


viktorlarsson:
Nvm, i saw that it was merged but not released

PR #4777, opened 2025-02-08
fix (provider/mistral): correctly parse complex content type responses

lgrammel:


matthiasbayer:
Nice! Thanks for the fix.

PR #4783, opened 2025-02-08
fix (ai/core): Add null check for OTEL attributes

kdawgwilk:
My mind was boggle today when i learned that in JS `(typeof null === 'object') === true` So this helps protect from trying to access `'input' in null` which throws below

### DIFF
--- a/packages/ai/core/telemetry/select-telemetry-attributes.test.ts
+++ b/packages/ai/core/telemetry/select-telemetry-attributes.test.ts
@@ -99,6 +99,7 @@ it('should handle mixed attribute types correctly', () => {
       input: { input: () => 'input value' },
       output: { output: () => 'output value' },
       undefined: undefined,
+      someNull: null,
     },
   });
   expect(result).toEqual({
--- a/packages/ai/core/telemetry/select-telemetry-attributes.ts
+++ b/packages/ai/core/telemetry/select-telemetry-attributes.ts
@@ -11,7 +11,8 @@ export function selectTelemetryAttributes({
       | AttributeValue
       | { input: () => AttributeValue | undefined }
       | { output: () => AttributeValue | undefined }
-      | undefined;
+      | undefined
+      | null;
   };
 }): Attributes {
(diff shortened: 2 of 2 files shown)


lgrammel:
Can you provide a reproduction code example for the bug that this fixes?

kdawgwilk:
It's not necessarily a bug because OTEL doesn't like undefined or null attributes but some the attributes we were passing to our tracing lib langsmith ended up being null and caused a crash.  I figured since this was already handling undefined values adding null check wouldn't be too big of an issue.  I can try and pull together a [MRE](https://en.wikipedia.org/wiki/Minimal_reproducible_example).  I thought a failing test would be enough to cover it.

lgrammel:
We need to support OTEL not LangSmith. Passing `null` is valid according and supported by the OTEL schema / standard (pls correct if wrong), and there are legit use cases (you may want to pass null for some custom values some times). If LangSmith crashes, but it's bc they don't handle things correctly, it needs to be fixed on their end.

kdawgwilk:
Alright here is the MRE without a dependency on langsmith
```ts
import { ollama } from 'ollama-ai-provider';

const model = ollama('llama3.2');

const prompt = 'Tell a joke';
  model,
  prompt,
  experimental_telemetry: {
    isEnabled: true,
    metadata: {
      // @ts-expect-error
      nullValue: null,
      prompt,
    }
  },
});
  
console.log(result.text);
```

lgrammel:
I'm getting the following (expected) typescript error:

CleanShot 2025-02-10 at 09 17 16

`AttributeValue` comes from OTEL and does not support `null`: 

`export declare type AttributeValue = string | number | boolean | Array<null | undefined | string> | Array<null | undefined | number> | Array<null | undefined | boolean>;`

https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/common/Attributes.ts#L35

I'm trying to understand the request here. Do you want us to support `null` (which is not supported by OTEL)?

kdawgwilk:
Any update on a decision if this will merge?

aayush-kapoor:

Issue #4788, opened 2025-02-09
Add data streaming support for tools


BrianHung:
### Feature Description

Add the ability for tools to stream data during their execution. This allows tools to send intermediate data or progress updates while they are running, rather than just returning a final result.

The feature adds:
- Split DataStreamWriter into basic and full interfaces
- Add dataStream parameter to tool execution and onToolCall

### Use Cases

Progress updates during long-running tool operations.

### Additional context

https://github.com/vercel/ai-chatbot/blob/main/app/(chat)/api/chat/route.ts

lgrammel:
You can now use preliminary tool results for this: https://github.com/vercel/ai/pull/7987

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This is an API and behavior enhancement request rather than a failure of existing documented behavior.

### Evidence
- The issue requests a new capability for tools to stream intermediate data or progress updates during execution.
- It proposes API changes such as a dataStream parameter for tool execution and onToolCall.
- A maintainer notes that preliminary tool results now provide this capability via PR #7987.

PR #4789, opened 2025-02-09
feat: add data stream support to tools and use-chat

BrianHung:
#4788

### DIFF
--- a/packages/ai/core/generate-text/run-tools-transformation.ts
+++ b/packages/ai/core/generate-text/run-tools-transformation.ts
@@ -1,8 +1,10 @@
 import { LanguageModelV1StreamPart } from '@ai-sdk/provider';
 import { generateId } from '@ai-sdk/ui-utils';
+import { formatStreamPart } from '@ai-sdk/ui-utils';
 import { Tracer } from '@opentelemetry/api';
 import { ToolExecutionError } from '../../errors';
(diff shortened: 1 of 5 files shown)


packages/ai/core/generate-text/run-tools-transformation.ts
```diff
@@ -133,6 +135,18 @@ export function runToolsTransformation<TOOLS extends ToolSet>({
...
+        writeData(data: unknown) {
+          controller.enqueue(formatStreamPart('data', [data]));
+        },
+        writeMessageAnnotation(annotation: unknown) {
+          controller.enqueue(
+            formatStreamPart('message_annotations', [annotation]),
+          );
+        },
+      };
+
```
lgrammel, line 149: at this stage we do not have a data stream so this goes against the architecture of streamText. agree with the general idea though - but the integration will not be that simple (thought about it quite a bit)

BrianHung: Ah; yeah I was looking at the ai chatbot over the weekend and how it handles custom data. 

Realized being able to write to a custom data stream would be pretty useful for execute and onToolCall, especially for longer tool calls with an api that supports streams.

Tried to find a good DX but yeah couldn’t exactly find best way to integrate data stream into stream text.

aayush-kapoor:

PR #4840, opened 2025-02-11
Modified append-response-messages to prevent exceptions when messages are passed as an empty array

m-rlin:
I am using this function to convert from Core Message to UI Message, and I have modified the code so that no exception occurs even when an empty array is passed to the messages parameter.


aayush-kapoor:

PR #4911, opened 2025-02-17
docs: Update o1-mini capabilities in openai provider to match getting started with o1 page

sandonl:
There are currently conflicting docs between the tables in [OpenAI Providers](https://sdk.vercel.ai/providers/ai-sdk-providers/openai) and [Get Started with o1](https://sdk.vercel.ai/docs/guides/o1) with the former suggesting that using `o1-mini` is capable of object generation and tool usage whereas the latter suggesting it isn't. 

Updates the discrepancy to ensure consistency between these two pages


lgrammel:
@sandonl did you check which one is correct? i.e. if `o1-mini` supports tools / object generation?

nicoalbanese:
Confirmed this is correct. Thanks @sandonl.

nicoalbanese (approved):

PR #4913, opened 2025-02-17
fix (provider/google): Add support for executableCode in Gemini response

BrandonWeng:
Something seems to have changed with the search tool, as it's consistently returning executableCode in the response. SearchGrounding is pretty much unusable for us because of this. More than 90% of the responses return `print(google_search.search(queries=["<original search query>"]))`.

bweng-Cursor-2025-02-17-at-00 28 33@2x

Gemini's doc on Search Grounding, it sounds like search grounding implicitly enables other tools as well now.. 
bweng-Google Chrome-2025-02-17-at-00 59 15@2x

Related to this: https://github.com/vercel/ai/issues/3205


BrandonWeng:
I think they reverted the change

Issue #4744, opened 2025-02-06
useCompletion submit handler not clearing input


nicoalbanese:
### Description

`handleSubmit` supposed to automatically resets the input field but the input field is never reset.
Specifically for `ai/react`

lgrammel:
### Identify issue type

**Identified type: Bug**  
Confidence: High

Rationale: This describes current SDK behavior allegedly contradicting expected or documented behavior, resulting in incorrect UI state.

Evidence:
- The reporter states that `useCompletion` / `ai/react` `handleSubmit` is supposed to automatically reset the input field.
- They report the input field is never reset after submit.

PR #4917, opened 2025-02-17
fix (ui) : useCompletion submit handler not clearing input

JacobRozak:
quick fix for https://github.com/vercel/ai/issues/4744


lgrammel:
Thanks - this needs to be optional (opt in flag on useChat) since it breaks backwards compat, and we need the change for svelte, vue, solid as well

gr2m:
Closing after no follow-up on the requested scope changes. Bug stays tracked in #4744. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

Issue #4472, opened 2025-01-22
Add support for Amazon Bedrock prompt caching


Henderson11:
### Feature Description

Was just added to Amazon Bedrock on December 4th. Please add this feature. Relevent website is linked.
I see that anthropopic has implemented this feature, but bedrock has not yet
https://aws.amazon.com/about-aws/whats-new/2024/12/amazon-bedrock-preview-prompt-caching/
https://aws.amazon.com/blogs/aws/reduce-costs-and-latency-with-amazon-bedrock-intelligent-prompt-routing-and-prompt-caching-preview/

petergok:
+1 This would be huge

dmahapatro:
@lgrammel Do we have a timeline for this feature request? I am the Bedrock Specialist SA for Startups at AWS and I see lot of interest around this feature being available in the AI SDK asap.

shaper:
Looking at this, referencing https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

shaper:
A first version of this is live in v2.0.3. The docs at https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock will be updated to describe the new cache points feature in a few hours.

lgrammel:
### Identify issue type

**Identified type: Feature**  
Confidence: High

Rationale: The issue requests new provider capability/support for Amazon Bedrock prompt caching, not a failure of existing behavior.

Evidence:
- Title asks to “Add support for Amazon Bedrock prompt caching.”
- Description says Amazon Bedrock added prompt caching and requests adding this feature because Anthropic has it but Bedrock does not.
- Comments express interest and ask for a timeline for the feature request; maintainer later notes a first version is live.

PR #4955, opened 2025-02-20
feat (provider/amazon-bedrock): add support for cache points

shaper:
Part of #4472

- [x] tests
- [x] docs
- [x] changeset
- [x] streaming

Tool checkpoints [ref](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) are planned for later work.


packages/amazon-bedrock/src/bedrock-api-types.ts
```diff
@@ -35,13 +42,15 @@ export interface BedrockToolInputSchema {
...
 }
 
-export interface BedrockTool {
-  toolSpec: {
-    name: string;
-    description?: string;
-    inputSchema: { json: JSONObject };
-  };
-}
+export type BedrockTool =
```
lgrammel, line 45: This is semantically weird. Would it be cleaner to do the `|` where `BedrockTool` is used instead?

lgrammel: Or introduce a separate type?

packages/amazon-bedrock/src/bedrock-chat-prompt.ts
```diff
@@ -2,6 +2,7 @@ import { BedrockContentBlock } from './bedrock-api-types';
 
 export type BedrockMessagesPrompt = {
   system?: string;
+  isSystemCachePoint: boolean;
   messages: BedrockMessages;
 };
```
lgrammel, line 7: Why do we have a `BedrockMessagesPrompt` that is different from the API types? Ideally we'd only have the API types and remove a level of indirection?

packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts
```diff
@@ -17,12 +23,19 @@ import {
...
 
+function getCachePoint(
+  providerMetadata: LanguageModelV1ProviderMetadata | undefined,
+): BedrockCachePoint | undefined {
+  return providerMetadata?.bedrock?.cachePoint as BedrockCachePoint | undefined;
+}
+
 export function convertToBedrockChatMessages(
   prompt: LanguageModelV1Prompt,
 ): BedrockMessagesPrompt {
```
lgrammel, line 34: Can we convert to API types directly here?

lgrammel (approved):

packages/amazon-bedrock/src/bedrock-chat-prompt.ts
shaper: Good suggestion, remnant from the rework, I restructured to use api-types.

PR #4971, opened 2025-02-24
feat (ai): add anthropic reasoning

lgrammel:


nicoalbanese (approved):

packages/provider/src/language-model/v1/language-model-v1.ts
```diff
@@ -1,3 +1,4 @@
+import { P } from 'vitest/dist/chunks/environment.LoooBwUu';
```
shaper, line 1: @lgrammel accidental commit?

lgrammel: thanks good catch

PR #4978, opened 2025-02-25
Update types for Anthropic `computer_20250124` tool

ezhlobo:
A new Anthropic's tool `computer_20250124` came out with a new schema:

https://docs.anthropic.com/en/docs/agents-and-tools/computer-use#computer-tool

### DIFF
--- a/packages/anthropic/src/anthropic-tools.ts
+++ b/packages/anthropic/src/anthropic-tools.ts
@@ -245,6 +245,33 @@ function computerTool_20241022<RESULT>(options: {
   };
 }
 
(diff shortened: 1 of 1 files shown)


OBannon37:
This would also requiring enabling the new beta `betas.add('computer-use-2025-01-24');` in `packages/anthropic/src/anthropic-prepare-tools.ts`

ezhlobo:
@OBannon37 that is true, and I did that too within separate PR: https://github.com/vercel/ai/pull/4979

I decided to split them because for me those are different issues.

PR #4980, opened 2025-02-25
feat (provider/amazon-bedrock): Add reasoning support

### LINKED ISSUE #4977: Add anthropic reasoning support to bedrock
### Feature Description

Add anthropic reasoning support to bedrock. `vertexAnthropic` supports reasoning via:

```typescript
providerOptions: {
    anthropic: {
        thinking: { type: 'enabled', budgetTokens: 1024 }
     }
},
```

Adding support to the bedrock provider as well would be beneficial

### Use Cases

Having reasoning support for anthropic claude models in bedrock would be nice.

### Additional context

Based off of my two second look, you can add an additional field like below to the `BedrockChatSettings` and have the reasoning enabled:

```
{
	"additionalModelRequestFields": {
		"reasoning_config": {
			"budget_tokens": 1024,
			"type": "enabled"
		}
	}
}
```

I took a quick stab at this in #4980 for reference. My next comment has some testing notes.

Und3rf10w:
Closes #4977

This adds reasoning support for Claude models in Amazon Bedrock.

I have tested these changes in my environment, and confirmed that we receive reasoning in both streaming and non-streaming configurations.

In streaming and non-streaming, both the signature and reasoning text are properly extracted. You may want to convert this to use `providerOptions` instead of relying on `additionalModelRequestFields` in `BedrockChatSettings`, but I don't know the best practice for that.


shaper:
Thank you! I will take a look at this soon.

Und3rf10w:
After some testing (from e7aeec0), it looks like something goes wrong when you have tool use involved with reasoning enabled.

I can at least get this error back when I'm doing a generation with a tool enabled, but not sure how to resolve this:

> The model returned the following errors: messages.1.content.0.type: Expected `thinking` or `redacted_thinking`, but found `text`. When `thinking` is enabled, a final `assistant` message must start with a thinking block (preceeding the lastmost set of `tool_use` and `tool_result` blocks). We recommend you include thinking blocks from previous turns. To avoid this requirement, disable `thinking`.

david-arteaga:
For what it's worth I took this patch and manually applied it locally and it works with sonnet 3.7 with both `generateText` and `streamText`

shaper:
This is on my radar and I aim to get to it soon, thank you for your continued work here and contributions. We plan to work with you to get this landed as soon as we can.

Und3rf10w:
This has literally been my experience with trying to write this, I swear:

image

I feel like every time I think I have it done right, I do some testing, then experience some random edge case and find I did something wrong.

---

@shaper, do you want to populate `reasoningDetails[]` on a response from `doGenerate` that includes a tool invocation?

For example, right now, as of [468a750](https://github.com/vercel/ai/pull/4980/commits/468a750e032451459e4aed463f29bd61db866d25), if we call `generateText` without any tools, then the reasoningDetails comes back populated:

image

However, if we call `generateText` with tools, then it will be empty, but can be extracted from the steps or response messages:

image


And later in the response.messages:

image

I genuinely have no idea how that's supposed to return, but something important to note.

Und3rf10w:
Commit [ac55421](https://github.com/vercel/ai/pull/4980/commits/ac5542186a87b8231f9df6fdb037424e4ff297b7) resolves the aforementioned issue, returning reasoning details and reasoning in doGenerate responses with tool invocations.

packages/ai/core/generate-text/generate-text.ts
```diff
@@ -469,9 +469,35 @@ A function that attempts to repair a tool call that failed to parse.
...
+                ...newReasoningDetails,
+              ];
+            } else {
+              currentReasoningDetails = [
+                ...currentReasoningDetails,
+                newReasoningDetails,
+              ];
+            }
+          }
+        }
```
lgrammel, line 500: Why is this necessary? my instinct says this should be solved in the provider. Is the bedrock provider returning duplicated reasoning?

packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
```diff
@@ -994,22 +1114,71 @@ describe('doGenerate', () => {
...
+    // Add reasoning content
+    if (reasoningContent) {
+      if (Array.isArray(reasoningContent)) {
+        reasoningContent.forEach(rc => {
+          contentBlocks.push(rc as any);
+        });
+      } else {
+        contentBlocks.push(reasoningContent as any);
+      }
+    }
```
lgrammel, line 1173: Ideally order should be kept - is there a way to map the content instead?

packages/ai/core/generate-text/generate-text.ts
Und3rf10w: I have that here because let's say you pass a tool to the aisdk with an `execute()` method. That means you'll get two responses from the sdk. The first response will have the currentReasoningDetails, but that isn't extracted out to the generateText response from the provider, because the SECOND run (after the tool result is done), will overwrite the reasoning details. This code just appends the reasoning details every time this is called for that interaction, ensuring we always get the reasoning

packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
Und3rf10w: fdb87b5

Und3rf10w:
Docs added and I tweaked the way the reasoning config is set. Before, I had it being set via the `additionalModelRequestFields`, e.g.:

```json
       additionalModelRequestFields: {
          reasoning_config: {
            type: 'enabled',
            budget_tokens: 1024
          }
        }
```

Now I added support for this via `providerOptions`, but behind the scenes it [injects it into`additionalModelRequestFields`](https://github.com/vercel/ai/pull/4980/commits/9988ee868b7eaef3a9b0e6abbe85a744d2268ea7#diff-0008edd898f4224ec017b5e1ffd37088adadec1ba41b61c5934fbe1ee3571da3R121-R179)

So you would pass in:

```json
providerOptions: {
  bedrock: {
    reasoning_config: {
      type: 'enabled',
      budget_tokens: 1024
    }
  }
}
```

.changeset/two-mayflies-shave.md
```diff
@@ -0,0 +1,8 @@
+---
+'@ai-sdk/amazon-bedrock': minor
+'ai': minor
+---
+
+feat (ai/core): Add all reasoning parts in a generate text call
+
+feat (providers/amazon-bedrock): Add reasoning support to amazon-bedrock
```
lgrammel, line 8: remove the ai part

content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
```diff
@@ -316,6 +316,38 @@ console.log(
 // }
 ```
 
+## Reasoning
+
+Amazon Bedrock has reasoning support for the `claude-3-7-sonnet-20250219` model.
+
+You can enable it using the `reasoning_config` provider option and specifying a thinking budget in tokens (minimum: `1024`, maximum: `64000`).
+
+````ts
```
lgrammel, line 325: three backticks

content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
```diff
@@ -362,7 +394,7 @@ using the `.embedding()` factory method.
 
 ```ts
 const model = bedrock.embedding('amazon.titan-embed-text-v1');
-```
+````
```
lgrammel, line 397: revert

packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
```diff
@@ -994,22 +1112,76 @@ describe('doGenerate', () => {
...
+              name: item.name,
+              input: item.input,
+            },
+          };
+        }
+      } else if ('reasoningContent' in item) {
+        return item;
+      }
+      return item;
+    });
```
lgrammel, line 1176: this looks a bit unclear (mixing `in` w/ type). why is this necessary? can this be solved in a switch? if not, consider adding some comments

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -508,8 +650,19 @@ const BedrockStreamSchema = z.object({
...
+            reasoningContent: z.object({
+              signature: z.string(),
+            }),
+          }),
+          z.object({
+            reasoningContent: z.object({ data: z.string() }),
+          }),
         ])
-        .nullish(),
+        .optional(),
```
lgrammel, line 665: prefer nullish over optional

lgrammel (approved):

content/providers/01-ai-sdk-providers/08-amazon-bedrock.mdx
Und3rf10w: lol that was definitely my editor, let me fix that

packages/ai/core/generate-text/generate-text.ts
Und3rf10w: Okay, do you want me to extract the reasoning from steps in the model provider, or would you like for me to leave as is and wait for an upgrade `ai` package?

lgrammel: leave as is and wait for `ai` package upgrade. the extraction would be in the user code

packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
Und3rf10w: [d85704e](https://github.com/vercel/ai/pull/4980/commits/d85704e65a68d8341f67440559037fc46a1c4d2f)

I was under the assumption that the content needed remapping, that's not the case

packages/ai/core/generate-text/generate-text.ts
Und3rf10w: I wonder if this was related to #4986...

Und3rf10w:
I will add an example shortly. I will try to add a few different ones, ones with reasoning and tool calling and one without, just to cover all the cases.

Und3rf10w:
@lgrammel, examples added in [8eed8cd](https://github.com/vercel/ai/pull/4980/commits/8eed8cdc3c67d213ab1aa7e3b0503a2e7e43c892). I have tested each and they work as expected. Please let me know if there's anything I can provide.

examples/ai-core/src/generate-text/amazon-bedrock-reasoning-chatbot.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const messages: CoreMessage[] = [];
+
+async function main() {
+  while (true) {
+    const userInput = await terminal.question('You: ');
+    messages.push({ role: 'user', content: userInput });
+
+    const { steps, response } = await generateText({
+      model: bedrock('anthropic.claude-3-7-sonnet-20250219-v1:0'),
```
shaper, line 20: Does this work for you with the given model id? I am still wrestling with Bedrock always telling me I need to use an ARN for some models and this is one, even across several accounts I have access to test.

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -457,6 +573,13 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
   }
 }
 
+const BedrockReasoningConfigOptionsSchema = z
+  .object({
+    type: z.union([z.literal('enabled'), z.literal('disabled')]),
+    budget_tokens: z.number().min(1024).max(64000).nullish(),
```
shaper, line 579: Where did you find 6400? On initial search I only see https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-37.html which doesn't give the upper bound, but I probably haven't found the right thing.

examples/ai-core/src/generate-text/amazon-bedrock-reasoning-chatbot.ts
Und3rf10w: I have to use `us.anthropic.claude-3-7-sonnet-20250219-v1:0` with Cross Region Inference enabled actually.

You MIGHT have that work if have it set up without CRIS, but try `us.anthropic.claude-3-7-sonnet-20250219-v1:0`

shaper: Yes, I have to add the `us.` prefix as well. If it doesn't work for you either without it, I think we should update the docs/examples to have that. I'll do it in another small change shortly.

shaper (commented):
This looks good, I am just making a couple of small minor edits (I just pushed one commit).

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
Und3rf10w: I pulled that from here: https://docs.anthropic.com/en/docs/about-claude/models/all-models#model-comparison-table

shaper:
Thanks, I'm done with edits -- I'll let @lgrammel do a final test and push the merge button, to make sure the examples and model ids work for him.

I think this work has turned up a potential gap in tool call testing for the language model, but if so the gap was already there, I'm making a note to look into it separately.

Thank you for soldiering through a very obtuse set of work, reasoning is not as simple to implement as it seems like it would be at face value!

shaper (approved):

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
lgrammel: I'd be in favor of removing the ranges, they will most likely change in the future.

lgrammel:
@Und3rf10w the reasoning chatbot stream text example is not working for me, it does not return the final answer:

```
❯ pnpm tsx src/stream-text/amazon-bedrock-reasoning-chatbot
You: what is the weather in paris

Assistant: The user is asking for the weather in Paris. I can use the `weather` function to answer this question.

The `weather` function requires a parameter called `location`. In this case, the location is "Paris".

I'll call this function to get the weather in Paris.I'll check the current weather in Paris for you.

You:
```

lgrammel:
@Und3rf10w for anthropic we add reasoning tokens to max tokens. please do the same here to prevent:

```
APICallError [AI_APICallError]: The model returned the following errors: max_tokens must be greater than thinking.budget_tokens
```

(check out the anthropic language model for an example)

lgrammel:
@Und3rf10w another error i'm getting with generate text reasoning chatbot when i specify maxTokens and sya "what is the weather in paris?":

```
APICallError [AI_APICallError]: The model returned the following errors: temperature may only be set to 1 when thinking is enabled.
```

lgrammel (changes requested):
Discovered bugs need to be resolved

Und3rf10w:
> @Und3rf10w for anthropic we add reasoning budget to max tokens. please do the same here to prevent:
> 
> ```
> APICallError [AI_APICallError]: The model returned the following errors: max_tokens must be greater than thinking.budget_tokens
> ```
> 
> (check out the anthropic language model for an example)

@shaper, @lgrammel, Okay, I think i've tracked this down, but I'm having some trouble here.

Basically, I am trying to do a tool call, and we get a tool result.

The model gets reinvoked once we get a tool result, however, the reinvocation doesn't keep the same parameters:

I think this was a regression with 7a157d8.

After a tool call and with that logic removed:

image


After a tool call with that logic in place:

image


I think this prevents us from resolving `isThinking` correctly. In this case, we remake a call to the model, but since reasoning is enabled, i guess stream-text in the background sets the temperature to 0, which breaks the flow.

I have added [4ac00d3](https://github.com/vercel/ai/pull/4980/commits/4ac00d3f11987119272abf8eb27e3d55a4f794b5), which removes this, and slightly tweaked the null check, and now the examples work properly for me. Please reverify.

I will review the budget configs now

Und3rf10w:
@lgrammel [4ac00d3](https://github.com/vercel/ai/pull/4980/commits/4ac00d3f11987119272abf8eb27e3d55a4f794b5) also fixed the generateText temperature error, although you will still receive a warning.

Und3rf10w:
> @Und3rf10w for anthropic we add reasoning budget to max tokens. please do the same here to prevent:
> 
> ```
> APICallError [AI_APICallError]: The model returned the following errors: max_tokens must be greater than thinking.budget_tokens
> ```
> 
> (check out the anthropic language model for an example)

Can you provide an example where this happened please? I was under the impression this was already happening here:

https://github.com/vercel/ai/blob/4ac00d3f11987119272abf8eb27e3d55a4f794b5/packages/amazon-bedrock/src/bedrock-chat-language-model.ts#L141-L153

Which for the most part mimics https://github.com/vercel/ai/blob/main/packages/anthropic/src/anthropic-messages-language-model.ts#L188-L189

lgrammel:
Tests are passing now:

- [x] generateText reasoning
- [x] generateText reasoning chatbot with tool call
- [x] streamText reasoning
- [x] streamText reasoning chatbot with tool call

Und3rf10w:
@lgrammel, sorry I made one small modification to the examples just to make the requests more reasonable, and noticed that the bedrock chatbot didn't behave as expected, so I added maxSteps there

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -113,13 +113,65 @@ export class BedrockChatLanguageModel implements LanguageModelV1 {
...
+      // Add them to additional model request fields
+      // Add reasoning config to additionalModelRequestFields
+      this.settings.additionalModelRequestFields = {
+        ...this.settings.additionalModelRequestFields,
+        reasoning_config: { ...reasoningConfigOptions.data },
+      };
+    }
+
+    // Remove temperature if thinking is enabled
+    if (isThinking && inferenceConfig.temperature !== null) {
```
lgrammel, line 156: what if temperature is undefined? why `!==` vs `!=`

lgrammel (approved):
Looks good! had some last minor nits. Will land once addressed

PR #5000, opened 2025-02-26
fix (provider/anthropic): add model setting to allow omitting reasoning content from model requests

shaper:
For the included example script with a tool call, claude 3 opus produces thinking-tagged content below which, when extracted with middleware into a `reasoning` part, produces `thinking` content. Anthropic API then throws an error on the request as it enforces a signature requirement on `thinking` content.

This change adds a `sendReasoning` model setting, defaulting to `true`. When set to `false`, we omit reasoning content from inclusion in model requests.

```
<thinking>
The user is asking for the weather in a specific location, San Francisco. The weather tool is directly relevant for answering this query.

Looking at the required parameters for the weather tool:
- location (required): The user directly provided the location of "San Francisco"

Since the required location parameter is present, we can proceed with calling the weather tool to get the forecast for San Francisco. No other tools are needed.
</thinking>
```

### DIFF
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
@@ -251,12 +251,18 @@ export function convertToAnthropicMessagesPrompt({
               }
 
               case 'reasoning': {
-                anthropicContent.push({
-                  type: 'thinking',
-                  thinking: part.text,
-                  signature: part.signature!,
-                  cache_control: cacheControl,
-                });
+                // omit reasoning content if no signature is present. middleware
+                // could have extracted it from an older model's text content
+                // with no associated signature, and anthropic API
+                // requires signature presence even for older models.
+                if (part.signature) {
+                  anthropicContent.push({
+                    type: 'thinking',
+                    thinking: part.text,
+                    signature: part.signature,
+                    cache_control: cacheControl,
+                  });
+                }
                 break;
               }
 
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
@@ -266,7 +266,8 @@ export function convertToAnthropicMessagesPrompt({
                 } else {
                   warnings.push({
                     type: 'other',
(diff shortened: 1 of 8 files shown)


lgrammel (approved):

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -251,12 +251,18 @@ export function convertToAnthropicMessagesPrompt({
...
+                // with no associated signature, and anthropic API
+                // requires signature presence even for older models.
+                if (part.signature) {
+                  anthropicContent.push({
+                    type: 'thinking',
+                    thinking: part.text,
+                    signature: part.signature,
+                    cache_control: cacheControl,
+                  });
+                }
```
lgrammel, line 265: We might want to add warnings, since we are omitting information some users might wonder about.

lgrammel (changes requested):
The current fix has downsides:
- silent swallowing of potential errors makes debugging hard
- removing fix is backwards compat breaking

Considering that the bug is for a very specific use case, I'd prefer an alternative design where we introduce a `sendThinking` (or similar) model option that defaults to true. When false, no thinking is forwarded.

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
shaper: Done.

shaper:
> The current fix has downsides:
> 
> * silent swallowing of potential errors makes debugging hard
> * removing fix is backwards compat breaking
> 
> Considering that the bug is for a very specific use case, I'd prefer an alternative design where we introduce a `sendThinking` (or similar) model option that defaults to true. When false, no thinking is forwarded.

Done. One note, the `signature: part.signature!` build-time assert in conversion is knowably not always valid at runtime, e.g. the example in this PR where we use middleware to extract `thinking` tags and there is no accompanying signature data. Even if one could view this as a provider-side model issue as a non-reasoning model is outputting `thinking` tags.

This makes me still wonder if we should also omit if signature is nullish, since it is in violation of the request spec.

We can consider the above for a later change as the current one is opt-in and addresses the immediate issue.

lgrammel (approved):

PR #5003, opened 2025-02-27
feat: add MCP tool conversion util

iteratetograceness:


packages/ai/package.json
```diff
@@ -63,6 +63,7 @@
     "@ai-sdk/provider-utils": "2.1.10",
     "@ai-sdk/react": "1.1.18",
     "@ai-sdk/ui-utils": "1.1.16",
+    "@modelcontextprotocol/sdk": "^1.6.0",
```
lgrammel, line 66: Hm this is a major challenge because their dependencies match our, which can lead to dependency version conflicts and nasty bugs: https://www.npmjs.com/package/@modelcontextprotocol/sdk?activeTab=dependencies

lgrammel: Will think about options how to address this. One is to move mcp tools to a separate pkg, another is to use peer dependencies.

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
```
lgrammel, line 1: strange that the imports are with `.js` - is this intentional?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+  options?: SSEClientTransportOptions;
+}
+type McpServerConfig = McpStdioServerConfig | McpSSEServerConfig;
+
+interface CreateMcpToolsConfig {
+  servers: Record<string, McpServerConfig>;
+  toolCallOptions?: RequestOptions;
+}
+
+type McpCallToolInputSchema = z.ZodType<any, any, any> | Schema<any>;
```
lgrammel, line 38: Does the `any` affect type inference down the line (when the mcp tool is used with generateText/streamText)?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+          env: {
+            ...process.env,
+            ...serverConfig.env,
+          } as Record<string, string>,
+        });
+
+  const client = new Client(
+    {
+      name: `${name}-client`,
+      version: '0.1.0',
```
lgrammel, line 69: what does the version indicate here?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+      inferParameters<TToolSchemas[K]['input']>,
+      CallToolResult
+    >;
+  };
+  clients: Record<string, Client>;
+}
+
+// Based off: https://github.com/dnakov/claude-code/blob/da716d77b251868918bb5776377f85a18a47ffdf/src/services/mcpClient.ts#L224
+async function connectToServer(
+  name: string,
```
lgrammel, line 52: what is name used for?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+    {
+      name: `${name}-client`,
+      version: '0.1.0',
+    },
+    {
+      capabilities: {},
+    },
+  );
+
+  const CONNECTION_TIMEOUT_MS = 5000;
```
lgrammel, line 76: this could be a configurable option

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+          } as Record<string, string>,
+        });
+
+  const client = new Client(
+    {
+      name: `${name}-client`,
+      version: '0.1.0',
+    },
+    {
+      capabilities: {},
```
lgrammel, line 72: what can be set here?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+    },
+  );
+
+  const CONNECTION_TIMEOUT_MS = 5000;
+  const connectPromise = client.connect(transport);
+  const timeoutPromise = new Promise<never>((_, reject) => {
+    const timeoutId = setTimeout(() => {
+      reject(
+        new Error(
```
lgrammel, line 81: please use `AISDKError` for now. we might need a special subclass at some point

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+ * Read: https://mswjs.io/docs/basics/handling-websocket-events
+ * Also add new example(s) for MCP stdio (and SSE?)
+ */
+
+/**
+ * Overall improvement of error handling
+ */
+
+export async function createMcpTools<
+  TToolSchemas extends McpToolSchemaMap = {},
```
lgrammel, line 150: I use upper case for generics e.g. `TOOL_SCHEMAS`

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+type McpServerConfig = McpStdioServerConfig | McpSSEServerConfig;
+
+interface CreateMcpToolsConfig {
+  servers: Record<string, McpServerConfig>;
+  toolCallOptions?: RequestOptions;
+}
+
+type McpCallToolInputSchema = z.ZodType<any, any, any> | Schema<any>;
+type McpToolSchemaMap = Record<string, { input: McpCallToolInputSchema }>;
+interface McpToolset<TToolSchemas extends McpToolSchemaMap = {}> {
```
lgrammel, line 40: We have a type `ToolSet` internally, so maybe change to `McpToolSet` (and it should match that type somewhat)

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+    tools: {} as McpToolset<TToolSchemas>['tools'],
+    clients: {},
+  };
+
+  const clientConnections = await Promise.allSettled(
+    Object.entries(config.servers).map(async ([name, serverConfig]) => {
+      const client = await connectToServer(name, serverConfig);
+      return { name, client };
+    }),
+  );
```
lgrammel, line 162: I'm confused - does a single McpToolSet not match a single client/server connection?

My understanding was 1 McpToolSet <> 1 McpClient <> 1 McpServer - is that incorrect?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,209 @@
...
+        } catch (error) {
+          const errorMessage =
+            error instanceof Error ? error.message : String(error);
+          throw new Error(
+            `Failed to list tools for server ${name}: ${errorMessage}`,
+          );
+        }
+      } else {
+        throw new Error(`Failed to connect to server: ${result.reason}`);
+      }
```
lgrammel, line 204: ai sdk errors

lgrammel:
@iteratetograceness actually would it be possible to call the MCP websocket api directly without their library? if so, how hard would it be? that would solve the dependency issues and allow us to include the MCP tool directly in the `ai` pkg. It would also be more aligned with our general direct api calling approach (avoid libraries).

packages/ai/package.json
iteratetograceness: Looking into the solution you proposed most recently around calling web socket API directly!

packages/ai/core/tool/mcp.ts
iteratetograceness: The conversion tool accepts multiple server configs, spawns a client for each one, and combines all available tools; but I'm open to being opinionated about focusing on a single server<>client relationship - lmk what you think!

iteratetograceness:
> would it be possible to call the MCP websocket api directly without their library? if so, how hard would it be? that would solve the dependency issues and allow us to include the MCP tool directly in the ai pkg. It would also be more aligned with our general direct api calling approach (avoid libraries).

Thinking about this: seems like what we want is a custom Client and Transport; the trade-off here being we would essentially be copying over chunks of their code that we want to use

packages/ai/core/tool/mcp.ts
iteratetograceness: yeah that was the auto import!

packages/ai/core/tool/mcp.ts
iteratetograceness: `version of an MCP implementation`

we could expose as config option to users!

packages/ai/core/tool/mcp.ts
iteratetograceness: this is used to name the client that will connect to the server!

packages/ai/core/tool/mcp.ts
iteratetograceness: https://github.com/modelcontextprotocol/typescript-sdk/blob/170e656dcbb154b1e1723ba7dbb850d73010fbe1/src/types.ts#L196

lgrammel:
> Thinking about this: seems like what we want is a custom Client and Transport; the trade-off here being we would essentially be copying over chunks of their code that we want to use

I usually re-develop from scratch against APIs to make sure the code is up to our standards.

packages/ai/core/tool/mcp.ts
iteratetograceness: Follow-up: converted to enforcing a single server; users can invoke the tool N times for every server they'd like to pull tools from!

examples/ai-core/package.json
```diff
@@ -48,5 +49,6 @@
     "@types/node": "20.11.20",
     "tsx": "4.19.2",
     "typescript": "5.6.3"
-  }
+  },
+  "type": "module"
```
lgrammel, line 53: why is this needed?

examples/ai-core/src/mcp/agent.ts
```diff
@@ -0,0 +1,78 @@
...
+    {
+      server: {
+        type: 'stdio',
+        command: 'node',
+        args: ['src/mcp/dist/server.js'],
+      },
+      customClient: client,
+    },
+    toolSchemas,
+  );
```
lgrammel, line 44: ideally I'd imagine an api like:

```
const myTools = createMCPTools({
  type: "stdio",
  command: 'node',
  args: ['src/mcp/dist/server.js'],
  tools: {
    'get-pokemon': {
      input: z.object({ name: z.string() }),
    },
  }
})
```

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+    ) => Promise<CallToolResult>;
+  };
+} & {
+  [key: string]: Tool<any, CallToolResult> & {
+    execute: (
+      args: any,
+      // options?
+    ) => Promise<CallToolResult>;
+  };
+};
```
lgrammel, line 98: we have this in toolset / tools, but slightly different. would be good to reuse what we have

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+  type: 'stdio';
+}
+interface McpSSEServerConfig {
+  type: 'sse';
+  url: string;
+  options?: {
+    // authProvider?: OAuthClientProvider;
+    eventSourceInit?: EventSourceInit;
+    requestInit?: RequestInit;
+  };
```
lgrammel, line 39: pull up eventSourceInit / requestInit (or remove for now)

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * An interface that defines the minimum required MCP client properties.
+ * Used when endusers provide their own MCP Client for more granular control.
+ */
+interface McpClientInterface {
+  listTools(
+    params?: ListToolsRequest['params'],
+    options?: RequestOptions,
+  ): Promise<ListToolsResult>;
```
lgrammel, line 57: do we need the tool listing? or can we limit ourselves to tool calling for the mvp? (since the user provides tool schemas to begin with)

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+  callTool(
+    params: CallToolRequest['params'],
+    resultSchema?:
+      | typeof CallToolResultSchema
+      | typeof CompatibilityCallToolResultSchema,
```
lgrammel, line 62: why 2?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+
+export async function createMcpTools<TOOL_SCHEMAS extends ToolSchemas = {}>(
+  config: AdapterConfig,
+  toolSchemas: TOOL_SCHEMAS,
+): Promise<AdapterReturnType<TOOL_SCHEMAS>> {
+  const tools: Record<string, Tool> = {};
+  const { client, cleanup } = await setupClient(config);
+
+  try {
+    const listToolsResult = await client.listTools();
```
lgrammel, line 108: if we have the tools from `toolSchemas`, do we need this? (maybe yes for more dynamic use cases, but want to understand better)

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+
+    for (const { name, description, inputSchema } of listToolsResult.tools) {
+      const parameters = toolSchemas[name]
+        ? toolSchemas[name].input
+        : jsonSchema(inputSchema as JSONSchema7);
```
lgrammel, line 113: hm this might lead to errors where toolSchemas is wrong

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+            },
+            CallToolResultSchema,
+            config.toolCallOptions,
+          );
+          const parsedResult = CallToolResultSchema.parse(result);
+          return parsedResult as CallToolResult;
+        },
+      };
+
+      tools[name] = toolWithExecute;
```
lgrammel, line 138: could this be just

```
tools[name] = tool({
   ...
})
```

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+}
+
+async function setupClient(config: AdapterConfig) {
+  if (config.customClient)
+    return {
+      client: config.customClient,
+      cleanup: async () => {
+        // noop - no clean up for custom clients
+        return Promise.resolve();
```
lgrammel, line 161: is promise.resolve needed?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,227 @@
...
+      },
+    };
+
+  try {
+    const client = new SimpleMcpClient(
```
lgrammel, line 166: rename to `McpClient`?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+ * It is a custom implementation of the MCP Client (derived from Protocol) class.
+ *
+ * Not supported:
+ * - Client options (e.g. sampling, roots) as they are not needed for tool conversion
+ * - Accepting notifications
+ */
+export class SimpleMcpClient {
+  private _transport: Transport | undefined;
+  private _requestMessageId = 0;
+  private _responseHandlers: Map<
```
lgrammel, line 41: pls remove `_` prefix

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+ * primarily for tool conversion between MCP<>AI SDK.
+ *
+ * It is a custom implementation of the MCP Client (derived from Protocol) class.
+ *
+ * Not supported:
+ * - Client options (e.g. sampling, roots) as they are not needed for tool conversion
+ * - Accepting notifications
+ */
+export class SimpleMcpClient {
+  private _transport: Transport | undefined;
```
lgrammel, line 39: must always be defined, set in constructor to simplify flows

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+  private requestTimeoutMs: number;
+  private _timeoutInfo: Map<number, TimeoutInfo> = new Map();
+  private _serverCapabilities?: ServerCapabilities;
+
+  constructor(private _clientInfo: Implementation) {
+    this.connectionTimeoutMs = _clientInfo.connectionTimeoutMs || 6000;
+    this.requestTimeoutMs = _clientInfo.requestTimeoutMs || 3000;
+  }
+
+  async connect(transport: Transport): Promise<void> {
```
lgrammel, line 55: merge into constructor?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+  private _responseHandlers: Map<
+    number,
+    (response: JSONRPCResponse | Error) => void
+  > = new Map();
+  private connectionTimeoutMs: number;
+  private requestTimeoutMs: number;
+  private _timeoutInfo: Map<number, TimeoutInfo> = new Map();
+  private _serverCapabilities?: ServerCapabilities;
+
+  constructor(private _clientInfo: Implementation) {
```
lgrammel, line 50: prefer object params

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+          name: 'McpClientError',
+          message: 'Server sent invalid initialize result',
+        });
+      }
+
+      if (!SUPPORTED_PROTOCOL_VERSIONS.includes(result.protocolVersion)) {
+        throw new AISDKError({
+          name: 'McpClientError',
+          message: `Server's protocol version is not supported: ${result.protocolVersion}`,
+        });
```
lgrammel, line 83: would be good to introduce a dedicate `MCPClientError`class (see examples for specific ai sdk errors)

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+  async request<T extends ZodType<object>>(
+    request: Request,
+    resultSchema: T,
+    options?: RequestOptions,
+  ): Promise<z.infer<T>> {
+    return new Promise((resolve, reject) => {
+      if (!this._transport) {
+        reject(new Error('Not connected'));
+        return;
+      }
```
lgrammel, line 111: there should always be a transport (or object creation should fail)

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+
+  async close(): Promise<void> {
+    await this._transport?.close();
+    this._onclose();
+  }
+
+  async request<T extends ZodType<object>>(
+    request: Request,
+    resultSchema: T,
+    options?: RequestOptions,
```
lgrammel, line 105: use param object

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+  async listTools(
+    params?: ListToolsRequest['params'],
+    options?: RequestOptions,
+  ): Promise<ListToolsResult> {
+    if (!this._serverCapabilities?.tools) {
+      throw new AISDKError({
+        name: 'McpClientError',
+        message: `Server does not support tools (required for tools/list)`,
+      });
+    }
```
lgrammel, line 175: we should not have that flag. always assume tools are supported (or the server is useless for the ai sdk)

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,318 @@
...
+
+    const jsonrpcNotification: JSONRPCNotification = {
+      ...notification,
+      jsonrpc: '2.0',
+    };
+
+    await this._transport.send(jsonrpcNotification);
+  }
```
lgrammel, line 219: do we need this?

packages/ai/core/tool/mcp/types.ts
```diff
@@ -0,0 +1,251 @@
...
+});
+export const CompatibilityCallToolResultSchema = CallToolResultSchema.or(
+  ResultSchema.extend({
+    toolResult: z.unknown(),
+  }),
+);
+export type CallToolResult = z.infer<typeof CallToolResultSchema>;
+export type CompatibilityCallToolResult = z.infer<
+  typeof CompatibilityCallToolResultSchema
+>;
```
lgrammel, line 251: is all of this needed or can parts be removed? do we need zod schemas for everything?

examples/ai-core/package.json
iteratetograceness: running the example i added errors without this! one of the newly introduced deps (pkce-challenge) lacks cjs compatibility

packages/ai/core/tool/mcp.ts
iteratetograceness: lmk your thoughts on the following: i had to add the explicit `execute` to enforce specific signature of the execute method!

```
type ToolSchemas = Record<string, { parameters: Parameters }>;

type McpToolSet<TOOL_SCHEMAS extends ToolSchemas = {}> =
  ToToolsWithDefinedExecute<{
    [K in keyof TOOL_SCHEMAS]: Tool<
      inferParameters<TOOL_SCHEMAS[K]['parameters']>,
      CallToolResult
    > & {
      execute: (
        args: inferParameters<TOOL_SCHEMAS[K]['parameters']>,
        options: ToolExecutionOptions,
      ) => PromiseLike<CallToolResult>;
    };
  }>;
```

packages/ai/core/tool/mcp.ts
iteratetograceness: while we do ask the user to provide tool schemas, i thought it may be worth pulling all available tools from the server; happy to pivot away from introducing this initially!

packages/ai/core/tool/mcp.ts
iteratetograceness: copied over directly from `mpc` ts sdk! seems like this was due to supporting backwards compatibility

packages/ai/core/tool/mcp.ts
iteratetograceness: > yes for more dynamic use cases

exactly! similar to the comment response above, happy to pivot away from this initially and add back based on user sentiment

packages/ai/core/tool/mcp.ts
iteratetograceness: very similar to the dynamic nature of fetching all available tools, regardless of whether the user explicitly defined in toolSchemas; would love to chat about this further

packages/ai/core/tool/mcp.ts
iteratetograceness: ah, a left over change from my battle with typescript signatures on `execute`

packages/ai/core/tool/mcp/client.ts
iteratetograceness: brilliant!

iteratetograceness: ope, silly me: we can't have an async constructor - lmk if i'm missing something!

packages/ai/core/tool/mcp/client.ts
iteratetograceness: yes! this is needed to finalize the handshake w/ the server after we send the `initialize` request:

```
await this.notification({
        method: 'notifications/initialized',
});
```

packages/ai/core/tool/mcp/types.ts
iteratetograceness: re-reviewing types to see what's _really_ needed!

packages/ai/core/tool/tool.ts
```diff
@@ -3,7 +3,7 @@ import { z } from 'zod';
 import { ToolResultContent } from '../prompt/tool-result-content';
 import { CoreMessage } from '../prompt/message';
 
-type Parameters = z.ZodTypeAny | Schema<any>;
+export type Parameters = z.ZodTypeAny | Schema<any>;
```
lgrammel, line 6: Rename to `ToolParameters` if exported?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,151 @@
...
+          return result;
+        },
+      });
+
+      tools[name] = toolWithExecute;
+    }
+
+    return {
+      toolSet: tools as McpToolSet<TOOL_SCHEMAS>,
+      cleanup,
```
lgrammel, line 86: rename to `close`?

packages/ai/core/tool/mcp.ts
```diff
@@ -0,0 +1,151 @@
...
+      },
+    };
+  } catch (error) {
+    throw new AISDKError({
+      name: 'McpClientSetupError',
+      message: `Failed to connect to ${config.transport.type} MCP server`,
+      cause: error,
+    });
+  }
+}
```
lgrammel, line 122: inline

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+ * primarily for tool conversion between MCP<>AI SDK.
+ *
+ * It is a custom implementation of the MCP Client (derived from Protocol) class.
+ *
+ * Not supported:
+ * - Client options (e.g. sampling, roots) as they are not needed for tool conversion
+ * - Accepting notifications
+ */
+export class MCPClient {
+  private transport: Transport;
```
lgrammel, line 48: rename `Transport` to `MCPTransport`

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+  private clientInfo: Implementation;
```
lgrammel, line 49: better type name?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+  }): Promise<z.infer<T>> {
+    return new Promise((resolve, reject) => {
+      options?.signal?.throwIfAborted();
+
+      const timeoutSignal = AbortSignal.timeout(
+        options?.timeout ?? this.requestTimeoutMs,
+      );
+      const signal = options?.signal
+        ? AbortSignal.any([options.signal, timeoutSignal])
+        : timeoutSignal;
```
lgrammel, line 147: since we pass in an `AbortSignal`, I was thinking we can fully remove the timeout code to keep things simple

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  constructor({
+    transport: transportConfig,
+    name = 'ai-sdk-mcp-client',
+    version = '1.0.0',
+    connectionTimeoutMs = CONNECTION_TIMEOUT_MS,
+    requestTimeoutMs = REQUEST_TIMEOUT_MS,
+  }: MCPClientConfig) {
+    this.transport = createMcpTransport(transportConfig);
+    this._initTransport();
```
lgrammel, line 66: remove _

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+      name: 'McpClientConnectionClosedError',
+      message: 'Connection closed',
+    });
+
+    for (const handler of responseHandlers.values()) {
+      handler(error);
+    }
+  }
+
+  private _onerror(error: Error): void {
```
lgrammel, line 246: remove _ from private methods

packages/ai/core/tool/mcp/client.ts
lgrammel: and inline if only used here?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+    this._initTransport();
+    this.clientInfo = {
+      name,
+      version,
+    };
+    this.connectionTimeoutMs = connectionTimeoutMs;
+    this.requestTimeoutMs = requestTimeoutMs;
+  }
+
+  async init(): Promise<this> {
```
lgrammel, line 75: why is there an init that's separate from the constructor and `initTransport`?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,302 @@
...
+      signal.addEventListener('abort', () => {
+        cleanup();
+        reject(
+          new AISDKError({
+            name: 'McpClientTimeoutError',
+            message: 'Request timed out',
+            cause: signal?.reason,
+          }),
+        );
+      });
```
lgrammel, line 169: e.g. not needed if we remove timeouts?

examples/ai-core/package.json
iteratetograceness: Fixed by updating build command! `npx tsc src/mcp/server.ts --outDir /dist --target es2023 --module nodenext`

packages/ai/core/tool/mcp/client.ts
iteratetograceness: ah that was in response to your comment regarding merging the construction + initialization of the connection! since constructor cannot have async calls!

packages/ai/core/tool/mcp/client.ts
```diff
@@ -20,7 +20,7 @@ import {
   SUPPORTED_PROTOCOL_VERSIONS,
   TransportConfig,
 } from './types';
-import { AISDKError } from '@ai-sdk/provider';
+import { MCPClientError } from '@ai-sdk/provider';
```
lgrammel, line 23: The error can be in the `ai` package

packages/provider/src/errors/index.ts
```diff
@@ -13,3 +13,4 @@ export { NoSuchModelError } from './no-such-model-error';
 export { TooManyEmbeddingValuesForCallError } from './too-many-embedding-values-for-call-error';
 export { TypeValidationError } from './type-validation-error';
 export { UnsupportedFunctionalityError } from './unsupported-functionality-error';
+export { MCPClientError } from './mcp-client-error';
```
lgrammel, line 16: mv to `ai` pkg

examples/mcp/.env.example
```diff
@@ -0,0 +1 @@
+OPENAI_API_KEY=""
```
lgrammel, line 1: +1 on having a separate example folder

examples/mcp/src/sse/client.ts
```diff
@@ -0,0 +1,40 @@
...
+  const client = await createMCPClient({
+    transport: {
+      type: 'sse',
+      url: 'http://localhost:8080/sse',
+    },
+  });
+
+  const tools = await client.tools({
+    schemas: sseToolSchemas,
+  });
```
lgrammel, line 22: think this looks good. `sseToolSchemas` could be inlined in this example

packages/ai/core/tool/mcp/client.test.ts
```diff
@@ -0,0 +1,192 @@
...
+          'mock-tool': {
+            parameters: z.object({ bar: z.string() }),
+          },
+        },
+      });
+      expect(tools).toHaveProperty('mock-tool');
+    });
+  });
+
+  describe.only('sse', () => {
```
lgrammel, line 91: why `.only`?

packages/ai/core/tool/mcp/client.test.ts
```diff
@@ -0,0 +1,192 @@
...
+          ],
+        },
+      },
+      'http://localhost:3000/messages': {
+        response: {
+          type: 'json-value',
+          body: { jsonrpc: '2.0', id: 1, result: {} },
+        },
+      },
+    });
```
lgrammel, line 122: move into `before`?

packages/ai/core/tool/mcp/client.test.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    it('should return AI SDK compatible tool set', async () => {
+      const tools = await client.tools();
+      expect(tools).toHaveProperty('mock-tool');
+      const tool = tools['mock-tool'];
+      expect(tool).toHaveProperty('parameters');
+      expect(tool).toHaveProperty('execute');
+      expect(tool.execute).toBeInstanceOf(Function);
+    });
```
lgrammel, line 190: more tests for the other cases?

packages/ai/core/tool/mcp/client.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    afterEach(async () => {
+      await client?.close();
+    });
+
+    it('should return AI SDK compatible tool set', async () => {
+      const tools = await client.tools();
+      expect(tools).toHaveProperty('mock-tool');
+      const tool = tools['mock-tool'];
+      expect(tool).toHaveProperty('parameters');
+      expect(tool).toHaveProperty('execute');
```
lgrammel, line 188: test the values?

packages/ai/core/tool/mcp/client.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    });
+
+    it('should return AI SDK compatible tool set', async () => {
+      const tools = await client.tools();
+      expect(tools).toHaveProperty('mock-tool');
+
+      const tool = tools['mock-tool'];
+      expect(tool).toHaveProperty('parameters');
+      expect(tool).toHaveProperty('execute');
+      expect(tool.execute).toBeInstanceOf(Function);
```
lgrammel, line 36: test values & function?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+} from './types';
+import { createMcpTransport } from './transport';
+
+interface MCPClientConfig {
+  /** Transport configuration for connecting to the MCP server */
+  transport: TransportConfig;
+  /** Optional client name, defaults to 'ai-sdk-mcp-client' */
+  name?: string;
+  /** Optional client version, defaults to '1.0.0' */
+  version?: string;
```
lgrammel, line 38: hm the client version is what we support, right? so should the user be able to specify it?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+ * but can later be extended to support other MCP features
+ *
+ * Tool parameters are automatically inferred from the server's JSON schema
+ * if not explicitly provided in the tools configuration
+ *
+ * Not supported:
+ * - Client options (e.g. sampling, roots) as they are not needed for tool conversion
+ * - Accepting notifications
+ */
+export class MCPClient {
```
lgrammel, line 62: is the export needed?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+        throw new MCPClientError({
+          message: `Server's protocol version is not supported: ${result.protocolVersion}`,
+        });
+      }
+
+      if (!result.capabilities?.tools) {
+        throw new MCPClientError({
+          message: `Server does not support tools`,
+        });
+      }
```
lgrammel, line 132: should this happen here or when the `tools` are actually requested?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+
+      // Complete initialization handshake:
+      await this.notification({
+        method: 'notifications/initialized',
+      });
+
+      return this;
+    } catch (error) {
+      void this.close();
```
lgrammel, line 141: `void` is not needed? also, what are the implications of calling close here?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+    const responseHandlers = this.responseHandlers;
+    this.responseHandlers = new Map();
+
+    const error = new MCPClientError({
+      message: 'Connection closed',
+    });
+
+    for (const handler of responseHandlers.values()) {
+      handler(error);
+    }
```
lgrammel, line 305: would the actual connection need to be closed? also, why do we need to swap response handlers? can we set an isClosed flag on the object instead?

also, i prefer `onError`, `onClose`

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+  }
+
+  private onerror(error: Error): void {
+    throw new MCPClientError({
+      message: error.message,
+      cause: error,
+    });
+  }
```
lgrammel, line 313: what will this do in practice (in terms of the stack and where the error shows up)?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,341 @@
...
+
+    if ('result' in response) {
+      handler(response);
+    } else {
+      const error = new MCPClientError({
+        message: response.error.message,
+        cause: response.error,
+      });
+      handler(error);
+    }
```
lgrammel, line 339: consider push into ternary or use guard clause:

```ts
handler(
  'result' in response
  ? response
  : new MCPClientError({
        message: response.error.message,
        cause: response.error,
      })
)
```

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,84 @@
...
+        foo: { type: 'string' },
+      },
+    },
+  },
+];
+
+export class MockStdioTransport {
+  onmessage?: (message: JSONRPCMessage) => void;
+  onclose?: () => void;
+  onerror?: (error: Error) => void;
```
lgrammel, line 19: I prefer onMessage, onClose, onError

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,84 @@
...
+          });
+        }, 0);
+      }
+    }
+
+    return Promise.resolve();
+  }
+
+  async close(): Promise<void> {
+    if (this.onclose) this.onclose();
```
lgrammel, line 81: `this?.onClose()`

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,84 @@
...
+    return Promise.resolve();
```
lgrammel, line 82: not needed if async

packages/ai/core/tool/mcp/transport.ts
```diff
@@ -0,0 +1,374 @@
...
+  onclose?: () => void;
+  onerror?: (error: Error) => void;
+  onmessage?: (message: JSONRPCMessage) => void;
+
+  constructor({ url }: McpSSEServerConfig) {
+    this.url = new URL(url);
+  }
+
+  async start(): Promise<void> {
+    console.log('>>>>>>SSEClientTransport start', this.url);
```
lgrammel, line 145: dangling log

packages/ai/core/tool/mcp/types.ts
```diff
@@ -0,0 +1,269 @@
...
+  JSONRPCRequestSchema,
+  JSONRPCNotificationSchema,
+  JSONRPCResponseSchema,
+  JSONRPCErrorSchema,
+]);
+export type JSONRPCMessage = z.infer<typeof JSONRPCMessageSchema>;
+
+export interface MCPTransport {
+  start(): Promise<void>;
+  send(message: JSONRPCMessage): Promise<void>;
```
lgrammel, line 141: do we need zod schemas for all of this (as in, we use the schemas), or can we go more lightweight and just define TS types (at least for some)?

packages/provider-utils/src/test/unified-test-server.ts
```diff
@@ -104,6 +122,7 @@ export function createTestServer<URLS extends { [url: string]: UrlHandler }>(
 ): {
   urls: FullHandlers<URLS>;
   calls: TestServerCall[];
+  addSseEvent: (url: string, data: any, event?: string, id?: string) => void;
```
lgrammel, line 125: wondering about this because it is sse specific. there is another pattern i used, see e.g. https://github.com/vercel/ai/blob/main/packages/react/src/use-chat.ui.test.tsx#L880

wondering if that would be an alternative (maybe it's the same thing and `addServerSentEvent` or `addSSE` (rename?) is more elegant

lgrammel: also do we want to be SSE specific here? this could just a generic stream and you could attach SSE formatted chunks

packages/ai/core/tool/mcp/client.test.ts
iteratetograceness: ah left over from testing!

packages/ai/core/tool/mcp/client.ts
iteratetograceness: client config is more informational, not really used in a critical capacity (e.g. compatibility checking b/w client and server) - i'd like to advocate for keeping it so that users can debug/track specific clients they create if necessary!

packages/ai/core/tool/mcp/client.ts
iteratetograceness: initial had it where we list tools but moved here given a server that does not support tools is not useful to the end user in our use case!

packages/ai/core/tool/mcp/client.ts
iteratetograceness: noted on `void`! calling close here ensures we clean up the transport/connection and clear our response handlers

packages/ai/core/tool/mcp/client.ts
iteratetograceness: yes! we invoke `this.transport?.close()` when `this.onclose` is called!

packages/ai/core/tool/mcp/client.ts
iteratetograceness: ah, good point! since the client's `onError` (which is triggered via the transport's `onerror` handler or `onResponse`) we aren't properly propagating this up within the client

iteratetograceness: instead of throwing `MCPClientError`, should we just close the client+transport? we need a way to properly bubble up the error to the user hmmmmm

packages/ai/core/tool/mcp/transport.ts
```diff
@@ -0,0 +1,374 @@
...
+
+    env[key] = value;
+  }
+
+  return env;
+}
+
+function isElectron() {
+  return 'type' in process;
+}
```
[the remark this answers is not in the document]
iteratetograceness: used for config when spawning child_process!

packages/ai/core/tool/mcp/transport.ts
```diff
@@ -0,0 +1,374 @@
...
+                    message: 'SSE connection closed unexpectedly',
+                  });
+                  this.onerror?.(error);
+                }
+                return;
+              }
+
+              buffer += textDecoder.decode(value, { stream: true });
+              const events = buffer.split('\n\n');
+              buffer = events.pop() || '';
```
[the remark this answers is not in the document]
iteratetograceness: added in case we get a chunk of that includes the beginning of the _next_ event!

packages/provider-utils/src/test/unified-test-server.ts
iteratetograceness: my approach is missing a sort of `streamController` like your approach is 👀 hoping to discuss in call tomo!

packages/ai/core/tool/mcp/client.ts
lgrammel: we may want this internally, agreed - the reason i bring this up is that i'm not convinced we should expose it to our users

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,354 @@
...
+
+  private async request<T extends ZodType<object>>({
+    request,
+    resultSchema,
+    options,
+  }: {
+    request: Request;
+    resultSchema: T;
+    options?: RequestOptions;
+  }): Promise<z.infer<T>> {
```
lgrammel, line 159: check if closed?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,354 @@
...
+          options: ToolExecutionOptions,
+        ): Promise<CallToolResult> => {
+          options?.abortSignal?.throwIfAborted();
+
+          const result = await this.callTool({
+            params: {
+              name,
+              arguments: args,
+            },
+            resultSchema: CallToolResultSchema,
```
lgrammel, line 291: if this is fixed, should it be pushed into callTool?

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,354 @@
...
+            options: {
+              signal: options.abortSignal,
+            },
```
lgrammel, line 294: callTool is our own function so we can flatten

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,354 @@
...
+          args: inferParameters<typeof parameters>,
+          options: ToolExecutionOptions,
+        ): Promise<CallToolResult> => {
+          options?.abortSignal?.throwIfAborted();
+
+          const result = await this.callTool({
+            params: {
+              name,
+              arguments: args,
+            },
```
lgrammel, line 290: same as below, we can flatten

packages/ai/core/tool/mcp/client.ts
```diff
@@ -0,0 +1,354 @@
...
+      const toolWithExecute = tool({
+        description,
+        parameters,
+        execute: async (
+          args: inferParameters<typeof parameters>,
+          options: ToolExecutionOptions,
+        ): Promise<CallToolResult> => {
+          options?.abortSignal?.throwIfAborted();
+
+          const result = await this.callTool({
```
lgrammel, line 286: `this` can lead binding issues. it might be better to have a `self` reference in scope and use that

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,94 @@
...
+                  text: `Mock tool call result`,
+                },
+              ],
+            },
+          });
+        }, 0);
+      }
+    }
+
+    return;
```
lgrammel, line 87: is `return` needed here?

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,94 @@
...
+    overrideTools = DEFAULT_TOOLS,
+  }: {
+    overrideTools?: MCPTool[];
+  } = {}) {
+    this.tools = overrideTools;
+  }
+
+  async start(): Promise<void> {
+    return;
+  }
```
lgrammel, line 33: why is this needed?

packages/ai/core/tool/mcp/mock-mcp-stdio-server.ts
```diff
@@ -0,0 +1,94 @@
...
+    inputSchema: {
+      type: 'object',
+      properties: {
+        foo: { type: 'string' },
+      },
+    },
+  },
+];
+
+export class MockStdioTransport {
```
lgrammel, line 16: implement mcp transport?

packages/ai/core/tool/mcp/transport.ts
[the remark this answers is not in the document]
lgrammel: i meant replace `||` w/ `??` because that's closer to the intended semantics?

RoversX:
Wonderful, just wondering when this will be integrated. 😀

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -624,6 +624,87 @@ export const weatherTool = tool({
 });
 ```
 
+## MCP Tools
+
+The AI SDK supports connecting to Model Context Protocol (MCP) servers to access their tools. This enables your AI applications to discover and use tools across various services through a standardized interface.
```
lgrammel, line 629: might be nice to link here to the mcp site

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -624,6 +624,87 @@ export const weatherTool = tool({
...
+
+### Using MCP Tools
+
+Create an MCP client using either `stdio` (for local servers) or `SSE` (for web servers):
```
lgrammel, line 633: a little explanation would stdio vs sse is could be nice (1 sentence each maybe)

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -624,6 +624,87 @@ export const weatherTool = tool({
...
+const client = await createMCPClient({
+  transport: {
+    type: 'stdio',
+    command: 'node',
+    args: ['src/stdio/dist/server.js'],
+  },
+});
+```
+
+The client's `tools` method acts as an adapter between MCP tools and AI SDK tools. It supports two approaches for working with tool schemas:
```
lgrammel, line 645: maybe have a subsection for this, and then explain the trade-offs as well. as a developer, when would i choose which approach?

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -624,6 +624,87 @@ export const weatherTool = tool({
...
+      query: z.string().describe('The data query'),
+      format: z.enum(['json', 'text']).optional(),
+    }),
+  },
+} as const;
+
+const tools = await client.tools({ schemas: toolSchemas });
+```
+
+Below is a complete example of using MCP tools with the AI SDK.
```
lgrammel, line 671: instead move under Example heading and explain example there. No need to link sections with placeholders

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -624,6 +624,87 @@ export const weatherTool = tool({
...
+        content: 'Find products under $100',
+      },
+    ],
+  });
+
+  console.log(response.text);
+} catch (error) {
+  console.error(error);
+} finally {
+  await client.close();
```
lgrammel, line 704: maybe explain closing and why

examples/mcp/README.md
```diff
@@ -0,0 +1,31 @@
+# MCP + AI SDK Example - TODO: REVISE
```
lgrammel, line 1: TODO here

content/cookbook/05-node/54-mcp-tools.mdx
```diff
@@ -0,0 +1,60 @@
...
+    },
+  });
+
+  const toolSetOne = await clientOne.tools();
+  const toolSetTwo = await clientTwo.tools();
+
+  const tools = {
+    ...toolSetOne,
+    ...toolSetTwo,
+  };
```
lgrammel, line 41: not sure we need to do something here, but a potential error scenario is that `toolSetTwo` silently overrides tools from `toolSetOne` when there  naming conflicts

examples/mcp/src/stdio/client.ts
```diff
@@ -0,0 +1,48 @@
...
+        console.log(`STEP RESULTS: ${JSON.stringify(toolResults, null, 2)}`);
+      },
+      system: 'You are an expert in Pokemon',
+      prompt:
+        'Which Pokemon could best defeat Feebas? Choose one and share details about it.',
+    });
+
+    console.log(`FINAL ANSWER: ${answer}`);
+  } catch (error) {
+    throw error;
```
lgrammel, line 42: does nothing, not needed

lgrammel:
Superseeded by https://github.com/vercel/ai/pull/5144

Issue #4741, opened 2025-02-06
useChat: Infinite re-renders when using messages as useEffect dependency


sebaFP:
### Bug Description
When using the messages array from useChat as a dependency in useEffect, it causes infinite re-renders and throws a "Maximum update depth exceeded" error.

### Steps to Reproduce
1. Use the useChat hook to get messages
2. Add messages as a dependency in a useEffect

```
const { messages } = useChat();
useEffect(() => {
  // Any logic here
}, [messages]); // Adding messages as dependency
```
### Current Behavior
- Component enters an infinite re-render loop
- Console shows error: "Maximum update depth exceeded"
- Full error: "Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render."

### Expected Behavior
- useEffect should only run when messages actually change from user interactions
- No infinite re-render loop should occur

### Environment
- vercel/ai version: 4.1.24
- React version: 19

### Additional Context
This seems to happen because messages reference might be changing on every render, causing the effect to run repeatedly.

![Image](https://github.com/user-attachments/assets/312a9707-19d9-4447-a872-b5cfdcfedac1)

calenwu:
I am also experiencing this bug, did you find workaround?

lgrammel:
@sebaFP did you encounter this issue as part of an ai sdk upgrade or independently?

Godrules500:
@lgrammel I just updated and ran into the issue after I updated the ai-sdk and vertex ai sdk. I uninstalled it and it started working again. The error seems to be happening after the final message is finished. It'll print out the whole long message I'm playing with, and then at the very end it throws that maximum depth error listed above

lgrammel:
@Godrules500 from which version did you update?

rafalzawadzki:
I'm experiencing something similar, but while using `setInput`:

```js
const {
    input,
    setInput
  } = useChat({ ... })
```

Also on `4.1.34` and downgrading to `4.1.16` fixed it.

lgrammel:
@rafalzawadzki could you provide more details so i can reproduce?

rafalzawadzki:
tbh the project is so old and complicated that it's hard to give a full rundown.

I upgraded from `3.4.33` to `4.1.34`, saw the error, this solution, and it worked again after downgrade to `4.1.16`.

my first attempts at fixing this were the typical React state fixes: add dependencies to hooks etc - but nothing worked.

[...]

PR #5013, opened 2025-02-27
fix (react): infinite re-render caused by `fillMessageParts` 

iteratetograceness:
Addresses https://github.com/vercel/ai/issues/4741, caused by referential instability of `initialMessages`

When initial messages are present in `useChat`, we would call `fillMessageParts` when setting SWR `fallbackData`; this would create a new array every render

Resolved by implementing the same logic used for `initialMessagesFallback`, where we store the array of initial messages (with parts, if applicable) instead of relying on the default parameter value

### DIFF
--- a/packages/react/src/use-chat.ui.test.tsx
+++ b/packages/react/src/use-chat.ui.test.tsx
@@ -9,7 +9,7 @@ import {
 import '@testing-library/jest-dom/vitest';
 import { cleanup, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import React, { useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useChat } from './use-chat';
 
 describe('data protocol stream', () => {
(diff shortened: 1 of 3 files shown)


gaspar09 (approved):

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -1888,3 +1888,83 @@ describe('test sending additional fields during message submission', () => {
...
+
+  it('should not cause infinite rerenders when initialMessages is defined and messages is a dependency of useEffect', async () => {
+    await waitFor(() => {
+      expect(screen.getByTestId('message-user')).toHaveTextContent(
+        'Test message',
+      );
+    });
+
+    const initialRenderCount = parseInt(
+      screen.getByTestId('render-count').textContent || '0',
```
lgrammel, line 1954: why `|| '0'`? if this should not happen, can you use `!` instead to fix type error (also below)?

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -1888,3 +1888,83 @@ describe('test sending additional fields during message submission', () => {
...
+    );
+
+    await new Promise(resolve => setTimeout(resolve, 100));
+
+    const finalRenderCount = parseInt(
+      screen.getByTestId('render-count').textContent || '0',
+    );
+
+    expect(finalRenderCount).toBeLessThan(3);
```
lgrammel, line 1963: test exact count?

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -1888,3 +1888,83 @@ describe('test sending additional fields during message submission', () => {
...
+      ],
+    });
+
+    useEffect(() => {
+      setDerivedState(messages.map(m => m.content));
+    }, [messages]);
+
+    if (renderCount > 10) {
+      throw new Error('Excessive renders detected; likely an infinite loop!');
+    }
```
lgrammel, line 1922: needed? or could we just let the test time out?

lgrammel (approved):
Looks great, just had a few minor questions on the test.

Can you add a patch changeset (with PR title)?

packages/react/src/use-chat.ui.test.tsx
iteratetograceness: Changing to 2, given useEffect will run twice (due to react strict mode) in test env

packages/react/src/use-chat.ui.test.tsx
iteratetograceness: this was a result of the test not timing out as expected! when we `render` the component, it just hangs (due to the infinite render) - when we detect infinite loop this `if` conditional forces the test to fail!

PR #5039, opened 2025-03-03
fix: handle empty object for content in responseSchema

BrianHung:
Not sure if it's a gemini-2.0-flash-lite API thing, but their endpoint started returning an empty object for content for `MALFORMED_FUNCTION_CALL`. This PR updates the responseSchema to handle it.

```
{
  "candidates": [
    {
      "content": {},
      "finishReason": "MALFORMED_FUNCTION_CALL"
    }
  ],
  "usageMetadata": {
    "promptTokenCount": 9056,
    "totalTokenCount": 9056,
    "promptTokensDetails": [
      {
        "modality": "TEXT",
        "tokenCount": 9056
      }
    ]
  },
  "modelVersion": "gemini-2.0-flash-lite"
}
```

### DIFF
--- a/packages/google/src/google-generative-ai-language-model.test.ts
+++ b/packages/google/src/google-generative-ai-language-model.test.ts
@@ -8,6 +8,7 @@ import { createGoogleGenerativeAI } from './google-provider';
 import {
   GoogleGenerativeAILanguageModel,
   groundingMetadataSchema,
+  responseSchema,
 } from './google-generative-ai-language-model';
 import { GoogleGenerativeAIGroundingMetadata } from './google-generative-ai-prompt';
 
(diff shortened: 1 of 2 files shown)


lgrammel:
@BrianHung did google publish any api docs that mention this state?

BrianHung:
@lgrammel https://cloud.google.com/vertex-ai/docs/reference/rest/v1/GenerateContentResponse#FinishReason has it, but nothing that mentions the empty object returned as content as well. The message above I got from snooping the network requests.

BrianHung:
@lgrammel Don't have a consistent reproduction example because I'm using multiple tools in a loop (not from using structured outputs) with experimental_activeTools on; but I'll try to log out the exact request body to gemini if it appears again.

Can't really control for it if the error occurs internally in google's API servers. Other developers (langchain/pydantic) have encountered it too: https://github.com/langchain-ai/langchain-google/issues/725

je2ryw:
I am facing the same "MALFORMED_FUNCTION_CALL" error with gemini-2.0-flash, and after many rounds of direct HTTP POST tests with curl command, it turns out the triple backticks "```" is the root cause of error, even in the user text part section, and it has nothing to do with "tools" or "functionDeclarations".

PR #5043, opened 2025-03-03
fix (ai/core): forward providerOptions for text, image, and file parts

### LINKED ISSUE #4998: Anthropic prompt caching does not work for user message when set with providerOptions
### Description

Looks like the function (convertPartToLanguageModelPart) (https://github.com/vercel/ai/blob/9197aa5b18c50668118b9f2b9551239d4ca3b06c/packages/ai/core/prompt/convert-to-language-model-prompt.ts#L214) only relies on experimental_providerMetadata and not on providerOptions.

```
if (part.type === 'text') {
    return {
      type: 'text',
      text: part.text,
      providerMetadata: part.experimental_providerMetadata,
    };
  }
```

### Code example

So in my code this does not work -
```
part.providerOptions = { anthropic: { cache_control: { type: 'ephemeral' } } };
```

While this is ok -

```
part.experimental_providerMetadata = { anthropic: { cache_control: { type: 'ephemeral' } } };
```

### AI provider

Anthropic

### Additional context

Trying to use prompt caching

lgrammel:
Fixes #4998

### DIFF
--- a/.changeset/chatty-icons-destroy.md
+++ b/.changeset/chatty-icons-destroy.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix (ai/core): forward providerOptions for text, image, and file parts
--- a/packages/ai/core/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/core/prompt/convert-to-language-model-prompt.ts
@@ -225,7 +225,8 @@ function convertPartToLanguageModelPart(
     return {
       type: 'text',
       text: part.text,
-      providerMetadata: part.experimental_providerMetadata,
+      providerMetadata:
+        part.providerOptions ?? part.experimental_providerMetadata,
     };
   }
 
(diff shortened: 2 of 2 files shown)


nicoalbanese (approved):

gianpaj:
Could somebody document `providerOptions` for `useChat()`
https://sdk.vercel.ai/docs/reference/ai-sdk-ui/use-chat

I see the type, but nothing on the doc's page
https://github.com/vercel/ai/blob/8fd997e6a53237aee108d95680b2b9e8ca93c91b/packages/ai/core/prompt/message.ts#L36-L40

Is that the right place to send `metadata` to OpenAI?
https://platform.openai.com/docs/api-reference/chat/create#chat-create-metadata

I'm looking to send some metadata to get in the OpenAI Dashboard logs.
SCR-20250314-tllv

thanks 🙏

PR #5044, opened 2025-03-03
fix (docs): update data stream data part documentation

valstu:
Minor fix to documentation related to data streams.

Data Part is not appended to the message, it is available through data array returned by the useChat.

I think it should be separate discussion but I think we should have data part that is actually related to certain message though. Similarly like we have annotations. Now you have to do all kinds of shenanigans to map data to certain message through the "global" data array returned by the `useChat`. Just as an example, I have an function call which makes separate llm call and I want to stream that content to main chat. Now I'm using custom data stream for that, I'm basically streaming deltas from nested llm call through data stream like this:
```
{
  toolCallId: '123',
  textDelta: 'something'
}
```
And then map them inside messages to correct function call part so I can render the stream of nested llm call. It would be nice if I could just access data through messages directly. Something like
```
messages.map(msg => msg.data)
```
Don't know if this makes sense or not but if it does I could do a discussion or feature request about this feature?


lgrammel (approved):

lgrammel:
@valstu feel free to file a separate issue with your thoughts. It's on my mind and we have some improvements planned in the area, but it's always good to have more input.

PR #5056, opened 2025-03-03
feat (provider/svelte): support Svelte 5

### LINKED ISSUE #3107: Svelte 5 Support
### Description

Currently, the `@ai-sdk/svelte` only supports Svelte 3 and Svelte 4 but could be completely usable by changing the `sswr` dependency with something like `@tanstack/svelte-query` or a built-in fetching strategy.

## Svelte 5: Runes mode

```typescript
let { data } = $props();
let { messages, input, handleSubmit } = useChat();
```

image

## Svelte 5: Compatibility mode

```typescript
export let data;
// No rune is used, does not throw an error.
let { messages, input, handleSubmit } = useChat();
```

### Additional context

These are the places where the library uses the Svelte 4 restrictive code:

https://github.com/ConsoleTVs/sswr/blob/836f3eab4faa4e22bd936f39ea66a42d56707175/src/sswr.ts#L29-L50

### LINKED ISSUE #3454: `@ai-sdk/svelte` - `beforeUpdate(...)` cannot be used in runes mode
### Description

After updating to Svelte 5, I get this error:
```
Uncaught Svelte error: lifecycle_legacy_only
`beforeUpdate(...)` cannot be used in runes mode
```
In the log it points to this code somewhere in `@ai-sdk/svelte`
brave_zjXBUTPpHb

### Code example

```svelte
<script lang="ts">
	import { useCompletion } from "@ai-sdk/svelte";

	const { input, handleSubmit, completion, isLoading } = useCompletion({
		api: "/api/openai/format",
	});

	let output = $state("");

	$effect(() => {
		output = $completion;
	});
</script>

```

### Additional context

_No response_

elliott-with-the-longest-name-on-github:
Closes #3107
Closes #3454 

`sswr` was the blocker, and I'm really, really unclear on why it was being used in the first place. It _appears_ it was being used as a global cache, but... we can just use a regular store for that? So I _think_ this replaces it.

I tested both the `useChat` and `useCompletion` hooks in Svelte 5 and they appear to work just fine. Since the only changes here are the removal of `sswr` and the addition of a global writable store, this should work with Svelte 3-5 just as it did with Svelte 3-4 before.

### DIFF
--- a/.changeset/selfish-moons-happen.md
+++ b/.changeset/selfish-moons-happen.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/svelte': patch
+---
+
+feat (provider/svelte): support Svelte 5
--- a/packages/svelte/package.json
+++ b/packages/svelte/package.json
@@ -28,8 +28,7 @@
   ],
   "dependencies": {
     "@ai-sdk/provider-utils": "2.1.10",
-    "@ai-sdk/ui-utils": "1.1.16",
-    "sswr": "^2.1.0"
+    "@ai-sdk/ui-utils": "1.1.16"
   },
   "devDependencies": {
     "@types/node": "^18",
--- a/packages/svelte/src/use-chat.ts
+++ b/packages/svelte/src/use-chat.ts
@@ -20,7 +20,6 @@ import {
   shouldResubmitMessages,
   updateToolCallResult,
 } from '@ai-sdk/ui-utils';
-import { useSWR } from 'sswr';
 import { Readable, Writable, derived, get, writable } from 'svelte/store';
 export type { CreateMessage, Message };
 
@@ -111,7 +110,7 @@ export type UseChatHelpers = {
   id: string;
 };
 
-const store: Record<string, UIMessage[] | undefined> = {};
+const store = writable<Record<string, UIMessage[] | undefined>>({});
 
 export function useChat({
   api = '/api/chat',
@@ -144,10 +143,7 @@ export function useChat({
   const chatId = id ?? generateId();
 
   const key = `${api}|${chatId}`;
-  const { data, mutate: originalMutate } = useSWR<UIMessage[]>(key, {
-    fetcher: () => store[key] ?? fillMessageParts(initialMessages),
-    fallbackData: fillMessageParts(initialMessages),
-  });
+  const data = derived([store], ([$store]) => $store[key] ?? initialMessages);
 
   const streamData = writable<JSONValue[] | undefined>(undefined);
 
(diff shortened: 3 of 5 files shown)


lgrammel:
1. how have you tested this?
2. can you add a changeset (patch for ai sdk svelte, with same content as pr title)?

lgrammel:
Does this also address https://github.com/vercel/ai/issues/3454 ?

Rich-Harris:
It does, because `beforeUpdate` is only used inside `sswr`

elliott-with-the-longest-name-on-github:
@lgrammel I put together a Svelte 5 application using OpenAI and tested the `useChat` and `useCompletion` hooks -- here's the code:

```svelte
<script>
	import { useChat, useCompletion } from '@ai-sdk/svelte';

	const { input: chatInput, handleSubmit: handleChatSubmit, messages } = useChat();
	const {
		completion,
		input: completionInput,
		handleSubmit: handleCompletionSubmit
	} = useCompletion();
</script>

<main>
	<ul>
		{#each $messages as message}
			<li>{message.role}: {message.content}</li>
		{/each}
	</ul>
	<form onsubmit={handleChatSubmit}>
		<input bind:value={$chatInput} />
		<button type="submit">Send</button>
	</form>
	<form onsubmit={handleCompletionSubmit}>
		<input name="prompt" bind:value={$completionInput} id="input" />
		<button type="submit">Submit</button>
		<div>{$completion}</div>
	</form>
</main>

```

I also made sure that multiple copies of the hook can exist in different places in the UI at the same time and continue to function.

perezd:
Did you try this code in "runes" mode?

elliott-with-the-longest-name-on-github:
@perezd yes!

lgrammel (approved):

wookiehangover:
I'm seeing errors stemming from this in `addToolResult()` calls. I'm assuming that it's because I accepted this patch upgrade but my component with `useChat` still uses legacy mode in svelte 5. I'm guessing these changes aren't backwards compatible? @lgrammel was there any testing with svelte 5 legacy mode to ensure that this wouldn't break people? Would have loved a major version number for this..

wookiehangover:
https://github.com/vercel/ai/blob/8353caf5fc7ea331dd64039b9c9edc02df57834c/packages/svelte/src/use-chat.ts#L422

messages.set is undefined

```
Uncaught (in promise) TypeError: messages.set is not a function
    at addToolResult (@ai-sdk_svelte.js?v=18889736:266:14)
    at Object.addToolResult
```

pinning to @ai-sdk/svelte v1.1.19 fixes this. TBH the next patch should roll this back entirely until it's either backwards compatible or included in a major version bump. This is super sloppy!

perezd:
Yeah, I agree this should roll back, and supporting svelte 5 should be a major bump. I don't think split legacy support needs to exist in the major bump, so that would be a good way to deal with this going forward.

elliott-with-the-longest-name-on-github:
Fixed in #5078, testing now -- the problem was that, to the shock of _absolutely everyone_, I'm sure, `as X` is evil and shouldn't exist. A `Readable` store was being cast as a `Writable` store.

PR #5078, opened 2025-03-05
fix (ui/svelte): error resulting from hard cast from readable to writable

elliott-with-the-longest-name-on-github:
This is just another example of why `as X` is evil and shouldn't ever be used. A `Readable` was being cast to a `Writable`.

I stuck a button in my test app with 
```
	<button
		onclick={() => {
			console.log('adding tool result');
			addToolResult({
				toolCallId: 'call_wLJFuAcFfygtlxd3nAtYRpt8',
				result: { location: 'Denver, CO', temperature: 68 }
			});
		}}>Add Tool Result</button
	>
```

on `main` this causes the failure reported in the #5056 comments. With this branch, the error is gone and everything seems to be working correctly.

### DIFF
--- a/.changeset/empty-moose-march.md
+++ b/.changeset/empty-moose-march.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/svelte': patch
+---
+
+fix: Remove hard-cast causing `Readable` to be treated as `Writable`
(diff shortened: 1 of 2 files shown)


wookiehangover:
thanks for the fix on this!

lgrammel (approved):

lgrammel:
Approving and merging.

@elliott-with-the-longest-name-on-github this might introduce a new bug. are the returned messages of type `Message` or of type `UIMesage`?

PR #5143, opened 2025-03-12
Bedrock Docs: Updated Reasoning Config Documentation to Match Actual Implementation

tosh99:
Fixed documentation to reflect the actual changes made to the Bedrock Reasoning Configuration.


lgrammel:
Prefer to fix the code instead

pri2si17-1997:
I agree. We should be consistent across all providers. I did mention about doc update in #4977 (after it was closed). But that's not a good idea. We should fix in code.

tosh99:
@pri2si17-1997: Request you to update the code if possible. Thank you.

PR #5159, opened 2025-03-12
feat (provider/google): add support for dynamic retrieval

d3lm:
This PR adds the ability to configure [dynamic retrieval](https://ai.google.dev/gemini-api/docs/grounding?lang=python#dynamic-retrieval) when using Google's search grounding.

Here's an example on how to use it:

```ts

const { text, providerMetadata } = await generateText({
  model: google('gemini-1.5-flash', {
    useSearchGrounding: true,
    dynamicRetrievalConfig: {
      mode: 'MODE_DYNAMIC',
      dynamicThreshold: 0.8,
    },
  }),
  prompt: 'Who won the latest F1 grand prix?',
});
```


lgrammel:
Please add a patch changeset with the pr title for the google provider

d3lm:
Yep will do in a minute.

d3lm:
Added a changeset.

packages/google/src/google-prepare-tools.ts
```diff
@@ -36,9 +45,19 @@ export function prepareTools(
...
     return {
-      tools: isGemini2 ? { googleSearch: {} } : { googleSearchRetrieval: {} },
+      tools: isGemini2
+        ? { googleSearch: {} }
+        : {
+            googleSearchRetrieval:
+              !supportsDynamicRetrieval || !dynamicRetrievalConfig
+                ? {}
+                : { dynamicRetrievalConfig },
+          },
```
lgrammel, line 60: what happens if you do not have the `supportsDynamicRetrieval` flag? wouldn't it be enough to check if the retrieval config is there? that would give the user enough control?

lgrammel:
Please run prettier.

packages/google/src/google-prepare-tools.ts
d3lm: The `supportsDynamicRetrieval` is to check if it's one of the Gemini Flash 1.5 models so if it's not and the user provided the retrieval config then we won't send that up to the Google API. I mean we could also just pass it along and let it fail. I don't have a strong opinion on this one. WDYT?

lgrammel: lets just leave as is

lgrammel (approved):

.changeset/fluffy-clouds-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/google': minor
```
lgrammel, line 2: please change to patch

Issue #5140, opened 2025-03-11
Support useChat Without an API Route (Client-side usage)


The-Best-Codes:
### Feature Description

Right now, the Vercel AI SDK's `useChat` hook is good for managing chat state, but it requires an API route (e.g., `app/api/chat/route.ts`). So even in simple projects or static sites, a backend is necessary just to proxy requests.
I would like to be able to use the `useChat` hook right on the client with a user-provided API key, with no need for an API route.

### Use Cases

I can think of a lot of use cases:

- Static site generators (Next.js static export, Astro, etc.)
- No-API web-to-desktop tools, e.g., Tauri
- Browser extensions
- Vite projects

The-Best-Codes:
Feel free to assign me!
I've got a solution I'm using locally (for the React `useChat` hook). Let me know how you guys would implement this. 😊

chewong:
For anyone who is looking for a pure-frontend hacky solution:

```ts
const model = createOpenAICompatible({
  name: "model",
  baseURL: "http://localhost:5000/v1/"
})
const customFetch = async (url, options) => {
  const m = JSON.parse(options.body) as any;
  console.log(m)
  const result = await streamText({
    model: model.chatModel('meta-llama/Llama-3.1-8B-Instruct'),
    messages: m.messages,
  });
  return result.toDataStreamResponse()
}

const { messages, input, handleInputChange, handleSubmit } = useChat({
  fetch: customFetch,
});
```

The-Best-Codes:
@chewong This looks great, I will try it 👍

xhiroga:
Thank you for the issue and PR. This feature is very necessary.
When creating application extensions, for example, we often need to attach a web server to an existing process.
In such cases, in order to use the AI SDK, it must be available from the hosted application in a statically exported state.

lgrammel:
ai sdk 5 supports custom chat transports, which enables client only usage

The-Best-Codes:
Awesome!! ❤
Should my PR be closed too?
https://github.com/vercel/ai/pull/5207

The-Best-Codes:
@lgrammel I've been experimenting with AI SDK v5 beta and can't really see how it will let me use a model without a backend. Is there any example in the docs?

The-Best-Codes:
Thanks to the code example from @chewong and some research, I was able to create a client-side only (no API backend) minimal chat example. In my example, I'm using `@ai-sdk/google`. If anyone needs it, here it is!

```tsx
[...]

PR #5207, opened 2025-03-13
Add `model` prop to useChat hook to remove the need for a backend

The-Best-Codes:
## Add `model` prop to useChat hook to remove the need for a backend

This PR introduces the `model` prop to the `useChat` hook, enabling direct interaction with Language Model V1 compliant AI providers (e.g., Google, Groq, and other models from `@ai-sdk`) directly from the client-side, thus removing the need for a dedicated backend in simple use cases.

**Addresses:** #5140

**Changes:**

-   **Added `model` prop to `UseChatOptions`:** The `UseChatOptions` interface now accepts an optional `model` property of type `LanguageModelV1`.
-   **Conditional API Call:**  The `triggerRequest` function now checks if a `model` is provided. If so, it uses the model's `doStream` method to interact with the AI provider. Otherwise, it defaults to the existing `callChatApi` function, which requires a backend endpoint.
-   **Streaming Implementation:** When a `model` is provided, the PR implements a streaming mechanism using `TransformStream` and `WritableStream` to handle the response from the AI provider. The response is processed chunk by chunk, updating the chat messages in real-time.
-   **Updated `chatKey`:** The `chatKey` used for SWR is updated to include the `model` instance if it is provided, ensuring unique cache keys for different model configurations.
-   **Typescript improvements:** Added the language model types to the existing types.

**Usage Example:**

```tsx
"use client";
import { createGoogleGenerativeAI } from "@ai-sdk/google";
import { useChat } from "@ai-sdk/react";

// Existing function, I made no changes here
const google = createGoogleGenerativeAI({
  apiKey: "(pass the key from somewhere)",
});

function ChatComponent() {
  const { messages, append, input, setInput, handleSubmit } = useChat({
    model: google("gemini-2.0-flash-001"), // Pass the model instance - this is the change I made. You can still remove this prop and use the API by default. This removes the need for a backend, not the option.
  });

  return (
    <div>
      {/* Some UI */}
    </div>
  );
}

export default ChatComponent;
```

Super open to changing this if it's not what you had in mind. :grinning:


The-Best-Codes:
Please do **not** merge this PR yet. I need to do some more tests. :face_with_diagonal_mouth:

The-Best-Codes:
I'm done testing, ready for feedback :ear:

qieman:
Hope the code can be merged quickly, as I want to use the API key directly in the client instead of relying on server forwarding.

rob-benzo:
Vercel when will this merge :eyes:

hanzalahwaheed:
yo vercel! review over my guys code asap (plz)

lgrammel:
It's important to remove the need for a backend, but baking the model in into useChat not the right approach - instead there needs to be flexibility on how requests get executed.

The-Best-Codes:
@lgrammel Right now you can use it exactly the same as before, or you can pass a Language Model V1 instance to the `model` prop. For example, `@ai-sdk/google` or `@ai-sdk/groq`.
So it's not necessarily 'baking in' models into the `useChat` hook, since you can choose not to use the `model` prop at all, and you can pass any language model from the AI SDK to it.
(I probably misunderstood what you were saying, sorry :roll_eyes:)

Of course I'll be happy to close this if it's not what you all have in mind. What approach does the team want to take to this?

lgrammel:
@The-Best-Codes I'm envisioning a transport abstraction where you can plug in transports other than http, and a potential transport could be local execution (with a simplified direct model execution that is close to what you show here). The main difference would be how it's configured such that the transport polymorphism works well.

lgrammel:
@The-Best-Codes you can just leave it as is for now - I will come back to it in time.

The-Best-Codes:
Merge conflicts. Apparently that rebase didn't go quite right... :flushed:

lauridskern:
any updates on this? still doesn't seem possible

lauridskern:
> What doesn't seem possible?

Sorry, I meant to ask what the plans are for this PR and the "transport abstraction", mainly directed towards the maintainers

The-Best-Codes:
@lauridskern 
> Sorry, I meant to ask what the plans are for this PR and the "transport abstraction", mainly directed towards the maintainers

You might find this issue helpful:
https://github.com/vercel/ai/issues/5140
And the transport docs here: https://ai-sdk.dev/docs/ai-sdk-ui/transport

Issue #5179, opened 2025-03-13
Allow custom MCP transport


phuctm97:
### Feature Description

The new MCP API is great but it's harded-coded with 2 default transport at the moment. Instead, it should accept custom transport.

### Use Cases

To use AI SDK in non-standard JS environment like Electron / Tauri browser code that has access to equivalent Node.js APIs for IO but not the same.

lgrammel:
Resolved by https://github.com/vercel/ai/pull/5209

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests a new API capability and broader transport extensibility rather than reporting a failure of documented behavior.

### Evidence
- The MCP API was limited to two hard-coded transports.
- The request asks the API to accept custom transports for environments such as Electron and Tauri.
- A maintainer states that the request was resolved by PR #5209.

PR #5209, opened 2025-03-14
feat(ai): allow custom mcp transports

iteratetograceness:
Enables [use of custom transports](https://github.com/vercel/ai/issues/5179) with AI SDK's `experimental_createMCPClient`

Additionally, addresses problems that arise with bundling node-only dependencies like `child_process` (related issue [here](https://github.com/vercel/ai/issues/5228))

This PR introduces a breaking change to how we support `stdio` transports. Instead of creating the `stdio` client transport internally, we export `Experimental_StdioMCPTransport` for users to pass in as a custom transport.


content/cookbook/05-node/54-mcp-tools.mdx
```diff
@@ -17,11 +17,12 @@ let clientTwo;
 
 try {
   // Initialize an MCP client to connect to a `stdio` MCP server:
+  const transport = new StdioClientTransport({
+    command: 'node',
+    args: ['src/stdio/dist/server.js'],
+  });
```
lgrammel, line 23: should we already change the bundling in this PR to have a single breaking change vs two?

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -639,35 +639,44 @@ Create an MCP client using either:
...
-const mcpClient = await createMCPClient({
-  transport: {
-    type: 'stdio',
-    command: 'node',
-    args: ['src/stdio/dist/server.js'],
-  },
+const transport = new StdioClientTransport({
+  command: 'node',
+  args: ['src/stdio/dist/server.js'],
+});
```
lgrammel, line 649: inline?

packages/ai/core/tool/mcp/mcp-client.ts
```diff
@@ -78,10 +78,16 @@ class MCPClient {
     onUncaughtError,
   }: MCPClientConfig) {
     this.onUncaughtError = onUncaughtError;
-    this.transport = createMcpTransport(transportConfig);
-    this.transport.onClose = () => this.onClose();
-    this.transport.onError = (error: Error) => this.onError(error);
-    this.transport.onMessage = message => {
+
+    if ('type' in transportConfig) {
```
lgrammel, line 82: wonder if this is good enough of a check - bc the `type` property is very common.

we could also turn it around and check for a few function properties of `MCPTransport`

packages/ai/core/tool/mcp/mcp-transport.ts
```diff
@@ -1,9 +1,14 @@
...
-    : new SSEClientTransport(config);
+  if (config.type === 'stdio') {
+    throw new MCPClientError({
+      message:
+        'The stdio transport configuration has been deprecated in favor of passing in a custom transport.',
+    });
+  }
+
+  return new SSEClientTransport(config);
 }
```
lgrammel, line 14: move the `MCPTransport` interface here? (personally I find `types` files confusing and prefer to have types, functions, etc in files with matching names)

lgrammel: and potentially transportconfig as well

lgrammel: related: this also applies for things like `utils.ts` etc. - not a huge fan of such collection files

packages/ai/core/tool/mcp/types.ts
```diff
@@ -42,11 +40,14 @@ export type McpToolSet<TOOL_SCHEMAS extends ToolSchemas = 'automatic'> =
...
+ */
+interface McpStdioServerConfig {
   command: string;
   args?: string[];
   env?: Record<string, string>;
-  stderr?: IOType | Stream | number;
+  stderr?: any | number;
   cwd?: string;
   type: 'stdio';
 }
```
lgrammel, line 53: we might need to make a hard breaking change and just remove it (fine bc things are experimental)

packages/ai/core/tool/mcp/types.ts
```diff
@@ -139,13 +140,40 @@ export const JSONRPCMessageSchema = z.union([
 ]);
 export type JSONRPCMessage = z.infer<typeof JSONRPCMessageSchema>;
```
lgrammel, line 141: Let's move all the JSON RPC related types to e.g. `json-rpc-message.ts`

packages/ai/mcp-stdio/index.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  StdioClientTransport,
+  type McpStdioServerConfig,
```
lgrammel, line 3: There is a mismatch here - for consistency we should either prefix both or none with `Mcp`

Also why `ClientTransport` vs `Transport`? could simplify, also if `Client` is used elsewhere. Same for `Server`:

```
export {
   StdioTransport,
   StdioConfig
}
```

packages/ai/mcp-stdio/create-child-process.ts
```diff
@@ -1,6 +1,6 @@
 import type { ChildProcess } from 'node:child_process';
-import { MCPClientError } from '../../../errors';
-import { McpStdioServerConfig } from './types';
+import { MCPClientError } from '../errors';
+import { McpStdioServerConfig } from './mcp-stdio-transport';
```
lgrammel, line 3: we might be able to reverse some of the hacks in here (no dynamic imports needed if we do separate bundling, so the hack become accidental complexity that should be removed)

packages/ai/core/tool/mcp/mcp-sse-transport.ts
```diff
@@ -16,9 +16,9 @@ export class SSEClientTransport implements MCPTransport {
     close: () => void;
   };
 
-  onClose?: () => void;
-  onError?: (error: unknown) => void;
-  onMessage?: (message: JSONRPCMessage) => void;
+  onclose?: () => void;
```
lgrammel, line 19: Can we rename the class to `SSEMCPTransport` (I don't like long caps sequence but it's important to have consistent patterns imo)

packages/ai/core/tool/mcp/types.ts
```diff
@@ -139,13 +140,40 @@ export const JSONRPCMessageSchema = z.union([
...
 
+/**
+ * Transport interface for MCP (Model Context Protocol) communication.
+ */
```
lgrammel, line 145: Is this similar to what the original MCP library has? if so might be worth leaving a // note for developers

packages/ai/tsup.config.ts
```diff
@@ -62,4 +62,12 @@ export default defineConfig([
...
     },
   },
+  // MCP stdio
+  {
+    entry: ['mcp-stdio/index.ts'],
+    outDir: 'mcp-stdio/dist',
+    format: ['cjs', 'esm'],
+    dts: true,
+    sourcemap: true,
+  },
```
lgrammel, line 72: awesome! would be good to try using this in a next application to see if there are any warnings/errors

content/cookbook/05-node/54-mcp-tools.mdx
lgrammel: nevermind, saw that you did the split already. import might be missing

lgrammel:
Please add a patch changeset.

packages/ai/mcp-stdio/index.ts
iteratetograceness: The naming here is explicit because:

- The transport we make available is specific to MCP clients!
- The config is essentially the server params you pass to the client transport!

I can see how the naming is a bit confusing but this aligns with official SDK naming conventions, happy to pivot away on our end!

packages/ai/core/tool/mcp/mcp-sse-transport.ts
iteratetograceness: Similar to the above, the reason I wanted to keep `Client` in the name is to ensure users are aware this isn't a server transport!

lgrammel: Hm - I don't think we'll add MCP servers anytime soon, so I'm inclined to vote for simplifying and not aligning to the original MCP lib on this (I don't want to much of their decisions bleed into the AI SDK, the lowercase callbacks were an exception)

iteratetograceness: Noted!

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -639,35 +639,44 @@ Create an MCP client using either:
...
+  transport,
 });
 
 // Example of using MCP with SSE
-const mcpClient = await createMCPClient({
+const mcpClient = await experimental_createMCPClient({
   transport: {
     type: 'sse',
     url: 'https://my-server.com/sse',
   },
```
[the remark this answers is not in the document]
iteratetograceness: On it! Adding reference docs for both SSEMCPTransport and StdioMCPTransport

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -639,35 +639,49 @@ Create an MCP client using either:
...
 });
+// Can also be written as:
+const mcpClient = await experimental_createMCPClient({
+  transport: new SSEMCPTransport({
```
lgrammel, line 662: it already looks ugly, so wdyt about `Experimental_SSEMCPTransport`? 

Also open to ideas on how to make it look nicer - CamelCase maybe?

lgrammel: Or `Sse`?

packages/ai/mcp-stdio/mcp-stdio-transport.test.ts
```diff
@@ -17,7 +17,7 @@ interface MockChildProcess {
 }
 
 describe('MCPStdIOTransport', () => {
```
lgrammel, line 19: nit, looks a bit weird

.changeset/cold-apples-design.md
```diff
@@ -0,0 +1,9 @@
+---
+'next-openai': patch
+'ai-sdk-mcp-example': patch
```
lgrammel, line 3: rm examples

.changeset/cold-apples-design.md
```diff
@@ -0,0 +1,9 @@
...
+'ai': patch
+---
+
+feat: enable custom mcp transports
+breaking change: remove internal stdio transport creation
+chore: update next-openai example with mcp tools
```
lgrammel, line 9: rm - we don't include example changes

lgrammel (approved):

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -639,35 +639,43 @@ Create an MCP client using either:
...
 - `stdio`: Uses standard input and output streams for communication, ideal for local tool servers running on the same machine (like CLI tools or local services)
 - `SSE` (Server-Sent Events): Uses HTTP-based real-time communication, better suited for remote servers that need to send data over the network
+- Custom transport: Bring your own transport by implementing the `MCPTransport` interface
 
 ```typescript
 // Example of using MCP with stdio
-const mcpClient = await createMCPClient({
-  transport: {
-    type: 'stdio',
+const mcpClient = await experimental_createMCPClient({
```
lgrammel, line 646: show import(s)? also possible to rename to non-experimental in import

PR #5211, opened 2025-03-14
feat (providers/cohere): add command-a model id to chat settings

shaper:


lgrammel (approved):

lgrammel:
Can you update the provider model list in the docs as well?

Issue #4996, opened 2025-02-26
Problems upgrading from Bedrock v1 to v2


elie222:
### Description

Same credentials used with Bedrock provider work in v1, but not in v2.

The problem may be related to this note, but I couldn't find a variation that worked:
https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock#migrating-to-ai-sdkamazon-bedrock-2x

I tried setting `sessionToken` to undefined explicitly too.

### Code example

v1:
```ts
      llmModel: createAmazonBedrock({
        bedrockOptions: {
          region: env.BEDROCK_REGION,
          credentials: {
            accessKeyId: env.BEDROCK_ACCESS_KEY,
            secretAccessKey: env.BEDROCK_SECRET_KEY,
          },
        },
      })(model),
```

v2:
```ts
    llmModel: createAmazonBedrock({
        region: env.BEDROCK_REGION,
        accessKeyId: env.BEDROCK_ACCESS_KEY,
        secretAccessKey: env.BEDROCK_SECRET_KEY,
      })(model),
```
https://github.com/elie222/inbox-zero/commit/18ffdad4ada2af7536507e9912a0392d3e5dcbbc

### AI provider

@ai-sdk/amazon-bedrock v2.0.5

shaper:
Hi, if you add `sessionToken: undefined` does it work?

See also this note: https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock#migrating-to-ai-sdkamazon-bedrock-2x and also mentioned in a note here: https://sdk.vercel.ai/providers/ai-sdk-providers/amazon-bedrock#provider-instance

shaper:
Try replacing:

`sessionToken: resolvedCredentials?.sessionToken || undefined,`

with

`sessionToken: undefined`

To force that it not be defined, since you likely aren't/weren't intending to use it (you weren't in your initial example), and the old logic using bedrockOptions would leave it undefined, but new logic may use an env var if present.

shaper:
Ack -- I re-read your intro and see you said you did, a couple of times actually:

> I tried setting sessionToken to undefined explicitly too.

Hmm. The absence of any errors is also not what I would expect. Can you share the code that's producing `resolvedCredentials`? If you log the credentials before create, do they look like what you'd expect?

ben-bwh:
I'm actually not the one who opened this issue, I'm just having the same problem so decided to piggyback on. 

[...]

PR #5220, opened 2025-03-14
feat (provider/amazon-bedrock): support AWS credential providers

### LINKED ISSUE #2780: Bedrock client does not work natively with profiles, instance roles, container roles, etc.

niranjan94:
### Add support for AWS credential providers in the amazon-bedrock provider

This PR adds a `credentialProvider` prop to the `AmazonBedrockProviderSettings`. This allows the consumers of this provider to make use of the credential providers from AWS SDK to ensure the amazon-bedrock provider can work dynamic short lived credentials such as those from instance profiles, instance roles, container roles, EKS IRSA, etc with are recommended practices for AWS applications to avoid usages of long lasting access keys/secret access keys. 

No additional dependency has been added to the amazon-bedrock provider package. 
The responsibility of creating the credential provider lies with the consumer of this package.

Resolves #2780 

#### Example usage

```ts
import { fromNodeProviderChain } from '@aws-sdk/credential-providers';

const bedrock = createAmazonBedrock({
  region: 'us-east-1',
  credentialProvider: fromNodeProviderChain()
});
```

---

Though there is a workaround suggested by @shaper at https://github.com/vercel/ai/issues/2780#issuecomment-2660787835 , that has some drawbacks:
- That causes the model creation to have to be done asynchronously. Which makes it a different behaviour than every other provider in the AI SDK (As a result, also makes it harder to use with other libraries such as https://github.com/mastra-ai/mastra which depend on model creation being non-async) 
- Since credentials in these cases are short-lived, the amazon bedrock provider has to be initialized on every usage to ensure the latest credentials are always loaded from the credentials chain

Related as well to some discussion in #4996


turlockmike:
This is really important for a lot of companies that rely on bedrock for hosting. Just want to +1 this

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -17,11 +17,11 @@ export interface BedrockCredentials {
 Creates a fetch function that applies AWS Signature Version 4 signing.
 
 @param getCredentials - Function that returns the AWS credentials to use when signing.
-@param originalFetch - Optional original fetch implementation to wrap. Defaults to global fetch.
+@param fetch - Optional original fetch implementation to wrap. Defaults to global fetch.
 @returns A FetchFunction that signs requests before passing them to the underlying fetch.
  */
 export function createSigV4FetchFunction(
-  getCredentials: () => BedrockCredentials,
+  getCredentials: () => BedrockCredentials | Promise<BedrockCredentials>,
```
lgrammel, line 24: PromiseLike?

niranjan94: Makes sense. Done ✅ 
Thanks :)

shaper (approved):

shaper:
Thanks! This is a really nice contribution. I appreciate your thoughtful PR description and iterating on it with us as well. I only added a couple of tests and a bit more tsdoc for the new setting. 🥇

PR #5232, opened 2025-03-16
Enhancement: Added single pass for cosine Calculation

akulabs8:
Added single pass for cosine calculation to improve the performance


lgrammel:
Doesn't this have the same computational complexity as before?

lgrammel:
That said, I like it because it's simpler, even tho the performance improvements are most likely marginal.

PR #5236, opened 2025-03-17
add support for custom headers in SSE transport configuration

mikiastilahun:


content/docs/07-reference/01-ai-sdk-core/21-create-mcp-client.mdx
```diff
@@ -91,6 +91,12 @@ This feature is experimental and may change or be removed in the future.
                       type: 'string',
                       description: 'URL of the MCP server',
                     },
+                    {
+                      name: 'headers',
+                      type: 'Record<string, string>',
+                      isOptional: true,
+                      description: 'An Object of headers',
```
iteratetograceness, line 98:
```suggestion
                      description:
        'Additional HTTP headers to be sent with requests.',
```

packages/ai/core/tool/mcp/mcp-sse-transport.ts
```diff
@@ -147,6 +150,9 @@ export class SSEClientTransport implements MCPTransport {
     try {
       const headers = new Headers();
       headers.set('Content-Type', 'application/json');
+      Object.entries(this.headers ?? {}).forEach(([key, value]) => {
```
iteratetograceness, line 153: nit: would be cleaner to do something like:

`const headers = new Headers({ ...this.headers })` and then set the content type

packages/ai/core/tool/mcp/mcp-sse-transport.ts
```diff
@@ -37,6 +39,7 @@ export class SSEClientTransport implements MCPTransport {
           const response = await fetch(this.url.href, {
             headers: {
               Accept: 'text/event-stream',
```
iteratetograceness, line 41: q: thoughts on spreading custom headers first and then setting `Accept` to avoid overriding?

iteratetograceness (commented):
we should add tests!

PR #5293, opened 2025-03-20
chore (ai): improve cosine similarity calculation

lgrammel:
Continues #5232


packages/ai/core/util/cosine-similarity.ts
```diff
@@ -22,8 +22,7 @@ export function cosineSimilarity(
   } = {
     throwErrorForEmptyVectors: false,
   },
-) {
-  // TODO: In the next major version, change the default value of throwErrorForEmptyVectors to true
```
lgrammel, line 26: fine to leave as is / not change in next major

nicoalbanese (approved):

PR #5295, opened 2025-03-20
feat (providers/xai): add xai image model support

shaper:


content/providers/01-ai-sdk-providers/01-xai.mdx
```diff
@@ -126,3 +126,45 @@ The following optional settings are available for xAI chat models:
...
+### Model-specific options
+
+You can customize the image generation behavior with model-specific settings:
+
+```ts
+import { xai } from '@ai-sdk/xai';
+import { experimental_generateImage as generateImage } from 'ai';
+
+const { image } = await generateImage({
+  model: xai.imageModel('grok-2-image', {
```
jeremyphilemon, line 158: shouldn't it be `xai.image('...')` for consistency?

shaper: The `ProviderV1` interface uses `imageModel` [ref](https://github.com/vercel/ai/blob/5bf05106162fd6f2cd74deb65251b2fe13f1375c/packages/provider/src/provider/v1/provider-v1.ts#L41)

I am not sure how @lgrammel is thinking about the shorthand methods that aren't formally part of the provider interface. I do like them, but this seems like the required/latest. We can add the shorthand in a later PR if we think appropriate?

shaper: Added `image` as well.

jeremyphilemon (approved):

PR #5303, opened 2025-03-21
feat (provider/openai): pdf support for chat language models

lgrammel:
Continues #5300

### DIFF
--- a/.changeset/good-dragons-do.md
+++ b/.changeset/good-dragons-do.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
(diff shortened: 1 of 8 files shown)


laneparton:
@lgrammel - Do you know if there will be support PDFs in the https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/tool-result-content.ts#L3 ?

lgrammel:
@laneparton are there any LLMs that support this?

laneparton:
@lgrammel - Ahhh I think I misunderstood how `experimental_toToolResultContent` works 😞 

I was hoping, just like images, we could return PDFs to pass along.

PR #5327, opened 2025-03-22
Add support for `taskType` and `autoTruncate` parameters for VertexAI Embeddings Model

abir-taheer:
Please review official Vertex AI Docs Here: 

https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api

More specifically it lists a `taskType` parameter: 

> Used to convey intended downstream application to help the model produce better embeddings. If left blank, the default used is RETRIEVAL_QUERY.

As well as an `autoTruncate` parameter:

> When set to true, input text will be truncated. When set to false, an error is returned if the input text is longer than the maximum length supported by the model. Defaults to true.

Currently these are inaccessible via the Vercel AI SDK so I've added them to the settings and updated the implementation of the embedding model for Vertex-AI to be able to pass in these parameters.

Before:

```typescript
const model = vertex.textEmbeddingModel('text-embedding-004', {
  outputDimensionality: 512, // optional, number of dimensions for the embedding
});
```

After:

```typescript
const model = vertex.textEmbeddingModel('text-embedding-004', {
  outputDimensionality: 512, // optional, number of dimensions for the embedding
  taskType: 'RETRIEVAL_DOCUMENT', // optional, task type for which embeddings will be used
  autoTruncate: true, // optional, whether to auto-truncate text exceeding token limit
});

```

### DIFF
--- a/content/providers/01-ai-sdk-providers/16-google-vertex.mdx
+++ b/content/providers/01-ai-sdk-providers/16-google-vertex.mdx
@@ -563,6 +563,8 @@ Google Vertex AI embedding models support additional settings. You can pass them
 ```ts
 const model = vertex.textEmbeddingModel('text-embedding-004', {
   outputDimensionality: 512, // optional, number of dimensions for the embedding
+  taskType: 'RETRIEVAL_DOCUMENT', // optional, task type for which embeddings will be used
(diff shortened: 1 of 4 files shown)


aayush-kapoor:

PR #5364, opened 2025-03-25
fix(amazon-bedrock): image conversion in tool use

### LINKED ISSUE #5133: Tool result content is incorrectly serialized as a string when calling bedrock
### Description

When converting messages to the bedrock API, tool responses appear to have their entire content serialized as a JSON string within a `text` content block, rather than handling each individual piece of content appropriately.

This breaks handling of image responses from tools, because the base64 of the image ends up just being sent to the model as text.

This seems very similar to https://github.com/vercel/ai/issues/3379, but with bedrock rather than anthropic. 

The current behaviour produces requests like:

```
{
  "role": "user",
  "content": [
    {
      "toolResult": {
        "toolUseId": "tooluse_PBqR3AclQ2eObQsXbNzx5w",
        "content": [
          {
            "text": "[{\"type\":\"text\",\"text\":\"Tool Response\"},{\"type\":\"image\",\"data\":\"MQ==\",\"mimeType\":\"image/png\"}]"
          }
        ]
      }
    }
  ]
}
```

### Code example

```js
import { z } from 'zod';


async function main() {

    model: bedrock("us.anthropic.claude-3-7-sonnet-20250219-v1:0"),
    messages: [{
      role: 'user',
      content: [{ type: "text", text: "Please download this image https://picsum.photos/200/300 and tell me what you see" }]
    }],
    tools: {
      submit: tool({
        description: 'Download an image',
        parameters: z.object({
          url: z
            .string()
            .describe('The image URL'),
        }),
        execute: async ({ url }) => {
          const bytes = await (await fetch(url)).bytes();
          return { bytes }
        },
        experimental_toToolResultContent(result) {
          return [{
            type: "image",
            data: Buffer.from(result.bytes).toString("base64"),
            mimeType: "image/jpeg",
          }]
        },
      }),
    },
    maxSteps: 5,
  });

  console.log(result.text);
}

main().catch(console.error);
```

The model will completely hallucinate the contents of the image, because it only has access to the base64 data as a string.

### AI provider

@ai-sdk/amazon-bedrock v2.1.1

### Additional context

I think the issue is caused by this line https://github.com/vercel/ai/blame/9b081a7a2241e9796d8b8561e71fc5ba9dd2bdf6/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts#L138, which should probably be reusing the existing content serialisation logic rather than JSON.stringify.

dreamorosi:
This PR fixes #5133.

This PR improves the handling of different content types in the `amazon-bedrock` package when working with tool usage and the experimental `toToolResultContent` function.

Specifically, the PR updates the logic that converts results to Bedrock messages by adding support for handling both text and image content.

Before this PR, as reported in the linked issue, this logic always assumed the result content was a text string and converted it to a JSON string, causing hallucinations.

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-api-types.ts
+++ b/packages/amazon-bedrock/src/bedrock-api-types.ts
@@ -115,7 +115,7 @@ export interface BedrockImageBlock {
 export interface BedrockToolResultBlock {
   toolResult: {
     toolUseId: string;
-    content: Array<{ text: string }>;
+    content: Array<BedrockTextBlock | BedrockImageBlock>;
   };
 }
 
--- a/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts
+++ b/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts
@@ -3,9 +3,11 @@ import {
   BedrockAssistantMessage,
   BedrockCachePoint,
   BedrockDocumentFormat,
+  BedrockImageBlock,
   BedrockImageFormat,
   BedrockMessages,
   BedrockSystemMessages,
+  BedrockTextBlock,
   BedrockUserMessage,
 } from './bedrock-api-types';
 import {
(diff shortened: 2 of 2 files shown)


shaper:
Continuing this work with your commits to track credit in https://github.com/vercel/ai/pull/5390

dreamorosi:
Thank you for looking at the PR, and sorry for not adding tests - I was unsure if the current test suite was enough because I saw it already tested for images, but after looking at your PR I agree I should have.

Will keep it in mind for the next PR.

PR #5370, opened 2025-03-25
feat (docs): add streaming not working when proxied troubleshooting page

chasewoo:
I struggled for a whole day to understand why Hono's stream had results but couldn't be generated one by one in the form of a stream. Finally, I found that the problem was caused by Next.js rewrites proxy. Next.js's proxy compresses all responses by default.

So, I added those lines into docs to prevent others from spending more time due to the compression middleware.


lgrammel:
@chasewoo this is a super important issue, I'm wondering if the Hono example is the right place for it. The content encoding affects what happens based on deployment, not on backend framework, and more users will run into. I'm wondering if there should be a dedicated page in the docs that discusses the response settings (content type, encoding) in relation to deployments.

chasewoo:
Screenshot 2025-03-26 at 07 55 42

Is it work if we add a new Troubleshooting item in Vercel Deployment Guide?

lgrammel:
@chasewoo yes, that's a good starting point (esp streaming not working). This is probably the right page to add to:

- https://sdk.vercel.ai/docs/troubleshooting/streaming-not-working-when-deployed

But I'm thinking it goes beyond that and might (eventually, after adding to the troubleshooting page) deserve a standalone page earlier in the guide (related to deployment) that could then also be linked from e.g. the Hono example.

chasewoo:
Hi @lgrammel I've updated the PR.

lgrammel (approved):

PR #5371, opened 2025-03-25
fix(core): response.messages is empty in onFinish and onStepFinish callbacks

mxzinke:
The issue occurred because in step-finish events, response.messages wasn't  properly populated for continuation steps. Now message data is correctly  included in both onFinish and onStepFinish callbacks.

Added tests to verify both callbacks receive the expected message data.

### DIFF
--- a/packages/ai/core/generate-text/__snapshots__/stream-text.test.ts.snap
+++ b/packages/ai/core/generate-text/__snapshots__/stream-text.test.ts.snap
@@ -1010,7 +1010,19 @@ exports[`streamText > options.maxSteps > 4 steps: initial, continue, continue, c
       "response": {
         "headers": undefined,
         "id": "id-0",
-        "messages": [],
+        "messages": [
+          {
+            "content": [
+              {
+                "text": "part 1 
+ ",
+                "type": "text",
+              },
+            ],
+            "id": "msg-0",
+            "role": "assistant",
+          },
+        ],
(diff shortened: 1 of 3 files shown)


mxzinke:
> What was the scenario that caused this PR? did you run into any issues with the existing behavior?

We were using the messages from response on the callbacks onStepFinished/onFinish when streaming the text. The use case was for saving the message history to our DB - also when interruption happened (e.g. by anthropic overloaded error).

I upgraded from v4.1.46 to v4.2.2, and this broke our application.

lgrammel:
Can you elaborate on the breakage? It could be unrelated.

The current implementation is there for a reason - so you can get correct information on each step.

mxzinke:
@lgrammel You were right, thank you. We are using Anthropic, and it includes empty text messages on tool calls like this:

```json
{
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": ""
    },
    {
      "type": "tool-call",
      "toolCallId": "toolu_01Dq5oaGRX8EVoMVNoac4Awy",
      "toolName": "search_knowledge",
      "args": {
        "query": "...."
      }
    }
  ],
  "id": "msg-23W2M3ds7brNnFacNYwtNSkC"
}
``` 

I just misinterpreted the issue. I don't know if this is a bug, but for now, we are just filtering them out. Thanks for helping!

PR #5379, opened 2025-03-25
feat (providers/openai-compatible): add openai-compatible image model and use as xai image model base

shaper:


packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -10,12 +11,15 @@ import { OpenAICompatibleCompletionLanguageModel } from './openai-compatible-com
...
+import { OpenAICompatibleImageSettings } from './openai-compatible-image-settings';
+import { OpenAICompatibleImageModel } from './openai-compatible-image-model';
 
 export interface OpenAICompatibleProvider<
   CHAT_MODEL_IDS extends string = string,
   COMPLETION_MODEL_IDS extends string = string,
   EMBEDDING_MODEL_IDS extends string = string,
-> extends ProviderV1 {
+  IMAGE_MODEL_IDS extends string = string,
+> extends Omit<ProviderV1, 'imageModel'> {
```
shaper, line 22: This oddity is due to the provider spec `imageModel` being `readonly` which I am sure was done for a reason. `openai-compatible` is kind of a unique thing in its intended-for-extension so I think this is ok, but open to other approaches.

lgrammel: was intentional; but worth revisiting and removing the readonly if it causes problems

lgrammel (approved):

packages/openai-compatible/src/openai-compatible-provider.ts
shaper: sg, will see if it crops up again, this `Omit` doesn't seem untoward yet, it feels like one of those "make sure you know what/why you're doing it" kind of things.

Issue #5378, opened 2025-03-25
Unable to set mimeType and providerOption for images


wenjunche:
### Description

Hello
I am using azure provider and trying to send a base64 encoded jpeg.   Here is my code for creating the image message:

```
                {
                    type: 'image',
                    image: imageBase64_string,
                    mimeType: 'image/jpeg',
                    providerOptions: {
                        azure: {
                            detail: 'low'
                        }
                    }
                }

//  calling generateText with the image in the messages

```
and network trace shows:

![Image](https://github.com/user-attachments/assets/c22d87dd-649a-44cd-9cf9-306119cc11ff)

looks like it is sending `png `instead, and missing `detail`.

I am using `"@ai-sdk/azure": "1.3.2"` and `"ai": "4.2.4"`

Thanks

iteratetograceness:
Hi @wenjunche! Updating our docs but a few things:

1. Internally, we have a util that detects image mimetype and converts accordingly, overriding the user-defined `mimeType`. I would verify that your `imageBase64_string` isn't actually a PNG!
2. For `providerOptions`: because you're using an OpenAI model via Azure, you can set image details like so:

```
providerOptions: { openai: { imageDetail: 'low' } }
```

wenjunche:
that works.   thanks

lgrammel:
## Identify issue type

Identified as **Documentation**  
Confidence: **High**

### Rationale
The reported behavior was resolved through usage clarification rather than an SDK behavior change; the remaining action was to document the MIME detection and provider option namespace.

### Evidence
- Maintainer explained that the SDK detects the image MIME type and overrides the supplied `mimeType`.
- Azure-hosted OpenAI models require `providerOptions.openai.imageDetail`, not `providerOptions.azure.detail`.
- The maintainer stated they were updating the documentation, and the reporter confirmed the corrected configuration works.

PR #5402, opened 2025-03-26
chore(azure): provider option docs and example

iteratetograceness:
related to: https://github.com/vercel/ai/issues/5378

### DIFF
--- a/examples/ai-core/src/generate-text/azure-image.ts
+++ b/examples/ai-core/src/generate-text/azure-image.ts
@@ -1,0 +1,39 @@
+import { azure } from '@ai-sdk/azure';
+import { generateText } from 'ai';
+import 'dotenv/config';
+import fs from 'node:fs';
+
+async function main() {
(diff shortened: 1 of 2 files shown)


examples/ai-core/src/generate-text/azure-image.ts
```diff
@@ -0,0 +1,39 @@
...
+    messages: [
+      {
+        role: 'user',
+        content: [
+          { type: 'text', text: 'Describe the image in detail.' },
+          {
+            type: 'image',
+            image: imageBase64_string,
+            // Even if you mislabel the mime type, internally the AI SDK will detect and use the correct mime type:
+            mimeType: 'image/svg+xml',
```
lgrammel, line 21: is possible to just leave out the mimetype (this should be the recommended way as it has the best dx and we auto-recognize images)

lgrammel (approved):

Issue #5435, opened 2025-03-28
Separate Text Parts with Merged Streams in processChatResponse


BrianHung:
### Feature Description

I'm merging data streams of individual streamTexts to enable a multi-step workflow.
https://x.com/aisdk/status/1896946680081302007

Right now, the behavior of `processChatResponse` is to accumulate the text parts and reasoning text parts of separate streams into the same part.

So for multiple streams, parts would always be [TextPart, ToolCallPart, ToolCallPart].

What I'm looking for is, [TextPart, ToolCallPart, TextPart, ToolCallPart].

### Use Cases

With merged streams, we want to keep the text interweaved and not accumulate.

Before

![Image](https://github.com/user-attachments/assets/a11b2756-dfbb-4eab-b781-d3b9a4701eef)

After

![Image](https://github.com/user-attachments/assets/703ef852-7a68-4c96-9685-48166e28f509)

BrianHung:
Turns out latest version of vercel ai already has this implemented here:
https://github.com/vercel/ai/blame/4dece70cbc9606ea0c0c7921b10f43b8cce96791/packages/ui-utils/src/process-chat-response.ts#L361-L364

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This was a request for a stream-processing capability rather than a failure of promised behavior. Although the capability was already available in the latest version, the issue's substantive request is still feature-oriented.

### Evidence
- The description explicitly requests separate text parts when multiple streams are merged, changing the resulting part sequence to preserve interleaving.
- No documented or promised behavior is cited as being violated; the request is framed as a desired workflow enhancement.
- The author later notes that the latest AI SDK version already implements the requested behavior.

PR #5436, opened 2025-03-28
allow for text part interweaving

BrianHung:
Implements #5435 

By resetting text state after tool result and finish, we allow for text interweaving within parts array.


lgrammel:
Ideally the step start / step end boundaries should be sufficient. The proposed approach can have unintended side effects and potentially introduce bugs in other scenarios.

`onFinishStepPart` already does reset text and reasoning. Why does this not work in your scenario?

BrianHung:
Oh it turns out I was using a slightly outdated branch and the reset text on finish step works; vercel ai moves fast!

PR #5441, opened 2025-03-29
Clarify useObject framework support in 01-overview.mdx

Sultron:
According to the documentation for useObject (https://sdk.vercel.ai/docs/reference/ai-sdk-ui/use-object),   "useObject is an experimental feature and only available in React and SolidJS."  

However, the overview page did not reflect this limitation.   To maintain consistency, I have updated the framework support table by adding a ❌ (cross) for useObject in Svelte.


lgrammel:
The documentation of useObject is incorrect. There is a Svelte version called StructuredObject. The overview table is correct.

gr2m:
> The documentation of useObject is incorrect. There is a Svelte version called StructuredObject. The overview table is correct.

https://github.com/vercel/ai/blob/9ae327d7ba4176993769509b37c32369f186821d/packages/svelte/src/index.ts#L5

PR #5450, opened 2025-03-29
Feat: Allow onStepFinish to optionally mutate AI Execution Context

theswerd:
This is a prototype for a feature I really want in the ai sdk. I want to be able to dynamically change the environment of the execution during execution runs, including everything around the message history.

use case for us is changing environment variables in code execution, but i think it could also be useful for changing the model dynamically based on arbitrary params, or changing the job based in the system prompt. 

Using this I can make the AI get progressively ruder in the system prompt after someone runs out of credits.

Don't think its mergable yet, just putting it here for thoughts


gr2m:
Since nobody chimed in since March, I don't think there is a need. Did you find a workable workaround for your use case?

aayush-kapoor:

PR #5483, opened 2025-04-01
fix: reset error in use assistant

### LINKED ISSUE #4615: Error doesn't clear on useAssistant
### Description

If a response has an error, then `error` from useAssistant will have a value, but even after a successful response, the error value persists. As far as I can tell, there is no way to clear this (no setError etc.).

### Code example

_No response_

### AI provider

_No response_

### Additional context

_No response_

ulentini:
Close #4615

When using the `useAssistant` hook, if an error occurs, submitting a subsequent request does not clear the previous error.


lgrammel:

Issue #5459, opened 2025-03-30
Chatbot 'Stop' Button Causes ResponseAborted Error


brandonhays3:
### Description

It looks like the chatbot should call onFinish with result.finishReason === "stop" when the user stops the chat, but instead, it's throwing a "ResponseAborted" error. This prevents chat messages from being saved correctly since onFinish is not called.

You can reproduce this issue on [chat.vercel.ai](https://chat.vercel.ai/) by requesting a long response and then pressing the stop button. Instead of handling the stop gracefully, it throws an error.

The expected behavior should be:

When the user stops the chat, the streaming process should terminate cleanly.

onFinish should be called with { finishReason: "stop" }, ensuring messages are saved properly.

The app should not throw a "ResponseAborted" error.

Let me know if you need help debugging or implementing a fix!

lgrammel:
Can you tell me more about your environment and how to reproduce?

tunde-alao:
Hi, I'm also experiencing the same issue on my end. I'm using Next.js, and simple starting a chat and stopping by calling the stop() from useChat

lgrammel:
@tunde-alao which next js version are you using? which `@ai-sdk/react` version?

tunde-alao:
@lgrammel I'm currently using @ai-sdk/react@1.1.20

tunde-alao:
I just updated to @ai-sdk/react@1.2.4, and the same issue is still there. This only happens when result.consumeStream(); is present before return result.toDataStreamResponse(). But we need to add that so that onSave is called when when the chat streaming is stopped.

brandonhays3:
You can reproduce the issue with Vercel's AI example chatbot: https://vercel.com/templates/next.js/nextjs-ai-chatbot
https://github.com/vercel/ai-chatbot/blob/main/app/(chat)/api/chat/route.ts

(My project is setup similarly, with the latest version installed) 

To reproduce:
1. Ask the chatbot to "write an essay about nextjs" on https://chat.vercel.ai/
2. Pause it before it completes
3. Then ask it a follow up question
4. You will notice it says "Oops, an error occurred!" 
5. The user cannot send any further messages, and the saving logic doesn't save the assistant's partial response

This is because an error is getting thrown, but I am assuming the library is setup to call onFinish instead, given that onFinish returns a result with finishReason with possible value "stop.". 

Ideally, when a user leaves the page or presses the stop button or the stream stops, it should call onFinish. 

[...]

PR #5492, opened 2025-04-01
fix (core): consume stream on abort

iteratetograceness:
Addresses https://github.com/vercel/ai/issues/5459

This PR ensures that the `consumeStream` method on `streamText` does not throw unhandled errors. 

NOTE: This is an **edge case** that occurs when an `abortSignal` is passed to `streamText`; the aborted signal propagates to `consumeStream` as a `ResponseAborted` error but we were not catching this error. This then triggered a network error, caught by the `useChat` hook (setting the hook status to `error`).


examples/next-openai/app/api/use-chat-resilient-persistence/route.ts
```diff
@@ -1,32 +1,38 @@
...
+    // consume the stream to ensure it runs to completion and triggers onFinish
+    // even when the client response is aborted (e.g. when the browser tab is closed).
+    result.consumeStream(); // no await
 
-  return result.toDataStreamResponse();
+    return result.toDataStreamResponse();
+  } catch (error) {
+    console.error('[CATCH]', error);
+    return new Response('Internal Server Error', { status: 500 });
+  }
```
lgrammel, line 37: try..catch is meaningless here bc `toDataStreamResponse` immediately returns

packages/ai/util/consume-stream.ts
```diff
@@ -9,9 +9,29 @@
...
+      try {
+        const { done, value } = await reader.read();
+        if (done) break;
+        await writer.write(value);
+      } catch (error) {
+        // Ignore abort errors, continue reading:
+        if (
+          error instanceof Error &&
+          error.name !== 'AbortError' &&
+          error.name !== 'ResponseAborted'
```
lgrammel, line 27: Would be nice to have an `isAbortError` helper.

packages/ai/util/consume-stream.ts
```diff
@@ -9,9 +9,29 @@
  * @returns {Promise<void>} A promise that resolves when the stream is fully consumed.
  */
 export async function consumeStream(stream: ReadableStream): Promise<void> {
```
lgrammel, line 11: Can you add tests for `consume-stream` that show the issue?

packages/ai/util/consume-stream.ts
```diff
@@ -9,9 +9,29 @@
...
-  while (true) {
-    const { done } = await reader.read();
-    if (done) break;
+
+  try {
+    while (true) {
+      try {
+        const { done, value } = await reader.read();
+        if (done) break;
+        await writer.write(value);
```
lgrammel, line 21: We do not need a writer?

packages/ai/core/generate-text/stream-text.ts
```diff
@@ -1593,8 +1593,19 @@ However, the LLM results are expected to be small enough to not cause issues.
...
+    } catch (error) {
+      // Ignore abort errors, continue processing:
+      if (
+        error instanceof Error &&
+        error.name !== 'AbortError' &&
+        error.name !== 'ResponseAborted'
+      ) {
+        throw error;
+      }
     }
```
lgrammel, line 1609: use `consumeStream` helper instead?

packages/ai/core/generate-text/stream-text.test.ts
```diff
@@ -1542,6 +1542,70 @@ describe('streamText', () => {
...
+            },
+          }),
+        }),
+        prompt: 'test-input',
+      });
+
+      await expect(result.consumeStream()).resolves.not.toThrow();
+    });
+
+    it('should propagate other errors during stream consumption', async () => {
```
lgrammel, line 1590: Wondering if that's the desired behavior - might be good to think through use cases

lgrammel:
How does this affect the useChat UI parts?

iteratetograceness:
> How does this affect the useChat UI parts?

Before this change, because `result.consumeStream` would throw an unhandled response aborted error, and useChat would update status to error (disabling sending of subsequent messages)!

More details [here](https://github.com/vercel/ai/issues/5459#issuecomment-2770037547)

lgrammel:
> > How does this affect the useChat UI parts?
> 
> Before this change, because `result.consumeStream` would throw an unhandled response aborted error, and useChat would update status to error (disabling sending of subsequent messages)!
> 
> More details [here](https://github.com/vercel/ai/issues/5459#issuecomment-2770037547)

So consume stream would break an ongoing response? (asking because at that point consume stream should run async and separately of the function)

lgrammel:
The consumeStream resilience is very good regardless (and maybe it should swallow all errors or have an error callback to not break any server).

iteratetograceness:
@lgrammel heard!

Also, I'm not crazy, I did see status set to `error` when we do not properly swallow abort errors from consumeStream! See video:

https://github.com/user-attachments/assets/4b7ee477-e8f3-4398-80a3-b489270b4ba7

If `stop` fires while long response is streaming, `consumeStream`'s unhandled abort error seems to bubble up in triggerRequest's `catch` as a network error!

lgrammel:
@iteratetograceness that is very interesting - do you know what happens (on the network level)? is the next js server forwarding the error through the stream somehow?

other server might behave differently

iteratetograceness:
> @iteratetograceness that is very interesting - do you know what happens (on the network level)? is the next js server forwarding the error through the stream somehow?
> 
> other server might behave differently

I don't think this is Next.js server specific! I believe the unhandled network error is bubbling up to the nearest error handler in the event loop (which prior to these changes would be triggerRequest's catch)

iteratetograceness:
@lgrammel oh but I am seeing a slightly different edge case now! If a _subsequent message isn't sent_ after the stopping, onFinish does not trigger. Any inklings on what causes this? I'm gonna take a closer look at the stream logic in stream-text.ts

iteratetograceness:
@lgrammel tldr, when the abort signal is passed into the streamText, consumeStream does not trigger onFinish (the onFinish callback is only called as a side effect of subsequent messages' flushes)

im stepping away for a bit but will come back to dig deeper and flesh out the appropriate solution (e.g. maybe something along the lines of adding new `user-cancellation` finishReason and adding logic for this case)

packages/ai/core/generate-text/stream-text.test.ts
iteratetograceness: esp. given the network error finding - i like the idea of swallowing all errors and exposing an onError callback option!

packages/ai/util/consume-stream.ts
iteratetograceness: just realized we already have one!

examples/next-openai/app/api/use-chat-resilient-persistence/route.ts
```diff
@@ -26,7 +26,10 @@ export async function POST(req: Request) {
 
   // consume the stream to ensure it runs to completion and triggers onFinish
   // even when the client response is aborted (e.g. when the browser tab is closed).
-  result.consumeStream(); // no await
+  // no await
+  result.consumeStream(error => {
+    console.log('Error during background stream consumption: ', error); // optional error callback
+  });
```
iteratetograceness, line 32: Added to illustrated optional error callback

packages/ai/util/consume-stream.ts
iteratetograceness: I reverted some of my changes here to keep this low-level! Error handling (e.g. checking for abort errors) should be the responsbility of the caller

content/docs/07-reference/01-ai-sdk-core/02-stream-text.mdx
```diff
@@ -2274,13 +2274,21 @@ To see `streamText` in action, check out [these examples](#examples).
     },
     {
       name: 'consumeStream',
-      type: '() => Promise<void>',
+      type: '(onError?: (error: unknown) => void) => Promise<void>',
```
lgrammel, line 2277: prefer to use optional options object `{ onError? }` to ensure future extensibility

packages/ai/core/generate-text/stream-text-result.ts
```diff
@@ -203,8 +203,10 @@ Consumes the stream without processing the parts.
 This is useful to force the stream to finish.
 It effectively removes the backpressure and allows the stream to finish,
 triggering the `onFinish` callback and the promise resolution.
+
+If an error occurs, it is passed to the optional `onError` callback.
   */
-  consumeStream(): Promise<void>;
+  consumeStream(onError?: (error: unknown) => void): Promise<void>;
```
lgrammel, line 209: prefer options object

packages/ai/util/consume-stream.ts
```diff
@@ -10,8 +10,12 @@
...
-  while (true) {
-    const { done } = await reader.read();
-    if (done) break;
+  try {
+    while (true) {
+      const { done } = await reader.read();
+      if (done) break;
+    }
+  } finally {
+    reader.releaseLock();
```
lgrammel, line 19: great find!

packages/ai/core/generate-text/stream-text.ts
```diff
@@ -1591,10 +1592,11 @@ However, the LLM results are expected to be small enough to not cause issues.
     );
   }
 
-  async consumeStream(): Promise<void> {
-    const stream = this.fullStream;
-    for await (const part of stream) {
-      // no op
+  async consumeStream(onError?: (error: unknown) => void): Promise<void> {
+    try {
+      await consumeStream(this.fullStream);
```
lgrammel, line 1597: wonder if it's worth pushing the onError object into `consumeStream` to keep the behavior between `consumeStream` here and the general helper aligned?

lgrammel: the helper is afaik only used internally so we can change the main `stream` object to become a parameter object 

```
{
  stream,
  onError?
}
```

packages/provider-utils/src/is-abort-error.ts
```diff
@@ -1,6 +1,8 @@
 export function isAbortError(error: unknown): error is Error {
   return (
     error instanceof Error &&
-    (error.name === 'AbortError' || error.name === 'TimeoutError')
+    (error.name === 'AbortError' ||
+      error.name === 'TimeoutError' ||
+      error.name === 'ResponseAborted')
```
lgrammel, line 6: wonder if we should document which name is triggered when / how

lgrammel: is this needed if consume streams swallows all errors?

iteratetograceness: good call!

lgrammel (approved):

ItayElgazar:
is it release already? I am still experiencing the bug, even after upgrading to v`4.3.9`

iteratetograceness:
@ItayElgazar this fix has already been released. I recommend opening a new discussion/issue with more context or minimal repro steps!

PR #5509, opened 2025-04-02
chore (ui/react,vue): update more tests to unified test server

samdenty:
This updates more tests to use the unified test server


packages/react/src/use-object.ui.test.tsx
```diff
@@ -69,208 +73,180 @@ describe('text stream', () => {
...
+        controller.close();
+
+        // wait for element "loading" to have text content "false":
+        await waitFor(() => {
+          expect(screen.getByTestId('loading')).toHaveTextContent('false');
+        });
+      });
+    });
+
+    describe('stop', () => {
```
lgrammel, line 130: is the separate `describe` needed here? if it's just a single `it` we can remove the `describe` also above)

packages/provider-utils/src/test/unified-test-server.ts
```diff
@@ -62,6 +62,10 @@ class TestServerCall {
     return this.request!.text().then(JSON.parse);
   }
 
+  get requestCredentials() {
```
lgrammel, line 65: needs a `provider-utils` patch changeset

packages/react/src/use-object.ui.test.tsx
samdenty: i'd use semantic diff, githubs diffs are wrong https://app.semanticdiff.com/gh/vercel/ai/pull/5509/changes#packages/react/src/use-object.ui.test.tsx?ignore_comments=false

samdenty: oh wait i see what you mean, yeah we can

lgrammel (approved):

PR #5517, opened 2025-04-02
docs: update baseten community provider docs

philipkiely-baseten:
Hi! I work at Baseten, and we recently removed our dependency on a bridge endpoint -- models are now OpenAI compatible by default when served with TensorRT-LLM/vLLM/etc

This simplifies our integration with tools like the Vercel AI SDK, so I updated the docs and example in the SDK to reflect this!


lgrammel (approved):

philipkiely-baseten:
Hey @lgrammel thank you for merging! qq when will the updated docs be live?

PR #5518, opened 2025-04-02
feat: typescript project references

samdenty:
This adds project references to the AI SDK codebase.

## Improvements

- Clone codebase, install and start getting intellisense without first running build prior
- Use find all references to see everywhere something is being used
- Go to definition now goes to source files
- Rename symbol across all examples, tests etc.
- **Faster full typecheck** 17sec -> 1sec

## Possibly breaking changes

Typescript requires the source folders to match output folders, so we would require a bunch of difference tsconfigs with circular dependencies which aren't supported.

So this PR changes the packages with multiple dist folders to instead from:

```
dist/index.js
anthropic/dist/index.js
...
```

to

```
dist/index.js
dist/anthropic/index.js
...
```

This could be a breaking change if people are depending on the filepath in the package. But if they're importing the package directly, it'll use the import map that i've updated to now use the new format (so should continue working)


.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.defaultFormatter": "esbenp.prettier-vscode"
```
samdenty, line 3: i thought i'd set this since we use prettier, not sure if needed

lgrammel (changes requested):
Let's chat on how to best move forward w/ this. Too risky to land on main given the impact and breakages.

packages/rsc/src/types.test-d.ts
```diff
@@ -1,16 +1,8 @@
 import { expectTypeOf } from 'vitest';
 
-import type { StreamableValue } from '../dist';
+import type { StreamableValue } from '.';
 
 describe('StreamableValue type', () => {
-  it('should not contain types marked with @internal after compilation', () => {
```
samdenty, line 6: this is now not possible to do since we don't have separate dist builds

.github/workflows/quality.yml
```diff
@@ -77,4 +77,4 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Run TypeScript type check
-        run: pnpm run type-check
+        run: pnpm run type-check:full
```
samdenty, line 80: full does it with examples, normal tsconfig does without so it's faster

examples/ai-core/package.json
```diff
@@ -42,11 +42,12 @@
   "scripts": {
     "test:e2e:all": "vitest run src/e2e/*.test.ts",
     "test:file": "vitest run",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --build"
```
lgrammel, line 45: where will the js files go?

packages/ai/core/prompt/convert-to-core-messages.ts
```diff
@@ -6,12 +6,7 @@ import {
   ToolInvocationUIPart,
 } from '../types';
 import { ToolSet } from '../generate-text/tool-set';
-import {
-  AssistantContent,
-  CoreMessage,
-  ToolCallPart,
-  ToolResultPart,
-} from '../prompt';
+import { AssistantContent, CoreMessage, ToolCallPart, ToolResultPart } from '.';
```
lgrammel, line 9: preference is to directly import from the source file (i.e. 4 imports here)

packages/ai/core/prompt/convert-to-language-model-prompt.ts
```diff
@@ -275,7 +275,7 @@ function convertPartToLanguageModelPart(
 
   let mediaType: string | undefined = part.mediaType ?? part.mimeType;
   let data: DataContent | URL;
-  let content: DataContent | URL | string;
+  let content: URL | DataContent;
```
lgrammel, line 278: same order?

packages/ai/package.json
```diff
@@ -9,17 +9,14 @@
   "types": "./dist/index.d.ts",
   "files": [
     "dist/**/*",
-    "mcp-stdio/**/*",
```
lgrammel, line 12: this needs to remain

packages/ai/core/prompt/convert-to-language-model-prompt.ts
samdenty: oops yep, i'll revert

lgrammel (approved):

PR #5523, opened 2025-04-03
fix (provider/google-vertex): fix anthropic support for image urls in messages

shaper:


lgrammel (approved):

packages/google-vertex/src/anthropic/google-vertex-anthropic-provider.ts
```diff
@@ -100,6 +100,7 @@ export function createVertexAnthropic(
         baseURL,
         headers: options.headers ?? {},
         fetch: options.fetch,
+        supportsImageUrls: false,
```
modestaspruckus, line 103: Why is this set to false?? @lgrammel

modestaspruckus: @shaper

modestaspruckus: SDK is throwing errors `URL sources are not supported` for vertex now, it is not converting urls to base64 anymore

modestaspruckus: It seems broken after .14 version https://github.com/vercel/ai/commit/aeba38eb490390dadb67a2c79665afbe5cc28d50#diff-6b314517fe13fd78e5d279e3bede3255a255a1ac00b6ca85e7f00835c282a892R57-R61 because this method `supportsUrl` was introduced.

PR #5528, opened 2025-04-03
fix: should trigger request again when throttle is not undefined

coverlv:
when `experimental_throttle` is set, occasionally useChat() will not trigger request again, because the `messageRef.current` is not the newest messages.

reproduction:
```ts
const { append } = useChat({
  maxSteps: 5,
  experimental_throttle: 100
});
```

Root cause:
```ts
// Keep the latest messages in a ref.
const messagesRef = useRef<UIMessage[]>(messages || []);
useEffect(() => {
  messagesRef.current = messages || [];
}, [messages]);

// ...

onUpdate({ message, data, replaceLastMessage }) {
   // has latency here
  throttledMutate(
    newMessages,
    false,
  );
});

const messages = messagesRef.current;
if (
  shouldResubmitMessages({
    originalMaxToolInvocationStep: maxStep,
    originalMessageCount: messageCount,
    maxSteps,
    messages,
  })
)// return false;
```

### DIFF
--- a/.changeset/silent-nails-taste.md
+++ b/.changeset/silent-nails-taste.md
@@ -1,6 +1,0 @@
----
-'@ai-sdk/google-vertex': patch
-'@ai-sdk/anthropic': patch
----
-
-fix (provider/google-vertex): fix anthropic support for image urls in messages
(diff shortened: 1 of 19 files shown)


aayush-kapoor:

PR #5539, opened 2025-04-03
feat: improved text splitter

### LINKED ISSUE #5231: smoothStream does not support CJK characters
### Description

https://github.com/user-attachments/assets/e6f499a5-f0ce-456f-adc0-9aabfff1b6aa

I checked the implementation of smoothStream and found that it chunks by English characters. For CJK languages, we need chunk by characters.

I think we can set a `chunking:'char'` to support CJK languages.

I am willing to implement this feature

### Code example

_No response_

### AI provider

_No response_

### Additional context

_No response_

samdenty:
This integrates the custom text-splitter API I created in react-ai-flow, for better handling of splitted elements.

Features:

Spaces are always trailing on the last element, the regex is created over the whole string so you can use lookahead 

Closes https://github.com/vercel/ai/issues/5231


packages/ai/core/generate-text/smooth-stream.ts
```diff
@@ -2,10 +2,12 @@ import { InvalidArgumentError } from '@ai-sdk/provider';
...
 import { TextStreamPart } from './stream-text-result';
 import { ToolSet } from './tool-set';
+import { TextSplit, splitText } from './text-splitter';
 
 const CHUNKING_REGEXPS = {
-  word: /\s*\S+\s+/m,
-  line: /[^\n]*\n/m,
+  character: /(?!\s)(?=.)/g,
+  word: /[\u4E00-\u9FFF\u3040-\u309F\u30A0-\u30FF]|\s+/gm,
+  line: /\r\n|\r|\n/g,
```
lgrammel, line 10: what will happen to custom user regexp?

samdenty: i'll add more tests for custom chunking

packages/ai/core/generate-text/text-splitter.ts
```diff
@@ -0,0 +1,106 @@
+export interface SplitOptions {
+  /**
+   * When true, delimiters will be separate matches rather then extending the matches near them
+   * @default false
+   */
+  separateDelimiters?: boolean;
```
lgrammel, line 6: is this needed? can we simplify?

lgrammel:
This breaks/changes the current behavior - might be okay but I wonder if there's something simpler we can do.

lgrammel:
@samdenty I think overall I'd like to understand what the simplest thing is that we could do to support kanjii etc, vs revamping the whole algorithm.

samdenty:
closing in favour of https://github.com/vercel/ai/pull/5548

PR #5542, opened 2025-04-03
chore: extract rsc to separate package

samdenty:
This removes & moves over the `ai/rsc` bundle to instead be in `@ai-sdk/react/rsc`. And also removes the dependency on react from the `ai` package

- [x] create separate `ai/internal` bundle with the newly needed exports (does duplication for now, separate PR will fix that)
- [x] update references in docs
   - [x] subtask - add migration guide TBD in future closer to release
- [x] remove react & other dependencies from `ai` package
- [x] move tests to `rsc` package (✅ confirmed to all be running in new package)


packages/ai/streams/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from '../core/index';
 export * from '../errors/index';
+export * from '../util/constants';
```
samdenty, line 3: i wasn't sure with stuff like this to mark as @internal, use a separate package or something else?

package.json
```diff
@@ -11,6 +11,7 @@
     "update-references": "update-ts-references",
     "prettier-check": "prettier --check \"**/*.{js,ts,tsx,md,mdx,svelte}\"",
     "type-check": "tsc --build",
+    "type-check:full": "tsc --build tsconfig.with-examples.json",
```
lgrammel, line 14: which one will run in CI? important that we keep full checks there

packages/ai/package.json
```diff
@@ -44,17 +44,6 @@
...
-      "types": "./dist/rsc/index.d.ts",
-      "react-server": "./dist/rsc/rsc-server.mjs",
-      "import": "./dist/rsc/rsc-client.mjs"
-    },
-    "./react": {
-      "types": "./dist/react/index.d.ts",
-      "react-server": "./dist/react/index.server.mjs",
-      "import": "./dist/react/index.mjs",
-      "require": "./dist/react/index.js"
-    },
```
lgrammel, line 57: prefer copy not full move on main

lgrammel: so we can deprecate

packages/ai/streams/index.ts
lgrammel: we will merge ui-utils and ai in v5, so some of this may not be relevant - what constants need exporting?

packages/react/src/rsc/streamable-ui/create-streamable-ui.tsx
```diff
@@ -1,4 +1,4 @@
-import { HANGING_STREAM_WARNING_TIME_MS } from '../../util/constants';
+import { HANGING_STREAM_WARNING_TIME_MS } from 'ai';
```
lgrammel, line 1: I see - let me think about it. We may as well duplicate the value and remove the constant.

lgrammel: (since I'll clean up the `ai` parts that use it most likely in v5)

tools/tsconfig/base.json
```diff
@@ -9,6 +9,7 @@
     "forceConsistentCasingInFileNames": true,
     "inlineSources": false,
     "isolatedModules": true,
+    "module": "ESNext",
```
lgrammel, line 12: is there a risk that this breaks common js usage of the ai sdk?

lgrammel: needed for the rsc move?

package.json
samdenty: i've updated it to run `type-check:full` in CI. I made type-checking faster in the IDE that uses the default tsconfig (including examples in it makes it way slower)

packages/ai/package.json
samdenty: sounds good 👍

packages/react/src/rsc/streamable-ui/create-streamable-ui.tsx
samdenty: fixed

tools/tsconfig/base.json
samdenty: i think this was required for one of the builds to work when moving to the react package. This just affects how typescript compiles to JS - but we pass it through tsup so it converted at later stage back to commonjs

packages/ai/streams/index.ts
samdenty: resolved by making ai/internal export

tools/tsconfig/base.json
samdenty: it should have no effect - but i'm going to thoroughly inspect the compiled package to ensure this next

content/cookbook/20-rsc/20-stream-text.mdx
```diff
@@ -20,14 +20,14 @@ Text generation can sometimes take a long time to complete, especially when you'
 
 ## Client
 
-Let's create a simple React component that will call the `generate` function when a button is clicked. The `generate` function will call the `streamText` function, which will then generate text based on the input prompt. To consume the stream of text in the client, we will use the `readStreamableValue` function from the `ai/rsc` module.
+Let's create a simple React component that will call the `generate` function when a button is clicked. The `generate` function will call the `streamText` function, which will then generate text based on the input prompt. To consume the stream of text in the client, we will use the `readStreamableValue` function from the `@ai-sdk/react/rsc` module.
```
lgrammel, line 23: wonder if we need to split it out into `@ai-sdk/react/rsc` or if we can integrate directly into `@ai-sdk/react`? (no need to change in this PR, can be a separate PR if we decide to do this)

packages/ai/package.json
```diff
@@ -21,18 +21,15 @@
     "lint": "eslint \"./**/*.ts*\"",
     "type-check": "tsc --noEmit",
     "prettier-check": "prettier --check \"./**/*.ts*\"",
-    "test": "pnpm test:node && pnpm test:edge && pnpm test:ui && pnpm test:e2e",
+    "test": "pnpm test:node && pnpm test:edge && pnpm test:e2e",
```
lgrammel, line 24: `e2e` tests are only for rsc - they can be moved

packages/ai/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./node_modules/@vercel/ai-tsconfig/react-library.json",
+  "extends": "./node_modules/@vercel/ai-tsconfig/base.json",
   "compilerOptions": {
```
lgrammel, line 3: can the `dom`, `dom.iterable` libs be removed below? (worth a try)

lgrammel (approved):

lgrammel (approved):
Please add a changeset for `ai` and `@ai-sdk/rsc`

PR #5543, opened 2025-04-03
refactor (provider-utils): move `customAlphabet()` method from `nanoid` into codebase

gr2m:
@lgrammel and I discussed that [`nanoid`](https://github.com/ai/nanoid/) could be removed as dependency, to further reduce the number of external dependencies.

`@ai-sdk/provider-utils` is the only package that had a dependency on `nanoid`, and it only used the [`customAlphabet()` method](https://github.com/ai/nanoid/blob/c49cc113499f2b44ab52b80314e18043ad0f3f79/non-secure/index.js#L12C12-L23) from the non-secure module export.

### Todos

- [x] merge target should be `v5` branch
- [x] no vercel ai copyright header needed; however we should keep original copyright if copied from nanoid - see example here: https://github.com/vercel/ai/blob/main/packages/ui-utils/src/deep-partial.ts
- [x] add a patch changeset for the `provider-utils` pkg (with PR title as description)

### DIFF
--- a/packages/provider-utils/src/generate-id-custom-alphabet.ts
+++ b/packages/provider-utils/src/generate-id-custom-alphabet.ts
@@ -3,7 +3,7 @@
 //
 // MIT License
 //
-// Copyright (c) 2020–2024 AI
+// Copyright (c) 2017 Andrey Sitnik <andrey@sitnik.ru>
 // Copyright (c) Vercel, Inc. (https://vercel.com)
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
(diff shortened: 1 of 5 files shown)


lgrammel:
Sweet! Can you make the following tweaks:

- merge target should be `v5` branch
- no vercel ai copyright header needed; however we should keep original copyright if copied from nanoid - see example here: https://github.com/vercel/ai/blob/main/packages/ui-utils/src/deep-partial.ts

packages/provider-utils/src/generate-id-custom-alphabet.ts
```diff
@@ -3,7 +3,7 @@
 //
 // MIT License
 //
-// Copyright (c) 2020–2024 AI
+// Copyright (c) 2017 Andrey Sitnik <andrey@sitnik.ru>
```
gr2m, line 6: honestly I have no idea where I got the `2020–2024 AI` from, I thought I copied it from nanoid's license text

lgrammel (approved):

Issue #5552, opened 2025-04-04
LangChainAdapter uses deprecated StreamData() method


jayadehart:
### Description

Using StreamData() currently gives this error:

⚠ Error (TS6385)  | 
StreamData is deprecated.
(alias) new StreamData(): StreamData
import StreamData
A stream wrapper to send custom JSON-encoded data back to the client.

@deprecated — Please use createDataStream, createDataStreamResponse, and pipeDataStreamToResponse instead.

### Code example

`packages/ai/streams/langchain-adapter.ts`

````
export function toDataStreamResponse(
  stream:
    | ReadableStream<LangChainStreamEvent>
    | ReadableStream<LangChainAIMessageChunk>
    | ReadableStream<string>,
  options?: {
    init?: ResponseInit;
    data?: StreamData;
    callbacks?: StreamCallbacks;
  },
) {
  const dataStream = toDataStreamInternal(
    stream,
    options?.callbacks,
  ).pipeThrough(new TextEncoderStream());
  const data = options?.data;
  const init = options?.init;

  const responseStream = data
    ? mergeStreams(data.stream, dataStream)
    : dataStream;

  return new Response(responseStream, {
    status: init?.status ?? 200,
    statusText: init?.statusText,
    headers: prepareResponseHeaders(init?.headers, {
      contentType: 'text/plain; charset=utf-8',
      dataStreamVersion: 'v1',
    }),
  });
}
````

lgrammel:
reworked in ai sdk 5

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
The issue reports an existing adapter relying on an API that the SDK itself deprecated, causing a concrete TypeScript warning. This is an implementation inconsistency rather than a request for documentation or a new capability.

### Evidence
- LangChainAdapter's `toDataStreamResponse` still accepts and merges the deprecated `StreamData` API, producing TypeScript deprecation warning TS6385.
- The maintainer stated that this was reworked in AI SDK 5, confirming the adapter implementation required a product-code change.

PR #5553, opened 2025-04-04
Langchain adapter fix deprecated

jayadehart:
I updated the langchain method to no longer use the deprecated StreamData() method. I also updated the documentation to reflect this change. I added a test case to ensure that the data stream is correctly being merged into the response stream!

Resolves this issue: #5552


lgrammel:
We are planning to make larger changes in this area in `v5`, not merging for now.

sicter:
Just checking, would this also fix: #5650?

PR #5555, opened 2025-04-04
chore: deprecate ai/rsc in favour of @ai-sdk/react/rsc

samdenty:
This moves over the `ai/rsc`  to instead be in `@ai-sdk/react/rsc`. This pull request deprecates but keeps the implementation. In the next [version v5 pull request](https://github.com/vercel/ai/pull/5542) we removed it entirely.

We're deprecating it beforehand to give users time to notice & update their code

PR #5556, opened 2025-04-04
chore: remove deprecated ai/react

samdenty:
This removes the deprecated `ai/react` bundle so we don't run into a cyclic error in the next PR https://github.com/vercel/ai/pull/5555 to deprecate ai/rsc.


samdenty:
closing for now as i'm not sold this is best way to tackle it since we probably want to give people more time to upgrade

PR #5583, opened 2025-04-07
fix (ai/mcp): prevent mutation of customEnv

cgoinglove:
## Description
This PR fixes an issue where the original `customEnv` object passed to functions in `get-environment.ts` was being mutated due to JavaScript's pass-by-reference behavior for objects.

## Problem
When a configuration object with an `env` property was passed to `createChildProcess`, any modifications to the environment inside the function would also affect the original object, potentially causing unexpected behavior in subsequent code.

In our project, we encountered this issue specifically when using `serverConfig` with the MCP client:

```typescript
const serverConfig = await findMcpServerConfig(id);
// {
//   command: 'node server.js',
//   env: { PORT: '3000' }
// };

      name,
      onUncaughtError,
      transport: new StdioMCPTransport(serverConfig),
});

return Response.json(...)
```

After the `createMCPClient` call, we discovered that the original `serverConfig.env` was being modified. This could cause problems when:
1. The same config object is reused for multiple operations
2. The config needs to be stored back to the database in its original form
3. The config is used for comparison or validation later in the code

This change preserves the immutability of configuration objects throughout the application, making the code more predictable and preventing hard-to-trace bugs that might occur when objects are unexpectedly modified.

image

### DIFF
--- a/.changeset/fix-env-mutation.md
+++ b/.changeset/fix-env-mutation.md
@@ -1,5 +1,0 @@
----
-"@ai/core": patch
----
-
-Fix Environment Object Mutation in `get-environment.ts`
--- a/.changeset/fix-env-mutation.md
+++ b/.changeset/fix-env-mutation.md
@@ -1,0 +1,5 @@
+---
+"@ai/core": patch
+---
+
+Fix Environment Object Mutation in `get-environment.ts`
--- a/.changeset/fix-env-mutation.md
+++ b/.changeset/fix-env-mutation.md
@@ -1,0 +1,5 @@
+---
+'@ai/core': patch
+---
+
+fix (ai/mcp): prevent mutation of customEnv
--- a/packages/ai/mcp-stdio/create-child-process.test.ts
+++ b/packages/ai/mcp-stdio/create-child-process.test.ts
@@ -52,6 +52,22 @@ describe('createChildProcess', () => {
     childProcessWithCustomEnv.kill();
   });
 
+  it('should not mutate the original environment object', async () => {
+    const originalConfig = {
+      command: process.execPath,
+      env: { CUSTOM_VAR: 'custom_value' },
+    };
+    const originalConfigCopy = JSON.parse(JSON.stringify(originalConfig));
+
+    const childProcessWithCustomEnv = await createChildProcess(
+      originalConfig,
+      new AbortController().signal,
+    );
+
+    expect(originalConfig.env).toEqual(originalConfigCopy.env);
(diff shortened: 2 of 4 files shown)


lgrammel:
1. please add a patch changeset for ai with the PR title as description
2. please add tests for get-environment.ts that reproduce the issue

.changeset/fix-env-mutation.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ai/core": patch
+---
+
+Fix Environment Object Mutation in `get-environment.ts`
```
lgrammel, line 5: change to `fix (ai/mcp): prevent mutation of customEnv`

packages/ai/mcp-stdio/create-child-process.test.ts
```diff
@@ -52,6 +52,22 @@ describe('createChildProcess', () => {
...
+    const originalConfigCopy = JSON.parse(JSON.stringify(originalConfig));
+
+    const childProcessWithCustomEnv = await createChildProcess(
+      originalConfig,
+      new AbortController().signal,
+    );
+
+    expect(originalConfig.env).toEqual(originalConfigCopy.env);
+    childProcessWithCustomEnv.kill();
+  });
```
lgrammel, line 69: move to separate file `get-environment.test.ts` and limit testing to `getEnvironment`

lgrammel (approved):

packages/ai/mcp-stdio/get-environment.test.ts
```diff
@@ -0,0 +1,13 @@
+import { describe, it, expect } from 'vitest';
+import { getEnvironment } from './get-environment';
+
+describe('getEnvironment', () => {
+  it('should not mutate the original custom environment object', () => {
+    const customEnv = { CUSTOM_VAR: 'custom_value' };
+
+    const result = getEnvironment(customEnv);
+
+    expect(customEnv).toEqual({ CUSTOM_VAR: 'custom_value' });
```
lgrammel, line 10: `toStrictEqual`

cgoinglove:
@lgrammel  Thanks for your review! I learned a lot from your feedback. 

Let's gooo 🚀

.changeset/fix-env-mutation.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai/core': patch
```
lgrammel, line 2: there was a bug here (fixed it on main) - for future reference, you can create changesets by running `pnpm changeset` in the project root

PR #5588, opened 2025-04-07
fix: support mixed message types in `convertToCoreMessages`

### LINKED ISSUE #5473: providerOptions are lost when UIMessages are included in messages passed to core functions
### Description

I believe this is the root cause of at least some of the issues seen in https://github.com/vercel/ai/issues/5227 .

The logic in [detectPromptType](https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/detect-prompt-type.ts) declares a prompt to be of form 'ui-messages' if *any* of the messages provided have "UI-specific parts." That means that if mixed message types are passed in, they are all treated as UIMessages.

That breaks this pattern, which I would expect to be a fairly common use case:

```javascript
messages: [
  {
    role: 'system',
    content: systemPrompt,
    providerOptions: {
      anthropic: { cacheControl: { type: 'ephemeral' } },
    },
  },
  ...userMessages, // messages passed in from the UI
];
```

In that case, the providerOptions field is stripped from all messages in the list by [convertToCoreMessages](https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/convert-to-core-messages.ts#L22), which reconstructs all the messages including only role, content, and experimentalAttachments.

Here's my simple repro case:

> ```javascript   
> const messages = [
>       {
>         role: 'user',
>         content: 'Please process this message',
>         providerOptions: {
>           anthropic: { cacheControl: { type: 'ephemeral' } },
>         },
>       },
>     ];
> 
> const generateResult = await generateText({
>       model: anthropic('claude-3-7-sonnet-20250219');,
>       messages: messages,
>  });
> ```
> 
> successfully sends the following to Anthropic:
> 
> ```
> {
>   role: 'user',
>   content: [
>     {
>       type: 'text',
>       text: 'Please process this message',
>       cache_control: [Object]
>     }
>   ]
> }
> ```
> However this:
> 
> ```javascript   
> const messages = [
>       {
>         role: 'user',
>         content: 'Please process this message',
>         providerOptions: {
>           anthropic: { cacheControl: { type: 'ephemeral' } },
>         },
>       },
>       {
>         role: 'data',
>         content: 'some data',
>       },
>     ];
> 
>  const generateResult = await generateText({
>       model: anthropic('claude-3-7-sonnet-20250219');,
>       messages: messages,
> });
> ```
> Still sends only the first message, but without the cache_control parameter:
> 
> ```
> {
>   role: 'user',
>   content: [ { type: 'text', text: 'Please process this message' } ]
> }
> ```

I could potentially fix this by calling convertToCoreMessages on the UIMessages before passing them to generateText (/streamText/etc), but there's not any indication that should be necessary. 

Potential solutions?

1. Core AI functions throw an error if the message array passed in contains an array of mixed message types, rather than defaulting to UIMessage processing, so users know they must call convertToCoreMessage on their UIMessages
2. convertToCoreMessage leaves messages unchanged if they are already in CoreMessage form
3. convertToCoreMessage retains providerOptions specifically

iteratetograceness:
Refactors `convertToCoreMessages` to avoid assumption that all messages passed in are of type `UIMessage`. Ensures that if any `CoreMessage` types are passed in, we return that as is to avoid removing core properties.

Resolves https://github.com/vercel/ai/issues/5473

### DIFF
--- a/content/docs/02-foundations/04-tools.mdx
+++ b/content/docs/02-foundations/04-tools.mdx
@@ -94,6 +94,7 @@ There are several providers that offer pre-built tools as **toolkits** that you 
 - **[agentic](https://github.com/transitive-bullshit/agentic)** - A collection of 20+ tools. Most tools connect to access external APIs such as [Exa](https://exa.ai/) or [E2B](https://e2b.dev/).
 - **[browserbase](https://docs.browserbase.com/integrations/vercel-ai/introduction)** - Browser tool that runs a headless browser
 - **[Stripe agent tools](https://docs.stripe.com/agents)** - Tools for interacting with Stripe.
+- **[StackOne ToolSet](https://docs.stackone.com/agents)** - Agentic integrations for hundreds of [enterprise SaaS](https://www.stackone.com/integrations)
 - **[Toolhouse](https://docs.toolhouse.ai/toolhouse/using-vercel-ai)** - AI function-calling in 3 lines of code for over 25 different actions.
 - **[Agent Tools](https://ai-sdk-agents.vercel.app/?item=introduction)** - A collection of tools for agents.
 - **[AI Tool Maker](https://github.com/nihaocami/ai-tool-maker)** - A CLI utility to generate AI SDK tools from OpenAPI specs.


packages/ai/core/prompt/convert-to-core-messages.test.ts
```diff
@@ -838,6 +838,38 @@ describe('convertToCoreMessages', () => {
...
+          content: 'It is sunny in Tokyo.',
+          providerOptions: {
+            anthropic: { cacheControl: { type: 'ephemeral' } },
+          },
+        },
+        {
+          role: 'user',
+          content: 'What is the weather in Tokyo?',
+        },
+      ]);
```
lgrammel, line 869: Aren't these all core messages?

packages/ai/core/prompt/convert-to-core-messages.ts
```diff
@@ -13,21 +13,30 @@ import {
...
+import { detectSingleMessageCharacteristics } from './detect-prompt-type';
 import { MessageConversionError } from './message-conversion-error';
 
 /**
 Converts an array of messages from useChat into an array of CoreMessages that can be used
 with the AI core functions (e.g. `streamText`).
  */
 export function convertToCoreMessages<TOOLS extends ToolSet = never>(
-  messages: Array<Omit<Message, 'id'>>,
+  messages: Array<Omit<Message, 'id'> | CoreMessage>,
```
lgrammel, line 24: I'm wondering if we should go down the route to allow mixed arrays as input, or if we should catch it earlier and throw and error (because it might be messed up data on the user side). What are your thoughts about the trade-offs?

iteratetograceness: I see your point! I had opted for this since we offer provider options at the message level _and_ function level (also will respond and let the user know they should pass to generateText)

But I'm not sure if it's worth introducing a breaking change by throwing / preventing a mixed array of messages, e.g. in situations where users are dynamically appending messages, because the `messages` parameter does technically accept all kinds. I wonder how many users are actually doing this!

Not a strongly held opinion, happy to enforce uniform messages array if you feel strongly!

lgrammel: > I had opted for this since we offer provider options at the message level and function level (also will respond and let the user know they should pass to generateText)

Interesting - my sense is that there is a larger problem here, we need to provider the user with a standardized way of adding provider options that are part-specific (e.g. openai image detail) when they have ui messages. Currently they would need to run `convertToCoreMessages` and then map the core messages, but I don't think we have this documented.

My concern is that we enable the user to have a myriad of patterns if we support mixed messages, and then later we have to support that (also in persistence patterns).

2 alternatives I can think of:
- supporting `providerOptions` on UI messages and parts (somewhat reluctant to do this)
- documenting a pattern of how to add image part details (even if it's clunky)

Not saying that we should not support mixed arrays as you suggested, just want to have a better understanding what it would mean for the users and also for our own maintenance / compatibility in the future.

iteratetograceness: For posterity:

Spoke outside PR to pivot to a documentation update approach in consideration of future maintenance!

PR #5612, opened 2025-04-08
chore: update docs on `providerOption` levels

iteratetograceness:
Adds section to [Prompts](https://sdk.vercel.ai/docs/foundations/prompts) docs page to differentiate between the different "levels" at which `providerOptions` are applied

Addresses confusion in https://github.com/vercel/ai/issues/5473#issuecomment-2787189564

### DIFF
--- a/content/docs/02-foundations/03-prompts.mdx
+++ b/content/docs/02-foundations/03-prompts.mdx
@@ -554,7 +554,3 @@ const result = await generateText({
   ],
 });
 ```
-
-```
-
-```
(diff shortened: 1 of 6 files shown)


content/docs/02-foundations/03-prompts.mdx
```diff
@@ -476,3 +554,7 @@ const result = await generateText({
   ],
 });
 ```
+
+```
+
+```
```
lgrammel, line 560: rm

lgrammel (approved):

PR #5626, opened 2025-04-09
Support `additionalProperties` in open api schema

### LINKED ISSUE #5628: Requests to vertex ai fail if `z.record(z.string())` used in tool schema
### Description

See the code example below. When using vertex ai if `z.record(z.string())` is used an error occurs because in the generated open api schema `something` becomes `undefined`

### Code example

```ts
import { createVertex } from '@ai-sdk/google-vertex'
import { CoreMessage, Tool, generateText, tool, LanguageModelV1 } from 'ai'

const vertex = createVertex({
  project: ...,
  location: ...,
})

const gemini2Flash: LanguageModelV1 = vertex.languageModel(
  'gemini-2.0-flash-001',
)

await generateText({
  model: gemini2Flash,
  prompt: 'Hi',
  tools: {
    test: tool({
      description: 'Test',
      parameters: z.object({
        // the issue
        something: z.record(z.string())
      }),
    }
  }
})
```

### AI provider

@ai-sdk/google-vertex 2.2.14

### Additional context

I opened a PR to fix this here: https://github.com/vercel/ai/pull/5626

tjenkinson:
`z.record(z.string())` as a property in tool parameters currently gets converted to `undefined`, which is invalid and causes an error from vertex ai.

This now gets converted to `{ type: 'object', additionalProperties: { type: 'string' } }`, which works

fixes https://github.com/vercel/ai/issues/5628


himanshusinghs (commented):
Looks good! Hope one of the code owners take a look at this soon :)

himanshusinghs:
Ok it seems that Google APIs do not support `additionalProperties`. Infact the docs for both [Gemini API](https://ai.google.dev/api/caching#Schema) and [Vertex API](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#schema-examples) mentions that only a subset of OpenAI Schema are supported and that subset does not mention `additionalProperties` at all. I suppose ideally, this property should then be omitted from the translated schema?

lgrammel:
Landed on `v5`: https://github.com/vercel/ai/pull/6919

lgrammel (approved):

himanshusinghs:
@lgrammel the patch in incorrect. Google function declarations does not support `additionalProperties` and so it should be removed, not added.

Here is a reference - https://ai.google.dev/api/caching#FunctionDeclaration

Just for reference - how langchain [handles it](https://github.com/langchain-ai/langchainjs/blob/451f4568d6f66d8feeeb79aa414bb41a8b913373/libs/langchain-google-genai/src/utils/zod_to_genai_parameters.ts#L26-L57) is by removing the property recursively entirely.

himanshusinghs:
This is what I was thinking to properly address this - https://github.com/vercel/ai/pull/6933

PR #5660, opened 2025-04-10
chore (provider): allow both binary and base64 file content (spec)

lgrammel:
## Background
We are currently converting all file parts to `base64` in `ai`. This can lead to double-conversion and performance issues if the provider needs to send binary data.

## Summary
This pull request changes the behavior to send either base64 or binary into the provider. The provider than does the necessary conversion, eliminating the need for double conversions.


samdenty (approved):
lgtm!

PR #5674, opened 2025-04-10
chore(providers/cohere): convert to providerOptions

samdenty:
## Tasks


packages/cohere/src/cohere-embedding-model.ts
```diff
@@ -50,9 +45,16 @@ export class CohereEmbeddingModel implements EmbeddingModelV1<string> {
     values,
     headers,
     abortSignal,
+    providerOptions,
```
lgrammel, line 48: this is an embedding model. provider options are not supported on embedding models

samdenty: resolving, as we added providerOptions to embedding models now

lgrammel (approved):

PR #5698, opened 2025-04-11
feat(embedding-model-v2): add providerOptions

samdenty:
## Summary

This adds providerOptions to embedding model v2 for future PRs

## Tasks


lgrammel:
This needs to be introduced end-to-end, i.e. `embed` and `embedMany` need support & tests.

packages/provider/src/language-model/v2/index.ts
```diff
@@ -10,7 +10,6 @@ export * from './language-model-v2-logprobs';
 export * from './language-model-v2-prompt';
 export * from './language-model-v2-provider-defined-tool';
 export * from './language-model-v2-provider-metadata';
-export * from './language-model-v2-provider-options';
```
samdenty, line 13: maybe i should add an alias we can later remove?

lgrammel: no, lets keep different spec (embedding, languagemodel, etc) fully separate for now, also across versions. it's only types so no runtime overhead, and it lets us change one spec without impacting other specs

packages/provider/src/language-model/v2/language-model-v2-call-options.ts
```diff
@@ -130,5 +130,5 @@ Only applicable for HTTP-based providers.
    * to the provider from the AI SDK and enable provider-specific
    * functionality that can be fully encapsulated in the provider.
    */
-  providerOptions?: LanguageModelV2ProviderOptions;
+  providerOptions?: ProviderV2Options;
```
lgrammel, line 133: Please keep specs fully separate for now.

packages/provider/package.json
```diff
@@ -34,7 +34,11 @@
     "@types/node": "20.17.24",
     "@vercel/ai-tsconfig": "workspace:*",
     "tsup": "^8",
-    "typescript": "5.8.3"
+    "typescript": "5.8.3",
+    "zod": "3.23.8"
+  },
+  "peerDependencies": {
+    "zod": "^3.23.8"
```
lgrammel, line 41: Please remove Zod. The provider pkg should not have this dependency.

lgrammel (changes requested):
Please remove Zod and revert shared types.

packages/provider/src/language-model/v2/language-model-v2-call-options.ts
lgrammel: I've extracted `SharedV2ProviderOptions` and `SharedV2ProviderMetadata` that we can use in all model specs: https://github.com/vercel/ai/pull/5733

samdenty: awesome, thanks - yeah think it makes sense

packages/provider/package.json
samdenty: will do 🙏

samdenty: done 👍

packages/ai/core/embed/embed-many.test.ts
```diff
@@ -123,6 +123,28 @@ describe('options.headers', () => {
...
+          return { embeddings: [[1, 2, 3]] };
+        },
+      }),
+      values: ['test-input'],
+      providerOptions: {
+        aProvider: { someKey: 'someValue' },
+      },
+    });
+
+    expect(result.embeddings).toStrictEqual([[1, 2, 3]]);
```
lgrammel, line 144: how good are the error messages that we get when the provider options are not passed through? also the final expect is misleading.

I wonder if a `spy` pattern would be a better alternative for these kinds of tests, where we use a spy for `doEmbed` and then expect that the value sent into the spy is correct.

lgrammel (approved):

PR #5737, opened 2025-04-13
fix(utils/detect-mimetype): add support for detecting id3 tags

haydenbleasel:
## Background

While testing the new Fal transcription functionality, I noticed that we weren't able to correctly detect the mimetype of an mp3 file. Some console logging showed that the headers didn't match. So after a bit of a deep dive, I learned that mp3 files can sometimes contain id3 tags, essentially metadata.

This PR strips that metadata if it exists, allowing the signature checks to function as intended.

## Summary

This pull request includes significant changes to the `detect-mimetype.ts` file to enhance the detection of MIME types by handling ID3 tags in audio files. The most important changes include the addition of functions to strip ID3 tags from binary and base64 data and modifications to the `detectMimeType` function to utilize these new functions.

Enhancements to MIME type detection:

* Added `getID3v2TagSize` function to calculate the size of the ID3v2 tag in an audio file.
* Added `stripID3` function to remove ID3 tags from a `Uint8Array`.
* Added `isBase64ID3v2` function to check if a base64 string starts with an ID3v2 header.
* Added `base64ToBytes` function to convert base64 encoded data to a `Uint8Array`.
* Modified `detectMimeType` function to use `stripID3TagsIfPresent` to process data before checking MIME type signatures.

## Tasks

## Future Work

None.


lgrammel:
Please add a test suite that covers the id3 behavior.

packages/ai/core/util/detect-mimetype.ts
```diff
@@ -83,19 +85,48 @@ export const audioMimeTypeSignatures = [
   },
 ] as const;
 
+const stripID3 = (data: Uint8Array | string) => {
+  const bytes =
+    typeof data === 'string' ? convertBase64ToUint8Array(data) : data;
```
lgrammel, line 90: This leads to unnecessary conversion of potentially large data entries. Can the edit in the base64 case happen w/o converting to UInt8Array?

haydenbleasel: I did spend quite a while on this today, I couldn't get raw base64 stripping working. You can detect the existence of id3 tags by the prefix `SURz` but I think the remaining data between the prefix and the start of the raw mp3 frames are variable.

I'd love to know whether it's possible but yeah I couldn't get it working.

lgrammel:
Please add a changeset (e.g. `fix (core): support mimetypes with id3 tags`) bc it changes the behavior.

lgrammel (approved):

lgrammel (approved):

Issue #5765, opened 2025-04-14
[v5] remove deprecated APIs


gr2m:
### Description

- [x] `CoreTool*` in `ai` package

  PR: #5768

  [packages/ai/core/generate-text/tool-call.ts](https://github.com/vercel/ai/blob/f10304b8259321079ab3dbb38eb2d570f6df5735/packages/ai/core/generate-text/tool-call.ts#L17-L21):

  ```
    19   */
    20  // TODO remove in v5
    21  export type CoreToolCallUnion<TOOLS extends ToolSet> = ToolCallUnion<ToolSet>;
  ```

  [packages/ai/core/generate-text/tool-result.ts](https://github.com/vercel/ai/blob/f10304b8259321079ab3dbb38eb2d570f6df5735/packages/ai/core/generate-text/tool-result.ts#L29-L33):

  ```
    31   */
    32  // TODO remove in v5
    33  export type CoreToolResultUnion<TOOLS extends ToolSet> = ToolResultUnion<TOOLS>;
  ```

  [packages/ai/core/tool/tool.ts](https://github.com/vercel/ai/blob/f10304b8259321079ab3dbb38eb2d570f6df5735/packages/ai/core/tool/tool.ts#L106-L113):

  ```
    108   */
    109  // TODO remove in v5
    110  export type CoreTool<
  ```

  [packages/ai/core/types/language-model.ts](https://github.com/vercel/ai/blob/5d1e3ba3625ad30ba286fb6c1c1292630fc1b23b/packages/ai/core/types/language-model.ts#L59-L64):

  ```
    61   */
    62  // TODO remove in v5
    63  export type CoreToolChoice<TOOLS extends Record<string, unknown>> =
  ```

- [x] Special handling of `temperature` in `ai`

  PR: https://github.com/vercel/ai/pull/5769

  [packages/ai/core/middleware/default-settings-middleware.ts](https://github.com/vercel/ai/blob/c4a2fec279aa6e316dfed650af4295a82c897d9e/packages/ai/core/middleware/default-settings-middleware.ts#L31-L36):

  TODO: verify if `temperature` now defaults to `undefined`

  ```
    31          // special case for temperature 0
    32          // TODO remove when temperature defaults to undefined
    33          temperature:
  ```

- [x] `experimental_wrapLanguageModel` in `ai`

  PR: https://github.com/vercel/ai/pull/5771

[...]

PR #5769, opened 2025-04-14
refactor (ai): remove obsolete handling of `temperature` when it is undefined

gr2m:
## Background

towards #5765 

## Summary

`temperature` now defaults to `undefined`, so the special handling is no longer necessary - at least that's my assumption given that tests pass. Please verify before merging.

## Tasks

- [ ] n/a ~~Docs have been added / updated (for bug fixes / features)~~


packages/ai/core/middleware/default-settings-middleware.ts
```diff
@@ -28,12 +28,7 @@ export function defaultSettingsMiddleware({
           params.providerOptions,
         ),
 
-        // special case for temperature 0
-        // TODO remove when temperature defaults to undefined
-        temperature:
-          params.temperature === 0 || params.temperature == null
-            ? (settings.temperature ?? 0)
-            : params.temperature,
+        temperature: params.temperature || settings.temperature,
```
lgrammel, line 31: prefer `??`

lgrammel:
Please add a changeset.

lgrammel:
Is that all that is needed to remove the default temperature?

packages/ai/core/middleware/default-settings-middleware.ts
gr2m: Right, makes sense 👍🏼 

```suggestion
        temperature: params.temperature ?? settings.temperature,
```

gr2m:
> Please add a changeset.

done: [b0ace14](https://github.com/vercel/ai/pull/5769/commits/b0ace14c9d104efcdf98a1e112943a3d2a0be6c1)

> Is that all that is needed to remove the default temperature?

I think so? I added a test for the settings middleware to verify the new behavior and to prevent future regressions: [756e597](https://github.com/vercel/ai/pull/5769/commits/756e597f90c4953ed237d02497340fcfd1591321)

packages/ai/core/middleware/default-settings-middleware.test.ts
```diff
@@ -41,6 +41,20 @@ describe('defaultSettingsMiddleware', () => {
       expect(result.temperature).toBe(0.5);
     });
 
+    // https://github.com/vercel/ai/pull/5769
```
lgrammel, line 44: pls rm (to keep overhead low) - the tests should be the source of truth / specification

lgrammel:
The main setting that defaults temp to `0` is in `packages/ai/core/prompt/prepare-call-settings.ts` line 101

packages/ai/core/middleware/default-settings-middleware.test.ts
gr2m:
```suggestion
```

packages/ai/core/middleware/default-settings-middleware.test.ts
```diff
@@ -41,6 +41,19 @@ describe('defaultSettingsMiddleware', () => {
...
+      });
+
+      const result = await middleware.transformParams!({
+        type: 'generate',
+        params: BASE_PARAMS,
+      });
+
+      expect(result.temperature).toBeUndefined();
+    });
+
```
gr2m, line 56: I think we can remove this test, [I added another one to `packages/ai/core/prompt/prepare-call-settings.test.ts`](https://github.com/vercel/ai/pull/5769/files#diff-ff1609a57554946701169e99ab02cda0ce1efd5325063d35ecbb8672cc59c66d)

```suggestion
```

gr2m:
> The main setting that defaults temp to `0` is in `packages/ai/core/prompt/prepare-call-settings.ts` line 101

🤦🏼 sorry I missed that. Addressed via https://github.com/vercel/ai/pull/5769/commits/39635866e5e9f465c43ba448c6eb3f2e787f228d. 

Should be ready now. Once merged we can close https://github.com/vercel/ai/issues/5765

lgrammel:
Before merging, I want to think about it more. The alternative would have been to default to `0` and allow users setting it to `null` to reset, which may or may not be the better DX.

lgrammel:
Preferring https://github.com/vercel/ai/pull/5890 because `temperature=0` is a great default for many models.

PR #5774, opened 2025-04-15
fix (provider): remove obsolete `.rawCall` from `doStream()` and `doGenerate()` results (now there is request)

gr2m:
## Background

Towards #5765

## Summary

Obsolete `.rawCall` from `doStream()` and `doGenerate()` results has been removed. Use `.request` instead.

## Tasks

- [ ] n/a ~~Tests for the changes have been added (for bug fixes / features)~~
- [ ] n/a ~~Docs have been added / updated (for bug fixes / features)~~

### DIFF
--- a/packages/provider/src/language-model/v1/language-model-v1.ts
+++ b/packages/provider/src/language-model/v1/language-model-v1.ts
@@ -146,24 +146,6 @@ Finish reason.
     };
 
     /**
-Raw prompt and setting information for observability provider integration.
-     */
-    // TODO v2: remove in v2 (now there is request)
-    rawCall: {
-      /**
-Raw prompt after expansion and conversion to the format that the
-provider uses to send the information to their API.
-       */
-      rawPrompt: unknown;
-
-      /**
-Raw settings that are used for the API call. Includes provider-specific
-settings.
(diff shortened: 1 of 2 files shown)


packages/provider/src/language-model/v1/language-model-v1.ts
```diff
@@ -145,24 +145,6 @@ Finish reason.
       completionTokens: number;
     };
 
-    /**
-Raw prompt and setting information for observability provider integration.
-     */
-    // TODO v2: remove in v2 (now there is request)
```
gr2m, line 151: I assume that `"request"` is referring to the `.request` key:
https://github.com/vercel/ai/blob/56faf0dc29a00c936b101a8e30bc05e63fb4dab9/packages/provider/src/language-model/v1/language-model-v1.ts#L164-L173

But that key is marked as optional, while `.rawCall` seems to have always been present.

Maybe the `.request` key should not be optional?

lgrammel (changes requested):
We do not change `LanguageModelV1` - it is left for reference and will be fully removed before the V5 release. This has been fixed in LangaugeModelV2

PR #5855, opened 2025-04-20
fix (providers/fal): improve model compatibility

shaper:
## Background

Some models on fal.ai do not return image width/height in the API response. The fal image response schema has these noted as required fields, causing otherwise-successful image responses to fail due to missing dimensions.

## Summary

This change removes the width/height fields from the expected response schema. We don't currently use these values, so though we could mark them `nullish` and have them available when possible, we may as well just omit them entirely until we have a use.

Also adds two example scripts for popular models 'recraft-v3' and `luma-photon' that exhibit the responses lacking dimensions.

## Tasks

### DIFF
--- a/.changeset/slimy-chefs-play.md
+++ b/.changeset/slimy-chefs-play.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/fal': patch
+---
+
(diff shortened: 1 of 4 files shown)


samdenty (approved):
LGTM! 👍

Issue #3521, opened 2024-11-07
Custom Tool Parser for Open Source Models


ShervK:
### Feature Description

When using LLM serving frameworks such as [vLLM](https://github.com/vllm-project/vllm)  or [MLC-LLM](https://github.com/mlc-ai/mlc-llm) , or services that host open-source models like [DeepInfra](https://deepinfra.com/), [Fireworks](https://fireworks.ai/), or [OpenRouter](https://openrouter.ai/), you sometimes run into an issue where the model being served doesn't have a dedicated tool parser yet the model does support tool use. This usually means the `tools` parameter in their OpenAI compliant API either doesn't work or causes an error and you'll have to parse the chat completion for any tool calls manually after the request. 

While creating a custom provider can address this, it'll need ongoing maintenance and may lead to missing out on new provider features unless manually implemented.

To address this, I suggest adding a setting for a custom tool parser that can be passed to the OpenAI provider when in `compatible` mode. This feature would allow you to define a function that processes either the response message when using `generateText` or a stream when using `streamText` to determine if the response includes a tool call. This way, you can still keep all the benefits of the tool features of the SDK while serving your own models or using an open source model hosting service. 

Example usage of a basic parser
```typescript
import { createOpenAI } from '@ai-sdk/openai';
import { isParsableJson } from "@ai-sdk/provider-utils";

const llama = createOpenAI({
  // other settings
  compatibility: 'compatible',
  textToolParser: (response: string) => {
      if (!response.startsWith("<|python_tag|>")) return [];
      response = response.replace("<|python_tag|>", "");
      if (!isParsableJson(response)) {
        return [];
      }
      const parsed: Array<{ name: string, arguments: Record<string, unknown>}> = JSON.parse(response) 
      return parsed;
    },
  streamToolParser: (chunk: LanguageModelV1StreamPart) => {
      if (chunk.type !== "text-delta") return;
      if (chunk.textDelta.startsWith("<|python_tag|>") {
      //rest of the implementation

    }
});

```

### Use Case

 - If you're hosting your own models and want to incorporate tools into your project but the serving framework doesn't support tool use for that model or tools are not included in the chat template.
 - When using a model hosting service that either doesn’t support tools for specific models or lacks tool functionality altogether.
 - In projects where you use different models with different tool response formats, making it difficult to parse and handle tool calls in a consistent way.

### Additional context

[...]

PR #5866, opened 2025-04-21
feat (core/middleware): Custom Tool Parser for Open Source Models

minpeter:
## Background

https://github.com/vercel/ai/issues/3521

## Summary

Transform the user's function call schema and input into a prompt, and parse the LLM's response into JSON to implement the function call as a middleware. This unlocks the function call ability for models like ollama's model or openrouter's gemma that do not support function calls.

## Future Work

- Provide basic middleware for more model types
- Could add documentation on how to use the middleware
- Add tests for tool call middleware and relaxed-json

## Change
- Added 2 middleware examples (I used Openrouter for convenience, but it looks like you could replace it with gemma from the google provider)
- Exposes three middlewares for function call functionality. (`createToolMiddleware`, `hermesToolMiddleware`, `gemmaToolMiddleware`)
- Port the implementation of relaxed-json to TypeScript and add it to the utils. (No alternative found as of now)


packages/ai/core/middleware/tool-call-middleware.ts
```diff
@@ -0,0 +1,293 @@
+import * as RJSON from 'relaxed-json';
+
+import { LanguageModelV1Middleware } from './language-model-v1-middleware';
```
lgrammel, line 3: middleware should be v2 and important from provider

minpeter: Thanks, I'm working on the V2 migration, I'll mark the PR as ready when it's ready

minpeter:
@lgrammel I'm ready for review, I've reviewed the license for relaxed-json and it looks fine.

lgrammel:
I've reviewed your PR at a high level and have 2 fundamental concerns:

1) The PR includes prompt modifications, something we try to avoid having in the AI SDK (with very minor exceptions).
2) Longer term, the AI providers will move towards stronger tool call support, rendering this middleware unnecessary or even counter-productive (see e.g. https://cookbook.openai.com/examples/gpt4-1_prompting_guide#tool-calls ).

That said, it is very valuable to have this middleware today, and I realize that the prompt modifications are fundamental to its functionality. My suggestion would therefore be to publish the middleware as a standalone npm package in a repository that you own. We are happy to include a link to your package from the middleware page, please file a PR for the docs change when ready.

minpeter:
> I've reviewed your PR at a high level and have 2 fundamental concerns:
> 
> 1. The PR includes prompt modifications, something we try to avoid having in the AI SDK (with very minor exceptions).
> 2. Longer term, the AI providers will move towards stronger tool call support, rendering this middleware unnecessary or even counter-productive (see e.g. https://cookbook.openai.com/examples/gpt4-1_prompting_guide#tool-calls ).
> 
> That said, it is very valuable to have this middleware today, and I realize that the prompt modifications are fundamental to its functionality. My suggestion would therefore be to publish the middleware as a standalone npm package in a repository that you own. We are happy to include a link to your package from the middleware page, please file a PR for the docs change when ready.

I think the more powerful function calls you mentioned in #2 are the direction that many companies, including OpenAI, are heading in, but there are still many API providers that do not support native function calls. How about writing documentation to strongly recommend using it only for models that "do not support function calls", or adding a bypass in the middleware that does not perform transformation for models that support native function calls?

If you give me feedback, I can reflect it and make small modifications. I already opened the package in @ai-sdk-tool/parser, but it was rejected by another project that used it because it was an external dependency.

lgrammel:
I'd much prefer having this live outside the AI SDK. You could publish a package & github repo under your own account, and we'd be more than happy to link. We can alslo reference the middleware for providers that have less sophisticated tool calling support. Again, I think your middleware is very useful, but we are pretty strict about avoiding specific prompts or prompt modifications in the AI SDK itself.

minpeter:
I understand perfectly, I will move the above v5 implementation out and post a new PR with proper documentation, thanks for reviewing

PR #5868, opened 2025-04-21
chore(providers/openai): update completion model to use providerOptions

samdenty:
## Tasks

- [ ] Tests for the changes have been added (for bug fixes / features)


content/providers/01-ai-sdk-providers/03-azure.mdx
```diff
@@ -377,18 +381,24 @@ OpenAI completion models support also some model specific settings that are not
...
-const model = azure.completion('your-gpt-35-turbo-instruct-deployment', {
-  echo: true, // optional, echo the prompt in addition to the completion
-  logitBias: {
-    // optional likelihood for specific tokens
-    '50256': -100,
+const model = azure.completion('your-gpt-35-turbo-instruct-deployment');
+
+await model.doGenerate({
+  providerOptions: {
+    'azure-openai': {
```
samdenty, line 388: this is the name of the model internally, maybe we want to rename it to just `azure` 🤔

samdenty: not sure if that would be a breaking change @lgrammel

lgrammel: lets check the different azure llm apis first - it might be openai specific. but in general i agree, azure would be better

lgrammel (approved):

content/providers/01-ai-sdk-providers/03-azure.mdx
samdenty: i've just checked and the azure package only has an openai provider. I think unifying the provider id to be just `azure` makes sense here since the package is called `@ai-sdk/azure`, and if in the future we need to add more azure stuff we can just overload it. I don't think it would make sense to have two different `azure-openai` or `azure-anthropic` provider options

packages/openai/src/openai-completion-language-model.ts
```diff
@@ -76,9 +78,24 @@ export class OpenAICompletionLanguageModel implements LanguageModelV2 {
...
+        provider: 'openai',
+        providerOptions,
+        schema: openaiCompletionProviderOptions,
+      }) ?? {},
+      parseProviderOptions({
+        provider: this.providerOptionsName,
+        providerOptions,
+        schema: openaiCompletionProviderOptions,
+      }) ?? {},
+    );
```
lgrammel, line 97: why is `Object.assign` required here?

lgrammel (approved):

packages/openai/src/openai-completion-language-model.ts
samdenty: ill make it use a spread instead

lgrammel (approved):

PR #5873, opened 2025-04-21
fix(providers/xai): return actual usage when streaming instead of NaN

samdenty:
## Background

To resolve #5736

## Summary

Added support for `includeUsage: boolean` in openai-compatible, and use this new option inside the xai provider to return proper usage instead of NaN numbers

## Tasks

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [x] If required, a _patch_ changeset for relevant packages has been added
- [x] You've run `pnpm prettier-fix` to fix any formatting issues

PR #5891, opened 2025-04-22
fix(providers/xai): return actual usage when streaming instead of NaN (#5873)

samdenty:
## Tasks

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If required, a _patch_ changeset for relevant packages has been added
- [ ] You've run `pnpm prettier-fix` to fix any formatting issues


lgrammel (approved):

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -612,7 +619,7 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV2 {
           },
         }),
       ),
-      request: { body },
+      request: { body: JSON.stringify(body) },
```
lgrammel, line 622: needs to be changed, regression

lgrammel: actually not a regression but a bug

samdenty: good catch, thanks lars

Issue #5933, opened 2025-04-23
Feature: Support OpenAI gpt-image-1


dillionverma:
### Description

OpenAI just announced their new image API. 

Would love ai sdk support :)

https://x.com/OpenAIDevs/status/1915097067023900883
https://platform.openai.com/docs/guides/image-generation
https://platform.openai.com/docs/api-reference/images/create?lang=node.js

<img width="593" alt="Image" src="https://github.com/user-attachments/assets/c66f65f1-beb6-47a1-9612-01f08404aa2f" />

dillionverma:
The current version gives an error for `response_format`

```ts

    const result = await generateImage({
      model: openai.image('gpt-image-1'),
      prompt,
      n,
      size,
      providerOptions: {
        openai: {
          // @ts-ignore
          response_format: undefined,
          ...(transparent && { background: 'transparent' }),
          ...(quality !== 'auto' && { quality })
        }
      }
    })
```

```text
Error in imageGeneratorTool: APICallError [AI_APICallError]: Unknown parameter: 'response_format'.
    at <anonymous> (/Users/dillionverma/src/refract/node_modules/.pnpm/@ai-sdk+provider-utils@2.2.7_zod@3.24.3/node_modules/@ai-sdk/provider-utils/src/response-handler.ts:59:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async postToApi (/Users/dillionverma/src/refract/node_modules/.pnpm/@ai-sdk+provider-utils@2.2.7_zod@3.24.3/node_modules/@ai-sdk/provider-utils/src/post-to-api.ts:111:28)
    at async OpenAIImageModel.doGenerate (/Users/dillionverma/src/refract/node_modules/.pnpm/@ai-sdk+openai@1.3.16_zod@3.24.3/node_modules/@ai-sdk/openai/src/openai-image-model.ts:69:50)
    at async _retryWithExponentialBackoff (/Users/dillionverma/src/refract/node_modules/.pnpm/ai@4.3.9_react@19.1.0_zod@3.24.3/node_modules/ai/util/retry-with-exponential-backoff.ts:36:12)
    at async Promise.all (index 0)
    at async generateImage (/Users/dillionverma/src/refract/node_modules/.pnpm/ai@4.3.9_react@19.1.0_zod@3.24.3/node_modules/ai/core/generate-image/generate-image.ts:125:19)
[...]

PR #5938, opened 2025-04-23
feat (providers/openai): support gpt-image-1 image generation

shaper:
## Background

OpenAI launched the new `gpt-image-1` model. The [API](https://platform.openai.com/docs/api-reference/images/create) has some revisions as part of this. If the `response_format` parameter is specified with `gpt-image-1` model the image generation requests fail. This means the current openai image generation support fails when devs try to use the new model.

## Summary

This change conditionally includes the `response_format` parameter so as to omit it if the new model id is specified. The new model only returns `base64_json` format and so there is no need to specify it.

Part of #5933 

## Tasks

## Future Work

- formally specified `providerOptions` schema
- review and potential incorporation of add'l options for the new model


samdenty:
@shaper looks like tests are failing & branch is out of date

content/docs/03-ai-sdk-core/35-image-generation.mdx
```diff
@@ -243,6 +243,7 @@ for (const file of result.files) {
...
 | ------------------------------------------------------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | [xAI Grok](/providers/ai-sdk-providers/xai#image-models)                  | `grok-2-image`                                               | 1024x768 (default)                                                                                                                                                  |
+| [OpenAI](/providers/ai-sdk-providers/openai#image-models)                 | `gpt-image-1`                                                | 1024x1024, 1536x1024, 1024x1536                                                                                                                                     |
```
samdenty, line 246: > 1024x1024, 1536x1024, 1024x1536

do you have a link to this data?

content/providers/01-ai-sdk-providers/02-openai.mdx
```diff
@@ -899,10 +899,23 @@ const model = openai.image('dall-e-3');
...
+| ------------- | ------------------------------- |
+| `gpt-image-1` | 1024x1024, 1536x1024, 1024x1536 |
+| `dall-e-3`    | 1024x1024, 1792x1024, 1024x1792 |
+| `dall-e-2`    | 256x256, 512x512, 1024x1024     |
+
+You can pass optional `providerOptions` to the image model. These are prone to change by OpenAI and aremodel dependent. For example, the `gpt-image-1` model supports the `quality` option:
+
+```ts
+const model = openai.image('gpt-image-1', {
+  providerOptions: {
```
samdenty, line 912: is this how you pass them? i thought you had to call a method cause this is settings? (either way I still want the ability to do this on all models, think it's handy)

samdenty:
other than that looks good :shipit:

samdenty (approved):

content/providers/01-ai-sdk-providers/02-openai.mdx
shaper: This works and is the way AFAIK, though here we don't have a schema defined with the known set of options, which we are moving to, I think we just haven't done the image models yet. Even then, not all options are supported for all models, so it is a world with some imperfections in any case.

content/docs/03-ai-sdk-core/35-image-generation.mdx
shaper: https://platform.openai.com/docs/api-reference/images/create#images-create-size

shaper:
> @shaper looks like tests are failing & branch is out of date

Thanks, I missed that Azure may pass (apparently) dall-e deployment model ids, so I inverted the image settings logic to instead tag specific model ids as ones that don't need response-format specified. Maybe we evolve this later, not sure if Azure has these deployment ids for `gpt-image-1` as well, and the current logic wouldn't do the right thing in that case.

packages/openai/src/openai-image-model.ts
```diff
@@ -78,7 +79,9 @@ export class OpenAIImageModel implements ImageModelV1 {
         n,
         size,
         ...(providerOptions.openai ?? {}),
-        response_format: 'b64_json',
+        ...(!hasDefaultResponseFormat.has(this.modelId)
+          ? { response_format: 'b64_json' }
+          : {}),
```
lgrammel, line 84: why can't we always set the response format?

shaper: If it is set for the `gpt-image-1` model their API reports an error due to unsupported argument. The docs note this behavior as well: https://platform.openai.com/docs/api-reference/images/create#images-create-response_format

> This parameter isn't supported for gpt-image-1 which will always return base64-encoded images.

Though they don't explicitly state what "isn't supported" means, the observed behavior is as above.

PR #5954, opened 2025-04-24
fix(packages): export node10 compatible types

samdenty:
(before the top row was unchecked)

https://arethetypeswrong.github.io/

CleanShot 2025-04-24 at 09 41 48@2x


packages/ai/core/generate-text/smooth-stream.test.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { convertArrayToReadableStream } from '../../test';
+import { convertArrayToReadableStream } from '../../test/index';
```
lgrammel, line 2: this is strange - can we directly import (from provider utils or the file)?

lgrammel (commented):
How have you tested that this fixes the observed problem?

lgrammel (approved):
We can land & release so @shaper can try it out.

packages/ai/core/generate-text/smooth-stream.test.ts
samdenty: this is so that the imports `test.d.ts` and `test/index.ts` don't clash, otherwise it'll pickup the wrong file

samdenty:
> How have you tested that this fixes the observed problem?

I've tested it with arethetypeswrong (before it highlighted error that shaper run into with node 10), i can try to do a local install just to make sure the packages are correct but there's quite a few of them. The key thing to check is that the `export * ` are from the right packages, and that they'd included in the bundle. If both of those are true, they should work with node 10

samdenty:
but yeah good to double check, I can try installing with npm pack locally or another option is for @shaper to test them

packages/ai/core/generate-text/smooth-stream.test.ts
lgrammel: can't you replace it with:

```
import { convertArrayToReadableStream } from '@ai-sdk/provider-utils/test';
```

lgrammel: ideally for imports we avoid barrel files as much as possible and go for the original source

samdenty: will do 🙏

lgrammel (approved):

PR #5977, opened 2025-04-25
feat (provider): add `providerMetadata` to `ImageModelV2` interface

gr2m:
## Background

Compare #5698

## Summary

Additional provider-specific options to the image model provider interface. They are passed through to the provider from the AI SDK and enable provider-specific functionality that can be fully encapsulated in the provider.

Unlike other models, ImageModel request return an array of images, and provider can return image-specific metadata for each. So far, this pull request passing through the revised prompt used for each image. In order to make that possible, I introduced a new type `ImageModelV2ProviderMetadata` which is the same as `SharedV2ProviderMetadata` plus it guarantees the presence of the `.images` key

```js
export type ImageModelV2ProviderMetadata = Record<
  string,
  {
    images: JSONArray;
  } & JSONValue
>;
```

That also makes it possible to deeply merge providerMetadata from multiple responses effectively.

## Verification

I updated the `examples/ai-core/src/generate-image/openai.ts` example to verify that the code is working.

## Tasks

## Future Work

1. STT and TTS models have similar behavior to Image Models. As we add provider metadata we will likely find patterns that will help us to improve the API for `.providerMetadata` for these types of models.
2. Implement `providerMetadata` in image models by other providers - **Follow up issue:** #6405


packages/ai/core/generate-image/generate-image.ts
```diff
@@ -159,29 +161,49 @@ Only applicable for HTTP-based providers.
...
+    if (result.providerMetadata) {
+      for (const [providerName, metadata] of Object.entries<{
+        images: unknown;
+      }>(result.providerMetadata)) {
+        providerMetadata[providerName] ??= { images: [] };
+        providerMetadata[providerName].images.push(
+          ...result.providerMetadata[providerName].images,
+        );
+      }
+    }
```
lgrammel, line 174: what about provider metadata that's not `images`?

gr2m: It would be merged into the `result.providerMetadata[providerName]` object, but I haven't seen any meta data yet. I can prepare the code for it already though if you like?

packages/provider/src/image-model/v2/image-model-v2.ts
```diff
@@ -1,6 +1,14 @@
+import { JSONArray, JSONValue } from '../../json-value';
 import { ImageModelV2CallOptions } from './image-model-v2-call-options';
 import { ImageModelV2CallWarning } from './image-model-v2-call-warning';
 
+export type ImageModelV2ProviderMetadata = Record<
+  string,
+  {
+    images: JSONArray;
```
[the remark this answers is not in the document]
gr2m: I think the dx would be better if it's always present, set to an array with the same length as the `result.images` array. But each item of the `JSONArray` can be `null`

lgrammel (approved):

PR #5995, opened 2025-04-25
feat(ai-core): exported types for the option in the generateObject function

### LINKED ISSUE #5615: export `generateObject` parameter `options` type
### Feature Description

The options type for the `generateObject` function is pretty complex, and there are multiple overloads in the packaged type.

Because of this, getting access to this type to nicely wrap this function is pretty difficult. Using TS utils like `ReturnType` doesn't seem to work because of the overloads. Getting all sorts of weird behaviors trying to hack this in myself.

Next best bet was to just copy/paste the types into my project, but a) this will break if the api shape changes (which is fine I suppose, as this would happen already, but this is still hacky), and b) this is a lot of type code to paste in.

Is there any way for this to be exposed externally?

### Use Cases

 This would enable me to do something like this, which is really handy at least for my use case
```ts
export async function safeGenerateObject(
  opts: GenerateObjectOptions // imported from the package
  errorHandler?: (error: unknown) => AISDKError
) {
  return ResultAsync.fromThrowable(
    () => generateObject(opts),
    errorHandler ?? handleAISdkError
  )();
}
```

### Additional context

_No response_

AVtheking:
## Summary

Exports type for the option parameter in the generateObject function

## Verification
Run the test cases and verified the working of the function

## Tasks

## Related Issues
Fixes #5615


lgrammel:
This area has significantly changed in `v5`. Please check the `v5` branch and see if the changes are still needed.

AVtheking:
I see. Though user has asked for explicit type for the options parameter be exported.

AVtheking:
One thing more @lgrammel do we need to create PR setting v5 as base ? or we can do it with main branch also ? Thanks

lgrammel:
`v5` as base, and make sure to branch off of `v5`

PR #6027, opened 2025-04-28
chore(ai): cleanup provider-utils dependency

samdenty:
## Summary

After more cross-referencing this was the only one


gr2m (approved):

lgrammel (changes requested):
I had left this particular dependency in `provider-utils` to facilitate testing of url support in providers. We can close this PR.

PR #6028, opened 2025-04-28
feat(streamObject): add enum support

samdenty:
## Background

We were missing enum on the streamObject method

## Summary

Added enum support to streamObject

## Tasks

### DIFF
--- a/packages/ai/core/generate-object/output-strategy.ts
+++ b/packages/ai/core/generate-object/output-strategy.ts
@@ -335,11 +335,50 @@ const enumOutputStrategy = <ENUM extends string>(
           };
     },
 
-    validatePartialResult() {
-      // no streaming in enum mode
-      throw new UnsupportedFunctionalityError({
-        functionality: 'partial results in enum mode',
-      });
+    async validatePartialResult({ value, textDelta }) {
+      if (!isJSONObject(value) || typeof value.result !== 'string') {
+        return {
+          success: false,
+          error: new TypeValidationError({
+            value,
+            cause:
+              'value must be an object that contains a string in the "result" property.',
+          }),
+        };
+      }
+
+      const result = value.result as string;
+      const possibleEnumValues = enumValues.filter(enumValue =>
+        enumValue.startsWith(result),
+      );
+
+      if (value.result.length === 0 || possibleEnumValues.length === 0) {
+        return {
+          success: false,
+          error: new TypeValidationError({
+            value,
+            cause: 'value must be a string in the enum',
+          }),
+        };
+      }
+
+      if (possibleEnumValues.length > 1) {
+        return {
+          success: false,
+          error: new TypeValidationError({
+            value,
+            cause: `ambiguous enum value: ${result}`,
+          }),
+        };
+      }
+
+      return {
+        success: true,
+        value: {
+          partial: possibleEnumValues[0],
+          textDelta,
+        },
+      };
     },
 
     createElementStream() {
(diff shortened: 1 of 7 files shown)


gr2m (approved):

lgrammel (changes requested):
Please add a test for the enum functionality (not just the types).

lgrammel:
Aps examples would be good and docs changes might be needed.

packages/ai/core/generate-object/output-strategy.ts
```diff
@@ -290,7 +290,7 @@ const arrayOutputStrategy = <ELEMENT>(
 
 const enumOutputStrategy = <ENUM extends string>(
   enumValues: Array<ENUM>,
-): OutputStrategy<ENUM, ENUM, never> => {
+): OutputStrategy<ENUM | (string & {}), ENUM, never> => {
```
lgrammel, line 293: why is this needed?

packages/ai/core/generate-object/output-strategy.ts
```diff
@@ -335,11 +335,41 @@ const enumOutputStrategy = <ENUM extends string>(
...
+      }
+
+      return {
+        success: true,
+        value: {
+          partial:
+            possibleEnumValues.length > 1 ? result : possibleEnumValues[0],
+          textDelta,
+        },
+      };
```
lgrammel, line 372: this can lead to issues where the wrong enum value is selected as a partial result, leading to changing enum values. i'd prefer a solution where that cannot happen (either by only returning fully matched enum values, or (more complicated) by returning enum values for which only one value is possible per prefix)

packages/ai/core/generate-object/stream-object.test-d.ts
```diff
@@ -1,11 +1,26 @@
...
+      enum: ['a', 'b', 'c'] as const,
+      model: undefined!,
+    });
+
+    expectTypeOf<typeof result.object>().toEqualTypeOf<
+      Promise<'a' | 'b' | 'c'>
+    >;
+
+    for await (const text of result.partialObjectStream) {
+      expectTypeOf(text).toEqualTypeOf<'a' | 'b' | 'c' | (string & {})>();
```
lgrammel, line 20: why the generic `string & {}`?

packages/ai/core/generate-object/output-strategy.ts
samdenty: yep good call out - I added tests to test for these cases. In this case I check for `possibleEnumValues.length > 1` which means there's more than one match, in that case i return just the partial value. If there's only one possible enum per prefix, i return the full value preemptively

packages/ai/core/generate-object/output-strategy.ts
samdenty: enums partial streaming can either return the full enum `foobar` or a prefix like `foo` (-`bar`) so I added a partial type to make autocomplete still work will making sure there's a `string` | type

lgrammel: I don't think this is what the users would expect - then they'd need to handle the string case. I would prefer a solution that is closer to what we do for arrays - i.e., only return ENUM values even as partials.

lgrammel: wonder if we can just return `string` then

samdenty: done :pray:

lgrammel (approved):

PR #6051, opened 2025-04-29
feat(dx): use raw source files in development

samdenty:
## Background

Before we were always using the compiled output, this meant you had to run a build to get the project to reflect updated source files.

We were also running the `build` steps serially, as they all depended on each other.

### Rough results 

Non-cached full build -> 4x faster
Non-cached tests run -> 2x faster (doesn't depend on build step)

For example, when running one of the examples

## Summary

Now we load the source files directly in development, only using compiled in the published packages only

## Tasks


samdenty:
> Great stuff! For verification you could create a snapshot release and test out the packages.

@lgrammel i've just checked and it works. Changesets publishes with pnpm if it's a pnpm project, and doing that with exports works

lgrammel:
Thought more about it and I'm wondering if this is a good path take.

We create another test gap. When we run the examples right now in development for manual testing, they are using the dist files of the packages. We don't run manual testing in CI. Therefore we'll not catch bugs anymore that relate to dist files, and our examples setup will be different from what our users have (since they use packages/dist files).

I'd much prefer exploring having a `build:watch` instead.

PR #6055, opened 2025-04-30
fix (react): integrate addToolResult into UseChatHelpers type without…

cgoinglove:
## Background
Unnecessary type intersection used for addToolResult in React's useChat implementation.

## Summary
Integrated addToolResult directly into UseChatHelpers type for cleaner type definition.

## Verification
Type checking works as expected with no functional changes.

## Tasks
- [x] Formatting issues have been fixed


lgrammel:
Can you make this change on the `v5` branch instead? we are trying to limit the work on `main` and this change is not critical.

PR #6083, opened 2025-05-01
fix (ai)!: move image model settings into `generateImage({providerOptions})`

### LINKED ISSUE #6077: [v5] ImageModelV2: move `settings` constructor options into `options.providerOptions` of `.doGenerate()` method
We want to update the constructor of classes implementing `ImageModelV2`, e.g. https://github.com/vercel/ai/blob/9bd5ab57a32142e55654ee805134b609ef33f605/packages/openai/src/openai-image-model.ts#L36-L40

Instead of passing `OpenAIImageSettings` as the 2nd argument in the constructor, we want to pass them to the `doGenerate()` method:

https://github.com/vercel/ai/blob/8919eff9b71e25c19f54dfa3097dc546c3a1c436/packages/openai/src/openai-image-model.ts#L42-L51

### Prior art

- language model
  - current implementation: [packages/openai/src/openai-chat-language-model.ts](https://github.com/vercel/ai/blob/7b3ae3f8fc2a72d29953b65b1f6c27f5366c8fe5/packages/openai/src/openai-chat-language-model.ts#L56-L59)
  - Pull request that made the change: [#5990](https://github.com/vercel/ai/pull/5990/files#diff-056bbb5c30a5e0416f2fadc19ac4d379645510d740475a3902546afadc569744)
- https://github.com/vercel/ai/pull/5989

gr2m:
## Background

Closes #6077

## Summary

Before

```js
await generateImage({
  model: openai.image("dall-e-2", { maxImagesPerCall: 5 }),
  prompt,
  n: 10,
});
```

after

```js
await generateImage({
  model: openai.image("dall-e-2"),
  prompt,
  n: 10,
  providerOptions: {
    openai: { maxImagesPerCall: 5 },
  },
});
```

## Verification

I tested using the altered `examples/ai-core/src/generate-image/openai.ts` code shown above and verified that `maxImagesPerCall` is correctly used to calculate the API requests, without sending `maxImagesPerCall` itself as part of the API request bodies.

## Task


packages/ai/core/generate-image/generate-image.test.ts
```diff
@@ -61,7 +61,13 @@ describe('generateImage', () => {
...
       aspectRatio: '16:9',
       seed: 12345,
-      providerOptions: { openai: { style: 'vivid' } },
+      providerOptions: {
+        'mock-provider': {
+          style: 'vivid',
+          // maxImagesPerCall is not passed to doGenerate
+          maxImagesPerCall: 3,
+        },
+      },
```
gr2m, line 70: I assume the intend was to pass `'mock-provider'` as `providerOptions` key to match the mock provider? Or is there a valid use case to pass `providerOptions` that don't match the passed model?

packages/ai/core/generate-image/generate-image.ts
```diff
@@ -110,20 +110,27 @@ Only applicable for HTTP-based providers.
 }): Promise<GenerateImageResult> {
   const { retry } = prepareRetries({ maxRetries: maxRetriesArg });
 
+  // extract maxImagesPerCall from providerOptions as it's not meant to be passed to `doGenerate()`
+  const [
+    [providerName, { maxImagesPerCall, ...generateProviderOptions } = {}] = [],
+  ] = Object.entries(providerOptions ?? {});
```
gr2m, line 116: I'm surprised that we don't know `providerName` at the time when `generateImage()` is called. Can't we somehow reliable extract it from `model`?

I also don't know if this change breaks a use case that I'm not aware of?

packages/ai/core/generate-image/generate-image.ts
```diff
@@ -110,20 +110,27 @@ Only applicable for HTTP-based providers.
...
+
   // default to 1 if the model has not specified limits on
   // how many images can be generated in a single call
-  const maxImagesPerCall = model.maxImagesPerCall ?? 1;
+  const maxImagesPerCallWithDefault =
+    (maxImagesPerCall as number) ?? model.maxImagesPerCall ?? 1;
```
gr2m, line 121: This is probably the main change. Before we had access to `model.settings` in the `maxImagesPerCall` getter. Now that `settings` has been removed from the `ImageModelV2` constructor, that is no longer the case.

packages/ai/core/generate-image/generate-image.ts
```diff
@@ -135,7 +142,9 @@ Only applicable for HTTP-based providers.
           size,
           aspectRatio,
           seed,
-          providerOptions: providerOptions ?? {},
+          providerOptions: providerName
+            ? { [providerName]: generateProviderOptions }
+            : {},
```
gr2m, line 147: Is there a use case where providerOptions has more than one key? If so this change would break it. We don't pass through provider options because we want to remove the `maxImagesPerCall` option first, otherwise it ends up in the API request body.

packages/amazon-bedrock/src/bedrock-provider.ts
```diff
@@ -85,15 +82,16 @@ export interface AmazonBedrockProvider extends ProviderV2 {
...
 
-  image(
-    modelId: BedrockImageModelId,
-    settings?: BedrockImageSettings,
-  ): ImageModelV2;
+  /**
+Creates a model for image generation.
+@deprecated Use `imageModel` instead.
+   */
+  image(modelId: BedrockImageModelId): ImageModelV2;
```
gr2m, line 89: `image` is deprecated, right? Is saw it in one of the models and applied it everywhere. Shall we remove it as part of v5 (in another PR?)

packages/luma/src/luma-image-model.ts
```diff
@@ -86,6 +76,10 @@ export class LumaImageModel implements ImageModelV2 {
       });
     }
 
+    // remove non-request options from providerOptions
+    const { pollIntervalMillis, maxPollAttempts, ...providerRequestOptions } =
+      providerOptions.luma ?? {};
```
gr2m, line 81: When providers want to support options that should not be passed through to the API calls then they have to manually extract them before sending the request. As an alternative we could use a special key like `providerOptions._.*` for options that should not be sent as part of the request body.

gr2m (commented):
Moving image model settings into provider options has several consequences

- we loose the type intellisense. We could bring it back to `providerOptions` but it would require additional work to infer the `providerOptions` type from the passed model
- Before, all `providerOptions` would be passed 1:1 to the API request body. Now we have to extract and remove non-request settings from `providerOptions` before passing it to `doGenerate()`

I didn't update docs yet, I'll wait until we agree the changes look good

gr2m:
superseded by #6180

PR #6108, opened 2025-05-02
feat(embed-many): respect supportsParallelCalls & concurrency

samdenty:
## Background

We didn't actually read the supportsParallelCalls field at all, and we did everything serially in embedding model

## Summary

This makes the embedding model actually respect supportsParallel calls

## Tasks

### DIFF
--- a/packages/ai/core/embed/embed-many.test.ts
+++ b/packages/ai/core/embed/embed-many.test.ts
@@ -5,6 +5,7 @@ import {
 } from '../test/mock-embedding-model-v2';
 import { MockTracer } from '../test/mock-tracer';
 import { embedMany } from './embed-many';
+import { createResolvablePromise } from '../../util/create-resolvable-promise';
 
 const dummyEmbeddings = [
   [0.1, 0.2, 0.3],
@@ -20,20 +21,25 @@ const testValues = [
 
 describe('model.supportsParallelCalls', () => {
   it('should not parallelize when false', async () => {
-    let currentValueIndex = 0;
+    const events: string[] = [];
+    let callCount = 0;
 
-    const { embeddings } = await embedMany({
+    const resolvables = [
+      createResolvablePromise<void>(),
+      createResolvablePromise<void>(),
+      createResolvablePromise<void>(),
+    ];
+
+    const embedManyPromise = embedMany({
       model: new MockEmbeddingModelV2({
         supportsParallelCalls: false,
         maxEmbeddingsPerCall: 1,
-        doEmbed: async ({ values }) => {
-          const index = currentValueIndex++;
+        doEmbed: async () => {
+          const index = callCount++;
+          events.push(`start-${index}`);
 
-          expect(values).toEqual([testValues[index]]);
-
-          await new Promise(resolve => setTimeout(resolve, 1));
-
-          expect(currentValueIndex).toEqual(index + 1);
+          await resolvables[index].promise;
+          events.push(`end-${index}`);
 
           return {
             embeddings: [dummyEmbeddings[index]],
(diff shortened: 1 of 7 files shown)


packages/provider/src/embedding-model/v2/embedding-model-v2.ts
```diff
@@ -42,7 +42,10 @@ Use Infinity for models that do not have a limit.
   /**
 True if the model can handle multiple embedding calls in parallel.
    */
-  readonly supportsParallelCalls: PromiseLike<boolean> | boolean;
+  readonly supportsParallelCalls:
+    | PromiseLike<boolean | number>
+    | boolean
+    | number;
```
lgrammel, line 48: why do we need a number here?

packages/ai/core/embed/embed-many.ts
```diff
@@ -100,7 +100,17 @@ Only applicable for HTTP-based providers.
...
+        model.maxEmbeddingsPerCall,
+        model.supportsParallelCalls,
+      ]);
+
+      const maxParallelChunks =
+        typeof supportsParallelCalls === 'number'
+          ? supportsParallelCalls
+          : supportsParallelCalls
+            ? Infinity
+            : 1;
```
lgrammel, line 113: Why do we need to cap the max tool calls based on the provider? for which provider would this matter vs boolean?

packages/provider/src/embedding-model/v2/embedding-model-v2.ts
lgrammel: maybe change to `maxParallelToolCalls: PromiseLike<number> | number` but unclear why we need to cap the max tool calls based on the provider? it's usually boolean?

packages/ai/core/embed/embed-many.test.ts
```diff
@@ -18,6 +18,68 @@ const testValues = [
...
+    const { embeddings } = await embedMany({
+      model: new MockEmbeddingModelV2({
+        supportsParallelCalls: false,
+        maxEmbeddingsPerCall: 1,
+        doEmbed: async ({ values }) => {
+          const index = i++;
+
+          expect(values).toEqual([testValues[index]]);
+
+          await new Promise(resolve => setTimeout(resolve, 1));
```
lgrammel, line 34: Use `delay` if you really need a timeout. Ideally timeouts should be avoided in tests.

packages/ai/core/embed/embed-many.test.ts
```diff
@@ -18,6 +18,68 @@ const testValues = [
...
+    const { embeddings } = await embedMany({
+      model: new MockEmbeddingModelV2({
+        supportsParallelCalls: true,
+        maxEmbeddingsPerCall: 1,
+        doEmbed: async ({ values }) => {
+          const index = i++;
+
+          expect(values).toEqual([testValues[index]]);
+
+          await new Promise(resolve => setTimeout(resolve, 1));
```
lgrammel, line 62: same - can we do this without delays that tend to be brittle? e.g. by resolving mock promises? (that's how i usually deal with parallel vs non-parallel without needing the brittleness of timeouts)

packages/ai/core/embed/embed-many.test.ts
```diff
@@ -18,6 +18,68 @@ const testValues = [
...
+        },
+      }),
+      values: testValues,
+    });
+
+    expect(embeddings.length).toEqual(testValues.length);
+  });
+
+  it('should parallelize when true', async () => {
+    let i = 0;
```
lgrammel, line 51: what is he meaning of `i` here? can there be a more descriptive name? (same above)

packages/ai/core/embed/embed-many.test.ts
lgrammel: better idea: track `start` and `end` of mock calls in array, and use promises to resolve in parallel case.

packages/provider/src/embedding-model/v2/embedding-model-v2.ts
samdenty: see https://community.openai.com/t/concurrent-request-restriction/1062443

samdenty: this is to cap the amount of parallel embed-many requests

packages/ai/core/embed/embed-many.test.ts
samdenty: yeah this is just a 1ms delay to ensure it runs async vs sync, I can also do start / end

packages/ai/core/embed/embed-many.test.ts
samdenty: yep, thought that too

packages/ai/core/embed/embed-many.test.ts
samdenty: done 🙏

packages/ai/core/embed/embed-many.ts
```diff
@@ -73,6 +74,13 @@ Only applicable for HTTP-based providers.
   functionality that can be fully encapsulated in the provider.
   */
   providerOptions?: ProviderOptions;
+
+  /**
+   * Maximum number of concurrent requests.
+   *
+   * @default Infinity
+   */
+  concurrency?: number;
```
lgrammel, line 83: rename to `maxParallelCalls` or `maxParallelRequests` or `maxConcurrentRequests`. If we rename to `Concurrent` here it would be good to do the same in the model spec

samdenty: yep that's a clearer name, updated

lgrammel (approved):

packages/amazon-bedrock/src/bedrock-embedding-model.test.ts
```diff
@@ -86,14 +86,6 @@ describe('doEmbed', () => {
     expect(usage?.tokens).toStrictEqual(8);
   });
 
-  it('should handle multiple input values and extract usage', async () => {
-    const { usage } = await model.doEmbed({
-      values: testValues,
-    });
-
-    expect(usage?.tokens).toStrictEqual(16);
-  });
```
lgrammel, line 95: why was  this test removed?

samdenty: removed because we now throw an error 👍 
image

PR #6121, opened 2025-05-02
chore(provider-utils): move ToolResultContent to provider-utils

samdenty:
## Background

After a lot of cross-referencing, the only cleanup we need is to unify this interface into provider-utils

## Summary

Move ToolResultContent to provider-utils

### DIFF
--- a/packages/anthropic/src/anthropic-tools.ts
+++ b/packages/anthropic/src/anthropic-tools.ts
@@ -1,25 +1,12 @@
+import { ToolResultContent } from '@ai-sdk/provider-utils';
 import { z } from 'zod';
 
-// Copied from ai package
 type ExecuteFunction<PARAMETERS, RESULT> =
   | undefined
   | ((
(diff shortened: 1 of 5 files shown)


packages/anthropic/src/anthropic-tools.ts
```diff
@@ -1,26 +1,13 @@
+import { ToolResultContent } from '@ai-sdk/provider-utils';
 import { z } from 'zod';
 
-// Copied from ai package
 type ExecuteFunction<PARAMETERS, RESULT> =
```
lgrammel, line 4: is ExecuteFunction not copied in some fashion?

samdenty: couldn't find a reference, tried searching multiple ways

lgrammel:
- needs a changeset bc things are moved crosspackage

lgrammel (approved):

Issue #5499, opened 2025-04-02
AI_InvalidPromptError: Invalid prompt: messages must be an array of CoreMessage or UIMessage


HomyeeKing:
### Description

After upgrade the ai-sdks to the latest, the exact version are list below, can't get pass the zod validation.

1. setMessages modify the original message object and fill the `part` property https://github.com/vercel/ai/blob/main/packages/react/src/use-chat.ts#L465
2. I thought the new `part` property failed the zod validation, so I use `convertToCoreMessages` this util to remove it, unfortunately still can't pass the validation
3. the `part` of the user message was removed, but the `assistant` message is formatted to mismatch the `CoreAssistantMessage`

### Code example
```ts
import { convertToCoreMessages } from 'ai';

const res = convertToCoreMessages([
  {
    role: 'assistant',
    content: [
      {
        type: 'text',
        text: 'xxxxxxxxx',
      },
    ],
    parts: [
      {
        type: 'text',
        text: [
          {
            type: 'text',
            text: 'xxxxxxxxxxx',
          },
        ],
      },
    ],
  },
]);

console.log('res', res);

```
### AI provider

  "@ai-sdk/provider": "^1.1.0",

### Additional context

```
"@ai-sdk/openai": "^1.3.6",
  "@ai-sdk/openai-compatible": "^0.2.4",
  "@ai-sdk/provider": "^1.1.0",
  "@ai-sdk/provider-utils": "^2.2.3",
  "@ai-sdk/react": "^1.2.5",
  "ai": "^4.2.10"
```

HomyeeKing:
so should fix the type definition of `CoreAssistantMessage `  or the `convertToCoreMessages` logic

iteratetograceness:
@HomyeeKing could you help me understand where you're getting the assistant message you're passing to `convertToCoreMessages`?

in the code snippet you shared there are some type discrepancies:

```
{
    role: 'assistant',
    content: [ // expected type is string
      {
        type: 'text',
        text: 'xxxxxxxxx',
      },
    ],
    parts: [
      {
        type: 'text',
        text: [ // expected type is string
          {
            type: 'text',
            text: 'xxxxxxxxxxx',
          },
        ],
      },
    ],
  }
```

HomyeeKing:
@iteratetograceness 
the assistant messages are from `generateText`
```ts
const { messages, setMessages } = useChat({
 // ....
})
const { response } = await generateText({
  model: openai('gpt-4o'),
  messages: messages
[...]

PR #6133, opened 2025-05-03
feat(message-validator): include more details in error messages

samdenty:
## Background

Before in issues like https://github.com/vercel/ai/issues/5499 & https://github.com/vercel/ai/discussions/5833, if users ran into an issue where the messages array failed to parse correctly they'd run into a difficult to debug error `Invalid prompt: messages must be an array of CoreMessage or UIMessage`

## Summary

This PR extends these error messages to include more details on what exactly went wrong, which is useful information for debugging when stuff breaks.

On security: I think this is an okay change since the data sent is not private

## Verification

E2E tests validating new error messages

## Tasks


packages/ai/core/prompt/standardize-prompt.ts
```diff
@@ -110,7 +103,10 @@ export function standardizePrompt<TOOLS extends ToolSet>({
     if (!validationResult.success) {
       throw new InvalidPromptError({
         prompt,
-        message: 'messages must be an array of CoreMessage or UIMessage',
+        message: [
+          'messages must be an array of CoreMessage or UIMessage',
```
lgrammel, line 107: maybe change msg to "message must be a CoreMessage or a UI message" and consider changing the InvalidPromptError so that the messages can be included and printed instead

lgrammel (commented):
Best practice is to include the raw information (messages, message) in the error object, vs adding to the error message itself.

samdenty:
> Best practice is to include the raw information (messages, message) in the error object, vs adding to the error message itself.

Yep I was unsure on the best way to approach this problem. The issue is what if it throws it inside the stream, so it ends up being serialised as a data stream message - in that case what is the best way to surface the error message to the user. Do the errors get printed to the console terminal output automatically? If not, should we print the messages there? What about if people run into this on production, how best can we allow for this error to be fixed by the application. I imagine they'd have error reporting built into the stack somewhere, it's just a question of how we hook that up correctly (and what methods they use)


I also want to make sure if we do this, that there's no possibility of leaking the system prompt as the user could malform a message.

lgrammel (approved):

PR #6136, opened 2025-05-04
fix(docs): Use the re-ask strategy for repair example code passed invalid tool call args

rychlis:
## Background

We're using `experimental_repairToolCall` with re-ask strategy to repair wrong tool calls and we hit a 400 error (`messages.X.content.0.tool_use.input: Input should be a valid dictionary`) in the generateText inside this repair function.

After some (pretty long) investigation, I found out that there is a bug in docs (which we used as base of our implementation):
- repairToolCall accepts toolCall of type LanguageModelV1FunctionToolCall which has args as a string [source](https://github.com/vercel/ai/blob/845080d80b8538bb9c7e527d2213acb5f33ac9c2/packages/provider/src/language-model/v1/language-model-v1-function-tool-call.ts#L10)
- experimental_repairToolCall is calling generateText and passing 'tool-call' message which expects object in `args` - the toolCall.args was passed as a string
- this causes 400 error response from LLMs (at least Anthropic in my case) 

## Summary

I've added a json parsing before passing args to generateText. If JSON parse is unsuccessful, fixing is skipped


rychlis:
Link to the relevant docs https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling#example-use-the-re-ask-strategy-for-repair

rychlis:
There might be same issue in `examples/ai-core/src/generate-text/mock-tool-call-repair-reask.ts` https://github.com/rychlis/ai/blob/830876d2bbae7a3a650be88ddd49381e26567033/examples/ai-core/src/generate-text/mock-tool-call-repair-reask.ts#L51 - can try to investigate in separate PR. However it might not matter for mocks

rychlis:
Also this would not happen if ToolCallPart was properly typed - now args is just `unknown` and I thinki it could be something like `Record<string, unknown>​​` - https://github.com/vercel/ai/blob/845080d80b8538bb9c7e527d2213acb5f33ac9c2/packages/ai/core/prompt/content-part.ts#L232

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -521,6 +521,13 @@ const result = await generateText({
     messages,
     system,
   }) => {
+    // Parse tool args without schema
+    const parsedArgsResult = safeParseJSON({
```
lgrammel, line 525: needs to be added to imports

lgrammel:
might need prettier fix

rychlis:
> might need prettier fix

Yeah, will fix later today

content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
rychlis: oops, fixed now

rychlis:
@lgrammel I fixed the prettier issue and missing import. Should I rebase with squash or leave 3 commits?

lgrammel (approved):

gr2m:

PR #6182, opened 2025-05-07
feat(provider): return image request body for openai (+ compatible) providers

haydenbleasel:
This is a first step towards returning usable body information for image generations.


lgrammel (changes requested):
Please change on `v5` branch instead.

PR #6184, opened 2025-05-07
chore (provider-utils): replace zod with standard-schema where possible

### LINKED ISSUE #5766: [v5] enable `standard-schema` usage where possible
## Background

We want to support [standard-schema](https://github.com/standard-schema/standard-schema) instead of zod schemas in custom validators.

We will keep the `zod` dependency for internal validations (e.g. for user inputs and parts of provider API responses, e.g. [packages/ai/core/generate-object/generate-object.ts](https://github.com/vercel/ai/blob/7ea4132624d8ab83f1f9713020e4c76bd6c2036b/packages/ai/core/generate-object/generate-object.ts#L61-L70)). The motivation to use `standard-schema` instead of `zod` in the validate type methods is to enable model providers to use their own preferred way of validation that is compatible with `standard-schema`. The same is true for users of our UI adapters like React, users should be able to utilize their schema validation library of choice instead of forcing them to use `zod`.

Users bring validators in two ways

1. Instruct providers to generate structural output following a provided schema (needs conversion to JSON Schema, see also [valibot/src/valibot-schema.ts](https://github.com/vercel/ai/blob/59e0fcdde18de92ffae75e1fdae5a0dd7ea9d0bd/packages/valibot/src/valibot-schema.ts))
2. Validation library for the frontend. Will become more important in the future for typed tool calls, typed custom data in React components

## Summary

This issue is about updating the type of [the `schema` argument of the `validateTypes()` method](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L16) used by providers. We want to replace the use of `z.Schema` [here](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L19) and [here](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L44) with standard schema. We can keep the custom validators.

The [`validator.validate()` call](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L55) is currently synchronous, but introducing `standard-schema` makes it possible to set schema requirements with async validations, which means that `validator.validate()` will have to be async, and in turn `safeValidateTypes()` and `validateTypes()` will have to be made async.

It might make sense to decouple the introduction of `standard-schema` to the aync API change, by starting a pull request that simply makes `safeValidateTypes()` async without any other changes.

## Tasks

- [x] #5779
- [ ] replace the use of `z.Schema` [here](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L19) and [here](https://github.com/vercel/ai/blob/4bd27a9b8e8f264bef3a9442906488a725609b16/packages/provider-utils/src/validate-types.ts#L44) with standard schema. Potentially other places like [parseJSON](https://github.com/vercel/ai/blob/7ea4132624d8ab83f1f9713020e4c76bd6c2036b/packages/provider-utils/src/parse-json.ts#L31) and [safeParseJSON](https://github.com/vercel/ai/blob/7ea4132624d8ab83f1f9713020e4c76bd6c2036b/packages/provider-utils/src/parse-json.ts#L84) and in the UI (e.g. [packages/react/src/use-object.ts](https://github.com/vercel/ai/blob/59e0fcdde18de92ffae75e1fdae5a0dd7ea9d0bd/packages/react/src/use-object.ts#L29))
- [ ] Add `standard-schema` validator next to [`zodValidator`](https://github.com/vercel/ai/blob/7ea4132624d8ab83f1f9713020e4c76bd6c2036b/packages/provider-utils/src/validator.ts#L52-L61) which would need to be async, therefore [the `Validator.validate()` type](https://github.com/vercel/ai/blob/7ea4132624d8ab83f1f9713020e4c76bd6c2036b/packages/provider-utils/src/validator.ts#L18-L22) would need to be async 
- [ ] potential follow up: look into replacing internal use of `zod` with a minimal version of `zod`

### Resources

- Compare https://github.com/vercel/ai/pull/5479

gr2m:
## Background
[Standard schema](https://github.com/standard-schema/standard-schema) allows validation with any schema library that implements it. While we still need to support specific schema libraries in many places because of the JSONSchema conversion, switching to standard schema support where possible will increase the flexibility of the AI SDK.

## Summary

Replace `zod` with `standard-schema` where users can pass their own schemas for validation.

## Related Issues

Closes #5766


packages/ai/core/generate-object/output-strategy.ts
```diff
@@ -129,9 +134,13 @@ const objectOutputStrategy = <OBJECT>(
   },
 });
 
-const arrayOutputStrategy = <ELEMENT>(
-  schema: Schema<ELEMENT>,
-): OutputStrategy<ELEMENT[], ELEMENT[], AsyncIterableStream<ELEMENT>> => {
+const arrayOutputStrategy = <OBJECT>(
+  schema: Schema<OBJECT>,
```
lgrammel, line 138: `ELEMENT` was on purpose because it is an array element

packages/provider-utils/src/validator.ts
```diff
@@ -19,7 +20,7 @@ export type Validator<OBJECT = unknown> = {
    * Optional. Validates that the structure of a value matches this schema,
    * and returns a typed version of the value if it does.
    */
-  readonly validate?: (value: unknown) => ValidationResult<OBJECT>;
+  readonly validate?: (value: unknown) => ValidationResult<OBJECT> | PromiseLike<ValidationResult<OBJECT>>;
```
gr2m: no sure if making the `validate()` method optionally async could cause any problems down the road, but it made the transition easier.

packages/ai/src/ui/process-ui-message-stream.ts
```diff
@@ -105,6 +105,7 @@ export function processUIMessageStream<MESSAGE_METADATA = unknown>({
               if (messageMetadataSchema != null) {
                 await validateTypes({
                   value: mergedMetadata,
+                  // @ts-ignore TODO: Couldn't figure this one out yet ~@gr2m
```
gr2m, line 108: @lgrammel if you like have a look at this type error. It's probably something obvious that I'll figure out quickly after some rest.

lgrammel (approved):

gr2m:
Thanks for getting it over the line @lgrammel 🙏🏼

PR #6189, opened 2025-05-07
feat (ai): support model message array in prompt

lgrammel:
## Background

Limiting the `prompt` property to `string` inputs makes it harder to change from text prompts to message prompts. There is no reason why `prompt` cannot accept `ModelMessage[]` inputs as well.

## Summary

Allow `ModelMessage[]` inputs on `prompt`.

### DIFF
--- a/.changeset/two-otters-whisper.md
+++ b/.changeset/two-otters-whisper.md
@@ -1,0 +1,5 @@
+---
+'ai': major
+---
+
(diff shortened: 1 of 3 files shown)


EurFelux:
#8363 

Hi, I just found that these changes make the type of `prompt` different from ai-sdk docs.

PR #5770, opened 2025-04-14
feat (core, react): refactor `useChat` to use `MessageStore`

iteratetograceness:
Open questions:

- We technically lose the SWR caching across chat id's by relying on a ChatStore that re-initializes when chat id changes. Do we need a hybrid approach?
- What do we mean when we say we want to support robust multi-chat support?
- How can we better handle: partial state clean up, and step calculation?

Things to note:
- How does this new approach translate over to Svelte/Solid/Vue?

TODO, document discussion on alternate considerations + why we landed on final direction

## Tasks

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] If required, a _patch_ changeset for relevant packages has been added
- [ ] You've run `pnpm prettier-fix` to fix any formatting issues

PR #6243, opened 2025-05-09
feat (ai/ui): `useChat` refactor

iteratetograceness:
Note: This is a new branch pulling latest changes from https://github.com/vercel/ai/pull/5770

## Background & Summary

Replaces use of `SWR` for managing CRUD for chat state (messages, status, and error) in `useChat` hook (the scope of this PR only includes changes to the React `useChat` hook). The new `ChatStore` class is now the single source of truth across: 

- `useChat`
- `callChatApi`
- `processChatTextResponse`/`processChatResponse`

**Key Changes**

**ChatStore:**
A subscription-based class that manages chat state and notifies consumers of updates.

**Unified State Management:**
All chat-related modules now interact with the same state instance, ensuring consistency and avoiding race conditions (note: this was commonly faced when users toggled experimental throttle).

**Improved State Updates:**
The subscription model allows for more granular and reliable updates, reducing the risk of race conditions or stale state.

Previously, chat state was managed in a disjointed manner, leading to inconsistent updates and potential synchronization issues between components and API calls. By centralizing state management in ChatStore, we ensure consistent state across all consumers, easier debugging and extension of chat logic, better performance by avoiding unnecessary re-renders or fetches.

## Considerations

1. **Abstraction Layer**: we briefly entertained further abstraction through the introduction of a `Chat` class; this would have meant `ChatStore` would act as a manager of chats instead of a holding the core business logic for per-chat CRUD. We decided against this to avoid de-centralizing state management.
2. [WIP]

## Tasks


packages/ai/core/generate-text/stream-text.ts
```diff
@@ -5,6 +5,7 @@ import {
 import { createIdGenerator, IdGenerator } from '@ai-sdk/provider-utils';
 import { Span } from '@opentelemetry/api';
 import { ServerResponse } from 'node:http';
+import { convertToModelMessages } from '../../src';
```
lgrammel, line 8: rm - users now need to call `convertToModelMessages` in their code.

packages/ai/core/generate-text/stream-text.ts
```diff
@@ -946,7 +947,8 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT, PARTIAL_OUTPUT>
           const initialPrompt = await standardizePrompt({
             system,
             prompt,
-            messages,
+            // @ts-expect-error
+            messages: messages ? convertToModelMessages(messages) : undefined,
```
lgrammel, line 951: revert

packages/ai/src/data-stream/process-data-stream.ts
```diff
@@ -28,6 +29,7 @@ export async function processDataStream({
   onStartStepPart,
 }: {
   stream: ReadableStream<Uint8Array>;
+  onStreamStart?: () => Promise<void> | void;
```
lgrammel, line 32: why is this needed? (just curious)

packages/ai/src/ui/call-chat-api.ts
```diff
@@ -1,8 +1,8 @@
 import { JSONValue } from '@ai-sdk/provider';
 import { IdGenerator } from '@ai-sdk/provider-utils';
+import { ChatStore } from '.';
```
lgrammel, line 3: use file import

packages/ai/src/ui/process-chat-text-response.test.ts
```diff
@@ -9,21 +10,13 @@ function createTextStream(chunks: string[]): ReadableStream<Uint8Array> {
   );
 }
 
-let updateCalls: Array<{
-  message: UIMessage;
-  data: any[] | undefined;
-  replaceLastMessage: boolean;
-}> = [];
+let updateDataCalls: any[] = [];
```
lgrammel, line 13: can `any` be avoided?

packages/ai/src/ui/chat-store.test.ts
```diff
@@ -0,0 +1,1097 @@
...
+            parts: [],
+            createdAt: new Date('2025-01-01'),
+          },
+        ],
+      });
+      expect(store.getMessages(id)).toMatchSnapshot();
+      expect(onChatChanged).toHaveBeenCalledOnce();
+      unsubscribe();
+      store.setMessages({ id, messages: [] });
+      expect(onChatChanged).toHaveBeenCalledOnce();
```
lgrammel, line 130: ideally record calls in an array and validate the call content

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+interface ChatStoreSubscriber {
+  id: string;
+  onChatChanged: (event: ChatStoreEvent) => void;
+}
+
+export enum ChatStoreEvent {
+  ChatMessagesChanged = 'chat-messages-changed',
+  ChatStatusChanged = 'chat-status-changed',
+  ChatErrorChanged = 'chat-error-changed',
+}
```
lgrammel, line 19: Avoid enum (string prefered)

Also would it make sense to have a richer event?

```ts
export interface ChatStoreEvent {
  type: 'chat-messages-changed' | 'chat-status-changed' | 'chat-error-changed',
  chatId: number
} 
```

This will also enable us to do discriminated unions later with different information per event type.

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+
+export interface ChatStoreInitialization {
+  chats?: Record<string, Pick<ChatState, 'messages'>>;
+  getCurrentDate?: () => Date;
```
lgrammel, line 23: testing related props can be hidden, I'm using `"~internal": {...}` nowadays

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+}
+
+export class ChatStore {
+  private chats: Map<string, ChatState>;
+  private subscribers: Map<string, Set<ChatStoreSubscriber>>;
+  private getCurrentDate: () => Date;
+
+  constructor({
+    chats = {},
+    getCurrentDate: customGetCurrentDate = () => new Date(),
```
lgrammel, line 47: no need to rename here i think

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+  TextUIPart,
+  ToolInvocation,
+  ToolInvocationUIPart,
+  UIMessage,
+} from './ui-messages';
+
+interface ChatStoreSubscriber {
+  id: string;
+  onChatChanged: (event: ChatStoreEvent) => void;
+}
```
lgrammel, line 13: I would prefer generic subscribers (not specific to chat `id`) and then push the id filtering into the hook (useChat) -- this will enable more flexible clients to chatStore.

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+    if (!subscribers) return;
+
+    for (const subscriber of subscribers) {
+      subscriber.onChatChanged(event);
+    }
+  }
+
+  private calculateStep(id: string) {
+    const chat = this.chats.get(id);
+    if (!chat) return 0;
```
lgrammel, line 179: should be an error?

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+
+  private resetActiveResponse(id: string) {
+    const chat = this.chats.get(id);
+    if (!chat || !chat.activeResponse) return;
+
+    chat.activeResponse = undefined;
+    chat.error = undefined;
+    chat.status = 'ready';
+    this.emitEvent({ id, event: ChatStoreEvent.ChatStatusChanged });
+    this.emitEvent({ id, event: ChatStoreEvent.ChatErrorChanged });
```
lgrammel, line 214: is this separation needed or can the status change just include the error (also as additional property on the event that can undefined)?

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,518 @@
...
+          existingToolInvocation.args =
+            state === 'successful-parse' ? partialArgs : args;
+          break;
+        }
+        case 'call': {
+          existingToolInvocation.args = toolInvocation.args;
+          break;
+        }
+        case 'result': {
+          // @ts-ignore - caused by `result` not existing on ToolCall type
```
lgrammel, line 428: strange it should have it

packages/react/src/use-chat.ts
```diff
@@ -190,38 +201,78 @@ Default is undefined, which disables throttling.
...
-  const { data: messages, mutate } = useSWR<UIMessage[]>(
-    [chatKey, 'messages'],
-    null,
-    { fallbackData: processedInitialMessages },
-  );
+  const newStoreRef = useRef<ChatStore>();
+  const chatStore = useMemo(() => {
+    if (store) {
+      if (!store.hasChat(chatId)) {
+        store.addChat(chatId, processedInitialMessages);
```
lgrammel, line 208: can we just get rid of initial messages?

packages/react/src/use-chat.ts
```diff
@@ -190,38 +201,78 @@ Default is undefined, which disables throttling.
...
+    [chatStore, chatId],
+  );
+  const messages = useSyncExternalStore(
+    subscribeToMessages,
+    () => chatStore.getMessages(chatId) ?? EMPTY_MESSAGES,
+    () => processedInitialMessages,
+  );
+  const status: 'submitted' | 'streaming' | 'ready' | 'error' =
+    useSyncExternalStore(
+      cb =>
```
lgrammel, line 239: better name?

packages/react/src/use-chat.ts
```diff
@@ -190,38 +201,78 @@ Default is undefined, which disables throttling.
...
+            }
+          },
+        }),
+      () => chatStore.getStatus(chatId) ?? 'ready',
+      () => 'ready',
+    );
+  const error = useSyncExternalStore(
+    cb =>
+      chatStore.subscribe({
+        id: chatId,
```
lgrammel, line 254: how would id changes work? (switching the useChat hook to a different id)

packages/ai/src/data-stream/process-data-stream.ts
iteratetograceness: for updating status! before we were setting status to streaming every time we processed a chunk

packages/ai/src/ui/chat-store.ts
iteratetograceness: this is expected due to the fact that the switch narrows down `existingToolInvocation` but not `toolInvocation` even though we set them to be the same earlier in the code

iteratetograceness: gonna leave a better comment here!

packages/react/src/use-chat.ts
iteratetograceness: good flag! would love your take on the hook approach to chat store i took (used inside useChat)

packages/ai/src/ui/chat-store.ts
```diff
@@ -0,0 +1,528 @@
...
+    if (id) {
+      this.setMessages({ id, messages: [] });
+      this.resetActiveResponse(id);
+    } else {
+      const ids = Array.from(this.chats.keys());
+      for (const id of ids) {
+        this.clear(id);
+      }
+    }
+  }
```
lgrammel, line 527: why not just evict from map? if not, error also needs to be cleared.

needs to abort ongoing streaming?

packages/react/src/use-chat.ts
```diff
@@ -296,19 +322,7 @@ Default is undefined, which disables throttling.
           },
           abortController: () => abortControllerRef.current,
           onResponse,
-          onUpdate({ message, data, replaceLastMessage }) {
-            mutateStatus('streaming');
-
-            throttledMutate(
```
lgrammel, line 302: how is throttling working now

packages/react/src/use-chat.ts
```diff
@@ -183,45 +190,68 @@ Default is undefined, which disables throttling.
...
 
-  // Keep the latest messages in a ref.
-  const messagesRef = useRef<UIMessage[]>(messages || []);
-  useEffect(() => {
-    messagesRef.current = messages || [];
-  }, [messages]);
+  const { subscribe, getMessages, getStatus, getError, setStatus } =
+    useChatStore({
+      store: chatStore.current,
+    });
```
lgrammel, line 214: why is this necessary vs. using chatStore current directly?

packages/react/src/use-chat.ts
```diff
@@ -183,45 +190,68 @@ Default is undefined, which disables throttling.
...
+
+  const messages = useSyncExternalStore(
+    callback =>
+      subscribe({
+        onStoreChange: callback,
+        eventType: 'chat-messages-changed',
+      }),
+    () => getMessages(chatId),
+    () => getMessages(chatId),
```
lgrammel, line 223: last arg can be removed? (optional on useSyncExternalStore)

packages/react/src/use-chat.ts
```diff
@@ -183,45 +190,68 @@ Default is undefined, which disables throttling.
...
+      store: chatStore.current,
+    });
+
+  const messages = useSyncExternalStore(
+    callback =>
+      subscribe({
+        onStoreChange: callback,
+        eventType: 'chat-messages-changed',
+      }),
+    () => getMessages(chatId),
```
lgrammel, line 222: Are the messages immutable? (even deeper in the tree, e.g. some tool invocations)

packages/react/src/use-chat.ts
iteratetograceness: pushed up!

packages/react/src/use-chat.ts
iteratetograceness: i'm leaning towards this so that we can encapsulate the stable selector functions (e.g. `getMessages`) in one spot! like a separation of concerns b/w subscription/reading (within hook) and imperative mutations

packages/react/src/use-chat.ts
iteratetograceness: they are not _truly_ immutable + should be treated as read-only (which we can add to documentation); only mutated via methods on chat store! from our end, to truly enforce immutability, we'd likely want to introduce something like `immer` - would love to discuss this further tomorrow!

PR #6244, opened 2025-05-09
docs: Fix tool calling repair re-ask strategy example

kde713:
## Background

The `ToolCallPart.arg` should be a JSON-serializable object matching the tool's input schema. When the `experimental_repairToolCall` function receives a toolCall parameter as `LanguageModelV1FunctionToolCall` type, its .arg property contains a stringified JSON object. Therefore, when using `toolCall.arg` in a generateText call, we need to parse this stringified JSON object first.

## Summary

Fixed example related to issue mentioned in background section.

## Tasks


gr2m:

PR #6327, opened 2025-05-14
Feat(mistral)  use actual response format

Autumnlight02:
## Background

Currently when calling the mistral AI we are passing on the response schema as an message, but do not actually make use of the structured output endpoint. This commit changes that.

## Summary

1. I implemented that instead of having to rely on the depricated mode.schema, that we now get proper responseFormat results. This should allow us now to stop using the depricated mode field in other packages (like openai) as well.
2. Inside of Mistral I removed the warning and added a handler to actually make use of the responseFormat field and to return the correct Response schema.

## Verification

I manually ran the mistral file inside of the ai core generate-object endpoint. And it worked now as expected. (I would prefer if someone could look over it as well to be certain and test if my ResponseFormat fix actually impacted other packages. (Many packages have error handling for ResponseFormat / warnings for it, but since it always returned undefined those we're never invoked (from what I know, I did not check other packages besides mistral and a bit openai)

## Tasks

## Future Work

People would now need to go around and update all other packages to move away from the depricated mode field. This should lay the first step to make ResponseFormat usuable. 

## Related Issues
Did not check.


Autumnlight02:
Sure can do (will do in a new PR tonight). Do we happen to have an ETA for v5? My company is running into issues during testing, and we would like to use this asap. 

(Also thank you for the quick reply)

lgrammel:
We can fix on main and v5. Having 2 prs works!

Autumnlight02:
Could you point me to a file I can use as a test base? (So I know in which way I should exactly write the test?)

lgrammel:
`mistral-chat-language-model.test.ts` should contain related tests. you can add a new test case.

Autumnlight02:
@lgrammel Fixed a bug and added a test case. Good to merge now.

.changeset/hot-dryers-shout.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/mistral': minor
+'ai': minor
```
lgrammel, line 3: pls change to patch

packages/ai/core/generate-object/generate-object.ts
```diff
@@ -537,6 +537,10 @@ export async function generateObject<SCHEMA, RESULT>({
                     name: schemaName,
                     description: schemaDescription,
                   },
+                  responseFormat: {
+                    schema: outputStrategy.jsonSchema,
+                    type: 'json',
+                  },
```
lgrammel, line 543: in the object json mode mode.schema should be used. pls remove

Autumnlight02: But is the mode not depricated? 
image

The reason why I added responseFormat there was that I've seen that mistral (and I assume others) handle it with a warning, but responseFormat was from what Ive seen never set if you passed on an schema. if I remove this I have to modify my code to handle the schema by the depricated mode field.

If you are still set on removal of this I can do that.

Autumnlight02: @lgrammel Could I get an answer on this

gr2m:

PR #6343, opened 2025-05-15
feat: `ImageModelV2#maxImagesPerCall` can be set to a function that returns a `number` or `undefined`, optionally as a promise

### LINKED ISSUE #6078: [v5] `ImageModelV2` - make `maxImagesPerCall` type: PromiseLike<number> | number
### Description

The `maxImagesPerCall` setting sets the number of images that are requested in a single API call

https://github.com/vercel/ai/blob/9bd5ab57a32142e55654ee805134b609ef33f605/packages/provider/src/image-model/v2/image-model-v2.ts#L35-L39

Providers have defaults per image model:

https://github.com/vercel/ai/blob/c15dfbf3d1f3ecacb2b43576668e9fc7ccdda774/packages/openai/src/openai-image-settings.ts#L8-L12

We want to support use cases where the `maxImagesPerCall` number is not known at the time of the `generateImage()` call, which will require `maxImagesPerCall` to be retrieved asynchronously before the actual API call is done

gr2m:
## Background

The `maxImagesPerCall` setting sets the number of images that are requested in a single API call

https://github.com/vercel/ai/blob/9bd5ab57a32142e55654ee805134b609ef33f605/packages/provider/src/image-model/v2/image-model-v2.ts#L35-L39

Providers have defaults per image model:

https://github.com/vercel/ai/blob/c15dfbf3d1f3ecacb2b43576668e9fc7ccdda774/packages/openai/src/openai-image-settings.ts#L8-L12

We want to support use cases where the `maxImagesPerCall` number is not known at the time of the `generateImage()` call, which will require `maxImagesPerCall` to be retrieved asynchronously before the actual API call is done

## Summary

@lgrammel [suggested](https://github.com/vercel/ai/issues/6078#issuecomment-2844187333) the following interface

```ts
readonly getMaxmagesPerCall(options: { modelId: string }): PromiseLike<number | undefined> | number | undefined
```

I changed it so that `getMaxmagesPerCall` can continue to be set to just a `number` or `undefined`

```ts
readonly maxImagesPerCall: number | undefined | GetMaxImagesPerCallFunction;
```

`GetMaxImagesPerCallFunction` is the type suggested by @lgrammel

## Verification

I altered `examples/ai-core/src/registry/generate-image.ts` to use the OpenAI Image Model and changed the implementation of the `maxImagesPerCall` getter to an async function, then ran the example to confirm that it works as expected.

## Tasks

## Related Issues

Closes #6078


gr2m:
@lgrammel confirmed that this approach will work, I'll continue with tests, docs, and a changeset

lgrammel (approved):

PR #6357, opened 2025-05-16
fix (telemetry): serialize UInt8Arrays as base64 for inner telemetry spans

### LINKED ISSUE #6210: `ImagePart`s sent via experimental_telemetry are bloated
### Description

For `doGenerate` spans, image data is sent as `Uint8Array`, on which `JSON.stringify` acts very weirdly, resulting in a 5-10x bloat on average:

```
Welcome to Node.js v23.3.0.
Type ".help" for more information.
> JSON.stringify(new Uint8Array([42, 33, 22]))
'{"0":42,"1":33,"2":22}'
// I would expect '[42, 33, 22]' probably
```

As a telemetry provider (Laminar), we can implement some conversions on our backend, but I believe this is rather a client side bug than a feature. My understanding is that Uint8Arrays are compact (about 3/4 the size of a base64 string) and easy to carry around in JS, that's why we convert image data to that format. But I'd say such stringification goes against the whole purpose of using a compact data structure.

I tried tracing where and why this happens. Roughly:

`JSON.stringify(promptMessages)` in [core/generate-text/generate-text.ts](https://github.com/vercel/ai/blob/main/packages/ai/core/generate-text/generate-text.ts#L373-L375)
->
`convertToLanguageModelPrompt` in [core/prompt/convert-to-language-model-prompt.ts](https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/convert-to-language-model-prompt.ts#L24)
-> (via several calls in the same file)
`convertDataContentToUint8Array` in [core/prompt/convert-to-language-model-prompt.ts](https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/convert-to-language-model-prompt.ts#L317)
->
`convertBase64ToUint8Array` in [core/prompt/data-content.ts](https://github.com/vercel/ai/blob/main/packages/ai/core/prompt/data-content.ts#L52)

See also [MDN docs examples of JSON.stringify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#examples)

dinmukhamedm:
## Background

`generateObject`, `generateText`, `streamText`, and `streamObject` currently call `JSON.stringify` on the input messages. If the input messages contain an image, it is most likely normalized into a `Uint8Array`.

`JSON.stringify` does not the most obvious things to TypedArrays including `Uint8Array`.

```javascript
// this returns '{"0": 1,"1": 2,"2": 3}', where I'd expect this to be '[1,2,3]'
JSON.stringify(new Uint8array([1, 2, 3]))
```

In practice, this results in bloating images by about 5-15x depending on the original image size. For Laminar, for example, a span with 3 avg sized images will not be able to be sent as it is larger than the (reasonably high) gRPC payload size for our traces endpoint.

From [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#examples):
```javascript
// TypedArray
JSON.stringify([new Int8Array([1]), new Int16Array([1]), new Int32Array([1])]);
// '[{"0":1},{"0":1},{"0":1}]'
JSON.stringify([
  new Uint8Array([1]),
  new Uint8ClampedArray([1]),
  new Uint16Array([1]),
  new Uint32Array([1]),
]);
// '[{"0":1},{"0":1},{"0":1},{"0":1}]'
JSON.stringify([new Float32Array([1]), new Float64Array([1])]);
// '[{"0":1},{"0":1}]'
```

## Summary

Added a function that maps over messages in a `LanguageModelV1Prompt` and maps over content parts in each message, replacing `UInt8Array`s with raw base64 strings instead.

Call this function when calling `recordSpan` for the inner (doStream/doGenerate) span in `generateObject`, `generateText`, `streamText`, and `streamObject`.

## Verification

Ran this small script against a local instance of Laminar and logged the Telemetry payloads (span attributes) on the backend to verify that they are indeed base64.

```javascript
import { Laminar, getTracer } from '@lmnr-ai/lmnr'

Laminar.initialize();

import { openai } from '@ai-sdk/openai'
import { generateText, generateObject, streamText, streamObject, tool } from "ai";
import { z } from "zod";
import dotenv from "dotenv";

dotenv.config();

const handle = async () => {
  const imageUrl = "https://upload.wikimedia.org/wikipedia/commons/b/bc/CoinEx.png"
  const imageData = await fetch(imageUrl)
    .then(response => response.arrayBuffer())
    .then(buffer => Buffer.from(buffer).toString('base64'));

  const o = streamObject({
    schema: z.object({
      text: z.string(),
      companyName: z.string().optional().nullable(),
    }),
    messages: [
      {
        role: "user",
        content: [
          {
            type: "text",
            text: "Describe this image briefly"
          },
          {
            type: "image",
            image: imageData,
            mimeType: "image/png"
          }
        ]
      }
    ],
    model: openai("gpt-4.1-nano"),
    experimental_telemetry: {
      isEnabled: true,
      tracer: getTracer()
    }
  });

  for await (const chunk of o.fullStream) {
    console.log(chunk);
  }
  await Laminar.shutdown();
};

handle().then((r) => {
    console.log(r);
});
```

## Tasks

    -  telemetry is experimental, so I reckon a doc update for this small fix is not required

## Related Issues

Fixes #6210

### DIFF
--- a/packages/ai/core/prompt/stringify-for-telemetry.ts
+++ b/packages/ai/core/prompt/stringify-for-telemetry.ts
@@ -1,0 +1,54 @@
+/**
+ * Helper utility to serialize prompt content for OpenTelemetry tracing.
+ * It is initially created because normalized LanguageModelV1Prompt carries
+ * images as Uint8Arrays, on which JSON.stringify acts weirdly, converting
+ * them to objects with stringified indices as keys, e.g. {"0": 42, "1": 69 }.
+ */
+
+import {
+  LanguageModelV1ImagePart,
+  LanguageModelV1Message,
+  LanguageModelV1Prompt,
+  LanguageModelV1ProviderMetadata,
+} from '@ai-sdk/provider';
+import { convertDataContentToBase64String } from './data-content';
+
+export function stringifyForTelemetry(prompt: LanguageModelV1Prompt): string {
+  const processedPrompt = prompt.map((message: LanguageModelV1Message) => {
+    return {
+      ...message,
+      content:
+        typeof message.content === 'string'
+          ? message.content
+          : message.content.map(processPart),
+    };
+  });
+
+  return JSON.stringify(processedPrompt);
+}
+
+type MessageContentPart = Exclude<
+  LanguageModelV1Message['content'],
+  string
+>[number];
+type ProcessedMessageContentPart =
+  | Exclude<MessageContentPart, LanguageModelV1ImagePart>
+  | {
+      type: 'image';
+      image: string | URL;
+      mimeType?: string;
+      providerMetadata?: LanguageModelV1ProviderMetadata;
+    };
+
+function processPart(part: MessageContentPart): ProcessedMessageContentPart {
+  if (part.type === 'image') {
+    return {
+      ...part,
+      image:
+        part.image instanceof Uint8Array
+          ? convertDataContentToBase64String(part.image)
+          : part.image,
+    };
+  }
+  return part;
+}
--- a/.changeset/rotten-apples-attend.md
+++ b/.changeset/rotten-apples-attend.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Avoid JSON.strinfigy on UInt8Arrays for telemetry
--- a/packages/ai/core/generate-object/generate-object.ts
+++ b/packages/ai/core/generate-object/generate-object.ts
@@ -35,6 +35,7 @@ import { GenerateObjectResult } from './generate-object-result';
 import { injectJsonInstruction } from './inject-json-instruction';
 import { getOutputStrategy } from './output-strategy';
 import { validateObjectGenerationInput } from './validate-object-generation-input';
+import { stringifyForTelemetry } from '../prompt/stringify-for-telemetry';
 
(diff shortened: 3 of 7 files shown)


packages/ai/core/prompt/stringify-for-telemetry.ts
```diff
@@ -0,0 +1,54 @@
...
+      providerMetadata?: LanguageModelV1ProviderMetadata;
+    };
+
+function processPart(part: MessageContentPart): ProcessedMessageContentPart {
+  if (part.type === 'image') {
+    return {
+      ...part,
+      image:
+        part.image instanceof Uint8Array
+          ? convertDataContentToBase64String(part.image)
```
dinmukhamedm, line 49: This will keep the raw base64 data, e.g. 
`IGcfqljA=`

And NOT `data:image/png;base64,IGcfqljA=`

I am open to suggestions on this, but I decided not to do so, because:
- `mimeType` is optional, and not always known
- current Uint8Array is raw data as well and is not aware of mimeTypes
- Telemetry backends can use the `mimeType` field, and there is lots of other heuristics to infer that this is base64

lgrammel (approved):

PR #6384, opened 2025-05-18
chore: wrap `useChat.handleInputChange` in `useCallback`

### LINKED ISSUE #6266: bug: infinite re-render loop because missing `useCallback` in `useChat => handleInputChange`
### Description

Was investigation why my chat re-render so much, and found that inside `useChat`, `handleInputChange` is created without `useCallback`, resulting on different function every time the hook is called.

https://github.com/vercel/ai/blob/396d0134e71b05b346a19fa02e8014e3ee3147e4/packages/react/src/use-chat.ts#L553-L555

## Exploration

```tsx
export const Chat = ({
  initialMessages,
  onStreamingComplete,
}: {
  initialMessages?: UIMessage[],
  onStreamingComplete: (args: { chat: ReturnType<typeof useChat>; }) => void,
}) => {

  // server state
  const useChatOptions = useMemo<UseChatOptions>(
    () => {
      return {
        id: 'FIXED',
        api: 'https://...',
        initialMessages: initialMessages ? initialMessages : undefined,
        onError: error => {
          console.error('An error occurred:', error);
          toast.error('An error occurred. We are investingating! Try different model. \nCODE ' + error.message);
        },
      };
    },
    [
      initialMessages,
    ]
  );
  const chat = useChat(useChatOptions);
  useLogDifferencesBetweenPrevAndCurrentValue(chat); // Utility hook 

  return <ChatUI chat={chat} />;

};

```

Image

**Utility hook used for debug**  

I created this utility debug hook to check what changes between re-render

```ts
import { useEffect, useRef } from "react";

type AnyObject = { [k: string]: unknown; };


const getDifferences = (
  prev: AnyObject,
  current: AnyObject
) => {

  type Difference = {
    key: string,
    type: 'added' | 'removed' | 'changed',
    v: {
      prev: unknown,
      current: unknown,
    },
    isEqual_withObjectIs?: boolean,
    isEqual_withJsonStringify?: boolean,
  };
  const differences: Difference[] = [];

  // get all keys
  const allKeys = Array.from(
    new Set([...Object.keys(prev), ...Object.keys(current)])
  );

  // create diff
  allKeys.forEach(key => {
    const prevV = prev[key];
    const currentV = current[key];

    // if new prop
    if (!prevV && currentV) {
      differences.push({
        key,
        type: 'added',
        v: {
          prev: prevV,
          current: currentV,
        }
      });
      return;
    }

    // if removed prop
    if (prevV && !currentV) {
      differences.push({
        key,
        type: 'removed',
        v: {
          prev: prevV,
          current: currentV,
        }
      });
      return;
    }

    // if not changed
    const isEqual_withObjectIs = Object.is(prevV, currentV);
    const isEqual_withJsonStringify = JSON.stringify(prevV) === JSON.stringify(currentV);
    if (isEqual_withObjectIs && isEqual_withJsonStringify) return;

    // if changed
    differences.push({
      key,
      type: 'changed',
      v: {
        prev: prevV,
        current: currentV,
      },
      isEqual_withObjectIs,
      isEqual_withJsonStringify
    });


  });

  return differences;

};


export const useLogDifferencesBetweenPrevAndCurrentValue = (currentValue: AnyObject) => {
  const prevValue = useRef(currentValue);

  // on every render => log differences
  useEffect(() => {
    const differences = getDifferences(prevValue.current, currentValue);
    prevValue.current = currentValue;
    console.log({ differences });
  });

  return null;
};

```

tresorama:
## Background
Missing useCallback here lead to unnecessary re-renders, in the consumer of `useChat`.

## Summary

Added useCallback to `handleInputChange`

## Verification

Not done.

## Related Issues

Fixes #6266


gr2m:

tresorama:
Will look into that in next days.
For what I remember , this pr makes sense only for v4.

PR #6385, opened 2025-05-19
feat add onParts support for chat API and processing functions

NoelJacob:
Currently this is made for my personal use. It's a hook for any transformation when each part of stream is recieved. Also made prepareBody support async functions. Currently only onFilePart is added and only on react.

Would this feature be appreciated? If so I can make a proper PR.


lgrammel:
thanks - in ai sdk 5 you can implement your own ChatTransport and transform incoming UIMessageStreamPart streams.

PR #6395, opened 2025-05-20
chore: upgrade to pnpm 10

lgrammel:
## Background

The Codex VM uses pnpm 10. I am running into build issues with Codex and want to rule out that they are caused by pnpm version differences.

## Summary

Upgrade pnpm to version 10.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,4 @@
 auto-install-peers = true
 link-workspace-packages = true
+public-hoist-pattern[]=*eslint*
+public-hoist-pattern[]=*prettier*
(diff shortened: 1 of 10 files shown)


.npmrc
```diff
@@ -1,2 +1,4 @@
 auto-install-peers = true
-link-workspace-packages = true
\ No newline at end of file
+link-workspace-packages = true
+public-hoist-pattern[]=*eslint*
```
lgrammel, line 3: see https://github.com/pnpm/pnpm/issues/8378#issuecomment-2636152421

Issue #6405, opened 2025-05-20
[v5] Implement `.providerMetadata` across all image models


gr2m:
### Description

Follow up to #5977

### Strategy

1. Set a break point in [/packages/provider-utils/src/response-handler.ts](https://github.com/vercel/ai/blob/ea7a7c965b4d13ac30d44aae001087b4b0c6e99f/packages/provider-utils/src/response-handler.ts#L29)
in order to inspect the API response body across all providers
1. Run the examples at [examples/ai-core/generate-image/*.ts](https://github.com/vercel/ai/tree/v5/examples/ai-core/src/generate-image)
1. For providers that include any meta data in the response body, update the response schema accordingly in separate pull requests

### Progress

- [x] `@ai-sdk/amazon-bedrock` 

  Update `packages/amazon-bedrock/src/bedrock-image-model.ts`

  - no meta data. The response body only has an `images` key which is an array of base64 strings

- [x] `@ai-sdk/deepinfra` 

  Update `packages/deepinfra/src/deepinfra-image-model.ts`

  - no meta data. The response body only has an `images` key which is an array of `data:image/png,...` strings

- [ ] #6406

- [x] `@ai-sdk/fireworks` 

  Update `packages/fireworks/src/fireworks-image-model.ts`

  - no meta data. Fireworks responds with a binary response body which is the image itself. However there is some information in response headers that might be of interest

    ```js
    {
        "access-control-allow-origin": "*",
        "alt-svc": "h3=\":443\"; ma=86400",
        "cf-cache-status": "DYNAMIC",
        "cf-ray": "942e60a20fb0979e-LAX",
        connection: "keep-alive",
        "content-length": "1348913",
        "content-type": "image/png",
        date: "Tue, 20 May 2025 19:52:59 GMT",
        "finish-reason": "SUCCESS",
        "fireworks-middleware-version": "v2",
        "fireworks-request-generation-ms": "750",
        "fireworks-request-postprocessing-ms": "398",
        "fireworks-request-preprocessing-ms": "1",
        "fireworks-request-queued-ms": "0",
        "fireworks-request-runtime-ms": "1150",
        id: "",
        seed: "1674419546",
        server: "cloudflare",
        "server-timing": "total;dur=1270.0;desc=\"Total Response Time\"",
        vary: "Accept-Encoding",
[...]

PR #6406, opened 2025-05-20
feat (fal): Set `.providerMetaData` for image model responses

gr2m:
## Background

`Fal` Image Models return extra information besides the images themselves. This pull request exposes that information as `.providerMetaData.fal.*` on the `generateImage()` return object.

Example response

```js
{
    images: [
        {
        url: "https://v3.fal.media/files/monkey/YFjatR9EwscWibTqF-NtK.png",
        width: 1024,
        height: 768,
        content_type: "image/jpeg",
        },
    ],
    timings: {
        inference: 0.35681894700974226,
    },
    seed: 731321029,
    has_nsfw_concepts: [
        false,
    ],
    prompt: "A cat wearing an intricate robe while gesticulating wildly, in the style of 80s pop art",
}
```

## Summary

I updated the response schema for Fal's Image Model API responses and set the `.providerMetaData` key to the return object of `FalImageModel#doGEnerate()`.

I normalized the `.nsfw` image meta data since Fal models have two variations

1. `has_nsfw_concepts` - seems to be the standard one
2. `nsfw_content_detected` - only used by https://fal.ai/models/fal-ai/lcm/api#schema-output as far as I can tell

## Verification

I altered `examples/ai-core/src/generate-image/fal.ts`

image

## Tasks

## Related Issues

Part of #6405


packages/fal/src/fal-image-model.ts
```diff
@@ -165,16 +197,54 @@ const falErrorSchema = z.union([falValidationErrorSchema, falHttpErrorSchema]);
...
+  height: z.number().optional(),
+  content_type: z.string().optional(),
+  // e.g. https://fal.ai/models/fal-ai/flowedit/api#schema-output
+  file_name: z.string().optional(),
+  file_data: z.string().optional(),
+  file_size: z.number().optional(),
+});
+
+// https://fal.ai/models/fal-ai/lora/api#type-File
+const LoraFileSchema = z.object({
```
lgrammel, line 210: convention: start non-classes/types (ie regular vars like schemas) with lowercase letter

packages/fal/src/fal-image-model.ts
```diff
@@ -83,6 +87,19 @@ export class FalImageModel implements ImageModelV2 {
     const downloadedImages = await Promise.all(
       targetImages.map(image => this.downloadImage(image.url, abortSignal)),
     );
+    const {
+      // @ts-expect-error - either image or images is present, not both.
```
lgrammel, line 91: is there a typesafe way to handle this?

packages/fal/src/fal-image-model.test.ts
```diff
@@ -173,6 +173,130 @@ describe('FalImageModel', () => {
...
+          seed: undefined,
+          aspectRatio: undefined,
+        });
+        expect(result.providerMetadata).toStrictEqual({
+          fal: {
+            images: [
+              {
+                width: 1024,
+                height: 1024,
+                content_type: 'image/png',
```
lgrammel, line 231: duplicates standardized prop?

packages/fal/src/fal-image-model.ts
gr2m: I'm not using `image` or `images`, the code is just to filter keys fro `responseMetaData`. It's the most efficient way as far as I know.

packages/fal/src/fal-image-model.ts
gr2m: resolved via 101cd0a0c

packages/fal/src/fal-image-model.ts
```diff
@@ -92,6 +109,34 @@ export class FalImageModel implements ImageModelV2 {
...
+            const nsfw =
+              value.has_nsfw_concepts?.[index] ??
+              value.nsfw_content_detected?.[index];
+
+            return {
+              ...imageMetaData,
+              ...(contentType !== undefined ? { contentType } : undefined),
+              ...(fileName !== undefined ? { fileName } : undefined),
+              ...(fileData !== undefined ? { fileData } : undefined),
+              ...(fileSize !== undefined ? { fileSize } : undefined),
```
gr2m, line 133: I normalized some of the `providerMetaData` properties that I think the way we normalized `revised_prompt` to `revisedPrompt` in the Open AI Image model. 

Shall I also camelize `debug_latents`, `debug_per_pass_latents`, and their respective properties?

lgrammel: wonder how far we should to here. usually we do this but it is also overhead

lgrammel (approved):

packages/fal/src/fal-image-model.ts
gr2m: > wonder how far we should to here. usually we do this but it is also overhead 

I agree. I think we should do it for properties that we see repeatedly across providers and models. But e.g. not go overboard with things like `debug_latents` which is specific to a single model from `fal`

I also wonder if we should pass through provider options that we don't handle ourselves. Otherwise we will always play catch up whenever some model from some provider exposes a new information. We could do it in a way that clearly communicates that we don't take responsibility, like `.providerMetaData.fal.unsafe_otherData`  or similar.

PR #6409, opened 2025-05-21
feat (google-vertex): Set `.providerMetaData` for image model responses

gr2m:
## Background

Google's Vertex Image Models return extra information besides the images themselves. This pull request exposes that information as `.providerMetaData.vertex.*` on the `generateImage()` return object.

Example response

```js
{
    predictions: [
        {
            mimeType: "image/png",
            prompt: "A vibrant, high-contrast image ...",
            bytesBase64Encoded: "iVBORw...",
        },
    ]
}
```

## Summary

I updated the response schema for Vertex' Image Model API responses and set the `.providerMetaData` key to the return object of `GoogleVertexImageModel#doGenerate()`.

## Verification

I altered `examples/ai-core/src/generate-image/google-vertex.ts`

image

## Tasks

## Related Issues

Part of #6405

### DIFF
--- a/packages/google-vertex/src/google-vertex-image-model.test.ts
+++ b/packages/google-vertex/src/google-vertex-image-model.test.ts
@@ -329,11 +329,9 @@ describe('GoogleVertexImageModel', () => {
       expect(result.providerMetadata?.vertex).toStrictEqual({
         images: [
           {
-            contentType: 'image/png',
             revisedPrompt: 'revised prompt 1',
           },
           {
-            contentType: 'image/png',
             revisedPrompt: 'revised prompt 2',
           },
         ],
--- a/packages/google-vertex/src/google-vertex-image-model.test.ts
+++ b/packages/google-vertex/src/google-vertex-image-model.test.ts
@@ -27,8 +27,17 @@ describe('GoogleVertexImageModel', () => {
         headers,
         body: {
           predictions: [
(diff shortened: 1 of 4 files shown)


packages/google-vertex/src/google-vertex-image-model.test.ts
```diff
@@ -304,5 +313,31 @@ describe('GoogleVertexImageModel', () => {
...
+        size: undefined,
+        aspectRatio: undefined,
+        seed: undefined,
+        providerOptions: {},
+      });
+
+      expect(result.providerMetadata?.vertex).toStrictEqual({
+        images: [
+          {
+            contentType: 'image/png',
```
lgrammel, line 332: is the content type already part of the standardized response? if yes, then we do not need to duplicate it  here. if no, should it be?

lgrammel (approved):

packages/google-vertex/src/google-vertex-image-model.test.ts
gr2m: Good catch, I missed that it already is: https://github.com/vercel/ai/blob/abf9a792c25f408bafc7e0c277ceafda5ac69b35/packages/ai/core/generate-text/generated-file.ts#L20-L25

I'll update the PR

gr2m: resolved via 2ee6b2c96

Issue #5682, opened 2025-04-11
Request: Support for Zod v4


AsharibAli:
### Feature Description

Hi Maintainers,

Zod v4 was recently announced ([Introducing Zod 4 beta](https://v4.zod.dev/v4)) and is currently in beta. It brings significant improvements over v3, including:

*   Major performance gains (parsing strings, arrays, objects)
*   Reduced core bundle size
*   Improved TypeScript compiler efficiency
*   New features like `z.interface()`, JSON schema conversion, metadata, etc.

While Zod v4 introduces breaking changes (detailed in the [Migration Guide](https://v4.zod.dev/v4/changelog)), upgrading would allow users of `ai` to benefit from these enhancements, especially when defining complex schemas for tools or input/output validation.

Could you please consider adding support for Zod v4 in `ai`? It would be helpful to understand if there are plans to address compatibility with the upcoming stable release of Zod v4.

Thank you for your great work on this library!

dannote:
Maybe it would be better to consider supporting [Standard Schema](https://standardschema.dev) (#3282), although it doesn't support the JSON schema generation out of the box. There is an ongoing discussion regarding schema emission  (standard-schema/standard-schema#7).

pattobrien:
FWIW yesterday zod v4 was released to stable, and is ready for production - so perhaps this request is even more relevant now.

see: https://zod.dev/v4

gr2m:
👋🏼 we are working on supporting both v3 and v4 for validation for the upcoming stable release of ai@5.0.0: https://ai-sdk.dev/docs/announcing-ai-sdk-5-alpha

> Maybe it would be better to consider supporting [Standard Schema](https://standardschema.dev)

This should already be addressed via https://github.com/vercel/ai/pull/6184. Please give `v5` a try to confirm. Note that it's not yet ready for production, but we very much appreciate feedback by early adopters.

gr2m:
I ran into a problem with z4's new `.toJSONSchema()` method. I suggested an approach how to address it in zod 4 itself instead of working around it within the `ai` SKD. Would love to hear your thoughts: https://github.com/colinhacks/zod/issues/4498

gr2m:
The latest v5 alpha release implements support for v4 schemas for validation:
[ai@5.0.0-alpha.8](https://github.com/vercel/ai/releases/tag/ai%405.0.0-alpha.8)

Please give it a try and let us know if it works, or if you run into any problem.

[...]

PR #6421, opened 2025-05-22
feat: support for zod v4 schema validation

gr2m:
## Background

zod 4 has now been released: https://zod.dev/v4

We want to support both zod 3 and zod 4 schemas for validation.

## Summary

- update `zod` to latest (`3.25.32` as of May 28)
- Update implementations of `generateObject()`, `streamObject()`, `generateText()`, `experimental_useObject()` from `@ai-sdk/react`, and `streamUI()` from `@ai-sdk/rsc` to accept both zod v3 and zod v4 (mini) schemas

## Verification

- [x] `generateObject()`: `examples/ai-core/src/generate-object/openai.ts`
- [x] `streamObject()`: `examples/ai-core/src/stream-object/openai.ts`
- [x] `generateText()`: `examples/ai-core/src/generate-text/openai-output-object.ts`
- [x] `experimental_useObject()`: `examples/next-openai/app/use-obj/page.tsx`
- [x] `streamUI()`: `examples/next-openai/app/stream-ui/actions.tsx`, update: [comment](https://github.com/vercel/ai/pull/6421#issuecomment-2917810611)

## Tasks

  All the docs use `import { z } from 'zod';` which is zod v3. Should we update it to `zod/v4` to align with the recommendation on https://zod.dev/?

## Future Work

- [x] When using a zod 4 schema in `examples/next-openai/app/api/use-object/schema.ts`, then `useObject({ api: '/api/use-object', schema: notificationSchema })` in `examples/next-openai/app/use-object/page.tsx` sets the type of `{ object }` to `any`, and I couldn't figure out why that is. For now, I explicitly used a zod 3 schema for `notificationSchema`.
- [x] `packages/svelte/src/structured-object.svelte.ts`
- [x] `packages/svelte/src/tests/structured-object-synchronization.svelte`
- [x] Upgrade all internal uses of `zod` to z4 (~102 files), try using `z4-mini`.

  - moved to #6540

## Related Issues

#5682


packages/ai/core/generate-object/generate-object.ts
```diff
@@ -115,16 +116,23 @@ functionality that can be fully encapsulated in the provider.
...
+      : z4.infer<SCHEMA>
+    : SCHEMA extends z3.Schema
       ? Output extends 'array'
-        ? Array<T>
-        : T
-      : never,
-  SCHEMA extends z.Schema | Schema = z.Schema<JSONValue>,
+        ? Array<
+            // @ts-expect-error - TODO: Type instantiation is excessively deep and possibly infinite
+            z3.infer<SCHEMA>
```
gr2m, line 127: make sure `SCHEMA` is not an instance of zod 4

packages/provider-utils/src/zod-schema.ts
gr2m, line 1: try alternative approach: replace `zodSchema` with `zod3Schema` and `zod4Schema`

gr2m:
One problem we ran into is that the JSON schema created by `z3.object()` and `zod-to-json-schema` differs from `z4.object().toJsonSchema()`

`zod-to-json-schema` seems to set `"additionalProperties": false,` by default on objects. z4's `.toJsonSchema()` does that only for `z4.strictObject().toJsonSchema()`

At least for OpenAI, the missing `"additionalProperties": false` causes an error

gr2m:
My remaining blocker is this error

> Type instantiation is excessively deep and possibly infinite. Likely caused by mixing v3 and v4 types.

That comes up at 3 places

1. https://github.com/vercel/ai/blob/bca63f96a0a7e85a9f713d8d371abf8182fcb682/packages/ai/core/tool/mcp/types.ts#L19
2. https://github.com/vercel/ai/blob/7324c21294042983f930f400d3466b3468d0f788/packages/ai/core/generate-object/stream-object.ts#L165
3. https://github.com/vercel/ai/blob/7324c21294042983f930f400d3466b3468d0f788/packages/ai/core/generate-object/generate-object.ts#L127

@lgrammel mentioned that he ran into this problem before and tracked down the cause being mixing v3 and v4 of Zod.

OscarCornish:
Hey - Super keen to start using this so I've been poking around, figure I'll offer my 2 cents

You appear to have two causes for the recursive type error, the first https://github.com/vercel/ai/blob/bca63f96a0a7e85a9f713d8d371abf8182fcb682/packages/ai/core/tool/mcp/types.ts#L20

Is because when `TOOL_SCHEMAS = 'automatic' | undefined`, the second branch is continually taken, as when it recurses it recurses with `TOOL_SCHEMAS = Record<string, { parameters: ToolParameters<unknown> }>` which is invariant to `Record<string, { parameters: ToolParameters<any> }>`... and so the second branch is taken again, infinitely recursing

I think this is as simple as switching both tool schema records to use unknown?


I'm not so sure about the second one: https://github.com/vercel/ai/blob/7324c21294042983f930f400d3466b3468d0f788/packages/ai/core/generate-object/stream-object.ts#L165 (identical to the other file).

I think it's because you are defining `RESULT` before `SCHEMA`, even though it depends on schema, which is triggering re-evaluation? in combination with the large union of any zod3 object & any zod4 object? I think your comment on the zod repo will yield much clearer answers.

Regardless, I have a fix. By pulling out some of the heavy lifting into helper types:

```ts
type InferSchema<S> =
  S extends z3.Schema ? z3.infer<S> :
  S extends z4.$ZodType ? z4.infer<S> :
  S extends Schema<infer T> ? T :
  never;

type InferFor<S, O extends 'object' | 'array'> =
  O extends 'array' ? Array<InferSchema<S>> : InferSchema<S>;
```

You can write the signature like:

```ts
...
export async function generateObject<
  SCHEMA extends z3.Schema | z4.$ZodType | Schema = z4.$ZodType<JSONValue>,
  Output extends 'object' | 'array' | 'enum' | 'no-schema' =
      InferSchema<SCHEMA> extends string ? 'enum' : 'object',
  RESULT = Output extends 'array' ? InferFor<SCHEMA, 'array'> : InferFor<SCHEMA, 'object'>
>(
  options: Omit<CallSettings, 'stopSequences'> &
 ...
 ```
 
 So that the generics are in order
 
 
 ---
 
Hopefully this is both clear & helpful, apologies that the explanation of the second issue is a bit fuzzy

**Comment by colinhacks** [comment]
Left some reviews. It's *extremely* thoughtful to try to ensure compatibility with Zod 3.24.x. Like, probably too thoughtful. :) Based on a quick scan, the current implementation will not work well for 3.24.x. I explain more in my reviews.

You're going to have a much easier time if you bump your Zod peer dep to 3.25. The timing with `ai@5` works out great for this. There are zero breaking changes from 3.24 -> 3.25 for Zod 3 users. It's just a matter of running `npm upgrade`. In my opinion, a minor version bump in a peer dependency does not qualify as a breaking change since (assuming the peer dependency library didn't break anything between minors) it requires no code changes on the part of your users.

File: packages/provider-utils/src/zod-schema.ts [hunk 2]
```diff
@@ -33,3 +34,64 @@ export function zodSchema<OBJECT>(
...
+  // default to no references (to support openapi conversion for google)
+  const useReferences = options?.useReferences ?? false;
+
+  const z4JSONSchema = z4.toJSONSchema(zodSchema, {
+    target: 'draft-7',
+    io: 'output',
+    reused: useReferences ? 'ref' : 'inline',
+  }) as JSONSchema7;
+
+  console.log({ z4JSONSchema });
```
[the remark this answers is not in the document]
gr2m: 😅 fixed in 26869fe1d

packages/react/src/use-object.ts
```diff
@@ -12,12 +13,16 @@ import {
...
-import z from 'zod';
+import * as z3 from 'zod/v3';
+import * as z4 from 'zod/v4/core';
 
 // use function to allow for mocking in tests:
 const getOriginalFetch = () => fetch;
 
-export type Experimental_UseObjectOptions<RESULT> = {
+export type Experimental_UseObjectOptions<
+  SCHEMA extends z4.$ZodType | z3.Schema | Schema,
```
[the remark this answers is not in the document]
gr2m: If we bumped the peer dependency to `^zod@3.25.x` we wouldn't need that workaround?

gr2m (commented):
Thanks a lot @colinhacks for reviewing the PR, much appreciated!

packages/react/src/use-object.ts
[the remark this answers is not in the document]
gr2m: just saw [your comment below](https://github.com/vercel/ai/pull/6421#issuecomment-2917977336) that answers this question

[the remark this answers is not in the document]
colinhacks: That's correct – because you can have complete confidence that `"zod/v4"` will exist. It was added in 3.25.0 and will never be removed. It's a "permalink" basically. 

I recommend reading these very carefully. This stuff is tricky.

https://github.com/colinhacks/zod/issues/4371
https://zod.dev/library-authors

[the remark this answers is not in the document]
gr2m: fixed via dfb5749ce 👍🏼

gislerro:
I've been playing around with the alpha a bit and noticed that providers (tried it with google gen ai) may reject `zod/v4` schemas - this issue occurs with `z.literal` and `z.enum` and possibly more.

Google expects `enum` properties in JSON schema to have a `type` property - otherwise an error is thrown. Zod does not currently add this property to the JSON schema.

So either the generated JSON schema has to be post-processed or zod's JSONschema generation has to be adapted.

Relevant issues in zod:
literals: https://github.com/colinhacks/zod/issues/4249
enums: https://github.com/colinhacks/zod/pull/4577

Here's a test that illustrates error:

```ts
import type { JSONSchema7 } from "ai@alpha";
import { jsonSchema } from "@ai-sdk/provider-utils@alpha";
import { generateObject } from "ai@alpha";
import { z } from "zod/v4";

import { providers } from "~/alpha/providers";

describe("zod v4 with literals & enum -> generateObject", () => {
  it("should work without patch", async () => {
    const schema = z.strictObject({
      food: z.literal(["pizza", "burger", "pasta"]),
      animal: z.enum(["cats", "dogs"]),
    });

    try {
      const JSONschema = z.toJSONSchema(schema, {
        io: "output",
        target: "draft-7",
      }) as JSONSchema7;

      const manual = jsonSchema(JSONschema, {
        validate(value) {
          const result = schema.safeParse(value);
          return result.success
            ? { success: true, value: result.data }
            : { success: false, error: result.error };
        },
      });
      
      // throws AI_APICallError
      const generated = await generateObject({
        schema: manual,
        model: providers.languageModel("google:gemini-1.5-flash"),
        system: "Generate an object based on the provided json schema",
        prompt: "Pick your favorite options among the available ones",
      });

      console.log(generated.object);
    } catch (e) {
      if (e instanceof Error) {
        console.log(`[API ERROR ${e.name}]\n${e.message}`);
      }
    }

    try {
      const JSONschema = z.toJSONSchema(schema, {
        io: "output",
        target: "draft-7",
      }) as JSONSchema7;

      // @ts-expect-error necessary patch
      JSONschema.properties.food.type = "string";

      // @ts-expect-error necessary patch
      JSONschema.properties.animal.type = "string";

      const manual = jsonSchema(JSONschema, {
        validate(value) {
          const result = schema.safeParse(value);
          return result.success
            ? { success: true, value: result.data }
            : { success: false, error: result.error };
        },
      });

      // doesn't throw
      const generated = await generateObject({
        schema: manual,
        model: providers.languageModel("google:gemini-1.5-flash"),
        system: "Generate an object based on the provided json schema",
        prompt: "Pick your favorite options among the available ones",
      });

      console.log(generated.object);
    } catch (e) {
      console.error(e);
      throw e;
    }
  });
});
```

Here's the output of the above test (I've intercepted the api calls in `providers`)
```
{
  input: 'https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent',
  method: 'POST',
  body: {
    generationConfig: {
      responseMimeType: 'application/json',
      responseSchema: {
        required: [ 'food', 'animal', [length]: 2 ],
        type: 'object',
        properties: {
          food: { enum: [ 'pizza', 'burger', 'pasta', [length]: 3 ] },
          animal: { enum: [ 'cats', 'dogs', [length]: 2 ] }
        }
      }
    },
    contents: [
      {
        role: 'user',
        parts: [
          {
            text: 'Pick your favorite options among the available ones'
          },
          [length]: 1
        ]
      },
      [length]: 1
    ],
    systemInstruction: {
      parts: [
        {
          text: 'Generate an object based on the provided json schema'
        },
        [length]: 1
      ]
    }
  }
}
[API ERROR AI_APICallError]
* GenerateContentRequest.generation_config.response_schema.properties[food].enum: only allowed for STRING type
* GenerateContentRequest.generation_config.response_schema.properties[animal].enum: only allowed for STRING type

{
  input: 'https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent',
  method: 'POST',
  body: {
    generationConfig: {
      responseMimeType: 'application/json',
      responseSchema: {
        required: [ 'food', 'animal', [length]: 2 ],
        type: 'object',
        properties: {
          food: {
            type: 'string',
            enum: [ 'pizza', 'burger', 'pasta', [length]: 3 ]
          },
          animal: { type: 'string', enum: [ 'cats', 'dogs', [length]: 2 ] }
        }
      }
    },
    contents: [
      {
        role: 'user',
        parts: [
          {
            text: 'Pick your favorite options among the available ones'
          },
          [length]: 1
        ]
      },
      [length]: 1
    ],
    systemInstruction: {
      parts: [
        {
          text: 'Generate an object based on the provided json schema'
        },
        [length]: 1
      ]
    }
  }
}
{ food: 'pizza', animal: 'cats' }
```

packages/provider-utils/src/zod-schema.ts
```diff
@@ -33,3 +34,60 @@ export function zodSchema<OBJECT>(
...
+     * However, not all language models and providers support such references.
+     * Defaults to `false`.
+     */
+    useReferences?: boolean;
+  },
+): Schema<OBJECT> {
+  // default to no references (to support openapi conversion for google)
+  const useReferences = options?.useReferences ?? false;
+
+  const z4JSONSchema = z4.toJSONSchema(zodSchema, {
```
gislerro, line 53: This would post process the schema, I don't know whether the type should be inferred from the zodSchema directly or just using `"string"` is enough since zod will parse it anyways

```ts
      const z4JSONSchema = z.toJSONSchema(schema, {
        io: "output",
        target: "draft-7",
        override({ zodSchema, jsonSchema }) {
          if (zodSchema instanceof z.ZodLiteral) {
            jsonSchema.type = "string";
          }

          if (zodSchema instanceof z.ZodEnum) {
            jsonSchema.type = "string";
          }
        },
      }) as JSONSchema7;

File: packages/provider-utils/src/zod-schema.ts [hunk 5]
```diff
@@ -33,3 +34,60 @@ export function zodSchema<OBJECT>(
...
+    target: 'draft-7',
+    io: 'output',
+    reused: useReferences ? 'ref' : 'inline',
```
gislerro, line 56: This would post process the schema so providers like google don't complain about the schema.

I don't know whether the type should be inferred from the zodSchema directly or just using "string" is enough since zod will parse it anyways

```ts
        override({ zodSchema, jsonSchema }) {
          if (zodSchema instanceof z.ZodLiteral) {
            jsonSchema.type = "string";

          if (zodSchema instanceof z.ZodEnum) {
            jsonSchema.type = "string";
          }
        },
 ```

**Comment by gr2m** [comment]
Thanks for the detailed comment!

Looks like the fix for `enums` is already merged, I'll make sure to bump the version range accordingly

> enums: [colinhacks/zod#4577](https://github.com/colinhacks/zod/pull/4577)

For literals, I'll look into implementing it using an override if the change does land in zod

File: packages/provider-utils/src/schema.ts [hunk 6]
```diff
@@ -62,7 +71,11 @@ function isSchema(value: unknown): value is Schema {
 }
 
 export function asSchema<OBJECT>(
-  schema: z.Schema<OBJECT, z.ZodTypeDef, any> | Schema<OBJECT> | undefined,
+  schema:
+    | z4.$ZodType<OBJECT, any>
+    | z3.Schema<OBJECT, z3.ZodTypeDef, any>
+    | Schema<OBJECT>
```
lgrammel, line 77: wondering if it makes sense to extract this into a dedicated type e.g. extended schema

lgrammel (approved):

packages/provider-utils/src/schema.ts
gr2m: I'm not 100% sure what you mean but happy to pair on it this week?

gr2m:
I confirmed that `z.enum()` is working with [adf1acc](https://github.com/vercel/ai/pull/6421/commits/adf1acc1bc5dc6f3ef99d9cc1dd09c5279c2a52f). I also confirmed that `z.literal([...])` is still failing. I'm tracking https://github.com/colinhacks/zod/pull/4590 to see if it gets merged or not. If it does, I'll bump the `zod` version in a follow up PR. If not, I'll create an issue to discuss how to best address it in the `ai` SDK

PR #6419, opened 2025-05-21
feat (providers/vercel): initial vercel provider

ctate:
## Background

Vercel is adding an API for the v0 model per https://vercel.com/docs/v0/api and an AI SDK provider would make it easier for developers to interact with it.

## Summary

Added an initial provider for the Vercel API.

## Verification

Added examples across several features and executed end to end manually.

## Tasks

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)

PR #6464, opened 2025-05-23
feat (providers/vercel): initial vercel provider

shaper:
## Background

Port of #6419 

Vercel is adding an API for the v0 model per https://vercel.com/docs/v0/api and an AI SDK provider would make it easier for developers to interact with it.

## Summary

Added an initial provider for the Vercel API.

## Verification

Added examples across several features and executed end to end manually.

## Tasks


.github/workflows/ci.yml
```diff
@@ -100,6 +100,9 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Build packages
```
shaper, line 103: Without this, the type-check fails as the `@ai-sdk/vercel` package hasn't been built. I'm not sure why this package is the first to trigger this failure. It seems logical that we would need the packages built for the type-check to succeed. Perhaps the existing packages are/were cached by some other build or action, and we've been getting lucky?

examples/ai-core/src/generate-text/vercel-image.ts
shaper, line 1: Both of the image example scripts fail with a v0-side server error on the v5 branch, but they work on ToT. Are there known image attachment issues, else I'll debug further with @ctate

packages/vercel/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@ai-sdk/vercel",
+  "version": "0.0.0",
```
lgrammel, line 3: 1.0.0-alpha.4

files under /.changeset might also need updates, and there are other link files (search for how other provider pkgs are referenced)

.github/workflows/ci.yml
lgrammel: the type check does not need pkgs to be built afaik

lgrammel (approved):

PR #6498, opened 2025-05-27
feat (ui): add onFinish to createUIMessageStream

lgrammel:
## Background

To enable persistence for customized UI message streams, `createUIMessageStream` needs an notification on finish with the reconciled messages.

## Summary

Add `onFinish` and related values to `createUIMessageStream`.

### DIFF
--- a/.changeset/gentle-mayflies-call.md
+++ b/.changeset/gentle-mayflies-call.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+feat (ui): add onFinish to createUIMessageStream
--- a/examples/next-openai/app/api/use-chat-custom-sources/route.ts
+++ b/examples/next-openai/app/api/use-chat-custom-sources/route.ts
@@ -4,13 +4,16 @@ import {
   createUIMessageStream,
   createUIMessageStreamResponse,
   streamText,
+  UIMessage,
 } from 'ai';
 
 export async function POST(req: Request) {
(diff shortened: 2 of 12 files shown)


remorses:
This means you now always need to call `createUIMessageStream` if you want to persist messages.

I don't like the direction `ai` is taking, I don't want to call `createUIMessageStream`, I want to implement my own serialization (or use the RSC format directly with the async iterator created by `streamText().fullStream`), `ai` now requires you to always create a `UIMessageStream` instead of working with the result of `streamText`. 

An `UIMessageStream` is yet another stream format added to the package when we already have the one from `streamText().fullStream` which people already know. Why add yet another one?

What I would like to see instead is for these features to not require the `UIMessageStream` serialization:
- Convert response messages to `UIMessage` for persistence
- Emit custom data parts from tools `execute`

This could be done by 
- supporting generators in the tools execute functions, yielded items will simply become items in `streamText().fullStream`
- keeping `appendResponseMessages` or something similar
- adding a function `messagesFromfullStream` that creates an `UIMessage` from a `streamText().fullStream` (now this function requires an ui stream instead of the result of `streamText().fullStream`, when both formats are basically the same thing)

remorses:
If `UIMessageStream` becomes such an important part of the framework maybe it's worth adding it as `result.uiStream`. This way you can remove `createUIMessageStream` and `writer`. Tools would be able to yield data parts via async generators

PR #6554, opened 2025-05-30
feat (codemods): updates for v5

gr2m:
## Tasks

## Related Issues

Closes #6552


packages/codemod/src/codemods/replace-generatetext-text-property.ts
```diff
@@ -0,0 +1,46 @@
...
+        return false;
+      }
+
+      // The object must be a simple identifier (not a member expression)
+      if (!j.Identifier.check(node.object)) {
+        return false;
+      }
+
+      // The identifier should be a common generateText result variable name
+      const commonNames = [
```
imranbarbhuiya, line 24: This is wrong approach and replaces all the places where result.text is used, for example using fetch with response variable then doing response.text() or any other result.text. Also misses places where common names arent used. I would like work on this to fix. I dont think I can make it perfect but atleast i can make it to get all variable names from this file that are result of `generateText` then transform usage of them only. This still has issue where same file uses same named variable twice, one from generateText and another from fetch but still better than transforming from all files where ai isnt even used

gr2m:
merged via #7264

PR #6559, opened 2025-05-31
fix (providers/google): update supportedUrls to only support native URL

### LINKED ISSUE #6558: Google provider does not download images provided through non-native URLs
### Description

The Google provider [includes](https://github.com/vercel/ai/blob/989ac756bf2097b15bf0596110496b815074b91c/packages/google/src/google-provider.ts#L96-L111) `https://*` urls as part of `supportedUrls` in its provider specification, as a result, it does not download non-native URLs resulting in the following error from the provider

```
error: {
  code: 400,
    message: 'Invalid or unsupported file uri: https://<redacted>',
    status: 'INVALID_ARGUMENT'
  }
}
```

The provider only supports URLs that start with `https://generativelanguage.googleapis.com/v1beta/files/...`, while others need to be downloaded.

### AI SDK Version

- ai: 5.0.0-alpha.7
- @ai-sdk/google: 2.0.0-alpha.7

jeremyphilemon:
## Background

Fixes the ability to use non-native links to files in file parts when using the Google provider.

## Summary

Updated the `supportedUrls` of the google provider specification to only include the native URL, which is `https://generativelanguage.googleapis.com/v1beta/files/...`.

## Verification
Manually verified by running the `next-openai` example, switching the provider to use `google("gemini-2.0-flash")`, and testing the example at `/use-chat-attachments-url`.

## Tasks

## Related Issues
Fixes #6558

### DIFF
--- a/packages/google/src/google-provider.ts
+++ b/packages/google/src/google-provider.ts
@@ -103,8 +103,9 @@ export function createGoogleGenerativeAI(
       generateId: options.generateId ?? generateId,
       supportedUrls: () => ({
         '*': [
-          // HTTP URLs:
-          /^https?:\/\/.*$/,
+          // Only allow requests to the Google Generative Language "files" endpoint
+          // e.g. https://generativelanguage.googleapis.com/v1beta/files/...
+          new RegExp(`^${baseURL}/files/.*$`),
         ],
       }),
       fetch: options.fetch,
--- a/.changeset/empty-flowers-sniff.md
+++ b/.changeset/empty-flowers-sniff.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+update supportedUrls to only support native URL


packages/google/src/google-provider.ts
```diff
@@ -103,8 +103,9 @@ export function createGoogleGenerativeAI(
       generateId: options.generateId ?? generateId,
       supportedUrls: () => ({
         '*': [
-          // HTTP URLs:
-          /^https?:\/\/.*$/,
+          // Only allow requests to the Google Generative Language "files" endpoint
+          // e.g. https://generativelanguage.googleapis.com/v1beta/files/...
+          new RegExp(`^${baseURL}/files/.*$`),
```
lgrammel, line 108: this would also apply to images - are web url images not supported?

jeremyphilemon: yes web url images are not supported, only image links with that prefix are supported unfortunately.

lgrammel (approved):

PR #6614, opened 2025-06-04
add optimistic routing and cache revalidation

jeremyphilemon:


examples/next/app/chat/[chatId]/chat.tsx
```diff
@@ -24,6 +25,9 @@ export default function Chat({
         }),
         messageMetadataSchema: myMessageMetadataSchema,
       }),
+      onFinish: async () => {
+        await invalidateRouterCache();
+      },
```
Comment (lgrammel) [inline examples/next/app/chat/[chatId]/chat.tsx:30]: why would we do this?

PR #6618, opened 2025-06-04
fix: add url validation for google vertex anthropic provider

dancer:
## background

google vertex anthropic provider was throwing cryptic "url sources are not supported" errors after commit aeba38e added url support to base anthropic provider. both providers share the same anthropicmessageslanguagemodel but vertex has supportsimageurls: false while base has true.  issue #6453 

## summary

removed unnecessary url validation from anthropic provider conversion logic:

- removed supportsImageUrls parameter from convertToAnthropicMessagesPrompt function
- removed url validation checks that threw errors for url-based images and pdf files
- updated AnthropicMessagesLanguageModel to no longer pass supportsImageUrls setting
- updated tests to verify url-based images and pdfs are properly passed through
- relies on AI SDK core convertToLanguageModelPrompt to handle url downloading when needed

## verification

- all 64 anthropic tests pass in both node and edge environments
- all 68 google vertex tests pass including anthropic provider tests
- url-based images and pdf files are properly converted to anthropic format
- ai sdk core automatically downloads urls when modelSupportsImageUrls is false
- provider-level conversion trusts that urls have been pre-processed by ai sdk core

## tasks

- [x] removed supportsImageUrls parameter from conversion function
- [x] removed url validation error throwing logic
- [x] updated anthropic language model to not pass supportsImageUrls
- [x] updated tests to verify url passthrough instead of error throwing
- [x] verified ai sdk core handles url downloading automatically
- [x] all anthropic tests passing (64 tests)
- [x] all google vertex tests passing (68 tests)

## implementation details

the fix relies on the two-layer architecture of the ai sdk:

- **ai sdk core layer**: `convertToLanguageModelPrompt()` handles url downloading when `modelSupportsImageUrls: false` or `modelSupportsUrl()` returns false
- **provider layer**: `convertToAnthropicMessagesPrompt()` trusts that urls have been pre-processed and focuses only on anthropic-specific formatting

this approach eliminates the error users were seeing while maintaining the proper url handling behavior through the established ai sdk patterns.

## personal note

simpler fix that leverages existing ai sdk architecture. url downloading happens automatically at the right level without provider-specific validation logic.

### DIFF
--- a/packages/google-vertex/src/anthropic/google-vertex-anthropic-provider.test.ts
+++ b/packages/google-vertex/src/anthropic/google-vertex-anthropic-provider.test.ts
@@ -105,4 +105,38 @@ describe('google-vertex-anthropic-provider', () => {
       }),
     );
   });
+
+  it('should create a Google Vertex Anthropic provider instance with custom settings', () => {
+    const customProvider = createVertexAnthropic({
+      project: 'custom-project',
+      location: 'custom-location',
+      baseURL: 'https://custom.base.url',
+      headers: { 'Custom-Header': 'value' },
+    });
+
+    expect(customProvider).toBeDefined();
+    expect(typeof customProvider).toBe('function');
+    expect(customProvider.languageModel).toBeDefined();
+  });
+
+  it('should not support URL sources to force base64 conversion', () => {
+    const provider = createVertexAnthropic();
+    provider('test-model-id');
+
+    // Assert that the model constructor was called with supportedUrls function
+    expect(AnthropicMessagesLanguageModel).toHaveBeenCalledWith(
+      'test-model-id',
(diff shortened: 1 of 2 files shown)


packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
```diff
@@ -213,7 +213,7 @@ describe('user messages', () => {
             content: [
               {
                 type: 'file',
-                data: 'base64data',
+                data: 'data:text/plain;base64,SGVsbG8gV29ybGQ=',
```
lgrammel: is this the right format here? file content parts do no contain string like that afaik

lgrammel (approved):

PR #6661, opened 2025-06-06
refactoring: push subscription mechanism into react chat

elliott-with-the-longest-name-on-github:
## Background

The chat subscription mechanism is only needed for the react abstraction.

## Summary

Make the subscription mechanism on Chat react only.


lgrammel:
I feel uneasy about this change. Currently sharing the same chat object with multiple useChat hooks through a react context is not covered by tests. How would this work with this version? passing in Chat objects into useChat seems to have been removed. I have some other concerns as well but that's the main one

elliott-with-the-longest-name-on-github:
@lgrammel I gave what you were saying some thought and revamped the PR to keep the `useChat` API exactly the same. The only difference now is that the `Chat` class from the React package is now the "container" for subscriptions rather than the `AbstractChat`. The major benefit of this is that `AbstractChat` is no longer multimodal / leaky as an abstraction -- it uses the `ChatState` abstraction passed to it and doesn't have to worry about _also_ firing events. This will prevent bugs in the future where we remember to call `pushMessage` but forget to emit an event, or vice versa.

(In reality there's really no difference in functionality whatsoever here -- previously, the "subscriber" logic lived on `AbstractChat`, and now it lives on `Chat`, which extends `AbstractChat` in the first place. The way that API is _expressed_ has changed a little bit, but the functionality has not.)

The only downside to this is that the `Chat` class now has a `~register*Callback` function on it for each of the three types of state. I prefixed it with that for the same reason the folks over at StandardSchema did: Because it sorts it to the bottom of the autocomplete list. The only reason they're necessary at all is because React needs some place to "anchor" state invalidations within `useChat`, and `useSyncExternalStore` is that "anchor" position. 

Overall this feels like a much better separation of concerns -- React keeps all of the React-y things contained in the React package, and the core package just says "hey, update your state like this!"

iteratetograceness (approved):
Thanks for adding the `Set`-based multi-subscriber approach! This feels more React "native" and simple (esp. with the hook encapsulation) over the previously event-driven approach that was only being using by React

iteratetograceness:
Smol poke on Prettier + changeset!

elliott-with-the-longest-name-on-github:
I'm not sure this should get a changeset bc the public API hasn't changed and no bugs have been fixed -- this is purely optimization

did fix prettier though

PR #6665, opened 2025-06-07
Add support for type safe tool calls to `UIMessage`, adding a generic parameter

remorses:


remorses:
using tool calls directly in UI is an anti pattern because it couples the UI with the AI tools implementation and it’s difficult to make it keep working for past chats, needing migrations

PR #6708, opened 2025-06-11
feat (provider/openai): add `reasoning-part-finish` to openai responses provider

### LINKED ISSUE #6629: Reasoning parts should include a `summary_index` field to distinguish between different reasoning summaries
### Description

Openai returns many reasoning summaries, each one has a different `summary_index` number, for example a reasoning delta can have the following shape:

```json
{
  type: 'response.reasoning_summary_text.delta',
  item_id: 'rs_6840931d804c81989497d944ce3842010581aba909b8922e',
  output_index: 0,
  summary_index: 2,
  delta: '!'
}
```

The reasoning items returned by `ai` `fullStream` instead have the following shape:

```json
{
  "type": "reasoning",
  "reasoning": "..."
}
```

Because of this issue if the UI tries to render the reasoning summaries there will be parts where 2 summaries are put one besides the other, where the next summaries starts after a dot without a space, for example:

`So now, I must call the function individually for each city and then summarize the weather for each in a final output.**Planning weather calls**\n\nI need to ca`

Without this field it's impossible to render the reasoning summaries nicely.

### AI SDK Version

Both v5 and v4 miss this field

remorses:
## Background

The openai responses provider was missing the reasoning-part-finish parts, which makes it impossible to distinguish reasoning summaries in the UI.

## Summary

Add `reasoning-part-finish` to openai responses provider.

## Related Issues

Fixes #6629

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.ts
@@ -398,6 +398,12 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
     > = {};
     let hasToolCalls = false;
 
+    // Track the last summary_index received in reasoning summary chunks.
+    // When the provider switches to a different summary_index we emit a
+    // `reasoning-part-finish` marker so downstream consumers can separate
+    // distinct reasoning parts.
+    let lastReasoningSummaryIndex: number | null = null;
+
     return {
       stream: response.pipeThrough(
         new TransformStream<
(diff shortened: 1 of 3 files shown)


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -615,6 +634,14 @@ const responseReasoningSummaryTextDeltaSchema = z.object({
   delta: z.string(),
 });
 
+const responseReasoningSummaryPartDoneSchema = z.object({
+  type: z.literal('response.reasoning_summary_part.done'),
+  item_id: z.string(),
+  output_index: z.number(),
+  summary_index: z.number(),
+  part: z.unknown().optional(),
```
lgrammel: prefer nullish for robustness

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -459,10 +465,23 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
...
+              // When the summary_index changes, mark the end of the previous
+              // reasoning part before streaming the new one.
+              if (
+                lastReasoningSummaryIndex !== null &&
+                value.summary_index !== lastReasoningSummaryIndex
+              ) {
+                controller.enqueue({ type: 'reasoning-part-finish' });
+              }
+
+              lastReasoningSummaryIndex = value.summary_index;
```
lgrammel, line 477: this assumes that reasoning chunks arrive sequentially (necessary because of how the language model is chunks are structured). fine for now but something to potentially look at in the future (for something like language model v3)

lgrammel:
Can you add a patch changeset for the openai provider?

lgrammel (approved):

PR #6716, opened 2025-06-11
build: Add `node` export for better compatibility

mydea:
## Background

On Next.js, we noticed that instrumentation via `import-in-the-middle` of this package was not working. After some deep debugging, I figured out that the problem seems to be that it keeps picking up the `.mjs` variant in Next.js, although Next itself is cjs. That seems to mess with import-in-the-middle (or actually require-in-the-middle, I suppose).

The only way to fix this and get it to load the cjs correctly in Next.js was to add a `node` variant export.

## Summary

Added a `node` variant export that points to the cjs outout.


lgrammel:
Hi can you rebase against the latest main? we are in the final stretch towards ai sdk 5 ga and i would love to see if we can get this in

gr2m:

PR #6747, opened 2025-06-14
chore (ai): Export MockImageModelV1

c-ehrlich:
## Background

I would like to use `MockImageModelV1` in my code. The equivalent `MockEmbeddingModelV1` and `MockLanguageModelV1` are already exported.

## Summary

I added the export


lgrammel:
#7199

PR #6759, opened 2025-06-16
fix(anthropic): resolve web search API validation errors with partial location

dancer:
## background

API validation errors when using Anthropic web search with partial location data. The API was rejecting requests with empty strings for optional fields, preventing flexible location-based search configurations.

## summary

- make all userLocation fields optional (city, region, country, timezone)
- add conditional field inclusion in tool preparation
- add validation warning for empty userLocation objects

## verification

- all tests pass including new optional field combinations
- real API calls work with any single field (city-only, country-only, timezone-only)
- empty userLocation generates proper unsupported-tool warning
- no userLocation at all works correctly

## tasks

- [x] updated userLocation types in anthropic-tools.ts and anthropic-api-types.ts
- [x] conditional field inclusion logic in anthropic-prepare-tools.ts
- [x] validation for empty userLocation with warning generation
- [x] removed redundant test cases
- [x] comprehensive test coverage for all field combinations


packages/anthropic/src/anthropic-prepare-tools.ts
```diff
@@ -134,14 +134,31 @@ export function prepareTools({
...
+              }
+              if (loc.region) {
+                webSearchTool.user_location.region = loc.region;
+              }
+              if (loc.country) {
+                webSearchTool.user_location.country = loc.country;
+              }
+              if (loc.timezone) {
+                webSearchTool.user_location.timezone = loc.timezone;
+              }
```
lgrammel: unnecessary, will be removed automatically by json stringify if undefined

packages/anthropic/src/anthropic-prepare-tools.ts
```diff
@@ -134,13 +134,20 @@ export function prepareTools({
                 .blockedDomains as string[];
             }
             if (tool.args.userLocation) {
-              const loc = tool.args.userLocation as any;
+              const loc = tool.args.userLocation as {
```
lgrammel: why is the cast needed?

lgrammel (approved):

PR #6762, opened 2025-06-16
fix (react): `structuredClone` message in `replaceMessage`

iteratetograceness:
## Background

When using the React Compiler (currently in RC), deeply nested mutations to message parts (and likely metadata) fail to trigger re-renders even though we are re-creating the entire array messages array upon every chunk (in `replaceMessage`) 

## Summary

In `replaceMessage`, we use `structuredClone` to guarantee the message being updated reflected deeply nested changes in the React state manager

Alternatives considered:

- Apply snapshot when executing mutate job (in `AbstractChat`); this approach was vetoed because the snapshotting should be collocated to the React implementation
- Add custom cloning approach to avoid deep cloning the entire message; if users report a huge impact on performance, we can potentially pivot to this approach. For now, this is the simplest solution!

## Verification

Tested in demo Next.js project with React Compiler enabled

## Tasks

## Future Work

See custom clone/snapshot method snippet above

## Related Issues

Addresses https://github.com/vercel/ai/issues/6466

### DIFF
--- a/packages/react/src/chat.react.ts
+++ b/packages/react/src/chat.react.ts
@@ -18,8 +18,6 @@ class ReactChatState<UI_MESSAGE extends UIMessage>
   #statusCallbacks = new Set<() => void>();
   #errorCallbacks = new Set<() => void>();
 
-  #clonedMessages = new WeakMap();
-
   constructor(initialMessages: UI_MESSAGE[] = []) {
     this.#messages = initialMessages;
   }
--- a/packages/react/src/chat.react.ts
+++ b/packages/react/src/chat.react.ts
@@ -1,13 +1,4 @@
-import {
-  AbstractChat,
-  ChatInit,
-  ChatState,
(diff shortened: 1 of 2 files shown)


lgrammel (approved):

packages/react/src/chat.react.ts
```diff
@@ -71,7 +62,7 @@ class ReactChatState<UI_MESSAGE extends UIMessage>
   replaceMessage = (index: number, message: UI_MESSAGE) => {
     this.#messages = [
       ...this.#messages.slice(0, index),
-      message,
+      this.snapshot(message),
```
lgrammel, line 66: since there is no test, can you add a comment here?

iteratetograceness: yes can do!

lgrammel:
pls add a changeset

PR #6773, opened 2025-06-17
feat (ui): extended regenerate support

lgrammel:
## Background

Reload was slightly misleading as a name. Regenerating any assistant message should be supported.

## Summary

* rename `reload` to `regenerate` on chat
* support optional `messageId` param and regenerate on user msgs
* rename transport `submitMessages` and related to `sendMessages`

## Verification

Tested with `next` example.

### DIFF
--- a/.changeset/seven-fans-speak.md
+++ b/.changeset/seven-fans-speak.md
@@ -1,0 +1,5 @@
+---
+'ai': major
+---
+
+feat (ui): extended regenerate support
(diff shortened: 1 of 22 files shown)


nicoalbanese (approved):

oagix:
Hey @lgrammel ,

The previous `reload` function had support for a `body` used to pass data with the regenerate request. The new `regenerate` doesn't.

image

image

Facing 2 issues with the `regenerate` to `reload` update:

1 - Getting errors like `Response Error Error: GraphQL error: expecting a value for non-nullable variable: "conversation_uuid"` because regenerate no longer sends all the per-message data sendMessage does

2 - It won't submit the custom body anymore (for model_slug for example).

What's the best way to pass ALL the custom data we normally pass via sendMessage per message, but in the `regenerate` function?

PR #6825, opened 2025-06-22
fix (ai): add tests and examples for openai responses

dancer:
## background

frontend testing was needed to reasoning for openai response model

## summary

- add frontend test pages for openai response reasoning
- added reasoning start delta end to ui-message-stream-parts

## verification

- all tests pass
- raw chunk capture tests

## tasks

- [x] test pages with useChat and DefaultChatTransport
- [x] raw streaming verification tests

## future work

* add any other providers if needed

### DIFF
--- a/packages/ai/src/ui-message-stream/ui-message-stream-parts.ts
+++ b/packages/ai/src/ui-message-stream/ui-message-stream-parts.ts
@@ -58,6 +58,22 @@ export const uiMessageStreamPartSchema = z.union([
     providerMetadata: z.record(z.any()).optional(),
   }),
   z.strictObject({
+    type: z.literal('reasoning-start'),
+    id: z.string(),
+    providerMetadata: z.record(z.any()).optional(),
+  }),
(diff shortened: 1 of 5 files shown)


packages/ai/src/ui-message-stream/ui-message-stream-parts.ts
```diff
@@ -57,6 +57,22 @@ export const uiMessageStreamPartSchema = z.union([
...
+    type: z.literal('reasoning-delta'),
+    id: z.string(),
+    delta: z.string(),
+    providerMetadata: z.record(z.any()).optional(),
+  }),
+  z.strictObject({
+    type: z.literal('reasoning-end'),
+    id: z.string(),
+    providerMetadata: z.record(z.any()).optional(),
+  }),
```
lgrammel, line 75: good find but this makes it a fix. can you add a changeset?

lgrammel (approved):

PR #6846, opened 2025-06-24
fix(ai): selectTelemetryAttributes attribute's value should ignore `u…

Gaubee:
## Background

The `selectTelemetryAttributes` function was previously filtering out attributes only if their value was strictly `undefined`. However, in dynamic JavaScript environments or when interoperating with other libraries, it's common for an attribute's value to be `null`.

The existing implementation would incorrectly retain attributes with `null` values, which could lead to unexpected behavior or errors in downstream telemetry processing. This change makes the attribute filtering more robust by handling both `null` and `undefined` cases, acknowledging that TypeScript type constraints alone cannot always guarantee runtime data integrity.

## Summary

The core change is in the `selectTelemetryAttributes` function. The condition to filter attribute values was updated from a strict check `if (value === undefined)` to a loose equality check `if (value == null)`.

This ensures that attributes with a value of either `null` or `undefined` are now correctly ignored and excluded from the returned `Attributes` object, making the function safer and more predictable.

## Verification

I manually verified this change by setting up a test case that calls the `selectTelemetryAttributes` function with an `attributes` object containing various value types:

```javascript
const attributes = {
  "ai.settings.state": null,
};

// Call selectTelemetryAttributes with the object above
```

1.  **Before Change:** Running the function with the old code (`value === undefined`) resulted in `{ defined: 'value', isNull: null }`. The `isNull` attribute was incorrectly included.
2.  **After Change:** Running the function with the new code (`value == null`) results in `{ defined: 'value' }`. Both `isNull` and `isUndefined` attributes were correctly filtered out, as expected.

This confirms the fix works as intended.

## Tasks

## Future Work

N/A

## Related Issues

N/A

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -84,6 +84,7 @@
     "big-panthers-judge",
     "brave-mails-taste",
     "brave-numbers-drive",
+    "bright-chefs-lick",
     "bright-readers-breathe",
     "bright-turtles-give",
     "brown-eagles-tickle",
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -1,5 +1,12 @@
 # ai
 
+## 5.0.0-beta.16
+
+### Patch Changes
+
+- Updated dependencies [9e16bfd]
+  - @ai-sdk/gateway@1.0.0-beta.6
+
 ## 5.0.0-beta.15
 
 ### Patch Changes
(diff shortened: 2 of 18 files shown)


lgrammel:
hi can you rebase against latest main?

Gaubee:
@lgrammel rebase and reopen PR in #7268

PR #6849, opened 2025-06-24
feat(openai-compatible): passing config to chat models

### LINKED ISSUE #6774: `providerOptions.stream_options` No Longer Effective in Latest SDK, Preventing Usage Data for DeepSeek/Volcengine
### Description

As suggested in issue #4929 (https://github.com/vercel/ai/issues/4929#issuecomment-2753145492), setting the custom parameter `stream_options` via `providerOptions` was the solution to correctly receive `usage` information when calling the DeepSeek API (Alibaba Cloud / Volcengine). This solution worked as recently as March.

However, after the merge of pull request #5873, `providerOptions.stream_options` is no longer effective. Consequently, even when setting `stream_options: { include_usage: true }`, the SDK fails to retrieve the correct `usage` data, which consistently returns as `null`.

**Steps to Reproduce**
1.  Use the latest versions of `ai-sdk` and `@ai-sdk/openai-compatible`.
2.  Pass `stream_options: { include_usage: true }` via `providerOptions`.
3.  Observe that the returned `usage` remains `null`.

**Expected Behavior**

We expect the ability to pass `stream_options` to the underlying request to be restored or provided via a new mechanism (either still within `providerOptions` or another parameter). This is essential to receive the `usage` information correctly.

*   Related Historical Issue: #4929
*   Related PR: #5873

### AI SDK Version

- ai: 4.3.16
- @ai-sdk/openai-compatible: 0.2.14

samurai00:
## Background

Cannot pass the `stream_options` to the OpenAICompatible Providers.

## Summary

- Allow passing config to chat models
- Refine the `stream_options` handling in `OpenAICompatibleChatLanguageModel` to correctly omit the `stream_options` field when `includeUsage` is false/undefined, preventing `undefined` from being sent in the request body.

## Tasks

## Related Issues

Fixes #6774

### DIFF
--- a/.changeset/hot-eggs-obey.md
+++ b/.changeset/hot-eggs-obey.md
@@ -2,4 +2,4 @@
 '@ai-sdk/openai-compatible': patch
 ---
 
-Allow passing config to chat models; Refine the `stream_options` handling in `OpenAICompatibleChatLanguageModel` to correctly omit the `stream_options` field when `includeUsage` is false/undefined, preventing `undefined` from being sent in the request body.
+Allow passing config to chat models
--- a/.changeset/hot-eggs-obey.md
+++ b/.changeset/hot-eggs-obey.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+Allow passing config to chat models; Refine the `stream_options` handling in `OpenAICompatibleChatLanguageModel` to correctly omit the `stream_options` field when `includeUsage` is false/undefined, preventing `undefined` from being sent in the request body.
--- a/packages/openai-compatible/src/openai-compatible-chat-language-model.ts
+++ b/packages/openai-compatible/src/openai-compatible-chat-language-model.ts
@@ -385,9 +385,9 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV1 {
       stream: true,
 
       // only include stream_options when in strict compatibility mode:
-      stream_options: this.config.includeUsage
-        ? { include_usage: true }
-        : undefined,
+      ...(this.config.includeUsage
+        ? { stream_options: { include_usage: true } }
+        : {}),
     };
 
     const metadataExtractor =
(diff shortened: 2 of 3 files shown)


packages/openai-compatible/src/openai-compatible-chat-language-model.ts
```diff
@@ -385,9 +385,9 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV1 {
       stream: true,
 
       // only include stream_options when in strict compatibility mode:
-      stream_options: this.config.includeUsage
-        ? { include_usage: true }
-        : undefined,
```
lgrammel: we use json stringify, which omits undefined values from objects, so how is this any different in terms of the body that is sent?

lgrammel (approved):

lgrammel (approved):

.changeset/hot-eggs-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+Allow passing config to chat models; Refine the `stream_options` handling in `OpenAICompatibleChatLanguageModel` to correctly omit the `stream_options` field when `includeUsage` is false/undefined, preventing `undefined` from being sent in the request body.
```
lgrammel: 2nd part can be removed

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
samurai00: Thanks for the feedback!

With the original code, if includeUsage is false, stream_options is always set to undefined, wiping out any value that providerOptions might have had. The new code only applies the override when includeUsage is true, leaving the original stream_options untouched otherwise.

.changeset/hot-eggs-obey.md
samurai00: So to confirm: should stream_options only be determined by config.includeUsage, and not by what's in providerOptions?

packages/openai-compatible/src/openai-compatible-chat-language-model.ts
lgrammel: i understand that. but the actual body that is sent will be the same because we json stringify later

lgrammel: when would stream_options be included in args?

lgrammel (approved):

.changeset/hot-eggs-obey.md
lgrammel: we can change that in v5 but for v4 this is sufficient imo

PR #6851, opened 2025-06-24
feat (providers/gateway): improve oidc api key client auth flow

joshualipman123:


packages/gateway/src/errors/as-gateway-error.ts
```diff
@@ -2,7 +2,10 @@ import { APICallError } from '@ai-sdk/provider';
 import { extractApiCallResponse, GatewayError } from '.';
 import { createGatewayErrorFromResponse } from './create-gateway-error';
 
-export function asGatewayError(error: unknown) {
+export function asGatewayError(
+  error: unknown,
+  authMethod?: 'api-key' | 'oidc',
```
shaper, line 7: this fn is intended to be error-subtype agnostic, wondering if we can keep it that way somehow, need to look at control flow more closely with you

packages/gateway/src/gateway-language-model.ts
```diff
@@ -71,7 +71,11 @@ export class GatewayLanguageModel implements LanguageModelV2 {
         warnings: [],
       };
     } catch (error) {
-      throw asGatewayError(error);
+      const headers = await resolve(this.config.headers());
+      const authMethod = headers['x-ai-gateway-auth-method'] as
+        | 'api-key'
+        | 'oidc';
+      throw asGatewayError(error, authMethod);
```
shaper: @lgrammel guidance here (more in PR, I will comment) is generally to inline as much as possible for better readability/maintenance of code, so:

```
throw asGatewayError(error, authMethod: (await resolve(this.config.headers()))headers['x-ai-gateway-auth-method']
```

also `as ...` preferred to avoid, it is a forced cast basically, maybe we zod-schema-parse the value to what we know is permitted, you can find examples of zod use for similar in the code base

packages/gateway/src/gateway-language-model.ts
```diff
@@ -128,7 +132,11 @@ export class GatewayLanguageModel implements LanguageModelV2 {
         response: { headers: responseHeaders },
       };
     } catch (error) {
-      throw asGatewayError(error);
+      const headers = await resolve(this.config.headers());
```
shaper: same as above

packages/gateway/src/gateway-provider.ts
```diff
@@ -142,8 +175,17 @@ export function createGatewayProvider(
...
           return metadata;
         })
-        .catch((error: unknown) => {
-          throw asGatewayError(error);
+        .catch(async (error: unknown) => {
+          try {
+            const headers = await getHeaders();
+            const authMethod =
+              'x-ai-gateway-auth-method' in headers
+                ? (headers['x-ai-gateway-auth-method'] as 'api-key' | 'oidc')
```
shaper: Same comment re: avoid hard cast, maybe we need a helper fn to do this via zod validation since we do it in a few places

shaper:
Two mechanical notes:
- prefix on PR description: `feat (providers/gateway): ...`
- add a changeset, run `pnpm changeset` at root of repo, choose gateway, make it a patch level, include PR description as the note

shaper (approved):

PR #6864, opened 2025-06-25
temperature does not default to 0

alexgleason:
I could not find any evidence that the `temperature` value is actually being set to 0 by the ai sdk if not provided by the caller to `generateText`.

The current jsdoc is misleading:

Screenshot from 2025-06-25 14-39-31

If the caller doesn't provide a temperature, no temperature value is passed to the API, and many models (such as claude) default to 1.0 in this case, not 0.

### DIFF
--- a/packages/ai/core/prompt/call-settings.ts
+++ b/packages/ai/core/prompt/call-settings.ts
@@ -9,8 +9,6 @@ Temperature setting. This is a number between 0 (almost no randomness) and
 1 (very random).
 
 It is recommended to set either `temperature` or `topP`, but not both.
-
-@default 0
    */
   temperature?: number;
 


lgrammel:
In AI SDK 5 the default was removed. Your fix is for the AI SDK 4 docs and would be incorrect there.

alexgleason:
@lgrammel So the default actually is 0 in version 4? This is what I'm trying to figure out.

lgrammel:
@alexgleason yes except for a few specific models where it is overridden for compatibility

alexgleason:
Thank you very much for clarifying. I also see it in the code now, in `prepare-call-settings.ts`. I think removing it in v5 is a good move!

Issue #6664, opened 2025-06-07
MIME Type Logic Inconsistency: v5 Branch vs. Main


adolphnov:
### Description

Current Behavior in v5 Branch:
Currently, in the v5 branch, there's an inconsistency in how file Content-Type is handled. In many scenarios, when a file's Content-Type is application/octet-stream (a generic binary type), the v5 branch's logic prioritizes extracting or interpreting this generic type. This often occurs even when a more specific type might be implicitly or explicitly available.

---
- At `main`:
https://github.com/vercel/ai/blob/65e042afde6aad4da9d7a62526ece839eb34f9a5/packages/ai/core/prompt/convert-to-language-model-prompt.ts#L327

- At `v5`:
https://github.com/vercel/ai/blob/2d1132abc7bab45342da908dba2d5265e15b2069/packages/ai/core/prompt/convert-to-language-model-prompt.ts#L284
---
Impact:
As a result, Google gemini frequently fail to correctly identify or render these files.

### AI SDK Version

ai: 5.0.0-alpha.9

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This reports a concrete MIME-type handling failure in the current v5 alpha behavior that causes incorrect provider behavior, so it is a bug.

### Evidence
- AI SDK 5.0.0-alpha.9 prioritizes the generic `application/octet-stream` MIME type even when a more specific type may be available.
- The behavior differs from the main branch's MIME-type handling.
- The reported result is that Google Gemini frequently fails to identify or render affected files.

PR #6900, opened 2025-06-27
fix (ai): use user-provided media type when available

dancer:
## background
fixed media type bug where user-provided specific media types 
were being overwritten by generic downloaded file media types.

## summary
before: mediaType = downloadedFile.mediaType ?? mediaType
- downloaded "application/octet-stream" would override user's "image/jpeg"
- caused Google Gemini and other providers to fail file identification

after: mediaType ??= downloadedFile.mediaType  
- user's "image/jpeg" is preserved, downloaded type used only as fallback
- providers now receive correct, specific media types for proper handling

## Related issues
related to issue #6664

### DIFF
--- a/.changeset/clean-cheetahs-pump.md
+++ b/.changeset/clean-cheetahs-pump.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix (ai): mediatype for providers
--- a/.changeset/clean-cheetahs-pump.md
+++ b/.changeset/clean-cheetahs-pump.md
@@ -2,4 +2,4 @@
 'ai': patch
 ---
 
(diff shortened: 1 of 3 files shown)


lgrammel:
can you add a test case and a changeset?

.changeset/clean-cheetahs-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix (ai): mediatype for providers
```
lgrammel: pls change title and this to 

`fix (ai): use user-provided media type when available`

lgrammel (approved):

PR #6908, opened 2025-06-28
fix (ui/vue): fix `Chat` class reactivity

gorango:
## Background

In the new Vue `Chat` class, `messages` that are passed as props or computed values will break the class' ability to update its internal state.

In Vue's reactivity system `ref.value.<push|pop>(item)` is problematic because it mutates the array directly and Vue won’t detect this change in some cases. Creating a new array allows Vue to correctly track and trigger updates to any reactive dependencies. (Vue tracks assignments (`=`), not mutations (`push`, `pop`, `splice`, etc.) on `.value`.)

## Summary

- Added failing tests in first commit to demonstrate the issue.
- Fixed failing tests in second commit with updates to the core class methods.

## Verification

Added tests.

## Tasks

### DIFF
--- a/examples/nuxt-openai/pages/use-chat-messages/index.vue
+++ b/examples/nuxt-openai/pages/use-chat-messages/index.vue
@@ -1,0 +1,51 @@
+<script setup lang="ts">
+import type { UIMessage } from 'ai';
+import { Chat } from '@ai-sdk/vue';
+import { createIdGenerator } from 'ai';
+import { computed, readonly, ref } from 'vue';
+
+const messages: UIMessage[] = [
+  { id: 'message-0', role: 'user', parts: [{ type: 'text', text: 'Greetings.' }] },
+  { id: 'message-1', role: 'assistant', parts: [{ type: 'text', text: 'Hello.' }] },
+];
+
+const chat = new Chat({
+  generateId: createIdGenerator({ prefix: 'msgc', size: 16 }),
+  // @ts-expect-error - simulate messages passed as props
+  messages: readonly(messages),
+});
+
+const messageList = computed(() => chat.messages); // computed property for type inference
+const input = ref('');
+
+const handleSubmit = (e: Event) => {
+  e.preventDefault();
+  chat.sendMessage({ text: input.value });
+  input.value = '';
+};
+</script>
+
+<template>
+  <div class="flex flex-col w-full max-w-md py-24 mx-auto stretch">
+    <div
+      v-for="message in messageList"
+      :key="message.id"
+      class="whitespace-pre-wrap"
+    >
+      <strong>{{ `${message.role}: ` }}</strong>
+      {{
+        message.parts
+          .map(part => (part.type === 'text' ? part.text : ''))
+          .join('')
+      }}
+    </div>
+
+    <form @submit="handleSubmit">
+      <input
+        class="fixed bottom-0 w-full max-w-md p-2 mb-8 border border-gray-300 rounded shadow-xl"
+        v-model="input"
+        placeholder="Say something..."
+      />
+    </form>
+  </div>
+</template>
--- a/examples/nuxt-openai/pages/use-chat-messages/index.vue
+++ b/examples/nuxt-openai/pages/use-chat-messages/index.vue
@@ -2,17 +2,16 @@
 import type { UIMessage } from 'ai';
 import { Chat } from '@ai-sdk/vue';
 import { createIdGenerator } from 'ai';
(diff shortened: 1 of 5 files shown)


lgrammel:
thanks! can you add a changeset and for verification try this with `examples/nuxt-openai` manually end-to-end?

gorango:
Added a page in the Nuxt dir and can confirm that running without my patch produced errors when sending a new message. Running with the patch solved the issue.

Thanks for the quick review!

packages/vue/src/TestChatInitMessages.vue
```diff
@@ -0,0 +1,34 @@
...
+import { Chat } from './chat.vue';
+import { readonly } from 'vue';
+
+const messages: UIMessage[] = [
+  { id: 'message-0', role: 'user', parts: [{ type: 'text', text: 'Greetings.' }] },
+  { id: 'message-1', role: 'assistant', parts: [{ type: 'text', text: 'Hello.' }] },
+];
+
+const chat = new Chat({
+  // @ts-expect-error - simulate messages passed as props
```
lgrammel: why is this throwing an error? can this be solved by adding a generic to Chat?

examples/nuxt-openai/pages/use-chat-messages/index.vue
```diff
@@ -0,0 +1,51 @@
...
+
+const messages: UIMessage[] = [
+  { id: 'message-0', role: 'user', parts: [{ type: 'text', text: 'Greetings.' }] },
+  { id: 'message-1', role: 'assistant', parts: [{ type: 'text', text: 'Hello.' }] },
+];
+
+const chat = new Chat({
+  generateId: createIdGenerator({ prefix: 'msgc', size: 16 }),
+  // @ts-expect-error - simulate messages passed as props
+  messages: readonly(messages),
```
lgrammel: same - why is this throwing an error

gorango:
I have replaced `readonly` values with `ref`s, which now satisfies the compiler. In practice, users will be passing `props`, `ref`s, or `computed` messages.

To allow explicit `readonly()` values would require some typescript gymnastics, as instantiating the `Chat` with a generic `Readonly<UIMessages>` would not work. The following changes to `chat.vue.ts` would allow passing explicit `readonly()` values but I think it would introduce needless complications to the base class:

```diff
} from 'ai';
import { Ref, ref } from 'vue';

+type DeepReadonlyUIMessage<MESSAGE extends UIMessage> = Omit<
+  Readonly<MESSAGE>,
+  'parts'
+> & {
+  readonly parts: ReadonlyArray<MESSAGE['parts'][number]>;
+};
+
class VueChatState<UI_MESSAGE extends UIMessage>
  implements ChatState<UI_MESSAGE>
{
  private messagesRef: Ref<UI_MESSAGE[]>;
  private statusRef = ref<ChatStatus>('ready');
  private errorRef = ref<Error | undefined>(undefined);

-  constructor(messages?: UI_MESSAGE[]) {
-    this.messagesRef = ref(messages ?? []) as Ref<UI_MESSAGE[]>;
+  constructor(messages?: +ReadonlyArray<DeepReadonlyUIMessage<UI_MESSAGE>>) {
+    const mutableMessages =
+      messages?.map(m => ({ ...m, parts: [...m.parts] }) as UI_MESSAGE) ?? [];
+    this.messagesRef = ref(mutableMessages) as Ref<UI_MESSAGE[]>;
  }

  get messages(): UI_MESSAGE[] {
  
  ...

  snapshot = <T>(value: T): T => value;
}

+export type ChatInit<UI_MESSAGE extends UIMessage> = Omit<
+  BaseChatInit<UI_MESSAGE>,
+  'messages'
+> & {
+  messages?: ReadonlyArray<DeepReadonlyUIMessage<UI_MESSAGE>>;
+};

export class Chat<
  UI_MESSAGE extends UIMessage,
> extends AbstractChat<UI_MESSAGE> {
-  constructor({ messages, ...init }: BaseChatInit<UI_MESSAGE>) {
+  constructor({ messages, ...init }: ChatInit<UI_MESSAGE>) {
    super({
      ...init,
      state: new VueChatState(messages),
    });
  }
}
```

I will push the changes to this file for comparison and review but will be ready to revert if you agree with my conclusions.

lgrammel:
I agree - reverting seems like the right approach. I like your idea of using refs in the unit tests, ideally their setup would match real code to a reasonable degree.

lgrammel (approved):

lgrammel (approved):

PR #6933, opened 2025-06-30
fix (provider/google): removes `additionalProperties` field from schema sent to Google APIs

himanshusinghs:
## Background
Google APIs support only a subset of OpenAPI schema, for both [Function Declarations](https://ai.google.dev/api/caching#FunctionDeclaration) and [Structured Outputs](https://ai.google.dev/gemini-api/docs/structured-output#json-schemas).

And `additionalProperties` is not one of them. Passing this property in the request leads to 400 BAD REQUEST error from Google APIs, complaining specifically about the invalid parameter: `additionalProperties`.

In the following PRs, [for v4](https://github.com/vercel/ai/pull/5626) and [for v5](https://github.com/vercel/ai/pull/6919), there was an important fix to identify properties incorrectly getting converted to undefined because they did not have `properties` field but we those PRs introduced `additionalProperties` in the converted schema.

## Summary
This PR keeps the change in `isEmptyObjectSchema` for identifying the objects with empty or no `properties` field and reverts the introduction of `additionalProperties`.

## Verification
Verified manually that the change helps in mitigating the 400 BAD REQUEST from Google APIs.

## Related Issues
Partially reverts #5626

### DIFF
--- a/packages/google/src/convert-json-schema-to-openapi-schema.test.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.test.ts
@@ -23,48 +23,28 @@ it('should remove additionalProperties and $schema', () => {
   expect(convertJSONSchemaToOpenAPISchema(input)).toEqual(expected);
 });
 
-it('supports additionalProperties: true', () => {
+it('should remove empty additionalProperties object from nested object schemas', function () {
   const input: JSONSchema7 = {
-    $schema: 'http://json-schema.org/draft-07/schema#',
     type: 'object',
     properties: {
-      name: { type: 'string' },
-      age: { type: 'number' },
+      keys: {
+        type: 'object',
+        additionalProperties: { type: 'string' },
+        description: 'Description for the key',
+      },
     },
-    additionalProperties: true,
+    additionalProperties: false,
+    $schema: 'http://json-schema.org/draft-07/schema#',
   };
 
   const expected = {
     type: 'object',
     properties: {
-      name: { type: 'string' },
-      age: { type: 'number' },
+      keys: {
+        type: 'object',
+        description: 'Description for the key',
+      },
     },
-    additionalProperties: true,
-  };
-
-  expect(convertJSONSchemaToOpenAPISchema(input)).toEqual(expected);
-});
-
-// z.record(z.string())
-it('supports `additionalProperties` as an object', () => {
-  const input: JSONSchema7 = {
-    $schema: 'http://json-schema.org/draft-07/schema#',
-    type: 'object',
-    properties: {
-      name: { type: 'string' },
-      age: { type: 'number' },
-    },
-    additionalProperties: { type: 'string' },
-  };
-
-  const expected = {
-    type: 'object',
-    properties: {
-      name: { type: 'string' },
-      age: { type: 'number' },
-    },
-    additionalProperties: { type: 'string' },
   };
 
   expect(convertJSONSchemaToOpenAPISchema(input)).toEqual(expected);
--- a/packages/google/src/convert-json-schema-to-openapi-schema.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.ts
@@ -28,7 +28,6 @@ export function convertJSONSchemaToOpenAPISchema(
     const: constValue,
     minLength,
     enum: enumValues,
-    additionalProperties,
   } = jsonSchema;
 
   const result: Record<string, unknown> = {};
@@ -114,13 +113,6 @@ export function convertJSONSchemaToOpenAPISchema(
 
   if (minLength !== undefined) {
     result.minLength = minLength;
-  }
-
-  if (additionalProperties === true) {
-    result.additionalProperties = true;
-  } else if (additionalProperties) {
-    result.additionalProperties =
-      convertJSONSchemaToOpenAPISchema(additionalProperties);
   }
 
   return result;
--- a/.changeset/flat-taxis-cheer.md
+++ b/.changeset/flat-taxis-cheer.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+removes (unsupported) `additionalProperties` from the Schema sent in the request payloads to Google APIs


lgrammel:
please open against v5 or v4 branches. we have stopped using main.

how is this related to the fix that was landed earlier today?

himanshusinghs:
Hey @lgrammel! I might have mentioned the problem in the previously closed PR [here](https://github.com/vercel/ai/pull/5626#issuecomment-3019062084).

But in short, Google APIs only support a subset of OpenAPI schema and `additionalProperties` is not mentioned as one of the supported ones, ref: https://ai.google.dev/api/caching#FunctionDeclaration

The proper handling for `additionalProperties` should be to remove them and not add them to the converted schema, similar to what langchain does in their [implementation](https://github.com/langchain-ai/langchainjs/blob/451f4568d6f66d8feeeb79aa414bb41a8b913373/libs/langchain-google-genai/src/utils/zod_to_genai_parameters.ts#L26-L57).

This PR builds on top of @tjenkinson's PR where he made the change to recognise objects with `additionalProperties` as non-empty object. That is crucial because otherwise you incorrectly remove them from the converted schema.

himanshusinghs:
Happy to rebase this for both v4 and v5 once we agree on the path forward here.

packages/google/src/convert-json-schema-to-openapi-schema.test.ts
```diff
@@ -23,6 +23,51 @@ it('should remove additionalProperties and $schema', () => {
...
+      type: 'object',
+      properties: {
+        name: { type: 'string' },
+        age: { type: 'number' },
+      },
+    };
+
+    expect(convertJSONSchemaToOpenAPISchema(input)).toEqual(expected);
+  },
+);
```
lgrammel: please convert into 3 simple tests.

packages/google/src/convert-json-schema-to-openapi-schema.ts
```diff
@@ -115,6 +116,14 @@ export function convertJSONSchemaToOpenAPISchema(
     result.minLength = minLength;
   }
 
+  if (additionalProperties) {
+    // Google Function declarations do not support additionalProperties so we
+    // should avoid adding them entirely.
+    // Ref: https://ai.google.dev/api/caching#FunctionDeclaration
+    // result.additionalProperties =
+    // convertJSONSchemaToOpenAPISchema(additionalProperties);
+  }
```
lgrammel: this does not do anything?

.changeset/cold-schools-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Support tool schemas that allow additional properties (e.g `z.record(z.string())`)
```
lgrammel: replace tool schemas with json schemas

this affects structured outputs as well

lgrammel:
please edit pr to base on v5

lgrammel:
can you add some of that explanation to background?

packages/google/src/convert-json-schema-to-openapi-schema.ts
himanshusinghs: Yea its intentional that it does nothing. I left the `if` to make it more explicit that we are intentionally removing the `additionalProperties` because Google APIs don't expect this property to be there. I am fine removing it as well as it is more for documentation than anything.

packages/google/src/convert-json-schema-to-openapi-schema.test.ts
himanshusinghs: sounds good - will do that

.changeset/cold-schools-sip.md
himanshusinghs: yea I noticed it just now - I will confirm first if its only the functionDeclaration that does not support `additionalProperties` or more.

himanshusinghs:
I will update the PR in a moment, thank you for the quick feedback :)

packages/google/src/convert-json-schema-to-openapi-schema.test.ts
himanshusinghs: I have removed the unnecessary tests.

packages/google/src/convert-json-schema-to-openapi-schema.ts
himanshusinghs: Got rid of the entire block.

.changeset/cold-schools-sip.md
himanshusinghs: I can confirm that the StructuredOutputs also follow the same limitation, referenced [here](https://ai.google.dev/gemini-api/docs/structured-output#json-schemas). So it should be safe to pull this change for both of them (Structured Outputs and Function Declarations).

himanshusinghs:
@lgrammel I have based this PR on v4, will raise the same patch for v5 once this is fine to go. Have also added some background for clarification.

lgrammel (approved):

lgrammel:
@himanshusinghs does this effectively revert #6919 ? if so, is there more code that should be removed (additionProperties arg)? also what about the original bug that triggered #6919 ?

himanshusinghs:
The original problem was that schemas built using `z.record()` were not getting recognised because the JSONSchema7 representation for this zod schema is:
```
{
  type: 'object',
  properties: {},
  additionalProperties: {} // what's inside depends on how z.record was called. For example:
   // when using z.record(z.string(), z.string()), the additionalProperties will be { type: "string" }
}
```

See that properties is an empty object? That is why the entire schema was not considered a candidate for conversion and returned as `undefined`. The fix for this is the method `isEmptyObjectSchema`, [these lines](https://github.com/vercel/ai/pull/6919/files#diff-db99b750a11626b268f6fea2df8c2f94414eaf3adbe3cfa358832b7f18d574a4R135-R136) particularly. Its only the rest of the changes in method `convertJSONSchemaToOpenAPISchema` that we are reverting.

lgrammel:
Awesome - would be great to have this fixed in v5 as well. Thank you for looking into the issue!

himanshusinghs:
Here is the PR for v5 patch https://github.com/vercel/ai/pull/6958

PR #6951, opened 2025-07-01
feat: add image editing functionality to OpenAI provider

### LINKED ISSUE #5949: Support image editing (openai/luma)
### Description

OpenAI has an "edit" image which allows to send multiple images to either refine, change, improve your image.
https://platform.openai.com/docs/api-reference/images/createEdit
https://platform.openai.com/docs/guides/image-generation?image-generation-model=gpt-image-1#edit-images

Luma also has this ability 
https://docs.lumalabs.ai/docs/image-generation#character-reference

It would be nice to have such options in the `generateImage` function

FranciscoMoretti:
## Background

This PR adds images as input for image generation (image editing).

Some providers support image editing capabilities to refine, change, or improve existing images. This requires sending multiple images as input. This functionality was missing from the AI SDK's `generateImage` function, limiting users to only image generation from text prompts.

## Summary

- Introduced image editing functionality to the `generateImage` function in the OpenAI provider
- Added support for single and multiple image edits
- OpenaAI: Enhanced the `doGenerate` method to intelligently route requests to the appropriate OpenAI endpoint (`/images/generations` vs `/images/edits`) based on the presence of input images
- Added support for optional mask images to specify areas for targeted editing
- Expanded test coverage to include image editing scenarios and proper request handling validation
- Added some examples `examples/ai-core/src/generate-image/openai-edit.ts`

## Verification

[x] Manually tested image editing with single images using both Uint8Array and base64 formats
[x] Verified multiple image editing functionality works correctly
[x] Tested mask image functionality for targeted area editing
[x] Confirmed proper endpoint routing (generation vs editing) based on input parameters
[x] Tested backward compatibility - existing `generateImage` calls continue to work as expected

## Tasks

## Future Work

- Add image editing support to other providers.

## Notes

- I couldn't get dall-e-2 to work even when following [OpenAI API reference](https://platform.openai.com/docs/api-reference/images/createEdit) with their example curl command. It doesn't respond with an error, but it returns the original input image without making edits.

- Not sure where to add this documentation

## Related Issues

Fixes [#5949](https://github.com/vercel/ai/issues/5949)


lgrammel:
Hi! can you rebase against main (previously v5)?

FranciscoMoretti:
> Hi! can you rebase against main (previously v5)?

@lgrammel rebased and passed build. I might want to run one final round of testing before merging. Does it look good overall?

lgrammel:
@FranciscoMoretti tests and prettier currently failing

FranciscoMoretti:
@lgrammel now passing tests and prettier check locally 👍

lgrammel:
My biggest concern is how to model different image editing operations going forward, and whether to have a single generate or more specific functions.

E.g. typical operations are to edit images, enlarge images, generate image variants. All of these would fall under edit. While this may not be relevant for the openai image api, there are many other image APIs that we will need to support.

What are your thoughts on this? how would this work with other image generation APIs?

FranciscoMoretti:
> My biggest concern is how to model different image editing operations going forward, and whether to have a single generate or more specific functions.

Thanks for the review @lgrammel. I appreciate you taking the time to look at this even during the v5 Beta. 

You ask a great question. I'm not too attached to this implementation, but here are my thoughts:


### Why as part of Generate Image

Fun fact, on the first iteration, instead of `generateImage`, I created an `editImage` function + a  `doEdit` method in the image model. However, I changed to the current implementation for the following reasons:

- Realized that "edit image" is just "images used as input" (together with a prompt). This is not too different from how the "generateText" can use images with a prompt.... Just different outputs.
- I liked the simplicity and consistency of `generateText`, `generateImage`, `generateSpeech`

### Compatibility of other providers

I agree that integrating functionality from many providers is a challenge. Not sure how to solve this. 

Let's consider a small example from Luma ["Generate an Image"](https://docs.lumalabs.ai/reference/generateimage):

The Generate Image endpoint has the parameters: "image_ref", "style_ref", "character_ref", and "modify_image_ref".

The question (IMHO) is not if these are image generations or image edits. Rather, how do we present parameters that are useful for some providers but not for others? 

I can think of the following options:
1. No validation: We use optional properties, and we let the API respond with an error when they are not allowed
2. Validate on the model: We validate them inside each image model and respond with a nicer error (at more development cost)
3. Typesafe params: we change some of the Call Options based on the model selected (like a discriminated union). 

I like option 3 and find it similar to the `ProviderOptions` approach. However, maybe 1 or 2 are better in the short term

amarant:
it would be great to have this functionality merged, that's a really important feature missing now.

ross0x01:
+1

xiaoiver:
> how do we present parameters that are useful for some providers but not for others?

I think we can use `providerOptions` like this: https://github.com/vercel/ai/pull/7470

```ts
const result = await generateSpeech({
  model: minimax.speech('speech-02-hd'),
  text: 'Hello from the AI SDK!',
  providerOptions: { // here
    minimax: {
      audio_setting: {
        format: 'mp3',
      },
    },
  },
});
```

DeJeune:
any progress?

gr2m:
We started to look more thoroughly into image editing for the upcoming v5.1 release, see https://github.com/vercel/ai/issues/8719

aayush-kapoor:

PR #6982, opened 2025-07-02
pkg(prettier): fixed version

dancer:
## background

updated prettier version because team inconsistencies in code formatting due different prettier versions

## summary

- fix prettier and prettier-plugin-svelte to specific versions (3.5.3 and 3.2.7)

## verification

- prettier version now consistent across all environments
- formatting behavior standardized for entire team

## tasks

- [x] update prettier from ^3.3.3 to 3.5.3
- [x] update prettier-plugin-svelte from ^3.3.3 to 3.2.7


package.json
```diff
@@ -37,8 +37,8 @@
     "lint-staged": "15.2.10",
     "next": "15.0.0-canary.23",
     "playwright": "^1.44.1",
-    "prettier": "^3.3.3",
-    "prettier-plugin-svelte": "^3.3.3",
+    "prettier": "3.5.3",
+    "prettier-plugin-svelte": "3.2.7",
```
lgrammel: this is a downgrade - why?

dancer: i am dumb i checked on google if 3.5.3 is compatible with svelte 3.3.3 and it said no don't trust google just checked it is compatible will fix

lgrammel (approved):

PR #7003, opened 2025-07-02
fix (providers/anthropic): remove fine grained tool streaming beta

http-samc:
## Background

Certain tool input schemas were causing JSON parsing errors for the tool input returned by the Anthropic provider when streaming. After further investigation, the [fine-grained-tool-streaming-2025-05-14](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/fine-grained-tool-streaming) beta, which is enabled by default in the language model and currently not guaranteed to return valid JSON, was causing the error.

## Summary

To reduce minimize added complexity, the beta header was removed entirely. Users who wish to use the beta may do so by passing in custom headers in the streamText call.

## Verification

~There's a minimal [example script](https://github.com/vercel/ai/blob/68445460b46fb28df780efe9f1d9d905da4b7ced/examples/ai-core/src/stream-text/anthropic-fine-grained-tool-call.ts) that breaks when Fine Grained Tool Streaming is enabled but works when its disabled.~

~This verifies the issue and tests the updated API options.~

**Update: I've removed the driver script above from the final PR after we decided to move from an opt-in for the flag to just not including it. It doesn't make sense to have a demo for something that isn't a direct option in the final provider**

## Tasks

- [ ] ~Documentation has been added / updated (for bug fixes / features)~ `N/A`


lgrammel:
If this is confirmed to cause issues, all that is needed is to remove the beta header, since users can already add headers and all this logic is unnecessary complexity.

lgrammel (approved):

PR #6929, opened 2025-06-30
feat(providers/openai): Add full reasoning support to OpenAI responses API

jonaslalin:
## Background

This PR adds comprehensive support for OpenAI's Responses API with full reasoning capabilities. The OpenAI Responses API is a new API that provides access to OpenAI's reasoning models (o1, o3, o3-mini, o4-mini) with enhanced reasoning summaries, encrypted content handling, and zero data retention options.

## Summary

This PR implements full reasoning support for the OpenAI Responses API, including:

- **Async message conversion**: Updated `convertToOpenAIResponsesMessages` to handle asynchronous reasoning message processing with provider options parsing
- **Reasoning message handling**: Added comprehensive support for reasoning parts with ID generation, encrypted content handling, and summary merging
- **Streaming support**: Full streaming implementation for reasoning summaries with encrypted content support
- **Enhanced API types**: Added `OpenAIResponsesReasoning` type with encrypted content and summary structures
- **Provider options**: Added support for `reasoningSummary`, `include`, and reasoning effort configuration
- **Zero data retention**: Support for stateless interactions with encrypted content retrieval
- **Comprehensive examples**: Added 4 new examples demonstrating different reasoning use cases
- **Robust error handling**: Added validation and warnings for malformed reasoning parts and mismatched encrypted content

### Key Changes:
- Made `convertToOpenAIResponsesMessages` async to support provider options parsing
- Added reasoning message merging logic for consecutive parts with the same ID
- Enhanced streaming to include reasoning-specific events and metadata
- Updated usage token handling to include reasoning tokens in the standard usage object
- Added comprehensive test coverage for all reasoning scenarios

### Manual Testing Performed:
1. **Basic reasoning generation**: Tested with o3-mini model generating reasoning summaries
2. **Streaming reasoning**: Verified real-time streaming of reasoning content with proper start/end events
3. **Encrypted content handling**: Tested zero data retention scenarios with encrypted content retrieval
4. **Message conversion**: Verified proper conversion of reasoning parts to OpenAI API format
5. **Error scenarios**: Tested malformed provider options and mismatched encrypted content handling
6. **Multi-step reasoning**: Tested complex reasoning flows with tool interactions
7. **All 4 new examples**: Ran and verified all example scripts work correctly with the API

### Test Coverage:
- Added 15+ new test cases covering reasoning message conversion
- Added streaming tests for reasoning with encrypted content
- Added comprehensive error handling tests
- All existing tests continue to pass

## Tasks

[...]

PR #7014, opened 2025-07-03
feat (provider/openai): add support for encrypted_reasoning to responses api

lgrammel:
## Background

Encrypted reasoning information is required for multi-turn interactions with the OpenAI responses API that include reasoning.

## Summary

* send openai reasoning information (id, encrypted reasoning) in reasoning parts
* reconstruct multi-summary openai reasoning messages from multiple reasoning parts
* support `include` provider option

## Verification

* run streamText and generateText zero data retention examples

## Tasks

- [x] test generate case
- [x] test stream case
- [x] fix streaming with encrypted content
- [x] refactoring

## Related Issues

Continues #6929


jonaslalin:
@lgrammel this part is important, because the reasoningIds from OpenAI will be the same for all the reasoning summary parts here, so it should not be combined into one reasoning part, but several, and they have different encrypted content. If we cannot reproduce the exact reply from OpenAI when we do multi turn conversations, we will not trigger the cached prompt feature.
```
if (part.type !== 'reasoning') {
            // Reset reasoning state when encountering non-reasoning content.
            // This ensures that reasoning parts separated by non-reasoning content are treated as separate reasoning sequences.
            // This supports alternating patterns like:
            // reasoning → reasoning → tool-call → tool-result → reasoning → reasoning → reasoning → tool-call → tool-result → ... → text.
            reasoningMessagesInScope = {};
          }
```

lgrammel:
> @lgrammel this part is important, because the reasoningIds from OpenAI will be the same for all the reasoning summary parts here, so it should not be combined into one reasoning part, but several, and they have different encrypted content. If we cannot reproduce the exact reply from OpenAI when we do multi turn conversations, we will not trigger the cached prompt feature.

do you have a reproducible live example where this happens and the openai reasoning ids are the same?

the scope has been changed to within a single assistant message, and my assumption is that different reasoning parts with separate summaries within the same assistant message would have different reasoning ids.

otherwise we can just key on the encrypted content

lgrammel:
Per the API docs the identifier is unique, so I am under the assumption this is not an issue. We can follow up with additional changes if there turn out to be problems in practice.

image

jonaslalin:
It looks like it is unique as you say. I double checked with a deep research model now:
image
Here we see the id change.

jonaslalin:
Should we then update the tests where I reused the same id, but it should in fact be different?

lgrammel:
@jonaslalin yes that is a good idea. i've removed & moved around the tests. i will land this pr and then you can follow up with a pr that adds more test cases / updates existing test cases.

Issue #6964, opened 2025-07-01
Support correct and lossless reconstruction of OpenAI messages when streaming for multi-turn conversation (needed for deep research etc.)


jonaslalin:
### Description

**Streaming Response Message Reconstruction Issue:** 

The current streaming implementation cannot correctly reconstruct the `(await result.response).messages` array from streaming deltas, which prevents proper multi-turn conversation continuity. Specific issues include:

- **Complex Reasoning Structure**: Reasoning items can contain arrays of summary texts that cannot be properly reconstructed from individual deltas
- **Missing Output Index**: The `output_index` needed to correctly place chunk deltas in the final array structure is not exposed to consumers
- **Incomplete State Building**: The streaming parser does not correctly build the final message state from deltas, particularly for complex reasoning content
- **Multi-turn Conversation Breakage**: For deep research models and similar advanced reasoning models, users cannot continue conversations with the correct message prefix without implementing custom parsing logic

This means that for streaming scenarios with reasoning models, developers currently need to either:
1. Use non-streaming mode for multi-turn conversations, or  
2. Implement custom delta parsing logic (which lacks access to necessary metadata like `output_index`)

See real output from responses API to highlight this issue:

```
curl https://api.openai.com/v1/responses \
  -H "Authorization: Bearer $OPENAI_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "o3-deep-research",
    "input": "Research the economic impact of semaglutide on global healthcare systems. Include specific figures, trends, statistics, and measurable outcomes. Prioritize reliable, up-to-date sources: peer-reviewed research, health organizations (e.g., WHO, CDC), regulatory agencies, or pharmaceutical earnings reports. Include inline citations and return all source metadata. Be analytical, avoid generalities, and ensure that each section supports data-backed reasoning that could inform healthcare policy or financial modeling.",
    "tools": [
      { "type": "web_search_preview" }
    ],
    "reasoning": {
      "effort": "medium", 
      "summary": "auto"
    },
    "include": ["reasoning.encrypted_content"],
    "stream": true
  }'
```

jonaslalin:
[...]

PR #7029, opened 2025-07-03
feat(provider/openai): fix streaming of reasoning events

jonaslalin:
## Background

OpenAI's reasoning models (like o3-mini) introduced a new streaming event structure for reasoning parts that wasn't properly handled in the existing implementation. The previous code was designed around the older event format and didn't correctly process the new `response.reasoning_summary_part.added` and `response.reasoning_summary_part.done` events, causing reasoning deltas to be dropped or improperly streamed.

## Summary

Fixed streaming of reasoning parts in `OpenAIResponsesLanguageModel` by:

- **Enhanced event handling**: Added support for new OpenAI streaming events (`response.reasoning_summary_part.added` and `response.reasoning_summary_part.done`)
- **Improved state management**: Introduced `activeReasoning` state tracking to properly manage reasoning blocks throughout their lifecycle
- **Fixed reasoning part lifecycle**: Now correctly emits `reasoning-start`, `reasoning-delta`, and `reasoning-end` events at the appropriate times
- **Enhanced schema validation**: Updated Zod schemas to handle the new event structure
- **Added comprehensive tests**: Implemented tests for multiple reasoning blocks and various edge cases

The fix ensures that reasoning summaries are properly streamed in real-time as they're generated, maintaining the expected event flow for consumers of the streaming API.

## Verification

Manually tested the fix with:
1. **Single reasoning block streaming**: Verified that reasoning deltas are properly streamed and events are emitted in the correct order
2. **Multiple reasoning blocks**: Confirmed that multiple reasoning blocks within a single response are handled correctly
3. **Edge cases**: Tested reasoning blocks with and without encrypted content, empty summaries, and mixed content scenarios
4. **Backward compatibility**: Verified that existing non-reasoning streaming functionality remains unaffected

The changes have been validated against the actual OpenAI API response format and ensure proper handling of all reasoning-related streaming events.

## Tasks

## Related Issues

Fixes streaming issues with OpenAI's reasoning models where reasoning deltas were not properly emitted during streaming responses #6964.


jonaslalin:
@lgrammel I am not completely done yet. Will add tests and extra polishing. What do you think about this solution? It will allow providers such as OpenAI to split reasoning, without polluting streamText with provider specific details, since I inject a strategy function instead. I am a little bit unsure about the best naming of that function and the variables. If you have any feedback, please feel free to share.

lgrammel:
What does this fix?

Please keep in mind that we cannot change the spec for the next 4-6 months.

packages/provider/src/language-model/v2/language-model-v2.ts
```diff
@@ -134,4 +134,20 @@ Response headers.
...
+
+@param options - The options for determining the reasoning part index.
+@param options.id - The ID of the reasoning part.
+@param options.providerMetadata - Optional provider metadata that may contain indexing information.
+@returns The array index where this reasoning part should be stored (0-based).
+   */
+  getReasoningPartIndex?: (options: {
+    id: string;
+    providerMetadata?: SharedV2ProviderMetadata;
+  }) => number;
```
lgrammel: the spec must not contain such functions to support providers such as gateway

jonaslalin:
> What does this fix?
> 
> Please keep in mind that we cannot change the spec for the next 4-6 months.

We cannot construct the correct reasoning messages for streaming cases with the openai provider right now. I don't know how we could fix this without appending to the spec. They are incorrectly combined into one reasoning part, whereas they should be several parts. If we don't fix this we will not get prompt caching to work, and we cannot really use openai in streaming mode for multi-turn conversations with reasoning models.

For the following stream, we incorrectly merge the different summaries into one, where they should be 2, if we look at the summary index property from openai. With my feature, we will split this so we can recombine into correct state and send again for multi turn conversations.

@lgrammel Do you have any idea how we could apply my fix in a different way without changing the spec?

```
{
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "encryptedContent": "encrypted_reasoning_data_abc123",
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                    },
                  },
                },
                "type": "reasoning-start",
              },
              {
                "delta": "**Exploring burrito origins**

            The user is",
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                      "summary_index": 0,
                    },
                  },
                },
                "type": "reasoning-delta",
              },
              {
                "delta": " curious about the debate regarding Taqueria La Cumbre and El Farolito.",
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                      "summary_index": 0,
                    },
                  },
                },
                "type": "reasoning-delta",
              },
              {
                "delta": "**Investigating burrito origins**

            There's a fascinating debate",
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                      "summary_index": 1,
                    },
                  },
                },
                "type": "reasoning-delta",
              },
              {
                "delta": " about who created the Mission burrito.",
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                      "summary_index": 1,
                    },
                  },
                },
                "type": "reasoning-delta",
              },
              {
                "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                "providerMetadata": {
                  "openai": {
                    "reasoning": {
                      "encryptedContent": "encrypted_reasoning_data_abc123",
                      "id": "rs_6808709f6fcc8191ad2e2fdd784017b3",
                    },
                  },
                },
                "type": "reasoning-end",
              },
             ```

File: packages/provider/src/language-model/v2/language-model-v2.ts (hunk 1 above)
Comment (jonaslalin) [reply packages/provider/src/language-model/v2/language-model-v2.ts]: It is optional, not required, with a default implementation inside streamText that maintains the current behaviour, so if you don't add something here, you will never see a difference.

**Comment by lgrammel** [comment]
is it just the summary index property that is missing?

**Comment by lgrammel** [comment]
or is this more of a bug when we send the reasoning parts (we need to aggregate based on summary index)?

**Comment by lgrammel** [comment]
Most important would be an example that reproduces the error. Then I can take a look next week when I'm back.

**Comment by jonaslalin** [comment]
I was thinking about this when going to sleep last night. I think I might be able to fix this inside the OpenAI provider instead. I will at least give it a shot, but yes this is a bug. When we reconstruct messages from the stream, which will be accessible by `.messages` after stream completion, the reasoning objects have been constructed in the wrong way in order to be able to reproduce the original response from OpenAI, which we will pass to `streamText` or `generateText` again to continue the conversation. Stay tuned. I will keep you updated if I am able to fix this inside the provider instead.

**Comment by jonaslalin** [comment]
Yes! With my fixes, we finally collect reasoning for the streaming case correctly too:
```
=== First request ===
[
  {
    "type": "reasoning",
    "text": "**Analyzing encrypted data**\n\nThe user is asking about the string \"U2VjcmV0UGFzc3dvcmQxMjM=\". This appears to be Base64 encoded data. When I decode it, I find it translates to \"SecretPassword123\". So, it's not encryption—just an easily reversible encoding. The secret it contains is indeed \"SecretPassword123\". To wrap up, I can say: this data is not encrypted but Base64 encoded, revealing the secret when decoded.",
    "providerMetadata": {
      "openai": {
        "reasoning": {
          "id": "rs_6867dd4f13788192b5d33d19919c545a0827e77773be1a41",
          "encryptedContent": "gAAAAABoZ91PiF1oJ3RFtG6P_X10HIZcN9INzrO_jU3NTfFOOXOPHetekrAPeZMC1Xj0sNd0bqfMEF1UXoIMokNeG2chuuEFL7pj4TvVWI9jYGWw4WGwcogO0r7O5OKvkqYJmzCe6-X_JTvNgqySw4mrPGYg4X6Is_mgFp0lz_-gP-HUxcSMA4SJt5mm9usOafB24hDspo6l1QeYeC4sMm650prPepvjpeWbV80Of5AwBaykza3tGEImcwciUxUa7NgzjjCFR9Cdg09qB2R1W8oUR7AD89ir93qdf8Pgerb7apvdwg81bl6NtmfjGCZrHrJg9hI0Nc2FBX5rqfB3aNSffJ-ZrYBUhg6rsKSPLfm1lbZT7jX3aO8QZ8IlyZJgDHHtDkWfliR_Z5BeM-3eFEnTsr_3kRlSzED0tkajlohT9yuMkqBU038="
        }
      }
    }
  },
  {
    "type": "reasoning",
    "text": "**Decoding the string**\n\nAfter I decode the string, I find that it reveals the plaintext message, which is \"SecretPassword123.\" Now, I can provide this answer clearly, ensuring that it conveys the information accurately. It's straightforward and directly addresses what was encoded. So, my main point is that this decoded message is now ready to be shared with the user.",
    "providerMetadata": {
      "openai": {
        "reasoning": {
          "id": "rs_6867dd4f13788192b5d33d19919c545a0827e77773be1a41",
          "encryptedContent": "gAAAAABoZ91PiF1oJ3RFtG6P_X10HIZcN9INzrO_jU3NTfFOOXOPHetekrAPeZMC1Xj0sNd0bqfMEF1UXoIMokNeG2chuuEFL7pj4TvVWI9jYGWw4WGwcogO0r7O5OKvkqYJmzCe6-X_JTvNgqySw4mrPGYg4X6Is_mgFp0lz_-gP-HUxcSMA4SJt5mm9usOafB24hDspo6l1QeYeC4sMm650prPepvjpeWbV80Of5AwBaykza3tGEImcwciUxUa7NgzjjCFR9Cdg09qB2R1W8oUR7AD89ir93qdf8Pgerb7apvdwg81bl6NtmfjGCZrHrJg9hI0Nc2FBX5rqfB3aNSffJ-ZrYBUhg6rsKSPLfm1lbZT7jX3aO8QZ8IlyZJgDHHtDkWfliR_Z5BeM-3eFEnTsr_3kRlSzED0tkajlohT9yuMkqBU038="
        }
      }
    }
  }
]
The string is encoded using Base64. When you decode U2VjcmV0UGFzc3dvcmQxMjM=, you get the plaintext "SecretPassword123".

Request body: {
  "model": "o3-mini",
  "input": [
    {
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "Analyze the following encrypted data: U2VjcmV0UGFzc3dvcmQxMjM=. What type of encryption is this and what secret does it contain?"
        }
      ]
    }
  ],
  "store": false,
  "include": [
    "reasoning.encrypted_content"
  ],
  "reasoning": {
    "effort": "medium",
    "summary": "auto"
  }
}
=== Second request ===
[
  {
    "type": "reasoning",
    "text": "**Expanding on security recommendations**\n\nI see the user asking about security recommendations based on previous encrypted data discussions. It's essential to highlight key points like avoiding base64 encoding for secrets and opting for strong salted hashing methods when storing any passwords. I'll stress the importance of using proper encryption instead of encoding and following secure password practices to prevent plaintext leaks. Essentially, passwords should never be stored as plaintext; strong algorithms like bcrypt or Argon2 should be used instead for security.",
    "providerMetadata": {
      "openai": {
        "reasoning": {
          "id": "rs_6867dd58417c8192a0dd22d3cb19d6710827e77773be1a41",
          "encryptedContent": "gAAAAABoZ91YBml6qoxvpMl_pTDmIoSKbhPgLpl4lBjy1H6vn-E4UDlQEpJ4R1hxqy0U8qq1NaEYorXEeSh3-1MmiGHQZ9DudzjkqRTJ0BaexxR7NJF0W_k795_-50bF1pKDksKV_mhDPQAf1rEID9xxsXIo8yB0ZdAnH4oHrxpL-M8ekwn-fsleYrgb7exGQvphgr1nl5FojGCqMeUyOtH6jbdJUxeiZ9yCQZnVr1SpeidcWYOwGloYvbhT4yo8xsVQ7f7kIKgOjOwZZuwy8pDeV0Bcznfny6gyGvpX1wG83f_MQAzs5r6N8-IAWpNHYQHi4FcMhNgi769LbwIrmiLzcIEAKzFlf8z7b05hkpJ6eUsxJM3auIJErJyMHq3ZNGxneGvk-Uqfib7zflCAeiavtguolgZ_mhtx8giSQAmvCav9V8HT70s="
        }
      }
    }
  },
  {
    "type": "reasoning",
    "text": "**Strengthening security measures**\n\nIt's crucial to emphasize that encoding is not enough for protecting sensitive information. Instead, using strong encryption methods is necessary, along with proper salting and hashing when storing passwords. Additionally, passwords should never be kept in plain text to mitigate vulnerabilities. For transmitting secrets, ensuring the use of secure channels is vital. Implementing multi-factor authentication can further enhance security. I'll compile these points into a final analysis to convey comprehensive recommendations.",
    "providerMetadata": {
      "openai": {
        "reasoning": {
          "id": "rs_6867dd58417c8192a0dd22d3cb19d6710827e77773be1a41",
          "encryptedContent": "gAAAAABoZ91YBml6qoxvpMl_pTDmIoSKbhPgLpl4lBjy1H6vn-E4UDlQEpJ4R1hxqy0U8qq1NaEYorXEeSh3-1MmiGHQZ9DudzjkqRTJ0BaexxR7NJF0W_k795_-50bF1pKDksKV_mhDPQAf1rEID9xxsXIo8yB0ZdAnH4oHrxpL-M8ekwn-fsleYrgb7exGQvphgr1nl5FojGCqMeUyOtH6jbdJUxeiZ9yCQZnVr1SpeidcWYOwGloYvbhT4yo8xsVQ7f7kIKgOjOwZZuwy8pDeV0Bcznfny6gyGvpX1wG83f_MQAzs5r6N8-IAWpNHYQHi4FcMhNgi769LbwIrmiLzcIEAKzFlf8z7b05hkpJ6eUsxJM3auIJErJyMHq3ZNGxneGvk-Uqfib7zflCAeiavtguolgZ_mhtx8giSQAmvCav9V8HT70s="
        }
      }
    }
  }
]
The analyzed string wasn't encrypted—it was simply encoded with Base64, which means it's trivial to decode. Given that it contained a sensitive string ("SecretPassword123"), here are some security recommendations:

1. Do not store or transmit passwords or other secrets in plaintext (even if encoded). Instead, use proper cryptographic measures.

2. If you must store passwords, use a strong, salted hash function (e.g., bcrypt, Argon2, or scrypt) rather than reversible encoding. This helps protect user credentials in the event of a data breach.

3. Always use secure encryption if you need confidentiality during data transmission or persistent storage. Rely on established protocols (like TLS for data in transit) and robust encryption algorithms for data at rest.

4. Keep secrets out of code repositories and configuration files. Use environment variables or dedicated secret management solutions (e.g., Vault or AWS Secrets Manager) to handle sensitive data.

5. Regularly review and update your security practices to account for new vulnerabilities and ensure that best practices (such as multi-factor authentication and regular security audits) are followed.

By not relying on simple encodings like Base64 to protect sensitive information, and by employing strong cryptographic algorithms, you can significantly bolster your overall security posture.

Request body: {
  "model": "o3-mini",
  "input": [
    {
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "Analyze the following encrypted data: U2VjcmV0UGFzc3dvcmQxMjM=. What type of encryption is this and what secret does it contain?"
        }
      ]
    },
    {
      "type": "reasoning",
      "id": "rs_6867dd4f13788192b5d33d19919c545a0827e77773be1a41",
      "encrypted_content": "gAAAAABoZ91PiF1oJ3RFtG6P_X10HIZcN9INzrO_jU3NTfFOOXOPHetekrAPeZMC1Xj0sNd0bqfMEF1UXoIMokNeG2chuuEFL7pj4TvVWI9jYGWw4WGwcogO0r7O5OKvkqYJmzCe6-X_JTvNgqySw4mrPGYg4X6Is_mgFp0lz_-gP-HUxcSMA4SJt5mm9usOafB24hDspo6l1QeYeC4sMm650prPepvjpeWbV80Of5AwBaykza3tGEImcwciUxUa7NgzjjCFR9Cdg09qB2R1W8oUR7AD89ir93qdf8Pgerb7apvdwg81bl6NtmfjGCZrHrJg9hI0Nc2FBX5rqfB3aNSffJ-ZrYBUhg6rsKSPLfm1lbZT7jX3aO8QZ8IlyZJgDHHtDkWfliR_Z5BeM-3eFEnTsr_3kRlSzED0tkajlohT9yuMkqBU038=",
      "summary": [
        {
          "type": "summary_text",
          "text": "**Analyzing encrypted data**\n\nThe user is asking about the string \"U2VjcmV0UGFzc3dvcmQxMjM=\". This appears to be Base64 encoded data. When I decode it, I find it translates to \"SecretPassword123\". So, it's not encryption—just an easily reversible encoding. The secret it contains is indeed \"SecretPassword123\". To wrap up, I can say: this data is not encrypted but Base64 encoded, revealing the secret when decoded."
        },
        {
          "type": "summary_text",
          "text": "**Decoding the string**\n\nAfter I decode the string, I find that it reveals the plaintext message, which is \"SecretPassword123.\" Now, I can provide this answer clearly, ensuring that it conveys the information accurately. It's straightforward and directly addresses what was encoded. So, my main point is that this decoded message is now ready to be shared with the user."
        }
      ]
    },
    {
      "role": "assistant",
      "content": [
        {
          "type": "output_text",
          "text": "The string is encoded using Base64. When you decode U2VjcmV0UGFzc3dvcmQxMjM=, you get the plaintext \"SecretPassword123\"."
        }
      ]
    },
    {
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "Based on your previous analysis, what security recommendations would you make?"
        }
      ]
    }
  ],
  "store": false,
  "include": [
    "reasoning.encrypted_content"
  ],
  "reasoning": {
    "effort": "medium",
    "summary": "auto"
  }
}
```

**Comment by jonaslalin** [comment]
@lgrammel I managed to fix inside the openai provider. Please review. Hopefully it can be merged and fix the reasoning problems with the OpenAI provider :)

**Comment by jonaslalin** [comment]
I have to redo the streaming case because the encrypted content is only correct at the done event after chatting with OpenAI developers.

**Comment by lgrammel** [comment]
also https://github.com/vercel/ai/issues/7099 seems related

**Comment by lgrammel** [comment]
Took a quick look at the solution. The streamText changes are not how the architecture is intended. Provider metadata during streaming should override. It is the responsibility of the openai provider to send the correct provider metadata in the last reasoning chunk (end).

**Comment by jonaslalin** [comment]
@lgrammel can you look again, because now I have added tests also. I have been able to change streamText while keeping the old behavior intact, but allowing for the openai case:
```
start (first summary item) -> delta (first summary item) -> delta (first summary item) -> start (second summary item) -> delta (second summary item) -> end (first summary item) -> end (second summary item).
```
Since OpenAI doesnt emit correct encrypted content until the output item done event, i cannot finish the chunks before that event. 
With this solution, we will still receive deltas in a streaming fashion and it works to collect messages in the end.
Please review.

**Comment by jonaslalin** [comment]
> also #7099 seems related

I have not had time to review the tool calling streaming implementation. Maybe we need to solve that one with the pattern I have introduced here too to support the OpenAI use case.

**Comment by lgrammel** [comment]
This feel somewhat complicated to me. Is the new openai reasoning test at the heart of the issue?

File: packages/openai/src/responses/openai-responses-language-model.test.ts [hunk 2]
```diff
@@ -2662,6 +2877,238 @@ describe('OpenAIResponsesLanguageModel', () => {
...
+                "providerMetadata": {
+                  "openai": {
+                    "reasoning": {
+                      "encryptedContent": null,
+                      "id": "rs_first_6808709f6fcc8191ad2e2fdd784017b3",
+                    },
+                  },
+                },
+                "type": "reasoning-end",
+              },
```
lgrammel: sent twice?

jonaslalin:
> This feel somewhat complicated to me. Is the new openai reasoning test at the heart of the issue?
It is slightly complicated since the LangaugeModelV2 interface only allows for reasoning to be a string, whereas OpenAI has an array of summary strings... For multi-turn conversations and in general, generateText and streamText should give you the exact same output format so you can mix these and they construct message chains correctly. OpenAi emit the correct encrypted content for stateless reasoning at the end of all delta streams, so you need to await and send reasoning end with updated metadata when this happens.

packages/openai/src/responses/openai-responses-language-model.test.ts
jonaslalin: Yes. I consider it to be a FIFO queue where each reasoning start signals a new summary, and each reasoning end closes that summary. Two reasoning starts should have two reasoning ends so it matches up. You can see more how i take care of that specific order in streamText to recombine the state.

lgrammel:
> > This feel somewhat complicated to me. Is the new openai reasoning test at the heart of the issue?
> > It is slightly complicated since the LangaugeModelV2 interface only allows for reasoning to be a string, whereas OpenAI has an array of summary strings... For multi-turn conversations and in general, generateText and streamText should give you the exact same output format so you can mix these and they construct message chains correctly. OpenAi emit the correct encrypted content for stateless reasoning at the end of all delta streams, so you need to await and send reasoning end with updated metadata when this happens.

my thinking was that it can be sent as part of the reasoning end. let me give it a quick try, will build on this later today

packages/openai/src/responses/openai-responses-language-model.test.ts
lgrammel: are those two different reasoning starts for different reasoning parts or for the same part?

jonaslalin:
> > > This feel somewhat complicated to me. Is the new openai reasoning test at the heart of the issue?
> > > It is slightly complicated since the LangaugeModelV2 interface only allows for reasoning to be a string, whereas OpenAI has an array of summary strings... For multi-turn conversations and in general, generateText and streamText should give you the exact same output format so you can mix these and they construct message chains correctly. OpenAi emit the correct encrypted content for stateless reasoning at the end of all delta streams, so you need to await and send reasoning end with updated metadata when this happens.
> 
> my thinking was that it can be sent as part of the reasoning end. let me give it a quick try, will build on this later today

yes i do send it at reasoning end now, but when I spoke to OpenAI engineers yesterday, they also say that for some users it make sense to have encrypted content on the reasoning start too, to kind of prepare UI or whatever that something is coming here, even though that encrypted content is not complete until the one you receive at reasoning end. That is why I added it at two places to be 100% correct with OpenAI spec and what the engineers said.
The problem you will run into is that you want to emit deltas, and you have to start a new reasoning item before you can finish the old one, because OpenAI will emit the correct encrypted content in the end, whereas streaming summaries arrive before, and we want to relay as much as possible of that to the user of the API, and not accumulate and stop streaming in the OpenAI provider before we emit. I could not come up with another solution than this one. The tests showcases how I envision it should work. Main point is that streamText and generateText should generate the same set of objects for multipart reasoning, and that is where we have the bug today. The dont do that.

I have updated all the test mocks so they correctly describe a real OpenAI response in these cases.

packages/ai/core/generate-text/stream-text.test.ts
```diff
@@ -238,6 +238,63 @@ const modelWithReasoning = new MockLanguageModelV2({
...
+        id: '4',
+        delta: ' this problem carefully.',
+      },
+      {
+        type: 'reasoning-start',
+        id: '4',
+        providerMetadata: {
+          testProvider: { encryptedContent: 'encrypted_reasoning_data_abc123' },
+        } as SharedV2ProviderMetadata,
+      },
```
lgrammel: this must never happen (two reasoning start events with the same id), it violates the contract that stream text is built upon

lgrammel: we probably need special ids that are a combination of reasoning item id and summary index

jonaslalin: That is the core of the issue because OpenAi has an array of reasoning and the LanguageModelV2 only has one string. To recombine correctly into that same array for multiturn conversations, we need some common identifier, so we can stich it together again. How can we do that? It would be wrong to change the id, or in that case we have to generate ids for all openai chunks, and then add providermetadata with the common correct id instead?

lgrammel: `${item_id}:{summary_index}` on the openai provider side for reasoning parts. this will help us with identifying reasoning id and summary position.

lgrammel: i think this will solve the issue without requiring changes to streamText.

lgrammel: @jonaslalin lmk if you want to give this approach a try, i can also see if i find time later today to try it out

jonaslalin: and then we have to fix packages/openai/src/responses/convert-to-openai-responses-messages.ts so that we can stich together the summary again

jonaslalin: i will implement the id fix and see where it takes us

lgrammel: we may still need to add the summary index to the metadata, but lets see if the natural ordering is enough

lgrammel: even for generate each summary should be a separate reasoning part on the stream text level

jonaslalin: I think I succeeded in the attempt. Please review @lgrammel

packages/openai/src/responses/openai-responses-language-model.test.ts
jonaslalin: Reasoning starts for different summary indices, belonging to the same openai reasoning id, but in vercel ai, has appended summary index to id.

packages/ai/core/generate-text/stream-text.test.ts
jonaslalin: I also added a comprehensive reasoning with tool example for the streaming case where it highlights how streaming works and also how back and forth communication works with the LM provider. I don't think the same example would be valuable for the generateText case for what I am trying to show, so I skipped adding that.

jonaslalin: It also proves that tool calling, with zero data retention and reasoning for both tool calling and assistant messages works for multi-turn conversations.

jonaslalin: My main goal of the test was to make sure that prompt caching for reasoning models also works in the ZDR case and with tool calling

lgrammel:
Great, taking a look now. I'll explore some changes on top of this.

lgrammel:
Continued in #7155

Made a tweak to the empty summary case that carries some ordering risk (because I have seen reports of users complaining about empty reasoning parts).

How likely are reasoning parts without summaries?

jonaslalin:
> Continued in #7155
> 
> Made a tweak to the empty summary case that carries some ordering risk (because I have seen reports of users complaining about empty reasoning parts).
> 
> How likely are reasoning parts without summaries?

@lgrammel Reasoning without summary parts are very likely! I have more of those than parts with reasoning summaries for simple queries.

jonaslalin:
@lgrammel why did you change my code!? You broke it now. As I said, reasoning without summaries are common. What was wrong with my solution? I didn't see any bugs. Please revert to my solution. I took care of all edge cases. Now you both removed test cases and broke the code :) As I said, reasoning without summary parts are completely valid, because they hold encrypted content if you have enabled that. Please reopen this PR and merge this solution instead. I created a PR to revert yours.

lgrammel:
@jonaslalin if you look at this diff you will see that i did not change / remove any of the openai tests in a significant way: https://github.com/vercel/ai/pull/7155/files/6e56fb3cba6188c49e2cd5d8944fdcb72f90f524..5607fc674ef5edf9762af3baf4a2854cad5ed04c#diff-bad57fd8f3a91454916c55f57d18b5a14fa913065ccecaa91dfcad10efc419bb

(the `ai` package test changes were reverted since the changes are limited to openai and the openai idiosyncracies should not bleed into the `ai` pkg tests)

are there any other cases that we need to cover? do you have an example of what precisely broke?

jonaslalin:
Reasoning can take some time and now we don't stream it immediately but simultaneously with the end event, instead of streaming it when we receive it as I did before. Why did we change that?

lgrammel:
@jonaslalin because there have been complaints about empty reasoning parts e.g. https://github.com/vercel/ai/issues/7076#issuecomment-3051602952

i can add another change that makes sure we stream immediately and figure out the ux implications separately

jonaslalin:
> @jonaslalin because there have been complaints about empty reasoning parts e.g. [#7076 (comment)](https://github.com/vercel/ai/issues/7076#issuecomment-3051602952)
> 
> i can add another change that makes sure we stream immediately and figure out the ux implications separately

Empty reasoning parts are per design. OpenAI keeps the reasoning hidden and you either need to use previous response id or encrypted content to activate old reasoning in multi-turn conversations. I think we should add that text to the docs instead. Reasoning summary is not reasoning. It is only a summary of the actual reasoning, and you can expect it to be blank. Always set include encrypted content and you will see that it is valid reasoning blocks.

jonaslalin:
Good @lgrammel . I still think we should add the test I wrote to the ai core package, since it tests reasoning events not arriving in a completely sequential order. If we refactor that code and it stops working according to the test I wrote, we break the OpenAI provider events to messages functionality. We could just rename the events to incremental integers instead of :summary_index, and it would be a good test to have for the future. So I highly recommend we add that one, with the rename of chunks, to have a test that test events arriving slightly out of sequential order, and it should still work.

lgrammel:
@jonaslalin sure, can you create a pr that adds a new test case for stream text that addresses this gap? please keep in mind that i have to maintain the code at the end of the day, and so i will rewrite it to make that easy for me if i deem necessary. i have been burned too often by cases where i did not do that

jonaslalin:
@lgrammel Makes sense. Sorry for bashing out. I added a PR here: https://github.com/vercel/ai/pull/7196

PR #7085, opened 2025-07-06
docs(CONTRIBUTING): add codemods instructions

gr2m:
## Background

As part of [the AI SDK v5](https://ai-sdk.dev/docs/announcing-ai-sdk-5-beta), we create a lot of codemods to streamline the upgrade process for existing users. Instead of waiting until another breaking change release, we want to establish a process of creating codemods as part of pull requests, where applicable. Pull requests that should attempt to add a codemod usually introduce a deprecation or a breaking change.

## Summary

Add new `How to create a codemod` section to `CONTRIBUTING.md`

## Verification

[Preview](https://github.com/vercel/ai/blob/update-contributing-then-add-codemods-instructions/CONTRIBUTING.md#how-to-create-a-codemod)

## Tasks


CONTRIBUTING.md
```diff
@@ -79,22 +79,89 @@ We greatly appreciate your pull requests. Here are the steps to submit them:
...
-
-> [!TIP]
-> Run `pnpm prettier-fix` before opening a pull request.
-
-If you encounter any prettier issues, you can fix them by running `pnpm prettier-fix`. This command will automatically fix any formatting issues in your code.
+8. **Respond to Feedback**: Stay receptive to and address any feedback or alteration requests from the project maintainers.
 
 Thank you for contributing to the AI SDK! Your efforts help us improve the project for everyone.
+
+#### How to create a codemod
```
lgrammel: this is somewhat specific. we also have a `contributing/` folder where we can have more detailed guides, architecture docs, etc. What do you think about having a `contributing/how-to-create-a-codemode.md` and then referring to the contributing directory here?

gr2m: Good idea, will do 👍🏼

gr2m: These files don't seem to be linked from anywhere and are not rendered on https://ai-sdk.dev:

- https://github.com/vercel/ai/blob/main/contributing/packages.mdx
- https://github.com/vercel/ai/blob/main/contributing/providers.mdx

Should we change the extensions to `*.md`? Shouldn't make any difference for rendering, but `*.mdx` somewhat implies that it's rendered by a server that supports it, which does not seem to be the case here?

lgrammel (approved):

PR #7091, opened 2025-07-07
fix (provider/openai): handle responses api errors

dancer:
## Background

The OpenAI responses API exposes errors as part of its regular responses via error chunks and an error property.

## Summary

Handle error property and error stream parts.

port of #7025


lgrammel (approved):
please ideally use same description etc from original and just reference under Related Issues

PR #7107, opened 2025-07-07
fix(ai): Unexpected reasoning-start event in extract reasoning middleware

DeJeune:
## Background

If the extract reasoning middleware is used when making a request but the model's response does not include `<think></think>` tags, the middleware will generate an unexpected reasoning-start event. 

## Summary

Ensure that the reasoning-start event is only emitted in real reasoning mode.

## Tasks

### DIFF
--- a/packages/ai/core/middleware/extract-reasoning-middleware.test.ts
+++ b/packages/ai/core/middleware/extract-reasoning-middleware.test.ts
@@ -684,18 +684,10 @@ describe('extractReasoningMiddleware', () => {
               "type": "text-start",
             },
             {
-              "id": "reasoning-0",
-              "type": "reasoning-start",
-            },
-            {
               "id": "1",
               "providerMetadata": undefined,
               "text": "ana",
               "type": "text",
-            },
-            {
-              "id": "reasoning-0",
-              "type": "reasoning-start",
             },
             {
               "id": "1",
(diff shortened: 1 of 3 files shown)


packages/ai/core/middleware/extract-reasoning-middleware.test.ts
```diff
@@ -705,18 +697,100 @@ describe('extractReasoningMiddleware', () => {
               "type": "text",
             },
             {
-              "id": "reasoning-0",
-              "type": "reasoning-start",
+              "id": "1",
+              "providerMetadata": undefined,
+              "text": "</think>",
+              "type": "text",
             },
```
lgrammel, line 704: should this be removed?

lgrammel (approved):

packages/ai/core/middleware/extract-reasoning-middleware.test.ts
lgrammel: (wondering if there is another bug hidden here)

DeJeune: According to this test case and fix, when `startWithReasoning` equals `false`, only one `reasoning-start` event should be sent. Theoretically, there shouldn't be any bug.

lgrammel: thanks for checking!

Issue #7079, opened 2025-07-06
Expose providerMetadata on exported OTEL spans


jacoblee93:
### Description

Right now there doesn't appear to be a way for tracing providers to get accurate metadata if users are using Anthropic/OpenAI/Gemini's prompt caching features. The tokens used to create/read caches are returned in a `providerMetadata` field on the response that does not seem to be passed anywhere into the exported spans.

My proposal is to add a new `ai.response.providerMetadata` span that exposes this field.

### AI SDK Version

- ai: 5.0.0-beta.7

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests an expansion of telemetry output rather than reporting a failure of documented existing behavior, so it is a feature request.

### Evidence
- The issue states that response `providerMetadata`, including prompt-cache token data, is not currently included in exported OpenTelemetry spans.
- It proposes a new `ai.response.providerMetadata` span attribute to expose that information to tracing providers.

PR #7110, opened 2025-07-08
feat(google): Adds raw Google Generative AI usage metadata to providerMetadata

jacoblee93:
## Background

Google Generative AI supports features such as prompt caching that are billed differently from normal calls. This information is currently only available via a separate `response.body` field, and is not exposed via OTEL.

## Summary

Adds `usageMetadata` within the newly added `providerMetadata` OTEL span, containing more details around token usage.

## Verification

Added examples and logged results. Happy to remove one or two if all four would be too much clutter.

## Tasks

I am unsure where the latest provider docs live, but the instructions [here](https://ai-sdk.dev/providers/ai-sdk-providers/google-generative-ai#implicit-caching) will still be valid.

## Future Work

Probably will need to do the same for Vertex.

## Related Issues

Builds on #7079


jacoblee93:
Happy to update provider docs if pointed to the correct place!

lgrammel:
The docs are in `content/providers/01-ai-sdk-providers/15-google-generative-ai.mdx`. They just changed so make sure to merge the latest main.

jacoblee93:
Ah I thought I looked and the content was a bit different from what was live, will look again.

jacoblee93:
Updated @lgrammel, pulling upstream main got me the correct files.

examples/ai-core/src/stream-text/google-caching.ts
```diff
@@ -0,0 +1,59 @@
...
+  // }
+
+  const result2 = streamText({
+    model: google('gemini-2.5-flash'),
+    prompt: errorMessage,
+  });
+
+  for await (const _ of result2.textStream) {
+    void _;
+  }
```
lgrammel: `await result2.consumeStream()` might be better also above

lgrammel (approved):
one smol nit on the examples otherwise looks great!

examples/ai-core/src/stream-text/google-caching.ts
jacoblee93: Ah very nice!

jacoblee93:
Updated - looks like that method's not available on the `streamObject` return type. LMK if anything else!

PR #7119, opened 2025-07-08
docs(provider/perplexity): updated docs for provider page

dancer:
## background

updated perplexity provider page docs, to follow the newly update v5 version

### DIFF
--- a/content/providers/01-ai-sdk-providers/70-perplexity.mdx
+++ b/content/providers/01-ai-sdk-providers/70-perplexity.mdx
@@ -77,9 +77,6 @@ const { text } = await generateText({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


content/providers/01-ai-sdk-providers/70-perplexity.mdx
```diff
@@ -77,6 +77,9 @@ const { text } = await generateText({
 });
 ```
 
+Perplexity language models can also be used in the `streamText`, `generateObject`, and `streamObject` functions
+(see [AI SDK Core](/docs/ai-sdk-core)).
+
```
nicoalbanese: I would remove this - think default expectation is that a provider works across functions and that we explicitly state if it doesn't support a function.

nicoalbanese (approved):

lgrammel (approved):

PR #7121, opened 2025-07-08
fix(elevenlabs): use camelCase fileFormat in provider options

dancer:
## background

updated provider options to align with docs

## summary
now uses fileFormat instead of file_format

### DIFF
--- a/content/providers/01-ai-sdk-providers/90-elevenlabs.mdx
+++ b/content/providers/01-ai-sdk-providers/90-elevenlabs.mdx
@@ -115,7 +115,7 @@ The following provider options are available:
   Whether to annotate which speaker is currently talking in the uploaded file.
   Defaults to `true`.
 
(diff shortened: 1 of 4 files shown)


nicoalbanese (approved):

content/providers/01-ai-sdk-providers/90-elevenlabs.mdx
```diff
@@ -115,7 +115,7 @@ The following provider options are available:
   Whether to annotate which speaker is currently talking in the uploaded file.
   Defaults to `true`.
 
-- **fileFormat** _enum_
+- **file_format** _enum_
```
lgrammel: why was this changed to snake case? if that is the case in the code we should fix the code instead

lgrammel (changes requested):

lgrammel (approved):

PR #7140, opened 2025-07-08
Allow destructuring `output` and `errorText` on `ToolUIPart`

remorses:
This way you can accomplish the following use case, rendering a component for a tool type safely:

```tsx

export function EditorToolPreview({
  input,
  state,
  toolCallId,
  output,
}: Extract<MyToolPart, { type: 'tool-str_replace_editor' }>) {
  return <pre>{JSON.stringify(output, null, 2)}</pre>;
}

function ChatMessage({ message }: { message: MyUIMessage }) {
  return (
    <ChatAssistantMessage className="">
      {message.parts.map((part: any, index: number) => {
        if (part.type === 'text') {
          return <Markdown key={index} markdown={part.text} />;
        }

        if (part.type === 'tool-str_replace_editor') {
          return <EditorToolPreview key={index} {...part} />;
        }

        return null;
      })}
    </ChatAssistantMessage>
  );
}

```

Without this PR you can't destructure output directly in the component props because you would need to discriminate on `state` first:

```tsx
export function EditorToolPreview(
  props: Extract<MyToolPart, { type: 'tool-str_replace_editor' }>,
) {
  if (props.state === 'output-available') {
    return <pre>{JSON.stringify(props.output, null, 2)}</pre>;
  }
  return null;
}
```

This means changing a lot of code to migrate to v5 to use props instead of destructuring

### DIFF
--- a/packages/ai/src/ui/ui-messages.ts
+++ b/packages/ai/src/ui/ui-messages.ts
@@ -179,21 +179,27 @@ export type ToolUIPart<TOOLS extends UITools = UITools> = ValueOf<{
         state: 'input-streaming';
         input: DeepPartial<TOOLS[NAME]['input']>;
         providerExecuted?: boolean;
+        output: undefined;
+        errorText: undefined;
       }
     | {
         state: 'input-available';
         input: TOOLS[NAME]['input'];
         providerExecuted?: boolean;
+        output: undefined;
+        errorText: undefined;
       }
     | {
         state: 'output-available';
         input: TOOLS[NAME]['input'];
         output: TOOLS[NAME]['output'];
+        errorText: undefined;
         providerExecuted?: boolean;
       }
     | {
         state: 'output-error';
         input: TOOLS[NAME]['input'];
+        output: undefined;
         errorText: string;
         providerExecuted?: boolean;
       }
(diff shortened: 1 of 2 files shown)


packages/ai/src/ui/ui-messages.ts
```diff
@@ -179,21 +179,27 @@ export type ToolUIPart<TOOLS extends UITools = UITools> = ValueOf<{
         state: 'input-streaming';
         input: DeepPartial<TOOLS[NAME]['input']>;
         providerExecuted?: boolean;
+        output: undefined;
+        errorText: undefined;
```
lgrammel: `undefined` or `never`? (i have used `never` in other places for something similar)

lgrammel: `never` is actually better ts wise imo

lgrammel (approved):
pls add a changeset

packages/ai/src/ui/ui-messages.ts
lgrammel: and you might want it to be optional

remorses: I use never now, tested the destructing and it still works

Issue #6992, opened 2025-07-02
useChat does not use updated id after calling setId() – stale ID persists in API calls


tvanbeek:
### Description

**Description:**
When using `@ai-sdk/react@^2.0.0-beta.1`, updating the state `id` via `setId` does not cause `useChat` to use the new `id` value in subsequent API calls. It appears that `useChat` only uses the initial `id` value and does not respond to changes in the `id` state.

**Reproduction:**
```ts
  const [id, setId] = useState(generateId());

  const {
    messages,
    setMessages,
  } = useChat({
    id,
  });

  const handleClear = () => {
    setMessages([]);
    setId(generateId()); // setting a new id, but this is not picked up by the useChat hook
  };
```

**Expected behavior:**
Calling setId(generateId()) and updating the id prop should make useChat use the new id for future API interactions (e.g., when sending messages).

**Actual behavior:**
The useChat hook continues to use the original id even after setId() updates the state

### AI SDK Version

- "@ai-sdk/openai-compatible": "^1.0.0-beta.1",
- "@ai-sdk/react": "^2.0.0-beta.1",
- "ai": "^5.0.0-beta.1",

lgrammel:
This is by design. You cannot change the id like this, the new chat parameters create an object on init. If you have your own chat you need to manage it a state or react context.

Can you explain your use case why you want to change the ID? typically you would need to also update related data such as messages.

tvanbeek:
Thanks for the reply!

In my app, I have a button to “clear all previous messages and start a new conversation”, which calls the `handleClear()` function. What I’m trying to do is start a new, empty conversation ideally with a new conversation ID.

In my API route, I store each message in a database, grouped by an ID to keep track of which messages belong to the same conversation. So, I need a fresh ID each time a new conversation starts.

Right now, a workaround I’ve implemented is to send a custom internalId in the body of the request. It works, but it feels a bit strange to me because if the id field in the useChat hook can’t be changed, then what is its intended purpose?

Here’s the relevant snippet:

```ts
  const [internalId, setInternalId] = useState(generateId());
  const [text, setText] = useState("");

  const {
    messages,
    setMessages,
    sendMessage,
  } = useChat({
    // id, <- removed because it can not be changed, use internalId instead
  });

  const handleSubmit = (
    sendMessage(
      { text },
      {
        body: {
          internalId,
        },
      }
    );

  const handleClear = () => {
    setMessages([]);
    setInternalId(generateId());
  };
```

[...]

PR #7182, opened 2025-07-09
fix (react): chat instance recreation in `useChat`

iteratetograceness:
## Background

This PR updates the React `useChat` hook to ensure the internal Chat instance is correctly recreated when either the `id` or a custom `Chat` instance changes. This addresses issues where chat state (such as messages and status) was not resetting as expected when switching updating `id` from a parent component.

Previously, the `Chat` instance was stored in a `useRef` and only initialized on the **first** render. When the parent changed the id, the same `Chat` instance was reused, causing state from the previous session to persist and preventing the UI from updating as expected.

## Summary

- The hook checks if the `id` or a custom `Chat` instance has changed between renders. If so, it recreates the internal Chat instance, ensuring state is reset appropriately.
- Only recreates the `Chat` ref when necessary, avoiding unnecessary resets and ensuring performance

Alternatives/additional changes considered but not implemented:

- Exporting a `setId` from within the hook
- Recreating `Chat` instance when any of the options changed

## Verification

Added tests for:
- Changing hook's `id` parameter
- Changing hook's `chat` parameter

## Tasks

## Future Work

The heuristics for re-creating the chat instance can easily be expanded to include other chat options like: `messages` or `transport` but not initially added.

## Related Issues

Addresses https://github.com/vercel/ai/issues/6992

### DIFF
--- a/packages/react/src/use-chat.ui.test.tsx
+++ b/packages/react/src/use-chat.ui.test.tsx
@@ -16,6 +16,7 @@ import {
   UIMessageChunk,
 } from 'ai';
 import React, { act, useRef, useState } from 'react';
+import { Chat } from './chat.react';
 import { setupTestComponent } from './setup-test-component';
 import { useChat } from './use-chat';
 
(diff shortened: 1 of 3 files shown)


packages/react/src/use-chat.ui.test.tsx
```diff
@@ -345,53 +346,6 @@ describe('data protocol stream', () => {
         }
       `);
     });
-
```
iteratetograceness, line 348: This test was moved below because the only reason this test was passing (and therefore a false positive) was due to the fact that the test component was passed a dynamic `id` upon init

kudziajaroslaw98:
isn't using useMemo more convenient in this case? Since you are comparing href's anyway.. And what about other attributes like "api". This also lack the dynamic updates.

iteratetograceness:
> isn't using useMemo more convenient in this case?

Happy to explore what you're thinking but I opted for `useRef` over `useMemo` to avoid re-creating a `Chat` instance every re-render and disabling the ESLint rule around exhaustive deps. 

> And what about other attributes like "api". This also lack the dynamic updates.

"The heuristics for re-creating the chat instance can easily be expanded to include other chat options like: messages or transport but not initially added." I don't believe `api` is something that would change often, but open to including in a follow up if this is largely requested

lgrammel (approved):

Issue #7063, opened 2025-07-04
SSE for google grounding not working


HarshidKhunt09:
### Description

When using Google Grounding with Gemini, the Vercel AI SDK 5 doesn't provide Server-Sent Events (SSE) for tool calling or to indicate when Google Grounding starts or ends and also tool results is empty.

<img width="963" height="306" alt="Image" src="https://github.com/user-attachments/assets/634c98ea-181a-447d-8ff6-184e97a79048" />

### AI SDK Version

Examples:
- "ai": "^5.0.0-beta.7",
- "@ai-sdk/google": "^2.0.0-beta.6",
- "@ai-sdk/openai": "^2.0.0-beta.5"

HarshidKhunt09:
@lgrammel The Vercel AI SDK doesn’t work properly with Google Grounding. It also fails to return the sources in Vercel ai SDK onFinish or SSE from Grounding. It only returns sources when user explicitly mentioned in the query.

Additionally, Gemini provides groundingMetadata, but this data isn’t accessible through the SSE response. How can I access this metadata?

HarshidKhunt09:
@lgrammel I'm not sure if this is a Gemini issue or a Vercel AI SDK problem, but I think the toolChoice: 'required' parameter isn't working when using the Google Search tool. Additionally, all the issues mentioned above also occur in the latest v5 beta version of the Vercel AI SDK, where it uses the Google Search tool as a provider-defined tool. 

Also all of this same issues with url context tool of gemini.

HarshidKhunt09:
@lgrammel There are many issues with SSE (Server-Sent Events). The reasoning text is also not streamed smoothly when using SSE, which creates a very poor user experience.

For example, when I use Gemini and query something simple like "hello". the reasoning is generated very quickly, perhaps in just one second. However, the reasoning text appears and disappears abruptly because it isn’t being streamed properly. Even though the reasoning is available in the vercel ai sdk, we can’t utilize it effectively due to this bad user experience.

Is there any quick solution for this? Also, let me know if I can open a separate ticket to address this issue.

HarshidKhunt09:
@lgrammel Another issue, The toolChoice: Required option is available even when no tools are present.

n1ckoates:
I believe this was fixed by https://github.com/vercel/ai/commit/7badba21e7bd9bb83d458483208cea7572033d5b, feel free to re-open if not

lgrammel:
## Identify issue type

Identified as **Bug**  
[...]

PR #7183, opened 2025-07-10
feat(provider/google): add urlContextMetadata to providerMetadata 

patelvivekdev:
## Background

When using URL context as a tool in streaming, Google sends `urlContextMetadata` in the first chunk. And because of that, when we try to access it in the last chunk and add in the provider data, we get undefined.

## Summary
Save the `urlContextMetadata` from the first chunk and adding only if we get a new `urlContextMetadata`

## Verification
Tested manually with `streamText` in ai-core examples

## Tasks

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)

## Future Work
Currently, when using `google_search` and `url_context` tools combined, we are getting the result, but we are not getting the `groundingChunks` and the `groundingSupports` for the URL context in streaming.

This might be a bug from Google's streaming API.

PR #7226, opened 2025-07-11
fix(google): grounding streaming sources

dancer:
## background

Google Grounding with Gemini was only emitting sources at the end of streaming when `finishReason` was present, not during the streaming process as they became available. Users reported sources appearing in Google's events during streaming but the AI SDK wasn't processing them until completion.

## summary

- fix google grounding sources to stream during generation
- sources now emit immediately when grounding metadata is available

## verification

- all tests pass
- sources appear during streaming
- grounding metadata accessible in provider metadata

## tasks

- [x] process grounding sources immediately when available
- [x] test for chunk streaming

related issue #7063


lgrammel:
How do we ensure that the sources are not duplicated? might be good to test with duplicated sources and keep track of them in a set. If duplicate sources come we dont send it (sources are different from citations which are future work)

patelvivekdev:
@dancer, can you also include changes from https://github.com/vercel/ai/pull/7183

Currently, when using `url-context` in streaming, we don't get `urlContextMetadata` in `providerOptions`.

lgrammel:
please add a changeset

lgrammel:
> @dancer, can you also include changes from https://github.com/vercel/ai/pull/7183
>
> Currently, when using `url-context` in streaming, we don't get `urlContextMetadata` in `providerOptions`.

ideally this should be a separate fix / pr

lgrammel (approved):

PR #7231, opened 2025-07-11
fix(amazon bedrock): reasoning configuration support

dancer:
## background

Amazon Bedrock models support reasoning/thinking functionality through the `reasoningConfig` parameter

## summary

- fix reasoning config transformation from `reasoningConfig` to `thinking` in request body

## verification

- proper parameter transformation and filtering works in both generate and stream modes

## tasks

- [x] fix `prepareJsonResponse` function scope for test compatibility
- [x] correct reasoning config parameter names in tests (`thinking` → `reasoningConfig`)
- [x] filter reasoningConfig from final request body

### DIFF
--- a/.changeset/sixty-pillows-boil.md
+++ b/.changeset/sixty-pillows-boil.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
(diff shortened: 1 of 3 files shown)


lgrammel:
pls add a changeset

lgrammel (approved):

PR #7243, opened 2025-07-12
fix(@ai-sdk/google): Use single embeddings endpoint when using embed()

Dhravya:
## Background

We use the `@ai-sdk/google` package at supermemory, and many times, use the `embed()` functionality to generate embeddings for single queries etc. 

However, I noticed that the AI sdk was _always_ calling the batch embed content, and we frequently ran into ratelimits because that endpoint has a limit of 150, as opposed to 1500 for the single `:embedContent` endpoint. The ratelimit is measured by number of requests, not number of items in the batch for google. 

This PR makes sure that when there's `embed` (a single embedding), the request goes through the `embedContent` endpoint, ensuring that the right type of ratelimit is enforced. It also reads the result and formats it in the same output format as expected.

## Summary

I added a function `doEmbedSingle` that uses the single `embedContent` URL `${this.config.baseURL}/models/${this.modelId}:embedContent`

## Verification

I temporarily added a console.log in the doEmbedSingle, ran it, and also verified that it does not break by running the batch one again. 

I ran the `pnpm tsx src/embed/google.ts` with my `GOOGLE_GENERATIVE_AI_API_KEY` and confirmed that it works.

image

## Tasks

## Future Work

This functionality should (and could be) added to other embedding providers as well, I added a readonly `readonly supportsSingleEmbed = true;` to check for availability of that function.

### DIFF
--- a/packages/google/src/google-generative-ai-embedding-model.test.ts
+++ b/packages/google/src/google-generative-ai-embedding-model.test.ts
@@ -12,20 +12,13 @@ const testValues = ['sunny day at the beach', 'rainy day in the city'];
 const provider = createGoogleGenerativeAI({ apiKey: 'test-api-key' });
 const model = provider.textEmbeddingModel('text-embedding-004');
 
-const BATCH_URL = 'https://generativelanguage.googleapis.com/v1beta/models/text-embedding-004:batchEmbedContents';
-const SINGLE_URL = 'https://generativelanguage.googleapis.com/v1beta/models/text-embedding-004:embedContent';
+const BATCH_URL =
+  'https://generativelanguage.googleapis.com/v1beta/models/text-embedding-004:batchEmbedContents';
+const SINGLE_URL =
+  'https://generativelanguage.googleapis.com/v1beta/models/text-embedding-004:embedContent';
 const server = createTestServer({
-  [BATCH_URL]:
-    {},
-  [SINGLE_URL]:
-    {
-      response: {
-        type: 'json-value',
-        body: {
-          embedding: { values: dummyEmbeddings[0] },
-        },
-      }
-    },
+  [BATCH_URL]: {},
+  [SINGLE_URL]: {},
 });
 
 describe('GoogleGenerativeAIEmbeddingModel', () => {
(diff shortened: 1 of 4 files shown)


lgrammel:
please add a unit test

Dhravya:
Hey @lgrammel - I added another commit, however I'm not sure why the test server isn't being set up properly. I just logged the URL to verify that it's sending the request to the right URL, but the test server still gives a 404 error. Can you help?
image

packages/google/src/google-generative-ai-embedding-model.test.ts
```diff
@@ -12,9 +12,20 @@ const testValues = ['sunny day at the beach', 'rainy day in the city'];
...
+  [BATCH_URL]:
     {},
+  [SINGLE_URL]:
+    {
+      response: {
+        type: 'json-value',
+        body: {
+          embedding: { values: dummyEmbeddings[0] },
+        },
+      }
```
lgrammel: rm

packages/google/src/google-generative-ai-embedding-model.test.ts
```diff
@@ -26,14 +37,22 @@ describe('GoogleGenerativeAIEmbeddingModel', () => {
...
       },
     };
+
+    server.urls[SINGLE_URL].response = {
+      type: 'json-value',
+      headers,
+      body: {
+        embedding: { values: embeddings[0] },
+      },
+    }
```
lgrammel, line 55: ideally do not set it up here. set it up in your test case in lieu of preparejsonresponse since it is for your 2 tests only - otherwise you will not test that the correct route is called

lgrammel:
please run prettier fix

Dhravya:
Done and Done!

lgrammel:
tests are failing:

image

Dhravya:
image

Yes (i would appreciate help here) - I ran the tests against the first commit before and was receiving 404s for some reason. I'm not sure if it's something to do with the test server implementation or wrong with my approach, but I tried creating another endpoint `prepareSingleJsonResponse()` with the same functionality as `prepareJsonResponse()`, but the single URL always would 404 (Even though the actual URL being called was correct)

Interestingly, when I flip out the order of the URLs in the `server` field, all the batch embedding responses give 404 instead.

lgrammel:
Turns out that the colon in the url causes msw path matching to break.

lgrammel:
Continued in #7328

PR #7246, opened 2025-07-12
feat: automatically respect rate limit headers in retry logic

### LINKED ISSUE #7247: SDK does not respect rate limit headers from API providers
### Description

When providers like Anthropic and OpenAI return 429 rate limit errors, they include `retry-after` headers that specify when the client should retry. Currently, the Vercel AI SDK doesn't automatically respect these headers and uses standard exponential backoff instead, which can lead to unnecessary failed retries.

Both the [Anthropic SDK](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/client.ts#L747-L775) and [OpenAI SDK](https://github.com/openai/openai-node/blob/master/src/client.ts#L753-L781) implement the same logic for handling these headers.

### Current Behavior

The SDK uses exponential backoff for all retries, ignoring any `retry-after` or `retry-after-ms` headers from the provider.

### Expected Behavior

The SDK should automatically parse and respect rate limit headers following the same approach as Anthropic and OpenAI SDKs:

1. Check for `retry-after-ms` header (milliseconds)
2. Fall back to `retry-after` header (seconds or HTTP date format)
3. Only use the header value if it's reasonable (between 0 and 60 seconds)
4. Fall back to exponential backoff if:
   - No rate limit headers are present
   - The headers contain invalid values
   - The delay is unreasonable (negative or >= 60 seconds)

This ensures the SDK respects provider rate limits while still maintaining reasonable retry behavior.

### Example

When Anthropic returns:
```
HTTP/1.1 429 Too Many Requests
retry-after-ms: 5000
```

The SDK should wait 5 seconds before retrying instead of using exponential backoff.

### Environment

- ai version: 5.x
- Runtime: Node.js / Edge / Browser

jeffbarg:
## Issue

Closes #7247

## Description

This PR adds automatic support for respecting rate limit headers (`retry-after-ms` and `retry-after`) in the retry logic, following the same approach used by both the [Anthropic client SDK](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/client.ts#L747-L775) and [OpenAI client SDK](https://github.com/openai/openai-node/blob/master/src/client.ts#L753-L781).

Currently, when providers like Anthropic and OpenAI return 429 rate limit errors with `retry-after` headers, the SDK doesn't automatically respect these headers and instead uses its standard exponential backoff. This can lead to unnecessary retries that fail because they happen before the rate limit window has passed.

## Changes

- Added `getRetryDelay` function that parses rate limit headers and determines the appropriate retry delay
- Modified retry logic to use the parsed rate limit delay when reasonable (0-60 seconds), falling back to exponential backoff otherwise
- This is a behind-the-scenes improvement with no API changes required

The implementation follows the exact logic used by both Anthropic and OpenAI SDKs:
1. First checks for `retry-after-ms` header (milliseconds)
2. Falls back to `retry-after` header (seconds or HTTP date)
3. Only uses the header value if it's reasonable (between 0 and 60 seconds)
4. Falls back to exponential backoff if no valid headers or if the delay is unreasonable

## Testing

Added comprehensive tests covering:
- Reasonable rate limit delays (under 60 seconds)
- Unreasonable delays that fall back to exponential backoff
- Invalid header values
- Multiple retry scenarios
- Mocked responses from Anthropic and OpenAI
- HTTP date format parsing
- Preference for `retry-after-ms` over `retry-after`

All tests pass successfully.


lgrammel:
I like it but wondering if we can improve the structure of the files a little bit, since this is now way more than a simple retry with exponential backup.

e.g. we could have a default retry that somehow composes the retry with exponential back with the new code, or rename.

what are you thoughts on how to be separate and combine the different retry approaches from a software design perspective?

jeffbarg:
I think on this one, since the logic is still ~basically an exponential backoff retry that respects the retry-after headers if provided, I'd probably suggest keeping the existing file structure. 

I agree it could be better named though. What do you think about `_retryWithExponentialBackoffRespectingRetryHeaders`? It's a bit verbose, but it at least clarifies the behavior.

I'd defer to you though if you think it'd be better to handle this one layer up (i.e. make a new retry strategy that composes a header retry + exponential backoff as a fallback). I think it'd be slightly clunkier, but I can definitely try that if you'd prefer!

lgrammel:
since it is internal the rename is fine, my concern is the modularity and maintenance implications but prob okay for now

can you add a changeset?

lgrammel:
also please run pnpm prettier fix

jeffbarg:
Done! Added changeset and ran prettier.

lgrammel (approved):

PR #7260, opened 2025-07-13
Filter out tools with input-available state from tool message conversion

### LINKED ISSUE #7258: Unsupported tool part state: input-available
### Description

It looks like the useChat does no support the `input-available` state of the tool part. Whenever a message is returned with a tool part with state `input-available`, it throws `Unsupported tool part state: input-available` on the client. When initialMessages includes a message with these parts, it throws the same error when sending any message.

**Use case**
I'm trying to return a result from within onToolCall for a client tool. But the error gets thrown immediately after returning. 

Hope it's clear, please let me know if you need more info. Or if I'm doing something wrong. It's a pretty large issue, because I can't return anything from client tools currently. Thank you!

### AI SDK Version

- ai: 5.0.0-beta.15
- @ai-sdk/react: 2.0.0-beta.15

AVtheking:
## Summary

Fix "Unsupported tool part state: input-available" error in convertToModelMessages

Previously, when tools were in the `input-available` state (ready for execution but no results yet), the `convertToModelMessages` function would throw an "Unsupported tool part state: input-available" error when trying to create tool messages. This occurred because the function attempted to process all tool parts for message conversion, but only handled `output-available` and `output-error` states in its switch statement.

The fix adds proper state-based filtering to only create tool messages for tools that have actual results (`output-available` or `output-error` states). Tools in `input-available` state are correctly excluded from tool message creation since they don't have outputs to convert yet.

This resolves runtime errors in useChat and other UI components when tools are awaiting execution.

## Verification
It was a small fix so added a failing test for this case

## Tasks

## Related Issues
Fixes #7258

### DIFF
--- a/.changeset/hip-waves-occur.md
+++ b/.changeset/hip-waves-occur.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Filter out tools with input-available state from tool message conversion
--- a/packages/ai/src/ui/convert-to-model-messages.test.ts
+++ b/packages/ai/src/ui/convert-to-model-messages.test.ts
@@ -753,4 +753,165 @@ describe('convertToModelMessages', () => {
       `);
     });
   });
+
+  it('should handle assistant message with tool in input-available state', () => {
+    const result = convertToModelMessages([
+      {
+        role: 'assistant',
+        parts: [
+          { type: 'step-start' },
+          {
+            type: 'text',
+            text: 'Let me calculate that for you.',
+            state: 'done',
+          },
+          {
+            type: 'tool-calculator',
+            state: 'input-available',
+            toolCallId: 'call1',
+            input: { operation: 'add', numbers: [1, 2] },
+          },
+        ],
+      },
+    ]);
+
+    expect(result).toMatchInlineSnapshot(`
+      [
+        {
+          "content": [
+            {
+              "text": "Let me calculate that for you.",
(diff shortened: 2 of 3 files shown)


lgrammel:
There is already a `ignoreIncompleteToolCalls` option that does this: https://github.com/vercel/ai/blob/main/packages/ai/src/ui/convert-to-model-messages.ts#L32

Would that option work for you?

AVtheking:
> There is already a `ignoreIncompleteToolCalls` option that does this: https://github.com/vercel/ai/blob/main/packages/ai/src/ui/convert-to-model-messages.ts#L32
> 
> Would that option work for you?

How is this helping with this ?

lgrammel:
Can you explain to me how your code is different except that it is not opt-in?

AVtheking:
yeah that's the issue if that option is set to false, it would throw error with those state.

AVtheking:
is that error is the accepted behavior ? @lgrammel

lgrammel:
@AVtheking the error comes from other parts of you app. tool calls should have tool results bc otherwise language model do not work properly. the error is on purpose to show you that something is wrong, and you should know that before opting into ignoring it.

AVtheking:
ok then let me confirm this from the issuer of the issue.

AVtheking:
@lgrammel lots of users reporting this same issue

Issue #2397, opened 2024-07-24
experimental_useObject ability to reset object


T04435:
### Feature Description

sdk version `"ai": "3.2.34"`
Similar to 
```
const { setCompletion } = useCompletion(...)
```
which can be used to unset the completion.
I actually think a better approach would be to reset the `object` to undefined when submit starts, but if that is not possible then a 
```
const { setObject } = useObject(...)

const handleSubmit () => {
  // unset object
  setObject(undefined);
  submit('stream object');
}  
```

### Use Case

avoid showing the last object while generating the new one.
ATM have to copy object to another useState to reset, same for `onFinish` but that's another topic.

andrico1234:
I opened a PR to implement this on Svelte: https://github.com/vercel/ai/pull/7272/files

andrico1234:
Hi @T04435 , just an FYI that this feature has been merged into the React, Angular, and Svelte packages

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests a new state-reset capability for `experimental_useObject`, rather than reporting behavior that violated an existing promise.

### Evidence
- The issue requests a `setObject` API, or automatic resetting of `object` to `undefined` when submission starts.
- The stated goal is to avoid displaying the previous object while a new object is generated.
- A later comment confirms that this capability was implemented and merged for React, Angular, and Svelte.

PR #7272, opened 2025-07-14
feat (ui): Add ability to "reset" to structured object

andrico1234:
## Background

I'm building an application, and I want a way to clear the existing object without reinitialising the StructuredObject class or submitting the object.

This is a wanted feature if the thumbs up on this [GH issue](https://github.com/vercel/ai/issues/2397) are any indication.

For me, I'm using `StructuredObject` instead of `Chat` to manage a conversation. It means that I'm manually managing the messages logic. 

Once the object finishes streaming, I want to clear it and append it to the list of messages.

## Summary

Added a new field to the object that allows us to reset the object.

I also spotted a bug where the error created via the AbortController was being returned falsy when pass through to `isAbortError`

## Verification

I added some tests, and everything passes.

## Tasks


.changeset/pretty-houses-whisper.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/provider-utils': patch
```
andrico1234, line 2: Happy to change this, not sure if this is the correct approach for beta packages

packages/provider-utils/src/is-abort-error.ts
```diff
@@ -1,6 +1,6 @@
 export function isAbortError(error: unknown): error is Error {
   return (
-    error instanceof Error &&
+    (error instanceof Error || error instanceof DOMException) &&
```
andrico1234, line 3: Please let me know if I'm missing context, since the AbortError error created in the structured-object.svelte instantiates a DOMException not Error, so `isAbortError` would return false.

Happy to change this if there's a better way

lgrammel:
this would need to be implemented for all 4 supported frameworks

andrico1234:
> this would need to be implemented for all 4 supported frameworks

Not a problem, I'd be happy to do that. 

Could I get a review on my approach beforehand? Just to ensure that the approach I've taken here will be accepted if I were to take the time to implement on the other frameworks.

packages/provider-utils/src/is-abort-error.ts
lgrammel: are there envs where DOMException is not available and this could cause errors?

lgrammel (approved):

lgrammel:
approach looks good!

packages/provider-utils/src/is-abort-error.ts
andrico1234: I just ran the `new DOMException()` in both the Node and Bun REPL and it behaved as expected. So should be fine across envs

andrico1234:
Thanks @lgrammel ! I'll resolve the merge conflicts. 

Do you want me to continue working on this branch and implement the feature across all frameworks, or should I open different PRs for each framework. (I'm happy for either approach, but the former will make it easier to group the changes and monitor the changes.

lgrammel:
ideally all in this pr

andrico1234:
@lgrammel 

Just to confirm, I'll be implementing the following frameworks
- [x] Svelte
- [x] React
- [x] Angular

[The Solid package no longer exists](https://github.com/vercel/ai/tree/main/examples/solidstart-openai)
[The Vue package doesn't have a working implementation of `useObject`](https://github.com/vercel/ai/tree/main/packages/vue)

You mentioned 4 frameworks. Is there another one that I've missed?

lgrammel:
Ah yes that makes sense, forgot that vue does not have useObject

andrico1234:
@lgrammel , I added the features and included tests. 

Do you have any suggestions for ways that I can validate that everything works as expected?

andrico1234:
I might change the name of the function to `clear` instead of `reset` as reset may imply that the object's value gets set to the initial value, which is not what the function does. What do you think?

lgrammel:
I like `clear` better, agreed. For validation I usually add/extend an example under e.g. `examples/next-openai`, `examples/sveltekit-openai` etc

andrico1234:
Great, here it is working in [React](https://www.loom.com/share/d34aa10e1fae41ddad214157dd65696f) and [Svelte](https://www.loom.com/share/6aa8eb96164e409a85aaf559d21bbd89)!

I think this PR is ready to go in :)

lgrammel (approved):

PR #7306, opened 2025-07-15
Enable auto completion and type safety for `providerOptions` without having to use `satisfies`

remorses:
## Background

Currently `providerOptions` needs to be typed manually with `satisfies`, this requires additional work from the user which has to import the provider type and then add a satisfies. 

## Summary

This PR uses `declare` to extend the type of `providerOptions` with the current provider options type.

The only thing the user needs to do is import the provider package and the provider options will be automatically extended.

This enables auto completion and type safety without any work from the user.

## Verification

Auto completion now works in the exaples for the provider openapi.


remorses:
To get the most out of this approach it would make sense to split the types of `providerOptions` between `embed`, `transcribe`, `generateImage`, etc

This way each function providerOptions will have the right type and there is no need to merge all the different provider option types using `&`

lgrammel:
NIce idea! currently lots of ts errors

lgrammel:
also yes the shared nature and different kind of models make this problematic. language model v2 spec is fixed at this point but this could be great for v2.1 of the spec

remorses:
What do you mean with language model v2 spec is fixed? Why is 2.1 needed to use a different type for non chat providerOptions?

lgrammel:
language model spec changes are highly breaking, esp for community providers (see all the ts errors on this pr), so we want to avoid them. language model v2 was freezed (except for major bugs) after ai sdk 5 alpha

Issue #2453, opened 2024-07-29
Ungraceful failure error when the anthropic api is overloaded


TechNickAI:
### Description

When anthropic api is busy/down, it responds with 

{"type":"error","error":{"details":null,"type":"overloaded_error","message":"Overloaded"}}

Which causes an unhandled rejection 

unhandledRejection: TypeValidationError [AI_TypeValidationError]: Type validation failed: Value: {"type":"error","error":{"details":null,"type":"overloaded_error","message":"Overloaded"}}.
Error message: [
  {
    "code": "invalid_union_discriminator",
    "options": [
      "message_start",
      "content_block_start",
      "content_block_delta",
      "content_block_stop",
      "message_delta",
      "message_stop",
      "ping"
    ],
    "path": [
      "type"
    ],
    "message": "Invalid discriminator value. Expected 'message_start' | 'content_block_start' | 'content_block_delta' | 'content_block_stop' | 'message_delta' | 'message_stop' | 'ping'"
  }
]
    at safeValidateTypes (/Users/nick/src/cora-morphic/.next/server/chunks/ssr/node_modules_d8329a._.js:2811:20)
    at safeParseJSON (/Users/nick/src/cora-morphic/.next/server/chunks/ssr/node_modules_d8329a._.js:2856:16)
    ... 5 lines matching cause stack trace ...
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  cause: ZodError: [
    {
      "code": "invalid_union_discriminator",
      "options": [
        "message_start",
        "content_block_start",
        "content_block_delta",
        "content_block_stop",
        "message_delta",
        "message_stop",
        "ping"
      ],
      "path": [
        "type"
      ],
      "message": "Invalid discriminator value. Expected 'message_start' | 'content_block_start' | 'content_block_delta' | 'content_block_stop' | 'message_delta' | 'message_stop' | 'ping'"
    }
  ]
      at get error (/Users/nick/src/cora-morphic/.next/server/chunks/ssr/node_modules_zod_lib_index_mjs_8e7072._.js:665:31)
      at safeValidateTypes (/Users/nick/src/cora-morphic/.next/server/chunks/ssr/node_modules_d8329a._.js:2813:41)
      at safeParseJSON (/Users/nick/src/cora-morphic/.next/server/chunks/ssr/node_modules_d8329a._.js:2856:16)
[...]

PR #7317, opened 2025-07-15
Fix parsing anthropic overloaded error

thdxr:
addresses #2453

it seems that for some reason the anthropic api is returning "details: null" in the error response for overloaded

```
{"type":"error","error":{"details":null,"type":"overloaded_error","message":"Overloaded"}}
```

this allows for parsing that properly and fixes retries being broken

### DIFF
--- a/packages/anthropic/src/anthropic-error.ts
+++ b/packages/anthropic/src/anthropic-error.ts
@@ -6,6 +6,7 @@ const anthropicErrorDataSchema = z.object({
   error: z.object({
     type: z.string(),
     message: z.string(),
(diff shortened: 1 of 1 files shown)


lgrammel:
this fix does not help with parsing. wrote a test that shows that the provided snippet parses fine https://github.com/vercel/ai/pull/7323 . can you provide a reproduction?

thdxr:
ah sorry about that thought i saw a .strict() call somewhere

let me try this again - hard to figure out because we can't force anthropic to produce an overloaded call

PR #7344, opened 2025-07-16
feat(ai): add provider and provider registry middleware functionality

jonaslalin:
## Background

The AI SDK needed a way to apply middleware to all language models accessed through providers and provider registries. Currently, middleware can only be applied to individual models, which becomes cumbersome when you want to apply the same middleware (e.g., logging, rate limiting, model ID overrides) to all models from a provider or registry.

## Summary

Added two new wrapper functions to enable middleware application at the provider and registry level:

- **`wrapProvider`**: Wraps a `ProviderV2` instance to apply middleware to all language models accessed through that provider
- **`wrapProviderRegistry`**: Wraps a `ProviderRegistryProvider` instance to apply middleware to all language models accessed through the registry

Both functions support:
- Single or multiple middleware chains
- Optional model ID and provider ID overrides
- Proper TypeScript typing with generics
- Only wrapping language models (other model types pass through unchanged)

## Verification

Added tests for the new wrapper functions.

## Tasks


jonaslalin:
@lgrammel please review.

I don't know if I have missed something with the types for the provider registry or the mocks. Because now you can pass any string for model, it only typechecks the provider key. But it seems to be the same case for the original provider registry too, so I don't think the wrapped version made any difference to the typing issue here.

image
and
image
which are similar in the string requirement for the model.

lgrammel:
what do you think about just adding the middleware to the provider registry and having it applied automatically in the registry?

that would be how i would design it. then you can pass languageModelMiddleware[] into the registry and dont need to worry. also simpler

jonaslalin:
sure we can do that, still i think i would need the provider middleware for people who dont use a registry? or you want to change that api too?

lgrammel:
provider wrapping could be useful. will review tmrw

jonaslalin:
the problem with that solution is that someone implementing their own provider registry, instead of the default one, would not get the middleware functionality for free, if we change directly there, compared to the current decorator solution which can be used as a building block for your own implementation. what do you think?

lgrammel:
someone implementing their own registry is an edge case. i do not expect many people to do this. if they have that level of skill, adding the wrapping should be easy for them in particular if it is in the default provider registry

lgrammel:
overall it looks good, reviewed. i would suggest moving the middlewares / wrapping into the registry as discussed

jonaslalin:
@lgrammel pushed changes

packages/ai/src/registry/provider-registry.ts
```diff
@@ -105,9 +129,29 @@ class DefaultProviderRegistry<
 {
   private providers: PROVIDERS = {} as PROVIDERS;
   private separator: SEPARATOR;
+  private languageModelMiddleware?:
+    | LanguageModelV2Middleware
+    | LanguageModelV2Middleware[];
+  private languageModelIdOverride?: string;
+  private languageModelProviderIdOverride?: string;
```
lgrammel: these 2 are strange because it would mean standardizing to a single provider/model name. are they needed or can they be removed?

packages/ai/src/registry/provider-registry.ts
```diff
@@ -171,15 +215,26 @@ class DefaultProviderRegistry<
...
+    ).languageModel?.(modelId);
 
-    if (model == null) {
+    if (originalModel == null) {
       throw new NoSuchModelError({ modelId: id, modelType: 'languageModel' });
     }
 
-    return model;
+    if (this.languageModelMiddleware != null) {
+      const wrappedModel = wrapLanguageModel({
```
lgrammel: inline variable (and keep `model` instead of `originalModel`)

packages/ai/src/test/mock-provider-v2.ts
```diff
@@ -0,0 +1,68 @@
+import {
```
lgrammel, line 1: worth exporting in the text index?

packages/ai/src/registry/provider-registry.ts
jonaslalin: yeah i agree, tried to be consistent with the wraplanguagemodel signature, but makes no sense in this case. will remove

packages/ai/src/registry/provider-registry.ts
jonaslalin: done

packages/ai/src/test/mock-provider-v2.ts
jonaslalin: i didnt find a test index file?

jonaslalin: aha, now i did, so done also

lgrammel (approved):

packages/ai/src/registry/provider-registry.ts
```diff
@@ -171,14 +199,21 @@ class DefaultProviderRegistry<
...
     }
 
+    if (this.languageModelMiddleware != null) {
+      model = wrapLanguageModel({
+        model,
+        middleware: this.languageModelMiddleware,
+      });
+    }
+
     return model;
```
lgrammel, line 217: would probably change this into const + ternary return but good to land unless you want to change this

jonaslalin: feel free to change that line if you want to :)

PR #7355, opened 2025-07-16
[draft] active chat context

iteratetograceness:
## Background

Minimal example in `next` uses `window.history.pushState` w/ a Next.js router cache invalidation which was polluting the navigation history + causing a reload after navigation. Also, by migrating away from SWR, we lost the benefit of having all hooks w/ the same id share the same state which was provided to us via the SWR global cache.

Unsolved:
- Resetting on `/` route

## Summary

A new minimal example added using a global state store (React Context, but could also use Zustand or Jotai) to illustrate solution for sharing active chat state across useChat instances. Can be tweaked to store a full chat registry instead of just the active chat.

This example also relies on `router.push` to avoid calling `window.history.pushState`.

## Verification

Tested new chat creation, navigation across chats, and back/forward navigation in development env

## Tasks

## Future Work

There is a **known bug** with this implementation that should be addressed as a follow up:

After starting a new chat by sending the first message from `/`, we route to `/chat/[chatId]`. After the assistant response streams in, when we navigate back to `/` and then return to `/chat/[chatId]`, the messages are stale.


lgrammel:
thank you! can you make this a separate example (dont want to override what is there since it is a good example in itself)?

lgrammel (approved):

hoanginc144:
@iteratetograceness Not sure if it helps, but regarding new chat flow and also navigating back to /chat, I solved it using React context to set the first message before navigating to /chat/id, then after router.push I use useEffect to send the first message. I also use a hasSentFirstMessage ref to make sure it only sends the first message once. Obviously the downside is now I have to call sendMessage within a useEffect, which I do not like, but so far it has worked well for my case.

PR #7359, opened 2025-07-16
feat (provider/gateway): Add AI Gateway provider options (ordering)

R-Taneja:
## Background

AI Gateway provider options were previously only documented but not supported in the package

## Summary

Added type-safe provider options support for the AI Gateway provider

## Tasks


packages/gateway/src/gateway-language-model.ts
```diff
@@ -38,11 +40,54 @@ export class GatewayLanguageModel implements LanguageModelV2 {
     return this.config.provider;
   }
 
+  private async prepareRequestBody(
+    body: any,
+    providerOptions: any,
```
shaper: avoid `any` throughout PR, `unknown` if needed.
I think this can be:
```
body: LanguageModelV2CallOptions,
providerOptions: Record<string, unknown> | undefined;
```

packages/gateway/src/gateway-language-model.ts
```diff
@@ -38,11 +40,54 @@ export class GatewayLanguageModel implements LanguageModelV2 {
     return this.config.provider;
   }
 
+  private async prepareRequestBody(
```
shaper: From looking at the Anthropic provider I think the pattern is to process the provider options in `getArgs` which is shared logic across both generate and stream. See:

https://github.com/vercel/ai/blob/e151efdab8ce7597a7f477170bdf526c828c460a/packages/anthropic/src/anthropic-messages-language-model.ts#L246

and then the way developers get type safety is through importing the exported provider options type and using `satisfies GatewayProviderOptions` on their `providerOptions` object.

GatewayLanguageModel hasn't had a `getArgs` yet but this might be the point where we should create one, to follow standard naming, vs. `prepareRequestBody`.

packages/gateway/src/gateway-language-model.ts
```diff
@@ -38,11 +40,54 @@ export class GatewayLanguageModel implements LanguageModelV2 {
...
+    body: any,
+    providerOptions: any,
+  ): Promise<any> {
+    const gatewayOptions = await parseProviderOptions({
+      providerOptions,
+      provider: 'gateway',
+      schema: gatewayProviderOptions,
+    });
+
+    const baseRequestBody = this.maybeEncodeFileParts(body);
```
shaper: the result here could be large if we encode a file, so we should avoid copying the object if possible below, e.g. mutate the object directly instead of using `{ ...etc }`.

A possible example approach:

```
 if (gatewayOptions || requestBody.providerOptions?.gateway) {
    const { gateway, ...otherProviderOptions } = requestBody.providerOptions || {};
    
    const newProviderOptions = {
      ...otherProviderOptions,
      ...(gatewayOptions && { gateway: gatewayOptions }),
    };
    
    if (Object.keys(newProviderOptions).length > 0) {
      requestBody.providerOptions = newProviderOptions;
    } else {
      delete requestBody.providerOptions;
    }
  }
  
  return requestBody;
}
```

But now that we're looking at this, I am not really sure we need to do this. The other AI SDK language models do this so that they can enact logic around the options specified, but for Gateway we would actually be fine just passing whatever the user specified on through to the server, where we'll need to parse/validate and scrutinize like this again anyway.

So from a minimize code complexity perspective I think, unless we get other feedback, perhaps we can remove the edits to the language model, pass straight through whatever the user has specified, and we have enough with just the added `gateway-provider-options.ts` and exporting that type for `satisfies`.

.changeset/stupid-wasps-pump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+Added support for Gateway options (ordering)
```
shaper: general naming pattern for changeset note is to use the PR description which is formatted as I've edited slightly now, `feat (provider/foo): bar` or sometimes `chore`, `fix`, etc.

packages/gateway/turbo.json
shaper, line 1: revert this change

packages/gateway/src/gateway-language-model.ts
```diff
@@ -38,11 +39,24 @@ export class GatewayLanguageModel implements LanguageModelV2 {
     return this.config.provider;
   }
 
+  private async getArgs(options: Parameters<LanguageModelV2['doGenerate']>[0]) {
+    const { abortSignal, ...optionsWithoutSignal } = options;
+    const args = this.maybeEncodeFileParts(optionsWithoutSignal);
```
shaper: inline this below so:
```
return {
  args: this.maybeEncodeFileParts(optionsWithoutSignal),
  warnings: [],
}
```

packages/gateway/src/gateway-language-model.ts
```diff
@@ -107,6 +124,11 @@ export class GatewayLanguageModel implements LanguageModelV2 {
             ParseResult<LanguageModelV2StreamPart>,
             LanguageModelV2StreamPart
           >({
+            start(controller) {
```
shaper, line 126: Is this a bug fix/add based on what you saw looking at other providers/language models?

packages/gateway/src/gateway-provider-options.ts
shaper, line 1: add an example of use of these options under a new script `examples/ai-core/src/stream-text/gateway-provider-options-order.ts`
you can follow the existing `gateway.ts` script we have under there for the basic logic

allenzhou101:
@R-Taneja Nice stuff! @shaper is much more familiar with the sdk so I'll leave most of the review to him, but maybe another good addition is to include the `GatewayProviderOptions` usage in the docs similar to the [Anthropic AI SDK Provide Options section](https://v5.ai-sdk.dev/providers/ai-sdk-providers/anthropic#reasoning).

Screenshot 2025-07-18 at 11 55 48 AM

Maybe in the [Vercel docs](https://vercel.com/docs/ai-gateway/provider-options#available-providers) and [AI SDK Docs](https://v5.ai-sdk.dev/providers/ai-sdk-providers/ai-gateway).

packages/gateway/src/gateway-language-model.ts
R-Taneja: Yes! I noticed many other `[provider]-language-model.ts` files had it

R-Taneja:
> @R-Taneja Nice stuff! @shaper is much more familiar with the sdk so I'll leave most of the review to him, but maybe another good addition is to include the `GatewayProviderOptions` usage in the docs similar to the [Anthropic AI SDK Provide Options section](https://v5.ai-sdk.dev/providers/ai-sdk-providers/anthropic#reasoning).
> 
> Screenshot 2025-07-18 at 11 55 48 AM
> Maybe in the [Vercel docs](https://vercel.com/docs/ai-gateway/provider-options#available-providers) and [AI SDK Docs](https://v5.ai-sdk.dev/providers/ai-sdk-providers/ai-gateway).

Gateway provider options is already documented with its own page, so I added a link out to it from the AI SDK docs

packages/gateway/src/gateway-provider-options.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from 'zod/v4';
+
+// https://vercel.com/docs/ai-gateway/provider-options
+export const gatewayProviderOptions = z.object({
+  /**
+   * Array of provider slugs that specifies the sequence in which providers should be attempted.
+   *
+   * Example: ['bedrock', 'anthropic'] will try Amazon Bedrock first, then Anthropic as fallback.
```
shaper: nit: wrap the code example snippet in backticks so it formats well in tooltips/doc rendering.

shaper (approved):

PR #7361, opened 2025-07-17
chore (provider/openai): standardize on itemId in provider metadata

lgrammel:
## Background

We have `itemId` on tool and text parts. However, reasoning parts use `reasoning.id`. We need to standardize now to avoid confusion and larger migrations in the future.

## Summary

* Change `reasoning.id` to `itemId` in the provider metadata to standardize on `itemId`
* Change `reasoning.encryptedContent` to `reasoningEncryptedContent` to flatten object

## Verification

Ran `examples/ai-core/src/stream-text/openai-responses-reasoning-zero-data-retention.ts`

### DIFF
--- a/.changeset/soft-forks-poke.md
+++ b/.changeset/soft-forks-poke.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+chore (provider/openai): standardize on itemId in provider metadata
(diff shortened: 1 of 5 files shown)


lgrammel:
cc @jonaslalin heads up on this change

jonaslalin:
@lgrammel I guess it makes sense to rename `reasoning: { encryptedContent: ..` to `reasoningEncryptedContent` also then? so it is flat?

PR #7376, opened 2025-07-17
feat(ai): implement wrapGenerateObject middleware for language models

jonaslalin:
- Added `wrapGenerateObject` function to create middleware wrappers for language model generation.
- Introduced tests for the middleware functionality to ensure correct behavior with temperature settings.
- Enhanced compatibility with Zod schemas for input validation.

## Tasks

### DIFF
--- a/packages/ai/src/middleware/wrap-generate-object.test.ts
+++ b/packages/ai/src/middleware/wrap-generate-object.test.ts
@@ -25,10 +25,7 @@ describe('wrapGenerateObject', () => {
       },
     });
 
-    const wrappedGenerateObject = wrapGenerateObject<{
-      model: LanguageModelV2;
-      hot?: boolean;
-    }>({
+    const wrappedGenerateObject = wrapGenerateObject<{ hot?: boolean }>({
       middleware: ({ options, doGenerateObject }) => {
         const { hot, ...baseOptions } = options;
         let temperature = baseOptions.temperature;
--- a/packages/ai/src/middleware/wrap-generate-object.test.ts
+++ b/packages/ai/src/middleware/wrap-generate-object.test.ts
@@ -1,0 +1,56 @@
+import { LanguageModelV2 } from '@ai-sdk/provider';
+import { z } from 'zod/v4';
+import { MockLanguageModelV2 } from '../test/mock-language-model-v2';
+import { wrapGenerateObject } from './wrap-generate-object';
+
+const dummyResponseValues = {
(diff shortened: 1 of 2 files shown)


jonaslalin:
Hello @lgrammel . Before I add wrapper methods for streamObject, generateText, streamText, I want your opinion on the solution. Basically, I want to be able to create my own version of these functions, add extra parameters, but reuse the existing one, plus modify the behavior by for example adding languagemodel middleware using these extra option values. What do you think about the pattern?

lgrammel:
You do not need this and wrapping should be independent of the ai functions. I feel strongly about not adding this to the framework to keep the wrapping / language models independent from the ai functions.

lgrammel:
Do you have concrete use cases that show the benefit?

lgrammel:
(one of the challenges with a library is to keep the API surface limited because we need to maintain, document, provide examples, support etc) - this seems better solved in application land / user code

jonaslalin:
Yes. I want to decorate these functions with extra parameters and logic, and it is a mess with the types, so I thought it would help others to have this possibility. My concrete use case is to expose for example generateText to other developers (we are a big team), forcing them to pass usage analytics parameters. Then the middleware automatically adds and configures an analytics service and attaches it to whatever languagemodel the developer is using. Lint rules then forbid the import from vercel ai sdk, but instead to these decorated functions. The developer can use the vercel ai sdk as usual, but is required to pass an extra object that configures everything and keeps everything dry. Every call needs a new set of parameters which is why the provider registry or provider doesnt work.

lgrammel:
My preference is to leave this outside of the ai sdk for now, especially since we are getting close to the v5 release. Might be worth revisiting in a few months if there are similar requests from the community

jonaslalin:
Okey, anyone who wants to decorate any of these functions with extra parameters and logic will need to rely on pretty deep typescript skills to be able to pull that task off. That is why i thought i could share the code i already had to write to make it work so others could modify the behavior if they had to as well. The wrapper is a swiss army knife and you can expand and change the underlying api in a flexible way, while keeping type safety and correct type inference.

lgrammel:
Not disagreeing, they may chime in on the PR or open separate issues. This is currently more of an API surface and timing issue, I prefer to wait and see more evidence/requests.

jonaslalin:
Okey, do you want it to stay open or should I close it?

lgrammel:
your call either way works for me

PR #7385, opened 2025-07-17
feat(ai): infer tool call types for the `onToolCall` callback

damianstasik:
## Background

This PR builds on top of the great improvements you made in v5 and extends UI tool inference to the `onToolCall` callback.

## Summary

Currently, the `onToolCall` callback does not infer any information about available tools, so handling tool calls on the client side doesn’t get the same nice type support as on the server side. This PR makes a small adjustment to the types to infer the possible tool calls and their inputs.

image

image

Let me know if the PR makes sense, I'll gladly add some tests and the changeset.

## Verification

I checked a few examples that define tools and tried a few of my own, using both the `InferUITools` utility type and the raw `{ input: ...; output: ... }` form. I also ran `pnpm run type-check:full` to make sure there are no type-related issues.

## Tasks

### DIFF
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -2,7 +2,7 @@ import {
   generateId as generateIdFunc,
   IdGenerator,
   StandardSchemaV1,
-  ToolCall,
+  Tool,
   Validator,
 } from '@ai-sdk/provider-utils';
 import { consumeStream } from '../util/consume-stream';
@@ -20,6 +20,7 @@ import {
   shouldResubmitMessages,
 } from './should-resubmit-messages';
 import {
+  InferUIMessageToolCall,
   isToolUIPart,
   type DataUIPart,
   type FileUIPart,
@@ -93,11 +94,10 @@ export interface ChatState<UI_MESSAGE extends UIMessage> {
 
 export type ChatOnErrorCallback = (error: Error) => void;
 
-export type ChatOnToolCallCallback = ({
-  toolCall,
-}: {
-  toolCall: ToolCall<string, unknown>;
-}) => void | Promise<unknown> | unknown;
+export type ChatOnToolCallCallback<UI_MESSAGE extends UIMessage = UIMessage> =
+  (options: {
+    toolCall: InferUIMessageToolCall<UI_MESSAGE>;
+  }) => void | Promise<unknown> | unknown;
 
 export type ChatOnDataCallback<UI_MESSAGE extends UIMessage> = (
   dataPart: DataUIPart<InferUIMessageData<UI_MESSAGE>>,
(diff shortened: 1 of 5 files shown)


packages/ai/src/ui/chat.ts
```diff
@@ -93,11 +94,10 @@ export interface ChatState<UI_MESSAGE extends UIMessage> {
 
 export type ChatOnErrorCallback = (error: Error) => void;
 
-export type ChatOnToolCallCallback = ({
-  toolCall,
-}: {
-  toolCall: ToolCall<string, unknown>;
-}) => void | Promise<unknown> | unknown;
+export type ChatOnToolCallCallback<UI_MESSAGE extends UIMessage = UIMessage> =
```
damianstasik, line 98: I've included the default value here to avoid breaking changes for users who might be using `ChatOnToolCallCallback`.

packages/ai/src/ui/process-ui-message-stream.ts
```diff
@@ -350,7 +351,7 @@ export function processUIMessageStream<UI_MESSAGE extends UIMessage>({
               // Skip calling onToolCall for provider-executed tools since they are already executed
               if (onToolCall && !chunk.providerExecuted) {
                 const result = await onToolCall({
-                  toolCall: chunk,
+                  toolCall: chunk as InferUIMessageToolCall<UI_MESSAGE>,
```
damianstasik, line 358: The `UIMessageChunk` type (passed to the `TransformStream`) doesn’t take any arguments and is used in several places across the codebase. Because of this, I opted for a simpler approach and just did a basic type cast.

packages/ai/src/ui/chat.ts
```diff
@@ -93,11 +94,10 @@ export interface ChatState<UI_MESSAGE extends UIMessage> {
...
+  (options: {
+    toolCall: InferUIMessageToolCall<UI_MESSAGE>;
+  }) => void | Promise<unknown> | unknown;
```
lgrammel: is there an easy way to type the output as well, i.e. replace unknown here?

lgrammel:
please add a changeset.

how can this be tested?

packages/ai/src/ui/chat.ts
damianstasik: @lgrammel I did a bit of experimenting yesterday, and it seems the easiest approach is to use a union type of all possible outputs as the return type. I was hoping TypeScript would support something more precise where a specific tool branch could enforce a matching return type, but that doesn't seem to be possible at the moment. I've pushed the change with the union type, please let me know if that's all right or whether I should revert it.

damianstasik:
> please add a changeset.

Done!

> how can this be tested?

I added a new test that focuses just on types, using Vitest's `toMatchTypeOf` and `toEqualTypeOf` to check the inferred type of `toolCall` in `onToolCall`, as well as its return type. I tried to cover a few scenarios like with and without helpers, without an output schema, and cases with slightly more complex types. Let me know if that makes sense or if you'd like me to change anything, happy to do so!

lgrammel (approved):

PR #7387, opened 2025-07-17
fix(ai-sdk/react) Add options.id to subscribeToMessages callback dependency array

### LINKED ISSUE #7378: Streaming not working on first load / hard refresh (fixes after fast-refresh) on local dev
### Description

Text Streaming is not working on first page load / hard refresh on nextjs in development mode.
If I send a message asking for a long response after a hard refresh, the network streaming tab shows the chunks coming in, but the `messages` array from useChat only updates with the first step-start and state: streaming, and then at the very end with the final message.

If I leave the page open and cause an update with Fast Refresh, then streaming works properly until I reload the page.

If the model makes tool calls before responding, the first tool call gets parsed in the messages array with "input-streaming". Then streaming stops being parsed, and the next update to `messages` contains the full tool call, it's output, any other tool calls and the final text response with state: "done".

Streaming works as expected after a hot/fast reload (If I add a console.log and then save)

Tested on Chrome, Safari, Arc

### AI SDK Version

ai-sdk/react: 2.0.0-beta.22
ai: 5.0.0-beta.22
next: 15.3.4

Potrock:
## Background

When useChat is mounted with id=undefined and the ID is updated later (e.g., from a backend fetch), streaming responses fail to display in real-time. The UI only updates after the entire response is received, even though the network shows chunks arriving correctly.

## Summary

Fixed stale subscription callbacks in useChat hook by adding the chat ID (if available) to the dependency array of the messages subscription callback.

## Verification

Built and tested locally in my app

## Tasks

## Related Issues

Fixes #7378

### DIFF
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -70,7 +70,7 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
   }
 
   const optionsId = 'id' in options ? options.id : null;
-  
+
   const subscribeToMessages = useCallback(
     (update: () => void) =>
       chatRef.current['~registerMessagesCallback'](update, throttleWaitMs),
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -69,10 +69,14 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
     chatRef.current = 'chat' in options ? options.chat : new Chat(options);
   }
 
+  const optionsId = 'id' in options ? options.id : null;
+  
   const subscribeToMessages = useCallback(
     (update: () => void) =>
       chatRef.current['~registerMessagesCallback'](update, throttleWaitMs),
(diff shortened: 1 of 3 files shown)


packages/react/src/use-chat.ts
```diff
@@ -72,7 +72,7 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
   const subscribeToMessages = useCallback(
     (update: () => void) =>
       chatRef.current['~registerMessagesCallback'](update, throttleWaitMs),
-    [throttleWaitMs],
+    [throttleWaitMs, 'id' in options ? options.id : null],
```
iteratetograceness: q: could we address this lint error? after fixing, we'll likely need get another, different lint error (unnecessary dependency) that we should disable w/ a comment on why

Potrock: Yes fixed, sorry I missed the warning! Added a comment and suppression on the exhaustive dep error.

lgrammel (approved):

PR #7414, opened 2025-07-20
feat: export fixJson from ai

BrianHung:
Exports `fixJSON` util. 

For context, I'm trying to only use the `fixJSON` method to repair a custom tool call. `parsePartialJson` is async so I didn't opt for that.

### DIFF
--- a/packages/ai/src/util/index.ts
+++ b/packages/ai/src/util/index.ts
@@ -8,3 +8,4 @@ export { isDeepEqualData } from './is-deep-equal-data';
 export { parsePartialJson } from './parse-partial-json';
 export { SerialJobExecutor } from './serial-job-executor';
 export { simulateReadableStream } from './simulate-readable-stream';
+export { fixJson } from './fix-json';


lgrammel:
fixJson is a misleading name unfortunately, so this would need some refinement before exposing it in a public api.

tool repair supports async functions, is there a particular reason why you dont want to use parsePartialJson?

BrianHung:
> tool repair supports async functions, is there a particular reason why you dont want to use parsePartialJson?

Oh, the custom tool repair logic I'm using is sync, so I'm validating the schema synchronously and would also want to fix the json synchronously. `parsePartialJson` is async so I would have to refactor my consumers.

lgrammel:
parsePartialJson is async to support async schema validation (e.g. Zod refineAsync). my recommendation would be to make that code path async instead, so you can easily switch in the future.

BrianHung:
Is fixJson intended to be used independently, or is it meant to be an internal utility that's only used by parsePartialJSON?

I have an alternative way of validating the schema (or skipping validation entirely), so I only need the fixing logic not the parsing logic. If it's expected to be internal only, then I can switch over to parsePartial.

lgrammel:
Yes that would be my preference. fix-json implies more than it does, see e.g. https://www.npmjs.com/package/jsonrepair for advanced solutions

PR #7423, opened 2025-07-20
feat: locale aware chunking in smoothStream via `Intl.Segmenter`

Rajaniraiyn:
## Background

Currently `smoothStream` uses regex based chunking, but for many languages that doesn't use space delimiter it doesn't work well and we need to manually provide chunks. This often times performant and robust then regex based approaches and unlocks new chucking strategies (grapheme and sentence). This API is also [widely supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter#browser_compatibility) for sometime now.

## Summary

Added new chunking options (word-intl, grapheme and sentence) using `Intl.Segmenter` and also added locale specific options via `segmenterOptions` option.

```ts
smoothStream({
  delayInMs: 10,
  chunking: 'word-intl',
  segmenterOptions: { locale: 'ja' },
})

smoothStream({
  chunking: 'grapheme',
})
```

## Tasks

## Future Work

`Intl.Segmenter` can be made default chunking instead of regex based chunking and overall API can be improved targeting DX

### DIFF
--- a/.changeset/grumpy-falcons-collect.md
+++ b/.changeset/grumpy-falcons-collect.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+feat(ai): locale-aware segmentation for `smoothStream`
--- a/content/docs/07-reference/01-ai-sdk-core/80-smooth-stream.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/80-smooth-stream.mdx
@@ -42,10 +42,17 @@ const result = streamText({
     },
     {
       name: 'chunking',
-      type: '"word" | "line" | RegExp | (buffer: string) => string | undefined | null',
+      type: '"word" | "line" | "grapheme" | "word-intl" | "sentence" | RegExp | (buffer: string) => string | undefined | null',
       isOptional: true,
       description:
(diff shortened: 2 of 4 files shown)


lgrammel:
Thanks, I am very weary about mixing regexp and intl segmenter and having things such as word.intl

smoothStream supports a chunk detector:

```
/**
 * Detects the first chunk in a buffer.
 *
 * @param buffer - The buffer to detect the first chunk in.
 *
 * @returns The first detected chunk, or `undefined` if no chunk was detected.
 */
export type ChunkDetector = (buffer: string) => string | undefined | null;
```

it should be possible to use Intl.Segmenter that way without changing smooth stream.

Can you confirm? ideally then the change would be limited to the documentation for now

Rajaniraiyn:
yes, we could do that. I felt its better to have the built-in word and line based chunking to use language agnostic Intl.Segmenter api.

```ts
const segmenter = new Intl.Segmenter('ja', { granularity: 'word' });

const intlChunker = (buffer: string) => {
    const segments = Array.from(segmenter.segment(buffer));
    if (segments.length === 0) return null;
    const { segment } = segments[0];
    return segment.length ? segment : null;
};
```

can create a separate PR just updating the docs o make a PR with Intl.Segmentor as default chunk detector for line and word options

aayush-kapoor:

PR #7435, opened 2025-07-21
feat(provider/openai-compatible):change providerOptions tools\ tool_choice

MyPrototypeWhat:
## Background

Tools specified in `providerOptions[providerName].tools` are being overridden by the standard `tools` parameter in `@ai-sdk/openai-compatible`, making it impossible to pass provider-specific tool configurations.

## Summary

**Fix:** Extract `tools` and `tool_choice` from `providerOptions` before spreading to prevent override conflicts.

- Use `providerOptions.tools` when present, fallback to standard tools otherwise
- Add warning when both tool types are specified
- Maintain full backward compatibility

```typescript
// Before: providerOptions.tools was always overridden
{
  ...providerOptions?.[this.providerOptionsName], // includes tools
  tools: openaiTools, // ❌ overwrites provider tools
}

// After: conditional tool selection
tools: hasProviderTools ? providerTools : openaiTools,
tool_choice: providerToolChoice ?? openaiToolChoice,
```

## Impact for Third-Party Providers

This change is crucial for third-party provider packages built on `@ai-sdk/openai-compatible`:

**Before this fix:**
- Third-party providers couldn't expose provider-specific tools (like web search, computer use)
- Aggregation platforms couldn't handle different tool formats per provider
- Provider packages had to choose between standard OpenAI tools OR provider-specific tools, not both

**After this fix:**
- **Provider packages** can now offer both standard and provider-specific tools
- **Aggregation platforms** can route different tool formats to appropriate providers

**Example for third-party providers:**
```typescript
// Provider package can now support both:
const result = streamText({
  model: customProvider.chatModel('model'),
  tools: [standardWeatherTool], // Standard tools work
  providerOptions: {
    custom: {
      tools: [{ type: 'provider_search', config: {...} }] // Provider tools also work
    }
  }
});
```

## Verification

Tested with various tool configurations:
- Standard tools only ✅
- Provider tools only ✅  
- Both present (with warning) ✅
- Empty/null provider tools ✅
- Streaming mode ✅

## Tasks


.changeset/unlucky-foxes-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': major
```
lgrammel: change to patch

lgrammel:
The solution with provider options seems wrong to me. We have provider defined tools in the AI SDK. Is there a way to solve this without providerOptions?

MyPrototypeWhat:
> The solution with provider options seems wrong to me. We have provider defined tools in the AI SDK. Is there a way to solve this without providerOptions?

Thanks for the feedback!
Some third-party proxy platforms (outside official/community providers) only work through @ai-sdk/openai-compatible, and their tools format can differ from the standard.
Currently, providerOptions.tools gets overridden by the request’s tools, so provider-specific tools can’t be preserved.
I used providerOptions here as the smallest, backward-compatible fix.
If you’d prefer a dedicated merge mechanism in the core, I’m happy to adjust this PR in that direction.

PR #7444, opened 2025-07-21
feat (provider/gateway): Add OpenAI embedding support

R-Taneja:
## Background

With this [PR](https://github.com/vercel/ai-gateway/pull/197), the AI Gateway will support embeddings calls (initially with just OpenAI).

## Summary

This PR adds embedding support on the AI SDK client-side for the Gateway provider

## Verification

Created and ran an example file, it returns embeddings

## Tasks


examples/ai-core/src/embed-many/gateway.ts
```diff
@@ -0,0 +1,21 @@
+import { createGatewayProvider } from '@ai-sdk/gateway';
```
shaper: can just do:

```
```

and rm line 5

examples/ai-core/src/embed/gateway.ts
```diff
@@ -0,0 +1,17 @@
+import { createGatewayProvider } from '@ai-sdk/gateway';
```
shaper: same as above -- `import { gateway } ...` and rm line 5

packages/gateway/src/gateway-provider.ts
```diff
@@ -30,6 +35,11 @@ Creates a model for text generation.
 Returns available providers and models for use with the remote provider.
  */
   getAvailableModels(): Promise<GatewayFetchMetadataResponse>;
+
+  /**
+Creates a model for generating text embeddings.
+*/
+  textEmbeddingModel(modelId: string): EmbeddingModelV2<string>;
```
shaper: I think we will want to define a new `gateway-embedding-options.ts` file which exports `GatewayEmbeddingModelId` and use that type for the `modelId` parameter here, so that we get model id autocomplete for the appropriate set of embedding model ids.

packages/gateway/src/gateway-provider.ts
```diff
@@ -173,7 +183,34 @@ export function createGatewayProvider(
   };
   provider.languageModel = createLanguageModel;
   provider.textEmbeddingModel = (modelId: string) => {
-    throw new NoSuchModelError({ modelId, modelType: 'textEmbeddingModel' });
+    const deploymentId = loadOptionalSetting({
```
shaper: I know you may have had it as shared previously and reworked to inline -- reading it now, enough of this is the same as above that I think it's worth moving to a common function, something like how common config is done here  https://github.com/vercel/ai/blob/main/packages/vercel/src/vercel-provider.ts#L68

So a peer function at the same level as getHeaders which contains roughly lines 186 - 197 and 200 - 212 and it's used both here and at line 130 above.

packages/gateway/src/gateway-provider.ts
shaper, line 1: Update unit tests to reflect new logic

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  private getUrl() {
+    return `${this.config.baseURL}/embedding-model`;
+  }
+
+  private getModelConfigHeaders() {
+    return {
+      'ai-embedding-model-specification-version': '2',
+      'ai-embedding-model-id': this.modelId,
+      'ai-embedding-model-streaming': 'false',
```
shaper: We can omit streaming since it isn't needed for embeddings -- if we need to edit server logic to reflect let's do that.

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -0,0 +1,91 @@
...
+import { z } from 'zod/v4';
+import type { GatewayConfig } from './gateway-config';
+import { asGatewayError } from './errors';
+import { parseAuthMethod } from './errors/parse-auth-method';
+
+export class GatewayEmbeddingModel implements EmbeddingModelV2<string> {
+  readonly specificationVersion = 'v2';
+  readonly supportedUrls = { '*/*': [/.*/] } as const;
+  readonly maxEmbeddingsPerCall = 2048;
+  readonly supportsParallelCalls = true;
```
shaper, line 18: mistral is an example where this is false -- let's test it and see how we behave, and consider vs. what we think a developer would want to happen.

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -0,0 +1,91 @@
...
+  type Resolvable,
+} from '@ai-sdk/provider-utils';
+import { z } from 'zod/v4';
+import type { GatewayConfig } from './gateway-config';
+import { asGatewayError } from './errors';
+import { parseAuthMethod } from './errors/parse-auth-method';
+
+export class GatewayEmbeddingModel implements EmbeddingModelV2<string> {
+  readonly specificationVersion = 'v2';
+  readonly supportedUrls = { '*/*': [/.*/] } as const;
```
shaper: I think we don't need this for embedding models?

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -0,0 +1,91 @@
...
+  readonly maxEmbeddingsPerCall = 2048;
```
shaper, line 17: bedrock is an example where this is 1 -- let's test it and see how we behave, and consider vs. what we think a developer would want to happen.

packages/gateway/src/gateway-embedding-model.ts
shaper, line 1: For the two properties I commented on below re: testing, what's available/returned property-wise here will be overridden by what's present in the model server side when instantiated.

The ai core `embed-many.ts` records the two cited properties in an otel span and uses `maxEmbeddingsPerCall` and `supportsParallelCalls` to split into chunks, and so if they differ from what ends up being specified server-side, the intended behavior could drift.

Let's look into this more outside of PR.

shaper, line 1: Add unit tests for this file

shaper (commented):
Comments while reading, I know it is still draft.

packages/gateway/src/gateway-embedding-model.ts
R-Taneja: A bit unclear on the implications of this. Is it still an issue after the server-side changes we made? Is this something to ITG on after this PR?

packages/gateway/src/gateway-embedding-model.ts
R-Taneja: It errors: "GatewayInternalServerError: Too many values for a single embedding call. The amazon-bedrock model "amazon.titan-embed-text-v1" can only embed up to 1 values per call, but 2 values were provided."

This isn't the best DX and isn't super Gateway-esque. I'm attempting to implement some kind of chunking solution to abstract that away from developers.

packages/gateway/src/gateway-embedding-model.ts
R-Taneja: Fixed in https://github.com/vercel/ai-gateway/pull/216

packages/gateway/src/gateway-embedding-model.ts
R-Taneja: Fixed in https://github.com/vercel/ai-gateway/pull/216

examples/ai-core/src/embed-many/gateway.ts
```diff
@@ -0,0 +1,19 @@
...
+    values: [
+      'sunny day at the beach',
+      'rainy afternoon in the city',
+      'snowy night in the mountains',
+    ],
+  });
+
+  console.log(embeddings);
+  console.log(usage);
+}
```
shaper, line 17: nit: please log providerMetadata as well with `JSON.stringify(..., null, 2)` so that we also show routing detail (and include in the embed example as well)

packages/gateway/src/gateway-embedding-model.ts
shaper: I think there may be more work here, but should be ok to land this and look at it more closely following.

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  private getUrl() {
+    return `${this.config.baseURL}/embedding-model`;
+  }
+
+  private getModelConfigHeaders() {
+    return {
+      'ai-embedding-model-specification-version': '2',
```
shaper, line 85: We can probably make this general like we did with model id, to look at later.

turbo.json
```diff
@@ -16,7 +16,7 @@
         "COHERE_API_KEY",
         "DEEPINFRA_API_KEY",
         "DEEPSEEK_API_KEY",
-        "DEPLOYMENT_ID",
+        "VERCEL_DEPLOYMENT_ID",
```
shaper: nit: alpha-sort with the rest

shaper (approved):

examples/ai-core/src/embed-many/gateway.ts
R-Taneja: This is a more involved change and is requires https://linear.app/vercel/issue/AI-4010/add-providermetadata-to-embed-responses. I'll add it to that ticket's description and make the change there

PR #7500, opened 2025-07-24
fix: Throw errors for error parts in `readUIMessageStream`

### LINKED ISSUE #7420: `readUIMessageStream` should throw an error when it encounters an `error` UI stream part
### Description

Currently there is no way to detect errors that happened during streaming when using `readUIMessageStream`, the message generator will just stop yielding new messages

Instead it should throw an error when it encounters an error stream part, for example it should throw an error when the UIStream contains something like this:

```
{
  "type": "error",
  "errorText": "Invalid input for tool strReplaceEditor: Type validation failed: Value: {\"command\":\"str_replace\",\"path\":\"essentials/images.md\",\"old_str\":\"<img\\n  style={{ borderRadius: '0.5rem' }}\\n  src=\\\"https://uploads.fumabase.com/Gui86K8XoAAZRb_.jpeg\\\"\\n/>\\n\\n\",\"new_str\":\"\"}.\nError message: [\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"file_text\"\n    ],\n    \"message\": \"Invalid input: expected string, received undefined\"\n  },\n  {\n    \"expected\": \"number\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"insert_line\"\n    ],\n    \"message\": \"Invalid input: expected number, received undefined\"\n  },\n  {\n    \"expected\": \"array\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"view_range\"\n    ],\n    \"message\": \"Invalid input: expected array, received undefined\"\n  }\n]"
}


```

### AI SDK Version

_No response_

remorses:
Fix https://github.com/vercel/ai/issues/7420


lgrammel:
one issue with this is that there could be multiple error parts and the stream may resume correctly after encountering an error, i.e. error chunks are not necessarily fatal for the stream. throwing would abort the stream.

i'll see if i can refine this / make it opt in

lgrammel:
Continued in #7503

PR #7520, opened 2025-07-24
feat (provider/gateway): Add providerMetadata to embeddings response

R-Taneja:
## Background

Provider metadata (Gateway routing) wasn't being returned in embedding responses

## Summary

Returned it in the embed route and printed it in 2 example Gateway scripts

## Verification

`pnpm test`
Manual verification using 2 example Gateway embeddings scripts

## Tasks

## Related Issues
https://linear.app/vercel/issue/AI-4010/add-providermetadata-to-embed-responses

### DIFF
--- a/.changeset/wise-bees-listen.md
+++ b/.changeset/wise-bees-listen.md
@@ -1,0 +1,7 @@
+---
+'@example/ai-core': patch
+'@ai-sdk/gateway': patch
+'ai': patch
+---
+
+feat (provider/gateway): Add providerMetadata to embeddings response
--- a/.changeset/wise-bees-listen.md
+++ b/.changeset/wise-bees-listen.md
@@ -1,5 +1,6 @@
 ---
 '@ai-sdk/gateway': patch
+'@ai-sdk/provider': patch
 'ai': patch
 ---
 
--- a/.changeset/wise-bees-listen.md
+++ b/.changeset/wise-bees-listen.md
@@ -1,5 +1,4 @@
 ---
-'@example/ai-core': patch
 '@ai-sdk/gateway': patch
 'ai': patch
 ---
--- a/examples/ai-core/src/embed-many/gateway.ts
+++ b/examples/ai-core/src/embed-many/gateway.ts
@@ -1,10 +1,6 @@
-import { createGatewayProvider } from '@ai-sdk/gateway';
+import { gateway } from '@ai-sdk/gateway';
 import { embedMany } from 'ai';
 import 'dotenv/config';
-
-const gateway = createGatewayProvider({
-  baseURL: 'http://localhost:3000/v1/ai',
-});
 
 async function main() {
   const result = await embedMany({
--- a/examples/ai-core/src/embed-many/gateway.ts
+++ b/examples/ai-core/src/embed-many/gateway.ts
@@ -1,6 +1,10 @@
-import { gateway } from '@ai-sdk/gateway';
+import { createGatewayProvider } from '@ai-sdk/gateway';
 import { embedMany } from 'ai';
 import 'dotenv/config';
+
+const gateway = createGatewayProvider({
+  baseURL: 'http://localhost:3000/v1/ai',
+});
 
 async function main() {
   const result = await embedMany({
@@ -14,7 +18,7 @@ async function main() {
 
   console.log('Embeddings:', result.embeddings);
   console.log('Usage:', result.usage);
-  
+
   if (result.providerMetadata) {
     console.log('\nProvider Metadata:');
     console.log(JSON.stringify(result.providerMetadata, null, 2));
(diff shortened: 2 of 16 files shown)


shaper:
Please update docs for `embed` and `embedMany` to include the provider metadata, ref:
https://v5.ai-sdk.dev/docs/reference/ai-sdk-core/embed
https://v5.ai-sdk.dev/docs/reference/ai-sdk-core/embed-many

you can see how we document it for `generateImage` response here:
https://v5.ai-sdk.dev/docs/reference/ai-sdk-core/generate-image

For images there is a specific `ImageModelProviderMetadata` type, but for `generateObject` and others there is not, I don't think we need a custom one for embeddings.
https://v5.ai-sdk.dev/docs/reference/ai-sdk-core/generate-object#provider-metadata

examples/ai-core/src/embed-many/gateway.ts
```diff
@@ -1,9 +1,13 @@
-import { gateway } from '@ai-sdk/gateway';
+import { createGatewayProvider } from '@ai-sdk/gateway';
 import { embedMany } from 'ai';
 import 'dotenv/config';
 
+const gateway = createGatewayProvider({
```
shaper: for merge need to remove this and just use `gateway` default provider instance import here and the other example. Maybe in a later PR we can add a lib to make it easier to switch via an env var, for use in these examples.

packages/ai/src/embed/embed-many.ts
```diff
@@ -195,6 +196,8 @@ Only applicable for HTTP-based providers.
           values,
           embeddings,
           usage,
+          providerMetadata: (response?.body as EmbeddingResponseBody)
```
shaper: do we use `as` in `generateObject`, `generateText`, and similar? Prefer type safety vs forcing as possible, same comment below. We usually do zod parsing to have known safe typing everywhere.

packages/gateway/src/gateway-embedding-model.ts
```diff
@@ -91,4 +91,5 @@ export class GatewayEmbeddingModel implements EmbeddingModelV2<string> {
 const gatewayEmbeddingResponseSchema = z.object({
   embeddings: z.array(z.array(z.number())),
   usage: z.object({ tokens: z.number() }).nullish(),
+  providerMetadata: z.any().optional(),
```
shaper: Avoid use of `any` -- revise to type this, look at how we handle it for `generateText`, `generateObject`, etc. It is probably set up as a `Record<string, unknown>` or similar.

packages/ai/src/embed/embed.ts
```diff
@@ -168,6 +173,8 @@ Only applicable for HTTP-based providers.
         value,
         embedding,
         usage,
+        providerMetadata: (response?.body as EmbeddingResponseBody)
```
shaper: Same comment as above re: avoid use of `as` -- explore type safety following pattern in existing similar code/functions.

examples/ai-core/src/embed-many/gateway.ts
R-Taneja: Ah this must've slipped through the cracks, I usually keep this change locally

packages/ai/src/embed/embed-many.ts
lgrammel: please extend the spec instead of adding this to the body. look at language model spec for examples

lgrammel (changes requested):
needs to extend embedding model spec similar to language model spec

packages/ai/src/embed/embed-many.ts
```diff
@@ -281,6 +289,9 @@ Only applicable for HTTP-based providers.
           embeddings.push(...result.embeddings);
           responses.push(result.response);
           tokens += result.usage.tokens;
+          if (!providerMetadata && result.providerMetadata) {
+            providerMetadata = result.providerMetadata;
+          }
```
lgrammel: so this means we would only return the provider metadata from the first request? or even a random request since it is parallel?

my preference would be to return either no provider metadata (leave it on the responses only) or all provider metadata

.changeset/wise-bees-listen.md
```diff
@@ -0,0 +1,7 @@
+---
+'@example/ai-core': patch
```
lgrammel: rm example

packages/ai/src/embed/embed-many.ts
R-Taneja: Good catch, fixed

.changeset/wise-bees-listen.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/gateway': patch
+'ai': patch
```
lgrammel, line 4: add provider changeset

lgrammel (approved):

PR #7580, opened 2025-07-27
fix(otel): change back `toolCall` attributes of input/output back to args/result for compatibility

neo:
in #6760, all the `.result` was supposed to change to `.output`; however, it seems like we missed this one place.

reference: https://github.com/vercel/ai/commit/63f9e9b45b2e6a3b53fe9097c817cbd7fd3882a4#diff-ae49190d75e3816588c4b5b0ea8fa7797fd775d1a70517f73b7a96f46c970cb6L249-R250


packages/ai/src/generate-text/generate-text.ts
```diff
@@ -606,8 +606,8 @@ async function executeTools<TOOLS extends ToolSet>({
                 selectTelemetryAttributes({
                   telemetry,
                   attributes: {
-                    'ai.toolCall.result': {
-                      output: () => JSON.stringify(result),
+                    'ai.toolCall.output': {
+                      output: () => JSON.stringify(output),
```
lgrammel: we are not changing telemetry attributes for backwards compatibility / otel provider compatibility. telemetry will be reworked in 5.1

neo: hey Lars! thank you for reviewing and sharing the decision

actually I only found out about this inconsistency because of a compatibility issue I encountered – some existing code (again, changed from #6760) seems to already have some of them changed:
https://github.com/vercel/ai/blob/f666efcc0cbc32f80e8ce78ed8f2994b8fa840bb/packages/ai/src/generate-text/run-tools-transformation.ts#L291-L300
https://github.com/vercel/ai/blob/f666efcc0cbc32f80e8ce78ed8f2994b8fa840bb/packages/ai/src/generate-text/run-tools-transformation.ts#L252-L254
https://github.com/vercel/ai/blob/f666efcc0cbc32f80e8ce78ed8f2994b8fa840bb/packages/ai/src/generate-text/generate-text.ts#L590-L592

do we want to either make this in or change those back?

lgrammel: hmm ideally we would change back to call/result if that was the case in ai sdk 4. telemetry will be revamped in ai sdk 5.1 and this change might have broken telemetry integrations

neo: it indeed had 😂 that's how I found out about it...

so do you want me to update this PR to change back? only the attribute names that start with `ai.toolCall.` are the ones that need to be changed back, right?

lgrammel: yes only toolCall.input and toolCall.output afaik. worth comparing to v4

neo: ok updated, the only thing I'm not sure is whether we also need to change the `input` inside `ai.response.toolCalls`:
```
"ai.response.toolCalls": "[{"toolCallId":"call-1","toolName":"tool1","input":"{ \\"value\\": \\"value\\" }"}]",
```

lgrammel:
Continued in #7623

Issue #7528, opened 2025-07-24
[SDK V5][Amazon Bedrock] Hitting "The toolConfig field must be defined when using toolUse and toolResult content blocks." when setting `toolChoice: 'none'`


brendan-kellam:
### Description

Heya - When using `streamText` with Amazon Bedrock, if a conversation has prior tool calls, any subsequent step with no active tools (`activeTools: []` or `toolChoice: none`) will fail. This only happens with Bedrock and no other provider I have tested.

**Steps to Reproduce:**

1.  Use `streamText` with a Bedrock model and a tool.
2.  In the `prepareStep` callback, set `activeTools` to an empty array for the second turn of the conversation to force a text-only response.
3.  The call fails when the agent tries to generate the final answer.

**Minimal Code Example:**

```typescript
const aws = createAmazonBedrock();

const stream = streamText({
    model: aws('anthropic.claude-3-5-sonnet-20241022-v2:0'),
    tools: {
        weather: tool({
            description: 'Get the weather in a location',
            inputSchema: z.object({ city: z.string() }),
            execute: async ({ city }) => ({
                result: `The weather in ${city} is 20°C.`,
            }),
        }),
    },
    stopWhen: [
        stepCountIs(5)
    ],
    prepareStep: ({ stepNumber }) => {
        if (stepNumber > 0) {
            return {
                activeTools: [] // or `toolChoice: 'none'`
            };
        }

        return undefined;
    },
    toolChoice: 'auto',
    prompt: 'What is the weather in Toronto, Calgary, and Vancouver?',
});
```

**Expected Behavior:**

The request should succeed, and the model should generate a final text response without calling any tools.

**Actual Behavior:**

The API call fails with the error: `The toolConfig field must be defined when using toolUse and toolResult content blocks.`

### AI SDK Version

- ai: 5.0.0-beta.26
- @ai-sdk/amazon-bedrock: 3.0.0-beta.9
- @ai-sdk/react:2.0.0-beta.26

dancer:
merged will be included in next beta release

brendan-kellam:
Hey @dancer thanks for the patch - I updated to bedrock version `3.0.0-beta.10` (and the ai package to`5.0.0-beta.28`), but now running into a different error message:

> 1 validation error detected: Value at 'toolConfig.tools' failed to satisfy constraint: Member must have length greater than or equal to 1

This happened on the first tool execution.

[...]

PR #7600, opened 2025-07-28
fix(amazon-bedrock): handle empty activeTools with tool conversation history

dancer:
## background

Users integrating Amazon Bedrock with multi-step agents hit validation errors when setting `activeTools: []` or `toolChoice: 'none'` in conversations that previously used tools. Bedrock requires toolConfig to be present when conversation contains tool content, but rejects empty tools arrays.

## summary

- add placeholder tool when activeTools is empty but conversation has tool content
- handle both `activeTools: []` and `toolChoice: 'none'` scenarios
- include helpful warning about workaround

## verification

- all existing tests pass with updated expectations
- multi-step agent scenarios work without validation errors
- placeholder tool doesn't interfere with final responses

## tasks

- [x] placeholder tool logic in bedrock-prepare-tools.ts
- [x] updated test expectations for both scenarios
- [x] warning messages for user awareness

## future work
* remove workaround if Amazon Bedrock API supports empty tools with conversation history

related issue #7528

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
+++ b/packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
@@ -2091,7 +2091,7 @@ describe('doGenerate', () => {
     `);
   });
 
-  it('should include toolConfig when conversation has tool calls but no active tools', async () => {
+  it('should omit toolConfig when conversation has tool calls but no active tools', async () => {
     prepareJsonResponse({});
 
     const conversationWithToolCalls: LanguageModelV2Prompt = [
@@ -2137,11 +2137,12 @@ describe('doGenerate', () => {
 
     const requestBody = await server.calls[0].requestBodyJson;
 
-    expect(requestBody.toolConfig).toMatchInlineSnapshot(`
-      {
-        "tools": [],
-      }
-    `);
+    // toolConfig should include placeholder tool when conversation has tool content
+    expect(requestBody.toolConfig).toBeDefined();
+    expect(requestBody.toolConfig.tools).toHaveLength(1);
+    expect(requestBody.toolConfig.tools[0].toolSpec.name).toBe(
+      '_ai_sdk_placeholder',
+    );
   });
 
   it('should handle JSON response format with schema', async () => {
@@ -2256,7 +2257,7 @@ describe('doGenerate', () => {
     ]);
   });
 
-  it('should include toolConfig when conversation has tool calls but toolChoice is none', async () => {
+  it('should omit toolConfig when conversation has tool calls but toolChoice is none', async () => {
     prepareJsonResponse({});
 
     const conversationWithToolCalls: LanguageModelV2Prompt = [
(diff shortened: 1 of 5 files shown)


packages/amazon-bedrock/src/bedrock-chat-language-model.test.ts
```diff
@@ -2137,11 +2137,12 @@ describe('doGenerate', () => {
...
-      {
-        "tools": [],
-      }
-    `);
+    // toolConfig should include placeholder tool when conversation has tool content
+    expect(requestBody.toolConfig).toBeDefined();
+    expect(requestBody.toolConfig.tools).toHaveLength(1);
+    expect(requestBody.toolConfig.tools[0].toolSpec.name).toBe(
+      '_ai_sdk_placeholder',
+    );
```
lgrammel: why replace the inline snapshot? ai?

packages/amazon-bedrock/src/bedrock-prepare-tools.ts
```diff
@@ -88,15 +122,49 @@ export function prepareTools({
...
-      };
+      // Bedrock does not support 'none' tool choice.
+      if (hasToolContent) {
+        // When conversation has tool content, provide a placeholder tool.
+        return {
+          toolConfig: {
+            tools: [
+              {
+                toolSpec: {
+                  name: '_ai_sdk_placeholder',
```
lgrammel: this seems like a really bad solution. is there no other way?

dancer: yep no other way

lgrammel (changes requested):
this will impact llm behavior with a fake tool

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -193,13 +191,63 @@ export class BedrockChatLanguageModel implements LanguageModelV2 {
...
+          }) as typeof prompt;
+
+        warnings.push({
+          type: 'unsupported-setting',
+          setting: 'toolContent',
+          details:
+            'Tool calls and results removed from conversation because Bedrock does not support tool content without active tools.',
+        });
+      }
+    }
```
lgrammel: could this be done with a message.map and parts.filter ? might be simpler and also avoid casting etc

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -197,37 +197,31 @@ export class BedrockChatLanguageModel implements LanguageModelV2 {
...
+                  ...message,
+                  content: message.content.filter(
+                    part =>
+                      part.type !== 'tool-call' && part.type !== 'tool-result',
+                  ),
+                },
+          )
+          .filter(
+            message => message.role === 'system' || message.content.length > 0,
+          ) as typeof prompt;
```
lgrammel, line 224: why is the type cast needed

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
```diff
@@ -193,13 +191,57 @@ export class BedrockChatLanguageModel implements LanguageModelV2 {
...
+
+    if (activeTools.length === 0) {
+      const hasToolContent = prompt.some(
+        message =>
+          'content' in message &&
+          Array.isArray(message.content) &&
+          message.content.some(
+            part => part.type === 'tool-call' || part.type === 'tool-result',
+          ),
+      );
```
lgrammel, line 207: why is this check needed

lgrammel (approved):

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
dancer: messages can contain only tool-call parts (no texts) and then tool messages contain only tool-results parts so after filtertering removes these we would get a empty content: [] array and bedrock rejects that

packages/amazon-bedrock/src/bedrock-chat-language-model.ts
dancer: ts can't prove that filtering maintains the exact prompt type structure, so we need the cast to tell it the result still matches LMV2Prompt

PR #7640, opened 2025-07-30
feat (ai): add experimental repairText function to streamObject

danielamitay:
@lgrammel please take a look! This pull request adds the `experimental_repairText` functionality to `streamObject`, matching the signature and behavior of `generateObject`. It does not attempt to repair the json/object during streaming, because that is assumed to always need repair and goes through a separate partial json repair flow. It will only perform the `experimental_repairText` on the final, accumulated output.

With the exception of three changes, everything is identical to the logic introduced by you for `generateObject`. The three changes:
1. I extracted the `processResult` logic into one/two new functions that both `generateObject` and `streamObject` can use. The implementation is the same as the original but it is restructured.
2. There were two existing unit tests/cases for `streamObject` that were erroring with the message `"No object generated: response did not match schema."` when the correct message is `"No object generated: could not parse the response."` because the two test cases in question were indeed providing invalid json that couldn't be parsed, regardless of the schema. The updated logic fixes this.
3. I added one extra unit test for the scenario where an LLM returns the json in a markdown code block, with repair functionality:
```
```json\n{ "content": "test message" }\n```
```

Which happens to be the scenario I'm looking to fix in my own use case.

## Background
- Implemented for `generateObject` in https://github.com/vercel/ai/pull/4937

## Summary
- Move RepairText interface
- Update StreamObjectResult interface
- Add unit tests for new behavior in `streamObject`, mirroring the existing unit tests for `generateObject`
- Extract the "processResult" logic from `generateObject` class to be shared with `streamObject`

## Verification
- Identical signature and implementation to match `generateObject`
- Unit tests for both `generateObject` and `streamObject` pass
- Artificially constructed llm request that returns incomplete json is patched in local testing


packages/ai/src/generate-object/generate-object.ts
```diff
@@ -421,66 +412,20 @@ Default and recommended: 'auto' (best mode for the model).
...
-              response,
-              usage,
-              finishReason,
-            });
-          }
-
-          const validationResult = await outputStrategy.validateFinalResult(
-            parseResult.value,
+        let object: RESULT;
+        try {
```
lgrammel: try catch can be removed if it only rethrows error

lgrammel:
looks good, just one smol comment

packages/ai/src/generate-object/generate-object.ts
danielamitay: Good catch ty!

danielamitay:
✅

packages/ai/src/generate-object/generate-object.ts
```diff
@@ -421,67 +412,16 @@ Default and recommended: 'auto' (best mode for the model).
...
-            if (repairedText === null) {
-              throw error;
-            }
-
-            object = await processResult(repairedText);
-          } else {
-            throw error;
-          }
-        }
+        let object: RESULT = await parseAndValidateObjectResultWithRepair(
```
lgrammel: can now be const and does not need a type

danielamitay: apologies, my work repo is set up to auto fix things like this, lost the muscle memory without it. updated

lgrammel (approved):

Issue #7636, opened 2025-07-29
Support for file_search_call.result in file_search tool for OpenAI provider


Andreas-velo:
### Description

Results from file search are not returned as messages by default from the File Search tool in OpenAI Responses tool https://platform.openai.com/docs/guides/tools-file-search?lang=javascript#include-search-results-in-the-response. 

There are no way to set the `include` option in AI SDK https://v5.ai-sdk.dev/providers/ai-sdk-providers/openai#file-search 

```
const response = await openai.responses.create({
    model: "gpt-4o-mini",
    input: "What is deep research by OpenAI?",
    tools: [{
        type: "file_search",
        vector_store_ids: ["<vector_store_id>"],
    }],
    include: ["file_search_call.results"],
});
```

### AI SDK Version

- ai: 5.0.0-beta.25
- @ai-sdk/openai: 2.0.0-beta.11

dancer:
will be fixed next patch

Andreas-velo:
Thanks for the update @dancer. 
I can now set the include parameter 
```
      providerOptions: {
        openai: {
          include: ['file_search_call.results'],
        },
      },
```
But I still don't get the file search result returned from the provider. Are the include input propagated to the actual provider call?

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
The follow-up reports a concrete failure of the newly supported `include` option: it can be configured but does not produce the requested provider results.

### Evidence
- AI SDK initially exposed no way to pass OpenAI Responses API's `include: ['file_search_call.results']` option.
- A maintainer stated it would be fixed in the next patch.
- After the patch, `providerOptions.openai.include` was accepted, but the requested file-search results were still absent, suggesting the option was not propagated or handled correctly.

PR #7643, opened 2025-07-30
feat(openai): add file_search_call.results support to include parameter

dancer:
## background

openai's responses api supports returning file search results in responses via the `include` parameter, but we only supported `reasoning.encrypted_content`. users needed access to file search results for debugging and transparency when using file search tools.

## summary

- add `file_search_call.results` to include parameter enum
- update ranking options to match current API (`auto`, `default-2024-08-21`)

## verification

- all tests pass
- include parameter correctly passes through to openai api
- file search ranking validation updated to current API spec

## tasks

- [x] extend include parameter schema for file_search_call.results  
- [x] update ranking options from deprecated values
- [x] fix test assertions for new ranking values

## future work

* monitor openai api for additional include parameter options

related issue #7636

### DIFF
--- a/.changeset/slimy-otters-tease.md
+++ b/.changeset/slimy-otters-tease.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+feat(openai): add file_search_call.results support to include parameter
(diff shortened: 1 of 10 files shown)


lgrammel:
pls add changeset

lgrammel (approved):

PR #7647, opened 2025-07-30
fix(examples/svelte): update tool call types for current api

dancer:
## background

SvelteKit examples were using outdated tool invocation patterns that caused TypeScript errors when opened in editors, breaking the developer experience for users trying the examples.

## summary

- fix tool part type checking patterns
- update onToolCall callback return types
- correct addToolResult api usage
- remove deprecated properties

## verification

- tool calls work with proper type safety
- client-server tool execution functions correctly
- multi-step tool flows execute as expected

## tasks

- [x] tool part access patterns updated from .toolInvocation to direct properties
- [x] addToolResult api corrected to use tool/output parameters
- [x] removed unsupported maxSteps from client chat initialization

### DIFF
--- a/.changeset/yellow-planes-sell.md
+++ b/.changeset/yellow-planes-sell.md
@@ -1,5 +1,0 @@
----
-'@example/sveltekit-openai': patch
----
-
-fix(examples/svelte): update tool call types for current api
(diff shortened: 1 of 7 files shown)


lgrammel (approved):

.changeset/yellow-planes-sell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@example/sveltekit-openai': patch
```
lgrammel: no changeset for examples

PR #7658, opened 2025-07-31
feat(ai): Export ai.response.usage span attribute

jacoblee93:
## Background

There used to be an OTEL `ai.usage.cachedInputTokens` span that the OpenAI language model populated which seems to have been dropped as part of AI SDK 5 (or possibly the default switched to the responses API?). In order for tracing providers to accurately track costs, we need more granular information about the usage of cached (and reasoning) tokens returned as part of OTEL.

## Summary

Adds `ai.response.usage` span attribute

## Verification

Updated and ran the examples changed in this PR.

## Tasks


lgrammel:
AI SDK 5 has all this information in the default usage. Would it be possible to use that instead?

jacoblee93:
Sure, if we'd rather export `usage` as a span I'm for that too.

Will revise and look into that.

jacoblee93:
How's this @lgrammel?

And congrats on the official AI SDK 5 launch!

lgrammel:
@jacoblee93 thanks!

the information is already there, i am a bit confused why this is needed?

image

jacoblee93:
Hmm seems you're right - I may have been on an older version while testing externally. Just bumped to `ai@5.0.0` and `@ai-sdk/openai@2.0.0` and got a run with caching. Will close.

jacoblee93:
Actually I think OpenAI's caching might just be very flaky.

PR #7264, opened 2025-07-13
feat: Add codemod suite for AI SDK v5 migration

dancer:
## background

AI SDK v5 introduces breaking changes across providers, streaming APIs, message types, and package structure. Users need automated migration tools to upgrade large codebases efficiently without manual refactoring. The existing codemod system needed better organization and version-specific commands.

## summary

- reorganize 62 codemods into v4/ and v5/ directories for better structure
- add version-specific CLI commands: `v4` and `v5` alongside existing `upgrade`
- update documentation with proper v4/ and v5/ prefixes in migration guides
- fix test fixtures and ensure all tests pass
- auto-generate README with version-categorized codemod listings

## verification

- all 62 codemods organized: 24 in v4/, 38 in v5/
- three working CLI commands: `upgrade` (all), `v4` (v3→v4), `v5` (v4→v5)
- migration guides updated with correct codemod names and new commands
- README auto-generates with proper v4/v5 categorization
- all tests passing with updated fixtures

## tasks

- [x] reorganize codemods into v4/ and v5/ subdirectories
- [x] update import paths in moved codemod files
- [x] add v4 and v5 CLI commands with separate upgrade functions
- [x] update README generation script for version categorization
- [x] fix migration guide documentation with correct codemod names
- [x] update test fixtures and resolve failing tests
- [x] ensure all 62 codemods work with new structure

## future work

* consider updating codemod categories

PR #7693, opened 2025-08-01
fix(codemod): track generateText usage before transform

imranbarbhuiya:
## Background

this is based https://github.com/vercel/ai/pull/6554#discussion_r2247413087
The current approach replaces all places where result.text, response.text is used. Its very common to use these functions and replacing all is incorrect. Also it doesnt transform when the variable name is different than result or response (and other common names). So this pr tracks for `generateText` usage, gets the variable name and replaces them only. This still has issue like if same name variable is there in the file for other use too, this'll transform that too, but ig this approach is still better than prev one.
## Summary

I'm tracking usage of `generateText`, getting the variables then transforming them.

## Verification

I ran the tests, all existing one are passing and added a new test to make sure it doesnt transform all. The jsshift is returning empty when no transforms are applied for me, ig its expected behaviour so the output file is kept empty
## Tasks

## Future Work

This pr still has issue like if same name variable is there in the file for other use too, this'll transform that too, but ig this approach is still better than prev one.


.changeset/hot-tomatoes-heal.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/codemod': major
```
imranbarbhuiya: Im not sure why it aded mafor here, I dont remember choosing any,  my commits also says `fix` only

gr2m (commented):
Thanks a lot Parbez for submitting the pull request! I'll have a look at it shortly and get back to you

gr2m:
@imranbarbhuiya I have not been working on codemods for the past few weeks but will pick the work on it again. Meanwhile, the team merged all codemods to `main` as part of the release of `ai` v5:

- https://github.com/vercel/ai/pull/7264

I changed the base to `main` and there are a few conflicts now. Also it has a ton of unrelated commits. I think best would be to re-create the pull request. One change that happened sinc you created this PR is that the codemod files are now in v4/v5 sub directories.

No worries if you don't have time or don't want to for any reason, just please let me know and I'll take care of it as soon as I can.

imranbarbhuiya:
I'm rebasing this pr, but seems like the tests for this codemod were removed from main, should I delete them too or readd?

dancer:
@imranbarbhuiya do me a favour if you don't mind make it one test file and i'll approve and merge once done ty

gr2m:
> I'm rebasing this pr, but seems like the tests for this codemod were removed from main, should I delete them too or readd?

yes we just realized, no idea what happened there. I see you re-added them. As a convention, we like having a 1:1 mapping of codemod to test file, and then use multiple `it` blocks in the test files with multiple fixtures as needed, as @dancer mentioned above

imranbarbhuiya:
Thanks, I'll do that now

imranbarbhuiya:
@dancer @gr2m thanks for review. I've included my test file in the existing one now

packages/codemod/src/test/__testfixtures__/replace-generatetext-text-property.output.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export async function generateTextWithDestructuring() {
+  const result = await generateText({
+    model: 'some-model',
+    prompt: 'Generate response'
+  });
+
+  const { text } = result;
+  return text;
```
gr2m, line 35: This should be `text.text`. I think you mentioned in another place that destructing is not yet support but your changes are definitely better than what we have today. I'll just add a comment here so future folks know we it's an open TODO and up for grabs if anyone is interested in fixing it

Maybe a comment like this in both input and output fixtures?

```suggestion
  // TODO: `text` should be transformed to `text.text`
  return text;
```

gr2m (approved):
> I've included my test file in the existing one now

wonderful, thanks a lot!

I have just one comment/suggestion but we can also merge as is

.changeset/hot-tomatoes-heal.md
gr2m: I got bitten by this several times myself.

When you run `pnpm changeset`, it will first ask you which packages are affected. Next question is which of the previously selected packages have major changes - don't select any. Then the same question for minor - again, don't select any. That will default to `patch`

packages/codemod/src/test/__testfixtures__/replace-generatetext-text-property.output.ts
imranbarbhuiya: Added

dancer:
thank you @imranbarbhuiya 🔥

PR #7698, opened 2025-08-01
Fix Documentation guides

philschmid:
# what does this PR do? 

Fixes the urls on the https://ai-sdk.dev/cookbook/guides page. Currently they redicted from /docs to /cookbook which works for all except Gemini. There it ends up on a 404. This fixes those links to be correct.


philschmid:
cc @nicoalbanese

nicoalbanese (approved):

PR #7702, opened 2025-08-01
feat(provider/gateway): Hide Cohere embedding models with no pricing info

R-Taneja:
## Background

4 Cohere embedding models were added as part of the Gateway embeddings feature, but there isn't any pricing info available.

## Summary

Commented them out (in the hopes that they can be re-added in the future)

## Tasks

### DIFF
--- a/packages/gateway/src/gateway-embedding-model-settings.ts
+++ b/packages/gateway/src/gateway-embedding-model-settings.ts
@@ -1,9 +1,5 @@
 export type GatewayEmbeddingModelId =
   | 'amazon/titan-embed-text-v2'
-  // | 'cohere/embed-english-light-v3.0'
-  // | 'cohere/embed-english-v3.0'
-  // | 'cohere/embed-multilingual-light-v3.0'
-  // | 'cohere/embed-multilingual-v3.0'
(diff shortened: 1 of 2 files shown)


packages/gateway/src/gateway-embedding-model-settings.ts
```diff
@@ -1,9 +1,9 @@
 export type GatewayEmbeddingModelId =
   | 'amazon/titan-embed-text-v2'
-  | 'cohere/embed-english-light-v3.0'
-  | 'cohere/embed-english-v3.0'
-  | 'cohere/embed-multilingual-light-v3.0'
-  | 'cohere/embed-multilingual-v3.0'
+  // | 'cohere/embed-english-light-v3.0'
```
shaper: We should just remove them, and edit the generation script to omit them either if the pricing info isn't sufficient or via an allow-list to omit if needed.

R-Taneja: Removed and edited the generation script in [this PR](https://github.com/vercel/ai-gateway/pull/253)

shaper (approved):

Issue #7740, opened 2025-08-04
Missing tool-file_search type in message parts when using OpenAI file_search tool


wassimAJ:
### Description

**Description:**
When invoking the OpenAI `file_search` tool, the response does not include a message part with type `"tool-file_search"` as expected. Instead, the response only contains a `text` type message part with the content, making it unclear that a tool call was involved.

**Steps to Reproduce:**

1. Use the `file_search` tool via the `createUIMessageStreamResponse()` or equivalent.
2. Inspect the response parts from the assistant.
3. Observe that the part indicating a tool call (`"tool-file_search"`) is missing.

**Expected Behavior:**
The response should include a message part with type `"tool-file_search"` or similar, to clearly reflect that the assistant used the file\_search tool.

**Actual Behavior:**
Only a `text` part is returned, even though a tool was invoked.

### AI SDK Version

- ai: 5.0.0
- @ai-sdk/openai: 2.0.0
- @ai-sdk/react: 2.0.0

wassimAJ:
@dancer I'm still receiving the type as "text" instead of "tool-file_search".

dancer:
@wassimAJ thank you for letting me sorry for the delay could you give me a code snippet reproducing the error

dancer:
i pushed a fix but i'll wait for your snippet to verify again just to make sure @wassimAJ @lgrammel

wassimAJ:
@dancer  Thank you for your support!

Here's the code snippet:

```
const result = streamText({
  model: aiModel,
  system: instructions + websearchPrompt,
  messages: convertToModelMessages(messages),
  tools: {
    file_search: openai.tools.fileSearch({
      vectorStoreIds: toolConfig.vector_store_ids || [],
      maxNumResults: 1,
    }),
  },
});
```


**Message logs:**

```
[
  {
    
    "id": "767aecba-2b7c-4d52-a1f9-f1b4fc157f55",
    "role": "user",
    "parts": [{
      "type": "text",
      "text": "tell me something random from the uploaded file"
    }]
  },
  {

    "id": "767aecba-2b7c-4d52-a1f9-c1b4fc147g77",
    "role": "assistant",
    "parts": [{
      "type": "step-start"
    }
  },
  {
    "state": "done",
    "text": "Here's something random from your uploaded file: It contains the word \"testing\".",
    "type": "text"
  }]
}
]
```

**Expected behavior:**

There should be a part with `"type": "tool-file_search"` present in the assistant response.

[...]

PR #7742, opened 2025-08-04
feat(openai): add file_search_call support to responses api

dancer:
## background

openai responses api introduced file_search_call tool calls that weren't supported causing only text part to be including

## summary

- add file_search_call support to openai responses api types
- add tests for file_search call handling

## verification

- all tests pass including new file_search-specific test
- inline snapshot correctly captures file_search tool call streaming
- file_search calls parsed as tool-call and tool-result pairs with providerExecuted: true

## tasks

- [x] openai-responses-api-types.ts updated with OpenAIFileSearchCall type
- [x] openai-responses-language-model.ts handles file_search_call in streaming
- [x] tests added for file_search tool call streaming scenario
- [x] inline snapshot validation for correct parsing

related issue - #7740

### DIFF
--- a/.changeset/little-doors-laugh.md
+++ b/.changeset/little-doors-laugh.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+feat(openai): add file_search_call support to responses api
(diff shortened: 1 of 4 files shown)


lgrammel:
changeset is missing

should those file search results be converted back to llm inputs in the next step?

lgrammel (approved):

Issue #7735, opened 2025-08-04
Type validation failed error with Fal.ai (kontext)


wong2:
### Description

After upgrading to AI SDK v5, I started to get this error when calling the Kontext image editing model with `fal` provider:

```
Type validation failed: Value: {"images":[{"url":"https://v3.fal.media/files/monkey/9A7PHLbs4H8rwEgxNXibU.png","content_type":"image/png","file_name":null,"file_size":null,"width":944,"height":1104}],"timings":{"inference":5.875932216644287},"seed":328395684,"has_nsfw_concepts":[false],"prompt":"A female model holding this book, keeping the book unchanged."}.
Error message: [{"code":"invalid_union","errors":[[{"expected":"string","code":"invalid_type","path":["images",0,"file_name"],"message":"Invalid input: expected string, received null"},{"expected":"number","code":"invalid_type","path":["images",0,"file_size"],"message":"Invalid input: expected number, received null"}],[{"expected":"object","code":"invalid_type","path":["image"],"message":"Invalid input: expected object, received undefined"}]],"path":[],"message":"Invalid input"}]
```

### AI SDK Version

ai: v5

wong2:
Still getting this error with @ai-sdk/fal 1.0.1

_TypeValidationError [AI_TypeValidationError]: Type validation failed: Value: {"images":[{"url":"https://fal.media/files/panda/G0v_9_OSALjG7PXAHtPL6_9b43c1118f6d4c459994a0ddb7046ba4.png","content_type":"image/png","file_name":null,"file_size":null,"width":880,"height":1184}],"timings":{},"seed":235205040,"has_nsfw_concepts":[false],"prompt":"Change the plates to colorful ones"}.
  Error message: [{"code":"invalid_union","errors":[[{"expected":"number","code":"invalid_type","path":["timings","inference"],"message":"Invalid input: expected number, received undefined"}],[{"expected":"object","code":"invalid_type","path":["image"],"message":"Invalid input: expected object, received undefined"},{"expected":"number","code":"invalid_type","path":["timings","inference"],"message":"Invalid input: expected number, received undefined"}]],"path":[],"message":"Invalid input"}]

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
The Fal provider's response schema rejects values currently returned by Fal, causing valid image-generation responses to throw a type-validation error. This is a concrete failure in existing provider behavior.

### Evidence
[...]

PR #7743, opened 2025-08-04
fix(fal): handle null file_name and file_size in image responses

dancer:
## background

fal.ai kontext image editing model returns null values for file_name and file_size fields in image responses, but the ai sdk validation schema expected these fields to be string/number or undefined, causing type validation errors.

## summary

- update falImageSchema to allow nullable file_name and file_size fields
- update loraFileSchema for consistency
- added tests for null values scenario

## verification

- all tests pass including new test for null values
- kontext model responses with null file_name/file_size now validate successfully
- null values preserved in provider metadata as expected

## tasks

- [x] falImageSchema updated with nullable() for file_name and file_size
- [x] loraFileSchema updated for consistency
- [x] test case added replicating exact kontext scenario
- [x] test validates null values are preserved in metadata

## future work

* monitor other fal models that may return null values in different fields

related issue - #7735

### DIFF
--- a/.changeset/chatty-candles-change.md
+++ b/.changeset/chatty-candles-change.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/fal': patch
+---
+
+fix(fal): handle null file_name and file_size in image responses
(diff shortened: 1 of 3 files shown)


lgrammel:
please fix you ai setup to generate changesets

lgrammel (approved):

Issue #7745, opened 2025-08-04
`lastAssistantMessageIsCompleteWithToolCalls` should support dynamic tool calls


vinayakakv:
## Problem

Currently, `lastAssistantMessageIsCompleteWithToolCalls` checks for `isToolUIPart` in the last tool invocation to see if it has finished streaming output.

https://github.com/vercel/ai/blob/6fb1b4378f5f9f390d08ef932660bc17e65c186b/packages/ai/src/ui/last-assistant-message-is-complete-with-tool-calls.ts#L29

However, `isToolUIPart` does not support `dynamic-tool`s.

https://github.com/vercel/ai/blob/6fb1b4378f5f9f390d08ef932660bc17e65c186b/packages/ai/src/ui/ui-messages.ts#L274-L278

This makes `lastAssistantMessageIsCompleteWithToolCalls` return `false` wherever dynamic tools are used.

## Possible solutions

Checking for `dynamic-tool` part along with `isToolUIPart` `lastAssistantMessageIsCompleteWithToolCalls` would solve this issue.

Changing `isToolUIPart` could be dangerous as it is used in many places other than `lastAssistantMessageIsCompleteWithToolCalls`

## Workarounds

Implementing `lastAssistantMessageIsCompleteWithToolCalls` with support for dynamic tools and plugging it into `sendAutomaticallyWhen` of `useChat` is one workaround.

### AI SDK Version

- `ai: 5.0.0`

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This describes a concrete failure in an existing completion helper when used with dynamic tool calls, along with a targeted correction. Although support is requested, the current behavior is presented as incorrect for an existing tool type.

### Evidence
- In AI SDK 5.0.0, `lastAssistantMessageIsCompleteWithToolCalls` checks `isToolUIPart`, which excludes `dynamic-tool` parts.
- As a result, the helper reportedly always returns `false` when dynamic tools are used, preventing the expected completion detection.
- The reporter provides a workaround by reimplementing the helper with dynamic-tool support.

PR #7746, opened 2025-08-04
Support dynamic tools in lastAssistantMessageIsCompleteWithToolCalls

vinayakakv:
Solves [!7745](https://github.com/vercel/ai/issues/7745)


lgrammel:
breaks backwards compat. i have a different fix. this was what ai suggested for me as well at first.

PR #7750, opened 2025-08-04
fix(anthropic): thinking blocks must precede tool_use in final assistant messages

dancer:
## background

claude sonnet 4 with extended thinking requires thinking blocks to appear before tool_use blocks in the final assistant message, but we were not enforcing this ordering requirement, causing API errors

## summary

- reorder thinking blocks to come first in final assistant messages when extended thinking is enabled
- preserve existing behavior when thinking is disabled

## verification

- all tests pass including new thinking block reordering test
- thinking blocks correctly appear before tool_use blocks in final messages

## tasks

- [x] add isThinking parameter to convertToAnthropicMessagesPrompt
- [x] implement content reordering logic for final assistant messages
- [x] tests added for thinking block ordering requirement

related issue - #7729

### DIFF
--- a/.changeset/tame-ducks-behave.md
+++ b/.changeset/tame-ducks-behave.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix(anthropic): thinking blocks must precede tool_use in final assistant messages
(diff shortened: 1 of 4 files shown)


lgrammel:
i feel this is not a good fix. do we know why the order is broken?

lgrammel (changes requested):
blocking for now; this fix is likely to cause more problems than it solves

jmif:
@lgrammel we're running into this issue and would like to try to fix it.  Can you elaborate on why this is not a good solution?  I want to make sure I incorporate your feedback in the PR we open.

PR #7755, opened 2025-08-04
fix(fal): handle empty timings object in image responses

dancer:
## background

fal.ai's kontext image editing models sometimes return an empty `timings` object `{}` instead of omitting the field or including the `inference` property. the current schema validation required `timings.inference` to be a number when `timings` was present, causing type validation errors.

## summary

- make `timings.inference` field optional in response schema
- added test for empty timings object scenario

## verification

- all existing tests pass
- new test verifies empty timings object handling
- schema now accepts both `timings: {}` and `timings: { inference: 5.875 }`

## tasks

- [x] update commonResponseSchema to make inference optional
- [x] add test case for empty timings object

related issue - #7735

### DIFF
--- a/.changeset/shy-beers-behave.md
+++ b/.changeset/shy-beers-behave.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/fal': patch
+---
+
+fix(fal): handle empty timings object in image responses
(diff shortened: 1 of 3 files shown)


lgrammel:
changeset

lgrammel (approved):

PR #7781, opened 2025-08-05
feat (ai): add experimental tool call context

### LINKED ISSUE #3938: send context into the tool execute function
### Feature Description

Similar to #3468 

When working on agents or tools that require retrieving, saving, or working with a session, you might need to use dynamic IDs or data that were sent with the original API call but not included as an option or argument on the tool call. It would be great if you can extend `ToolExecutionOptions` and send in your own variables to be used when the `tool.execute` function is running. 

To keep it type safe, you can define a type for the variables you're going to send and add the type as a generic type parameter. So an example might look like:

```typescript
// set up type
export type ExtendedOptions = {
  chatId: string;
  userId: string;
}

// define your tools
export const fileSearch = tool<ExtendedOptions>({
  description: 'Semantically search for relevant files the user owns',
  parameters: z.object({
    query: z.string().describe('The search query'),
  }),
  execute: async ({ query }, { userId }) => {
    const result = await searchFiles({
      query,
      userId,
    });
    return {
      files: result.files,
    };
  },
});

export const codeExecution = tool<ExtendedOptions>({
  description:
    "Execute code in a sandboxed Python Notebook environment dedicated to the user's project",
  parameters: z.object({
    code: z.string().describe('The code to execute'),
  }),
  execute: async ({ code }, { chatId, userId, messages }) => {
    const codeCells = getAllCodeToolCalls(messages);
    codeCells.push(code);
    const result = await executeCodeInSandbox({
      codeCells,
      userId,
      chatId,
    });
    return {
      result: result.stdout,
    };
  },
});
```

Then in the API route, you could send in the dynamic data with the tools

```typescript
    const { chatId } = await request.json();
    const session = await auth();

    // set up the dynamic option object
    const options: ExtendedOptions = {
      chatId: chatId,
      userId: session.user.id,
    };

    // set up the tools with the dynamic options
    const tools = {
      'file-search': {
        ...fileSearch,
        options,
      },
      'code-execution': {
        ...codeExecution,
        options,
      },
    };

    const { result } = streamText(
      // rest of the parameters
      ...
      tools,
    )
```

This is a pretty basic example, a feature like this would be more useful  for agentic workloads where your agents need access to dynamic data, like `chatId` or `userId` for example.

### Use Cases

_No response_

### Additional context

There is a workaround where you have a function that takes the dynamic data as arguments which then creates the tool. It's doable but becomes harder to manage as your toolset grows.

lgrammel:
## Background

Tools often need custom context. The current workaround is to create tools with a closure. However, this causes issues with the AI SDK 5 type inference helpers.

## Summary

Introduce `experimental_context` argument on `generateText` and `streamText` that is passed into the tools.

## Manual Verification

Run generate example.

## Tasks

- [x] add context to tool interface
- [x] add context arg to generateText
- [x] add context arg to streamText
- [x] add to agent
- [x] example generate
- [x] test generate text
- [x] test stream text
- [x] update docs
- [x] update ref docs
- [x] changeset

## Future Work

Exploring adding context to `prepareStep`. Explore context modifications (parallel tool calls make it difficult).

## Related Issues

Fixes #3938


dancer (approved):

minostauros:
What should I do to catch the experimental_context in mcp servers, e.g., FastMCP streamable http servers?

PR #7829, opened 2025-08-06
Allow passing model string for embeddings

R-Taneja:
## Background

Previously, embeddings models could only be used by calling `.textEmbeddingModel()` on a `gateway` instance

## Summary

Now, users can pass in model strings directly. LIMITATION: If users need to specify custom params (e.g. base url, api key, etc.) or call `createOpenAICompatible`, they'd still need to use `gateway.textEmbeddingModel()` because `gateway` alone returns a language model type. In order to change that, I'd have to make a breaking change to how language models work - probably by returning a union type. Worth it?

## Manual Verification

Tests and example scripts pass/work

## Tasks

### DIFF
--- a/examples/ai-core/src/embed-many/gateway.ts
+++ b/examples/ai-core/src/embed-many/gateway.ts
@@ -1,10 +1,9 @@
-import { gateway } from '@ai-sdk/gateway';
 import { embedMany } from 'ai';
 import 'dotenv/config';
 
 async function main() {
   const result = await embedMany({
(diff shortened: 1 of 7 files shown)


gr2m (commented):
changes look good to me. Maybe put the PR in draft if you are waiting for more feedback on the changes themselves before updating docs, tests, and changeset?

lgrammel (changes requested):
i'll continue this pr

PR #7852, opened 2025-08-07
react: useObject – add non-streaming fallback for RN/Expo (fixes #7817)

### LINKED ISSUE #7817: React Native / Expo - [Error: The response body is empty.]
Hello
I am trying to use the Vercel ai-sdk in order to extract specific fields values (first-name, last-name, sex) from a set of patients.
I am developing an Expo app (quite advanced right now) and try to integrate this AI sdk for importing patients into it.
I have extracted my code to show it in a reproducable way [here](https://github.com/sgandon/react-native-sandbox/tree/master/ai-sdk), in order to run it it requires an .env.local set with the env-var EXPO_PUBLIC_OPENAI_API_KEY, and an update of the [server url](https://github.com/sgandon/react-native-sandbox/blob/94e894f5467fdb4bee7c612b5e3f69abe2e26bee/ai-sdk/app/(tabs)/ai.tsx#L9)).

### Server side
I have developed a server side API [here](https://github.com/sgandon/react-native-sandbox/blob/master/ai-sdk/app/api/extract-patients%2Bapi.tsx) and it seems to be working at least when I do a curl.
```
curl -X POST http://localhost:8082/api/extract-patients
[{"firstname":"Mildred","lastname":"Hudson","male":false},{"firstname":"WALTER","lastname":"MYERS","male":true},{"firstname":"Eugene","lastname":"Cook","male":true},{"firstname":"Pamela","lastname":"Owens","male":false}]
```
but the client fails.

### client side
I just implemented a simple client using `useObject` [here](https://github.com/sgandon/react-native-sandbox/blob/94e894f5467fdb4bee7c612b5e3f69abe2e26bee/ai-sdk/app/(tabs)/ai.tsx#L8) and it is failing with the following error
```
 ERROR  === useObject ERROR ===
 ERROR  useObject error: [Error: The response body is empty.]
```
The call seem to be waiting for the end of the payload (so the isLoading is working). But I would like to have streaming of each object. The object returned seems to be never filled with the data.  
And it eventually fails at the end.

Any help would be much appreciated.

Thanks.

_Originally posted by @sgandon in https://github.com/vercel/ai/discussions/7782_

dipampaul17:
This PR fixes an error in `experimental_useObject()` when running in environments without `ReadableStream`/`TextDecoderStream` (e.g., Expo/React Native), which caused `Error: The response body is empty.`

Changes:
- Detect streaming capability; if unavailable, fall back to `response.text()` and parse once.
- Preserve progressive updates when streaming is available.
- Add UI tests for non-stream JSON and empty responses in `@ai-sdk/react`.
- Add changeset for `@ai-sdk/react`.

This addresses the issue reported in #7817 and allows RN/Expo users to successfully use `useObject` without stream support while still preferring streaming where available.

### DIFF
--- a/.changeset/use-object-fallback-react.md
+++ b/.changeset/use-object-fallback-react.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/react': patch
+---
+
+Fix `experimental_useObject()` to support environments without `ReadableStream`/`TextDecoderStream` (e.g., React Native/Expo) by adding a non-streaming JSON fallback. Also add tests for non-stream responses and empty bodies. Fixes #7817.
+
+
--- a/packages/react/src/use-object.ts
+++ b/packages/react/src/use-object.ts
@@ -189,47 +189,81 @@ function useObject<
         );
       }
 
-      if (response.body == null) {
-        throw new Error('The response body is empty.');
-      }
-
       let accumulatedText = '';
       let latestObject: DeepPartial<RESULT> | undefined = undefined;
 
-      await response.body.pipeThrough(new TextDecoderStream()).pipeTo(
-        new WritableStream<string>({
-          async write(chunk) {
-            accumulatedText += chunk;
+      const canStream =
+        response.body != null &&
+        typeof (response.body as any).pipeThrough === 'function' &&
+        typeof (globalThis as any).TextDecoderStream !== 'undefined' &&
+        typeof (globalThis as any).WritableStream !== 'undefined';
 
-            const { value } = await parsePartialJson(accumulatedText);
-            const currentObject = value as DeepPartial<RESULT>;
+      if (canStream) {
+        await response.body
+          .pipeThrough(new TextDecoderStream())
+          .pipeTo(
+            new WritableStream<string>({
+              async write(chunk) {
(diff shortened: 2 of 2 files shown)


gr2m:
Hi Dipam, thanks for sending the pull request!

I'm not very familiar with React Native, I probably miss something obvious. Based on [this comment](https://github.com/vercel/ai/issues/7817#issuecomment-3160825942), the error does not seem to occur when using `fetch` from the `expo` package? Would using the `expo` package be a workable workaround for React Native that wouldn't require changes in `@ai-sdk/react`?

lgrammel (changes requested):
strongly object to putting in code to work around issues stemming from other frameworks. this should be fixed on the expo side or with polyfills see https://ai-sdk.dev/docs/getting-started/expo#polyfills

gr2m:
Closing as per @lgrammel's comment. We are trying to stay aligned with web standards to maximize compatibility with today's and future JS runtimes. This seems to be a case where React Native is not compatible, and a problem that should be worked around outside of the `ai` SDK.

Let's keep the discussion in #7817. Maybe we can at least document said workaround for React Native and Expo in the AI SDK docs

PR #7880, opened 2025-08-08
feat(openai): add verbosity parameter support for responses api

dancer:
## background

openai released documentation for a new `verbosity` parameter in their responses api that allows controlling output length with values 'low', 'medium', and 'high'. this parameter enables developers to get more concise or verbose responses without modifying prompts

## summary

- add verbosity parameter to openai responses api provider options
- update zod schema to validate verbosity values
- add documentation with usage examples

## verification

- verbosity parameter is correctly passed 
- documentation follows existing patterns
- example demonstrates verbosity

PR #7897, opened 2025-08-08
feat(openai): add verbosity parameter support for chat api

imranbarbhuiya:
## background

openai released documentation for a new `verbosity` parameter in their chat api that allows controlling output length with values 'low', 'medium', and 'high'. this parameter enables developers to get more concise or verbose responses without modifying prompts

## summary

- add verbosity parameter to openai chat api
- update zod schema to validate verbosity values
- add documentation with usage examples

## verification

- verbosity parameter is correctly passed 
- documentation follows existing patterns
- example demonstrates verbosity

### DIFF
--- a/.changeset/thin-eyes-sort.md
+++ b/.changeset/thin-eyes-sort.md
@@ -2,4 +2,4 @@
 '@ai-sdk/openai': patch
 ---
 
-add verbosity settings in openai provider options
+feat(openai): add verbosity parameter support for chat api
--- a/.changeset/thin-eyes-sort.md
+++ b/.changeset/thin-eyes-sort.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+add verbosity settings in openai provider options
--- a/content/providers/01-ai-sdk-providers/03-openai.mdx
+++ b/content/providers/01-ai-sdk-providers/03-openai.mdx
@@ -228,7 +228,7 @@ The following optional provider options are available for OpenAI chat models:
   Whether to use strict JSON schema validation.
   Defaults to `false`.
 
-- **verbosity** _'low' | 'medium' | 'high'_
+- **textVerbosity** _'low' | 'medium' | 'high'_
 
   Controls the verbosity of the model's responses. Lower values will result in more concise responses, while higher values will result in more verbose responses.
 
--- a/content/providers/01-ai-sdk-providers/03-openai.mdx
+++ b/content/providers/01-ai-sdk-providers/03-openai.mdx
@@ -188,7 +188,7 @@ The following optional provider options are available for OpenAI chat models:
   A unique identifier representing your end-user, which can help OpenAI to
   monitor and detect abuse. [Learn more](https://platform.openai.com/docs/guides/safety-best-practices/end-user-ids).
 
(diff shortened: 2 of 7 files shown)


MonkeyLeeT:
Seems they prefer slightly different naming - https://github.com/vercel/ai/pull/7880

imranbarbhuiya:
Thanks for linking, somehow I missed it when checking the code lol

MonkeyLeeT:
@imranbarbhuiya sorry for not making it clear - your change is still needed for chat model since they only update response model to support that param.

@dancer FYI.

imranbarbhuiya:
Reopened, thanks @MonkeyLeeT . I'll wait for @dancer before changing the prop name

MonkeyLeeT:
@imranbarbhuiya do you mind also adding "minimal" to chat provider option while you're at this? https://github.com/vercel/ai/blob/e284ba652116335f26e9967c614b4e6a5c337b11/packages/openai/src/openai-chat-options.ts#L88

volkandkaya:
> @imranbarbhuiya do you mind also adding "minimal" to chat provider option while you're at this?
> 
> https://github.com/vercel/ai/blob/e284ba652116335f26e9967c614b4e6a5c337b11/packages/openai/src/openai-chat-options.ts#L88

Very much needed as I go from 50 output tokens to 750 with low on.

lgrammel:
`textVerbosity` was added in #7880

fwiw there was an error in the openai docs, verbosity is not a top level option in their api, it is under `text` and we flattened it to align with our other settings.

imranbarbhuiya:
@lgrammel correct me if Im wrong but ig that pr changes in responses and this pr is still needed for chat model file right?

for example here's openai sdk completion has it as top level https://github.com/openai/openai-node/blob/40b79cb8a372dce925b14a26e62110f9de6786ba/src/resources/chat/completions/completions.ts#L1785
and then responses has inside text https://github.com/openai/openai-node/blob/40b79cb8a372dce925b14a26e62110f9de6786ba/src/resources/responses/responses.ts#L4303

lgrammel:
@imranbarbhuiya ah i see. have you tested this manually against the chat model? can you point me to the openai api docs for chat/completion that show this flag? please change the title to make it clear that this is for the chat api

imranbarbhuiya:
Here's the doc link https://platform.openai.com/docs/api-reference/responses/create#responses_create-verbosity

I'll send u an ss of using it in a few mins

imranbarbhuiya:
https://platform.openai.com/docs/guides/latest-model?verbosity-mode=chat#verbosity:~:text=minimal%22%0A%20%20%7D%0A%7D%27-,Verbosity,-Responses%20API

text highlight doesnt seem to work so here's the image
image


this link shows verbosity as top level for chat completion and inside text for response. I can't seems to test my changes inside example folder, even if I add console logs in the `@ai-sdk/openai`'s `doGenerate`, it doesnt appear, lemme know how to test it

dancer:
@lgrammel tested works great ty @imranbarbhuiya for the help adjusted just a tiny bit so it matches responses but thank you

removed minimal reasoning will be in a different pr

dancer (approved):

MonkeyLeeT:
@dancer Would appreciate if you could add minimal reasoning asap. It's causing a lot of token waste in many non-reasoning use case right now across the board.

PR #7902, opened 2025-08-08
fix(ai): Call `onFinish` when message ui stream is cancelled in `toUIMessageStream`

### LINKED ISSUE #7469: `onFinish` from `createUIMessageStream` not called when aborted
### Description

I noticed that the `onFinish` callback from `createUIMessageStream` isn't called when `streamText` gets aborted through the `abortSignal`. However, `onFinish` from `toUIMessageStream` does get called. Example:

```ts
const resultStream = createUIMessageStream<ChatMessage>({
  execute: ({ writer: dataStream }) => {
    const result = streamText({
      abortSignal: AbortSignal.any([requestAbortSignal, responseAbortSignal.signal]),
      ...
    });

    const uiMessageStream = result.toUIMessageStream<ChatMessage>({
      originalMessages: uiMessages,
      onFinish: () => {
        // DOES get called when aborted

        // also when the request completes normally the response has no id here!!
      }
    });

    dataStream.merge(uiMessageStream);
  },
  onFinish: async () => {
    // NOT called when aborted
  },
});
```

I also noticed that the response message in the `onFinish` from `toUIMessageStream` doesn't have an `id` and it's an empty string when the response completes successfully.

### AI SDK Version

Using `ai@5.0.0-beta.25`.

### LINKED ISSUE #7900: `onFinish` in `toUIMessageStream` should be called on abort with the currently available partial ui message
### Description

Currently, `onFinish` is **not** called when a stream is aborted. It only gets called in the `flush()` method when the stream completes normally:

```typescript
// packages/ai/src/ui-message-stream/handle-ui-message-stream-finish.ts
async flush() {
  // onFinish only called here - not on abort
  await onFinish({
    isAborted,
    isContinuation,
    responseMessage: state.message as UI_MESSAGE,
    messages: [...],
  });
}
```

When an abort occurs:
- `onAbort` callback is triggered (if provided)
- Stream emits an `abort` event
- Stream closes without calling `flush()`
- **`onFinish` is never called**

## Problem

This behavior is problematic for chat interfaces where users may:
- Close the browser window/tab during streaming
- Navigate away from the page
- Lose network connection
- Explicitly cancel the request

In these scenarios, we need to persist the **partial message** to the database for conversation continuity. The current implementation assumes streams always complete normally, which doesn't reflect real-world usage patterns.

## Use Case

```typescript
  model: yourModel,
  messages: chatHistory,
});

return result.toUIMessageStream({
  onFinish: async ({ responseMessage, isAborted }) => {
    // Need this to be called even on abort with partial message
    await db.messages.save({
      ...responseMessage,
      status: isAborted ? 'partial' : 'complete'
    });
  }
});
```

## Expected Behavior

`onFinish` should be called in **both** scenarios:
1. Normal completion (current behavior)
2. On abort with the partially accumulated message

The callback already receives an `isAborted` flag, suggesting this was the intended design. The partial message data is already available in the `state.message` when abort occurs.

## Proposed Solution

Modify the stream handling to call `onFinish` with `isAborted: true` when an abort occurs, passing the current partial message state. This would provide a consistent interface for handling both normal and abnormal stream terminations.

Related to https://github.com/vercel/ai/issues/7469

### AI SDK Version

"ai": "5.0.8"

remorses:
Fix https://github.com/vercel/ai/issues/7469
Fix #7900

Call `onFinish` when abort controller aborts, via the TransformStream cancel method: https://developer.mozilla.org/en-US/docs/Web/API/TransformStream#browser_compatibility

That method is currently only available in Node and Deno


grant0417:
Anything blocking here? Also hitting this issue and would love if this was merged.

remorses:
I am not sure this PR solves the problem and it's a bit difficult to test.

The problem here is that ai sdk currently uses a type: "abort" chunk to detect aborts but it looks like this chunk cannot be sent in case of a client abort because the stream can no longer be written to after the abort happens.

Fixing this may require a big change, updating how aborts are detected in a lot of places

remorses:
I found a little documented field on TransformStream that should fix the problem @lgrammel 

The tests I added fail without the cancel fix

lgrammel:
I want to understand why this is a problem to begin with. I changed the abort behavior in streamText to not abort a stream any more; instead abort gets turned into regular stream behavior with an abort chunk to prevent special cases like this one.

Can you explain or even better share reproduction code for how such a situation can even occur?

remorses:
The problem was not strictly related to the abort controller but interrupting the async generator.

While writing this comment I also found another issue which was related to consuming the stream with await for and doing an early break.

I updated and added a new test file with the reproduction which do not pass without the fix and use the full streamText function to make sure they reproduce the real problem.

Basically the smallest repro is this code:

```ts
const result = streamText({
  model,
  prompt: 'Generate text',
});

  onFinish: onFinishCallback,
  generateMessageId: () => 'msg-async-iter',
});

// Consume only first few chunks using async iteration, then break
let chunkCount = 0;

  chunkCount++;

  if (chunkCount >= 5) {
    break; // This was not triggering the stream flush(). Instead it just stops consuming the stream and onfinish is never called
  }
}
// onFinish is never called
```

This code will cause the generator to stop and the stream to stop being consumed, the problem was in `createAsyncIterableStream` which does not pass a `return` method in the async iterator. return is necessary to cleanup resources when a `await for` loop breaks

I updated the code to cancel the stream when the async generator breaks. With the added cancel method in the `TransformStream` `onFinish` will be called.

Adding cancel in `TransofrmStream` now also handles the case when the user cancels the `UIMessageStreamResponse`, which is another problem that some users could face, some frameworks probably cancel responses when the request abort controller aborts to prevent memory leaks. Smallest repro is this code:

```tsx
const res = result.toUIMessageStreamResponse()
res.body?.cancel()
```

lgrammel:
```ts
const result = streamText({
  model,
  prompt: 'Generate text',
});

  onFinish: onFinishCallback,
  generateMessageId: () => 'msg-async-iter',
});

// Consume only first few chunks using async iteration, then break
let chunkCount = 0;

  chunkCount++;

  if (chunkCount >= 5) {
    break; // This was not triggering the stream flush(). Instead it just stops consuming the stream and onfinish is never called
  }
}
// onFinish is never called
```

@remorses isn't this expected behavior? you never fully consume the stream, so onFinish is not called, since you may continue consuming the stream at any point. if you want to abort stream consumption, you could use an abort controller and abort streamText when you break the loop?

lgrammel:
@remorses if this is a problem with the asynciterablestream, a good way may be to have extended unit tests coverage for that file so we can see each individual edge case and how it behaves. this could be a first good improvement in a separate pr, maybe it even solves the issues that you are encountering

remorses:
>  isn't this expected behavior? you never fully consume the stream, so onFinish is not called, since you may continue consuming the stream at any point

That is currently not possible with the current implementation. A for await creates a new iterator each time, but that iterator calls stream.getReader() which fails if the stream is still locked by the previous loop. You would need to add `releaseLock` to add support for that case

I think cancelling the stream when you exit the for loop is the better interface. Otherwise it is too easy to create memory leaks where you never consume the stream when using await for.

Do you want me to add more tests for that file in this PR?

lgrammel:
@remorses either here or in a separate pr just for the asynciterablestream

remorses:
I added more tests for `createAsyncIterableStream`, all results seem intuitive to me

Trying to iterate on the async iterable twice will simply yield no chunks the second time. This is not a breaking change because with the previous implementation it would just fail because of the missing reader.releaseLock() (throws with `TypeError: Invalid state: ReadableStream is locked`)

remorses:
I also updated packages/ai/src/ui/convert-to-model-messages.ts to not error when the messages contain non finished tool calls. Previously it would error if the client sends unfinished messages with tool calls still in input-streaming state, now it ignores tools that are in input-streaming state instead

lgrammel:
sorry for the delay; will try to find time to take a look today or tomorrow

lgrammel:
As a first step, I have extracted the async iterable stream changed into a separate pr: https://github.com/vercel/ai/pull/8220

Once that lands and this PR is updated, I will continue looking into the other parts.

remorses:
I fixed the conflicts with upstream, ready to review

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -149,12 +149,7 @@ export function convertToModelMessages(
               } else if (part.type === 'dynamic-tool') {
                 const toolName = part.toolName;
 
-                if (part.state === 'input-streaming') {
-                  throw new MessageConversionError({
-                    originalMessage: message,
-                    message: `incomplete tool input is not supported: ${part.toolCallId}`,
-                  });
-                } else {
+                if (part.state !== 'input-streaming') {
```
lgrammel, line 152: how are the changes in this file and chat.ts above related to the issue? can these changes be separated?

remorses: without this change you cannot resume an aborted message if you abort during a tool call. because when the client sends the messages back to the ai route it can contain unfinished tool calls

lgrammel:
`streamText` has an `onFinish` and an `onAbort` callback. The `onFinish` callback receives usage information that is not available / incorrect in the case of abort. Therefore only `onAbort` should be called.

Is there a bug where `onAbort` is not called? `onFinish` is not called by design atm.

lgrammel:
`streamText` has an `onFinish` and an `onAbort` callback. The `onFinish` callback receives usage information that is not available / incorrect in the case of abort. Therefore only `onAbort` should be called.

Is there a bug where `onAbort` is not called? `onFinish` is not called by design atm.

(this is slightly different for the ui message stream onFinish so there may be more here, just reviewed the first test part)

remorses:
Are you referring to the test I added in the last commit right?

I added this test to make sure I was understanding when StreamText `onFinish` runs and this was the wanted behaviour. Notice this test is not about aborts but errors. I tried running this same test in main and it still passes so it is unrelated to my changes. 

Do you want me to remove that test?

lgrammel:
@remorses maybe i got confused by that test. if it is valuable, would it be possible to move it to a separate pr and integrate it into `stream-text.test.ts`?

remorses:
Sure. I can confirm onFinish is still not called during aborts, I can add test for it

lgrammel:
@remorses ideally move as much as possible into separate prs. overload is real so i appreciate minimal prs focussed on 1 problem.

packages/ai/src/generate-text/stream-text-onfinish-abort.test.ts
```diff
@@ -0,0 +1,224 @@
+import { convertArrayToReadableStream } from '@ai-sdk/provider-utils/test';
+import { describe, expect, it, vi } from 'vitest';
+import { MockLanguageModelV2 } from '../test/mock-language-model-v2';
+import { streamText } from './stream-text';
+
+describe('streamText onFinish with abort', () => {
```
lgrammel: can you integrate this test into stream-text test and in particular into the ui message stream section (since it is specifically for toUIMessageStream)

packages/ai/src/ui-message-stream/handle-ui-message-stream-finish.test.ts
```diff
@@ -2,7 +2,7 @@ import {
   convertArrayToReadableStream,
   convertReadableStreamToArray,
 } from '@ai-sdk/provider-utils/test';
-import { describe, expect, it, vi } from 'vitest';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
```
lgrammel: can remove whole import line

packages/ai/src/ui-message-stream/handle-ui-message-stream-finish.ts
```diff
@@ -101,20 +121,13 @@ export function handleUIMessageStreamFinish<UI_MESSAGE extends UIMessage>({
       transform(chunk, controller) {
         controller.enqueue(chunk);
       },
+      // @ts-expect-error cancel is still new and missing from types https://developer.mozilla.org/en-US/docs/Web/API/TransformStream#browser_compatibility
```
lgrammel, line 124: we support node 20 and this was added in 21

image

will this cause any issues on node 20?

packages/ai/src/ui/convert-to-model-messages.ts
lgrammel: i wonder how to best handle/document this. right now from what i understand input-streaming would be dropped automatically, correct?

packages/ai/src/ui-message-stream/handle-ui-message-stream-finish.ts
remorses: No issues, it simply uses the previous behaviour where onFinish is not called

packages/ai/src/ui-message-stream/handle-ui-message-stream-finish.test.ts
remorses: fixed

packages/ai/src/generate-text/stream-text-onfinish-abort.test.ts
remorses: moved to existing file

lgrammel (approved):

gr2m (approved):
Great PR, thank you Tommy 💐

gr2m:
@remorses could you look into the merge conflict? Sorry for that, we can look into it ourselves but it might take a bit longer

remorses:
Fixed the merge conflict

PR #7922, opened 2025-08-09
fix(gateway): always download assets for language model

FranciscoMoretti:
This forces assets to be downloaded for the gateway language model before sending them downstream.

- Some providers accept asset/file parts as URLs, but others don't. The gateway currently rejects URL parts for some backends, causing inconsistent behavior.
- Normalizes behavior so assets are always accepted regardless of provider.

Context:
- Using the `gateway` provider with `openai/gpt-5-mini` failed with URL asset parts, but using `openai` directly with `gpt-5-mini` worked. The gateway returned:

```
GatewayInternalServerError: 'PDF file parts with URLs' functionality not supported.
    at createGatewayErrorFromResponse (...)
    at asGatewayError (...)
    at GatewayLanguageModel.doStream (...)
```

Openai has `{'image/*': [/^https?:\/\/.*$/],}` as `supportedURLs`, but gateway-language-model.ts has `{ '*/*': [/.*/] };`

### Notes:
Another solution could be to generate a complex `supportedUrl` setting based on the `owned_by` of the `modelId` and the providers that can serve it, but it seems quite error-prone and hard to maintain. The solution proposed in this PR downloads assets even when sometimes not needed, but shines in simplicity.

### DIFF
--- a/.changeset/rich-mayflies-wash.md
+++ b/.changeset/rich-mayflies-wash.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+fix(gateway): Download assets before request instead of sending as URLs
--- a/packages/gateway/src/gateway-language-model.ts
+++ b/packages/gateway/src/gateway-language-model.ts
@@ -28,7 +28,9 @@ type GatewayChatConfig = GatewayConfig & {
 
 export class GatewayLanguageModel implements LanguageModelV2 {
   readonly specificationVersion = 'v2';
-  readonly supportedUrls = { '*/*': [/.*/] };
+  readonly supportedUrls: Record<string, RegExp[]> = {
+    // No URLs are supported. For URLs to be supported, all providers available in the gateway should support them.
+  };
 
   constructor(
     readonly modelId: GatewayModelId,


shaper:
Hi, thank you for the PR. Would prefer to avoid forcing all providers to download urls in all cases, and instead look into revising AI Gateway to better support its stated contract by checking the provider and downloading only if needed server side. Will look into this.

Repro:
```
import 'dotenv/config';

async function main() {
    model: 'gpt-5-mini',
    messages: [
      {
        role: 'user',
        content: [
          {
            type: 'text',
            text: 'What is an embedding model according to this document?',
          },
          {
            type: 'file',
            data: new URL(
              'https://github.com/vercel/ai/blob/main/examples/ai-core/data/ai.pdf?raw=true',
            ),
            mediaType: 'application/pdf',
            filename: 'ai.pdf',
          },
        ],
      },
    ],
  });

  console.log(result.text);
}

main().catch(console.error);
```

FranciscoMoretti:
Sounds great @shaper! Feel free to close or rename this issue

Here's the code I've used to `replaceFilePartUrlByBinaryDataInMessages` until that's fixed. Posting in case others encounter the same issue.

```tsx
import type {
  ModelMessage,
  TextPart,
  ImagePart,
  FilePart,
  DataContent,
} from 'ai';

// Minimal utilities to download assets from URL-based parts and inline them.

export type DownloadResult = {
  mediaType: string | undefined;
  data: Uint8Array;
};

export type DownloadImplementation = (args: {
  url: URL;
}) => Promise<DownloadResult>;

async function defaultDownload({ url }: { url: URL }): Promise<DownloadResult> {
  const response = await fetch(url);
  if (!response.ok) {
    throw new Error(
      `Failed to download asset: ${url.toString()} (${response.status})`,
    );
  }
  const contentType = response.headers.get('content-type') || undefined;
  const arrayBuffer = await response.arrayBuffer();
  return { mediaType: contentType, data: new Uint8Array(arrayBuffer) };
}

function toHttpUrl(value: unknown): URL | null {
  if (value instanceof URL)
    return value.protocol.startsWith('http') ? value : null;
  if (typeof value === 'string') {
    try {
      const url = new URL(value);
      return url.protocol === 'http:' || url.protocol === 'https:' ? url : null;
    } catch {
      return null;
    }
  }
  return null;
}

/**
 * Collects all http(s) URLs from file/image parts in the provided messages and downloads them.
 * Returns a map keyed by the normalized URL string.
 */
export async function downloadAssetsFromModelMessages(
  messages: ModelMessage[],
  downloadImplementation: DownloadImplementation = defaultDownload,
): Promise<Record<string, DownloadResult>> {
  const urlSet = new Set<string>();

  for (const message of messages) {
    if (typeof message.content === 'string') continue;
    for (const part of message.content) {
      if (part.type !== 'file' && part.type !== 'image') continue;
      const dataOrUrl: DataContent | URL =
        part.type === 'file'
          ? (part as FilePart).data
          : (part as ImagePart).image;
      const url = toHttpUrl(dataOrUrl);
      if (url) urlSet.add(url.toString());
    }
  }

  const urls = Array.from(urlSet).map((u) => new URL(u));
  const downloaded = await Promise.all(
    urls.map(async (url) => ({
      url,
      data: await downloadImplementation({ url }),
    })),
  );
  return Object.fromEntries(
    downloaded.map(({ url, data }) => [url.toString(), data]),
  );
}

/**
 * Inlines any URL-based file/image parts within ModelMessage[] by replacing the URLs
 * with downloaded binary data. This ensures providers receive actual bytes.
 */
export async function replaceFilePartUrlByBinaryDataInMessages(
  messages: ModelMessage[],
  downloadImplementation: DownloadImplementation = defaultDownload,
): Promise<ModelMessage[]> {
  const downloaded = await downloadAssetsFromModelMessages(
    messages,
    downloadImplementation,
  );

  const mapPart = (
    part: TextPart | ImagePart | FilePart | any,
  ): TextPart | ImagePart | FilePart | any => {
    if (part.type === 'file') {
      const url = toHttpUrl((part as FilePart).data);
      if (url) {
        const found = downloaded[url.toString()];
        if (found) {
          const newPart: FilePart = {
            ...part,
            data: found.data,
            // keep existing mediaType; if missing, prefer detected content-type
            mediaType: part.mediaType ?? found.mediaType ?? part.mediaType,
          };
          return newPart;
        }
      }
      return part;
    }
    if (part.type === 'image') {
      const url = toHttpUrl((part as ImagePart).image);
      if (url) {
        const found = downloaded[url.toString()];
        if (found) {
          const newPart: ImagePart = {
            ...part,
            image: found.data,
            mediaType: part.mediaType ?? found.mediaType ?? part.mediaType,
          };
          return newPart;
        }
      }
      return part;
    }
    // pass-through for text/tool/reasoning/etc
    return part;
  };

  return messages.map((message) => {
    if (typeof message.content === 'string') return message;
    return {
      ...message,
      content: (
        message.content as Array<TextPart | ImagePart | FilePart | any>
      ).map(mapPart),
    } as ModelMessage;
  });
}
```

gr2m:
Closing after the requested gateway-side approach wasn't pursued. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

PR #7930, opened 2025-08-10
fix(providers): correct Base64 for Uint8Array data parts in chat messages 

duz52:
## Background

When a user message contains a file part whose `data` is a Uint8Array, the previous implementation created an incorrect Base64 string when constructing data URLs for image parts. This led to malformed `data:<mediaType>;base64,...` strings and downstream provider validation/decoding errors in chat APIs.

## Summary

Bug fix to ensure correct Base64 encoding for binary/file content parts across providers by using `convertToBase64(part.data)`:

- Use `convertToBase64` when building `image_url.url` for non-URL data.
- Applied consistently across affected providers.

### DIFF
--- a/.changeset/four-bugs-reflect.md
+++ b/.changeset/four-bugs-reflect.md
@@ -1,0 +1,8 @@
+---
+'@ai-sdk/openai-compatible': patch
+'@ai-sdk/mistral': patch
+'@ai-sdk/openai': patch
+'@ai-sdk/groq': patch
+---
+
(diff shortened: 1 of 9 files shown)


lgrammel:
Can you add unit tests and a patch changeset?

duz52:
> Can you add unit tests and a patch changeset?

Added unit tests and a patch changeset.

lgrammel (approved):

PR #7967, opened 2025-08-11
refactor(provider/openai): restructure files

lgrammel:
## Background

The OpenAI provider is quite complex and has 3 different LLM APIs (chat, completion, responses). There have been inconsistencies in the provider tool implementations between chat and responses.

## Summary

* move files into subfolders (e.g. `chat`)
* duplicate some files to better separate implementations for different APIs
* fix tool inconsistencies


packages/openai/src/internal/index.ts
```diff
@@ -1,13 +1,13 @@
...
+export * from '../completion/openai-completion-language-model';
+export * from '../completion/openai-completion-options';
+export * from '../embedding/openai-embedding-model';
+export * from '../embedding/openai-embedding-options';
+export * from '../image/openai-image-model';
+export * from '../image/openai-image-options';
+export * from '../transcription/openai-transcription-model';
+export * from '../transcription/openai-transcription-options';
+export * from '../speech/openai-speech-model';
+export * from '../speech/openai-speech-options';
```
gr2m, line 12: love. it! So clean ✨

gr2m (approved):
This is great, thanks a lot. Directly addresses some confusion I had yesterday about the 3 different endpoints that the `openai` provider uses.

packages/openai/src/responses/openai-responses-prepare-tools.ts
```diff
@@ -64,22 +65,16 @@ export function prepareResponsesTools({
             });
             break;
           }
-          case 'openai.web_search_preview':
-            // TODO update this with proper validation
+          case 'openai.web_search_preview': {
+            const args = webSearchPreviewArgsSchema.parse(tool.args);
```
gr2m, line 69: I assume this is just an additional refactoring without any change in behavior?

lgrammel: it adds validation and cleans up the types (before the api types were incorrect and we did not validate). no API surface changes though.

PR #7971, opened 2025-08-11
`convertToModelMessages()` result does not validate for `tool-call` if providerExecuted is null

wille:
## Background

A UIMessage returned from streamText or generateText can have `providerExecuted: null` which will fail validation when passed back to `streamText` or `generateText` using `convertToModelMessages`

Using `ai@5.0.9` and `@ai-sdk/openai@2.0.9`

## Summary

`toolCallPartSchema.providerExecuted` is now nullable and having a UIMessage with this field set to null will not fail validation.

I also looked into fixing it in processUIMessageStream(), toResponseMessages() and convertToModelMessages()

## Manual Verification

This is a test script to recreate the situation
The UI message will run through convertToModelMessages which will retain providerExecuted as null and then fail the validation

```js
import { convertToModelMessages, modelMessageSchema } from 'ai'
import { z as z6 } from "zod/v4";
import { safeValidateTypes } from '@ai-sdk/provider-utils'

const parsed = convertToModelMessages([
  {
    "id": "zz",
    "role": "assistant",
    "parts": [
      {
        "type": "step-start"
      },
      {
        "type": "reasoning",
        "text": "",
        "providerMetadata": {
          "openai": {
            "itemId": "rs_689a0675997481a1a390556b17fbcdbf0d60f55498a35479",
            "reasoningEncryptedContent": null
          }
        },
        "state": "done"
      },
      {
        "type": "tool-get_definition",
        "toolCallId": "call_IUPLg8KAna4uHSkF3sMZHsa4",
        "state": "output-available",
        "input": {
          "functionName": "cube"
        },
        "output": "Nothing implemented yet for cube",
        "rawInput": null,
        "errorText": null,
        "providerExecuted": null, // This is retained in the ModelMessage and fails validatiton below
        "callProviderMetadata": {
          "openai": {
            "itemId": "fc_689a06811a5c81a1a4d52a2b60f76c1f0d60f55498a35479"
          }
        }
      }
    ]
  }
]
);

// Taken from src/prompt/standardize-prompt.ts
const validationResult = await safeValidateTypes({
  value: parsed,
  schema: z6.array(modelMessageSchema)
});

if (validationResult.error)
console.log('Error')
console.dir(validationResult.error, { depth: 100 })
}

```

Manual patch until it's fixed upstreams

```js
const messages: UIMessage[] = [ /* ... */ ]
const modelMessages = convertToModelMessages(messages)

for (const msg of modelMessages) {
  if (msg.role === 'assistant') {
    for (const part of msg.content) {
      if (typeof part !== 'string' && part.type === 'tool-call' && part.providerExecuted === null) {
        delete part.providerExecuted;
      }
    }
  }
}
```

## Tasks

## Related Issues

I didn't open an issue for this bug.


packages/ai/src/prompt/content-part.ts
```diff
@@ -89,7 +89,7 @@ export const toolCallPartSchema: z.ZodType<ToolCallPart> = z.object({
   toolName: z.string(),
   input: z.unknown(),
   providerOptions: providerMetadataSchema.optional(),
-  providerExecuted: z.boolean().optional(),
+  providerExecuted: z.boolean().nullable().optional(),
```
lgrammel: use `nullish`

lgrammel:
Can you add a test case that reproduces the bug?

wille:
@lgrammel I added a test in convert-to-model-messages.test.ts. UIMessage providerExecuted type is not nullable, not sure if we should fix that.

I also added a call to standardizePrompt in every convert-to-model-messages.test.ts unit test to make sure that the converted messages always validate to avoid this situation again.

Let me know what you think

packages/ai/src/ui/convert-to-model-messages.test.ts
```diff
@@ -1,9 +1,10 @@
 import { ModelMessage } from '@ai-sdk/provider-utils';
 import { convertToModelMessages } from './convert-to-model-messages';
+import { standardizePrompt } from '../prompt/standardize-prompt';
```
lgrammel: why are standardizedPrompt tests mixed into this test? please revert

packages/ai/src/ui/convert-to-model-messages.test.ts
```diff
@@ -1176,4 +1176,61 @@ describe('convertToModelMessages', () => {
...
+            {
+              type: "tool-x",
+              toolCallId: "call_1",
+              state: "output-available",
+              input: {
+                x: "y"
+              },
+              output: "nothing",
+              rawInput: null,
+              // @ts-expect-error
```
lgrammel, line 1195: this would also need a change of the type in provider utils.

however, this raises a point: is this pr fighting symptoms instead of root causes?

can you show me code that produces `providerExecute: null` in ui messages? most likely that is what needs to be fixed instead of adding flexibility here

nicolaslopezj: I'm using a pretty standard implementation of ai v5 + gemini and I'm also having this error

wille: Trying to recreate it and fix the root cause

aayush-kapoor:

PR #7977, opened 2025-08-12
feat (provider/openai): provider defined image generation tool support

shaper:
## Background

OpenAI offers [image generation](https://platform.openai.com/docs/guides/tools-image-generation) for some of their language models. GPT-5 model launch includes the latest version of this feature. Users would like to be able to use these models to generate images.

## Summary

Added support for the image generation provider-defined tool. Note:

- includes tool options/arguments
- examples to show use with generate and stream text
- docs add to describe the new feature

## Manual Verification

Ran generate and stream examples.

## Tasks

- [x] A _patch_changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset`in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix`in the project root)

## Future Work

No support yet for [multi-turn editing](https://platform.openai.com/docs/guides/tools-image-generation#multi-turn-editing), partial images, or returning richer image detail in response via provider metadata on file parts.


shaper:
* **#7977** <a href="https://app.graphite.dev/github/pr/vercel/ai/7977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/ai/7977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/provider/src/language-model/v2/language-model-v2-file.ts
```diff
@@ -22,4 +23,9 @@ as base64 encoded strings. If the API returns binary data, the file data should
 be returned as binary data.
  */
   data: string | Uint8Array;
+
+  /**
+   * Additional provider-specific metadata for the file.
+   */
+  providerMetadata?: SharedV2ProviderMetadata;
```
shaper: This is a proposal -- if this is non-breaking and feasible, we need to add documentation for the type change to this PR. Waiting for feedback before doing so, since I realize we may want to revert this and consider as work for the future instead.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -104,6 +104,14 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
 
     const strictJsonSchema = openaiOptions?.strictJsonSchema ?? false;
 
+    // Determine which includes are supported by the OpenAI API and which are SDK-only.
```
shaper: This may need cleanup/revision -- first pass here added an `includes` we use only AI SDK-side to let a user specify whether they want to receive notice on partial image response parts, so here we split the include options into OpenAI-pass-through ones, and ones we use ourselves.

We could do it as a separate provider option. You could also tell me that we're better off with a minimal first implementation that only responds with the final full image part, and we can clean up the partial image response logic. Since these images are slow to produce, it seems valuable to me to return the incremental insight as it comes through, but it would be something you'd want to opt in to.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1082,6 +1220,28 @@ function isResponseOutputItemAddedReasoningChunk(
   );
 }
 
+function mapOpenAIImageFormatToMediaType(format: string): string {
```
shaper, line 1223: I know we have `detectMediaType` but it doesn't look exported, and here we actually have a format, we're just trying to make it fully qualified

lgrammel (dismissed):
please do not change language model v2 before 5.1

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1106,6 +1242,28 @@ function isErrorChunk(
...
+  sequence_number: number;
+  output_index: number;
+  item_id: string;
+  partial_image_index: number;
+  partial_image_b64: string;
+  size?: string;
+  quality?: string;
+  background?: string;
+  output_format?: string;
+};
```
lgrammel: should this be part of the response schema so the structure is validated?

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -802,6 +892,36 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
...
+                          background: value.background ?? null,
+                          quality: value.quality ?? null,
+                          size: value.size ?? null,
+                          partial_image_index: value.partial_image_index,
+                        },
+                      },
+                    },
+                  });
+                }
+              }
```
lgrammel: not sure if partial support is a good idea at this point. we would send many files that are just variants of the same partial. my hunch is that we need to extend the model spec for partial files and also tool call results. for now we may be better off without partials support.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -104,6 +104,14 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
 
     const strictJsonSchema = openaiOptions?.strictJsonSchema ?? false;
 
+    // Determine which includes are supported by the OpenAI API and which are SDK-only.
+    const requestedIncludes = openaiOptions?.include ?? undefined;
+    const apiSupportedIncludes = (requestedIncludes ?? []).filter(include =>
+      ['reasoning.encrypted_content', 'file_search_call.results'].includes(
+        include,
+      ),
+    );
```
lgrammel: would be nice to extract a generic intersection function for 2 arrays

lgrammel:
before ai sdk 5.1 i would recommend a simplified version without partials and no change to the spec

we may also not need any spec change, for multi-turn interactions we would most likely need to add provider executed tool calls/results which support metadata (ideally separate PR)

packages/openai/src/responses/openai-responses-language-model.ts
shaper: removed for now, no more partial images support

packages/openai/src/responses/openai-responses-language-model.ts
shaper: removed for now.

packages/openai/src/responses/openai-responses-language-model.ts
shaper: removed for now

gr2m (approved):
Changes look good to me now. Examples are working as well 👍🏼

content/providers/01-ai-sdk-providers/03-openai.mdx
```diff
@@ -899,6 +899,73 @@ const result = await generateText({
...
+import { openai } from '@ai-sdk/openai';
+import { generateText } from 'ai';
+import 'dotenv/config';
+
+const result = await generateText({
+  model: openai('gpt-5'),
+  prompt:
+    'Generate an image of an echidna swimming across the Mozambique channel.',
+  tools: {
+    image_generation: openai.tools.generateImage({ outputFormat: 'webp' }),
```
lgrammel: we align our tool names with the openai tool names except for the snake case

`openai.tools.imageGeneration`

content/providers/01-ai-sdk-providers/03-openai.mdx
```diff
@@ -899,6 +899,73 @@ const result = await generateText({
...
+import { openai } from '@ai-sdk/openai';
+import { streamText } from 'ai';
+import 'dotenv/config';
+
+const result = streamText({
+  model: openai('gpt-5'),
+  prompt:
+    'Generate an image of an echidna swimming across the Mozambique channel.',
+  tools: {
+    image_generation: openai.tools.generateImage({
```
lgrammel: same

examples/ai-core/src/generate-text/openai-responses-image-output.ts
```diff
@@ -0,0 +1,27 @@
...
+import 'dotenv/config';
+import { presentImages } from '../lib/present-image';
+
+async function main() {
+  const result = await generateText({
+    model: openai('gpt-5'),
+    prompt:
+      'Generate an image of an echidna swimming across the Mozambique channel.',
+    tools: {
+      image_generation: openai.tools.generateImage({
```
lgrammel: same

examples/ai-core/src/stream-text/openai-responses-image-output.ts
```diff
@@ -0,0 +1,36 @@
...
+import { streamText } from 'ai';
+import { presentImages } from '../lib/present-image';
+
+async function main() {
+  const result = streamText({
+    model: openai('gpt-5'),
+    prompt:
+      'Generate an image of an echidna swimming across the Mozambique channel.',
+    tools: {
+      image_generation: openai.tools.generateImage({
```
lgrammel: same

packages/openai/src/tool/image-generation.ts
```diff
@@ -0,0 +1,31 @@
...
+      .union([z.literal('auto'), z.string().regex(/^\d+x\d+$/)])
+      .optional(),
+    quality: z.enum(['auto', 'low', 'medium', 'high']).optional(),
+    moderation: z.enum(['auto']).optional(),
+    outputFormat: z.enum(['png', 'jpeg', 'webp']).optional(),
+    outputCompression: z.number().int().min(0).max(100).optional(),
+  })
+  .strict();
+
+export const generateImage = createProviderDefinedToolFactory<
```
lgrammel: rename to imageGeneration

packages/openai/src/openai-tools.ts
```diff
@@ -1,9 +1,11 @@
 import { codeInterpreter } from './tool/code-interpreter';
 import { fileSearch } from './tool/file-search';
+import { generateImage } from './tool/image-generation';
 import { webSearchPreview } from './tool/web-search-preview';
 
 export const openaiTools = {
   codeInterpreter,
   fileSearch,
+  generateImage,
```
lgrammel: imageGeneration

lgrammel:
Have you tested multi-step behavior, in particular changing a previously generated image, and also the same with the `store: false` provider option?

content/providers/01-ai-sdk-providers/03-openai.mdx
```diff
@@ -899,6 +899,73 @@ const result = await generateText({
   be customized.
 </Note>
 
+#### Image Generation (Responses API)
+
+OpenAI's Responses API supports multimodal image generation as a provider-defined tool. Availability is restricted to specific models (for example, `gpt-5` variants). Refer to the OpenAI docs for the current list of supported models.
```
gr2m, line 904: > Refer to the OpenAI docs

should we link to a specific section in the OpenAI docs?

Maybe:

```suggestion
OpenAI's Responses API supports multimodal image generation as a provider-defined tool. Availability is restricted to specific models (for example, `gpt-5` variants). [Refer to the OpenAI docs](https://platform.openai.com/docs/guides/image-generation#supported-models) for the current list of supported models.
```

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1374,7 +1409,11 @@ function getResponsesModelConfig(modelId: string): ResponsesModelConfig {
   };
 }
 
-// TODO AI SDK 6: use optional here instead of nullish
+function mapOpenAIImageFormatToMediaTypeSimple(formatOrType: string): string {
+  const lower = (formatOrType ?? 'png').toLowerCase().trim();
+  return lower.includes('/') ? lower : `image/${lower}`;
+}
+
```
gr2m, line 1416: Keep the comment

```suggestion
function mapOpenAIImageFormatToMediaTypeSimple(formatOrType: string): string {
  const lower = (formatOrType ?? 'png').toLowerCase().trim();
  return lower.includes('/') ? lower : `image/${lower}`;
}

// TODO AI SDK 6: use optional here instead of nullish
```

gr2m (approved):

gr2m:
We've been looking into making the image generation work with `store: false`.

I can fake it by constructing the request manually

```ts
import { readFile } from 'node:fs/promises';

import { openai, OpenAIResponsesProviderOptions } from '@ai-sdk/openai';

import { presentImages } from '../lib/present-image';

import 'dotenv/config';

async function main() {
    model: openai('gpt-5'),
    tools: {
      image_generation: openai.tools.imageGeneration({
        outputFormat: 'webp',
        quality: 'low',
      }),
    },
    messages: [
      {
        role: 'user',
        content: [
          {
            type: 'text',
            text: 'Now flip it upside down',
          },
          {
            type: 'image',
            image: await readFile('output/image-1756250346508-0.webp'),
          },
        ],
      },
    ],
    providerOptions: {
      openai: {
        store: false,
        include: ['reasoning.encrypted_content'],
      } satisfies OpenAIResponsesProviderOptions,
    },
  });

  console.log(result.text);

  for (const file of result.files) {
    if (file.mediaType.startsWith('image/')) {
      await presentImages([file]);
    }
  }
}

main().catch(console.error);
```

But when running `streamText()` in a loop appending new messages, then the `messages` argument will look something like this (simplified)

```js
      messages: [
        {
          role: 'user',
          content: 'A night stand desk but the bulb is a glowing hummingbird',
        },
        {
          role: 'assistant',
          content: [
            {
              type: 'reasoning',
              text: '',
              providerOptions: {
                openai: {
                  itemId: 'rs_68a...37a',
                  reasoningEncryptedContent:
                    'gAAAA...9jDHMiEJg==',
                },
              },
            },
            {
              type: 'file',
              data: BASE_64_IMAGE_DATA,
              mediaType: 'image/webp',
              providerOptions: undefined,
            },
          ],
        },
        {
          role: 'user',
          content: 'now make this exact image black and white',
        },
      ],
```

And when sending that request, the API is ignoring the image sent in the `type: 'file'` item and instead is generating a new image based on the original prompt. Not sure if there is a way to make this work auto-magically or if it's just not meant to be.

gr2m:
@lgrammel and I paired on this and created a draft PR with our current state:
https://github.com/vercel/ai/pull/8324

We found a bigger problem that `tool-result` contents are entirely ignored in our OpenAI provider, this is probably affecting all tool calls beyond image generation, that will need to be addressed anyway:
https://github.com/vercel/ai/pull/8324/files#diff-dffcdddf53bd204cf98b98ea32820d29386ff652b902f117abb6151030c57f22L155

But following up with my comment above, when using `store: false` the multi-step image editing is currently blocked by the fact that `id` is required in `image_generation_call` inputs by the OpenAI API ([source](https://github.com/vercel/ai/pull/8324/files#diff-dffcdddf53bd204cf98b98ea32820d29386ff652b902f117abb6151030c57f22R158)), but we shouldn't send it if `store` is set to false.

content/providers/01-ai-sdk-providers/03-openai.mdx
```diff
@@ -899,6 +899,73 @@ const result = await generateText({
   be customized.
 </Note>
 
+#### Image Generation (Responses API)
```
lgrammel, line 902: if we leave the pr as is, it would be good to add a note that says that multi turn is not supported

lgrammel:
I am concerned about the following error scenario:

* `store: false`, `encrypted_reasoning` included, reasoning model e.g. `gpt-5`
* generate image in the conversation
* subsequent messages fail with `Item 'rs_...' of type 'reasoning' was provided without its required following item`

It would be good to test that that wont happen.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -371,6 +371,11 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
                   }),
                 ),
               }),
+              z.object({
+                type: z.literal('image_generation_call'),
+                output_format: z.string().nullish(),
```
lgrammel, line 376: this property is not documented. can we use our own mime type detector instead?

gr2m (changes requested):
As per Lars' comment, let's add update the examples to also send a 3nd message to make sure that that works

gr2m:
using with multi step (without setting `store: false`) fails with

>  Item 'rs_68af7328fab88194af0aa10740efe95707788fdc7d719264' of type 'reasoning' was provided without its required following item

<details><summary><code>generateText()</code></summary>

```ts
import 'dotenv/config';
import { presentImages } from '../lib/present-image';

async function main() {
  const result1 = await generateText({
    model: openai('gpt-5'),
    prompt:
      'Generate an image of an echidna swimming across the Mozambique channel.',
    tools: {
      image_generation: openai.tools.imageGeneration({
        outputFormat: 'webp',
      }),
    },
  });

  console.log(result1.text);

  for (const file of result1.files) {
    if (file.mediaType.startsWith('image/')) {
      await presentImages([file]);
    }
  }

  const result2 = await generateText({
    model: openai('gpt-5'),
        messages: [
      ...(await result1.response).messages,
      { role: 'user', content: 'What is the weather there right now?' }
    ]
  });


  console.log(result2.text);
}

main().catch(console.error);
```

</details>

<details><summary><code>streamText()</code></summary>

```ts
import 'dotenv/config';
import { presentImages } from '../lib/present-image';

async function main() {
  const result1 = streamText({
    model: openai('gpt-5'),
    prompt:
      'Generate an image of an echidna swimming across the Mozambique channel.',
    tools: {
      image_generation: openai.tools.imageGeneration({
        outputFormat: 'webp',
        quality: 'low',
      }),
    },
  });

  for await (const part of result1.fullStream) {
    if (part.type == 'file' && part.file.mediaType.startsWith('image/')) {
      console.log('Image part', {
        mediaType: part.file.mediaType,
        length: part.file.base64.length,
      });
      await presentImages([part.file]);
    }
  }

  console.log();
  console.log('Finish reason:', await result1.finishReason);
  console.log('Usage:', await result1.usage);
  console.log();
  console.log((await result1.request).body);

  // make sure that multi-turn messages work
  const result2 = streamText({
    model: openai('gpt-5'),
    messages: [
      ...(await result1.response).messages,
      { role: 'user', content: 'What is the weather there right now?' }
    ],
  });

  console.log();
  console.log('Finish reason:', await result2.finishReason);
  console.log('Usage:', await result2.usage);
  console.log();
  console.log((await result2.request).body);
}

main().catch(console.error);
```

</details>

Which is most likely related to the fact that we don't include tool results in the results, we started to address that in https://github.com/vercel/ai/pull/8324

PR #7986, opened 2025-08-12
docs(examples): Add benchmark test for `streamText`

remorses:
I wanted to see what is the overhead of ai and try to improve it.

I added a benchmark that logs the troughput of the ai sdk procssing for 1 stream and 100 concurrent streams, these are the results:

  1. Single stream (10,000 tokens): ~2.39 MB/s
  2. 100 concurrent streams (100,000 tokens total): ~2.02 MB/s
  3. Stream with tool calls (5,000 tokens + tool chunks): ~4.01 MB/s


packages/ai/src/ui/default-chat-transport.ts
```diff
@@ -24,7 +24,7 @@ export class DefaultChatTransport<
       schema: uiMessageChunkSchema,
     }).pipeThrough(
       new TransformStream<ParseResult<UIMessageChunk>, UIMessageChunk>({
-        async transform(chunk, controller) {
+        transform(chunk, controller) {
```
gr2m: That change seems unrelated. Does it improve the throughput? We could land that in a separate PR.

gr2m (commented):
The benchmark tests lack expectations, right?

Running benchmark tests as part of test suits will also slow them down quite a bit. I like the idea of setting benchmark goals and testing against it to prevent performance regressions, but we should test it outside the standard unit/integration test suite.

packages/ai/src/ui/default-chat-transport.ts
remorses: yes it did improve by 13% by doing a quick test run

gr2m: Very cool! If you like, create a separate PR for that, together with your findings and how you tested. It will be much easier to land it outside the discussion of introducing benchmark tests.

lgrammel: does this impact ui performance in any way? from what i have seen rendering in the browser is the bottleneck, esp. with markdown, which is why improving the stream performance itself was not a high priority so far

remorses: Ui will still be the bottleneck for sure. Improving performance of server side processing will increase throughput per server more than the actual latency to the user. 

It will also make it faster to run evals if you cache responses. For example right now I use the cache middleware to run evals when prompt did not change and the ai processing still takes seconds what in my opinion should be milliseconds

lgrammel: agree, would be great getting the improvements added

lgrammel:
For the benchmark, I am wondering where to put it. Not sure I would want to include it as a test suite. We could put it under `examples/ai-core/src/benchmark` for now and convert into a regular script, or introduce `/benchmarks` but that might be a larger effort.

lgrammel (approved):

lgrammel:
prettier needs fixing

@gr2m lgtm, wdyt?

lgrammel:
also @remorses would be nice to get in those perf improvements you identified in a separate pr.

gr2m (approved):
This is great, thanks a lot!

PR #7987, opened 2025-08-12
feat(ai): preliminary tool results

### LINKED ISSUE #3063: Support Partial Tool Results / AsyncGenerator for Tool Calls in streamText
### Feature Description

For tool calls which take a couple of seconds to finish, it would be useful if partial tool results were supported so that the UI can reflect incremental progress.

For server side tools, this could be implemented by allowing `execute` to be an `AsyncGenerator`. This is similar to how the `generate` function in `streamUI` allows for intermediate updates to the UI:
https://github.com/vercel/ai/blob/8653fe5a9c01e7a9f9e0df135652bc5514054f4f/packages/ai/README.md?plain=1#L136-L141

For client side tools, it could be done with an `addPartialToolResult`.

```js
import { stream } from "fetch-event-stream"

onToolCall: async ({ toolCall }) => {

  if (toolCall.name == "searchWeb") {
	const events = await stream('http://websearch.tool?q=' + encodeURIComponent(toolCall.args.query), {
		method: 'POST',
		body: JSON.stringify({
			stream: true,
		}),
	});
	for await (const event of events) {
		addPartialToolResult({data: JSON.parse(event.data) })
	}
  }

}
```

In both cases, we could add a new stream part `toolResultDeltaStreamPart` that handles this. Similar to how partial tool calls `toolCallDeltaStreamPart` are supported when arguments are being generated:
https://github.com/vercel/ai/blob/8653fe5a9c01e7a9f9e0df135652bc5514054f4f/packages/ui-utils/src/stream-parts.ts#L334

### Use Case

Take for example a web search tool. The flow would be something like

```
{"message":"searching web"}
{"message":"links found!"}
{"message":"crawling links..."}
{"message":"results"}
```

Huggingface has a good reference implementation [here](https://github.com/huggingface/chat-ui/blob/main/src/lib/server/websearch/runWebSearch.ts) of that behavior, where it emits intermediate [updates to its UI](https://huggingface.co/chat/).

### Additional context

Tools implemented with edge functions typically time out if no response is given within a fixed limit like 10s or 30s.

A mix of returning an `event-stream` with `waitUntil` gets around that limitation.

```
      return new Response(readable, {
        headers: {
          "Content-Type": "text/event-stream",
          "Transfer-Encoding": "chunked",
          "content-encoding": "identity",
        },
      });
```

lgrammel:
## Background

Tools often need to send preliminary results, e.g. status information. We currently recommend using data parts for such use cases, but that requires creating a writer and related boilerplate code.

In AI SDK 5, we have a `toModelOutput` method, so tool outputs and what is being sent to LLMs can be easily decoupled. This opens the path for adding status information and more to tool results, and thus preliminary tool outputs with the same shape as the final output are possible without sending irrelevant information to the LLM.

## Summary

* Add support for `AsyncIterable` outputs to tool execute functions
* Add `preliminary` flag to streamed tool results
* Add `preliminary` flag to ui message tool results

## Example

```ts
tool({
  description: "Get the current weather.",
  inputSchema: z.object({ location: z.string() }),
  async *execute({ location }) {
    yield {
      status: "loading" as const,
      text: `Getting weather for ${location}`,
      weather: undefined,
    };

    await new Promise((resolve) => setTimeout(resolve, 3000));
    const temperature = 72 + Math.floor(Math.random() * 21) - 10;

    yield {
      status: "success" as const,
      text: `The weather in ${location} is ${temperature}°F`,
      weather: { temperature },
    };
  },
});
```

## Manual Verification

- [x] streamText console example
- [x] ui example

## Tasks

- [x] basic prototype
- [x] add streamText example
- [x] test streamText
- [x] fix type errors
- [x] add ui stream chunk support
- [x] add ui message support
- [x] add ui message support dynamic tools
- [x] changeset
- [x] add ui example
- [x] test generateText
- [x] fix any type inference
- [x] documentation tool reference
- [x] documentation tools page

## Related Issues
Fixes #3063


gr2m (approved):
love the use of generators and yields!

examples/next-openai/app/api/use-chat-tools/route.ts
```diff
@@ -16,21 +16,30 @@ export const maxDuration = 30;
...
     const assistantMessageCount = messages.filter(
       message => message.role === 'assistant',
     ).length;
 
-    if (assistantMessageCount <= 2) {
-      throw new Error('could not get weather information');
-    }
+    // if (assistantMessageCount <= 2) {
+    //   throw new Error('could not get weather information');
+    // }
```
gr2m, line 29: remove?

packages/provider-utils/src/types/execute-tool.ts
```diff
@@ -0,0 +1,27 @@
...
+> {
+  const result = execute(input, options);
+
+  if (isAsyncIterable(result)) {
+    let lastOutput: OUTPUT | undefined;
+    for await (const output of result) {
+      lastOutput = output;
+      yield { type: 'preliminary', output };
+    }
+    yield { type: 'final', output: lastOutput! };
```
gr2m, line 23: from call with Lars: last output is doubled so that output shows immediately in the UI

mattpocock:
@lgrammel @nicoalbanese Need some docs/advice on how best to consume this in the frontend.

lgrammel:
@mattpocock will add. not much should have changed. the tool result shape is the same (user controlled). you can check the `preliminary` flag if desired but most likely better solved through a user controlled discriminated union

Issue #6562, opened 2025-05-31
@ai-sdk/openai transcription response_format is missing from the providerOptions


ksanderer:
### Description

It seems that `response_format: 'verbose_json'` is missing from the OpenAI provider options at all:

https://github.com/vercel/ai/blob/2e6e5d31e0ed0726b60652bc00a5e59a349c3d2b/packages/openai/src/openai-transcription-model.ts#L147

Does this make `response.words?` unreachable for the OpenAI Whisper-1 model?

https://github.com/vercel/ai/blob/2e6e5d31e0ed0726b60652bc00a5e59a349c3d2b/packages/openai/src/openai-transcription-model.ts#L206

Am i missing something? 

![Image](https://github.com/user-attachments/assets/68e85d20-a937-4fae-b9bb-879b70463648)

https://platform.openai.com/docs/guides/speech-to-text#timestamps

### AI SDK Version

@ai-sdk/openai: 1.3.22
ai: 4.3.15

Abdillah:
The docs says transcriptions will be able to return segments data as timestamped transcription. But, it only happens when `response_format: 'verbose_json'`. Therefor, this is an essential option to be set a default on.

iamstevendao:
Upvote this. I have encountered a similar issue.

```
@ai-sdk/openai@2.0.0
ai@5.0.0
```

```js
const { segments } = await transcribe({
  model: openai.transcription('whisper-1'),
  audio: audioBuffer,
  providerOptions: {
    openai: {
      timestampGranularities: ['segment'],
    },
  },
});

// segments is empty

const { segments } = await openai.audio.transcriptions.create({
  file: new File([audioBuffer], 'audio.mp3', { type: 'audio/mpeg' }),
  model: 'whisper-1',
  timestamp_granularities: ['segment'],
  response_format: 'verbose_json',
});

// segments is not empty
```

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
This is a concrete failure in the existing transcription behavior: the SDK exposes timestamp granularities and segment/word results but omits the required OpenAI request format, making those results empty or unreachable.

### Evidence
- The OpenAI transcription adapter does not send `response_format: 'verbose_json'`, which Whisper-1 requires for timestamped segment/word data.
- Using AI SDK `timestampGranularities: ['segment']` returns an empty `segments` array, while the equivalent direct OpenAI request with `response_format: 'verbose_json'` returns segments.
[...]

PR #8011, opened 2025-08-13
feat(provider/openai): add response_format to be supported by default

dancer:
## background

openai transcription response_format parameter was missing from provider options, making word-level timestamps unreachable since they require response_format: 'verbose_json'

## summary

- add response_format parameter to transcription provider options  
- fix response parsing to handle segments from verbose_json format

## verification

- all tests pass including new response_format parameter tests
- word-level timestamps now accessible with responseFormat: 'verbose_json' + timestampGranularities: ['word']
- examples demonstrate working timestamped transcription

## tasks

- [x] response_format parameter added to openai-transcription-options.ts
- [x] tests for response_format parameter
- [x] documentation updated with verbose_json examples

related issue - #6562


packages/openai/src/transcription/openai-transcription-options.ts
```diff
@@ -24,6 +24,16 @@ export const openAITranscriptionProviderOptions = z.object({
...
 
+  /**
+   * The format of the output, in one of these options: json, text, srt, verbose_json, or vtt.
+   * For gpt-4o-transcribe and gpt-4o-mini-transcribe, the only supported format is json.
+   * @default 'json'
+   */
+  responseFormat: z
+    .enum(['json', 'text', 'srt', 'verbose_json', 'vtt'])
+    .default('json')
+    .optional(),
```
lgrammel: most formats would break our provider?

another option would be to always use `verbose_json` and not make this configurable

.changeset/real-coats-yawn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+feat(provider/openai): add response_format support
```
lgrammel: pr title and text here need to be updated

packages/openai/src/transcription/openai-transcription-model.test.ts
```diff
@@ -182,6 +182,62 @@ describe('doGenerate', () => {
...
+          Symbol(kLength): 40169,
+          Symbol(kType): "audio/wav",
+        },
+        "model": "whisper-1",
+        "response_format": "verbose_json",
+        "temperature": "0",
+        "timestamp_granularities": "segment",
+      }
+    `);
+  });
```
lgrammel, line 239: can you add tests for output parsing of segments?

lgrammel (approved):

PR #8032, opened 2025-08-13
test: replace `toStrictEqual()` with `toMatchInlineSnapshot()`

gr2m:
Just continuing some test refactoring inspired by #7898 as I learn the code base


gr2m:
🤣 nevermind I got totally off track need more coffee.

PR #8038, opened 2025-08-14
refactor(openai): consolidate model config into `getResponsesModelConfig()`

gr2m:
follow up to https://github.com/vercel/ai/pull/7798/files#r2254961289


lgrammel (approved):
maybe add a changeset even though it is a refactor, so we can trace it more easily in case something broke

Issue #8030, opened 2025-08-13
Mixing `web_search_preview` and `file_search` results in `invalid_value "url_citation"` type validation error


Irvel:
### Description

When I attempt to call both tools within a single `generateText` (or `streamText`) call, the Responses API returns annotations that include both `url_citation` and `file_citation` types. The SDK’s validator expects only `url_citation` in certain paths, which leads to a type validation error.

**Reproduction:**
```ts
const result = await generateText({
  model: openai('gpt-5'),
  system,
  prompt,
  tools: {
    web_search_preview: openai.tools.webSearchPreview({ searchContextSize: 'high' }),
    file_search: openai.tools.fileSearch({ vectorStoreIds: ['myvectorstoreid'], maxNumResults: 3 }),
  },
  providerOptions: {
      openai: {
        reasoningEffort: "high",
        textVerbosity: "low",
        serviceTier: "flex",
      },
    },
});
```

Error:
```
APICallError [AI_APICallError]: Failed to process successful response
    at postToApi (file:///repo/node_modules/@ai-sdk/provider-utils/dist/index.mjs:636:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    ... 2 lines matching cause stack trace ...
    at async file:///repo/node_modules/ai/dist/index.mjs:1426:22
    at async _retryWithExponentialBackoff (file:///repo/node_modules/ai/dist/index.mjs:1577:12)
    at async streamStep (file:///repo/node_modules/ai/dist/index.mjs:4719:15)
    at async fn (file:///repo/node_modules/ai/dist/index.mjs:5060:9)
    at async file:///repo/node_modules/ai/dist/index.mjs:1426:22 {
  cause: TypeError: First parameter has member 'readable' that is not a ReadableStream.
      at assertReadableStream (/repo/node_modules/web-streams-polyfill/dist/ponyfill.js:466:19)
      at convertReadableWritablePair (/repo/node_modules/web-streams-polyfill/dist/ponyfill.js:4035:9)
      at ReadableStream.pipeThrough (/repo/node_modules/web-streams-polyfill/dist/ponyfill.js:4119:29)
      at parseJsonEventStream (file:///repo/node_modules/@ai-sdk/provider-utils/dist/index.mjs:502:17)
      at file:///repo/node_modules/@ai-sdk/provider-utils/dist/index.mjs:785:12
[...]

PR #8047, opened 2025-08-14
fix(provider/openai): support file_citation annotations in responses api

dancer:
## background

when mixing `web_search_preview` and `file_search` tools in openai responses api, we threw schema validation errors because the response schema only supported `url_citation` annotations but the api was returning both `url_citation` and `file_citation` types

## summary

- add support for `file_citation` annotations in openai responses api schema
- update annotation processing to handle both citation types
- added tests for mixed citation scenarios

## verification

- all tests pass including new citation-specific tests
- mixed web_search_preview and file_search tools work without validation errors
- file citations map correctly to document sources

## tasks

- [x] update response schema to support discriminated union of url_citation and file_citation
- [x] update streaming annotation schema for both citation types
- [x] add doGenerate test for mixed citations
- [x] add doStream test for mixed citations  
- [x] add test for file_citation only scenario

related issue - #8030

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.ts
@@ -328,13 +328,22 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
                     text: z.string(),
                     logprobs: LOGPROBS_SCHEMA.nullish(),
                     annotations: z.array(
-                      z.object({
-                        type: z.literal('url_citation'),
-                        start_index: z.number(),
-                        end_index: z.number(),
-                        url: z.string(),
-                        title: z.string(),
-                      }),
+                      z.discriminatedUnion('type', [
+                        z.object({
+                          type: z.literal('url_citation'),
+                          start_index: z.number(),
+                          end_index: z.number(),
+                          url: z.string(),
+                          title: z.string(),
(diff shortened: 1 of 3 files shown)


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -933,13 +953,24 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
...
+                  id: self.config.generateId?.() ?? generateId(),
+                  url: value.annotation.url,
+                  title: value.annotation.title,
+                });
+              } else if (value.annotation.type === 'file_citation') {
+                controller.enqueue({
+                  type: 'source',
+                  sourceType: 'document',
+                  id: self.config.generateId?.() ?? generateId(),
+                  mediaType: 'text/plain',
```
lgrammel, line 969: is the media type always text/plain ?

lgrammel (approved):

packages/openai/src/responses/openai-responses-language-model.ts
dancer: i'll double check this but afaik yes

dancer: yeah it is

PR #8065, opened 2025-08-15
feat(ai): add validateUIMessages function

lgrammel:
## Background

Validating UI messages is often necessary, e.g. when they are loaded from the DB or received through REST requests.

## Summary

Add `validateUIMessages` function.

## Example

```ts
const messages = await validateUIMessages<UseChatToolsMessage>({
  messages: body.messages,
  tools,
});
```  

## Manual Verification

- [x] test ui tool calling example with validation

## Tasks

- [x] implementation TextUIPart
- [x] implementation ReasoningUIPart
- [x] implementation SourceUrlUIPart
- [x] implementation SourceDocumentUIPart
- [x] implementation FileUIPart
- [x] metadata
- [x] implementation StepStartUIPart
- [x] implementation  DataUIPart<DATA_TYPES>
- [x] implementation  DynamicToolUIPart
- [x] implementation ToolUIPart<TOOLS>
- [x] test provider metadata
- [x] test data ui part missing schema
- [x] documentation (reference)
- [x] documentation (guide)
- [x] example updates
- [x] changeset

## Related Issues

Fixes #6763

### DIFF
--- a/content/docs/07-reference/01-ai-sdk-core/32-validate-ui-messages.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/32-validate-ui-messages.mdx
@@ -1,0 +1,117 @@
+---
+title: validateUIMessages
+description: API Reference for validateUIMessages
+---
+
+# `validateUIMessages`
+
+`validateUIMessages` is an async function that validates UI messages against schemas for metadata, data parts, and tools. It ensures type safety and data integrity for your message arrays before processing or rendering.
+
+## Basic Usage
+
+Simple validation without custom schemas:
+
+```typescript
+import { validateUIMessages } from 'ai';
+
+const messages = [
+  {
+    id: '1',
+    role: 'user',
+    parts: [{ type: 'text', text: 'Hello!' }],
+  },
+];
+
+const validatedMessages = await validateUIMessages({
+  messages,
+});
+```
+
+## Advanced Usage
+
+Comprehensive validation with custom metadata, data parts, and tools:
+
+```typescript
+import { validateUIMessages, tool } from 'ai';
+import { z } from 'zod';
+
+// Define schemas
+const metadataSchema = z.object({
+  timestamp: z.string().datetime(),
+  userId: z.string(),
+});
+
+const dataSchemas = {
+  chart: z.object({
+    data: z.array(z.number()),
+    labels: z.array(z.string()),
+  }),
+  image: z.object({
+    url: z.string().url(),
+    caption: z.string(),
+  }),
+};
+
+const tools = {
+  weather: tool({
+    description: 'Get weather info',
+    parameters: z.object({
+      location: z.string(),
+    }),
+    execute: async ({ location }) => `Weather in ${location}: sunny`,
+  }),
+};
+
+// Messages with custom parts
+const messages = [
+  {
+    id: '1',
+    role: 'user',
+    metadata: { timestamp: '2024-01-01T00:00:00Z', userId: 'user123' },
+    parts: [
+      { type: 'text', text: 'Show me a chart' },
+      {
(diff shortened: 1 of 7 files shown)


imranbarbhuiya:
I tried this here https://github.com/imranbarbhuiya/ai/commit/caccd036bfe8b348422919fbeb2afe2b02c4b67f#diff-359eed88eef8047dde41894dee7f556365015932d5dfcdb994e5c13db9c2c1fb but there were some deep typing issues in test so I stopped, but it has most of the schemas, u can probably use some from there

packages/ai/src/ui/validate-ui-messages.ts
```diff
@@ -0,0 +1,128 @@
...
+  url: z.string(),
+  providerMetadata: providerMetadataSchema.optional(),
+});
+
+const stepStartUIPartSchema = z.object({
+  type: z.literal('step-start'),
+});
+
+const dataUIPartSchema = z.object({
+  type: z.string().startsWith('data-'),
```
imranbarbhuiya, line 62: since u r using zod v4, u can use

```suggestion
  type: z.templateLiteral([z.literal('data-'), z.string()]),
```

lgrammel: and make it less readable when it is not needed? why would i do that?

imranbarbhuiya: it gives proper type like the interface currently has `data-${string}` instead of just string, other then than yeah current one is better readable

lgrammel: here we wont push everything into zod because the dynamic union types are too complex, so it doesnt matter, hence the simpler version

content/docs/07-reference/01-ai-sdk-core/32-validate-ui-messages.mdx
```diff
@@ -0,0 +1,117 @@
...
+- **Error handling**: Throws `TypeValidationError` for invalid data or missing schemas
+- **Flexible validation**: Validates metadata, custom data parts, and tool inputs/outputs independently
+
+## Validation Behavior
+
+- **Messages structure**: Always validates basic message structure (id, role, parts)
+- **Metadata**: Only validated if `metadataSchema` is provided
+- **Data parts**: Only validated if matching schema exists in `dataSchemas`
+- **Tools**: Only validated if matching tool exists in `tools` object
+- **Missing schemas**: Throws error if data/tool parts exist without corresponding schemas
```
nicoalbanese: think we can remove this - doesn't really add anything beyond obvious

examples/next-openai/app/api/use-chat-tools/route.ts
```diff
@@ -84,7 +85,12 @@ export type UseChatToolsMessage = UIMessage<
 >;
 
 export async function POST(req: Request) {
-  const { messages } = await req.json();
+  const body = await req.json();
+
+  const messages = await validateUIMessages<UseChatToolsMessage>({
```
nicoalbanese, line 90: if you're passing in tools, do you need to pass generic argument? or will help with type mismatch if no metadata, data parts passed in?

lgrammel: yes generics are not inferred

lgrammel: our best practice is defining a ui message type so it should be easy

nicoalbanese (approved):

examples/next-openai/app/api/use-chat-tools/route.ts
nicoalbanese: ack

PR #6750, opened 2025-06-15
onUpdateToolsList

eliaweiss:
🛠️ Add onUpdateToolsList callback to streamText function
Overview
This PR introduces a new onUpdateToolsList callback to the streamText function, enabling dynamic tool management during multi-step conversations. This feature allows updating available tools and optionally providing system instructions after each tool execution step.
🚀 Key Features
1. Dynamic Tool Updates
Update tools dynamically after each tool execution step
Maintain full backward compatibility with existing code
Support both legacy format (tools only) and new format (tools + system instruction)
2. System Instruction Support
Optionally provide system instructions alongside tool updates
System instructions are inserted before response messages in the prompt flow
Clean separation of concerns with parameter-based approach
3. Type Safety
Full TypeScript support with proper type definitions
Maintains type inference for tool calls and results
Uses @ts-ignore strategically where type system limitations exist
📝 API Usage
Apply to ai-4.3.17.tg...
;
🏗️ Implementation Details
Message Flow Architecture
Apply to ai-4.3.17.tg...
;
Type Definition
Apply to ai-4.3.17.tg...
;
🔄 Workflow
Tool calls are executed in current step
onUpdateToolsList callback is invoked with tool execution context
Callback can return updated tools and optional system instruction
System instruction is inserted before response messages
Next step proceeds with updated tools and enhanced context
✅ Benefits
Dynamic Tool Management: Adapt available tools based on conversation context
Enhanced Context: Provide relevant instructions when tools change
Backward Compatible: Existing code continues to work unchanged
Type Safe: Full TypeScript support with proper inference
Clean Architecture: Parameter-based approach vs message pollution
🧪 Testing
✅ Builds successfully with TypeScript
✅ Maintains backward compatibility
✅ Supports both callback formats
✅ Proper error handling for failed tool updates

Issue #8000, opened 2025-08-12
V5 broke the entier API WTF


eliaweiss:
### Description

V5 broke the entier API WTF

eliaweiss:
ever hear of backward compatability?

lgrammel:
## Identify issue type

Identified as **None**  
Confidence: **Medium**

### Rationale
This is an unclear complaint rather than an actionable report of a specific SDK defect or a defined feature request.

### Evidence
- The description only states that V5 “broke the entire API” without identifying a specific API, failure, expected behavior, or reproduction.
- The comment mentions backward compatibility but provides no concrete technical details.

PR #8071, opened 2025-08-15
Enable Updating Tool Set in prepareStep for Improved Performance

eliaweiss:
---

## Background

This PR continues the work from [#6750](https://github.com/vercel/ai/pull/6750), originally targeting v4.
In v5, the `prepareStep` callback partly addressed the issue by allowing modification of `activeTools` after a tool call. However, it still does not provide a way to change the **original** tool set once the process starts.

In theory, I could send the full tool set (120 tools) to `streamText`, but this significantly impacts performance. From my measurements, a sequence of 3 tool calls takes around **14 seconds** with the full tool set, compared to **7 seconds** with this fix.

This change allows passing an updated tool set to the result of `prepareStep`, enabling more efficient tool handling without resending the entire list each time.

> **Note:** These results suggest there may be a deeper inefficiency in tool handling within the SDK. Even when only a small subset of tools is active, sending the full set still appears to slow down processing.

---

## Summary

* Added the ability to pass a new tool set to the result of `prepareStep`.
* Improves performance in cases where only a small subset of tools is required after initial calls.
* Avoids the overhead of repeatedly sending the full tool list.

---

## Manual Verification

* Tested with a scenario involving 120 tools and multiple consecutive tool calls.
* Measured execution time before and after the change:

  * **Before:** \~14 seconds for 3 consecutive calls
  * **After:** \~7 seconds for the same workload

---

## Tasks

* [ ] Add/update tests for the new behavior
* [ ] Update documentation where relevant
* [ ] Add a *patch* changeset (`pnpm changeset`)
* [ ] Fix any formatting issues (`pnpm prettier-fix`)

---

## Future Work

* Investigate the root cause of performance degradation when sending the full tool list.
* Explore further optimizations in tool management within the SDK.

---

## Related Issues

* Continues [#6750](https://github.com/vercel/ai/pull/6750)

---


lgrammel:
Changing the tools was never allowed because of type safety (which is why there are activeTools instead).

To address the underlying issue, I recommend first creating a benchmark that shows the inefficiency and then identifying the actual reason (e.g. missing narrowing of tools to active tools when it comes to some validation)

lgrammel:
image

lgrammel:
also https://github.com/vercel/ai/issues/8000

please stay civil we are all trying to do our work here

PR #8091, opened 2025-08-16
fix(bedrock): modify payload for cohere.* models.

Ray0907:
## Background

Bedrock Cohere embedding models (cohere.*) might use a different API payload format compared to Titan models. The current implementation sends {inputText: string} payload for all Bedrock embedding models, but Cohere models appear to expect {texts: string[], input_type: string}. This could cause API validation errors when using Cohere embedding models through the AI SDK.
## Summary

Added conditional payload formatting for Bedrock Cohere embedding models:

checking if a model ID starts with 'cohere.' and then handling values accordingly.

## Manual Verification

Tested with Cohere embedding model using the following code:
```typescript
import { embed } from 'ai';

const { embedding } = await embed({
    model,
    value: ['hello world']
  });
```
encountered the error
>  Error: undefined: Malformed input request: #: required key [images] not found#: required key [texts] not found#: required key [texts] not found#: required key [texts] not found#: required key [texts] not found#: required key [input_type] not found#: extraneous key [inputText] is not permitted, please reformat your input and try again.
## Tasks

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-embedding-model.ts
+++ b/packages/amazon-bedrock/src/bedrock-embedding-model.ts
@@ -70,7 +70,7 @@ export class BedrockEmbeddingModel implements EmbeddingModelV2<string> {
 
     // https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html
     const args = {
-      inputText: values[0],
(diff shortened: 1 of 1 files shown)


owen-yap:
Thanks for this

Issue #309, opened 2023-07-10
streamText: add callback for handling aborted streams


jvandenaardweg:
Now we have stream cancellation working properly (https://github.com/vercel-labs/ai/issues/90#issuecomment-1618915409 ) I think we need a way to detect and handle the cancellation server side.

In the current situation when a user cancels a stream, the [server errors](https://github.com/vercel-labs/ai/issues/90#issuecomment-1619424795) and there's no way to catch the cancellation and handle what should happen on the server. But I want to know how many tokens are generated when the user cancels the stream, so I can then store that information in my external database.

Currently we have `onStart`, `onCompletion` and `onToken` callback methods on `OpenAIStream()` / `AIStream()`. Ideally a new callback method would be great, something like `onStop` (to follow the `stop()` method naming in the React Hooks) or `onCancel`, which passes through the completion text at that point and allows to calculate the tokens and write that to a external database.

Or a different way to just catch the abort error in our (Edge) Functions.

From my understanding, I guess we need to postpone the abort until whatever method is doing something async is done, otherwise the serverless function might not exist anymore.

Is this even possible?

hassanbazzi:
Having the same issue. This is very important.

gartia:
This would be nice. to have an OnStop. I am finding it annoying to deal with function calls, when i want to abort after a function call. But i getting errors from using abort controllers via server methods.

lassegit:
Is it really necessary to have all these callbacks? Having `onStop(completion, error)` call should be enough. My problem is that I don't have a callback when the abort signal is called. I need a callback to be able to track token usage.

GorvGoyl:
What are some current workarounds till the team fixes the issue?

GorvGoyl:
related issue https://github.com/vercel/next.js/discussions/48381

peterdresslar:
Hello @GorvGoyl, are you calling the OpenAI using the OpenAI Node library (as documented in the examples from this SDK) or using your own custom logic to access the API?

GorvGoyl:
> Hello @GorvGoyl, are you calling the OpenAI API using the OpenAI Node library (as documented in the examples from this SDK) or using your own custom logic to access the API?

I'm using vercel edge function like this:

```
 const completion = await openai.chat.completions.create({
            model: 'gpt-3.5-turbo',
            stream: true,
            messages: [{ role: 'user', content: question }],
        });

        const stream = OpenAIStream(completion, {...});
[...]

PR #8106, opened 2025-08-17
Failing Test: streamText does not call onAbort if the abort signal is triggered during a tool call.

istarkov:
## Background

This PR adds a failing test case showing that AbortController does not trigger onAbort when the abort signal is executed during a tool call.

## Summary

Just failing test.

## Manual Verification

We encountered this bug in our product: calling .abort() sometimes does not trigger onAbort as expected.

## Tasks

## Future Work

Fix it

## PS

I feel the current abort implementation need additional checks.
The issue is that the stream consumer may already be closed due to the abort.
For example, if the client disconnects, nothing is consuming the stream, so probably the check in the stream transform will never be executed.

I mean this code
```
    stream = filterStreamErrors(stream, ({ error, controller }) => {
      if (isAbortError(error) && abortSignal?.aborted) {
        onAbort?.({ steps: recordedSteps });
        controller.enqueue({ type: 'abort' });
        controller.close();
      } else {
        controller.error(error);
      }
    });
```

## Related Issues

https://github.com/vercel/ai/issues/309#issuecomment-3194333386
https://github.com/vercel/ai/issues/8107

### DIFF
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -12449,6 +12449,123 @@ describe('streamText', () => {
         `);
       });
     });
+
+    describe('abort in 2nd step during tool call', () => {
+      let result: StreamTextResult<any, TextStreamPart<any>>;
+      let onErrorCalls: Array<{ error: unknown }> = [];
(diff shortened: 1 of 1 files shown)


lgrammel:
Continued in #8197

PR #8125, opened 2025-08-18
fix(provider/mistral): inject JSON generation instruction when response format is JSON

lgrammel:
## Background

Mistral object generation was broken in AI SDK 5, because the JSON generation instruction was no longer added to the system prompt.

## Manual Verification

* Ran `examples/ai-core/src/generate-object/mistral.ts`
* Ran `examples/ai-core/src/stream-object/mistral.ts`

## Future Work

- Add support for Mistral `json_schema` [response format](https://docs.mistral.ai/api/#tag/chat/operation/chat_completion_v1_chat_completions_post)
- Add injection support to OpenAI compatible provider and apply in DeepSeek provider

## Related Issues

Fixes https://github.com/vercel/ai/issues/8119

### DIFF
--- a/packages/mistral/src/mistral-chat-language-model.ts
+++ b/packages/mistral/src/mistral-chat-language-model.ts
@@ -12,7 +12,7 @@ import {
   createJsonResponseHandler,
   FetchFunction,
   generateId,
-  injectJsonInstruction,
+  injectJsonInstructionIntoMessages,
   parseProviderOptions,
   ParseResult,
   postJsonToApi,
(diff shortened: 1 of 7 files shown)


gr2m (approved):
we paired on it

packages/mistral/src/mistral-chat-language-model.ts
```diff
@@ -109,6 +110,7 @@ export class MistralChatLanguageModel implements LanguageModelV2 {
       });
     }
 
+    // TODO remove when we have JSON schema support (see OpenAI implementation)
```
gr2m, line 113: 🔜  https://github.com/vercel/ai/issues/8127

wong2:
Got this error after upgrading ai sdk deps today

`The requested module '@ai-sdk/provider-utils' does not provide an export named 'injectJsonInstructionIntoMessages'`

I guess it's related to this change

PR #8136, opened 2025-08-19
feat(baseten): Add Baseten as a Provider to AI SDK providers

AlexKer:
## Summary

This change adds Baseten to the AI SDK providers list. The provider supports:

- **Model APIs**: OpenAI-compatible chat models via default inference endpoints  
- **Custom Deployments**: Dedicated model deployments with /sync/v1 endpoints
- **High-Performance Embeddings**: 12x faster embedding generation via Baseten's Performance Client

Users can import `@ai-sdk/baseten` and access all Baseten capabilities through a consistent interface.

## Manual Verification

- 24/24 unit tests passing (mocked API responses)
- 12/12 integration tests passing (real API calls with embeddings + chat)
- Tested both /sync and /sync/v1 endpoint formats
- Verified Performance Client integration for embeddings
- Internal review by Baseten engineering team

## Tasks

## Related Issues


lgrammel:
Hi! Can you publish the baseten provider in your github repository and your npm module? we would be happy to list it in our community providers section if you send a docs PR.

AlexKer:
@lgrammel Hi Lars, appreciate you for opening this back up. lmk if you'd like any changes or have questions!

content/providers/01-ai-sdk-providers/170-baseten.mdx
```diff
@@ -0,0 +1,241 @@
+---
+title: Baseten
+description: Learn how to use Baseten models with the AI SDK.
+---
+
+# Baseten Provider
+
+[Baseten](https://baseten.ai/) is an inference platform for serving frontier, enterprise-grade open source AI models[API](https://docs.baseten.co/overview).
```
gr2m: do you want to stick to `baseten.ai` or `baseten.co`? fyi `baseten.ai` wouldn't load for me about 5 min ago, now it redirects to baseten.co

content/providers/01-ai-sdk-providers/170-baseten.mdx
```diff
@@ -0,0 +1,241 @@
...
+
+- **baseURL** _string_
+
+  Use a different URL prefix for API calls, e.g. to use proxy servers.
+  The default prefix is `https://inference.baseten.co/v1`.
+
+- **apiKey** _string_
+
+  API key that is being sent using the `Authorization` header. It defaults to
+  the `BASETEN_API_KEY` environment variable. It is recommended you set the environment variable using `export` so you do not need to include the field everytime.
```
gr2m, line 55: maybe link to https://app.baseten.co/settings/api_keys from here or further up so folks have a direct link from where to get an API key?

packages/baseten/src/baseten-provider.integration.test.ts
```diff
@@ -0,0 +1,307 @@
+import { describe, it, expect, beforeAll } from 'vitest';
+import { baseten, createBaseten } from '@ai-sdk/baseten';
+import { embed, embedMany, generateText } from '../../ai';
+
+// You need to fill these values to run this test suite
+const EMBEDDING_MODEL_ID = '03y7n6e3';
+const CHAT_MODEL_ID = '6wg17egw';
```
gr2m: are these specific to an account? All the integration tests failed for me

gr2m (commented):
changes look mostly good, thanks! I'm not sure about the integration tests that require an API key, I'll chat with Lars. Just have a few notes

content/providers/01-ai-sdk-providers/170-baseten.mdx
AlexKer: thanks for catching! edited to be `baseten.co` for consistency

packages/baseten/src/baseten-provider.integration.test.ts
AlexKer: You would need to deploy your own dedicated embeddings and chat model on Baseten and replace these values with your model ID as specified on line 5 to run the integration test. This is mostly for internal sanity check and something I wanted to add for future maintainers. All tests were passing as of last month. The other providers only provided unit tests with mock URLs, so happy to remove this as well.

gr2m: okay I talked to Lars, and we would prefer you remove these tests and add `examples/*/baseten*.ts` files instead, for consistency.

packages/baseten/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "dependencies": {
+    "@ai-sdk/openai-compatible": "workspace:*",
+    "@ai-sdk/provider": "workspace:*",
+    "@ai-sdk/provider-utils": "workspace:*",
+    "@basetenlabs/performance-client": "^0.0.8"
+  },
+  "devDependencies": {
+    "@types/node": "20.17.24",
+    "@vercel/ai-tsconfig": "workspace:*",
+    "ai": "workspace:*",
```
gr2m: we can't do that, it's a circular dependency

packages/baseten/package.json
```diff
@@ -0,0 +1,68 @@
...
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js"
+    }
+  },
+  "dependencies": {
+    "@ai-sdk/openai-compatible": "workspace:*",
+    "@ai-sdk/provider": "workspace:*",
+    "@ai-sdk/provider-utils": "workspace:*",
+    "@basetenlabs/performance-client": "^0.0.8"
```
gr2m, line 38: could you clarify why this is needed? We will need to install the package server-side for gateway and need to make sure we understand the impact

gr2m (changes requested):
blocking merge because we are in pre-release mode which means merging requires a bit of extra work from our side.

packages/baseten/package.json
AlexKer: We need the performance client because the existing `OpenAICompatibleEmbeddingModel` is sending large batches of 2048 strings for embedding, but Baseten has a maximum batch size limit of 128 strings. The `OpenAICompatibleEmbeddingModel` approach will fail with HTTP 422 errors. The performance client acts as an intelligent "reverse splitter/fanout" that takes large batches and distributes them into smaller chunks (e.g., 32 batches of 64 strings) across multiple replicas/machines, enabling parallel processing and faster throughput. https://www.baseten.co/blog/your-client-code-matters-10x-higher-embedding-throughput-with-python-and-rust/

packages/baseten/src/baseten-provider.integration.test.ts
AlexKer: sounds good, addressed this in my latest commit.

packages/baseten/package.json
michaelfeil: Hey @gr2m, the package is cross compiled to all platforms with node 14+. The package has no dependecies to other node packages (rust binary bundled with NAPI) 

Its mostly a performance optimization. For embeddings, splitting up e.g. 2048 embeddings into sizes of 4, results in better horizontal load balancing (you can queue them, estimate number of replicas for your deployment better). 

Adding an additional hope (reverse proxy as alex mentioned above) leads to issues with propagating cancellation. 

Usually, this results in around ~4x thoughput for the customer. We have a customer that runs ~1M RPM on this package.

packages/baseten/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "build:watch": "pnpm clean && tsup --watch",
+    "clean": "rm -rf dist *.tsbuildinfo",
+    "lint": "eslint \"./**/*.ts*\"",
+    "type-check": "tsc --build",
+    "prettier-check": "prettier --check \"./**/*.ts*\"",
+    "test": "pnpm test:unit && pnpm test:integration",
+    "test:update": "pnpm test:unit -u",
+    "test:watch": "vitest --config vitest.node.config.js",
+    "test:unit": "vitest --config vitest.node.config.js --run --exclude src/baseten-provider.integration.test.ts",
+    "test:integration": "vitest --config vitest.node.config.js --run src/baseten-provider.integration.test.ts"
```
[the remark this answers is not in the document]
AlexKer: updated.

packages/baseten/package.json
gr2m: fyi @shaper any questions / concerns?

shaper: There is support for specifying `maxEmbeddingsPerCall` in the `OpenAICompatibleEmbeddingModel` [here](https://github.com/vercel/ai/blob/main/packages/openai-compatible/src/embedding/openai-compatible-embedding-model.ts#L54). If we set that to 128 would that help? I think you can do it when you create the embedding model in your provider class.

However I do get that folks would generally benefit from using this performance client when interacting with your service.

The package should be fine for AI Gateway. We could try packing and testing it, but I don't see embedding models currently available in Baseten as serverless models we can test without a deploy.

AlexKer: We do want to use the client for performance reasons, so would prefer to keep if possible. I've also refactored the code to support that. You are correct, embeddings are only available through dedicated deployments today.

.changeset/quiet-pens-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/baseten': patch
```
gr2m: For initial release

```suggestion
'@ai-sdk/baseten': major
```

packages/baseten/package.json
```diff
@@ -0,0 +1,67 @@
+{
+  "name": "@ai-sdk/baseten",
+  "version": "1.0.0-beta.2",
```
gr2m: for initial release

```suggestion
  "version": "0.0.0",
```

packages/baseten/package.json
michaelfeil: @shaper Happy to provide replicas as needed for free. 

Two things:
- optimal packing. e.g. we have a character per request policy. Using this, you can send a list of say 2048 in say request that correspond to ca. 4000 characters per request. https://github.com/basetenlabs/truss/blob/4e770823ff20aacd96669f68ff726f609e01c309/baseten-performance-client/core/src/split_policy.rs#L27 
- request hedging: If the request does not succeed until 10.0s beyond start, you can optionally hedge it against a running request and yield the first requests. Avoids full batch retries if e.g. 1 request fails. 
- indiviual request retry: e.g. on 429, only individual requests are backed off.

shaper: Does the performance client support/respect the custom `fetch` passed to the provider, and the abort signal that can be optionally passed to the `embed` or `embedMany` functions? https://ai-sdk.dev/docs/reference/ai-sdk-core/embed#abort-signal

michaelfeil: We don't have abort signal at this point.

gr2m (approved):
I tested the examples, and made updates for the initial release. This should be good to ship now

PR #8181, opened 2025-08-20
feat(provider/gateway): Expose model type in model spec

R-Taneja:
## Background

Re: https://vercel.slack.com/archives/C08CCJK0484/p1755642299328509 @haydenbleasel 

## Summary

Exposes the model type in the model specification as a field

## Manual Verification

```
const { models } = await gatewayProvider.getAvailableModels();
const textModels = models.filter(m => m.modelType === 'language');
const embeddingModels = models.filter(m => m.modelType === 'embedding');
```

This code works now

## Tasks

### DIFF
--- a/packages/gateway/src/gateway-fetch-metadata.ts
+++ b/packages/gateway/src/gateway-fetch-metadata.ts
@@ -44,7 +44,6 @@ const gatewayLanguageModelSpecificationSchema = z.object({
   specificationVersion: z.literal('v2'),
   provider: z.string(),
   modelId: z.string(),
-  type: z.enum(['language', 'embedding', 'image']).optional(),
 });
 
 const gatewayLanguageModelPricingSchema = z.object({
@@ -58,6 +57,7 @@ const gatewayLanguageModelEntrySchema = z.object({
   description: z.string().nullish(),
   pricing: gatewayLanguageModelPricingSchema.nullish(),
   specification: gatewayLanguageModelSpecificationSchema,
+  modelType: z.enum(['language', 'embedding', 'image']).optional(),
 });
 
 const gatewayFetchMetadataSchema = z.object({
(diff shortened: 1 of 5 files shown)


packages/gateway/src/gateway-model-entry.ts
```diff
@@ -40,4 +40,6 @@ export interface GatewayLanguageModelEntry {
 export type GatewayLanguageModelSpecification = Pick<
   LanguageModelV2,
   'specificationVersion' | 'provider' | 'modelId'
->;
+> & {
+  type?: 'language' | 'embedding' | 'image';
```
shaper: I don't think we should put this here -- it is not part of the AI SDK LanguageModelV2 type/specification, which is what this was used for (it is now obsolete and deprecated, the `GatewayLanguageModelSpecification` type and fields), and is by its name and use specifically for language models, but here we are defining other model types.

Sadly, the type above, `GatewayLanguageModelEntry` is also language model specific by name 😬 

Perhaps we make a type alias to map `GatewayLanguageModelEntry` to just `GatewayModelEntry`, and make this `modelType` field on that?

R-Taneja: I updated the solution. It works in conjunction with https://github.com/vercel/ai-gateway/pull/327

I thought of a few slightly different interpretations of your feedback, but this is what I think is the cleanest solution/best interpretation. Let me know if you had something else in mind.

packages/gateway/src/gateway-model-entry.ts
```diff
@@ -35,6 +35,10 @@ export interface GatewayLanguageModelEntry {
    * Additional AI SDK language model specifications for the model.
    */
   specification: GatewayLanguageModelSpecification;
+  /**
```
shaper, line 39: nit: 1 blank above

packages/gateway/src/gateway-fetch-metadata.ts
```diff
@@ -57,6 +57,7 @@ const gatewayLanguageModelEntrySchema = z.object({
   description: z.string().nullish(),
   pricing: gatewayLanguageModelPricingSchema.nullish(),
   specification: gatewayLanguageModelSpecificationSchema,
+  modelType: z.enum(['language', 'embedding', 'image']).optional(),
```
shaper: AI SDK prefers `nullish` -- may need edits to code where we reference, but best for us to follow esp. since we use nullish in other fields above

shaper (approved):

PR #8196, opened 2025-08-21
make prompt type bit strict

Rajaniraiyn:
## Background

This restriction is mentioned in docs but enforced in types

## Summary

Updated `Prompt` type to enforce either `messages` or `prompt` fields at type level


lgrammel:
Did not do this on purpose to not affect autocomplete.

lgrammel:
maybe a `never` trick will do both, i'll build on it

lgrammel:
continued in #8198

PR #8198, opened 2025-08-21
feat(ai): improved type checking for prompt/messages input

lgrammel:
## Background

It was possible to enter both messages and prompt at the same time. While this was checked at runtime and threw errors, proper type validation would be preferable, as long as it does not affect autocomplete in the IDE.

## Summary

Enhance the `Prompt` type to limit to either messages or prompt, while retaining autocomplete through the use of `never`.

## Manual Verification

Tested autocomplete and error behavior using examples.

## Related Issues

Continues #8196


gr2m (approved):
Looks good :shipit:

PR #8206, opened 2025-08-21
feat(codemod): add zod import v3 transformation

dancer:
## background

users need to import zod v3 explicitly for compatibility with certain environments and to avoid version conflicts when multiple zod versions are present

## summary

- add codemod to transform `import z from 'zod'` to `import { z } from 'zod/v3'`

## verification

- codemod test passes
- transformation preserves comments and only affects default imports named 'z'
- included in v5 upgrade bundle

## tasks

- [x] create v5/replace-zod-import-with-v3 codemod
- [x] add test fixtures and test file
- [x] add to v5 bundle in upgrade.ts
- [x] verify codemod works correctly

### DIFF
--- a/packages/codemod/src/test/__testfixtures__/replace-zod-import-with-v3.output.ts
+++ b/packages/codemod/src/test/__testfixtures__/replace-zod-import-with-v3.output.ts
@@ -1,0 +1,21 @@
+// @ts-nocheck
+import { z } from 'zod/v3';
+import { generateText } from 'ai';
+
+const schema = z.object({
+  name: z.string(),
+  age: z.number(),
+});
+
+const result = await generateText({
+  model: openai('gpt-4'),
+  prompt: 'Generate a person',
+  schema,
+});
+
+// Other imports should not be affected
+import x from 'other-package';
+import { y } from 'zod';
+
+// Default import with different name should not be transformed
+import zod from 'zod';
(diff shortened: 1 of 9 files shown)


packages/codemod/src/test/__testfixtures__/replace-zod-import-with-v3.output.ts
```diff
@@ -0,0 +1,21 @@
...
+  prompt: 'Generate a person',
+  schema,
+});
+
+// Other imports should not be affected
+import x from 'other-package';
+import { y } from 'zod';
+
+// Default import with different name should not be transformed
+import zod from 'zod';
```
gr2m: I'm not sure about these two. Are there other imports from `zod` that should now be imported form `zod/v3` - can you please double check if you haven't yet?

If we only want to import `z` from `zod/3`, but there are other imports that people should keep importing from `zod`, then I'd cover that case to fixtures. It would need to split up a single import into 2

Same about the default import. Is that even a thing that is possible right now?

gr2m (commented):
See my comment. I'll also defer final approval to @lgrammel since Lars is working on related work right now

lgrammel:
lets wait until the other zod changes is done to see if we want to migrate to zod/v3 or zod/v4

lgrammel (approved):
lets go with zod/v3 for now until we understand the exact differences in the zod/v4 json schema mapping

aversini:
@dancer / @lgrammel thanks for the codemod! 

While the [docs](https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#typescript-performance-issues-with-zod) are for consumers and you recommend v3 over v4, I can see a lot of v4 still being used throughout the project. Shouldn't "ai" and all its packages also convert to importing ONLY v3?

The reason I found this is because v4 is way bigger in size than v3 and when I analyze the chunk created for "ai", I can see both v3 and v4:

Screenshot 2025-08-31 at 6 19 39 PM

It may be worth a shot to convert the whole code base to v3...

PR #8220, opened 2025-08-22
fix(ai): support loop breaking behavior in async iterable stream

lgrammel:
## Background

Async iterators can be exited early, e.g. by breaking out of the for loop. When that happened, the async iterator stream was not correctly closed.

## Summary

* implement `return` and `throw` async iterable methods to support loop-breaking behavior
* add jsdoc

## Future Work

* Investigate/review other parts of #7902

## Related Issues

Build on #7902


packages/ai/src/util/async-iterable-stream.ts
```diff
@@ -1,16 +1,87 @@
...
+
+    let finished = false;
+
+    /**
+     * Cleans up the reader by cancelling and releasing the lock.
+     */
+    async function cleanup() {
+      finished = true;
+      try {
+        await reader.cancel?.();
```
lgrammel: @remorses this is now also called from the regular `done`, which might be a bug that i might have introduced. would calling cancel on done be okay or would it need to be avoided (in which case i would add another test case for that scenario)?

remorses: cancel should not be called when the stream completes normally, even if it works in Node there could be other streams implementation like Bun or Deno that fail in this case.

.changeset/shy-shoes-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): support loop breaking behavior in async iterable stream
```
lgrammel, line 5: need to talk about risk of changing behavior (breaking change)

lgrammel:
need to check behavior on deno / bun

remorses:
That was the only problematic part. I don't think Bun or Deno would treat cancel differently

PR #8231, opened 2025-08-22
fix(ai): make `chat.addToolResult()` compatible with dynamic tool calls

### LINKED ISSUE #7896: Add `addToolResult` as input for `onToolCall`.
### Description

The current API design for the `useChat` hook requires the `onToolCall` handler to rely on the `addToolResult` function returned from the hook's outer scope.

#### Current Pattern

```ts
const { addToolResult } = useChat({
    // Logic here uses `addToolResult` from the closure
    const result = await executeToolCall(toolCall);
    addToolResult({
      toolName: toolCall.toolName,
      result,
    });
  },
});
```

This dependency makes it difficult to extract the `onToolCall` logic into a separate, reusable function, especially for complex implementations. The common workaround is to pass `addToolResult` manually, which is cumbersome.

#### Proposed Change

I suggest passing `addToolResult` as a parameter directly to the `onToolCall` callback. This would make the handler self-contained and significantly improve modularity.

```ts
useChat({
  async onToolCall({ toolCall, addToolResult }) {
    const result = await executeToolCall(toolCall);
    addToolResult({
      toolName: toolCall.toolName,
      result,
    });
  },
});
```

### AI SDK Version

ai: 5.0.5

dvoytenko:
## Background

`dynamic-tool` parts are ignored in the current version of `addToolResult`, but following other APIs, it should be supported just as any other tool.

## Summary

Add support for dynamic tools in `chat.addToolResult()`.

## Future Work

- Revisit function name `isToolOrDynamicToolUIPart` and similar in AI SDK 6, aiming for different naming re static/dynamic tools with Tools as unified name.

## Related Issues

Fixes #7896

### DIFF
--- a/packages/ai/src/ui/ui-messages.ts
+++ b/packages/ai/src/ui/ui-messages.ts
@@ -293,7 +293,7 @@ export function isToolOrDynamicToolUIPart<TOOLS extends UITools>(
 }
 
 export function getToolName<TOOLS extends UITools>(
-  part: ToolUIPart<TOOLS>
+  part: ToolUIPart<TOOLS>,
 ): keyof TOOLS {
   return part.type.split('-').slice(1).join('-') as keyof TOOLS;
 }
--- a/packages/ai/src/ui/ui-messages.ts
+++ b/packages/ai/src/ui/ui-messages.ts
@@ -280,10 +280,31 @@ export function isToolUIPart<TOOLS extends UITools>(
   return part.type.startsWith('tool-');
 }
 
+export function isDynamicToolUIPart(
+  part: UIMessagePart<UIDataTypes, UITools>,
+): part is DynamicToolUIPart {
+  return part.type === 'dynamic-tool';
+}
+
+export function isToolOrDynamicToolUIPart<TOOLS extends UITools>(
+  part: UIMessagePart<UIDataTypes, TOOLS>,
+): part is ToolUIPart<TOOLS> | DynamicToolUIPart {
+  return isToolUIPart(part) || isDynamicToolUIPart(part);
+}
+
 export function getToolName<TOOLS extends UITools>(
-  part: ToolUIPart<TOOLS>,
+  part: ToolUIPart<TOOLS> | DynamicToolUIPart,
 ): keyof TOOLS {
   return part.type.split('-').slice(1).join('-') as keyof TOOLS;
+}
+
+export function getToolOrDynamicToolName(
+  part: ToolUIPart<UITools> | DynamicToolUIPart,
+): string {
+  if (isDynamicToolUIPart(part)) {
+    return part.toolName;
+  }
+  return part.type.split('-').slice(1).join('-');
 }
 
 export type InferUIMessageMetadata<T extends UIMessage> =
--- a/packages/ai/src/ui/ui-messages.ts
+++ b/packages/ai/src/ui/ui-messages.ts
@@ -293,7 +293,7 @@ export function isToolOrDynamicToolUIPart<TOOLS extends UITools>(
 }
 
 export function getToolName<TOOLS extends UITools>(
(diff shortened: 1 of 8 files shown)


gr2m:
The change does break types for `useChat()` when used in conjunction with `getToolName()` and `addToolResult()`.

It's what causing the build errors. You can see it in this example:
[examples/next-openai/app/use-chat-human-in-the-loop/page.tsx](https://github.com/vercel/ai/blob/ca67e5a06eb7c4a300cbe3fd7fc995ac2746d3f4/examples/next-openai/app/use-chat-human-in-the-loop/page.tsx)

Before this change, `getToolName` would return a union of strings

image

With this change, it returns a generic string type.

image

And that's incompatible with `addToolResult()` which still requires the union

image

We could make it work by updating `addToolResult()` but would loose type intelligence

I'm looking into an alternative way

packages/ai/src/ui/ui-messages.ts
```diff
@@ -280,12 +280,33 @@ export function isToolUIPart<TOOLS extends UITools>(
...
 }
 
+export function getToolOrDynamicToolName(
+  part: ToolUIPart<UITools> | DynamicToolUIPart,
+): string {
+  if (isDynamicToolUIPart(part)) {
+    return part.toolName;
+  }
+  return part.type.split('-').slice(1).join('-');
+}
```
gr2m, line 308: this PR adds this method and exports it from the `ai` package, but does not consume it. 

We should either remove it or add tests.

gr2m (commented):
I'm still not 100% sure if this is the right approach, deferring to @lgrammel 

Should we add or update an example?

packages/ai/src/ui/ui-messages.ts
```diff
@@ -280,12 +280,33 @@ export function isToolUIPart<TOOLS extends UITools>(
...
   return part.type.split('-').slice(1).join('-') as keyof TOOLS;
 }
 
+export function getToolOrDynamicToolName(
+  part: ToolUIPart<UITools> | DynamicToolUIPart,
+): string {
+  if (isDynamicToolUIPart(part)) {
+    return part.toolName;
+  }
+  return part.type.split('-').slice(1).join('-');
```
lgrammel, line 307: call getToolName instead of duplicating logic

lgrammel (approved):
Not a huge fan of the name `isToolOrDynamicToolUIPart` but I dont have a good alternative; this may be something we want to rename in AI SDK 6

lgrammel (changes requested):
for the get method we can do better

lgrammel (approved):
or not

lgrammel:
Thanks!

dvoytenko:
Thanks!

dvoytenko:
Does `5.0.23` include this change?

lgrammel:
> Does `5.0.23` include this change?

yes https://github.com/vercel/ai/blob/main/packages/ai/CHANGELOG.md#5023

dvoytenko:
> > Does `5.0.23` include this change?
> 
> yes https://github.com/vercel/ai/blob/main/packages/ai/CHANGELOG.md#5023

@lgrammel do we also need to update `@ai-sdk/react`?

lgrammel:
@dvoytenko yes update to latest version

gr2m:
> do we also need to update `@ai-sdk/react`?

Good question. It seems like the [`workspace:*`](https://github.com/vercel/ai/blob/15271d6685d681870a025b21115b5e6ed207132a/packages/react/package.json#L33) version in dependencies is resolved to a pinned version instead of a range:
https://www.npmjs.com/package/@ai-sdk/react?activeTab=code

I'll discuss this with the team. For now I'll get a new version of `@ai-sdk/react` released to get you unblocked.

gr2m:
nevermind my last comment, latest `@ai-sdk/react` is already pointing to `ai@5.0.23`

PR #8256, opened 2025-08-24
Add support for `retry` callback in `streamText` `onError`. Adds support for retry on overloaded messages and fallback model logic

### LINKED ISSUE #2636: feat: ability to define fallback model / provider
### Feature Description

It would be helpful to be able to define "fallback" models / providers when building AI applications.

ie. if primary provider (x) errors out, revert to provider (y).

### Use Case

_No response_

### Additional context

_No response_

remorses:
Not ready for review! Still experimental

I added a `retry` callback that users can call inside `onError` to retry errors that happen in the middle of a stream, for example Anthropic overloaded errors. When the user calls retry streamText will send the partial messages up until that point 

- Added `retry` function to the `onError` callback parameters
- When `retry()` is called, the error is not forwarded to the stream
- Added `retriedError` parameter to `prepareStep` function to enable configuration changes based on the error that triggered the retry (e.g., switching to a backup model on overloaded errors)
- Properly clears active content tracking on retry to avoid ID conflicts

### Example usage:

```typescript
const result = streamText({
  model: primaryModel,
  prompt: 'Hello',
  prepareStep: ({ retriedError }) => {
    // Switch to backup model on overload error
    if (retriedError?.type === 'overloaded') {
      return { model: backupModel };
    }
  },
  onError: async (error, { retry }) => {
    if (error.type === 'overloaded' || error.type === 'network') {
      // Retry the request transparently
      await retry();
    } else {
      // Log non-retryable errors
      console.error('Fatal error:', error);
    }
  },
});
```

Fix https://github.com/vercel/ai/issues/2636


tjazsilovsek:
Great PR, looking forward to it being merged!

remorses:
Some things that were not clear while implementing this, and should be clarified before resuming this PR

- In this PR I tried handling error that happen in the middle of the stream. This is supported only for a few model providers like Anthropic, OpenAI and others instead don't support resuming an assistant message. Instead other providers can only resume a stream at step boundaries (after tool call end, or start of stream). This needs to be handled, probably only resuming at last step instead of middle of text like it does now
- In this PR I am blocking the sending of error, finish and finish-step parts because these should not appear in the response in case of a retry. This made logic very complex and difficult to follow. An alternative way to do this should probably be used

gr2m:
Closing this inactive experimental draft; the unresolved retry semantics need a fresh design. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

PR #8257, opened 2025-08-24
feat(ai): support URL to base64 conversion in tool `toModelOutput`

xinyao27:
## Background

Tools that generate media content (screenshots, images, etc.) currently need to return base64-encoded data directly, which makes tool implementations more complex and less efficient. Many tools naturally produce URLs (from CDN, cloud storage, or generation services) but have to download and encode the content themselves to work with models that don't support URLs.

## Summary

Added automatic URL to base64 conversion for media content in tool result outputs. When a tool returns a URL in its media content and the model doesn't support URLs, the SDK will automatically:
- Download the content from the URL
- Convert it to base64 format
- Include it in the prompt sent to the model

This allows tools to return URLs instead of base64 data, making tool implementations cleaner while ensuring compatibility with all models.

## Manual Verification

- Created a test tool that returns image URLs in its output
- Verified that URLs are automatically downloaded when the model doesn't support them
- Confirmed that base64 conversion happens correctly for downloaded content
- Tested with both URL-supporting and non-URL-supporting model configurations
- Verified existing functionality remains unchanged

## Tasks


lgrammel:
can you resolve merge conflicts?

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -347,3 +396,38 @@ function convertPartToLanguageModelPart(
...
+                data: convertToBase64(downloadedFile.data),
+                mediaType:
+                  downloadedFile.mediaType ??
+                  convertedMediaType ??
+                  result.mediaType,
+              };
+            }
+          }
+        }
+        return result;
```
gr2m, line 428: to avoid nesting, can you please implement an early return pattern?

gr2m (commented):
I think there are some ways to simplify the implementation, it's a bit hard to read. Could you have another look?

PR #8274, opened 2025-08-25
feat(provider/gateway): add support for image models

shaper:
## Background
Adding image model support to the Gateway

## Summary
- Adds image model support to the Gateway package
- Updates `generate-image` to preserve and pass through to user Gateway provider metadata (routing/cost info)
- Adds special case for Gateway provider in `providerMetadata` to not include the `images` array, which will always be empty as the images come from the upstream model provider
- Allows just passing in the model string (without `gateway.imageModel()`) to the `model` field of `generateImage`
- I don't believe any docs changes are needed

## Manual Verification

- All tests pass
- I ran `pnpm build` and `pnpm pack` in both the `ai` and `gateway` packages, moved the tgz files to the ai-gateway repo, updated `package.json` in `ai-gateway` to reference the local files with the updated code, and ran `generateImage` calls to verify images work and providerMetadata comes back as expected

## Tasks

- [X] A _patch_changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset`in the project root)
- [X] Formatting issues have been fixed (run `pnpm prettier-fix`in the project root)


shaper:
* **#8274** <a href="https://app.graphite.com/github/pr/vercel/ai/8274?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/ai/8274?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `release-v5.0`


<h2></h2>

packages/ai/src/model/resolve-model.ts
shaper, line 1: add tests to `resolve-model.test.ts`

packages/ai/src/types/image-model.ts
```diff
@@ -7,7 +7,7 @@ import {
 /**
 Image model that is used by the AI SDK Core functions.
   */
-export type ImageModel = ImageModelV2;
```
shaper, line 10: I am not sure whether this is considered a breaking change -- let's vet with @gr2m 

It also may warrant docs updates. Do we describe or document plain strings for embedding and language models anywhere today?

packages/gateway/src/gateway-image-model.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  private getUrl() {
+    return `${this.config.baseURL}/image-model`;
+  }
+
+  private getModelConfigHeaders() {
+    return {
+      'ai-image-model-specification-version': '2',
```
shaper, line 102: @gr2m should we forward-port this to v6, or land it in v6 and backport to v5?

packages/gateway/src/gateway-provider.ts
shaper, line 1: add unit tests for edits

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -171,7 +171,8 @@ Only applicable for HTTP-based providers.
   const images: Array<DefaultGeneratedFile> = [];
   const warnings: Array<ImageGenerationWarning> = [];
   const responses: Array<ImageModelResponseMetadata> = [];
-  const providerMetadata: ImageModelV2ProviderMetadata = {};
+  const providerMetadata: Record<string, JSONObject & { images?: JSONArray }> =
```
shaper: Why did we lose the spec type here? I would think we would still want to keep the type, even if later we need to zod-parse it or inspect it to do something more complex.

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -189,13 +190,37 @@ Only applicable for HTTP-based providers.
...
 
     if (result.providerMetadata) {
-      for (const [providerName, metadata] of Object.entries<{
-        images: unknown;
-      }>(result.providerMetadata)) {
-        providerMetadata[providerName] ??= { images: [] };
-        providerMetadata[providerName].images.push(
-          ...result.providerMetadata[providerName].images,
-        );
+      const entries = Object.entries(result.providerMetadata) as Array<
```
shaper: Can you summarize what we are doing here and why? I can see we are merging incoming images with existing images. Why did we need to do this -- what is new and changed due to gateway image model work and behavior?

Perhaps Gateway is actually doing something wrong, confusing, or surprising and there is another approach.

packages/ai/src/types/image-model.ts
gr2m: I see the change was reverted. But in general we cannot do any changes in `packages/provider` in our stable releases

packages/gateway/src/gateway-image-model.ts
gr2m: starting with v6 would benefit from our backport automations. The other way around would be more manual

packages/ai/src/generate-image/generate-image.ts
R-Taneja: I've since simplified the code quite a bit, see new code in the files tab. I now leave the original providerMetadata logic unchanged and treat gateway as a special case to accomplish what I want.

The issue I ran into was that gateway `providerMetadata` was coming back as:

```
gateway: {
  images: [],
}
```

The images array would always be empty (as expected), and none of the routing/cost info would come through.

I dug through the code and realized this is because the `generate-image` providerMetadata-building code would drop all non `images` fields in this line: `providerMetadata[providerName] ??= { images: [] };`. This also had the side effect of producing an images array for every provider, which doesn't apply to gateway.

To solve this, my code preserves the `metadata` (routing/cost info) and deletes the `images` field if present and empty for the gateway provider only

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -192,10 +187,23 @@ Only applicable for HTTP-based providers.
       for (const [providerName, metadata] of Object.entries<{
         images: unknown;
       }>(result.providerMetadata)) {
-        providerMetadata[providerName] ??= { images: [] };
```
shaper, line 195: I think some edit like this makes sense based on the below summary and some contemplation. @gr2m may have more background or other thoughts on approach and best practices.

Sharing as it helped me -- a sonnet-4.5 summary:

The original code had a simple merging strategy for provider metadata across multiple image generation calls...

- For __every provider__ in the metadata, it would:

  1. Initialize with `{ images: [] }` if not already present
  2. Append all images from that provider's metadata to the accumulated `images` array

- This worked fine for providers like OpenAI/Vertex that return per-image metadata in the `images` array

The new code adds special handling for the gateway provider:

__For `gateway` provider:__

- __Spreads all metadata fields__ (not just `images`) - this preserves fields like `routing`, `cost`, `marketCost`, `generationId`
- __Removes empty `images` arrays__ - the gateway doesn't provide per-image metadata, so empty arrays are noise
- __Last-write-wins__ for non-image fields across multiple calls (see test "should merge non-image provider metadata fields")

__For other providers (OpenAI, Vertex, etc.):__

- __Keeps the original behavior__ - accumulates `images` arrays across calls
- This preserves per-image metadata like `revisedPrompt` that these providers return

- __Gateway-specific metadata structure__: The AI Gateway returns operational metadata (`routing`, `cost`, `generationId`) that shouldn't be treated like per-image metadata. The old code would try to push these into an `images` array, which doesn't make sense.

- __Cleaner API surface__: Removing empty `images: []` from gateway metadata prevents confusion - if there's no per-image data, don't include the field.

- __Multi-call scenarios__: When `n > maxImagesPerCall`, multiple API calls are made. The new logic correctly:

  - Accumulates per-image metadata for providers that support it
  - Merges operational metadata for the gateway (with last-write-wins for conflicts)

- __Backward compatibility__: Non-gateway providers continue to work exactly as before.

shaper: I expect the issue is not specific to gateway per se (and thus rohan's original edits) -- if a provider did happen to return other metadata it would probably get thrown away since there is not merging logic. But this may be theoretical as it is not clear any other provider actually does this today.

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -192,10 +187,23 @@ Only applicable for HTTP-based providers.
...
+          if (Array.isArray(imagesValue) && imagesValue.length === 0) {
+            delete (providerMetadata[providerName] as { images?: unknown })
+              .images;
+          }
+        } else {
+          providerMetadata[providerName] ??= { images: [] };
+          providerMetadata[providerName].images.push(
+            ...result.providerMetadata[providerName].images,
+          );
+        }
```
gr2m, line 213: Sorry I've been starring at this code and I don't get it yet.

I understand that gateway returns request-specific metadata. But a request can yield more than one images, including with gateway, correct? And each of these images might yield their own meta data?

What confuses me is that for gateway images, the code still iterates through all images, and gets metadata for each image, but then ends up merging the metadata from all images into the single `providerMetadata['gateway']` object?

gr2m (commented):
Can you please add at least one example for generating images with gateway? If there are different edge cases or variants we want cover then create one for each

gr2m (approved):
Okay to merge as is to get it to work. I'll do a follow up to simplify the code once I can test it against production gateway

PR #8277, opened 2025-08-25
test: explicitly import vitest functions

### LINKED ISSUE #8271: Maintenance: explicitly import `describe`, `it`, `expect`, etc from `vitest`
Remove the implicit injection of `vivest`'s global APIs:
https://vitest.dev/config/#globals

The steps are

1. Search for `globals: true` in all `vitest.*config.js` files and remove it
2. Remove `vitest/globals` from `"types"` in TypeScript configuration files (`tsconfig.json` files, `tools/tsconfig/base.json`)
3. Add explicit imports (like `import { beforeEach, describe, expect, it, vi } from 'vitest';`) to all test files, import only what's needed

Comment below if you'd like to work on this issue.

emmperez:
* Explicitly import `describe`, `it`, `expect` from Vitest to fix TS build errors.
* Externalize `vitest` and `chai` for builds

### Verification:

* `pnpm build` succeeds without errors.
* `pnpm type-check:full` succeeds without errors.
* Ran pnpm test: all suites passed except @ai-sdk/assemblyai#test, which is a known Vitest bug and unrelated to this change.
* DTS files generate correctly.

### Related Issue:

Fixes #8271


emmperez:
Hi @gr2m 
I went through and updated all test files to use explicit imports from vitest. I believe I’ve covered them all, but please let me know if I missed anything. 

ps: this is my first open source PR

gr2m:
> * `pnpm build` succeeds without errors

can you run `pnpm run type-check:full`? It throws errors in CI

Can you please also run `pnpm prettier-fix` to resolve the code formatting problems that CI is complaining about?

> * Update TypeScript/tsup build target from `es2018` → `esnext` for modern JS features (BigInt, optional chaining, etc.).
> * Fix `ArrayBuffer` typing in test server code for DTS compatibility.

can you remove these changes since they are unrelated? Feel free to open separate PRs for these. I'm not sure if we can change `es2018` to `esnext`, it might be a breaking change (we consider type changes breaking changes)

emmperez:
> can you run `pnpm run type-check:full`? It throws errors in CI

> Can you please also run `pnpm prettier-fix` to resolve the code formatting problems that CI is complaining about?

I went through the errors, corrected the missed files and ran prettier 
> can you remove these changes since they are unrelated? Feel free to open separate PRs for these. I'm not sure if we can change `es2018` to `esnext`, it might be a breaking change (we consider type changes breaking changes)

Makes sense, these are now removed

gr2m:
> * `pnpm build` succeeds without errors

I see build errors, all related to `chai` with this error

> Big integer literals are not available in the configured target environment

Bumping the target resolves this but I'm not sure if that is something we can do. Got any idea? Maybe we can bump target but only for the test builds, not for production? 

I assume we see these build errors now due to the explicit import of `vitest`, which now becomes part of the test builds, which seems to be incompatible with `"target": "ES2018"`.

emmperez:
> I see build errors, all related to `chai` with this error
>
> Bumping the target resolves this but I'm not sure if that is something we can do. Got any idea? Maybe we can bump target but only for the test builds, not for production?
> 
> I assume we see these build errors now due to the explicit import of `vitest`, which now becomes part of the test builds, which seems to be incompatible with `"target": "ES2018"`.

Yeah, separating the build targets should work. Is excluding them from the builds an option?

emmperez:
Separated targets so:

Updated tsup.config.ts so test/ builds use target: "es2020" while everything else stays on es2018.

Marked chai as external for test builds to avoid bundling it.

I am able to build successfully now, can you review?

gr2m:
Looks like excluding `chai` from the build is causing problems now

```
@ai-sdk/lmnt:test: Error: Failed to load url chai (resolved id: chai) in /Users/gr2m/code/vercel/ai/packages/provider-utils/dist/test/index.mjs. Does the file exist?
@ai-sdk/lmnt:test:  ❯ loadAndTransform ../../node_modules/.pnpm/vite@5.4.11_@types+node@22.7.4_less@4.2.2_sass@1.85.0_terser@5.39.0/node_modules/vite/dist/node/chunks/dep-CB_7IfJ-.js:5
```

Let me see if we can just bump the compile target in TS Config

gr2m:
FYI 

- https://github.com/vercel/ai/pull/8307

Unfortunately it has its own build errors, and I'm not sure what's up:
https://github.com/vercel/ai/actions/runs/17249017736/job/48946188831?pr=8307#step:7:2452

I'll have to focus on other work for a bit, but will get back to it. If you could have a look that'd be great

emmperez:
Did some digging and figured because vitest was bundled it it couldn't resolve the chai dependency. I’ve updated the tsup config so that Vitest is externalized, just like chai. 

Looks like build sizes also reduced but I wanted to get your thoughts here. I'm not seeing any build or test errors.

gr2m (approved):
Thanks a lot Emmanuel, this is a great PR, would merge again 💯 Thank you for working through all the issues, digging into the root causes and resolving them elegantly. Much appreciated 💐

emmperez:
Thanks for all your help Greg!

PR #8284, opened 2025-08-26
fix(ai): remove use of `expect()` from production code

gr2m:
## Background

`expect()` was used in file that is imported from production code. This slipped through because the `expect()` method is currently imported implicitly, which will change once #8277 lands.

## Summary

Move function that uses `expect()` into its own file

## Manual Verification

n/a

## Tasks

## Future Work

- #8277

### DIFF
--- a/packages/ai/src/error/no-object-generated-error.ts
+++ b/packages/ai/src/error/no-object-generated-error.ts
@@ -68,22 +68,3 @@ export class NoObjectGeneratedError extends AISDKError {
     return AISDKError.hasMarker(error, marker);
   }
 }
-
-export function verifyNoObjectGeneratedError(
-  error: unknown,
-  expected: {
-    message: string;
-    response: LanguageModelResponseMetadata & {
-      body?: string;
-    };
-    usage: LanguageModelUsage;
-    finishReason: FinishReason;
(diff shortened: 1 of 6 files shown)


packages/ai/src/error/no-object-generated-error.ts
```diff
@@ -68,22 +68,3 @@ export class NoObjectGeneratedError extends AISDKError {
...
-    finishReason: FinishReason;
-  },
-) {
-  expect(NoObjectGeneratedError.isInstance(error)).toBeTruthy();
-  const noObjectGeneratedError = error as NoObjectGeneratedError;
-  expect(noObjectGeneratedError.message).toEqual(expected.message);
-  expect(noObjectGeneratedError.response).toEqual(expected.response);
-  expect(noObjectGeneratedError.usage).toEqual(expected.usage);
-  expect(noObjectGeneratedError.finishReason).toEqual(expected.finishReason);
-}
```
gr2m, line 89: This code is only used in tests, but the file itself is imported into production builds. I found out thanks to the work by @emmperez in #8277 that makes all `vitest` imports explicit.

lgrammel (approved):

Issue #8079, opened 2025-08-15
Mixing `file_search` results in type validation error; `file_citation` annotations lack `start_index`, `end_index`, `quote`


Irvel:
### Description

follow up to https://github.com/vercel/ai/issues/8030#issuecomment-3191527471

When using `generateText` with both `web_search_preview` and `file_search` tools in AI SDK v5, the call fails with an `AI_TypeValidationError` sometimes.

This happens when the OpenAI Responses API does not return `start_index`, `end_index`, and `quote` fields.

The OpenAI Responses API returns valid `file_citation` objects with `file_id`, `filename`, and `index`, but the AI SDK validator expects them to also have the `start_index`, `end_index`, and `quote` properties (like `url_citation` annotations). This leads to `AI_TypeValidationError` failures during parsing.

**Reproduction**

```ts
import { generateText, openai } from 'ai';

const result = await generateText({
  model: openai('gpt-5'),
  system: 'system prompt here',
  prompt: 'prompt here',
  tools: {
    web_search_preview: openai.tools.webSearchPreview({ searchContextSize: 'high' }),
    file_search: openai.tools.fileSearch({
      vectorStoreIds: ['myvectorstoreid'],
      maxNumResults: 3,
    }),
  },
});
```

### Observed Behavior
```
AI_APICallError: Invalid JSON response
AI_TypeValidationError: Invalid input: expected number, received undefined
Path: ["output", 11, "content", 0, "annotations", 0, "start_index"]
```
This repeats for start_index, end_index, and quote on each file_citation annotation.

### Example of Returned `file_citation` Objects
```json
{
  "type": "file_citation",
  "file_id": "file-XXXX",
  "filename": "ev-database-BEV-m1.json",
  "index": 145
}
```

### Full error paste

```json
error: {
      "name": "AI_APICallError",
      "cause": {
        "name": "AI_TypeValidationError",
        "cause": [
          {
            "expected": "number",
            "code": "invalid_type",
            "path": [
              "output",
              11,
              "content",
              0,
              "annotations",
              0,
              "start_index"
            ],
            "message": "Invalid input: expected number, received undefined"
          },
          {
            "expected": "number",
            "code": "invalid_type",
            "path": [
              "output",
              11,
              "content",
              0,
              "annotations",
              0,
              "end_index"
            ],
            "message": "Invalid input: expected number, received undefined"
          },
          {
            "expected": "string",
            "code": "invalid_type",
            "path": [
              "output",
              11,
              "content",
[...]

PR #8286, opened 2025-08-26
fix(provider/openai): make file_citation annotation fields optional for responses api compatibility

dancer:
## background

openai responses api sometimes returns file_citation annotations without start_index, end_index, and quote fields. the ai sdk schema was treating these as required, causing type validation errors when using file_search with web_search_preview tools

## summary

- made start_index, end_index, and quote optional in file_citation schema
- added filename and index field support
- added tests for file_citation data

## verification

tested with both streaming and non-streaming scenarios. all existing tests pass

## tasks

- [x] updated schema validation for file_citation annotations
- [x] added tests for annotation data

closes - #8079


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -361,10 +361,12 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
...
                           type: z.literal('file_citation'),
-                          start_index: z.number(),
-                          end_index: z.number(),
                           file_id: z.string(),
-                          quote: z.string(),
+                          filename: z.string().optional(),
+                          index: z.number().optional(),
+                          start_index: z.number().optional(),
+                          end_index: z.number().optional(),
+                          quote: z.string().optional(),
```
lgrammel: prefer nullish in response schemas

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1184,7 +1190,11 @@ const responseAnnotationAddedSchema = z.object({
     z.object({
       type: z.literal('file_citation'),
       file_id: z.string(),
-      quote: z.string(),
+      filename: z.string().optional(),
+      index: z.number().optional(),
+      start_index: z.number().optional(),
+      end_index: z.number().optional(),
+      quote: z.string().optional(),
```
lgrammel: prefer nullish in response schemas

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -496,8 +498,8 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
                   sourceType: 'document',
                   id: this.config.generateId?.() ?? generateId(),
                   mediaType: 'text/plain',
-                  title: annotation.quote,
-                  filename: annotation.file_id,
+                  title: annotation.quote ?? annotation.filename ?? 'Document',
```
lgrammel, line 501: why was the title the quote to begin with? does that make sense?

dancer: it kinda make sense openai used to use it in the previous api, ig it is just a good fallback to have just in case? they moved away from it though

lgrammel (approved):

PR #8312, opened 2025-08-26
feat(provider/openai): Adds service tier to OpenAI responses provider metadata

jacoblee93:
## Background

There's currently no great way to get the `service_tier` out of OpenAI's response when streaming without using `fullStream` and having the user pass `includeRawChunks`. For `generate` it's possible to get this out of the response body, but it's somewhat cumbersome. This field is important for token cost tracking.

OpenAI can downgrade you depending on rate limits, so getting this info out of the request params is not enough.

## Summary

This adds `serviceTier` to the returned `providerMetadata` for OpenAI streaming (and also generate for completeness).

## Manual Verification

Ran an existing example (`ai/examples/ai-core/src/stream-text/openai-responses-reasoning-summary.ts`) and saw:

```ts
Provider metadata: {
  openai: {
    serviceTier: 'default',
    responseId: 'resp_68ae3c510d9c8195a74eed6a01736ee30fd80228337cb053'
  }
}
```

Changing to `priority` also passed this through.

## Tasks

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.test.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.test.ts
@@ -3351,7 +3351,7 @@ describe('OpenAIResponsesLanguageModel', () => {
       `);
     });
 
-    it('should should handle logprops', async () => {
+    it('should should handle logprobs and service tier', async () => {
       server.urls['https://api.openai.com/v1/responses'].response = {
         type: 'stream-chunks',
         chunks: [
@@ -3441,6 +3441,7 @@ describe('OpenAIResponsesLanguageModel', () => {
                   ],
                 ],
                 "responseId": "resp_689cec4cf608819583c56813ccb0f5040f92af1765dd5aad",
+                "serviceTier": "default",
               },
             },
             "type": "finish",
--- a/packages/openai/src/responses/openai-responses-language-model.test.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.test.ts
@@ -3351,7 +3351,105 @@ describe('OpenAIResponsesLanguageModel', () => {
       `);
     });
 
-    it('should should handle logprobs and service tier', async () => {
+    it('Should handle service tier', async () => {
+      server.urls['https://api.openai.com/v1/responses'].response = {
+        type: 'stream-chunks',
+        chunks: [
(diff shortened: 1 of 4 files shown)


packages/openai/src/responses/openai-responses-language-model.test.ts
```diff
@@ -3351,7 +3351,7 @@ describe('OpenAIResponsesLanguageModel', () => {
       `);
     });
 
-    it('should should handle logprops', async () => {
+    it('should should handle logprobs and service tier', async () => {
```
lgrammel: can you separate this out into a different test?

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -673,6 +678,10 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
       }
     > = {};
 
+    const finishChunkProviderMetadata: SharedV2ProviderMetadata = {
+      openai: {},
+    };
```
lgrammel: can the service tier be a variable instead of keeping the full metadata?

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -999,21 +1011,17 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
...
-              openai: {
-                responseId,
-              },
-            };
+            finishChunkProviderMetadata.openai.responseId = responseId;
 
             if (logprobs.length > 0) {
-              providerMetadata.openai.logprobs = logprobs;
+              finishChunkProviderMetadata.openai.logprobs = logprobs;
             }
```
lgrammel: would prefer to keep this and just add the service tier based on the new variable

packages/openai/src/responses/openai-responses-language-model.ts
jacoblee93: Sure, can update

jacoblee93:
Updated, yeah this is nicer. PTAL.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1055,6 +1069,7 @@ const responseFinishedChunkSchema = z.object({
   response: z.object({
     incomplete_details: z.object({ reason: z.string() }).nullish(),
     usage: usageSchema,
+    service_tier: z.string().nullish(),
```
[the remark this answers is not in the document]
jacoblee93: Oh, interesting! Don't think this would affect the output since the PR uses the final returned tier only but good for completeness nonetheless

lgrammel (approved):

PR #8360, opened 2025-08-29
fix(ai): remove vitest dependency from test exports

### LINKED ISSUE #8356: `MockLanguageModelV2` Runtime Regression (ai@5.0.27)
### Summary

Importing `MockLanguageModelV2` from `ai/test` at **application runtime** (outside a Vitest test runner) started crashing in `ai@5.0.27` with:

```
Error: Vitest failed to access its internal state.
```

This is a regression from earlier versions (e.g. `ai@5.0.26` and before) where the same usage worked fine. In `5.0.27`, the bundled `ai/test` entry now hard-imports `vitest`, which assumes a Vitest worker context.

Our use case: We intentionally use the mock model as a lightweight, _deterministic development engine_ to exercise full chat workflows (UI ↔ server ↔ streaming) without calling real providers. This is not a unit test scenario, it's a dev-mode runtime.

### Environment

| Item            | Value                                |
| --------------- | ------------------------------------ |
| AI SDK Core     | `ai@5.0.27`                          |
| Node            | `v22.12.0` (also reproduced on 20.x) |
| Package Manager | pnpm                                 |
| OS              | macOS (Apple Silicon)                |

### Before (Working Behavior)

Using `ai@5.0.26`:

```js
import { simulateReadableStream } from "ai";
import { MockLanguageModelV2 } from "ai/test";

const model = new MockLanguageModelV2({
  doStream: async () => ({
      chunks: [
        { type: "text-start", id: "text-1" },
        { type: "text-delta", id: "text-1", delta: "Hello dev!" },
        { type: "text-end", id: "text-1" },
        {
          type: "finish",
          finishReason: "stop",
          usage: { inputTokens: 2, outputTokens: 3, totalTokens: 5 }
        }
      ]
    })
  })
});

console.log("Mock instantiated ok", model.specificationVersion);
```

Running with plain `node dev-mock.js` succeeds (no Vitest context required).

The older published `dist/test/index.js` had **no** `require("vitest")`.

### After (Broken in 5.0.27)

The same script now throws immediately:

```
node:internal/modules/esm/resolve:857
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'vitest' imported from /Volumes/SSD-2T/tmp/v27-vitest-issue/node_modules/ai/dist/test/index.mjs
    at packageResolve (node:internal/modules/esm/resolve:857:9)
    at moduleResolve (node:internal/modules/esm/resolve:926:18)
    at defaultResolve (node:internal/modules/esm/resolve:1056:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:654:12)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:603:25)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:586:38)
    at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:242:38)
    at ModuleJob._link (node:internal/modules/esm/module_job:135:49) {
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v22.12.0
```

Relevant extract from `node_modules/ai/dist/test/index.js` in `5.0.27`:

```js
// src/test/mock-embedding-model-v2.ts
var import_vitest = require("vitest"); // <— new hard import (unused for runtime mocking needs)
```

This unconditional import triggers Vitest’s internal state check when _any_ `ai/test` symbol is imported at runtime.

### Minimal Reproduction Steps

```bash
mkdir ai-mock-runtime-regression && cd ai-mock-runtime-regression
pnpm init --init-type module
pnpm add ai@5.0.27
cat > dev-mock.js <<'EOF'
import { simulateReadableStream } from "ai";
import { MockLanguageModelV2 } from 'ai/test';

new MockLanguageModelV2({
  doStream: async () => ({
      chunks: [
        { type: 'text-start', id: 'text-1' },
        { type: 'text-delta', id: 'text-1', delta: 'Hello' },
        { type: 'text-end', id: 'text-1' },
        { type: 'finish', finishReason: 'stop', usage: { inputTokens: 1, outputTokens: 5, totalTokens: 6 } },
      ],
    }),
  }),
});
console.log('OK');
EOF
node dev-mock.js
```

Observed: runtime crash with Vitest internal state error.

Now downgrade:

```bash
pnpm add ai@5.0.26
node dev-mock.js
```

Observed: runs successfully (no Vitest error).

### Impact

- Breaks legitimate dev-mode workflows that rely on mock models for integrated end-to-end local testing / demos.

### Workaround Implemented

We created a lightweight runtime clone (file: `runtimeMockLanguageModel.ts`) mirroring only the needed surface:

```ts
export class RuntimeMockLanguageModelV2 {
  specificationVersion = "v2";
  constructor({ doGenerate, doStream } = {}) {
    /* records calls & delegates */
  }
  // doGenerate / doStream call tracking
}
```

and replaced:

```diff
- import { MockLanguageModelV2 } from 'ai/test';
+ import { RuntimeMockLanguageModelV2 } from './runtimeMockLanguageModel';
```

This restores development ergonomics but duplicates upstream logic.

### Suggested Fixes

Any of the following would resolve the issue:

1. **Remove the unconditional `require('vitest')`** from the published `ai/test` bundle (if it’s unused in production-like mocking scenarios).
2. **Lazy / dynamic import `vitest`** only when the specific features that need it are invoked (e.g. assertions utilities), or guard with a try/catch.
3. **Split exports**: Provide a subpath (e.g. `ai/mock`) that excludes testing framework bindings and only contains pure mock model classes & stream helpers.
4. **Document** that `ai/test` must not be imported outside Vitest and provide an officially blessed alternative for runtime demo mocks.

### Rationale

Mocks are broadly useful beyond unit tests (demos, smoke environments, offline dev). Making them strictly dependent on an active Vitest worker reduces flexibility and introduces unexpected crashes for non-test contexts.

### Request

Please consider decoupling the mock model helpers from Vitest so they can be safely imported at runtime.

### Thanks

Thanks for the excellent SDK! This small adjustment would smooth out dev UX for teams relying on deterministic local mock flows.

---

If more diagnostics are needed (full stack trace, etc.), let me know and I can append them.

### AI SDK Version

- ai: 5.0.27

### Code of Conduct

lgrammel:
## Background

In `ai@5.0.27` the `ai/test` export requires vitest, leading to a breaking regression (see #8356 ).

## Summary

Remove `mockEmbed` function from `ai/test` (it should have never been exported to begin with).

## Related Issues

Fixes #8356

### DIFF
--- a/.changeset/chatty-rocks-rest.md
+++ b/.changeset/chatty-rocks-rest.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): remove vitest dependency from test exports
--- a/packages/ai/src/embed/embed-many.test.ts
+++ b/packages/ai/src/embed/embed-many.test.ts
@@ -1,6 +1,6 @@
+import { EmbeddingModelV2 } from '@ai-sdk/provider';
 import assert from 'node:assert';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-import { EmbeddingModelV2 } from '../../../provider/src/embedding-model';
 import { MockEmbeddingModelV2 } from '../test/mock-embedding-model-v2';
 import { MockTracer } from '../test/mock-tracer';
 import { Embedding, EmbeddingModelUsage } from '../types';
@@ -472,7 +472,7 @@ describe('result.providerMetadata', () => {
   });
 });
 
-export function mockEmbed<VALUE>(
+function mockEmbed<VALUE>(
   expectedValues: Array<VALUE>,
   embeddings: Array<Embedding>,
   usage?: EmbeddingModelUsage,
--- a/packages/ai/src/embed/embed-many.test.ts
+++ b/packages/ai/src/embed/embed-many.test.ts
@@ -1,10 +1,9 @@
 import assert from 'node:assert';
-import { describe, it, expect, vi, beforeEach } from 'vitest';
-import {
-  MockEmbeddingModelV2,
-  mockEmbed,
-} from '../test/mock-embedding-model-v2';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { EmbeddingModelV2 } from '../../../provider/src/embedding-model';
+import { MockEmbeddingModelV2 } from '../test/mock-embedding-model-v2';
 import { MockTracer } from '../test/mock-tracer';
+import { Embedding, EmbeddingModelUsage } from '../types';
 import { createResolvablePromise } from '../util/create-resolvable-promise';
 import { embedMany } from './embed-many';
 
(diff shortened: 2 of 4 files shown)


blyszcz:
@lgrammel I'm still experiencing this issue on version 5.0.108. It appears to have been fixed in 5.0.28, so this seems to be a regression.

PR #8375, opened 2025-08-30
feat(ai): export PartialObject from util to fix TS declaration consumers error

ItsWendell:
## Background

This PR exports PartialObject from the AI SDK public util index to avoid TypeScript declaration naming errors when consumer projects enable declaration output.

When using the AI SDK in a project with declaration: true, TypeScript may emit:

```log
Return type of public method from exported class has or is using name PartialObject from external module "/node_modules/ai/dist/index" but cannot be named.
```

## Summary

Expose PartialObject alongside DeepPartial in packages/ai/src/util/index.ts.

## Manual Verification

Repo builds, lints, and tests succeed locally. No runtime behavior changes; this is a type-only surface export.

## Tasks

### DIFF
--- a/.changeset/rare-cycles-scream.md
+++ b/.changeset/rare-cycles-scream.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+export PartialObject from util to fix TS declaration consumers error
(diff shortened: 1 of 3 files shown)


lgrammel:
Can you show a code snippet that reproduces the issue?

My main concern is that with a similar logic, all internal types should be exposed, vastly increasing the API surface of the project (since the types also need to be documented).

I would like to explore solutions that do not require us to, one-by-one, expose all internal types.

gr2m:
Closing after no reproduction or follow-up on the API-surface concern. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

PR #8253, opened 2025-08-24
feat(codemod): add not implemented comment and warnings

imranbarbhuiya:
## Background

Context: https://github.com/vercel/ai/issues/6552#issuecomment-3217736137
## Summary
This pr adds comments and a warning for things that need change but no codemod is available/possible

## Manual Verification

@gr2m ran it on `vercel/ai-chatbot`

## Tasks

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)
- [x] I have reviewed this pull request (self-review)

## Future Work

Ive moved `not-implemented` into a folder so if there's a complex codemod that cant be done like this then that can be placed in that folder

## Related Issues

towards #6552

PR #8398, opened 2025-09-02
docs(codemod): update readme

gr2m:
Follow up to #8253


imranbarbhuiya:
Sorry I missed this, the `not-implemented` check should be moved from https://github.com/vercel/ai/blob/d1164df79d55a60a7e4ea98ebd475d1b97db05e7/packages/codemod/scripts/generate-readme.ts#L25 to https://github.com/vercel/ai/blob/d1164df79d55a60a7e4ea98ebd475d1b97db05e7/packages/codemod/scripts/generate-readme.ts#L19

just noticed it too while creating #8399

PR #8412, opened 2025-09-03
feat(providers/anthropic): map 'refusal' stop reason to 'content-filter' finishReason

### LINKED ISSUE #6529: Support new "refusal" stop reason for Claude 4 models
### Description

According to https://docs.anthropic.com/en/docs/about-claude/models/migrating-to-claude-4#new-refusal-stop-reason,
_Claude 4 models introduce a new refusal stop reason for content that the model declines to generate for safety reasons, due to the increased intelligence of Claude 4 models_

It looks like [`packages/anthropic/src/map-anthropic-stop-reason.ts`](https://github.com/vercel/ai/blob/9b09f85143986636570e597c31903daf160608cd/packages/anthropic/src/map-anthropic-stop-reason.ts#L3) should be updated to support this new stop reason.

I think that `refusal` should be converted to `content-filter`.

### AI SDK Version

_No response_

danielamitay:
## Background
- Fixes https://github.com/vercel/ai/issues/6529

## Summary

This PR adds support for the `refusal` stop reason by converting it to `content-filter`. This maps well according to Anthropic's own documentation, which says it is returned `"During streaming when content violates policies"`
Screenshot 2025-09-02 at 11 32 07 PM

https://docs.anthropic.com/en/docs/test-and-evaluate/strengthen-guardrails/handle-streaming-refusals#current-refusal-types

## Manual Verification
- I have a test case for this but it is not appropriate to post publicly. Running it produces `refusal` stop reason from Anthropic which now correctly returns `content-filter` via AI-SDK

### DIFF
--- a/.changeset/thirty-hotels-mix.md
+++ b/.changeset/thirty-hotels-mix.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+feat(providers/anthropic): map 'refusal' stop reason to 'content-filter' finishReason
--- a/packages/anthropic/src/map-anthropic-stop-reason.ts
+++ b/packages/anthropic/src/map-anthropic-stop-reason.ts
@@ -11,6 +11,8 @@ export function mapAnthropicStopReason({
     case 'end_turn':
     case 'stop_sequence':
       return 'stop';
+    case 'refusal':
+      return 'content-filter';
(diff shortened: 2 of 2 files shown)


dmtran-g:
Thank you @danielamitay for opening this PR!

lgrammel (approved):

PR #8429, opened 2025-09-03
fix(provider/gladia): Invalid `model` parameter being passed & `experimental_transcribe` fails with valid Buffer

### LINKED ISSUE #8406: gladia: `experimental_transcribe fails with valid Buffer`
### Description

Prepared this issue to onboard @aayush-kapoor - latest addition to the AI SDK team 🥳 

Prior art for reference: See https://github.com/vercel/ai/pull/8159

Code that needs to be fixed

https://github.com/vercel/ai/blob/9e684f4bf223abfb7d4a957aefc50018dc8b6b24/packages/gladia/src/gladia-transcription-model.ts#L501-L504

### AI SDK Version

_No response_

### Code of Conduct

aayush-kapoor:
## Background

- Invalid `model` argument being passed to the gladia api during transcription
- `experimental_transcribe` fails with valid buffer input in v5

## Summary

- Commented out the `model` param being passed in the request
- Set `filename` when uploading audio blob to Gladia's transcribe API

## Manual Verification

Ran test file `examples/ai-core/src/transcribe/gladia.ts`

## Tasks

- [ ] n/a ~~Documentation has been added / updated (for bug fixes / features)~~

## Future Work

Apply the same fix to other places where we upload a blob without defining a file name

```
packages/openai/src/transcription/openai-transcription-model.ts:
  131      formData.append('model', this.modelId);
  132:     formData.append('file', new File([blob], 'audio', { type: mediaType }));
  133  

packages/revai/src/revai-transcription-model.ts:
  254  
  255:     formData.append('media', new File([blob], 'audio', { type: mediaType }));
  256      const transcriptionModelOptions: RevaiTranscriptionAPITypes = {
```

## Related Issues

fixes #8406


packages/gladia/src/gladia-transcription-model.ts
```diff
@@ -497,10 +498,13 @@ export class GladiaTranscriptionModel implements TranscriptionModelV2 {
         ? new Blob([options.audio])
         : new Blob([convertBase64ToUint8Array(options.audio)]);
 
-    formData.append('model', this.modelId);
+    // formData.append('model', this.modelId);
```
gr2m: I confirmed that leaving that line in results in a 400 error:

> property model should not exist

Let's just remove the line entirely.

.changeset/eighty-balloons-sing.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/gladia': patch
+---
+
+fix(provider/gladia): Invalid `model` parameter being passed
+fix(provider/gladia): `experimental_transcribe` fails with valid Buffer
```
gr2m, line 6: I was not able to reproduce that error by removing the 3rd argument to `formData.append`. How did you get the error?

gr2m (changes requested):
Two things

.changeset/eighty-balloons-sing.md
aayush-kapoor: If you're talking about the `experimental_transcribe` error that was discussed in previous issues, I actually wasn't able to reproduce that error for this provider either. But I assumed that this fix was still needed the same way we did it for other providers. let me know if I missed anything

packages/gladia/src/gladia-transcription-model.ts
aayush-kapoor: will do!

.changeset/eighty-balloons-sing.md
gr2m: Ah okay, got it. I think Gladia's API does not throw an error if the filename is not set explicitly, but it's still good practice to do it. No further changes needed.

gr2m (approved):
Let's gooo 🚀

PR #8449, opened 2025-09-04
docs: clarify `stepCountIs()`

R-Taneja:
## Background

A friend was reading the AI SDK docs and let me know that there was a small point of confusion because the naming of `stopWhen` and `stepCountIs(5)` doesn't make it clear that it's a maximum of 5 steps, not that it will take exactly 5 steps every time. His message to me:

> the info tag actually does explain it, but you kind have to assume that a text stop reason will cause the execution loop to terminate (which seems obvious now but it didn’t click for some reason). i think the comment in the codeblock + the naming of the utilities is the primary source of confusion, but probably just changing comment in the codeblock to “// stop after a maximum of 5 steps if tools were called” works.
>
> i would probably make the interface { stopCondition: stepCountMax(i: number) } but thats prolly not worth it

## Summary

Makes this more clear in the docs

## Tasks

### DIFF
--- a/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
+++ b/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
@@ -94,7 +94,7 @@ const { text, steps } = await generateText({
       }),
     }),
   },
-  stopWhen: stepCountIs(5), // stop after 5 steps if tools were called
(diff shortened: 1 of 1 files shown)


allenzhou101 (approved):

gr2m (approved):
great clarification, thanks!

PR #8452, opened 2025-09-04
chore (provider/vertex): update GoogleVertexModelId

danielamitay:
Simply updates the `GoogleVertexModelId` with the more recent stable models: https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions

I'll do my best to make sure Vertex docs and such get some love 🥹


packages/google-vertex/src/google-vertex-options.ts
```diff
@@ -1,8 +1,13 @@
-// https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#supported-models
+// https://cloud.google.com/vertex-ai/generative-ai/docs/learn/model-versions
```
gr2m, line 1: nice thanks for that 👍🏼

gr2m (approved):
Thanks a lot! 

I wish we could monitor this page and just get notified when a new model gets listed 🤔

PR #8493, opened 2025-09-07
feat(vue): add useObject support for Vue

### LINKED ISSUE #2903: Vue.js: useObject support
### Feature Description

Implement useObject support in Vue.js similar to the functionality for React.

### Use Case

_No response_

### Additional context

_No response_

taugr:
## Background

Adds useObject support to the Vue package. 

## Summary

* Added experimental_useObject to the Vue package, following the same pattern as the React version.
* Added unit tests similar to the corresponding React tests.
* Added example page to the nuxt-openai example project
* Updated the docs.

## Manual Verification

Added an example page `use-object` to the `nuxt-openai` example project to run and verify streaming objects works as expected.

## Tasks

## Related Issues

Fixes #2903


packages/vue/package.json
```diff
@@ -47,14 +47,19 @@
     "msw": "2.6.4",
     "tsup": "^7.2.0",
     "typescript": "5.8.3",
-    "vitest": "2.1.4"
+    "vitest": "2.1.4",
+    "zod": "3.25.76"
```
gr2m: we don't want to add a dependency on `zod`, a `peerDependency` that covers both `zod@3` and `zod@4` should suffice

packages/vue/src/use-object.ts
```diff
@@ -0,0 +1,234 @@
...
+  type DeepPartial,
+  isDeepEqualData,
+  parsePartialJson,
+  type Schema,
+} from 'ai';
+import * as z3 from 'zod/v3';
+import * as z4 from 'zod/v4';
+
+// use function to allow for mocking in tests
+const getOriginalFetch = () => fetch;
```
gr2m, line 20: @lgrammel I see that in the other implementations as well, but I don't think we actually utilize it in any tests? We would need to export it anyway in order to use it?

gr2m (commented):
Hi @tom-auger - thanks for sending your first pull request! It looks great to me! 

I'm not super familiar with the UI parts of AI SDK yet, I'll reach out for a second set of eyes to make sure I'm not missing anything

packages/vue/src/use-object.ts
taugr: It's not needed here specifically as useObject has an parameter to pass a custom fetch anyway. But it would allow tests to mock the global fetch - it doesn't need to be exported to do that.

gr2m: true, cheers.

gr2m (approved):
Great pull request Tom, thank you! Congratulations on your first contribution to the AI SDK 💐

PR #8505, opened 2025-09-08
fix (provider/openai): fix code interpreter tool in doGenerate

lgrammel:
## Background

The [OpenAI code interpreter tool](https://platform.openai.com/docs/guides/tools-code-interpreter) does not work with `generateText` (OpenAI responses API).

## Summary

* support code interpreter tool construction without arguments
* fix code interpreter tool in generate text

## Manual Verification

* [x] run code interpreter generate text example

## Tasks

- [x] review code interpreter tool
- [x] support no args code interpreter constructor
- [x] update / add examples for code interpreter tool
- [x] fix code interpreter tool for generate text case
- [x] add changeset

## Future Work
- add tool call and result parts for code interpreter tool


examples/ai-core/src/lib/run.ts
```diff
@@ -0,0 +1,11 @@
...
+import { APICallError } from 'ai';
+
+export function run(fn: () => Promise<void>) {
+  fn().catch(error => {
+    if (APICallError.isInstance(error)) {
+      console.error(error.responseBody);
+    }
+    console.error(error);
+  });
+}
```
nicoalbanese, line 11: is this necessary?

lgrammel: yes it will help with debugging in the future

nicoalbanese (approved):

PR #8472, opened 2025-09-05
feat(ai): custom user agent header with version information

aayush-kapoor:
## Background

Setting a custom user agent is considered best practice and would help debug problems in case any occur. As suggested by @gr2m, `ai` is not setting a custom user agent right now.

## Summary

- add `__PACKAGE_VERSION__` to each core package and inject it at build time
- create a shared user-agent utility in provider-utils
- create helper to attach user-agent and wire it across all core calls of the package
- updated test files, with updated snapshots
- expected behavior after this change is:
    - `user-agent` in the `Headers` with the syntax: `ai/<version> @ai-sdk/provider/<version> @ai-sdk/provider-utils/<version> node/<ver> os/<platform> arch/<arch>`
    - Plus any suffix if you set `X-AI-SDK-User-Agent-Extra`

## Manual Verification

Complete verification still needed (incomplete) but created custom scripts that simulates a server and a call being made to that server to verify that the header tag with `user-agent` is present.

## Tasks

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)

## Future work

- support browsers, deno, bun
- add `userAgent` argument to methods exported by `ai` and make it a recommendation

## Related Issues

#8419

PR #8530, opened 2025-09-09
feat(ai): set `ai`, `@ai-sdk/provider-utils`, and runtime in `user-agent` header

### LINKED ISSUE #8419: Set `ai`, `@ai-sdk/provider-utils`, and runtime in `user-agent` header
The AI SDK is currently not setting any user-agent header. 

For the initial implementation we want to set a user-agent header following the format

> ai/5.0.40 ai-sdk/provider-utils/3.0.8 node/24.8.0

gr2m:
## Background

Setting a custom user agent is considered best practice and would help debug problems in case any occur. As suggested by @gr2m, `ai` is not setting a custom user agent right now.

This is an alternative approach to #8472. 

## Summary

- export new `createUserAgentFetch()` and `withUserAgentSuffix()` helper methods
- add `__PACKAGE_VERSION__` to each core package and inject it at build time
- pass current `ai/x.y.z` version through the headers argument in all methods exported by `ai`
- updated test files, with updated snapshots
- expected behavior after this change is:
    - `user-agent` in the `Headers` with the syntax: `ai/<version> @ai-sdk/provider/<version> @ai-sdk/provider-utils/<version> <navigator.userAgent>`

## Manual Verification

Running `examples/ai-core/fetch-user-agent/user-agent.ts` simulates a mock call being made to an echo server with our new `createUserAgentFetch()` method. Output shows a user-agent header with the correct versions being set.

## Tasks

## Future work

- https://github.com/vercel/ai/issues/8595

## Related Issues

closes #8419 
closes #8472


.changeset/seven-zoos-suffer.md
```diff
@@ -0,0 +1,8 @@
+---
+'@ai-sdk/provider-utils': patch
+'@ai-sdk/provider': patch
```
lgrammel: not added

gr2m:
@aayush-kapoor and I talked today, we will continue with this approach. Aayush will implement the feature in all the methods exported by `ai`

packages/provider-utils/src/test/test-server.ts
```diff
@@ -88,6 +88,7 @@ class TestServerCall {
     // convert headers to object for easier comparison
     const headersObject: Record<string, string> = {};
     requestHeaders.forEach((value, key) => {
+      if (key.toLowerCase() === 'user-agent') return;
```
lgrammel, line 91: why

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+  const normalizedHeaders = new Headers(headers);
+  const currentUserAgentHeader = normalizedHeaders.get('user-agent') || '';
+
+  normalizedHeaders.set(
+    'user-agent',
+    [currentUserAgentHeader, ...userAgentSuffixParts].filter(Boolean).join(' '),
+  );
+
+  return Object.fromEntries(normalizedHeaders);
+}
```
aayush-kapoor: for this function `withUserAgentSuffix`, should we add an additional check for when we are appending a new user-agent to see if it is the same one as the existing user-agent?

for eg: i ran into this while testing. for the given test:
```
        model: new MockImageModelV2({
          maxImagesPerCall: 2,
          doGenerate: async options => {
            switch (callCount++) {
              case 0:
                expect(options).toStrictEqual({
                  prompt,
                  n: 2,
                  seed: 12345,
                  size: '1024x1024',
                  aspectRatio: '16:9',
                  providerOptions: { 'mock-provider': { style: 'vivid' } },
                  headers: { 'custom-request-header': 'request-header-value', 'user-agent': '' }, 
                  abortSignal: undefined,
                });
                return createMockResponse({
                  images: base64Images.slice(0, 2),
                  warnings: [{ type: 'other', message: '1' }],
                });
```

when we set 
`user-agent` to `ai/0.0.0-test` in `headers: { 'custom-request-header': 'request-header-value', 'user-agent': '' }, `

on running the test, output is `"user-agent": "ai/0.0.0-test ai/0.0.0-test"`

it's running as expected (since it appends) but I was wondering if an extra check would be helpful here or not. lmk

packages/provider-utils/src/test/test-server.ts
aayush-kapoor: this was for standard convention for the headers to be in lower-case as suggested [here](https://github.com/vercel/ai/pull/8472#discussion_r2325841114).

examples/ai-core/src/fetch-user-agent/user-agent.ts
```diff
@@ -0,0 +1,55 @@
...
+    const result = {
+      method: parsed.method,
+      path: parsed.path,
+      userAgent: echoedUserAgent,
+      headers: parsed.headers,
+      parsedQueryParams: parsed.parsedQueryParams ?? {},
+      parsedBody: parsed.parsedBody ?? null,
+    };
+
+    console.log(JSON.stringify(result, null, 2));
```
gr2m: example is great. I think we can simplify it though, I would only test the user agent since it is the only thing we are testing here.

I would leave out zod and schema parsing, just log out the user agent header

```
console.log(`User-Agent Header: ${result.headers['user-agent']}`)

File: packages/ai/src/generate-object/generate-object.test.ts [hunk 5]
```diff
@@ -1031,6 +1048,7 @@ describe('generateObject', () => {
         headers: {
           header1: 'value1',
           header2: 'value2',
+          'user-agent': '',
```
gr2m: Why is `user-agent` empty here?

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: nvm, this situation won't arise and has been taken care of

packages/ai/src/generate-object/generate-object.test.ts
aayush-kapoor: fixed!

examples/ai-core/src/fetch-user-agent/user-agent.ts
aayush-kapoor: this has been made simpler!

.changeset/seven-zoos-suffer.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ai-sdk/provider-utils': patch
+'@ai-sdk/gateway': patch
```
gr2m: this PR no longer has any changes in `@ai-sdk/gateway`


```suggestion
```

examples/ai-core/src/fetch-user-agent/user-agent.ts
```diff
@@ -0,0 +1,35 @@
+import { createUserAgentFetch } from '@ai-sdk/provider-utils';
+import 'dotenv/config';
+
+async function main() {
+  try {
+    const userAgentFetch = createUserAgentFetch();
+
+    const response = await userAgentFetch(
+      'https://echo.free.beeceptor.com/sample-request?test=1',
```
gr2m: In future, we might want to just build our own local echo server, just in case this one disappears or is down. But this is good for now as is, it's a problem when it's a problem

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (globalThis.navigator?.userAgent) {
+    return navigator.userAgent;
+  }
+
+  // Nodes.js < 21.1
+  if (globalThis.process?.versions?.node) {
+    return `Node.js/${process.version.substring(1)}`;
+  }
+
+  return '<unknown runtime>';
```
gr2m: Sorry I only caught this now. I think we should avoid using `<>` characters

See https://www.rfc-editor.org/rfc/rfc9110.html#field.user-agent for more information

How about we prefix all 3 possible returns with `provider/`

```suggestion
  // Browsers / Deno / Bun / Node.js >= 21.1
  if (globalThis.navigator?.userAgent) {
    return `runtime/${navigator.userAgent}`;
  }

  // Nodes.js < 21.1
  if (globalThis.process?.versions?.node) {
    return `runtime/Node.js/${process.version.substring(1)}`;
  }

  return 'runtime/unknown';
```

gr2m (commented):
Sorry found a few minor things, after that it should really be good to ship

examples/ai-core/src/fetch-user-agent/user-agent.ts
aayush-kapoor: i'll add it into future work!

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: yup sounds good! committing this change

examples/ai-core/src/fetch-user-agent/user-agent.ts
aayush-kapoor: added

gr2m (commented):
Looks good to me now! I'm the author so you'll have to approve the PR yourself @aayush-kapoor

packages/provider-utils/src/index.ts
```diff
@@ -3,6 +3,10 @@ export { convertAsyncIteratorToReadableStream } from './convert-async-iterator-t
 export * from './delay';
 export * from './extract-response-headers';
 export * from './fetch-function';
+export {
+  createUserAgentFetch,
+  withUserAgentSuffix,
+} from './fetch-with-user-agent';
```
lgrammel: do we need to expose them?

packages/provider-utils/src/test/test-server.ts
lgrammel: by returning you omit the user agent header tho?

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
+import type { FetchFunction } from './fetch-function';
+import { VERSION as PROVIDER_UTILS_VERSION } from './version';
+
+export function getRuntimeEnvironmentUserAgent(): string {
```
lgrammel: there is more, see https://github.com/vercel/modelfusion/blob/main/packages/modelfusion/src/util/detectRuntime.ts

(also consider extracting into e.g. a structured detectRuntime function)

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+  );
+
+  return Object.fromEntries(normalizedHeaders);
+}
+
+/**
+ * Creates a fetch function that adds a `user-agent` header to each request.
+ */
+export function createUserAgentFetch(baseFetch?: FetchFunction): FetchFunction {
+  const effectiveFetch: FetchFunction = baseFetch ?? (globalThis.fetch as any);
```
lgrammel: pls avoid `any`

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+      `ai-sdk/provider-utils/${PROVIDER_UTILS_VERSION}`,
+      getRuntimeEnvironmentUserAgent(),
+    );
+
+    return effectiveFetch(url, {
+      ...init,
+      headers,
+    });
+  };
+}
```
lgrammel: not in love with this approach bc it feels overly complex, why don't we just have a function to calculate the user agent header and use it in get-from-api/post-to-api?

packages/provider-utils/src/index.ts
aayush-kapoor: not really, will just star (*) export all

packages/provider-utils/src/test/test-server.ts
aayush-kapoor: i realise now that this particular line isn't for lower case but for converting the `Header` into a dictionary object.
this check here is to see if the user-agent header is already set, then the Header object is already a dict. That would mean that we don't need to do any changes to the Header and hence we return back.

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: changed to `FetchFunction`

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: I believe 
```
if (globalThis.navigator?.userAgent) {
    return `runtime/${navigator.userAgent}`;
  }
```
takes care of most, but will still add for `vercel-edge` if detected

packages/provider-utils/src/fetch-with-user-agent.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Creates a fetch function that adds a `user-agent` header to each request.
+ */
+export function createUserAgentFetch(baseFetch?: FetchFunction): FetchFunction {
+  const effectiveFetch: FetchFunction = baseFetch ?? (globalThis.fetch as any);
+
+  return async (input: string | URL | Request, init?: RequestInit) => {
+    // normalize arguments
+    if (input instanceof Request) {
+      init = input;
+      input = input.url;
```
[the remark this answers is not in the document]
aayush-kapoor: fixed with new commit

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: function is now removed altogether

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: fixed!

packages/provider-utils/src/fetch-with-user-agent.ts
gr2m, line 1: we can rename this file now, or maybe split up the code into two files.

packages/provider-utils/src/get-from-api.ts
```diff
@@ -27,7 +31,10 @@ export const getFromApi = async <T>({
   try {
     const response = await fetch(url, {
       method: 'GET',
-      headers: removeUndefinedEntries(headers),
+      headers: withUserAgentSuffix(
+        removeUndefinedEntries(headers),
+        getRuntimeEnvironmentUserAgent(),
```
gr2m: we forgot adding the provider-utils version now. We have to add it back.

`withUserAgentSuffix()` accepts multiple suffix arguments, so we can do something like

```ts
      headers: withUserAgentSuffix(
        removeUndefinedEntries(headers),
        `ai-sdk/provider-utils/${VERSION}`,
        getRuntimeEnvironmentUserAgent()
```

packages/provider-utils/src/post-to-api.ts
```diff
@@ -95,7 +99,10 @@ export const postToApi = async <T>({
   try {
     const response = await fetch(url, {
       method: 'POST',
-      headers: removeUndefinedEntries(headers),
+      headers: withUserAgentSuffix(
+        removeUndefinedEntries(headers),
+        getRuntimeEnvironmentUserAgent(),
+      ),
```
gr2m: same as above, we forgot adding the provider-utils version now. We have to add it back.

aayush-kapoor: added!

packages/provider-utils/src/get-from-api.ts
aayush-kapoor: added!

packages/provider-utils/src/fetch-with-user-agent.ts
aayush-kapoor: split into 2 files for now

packages/provider-utils/src/with-user-agent-suffix.ts
```diff
@@ -0,0 +1,23 @@
...
+  const normalizedHeaders = new Headers(headers);
+  const currentUserAgentHeader = normalizedHeaders.get('user-agent') || '';
+
+  normalizedHeaders.set(
+    'user-agent',
+    [currentUserAgentHeader, ...userAgentSuffixParts].filter(Boolean).join(' '),
+  );
+
+  return Object.fromEntries(normalizedHeaders);
+}
```
lgrammel: can you add unit tests that show what this does?

packages/provider-utils/src/index.ts
```diff
@@ -3,6 +3,8 @@ export { convertAsyncIteratorToReadableStream } from './convert-async-iterator-t
 export * from './delay';
 export * from './extract-response-headers';
 export * from './fetch-function';
+export * from './get-runtime-environment-user-agent';
+export * from './with-user-agent-suffix';
```
lgrammel: we aim to have explicit exports (instead of `*`), bc `*` has led to uncontrolled API surface extensions in the past (also below for version).

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -255,10 +258,15 @@ A function that attempts to repair a tool call that failed to parse.
 
   const callSettings = prepareCallSettings(settings);
 
+  const headersWithUserAgent = withUserAgentSuffix(
+    removeUndefinedEntries(headers ?? {}),
```
lgrammel: can removeUndefinedEntries be used inside withUserAgentSuffix to simplify the usage?

gr2m: Good idea 👍🏼

aayush-kapoor: we could do that but does that not just add more abstraction? this way tests for both are defined separately. but if you think it's fine for this scenario, we'll change it

aayush-kapoor: alright changing it!

.changeset/seven-zoos-suffer.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/provider-utils': patch
+'ai': patch
+---
+
+feat(ai): custom user agent header with version information
```
gr2m:
```suggestion
feat(ai): set `ai`, `@ai-sdk/provider-utils`, and runtime in `user-agent` header
```

examples/ai-core/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.0.0",
   "private": true,
   "dependencies": {
+    "@ai-sdk/provider-utils": "workspace:*",
```
gr2m: This is no longer necessary, right?

examples/ai-core/tsconfig.json
```diff
@@ -100,6 +100,9 @@
     {
       "path": "../../packages/provider"
     },
+    {
+      "path": "../../packages/provider-utils"
+    },
```
gr2m: same as above

packages/provider-utils/src/test/test-server.ts
gr2m: This is for snapshot testing. Otherwise snapshots will fail based on runtime (e.g. different node version) or change each time `ai` or `@ai-sdk/provider-utils` change

packages/provider-utils/src/with-user-agent-suffix.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      'cache-control': null,
+    };
+
+    const result = withUserAgentSuffix(headers as any, 'ai-sdk/0.0.0-test');
+
+    expect(result['user-agent']).toBe('TestApp/0.0.0-test ai-sdk/0.0.0-test');
+    expect(result['content-type']).toBe('application/json');
+    expect(result['accept']).toBe('application/json');
+    expect(result['authorization']).toBeUndefined();
+    expect(result['cache-control']).toBeUndefined();
```
gr2m, line 56: I'd use inline snap shots, but these assertions are find, I don't see us updating these often

gr2m (commented):
few more nits

examples/ai-core/package.json
aayush-kapoor: yeah, don't think so

packages/provider-utils/src/with-user-agent-suffix.test.ts
aayush-kapoor: not updating these for now then 👍

gr2m (commented):
ship it 🚀

aayush-kapoor (approved):
LGTM

PR #8536, opened 2025-09-09
fix (provider/anthropic): revert cd458a8c1667df86e6987a1f2e06159823453864

lgrammel:
## Background

https://github.com/vercel/ai/commit/cd458a8c1667df86e6987a1f2e06159823453864 did not fix the underlying issue and caused https://github.com/vercel/ai/issues/8516

## Summary

Revert #cd458a8c1667df86e6987a1f2e06159823453864

## Related Issues

#8516 #8474

### DIFF
--- a/.changeset/early-bugs-raise.md
+++ b/.changeset/early-bugs-raise.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix (provider/anthropic): revert cd458a8c1667df86e6987a1f2e06159823453864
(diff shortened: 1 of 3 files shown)


nicoalbanese (approved):

gr2m:
My bad, sorry for that. FYI @tayyab3245

tayyab3245:
​Hi @gr2m, no worries at all, and thank you for the apology and for taking ownership. I really appreciate it. These things happen in complex projects.
​That said, I did spend a lot of time in a deep dive on this and wanted to share my final thoughts on the root cause. It seems like these recurring bugs are all symptoms of a deeper architectural mismatch between the SDK's universal format and Anthropic's uniqlue API, which has led to a cycle of "band-aid" fixes.
 
For what it's worth, I believe the only permanent solution would be to evolve towards a provider-specific "builder" pattern that constructs a compliant payload from the start. This is similar to the more direct and stable approach that is already used successfully in the Google provider's conversion logic.
​Adopting this pattern for Anthropic would likely eliminate this entire class of bugs and pay down the accumulated technical debt.

PR #8537, opened 2025-09-09
fix(spec): allow `undefined` values in `JSONObject` type

nvie:
## Background

This type change more accurately matches reality as it's not the case that every thinkable key is actually defined on objects of this type. Accessing keys that don't exist on this object will in fact return `undefined` at runtime, and the type should reflect this.

See this [TypeScript Playground](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcBSBlA8gOSwIwCtgBjeAXjgG84BtAa2EQC44BnGKASwDsBzALqsO3fnAC+AbjhwA9LLgYuAWzAAbLgDMuwACZwSAVyhRgPeLuDaeXGFwg8AUPLihIsBMlSZcBYmThKGgYmYU5eQTDRPjgAHzhDHktrPQlpOQUABSgISDZU7RBHRxIHDjgQVh88IlIKailHEAA6LgB9XQcYNtAuDmaYCABVMBQoAGEAQ3yACgBKdJcMAAsIQzV9fFQkFFcTaDhAPg3AFF3HIA) with a simplified example that only uses strings.

## Summary

Added `undefined` as a possible value for all keys in `JSONObject`'s.

## Manual Verification

I've run `pnpm build` in the root of this monorepo and got no failures.

## Future work

- simplify code where possible: https://github.com/vercel/ai/pull/8537#discussion_r2445551544
- use shared `SharedV3ProviderOptions`: https://github.com/vercel/ai/pull/8537#discussion_r2445558842

### DIFF
--- a/packages/ai/src/types/json-value.ts
+++ b/packages/ai/src/types/json-value.ts
@@ -7,7 +7,7 @@ export const jsonValueSchema: z.ZodType<JSONValue> = z.lazy(() =>
     z.string(),
     z.number(),
     z.boolean(),
-    z.record(z.string(), jsonValueSchema),
+    z.record(z.string(), jsonValueSchema.optional()),
     z.array(jsonValueSchema),
   ]),
 );
--- a/packages/provider/src/json-value/json-value.ts
+++ b/packages/provider/src/json-value/json-value.ts
@@ -11,7 +11,7 @@ export type JSONValue =
   | JSONArray;
 
 export type JSONObject = {
-  [key: string]: JSONValue;
+  [key: string]: JSONValue | undefined;
 };
 
 export type JSONArray = JSONValue[];
--- a/packages/provider/src/speech-model/v3/speech-model-v3-call-options.ts
+++ b/packages/provider/src/speech-model/v3/speech-model-v3-call-options.ts
@@ -1,6 +1,6 @@
-import { JSONValue } from '../../json-value/json-value';
+import { JSONObject } from '../../json-value/json-value';
 
-type SpeechModelV3ProviderOptions = Record<string, Record<string, JSONValue>>;
+type SpeechModelV3ProviderOptions = Record<string, JSONObject>;
 
 export type SpeechModelV3CallOptions = {
   /**
--- a/packages/rsc/src/types.ts
+++ b/packages/rsc/src/types.ts
@@ -1,7 +1,7 @@
 export type JSONValue = string | number | boolean | JSONObject | JSONArray;
 
 interface JSONObject {
-  [x: string]: JSONValue;
+  [x: string]: JSONValue | undefined;
 }
 
 interface JSONArray extends Array<JSONValue> {}
--- a/.changeset/hot-pigs-live.md
+++ b/.changeset/hot-pigs-live.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/provider': patch
+---
+
+fix(spec): allow `undefined` values in `JSONObject` type
--- a/content/docs/07-reference/01-ai-sdk-core/01-generate-text.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/01-generate-text.mdx
@@ -492,7 +492,7 @@ To see `generateText` in action, check out [these examples](#examples).
     },
     {
       name: 'providerOptions',
-      type: 'Record<string,Record<string,JSONValue>> | undefined',
+      type: 'Record<string,JSONObject> | undefined',
       isOptional: true,
       description:
         'Provider-specific options. The outer key is the provider name. The inner values are the metadata. Details depend on the provider.',
@@ -860,7 +860,7 @@ To see `generateText` in action, check out [these examples](#examples).
             },
             {
               name: 'providerMetadata',
-              type: 'Record<string,Record<string,JSONValue>> | undefined',
+              type: 'Record<string,JSONObject> | undefined',
               isOptional: true,
               description:
                 'Optional metadata from the provider. The outer key is the provider name. The inner values are the metadata. Details depend on the provider.',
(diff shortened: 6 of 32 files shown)


packages/provider/src/json-value/json-value.ts
```diff
@@ -11,7 +11,7 @@ export type JSONValue =
   | JSONArray;
 
 export type JSONObject = {
-  [key: string]: JSONValue;
+  [key: string]: JSONValue | undefined;
```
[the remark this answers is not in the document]
nvie: While `undefined` is indeed not a value that can ever be serialized in a JSON string, it does signify a missing key. This is true both ways:

```ts
JSON.stringify({ a: 1, b: undefined })  // Perfectly fine: '{"a":1}'
JSON.parse('{"a":1}')                   // Perfectly fine: { a: 1 }

// And also
JSON.parse('{"a":1}').b   // undefined
```

It would be nice if the types would reflect this very typical usage of JSON data and embrace this natural uncertainty, rather than use an incorrect type that guarantees that all object keys will contain a legal JSON value.

packages/ai/src/types/json-value.ts
```diff
@@ -7,7 +7,7 @@ export const jsonValueSchema: z.ZodType<JSONValue> = z.lazy(() =>
     z.string(),
     z.number(),
     z.boolean(),
-    z.record(z.string(), jsonValueSchema),
+    z.record(z.string(), jsonValueSchema.optional()),
```
nvie, line 10: This subtle change will now relax validation of tool result outputs that are constructed as such:

```ts
const output: LanguageModelV2ToolResultOutput = {
  type: "json",
  value: {
    foo: myResults.foo,
    bar: myResults?.bar,
  }
}
```

Whereas without this relaxation, one has to do this:

```ts
const output: LanguageModelV2ToolResultOutput = {
  type: "json",
  value: {
    foo: myResults.foo,
    ...(myResults?.bar !== undefined ? { bar: myResults.bar } : {})
  }
}
```

…which is a lot less pragmatic.

Allowing explicit-undefined in JSON objects behaves no differently than missing fields, i.e. the results of `JSON.parse(JSON.stringify(validatedJsonData))` will still be deeply-equal to `validatedJsonData`, but makes it a lot more pragmatic to use.

packages/rsc/src/types.ts
```diff
@@ -1,7 +1,7 @@
 export type JSONValue = string | number | boolean | JSONObject | JSONArray;
 
 interface JSONObject {
-  [x: string]: JSONValue;
+  [x: string]: JSONValue | undefined;
```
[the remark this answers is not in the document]
nvie: As I [mentioned here earlier](https://github.com/vercel/ai/pull/8537#discussion_r2332980319), I disagree with this automated response.

nvie:
@lgrammel Curious to hear your take on this patch! 🙏

gr2m:
@nvie we are very reluctant because it's hard to say what might break because of it. We will start preparing the 5.1 release soon with beta releases, and will trial that change as part of that

nvie:
@gr2m That sounds perfect! Let me know when you have the first beta release for it ready and we’ll take it for a spin.

gr2m:
Could you please look into resolving the conflicts? We can try out merging it in now, and see if folks run into any problems with the beta

nvie:
@gr2m I've just rebased and force-pushed my changes onto the latest `main` 👍

gr2m:
Hi @nvie we prepared the v3 specs now. I reverted changes to v2 and updated some more v3 specs. Can you have a final glance my latest changes look good to you? 

- [`fa1bfe8` (#8537)](https://github.com/vercel/ai/pull/8537/commits/fa1bfe8b45dbf94389a32ec73819922712e6428e)
- [`0445fce` (#8537)](https://github.com/vercel/ai/pull/8537/commits/0445fce3aa17aef272f40b6b1b1859b191a21306)

Then we can ship it

packages/ai/src/types/json-value.ts
gr2m: We probably can simplify our code in a lot of places once we ship it. I suggest not to dive into that yet, let's wait for a bit to see if any problems occur with the spec change for users. We expect a growing amount of people to use the v6 beta after https://vercel.com/ship/ai (Oct 24).

But we can create a follow up issue to not forget about it, with links to or a generic list of places that we should simplify once we decide to keep the change

lgrammel (dismissed):
needs changesets

packages/provider/src/speech-model/v3/speech-model-v3-call-options.ts
```diff
@@ -1,6 +1,6 @@
-import { JSONValue } from '../../json-value/json-value';
+import { JSONObject } from '../../json-value/json-value';
 
-type SpeechModelV3ProviderOptions = Record<string, Record<string, JSONValue>>;
+type SpeechModelV3ProviderOptions = Record<string, JSONObject>;
```
lgrammel, line 3: ideally we'd use shared options (also fro transcription) but that could be for separate prs

gr2m: yes I thought the same thing

lgrammel (approved):

nvie:
> we prepared the v3 specs now. I reverted changes to v2 and updated some more v3 specs. Can you have a final glance my latest changes look good to you?

@gr2m Your changes all look sound to me — let's ship it! 👍

nvie:
Hey! I saw the first v6 betas are shipping out to NPM already. Any chance this could get merged in so we can give it an early test? Thanks! 🙏

gr2m:
> Hey! I saw the first v6 betas are shipping out to NPM already. Any chance this could get merged in so we can give it an early test? Thanks! 🙏

yes! We had https://vercel.com/ship/ai last week where we announced the beta, I didn't want to ship it before a weekend. Sorry for the delay!

PR #8541, opened 2025-09-09
feat: tool execution approval

### LINKED ISSUE #8700: feature: allow tool.execute to delegate to client side execution
### Description

Right now, if a tool has an `execute` function, it will always be called server-side during `streamText`.

```ts
  // Only execute tools that are not provider-executed:
            if (tool.execute != null && toolCall.providerExecuted !== true) {
              const toolExecutionId = generateId(); // use our own id to guarantee uniqueness
              outstandingToolResults.add(toolExecutionId);

             // more code here tall calls tool.execute()

          }
```

It would be very useful for complex use cases if there were a way to dynamically decide if a tool call should be executed client side.

Looking at the current code in `run-tool-transformations.ts`, the easiest addition would be some sort of `canExecuteServer` function, that receives the input, and return a result object, or `boolean`.

It could be something like this:

```ts
// Only execute tools that are not provider-executed:
            if (
              tool.execute != null 
              && toolCall.providerExecuted !== true 
              && tool.shouldExecuteServer(toolCall.input) // THIS IS NEW.
            ) {
              const toolExecutionId = generateId(); // use our own id to guarantee uniqueness
              outstandingToolResults.add(toolExecutionId);


           }
```

This would make use cases like human in loop, and complex back and forth between client/server so much easier.

Right now, human in the loop workflows require these complex workarounds with custom data-streaming of results.

### AI SDK Version

_No response_

### Code of Conduct

### LINKED ISSUE #8718: Tool Execution Approval (human-in-the-loop)

lgrammel:
## Background

Tool execution approval (e.g. for human-in-the-loop) requires complex logic for checking and tool execution.

## Summary

- introduce `needsApproval` flag on `Tool`
- introduce `ToolApprovalRequest` part on assistant model message 
- introduce `ToolApprovalResponse` part on tool model message
- introduce `ToolApprovalRequestOutput`
- add `execution-denied` tool output to language model v3 prompt specification
- add tool approval logic to `generateText` and `streamText`
- extract `executeToolCall` function
- add `UIMessage` and UI message chunk support for tool approvals
- add `addToolApprovalResponse` function to `Chat`
- add `ChatAddToolApproveResponseFunction` type
- update conversion and validation function for UI messages, model messages, and language model emssages 

Missing tool approvals are not handled in any special way, because some models may support tool calls without results. If the used model requires tool results for tool calls, the corresponding API will throw errors.

## Example
Screen Recording 2025-10-06 at 15 58 40

## Manual Verification

- [x] run `examples/ai-core/src/generate-text/openai-tool-approval.ts`
   - [x] with positive approval responses
   - [x] with negative approval responses
- [x] run `examples/ai-core/src/stream-text/openai-tool-approval.ts`
   - [x] with positive approval responses
   - [x] with negative approval responses
- [x] run UI test `next-openai` http://localhost:3000/test-tool-approval
   - [x] approved with follow up message
   - [x] denied with follow up message
   - [x] multiple approvals in one message

## Tasks

- [x] add `needsApproval` to tool definition
- [x] command line approval example for model message loop with generateText
- [x] generateText: add test case for tool approval output
- [x] generateText: add test case for tool approval request in response messages
- [x] generateText: add test case for execution of approved tools in initial step
- [x] separate function that gathers tool approvals
- [x] figure out how to solve consecutive tool messages
- [x] include tool execution messages in response messages
- [x] generateText: implement rejected tool approval handling in initial step
- [x] investigate how to prevent openai retries when tool execution is not approved
- [x] streamText: add example
- [x] streamText: approval requests in full stream
- [x] streamText: approval requests in content
- [x] streamText: approval requests in response messages
- [x] generateText: call createToolModelOutput
- [x] UI: add example
- [x] streamText: send approval requests in ui message stream
- [x] UI: parse approval requests from ui message stream
- [x] UI: sendApprovalResponse updates ui
- [x] UI: add lastAssistantMessageIsCompleteWithApprovalResponses helper
- [x] UI: map UI messages with approvals to model messages
- [x] UI: sendApprovalResponse send request
- [x] fix bug in stitchable stream
- [x] streamText: stream initial tool execution (stream tool calling, response messages)
- [x] streamText: preliminary tool results for approved tool executions (incl. ui testing)
- [x] streamText: stream tool denials
- [x] UI: check that step ordering is maintained and information is included (approvals when tool output)
- [x] add tool approval rejection as option to language model specification
- [x] test and update tool approval collection to take into account available tool-result parts
- [x] bug in run-tools-transformation: onInputAvailable needs to be called even when approval is needed
- [x] UI: sendApprovalResponse works while incoming stream is active
- [x] update summary
- [x] jsdoc
- [x] changeset

## Future Work

* discovered OpenAI responses bug: need to ensure tool result order matches tool call order
* add tool execution approvals for dynamic tools
* allow functions as values for the `Tool.needsApproval` property
* needsApproval can return metadata that becomes part of the approval request
* approval can have options
* approval is available during tool execution (incl. options)
* access initial tool execution step result similar to steps
* different id generator overrides for different ids
* UI sync when approvals happen server side (e.g. via email)
* infinite client/server loop when there are errors
* ai sdk stream versioning
* user facing documentation @nicoalbanese 

## Related Issues

Closes #8718
Closes #8700


packages/ai/src/generate-text/generate-text.ts
```diff
@@ -299,21 +302,71 @@ A function that attempts to repair a tool call that failed to parse.
...
+            content: [
+              // add tool results for approved tool calls:
+              ...toolOutputs.map(output => ({
+                type: 'tool-result' as const,
+                toolCallId: output.toolCallId,
+                toolName: output.toolName,
+                input: output.input,
+                output: {
+                  type: 'json' as const,
+                  value: (output as any).output,
```
lgrammel: any

lgrammel: `toModelOutput` should be called

gr2m:
might have interesting insights: https://x.com/pontusab/status/1971901833712226787 by @pontusab

muniter:
This looks great—thanks for working on it. I wanted to share a few thoughts after experimenting with different approaches to tool approval and complex tools with the AI SDK.

## Approval use cases

When thinking about tool approval, two use cases come to mind:

1. Getting authorization from the user (tool: send_email)

- This is the simple case. The UI shows a dialog with the LLM‑suggested inputs. If you want to go deeper, you can allow the user to edit the inputs, and then approve or deny. This can be conditional, depending on tool inputs and program context; for example, if the email is to a domain different from the user's and the user is not an admin, require approval. (The AI SDK has what we need via `experimental_context` (or roll your own `AsyncLocalStorage`), and we have the inputs and callbacks (`onInputAvailable` and `execute`).)

2. Setting up program state so that the tool has the required prerequisites for execution (tool: read_calendar)

- A function receives the input and context and determines whether we already have the OAuth scopes. If we do, no approval step is required. Otherwise, it generates a URL that the user can use to grant the scopes. Ideally, this is passed as metadata so the UI can properly guide the user through the approval process.

Example workflow with general approach of this PR:

```mermaid
sequenceDiagram

participant AI
participant AISdk
participant ComplexTool
participant UI
participant User

AI->>AISdk: read_calendar(input)
AISdk->>ComplexTool: [needsApproval callback]
alt requiresApproval
    ComplexTool->>AISdk: { requiresApproval: true, meta: { oauthUrl: "https://..." } }
    AISdk->>UI: { type: "tool-approval-request", ..., meta: { oauthUrl: "https://..." } }
    UI->>User: Take user through OAuth flow
    User->>UI: Approved: finished OAuth flow
    UI->>AISdk: { type: "tool-approval-result", ..., approve: true, meta: { oauthUrl: "https://..." } }
else
    ComplexTool->>ComplexTool: { requiresApproval: false }
    AISdk->>ComplexTool: execute(input)
    ComplexTool->>AISdk: result
    AISdk->>AI: result
    AISdk->>UI: { type: "tool-output-available", ..., output: result }
    UI->>User: Show result
end
```

Without making `needsApproval` a callback and being able to provide custom metadata, it's hard to build more complex interactions like the one above.


## Alternative idea

Overall, the AI SDK is really powerful, but the part that doesn't expose enough is tool execution. We have hooks like `onInputAvailable`, `onInputDelta`, and `onInputStart`, but we cannot affect the tool (`input`, `execute`, etc) or the program flow (`streamText`) based on these callbacks like we can with `prepareStep`. If we want to handle tool execution ourselves, we need to drive the whole loop manually, which makes using the UI message stream and other features harder.

A different, more extensible future:

```mermaid
sequenceDiagram

participant AI
participant AISdk
participant ComplexTool
participant UI
participant User

AI->>AISdk: read_calendar(input)
AISdk->>ComplexTool: [prepareTool callback] We don't have the OAuth scopes,
ComplexTool->>ComplexTool: { toolExecution: { stopReason: "approval", meta: { usefulToUI: "require_oauth", oauthUrl: "https://..." } } }
AISdk->>UI: Update `part.type` to `tool-approval-request` with metadata returned by the tool
AISdk->>AISdk: Stop `streamText` loop
UI->>User: Complex interaction happens
UI->>AISdk: { type: "tool-approval-result", ..., approve: true, meta: { calendarId: "123" } }
AISdk->>ComplexTool: [prepareTool callback] OK, let's override our input (return it) { execute: async (input, opts) => ..., input: { ... } }
```

The current `needsApproval: boolean` can be implemented by adding a `prepareStep` that returns `toolExecution: { stopReason: "approval" }`.

This approach aims to:

- Allow tools to return data to the UI that does not go to the LLM (`meta`). This is currently not possible unless you hack the `onFinish` callback (piggybacking on `tool-input-available` `providerMetadata`):

```typescript
        onFinish: (stepResult) => {
          if (stepResult.finishReason === "tool-calls") {
            for (const entry of stepResult.content) {
              if (entry.type === "tool-call" && entry.input && !entry.providerExecuted) {
                writer.write({
                  type: "tool-input-available",
                  toolCallId: entry.toolCallId,
                  toolName: entry.toolName,
                  input: entry.input,
                  providerMetadata: {
                    ...entry.providerMetadata,
                    app: {
                      waitingForApproval: true,
                      // Used by the UI to render and validate a form
                      inputSchema: getToolInputJSONSchema(entry.toolName) as Record<string, any>,
                    },
                  },
                });
              }
            }
          }
```

- Allow the UI to provide custom data to stopped tool calls. Same as above but in the other direction, the only way to do this now is handling tool execution manually outside of `streamText`, like the [existing example](https://ai-sdk.dev/cookbook/next/human-in-the-loop).

- Stop the `streamText` loop when the tool definition determines it should be stopped for various reasons (e.g., approval). Currently, the only way to do this is by not providing an `execute` function on the tool, which cannot easily be done conditionally.

Overall there's a need for tools to be "stopped" for various reasons approval being the most common one. But approval can be very complex like requiring a user to grant OAuth scopes, dependant on input, program context, timeout, resource availability, etc. Perhaps thinking about tools being in a "stopped" state for various reasons is a better way to think about it.

nicoalbanese (approved):

bfan1256:
Hey - has this been deployed to the latest beta? It seems that the latest installable beta for openai-sdk: `3.0.0-beta.64` is still outputting the old "input-available" results.

patrikdevlin:
Thanks for you work on this @lgrammel! I've got a small PR to add support for customizing the api request similar to the other functions. Let me know your thoughts #11048

PR #8588, opened 2025-09-11
feat(packages/test-server): Add `test-server` as a package

### LINKED ISSUE #8469: test-server.ts should not be exported from `ai/test`
### Description

If you import from `ai/test`, you get some nice test utils to test the ai package.

However, this seems to re-export from `test-server.ts` which seems to me as if it was an actual test, not a test util? This breaks somehow in our library now because it complains about the `msw` dependency and vitest not being setup etc 🤔 Is this exported here: https://github.com/vercel/ai/blob/37ae9e715502bcddaaecf46c06bc89614b8051ee/packages/provider-utils/src/test/index.ts on purpose?

### AI SDK Version

_No response_

### Code of Conduct

aayush-kapoor:
## Background

When exporting `test-server.ts` from `packages/provider-utils/src/test`, users were facing issues with added dependencies on `msw` and `vitest`

See [8469](https://github.com/vercel/ai/issues/8469)

## Summary

Created `test-server` package so that it can have the added dependency to msw and vitest, so that users don't face issues exporting `ai/test` 

## Manual Verification

Verified by e2e testing by installing said package

## Tasks

## Future Work

n/a

## Related Issues

Fixes #8469


lgrammel:
`/test` feels semantically more correct than `/internal`, and dependent providers should use it for testing.

why does it work with internal but not with test?

lgrammel:
Another fix: do not require vitest imports so it can be used with jest etc. Ofc debatable.

aayush-kapoor:
> `/test` feels semantically more correct than `/internal`, and dependent providers should use it for testing.
> 
> why does it work with internal but not with test?

so we thought of switching it to `/internal` because some people import the tests we defined in `ai/test` when they don've vitest configured, it breaks. Just added this as a support option for the users since some of them reported that they faced this issue.

why it works with `/internal` now is that we removed the entire exports of the classes from `test-server.ts` so that it isn't included by `ai/test`. so people don't face any issues but then we can still use the defined classes internally for our tests

aayush-kapoor:
> Another fix: do not require vitest imports so it can be used with jest etc. Ofc debatable.

yup @gr2m suggested that as well initially (so that there's basically only one added dependency for this particular issue on `msw`) but then we decided to go with this approach instead.

lgrammel:
okay instead of `/internal` would `/test-server` make more sense? it is not just internal, community providers should be free to use it.

aayush-kapoor:
sure yeah, i think that might be better nomenclature. i'll make the changes

.changeset/eleven-bags-smash.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+fix(provider-utils/test-server): exporting now from `./internal`
```
gr2m: Current implementation exports from `/test-server`

```suggestion
fix(provider-utils/test-server): exporting now from `@ai-sdk/provider-utils/test-server`
```

packages/provider-utils/src/test-server/index.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  createTestServer,
+  TestResponseController,
+  type UrlResponse,
+  type UrlHandler,
+  type UrlHandlers,
```
gr2m: @lgrammel I wonder if we need to export the types at all here? Are these useful to users at all? We do not use them anywhere in the code.

Also unlike `createTestServer()`, the `TestResponseController` does not depend on any dependencies, we could continue exporting this one at `ai/test` without a problem, this one is useful by itself

What do you think?

gr2m:
> okay instead of `/internal` would `/test-server` make more sense? it is not just internal, community providers should be free to use it.

I'm not sure, because `/test-server` makes it sound like a public API that we shouldn't break. The problem with `createTestServer()` is that it will only work for people that also use `vitest` and a compatible version of `msw` in their project. And I'm not sure if people use `createTestServer()` at all? In #8469 people ask for it to be removed entirely.

lgrammel:
> > okay instead of `/internal` would `/test-server` make more sense? it is not just internal, community providers should be free to use it.
> 
> I'm not sure, because `/test-server` makes it sound like a public API that we shouldn't break. The problem with `createTestServer()` is that it will only work for people that also use `vitest` and a compatible version of `msw` in their project. And I'm not sure if people use `createTestServer()` at all? In #8469 people ask for it to be removed entirely.

test server is critical for our testing. it is the result of refactoring 3 different testing/mock server approaches and unifying our codebase to be able to easily test provider api integrations (and more, e.g. chat). lets chat offline about it; it would love to allow community providers to benefit from this work as it is very hard to create high quality providers w/o it and painful to set up your own testing

we need to distinguish requests from ai sdk users ( i think #8469 is one of them ) from the needs of community provider developers

packages/provider-utils/src/test-server/index.ts
lgrammel: a) if we do not use the types they could be removed
b) does the testresponsecontroller make sense on its own? i have always seen it as part of the test server

gr2m: you are right, I thought we used `TestResponseController` on its own, but we only use it together with `createTestServer`

gr2m:
> it would love to allow community providers to benefit from this work as it is very hard to create high quality providers w/o it and painful to set up your own testing

One thing @aayush-kapoor and I discussed was to create a dedicated package for the test server, which then can have its own `msw` dependency, and we can externalize the `beforeAll`/`afterAll`,`beforeEach` blocks so it can be used without vitest

gr2m:
@lgrammel and I talked about the idea of creating a new package `@ai-sdk/test-server` which can have a dependency on `msw`. We think it's the best approach, it will also allow us to document it for people who build their own providers. That package will update the return of createTestServer() to add the server: { start, stop, reset } key as we discussed. But then we also want an export that behaves exactly like right now, so we don't need to update hundreds of our tests files. Something like @ai-sdk/test-server/with-vitest which implements the beforeAll / afterAll / beforeEach blocks using the new start / stop / reset functions, and we can use that export internally in our tests

aayush-kapoor:
Sounds good! I'll get started on the implementation

gr2m (changes requested):
blocking merge as a reminder to self: we are in pre-release mode and since this PR will introduce a new package for the test server, it will require a bit of extra work.

packages/test-server/package.json
```diff
@@ -0,0 +1,68 @@
+{
+    "name": "@ai-sdk/test-server",
+    "version": "1.0.0-beta.0",
```
gr2m:
```suggestion
    "version": "0.0.0",
```

I verified now that this is what we want. Then in the `.changeset` file use `major` for the initial release of a package

.changeset/eleven-bags-smash.md
gr2m:
```suggestion
'@ai-sdk/test-server': major
---

feat(test-server): initial version
```

packages/test-server/package.json
aayush-kapoor: understood, changed!

.changeset/eleven-bags-smash.md
aayush-kapoor: changed!

aayush-kapoor: i changed this [here](https://github.com/vercel/ai/pull/8588/files#diff-8074ff10d60c13351377d1f8411f0f05c7e73ea179f94a956c6d00d7f80972f8) but then CI is failing for some reason? [Log here](https://github.com/vercel/ai/actions/runs/17833740843/job/50705326376?pr=8588)

I have initialised the package version to be 0.0.0 [here](https://github.com/vercel/ai/pull/8588/files#diff-509a1d8cccbd15a0a1d6c9ab68999c4b43005564580ce6eb780df6b06b2a696c:~:text=server%22%2C-,%22version%22%3A%20%220.0.0%22%2C,-%22license%22%3A)

aayush-kapoor: this has been resolved my marking the PR with label `major`

packages/provider-utils/package.json
```diff
@@ -36,6 +36,12 @@
       "import": "./dist/test/index.mjs",
       "module": "./dist/test/index.mjs",
       "require": "./dist/test/index.js"
+    },
+    "./test-server": {
+      "types": "./dist/test-server/index.d.ts",
+      "import": "./dist/test-server/index.mjs",
+      "module": "./dist/test-server/index.mjs",
+      "require": "./dist/test-server/index.js"
```
gr2m: This can now be removed since we have a new package, and the `test-server` folder can be removed from `provider-utils`

aayush-kapoor:
I have not yet removed the files `provider-utils/src/test/test-server.ts` and `provider-utils/src/test-server/index.ts` btw

packages/provider-utils/package.json
aayush-kapoor: ahh yes was just typing that. will remove!

gr2m: okay ping me when done and I'll have another look

packages/test-server/src/index.ts
```diff
@@ -0,0 +1,8 @@
+console.log('USING NEW TEST SERVER');
```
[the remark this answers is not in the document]
aayush-kapoor: done

gr2m:
FYI not going to backport this one to 5.0 since it introduces a new package and that will make things complicated ™️

aayush-kapoor:
sure, wtv works best here

gr2m (approved):
great job on this one 👍🏼

Issue #8358, opened 2025-08-29
experimental_generateImage token usage


Morriz:
### Description

Currently experimental_generateImage does not return usage date. Is that coming soon? I hope so...

### AI SDK Version

- ai: 5.0.27
- @ai-sdk/openai: 2.0.22
- @ai-sdk/xai: 2.0.13

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gr2m:
Would you like to have a look at it on why token usage is not exposed yet? Might be a simple fix. A good starting point would be a pull request adding or updating an example (like [examples/ai-core/build/generate-image/openai.js](https://github.com/vercel/ai/blob/main/examples/ai-core/build/generate-image/openai.js)) to showcase what you would like to see working, and we can take it from there?

kingsleytorlowei:
I'm looking into this. Something like this would be the expected usage right? 
```
import { openai } from '@ai-sdk/openai';
import { experimental_generateImage as generateImage } from 'ai';
import 'dotenv/config';

async function main() {
  const prompt = 'Santa Claus driving a Cadillac';
  const result = await generateImage({
    model: openai.image('dall-e-3'),
    prompt,
  });

  console.log(result.usage);

}

main().catch(console.error);
```

jonaslalin:
I think it would make sense to update the v3 spec for packages/provider/src/image-model/v2/image-model-v2.ts to also have something like packages/provider/src/language-model/v2/language-model-v2-usage.ts. We also need usage tracking for image generation.

<img width="406" height="370" alt="Image" src="https://github.com/user-attachments/assets/a09d7ce6-98d3-46b8-b593-d1b75a219342" />

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests a new usage-tracking capability and API expansion for image generation.

### Evidence
- `experimental_generateImage` currently has no `usage` field, and the reporter asks whether usage data will be added.
- A commenter proposes extending the image model specification with usage information analogous to language-model usage tracking.
- The discussion describes an expected new `result.usage` API rather than a failure of documented behavior.

PR #8538, opened 2025-09-09
fix: return raw response body in image model v2

xiaoiver:
## Background

Usage costs will be included in response like gpt-image-1. Therefore, I believe we can return the raw response, just like SpeechModelV2 does.

## Tasks

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ ] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

Fixes https://github.com/vercel/ai/issues/6275

PR #8593, opened 2025-09-11
Feature/add usage data to image generation

kingsleytorlowei:
## Background
Make it possible to view the token usage cost after generating an image with openai ior xai provider

## Summary
examples/ai-core/src/generate-image/open-usage.ts // This currently results in a typescript error   Property 'usage' does not exist on type 'GenerateImageResult'.ts(2339)
In further commits this error will be fixed after the usgae functionality is addressed 

## Tasks
* Add usage function to generateImage
* Add more examples of usage to other providers

## Related Issues
Addresses: https://github.com/vercel/ai/issues/8358


gr2m:
FYI https://github.com/vercel/ai/pull/8538. Problem is that we didn't consider usage in the v2 Image Model Spec, and we cannot change the spec in patch versions. Good news is that we will start working on v5.1 next week, meaning we will start shipping `ai@5.1.0-beta.X` releases, where we can update the spec. 

https://github.com/vercel/ai/pull/8538 is exposing the raw response body, but we might consider exposing usage data explicitly like we do in other models, e.g. 

https://github.com/vercel/ai/blob/a3b2e2791f58e8e9270a31e2bed538a718645541/packages/provider/src/language-model/v2/language-model-v2.ts#L66

https://github.com/vercel/ai/blob/7979f7fc95d4b132d7b6e820b07a24f80d72cb23/packages/provider/src/language-model/v2/language-model-v2-usage.ts#L7-L33

It would be helpful to research what usage tokens are currently returned across multiple providers and models

examples/ai-core/src/generate-image/open-usage.ts
```diff
@@ -0,0 +1,17 @@
...
+
+async function main() {
+  const prompt = 'Santa Claus driving a Cadillac';
+  const result = await generateImage({
+    model: openai.image('dall-e-3'),
+    prompt,
+  });
+
+  // TODO: Usage tracking is not yet implemented for image generation
+  // console.log(result.usage);
```
gr2m: good example 👍🏼 just rename the file to `openai-usage.ts`

kingsleytorlowei:
@gr2m after adding the usage to the ImageModelV2, the plan would be to implement the usage data for all the providers and models right? Can i put that all in this PR?

gr2m:
> after adding the usage to the ImageModelV2, the plan would be to implement the usage data for all the providers and models right? Can i put that all in this PR?

hold on for a few days. We will introduce a new v3 specs (side-by-side) where we can add new properties. Then we can update all the providers to use the new spec. This will be part of #8662

kingsleytorlowei:
> > after adding the usage to the ImageModelV2, the plan would be to implement the usage data for all the providers and models right? Can i put that all in this PR?
> 
> hold on for a few days. We will introduce a new v3 specs (side-by-side) where we can add new properties. Then we can update all the providers to use the new spec. This will be part of #8662

Got it, do you need any help with the v3 spec? Anything i can take on?

gr2m:
Maybe, I'll keep you posted. Thanks for offering!

aayush-kapoor:
wondering if it makes sense to close this PR since different providers have different ways of calculating usage. 

This approach isn't a unified approach that we had discussed. maybe something to tackle after v6? 

cc @gr2m @lgrammel

PR #8605, opened 2025-09-12
RFC: Pass addToolResult into onToolCall callback

aron:
## Background

This is a request for comment with a potential solution for an issue we've been seeing around tool calling.

We have an application that utilizes `onToolCall` to handle client side processing of tools, these are potentially long running and block the agent conversation until completion.

```ts
const { messages, sendMessage, addToolResult } = useChat({
    api: '/api/chat',
  }),

  sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithToolCalls,

  // run client-side tools that are automatically executed:
    // Check if it's a dynamic tool first for proper type narrowing
    if (toolCall.dynamic) {
      return;
    }

    if (toolCall.toolName === 'getWeatherInformation') {
      // Long running task
      const weather = await getWeatherInformation(toolCall.input);

      // No await - avoids potential deadlocks
      addToolResult({
        tool: 'getWeatherInformation',
        output: cities[Math.floor(Math.random() * cities.length)],
      });
    }
  },
});
```

We also support the user canceling the conversation mid flow using `chat.stop()`. This works well in most standard cases as the stream is aborted and the `sendAutomaticallyWhen()` check in the `makeRequest` call is skipped.

However if we have a long running tool, it's possible that the tool will return _after_ the cancellation happens.

The tool itself is not aware of the response stream nor of the cancellation so in the above case the `addToolResult` call will update `messages` with stale data and write that stale data to the currently active response if it exists. It will also then call `sendMessage()` with the stale data initiating a new conversation.

The end result for the user is that the conversation appears canceled and then suddenly the agent responds a short while later when the tool completes due to the additional `sendMessage` call in `addToolResult()`.

We currently work around this by maintaining our own state for canceled messages and suppress the `addToolResult` call in `onToolCall`:

```ts
    // cancellation.promise resolves when `stop()` is called so that the tool
    // call doesn't block the stream.
    return Promise.race([cancellation.promise, () => {
     // Check if it's a dynamic tool first for proper type narrowing
     if (toolCall.dynamic) {
       return;
     }

     if (toolCall.toolName === 'getWeatherInformation') {
       // Long running task
       const weather = await getWeatherInformation(toolCall.input);

       if (!cancellation.isCanceled) {
        // No await - avoids potential deadlocks
        addToolResult({
          tool: 'getWeatherInformation',
          output: weather,
        });
       }
     }
    }])
  },
```

## Summary

This PR adds two new additional arguments to `onToolCall` the `addToolResult` function bound to the currently active response and the `AbortSignal` associated with the active response.

The `addToolResult` call will not run if the response stream associated with the current `onToolCall` call is terminated.

The `signal` is provided so that users can implement their own tear down code if needed if the stream is canceled.

```ts
  async onToolCall({ toolCall, addToolResult, signal }) {
    // Check if it's a dynamic tool first for proper type narrowing
    if (toolCall.dynamic) {
      return;
    }

    if (toolCall.toolName === 'getWeatherInformation') {
      // Signal can be used potentially in multi step flows...
      const location = await getLocation(toolCall.input);
      if (signal?.aborted) {
        return;
      }

      // Long running task, can tear down if signal is aborted...
      const weather = await getWeatherInformation(location, { signal });

      const summary = await summarize(weather);

      // Regardless, result will now be a no-op if signal is aborted.
      addToolResult({
        tool: 'getWeatherInformation',
        output: weather,
      });
    }
  },
```

This PR leaves the existing `addToolResult` unbound on the public interface because I think there are valid use cases for updating a tool call outside of the current request cycle.

## Tasks

### DIFF
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -88,10 +88,29 @@ export interface ChatState<UI_MESSAGE extends UIMessage> {
 
 export type ChatOnErrorCallback = (error: Error) => void;
 
-export type ChatOnToolCallCallback<UI_MESSAGE extends UIMessage = UIMessage> =
-  (options: {
-    toolCall: InferUIMessageToolCall<UI_MESSAGE>;
-  }) => void | PromiseLike<void>;
+type ChatAddToolResult<
+  UI_MESSAGE extends UIMessage = UIMessage,
+  TOOL extends
+    keyof InferUIMessageTools<UI_MESSAGE> = keyof InferUIMessageTools<UI_MESSAGE>,
+> = ({
+  tool,
+  toolCallId,
+  output,
+}: {
+  tool: TOOL;
+  toolCallId: string;
+  output: InferUIMessageTools<UI_MESSAGE>[TOOL]['output'];
+}) => Promise<void>;
+
+export type ChatOnToolCallCallback<
+  UI_MESSAGE extends UIMessage = UIMessage,
+  TOOL extends
+    keyof InferUIMessageTools<UI_MESSAGE> = keyof InferUIMessageTools<UI_MESSAGE>,
+> = (options: {
+  toolCall: InferUIMessageToolCall<UI_MESSAGE>;
+  addToolResult: ChatAddToolResult<UI_MESSAGE, TOOL>;
+  signal: AbortSignal;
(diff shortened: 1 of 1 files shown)


aron:
Closing this as stale.

PR #8616, opened 2025-09-12
feat(provider/openai-compatible): set `user-agent` header for `createOpenAICompatible` 

### LINKED ISSUE #8612: set `user-agent` header in `createOpenAICompatible`
As added for other providers/provider-utils/`ai`, we're setting `user-agent` header for `openai-compatible`

aayush-kapoor:
## Background

As added for other providers/provider-utils/`ai`, we're setting user-agent property for `openai-compatible`

## Summary

- added a `version.ts` file to extract that package's version
- inject that version number to the user-agent string when we build and test this package
- modified the `getHeaders()` function so that we append this package's string to existing user-agent header (if any)

## Manual Verification

updated test file `openai-compatible-provider.test.ts` to ensure the package version is being attached to the header

## Tasks

## Related Issues

Fixes #8612


packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -96,10 +102,15 @@ export function createOpenAICompatible<
...
-    ...options.headers,
-  });
+  const getHeaders = () =>
+    withUserAgentSuffix(
+      {
+        ...(options.apiKey && { Authorization: `Bearer ${options.apiKey}` }),
+        ...options.headers,
+      },
+      `ai-sdk/openai-compatible/${VERSION}`,
+      getRuntimeEnvironmentUserAgent(),
```
gr2m: I think we will end up having the runtime environment twice in the user agent. This already gets added in `@ai-sdk/provider-utils`, we don't need to do that anywhere else, unless maybe in cases where a provider uses `fetch` directly instead of going through `ai`

packages/openai-compatible/src/openai-compatible-provider.ts
```diff
@@ -96,10 +102,15 @@ export function createOpenAICompatible<
...
-    ...(options.apiKey && { Authorization: `Bearer ${options.apiKey}` }),
-    ...options.headers,
-  });
+  const getHeaders = () =>
+    withUserAgentSuffix(
+      {
+        ...(options.apiKey && { Authorization: `Bearer ${options.apiKey}` }),
+        ...options.headers,
+      },
+      `ai-sdk/openai-compatible/${VERSION}`,
```
gr2m:
```suggestion
    const headers = {
      ...(options.apiKey && { Authorization: `Bearer ${options.apiKey}` }),
      ...options.headers,
    }
    withUserAgentSuffix(
      headers,
      `ai-sdk/openai-compatible/${VERSION}`,
```

packages/openai-compatible/src/openai-compatible-provider.ts
aayush-kapoor: rightt - i didn't realise that this would be added twice in the flow of postJsonToApi -> postToApi -> suffix added (same for get)

will fix and keep in mind

packages/openai-compatible/src/openai-compatible-provider.ts
aayush-kapoor: changed!

gr2m (approved):
Nice!

PR #8622, opened 2025-09-12
feat: adding user-agent to all packages that use global fetch directly

aayush-kapoor:
## Background

All the packages that use global fetch directly have to be updated so that user-agent header is set properly

## Summary

- updated tests to reflect the new user-agent header
- package version is updated during build and test phases
- exports added to use the `__PACKAGE_VERSION__`

## Manual Verification

updated tests to reflect the new user-agent header

## Tasks

## Future Work

- [x] amazon-bedrock
- [x] google-vertex

## Related Issues

fixes #8617


packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
```diff
@@ -40,12 +43,17 @@ export function createSigV4FetchFunction(
...
+      `ai-sdk/amazon-bedrock/${VERSION}`,
+      getRuntimeEnvironmentUserAgent(), // adding runtime environment user agent since this it won't be updated
+    );
     const body = prepareBodyString(init.body);
     const credentials = await getCredentials();
     const signer = new AwsV4Signer({
       url,
       method: 'POST',
-      headers: Object.entries(removeUndefinedEntries(originalHeaders)),
+      headers: Object.entries(removeUndefinedEntries(headersWithUserAgent)),
```
gr2m: I think we can remove `removeUndefinedEntries` now since we do it in `withUserAgentSuffix`

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
gr2m, line 1: the current implementation misses the `fetch()` call in line 35.

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
aayush-kapoor: oh yes, removed!

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
aayush-kapoor: added!

packages/amazon-bedrock/src/bedrock-sigv4-fetch.test.ts
```diff
@@ -4,6 +4,23 @@ import {
...
+
+// Mock provider-utils to control runtime environment detection
+vi.mock('@ai-sdk/provider-utils', async () => {
+  const actual = await vi.importActual('@ai-sdk/provider-utils');
+  return {
+    ...actual,
+    getRuntimeEnvironmentUserAgent: vi.fn(() => 'runtime/testenv'),
+    withUserAgentSuffix: actual.withUserAgentSuffix,
+    combineHeaders: actual.combineHeaders,
+    removeUndefinedEntries: actual.removeUndefinedEntries,
```
gr2m: shouldn't this work the same? 

```suggestion
    ...actual,
    getRuntimeEnvironmentUserAgent: vi.fn(() => 'runtime/testenv'),
```

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
gr2m, line 1: There is a potential bug in our current implementation. It's unrelated to this PR, but could you add it to future work and create a follow up issues?

We normalize the `url` property already

```ts
    const url =
      typeof input === 'string'
        ? input
        : input instanceof URL
          ? input.href
          : input.url // input is a `Request` object
```

We should do the same with `headers` and `body`. Both can be set on input if input is a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object.

```ts
const request = new Request("https://example.com", {
  method: "POST",
  body: '{"foo": "bar"}',
  headers: {'x-foo':'bar'},
});
const response = await fetch(request)
```

This is low priority, we didn't run into it yet, but the way we use the `fetch` function inside our models might change and then this will start failing.

gr2m (approved):

packages/amazon-bedrock/src/bedrock-sigv4-fetch.test.ts
aayush-kapoor: yupp it should and it does. removed these extra lines

packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
aayush-kapoor: created as a sub issue here: #8658

aayush-kapoor:
user-agent header is implemented for all other misc. fetch calls that we were making. but still let me know if i missed anywhere

gr2m (approved):
### ship it :shipit:

PR #8635, opened 2025-09-13
Safely handles missing reasoning parts

cdalton713:
Adds a null check to prevent errors when processing a reasoning-end message for a non-existent or already removed reasoning part.

## Summary

Getting runtime errors when two reasoning end events come in from OpenAI. 

## Manual Verification

Added validation that the reasoning event exists.


gr2m:
can you please share a full example where the error occurs, so we can test your fix? If we go ahead, this PR will need a test, and a changeset

gr2m:
Closing after no follow-up on the requested reproduction and tests. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

PR #8638, opened 2025-09-14
build: add `"type": "module"` to `package.json`

iiio2:
It is recommended to mention `"type": "module"` explicitly since we are using ESM everywhere. Otherwise it incurs some performance issue. Node.js doc says, "Syntax detection should have no performance impact on CommonJS modules, but it incurs a slight performance penalty for ES modules; add "type": "module" to the nearest parent package.json file to eliminate the performance cost. "

https://nodejs.org/en/blog/release/v20.19.0#module-syntax-detection-is-now-enabled-by-default


gr2m:
The `package.json` file in the root is not published. If anything, we'd need to update `packages/*/package.json` files. But we are good for now, thank you.

PR #8639, opened 2025-09-14
feat: typed provider options

### LINKED ISSUE #8637: We need provider-specific types for providerOptions
### Description

With the advent of GPT-5 and more reasoning configuration i.e. { reasoningEffort: 'high' } impacting TTFT dramatically, bugs can be introduced where reasoining isn't properly configured. In this case we had a bug where the API wasn't being sent reasoningEffort / reasoning.effort:

```ts
{
  provider: {
    openai: {
    } as ThisShouldBeTyped
}
```

### AI SDK Version

^4.3.16

### Code of Conduct

imranbarbhuiya:
## Background

As mentioned in #8637 provider options isn't typed, only runtime validation is there, typed would help give autocomplete and prevent more bugs

## Summary

This pr allows providers to augment the provider options to add their own type

## Tasks

## Related Issues

fixes: #8637

## Alternative

track provider options in `LanguageModelV2` generic and use it in provider options, it'll be more strict but I didnt wanted to update an existing interface so using this way. If you prefer generic option then I can change it

### DIFF
--- a/packages/openai/src/openai-provider.ts
+++ b/packages/openai/src/openai-provider.ts
@@ -11,16 +11,25 @@ import {
   loadApiKey,
   withoutTrailingSlash,
 } from '@ai-sdk/provider-utils';
-import { OpenAIChatLanguageModel } from './chat/openai-chat-language-model';
+import {
+  OpenAIChatLanguageModel,
+  OpenAIChatProviderOptions,
+} from './chat/openai-chat-language-model';
 import { OpenAIChatModelId } from './chat/openai-chat-options';
-import { OpenAICompletionLanguageModel } from './completion/openai-completion-language-model';
+import {
+  OpenAICompletionLanguageModel,
+  OpenAICompletionProviderOptions,
+} from './completion/openai-completion-language-model';
 import { OpenAICompletionModelId } from './completion/openai-completion-options';
 import { OpenAIEmbeddingModel } from './embedding/openai-embedding-model';
 import { OpenAIEmbeddingModelId } from './embedding/openai-embedding-options';
 import { OpenAIImageModel } from './image/openai-image-model';
 import { OpenAIImageModelId } from './image/openai-image-options';
 import { openaiTools } from './openai-tools';
-import { OpenAIResponsesLanguageModel } from './responses/openai-responses-language-model';
+import {
+  OpenAIResponsesLanguageModel,
+  OpenAIResponsesProviderOptions,
+} from './responses/openai-responses-language-model';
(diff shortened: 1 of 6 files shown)


imranbarbhuiya:
I would like you guys to check and let me know if this approach is fine then i'll go ahead with updating all other providers
image

@dancer

packages/openai/src/openai-provider.ts
```diff
@@ -249,3 +258,12 @@ export function createOpenAI(
 Default OpenAI provider instance.
  */
 export const openai = createOpenAI();
+
+declare module '@ai-sdk/provider-utils' {
+  interface RegisterProviderOptions {
+    openai?:
+      | OpenAIResponsesProviderOptions
+      | OpenAIChatProviderOptions
+      | OpenAICompletionProviderOptions;
```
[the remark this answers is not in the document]
imranbarbhuiya: This is invalid and incorrect too. It does shows autocomplete for fields that are in 1 union only

gr2m:
Hi @imranbarbhuiya in general I do love the idea of having providerOptions typed automatically. However I do remember @lgrammel mentioning that he tried that in the past and gave up due to complexity and I think the fact that the AI SDK's model registry can break the assumptions you are making in this PR. I wanted to talk to Lars about your PR today but we didn't get to it. I'll make sure to chat with him about it tomorrow and get back to you

imranbarbhuiya:
Can you explain more about this part
> I think the fact that the AI SDK's model registry can break the assumptions you are making in this PR.

Since all the options passed to it is already getting validated using the same zod schema so if anything is outdated/not matching, it'll give same runtime error too

lgrammel:
Fwiw #8637 did not seem aware of the available ways to check the type safety.

imranbarbhuiya:
Imo while that works but providing proper typed options without asking users to import will be better dx

packages/openai/src/openai-provider.ts
gr2m, line 267: We think this might not be an improvement, the number of `*ProviderOptions` will grow. We think we want to invest into the explicit `import ...` and `... satisfies ...` approach for now. We do agree eventually we should go down the path of your PR eventually, we are just not ready yet.

imranbarbhuiya: Got it, thanks for the explanation. Also I think not all provider options are exported and only the response one is exported (not checked properly) so is it intended or others needs to be exported too?

gr2m: yes, I have a set of issues prepared for it, maybe for Hacktoberfest :)

PR #8654, opened 2025-09-15
docs(examples): switch next example to use AI Gateway provider

fveiraswww:
## Background

the Next.js example was using the OpenAI provider directly, which requires users to have an OpenAI API key and limits them to only OpenAI models. By switching to AI Gateway, users can access multiple AI providers

## Summary

- Replaced `@ai-sdk/openai` provider with a direct model string
- Updated the Next.js example to use the cheaper `gpt-5-nano` model

## Manual Verification

✅  @gr2m verified

### DIFF
--- a/examples/next/app/api/chat/route.ts
+++ b/examples/next/app/api/chat/route.ts
@@ -1,5 +1,4 @@
 import type { MyUIMessage } from '@/util/chat-schema';
-import { openai } from '@ai-sdk/openai';
 import { readChat, saveChat } from '@util/chat-store';
 import { convertToModelMessages, generateId, streamText } from 'ai';
 import { after } from 'next/server';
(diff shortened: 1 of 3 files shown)


gr2m (approved):
I think that makes sense but will double check with @lgrammel. Can you also update the `package.json` file since this change removes the dependency on `@ai-sdk/openai`?

examples/next/app/api/chat/route.ts
```diff
@@ -57,7 +56,7 @@ export async function POST(req: Request) {
   saveChat({ id, messages, activeStreamId: null });
 
   const result = streamText({
-    model: openai('gpt-4o-mini'),
+    model: 'openai/gpt-5-nano',
```
nicoalbanese:
```suggestion
    model: 'openai/gpt-5-mini',
```

gr2m: I verified that this works 👍🏼 let's do it.

fveiraswww:
> I think that makes sense but will double check with @lgrammel. Can you also update the `package.json` file since this change removes the dependency on `@ai-sdk/openai`?

done ✅

lgrammel:
examples are for our internal use, not for documentation. this now hinders my work. can we revert the pr?

lgrammel:
pls ignore, mistook this for `examples/next-openai`

PR #8680, opened 2025-09-16
feat(ai): add safeValidateUIMessages

### LINKED ISSUE #8673: [feature request] Safe variant of validateUIMessages
## Problem

Currently, validateUIMessages throws when validation fails.

This behavior matches zod's .parse function, so is expected and should be kept the same.

However, in API endpoints, it's more ergonomic to use Zod's `.safeParse` function, which returns a discriminated union of success and failure states. The reason is that without it, you need to wrap just the validateUIMessages section in a try/catch.

```ts
let messages: UIMessage[];

try {
  messages = await validateUIMessages({
    messages: body.messages,
  });
} catch (e) {
  return new Response('Invalid messages', { status: 400 });
}
```

This is without extracting the specific thing that failed from the `e` variable, which is a nightmare and either involves a cast or some complex narrowing logic.

## The Solution

Having a `safeValidateUIMessages` function (or `validateUIMessages.safe()`) would match the ergonomics of `.safeParse`:

```ts
const result = await safeValidateUIMessages({
  messages: body.messages,
});

if (!result.success) {
  console.log(result.reason); // Type safe errors!
  return new Response('Invalid messages', { status: 400 });
}

console.log(result.data);
``` 

## Code of Conduct

fveiraswww:
## Background

The current validateUIMessages throws on validation failure. For API endpoints and similar contexts it's ergonomically preferable to use a safe-return pattern (like zod's safeParse) to avoid try/catch and get a typed success/failure result. 

see context in #8673 

## Summary

- Add `safeValidateUIMessages`
- Returns a discriminated union: `{ success: true, data }` || `{ success: false, reason, error }`

## Manual Verification

- Ran unit tests in packages/ai

## Tasks

## Related Issues

fixes #8673


packages/ai/src/ui/validate-ui-messages.test.ts
```diff
@@ -1106,3 +1109,139 @@ describe('validateUIMessages', () => {
...
+        {
+          id: '1',
+          role: 'user',
+          parts: [{ type: 'text', text: 'Hello, world!' }],
+        },
+      ],
+    });
+
+    expect(result.success).toBe(true);
+    if (result.success) {
```
gr2m: the if check is not necessary. If it's not true, the assertion above will throw. Same for the other tests

gr2m (commented):
could you add documentation as we have for `validateUIMessages()`: https://ai-sdk.dev/docs/reference/ai-sdk-core/validate-ui-messages

packages/ai/src/ui/validate-ui-messages.ts
```diff
@@ -286,3 +286,61 @@ export async function validateUIMessages<UI_MESSAGE extends UIMessage>({
...
+      reason: string;
+      error: Error;
+    };
+
+/**
+ * Validates a list of UI messages without throwing.
+ *
+ * Metadata, data parts, and generic tool call structures are only validated if
+ * the corresponding schemas are provided. Otherwise, they are assumed to be
+ * valid.
```
gr2m: Instead of mostly copying the description, I think I'd describe it as being the same as `validateUIMessages`, but instead of throwing it returns an object with a `success` key and either `error` or `data`

packages/ai/src/ui/validate-ui-messages.ts
```diff
@@ -286,3 +286,61 @@ export async function validateUIMessages<UI_MESSAGE extends UIMessage>({
...
 }
+
+type SafeValidateUIMessagesResult<UI_MESSAGE extends UIMessage> =
+  | {
+      success: true;
+      data: Array<UI_MESSAGE>;
+    }
+  | {
+      success: false;
+      reason: string;
```
gr2m: do we need `reason`? I think using `error.message` is just as ergonomic, and more clear.

gr2m:
thanks for the pull request 👍🏼 if you could comment on the issue that you are working on, that would be great! For one, it allows us to assign the issue to you, and second people know that somebody is already working on it and won't duplicate your work

packages/ai/src/ui/validate-ui-messages.ts
```diff
@@ -286,3 +286,61 @@ export async function validateUIMessages<UI_MESSAGE extends UIMessage>({
...
+    return { success: true, data };
+  } catch (error) {
+    const err = error as Error;
+    return {
+      success: false,
+      reason: err.message,
+      error: err,
+    };
+  }
+}
```
lgrammel, line 358: instead of try catch around validateUIMessages, can we look into using safeValidate as default and have validateUIMessages wrap safeValidate so we do not throw/catch errors unecessarily

fveiraswww: sounds good, so something like:

```ts
export async function validateUIMessages<>(...): Promise<Array<UI_MESSAGE>> {  
  const res = await safeValidateUIMessages<UI_MESSAGE>(args);  
  if (!res.success) throw res.error;  
  return res.data;  
}
```

that way `safeValidateUIMessages` is the default and `validateUIMessages` just wraps it

packages/ai/src/ui/validate-ui-messages.test.ts
fveiraswww: if I remove the if I get a ts error because expect().toBe(true) doesn't narrow the type. I may be missing something.

`Property 'data' does not exist on type '{ success: false; error: Error; }'`

the workarounds I see look a bit ugly, what do you suggest?

fveiraswww:
@gr2m I added the docs following the same style as validateUIMessages and applied the other fixes

@lgrammel I refactored validateUIMessages to wrap safeValidateUIMessages as suggested ✅

packages/ai/src/ui/validate-ui-messages.test.ts
gr2m: Ah okay I see. Bummer that vitest doesn't have an assertion with type-guard

Maybe add a helper method like this

```ts
export function expectToBe<T extends boolean>(
  value: boolean,
  expected: T,
): asserts value is T {
  expect(value).toBe(expected);
}
```

And use that instead of `expect().toBe()`

```diff
- expect(result.success).toBe(true);
+ expectToBe(result.success, true);
```

I think either works. I personally prefer the helper method, conditional code in tests feels wrong to me

fveiraswww: thxx, I added the helper as suggested, looks much cleaner now

gr2m (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/8742

PR #8692, opened 2025-09-17
feat(codemod): add datastream to uimessagestream transformation

dancer:
## background

some datastream to uimessagestream transformations were missing from the v5 codemods. users migrating from v4 to v5 needed to manually update DataStream, DataStreamWriter, DataStreamOptions, and related functions that weren't covered by existing codemods

## summary

- added codemod to transform remaining datastream types and functions to uimessagestream equivalents
- covers DataStream, DataStreamWriter, DataStreamOptions, createDataStream, createDataStreamResponse

## verification

tested transformation with fixtures

## tasks

- [x] created replace-datastream-to-uimessagestream codemod
- [x] added test fixtures and tests
- [x] integrated into v5 upgrade bundle
- [x] updated readme with new codemod entry


packages/codemod/src/test/__testfixtures__/replace-datastream-to-uimessagestream.output.ts
```diff
@@ -0,0 +1,42 @@
+// @ts-nocheck
+import {
+  createUIMessageStreamResponse,
+  createUIMessageStream,
+  UIMessageStream,
```
gr2m: This export does not exist. Nor does `DataStream` from v4. 

I couldn't find a decleration of `UIMessageStream` anywhere in our code. Maybe it should exist [here](https://github.com/vercel/ai/blob/383cbfa9e9c57b7582a752a006ab8f20b9150de3/packages/ai/src/ui-message-stream/create-ui-message-stream.ts#L12) as return type of `createUIMessageStream()` but it doesn't

dancer: Fixed

gr2m (approved):

PR #8725, opened 2025-09-18
feat(codemod): add usechat input state transformation for v5

dancer:
## background

the v5 migration removes managed input state from the usechat hook. users need to manually manage input state with usestate instead of relying on the built-in input and handleinputchange from usechat. without this codemod, users would need to manually update every usechat usage in their codebase

## summary

- added codemod to transform usechat destructuring patterns
- removes input and handleinputchange from usechat
- adds usestate for local input management
- replaces handleinputchange with inline setters

## verification

tested with real-world patterns including aliased destructuring, partial extraction, and function-scoped replacements. all tests pass

## tasks

- [x] created replace-usechat-input-with-state codemod
- [x] handled import alias preservation
- [x] added function-scoped identifier replacement
- [x] created test fixtures for all edge cases
- [x] integrated into v5 upgrade bundle


packages/codemod/src/test/__testfixtures__/replace-usechat-input-with-state.input.ts
```diff
@@ -0,0 +1,48 @@
+// @ts-nocheck
+import { useChat } from '@ai-sdk/react';
```
gr2m: This codemod should also handle/upgrade imports from `ai/react`, to account for the possibility that the other codemod is not run first

https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#react-ui-hooks-moved-to-ai-sdkreact

gr2m: addressed via 056cd8809 and d8389e1b1

gr2m (approved):

PR #8847, opened 2025-09-23
ci(backport): comment in case of error

gr2m:
towards #8845


gr2m:
gonna self merge is it's just maintenance and it would have helped me already once today

PR #8439, opened 2025-09-04
feat(gateway): add automatic oidc token refresh 

dancer:
## background

vercel oidc tokens expire during development causing authentication failures that require manual `vc env pull` commands

## summary

- automatically refresh expired oidc tokens using cached tokens and vercel cli credentials
- check expiry to prevent mid-request failures
- fallback to original token if refresh fails
- edge-compatible with dynamic imports for node.js modules

## verification

all tests pass including new token refresh scenarios. compatibility maintained with dynamic imports

## tasks

- [x] tests have been added for token refresh functionality
- [x] browser compatibility ensured with dynamic node.js module imports

PR #8975, opened 2025-09-27
feat(gateway): oidc refresh

gr2m:
supersedes #8439.

My approach is to 

1. utilize conditional imports/exports in order to run different code in different environments
2. I pretty much embedded the [`@vercel/oidc` package](https://www.npmjs.com/package/@vercel/oidc?activeTab=code).
3. If this works, it would be nice to implement conditional exports upstream and add `@vercel/oidc` as dependency, which would remove the need of any conditional imports/exports in our code
4. As an alternative/first step, we could add `@vercel/oidc` as a dependency, but use conditional imports for browsers to replace the module with a no-op mock.

Still need to test this 👋🏼


gr2m:
I cannot reproduce the error with `pnpm run type-check:full` locally 🤷🏼😭

gr2m:
Looking into using `@vercel/oidc` as dependency instead of vendoring in all the code:

- vercel/vercel#14003
- vercel/vercel#14004
- vercel/vercel#14005

gr2m:
closing in favor of https://github.com/vercel/ai/pull/8982

PR #8985, opened 2025-09-28
fix(react): make useChat callbacks not stale

### LINKED ISSUE #8148: useChat V5 appears to cache the onToolCall callback but offers no possibility to invalidate
### Description

Despite **not** offering a dependency array, useChat appears to cache the initial version / value of the passed onToolCall callback.

## What does not work:

```ts
// users can change the used resources in a different component inside the same
// context. the resources get used by executeGenerateText within onToolCall
const {resources} = useResources()

const {messages, sendMessage, stop, status, setMessages, addToolResult} = useChat({
    api: ApiEndpoints.aiChat,
  }),

  sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithToolCalls,

  onToolCall: async ({toolCall}) => {
    let result: string = 'Done'

    // BUG
    // retains the initial value of resources on component mount, even after
    // mutations by the user
    console.log(resources)

    switch (toolCall.toolName) {
      case TextTools.generate_new_text:
        result = await executeGenerateText(
          toolCall.toolCallId,
          toolCall.input as GenerateNewTextArgs,
          // BUG
          // retains the initial value of resources on component mount, even
          // after mutations by the user
          resources,
        )
        break
    }

    addToolResult({
      tool: toolCall.toolName,
      output: result,
    })
  },
  onError: console.error,
})
```

## What does work:

```ts
// users can change the used resources in a different component inside the same
// context. the resources get used by executeGenerateText within onToolCall
const {resources} = useResources()

// writing the resources inside a ref
const resourcesRef = useRef<Resource[]>(resources)
resourcesRef.current = resources

const {messages, sendMessage, stop, status, setMessages, addToolResult} = useChat({
    api: ApiEndpoints.aiChat,
  }),

  sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithToolCalls,

  onToolCall: async ({toolCall}) => {
    let result: string = 'Done'

    // WORKS
    // has the the correct value, even after mutations by the user
    console.log(resourcesRef.current)

    switch (toolCall.toolName) {
      case TextTools.generate_new_text:
        result = await executeGenerateText(
          toolCall.toolCallId,
          toolCall.input as GenerateNewTextArgs,
          // WORKS
          // has the the correct value, even after mutations by the user
          resourcesRef.current,
        )
        break
    }

    addToolResult({
      tool: toolCall.toolName,
      output: result,
    })
  },
  onError: console.error,
})
```

### AI SDK Version

- "ai": "^5.0.11",
- "@ai-sdk/react": "^2.0.11",

### Code of Conduct

arnaugomez:
## Background

`useChat` captures the initial `onToolCall` callback via closures, offering no mechanism to refresh the callback during a chat’s lifetime. When the React component changes context/state (e.g., changing the value of a `useState` hook), `onToolCall` continues using stale values, leading to incorrect client-side tool execution and unexpected behavior. The only way to fix this was to wrap the values inside `useRef`.

## Summary

- Core (`ai`): add runtime setters on `AbstractChat` to update callbacks without recreating the chat instance:
  - `setOnToolCall`, `setOnData`, `setOnFinish`, `setOnError`, `setSendAutomaticallyWhen`.
- React (`@ai-sdk/react`): update `useChat` to call these setters inside an effect when options change, ensuring the latest callbacks are always used.
- Tests: add a regression test ensuring the latest `onToolCall` is invoked after a prop change.

## Manual Verification

- Local example with a toggle that switches the `onToolCall` implementation at runtime (simulates changed `resources` context). After the toggle, subsequent tool calls were handled by the updated callback and used the latest context values.
- Confirmed chat instances were not re-created when only callback props changed.
- Verified streaming tool calls still dispatch and complete as expected.

## Checklist

## Future Work

It appears the transport is cached, too. This has been reported in the following issue: #7819

## Related Issues

Fixes #8148


arnaugomez:
plz review plz 🐱 👉 👈

gr2m:
@arnaugomez apologies for the wait time, please know we appreciate your pull request. The core team has to focus on what we want to finish for the https://vercel.com/ship/ai conference, we will take some time to catch up with community issues and pull requests after that

arnaugomez:
Thanks @gr2m , please take your time and enjoy the Vercel Ship conference! I love Next.js and Vercel, you guys make great innovations. 

Fluid compute in particular has been really helpful for AI workflows, because it lets me make a call to the AI provider that might take a long time to respond, without worrying too much about getting billed heavily or receiving a timeout for the serverless function execution. I'll watch the conference 💯

taranek:
@gr2m Any idea when this can be merged?

lgrammel:
Instead of changing the chat interface in the ai package, would it be possible to keep intermediary event listeners that forward in useChat? those could then forward to the correct current listeners that the user provides

this would limit the change to the `@ai-sdk/react` package

arnaugomez:
Yes, I was thinking the same! Great idea @lgrammel . I'll modify this PR.

arnaugomez:
Thanks for the feedback @lgrammel! I've updated the PR to use intermediary proxy callbacks in `useChat`. This ensures the `ai` package is not modified and all changes are now isolated to `@ai-sdk/react`.

The callbacks ref is updated on each render, and the wrapper functions passed to the Chat instance read from this ref, so they always use the latest user-provided callbacks without needing setter methods on `AbstractChat`.

packages/react/src/use-chat.ts
```diff
@@ -60,16 +60,52 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
...
+    !('chat' in options)
+      ? {
+          onToolCall: options.onToolCall,
+          onData: options.onData,
+          onFinish: options.onFinish,
+          onError: options.onError,
+          sendAutomaticallyWhen: options.sendAutomaticallyWhen,
+        }
+      : {},
+  );
```
lgrammel, line 74: since it is updated below can you just init to `{}`?

lgrammel (approved):

packages/react/src/use-chat.ts
arnaugomez: Yes, although then it would lose type inference.

lgrammel: guess it is fine, ideally we would use generics for the types and init to `{}` to reduce the bundled code size

arnaugomez: I can create a PR refactoring it. I'll create one later.

PR #8988, opened 2025-09-28
Model id with '/' not work

### LINKED ISSUE #8931: Model id with '/' not work
### Description

For example: google/gemini-2.5-pro

Image

The request path will lose the models parameter

### AI SDK Version

@ai-sdk/google latest

### Code of Conduct

sujal12344:
Fixes #8931

## Background

getModelPath was returning incorrect paths for some model IDs.
IDs containing slashes were passed through correctly, but IDs without slashes needed the 'models/' prefix. 
Unit tests were missing for edge cases.

## Summary

- Updated getModelPath to handle both cases properly.
- Added unit tests:
  - IDs with '/'
  - IDs without '/'
  - Optional: empty string (edge case)

## Manual Verification

- Ran `vitest` and verified all tests pass.
- Checked locally by calling getModelPath with different model IDs.

## Checklist

- [ ] A _patch_ changeset for relevant packages has been added (not needed)
- [ ] Documentation has been added / updated (not needed for this change)
- [ ] Formatting issues have been fixed

## Future Work

- Consider validating empty string model IDs more strictly in future PRs.

## Related Issues

- Fixes #8931


packages/google/src/get-model-path.ts
```diff
@@ -1,3 +1,6 @@
 export function getModelPath(modelId: string): string {
+  if (modelId.startsWith('google/')) {
```
kangfenmao, line 2: https://github.com/CherryHQ/cherry-studio/blob/main/.yarn/patches/%40ai-sdk-google-npm-2.0.14-376d8b03cc.patch

This approach would be better, as it can handle model names with other prefixes.

sujal12344: Yes, I’m waiting for reviews from @gr2m and @lgrammel.

kangfenmao:
```
diff --git a/dist/index.mjs b/dist/index.mjs
index b957cb824faa79cf01ba3a504f221870bd8e306a..4d71d30f655775d61537d9d8b73f6e17d41fa67e 100644
--- a/dist/index.mjs
+++ b/dist/index.mjs
@@ -452,7 +452,7 @@ function convertToGoogleGenerativeAIMessages(prompt, options) {
 
 // src/get-model-path.ts
 function getModelPath(modelId) {
-  return modelId.includes("/") ? modelId : `models/${modelId}`;
+  return modelId?.includes("models/") ? modelId : `models/${modelId}`;
 }
 
 // src/google-generative-ai-options.ts
```

Any updates?

packages/google/src/get-model-path.ts
kangfenmao: > Yes, I’m waiting for reviews from @gr2m and @lgrammel.

The patch preserves the original logic and is compatible with model names containing '/', and many of our users are using it normally and it's working well.

kangfenmao: > Yes, I’m waiting for reviews from @gr2m and @lgrammel.

You can consider using the version I wrote, as it is compatible with model names containing "/".

n1ckoates:
Original issue closed

gr2m:
Closing: this model ID format is for aggregation/proxy services.

PR #9010, opened 2025-09-29
fix(cookbook): update import paths to include file extensions for MCP tools

xn1cklas:
## Background

The imports in the cookbook miss the `.js` for the anthopic mcp imports. Without them, copying the code results in import errors.

## Summary

Correcting the imports to match the exports described in the anthropic mcp js sdk docs.

## Checklist


gr2m:
I think it depends on whether you import it in plain JS or TS code, is that possible? I don't think the `.js` extension causes any harm though

xn1cklas:
> I think it depends on whether you import it in plain JS or TS code, is that possible? I don't think the `.js` extension causes any harm though

The anthropic docs list the `.js` endings and my ts compiler couldn't find the types when I left it out. Might be a setup problem on my side though!

Issue #8909, opened 2025-09-25
V2 to V3 adapter


aayush-kapoor:
Follow up issue for making changes on already merged PRs

Addressing comments:
- [#1 ](https://github.com/vercel/ai/pull/8837#issuecomment-3326883479)
- [#2](https://github.com/vercel/ai/pull/8893#pullrequestreview-3266032968)

And similar possible changes on #8777

aayush-kapoor:
[#1](https://github.com/vercel/ai/pull/8837#issuecomment-3326883479) has been addressed.

lgrammel:
## Identify issue type

Identified as **None**  
Confidence: **Medium**

### Rationale
This is primarily a maintenance/bookkeeping task with insufficient standalone detail to classify as a bug, feature, documentation, or examples request.

### Evidence
- The issue is a follow-up tracker for changes requested on already merged pull requests.
- The description provides only links to review comments and mentions similar possible changes, without describing a concrete failure or requested capability.
- The sole comment states that one linked item has already been addressed.

PR #9022, opened 2025-09-29
feat: add support for v2 specs

### LINKED ISSUE #9018: v3 spec changes: export type LanguageModel = string | LanguageModelV2 | LanguageModelV3
This line needs to be updated

https://github.com/vercel/ai/blob/96fd4c5e92075e7dd34350d194bb5676c3833471/packages/ai/src/types/language-model.ts#L11C13-L11C53

aayush-kapoor:
## Background

To allow for community providers to still work with V2 specs.
#9018

## Summary

Added support for V2 specs for LanguageModel and EmbeddingModel

## Checklist

## Future Work

Add adapter for v2->v3 (#8909)

## Related Issues

fixes #9018


packages/ai/src/model/resolve-model.ts
```diff
@@ -10,10 +10,7 @@ import { LanguageModel } from '../types/language-model';
 
 export function resolveLanguageModel(model: LanguageModel): LanguageModelV3 {
   if (typeof model !== 'string') {
-    if (
-      model.specificationVersion !== 'v3' &&
-      model.specificationVersion !== 'v2'
-    ) {
+    if (model.specificationVersion !== 'v3') {
```
aayush-kapoor: @gr2m because of this change, we indirectly still don't support v2. Leaving the line as:
```
if (
      model.specificationVersion !== 'v3' &&
      model.specificationVersion !== 'v2'
    ) {
```
gives a type error

we will eventually need an adapter. do we have clarity on what the structure of that might look like?

gr2m: The type error is probably something like this?

Image

I'm not sure why typescript infers `model` as `never` when we set

```
export type LanguageModel = string | LanguageModelV3 | LanguageModelV2;
```

but not with the current code when we do

```
export type LanguageModel = string | LanguageModelV3;
```

In the current code, even when checking that `model.specificationVersion` is not `'v3'`, it still infers the `model` type as `LanguageModelV3`, which makes no sense to me. `model` should have type `never` inside this `if` block

Image

It would be a fun exercise to dive deep and figure out what is happening, but I think the type error we are getting makes sense, and we should workaround it like this

```ts
const unsupportedModel: any = model;
throw new UnsupportedModelVersionError({
  version: unsupportedModel.specificationVersion,
  provider: unsupportedModel.provider,
  modelId: unsupportedModel.modelId,
});
```

aayush-kapoor: yeah that's weird.

applying this workaround for now and will deepdive into it as future work

aayush-kapoor: the workaround also leads to a type error:

Screenshot 2025-09-29 at 2 50 25 PM

Screenshot 2025-09-29 at 2 50 43 PM

i think this is also forcing an adapter to be implemented

gr2m: ah yes, sorry. Can you add a method to the same for now, sth like this?

```ts
function transformToV3LanguageModel(model: LanguageModelV2): LanguageModelV3 {
  return {
    ...model,
    specificationVersion: 'v3',
  };
}
```

gr2m (approved):

PR #9041, opened 2025-09-29
fix(provider/anthropic): add document support in tool results

### LINKED ISSUE #8014: Unable to pass PDF files from the tool calls
### Description

```typescript
import { tool } from 'ai';
import { z } from 'zod';

  return tool({
    inputSchema: z.object({
    }),
      // Logic of fetching the file
      // ...

      return {
        name: displayName,
        type: 'pdf',
        data: base64,
        text: '',
      };
    },
    toModelOutput: (result) => {
        type: string;
        data: string;
        text: string;
      };
      switch (type) {
        case 'text':
        case 'pdf':
          return {
            type: 'content',
            value: [
              {
                type: 'media',
                data,
                mediaType: 'application/pdf',
              },
            ],
          };
        default:
      }
    },
  });
};
```

I'm running on bun. 

### AI SDK Version

- "ai": "5.0.0",
- "@ai-sdk/google": "2.0.0",

nlaz:
## Background

Right now we are unable to pass documents by url back from a tool call to be processed by Anthropic. Anthropic's API handles this [out of the box](https://docs.claude.com/en/docs/build-with-claude/pdf-support). Just needed to update the types on the tool result handling.

## Summary

This PR updates the `convertToAnthropicMessagesPrompt` method to support document types in the tool results. The `AnthropicDocumentContent` type was already there so just needed to add it to the tool result type.

## Manual Verification

Tested this locally with a script to fetch a document and used the toModelOutput to convert to this format. Example tool call. With these changes agent went from hallucinating about the document to actually analyzing it.

```typescript
const fetchDocumentTool = () =>
  tool({
    name: 'fetchDocument',
    description: 'Fetch a document URL',
    inputSchema: z.object({}),
    execute: async () => {
      return [{
        type: 'media',
        data: new URL('https://arxiv.org/pdf/2410.00749'),
        mediaType: 'application/pdf',
      }, {
        type: 'text',
        text: `Successfully loaded document`,
      }];
    },
    toModelOutput: (output) => {
      return { type: 'content', value: output };
    },
  });
  ```

## Checklist

## Related Issues
Fixes https://github.com/vercel/ai/issues/8014

### CODE REVIEW

**Review by gr2m — COMMENTED** [review]
approach looks good, thank you! Can you please follow the pull request check list?

**Comment by nlaz** [comment]
@gr2m yep! on it

File: packages/provider/src/language-model/v3/language-model-v3-prompt.ts [hunk 1]
```diff
@@ -204,9 +204,9 @@ Text content.
             type: 'media';
 
             /**
-Base-64 encoded media data.
-*/
-            data: string;
+            Media data. Can be a base64 encoded string or a URL.
+            */
+            data: string | URL;
```
nlaz, line 209: chose to overload the media part here instead of creating a new document content for tool results since documents are not supported by every provider. lmk if that is reasonable

gr2m: I'll get back to you on that

gr2m: fyi in general we cannot do any spec changes for standard releases. But we are in preparation of 5.1 now, for which we are preparing the new v3 spec. So it might be possible, but I have to discuss it with the team

nlaz: makes sense! alternatively the data field can stay a string and we can test if the string is a base64 type or url type but that also breaks the spirit of this type contract

gr2m: yeah sometimes we do hacks like that to avoid changing the spec. But as we are in the process of preparing the v5.1 release, we can as well do it properly

gr2m: we are looking into it as part of a bigger overhaul of `LanguageModelV3ToolResultOutput`. This is a good API requirement though, we will make sure to incorporate it

nlaz:
@gr2m I saw the AI SDK team is aiming for the v6 release instead of the v5.1. I also saw the backport label got added to this PR does that mean it will be added into v5? Anything I can help push that forward?

gr2m:
we can't backport it to v5.0 as long as the PR changes the spec. Can you make it work with the existing spec, and then do a follow up pull request where we change the spec? 

I also think that instead of changing the type to `string | url` we should probably add a new property `url` and make `data` optional.

nlaz:
@gr2m New PR without the spec changes opened here: https://github.com/vercel/ai/pull/9503

lgrammel:
We first need to do this properly on main. Then in the backport we can consider hacks (or have a separate pr that target release v5)

lgrammel:
hm the other pr uses the same branch it seems so i cant continue this one

packages/provider/src/language-model/v3/language-model-v3-prompt.ts
lgrammel: updated spec is now ready

nlaz:
🤦‍♂️ embarrassing sorry! @lgrammel anything you need me to do?

cloudycotton:
@gr2m @lgrammel https://github.com/vercel/ai/issues/9632

I'm running into same issue when using computer-use with AI-SDK. Request payload size is exhausted with base64 images. I see the changes to spec that adds image-url as a separate tool output (not yet implemented on anthropic conversion yet). 

Should i make a PR to add support for it in anthropic side? This is a pretty big blocker for us

Issue #8868, opened 2025-09-24
Undocumented breaking change from SDK 4: structured output strict schemas went from default true to default false


AndrewPrifer:
### Description

In SDK 4, strict: true was the default when structured outputs were enabled, both for the chat and the responses models. This changed to false in SDK 5, which can now break schema validation.

A related thing that makes this confusing is that the one related entry in the migration doc talks about structuredOutputs defaulting to true in SDK 5, which was a big red herring when debugging this issue, since responses models always defaulted to structuredOutputs: true, even in SDK 4.

I propose adding a conspicuous warning about the schema strictness change to the docs, both in the migration guide and the provider docs, and clarifying that the default for structuredOutputs only changed for chat models.

### AI SDK Version

ai: 5.0.0
@ai-sdk/openai: 2.0.0

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

xoyseau:
True that.

I noticed that some of my `streamObject` and `generateObject` calls started failing after I migrated from SDK v4 –> SDK v5 (and from `@ai-sdk/openai` v1 –> `@ai-sdk/openai` v2).

It started passing `strict: false` instead of `strict: true` to the requests, and because of that the model's responses didn't always conform to the provided Zod schema, which in turn made it fail.

After an hour of googling and reading the source code, I tried making it like that:

```ts
        const { object } = await generateObject({
          schema: ...,
          system: ...,
          prompt: ...,
          model: openai("gpt-4.1"),
          providerOptions: {
            openai: {
              structuredOutputs: true,
              strictJsonSchema: true,
            },
          },
        });
```

And it worked.

I couldn't find this info anywhere in the AI SDK documentation nor in the Google. I only found PR https://github.com/vercel/ai/pull/6925 and [this mention in the documentation](https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#structured-outputs-default) that `structuredOutputs: true` is now the default (which is not about the `strict` change per se).

But providing those two flags in `providerOptions.openai` helped me.

AndrewPrifer:
@xoyseau `strictJsonSchema` is actually documented in the openai provider options page (used to be called `strictSchema` in v4, which defaulted to true), it's just not documented in the migrations page and as you said it's really confusing to figure out what's happening. I also had to dig into the code before I figured out that this option is what's messing things up.

Indeed the issue is that https://github.com/vercel/ai/pull/6925 introduced several breaking changes, none of which found their way into the migration doc.

[...]

PR #6925, opened 2025-06-30
fix (provider/openai): default strict mode to false

lgrammel:
## Background

In AI SDK 5, strict mode for tool calling and output generation was enabled by default for the OpenAI responses and chat language models.

However, this caused several issues, e.g. #6913 #6642 #4662

## Summary

* default strict mode to false for openai responses and chat language models (breaking change)
* introduce provider option `strictJsonSchema` for chat language models
* rename provider option `strictSchemas` to `strictJsonSchema` for responses language model (breaking change)

## Related Issues

Fixes #6913 #6642 #4662

PR #9099, opened 2025-10-01
fix: update migration guide for strict mode and structured output changes

nicoalbanese:
## Background

Addresses #8868.

## Summary

Updated migration guide based on missing information.

## Checklist

### DIFF
--- a/content/docs/08-migration-guides/26-migration-guide-5-0.mdx
+++ b/content/docs/08-migration-guides/26-migration-guide-5-0.mdx
@@ -2941,24 +2941,37 @@ return createUIMessageStreamResponse({ stream });
 
 #### Structured Outputs Default
 
-Structured outputs are now enabled by default for supported OpenAI models.
+The `structuredOutputs` parameter has been removed from the model instance and is now disabled by default. It can be enabled via provider options.
 
 ```tsx filename="AI SDK 4.0"
+import { openai } from '@ai-sdk/openai';
+
 const result = await generateText({
   model: openai('gpt-4.1-2024-08-06', { structuredOutputs: true }),
 });
 ```
 
 ```tsx filename="AI SDK 5.0"
+import { openai, type OpenAIResponsesProviderOptions } from '@ai-sdk/openai';
+
 const result = await generateText({
   model: openai('gpt-4.1-2024-08-06'),
-  // structuredOutputs: true is now the default
+  providerOptions: {
+    openai: {
+      structuredOutputs: true, // opt-in via provider options
+    } satisfies OpenAIResponsesProviderOptions,
+  },
 });
 ```
 
-#### Compatibility Option Removal
+#### Strict Mode Changes & Compatibility Option Removal
 
-The `compatibility` option has been removed; strict mode is now the default.
+The `compatibility` option has been removed.
+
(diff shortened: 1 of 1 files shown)


content/docs/08-migration-guides/26-migration-guide-5-0.mdx
```diff
@@ -2941,24 +2941,37 @@ return createUIMessageStreamResponse({ stream });
...
 
 ```tsx filename="AI SDK 5.0"
+import { openai, type OpenAIResponsesProviderOptions } from '@ai-sdk/openai';
+
   model: openai('gpt-4.1-2024-08-06'),
-  // structuredOutputs: true is now the default
+  providerOptions: {
+    openai: {
+      structuredOutputs: true, // opt-in via provider options
```
lgrammel: wrong prop name?

lgrammel: should be `strictJsonSchema`

lgrammel (approved):

AndrewPrifer:
@lgrammel  Correct me if I'm wrong but this seems incorrect. `structuredOutputs` wasn't replaced with `strictJsonSchema`, `strictSchemas` was. And the real breaking change is that it now defaults to false instead of true.

It is also more nuanced than that since `strictJsonSchema` is a new property for chat models (which also defaulted to true, just without a property to change it to false), whereas it is a rename of `strictSchemas` for responses models.

Moreover, responses models always defaulted `structuredOutputs` to true, even in 4.0. In v5, both chat and responses default structuredOutputs to true.

So structured outputs migration is confusing on multiple dimensions: responses vs chat models, property renames, different property defaults... _and_ the new default API is now responses instead of chat when calling the plain `openai()` function. Massive 🤯.

I think the migration doc should also emphasize that this _will_ break structured outputs and to preserve v4 behavior, `strictJsonSchema` should be set to `true`. In fact, I'm not quite sure what the use case for `strictJsonSchema: false` even is in the context of the AI SDK, since the Zod schema validation will be strict _regardless_, and it will throw an error if the non-strict openai response doesn't match, which is likely.

AndrewPrifer:
This PR by @lgrammel contains the changes related to schema strictness: https://github.com/vercel/ai/pull/6925

AtAFork:
Agreed with @AndrewPrifer and seems like this should be a high priority fix to the docs @nicoalbanese, it was a big footgun for me as well upon upgrading. A few things to change in the migration docs (echoing Andrew):

- Mentioning the change in default API from `Responses` to `Chat`, which means the examples for `providerOptions` should show the type `OpenAIResponsesProviderOptions` and mention it's no longer `OpenAIChatLanguageModelOptions`
- Mentioning that `strictJsonSchema` defaults to `false`, so to keep v4 behavior you have to explicitly set `strictJsonSchema` to `true`
```
    providerOptions: {
      openai: {
        strictJsonSchema: true
      } satisfies OpenAIResponsesProviderOptions
    }
```
- Making sure the migration docs are consistent with the openai specific docs pages too

AtAFork:
cc'ing @lgrammel too

nicoalbanese:
@AtAFork @AndrewPrifer thanks for flagging! Hopefully #9590 resolves. Sorry for the confusion!

PR #9148, opened 2025-10-02
feat (provider/anthropic): Add support for pdf files in tool result.

AVtheking:
## Background
Anthropic API supports handling of pdf files in the tool results which was not supported by AI SDK currently, so added the support for that.

## Summary
Added support for handling pdf files in tool response in anthropic provider.

## Manual Verification
Wrote a eg with a tool returning pdf file in tool result as base64 string.

## Checklist

## Related Issues
Fixes #8986


packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -274,6 +274,23 @@ export async function convertToAnthropicMessagesPrompt({
                             };
                           }
 
+                          if (contentPart.mediaType === 'application/pdf') {
+                            betas.add('pdfs-2024-09-25');
```
i-m-sid, line 282: we should let client handle this.

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -274,6 +274,23 @@ export async function convertToAnthropicMessagesPrompt({
...
+
+                            return {
+                              type: 'document',
+                              source: {
+                                type: 'base64',
+                                media_type: contentPart.mediaType,
+                                data:
+                                  typeof contentPart.data === 'string'
+                                    ? contentPart.data
```
i-m-sid: let's directly use contentPart.data similar to image

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
AVtheking: this betas thing ? it was used like this in the message content entered by user too, I followed the same pattern.

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
gr2m: agreed. Is there a reason you added the check? I'd assue it should work without it since we do the same for image

AVtheking: just added that incase provided data is not string but yeah make sense to use that way like image

gr2m (approved):
Thank you for another great PR Ankit 🥇

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9406

PR #9191, opened 2025-10-04
feat(mcp): Add an optional fetch property to MCPTransportConfig

jobrk:
## Background

When running some unit tests, I can provide a custom fetch function for the providers I'm using but not for the MCP client. (And I don't want to move my project and install [`@modelcontextprotocol/sdk` just for this](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/sse.ts#L54))

## Summary

Adding `fetch` as an optional property on `MCPTransportConfig` and the logic / tests necessary.

## Manual Verification

Not manually verified apart from unit tests.

## Checklist

### DIFF
--- a/.changeset/forty-files-approve.md
+++ b/.changeset/forty-files-approve.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/mcp': patch
+---
+
+Add an optional fetch property to MCPTransportConfig
--- a/content/docs/07-reference/01-ai-sdk-core/23-create-mcp-client.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/23-create-mcp-client.mdx
@@ -108,6 +108,13 @@ It currently does not support accepting notifications from an MCP server, and cu
                       description:
                         'Optional OAuth provider for authorization to access protected remote MCP servers.',
                     },
+                    {
+                      name: 'fetch',
+                      type: 'FetchFunction',
+                      isOptional: true,
+                      description:
+                        'Custom fetch implementation. You can use it as a middleware to intercept requests, or to provide a custom fetch implementation for e.g. testing.',
+                    },
                   ],
                 },
               ],
--- a/examples/mcp/README.md
+++ b/examples/mcp/README.md
@@ -43,6 +43,7 @@ Available examples/folders:
 - `elicitation` - MCP elicitation example
 - `elicitation-multi-step` - MCP multi-step elicitation example
 - `elicitation-ui` - MCP elicitation with UI (server only)
+- `sse-custom-fetch` - SSE Transport with custom fetch (Legacy)
 
 Example usage:
 
@@ -64,3 +65,17 @@ pnpm server:elicitation-ui
 ```
 
 and then start the dev server in a new terminal in `examples/next-openai` and navigate to `localhost:3000/mcp-elicitation`
+
+## SSE Transport with custom fetch (Legacy)
+
+Start server
+
+```sh
+pnpm server:sse
+```
+
+Run example:
+
+```sh
+pnpm client:sse-custom-fetch
+```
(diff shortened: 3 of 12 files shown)


packages/ai/src/tool/mcp/mcp-sse-transport.ts
```diff
@@ -17,6 +21,7 @@ export class SseMCPTransport implements MCPTransport {
     close: () => void;
   };
   private headers?: Record<string, string>;
+  private fetch: FetchFunction;
```
aayush-kapoor: could you make this optional as well? just for uniformity

aayush-kapoor:
Hi @jobrk, PR looks good but could you attach a small code example where you'd need to implement a custom fetch function? you can do so under the `/example/mcp` folder (feel free to create a new sub-directory) so that we can test it end-to-end. thanks!

packages/ai/src/tool/mcp/mcp-sse-transport.ts
jobrk: i don't think this should be optional given it always is set to something in the constructor, this would lead to unnecessary null check whenever the fetch property is being used down in the class methods

jobrk:
Added an example for testing @aayush-kapoor

aayush-kapoor (commented):
hey, the MCP components have been refactored into their own package. please merge the latest changes from main and resolve conflicts. let me know if you need any help with  that or if you have any questions

jobrk:
> **Additional Suggestion:**
> 
> The HttpMCPTransport constructor doesn't accept the `fetch` parameter from MCPTransportConfig, even though SseMCPTransport does. This creates an inconsistent API where the custom fetch feature only works with SSE transport.
> View Details
> 
> 📝 Patch Details
> 
> ### Analysis
> ## HttpMCPTransport doesn't accept custom fetch parameter from MCPTransportConfig
> 
> **What fails:** `HttpMCPTransport` constructor ignores the `fetch` parameter from `MCPTransportConfig`, creating an inconsistent API where custom fetch implementations only work with `SseMCPTransport`. Direct calls to the global `fetch` function on lines 110, 137, and 304 bypass any custom implementation.
> 
> **How to reproduce:**
> 
> ```ts
> const mockFetch = vi.fn(fetch);
> const transport = new HttpMCPTransport({
>   url: 'http://example.com/mcp',
>   fetch: mockFetch,  // This parameter was being ignored
> });
> await transport.start();  // mockFetch never gets called - global fetch is used instead
> ```
> 
> **Result:** The custom fetch function is silently discarded and never invoked. All HTTP requests use the global `fetch` directly.
> 
> **Expected:** Following the pattern of `SseMCPTransport`, `HttpMCPTransport` should accept and use the `fetch` parameter, allowing middleware and testing implementations to intercept HTTP calls.
> 
> **Fix implemented:**
> 
>     * Added `FetchFunction` import from `@ai-sdk/provider-utils`
> 
>     * Added `fetch?: FetchFunction` parameter to constructor
> 
>     * Stored it as private field `private fetch?: FetchFunction`
> 
>     * Implemented `private getFetch()` method matching `SseMCPTransport` pattern
> 
>     * Updated all three direct `fetch()` calls to use `this.getFetch()` instead
> 
>     * Added test case verifying custom fetch is invoked during start(), send(), and close() operations
> 
> 
> [ Fix on Vercel](https://vercel.com/vercel-labs/~/agent/tasks/a86adebc-e56f-4720-8f0a-99d4f798f100)

will fix

jobrk:
> **Additional Suggestion:**
> 
> The HttpMCPTransport constructor doesn't accept the `fetch` parameter from MCPTransportConfig, even though SseMCPTransport does. This creates an inconsistent API where the custom fetch feature only works with SSE transport.
> View Details
> 
> 
> [ Fix on Vercel](https://vercel.com/vercel-labs/~/agent/tasks/a86adebc-e56f-4720-8f0a-99d4f798f100)

fixed

PR #9274, opened 2025-10-08
fix(provider/openai): Fix handling of successful and failed web search tool call results

### LINKED ISSUE #9270: OpenAI: Both completed and failed web tool call results are enqueued as `tool-result` chunk
### Description

When web search results are returned, AI SDK enqueues either a `tool-result` or `tool-error` chunk based on whether web search was successful or not. However, the OpenAI implementation in AI SDK doesn't check if the web search was successful or not and always enqueues a `tool-result` chunk. See:

https://github.com/vercel/ai/blob/418e18f15129e4615f5200c64c535c40b688cff6/packages/openai/src/responses/openai-responses-language-model.ts#L1011-L1034

A fix would be to check the `status` field (described briefly in the example in OpenAI [docs](https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses#output-and-citations)) and enqueue a `tool-result` or `error` based on the value of the status field.

Also another issue I noticed there isn't a guideline on what should be included in the input/output of a web search tool call result, so AI SDK seems to currently set the input to the action returned in the web search result (which could be of various types, like 'search', 'find', etc and may include query and sources), so it feels slightly incorrect to include the whole thing into input and set output to just the status. This is also inconsistent with Anthropic's implementation. I can create another issue discussing this in detail if you'd like.

### AI SDK Version

- @ai-sdk/openai: 2.0.44
- ai: 5.0.60

### Code of Conduct

nimeshnayaju:
Fixes https://github.com/vercel/ai/issues/9270

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.ts
@@ -40,7 +40,7 @@ import { localShellInputSchema } from '../tool/local-shell';
 const webSearchCallItem = z.object({
   type: z.literal('web_search_call'),
   id: z.string(),
-  status: z.string(),
+  status: z.enum(['completed', 'failed']),
   action: z
     .discriminatedUnion('type', [
       z.object({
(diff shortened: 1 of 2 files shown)


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -40,7 +40,7 @@ import { localShellInputSchema } from '../tool/local-shell';
 const webSearchCallItem = z.object({
   type: z.literal('web_search_call'),
   id: z.string(),
-  status: z.string(),
+  status: z.enum(['completed', 'failed']),
```
nimeshnayaju, line 43: According to their [node package](https://github.com/openai/openai-node/blob/c8ee824c536ec36d2c554299275f9694040de67c/src/resources/responses/responses.ts#L2003), which is based on their OpenAPI specs, status can be either `in-progress`, `searching`, `completed` or `failed`. 

The `response.output_item.done` event with item of type `web_search_call` comes after  `response.web_search_call.completed` event, so I'm assuming that it's safe to believe that the status at the end of the tool call can be either `failed` or `completed`, but I'd love for someone to also check this.

I also noticed that this Zod schema is used to validate the message returned by `doGenerate`, so please check if it interferes with that implementation.

aayush-kapoor:
@nimeshnayaju could you please add tests to ensure this change will work as intended

aayush-kapoor:
also wondering if this was an issue you experienced or just noticed as a logic mismatch?

gr2m:
Closing after no follow-up on requested reproduction and tests. Bug stays tracked in #9270. See our [updated contribution guidelines](https://github.com/vercel/ai/issues/17243). Thanks!

PR #9291, opened 2025-10-08
feat(openai): add 'default' as service tier option

### LINKED ISSUE #9289: OpenAI default service tier validation error
### Description

https://platform.openai.com/docs/api-reference/responses/create#responses-create-service_tier

https://github.com/vercel/ai/blob/ec64c2793f94d00a229ee3e70aec3d497a22a670/packages/openai/src/chat/openai-chat-options.ts#L105-L113

`default` is not supported.

### AI SDK Version

2025/10/08 latest

### Code of Conduct

EurFelux:
## Background

OpenAI added a `default` option to the `service_tier` parameter. See #9289 

## Summary

Add 'default' service tier to enum options and update documentation

## Manual Verification

Add the `default` to `z.enum` so it could pass validation.

## Related Issues

Fixes #9289


packages/openai/src/responses/openai-responses-api.ts
```diff
@@ -669,7 +669,7 @@ export const openaiResponsesResponseSchema = lazyValidator(() =>
           }),
         ]),
       ),
-      service_tier: z.string().nullish(),
+      service_tier: z.enum(['auto', 'flex', 'priority', 'default']).nullish(),
```
EurFelux: I'm not sure if we should keep it loose here.

lgrammel: yes we should. otherwise it will lead to breakages when openai adds more items

EurFelux: OK

lgrammel (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9321

Issue #9272, opened 2025-10-08
OpenAI: Web search tool call sources are not included in tool result chunk


nimeshnayaju:
### Description

When a web call tool result is received, AI SDK currently enqueues a `tool-result` chunk (irrespective of success or failure - https://github.com/vercel/ai/issues/9270). The input field of this `tool-result` chunk is set to the action returned in the tool call result. However, `sources` property is not defined in the Zod schema of the `search` action type. See:

https://github.com/vercel/ai/blob/418e18f15129e4615f5200c64c535c40b688cff6/packages/openai/src/responses/openai-responses-language-model.ts#L1011-L1025

https://github.com/vercel/ai/blob/418e18f15129e4615f5200c64c535c40b688cff6/packages/openai/src/responses/openai-responses-language-model.ts#L40-L49

So, sources array is completely missed.

### AI SDK Version

- @ai-sdk/openai: 2.0.44
- ai: 5.0.60

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nimeshnayaju:
Maybe we could take this opportunity to discuss what should go in the `input` and `output` of a tool-call result for provider executed tool calls like web search.

Currently, for Anthropic, it's quite straightforward - input is an object with property query representing the web search query

```ts
{ query: string }
```

And, success result is an array of sources that were visited as part of the web search call.

https://github.com/vercel/ai/blob/ec64c2793f94d00a229ee3e70aec3d497a22a670/packages/anthropic/src/anthropic-messages-language-model.ts#L891-L904

Similarly, error result is an object that includes the error reason

https://github.com/vercel/ai/blob/ec64c2793f94d00a229ee3e70aec3d497a22a670/packages/anthropic/src/anthropic-messages-language-model.ts#L921-L931

It is not so formulated for OpenAI web search: the input field of this tool-result chunk is set to the action returned in the tool call result and output is an object containing the status of the tool call. This action object is a discriminated union and can either be a search action or page open action or find action, depending on the model (more in the OpenAI's [docs](https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses#output-and-citations) and OpenAI's [node repository](https://github.com/openai/openai-node/blob/c8ee824c536ec36d2c554299275f9694040de67c/src/resources/responses/responses.ts#L2032-L2083)).

```ts
{
    type: "search";
    query?: string;
    sources?: Array<{ type: "url", url: string }>;
} | {
    type: "open_page";
    url: string;
} | {
[...]

PR #9299, opened 2025-10-08
fix(provider/openai): Include sources field in search action during `web_search_call`

nimeshnayaju:
Addresses https://github.com/vercel/ai/issues/9272


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -46,6 +46,7 @@ const webSearchCallItem = z.object({
       z.object({
         type: z.literal('search'),
         query: z.string().nullish(),
+        sources: z.array(z.looseObject({})).nullish(),
```
nimeshnayaju: We could further validate each source in the `sources` array. OpenAI currently supports `url` and `api` sources. I obtained it from their [docs](https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses#sources) and their node [types](https://github.com/openai/openai-node/blob/c8ee824c536ec36d2c554299275f9694040de67c/src/resources/responses/responses.ts#L2036-L2046) although the node types doesn't list 'api' as a source.

```ts
{ type: "url", url: string } | { type: "api", name: 'oai-weather' | 'oai-sports' | 'oai-finance' }
```

The reason I didn't add a `discriminatedUnion` validation on sources is because OpenAI may introduce to add a new source type in future, and validation would fail because the type isn't recognized, which is not what we want.

lgrammel:
They are not part of the result because they are already part of the input and we dont want to send information twice.

nimeshnayaju:
> They are not part of the result because they are already part of the input and we dont want to send information twice.

Maybe I am missing something, but I see in the code that the result is simply set to the tool call status and input is set to action and the action schema currently strips away the sources array in `query` sources.

https://github.com/vercel/ai/blob/5366d009388147a90a5662250dd430ccd9b67cfb/packages/openai/src/responses/openai-responses-language-model.ts#L675-L681

packages/openai/src/responses/openai-responses-language-model.ts
nimeshnayaju: Maybe, in future, it'll make sense to refine it further and enqueue `source` chunks by iterating over this sources array (like we currently do for Anthropic's web search results), but that is out of scope for this PR.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1016,19 +1017,84 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV3 {
...
+                      },
+                    },
+                    output: { status: value.item.status },
+                  };
+                } else {
+                  result = {
+                    input: { action: undefined },
+                    output: { status: value.item.status },
+                  };
+                }
```
nimeshnayaju, line 913: With this change, if the action type is `search`, we keep the input and output as it previously was, except if the action type is `search`, we also include a `sources` field in the output object. This maintains backward compatibility and is consistent with Anthropic's implementation too.

lgrammel:
I've separated web search inputs and outputs in #9440 - adding source to the output should be straightforward now but this pr would need to be redone on top of 9440

PR #9411, opened 2025-10-11
ci(bundle-size): initial version

gr2m:
This is just a trivial check to start. It generates two bundles 

1. with condition "node"
2. with condition "browser"

I was actually surprised to see that the browser build is _bigger_ than the node build. It seems mostly be caused by the open telemetry dependency.

and it generates meta files that can be uploaded to https://esbuild.github.io/analyze/ for analysis. They are uploaded as artifacts to the CI check run.

The max size is currently configured as repository variable: https://github.com/vercel/ai/settings/variables/actions/BUNDLE_SIZE_LIMIT_KB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,7 @@ jobs:
         run: cd packages/ai && pnpm run check-bundle-size ${{ vars.BUNDLE_SIZE_LIMIT_KB }}
 
       - name: Upload bundle size metafiles
+        if: ${{ always() }}
         uses: actions/upload-artifact@v4
         with:
           name: bundle-size-metafiles
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
         run: pnpm run build:packages
 
       - name: Check bundle size
-        run: cd packages/ai && pnpm run check-bundle-size ${{ vars.BUNDLE_SIZE_LIMIT_KB }}
+        run: cd packages/ai && pnpm run check-bundle-size
 
       - name: Upload bundle size metafiles
         if: ${{ always() }}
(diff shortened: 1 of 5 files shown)


ghost (dismissed):

gr2m:
Bundle size meta files can be downloaded here:
https://github.com/vercel/ai/actions/runs/18425412417/artifacts/4243615987
(you can also download them from the over CI build overview page: https://github.com/vercel/ai/actions/runs/18425412417?pr=9411)

The JSON files can be uploaded here for analysis: https://esbuild.github.io/analyze/

.github/workflows/ci.yml
```diff
@@ -103,6 +107,40 @@ jobs:
...
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build packages
+        run: pnpm run build:packages
+
+      - name: Check bundle size
+        run: cd packages/ai && pnpm run check-bundle-size ${{ vars.BUNDLE_SIZE_LIMIT_KB }}
```
lgrammel: alternative would be to have the limit hardcoded in our code somewhere so we can track it in version control and update it in pull requests that affect the bundle size

lgrammel (approved):

.github/workflows/ci.yml
gr2m: ah yes, good point about having it in version control next to the changes that caused an increase / decrease. I like it!

gr2m:
Some interesting insights

### 60% is vendor code

image

### Dependency sizes

internal vs all external 

image

external

image

so yeah zod is currently ~~over 80%~~ nearly 60% of our bundle size. Not pointing at `zod`, Colin recommends to use zod 4 mini which we couldn't do yet since we want to support both v3 and v4. But we consider dropping `zod@3` as part of the upcoming `ai@6`

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/9419

gr2m:
Maybe one more insight: 

image

PR #9412, opened 2025-10-11
feat(provider/openai): automatically add reasoning.encrypted_content include when store = false

lgrammel:
## Background

Currently users need to specify the `reasoning.encrypted_content` include on the OpenAI provider options when they want reasoning with `store = false`. However, it only works with reasoning models, impeding model switching, and also requires users to look up that include, leading to worse DX.

## Summary

Automatically add `reasoning.encrypted_content` include when `store = false` with reasoning models.

## Manual Verification

- [x] run `examples/ai-core/src/stream-text/openai-chatbot.ts` with reasoning model
- [x] run `examples/ai-core/src/stream-text/openai-chatbot.ts` with non-reasoning model


nicoalbanese (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9414

PR #9413, opened 2025-10-11
feat(provider/openai): send assistant text and tool call parts as reference ids when store: true

lgrammel:
## Background

When using the OpenAI Responses API, assistant text and tool call parts are currently send with both content and ids when `store: true`. However, the content is redundant (stored by OpenAI) and the payload size can be reduced by sending item references instead.

## Summary

Send item references instead of regular parts for OpenAI responses assistant text and tool call parts when store: true and id present.

## Manual Verification

- [x] test web search ui example with reasoning and store: true (multi-step)
- [x] test web search ui example with reasoning and store: false (multi-step)


nicoalbanese (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9424

PR #9421, opened 2025-10-11
feat(gemini, anthropic): export message conversion utilities

### LINKED ISSUE #9391: Export `convert-to-anthropic-messages-prompt.ts` and `convert-to-google-generative-ai-messages.ts`
### Description

There are useful utilities in `vercel/ai` that convert a `ModelMessage` into a format that Anthropic API or Google Generative AI API expects. 

For our use-case, we're using `vercel/ai` to make inference calls to LLMs in a model-agnostic way. We'd also like to use Anthropic and Google's `count_tokens` API before making the call to make sure we don't exceed the context window

[1] Claude - https://docs.claude.com/en/docs/build-with-claude/token-counting
[2] Gemini - https://ai.google.dev/gemini-api/docs/tokens?lang=python

Converting from `ModelMessage` to these formats is annoying and it would be great to re-use the rails that already exist in this library. Can we export these utilities for developer use?

### AI SDK Version

- ai: ^5.0.45

### Code of Conduct

valenradovich:
## Background
Users need to call token-counting APIs (Gemini’s `count_tokens`, Claude’s `count_tokens`) before inference to check context window limits and estimate costs. Conversion functions to transform AI SDK messages into provider-specific formats already exist internally but are not exported.

## Summary
This PR exports the message conversion utilities for both Google and Anthropic:

- `convertToGoogleGenerativeAIMessages` from `@ai-sdk/google`
- `convertToAnthropicMessagesPrompt` from `@ai-sdk/anthropic`

And their corresponding type exports:
- `GoogleGenerativeAIPrompt`
- `AnthropicMessagesPrompt`

## Manual Verification
- ✅ Built both packages successfully
- ✅ Verified exports in type declarations (`packages/*/dist/index.d.ts`)
- ✅ All tests passing (Google: 153, Anthropic: 113)
- ✅ Documentation added with token-counting examples for both providers

## Checklist
- [x] Tests have been added/updated (tests already exist)
- [x] Documentation has been added/updated
- [x] A patch changeset for relevant packages has been added
- [x] Formatting issues fixed (`pnpm prettier-fix`)
- [x] Self-review completed

## Related Issues
Fixes #9391


lgrammel:
have you tested that you can convert a ModelMessage if you export those?

valenradovich:
good point @lgrammel - it will work with `LanguageModelV3Prompt` only. 

What do you recommend? I was thinking about a wrapper function that accept `ModelMessage`, calls the existing conversion functions and export these as the primary public API

lgrammel:
@valenradovich unfortunately it is not as straight forward afaik, bc translation also depends on e.g. provider options (see openai responses provider) and the system prompt (and how it is included). i understand the use case, trying to figure out an elegant solution here.

Brainstorming here
- Model classes would need to have an optional async function that supports converting to some sort of mode specific format. Alternatively this could be docoupled into standalone functions (maybe better)
- That conversion function would need to be standardized
- A common function in AI could take a prompt (incl. system and provider options) and a function and resolve this.

Something like

```ts
const apiMessages = convertToApiMessages({
  model: LanguageModelV3 | string, // or apiConverter: function(LanguageModelV3Prompt + providerOptions) => unknown
  prompt: string | ModelMessage[],
  system: string;
  providerOptions
})
```

Other options may be important as well though

SamStenner:
Would love this

Issue #7269, opened 2025-07-14
BUG: OpenTelemetry with @ai-sdk/azure is sending incorrect traces


christophmeise:
I have set up an app using @ai-sdk/azure and the default @vercel/otel OpenTelemetry configuration to send data to Braintrust. The setup is exactly as in the example docs for the Braintrust integration here: https://ai-sdk.dev/providers/observability/braintrust#opentelemetry

In Braintrust I can see the log traces from the endpoint and I can also see the LLM calls to Azure.

The problem: the traces are not correct, all LLM calls have duration, token usage and cost of 0.

I already spoke with @ankrgyl (Founder/CEO Braintrust) and he investigated the issue on their side.
**He confirmed that this has nothing to do with their code and recommended to create this issue here.**

The following image shows that there is a **fetch call inside of the LLM call, which I suspect as the underlying issue**.
The results are the same with streamObject and streamText.

<img width="1037" height="590" alt="Image" src="https://github.com/user-attachments/assets/4078ebc8-a0ab-42b9-91b0-2e480cef95ac" />

Compare this to a screenshot using the vercel openai package, where the LLM calls have not nested fetch, showing all data correctly:

<img width="1029" height="956" alt="Image" src="https://github.com/user-attachments/assets/9ae96eb3-4259-4cbe-bb85-6c8c43698660" />

christophmeise:
Hey @gr2m - it has been nearly half a year, is there any update or timeline on this?

mattschaller:
I ran into this exact attribute mapping mess — built [ai-sdk-otel-adapter](https://github.com/mattschaller/ai-sdk-otel-adapter) to solve it. It's a vendor-neutral OTel SpanProcessor that remaps Vercel AI SDK ai.* attributes to the gen_ai.* semantic conventions, so your backend (Jaeger, Honeycomb, etc.) sees clean, standard traces. Might unblock you: npm install ai-sdk-otel-adapter

khalidsaidi:
A2ABench has an accepted answer for this imported thread.
- Thread: https://a2abench-api.web.app/q/cmmqmwl1t006y9cs2670rqwqq
- Accepted at: 2026-03-25T19:51:13.974Z
- Accepted answer agent: `ai village`
- Answer preview: "No matching threads were found for this query."

lgrammel:
### Identify issue type

**Identified type: Bug**  
Confidence: High

[...]

PR #9497, opened 2025-10-14
feat(azure): decouple deployment id from reported model id

### LINKED ISSUE #7383: Feature: Add optional modelName param to Azure .languageModel() for accurate telemetry (e.g., fix Langfuse cost calculation)
### Description

When using the Vercel AI SDK with Azure OpenAI, the `.languageModel(deploymentName)` method requires passing the custom Azure deployment name (e.g., 'custom-deployment') for the API call to succeed, as Azure strictly validates this in the endpoint URL. However, this sets the OpenTelemetry span attribute `ai.model.id` to the deployment name, which is not a standard OpenAI model name (e.g., 'gpt-4o'). As a result, observability tools like Langfuse cannot match it to their pricing tables, leading to missing or zeroed-out cost calculations in traces, even though the LLM call works and token usage is ingested.

This is a known integration gap with Azure, as the SDK doesn't separate the deployment name (for API) from the model name (for telemetry). Libraries like LangChain/LangGraph handle this out-of-the-box with separate params: `azure_deployment` for the API and `model_name` for tokenizers/metadata/cost. Adding an optional `modelName` param to `.languageModel()` (e.g., `.languageModel(deploymentName, { modelName: 'gpt-4.1' })`) would allow using the deployment for the API while setting `ai.model.id` (or a new metadata field) to the model name for better telemetry compatibility.

**Reproduction Steps**  
1. Set up an Azure OpenAI resource with a deployment named 'custom-deployment' (underlying model: gpt-4o or similar).  
2. In code:  
   ```ts  
   import { createAzure } from '@ai-sdk/azure';  
   const azure = createAzure({ resourceName: 'your-resource', apiKey: 'your-key' });  
   const model = azure.languageModel('custom-deployment'); // Or 'gpt-4.1' to trigger the alternative behavior  
   // Use in a call, e.g., generateText({ model, prompt: 'Hello' })  
   ```  
3. Integrate with Langfuse (or similar) via OpenTelemetry.  
4. Observe traces: With deployment name, API succeeds but Langfuse cost is 0/missing. With model name, API errors but cost shows.  

**Expected Behavior**  
The SDK should support an optional param to specify the underlying model name for telemetry, decoupling it from the deployment name required by Azure. This could set `ai.model.id` to the model name or add it to metadata (e.g., `experimental_telemetry.metadata.actualModel`).  

**Related Issues**  
- #7269 (Telemetry bugs with Azure showing 0 cost/duration in observability tools)  
- #6378 (Expose deploymentName for tracing)

### AI SDK Version

**Environment**  
- Node.js version: 20.14.0 
- Azure API version: 2024-12-01-preview
- @ai-sdk/azure: 1.3.17
- ai: 4.3.9

rahulbhadja:
## Background
Azure deployments require the custom deployment name in the API URL, but telemetry and pricing tools (e.g. Langfuse) expect the underlying OpenAI model id. The SDK currently reports the deployment id, so Azure traces end up with missing data.

## Summary

- add optional `modelName` support to the Azure provider factories so telemetry can report the real OpenAI model id while still using the deployment id for API calls
- wrap the OpenAI models to expose the override
- added a new generate example 
- export the new settings type for azure

## Manual Verification

 - [x]  run generate text example `examples/ai-core/src/generate-text/azure-model-name.ts`

## Checklist

## Future Work
Update documentation to add modelName usage in Azure Provider.  

## Related Issues
Fixes #7383 and #7269


rahulbhadja:
Did a brainstorm on passing modelId instead of deploymentId in the Azure provider. Let me know if this change is in the correct direction, @gr2m. Please lmk if any changes are needed.

also tagging @tsuzaki430  and  `@vercel/ai-sdk-azure` for review.

packages/azure/src/azure-openai-provider.ts
```diff
@@ -25,59 +33,89 @@ import { azureOpenaiTools } from './azure-openai-tools';
...
+  ): ImageModelV3;
 
   /**
    * Creates an Azure OpenAI DALL-E model for image generation.
    */
-  imageModel(deploymentId: string): ImageModelV3;
+  imageModel(
+    deploymentId: string,
+    settings?: AzureOpenAIModelSettings,
+  ): ImageModelV3;
```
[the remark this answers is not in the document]
rahulbhadja: nice catch :)

tsuzaki430:
@rahulbhadja 

Thank you for let me know this PR.
I tried it and it work expectedly.
I think it’s good that the process works without affecting @ai-sdk/openai.
I also confirmed that it’s only used in the `get modelId()` process and is unrelated to the actual AI functionality.

I want test about new `modelName` properties.
How about like this test?
If you OK, please add this test in your PR .test.ts file?

``` typescript
    it('should return modelId from the deployment name exactly.', async () => {
      const modelId = provider('test-deployment').modelId;
      expect(modelId).toStrictEqual('test-deployment');
    });

    it('should return modelId from the deployment name when modelName is undefined.', async () => {
      const modelId = provider('test-deployment', {
        modelName: undefined,
      }).modelId;
      expect(modelId).toStrictEqual('test-deployment');
    });

    it('should return modelId from modelName when it is set.', async () => {
      const modelId = provider('test-deployment', {
        modelName: 'gpt-4.1-mini',
      }).modelId;
      expect(modelId).toStrictEqual('gpt-4.1-mini');
    });
```

And I hope `modelName` property will be added in the document.
https://ai-sdk.dev/providers/ai-sdk-providers/azure
https://github.com/vercel/ai/blob/main/content/providers/01-ai-sdk-providers/04-azure.mdx
( or Future Work after Merge ?)

rahulbhadja:
@tsuzaki430  , thanks for the review. Yes, I will add those tests; it was on my mind and I was waiting for a review from you or @gr2m .  For the docs part, let's wait until this is merged, and then I will add documentation regarding this. What do you think?

tsuzaki430:
I’ve confirmed that you’ve added the tests — thank you!
I’ve rechecked everything and it looks all good to me.

As for the documentation, I agree — it’s fine to add it after the merge.
Currently, the `Future Work` section is marked as N/A, but you might update it to mention that you plan to add documentation once this PR is merged.

It’ll be great to see your excellent proposal reflected in the docs so other Langfuse users can benefit from it! 🚀

rahulbhadja:
thanks @tsuzaki430  , I will update the future work in PR message.

examples/ai-core/src/generate-text/azure-model-name.ts
```diff
@@ -0,0 +1,23 @@
...
+import { generateText } from 'ai';
+import 'dotenv/config';
+
+async function main() {
+  const model = azure(
+    'gpt-nano-flex', // Replace 'gpt-nano-flex' with your actual deployment name
+    {
+      modelName: 'gpt-4.1-nano', // and specify the underlying model ex. for telemetry
+    },
+  );
```
gr2m, line 11: I fear this won't work because we need to be able to configure a model with only a string.

Could we maybe use a separator to configure both the deployment ID and model ID in the same string, e.g. like this

```suggestion
  const model = azure(
    // Replace 'gpt-nano-flex' with your actual deployment name
    // and optionally specify the underlying model e.g. for telemetry
    'gpt-nano-flex::gpt-4.1-nano',
  );
```

rahulbhadja: Thank you for the review, @gr2m. This is a good suggestion; it could work as well. What do you think about changes to the internal model ID? We will need to send the actual model for telementry for example this
typescript ```  get modelId(): string {
    return this._modelName ?? this.openaiModel.modelId;
  }```

gr2m: I'd make it more explicit. Since this is specific to Azure and deployments, I would call it something like `.azureDeploymentModelId`.

Und3rf10w: For _me_, _personally_, this approach would interfere with our current model logic.

We have a capability, built prior to the Provider Middleware introduced in I think v4?, that allows us to route logically across a single provider and model identifier. Think like `azure:gpt-4`, but behind the scenes it's pointed to `azure-us-east1:gpt-4`, `azure-us-west1:gpt-4`, etc.

I think with this there might be some confusion around how this would affect this kind of setup.

We've resorted currently to the user passing a fully constructed `${provider.providerName}:${model.modelName}` string in their request, which we parse with:

```typescript
interface ParsedModelString {
  providerName: string;
  modelName: string;
}

/**
 * Splits a combined provider and model ID into separate parts.
 *
 * @param {string} id - The combined provider and model ID
 * @returns {[string, string]} An array containing the provider ID and model ID
 * @throws {InvalidModelIdError} If the ID format is invalid
 */
export default function splitId(id: string): ParsedModelString {
  const index = id.indexOf(':');
  if (index === -1) {
    return {
      providerName: '',
      modelName: '',
    };
  }
  return {
    providerName: id.slice(0, index),
    modelName: id.slice(index + 1),
  };
}
```

Then use the `providerName` to key off of a mapping then use that to call the apporpriate `create${provider.name}` function (e.g. `createAzure()`).

Would the concept of an optional `LanguageModelMetadata` not suffice here? Something like a 

```typescript
model = createAzure({
  ...options, 
  headers: { 
    ...decryptedHeaders 
   }
})
  .languageModel(
   `${modelName}`, 
    { 
      providerOptions.azure.deploymentName = 'foo' 
    }
})

// ...

resourceNameForTracing = model.providerOptions?.azure?.deploymentName
```

could be set on model instantiation, and then read later by whatever your tracing setup call is?

@gr2m, I'd be happy to share our registry implementation with you on a private channel if it would help at all.

Realistically, the answer for us is probably _migrate to provider middleware_, but curious why we can't make a `LanguageModelMetadata` concept here.

gr2m: @Und3rf10w thanks for sharing your perspective!

@rahulbhadja hold off making changes for now, I'll give this some more thought

christophmeise: @rahulbhadja @gr2m is there any update or progress on this? We would really like to see this fixed so that we can properly monitor our llm pricing with braintrust, but this is currently blocking it.

PR #9503, opened 2025-10-14
fix(provider/anthropic): add url-based PDF support in tool results

### LINKED ISSUE #8014: Unable to pass PDF files from the tool calls

nlaz:
## Background

This PR adds support for url-based PDFs in tool results. This is a pretty common way to fetch and insert PDF without blowing up the context window. Anthropic's API handles this [out of the box](https://docs.claude.com/en/docs/build-with-claude/pdf-support). Just needed to update the types on the tool result handling.

## Summary

This PR adds support for url-based PDF support without changing the language model spec. It checks if the data passed back from the tool is a valid url and if so converts the tool result block into the correct PDF content block.

This is a new PR instead of this previous PR with spec changes: https://github.com/vercel/ai/pull/9041

It also built off of work from this PR that supports base64 PDFs: https://github.com/vercel/ai/pull/9148

## Manual Verification

Added an example file that shows that the changes work.

## Checklist

## Related Issues

Related: https://github.com/vercel/ai/pull/9041
Fixes https://github.com/vercel/ai/issues/8014


examples/ai-core/src/generate-text/anthropic-pdf-tool-results-base64.ts
nlaz, line 1: Differentiated between the base64 example and the url example

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -46,6 +46,15 @@ function convertToString(data: LanguageModelV3DataContent): string {
   });
 }
 
+function isValidUrl(string: string): boolean {
```
nlaz, line 49: Necessary to avoid changing the spec

gr2m:
Already implemented in #9733.

PR #9514, opened 2025-10-15
build(tools): use stats API fro `analyze-providers.ts` script

gr2m:


gr2m:
added some more data 
image

ghost (approved):

PR #9519, opened 2025-10-15
fix(ui): do not submit automatically when server return with error

lgrammel:
## Background

When `sendAutomaticallyWhen` is configured and the server responds with an error, an infinite client-server loop is triggered (see #9384 ).

## Summary

Only resubmit automatically when there is no error.

## Manual Verification

Fix manually verified as part of #9518

## Related Issues

Fixes #9384


ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9520

PR #9671, opened 2025-10-20
chore(agent): limit agent call parameters

lgrammel:
## Background

It was possible to call agents with a system parameter. This should not be possible. The instructions / system prompt of the agent should be controlled separately, some agent implementations may not even support this.

## Summary

Limit agent call parameters to messages and prompt.

### DIFF
--- a/.changeset/breezy-masks-occur.md
+++ b/.changeset/breezy-masks-occur.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+chore(agent): limit agent call parameters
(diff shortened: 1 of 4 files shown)


gr2m (approved):
do we have type-only tests that we could utilize here to make sure we don't accidentally regress to allow for a `system` key again? Unlikely, but if it's easy to add why not

nicoalbanese (approved):

lgrammel:
@gr2m good call, adding tests caught a bug

PR #9719, opened 2025-10-22
feat(provider/openai): support openai chat model's reasoning_content

DeJeune:
## Background

For non-official OpenAI models deployed in Azure OpenAI (such as DeepSeek R1), the `packages/openai` is used, so the `reasoning_content` field needs to be introduced in the responseSchema.

## Summary

Added the `reasoning_content` field, updated relevant tests, and aligned with the `openai-compatible` logic.

## Checklist

### DIFF
--- a/.changeset/eight-hounds-divide.md
+++ b/.changeset/eight-hounds-divide.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+support openai-chat-language-model's reasoning_content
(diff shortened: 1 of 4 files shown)


lgrammel:
Is it possible to use openai compatible instead?

DeJeune:
> Is it possible to use openai compatible instead?

`azure-openai-provider` does not extends from `openai-compatible`.

DeJeune:
Are there any plans to make the provider more robust in the future? For example, `openai-response-compatible`.
Related documentation: https://docs.x.ai/docs/api-reference#create-new-response.

lgrammel:
@DeJeune I see - then ideally we need to make changes such that the Azure provider is more flexible. The OpenAI provider should be optimized for OpenAIs models and API without having to make tradeoffs for compatible providers. We would need to see how that impacts Azure. CC @gr2m

gr2m:

PR #9720, opened 2025-10-22
feat(ai): add convertDataPart option to convertToModelMessages

### LINKED ISSUE #9617: Support for Custom Context Attachments in UserMessage
### Description

There currently isn’t a straightforward way to store and rehydrate additional context attachments within a UserMessage in the AI SDK.
This becomes relevant when building advanced chat UIs (e.g., similar to [Cursor’s `<PromptInput />` component](https://ai-sdk.dev/elements/components/prompt-input#cursor-style)), where a message may include not only text or files, but also structured context such as URLs that the system needs to fetch, JSON data, or tool toggles (like enabling webSearch for a specific prompt).

Existing mechanisms (FilePart, custom data parts, Message.metadata) only partially address this and don’t provide a consistent way to persist, rehydrate, and send non-file or text attachments.

**Use Case**

Image

In chat applications where users can attach extra context to a message — such as files, URLs, structured JSON, or tool toggles - it would be helpful if these attachments could be:

- Stored alongside the user message
- Displayed again when editing that message, and
- Optionally included when converting messages to model input.

At the moment:

- Custom parts are stripped out by `convertToModelMessages()`.
- metadata is not sent to the model and is mainly used for auxiliary data like timestamps or model info.
- Using `prepareStep()` to call a custom tool (e.g., `getContext(attachmentIds)`) results in that fetched context being stored in the assistant message rather than the user message, meaning those attachments cannot later be removed or modified.

**What's would be helpful**

It would be valuable to have a more flexible pattern to:

1. Persist arbitrary contextual data alongside UserMessage objects.
2. Rehydrate and display those attachments when editing messages.
3. Optionally include this data in the conversion process for model messages.
4. Achieve this without manually overriding internal conversion utilities like `converToModelMessages()`

**Potential solution options**

- Extending `Message.parts` to accept additional part types beyond text and file (e.g., data, tool, or url).
- Allowing custom serialization logic to map rich `UserMessage` structures to `ModelMessage`.
- Introducing middleware or hooks that can modify message payloads before conversion while retaining the ability to rehydrate those parts.
- Providing a mechanism to selectively include certain metadata fields in model messages when appropriate.

**References**
- https://github.com/vercel/ai/discussions/2820#discussioncomment-14616465
- https://ai-sdk.dev/docs/reference/ai-sdk-core/model-message
- https://ai-sdk.dev/docs/ai-sdk-ui/message-metadata#common-use-cases

### AI SDK Version

- [ai@5.0.76](https://github.com/vercel/ai/releases/tag/ai%405.0.76)

### Code of Conduct

drew-foxall:
## Summary
Adds optional `convertDataPart` callback to `convertToModelMessages` for converting custom data parts (URLs, code files, etc.) into text or file parts that models can process.

## Changes
- Add `convertDataPart` option to `convertToModelMessages` function
- Add `isTextUIPart` and `isFileUIPart` to `ui-messages.ts`

## Testing
- 8 new test cases covering all conversion scenarios
- All 46 tests passing in `convert-to-model-messages.test.ts`

## Documentation
- Updated API reference with examples
- Includes usage examples for URL content, code files, and selective inclusion
- Type safety guidance included

## Related Issues
Closes #9617


packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -76,34 +91,40 @@ export function convertToModelMessages(
...
+                  ...(part.providerMetadata != null
+                    ? { providerOptions: part.providerMetadata }
+                    : {}),
+                };
+              }
+
+              // Process data parts with converter if provided
+              if (isDataUIPart(part) && options?.convertDataPart) {
+                return options.convertDataPart(part);
+              }
```
lgrammel: small change:

```ts
if (isDataUIPart(part)) {
  return options?.convertDataPart(part) ?? null;
}
```

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -291,10 +312,11 @@ export function convertToModelMessages(
               part.type === 'file' ||
               isToolOrDynamicToolUIPart(part)
             ) {
-              block.push(part);
+              block.push(part as any); // Type assertion needed due to union type complexity
```
lgrammel: this is a bit of an orange flag re type safety, it should not happen.

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -76,34 +91,40 @@ export function convertToModelMessages(
         modelMessages.push({
           role: 'user',
           content: message.parts
-            .filter(
-              (part): part is TextUIPart | FileUIPart =>
-                part.type === 'text' || part.type === 'file',
-            )
```
lgrammel, line 82: is it possible to keep the filter and add data parts to it? it was done for type safety reasons

packages/ai/src/ui/index.ts
```diff
@@ -34,6 +34,8 @@ export {
   getToolName,
   getToolOrDynamicToolName,
   isDataUIPart,
+  isFileUIPart,
+  isTextUIPart,
```
lgrammel, line 39: why is this needed?

packages/ai/src/ui/ui-messages.ts
```diff
@@ -211,6 +211,18 @@ export function isDataUIPart<DATA_TYPES extends UIDataTypes>(
...
+  part: UIMessagePart<UIDataTypes, UITools>,
+): part is TextUIPart {
+  return part.type === 'text';
+}
+
+export function isFileUIPart(
+  part: UIMessagePart<UIDataTypes, UITools>,
+): part is FileUIPart {
+  return part.type === 'file';
+}
```
lgrammel: why are these needed?

lgrammel: i'd much rather not have functions for something this trivial

content/docs/07-reference/02-ai-sdk-ui/31-convert-to-model-messages.mdx
```diff
@@ -87,3 +87,131 @@ const result = streamText({
...
+        url: 'https://example.com/article',
+        title: 'Important Article',
+        content: '...',
+      },
+    },
+  ],
+});
+```
+
+**Including Code Files as Context**
```
tobiasbueschel, line 165: Love this, thanks for putting this together  @Drew-Garratt 👏 

Would it be worth to include an example here for toggling on custom tools that a user might turn on when sending the user message? In this way, it can be stored alongside the other user message parts inside the DB and one can easily restore this context when the user wants to edit the message later on.

packages/ai/src/ui/convert-to-model-messages.ts
drew-foxall: I did account for that using a slightly different method. The `isTextUIPart` helper narrows the type allowing is to handle it as two seperate if's without filtering to only these two part types

```
export function isTextUIPart(
  part: UIMessagePart<UIDataTypes, UITools>,
): part is TextUIPart {
  return part.type === 'text';
}

export function isFileUIPart(
  part: UIMessagePart<UIDataTypes, UITools>,
): part is FileUIPart {
  return part.type === 'file';
}
```

I agree their functionality is trivial, the intention was that the signatures in this function were all clear and from the same location. So that `ui-messages` dictated what types of message are parsed.

Happy to move the functions here directly though if the extraction feels to verbose.

packages/ai/src/ui/convert-to-model-messages.ts
drew-foxall: Agree I'll simplify ^-^

packages/ai/src/ui/convert-to-model-messages.ts
drew-foxall: Absolutley and not intentional or needed, removed.

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -1,39 +1,79 @@
...
+
+/**
+ * Type guard to check if a message part is a reasoning part.
+ */
+function isReasoningUIPart(
+  part: UIMessagePart<UIDataTypes, UITools>,
+): part is ReasoningUIPart {
+  return part.type === 'reasoning';
+}
+
```
lgrammel: this is trivial code. pls inline, extracting the function is not worth it and hides type inference issues

packages/ai/src/ui/convert-to-model-messages.ts
lgrammel: I am not a fan of extracting trivial functions and putting type inference on them unless absolutely needed. It hides what is going on and can lead to future issues when the types become more complex. Overall I like this PR a lot, will see if I can find a better way of doing this.

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -1,39 +1,79 @@
...
 @param options.ignoreIncompleteToolCalls - Whether to ignore incomplete tool calls. Default is `false`.
+@param options.convertDataPart - Optional function to convert data parts to text or file parts for the model.
 
 @returns An array of ModelMessages.
  */
-export function convertToModelMessages(
-  messages: Array<Omit<UIMessage, 'id'>>,
+export function convertToModelMessages<
+  DATA_PARTS extends UIDataTypes = UIDataTypes,
+  TOOLS extends UITools = UITools,
```
lgrammel: should be derived from the UI message type instead (make that generic)

lgrammel:
Thanks for the PR! i like where it is headed but some details need to be changed. I'll see if I can make the changes today on this branch.

packages/ai/src/ui/convert-to-model-messages.ts
lgrammel: nevermind, after exploring simpler solutions i ran into inference issues due to the template types

packages/ai/src/ui/convert-to-model-messages.ts
lgrammel: after exploring other solutions, i agree this is the best way here

lgrammel:
Changed the generics, simplified the tests, changed return from null to undefined, exposed is* helpers again, supported assistant data parts.

lgrammel (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/9748

PR #9723, opened 2025-10-22
feat(provider/openai-compatible): OpenAI Compatible Response support

DeJeune:
## Background

more and more provider supports `v1/responses` endpoint:

https://openrouter.ai/docs/api-reference/responses-api/overview
https://docs.x.ai/docs/api-reference#create-new-response

ToDo List:

- [x] getArgs()
- [x] doGenerate()
- [x] doStream()
- [x] unitTest
- [x] examples
- [x] docs

## Checklist


DeJeune:
I'm not sure whether some parts of the `OpenAIResponseSchema` should be dropped or be more flexible under "compatible"; perhaps @lgrammel needs to confirm this.

lgrammel:
@DeJeune ideally it should be minimal to support the core set of what is available in the responses apis compatible providers. I think tests will be extremely critical. please adopt the fixture/snapshot model from openai responses and have fixtures for different providers.

fixtures can be created as follows:
- for generate: take result.request.body from example and make it the fixture
- for stream: enable includeRawChunks: true on the streamText call, and then use saveRawChunks to save the chunks (will be in output dir), all in examples

ideally we have fully realistic test input fixtures for various compatible providers, otherwise it will be hard to make this model stable and solid

PR #9750, opened 2025-10-23
Allow multiple events per JSON stream chunk when streaming response

charypar:
Previously, `createJsonStreamResponseHandler` assumed each steam chunk will contain at most one JSON event. Some streaming backends can buffer events if they are generated faster than they are consumed, resulting in multiple lines per chunk. This adds support for such streams

## Background

Previously, `createJsonStreamResponseHandler` assumed each steam chunk will contain at most one JSON event. Some streaming backends can buffer events if they are generated faster than they are consumed, resulting in multiple lines per chunk. Currently, such stream will cause a JSONParseError to be thrown

## Summary

The PR changes the processing of the stream to look for newlines within the chunks as well as at the ends, and should process any stream, regardless of how it's chunked up.

## Manual Verification

I've added a test which generates a stream with the failing format, the test failed at first, with the fix it passed, alongside all other tests.

Happy to do further verification, we're likely to use a local patch before this is released.

## Checklist

### DIFF
--- a/.changeset/thick-glasses-develop.md
+++ b/.changeset/thick-glasses-develop.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+Fix processing of a Json stream with multiple events per chunk
--- a/packages/provider-utils/src/response-handler.test.ts
+++ b/packages/provider-utils/src/response-handler.test.ts
@@ -54,6 +54,38 @@ describe('createJsonStreamResponseHandler', () => {
       { success: true, value: { a: 1 }, rawValue: { a: 1 } },
     ]);
   });
+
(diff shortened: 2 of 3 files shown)


lgrammel:
`createJsonStreamResponseHandler` is used nowhere and should be removed instead.

providers should use ``createEventSourceResponseHandler` which handles this

lgrammel:
curious how you encountered this. was it a bug that you faced?

charypar:
@lgrammel ah, that's good to know, will switch to that instead! 

I have encountered it when talking to a FastAPI based streaming backend which we use to coordinate a number of 3rd party models an custom jobs. The easiers way to integrate it with the AI SDK was to treat it as a provider.

PR #9763, opened 2025-10-23
fix(google): properly extract JSON schema from tool.inputSchema for Vertex AI

### LINKED ISSUE #9761: Bug: Vertex AI strips tool parameters schema to empty properties
# Bug Report: Vertex AI function calling parameters schema stripped to empty properties

## Description

When using `tool()` with custom parameters (either via `jsonSchema()` or dynamically-generated Zod schemas) with the Google Vertex AI provider, the tool parameters schema is incorrectly serialized. All property definitions are stripped out, leaving only an empty `properties: {}` object in the request sent to Vertex AI.

This causes all custom tools to fail with the error:
```
Unable to submit request because `<tool_name>` functionDeclaration parameters schema should be of type OBJECT
```

## Environment

- **AI SDK Version**: `ai@4.1.0` (latest)
- **Provider**: `@ai-sdk/google-vertex@3.0.53`
- **Runtime**: Bun (also tested in Node.js)
- **Model**: `gemini-2.0-flash-exp`

### Minimal Test Case

```typescript
import { tool, streamText, jsonSchema } from 'ai';
import { vertex } from '@ai-sdk/google-vertex';

// Define a simple tool with parameters
const testTool = tool({
  description: 'Update a workflow step',
  parameters: jsonSchema({
    type: 'object',
    properties: {
      step_id: {
        type: 'string',
        description: 'Step ID',
      },
      new_name: {
        type: 'string',
        description: 'New step name',
      },
    },
    required: ['step_id', 'new_name'],
  }, { validate: false }),
  execute: async (params) => {
    return { success: true };
  },
});

// Try to use the tool
const model = vertex('gemini-2.0-flash-exp');

  model,
  messages: [{ role: 'user', content: 'Test' }],
  tools: { testTool },
});

// This fails with:
// Unable to submit request because `testTool` functionDeclaration
// parameters schema should be of type OBJECT
```

### Also Fails with Zod

The same issue occurs when using Zod schemas:

```typescript
import { z } from 'zod';

const testTool = tool({
  description: 'Update a workflow step',
  parameters: z.object({
    step_id: z.string().describe('Step ID'),
    new_name: z.string().describe('New step name'),
  }),
  execute: async (params) => {
    return { success: true };
  },
});

// Same error occurs
```

## Expected Behavior

The request body sent to Vertex AI should contain the complete schema:

```json
{
  "functionDeclarations": [
    {
      "name": "testTool",
      "description": "Update a workflow step",
      "parameters": {
        "type": "OBJECT",
        "properties": {
          "step_id": {
            "type": "STRING",
            "description": "Step ID"
          },
          "new_name": {
            "type": "STRING",
            "description": "New step name"
          }
        },
        "required": ["step_id", "new_name"]
      }
    }
  ]
}
```

## Actual Behavior

The request body sent to Vertex AI has empty properties:

```json
{
  "functionDeclarations": [
    {
      "name": "testTool",
      "description": "Update a workflow step",
      "parameters": {
        "properties": {}
      }
    }
  ]
}
```

The `type`, `required`, and all property definitions are missing.

## Evidence

I captured the actual request body from the error object:

```typescript
for await (const chunk of result.fullStream) {
  if (chunk.type === 'error') {
    const error = chunk.error;
    console.log('Request body:',
      JSON.stringify(error.requestBodyValues.tools, null, 2)
    );
  }
}
```

Output shows empty properties for all custom tools.

## Working Comparison

MCP tools created via `experimental_createMCPClient().tools()` work correctly with the same Vertex AI provider, suggesting the issue is specific to the schema serialization path used by `tool()`.

## Root Cause

The issue appears to be in how the Vertex AI provider serializes tool schemas. The schema is correctly constructed in memory (verified via logging), but gets stripped during serialization to the Vertex AI API format.

This may be related to issue #6572 which mentions "parameters.properties field missing in request body when using Vertex AI".

## Impact

This bug makes it **impossible to use custom tools with the Vertex AI provider** unless they have no parameters. Only MCP-provided tools work correctly.

## Workaround

Currently, the only workaround is to use `experimental_createMCPClient()` for all tools, which is not practical for custom application-specific tools.

## Additional Context

- The schema cleaning/conversion logic works correctly (tested extensively)
- Both `jsonSchema()` and Zod approaches fail identically
- The error occurs during the HTTP request to Vertex AI
- MCP tools using the same provider work fine
- Error response from Vertex AI: `INVALID_ARGUMENT` with message about schema should be type OBJECT

## Request

Please fix the schema serialization in the Vertex AI provider to preserve all schema properties when building function declarations for the Vertex AI API.

sujal12344:
## Overview

This PR fixes a bug where custom tools with parameters fail when using the Google/Vertex AI provider. The issue was that `tool.inputSchema` was not properly converted to a plain JSON Schema before being passed to the schema converter.

## Root Cause

The `prepareTools` function in the Google provider was directly using `tool.inputSchema` without calling `asSchema().jsonSchema` to extract the actual JSON Schema. Since `tool.inputSchema` is a `FlexibleSchema` type (which includes Schema objects with a `.jsonSchema` getter), this caused a `Schema` object to be passed instead of a plain `JSONSchema7` object.  

During serialization for the Vertex AI API, only an empty `properties` object was preserved, while all actual property definitions, types, and required fields were lost. This made custom tools with parameters fail.

## Changes

- Added `import { asSchema } from '@ai-sdk/provider-utils'`.
- Updated line 147 in `prepare-tools.ts` from:
  ```ts
  parameters: convertJSONSchemaToOpenAPISchema(tool.inputSchema),
  
  to 
  
  parameters: convertJSONSchemaToOpenAPISchema(
    await asSchema(tool.inputSchema).jsonSchema,
  ),


- Made prepareTools async and updated its return type to Promise<...>.

- Added await to prepareTools calls in google-generative-ai-language-model.ts.

This now matches the pattern used in packages/ai/src/prompt/prepare-tools-and-tool-choice.ts.

# Testing

## Tested with custom tools that have:

- Simple string/number parameters

- Nested object parameters

- Required fields

- Tools created via both jsonSchema() and Zod schemas

All scenarios now work correctly with the Vertex AI provider.

## Related Issues

Fixes #9761

### DIFF
--- a/packages/google/src/google-generative-ai-language-model.ts
+++ b/packages/google/src/google-generative-ai-language-model.ts
@@ -122,7 +122,7 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
       tools: googleTools,
       toolConfig: googleToolConfig,
       toolWarnings,
-    } = prepareTools({
+    } = await prepareTools({
       tools,
       toolChoice,
       modelId: this.modelId,
--- a/packages/google/src/google-prepare-tools.ts
+++ b/packages/google/src/google-prepare-tools.ts
@@ -3,6 +3,7 @@ import {
   LanguageModelV3CallWarning,
   UnsupportedFunctionalityError,
 } from '@ai-sdk/provider';
+import { asSchema } from '@ai-sdk/provider-utils';
 import { convertJSONSchemaToOpenAPISchema } from './convert-json-schema-to-openapi-schema';
 import { GoogleGenerativeAIModelId } from './google-generative-ai-options';
 
@@ -14,7 +15,7 @@ export function prepareTools({
   tools: LanguageModelV3CallOptions['tools'];
   toolChoice?: LanguageModelV3CallOptions['toolChoice'];
   modelId: GoogleGenerativeAIModelId;
-}): {
+}): Promise<{
   tools:
     | {
         functionDeclarations: Array<{
@@ -34,7 +35,7 @@ export function prepareTools({
         };
       };
   toolWarnings: LanguageModelV3CallWarning[];
-} {
+}> {
   // when the tools array is empty, change it to undefined to prevent errors:
   tools = tools?.length ? tools : undefined;
 
@@ -144,7 +145,9 @@ export function prepareTools({
         functionDeclarations.push({
           name: tool.name,
           description: tool.description ?? '',
-          parameters: convertJSONSchemaToOpenAPISchema(tool.inputSchema),
+          parameters: convertJSONSchemaToOpenAPISchema(
+            await asSchema(tool.inputSchema).jsonSchema,
+          ),
         });
         break;
       default:


lgrammel:
duplicate of #9762

sujal12344:
I haven’t noticed his PR

PR #9793, opened 2025-10-25
feat(provider/anthropic): tool calling with structured output

### LINKED ISSUE #9789: Not able to use tools with strucutred output in anthropic provider
### Description

When using output with tools in anthropic, it ignores tools and just passes json response tool to get the structured output, why is this like this ?

Image

### AI SDK Version

_No response_

### Code of Conduct

AVtheking:
## Background

Currently tool calling cannot be combined with structured outputs when using the Anthropic provider.

## Summary

Added support for tool calling with structured output in the Anthropic provider.

## Manual Verification

- [x] run `examples/ai-core/src/generate-text/anthropic-output-array-tools.ts`
- [x] run `examples/ai-core/src/stream-text/anthropic-output-array-tools.ts`

## Checklist

## Tasks

- [x] generate snapshot test with fixture
- [x] stream snapshot test with fixture
- [x] manual testing
- [x] example tweaks
- [x] check if docs updates are needed
- [x] code review
- [x] add test for streaming error scenario
- [x] add agent example

## Related Issues
Fixes #9789


examples/ai-core/src/generate-text/anthropic-tool-call-with-output.ts
```diff
@@ -0,0 +1,31 @@
+import { anthropic } from '@ai-sdk/anthropic';
+import { generateText, Output, stepCountIs, tool } from 'ai';
+import 'dotenv/config';
+import { z } from 'zod';
+import { weatherTool } from '../tools/weather-tool';
+
+async function main() {
```
lgrammel: pls use the `run` helper (see other example)

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -360,9 +352,9 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
     } = await prepareTools(
       jsonResponseTool != null
         ? {
-            tools: [jsonResponseTool],
-            toolChoice: { type: 'tool', toolName: jsonResponseTool.name },
-            disableParallelToolUse: true,
+            tools: tools ? [...tools, jsonResponseTool] : [jsonResponseTool],
+            toolChoice: { type: 'required' },
+            disableParallelToolUse: anthropicOptions?.disableParallelToolUse,
```
lgrammel: the setting was `true` on purpose so we can guarantee that only the json response tool is called in the last step, and only one time (not in parallel)

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -360,9 +352,9 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
     } = await prepareTools(
       jsonResponseTool != null
         ? {
-            tools: [jsonResponseTool],
-            toolChoice: { type: 'tool', toolName: jsonResponseTool.name },
-            disableParallelToolUse: true,
+            tools: tools ? [...tools, jsonResponseTool] : [jsonResponseTool],
```
lgrammel: `tools: [...(tools ?? []), jsonResponseTool]` is more concise

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -483,28 +475,25 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
 
     const content: Array<LanguageModelV3Content> = [];
     const mcpToolCalls: Record<string, LanguageModelV3ToolCall> = {};
+    let isJsonToolCalled = false;
```
lgrammel: this is why turning off parallel is important

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -483,28 +475,25 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
...
-                const source = createCitationSource(
-                  citation,
-                  citationDocuments,
-                  this.generateId,
-                );
-
-                if (source) {
-                  content.push(source);
-                }
+          content.push({ type: 'text', text: part.text });
```
lgrammel: this is a larger problem and a potential dealbreaker. I remember this is why tools are currently not supported with json output. Some anthropic models return text independent of tool calling, before the tool call. When that happens all output parsing is broken.

This needs to be reverted to the previous version, when structured output is active all text must be ignored.

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -535,9 +524,16 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
           break;
         }
         case 'tool_use': {
+          const isJsonResponseTool =
+            usesJsonResponseTool && part.name === 'json';
+
+          if (isJsonResponseTool) {
+            isJsonToolCalled = true;
+          }
```
lgrammel: rm flag, use inline check for `usesJsonResponseTool` and check for tool name instead

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -746,7 +742,7 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
       content,
       finishReason: mapAnthropicStopReason({
         finishReason: response.stop_reason,
-        isJsonResponseFromTool: usesJsonResponseTool,
+        isJsonResponseFromTool: isJsonToolCalled,
```
lgrammel: revert to `usesJsonResponseTool` and checking the name of the invoked tool (parallel needs to be disabled)

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -918,7 +914,14 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
                   }
 
                   case 'tool_use': {
-                    contentBlocks[value.index] = usesJsonResponseTool
+                    const isJsonResponseTool =
```
lgrammel, line 923: similar concerns as above apply here

packages/anthropic/src/anthropic-messages-language-model.ts
AVtheking: I am sorry but didn't get it, this flag was doing same thing I suppose to check if the json response tool is invoked right?

lgrammel: We need to distinguish 3 states:
- `usesJsonResponseTool` false (regular outputs)
- `usesJsonResponseTool` true, `json` tool invoked: json tool should be streamed as text, turn should end with stop finish reason
- `usesJsonResponseTool` true, other tool invoked: no text outputs must be streamed, tool should be stream as tool, turn should end with tool-use finish reason

AVtheking: so that is what this flag `isJsonToolCalled` doing for right ? 
 - for regular outputs this would be false
 - if true and json tool invoked this flag would be true and end with stop finsih reason.
 - structured output is present and other tools are used, this will end with tool-call finsih reason. (with parallel disabled)

lgrammel: here it is correct, but since you need to manage 3 states (to avoid sending text parts) it needs to be changed in other places.

examples/ai-core/src/generate-text/anthropic-tool-call-with-output.ts
lgrammel: still showing main

AVtheking: sorry I didn't pushed at that time, now pushed

packages/anthropic/src/anthropic-messages-language-model.ts
AVtheking: but I have doubt here, let say llm did return some text content before calling tool but we are doing output parsing for the last step text message right? then how would this break, I could be wrong here

packages/anthropic/src/anthropic-messages-language-model.ts
AVtheking: not needed to remove this flag, as it would be used in mapping finish response.

packages/anthropic/src/anthropic-messages-language-model.ts
AVtheking: that sending text parts is fixed, and this flag could be used at other parts too in a similar fashion and then used here to map the finsih reason

packages/anthropic/src/anthropic-messages-language-model.ts
lgrammel: we dont know if the result will be tool calling or not. the error scenario is the following:

anthropic send text: "i will use json response tool" before the tool is called. we pass through that text
then the json response tool is invoked, it's json content is also stream as text

the ai lib will receive "i will use json response tool" and the json. both together do not parse as json, leading to an error.

AVtheking: got you! yeah right, I did added this check here

lgrammel:
@AVtheking thanks! i will take it from here and make a few further changes to the PR.

lgrammel (approved):

PR #9794, opened 2025-10-25
fix(ai): do not drop custom headers in HttpChatTransport

### LINKED ISSUE #9792: HttpChatTransport drops all user headers when given a Headers instance
### Description

I noticed that passing a `Headers` instance into the Vue chat transport silently drops everything except the SDK-added `user-agent`. That breaks any request that relies on `Authorization` or other custom headers.

**Steps to reproduce**
1. Create a `Chat`/`HttpChatTransport` with `headers: new Headers({ Authorization: 'Bearer demo' })`.
2. Call `sendMessages` (any trigger works).
3. Inspect the outgoing request in a mocked `fetch`.

**What happens**
- The request headers contain only `content-type` plus the SDK `user-agent`. `Authorization` (and any other entries) disappear.

**What I expected**
- All existing headers should be preserved. Docs for `useChat` list the type as `Record<string, string> | Headers`, so handing over a `Headers` object should work fine.

**Why I think this is a bug**
- `withUserAgentSuffix` calls `removeUndefinedEntries` before creating `new Headers(...)`. That helper does `Object.entries(record)`, which returns an empty array for `Headers`. The net effect is that we always lose the originals.
- Other helpers such as `prepareHeaders` handle `HeadersInit` correctly.
- Common examples (e.g. the Kasada proxy in `examples/next-openai-kasada-bot-protection`) already use `new Headers(request.headers)`, so anyone forwarding those headers will get bit.

### Environment

- Node 22.11.0
- pnpm 10.11.0

### AI SDK Version

- `ai@6.0.0-beta.72`
- `@ai-sdk/vue@3.0.0-beta.72`

### Code of Conduct

teunlao:
## Summary
  Fixes the loss of user-provided headers when `HttpChatTransport` receives a
  `Headers` instance.

  ## Root Cause
  `withUserAgentSuffix` called `removeUndefinedEntries` on the incoming headers. That
  helper runs `Object.entries`, which returns an empty array for `Headers`, so every
  custom header vanished before the request was sent.

  ## Changes
  1. Added `normalizeHeaders()` in `packages/provider-utils/src/normalize-headers.ts`
  to unwrap any `HeadersInit` form (actual `Headers`, tuple arrays, or plain objects)
  and filter out nullish values.
  2. Updated `withUserAgentSuffix()` to rely on `normalizeHeaders()`, so the original
  headers survive when the user passes a `Headers` instance.
  3. In `packages/ai/src/ui/http-chat-transport.ts`, normalized the headers once
  before handing them to the optional `prepare*` callbacks. This keeps the data format
  consistent with the new helper; the transport’s behavior is otherwise unchanged.
  4. Added unit tests for `normalizeHeaders`.
  5. Extended the existing `with-user-agent-suffix` test to cover the `Headers` input
  case.
  
  ## Related Issues
  Fixes #9792


packages/provider-utils/src/normalize-headers.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (typeof headers === 'object') {
+    return Object.fromEntries(
+      Object.entries(headers).filter(
+        ([_, value]) => value !== undefined && value !== null,
+      ),
+    ) as Record<string, string>;
+  }
+
+  return {};
+}
```
lgrammel, line 38: We have a `headers-utils.ts` with some similar functionality. It might be better to extend `extractHeaders` to support ` Record<string, string | undefined>` as input as well, add the tests to `headers-utils.test.ts`, and then use that (potentially `extractHeaders` could be renamed to `normalizeHeaders` as well).

teunlao: Thanks for the suggestion! Moved the old Bedrock `extractHeaders` logic into `@ai-sdk/provider-utils` as `normalizeHeaders`, so everything now normalizes Headers / tuple arrays / plain records in one place.

One thing though: Bedrock's helper already lower-cased the header keys, but the rest of the SDK kept whatever casing the user provided. I kept the lower-case normalization here since it matches HTTP spec anyway (header names are case-insensitive), but wasn't 100% sure if any integration or monitoring setups might depend on the original casing. Happy to switch back if you'd prefer not to normalize the case.

lgrammel (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/9808

PR #3764, opened 2024-11-19
feat(reranking): support reranking models

patelvivekdev:
## Summary
This PR adds support for re-ranking models in the Vercel AI SDK provider. Currently, the SDK supports embedding models from providers like `Cohere` and `Mixedbread AI` for building Retrieval-Augmented Generation (RAG) systems. By integrating re-ranking models, developers can enhance these systems by improving the relevance of the retrieved results without additional manual setup or using another SDK.

## Manual Verification

- [x] cohere string
- [x] cohere object
- [x] bedrock string
- [x] bedrock object
- [x] together string
- [x] together object

## Tasks

- [x] rm maxDocsPerCall and error (no automatic handling; rely on API errors instead)
- [x] warnings and warning logging
- [x] response model id and response id support
- [x] rm provider metadata (you can just use the result.response)
- [x] remove generics from spec
- [x] update telemetry tests 
- [x] snapshot testing
- [x] cleanup provider v2/v3 code (separate PR)
- [x] extract warning logger changes
- [x] examples should use new format
- [x] verify model ids match current provider reranking models

## Future Work
- docs: rerank page (similar to embedding) @nicoalbanese 
- docs: rerank function reference page @nicoalbanese 
- docs: provider page sections on reranking (bedrock, cohere, togetherai) @nicoalbanese 

## Related Issues
closes #3584
extracted #9815
extracted #9820

Issue #8662, opened 2025-09-15
v6


gr2m:
### Milestone

[v6.0](https://github.com/vercel/ai/milestone/1)

### Setup

- [x] branching
  - [x] Create `v5.0` branch from latest main, setup releases to `latest` - patch only (as always)
  - [x] Set releases from `main` to be `<next minor version>-beta.X`, e.g. `ai@5.1.0-beta.1`
  - [x] document new branching in `contributing/branches.md`

### Main features

See sub issues below

- [x] Direct support for multiple validation libraries
- [x] Preliminary Tool Outputs
- [x] Experimental Context Object
- [x] Experimental Download Function
- [x] Warning Logger (#8343)

### Secondary features

- [ ] #8889

### Release process

- [ ] search for `// TODO ai sdk 5` and `// TODO ai sdk 6` to be sure everything we wanted to was addressed, create issues for remaining comments
- [ ] Create blog post
- [ ] Create demo(s)
- [ ] update release config for `main` branch to release to latest
- [ ] Release `ai@6.0.0` and bump all other packages accordingly in latest
- [x] Make sure `@ai-sdk/provider` gets released as 3.0.0

### Breaking changes

- [ ] in general, compare exports of `ai@5` and `ai@6`
- [ ] `ai/test` no longer exports v2 mock classes now. Use the new v3 mock classes instead.

  Before

  ```js
  import {
    MockEmbeddingModelV2,
    MockImageModelV2,
    MockLanguageModelV2,
    MockProviderV2,
    MockSpeechModelV2,
    MockTranscriptionModelV2,
  } from "ai/test";
  ```

  After

  ```js
  import {
    MockEmbeddingModelV3,
    MockImageModelV3,
    MockLanguageModelV3,
    MockProviderV3,
    MockSpeechModelV3,
    MockTranscriptionModelV3,
  } from "ai/test";
  ```

calvinalkan:
Relevant for human in the loop #8700

gr2m:
### Update September 23

We identified 3 features as must haves for the 5.1 stable release:

1. Telemetry v2 (top priority)
2. Tool Execution Approval (human-in-the-loop)
3. Image Editing

We need to implement these by October 10th in order to have enough time for a release blog post and demos before the planned 5.1 stable release at [Ship AI](https://vercel.com/ship/ai). We are on track to achieve that.

All other features are nice-to-haves, but we can work on these up to the release date.

evelynevervillefelix-hue:
[...]

PR #9820, opened 2025-10-27
fix(provider): add specificationVersion to ProviderV3

lgrammel:
## Background

`ProviderV3` was missing specification version.
We had no `ProviderV2` to `ProviderV3` adapter.

## Summary

* introduce `specificationVersion` on `ProviderV3`
* add `asProviderV3` adapter
* extract adapter functions for the models

## Future Work

* ensure that we have correct model v2 to v3 adapters

## Related Issues

Extracted from #3764

### DIFF
--- a/.changeset/angry-cobras-care.md
+++ b/.changeset/angry-cobras-care.md
@@ -1,0 +1,35 @@
+---
+'@ai-sdk/openai-compatible': patch
+'@ai-sdk/amazon-bedrock': patch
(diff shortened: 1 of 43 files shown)


nicoalbanese (approved):

gr2m (changes requested):
blocking so I can review before merge

gr2m (approved):
Looks good. We will add tests for v2 -> v3 transforms as part of #8662.

PR #9832, opened 2025-10-27
chore(ai): rename relevanceScore to score, add shorthand names

lgrammel:
## Background

All models except reranking models have shorthands. The `relevanceScore` property name can be shorter without reducing meaning.

## Summary

* introduce `reranking` shorthand names on the together, cohere, and bedrock providers
* rename `relevanceScore` to `score` on `RerankResult`

## Related Issues

DX improvements after demoing #3764


nicoalbanese (approved):

Snapp949 (commented):
Enable

PR #9853, opened 2025-10-28
feat(provider/anthropic): enable fine grained tool streaming by default

### LINKED ISSUE #3422: StreamObject doesn't work with Anthropic
### Description

The streamObject function does not actually stream for anthropic. It resolves at the end instead. streamText works fine for anthropic unless a tool is required. It works fine for OpenAi as well. Tested with Haiku and Sonnet.

Related issue: https://github.com/vercel/ai/discussions/1980

### Code example

```typescript
import { AnthropicProvider, createAnthropic } from '@ai-sdk/anthropic';
import { createOpenAI, OpenAIProvider } from '@ai-sdk/openai';
import { CoreMessage, DeepPartial, streamObject } from 'ai';

//  Other code...

const model = this.anthropic(CLAUDE_MODELS.HAIKU);

// This streams fine
// const model = this.openai(OPEN_AI_MODELS.GPT_4_TURBO);

            const result = await streamObject({
                model,
                system: 'You are a seasoned React and Tailwind expert.',
                schema: StreamReponseObject,
                messages,
            });

// This actually wait the whole time, resolving everything at the end 

            for await (const partialObject of result.partialObjectStream) {
                console.log(partialObject)
            }
```

### Additional context

Using main node.js process in an electron app

lgrammel:
## Background

Anthropic tool streaming is not enabled by default. You currently need to set the `fine-grained-tool-streaming-2025-05-14` beta header, but this does not work with the AI gateway.

## Summary

- enable Anthropic tool streaming by default
- add `toolStreaming` provider option to opt out
- **breaking semantic change**: will start tool call streaming (not to be backported)

## Manual Verification

Ran examples.

## Related Issues

Fixes #3422


iNishant:
@lgrammel is it a good idea to have tool streaming enabled by default for anthropic? This often leads to more invalid tool calls. For eg I see quite a lot of schema errors where strings were not wrapped with quotes. Something similar was documented here as well https://github.com/vercel/ai/pull/7003

lgrammel:
@iNishant do you have examples where this fails? have you tried enabling strict mode on the tools (for newer anthropic models)? https://platform.claude.com/docs/en/build-with-claude/structured-outputs

iNishant:
@lgrammel strict mode from anthropic still has some issues and constraints if you have a lot of tools. I tried to create a small repro but wasn't able to reproduce the same error. Could be that this issue pops up with a lot of input tokens / tools. 

Its just that in my project, the issue is consistently there and goes away when tool input streaming is disabled. Lets see if other folks report similar issues.

lgrammel:
@iNishant either way people will have issues. either streaming does not work or this happens. there were constant complaints about tool streaming not working so my preference is to leave this on. plus it would be a breaking change to turn it off. you can opt out if you need

PR #9922, opened 2025-10-30
feat(ai): Output.text() is default output mode

lgrammel:
## Background

When no `output` option is specified, the default output mode should be `Output.text()` so that `output` and `partialOutputStream` is still usable.

## Summary

* change default output to `Output.text()`
* remove `NoOutputSpecifiedError`

## Manual Verification

- [x] run `examples/ai-core/src/generate-text/openai-output-default.ts`
- [x] run `examples/ai-core/src/stream-text/openai-output-default.ts`

## Tasks

- [x] generateText
- [x] streamText
- [x] examples
- [x] type tests
- [x] self review
- [x] docs
- [x] changeset


gr2m (approved):
elegant!

PR #9896, opened 2025-10-29
feat: xai server-side tool calling

dancer:
## Background

xai supports server-side agentic tool calling through their responses api. this allows models to autonomously execute tools like web search, x search, and code execution on their servers without requiring client-side tool handling

https://docs.x.ai/docs/api-reference#create-new-response
https://docs.x.ai/docs/guides/tools/overview

## Summary

- added xai responses api with server-side tool execution
- implemented 5 server-side tools: web_search, x_search, code_execution, view_image, view_x_video
- added tool name mapping to handle sub-tool variants (browse_page → web_search, x_semantic_search → x_search)
- added citation extraction from response annotations
- added multi-turn conversation support with function_call items
- created next.js example with toolloopagent for multi-step workflows

## Manual Verification

tested with next.js example app using toolloopagent:
- multi-step research queries with web_search and x_search working
- tool sub-calls properly mapped to parent tools (no infinite loops)
- citations displaying with url fallback when title missing
- streaming events handling all 11+ event types correctly
- 99 tests passing with real api fixtures

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## follow up

use lazySchema, see https://github.com/vercel/ai/issues/9339 on all of xai

PR #9931, opened 2025-10-30
fix(spec): `LanguageModelV3ToolResult["result"]` change from `unknown` to `NonNullable<JSONValue>`

gr2m:
## Background

Discovered in https://github.com/vercel/ai/pull/9896#pullrequestreview-3402058357

## Summary

As far as I can tell, `result` has to be set to something that is not null or undefined. This makes the spec more precise and would have surfaced the problem of https://github.com/vercel/ai/pull/9896#pullrequestreview-3402058357 immediately

## Manual Verification

When the fix from this PR is applied, a type error is thrown in https://github.com/vercel/ai/pull/9896

## Checklist


lgrammel (approved):

lgrammel:
We also need to remove the `providerExecuted` flag in this file that was somehow added to the spec. It would always be true i.e. not needed.

lgrammel:
Removed the `providerExecuted` flag in https://github.com/vercel/ai/pull/9936

packages/provider/src/language-model/v3/language-model-v3-tool-result.ts
```diff
@@ -19,7 +19,7 @@ export type LanguageModelV3ToolResult = {
   /**
    * Result of the tool call. This is a JSON-serializable object.
    */
-  result: unknown;
+  result: NonNullable<unknown>;
```
lgrammel: we may want to also consider `NonNullable<JSONValue>` now that JSONValue supports undefined but the implications are unclear

gr2m: yeah I. thought so too. I'll do that and see how it goes

gr2m: shouldn't it work with `JSONObject | JSONArray`?

gr2m: Hmm it's tricky because it means we cannot use `unknown` for any (deep) value of the result object

like in line 50 here
https://github.com/vercel/ai/blob/f0b21570e587c9865a37b655cf1903f70a965f6c/packages/openai/src/tool/file-search.ts#L43-L60

gr2m: the `z.unknown()` should be `z.union([z.string(), z.number(), z.boolean()])` as per OpenAPI spec, will update

packages/openai/src/responses/openai-responses-api.ts
```diff
@@ -398,7 +398,10 @@ export const openaiResponsesChunkSchema = lazySchema(() =>
             results: z
               .array(
                 z.object({
-                  attributes: z.record(z.string(), z.unknown()),
+                  attributes: z.record(
+                    z.string(),
+                    z.union([z.string(), z.number(), z.boolean()]),
+                  ),
```
gr2m, line 409: as a side note: JSON API / JSON Schema specs do not have an "unknown" type. I haven't checked all providers we support yet, but all the important ones do have a JSON API spec for their APIs.

We still use `.unknown()` in other places related to provider requests/responses that we could probably all get rid off with some research.

packages/provider/src/language-model/v3/language-model-v3-tool-result.ts
```diff
@@ -19,7 +20,7 @@ export type LanguageModelV3ToolResult = {
   /**
    * Result of the tool call. This is a JSON-serializable object.
    */
-  result: unknown;
+  result: JSONObject | JSONArray;
```
gr2m: will change to `NonNullable<JSONValue>` after discussing with @lgrammel

PR #9936, opened 2025-10-31
chore(provider): remove providerExecuted from LanguageModelV3ToolResult

lgrammel:
## Background

`LanguageModelV3ToolResult` has a `providerExecuted` property. However, every tool result produced by the language model is always provider executed. The information is therefore redundant, a source of errors, and leads to unnecessary code bloat.

## Summary

Remove `providerExecuted` property from `LanguageModelV3ToolResult`.


josh-williams:
@lgrammel is there any chance of this change being backported? I'm hitting a bug specifically because this line ([web_fetch_tool_result](https://github.com/vercel/ai/blob/36b175c20415080a5e330f4616ba71ff9a9db267/packages/anthropic/src/anthropic-messages-language-model.ts#L1009)) doesn't set providerExecuted true. If it can't be backported, do you have any ETA on when the 6.0.0 will be out?

josh-williams:
I just put up this PR to backport this: https://github.com/vercel/ai/pull/10074

PR #9960, opened 2025-11-01
feat(ui): add output streaming to tool invocations

### LINKED ISSUE #9306: Feature request: add `output-streaming` to tool invocations
### Description

Currently, there's `output-available` with `preliminary: true`. It'd be nice to instead have an explicit state for `output-streaming` (similar to `input-available`) when `yield`ing progress updates from a tool.

(This would be a slight breaking change, which likely will block this for a bit)

Related: #7987

### AI SDK Version

_No response_

### Code of Conduct

Nikunj0601:
## Background

Currently, when tools yield multiple intermediate results using async generators, all states are shown as `output-available` with `preliminary: true`. The `output` property only contains the last yielded value, and the final return value from the async generator is lost. This makes it difficult for UI consumers to:
- Distinguish between streaming progress and final results
- Access the complete history of intermediate yields
- Retrieve the actual final return value from the tool

## Summary

This PR introduces enhanced tool progress tracking by:

1. **Adding `output-streaming` state**: A new explicit state that indicates a tool is actively streaming intermediate results, distinct from `output-available`.

2. **Adding `yields` array**: A new property that accumulates all yielded values during tool execution, allowing UI consumers to access the full history of intermediate results.

3. **Separating concerns**: The `output` property now contains only the final returned value from the tool's `execute` function, while all intermediate yields are available in the `yields` array.

4. **Fixing async generator return value**: Corrected `executeTool` to properly capture the final return value of async generators (when `done === true`) rather than the last yielded value.

The implementation uses a UI-only approach, interpreting existing `tool-output-available` chunks with `preliminary: true` as `output-streaming` states. This ensures full backward compatibility with existing `streamText` functionality.

## Checklist

## Related Issues

Fixes #9306


rfsmart-eland:
Very excited to see this PR come in!

I think this would help my issue with preliminary tool results where when a long running tool is aborted and I have preliminary results, my tool state won't update to error to show users it has stopped.

ChrisCindy:
Great PR! We've encountered similar needs in our scenario! But I think using `tool-output-delta` might be more appropriate? This way it would perfectly correspond with `tool-input-delta`, lol

lgrammel:
I don't understand the premise of why this PR was created in the first place. `preliminary: true/false` is how you distinguish streaming from the final result. Can you show an example that does not work with the current implementation?

rfsmart-eland:
> I don't understand the premise of why this PR was created in the first place. `preliminary: true/false` is how you distinguish streaming from the final result. Can you show an example that does not work with the current implementation?

I am not the author, but for me relying on the `preliminary: true` value in a stream response seems like an anti-pattern. The AI SDK docs and AI Elements Tool component all use the tool state as the primary indicator of tool status, and this PR adds a proper state that applies to that pattern. Without it, I think all AI SDK users would have to build additional, custom logic to read a tool response for a preliminary value and handle it accordingly--this seems like the exact boilerplate the AI SDK tries to prevent.

gr2m:

PR #9733, opened 2025-10-23
feat(provider/anthropic): Url support for pdfs and images in tool results

nlaz:
## Background

Now that v6 has `image-url` and `file-url` this makes it simple to support pdfs and images in the anthropic tool results blocks. 

This PR adds support for url-based PDFs in tool results. This is a pretty common way to fetch and insert PDF without blowing up the context window. Anthropic's API handles this [out of the box](https://docs.claude.com/en/docs/build-with-claude/pdf-support). Just needed to update the types on the tool result handling.

## Summary

Uses the `image-url` and `file-url` types to send urls to the provider.

## Manual Verification

Used the example files to verify the functionality.

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

Hacky Backport: https://github.com/vercel/ai/pull/9503
Related: https://github.com/vercel/ai/pull/9041
Fixes https://github.com/vercel/ai/issues/8014

PR #9963, opened 2025-11-01
feat(provider/google): Url support for pdfs and images in tool results

### LINKED ISSUE #8014: Unable to pass PDF files from the tool calls

songkeys:
## Background

This PR solves issue #8014 for google provider, mirroring the changes from #9733 .

Google gemini API supports uploading pdf with url: <https://ai.google.dev/gemini-api/docs/document-processing#large-pdfs-urls>

## Summary

Uses the `image-url` and `file-url` types to send urls to the provider.

## Manual Verification

Used the example files to verify the functionality.

## Checklist

## Related Issues

Related: #9733 
Fixes #8014


examples/ai-core/src/generate-text/google-image-tool-result-url.ts
gr2m, line 1: I get

> Unsupported file uri: https://github.com/vercel/ai/blob/main/examples/ai-core/data/comic-cat.png?raw=true

examples/ai-core/src/generate-text/google-pdf-tool-results-base64.ts
gr2m, line 1: I get no error, but

> Assistant response: "\nI have successfully read the PDF document. However, I cannot summarize its content using the current tools. Please let me know if you would like me to do anything else with the document's content, or if you have another request!"

examples/ai-core/src/generate-text/google-pdf-tool-results-url.ts
gr2m, line 1: same thing

> Assistant response: "I can read the PDF document for you using the `readPDFDocument` tool, but I do not have the ability to summarize its content. Would you like me to proceed with just reading the PDF?"

gr2m:
Already implemented in #12777.

PR #9976, opened 2025-11-02
fix(openai): treat unknown models as reasoning

gr2m:
## Background

Most new models coming from OpenAI are reasoning now

## Summary

If a model is not recognized, it is now treated as a reasoning model.
I also refactored the code so that responses and chat models are sharing the same logic.

## Manual Verification

tested with `openai` examples

## Checklist

- [ ] n/a Documentation has been added / updated (for bug fixes / features)

## Future Work

Look into adding a flag to manually set if a model is treated as reasoning.


packages/openai/src/openai-is-reasoning-model.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    ['gpt-5', 'true'],
+    ['gpt-5-2025-08-07', 'true'],
+    ['gpt-5-codex', 'true'],
+    ['gpt-5-mini', 'true'],
+    ['gpt-5-mini-2025-08-07', 'true'],
+    ['gpt-5-nano', 'true'],
+    ['gpt-5-nano-2025-08-07', 'true'],
+    ['gpt-5-pro', 'true'],
+    ['gpt-5-pro-2025-10-06', 'true'],
+    ['new-unknown-model', 'true'],
```
gr2m, line 58: I will shorten the list before merging, just wanted to keep it complete for review

lgrammel (approved):

lgrammel:
i like this solution

lgrammel:
it changes the defaults for unknown models, so it might be a breaking change for folks who use the openai provider with a different baseurl and provider (instead of openaicompat) for historical reasons. might be good to keep ai sdk 6 only and mention in the migration guide

MonkeyLeeT:
@gr2m I like this solution too (really helpful for the tests we're doing), but is it possible to allow caller to tell AI SDK if this model is reasoning or not? There could be cases where we indeed want to test non-reasoning stealth model.

Also can we considering porting this back to v2 packages so we dont need to upgrade to v3?

gr2m:
> but is it possible to allow caller to tell AI SDK if this model is reasoning or not? There could be cases where we indeed want to test non-reasoning stealth model.

Not yet, but we have a plan ™️ 

> Also can we considering porting this back to v2 packages so we dont need to upgrade to v3?

It's a breaking change, we can't backport it, unfortunately. But the upgrade should be very simple this time

MonkeyLeeT:
> > but is it possible to allow caller to tell AI SDK if this model is reasoning or not? There could be cases where we indeed want to test non-reasoning stealth model.
> 
> Not yet, but we have a plan ™️
> 
> > Also can we considering porting this back to v2 packages so we dont need to upgrade to v3?
> 
> It's a breaking change, we can't backport it, unfortunately. But the upgrade should be very simple this time

Thanks I'll spare some time for updating to beta then.

As of today there's unfortunately one more complexity. Though not documented in https://openai.com/index/gpt-5-1-for-developers/, based on curl testing, 5.1 actually support temperature and log_probs **if you use None as reasoning effort**.

That'd further require more flexibility around reasoning model check beyond checking name, so might as well benefit from user-defined reasoning check after all.

MonkeyLeeT:
@gr2m Here's the official confirmation - https://x.com/bfioca/status/1990151789027721228

PR #9988, opened 2025-11-03
backport: test server

dancer:
## Background

back porting test server from v6 to v5 

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

#8994, #8588

PR #9999, opened 2025-11-03
fix(test-server): make `msw` a production dependency

gr2m:
## Background

We introduced a new package for the test server in #8588. The test server always depended on `msw`, but we didn't add it as production dependency because before we created a dedicated package for it, it was part of provider-utils. Making `msw` a production dependency would have meant that installing `ai` would install `msw` without ever needing it.

## Summary

`msw` will now automatically installed when installing `@ai-sdk/test-server`. It's already fixed it in the `main` branch via #9988

## Manual Verification

I'll try after the release but I'm sure enough this works as designed.

## Related Issues

#8588, #9988

### DIFF
--- a/packages/test-server/tsconfig.json
+++ b/packages/test-server/tsconfig.json
@@ -13,11 +13,6 @@
       "build",
       "node_modules",
       "tsup.config.ts"
-    ],
-    "references": [
-        {
-        "path": "../provider"
-        }
     ]
   }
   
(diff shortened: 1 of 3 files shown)


packages/test-server/tsconfig.json
```diff
@@ -13,11 +13,6 @@
       "build",
       "node_modules",
       "tsup.config.ts"
-    ],
-    "references": [
-        {
-        "path": "../provider"
```
gr2m, line 19: the package does not have a dependency on `@ai-sdk/provider`

aayush-kapoor (approved):
Thanks for the cleanup!

PR #10019, opened 2025-11-05
fix(openai): extract meta data from first chunk that contains any

gr2m:
> [!NOTE]
> I'm not sure how we feel about adding Azure-specific code to the OpenAI Provider implementation.

## Background

- We read the `modelId` and other meta information from the first chunk in the OpenAI provider ([source](https://github.com/vercel/ai/blob/348fd104aa18adc5e423f9c2bb2c8d50d690e10a/packages/openai/src/chat/openai-chat-language-model.ts#L484-L491))
- The first chunk from the `model-router` response has `"id":"","model":""`
- The first chunk includes `prompt_filter_results` which I think is Azure-specific:

  > The `$response` object contains a property named `prompt_filter_results` that contains annotations about the filter results. If you prefer JSON to a .NET object, pipe the output to ConvertTo-JSON instead of Format-List.

  —https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/content-filter-annotations?tabs=powershell

## Summary

Instead of reading metadata from only the first chunk, keep trying to find metadata until any was returned. Also treat `0` as falsy for `"created"`.

## Manual Verification

`examples/ai-core/src/stream-text/azure-model-router.ts`

## Checklist

- [ ] n/a ~~Documentation has been added / updated (for bug fixes / features)~~

## Future Work

Review other models and providers that might need a similar fix

## Related Issues

closes #9966


packages/azure/src/__snapshots__/azure-openai-provider.test.ts.snap
```diff
@@ -1,5 +1,70 @@
...
+exports[`chat > doStream > should set .modelId for model-router request 1`] = `
+[
+  {
+    "type": "stream-start",
+    "warnings": [],
+  },
+  {
+    "id": "chatcmpl-CYPS1lijGoK8gd9lYzY3r9Sx50nbt",
+    "modelId": "gpt-5-nano-2025-08-07",
+    "timestamp": 2025-11-05T04:30:21.000Z,
```
gr2m: this is the relevant part. Without the fix all 3 properties would be unset (timestamp would be `1970-01-01T00:00:00.000Z`

packages/openai/src/chat/get-response-metadata.ts
```diff
@@ -10,6 +10,6 @@ export function getResponseMetadata({
   return {
     id: id ?? undefined,
     modelId: model ?? undefined,
-    timestamp: created != null ? new Date(created * 1000) : undefined,
```
gr2m, line 13: this change was needed because Azure sets `"created": 0` in its [content filter annotation](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/content-filter-annotations?tabs=powershell) chunk

lgrammel (approved):
imo it would be better to have the tests in the openai package so we get immediate feedback using pnpm watch if we break something

gr2m:
> imo it would be better to have the tests in the openai package so we get immediate feedback using pnpm watch if we break something

agreed, will do.

gr2m:
> Review other models and providers that might need a similar fix

I tried other Azure model types but get this error message:

> ModelRouter is only supported for chat completions, assistants API and Azure Foundry Agents.

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10027

PR #10021, opened 2025-11-05
fix(ai): Preventing multiple concurrent chat resume calls

jonjomckay:
## Background

When using `React.StrictMode` and `useChat.resume=true`, calls to the stream resumption endpoint are made twice, which causes multiple duplicate messages to appear.

## Summary

This fixes the issue by preventing multiple calls to the resume endpoint, if one is already in progress.

## Manual Verification

I built the package locally with my changes, included in my React project, and verified that calls to the stream resumption endpoint are only made once - with no duplicate messages shown on screen either.

## Checklist

### DIFF
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -193,6 +193,7 @@ export abstract class AbstractChat<UI_MESSAGE extends UIMessage> {
 
   protected state: ChatState<UI_MESSAGE>;
 
+  private isResuming: boolean = false;
   private messageMetadataSchema:
     | FlexibleSchema<InferUIMessageMetadata<UI_MESSAGE>>
     | undefined;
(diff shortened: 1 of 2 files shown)


jonjomckay:
Not sure how to mark this in the issue, but it would be great if this could also be backported to the v5 SDK!

packages/ai/src/ui/chat.ts
```diff
@@ -414,7 +415,17 @@ export abstract class AbstractChat<UI_MESSAGE extends UIMessage> {
...
+      return;
+    }
+
+    this.isResuming = true;
+
+    try {
+      await this.makeRequest({ trigger: 'resume-stream', ...options });
+    } finally {
+      this.isResuming = false;
+    }
```
gr2m, line 428: I think the implementation might lead to unexpected behavior. When called the 2nd time, it will resolve immediately.

Instead of implementing a `true/false` flag, I would prefer we implement `this.resumeRequestPromise` that is set to `null` by default and after the resume request concludes. If it's truthy, return it. That way a 2nd parallel call will also await the resume request

gr2m (commented):
can you please add tests

gr2m:
Superseded by #11509.

PR #10050, opened 2025-11-06
fix: handle mode='tool', issue #10023

### LINKED ISSUE #10023: `generateText()` errors in v5 and v6 when using `(experimental_)output`
Moved from https://github.com/vercel/ai/issues/8662#issuecomment-3487226172 @gabimoncha

---

Hi @gr2m! I've installed ai-sdk v6 and tried to genereteObject with tool use, but I get some errors. These errors are exactly the same when I tried `experimental_output` from v5

Is it because of the providers? Would the vercel gateway work with it?

For `"@ai-sdk/groq": "^3.0.0-beta.31"` I'm getting this error

`json mode cannot be combined with tool/function calling`

For `"@ai-sdk/google": "^3.0.0-beta.39"` I'm getting this error

`Forced function calling (ANY mode) with a response mime type: 'application/json' is unsupported`

`"@openrouter/ai-sdk-provider": "^1.2.0"` doesn't seem to work as expected

And the `gateway` agent loops 20 times, if i set stopWhen 20, instead of stopping early like in v5

Sahilbhatane:
## Background

The `mode` parameter in `generateObject()` and `streamObject()` was defined in the function signature but never actually used in the implementation. This caused the functions to always use JSON response format (`responseFormat`), even when `mode='tool'` was specified. 

Providers like Groq, Google, and OpenRouter reject requests that combine JSON response format with tool calling, causing errors such as "json mode cannot be combined with tool/function calling" (Groq) or "Forced function calling with response mime type: 'application/json' unsupported" (Google).

## Summary

- Extracted and used the `mode` parameter in both `generateObject()` and `streamObject()` (defaults to `'auto'`)
- Implemented conditional logic:
  - When `mode='tool'`: Uses `tools` + `toolChoice` instead of `responseFormat`
  - When `mode='json'` or `mode='auto'`: Uses `responseFormat` (existing behavior)
- Tool mode creates a function tool with schema as `inputSchema` and extracts result from tool call's `input` property
- Added `tool-input-delta` handling for streaming tool responses

## Manual Verification

Manually tested with:
- ✅ Groq provider (`llama-3.3-70b-versatile`) with `mode='tool'` - successfully generates structured objects using tool calling
- ✅ Google provider (`gemini-2.0-flash`) with `mode='tool'` - successfully generates structured objects using tool calling
- Both providers now return `finish reason: tool-calls` instead of throwing errors

## Checklist

## Related Issues

Closes #10023

### DIFF
--- a/packages/ai/src/generate-object/generate-object.test.ts
+++ b/packages/ai/src/generate-object/generate-object.test.ts
@@ -994,12 +994,7 @@ describe('generateObject', () => {
       ]
     `);
       expect(model.doGenerateCalls[0].responseFormat).toMatchInlineSnapshot(`
-      {
-        "description": undefined,
-        "name": undefined,
-        "schema": undefined,
-        "type": "json",
-      }
+      undefined
     `);
     });
   });
--- a/packages/ai/src/generate-object/stream-object.test.ts
+++ b/packages/ai/src/generate-object/stream-object.test.ts
@@ -1367,12 +1367,7 @@ describe('streamObject', () => {
       `);
 
       expect(mockModel.doStreamCalls[0].responseFormat).toMatchInlineSnapshot(`
-        {
-          "description": undefined,
-          "name": undefined,
-          "schema": undefined,
-          "type": "json",
-        }
+        undefined
       `);
     });
   });
--- a/packages/ai/src/generate-object/stream-object.ts
+++ b/packages/ai/src/generate-object/stream-object.ts
@@ -324,6 +324,7 @@ Callback that is called when the LLM response and the final object validation ar
     schema: inputSchema,
     schemaDescription,
     schemaName,
+    mode = 'auto',
   } = 'schema' in options ? options : {};
 
   validateObjectGenerationInput({
@@ -353,6 +354,7 @@ Callback that is called when the LLM response and the final object validation ar
     messages,
     schemaName,
     schemaDescription,
+    mode,
     providerOptions,
     repairText,
     onError,
@@ -400,6 +402,7 @@ class DefaultStreamObjectResult<PARTIAL, RESULT, ELEMENT_STREAM>
     messages,
     schemaName,
     schemaDescription,
+    mode = 'auto',
     providerOptions,
     repairText,
     onError,
@@ -421,6 +424,7 @@ class DefaultStreamObjectResult<PARTIAL, RESULT, ELEMENT_STREAM>
     messages: Prompt['messages'];
     schemaName: string | undefined;
     schemaDescription: string | undefined;
+    mode?: 'auto' | 'json' | 'tool';
     providerOptions: ProviderOptions | undefined;
     repairText: RepairTextFunction | undefined;
     onError: StreamObjectOnErrorCallback;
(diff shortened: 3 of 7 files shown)


gr2m:
> I am a student so if there is any problem then please guide me when needed

This is a very good PR as far as I can tell so far, great work! I'm just curious, it's fine either way, but did you use any AI to help create this pull request. If so, how? I'm curious from a maintainer perspective how our PRs are created these days

packages/ai/src/generate-object/stream-object.test.ts
```diff
@@ -1367,12 +1367,7 @@ describe('streamObject', () => {
       `);
 
       expect(mockModel.doStreamCalls[0].responseFormat).toMatchInlineSnapshot(`
-        {
-          "description": undefined,
-          "name": undefined,
-          "schema": undefined,
-          "type": "json",
-        }
```
gr2m, line 1375: this is a change in behavior, any chance we can avoid it?

packages/ai/src/generate-object/stream-object.ts
```diff
@@ -499,13 +503,35 @@ class DefaultStreamObjectResult<PARTIAL, RESULT, ELEMENT_STREAM>
...
+              ? [
+                  {
+                    type: 'function' as const,
+                    name: toolName,
+                    description:
+                      schemaDescription ?? 'Respond with a JSON object.',
+                    inputSchema: jsonSchema,
+                  },
+                ]
+              : undefined,
```
gr2m, line 531: I'm not sure if we can inject a tool call like this. I think it's something that the AI SDK did in v4 and we moved away from it, please give me some time to investigate what happened

packages/ai/src/generate-object/generate-object.test.ts
```diff
@@ -994,12 +994,7 @@ describe('generateObject', () => {
       ]
     `);
       expect(model.doGenerateCalls[0].responseFormat).toMatchInlineSnapshot(`
-      {
-        "description": undefined,
-        "name": undefined,
-        "schema": undefined,
-        "type": "json",
-      }
```
gr2m, line 1002: change of behavior, can we avoid it?

gr2m (commented):
Thank you for the great PR so far. I am curious why the `mode` argument exists but is no longer used, I will investigate. 

Not also that we will deprecate `generateObject()` and `streamObject()` in favor of `generateText({ output })` and `streamText({ output })`, see https://github.com/vercel/ai/issues/10025. We might need to apply a similar fix to `generateText()` and `streamText()` for when `output` is set to a schema

gr2m:
Keeping track of what I've found

- `mode` was removed in #5580
- the `Respond with a JSON object` tool injection was removed in #5834

We need to figure out why we did that but I don't think we should bring back functionality that was removed as part of v5. If I recall correctly, we responsibility for the correct parsing was moved from the `ai` core methods like `generateObject()` into the providers. Could you review the PRs, I assume there are some clues in there.

gr2m:
The background and summary in https://github.com/vercel/ai/pull/5834 sum it up pretty well

> ## Background
> Almost all providers support structured output generation via JSON mode or JSON schema. The distinction between JSON and tool generation modes prevents streamlining flows and leads to additional calls when building remote providers.
> 
> ## Summary
> Always send JSON schema from `ai` to the providers when generating structured outputs. Push responsibility for prompt modification and forced tool usage (when JSON schema is not available) into providers.

So I'd to the following

1. We can close this PR
2. Create a separate PR that removes the `mode` types because it should have been removed as part of #https://github.com/vercel/ai/pull/5580 already
3. Look into implementing the tool call for the provider that needs it, or ask users to do it for providers that need it, if there are only a few left. In which case I'd update our docs

Sahilbhatane:
> > I am a student so if there is any problem then please guide me when needed
> 
> This is a very good PR as far as I can tell so far, great work! I'm just curious, it's fine either way, but did you use any AI to help create this pull request. If so, how? I'm curious from a maintainer perspective how our PRs are created these days

I didn't properly understood the question but I will tell you where I have used AI for this Issue - 
1. while building - AI changed nuxt config file to run as normal serverless function on vercel.
2. to generate test script - after replicating the issue I asked AI to create test file for google and groq to check if issue is solved or not,
3. for documentation - the above pull request description is enhanced by AI to look similar to default description.

none of them are used in Main branch are on other branch in my forked repo.

Sahilbhatane:
Thank you @gr2m for the details, I understand now that the `mode` parameter and tool injection were intentionally removed as part of the v5 architecture changes.

## Addressing Your Questions:
1. **Test snapshot changes** : These changes occurred because when no schema is provided, my implementation doesn't set `responseFormat`.
2. **Tool injection concern**: You're absolutely right - this was the v4 approach that v5 intentionally moved away from. The responsibility should be in the providers, not in the core.

## The Real Fix:
The issue #10023 should be fixed in the provider packages **@ai-sdk/groq** and **@ai-sdk/google**, Should detect when JSON mode + schema conflicts and fall back to tool calling

I agree with your recommendation to close this PR, Thanks for letting me work and for knowledge, I am new to Open source and open source contribution and this helped me understand it (they are not like any "good first issue" kinda contributions)

PR #10074, opened 2025-11-07
fix(anthropic): set providerExecuted on Anthropic web_fetch tool results

josh-williams:
## Background

This tool is providerExecuted and the tool call has that property set correctly, but the tool result does not. This fixed in main (and thus will be fixed in v6.x) by this change: https://github.com/vercel/ai/pull/9936 . But it is still broken in v5.x, so this fixes it there.

## Summary

Added `"providerExecuted": true` to the `tool-result` object for the Anthropic web fetch tool.

## Manual Verification

I didn't have a good way to verify manually.

## Checklist

## Related Issues

N/A

### DIFF
--- a/.changeset/wet-turkeys-rescue.md
+++ b/.changeset/wet-turkeys-rescue.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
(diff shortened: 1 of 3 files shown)


gr2m (approved):
cheers! Thanks again for contributing Josh!

PR #10093, opened 2025-11-08
Fix(ai): correctly valiate empty arrays

### LINKED ISSUE #10003: `validateUIMessages` doesn't catch empty arrays
### Description

The `validateUIMessages` and `safeValidateUIMessages` functions don't properly check for arrays being not-empty, for example:

```ts
await safeValidateUIMessages({ messages: [] });
// ^ => { success: true, data: [] }
```

but empty arrays aren't valid input for `streamText` and other operations. Passing those as messages for `streamText` for example will throw

`InvalidPromptError [AI_InvalidPromptError]: Invalid prompt: messages must not be empty`

### AI SDK Version

- ai: 5.0.86

### Code of Conduct

AVtheking:
## Background

The `validateUIMessages` and `safeValidateUIMessages` functions don't properly check for arrays being not-empty,

## Summary

Updated the zod schema to catch empty arrays passed in messages.

## Manual Verification

Wrote tests with empty arrays throwing errors, this should be sufficient right ?

## Checklist

## Related Issues
Fixes #10003


gr2m (commented):
I wonder if this is worth validating with `zod`, or if we should update this check here (or add a new check)

https://github.com/vercel/ai/blob/81d43085f6adfb463458c980b10c0f0e118db824/packages/ai/src/ui/validate-ui-messages.ts#L324

AVtheking:
> I wonder if this is worth validating with `zod`, or if we should update this check here (or add a new check)
> 
> https://github.com/vercel/ai/blob/81d43085f6adfb463458c980b10c0f0e118db824/packages/ai/src/ui/validate-ui-messages.ts#L324

Yeah could update here also, but I thought zod is already being used for validation so it should be good.


Also this null check would also be caught by zod isn't  it ? I assume reason being checking this separately is to provide clear error message, which is for non empty array is provided by zod

gr2m:
you know what let's just go ahead. If for some reason a dedicated check will be better we can change that later

gr2m (approved):

PR #10097, opened 2025-11-09
docs(tools): minor grammar fix

neallseth:
## Background

https://www.merriam-webster.com/grammar/when-to-use-its-vs-its

## Summary

it's -> its


gr2m (approved):
Indeed, thank you for taking the time to send a pull request 🖤

PR #10158, opened 2025-11-11
fix(provider/anthropic): retain user supplied beta headers

### LINKED ISSUE #10018: [Bug] Anthropic provider: custom anthropic-beta header dropped due to overwrite behavior
## Description

While upgrading to AI SDK v6 beta, we noticed that our custom `anthropic-beta` header (`context-1m-2025-08-07`) is unexpectedly being removed when making requests to Anthropic.

After some investigation, we believe this behavior is caused by how betas are inferred and how headers are combined internally.

## Findings

`AnthropicMessagesLanguageModel` automatically infers beta headers:
https://github.com/vercel/ai/blob/e81d017643008eea7e44938e31f032e3bbc9cb3d/packages/anthropic/src/anthropic-messages-language-model.ts#L318-L332

Later, the headers are merged using `combineHeaders`:
https://github.com/vercel/ai/blob/e81d017643008eea7e44938e31f032e3bbc9cb3d/packages/anthropic/src/anthropic-messages-language-model.ts#L397-L401

https://github.com/vercel/ai/blob/e81d017643008eea7e44938e31f032e3bbc9cb3d/packages/provider-utils/src/combine-headers.ts#L1-L11

If multiple headers contain the 'anthropic-beta' key, only the last one wins.
However, according to Anthropic’s API, multiple beta flags should be comma-separated.

Recently, a new inferred beta was introduced: `fine-grained-tool-streaming-2025-05-14`
https://github.com/vercel/ai/pull/9853

As a result, our explicitly passed beta (e.g. `context-1m-2025-08-07`) gets overwritten.

## Expected Behavior

The docs mention that we can pass custom headers via `createAnthropic`:
https://ai-sdk.dev/providers/ai-sdk-providers/anthropic#provider-instance

We expect this usage to work:

```typescript
const anthropic = createAnthropic({
  headers: {
    'anthropic-beta': 'context-1m-2025-08-07',
  },
});
```

The resulting request should include both beta flags merged correctly, e.g.:

```
anthropic-beta: context-1m-2025-08-07,fine-grained-tool-streaming-2025-05-14
```

## Summary

Custom anthropic-beta headers are currently dropped when inferred betas are also present.
We suggest supporting comma-based merging for this header (similar to Anthropic’s own API expectations).

### AI SDK Version

_No response_

### Code of Conduct

chentsulin:
## Background

User-supplied Anthropic beta headers were dropped when internal betas were used. Additionally, betas from the request would override automatic betas (see #10018 ).

## Summary

* change order when combining header to prevent user supplied headers from overriding automatic beta headers
* parse and use config and request beta headers

## Checklist

## Future Work

 - Check if there are any other headers that have the same overwriting issue as `anthropic-beta`.

## Related Issues

Fixes #10018


packages/provider-utils/src/combine-headers.ts
gr2m, line 1: hey just quick early feedback as I didn't have time to dive deeper into it yet. The solution feels overly complicated, do you see a way to simplify it? I might be missing how complex the problem is, but it feels like the problem is simple enough and shouldn't need a function with 30 lines of code 🤔

chentsulin: Or we could use TypeScript overloading to extend `combineHeaders` and support the optional parameter:

```typescript
type Header = Record<string, string | undefined>;

type CombineHeadersOptions = {
  mergeableKeys?: string[];
};

// Overload
export function combineHeaders(
  ...headers: Array<Header | undefined>
): Header;

export function combineHeaders(
  headers: Array<Header | undefined>,
  options: CombineHeadersOptions,
): Header;

export function combineHeaders(
  ...args: Array<Header | undefined> | [Array<Header | undefined>, CombineHeadersOptions]
): Header {
  let headers: Array<Header | undefined>;
  let options: CombineHeadersOptions;
  if (Array.isArray(args[0])) {
    headers = args[0];
    options = args[1] || {};
  } else {
    headers = args as Array<Header | undefined>;
    options = {};
  }

  return headers.reduce(
    (combinedHeaders, currentHeaders) => {
      if (!currentHeaders) return combinedHeaders;

      const result = { ...combinedHeaders };
      const mergeableLowerCaseKeys = options.mergeableKeys?.map(key => key.toLowerCase()) || [];

      for (const [key, value] of Object.entries(currentHeaders)) {
        if (value === undefined) continue;

        // Check if the key is mergeable (case-insensitive) and already exists
        if (mergeableLowerCaseKeys.includes(key.toLowerCase()) && result[key]) {
          const existingValues = result[key].split(',').map(val => val.trim());
          const newValues = value.split(',').map(val => val.trim());
          // Combine and deduplicate values
          const allValues = Array.from(new Set([...existingValues, ...newValues]));
          result[key] = allValues.join(',');
        } else {
          result[key] = value;
        }
      }

      return result;
    },
    {},
  ) as Record<string, string | undefined>;
}
```

However, I think it’s difficult to simplify it further for the following reasons:

- It needs to merge headers regardless of casing differences, such as `anthropic-beta` vs `Anthropic-Beta`.
- It needs to accept whitespace after commas, e.g., `anthropic-beta: feature1, feature2, feature3`.
- It needs to deduplicate values when merging, e.g., merging `anthropic-beta: feature1,feature2,feature3` with `anthropic-beta: feature2,feature3`.
- It needs to handle cases where both the header name and header value may be `undefined` in the original type definition.

lgrammel:
Thanks - I will try to simplify this (pls do not modify the branch)

lgrammel (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10597

PR #10170, opened 2025-11-11
fix(packages/openai-compatible): remove special case for single text block

charleslien:
Note that this code was manually coded (not vibe coded)

## Background

There seems to be a special case for a message with a single text block being converted from
```
{
  type: 'user',
  content: [{type: 'text', text: message}],
}
```
to
```
{
  type: 'user',
  content: message,
}
```
I feel like this behavior is actually unexpected, but seems to be there since the first version and untouched since then.

This would be fine for most cases, but it also converts the provider metadata from a `content`-level object to a `message`-level object.
For example:
```
{
  type: 'user',
  content: [{type: 'text', text: message, providerMetadata: myMetadata}],
}
```
becomes
```
{
  type: 'user',
  content: message,
  providerMetadata: myMetadata,
}
```
when these could potentially mean different things. OpenRouter, for example, requires that `cache_control` metadata be on the `content`-level and not the `message`-level.

## Summary

Removed the special case for a single text block.

## Manual Verification

I sent requests using this change and verified that the content type was correct by listening to the request body. Also verified that OpenRouter requests to Anthropic models now cache correctly.

## Checklist

## Future Work

N/A

## Related Issues

N/A


aayush-kapoor (commented):
lot of tests failing - please look into them. 

also i wonder why can't we keep current behaviour and then have special handling for when provider Metadata is passed (or cacheControl) and then we pass that on the content level. 

the problem with this kind of change is we don't want regressions for cases where this might be expected behaviour

PR #6568, opened 2025-06-01
feat (provider/anthropic): json response schema support via tool calls

lgrammel:
## Background

In AI SDK 5, the different object generation modes have been removed. It is now the responsibility of the provider implementation to map JSON response format requirements to what is needed by the provider API, e.g. tool calls.

## Summary

Add JSON response support with schemas through tool use to the Anthropic provider.

## Verification

Run the anthropic object generation examples. 

## Tasks

- [x] implementation generate
   - [x] change
   - [x] test
- [x] implementation stream
   - [x] change
   - [x] test
- [x] warnings
   - [x] existing tools
   - [x] json without schema
- [x] changeset

## Related Issues

Fixes #6510

PR #10274, opened 2025-11-16
feat(groq,google,openrouter,gateway): fix structured output with tools (#10023)

Sahilbhatane:
## Summary

- add an Anthropic-style JSON-tool fallback so Groq/Google/OpenRouter stop rejecting `generateText` + tools + schema
- make the gateway/agent paths consume the JSON tool output so structured responses succeed instead of looping
- extend tests to cover the reproduction scenarios from https://github.com/gabimoncha/ai-sdk-generate-text-with-schema-issue

## Future Work

- Finalize the provider-level design : keep core mode-less, rely on an anthropic-style synthetic JSON tool whenever a schema and tools are requested together, ensure this covers `generateText`, agents and gateway routing
- Groq/Google fallback scaffolding is done and committed on my branch, next I'll audit them alongside new implementations for OpenRouter and the AI gateway so tool calls actually execute and structured output is returned (matching the repro’s failing cases).
- Add/extend tests mirroring gabimoncha’s scenarios (schema+tools for `generateText/agent` in v5 & v6 paths), then run pnpm test to verify.
- Finish with patch changesets and summary update referencing the repo and confirmed test results.

Partially Fixes issue #10023

### DIFF
--- a/packages/google/src/google-generative-ai-language-model.ts
+++ b/packages/google/src/google-generative-ai-language-model.ts
@@ -3,11 +3,11 @@ import {
   LanguageModelV4CallOptions,
   LanguageModelV4Content,
   LanguageModelV4FinishReason,
-  LanguageModelV4FunctionTool,
   LanguageModelV4GenerateResult,
   LanguageModelV4Source,
   LanguageModelV4StreamPart,
   LanguageModelV4StreamResult,
+  JSONObject,
   SharedV4ProviderMetadata,
   SharedV4Warning,
 } from '@ai-sdk/provider';
@@ -18,7 +18,10 @@ import {
   FetchFunction,
   generateId,
   InferSchema,
+  isCustomReasoning,
   lazySchema,
+  mapReasoningToProviderBudget,
+  mapReasoningToProviderEffort,
   parseProviderOptions,
   ParseResult,
   postJsonToApi,
@@ -97,6 +100,7 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
     seed,
     tools,
     toolChoice,
+    reasoning,
     providerOptions,
   }: LanguageModelV4CallOptions) {
     const warnings: SharedV4Warning[] = [];
@@ -136,57 +140,36 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
     }
 
     const isGemmaModel = this.modelId.toLowerCase().startsWith('gemma-');
+    const supportsFunctionResponseParts = this.modelId.startsWith('gemini-3');
 
     const { contents, systemInstruction } = convertToGoogleGenerativeAIMessages(
       prompt,
-      { isGemmaModel, providerOptionsName },
+      {
+        isGemmaModel,
+        providerOptionsName,
+        supportsFunctionResponseParts,
+      },
     );
-
-    const { supportsStructuredOutputWithTools } =
-      getGoogleModelCapabilities(this.modelId);
-    const structuredOutputMode =
-      googleOptions?.structuredOutputMode ?? 'auto';
-    const useStructuredOutputWithTools =
-      structuredOutputMode === 'outputFormat' ||
-      (structuredOutputMode === 'auto' && supportsStructuredOutputWithTools);
-
-    const needsJsonToolFallback =
-      responseFormat?.type === 'json' &&
-      responseFormat.schema != null &&
-      tools != null &&
-      tools.length > 0 &&
-      !useStructuredOutputWithTools;
-
-    const jsonResponseToolName = needsJsonToolFallback
-      ? tools.some(
-          tool => tool.type === 'function' && tool.name === 'json',
-        )
-        ? '__ai_sdk_json_response'
-        : 'json'
-      : undefined;
-
-    const jsonResponseTool: LanguageModelV4FunctionTool | undefined =
-      needsJsonToolFallback && jsonResponseToolName != null
-        ? {
-            type: 'function',
-            name: jsonResponseToolName,
-            description: 'Respond with a JSON object.',
-            inputSchema: responseFormat.schema,
-          }
-        : undefined;
 
     const {
       tools: googleTools,
       toolConfig: googleToolConfig,
       toolWarnings,
     } = prepareTools({
-      tools:
-        jsonResponseTool != null
-          ? [...(tools ?? []), jsonResponseTool]
-          : tools,
-      toolChoice: jsonResponseTool != null ? { type: 'required' } : toolChoice,
+      tools,
+      toolChoice,
       modelId: this.modelId,
     });
+
+    const resolvedThinking = resolveThinkingConfig({
+      reasoning,
+      modelId: this.modelId,
+      warnings,
+    });
+    const thinkingConfig =
+      googleOptions?.thinkingConfig || resolvedThinking
+        ? { ...resolvedThinking, ...googleOptions?.thinkingConfig }
+        : undefined;
 
     return {
       args: {
@@ -203,13 +186,10 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
 
           // response format:
           responseMimeType:
-            responseFormat?.type === 'json' && jsonResponseTool == null
-              ? 'application/json'
-              : undefined,
+            responseFormat?.type === 'json' ? 'application/json' : undefined,
           responseSchema:
             responseFormat?.type === 'json' &&
             responseFormat.schema != null &&
-            jsonResponseTool == null &&
             // Google GenAI does not support all OpenAPI Schema features,
             // so this is needed as an escape hatch:
             // TODO convert into provider option
@@ -222,7 +202,7 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
 
           // provider options:
           responseModalities: googleOptions?.responseModalities,
-          thinkingConfig: googleOptions?.thinkingConfig,
+          thinkingConfig,
           ...(googleOptions?.mediaResolution && {
             mediaResolution: googleOptions.mediaResolution,
           }),
@@ -242,24 +222,17 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
           : googleToolConfig,
         cachedContent: googleOptions?.cachedContent,
         labels: googleOptions?.labels,
+        serviceTier: googleOptions?.serviceTier,
       },
       warnings: [...warnings, ...toolWarnings],
       providerOptionsName,
-      usesJsonResponseTool: jsonResponseTool != null,
-      jsonResponseToolName,
     };
   }
 
   async doGenerate(
     options: LanguageModelV4CallOptions,
   ): Promise<LanguageModelV4GenerateResult> {
-    const {
-      args,
-      warnings,
-      providerOptionsName,
-      usesJsonResponseTool,
-      jsonResponseToolName,
-    } = await this.getArgs(options);
+    const { args, warnings, providerOptionsName } = await this.getArgs(options);
 
     const mergedHeaders = combineHeaders(
       await resolve(this.config.headers),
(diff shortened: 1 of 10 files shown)


packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -118,13 +119,29 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
       { isGemmaModel },
     );
 
+    const jsonResponseTool: LanguageModelV3FunctionTool | undefined =
+      responseFormat?.type === 'json' &&
+      responseFormat.schema != null &&
+      tools != null &&
+      tools.length > 0
```
AVtheking: I think it's better to have one approach, may be use json response tool for structured output only ?

AVtheking (commented):
Could default to jsonResponseTool for structured output

Sahilbhatane:
@AVtheking done defaulting to jsonResponseTool for structured output

elierotenberg:
Hi,
Any update on this PR? This is very blocking for me and I'm happy to help.

n-siddarth:
Just following up here. Is there anything left to implement before this can be merged?

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -122,13 +123,29 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
...
+        responseFormat.schema != null &&
+        tools != null &&
+        tools.length > 0
+        ? {
+          type: 'function',
+          name: 'json',
+          description: 'Respond with a JSON object.',
+          inputSchema: responseFormat.schema,
+        }
+        : undefined;
```
felixarntz: I think it's problematic that the tool is injected unconditionally here. While this is in line with how it _used_ to work for the Anthropic provider, that was only the case because Anthropic did not have any support whatsoever for structured JSON output.

That has of course changed since, please take a look at #10502 for reference.

Both Google and Groq support structured output natively. It might be the case that it's not supported by all models or under all circumstances, but we need to ensure the tool workaround here is only used when it's actually needed. Otherwise it may cause unexpected behavior.

cc @lgrammel since you worked on the above PR and also the original idea in #6568

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -122,13 +123,29 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
       { isGemmaModel },
     );
 
+    const jsonResponseTool: LanguageModelV3FunctionTool | undefined =
+      responseFormat?.type === 'json' &&
+        responseFormat.schema != null &&
+        tools != null &&
+        tools.length > 0
```
felixarntz: Why do we only add this when other tools are provided?

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -122,13 +123,29 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
...
       toolConfig: googleToolConfig,
       toolWarnings,
     } = prepareTools({
-      tools,
-      toolChoice,
+      tools:
+        jsonResponseTool != null
+          ? [...(tools ?? []), jsonResponseTool]
+          : tools,
+      toolChoice: jsonResponseTool != null ? { type: 'required' } : toolChoice,
```
felixarntz, line 199: See above, for cases where structured outputs are actually supported, forcing a required tool call may throw a wrench in things.

examples/nuxt-openai/nuxt.config.ts
```diff
@@ -4,7 +4,7 @@ export default defineNuxtConfig({
   modules: ['@nuxtjs/tailwindcss'],
 
   nitro: {
-    preset: 'vercel-edge', // you can use 'vercel' or other providers here
+    preset: 'vercel', // you can use 'vercel' or other providers here
```
felixarntz: this seems unrelated?

.changeset/groq-structured-output-tools.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ai-sdk/groq": patch
+---
+
+Support structured output when combining with tool calling by injecting synthetic JSON function tool
```
felixarntz: Please include only a single changeset file that includes both `patch` annotations for the packages.

Ideally you use `pnpm changeset`.

felixarntz (changes requested):
@Sahilbhatane Thank you for implementing the PR, and apologies for the delayed review.

I think the idea here is reasonable, but it's currently not implemented as a fallback as far as I can tell. This will need some revision.

The relevant logic in the Anthropic provider has also been updated since you opened this PR, so it might be helpful to take another look.

.changeset/groq-structured-output-tools.md
Sahilbhatane: Understood.
I’ll replace the current changesets with a single combined changeset file containing both patch entries, generated via `pnpm changeset`

examples/nuxt-openai/nuxt.config.ts
Sahilbhatane: I’ll drop it from this PR

packages/google/src/google-generative-ai-language-model.ts
Sahilbhatane: I’ll avoid forcing required tool choice unless we are in an explicit fallback mode where that behavior is necessary, native structured-output paths should keep their normal tool-choice behavior.

packages/google/src/google-generative-ai-language-model.ts
Sahilbhatane: The intent was to cover the schema+tools conflict path, but I agree the condition and behavior need to be clearer and tied to capability detection rather than a broad tools-length check. I’ll update the logic and add tests that document the exact activation criteria.

packages/google/src/google-generative-ai-language-model.ts
Sahilbhatane: yes it is unconditional, I can gate fallback activation behind explicit compatibility checks so native structured output remains the default where supported, and only use synthetic JSON-tool mode when required by provider/model constraints.

Sahilbhatane:
Thanks for review @felixarntz,
This was partial implementation and with the review i can rework it so the synthetic JSON tool is used strictly as a fallback path only, not as the default behavior.
and:
- align with the current Anthropic/provider pattern referenced in #10502/#6568
- keep native structured output as primary path for Google/Groq/OpenRouter
- only activate fallback for model/provider combinations where schema+tools cannot be handled natively remove the unrelated Nuxt config change
- regenerate changesets into a single file via pnpm changeset

I’ll push an updated revision and tag you for re-review.

Sahilbhatane:
hey @felixarntz, after the above request changes, I did tried this approach - Instead of guessing automatically, expose a provider option that lets the user explicitly choose the strategy.

`structuredOutputMode` supports:
1. `auto/native` (default)
Use native structured output (responseSchema / json_schema) — identical to current behavior.
2. `jsonTool`
Replace the response format with the synthetic JSON tool. This is intended for models that reject schema+tools.

But I still have multiple open question and other 2 approaches that i didn't specific choose but still want to confirm with maintainers.

I can give a detailed comment here but is there a better communication channel like discord or slack or community (i can create new topic there), etc. 

#### not pushed this approach, but all test passes, still - From my perspective, this approach is the strongest choice because:

- It doesn't break any existing behavior (zero risk regression)
- It mirrors what the Anthropic provider does for the tool mechanics, so the patterns are consistent across the SDK
The isJsonResponseFromTool dynamic flag (computed from actual response, not config) correctly handles edge cases that a static flag misses — this is the same pattern Anthropic uses
- The Groq streaming restructure (deltas before finish-reason) fixes a real timing bug regardless of which approach is chosen
- The weakest point is that auto doesn't actually auto-detect, which means users still get the original #10023 error by default and need to discover the structuredOutputMode: 'jsonTool' option. 

That's why i have some questions that will fix this (may fix this).

felixarntz:
@Sahilbhatane Ideally we align with the approach from #10502. What you're proposing is close to it, but not quite.

I think it's crucial that we're aware here which models support _actual_ structured output vs which do not - unless an entire provider doesn't support it.

While normally I would advise against specific model support maps like this, we're only building this JSON tool as a workaround, and in a perfect world it would never exist. Requiring the user to be aware of this low-level nuance would not be great DX, and that's why in the Anthropic provider there's also the awareness for whether a model supports structured output or not.

Please feel free to iterate on this PR. Also note this needs a refresh from latest `main`, as there are currently merge conflicts.

Sahilbhatane:
@felixarntz updated based on your review feedback. here is what changed:

#### Approach:

- The synthetic JSON tool is now used strictly as a fallback, not as the default behavior. Native structured output remains the primary path.
- Added `structuredOutputMode` provider option (auto , `outputFormat` , `jsonTool`) for both Google and Groq, aligning with the Anthropic pattern from #10502.
- `getGoogleModelCapabilities()` / `getGroqModelCapabilities()` provide the model-awareness layer. Currently all models return `supportsStructuredOutputWithTools`: false since no Google/Groq model is documented to support `schema+tools` together. When models add support, these functions can be updated with model-specific checks (like Anthropic's `getModelCapabilities()`).

#### Activation logic: The JSON tool is injected only when ALL of these are true:

1. `responseFormat.type === 'json'` with a schema
Tools are present
2. The model does NOT support native `schema+tools` (per capabilities + `structuredOutputMode`)
3. When the JSON tool is NOT needed, the request is unchanged from current behavior (zero regression risk).

#### Bug fixes included:

Fixed a streaming bug in both providers where return inside the transform function would exit the entire chunk processing (skipping tool call conversion and finish reason), changed to continue (Google) / conditional wrapper (Groq).

Tests: 287 Google tests passing, 84 Groq tests passing. Two test cases added (auto fallback + `outputForma`t override) per provider. Happy to add more test coverage for streaming, explicit `jsonTool` mode, and response mapping if desired.

*and* : OpenRouter and AI Gateway (from the original issue #10023) are not addressed in this PR, just because they were out of scope

I also did AI review on change code for any fallbacks or error with cursor.

packages/groq/src/groq-chat-language-model.ts
```diff
@@ -435,11 +483,22 @@ export class GroqChatLanguageModel implements LanguageModelV4 {
...
+                      type: 'tool-input-start',
+                      id: toolCallDelta.id,
+                      toolName: toolCallDelta.function.name,
+                    });
+                  } else {
+                    if (!isActiveText) {
+                      controller.enqueue({ type: 'text-start', id: 'txt-0' });
+                      isActiveText = true;
+                    }
+                  }
```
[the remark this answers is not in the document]
Sahilbhatane: Addressed. If a user tool is already named `json`, the synthetic tool is named `__ai_sdk_json_response` and response parsing uses `jsonResponseToolName` everywhere. Covered by should use reserved synthetic tool name when user already defines a `json` tool (Groq and same for Google).

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -141,13 +142,41 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV4 {
...
+
+    const jsonResponseTool: LanguageModelV4FunctionTool | undefined =
+      needsJsonToolFallback
+        ? {
+            type: 'function',
+            name: 'json',
+            description: 'Respond with a JSON object.',
+            inputSchema: responseFormat.schema,
+          }
+        : undefined;
```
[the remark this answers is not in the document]
Sahilbhatane: same as groq addressed above

packages/groq/src/groq-chat-language-model.ts
```diff
@@ -103,11 +104,42 @@ export class GroqChatLanguageModel implements LanguageModelV4 {
...
+
+    const jsonResponseTool: LanguageModelV4FunctionTool | undefined =
+      needsJsonToolFallback
+        ? {
+            type: 'function',
+            name: 'json',
+            description: 'Respond with a JSON object.',
+            inputSchema: responseFormat.schema,
+          }
+        : undefined;
```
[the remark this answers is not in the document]
Sahilbhatane: done, picked a reserved unique name `__ai_sdk_json_response`

PR #10281, opened 2025-11-16
feat: image editing

### LINKED ISSUE #8719: Image Editing
### Research

image editing types:

- **Outpainting** (“image expansion”, “uncrop”) — extend canvas beyond bounds; usually via prompt, sometimes prompt-free. 
- **Inpainting** (with/without prompt & mask/overlay) — edit inside a region; classic mask-based “generative fill.” 
- **Upscaling** / Super-resolution — increase resolution with minimal changes (ESRGAN, etc.). 
- **Variations** — small changes around a given image (sometimes via “image reference” or adapters like FLUX Redux). 
- **Restyle** (img2img) — preserve content/layout, change style via a prompt; many APIs call this “image-to-image.” 
- **Masked restyle** / local edits — prompt-guided changes bound to a region, or “local editing” without an explicit binary mask (e.g., FLUX Kontext). 
- **Background remove/replace** (RMBG / background swap). 
- **Object remove/add** (“gen-fill” / “erase” / “insert”) — targeted content edits. 
- **Control-guided edits** (ControlNet-style) — pose, depth, edge, scribble guidance. (Varies by host; often listed as “tools / adapters”.) 
- **Text editing** in images — replace text content cleanly. 

providers

- amazon-bedrock
- deepinfra
- fal
- fireworks
- google-vertex
- openai
- replicate
- togetherai
- also: [alibaba](https://x.com/_akhaliq/status/1970192728282505295)

gr2m:
## Background

As part of our [v6 release](https://github.com/vercel/ai/issues/8662), we want to expand our APIs to support image editing and other image-to-image operations.

We were discussing separate approaches

1. introduce a new API: `editImage()`
2. update the API of the existing `generateImage()`

After exploring both approached and researching the APIs for image editing across our providers, I recommend that we update `generateImage()` because of these reasons

1. Will will most likely consolidate our APIs further. In a future version of the `ai` SDK we might just have two methods: `stream()` and `generate()`, instead of its variants like `generateText()`, `generateObject` (already deprecated), and `generateImage()`. We might add APIs to generate audio and videos as well.
2. Editing an image always means a new image (or multiple new images) are being generated. But instead of text prompt, reference image(s) are provided.
3. There are a lot of different operations for image editing (inpaint, outpaint, scale, replace, remove background, variations, restyle, edits with masks, etc). But most of these are controlled by prompts. There is no need to introduce a big new API surface in order to support the most common use cases of image editing, and for edge cases provider implementations can utilize `providerOptions`

## Summary

Update `generateImage({ prompt })` argument to be an object in order to support additional arguments required for image editing / image-to-image operations.

Example

```ts
const { images } = await generateImage({
  model: blackForestLabs.image('flux-2-pro'),
  prompt: {
    text: 'A baby elephant with a shirt that has the logo from input image 1. Do not change the text of the logo.',
    images: [
      'https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png',
    ],
  },
});
```

`images` items can be one of

- url
- base64-encoded string
- a Uint8Array
- an ArrayBuffer
- a Buffer.

This pull request also promotes `experimental_generateImage` to the stable `generateImage` and deprecates `experimental_generateImage` (will be removed in a future version)

## Manual Verification

I tested all examples

## Progress

- [x] Provider API research
- [x] API Design
- [x] update spec and `generateImage()` implementation
- [x] [openai](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/openai/src/image/openai-image-model.ts)
- [x] [black-forest-labs](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/black-forest-labs/src/black-forest-labs-image-model.ts)
- [x] [google-vertex](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/google-vertex/src/google-vertex-image-model.ts)
- [x] [google](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/google/src/google-generative-ai-image-model.ts)
- [x] [openai-compatible](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/openai-compatible/src/image/openai-compatible-image-model.ts)
- [x] [amazon-bedrock](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/amazon-bedrock/src/bedrock-image-model.ts)
- [x] [deepinfra](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/deepinfra/src/deepinfra-image-model.ts)
- [x] [fal](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/fal/src/fal-image-model.ts)
- [x] [fireworks](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/fireworks/src/fireworks-image-model.ts)
- [x] [luma](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/luma/src/luma-image-model.ts)
- [x] [replicate](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/replicate/src/replicate-image-model.ts)
- [x] [togetherai](https://github.com/vercel/ai/blob/d91969f2cd2afc0d96e608ecc911e14b3467e2f2/packages/togetherai/src/togetherai-image-model.ts)
- [x] clean up examples, verify they all still work
- [x] clean up implementation
- [x] documentation

## Checklist

## Learnings

- Amazon Bedrock: edits fail with "Invalid Input: The input does not adhere to the expected standards. Please refer to the model user guide and adjust the input before trying again." when using mask image that is not black and white only. Also unlike any other provider, white is the color to mark what is to be preserved, and black is the color to mask where to replace.
- luma does not support base64 image inputs, only URLs.
- luma has a feature I haven't seen from others: style reference (https://docs.lumalabs.ai/docs/image-generation#style-reference) and character reference (https://docs.lumalabs.ai/docs/image-generation#character-reference)

<a name="future-work"></a>
## Future Work

The `generateImage()` method is only working with dedicated image models and provider APIs. But increasingly there are also multimodal AI models such as Google's Nano Banana (e.g. `gemini-2.5-flash-image-preview`). Naturally, people try to use it with `generateImage()` but it won't work. It only works with `generateText()` ([example](https://github.com/vercel/ai/blob/db6051bf93c7e9b26680cd86b3aeb7c01b6e1220/examples/ai-core/src/generate-image/google-gemini-minimal.ts)).

We are aware of the confusion this causes and are thinking about how we can abstract away the differences of dedicated image models and multimodal models for the purpose of image generation.

Other things

- [x] luma: deprecate old way of editing images, see `examples/ai-core/src/generate-image/luma-character-reference.ts`, `examples/ai-core/src/generate-image/luma-image-reference.ts`, `examples/ai-core/src/generate-image/luma-modify-image.ts`, `examples/ai-core/src/generate-image/luma-style-reference.ts`
- [x] fal: deprecate `imageUrl` provider option
- [x] bfl: deprecate `providerOptions.bfl.inputImage*` in favor of `prompt.images`
- [x] deepinfra: [examples/ai-core/src/generate-image/deepinfra-edit-multi-image.ts](https://github.com/vercel/ai/blob/gr2m/8719-image-editing/examples/ai-core/src/generate-image/deepinfra-edit-multi-image.ts) started failing, needs investigating
- [ ] fireworks: using `generateImage()` with `accounts/fireworks/models/flux-kontext-pro` and `accounts/fireworks/models/flux-kontext-max` models fails, because for these two models, the API responds with task ID that needs to be pulled until the image was generated, see https://docs.bfl.ml/kontext/kontext_text_to_image#poll-for-result. See https://github.com/vercel/ai/issues/11149
- [ ] fireworks + togetherai: do not simply pass through all provider options. Type all supported options as `camelCase`, then map them 1:1 explicitly. Deprecate all known arguments that have an `_` to avoid a breaking change.
- [x] replicate: APIs differentiate based on used models, e.g. compare https://replicate.com/black-forest-labs/flux-fill-pro/api/schema vs https://replicate.com/black-forest-labs/flux-2-pro/api/schema

## Related Issues

closes #8719


content/docs/07-reference/01-ai-sdk-core/10-generate-image.mdx
```diff
@@ -45,7 +45,7 @@ console.log(images);
     },
     {
       name: 'prompt',
-      type: 'string',
+      type: 'GenerateImagePrompt',
```
gr2m: @nicoalbanese this type is defined in here:
https://github.com/vercel/ai/pull/10281/changes#diff-d2fc3ae8af9e7e580348e659e3a620bd1e7edbb1cfce40d6985da3389d5a5938R32

and exported from the `ai` package. Will that just work or do I need to do anything else?

nicoalbanese: https://github.com/vercel/ai/blob/main/content/docs/07-reference/01-ai-sdk-core/02-stream-text.mdx?plain=1#L455

Unfortunately no, you will have to add the type manually

gr2m: could you double check if that looks good now? 199fbe015

content/providers/01-ai-sdk-providers/10-fal.mdx
```diff
@@ -146,15 +146,27 @@ Key features of Fal models include:
...
+  prompt: {
+    text: 'Put a donut next to the flour.',
+    images: [
+      'https://v3.fal.media/files/rabbit/rmgBxhwGYb2d3pl3x9sKf_output.png',
+    ],
+  },
+});
+```
+
+Images cann also be passed as base64-encoded string, a `Uint8Array`, an `ArrayBuffer`, or a `Buffer`.
```
gr2m:
```suggestion
Images can also be passed as base64-encoded string, a `Uint8Array`, an `ArrayBuffer`, or a `Buffer`.
```

content/providers/01-ai-sdk-providers/03-openai.mdx
```diff
@@ -1527,6 +1527,107 @@ const model = openai.image('dall-e-3');
   parameter instead.
 </Note>
 
+### Image Editing
+
+OpenAI's `gpt-image-1` model supports powerful image editing capabilities. Pass input images via `prompt.images` to transform, combine, or edit existing images.
```
gr2m, line 1532: update to `gpt-image-1.5` once available through API
https://platform.openai.com/docs/models/gpt-image-1.5

content/providers/01-ai-sdk-providers/60-replicate.mdx
```diff
@@ -169,4 +169,80 @@ const { image } = await generateImage({
...
+### Provider Options
+
+Common provider options for image generation:
+
+- **guidance_scale** _number_ - Guidance scale for classifier-free guidance. Higher values make the output more closely match the prompt.
+- **num_inference_steps** _number_ - Number of denoising steps. More steps = higher quality but slower.
+- **negative_prompt** _string_ - Negative prompt to guide what to avoid in the generation.
+- **output_format** _'png' | 'jpg' | 'webp'_ - Output image format.
+- **output_quality** _number (1-100)_ - Output image quality. Only applies to jpg and webp.
+- **strength** _number (0-1)_ - Strength of the transformation for img2img. Lower values keep more of the original image.
```
gr2m, line 238: we currently just pass through provider options
https://github.com/vercel/ai/blob/95b7560e102f49fe58f06b48a928a8656ea1a1aa/packages/replicate/src/replicate-image-model.ts#L117

Correct way would be to type all supported arguments as camelCase and map them explicitly, but it would be a bigger effort if we don't want to break current usage. I'd do it in a follow up and deprecate the supported underscore arguments in v6, remove in v7

packages/amazon-bedrock/src/bedrock-image-model.ts
```diff
@@ -124,8 +264,41 @@ export class BedrockImageModel implements ImageModelV3 {
...
+  // Normal successful response
+  images: z.array(z.string()).optional(),
+
+  // Moderation response fields
+  id: z.string().optional(),
+  status: z.string().optional(),
+  result: z.unknown().optional(),
+  progress: z.unknown().optional(),
+  details: z.record(z.string(), z.unknown()).optional(),
+  preview: z.unknown().optional(),
```
gr2m, line 296: ran into a moderation response while testing

packages/fal/src/fal-image-model.ts
```diff
@@ -70,6 +76,26 @@ export class FalImageModel implements ImageModelV3 {
       num_images: n,
     };
 
+    // Handle image editing: convert files to image_url
+    if (files != null && files.length > 0) {
+      // Use first file as the primary image_url
+      requestBody.image_url = await fileToDataUri(files[0]);
+
+      // Additional images can be passed via provider options if needed
```
gr2m:
```suggestion
```

packages/openai/src/image/openai-image-model.ts
```diff
@@ -63,6 +69,81 @@ export class OpenAIImageModel implements ImageModelV3 {
     }
 
     const currentDate = this.config._internal?.currentDate?.() ?? new Date();
+
+    if (files != null) {
+      const { value: response, responseHeaders } = await postFormDataToApi({
+        url: this.config.url({
+          // TODO: Consider adding support ofr /images/variations API. It's DALL·E 2 only, though. The response is different as well, it would be quite some work.
```
dancer:
```suggestion
          // TODO: Consider adding support of /images/variations API. It's DALL·E 2 only, though. The response is different as well, it would be quite some work.
```

packages/togetherai/src/togetherai-image-model.ts
```diff
@@ -40,11 +45,21 @@ export class TogetherAIImageModel implements ImageModelV3 {
...
 
+    if (mask != null) {
+      throw new Error(
+        'Together AI does not support mask-based image editing. ' +
+          'Use FLUX Kontext models (e.g., black-forest-labs/FLUX.1-kontext-pro) ' +
+          'with a reference image and descriptive prompt instead.',
+      );
+    }
+
     if (size != null) {
```
dancer: should this be `if (aspectRatio != null) {`

packages/provider-utils/src/convert-image-model-file-to-data-uri.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  // file.type === 'file' - data can be base64 string or Uint8Array
+  if (typeof file.data === 'string') {
+    // Already base64 encoded
+    return `data:${file.mediaType};base64,${file.data}`;
+  }
+
+  // Uint8Array - convert to base64
+  const base64 = convertUint8ArrayToBase64(file.data);
+  return `data:${file.mediaType};base64,${base64}`;
```
lgrammel: can be collapsed into ternary:

```
return `data:${file.mediaType};base64,${
  typeof file.data === 'string' ? file.data :  convertUint8ArrayToBase64(file.data)
}`;
```

packages/provider/src/image-model/v3/image-model-v3-file.ts
```diff
@@ -0,0 +1,44 @@
...
+       * as base64 encoded strings. If the API returns binary data, the file data should
+       * be returned as binary data.
+       */
+      data: string | Uint8Array;
+
+      /**
+       * Optional provider-specific metadata for the file part.
+       */
+      providerOptions?: SharedV3ProviderMetadata;
+    }
```
lgrammel, line 31: this match `LanguageModelV3File`.

i feel there is potential for unification here, and moving things to shared. this has the benefit that the provider-utils code would also be reusable and we would reduce api surface

packages/provider/src/image-model/v3/image-model-v3-file.ts
```diff
@@ -0,0 +1,44 @@
...
+      /**
+       * The URL of the image file.
+       */
+      url: string;
+
+      /**
+       * Optional provider-specific metadata for the file part.
+       */
+      providerOptions?: SharedV3ProviderMetadata;
+    };
```
lgrammel, line 44: something like this could make sense in LanguageModelV3 as well (vs using `Url` type). would prefer to standardize, e.g. by changing LanguageModelV3 to match

packages/provider/src/image-model/v3/image-model-v3-file.ts
gr2m: the difference is: `LanguageModelV3File` is output, while ImageModelV3File is `input`. It's `providerMetadata` vs `providerOptions`. I had the same thought as you but didn't know how because of this slight difference

lgrammel: you are right. we have the following input that we should standardize:

image

lgrammel: but given `filename` and `LanguageModelV3DataContent` (which supports url and data) maybe not worth it

lgrammel: image

lgrammel: just feel we have 2 almost similar yet different abtractions for the same thing, which leads to code duplication downstream

lgrammel: `SharedFileInputV3` could be an option, unclear how to deal with `filename` though

lgrammel:
```ts
type LanguageModelV3FilePart = SharedFileInputV3 & {
  filename?: string
}
```

would be an option

packages/provider-utils/src/convert-image-model-file-to-data-uri.ts
gr2m: a1fceada9

packages/provider/src/image-model/v3/image-model-v3-file.ts
gr2m: you mean adding a new dedicated URL-type like `LanguageModelV3UrlFilePart` and removing `URL` from the `LanguageModelV3DataContent` type?

packages/ai/src/error/index.ts
```diff
@@ -28,5 +28,5 @@ export { UnsupportedModelVersionError } from './unsupported-model-version-error'
 export { InvalidDataContentError } from '../prompt/invalid-data-content-error';
 export { InvalidMessageRoleError } from '../prompt/invalid-message-role-error';
 export { MessageConversionError } from '../prompt/message-conversion-error';
-export { DownloadError } from '../util/download/download-error';
+export { DownloadError } from '@ai-sdk/provider-utils';
```
gr2m, line 31: move to top

packages/provider/src/image-model/v3/image-model-v3-file.ts
gr2m: I'll create a follow up issue for v4 spec, for a shared File Spec

lgrammel (approved):

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -279,3 +301,36 @@ async function invokeModelMaxImagesPerCall(model: ImageModelV3) {
...
+      url: dataContent,
+    };
+  }
+
+  return {
+    type: 'file',
+    data: convertDataContentToUint8Array(dataContent),
+    mediaType:
+      detectMediaType({
+        data: convertDataContentToUint8Array(dataContent),
```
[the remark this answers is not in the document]
gr2m: wow, good catch, thank you 😅

PR #10440, opened 2025-11-20
feat(provider/gateway): Change default maxImagesPerCall per-provider

R-Taneja:
## Background
I discovered that passing in `n` to set the number of images to `generateImage` doesn't work with Black Forest Labs unless `maxImagesPerCall` is explicitly set to 1, because that's the max number of images BFL returns per request. It would return 1 image every time regardless of `n`. Users could solve this by explicitly setting `maxImagesPerCall: 1` themselves but the behavior is confusing. We previously defaulted to 4 because that's the value for Vertex Imagen. We need a new default for other providers.

## Summary
Added a mapping of provider to default `maxImagesPerCall` and adjust the default based on the model slug

## Manual Verification
`n` now works

## Checklist

### DIFF
--- a/.changeset/little-tomatoes-attack.md
+++ b/.changeset/little-tomatoes-attack.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+feat(provider/gateway): Change default maxImagesPerCall per-provider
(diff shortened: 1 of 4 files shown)


gr2m (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10493

PR #10463, opened 2025-11-21
feat(mcp): preserve MCP tool annotations including readOnlyHint

tonypan2:
## Summary

Preserve all MCP tool annotations (readOnlyHint, destructiveHint, idempotentHint, openWorldHint) when converting MCP tools to AI SDK tools.

## Changes

Previously, only the `title` annotation was extracted and preserved from MCP servers. This PR adds support for passing through all five annotation fields defined in the MCP specification (2025-03-26):

- **`title`** - Human-readable tool title
- **`readOnlyHint`** - Tool doesn't modify its environment  
- **`destructiveHint`** - Tool may perform destructive updates
- **`idempotentHint`** - Repeated calls with same args have no additional effect
- **`openWorldHint`** - Tool interacts with external entities

### Implementation

1. Added `annotations` field to `Tool` type definition in `@ai-sdk/provider-utils`
2. Updated `dynamicTool()` function signature to accept annotations parameter
3. Pass annotations from MCP `listTools` response to tool creation in MCP client

### Use Cases

These annotations enable clients to:
- Filter tools by behavior characteristics (e.g., only allow read-only tools in production)
- Show UI warnings before executing destructive operations
- Implement smart retry logic for idempotent tools
- Track/log external system interactions

### Security Note

Per the MCP specification, these annotations are **informational hints only** and should not be used for security-critical decisions. The documentation explicitly states this caveat.

## Testing

- ✅ Existing tests pass
- ✅ TypeScript types are properly defined
- ✅ Backward compatible (all fields are optional)

## Related

- MCP Specification: https://modelcontextprotocol.io/specification/2025-06-18
- Branch follows MCP spec version 2025-03-26 which introduced tool annotations

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.changeset/strange-suits-lick.md
+++ b/.changeset/strange-suits-lick.md
@@ -1,0 +1,8 @@
+---
+'@ai-sdk/provider-utils': patch
+'@ai-sdk/mcp': patch
+---
+
+feat(mcp): preserve MCP tool annotations including readOnlyHint
+
+Preserve all MCP tool annotations (readOnlyHint, destructiveHint, idempotentHint, openWorldHint) when converting MCP tools to AI SDK tools. These annotations are informational hints that enable clients to make informed decisions about tool behavior.
--- a/packages/mcp/src/tool/mcp-client.ts
+++ b/packages/mcp/src/tool/mcp-client.ts
@@ -501,12 +501,14 @@ class DefaultMCPClient implements MCPClient {
                   additionalProperties: false,
                 } as JSONSchema7),
                 execute,
+                annotations,
               })
             : tool({
                 description,
                 title,
                 inputSchema: schemas[name].inputSchema,
                 execute,
+                annotations,
               });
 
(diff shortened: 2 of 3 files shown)


dvoytenko (approved):

lgrammel:
a) I am not sure how much we should tie MCP to our more abstract tools. This PR bleeds the MCP spec into our tool spec
b) if we want to proceed in the current form, tests and documentation would be good to have

For your use case, is `readOnlyHint` the main thing you need? If so, we could add that one specifically to our spec because it's usefulness is clear. The other hints are more MCP specific imo.

If that is potential path, then tools could have a `metadata` or `hints` property.

nicoalbanese:
Agree with @lgrammel and feel strongly that this would be best implemented in some kind of generic tool metadata. This might require some further thinking though

tonypan2:
closing as this needs more thinking

PR #10478, opened 2025-11-22
fix(package/ai): fix Gateway image cost reporting bug

R-Taneja:
## Background
I discovered a bug in the cost reporting for images generated via Gateway. If n > 1, the cost in the provider metadata would be reported as if n = 1 (but the correct amount would be debited). This is because the cost fields in the provider metadata were being merged without being summed across requests (since maxImagesPerCall = 1)

## Summary
Merged with summation

## Manual Verification
Ran Gateway requests and cost is now reported accurately

## Checklist

### DIFF
--- a/packages/ai/src/generate-image/generate-image.ts
+++ b/packages/ai/src/generate-image/generate-image.ts
@@ -200,10 +200,25 @@ Only applicable for HTTP-based providers.
         if (providerName === 'gateway') {
           const currentEntry = providerMetadata[providerName];
           if (currentEntry != null && typeof currentEntry === 'object') {
-            providerMetadata[providerName] = {
+            const merged = {
               ...(currentEntry as object),
               ...metadata,
-            } as ImageModelV3ProviderMetadata[string];
+            } as Record<string, unknown>;
+
+            for (const key of ['cost', 'marketCost']) {
+              const previousValue = (currentEntry as Record<string, unknown>)[
+                key
+              ];
+              const newValue = (metadata as Record<string, unknown>)[key];
+              const sum = addDecimal(
+                typeof previousValue === 'string' ? previousValue : undefined,
+                typeof newValue === 'string' ? newValue : undefined,
+              );
+              if (sum) merged[key] = sum;
+            }
+
+            providerMetadata[providerName] =
+              merged as ImageModelV3ProviderMetadata[string];
           } else {
             providerMetadata[providerName] =
               metadata as ImageModelV3ProviderMetadata[string];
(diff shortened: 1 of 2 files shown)


packages/ai/src/generate-image/generate-image.ts
```diff
@@ -200,10 +200,25 @@ Only applicable for HTTP-based providers.
         if (providerName === 'gateway') {
           const currentEntry = providerMetadata[providerName];
           if (currentEntry != null && typeof currentEntry === 'object') {
-            providerMetadata[providerName] = {
+            const merged = {
```
shaper, line 203: Since this is gateway-specific logic, ideally there would be a way to solve this on the Gateway server or provider side. Earlier in the flow, summing things before we get into this code.

Since this is just reporting, and I get that I think the issue is this ai-core code is the only place we have all of the data come together, can we report the gateway provider metadata for each chunk/image-response within that chunk/response and leave it up to the caller to sum or aggregate or look across it if they choose to do so?

gr2m: The whole ` if (providerName === 'gateway') {}` block is already gateway specific and needs to be refactored.

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -200,13 +200,29 @@ Only applicable for HTTP-based providers.
...
```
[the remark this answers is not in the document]
R-Taneja: Intentional with the latest commit, we've taken a different approach.

packages/ai/src/generate-image/generate-image.ts
R-Taneja: Agreed that the whole block needs to be refactored. As a temporary solution to ship image models asap, 2 paths I see are:

1. Go with my code in an [earlier commit](https://github.com/vercel/ai/pull/10478/commits/e31ac1826615266316f69400771a23c89a2f194b) that sums the cost and sets the cost field to the true total cost
2. Go with the current code in the PR that adds an `allCalls` array to the provider metadata containing the metadata for each request, and let the user sum the costs if they want. The only real issue with this is the metadata output gets quite long, and can be somewhat confusing because the data outside the `allCalls` array is just for the last request and shows the cost of one image (same issue), so if a user doesn’t read into the `allCalls` array they won’t realize what the actual cost is.

I was leaning towards option 1, but I realized that other fields other than `cost` are also being overwritten with the last request's metadata. This could be an issue as we'd only return fields like `modelAttempts`, `generationId`, `resolvedProvider`, etc for the last request if n > 1. So maybe option 2 is the best because it returns metadata from all calls. I'm just concerned about the duplicative-ness of the last call's metadata being outside `allCalls` (in the top level `providerMetadata` object) and also in the last position of the `allCalls` array.

shaper:
Can we arrange things so that on ai side n is always 1 -- didn't we do something similar for embeddings?

Issue #5784, opened 2025-04-15
`smoothStream` not smooth reasoning text.


NeroBlackstone:
### Description

https://github.com/user-attachments/assets/3df33b36-29a3-49b0-9684-aa9a1338e1d6

`extractReasoningMiddleware` not export from `ai`, but I didn't make any changes except the log.

What is the calling order of `smoothStream` and `extractReasoningMiddleware`, why smoothStream does not work on reason text.

NeroBlackstone:
okay, I know the reason: 

> https://github.com/vercel/ai/blob/e6e1cd94ebab5259026719ac90a1dc2367070592/packages/ai/core/generate-text/smooth-stream.ts#L93

smoothStream not process `reasoning` type.

Am I right? If so, I can open a PR

NeroBlackstone:
change two line fix this problem, now reason output smoother! But I don't know if this is correct.

``` ts
      async transform(chunk, controller) {
        // fix here
        if (chunk.type !== 'text-delta' && chunk.type !== 'reasoning') {
          if (buffer.length > 0) {
            controller.enqueue({ type: 'text-delta', textDelta: buffer });
            buffer = '';
          }

          controller.enqueue(chunk);
          return;
        }

        buffer += chunk.textDelta;

        let match;

        while ((match = detectChunk(buffer)) != null) {
           // fix here
          controller.enqueue({ type: chunk.type, textDelta: match });
          buffer = buffer.slice(match.length);

          await delay(delayInMs);
        }
      },
```

https://github.com/user-attachments/assets/d08bbe6d-de15-4927-83e6-fe7190a06b9f

NeroBlackstone:
> Have you tried https://sdk.vercel.ai/docs/ai-sdk-ui/smooth-stream-chinese ?

Sure.

NeroBlackstone:
Still exist in ai sdk 5.0 beta 3.
smoothStream smooth text, but not reasoning text

NeroBlackstone:
okay I know how to fix it in ai sdk v5

```ts
      async transform(chunk, controller) {
        // fix
        if (chunk.type !== 'text' && chunk.type !== 'reasoning') {
          if (buffer.length > 0) {
            controller.enqueue({ type: 'text', text: buffer, id });
            buffer = '';
          }

          controller.enqueue(chunk);
          return;
        }

        if (chunk.id !== id && buffer.length > 0) {
          controller.enqueue({ type: 'text', text: buffer, id });
[...]

PR #10504, opened 2025-11-23
feat(agent): support experimental stream transforms

lgrammel:
## Background

Stream transformation such as stream smoothing were not supported when using the agent abstraction.

## Summary

Add `experimental_transform` support to `Agent.stream` and related functions.

Not implemented as a property on `Agent` because transformations specific to streaming and not available for generate.

## Manual Verification

Ran `examples/next-openai` with http://localhost:3000/chat-openai-smooth-stream


gr2m (approved):
I verified it's working. Just one observation: is it possible that the smooth streaming only applies to output text, not to the reasoning text?

lgrammel:
> I verified it's working. Just one observation: is it possible that the smooth streaming only applies to output text, not to the reasoning text?

Yes see https://github.com/vercel/ai/issues/5784

PR #10462, opened 2025-11-21
fix(google): add thought signature to gemini 3 pro image parts

gr2m:
## Background

`generateText()` / `streamText()` doesn't pass thought signature for image parts in response messages. That breaks multi-turn chats with `gemini-3-pro-image-preview`

## Summary

- Gemini now supports non-PNG images in assistant messages
- Add thought signature to image parts in `response.messages`
- spec update: `LanguageModelV3File#providerMetadata` (optional field)

## Manual Verification

```ts
import { google } from '@ai-sdk/google';
import { generateText } from 'ai';
import 'dotenv/config';
import { presentImages } from '../lib/present-image';

async function main() {
  const step1 = await generateText({
    model: google('gemini-3-pro-image-preview'),
    prompt: 'Create an image of Los Angeles where all car infrastructure has been replaced with bike infrastructure, trains, pedestrian zones, and parks. The image should be photorealistic and vibrant.',
  });

  await presentImages(step1.files);

  const step2 = await generateText({
    model: google('gemini-3-pro-image-preview'),
    messages: [
      ...step1.response.messages,
      {
        role: 'user',
        content: 'Now create a variation of the image, but in the style of a watercolor painting.',
      },
    ]
  });

  await presentImages(step2.files);
}

main().catch(console.error);
```

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

Fixes #10441
Closes #7979

PR #10508, opened 2025-11-23
Backport: fix(google): add thought signature to gemini 3 pro image parts

ghost:
This is an automated backport of #10462 to the release-v5.0 branch.

🚨 this pull request includes a spec change which is a breaking change. We might need to backport it anyway since it is needed for a fix that enables multi-turn chats with gemini-3-pro-image-preview.


packages/provider/src/language-model/v2/language-model-v2-file.ts
```diff
@@ -22,4 +24,6 @@ as base64 encoded strings. If the API returns binary data, the file data should
 be returned as binary data.
  */
   data: string | Uint8Array;
+  
+  providerMetadata?: SharedV2ProviderMetadata;
```
gr2m, line 31: breaking change

gr2m (changes requested):
blocking merge until deemed necessary

PR #10529, opened 2025-11-24
Changed example model from 'gpt-5.1' to 'gpt-5.1-instant' in Expo Quickstart docs page

roshidoni:
model "gpt-5.1" is not available in the AI gateway. Changed to working model "gpt-5.1-instant"

## Background

I was creating new expo app and I followed the docs in https://ai-sdk.dev/docs/getting-started/expo . I decided to use AI gateway. But It kept erroring even though I had the exact same code in my codebase as docs page had. Then found out gateway does not expect model named "gpt-5.1". I changed the model name to "gpt-5.1-instant" and it started to work immediately.

## Summary

I changed example model name to "gpt-5.1-instant" in the "AI gateway functions" (Idk if I am calling them correct)


nicoalbanese:
Thanks - fixed across the whole site!

PR #10592, opened 2025-11-25
fix(provider)!: remove generics from EmbeddingModelV3

lgrammel:
## Background

All embedding models that are used with the AI SDK are text embedding models. The generic abstraction in the provider specification has been premature. It makes our code unnecessarily complex and does not work well with custom models and provider registries. If additional embedding models (e.g. for images) are needed in the future, it would be better to add a dedicated model type for those.

## Summary

* Remove generics from `EmbeddingModel`
* Remove generics from `embed` and `embedMany`
* Rename `textEmbedding` / `textEmbeddingModel` on providers to `embedding` / `embeddingModel`

## Manual Verification

- [x] run `examples/ai-core/src/embed-many/openai.ts`
- [x] run `examples/ai-core/src/embed/openai.ts`

## Tasks

- [x] update provider
- [x] change all downstream usages
- [x] fix tests
- [x] update docs
- [x] changeset

## Future Work

* automated code migration for changing `textEmbedding` to `embedding` (provider method)
* automated code migration for changing `textEmbeddingModel` to `embeddingModel` (provider method)
* automatic code migration for custom provider constructor param

## Related Issues

Closes #9819


.changeset/curly-carrots-jog.md
```diff
@@ -0,0 +1,36 @@
...
+'@ai-sdk/azure': patch
+'@ai-sdk/revai': patch
+'@ai-sdk/groq': patch
+'@ai-sdk/luma': patch
+'@ai-sdk/fal': patch
+'@ai-sdk/xai': patch
+'ai': patch
+---
+
+chore(provider): remove generics from EmbeddingModelV3
```
gr2m, line 36: can we add before/after code examples, similar to something that we later want to add to our migration guides?

````suggestion
chore(provider): remove generics from EmbeddingModelV3

Before

```ts
model.textEmbeddingModel('my-model-id')
```

After

```ts
model.embeddingModel('my-model-id')
```
````

gr2m (commented):
~~Can we deprecate `.textEmbeddingModel` in v6 and remove it only in v7 to ease migration?~~

it's part of the provider spec, we cannot deprecate it

gr2m (approved):

PR #10716, opened 2025-11-29
feat (ai): propagate all metadata in generate-image

shaper:
## Background

The `generateImage` function throws away provider metadata other than data in the `images` field. Providers may include metadata the developer would like to reference outside of any `images` field.

## Summary

Generalize `gateway`-specific logic to preserve metadata to apply to all providers. Remove logic to "clean" empty `images` arrays as unnecessary complexity for a too-small win.

## Manual Verification

Ran example scripts with and without `foo: 'bar'` example top level output in OpenAI image model provider metadata and observe its presence with the change and absence without.

## Checklist

### DIFF
--- a/examples/ai-core/src/generate-image/openai.ts
+++ b/examples/ai-core/src/generate-image/openai.ts
@@ -19,7 +19,13 @@ async function main() {
     usage: result.usage,
   });
 
-  await presentImages([result.image]);
+  await presentImages(result.images);
+
+  console.log('Generated', result.images.length, 'image(s)');
+  console.log(
+    'Provider metadata:',
+    JSON.stringify(result.providerMetadata, null, 2),
+  );
 }
 
 main().catch(console.error);
--- a/packages/ai/src/generate-image/generate-image.test.ts
+++ b/packages/ai/src/generate-image/generate-image.test.ts
@@ -816,13 +816,14 @@ describe('generateImage', () => {
       });
 
       expect(result.providerMetadata.gateway).toStrictEqual({
+        images: [],
         routing: { provider: 'test2' },
         generationId: 'gen-123',
         cost: '0.01',
       });
     });
 
-    it('should drop empty images array for gateway provider', async () => {
+    it('should preserve empty images array for gateway provider', async () => {
       const result = await generateImage({
         model: new MockImageModelV3({
           doGenerate: async () =>
@@ -841,13 +842,13 @@ describe('generateImage', () => {
       });
 
       expect(result.providerMetadata.gateway).toStrictEqual({
+        images: [],
         routing: { provider: 'vertex' },
         cost: '0.04',
       });
-      expect(result.providerMetadata.gateway).not.toHaveProperty('images');
     });
 
-    it('should not drop empty images array for non-gateway providers', async () => {
+    it('should preserve empty images array for all providers', async () => {
       const result = await generateImage({
         model: new MockImageModelV3({
           doGenerate: async () =>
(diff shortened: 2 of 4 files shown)


shaper:
* **#10716** <a href="https://app.graphite.com/github/pr/vercel/ai/10716?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/ai/10716?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

shaper:
Adding a `foo: 'bar'` in openai image model provider metadata, before this change, running the `src/generate-image/openai.ts` example script and logging provider metadata gives:

```
Provider metadata: {
  "openai": {
    "images": [
      null
    ]
  }
}
```

After:
```
Provider metadata: {
  "openai": {
    "images": [
      null
    ],
    "foo": "bar"
  }
}
```

However, this change is still data-lossy in that we effectively apply "last one wins" to provider metadata fields other than `images`. We may make multiple image model requests when needed to generate more images than a provider can execute in one request. In this case we can end up with more than one `providerMetadata` in need of merging. For `images` we carefully preserve order and concatenate, but for the rest of the data we just merge as last-one-wins.

Sonnet 4.5 summary:

## The Challenge

When `generateImage` needs to make multiple calls to the provider (e.g., requesting 3 images but `maxImagesPerCall` is 2), the function must aggregate results from multiple provider responses. The key challenge is: **how should provider metadata be merged when each call returns its own metadata?**

## Current Merging Logic

Looking at lines 169-191 in `generate-image.ts`, the current logic:

1. **For the `images` field specifically**: Concatenates arrays to preserve order
   ```typescript
   if ('images' in existingMetadata && 
       Array.isArray(existingMetadata.images) &&
       Array.isArray(metadata.images)) {
     (providerMetadata[providerName] as { images: unknown[] }).images = [
       ...existingMetadata.images,
       ...metadata.images,
     ];
   }
   ```

2. **For all other fields**: Uses **last-one-wins** via object spread
   ```typescript
     ...(existingMetadata as object),
     ...metadata,  // This overwrites any conflicting keys
   } as ImageModelV3ProviderMetadata[string];
   ```

## What This Means

**Images field**: Properly merged and order-preserved across calls ✓

**Non-images fields**: Last call wins. For example:
- Call 1 returns `{ routing: { provider: 'openai' }, cost: '0.04' }`
- Call 2 returns `{ routing: { provider: 'vertex' }, generationId: 'gen-123' }`
- **Result**: `{ routing: { provider: 'vertex' }, cost: '0.04', generationId: 'gen-123' }`

The `routing.provider` from call 2 overwrites call 1's value, but `cost` from call 1 is preserved since call 2 didn't include it.

## Potential Issues

1. **Semantic correctness**: If  a value differs between calls, the final metadata only shows the last value, losing the history
2. **Aggregatable fields**: Fields that one might want to sum are instead overwritten/lost
3. **Nested objects**: Deep merging doesn't happen—entire nested objects are replaced

## Test Coverage

The tests verify this behavior works as designed (see test cases starting line 665), but they don't explicitly call out the last-one-wins semantics for non-images fields. The tests show it working correctly for the gateway use case where later calls add new fields rather than conflicting ones.

**Summary for code review**: The current implementation uses last-one-wins for all non-images metadata fields. This works for the gateway's current usage pattern (where successive calls add complementary fields), but could lose information if providers return conflicting values for the same field across multiple calls.

gr2m:
Already implemented in the latest AI SDK.

PR #10746, opened 2025-12-01
fix(mcp) : convert JSON-RPC errors to CallToolResult format

### LINKED ISSUE #10731: [@ai-sdk/react] Cannot read properties of undefined (reading 'text') after MCP tool error response
## Description

When using `@ai-sdk/mcp` with `@ai-sdk/react`, if an MCP tool returns an error response (JSON-RPC error with `error` field instead of `result`), subsequent tool calls may fail with:

```
    at @ai-sdk_react.js:21874:15
    at @ai-sdk_react.js:22878:17
    at Array.<anonymous> (@ai-sdk_react.js:22255:17)
    at SerialJobExecutor.processQueue (@ai-sdk_react.js:22246:26)
```

## Environment

- `@ai-sdk/mcp`: 0.0.11
- `@ai-sdk/react`: 2.0.104
- `ai`: 5.0.104
- Platform: Windows 11
- Node.js: v22.x

## Steps to Reproduce

1. Set up an MCP server that returns error responses (e.g., Elasticsearch MCP server with ES|QL)
2. Execute a tool call that returns an error (JSON-RPC response with `error` field)
3. The AI model attempts to call another tool
4. Error occurs when processing the second tool's result

## MCP Log Example

The MCP transport layer receives and sends messages correctly:

```json
// First tool call - returns error
{"method":"tools/call","params":{"name":"esql","arguments":{"query":"..."}},"jsonrpc":"2.0","id":14}
{"jsonrpc":"2.0","id":14,"error":{"code":-32603,"message":"HTTP status client error (400 Bad Request) for url ..."}}

// Second tool call - server responds correctly
{"method":"tools/call","params":{"name":"list_indices","arguments":{"index_pattern":"*"}},"jsonrpc":"2.0","id":15}
{"jsonrpc":"2.0","id":15,"result":{"content":[{"type":"text","text":"Found 5 indices:"},{"type":"text","text":"[...]"}],"isError":false}}
```

The MCP server returns valid responses, but `@ai-sdk/react` fails to process the second tool result.

## Expected Behavior

After an MCP tool returns an error response, subsequent tool calls should be processed correctly.

## Actual Behavior

The error `Cannot read properties of undefined (reading 'text')` is thrown, and the stream processing stops. The UI does not show the second tool call result.

## Analysis

The issue appears to be in the stream processing logic within `@ai-sdk/react`. When an MCP tool returns a JSON-RPC error (with `error` field instead of `result`), the internal state may not be properly updated, causing the next tool result's `content[].text` access to fail.

This is an intermittent issue - it doesn't always occur, suggesting a possible race condition or timing issue.

shubham-021:
Fixes : #10731 

## Background

When an MCP tool returns a JSON-RPC error response, the application crashes with:
`TypeError: Cannot read properties of undefined (reading 'text')`

Root cause:

 - MCP server returns JSON-RPC error: { error: { code, message, data } }
 - MCPClient.onResponse() converts this to MCPClientError exception
 - Exception propagates to React/AI SDK core
 - React expects CallToolResult with content array structure
 - Tries to access toolResult.cont ent[0].text → crashes (content is undefined)
 - This failure corrupts the tool execution queue, causing subsequent tool calls to also fail

## Summary

**Solution Implemented:**

 - first , the this.request() method in callTool was missing await , it was necessary since request return a promise.
 - then , in the try catch block , specifically in the catch block , wrote the logic to check if error is an instance of MCPClientError , if yes then return response satisfying the format of CallToolResult.
 
 ```
catch (error) {
      if (MCPClientError.isInstance(error)) {
        return {
          content: [{
            type: 'text',
            text: error.message,
          }],
          isError: true,
          _meta: {
            errorCode: error.code,
            errorData: error.data
          },
        };
      }
      throw error;
    }
```
 - `content` array containing error message as text
 - `isError` : true flag
 - `_meta` field preserving error code and data

**Updated Tests:**
 - Updated test `should error when calling tool with misconfigured parameters` to expect CallToolResult instead of thrown error
 - Updated test `should include JSON-RPC error data in MCPClientError` to `should include JSON-RPC error metadata in CallToolResult._meta` to verify metadata preservation
 
## Checklist


aayush-kapoor:
thanks for the PR! did you verify by reproducing the original issue exactly and confirmed that your PR resolves it? if so, it would help for reviewing if you could add the repro code as an example (add under `example/ai-core/src/generate-text`) in this PR and then we can later remove it

shubham-021:
> thanks for the PR! did you verify by reproducing the original issue exactly and confirmed that your PR resolves it? if so, it would help for reviewing if you could add the repro code as an example (add under `example/ai-core/src/generate-text`) in this PR and then we can later remove it

i verified the issue tracing through the execution flow and identified that when an mcp tool returns a JSON-RPC error response , the callTool function in mcp-client.ts throws an MCPClientError exception without converting it to the expected format. This causes the crash when trying to access result.content[0].text since error doesnt have any content field.

Then to confirm this , I wrote a test that simulates the issue by mocking mcp tool call that returns an error attempting to access result.content[0].text , without the fix, the test fails because an exception is thrown instead of returning a properly formatted result , with the fix, the error is converted to a valid callToolResult and the test passes.

aayush-kapoor:
even so, we highly recommend trying to reproduce the original issue. if reproduction code isn't provided, ask the OP to send some. not saying this is the case here but simulated unit tests are easy to bypass. we want to ensure tests and changes reflect in a real-world example

shubham-021:
> even so, we highly recommend trying to reproduce the original issue. if reproduction code isn't provided, ask the OP to send some. not saying this is the case here but simulated unit tests are easy to bypass. we want to ensure tests and changes reflect in a real-world example

yes absolutely , i'm on it

shubham-021:
@aayush-kapoor I wasn’t able to reproduce the issue. While retracing the steps, I realized my solution wasn’t actually consistent with it. Adding the test felt like a good way to validate, but as you pointed out, it didn’t capture the full scope of it across the repo.

Thanks for helping me understand why reproducing issues in a real-world environment is important. I’m closing this PR since I couldn’t identify the core cause.

PR #10747, opened 2025-12-01
fix(anthropic): send {} as tool input when streaming tool calls without arguments

lgrammel:
## Background

When streaming tool calls to tools without parameters, the Anthropic provider did send an empty string and not `{}` as the tool input (see #10295 ). This breaks e.g. the OpenAI agents integration.

## Summary

Send `{}` as tool input when streaming tool calls with no args.

## Tasks

- [x] generate test case with fixture
- [x] stream test case with fixture
- [x] stream fix
- [x] changeset

## Related Issues

Fixes #10295


nicoalbanese (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10749

PR #10756, opened 2025-12-01
fix(anthropic): simplify pulling first chunk

### LINKED ISSUE #10725: fix(anthropic): streamText() hangs in Cloudflare Workers due to tee() + async IIFE deadlock
## Description

`streamText()` with the Anthropic provider hangs indefinitely when deployed to Cloudflare Workers. The issue is in `@ai-sdk/anthropic`'s `doStream()` method.

## Root Cause

In `anthropic-messages-language-model.ts`, the `doStream()` method uses a `tee()` + async IIFE pattern to check if the first chunk is an error:

```typescript
// Lines 1551-1571
const [streamForFirstChunk, streamForConsumer] = transformedStream.tee();
stream = streamForConsumer;
const reader = streamForFirstChunk.getReader();
(async () => {
  try {
    const { done } = await reader.read();  // Hangs here in CF Workers!
    if (!done) {
      await reader.cancel();
    }
  } catch (error) { /* ... */ }
  finally { reader.releaseLock(); }
})();

return returnPromise.promise;
```

This creates a deadlock in Cloudflare Workers:

1. `returnPromise.promise` waits for the promise to be resolved
2. The promise is resolved inside `transform()` when data flows through
3. Data flows when someone reads from the stream
4. But nobody is reading from `streamForConsumer` because `doStream()` hasn't returned yet
5. The async IIFE calls `reader.read()` on `streamForFirstChunk`, which waits for data
6. In CF Workers' pull-based stream model, `read()` triggers `pull()`, but the tee'd stream doesn't properly propagate backpressure

**Result**: Deadlock - `reader.read()` never completes, `returnPromise` is never resolved, `doStream()` never returns.

## Environment

- Cloudflare Workers (workerd runtime)
- `@ai-sdk/anthropic` v3.0.0-beta.66
- AI SDK v6 beta

## Proposed Fix

The fix is to resolve `returnPromise` immediately without the tee + first chunk read pattern:

```typescript
// Remove the tee() pattern entirely
stream = transformedStream;
returnPromise.resolve({
  stream,
  request: { body },
  response: { headers: responseHeaders }
});
return returnPromise.promise;
```

The error handling in `transform()` (lines 964-984) already handles the case where Anthropic returns an error as a 200 response - it will emit an error event to the consumer.

**Trade-off**: With this fix, if Anthropic returns an overloaded error as 200, `doStream()` will return successfully and the consumer will get the error when reading from the stream, rather than having `doStream()` throw immediately. This is arguably more consistent with streaming semantics.

## Reproduction

1. Deploy a Cloudflare Worker using `streamText()` with Anthropic
2. The request will hang indefinitely with no response

## Workaround

Use `patch-package` to patch `@ai-sdk/anthropic`:

```diff
--- a/node_modules/@ai-sdk/anthropic/dist/index.mjs
+++ b/node_modules/@ai-sdk/anthropic/dist/index.mjs
-    const [streamForFirstChunk, streamForConsumer] = transformedStream.tee();
-    stream = streamForConsumer;
-    const reader = streamForFirstChunk.getReader();
-    (async () => { ... })();
+    stream = transformedStream;
+    returnPromise.resolve({
+      stream,
+      request: { body },
+      response: { headers: responseHeaders }
+    });
```

lgrammel:
## Background

#10739 did not resolve #10725 - Cloudflare still detects a promise that will never resolve.

## Summary

Rewrite logic to remove delayed promise.

## Manual Verification

- [x] test resolution on cloudflare workers

## Related Issues

Closes #10725
Related #10739

### DIFF
--- a/.changeset/moody-gorillas-clean.md
+++ b/.changeset/moody-gorillas-clean.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix(anthropic): simplify pulling first chunk
--- a/packages/anthropic/src/anthropic-messages-language-model.ts
+++ b/packages/anthropic/src/anthropic-messages-language-model.ts
@@ -1,7 +1,7 @@
 import {
+  APICallError,
   JSONObject,
   LanguageModelV3,
-  SharedV3Warning,
   LanguageModelV3Content,
   LanguageModelV3FinishReason,
   LanguageModelV3FunctionTool,
@@ -11,14 +11,13 @@ import {
   LanguageModelV3ToolCall,
   LanguageModelV3Usage,
   SharedV3ProviderMetadata,
+  SharedV3Warning,
   UnsupportedFunctionalityError,
-  APICallError,
 } from '@ai-sdk/provider';
 import {
   combineHeaders,
   createEventSourceResponseHandler,
   createJsonResponseHandler,
-  DelayedPromise,
   FetchFunction,
   generateId,
   InferSchema,
--- a/packages/anthropic/src/anthropic-messages-language-model.ts
+++ b/packages/anthropic/src/anthropic-messages-language-model.ts
@@ -1504,27 +1504,30 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
 
     const firstChunkReader = streamForFirstChunk.getReader();
     try {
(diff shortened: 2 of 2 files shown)


nicoalbanese (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10758

PR #10770, opened 2025-12-01
feat(ai): add warning when using v2 models with AISDK v6

aayush-kapoor:
## Background

#10689 

Add warning when a model with spec version v2 is used with ai-sdk v6

## Summary

Used the warning logger system.

## Checklist

## Related Issues

fixes #10689


packages/ai/src/model/as-embedding-model-v3.ts
```diff
@@ -7,6 +8,18 @@ export function asEmbeddingModelV3(
     return model;
   }
 
+  logWarnings({
+    warnings: [
+      {
+        type: 'compatibility',
+        feature: 'specificationVersion',
+        details: `This model is using specification version ${model.specificationVersion}. Please upgrade the package to the latest version.`,
```
gr2m: Will it be clear to the user what "the package" means? 

Also it reads like as a user I need to upgrade "the package" to the latest version in order to address the warning, but it really means that the owner of "the package" needs to upgrade it to use the latest specification version.

And I think so far we should be able to fully support providers that use v2 of the spec.

lgrammel: The issue this addresses is that people stay on older versions with limited functionality for a provider, leaving them unable to use v3 features. We could be descriptive instead of prescriptive in the message:

`Using v2 specification compatibility mode. Some features may not be available.`

aayush-kapoor: changed it to what Lars wrote.

lgrammel: still showing the old text

packages/ai/src/model/as-transcription-model-v3.ts
```diff
@@ -7,6 +8,18 @@ export function asTranscriptionModelV3(
     return model;
   }
 
+  logWarnings({
+    warnings: [
+      {
+        type: 'compatibility',
+        feature: 'specificationVersion',
+        details: `This model is using specification version ${model.specificationVersion}. Please upgrade the package to the latest version.`,
```
lgrammel: text

packages/ai/src/model/as-speech-model-v3.ts
```diff
@@ -7,6 +8,18 @@ export function asSpeechModelV3(
     return model;
   }
 
+  logWarnings({
+    warnings: [
+      {
+        type: 'compatibility',
+        feature: 'specificationVersion',
+        details: `This model is using specification version ${model.specificationVersion}. Please upgrade the package to the latest version.`,
```
lgrammel: text

packages/ai/src/model/as-language-model-v3.ts
```diff
@@ -7,6 +8,18 @@ export function asLanguageModelV3(
     return model;
   }
 
+  logWarnings({
+    warnings: [
+      {
+        type: 'compatibility',
+        feature: 'specificationVersion',
+        details: `This model is using specification version ${model.specificationVersion}. Please upgrade the package to the latest version.`,
+      },
```
lgrammel: text (maybe consider extracting a function that generates the warning and calling it from all as model v3 functions)

packages/ai/src/model/as-image-model-v3.ts
```diff
@@ -7,6 +8,18 @@ export function asImageModelV3(
     return model;
   }
 
+  logWarnings({
+    warnings: [
+      {
+        type: 'compatibility',
+        feature: 'specificationVersion',
+        details: `This model is using specification version ${model.specificationVersion}. Please upgrade the package to the latest version.`,
```
lgrammel: same

packages/ai/src/model/as-language-model-v3.ts
aayush-kapoor: that works too

lgrammel (approved):

PR #10812, opened 2025-12-02
feat(anthropic): advanced tool use

ussef083:
## Background

Anthropic recently introduced new advanced built-in tools and capabilities:
https://www.anthropic.com/engineering/advanced-tool-use

## Tasks

- [ ] changeset
- [ ] docs
- [ ] review tool regex
- [ ] review tool bm25
- [ ] examples
- [ ] review tool options


ussef083:
@lgrammel I'd really appreciate it if you could review this PR.

lgrammel:
@ussef083 thanks. i will see if i can get this shipped today. please do not make any more changes to this branch

lgrammel:
I prefer to add individual features (tool example, one search tool, the other search tool, etc) as individual pull requests to more thoroughly test and document them.

ussef083:
thank you very much , if there is an task i can do, I'm here.

packages/anthropic/src/anthropic-prepare-tools.ts
```diff
@@ -53,14 +57,17 @@ export async function prepareTools({
...
+
         anthropicTools.push({
           name: tool.name,
           description: tool.description,
           input_schema: tool.inputSchema,
           cache_control: cacheControl,
-          ...(supportsStructuredOutput === true && tool.strict != null
-            ? { strict: tool.strict }
-            : {}),
+          ...advancedToolFeatures,
```
AVtheking, line 70: wouldn't this make all tool follow same `defer_loading` property ? if we gonna go this way then how could we have mix of tools with some being defered and some (could be most used or critical ones) not defered ?

ussef083: no since `advancedToolFeatures` returns only the properties specified by the user , as example if `defer_loading ` is not specified, so it won't be there.

AVtheking: This would still be specified in provider options right, and that make  it applied to all tools then. I mean how could you have mix of tools with some having defer_loading set to true and some with false

AVtheking: Oops sorry, providerMetadata is tool property, I missed that,  this works then

raphaelcunha:
Appreciate this contribution. Very helpful.

gr2m:

PR #10817, opened 2025-12-03
feat: tool-specific strict mode

### LINKED ISSUE #7948: Add in v5 support on strict schema per tool
### Description

In v4, some logic based on the model name decided if a strict schema should be applied to function calls (enforcing a strict schema will not allow optional parameters). I think this should be disabled by default, and if users want to enforce it, they should be able to do so on the tool level. 

I think the library should not make assumptions on this (like based on model name), but let the user decide per tool.

https://platform.openai.com/docs/guides/function-calling#strict-mode

Example:

Image

Image

Original issue in v4: 
https://github.com/vercel/ai/issues/7888

lgrammel:
## Background

Many providers support tool specific `strict` mode. It guarantees correct tool calls. Strict mode depends on a specific tool and its input schema, and thus needs to be controlled on a tool level, so it can be turned off for tools with incompatible schemas.

## Summary

* add optional `strict` property to `Tool`
* add support in deepseek, mistral, anthropic, openai chat, openai responses, openai compatible chat
* remove using `strictJsonSchema` for strict tool schemas in openai chat and responses (**breaking change**)

## Manual Verification

- [x] test with anthropic `examples/ai-core/src/stream-text/anthropic-tool-call-strict.ts`
- [x] test with openai responses `examples/ai-core/src/stream-text/openai-tool-call-strict.ts`

## Tasks

- [x] specification
- [x] send strict from ai to provider
- [x] deepseek
- [x] remove strictJsonSchema default in openai chat and responses
- [x] openai chat
- [x] openai responses
- [x] openai compatible chat
- [x] anthropic
- [x] mistral
- [x] migration guide
- [x] examples and testing
- [x] docs: tool guide
- [x] docs: tool spec
- [x] changeset

## Future Work

* update other providers that support strict tool calling

## Related Issues

Resolves #7948


content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -49,6 +50,32 @@ const result = await generateText({
 Tool calling is not restricted to only text generation.
 You can also use it to render user interfaces (Generative UI).
 
+---
+
```
lgrammel: rm

nicoalbanese (approved):

Issue #8379, opened 2025-08-30
OpenAI provider: Item of type 'reasoning' was provided without its required following item


ShadowWalker2014:
### Description

we keep getting this error when using openai gpt-5 model with reasoning:
Item 'rs_68b3836e52148191a533bfa8f266c7060be42a56df013daa' of type 'reasoning' was provided without its required following item.

<img width="2860" height="1736" alt="Image" src="https://github.com/user-attachments/assets/63f6de89-563e-4053-b8ae-aaa34d241d5a" />

### AI SDK Version

- ai: 5.0.0
    "@ai-sdk/anthropic": "^2.0.6",
    "@ai-sdk/google": "^2.0.8",
    "@ai-sdk/google-vertex": "^2.2.27",
    "@ai-sdk/groq": "^2.0.14",
    "@ai-sdk/mistral": "^2.0.9",
    "@ai-sdk/openai": "^2.0.19",
    "@ai-sdk/openai-compatible": "^1.0.11",
    "@ai-sdk/react": "^2.0.22",

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

lgrammel:
Can you share a full code snippet that reproduces the issue? Is it always related to invalid tool calls? Which model are you using?

ben-edge:
Getting same issue - happens when a tool call fails.

<img width="804" height="882" alt="Image" src="https://github.com/user-attachments/assets/cceee477-dbbf-4aa0-8115-a184c78ba899" />

<img width="791" height="136" alt="Image" src="https://github.com/user-attachments/assets/7679b1dc-42bd-4016-bad9-38e47cba96f8" />

gr2m:
if one of you could share a minimal reproducible test case it would help us to prioritize this issue

Theonlyhamstertoh:
Getting same error related:

Using GPT-5, low reasoning

```
Error [AI_APICallError]: Item 'rs_68b72f0a0fd48195b0fe82e74c730f700d2f7ee01550624d' of type 'reasoning' was provided without its required following item.
    at M.errorSchema (.next/server/chunks/208.js:33:7522)
    at async z (.next/server/chunks/208.js:33:4039)
    at async as.doStream (.next/server/chunks/208.js:43:20508)
    at async fn (.next/server/chunks/208.js:29:24427)
    at async (.next/server/chunks/208.js:27:26438)
    at async bS (.next/server/chunks/208.js:27:27987)
[...]

PR #10733, opened 2025-12-01
fix(ai): v5: preserve providerMetadata on invalid tool calls and tool results

Und3rf10w:
## Background

Provider metadata (such as Google Gemini's `thoughtSignature`) was being lost in two scenarios:

1. **Tool result propagation**: When tool results were converted to model messages, the `callProviderMetadata` from tool UI parts was not being propagated to tool-result parts. This was fixed in main branch (PR #10361, merged November 19, 2025) but not backported to v5.0.

2. **Invalid tool call handling**: When a tool call failed input validation (e.g., Zod schema mismatch), the error handler was creating an invalid tool call object without preserving the `providerMetadata` field from the original tool call.

This is particularly problematic for providers like Google Gemini that use provider metadata to track `thoughtSignature` through reasoning and tool execution cycles.

## Summary

This PR includes two related fixes for provider metadata preservation on the `release-v5.0` branch:

### Fix 1: Tool Result Metadata Propagation (Backport from main)

Backported changes ensure that `callProviderMetadata` from tool UI parts is correctly propagated to tool-result parts as `providerOptions` in two locations:

1. **Provider-executed tool results** (within assistant message content) - [convert-to-model-messages.ts:223-225](packages/ai/src/ui/convert-to-model-messages.ts#L223-L225)
2. **Client-executed tool results** (in separate tool messages) - [convert-to-model-messages.ts:281-283](packages/ai/src/ui/convert-to-model-messages.ts#L281-L283)

### Fix 2: Invalid Tool Call Metadata Preservation (New)

Added `providerMetadata` preservation in the error handler of `parse-tool-call.ts` (line 88). When tool input validation fails, the invalid tool call object now includes the `providerMetadata` from the original `LanguageModelV2ToolCall`, matching the behavior of the success path.

**File**: [packages/ai/src/generate-text/parse-tool-call.ts:88](packages/ai/src/generate-text/parse-tool-call.ts#L88)

Both fixes ensure consistency in how provider metadata flows through the entire tool execution lifecycle (tool-call → tool-result) and error paths.

## Manual Verification

Added comprehensive test coverage:
- Client-executed tool result with provider metadata
- Provider-executed tool result with provider metadata  
- Error state tool result with provider metadata
- Dynamic tool result with provider metadata
- Updated existing test snapshot to reflect the fix

In addition, I manually ran: 

[google-vertex-code-execution.](examples/ai-core/src/stream-text/google-vertex-code-execution.ts)
[...]

PR #10827, opened 2025-12-03
fix(openai): v5 - skip unpaired reasoning item_reference in Responses API

test137E29B:
## Background

When using the OpenAI Responses API with `store: true` and reasoning models (e.g., o3, o4-mini), the API returns errors when `item_reference` items are unpaired:

- `Item 'rs_xxx' of type 'reasoning' was provided without its required following item`
- `Item 'fc_xxx' of type 'function_call' was provided without its required 'reasoning' item`

This occurs in multi-turn conversations when:

- A reasoning item is followed by a tool call that lacks an itemId (i.e., the tool call wasn't stored by the API)
- The tool call emits a full function_call object instead of an item_reference
- The reasoning item_reference is now "unpaired" because OpenAI expects both reasoning and its following item to either both use item_reference or both use full objects

The typical scenario is that the agent reasons, calls a tool, the tool returns an error (e.g., validation failure), and on retry the SDK rebuilds the conversation context. The reasoning has an `itemId` from the previous turn, but the failed tool call does not, causing the mismatch.

## Summary

Added look-ahead logic when processing reasoning parts with `store: true`. Before including a reasoning `item_reference`, the code now scans past consecutive reasoning parts to find the first non-reasoning part. The reference is only included if that part has a valid itemId that it will then include in it's iteration. Failed tool calls do not have the `itemId` and thus do not include the `item_reference`, causing the issue.

## Manual Verification

Tested with a reasoning model that calls a custom tool with invalid input. Previously this caused the API error; now the reasoning `item_reference` is correctly skipped and the conversation continues.

## Checklist

## Future Work

Port into v6 main branch. https://github.com/vercel/ai/pull/10829

## Related Issues

https://github.com/vercel/ai/issues/8379

### DIFF
--- a/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
@@ -1124,7 +1124,15 @@ describe('convertToOpenAIResponsesInput', () => {
                       },
                     },
                   },
-                  { type: 'text', text: 'First response' },
+                  {
+                    type: 'text',
+                    text: 'First response',
+                    providerOptions: {
+                      openai: {
+                        itemId: 'msg_001',
+                      },
+                    },
+                  },
                 ],
               },
               {
@@ -1143,7 +1151,15 @@ describe('convertToOpenAIResponsesInput', () => {
                       },
                     },
                   },
-                  { type: 'text', text: 'Second response' },
+                  {
+                    type: 'text',
+                    text: 'Second response',
+                    providerOptions: {
+                      openai: {
+                        itemId: 'msg_002',
+                      },
+                    },
+                  },
                 ],
               },
             ],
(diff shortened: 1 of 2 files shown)


test137E29B:
tagging @aayush-kapoor for visibility since I note you are assigned to the issue that this resolves 😄

aayush-kapoor:
cc @lgrammel

lgrammel:
can you provide an example under examples/ai-core with a tool that throws an error which leads to this problem?

i want to make sure we fix the root cause and i am not convinced yet that this isn't patching the symptom

lgrammel:
also please do all work in a pr against main; we have tooling for backports

test137E29B:
@lgrammel This one got a lot harder to reproduce, however, eventually I managed to track down the root cause of the issue itself - which was resolved a couple of days ago in https://github.com/vercel/ai/issues/10733

So we're good to close this, and related issues, as resolved!

The chain was:
InvalidToolInputError -> tool-call with no providerMetadata -> no providerMetadata.openai.itemId (even though one existed on the response from openai).

PR #10833, opened 2025-12-03
fix(rsc): require minimal versions for RSC to address CVE-2025-55182

gr2m:
https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -49,13 +49,13 @@
     "eslint-config-vercel-ai": "workspace:*",
     "jsdom": "^24.0.0",
     "msw": "2.6.4",
-    "react-dom": "^18 || ^19 || ^19.0.0-rc",
+    "react-dom": "^18 || ^19",
     "tsup": "^7.2.0",
     "typescript": "5.8.3",
     "zod": "3.25.76"
   },
   "peerDependencies": {
-    "react": "^18 || ^19 || ^19.0.0-rc"
+    "react": "^18 || ^19"
   },
   "engines": {
     "node": ">=18"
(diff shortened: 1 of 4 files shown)


packages/rsc/package.json
```diff
@@ -65,7 +65,7 @@
     "zod": "3.25.76"
   },
   "peerDependencies": {
-    "react": "^18 || ^19 || ^19.0.0-rc",
+    "react": "^18 || ~19.0.1 || ~19.1.2 || ^19.2.1",
```
gr2m, line 68: since the CVE is only related to server components, I don't think we need to update the peer dependency version range in the `react` package, `rsc` should suffice.

packages/rsc/package.json
```diff
@@ -65,7 +65,7 @@
     "zod": "3.25.76"
   },
   "peerDependencies": {
-    "react": "^18 || ^19 || ^19.0.0-rc",
```
gr2m, line 68: I think it makes sense to drop support for 19-rc in ai v6. We can keep supporting it in v5 though

packages/react/package.json
```diff
@@ -49,13 +49,13 @@
...
     "msw": "2.6.4",
-    "react-dom": "^18 || ^19 || ^19.0.0-rc",
+    "react-dom": "^18 || ^19",
     "tsup": "^7.2.0",
     "typescript": "5.8.3",
     "zod": "3.25.76"
   },
   "peerDependencies": {
-    "react": "^18 || ^19 || ^19.0.0-rc"
+    "react": "^18 || ^19"
```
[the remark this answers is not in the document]
gr2m: versions ranges not incompatible, I'm just limiting friction to where it's needed.

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/10836

PR #10858, opened 2025-12-04
add public serverInfo

amirbilu:
This pull request adds support for exposing server information in the MCP client. At Tabnine, customers use `serverInfo.name` to identify and control their servers, so this property is required.

## Server Information

* Added `ClientOrServerImplementation` and updated `MCPClient` to include a `serverInfo` property.  
* `DefaultMCPClient` now sets `serverInfo` during initialization.  

## Testing

* Added a test to check that `serverInfo` is correctly exposed.  

## Type Imports

* Imported `ClientOrServerImplementation` in the client file to support the new property.


gr2m:
Already implemented in #14288.

PR #10908, opened 2025-12-05
fix(ai-sdk/react): Fix subscribeToMessages callback dependency in useChat

jeffcarbs:
## Background

Follow-up to:
- #7387

That PR fixed stale subscriptions when the chat ID changes. However, it was incomplete since it only handled the case where the new ID was passed in the options. It's also possible for you to pass in a `chat` instance directly which could have a new ID.

## Summary

Tweak the logic there to use the `chatRef.current.id` in the dependency array, so no matter how the ID changes we'll refresh the subscription callbacks.

## Manual Verification

Didn't do manually testing since the unit tests for this specific case (plus the one I added) seemed sufficient.

## Checklist

### DIFF
--- a/packages/react/src/use-chat.ui.test.tsx
+++ b/packages/react/src/use-chat.ui.test.tsx
@@ -2314,6 +2314,71 @@ describe('chat instance changes', () => {
 
     expect(screen.queryByTestId('message-0')).not.toBeInTheDocument();
   });
+
+  it('should handle streaming correctly when the id changes', async () => {
+    const controller = new TestResponseController();
+    server.urls['/api/chat'].response = {
+      type: 'controlled-stream',
+      controller,
+    };
+
+    // First, change the ID
+    await userEvent.click(screen.getByTestId('do-change-chat'));
+
+    // Then send a message
+    await userEvent.click(screen.getByTestId('do-send'));
+
+    await waitFor(() => {
+      expect(screen.getByTestId('status')).toHaveTextContent('submitted');
+    });
+
+    controller.write(formatChunk({ type: 'text-start', id: '0' }));
+    controller.write(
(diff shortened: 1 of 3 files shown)


packages/react/src/use-chat.ui.test.tsx
```diff
@@ -2314,6 +2314,71 @@ describe('chat instance changes', () => {
 
     expect(screen.queryByTestId('message-0')).not.toBeInTheDocument();
   });
+
+  it('should handle streaming correctly when the id changes', async () => {
```
jeffcarbs, line 2318: This test is largely copied from the test added in #7387: https://github.com/vercel/ai/blob/68d88037b17d14d55ac0d2a727824fa2449b92af/packages/react/src/use-chat.ui.test.tsx#L2352-L2415

packages/react/src/use-chat.ui.test.tsx
```diff
@@ -2314,6 +2314,71 @@ describe('chat instance changes', () => {
...
+    const controller = new TestResponseController();
+    server.urls['/api/chat'].response = {
+      type: 'controlled-stream',
+      controller,
+    };
+
+    // First, change the ID
+    await userEvent.click(screen.getByTestId('do-change-chat'));
```
lgrammel, line 2326: is this changing the id on the chat instance in the way you describe in the bug report? ie without your fix would this test fail?

jeffcarbs: Yes, you can see here it resets the chat instance to one with a new ID: https://github.com/vercel/ai/blob/fa8e472acfc59a9bafc22f9d50181157d8425cfc/packages/react/src/use-chat.ui.test.tsx#L2252-L2262

The test fails before the change:
Screenshot 2025-12-05 at 10 24 01 AM

lgrammel (approved):

PR #10919, opened 2025-12-06
fix: handle SSE events without explicit message event

### LINKED ISSUE #10915: MCP client doesn't default to `message` for SSE events
### Description

By default an SSE event without an event type should be message but the process event function is expecting event to be message before proceeding.

This is wrong since servers can send just the data part and it should default to `message`.

I'll open up a PR tomorrow to fix this if someone doesn't beat me to it.

### AI SDK Version

@ai-sdk/mcp@0.0.11

### Code of Conduct

paoloricciuti:
## Background

SSE events should default to `message` as the event...both the SSE transport and the HTTP transport were explicitly checking for event `message` which means that servers that didn't explicitly emit events with an event name where hanging forever.

## Summary

I've expanded the check to invoke the onmessage callback even when the event is `undefined` which is how SSE events should work.

## Manual Verification

I've patched my version of ai sdk with the same changes and used a server that wasn't returning explicit event messages and verified that it worked.

## Checklist

## Related Issues

Closes #10915


aayush-kapoor:
hey @paoloricciuti , do you mind sending the server you tested with? if possible also the code snippet you tested with. will test it e2e

paoloricciuti:
I actually fixed the SDK I was using but I can probably use some old version and deploy a "bugged" server with it.

Will send this in a bit

paoloricciuti:
@aayush-kapoor here's a repo 

https://github.com/paoloricciuti/ai-sdk-bug

you can run it locally by doing `pnpm test`. The ai-sdk part looks like this

```ts
import { Experimental_Agent as Agent, stepCountIs } from 'ai';
import { experimental_createMCPClient as createMCPClient } from '@ai-sdk/mcp';
import { writeFileSync } from 'node:fs';
import 'dotenv/config';

const mcp_client = await createMCPClient({
	transport: {
		type: 'http',
		url: 'http://localhost:3000/mcp',
	},
});

console.log('Does it reach here?'); // spoiler: no

const agent = new Agent({
	model: 'anthropic/claude-haiku-4.5',
	tools: await mcp_client.tools(),
	stopWhen: stepCountIs(10),
});

const result = await agent.generate({
	prompt: 'Can you greet me?',
});

writeFileSync(`result-${Date.now()}.json`, JSON.stringify(result, null, 2));
```

packages/mcp/src/tool/mcp-http-transport.test.ts
```diff
@@ -99,6 +99,53 @@ describe('HttpMCPTransport', () => {
     });
   });
 
+  it('should handle text/event-stream responses without an explicit event', async () => {
```
lgrammel, line 102: can we use snapshot testing with fixtures

paoloricciuti: Why tho? This is basically the same test as above no?

lgrammel: because i am dumb and it easier for me to see and understand the exact fixture

lgrammel (changes requested):
please use test fixtures

packages/mcp/src/tool/mcp-http-transport.test.ts
lgrammel: you might need a different approach for capturing them but here is how to use the fixtures: https://github.com/vercel/ai/blob/main/contributing/testing.md#test-fixtures

paoloricciuti: Wait this doesn't concern text generation or stream at all...this is just a behaviour of the MCP client (the generation doesn't even start)...do I still need fixtures considering the rest of the tests for the MCP are not using fixtures AND the fixtures folder is regarding the text generation?

paoloricciuti:
@lgrammel can i get an update on this? https://github.com/vercel/ai/pull/10919#discussion_r2638229005

lgrammel:
Thank you for the pull request. Issue #10915 was closed as not planned. Closing this pull request.

Issue #10329, opened 2025-11-18
Agent interface: allow to specify instructions as system model messages in addition to string


markusjura:
#### Problem statement

I’m using the Vercel AI SDK v6 agent abstraction and would like to use provider-side cached system messages (e.g. [Anthropic cache control](https://v6.ai-sdk.dev/providers/ai-sdk-providers/anthropic#cache-control), but the current agent API only seems to support plain instructions: string on the agent declaration. To use cached system messages, I need to construct `ModelMessage[]` / `SystemModelMessage[]` with cache control metadata, but that’s only possible when calling agent.stream directly. I’d like to be able to define these system messages on the agent itself.

#### Current behavior

Agent declaration accepts instructions: string, but not structured model messages:

```
const chatAgent = new ToolLoopAgent({
  model: anthropic("claude-sonnet-4-5-20250929"),
  // Only string instructions supported here
  instructions: 'You are a helpful assistant...',
  // ...
});

// Later, in a workflow:
const result = await chatAgent.stream({
  messages: [
    { role: 'user', content: 'Hello!' },
  ],
});

```

To use Anthropic cache control for system messages (per docs: [AI SDK Anthropic Cache Control](https://v6.ai-sdk.dev/providers/ai-sdk-providers/anthropic#cache-control)), I instead need to construct SystemModelMessage[] and pass them into agent.stream, bypassing the agent’s instructions field:

```
const result = await chatAgent.stream({
  messages: [
    {
      role: "system",
      content: "You are a helpful assistant...",
      cacheControl: { type: 'ephemeral' }, // Anthropic cache control
    } satisfies SystemModelMessage,
    {
      role: "user",
      content: "Hello!",
    },
  ],
});
```

This works, but it means:
- The system behavior (and cache configuration) lives in the call site instead of in the agent definition.
- I can’t declaratively define or reuse cached system messages at the agent level.

#### Expected behavior

I’d like to be able to define the system messages in the agent declaration using ModelMessage[] or SystemModelMessage[], so I can attach cache control metadata (and other provider-specific properties) there.

Example of a possible API:

```
// Pseudo code – desired API
const chatAgent = new ToolLoopAgent({
  model: anthropic("claude-sonnet-4-5-20250929"),

  // support `systemMessages` in addition to `instructions`
  systemMessages: [
    {
      role: 'system',
      content: 'You are a helpful assistant...',
      cacheControl: { type: 'ephemeral' },
    } satisfies SystemModelMessage,
  ],
});

[...]

PR #10924, opened 2025-12-06
feat(ai): support SystemModelMessage[] in system and instructions properties

### LINKED ISSUE #10914: Allow passing messages array in ToolLoopAgent
### Description

Some models like claude haiku 4.5 perform better when you have instructions split into multiple messages  and you can also control the cache blocks with difference in ttl for things like system environment vs user specific data. 

### AI SDK Version

_No response_

### Code of Conduct

lgrammel:
## Background

For fine grained cache control (Anthropic), system prompts need to be split into multiple system messages (see #10914 ).

## Summary

* allow `SystemModelMessage[]` in `system` property on `Prompt` (**breaking change** on tool call repair functions)
* allow `SystemModelMessage[]` in agent instructions

## Manual Verification

- [x] run `examples/ai-core/src/agent/anthropic-cache-instruction.ts` twice and check request

## Future Work

* introduce type, e.g. `Instruction`, that is an alias for `string | SystemModelMessage | SystemModelMessage[]`

## Related Issues

Resolves #10914 
Builds on #10635 
Relates to #10329

### DIFF
--- a/.changeset/kind-birds-sniff.md
+++ b/.changeset/kind-birds-sniff.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+feat(ai): support SystemModelMessage[] in system and instructions properties
--- a/content/docs/07-reference/01-ai-sdk-core/01-generate-text.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/01-generate-text.mdx
@@ -40,7 +40,7 @@ To see `generateText` in action, check out [these examples](#examples).
     },
     {
       name: 'system',
-      type: 'string | SystemModelMessage',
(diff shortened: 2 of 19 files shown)


nicoalbanese (approved):

lordpotato89:
The comment 

AI SDK UI hooks like useChat return arrays of UIMessage objects, which do not support provider options. We recommend using the convertToModelMessages function to convert UIMessage objects to ModelMessage objects before applying or appending message(s) or message parts with providerOptions.

in 
https://ai-sdk.dev/docs/foundations/prompts

is then not valid anymore? With the fix provided you can pass provider options directly, correct @lgrammel ?

PR #10983, opened 2025-12-08
fix(google-vertex): allow 'vertex' as a key for providerOptions

### LINKED ISSUE #7783: Unexpected behavior when passing providerOptions to vertex/google provider
### Description

The chat model is imported from google (`GoogleGenerativeAILanguageModel`) so it looks for the `google` provider. When doing this from vertex it is kind of weird. For example when using `GoogleVertexImageModel` it does look for the `vertex` provider in the providerOptions

### AI SDK Version

- ai: 5.0.2

aayush-kapoor:
## Background

#7783

When providerOptions are defined in the vertex provider with the key `vertex`, they are ignored and not sent in the request to the model; they only work with `google` key. 

## Summary

Allow the `vertex` to be a valid key for setting providerOptions for vertex provider for consistency sake

## Manual Verification

test cases added

## Checklist

## Future Work

Need to add codemods for the changes in the migration guide

## Related Issues 

Fixes #7783


examples/ai-core/src/generate-text/google-vertex-safety.ts
```diff
@@ -4,9 +4,9 @@ import 'dotenv/config';
 
 async function main() {
   const result = await generateText({
-    model: vertex('gemini-1.5-pro'),
+    model: vertex('gemini-2.5-flash'),
     providerOptions: {
-      google: {
+      vertex: {
```
lgrammel, line 9: will this be a breaking change? if so we need to limit to ai sdk 6 and add it to the migration guide (and potentially add a codemod)

packages/google-vertex/src/google-vertex-provider.ts
```diff
@@ -140,6 +140,7 @@ export function createVertex(
           /^gs:\/\/.*$/,
         ],
       }),
+      providerOptionsName: 'vertex',
```
lgrammel: does this also affect `providerMetadata` (it should because we automatically mv metadata to options) - so the property name should be rename to providerName (which may already exist)

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -47,6 +47,12 @@ type GoogleGenerativeAIConfig = {
    * The supported URLs for the model.
    */
   supportedUrls?: () => LanguageModelV3['supportedUrls'];
+
+  /**
+   * This allows Vertex to use 'vertex' as the primary provider options key
+   * while maintaining backward compatibility with 'google'.
+   */
+  providerOptionsName?: string;
```
lgrammel: extract from `provider` string instead (there are examples eg openai compat afaik)

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -91,11 +97,22 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
...
+        })
+      : undefined;
+
+    if (googleOptions == null) {
+      googleOptions = await parseProviderOptions({
+        provider: 'google',
+        providerOptions,
+        schema: googleGenerativeAIProviderOptions,
+      });
+    }
```
lgrammel: need to change metadata as well

lgrammel: providerName should always be available, no conditional handling needed

packages/google/src/google-generative-ai-language-model.ts
aayush-kapoor: done

packages/google-vertex/src/google-vertex-provider.ts
aayush-kapoor: changed for `providerMetadata` as well

packages/google/src/google-generative-ai-language-model.ts
```diff
@@ -94,12 +94,23 @@ export class GoogleGenerativeAILanguageModel implements LanguageModelV3 {
   }: Parameters<LanguageModelV3['doGenerate']>[0]) {
     const warnings: SharedV3Warning[] = [];
 
-    const googleOptions = await parseProviderOptions({
-      provider: 'google',
+    const providerOptionsName = this.config.provider.includes('vertex')
+      ? 'vertex'
+      : 'google';
```
lgrammel, line 99: can extraction code similar to openai compatible be used here?

lgrammel: or is this challenging bc of `   provider: `google.vertex.${name}`,`?

lgrammel: (then might be fine as is but google.vertex is a problem bc it is different from our standard convention)

lgrammel:
Please add to migration guide. Are any changes to the google vertex provider docs necessary?

lgrammel (approved):

packages/google/src/google-generative-ai-language-model.ts
aayush-kapoor: i actually didn't refer to the openai compat code. do you think i still should?

aayush-kapoor: i think the current method works; extracting from provider via splitting could become problematic.

aayush-kapoor:
added a migration guide - will be adding codemods in a follow up PR once this lands

PR #11015, opened 2025-12-09
fix(vertex/anthropic): passing beta header only for structured outputs

### LINKED ISSUE #10981: Vertex: Unexpected value(s) `structured-outputs-2025-11-13` for the `anthropic-beta` header
### Description

Vertex AI does currently not support the `structured-outputs-2025-11-13` for the `anthropic-beta` header when using sonnet-4.5 in vertex. 

Reproducible example: 
```
import { createVertexAnthropic } from "ai-sdk-anthropic"
import {  generateText, } from "ai"

export const vertexAnthropic = createVertexAnthropic({
    project: "project-id",
    location: "europe-west1", 
    googleAuthOptions: {
        credentials: credentials ,
    },
})

export const test = async () => {
    const res = await generateText({
        model: vertexAnthropic("claude-sonnet-4-5@20250929"),
        messages: [
            {
                role: "user",
                content: "Say hi!",
            },
        ],
    })
    console.log(res)
}
```

### AI SDK Version

- ai@6.0.0-beta.138
- @ai-sdk/google-vertex@4.0.0-beta.102/anthropic

### Code of Conduct

aayush-kapoor:
## Background

Currently, the vertex anthropic provider only checked the model ID and would pass the beta header for all models that allow structured outputs.

But vertex doesn't yet support passing these beta headers. So we would get an [error like this](https://github.com/vercel/ai/issues/10981) even for a simple generateText function

This PR also addresses a bug when strict mode is enabled and the header is passed (shouldn't happen)

## Summary

- only pass the beta header when actually using native output format and when using strict tools

## Manual Verification

the example `google-vertex-anthropic-sonnet-4-5.ts` was used to verify the fix (was failing before)

## Checklist

## Future Work

ideally, we handle this shortcoming in our implementation of vertex provider
- if beta header is to be passed, we intercept it and then we could pass a providerOption saying `structuredOutputMode: 'jsonTool'` (continued in this PR after the comment https://github.com/vercel/ai/pull/11015#issuecomment-3634356904)

## Related Issues

Fixes #10981


aayush-kapoor:
all the commits till here fix the original issue. 

commits after this point are to account for the change in example `/examples/ai-core/src/generate-object/google-vertex-anthropic.ts` which fails in the current flow

that example will fail and give an error. but to enable it for users who want to use the latest anthropic models with vertex, we force the use of JSON tool

examples/ai-core/src/generate-object/google-vertex-anthropic-sonnet-4-5.ts
aayush-kapoor, line 1: was failing before fix

examples/ai-core/src/generate-object/google-vertex-anthropic.ts
aayush-kapoor, line 1: this file (after changing the model ID) was failing even after the prev fix was applied)

packages/anthropic/src/anthropic-messages-language-model.test.ts
```diff
@@ -2335,7 +2529,7 @@ describe('AnthropicMessagesLanguageModel', () => {
         it('should include advanced-tool-use beta header', async () => {
           expect(server.calls[0].requestHeaders).toMatchInlineSnapshot(`
             {
-              "anthropic-beta": "structured-outputs-2025-11-13,advanced-tool-use-2025-11-20",
+              "anthropic-beta": "advanced-tool-use-2025-11-20",
```
lgrammel: why is this dropped?

packages/anthropic/src/anthropic-messages-language-model.test.ts
```diff
@@ -2441,7 +2635,7 @@ describe('AnthropicMessagesLanguageModel', () => {
...
```
lgrammel: why is this dropped?

examples/ai-core/src/generate-object/google-vertex-anthropic-sonnet-4-5.ts
```diff
@@ -0,0 +1,18 @@
+import 'dotenv/config';
+import { vertexAnthropic } from '@ai-sdk/google-vertex/anthropic';
+import { generateText } from 'ai';
+import { z } from 'zod';
+
+async function main() {
```
lgrammel: use `run`

lgrammel (commented):
code looks good but test changes indicate a regression might have been introduced

packages/anthropic/src/anthropic-messages-language-model.test.ts
aayush-kapoor: expected behaviour actually - if this header is passed in vertex provider, it throws an error saying "Unexpected value(s) structured-outputs-2025-11-13 for the anthropic-beta header"

the PR added conditions so that this header isn't passed for functions that don't need it

packages/anthropic/src/anthropic-messages-language-model.test.ts
aayush-kapoor: same as above, expected behaviour

examples/ai-core/src/generate-object/google-vertex-anthropic-sonnet-4-5.ts
aayush-kapoor: actually will be removing this example - only for testing

aayush-kapoor:
> code looks good but test changes indicate a regression might have been introduced

expected behaviour, see comment above!

.changeset/lemon-sheep-behave.md
gr2m, line 1: why two changeset files?

gr2m (approved):
code changes look good but haven't tested end-to-end, I assume you did

.changeset/lemon-sheep-behave.md
aayush-kapoor: will merge into one - created while iterating

aayush-kapoor:
> code changes look good but haven't tested end-to-end, I assume you did

not with UI example (there isn't one for vertex or structured outputs) but yes, tested with different variations of examples (passing tools, tools + object output)

packages/anthropic/src/anthropic-messages-language-model.ts
```diff
@@ -477,7 +486,13 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
     }
 
     // structured output:
-    if (useStructuredOutput) {
+    // Only pass beta when actually using native output_format
+    const usingNativeOutputFormat =
+      useStructuredOutput &&
+      responseFormat?.type === 'json' &&
+      responseFormat.schema != null;
```
lgrammel, line 493: why is this additional condition needed?

lgrammel:
This also seems to fix a bug around strict mode and setting the header. Can this either be separated out into a different PR or added to the description?

lgrammel (approved):

aayush-kapoor:
> This also seems to fix a bug around strict mode and setting the header. Can this either be separated out into a different PR or added to the description?

EDIT: added in the description

packages/anthropic/src/anthropic-messages-language-model.ts
aayush-kapoor: just as an extra check - to completely rule out header being sent for functions that don't need it. otherwise it could send it for simple text gen too. i verified with expected behaviour for approx 6 different cases (combinations essentially)

AVtheking:
@aayush-kapoor why it is required to send the structured output header when strict mode is off ?

PR #11016, opened 2025-12-09
docs: update output schema type description

### LINKED ISSUE #10222: What is the purpose of outputSchema in toolsets? Contrary to what docs says, it is not used to validate the response output
### Description

Check the docs here:

https://v6.ai-sdk.dev/docs/reference/ai-sdk-core/tool#tool.tool.output-schema

It says that it is used for 
> The schema of the output that the tool produces. Used for validation and type inference.

However, I have not 1 or two but dozens of tools that declare the output schema, and then return a different value than the declared output schema and there is no runtime validation error. 
It's worth noting that those tools are not typed using the tool helper function, so there is no type error either. However, now when I'm migrating some of those tools to be on their own module by using the helper function tool for declaring them, I see their responses doesn't match the schema, and of course TypeScript is complaining with a very confusing message to be honest about it.

This is an application that has been working in production since version 5, and we are now on version 6. Everything just works fine. So, I guess the output schema is not actually used to validate the execute function response. 

### AI SDK Version

    "ai": "6.0.0-beta.93",
    "@ai-sdk/azure": "3.0.0-beta.53",

### Code of Conduct

nicoalbanese:
Closes #10222

### DIFF
--- a/content/docs/07-reference/01-ai-sdk-core/20-tool.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/20-tool.mdx
@@ -119,7 +119,7 @@ export const weatherTool = tool({
               isOptional: true,
               type: 'Zod Schema | JSON Schema',
               description:
-                'The schema of the output that the tool produces. Used for validation and type inference.',
+                'The schema of the output that the tool produces. Used for type inference.',
             },
             {
               name: 'toModelOutput',


gr2m (approved):
I trust you over what what agent vercel is saying 😁

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/12454

PR #11020, opened 2025-12-09
support `name` in OpenAI, xAI, and DeepSeek user and assistant messages via `ModelMessage` `ProviderOptions`

clarkg:
## Background

 the xAI chat completions API ( https://docs.x.ai/docs/api-reference#chat-completions ),
 openAI chat completions API (`https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages-user_message-name`), and
 DeepSeek chat completions API ( https://api-docs.deepseek.com/api/create-chat-completion#request )
 all  supports user or assistant messages with an optional `name` parameter , but there is no way to pass that via Vercel AI SDK right now

## Summary

adds support for `name` in xAI user and assistant messages via `ModelMessage` `ProviderOptions`

## Manual Verification

added tests

## Checklist

### DIFF
--- a/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
+++ b/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
@@ -329,4 +329,45 @@ describe('convertToDeepSeekChatMessages', () => {
       `);
     });
   });
+
+  it('passes through openai name on user messages', () => {
+    const { messages, warnings } = convertToDeepSeekChatMessages({
+      prompt: [
+      {
+        role: 'user',
+        content: [{ type: 'text', text: 'Hello' }],
+        providerOptions: { openai: { name: 'alice' } },
+      },
+    ],
+    responseFormat: undefined,
+  });
+
+    expect(warnings).toEqual([]);
+    expect(messages).toEqual([
+      { role: 'user', content: 'Hello', name: 'alice' },
+    ]);
+  });
+
+  it('passes through openai name on assistant messages', () => {
+    const { messages, warnings } = convertToDeepSeekChatMessages({
+      prompt: [
+      {
+        role: 'assistant',
+        content: [{ type: 'text', text: 'Hi' }],
+        providerOptions: { deepseek: { name: 'deepseek-bot' } },
+      },
+    ],
+    responseFormat: undefined,
+    });
+
+    expect(warnings).toEqual([]);
+    expect(messages).toEqual([
+      {
+        role: 'assistant',
+        content: 'Hi',
+        tool_calls: undefined,
+        name: 'deepseek-bot',
+      },
+    ]);
+  });
 });
--- a/packages/openai/src/chat/convert-to-openai-chat-messages.ts
+++ b/packages/openai/src/chat/convert-to-openai-chat-messages.ts
@@ -2,9 +2,19 @@ import {
   SharedV3Warning,
   LanguageModelV3Prompt,
   UnsupportedFunctionalityError,
+  SharedV3ProviderOptions,
 } from '@ai-sdk/provider';
 import { OpenAIChatPrompt } from './openai-chat-prompt';
 import { convertToBase64 } from '@ai-sdk/provider-utils';
+
+function getOpenAIMessageName(
+  providerOptions?: SharedV3ProviderOptions,
+): string | undefined {
+  const name = providerOptions?.openai?.name;
+  return typeof name === 'string' && name.trim().length > 0
+    ? name
+    : undefined;
+}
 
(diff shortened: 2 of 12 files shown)


packages/openai/src/chat/convert-to-openai-chat-messages.ts
```diff
@@ -140,6 +151,7 @@ export function convertToOpenAIChatMessages({
               }
             }
           }),
+          ...(name ? { name: name } : {}),
```
ManyaS-Git, line 154: `name` is only spread into the array-content path. The single-text shortcut above (`messages.push({ role: 'user', content: content[0].text })`) doesn't include it, and the new "passes through openai name on user messages" test exercises exactly that path (a single text part), so the test fails against this implementation. Please add the name spread there too — the xAI converter does this correctly.

packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
```diff
@@ -329,4 +329,45 @@ describe('convertToDeepSeekChatMessages', () => {
...
     });
   });
+
+  it('passes through openai name on user messages', () => {
+    const { messages, warnings } = convertToDeepSeekChatMessages({
+      prompt: [
+      {
+        role: 'user',
+        content: [{ type: 'text', text: 'Hello' }],
+        providerOptions: { openai: { name: 'alice' } },
```
ManyaS-Git, line 339: `getDeepSeekMessageName` reads `providerOptions?.deepseek?.name`, but this test passes `providerOptions: { openai: { name: 'alice' } }` — the wrong namespace (the assistant test just below correctly uses `deepseek`). As written, this test fails against the implementation. Also, the test title says "passes through openai name" while it tests the DeepSeek converter. Please make the test and implementation consistent.

packages/openai/src/chat/convert-to-openai-chat-messages.ts
```diff
@@ -2,10 +2,20 @@ import {
...
+  SharedV3ProviderOptions,
 } from '@ai-sdk/provider';
 import { OpenAIChatPrompt } from './openai-chat-prompt';
 import { convertToBase64 } from '@ai-sdk/provider-utils';
 
+function getOpenAIMessageName(
+  providerOptions?: SharedV3ProviderOptions,
+): string | undefined {
+  const name = providerOptions?.openai?.name;
+  return typeof name === 'string' && name.trim().length > 0
```
ManyaS-Git, line 14: Minor: this validates `name.trim().length > 0` but returns the raw `name`, so `' alice '` would pass validation and be sent with surrounding whitespace. Consider returning `name.trim()`. (Same pattern in the xai and deepseek helpers.)

ManyaS-Git (commented):
Thanks for the PR — adding `name` support across these providers is genuinely useful, and the xAI converter (which handles the single-text shortcut correctly) is a good reference. However, two of the added tests don't match the implementation, plus a few refinements:

1. **OpenAI: `name` is dropped for single-text user messages (inline).** The converter still has the special case `messages.push({ role: 'user', content: content[0].text })`, which bypasses the new `name` spread. The added user test uses exactly this single-text path, so it fails against the current implementation.

2. **DeepSeek: user test uses the wrong provider namespace (inline).** `getDeepSeekMessageName` reads `providerOptions?.deepseek?.name`, but the new user test passes `providerOptions: { openai: { name: 'alice' } }`. The assistant test below correctly uses `deepseek`. As written, the user test fails — please make test and implementation consistent.

3. **Trim validation returns the untrimmed value (inline).** The helpers validate `name.trim().length > 0` but return the raw `name`, so `' alice '` passes validation and is sent with surrounding whitespace.

4. **Code duplication.** `getOpenAIMessageName`, `getXaiMessageName`, and `getDeepSeekMessageName` are identical except for the provider key. Consider a single shared helper (e.g. in `@ai-sdk/provider-utils`) or a `(providerOptions, key)` factory.

5. **Responses API not covered.** The xAI/OpenAI Responses APIs (`xai.responses()`, `openai.responses()`) use different prompt converters, so `name` passed via providerOptions would be silently dropped there. Fine if intentional, but worth noting in the changeset/docs.

6. **xAI doc description looks wrong.** "A per-message unique identifier representing your end-user, which can help xAI to monitor and detect abuse" describes the `user` safety parameter, not the participant `name` field. Please verify the wording.

7. **Minor nits.** Changeset has a typo ("asssistant"), the new test blocks have inconsistent indentation that oxfmt (`pnpm check`) would flag, and the xAI single-line push should be wrapped.

Overall: good direction — once the two failing tests and the single-text path are addressed, this should be ready.

PR #11045, opened 2025-12-10
docs: document tool execution approval

nicoalbanese:
Adds documentation for tool execution approval in AI SDK 6.

Future:
- Update human in the loop recipe
- Add recipe for node

### DIFF
--- a/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
+++ b/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
@@ -105,6 +105,124 @@ tool({
   providers ignore the setting.
 </Note>
 
+## Tool Execution Approval
+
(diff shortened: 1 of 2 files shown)


content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
```diff
@@ -105,6 +105,128 @@ tool({
...
+  inputSchema: z.object({
+    command: z.string().describe('The shell command to execute'),
+  }),
+  needsApproval: true,
+  execute: async ({ command }) => {
+    const { stdout, stderr } = await execAsync(command);
+    return { stdout, stderr };
+  },
+});
+```
```
lgrammel, line 126: needs more context about this particular tool, otherwise there may be confusion around exec etc. also this might not work in other environments. how about an example that is more fake but does not have childprocess deps?

lgrammel (approved):

PR #11048, opened 2025-12-10
chore(ai): add optional ChatRequestOptions to `addToolApprovalResponse` and `addToolOutput`

### LINKED ISSUE #11423: `addToolApprovalResponse` on `sendAutomaticallyWhen` doesn't send `body`
## Description
```ts
const { messages, sendMessage, addToolApprovalResponse } = useChat({
  sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithApprovalResponses, // will be turned off later (see below)
})

const model = 'gpt-4o'

// I'm sending model info to backend
sendMessage({ text: 'hey' }, { body: { model }})

// ...later, on approval
// seems like `addToolApprovalResponse` on `sendAutomaticallyWhen` doesn't send `body` like sendMessage does
addToolApprovalResponse({ id, approved: true })
// so as a workaround, I turned off `sendAutomaticalyWhen` and sendMessage manually after addToolApprovalResponse
sendMessage(undefined, { body: { model }})
```

## Why this happens?
`packages\ai\src\ui\chat.ts > AbstractChat > sendMessage` executes `makeRequest` with options which has .body in it

```ts
    await this.makeRequest({
      trigger: 'submit-message',
      messageId: message.messageId,
      ...options,
    });
```

but `AbstractChat > addToolApprovalResponse` does not.
```ts
// automatically send the message if the sendAutomaticallyWhen function returns true
      if (
        this.status !== 'streaming' &&
        this.status !== 'submitted' &&
        this.sendAutomaticallyWhen?.({ messages: this.state.messages })
      ) {
        // no await to avoid deadlocking
        this.makeRequest({
          trigger: 'submit-message',
          messageId: this.lastMessage?.id,
          // no options here! <-
        });
      }
```

## Idea
You can keep previous body in AbstractChat.body property?

## Related
- https://github.com/vercel/ai/pull/11048 - make it possible to pass body to addToolApprovalResponse
- https://github.com/vercel/ai/issues/10980 - using `sendMessage({text:""})` as a workaround
- https://github.com/vercel/ai/issues/11024 - with `addToolOutput`
- `Cannot read properties of undefined (reading 'startsWith')` - happens when failed to pass model to `ToolLoopAgent` on backend

### AI SDK Version

"ai": "^6.0.0-beta.151",

### Code of Conduct

patrikdevlin:
## Background

Users should be able to customize the api request similar to sendMessage & regenerate

Fixes: https://github.com/vercel/ai/issues/11423

## Summary

Adding optional ChatRequestOptions parameter to addToolApprovalResponse. Add-on work to #8541.

## Checklist

- ~Tests have been added / updated (for bug fixes / features)~ - n/a
- ~Documentation has been added / updated (for bug fixes / features)~ - n/a

## Future Work

n/a


packages/ai/src/ui/chat.ts
```diff
@@ -81,6 +81,11 @@ export type ChatAddToolApproveResponseFunction = ({
    * Optional reason for the approval or denial.
    */
   reason?: string;
+
+  /**
+   * Optional request options to be used when sending the message
+   */
+  options?: ChatRequestOptions;
```
felixarntz, line 89: I'm not sure about this approach because it makes it seem that `options` is relevant to the tool approval when in reality it's only needed to automatically re-submit _if_ that's enabled. Curious what you think @lgrammel 

My initial idea was what I shared in https://github.com/vercel/ai/issues/11423#issuecomment-3825046274 - overloading in that way may not be the most intuitive either, but it would be closer to the root of the problem and directly associated with re-submitting.

lgrammel: My take is that it boils down to the concrete use cases, and what data is available when/where. I.e. are the ChatRequestOption values available only when sendApprovalResponse is invoked? then this is a good place (some of the concerns can be addressed by expanding the jsdoc). If not, then another place might be better.

felixarntz: Makes sense. They're available when invoking `addToolApprovalResponse`, so I think it's a good place, but the same problem also applies when calling `addToolOutput`. So I think we'll need to add it to both methods.

packages/ai/src/ui/chat.ts
```diff
@@ -81,6 +81,11 @@ export type ChatAddToolApproveResponseFunction = ({
    * Optional reason for the approval or denial.
    */
   reason?: string;
+
+  /**
+   * Optional request options to be used when sending the message
```
felixarntz: worth clarifying that this is only relevant _if_ a message is actually being sent as part of this.
```suggestion
   * Optional request options to be used if `sendAutomaticallyWhen` callback returns true
```

packages/ai/src/ui/chat.ts
felixarntz, line 89: Per https://github.com/vercel/ai/pull/11048/changes#r2760036163, can you please add this to `addToolOutput` as well?

felixarntz (commented):
@patrikdevlin upon further review, the approach looks good, but can you please expand it to also cover `addToolOutput`?

If you can, adding test cases for the two changes would be great as well.

felixarntz (approved):
@patrikdevlin I have pushed a few more changes to this branch:

- support `options` in `addToolOutput` too - equally relevant there
- add unit tests for the changes
- add a UI example where this is testable end-to-end

Should be good to go shortly. Thanks for helping to get this fixed!

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/13107

patrikdevlin:
@felixarntz Thanks for wrapping this up and moving things forward here -- Appreciate the work on it!

PR #11104, opened 2025-12-11
fix(devtools): prevent crash when displaying token usage across all models

### LINKED ISSUE #11096: Devtool middleware UI is not rendering on success call @ai-v6
### Description

- The devTools feature in ai SDK core (v6) version, is capturing the responses, but not rendering the UI for success responses
- It is showing data for error responses and blank screen for success responses (irrespective of tool calls)
- for success response, the data is correct, and the data is send from `api/runs/{:id}` api call.
- The issue is with the frontend

Image

### AI SDK Version

- ai: ^6.0.0-beta.0
- ai-sdk/anthropic: ^3.0.0-beta.82,
- ai-sdk/google : ^3.0.0-beta.71

### Code of Conduct

JyothiKumar03:
## Background

The codebase had potential runtime errors when handling usage token calculations. When `usage.inputTokens` or `usage.outputTokens` were non-numeric values (like null, undefined, or strings), the fallback operator ?? would only catch null and undefined, but wouldn't handle other edge cases. This could lead to NaN values being accumulated, breaking token usage tracking, and ultimately breaking the UI.

## Summary

- Replaced unsafe fallback operators with explicit type checking for token usage calculations.
- Changed from using the nullish coalescing operator (??) to properly validating that token values are numbers before using them. 

Now we can:

- Check if the value is actually a number using typeof
- Use 0 as the fallback if it's not
- Store the validated value in a variable for cleaner accumulation

### core fix
```
// Before (unsafe)
input: acc.input + (usage?.inputTokens ?? 0)

// After (type-safe)
const inputTokens = typeof usage?.inputTokens === 'number' ? usage.inputTokens : 0;
input: acc.input + inputTokens

```

## Manual Verification

- Tested with valid numeric token values - accumulation works correctly
- Tested with null/undefined token values - properly defaults to 0
- Tested with string/object token values - properly defaults to 0 instead of producing NaN
- Verified accumulated totals remain accurate across multiple usage objects
- Conducted UI rendering of all 3 types - `error` , `response` , `tool-call` with major providers like `Google` , `Anthropic` and `Openai`

Verified:

- No React runtime errors
- DevTools no longer crashes when viewing usage
- Token usage displays correctly when present
- Missing or non-numeric fields no longer break the UI
- Confirmed production build via pnpm build and pnpm start

## Checklist

- [] Tests have been added / updated (for bug fixes / features)

## Future Work

This is more like a simple bug, not much work on this maybe. The token-usage is already in a standardised format 

## Related Issues

Fixes https://github.com/vercel/ai/issues/11096

### DIFF
--- a/.changeset/healthy-jars-sneeze.md
+++ b/.changeset/healthy-jars-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/devtools': patch
+---
+
+Fix crash caused by invalid token usage values in DevTools
--- a/packages/devtools/src/viewer/client/app.tsx
+++ b/packages/devtools/src/viewer/client/app.tsx
@@ -411,9 +411,13 @@ function App() {
     return steps.reduce(
       (acc, s) => {
         const usage = parseJson(s.usage);
+        const inputTokens =
+          typeof usage?.inputTokens === 'number' ? usage.inputTokens : 0;
+        const outputTokens =
+          typeof usage?.outputTokens === 'number' ? usage.outputTokens : 0;
         return {
-          input: acc.input + (usage?.inputTokens ?? 0),
-          output: acc.output + (usage?.outputTokens ?? 0),
+          input: acc.input + inputTokens,
+          output: acc.output + outputTokens,
         };
       },
       { input: 0, output: 0 },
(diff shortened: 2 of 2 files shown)


martinsione:
Running into the same issue with the anthropic provider, tested locally and confirmed this fixed it for me.

JyothiKumar03:
@martinsione thanks for the confirmation, glad to know that this is tested by others !

lgrammel:
Closing this PR because issue #11096 has already been closed.

PR #11111, opened 2025-12-11
fix(gateway): remove `cause` from custom handling of `GatewayAuthenticationError` error

gr2m:


packages/ai/src/prompt/wrap-gateway-error.ts
```diff
@@ -1,19 +1,12 @@
...
-  GatewayAuthenticationError,
-  GatewayModelNotFoundError,
-} from '@ai-sdk/gateway';
+import { GatewayAuthenticationError } from '@ai-sdk/gateway';
 import { AISDKError } from '@ai-sdk/provider';
 
 export function wrapGatewayError(error: unknown): unknown {
-  if (
-    GatewayAuthenticationError.isInstance(error) ||
-    GatewayModelNotFoundError.isInstance(error)
```
gr2m, line 10: @lars @mclenhard throwing the custom `unauthenticated` error for `GatewayModelNotFoundError` was incorrect. Do we want a custom error for when a model was not found? I don't think we need it?

ghost (approved):

gr2m:
superseded by https://github.com/vercel/ai/pull/11113

PR #11114, opened 2025-12-12
fix(openai): don't use item_reference for client-executed tool calls

### LINKED ISSUE #8216: AI SDK returns "No tool call found for function call output" when state is "output-available"
### Description

I'm encountering an error when processing a response from the AI SDK on Vercel.
The error states that no tool call was found for a function call output, even though the tool's `state` is set to `output-available`.

**Steps to Reproduce:**

1. Send a user message to the AI SDK.
2. Get a response containing multiple tool calls with `state: "output-available"`.
3. Send a follow-up user message.
4. Receive the following error:

```
Error [AI_APICallError]: No tool call found for function call output with call_id call_i1njW2L3AzcgxnVh5SP1m5sK.
    at ...
statusCode: 400
responseBody: {
  "error": {
    "message": "No tool call found for function call output with call_id call_i1njW2L3AzcgxnVh5SP1m5sK.",
    "type": "invalid_request_error",
    "param": "input"
  }
}
```

**Relevant Data (processedMessages):**

```json
{
  "processedMessages": [
    {
      "id": "1c94d843-d42d-438c-aa69-45a8453e4e78",
      "role": "user",
      "parts": [{ "text": "Introduce the store", "type": "text" }],
      "metadata": {
        "selectedData": {},
        "createdAt": "2025-08-22T13:17:36+03:30",
        "type": "ai"
      }
    },
    {
      "id": "5d637201-f2e8-4271-a67e-f97da69fdb7f",
      "role": "assistant",
      "parts": [
        { "type": "step-start" },
        {
          "type": "tool-getProductsFromChatMessage",
          "toolCallId": "call_0CH0G84eu1DLdLDZohiaxGyz",
          "state": "output-available",
          "input": {},
          "output": "10am"
        },
        { "type": "step-start" },
        {
          "type": "tool-textGenerate",
          "toolCallId": "call_56i6ijoDMLHWRTRX05O3GQEp",
          "state": "output-available",
          "input": {
            "role": "",
            "text": "Here is a comparison of the three perfumes based on their features and prices."
          },
          "output": null
        },
        { "type": "step-start" },
        {
          "type": "tool-getRelatedPrompt",
          "toolCallId": "call_i1njW2L3AzcgxnVh5SP1m5sK",
          "state": "output-available",
          "input": {
            "result": [
              { "type": "query", "text": "What are the best-selling perfumes?", "url": "" },
              { "type": "query", "text": "Can you recommend perfumes for special occasions?", "url": "" },
              { "type": "query", "text": "What are the top-rated perfumes of 2025?", "url": "" }
            ]
          },
          "output": {
            "items": [
              { "type": "query", "text": "What are the best-selling perfumes?", "url": "" },
              { "type": "query", "text": "Can you recommend perfumes for special occasions?", "url": "" },
              { "type": "query", "text": "What are the top-rated perfumes of 2025?", "url": "" }
            ]
          }
        }
      ]
    },
    {
      "id": "8422dc10-e8ea-468c-8abe-acfc807852a5",
      "role": "user",
      "parts": [{ "text": "What are the top-rated perfumes of 2025?", "type": "text" }],
      "metadata": { "selectedData": {} }
    }
  ]
}
```

**Expected Behavior:**
The tool call with `state: "output-available"` should be recognized and handled properly without causing a `No tool call found` error.

**Actual Behavior:**
The API responds with a 400 error stating that the tool call ID does not exist.

**Environment:**

* AI SDK Vercel (latest)
* Model: `openai:gpt-4o-mini`
* Node.js version: \[Your Node version]

**Question:**

* Is there an extra step required to register or persist tool calls between responses?
* Should `output-available` parts automatically map to function call outputs?

### AI SDK Version

    "ai": "^5.0.19",
    "@ai-sdk/react": "^2.0.19",
    "@ai-sdk/openai": "^2.0.17",

### Code of Conduct

CalebBarnes:
## Summary

When `store=true` (the default for the OpenAI Responses API), using `item_reference` with `itemId` (fc_...) for client-executed tool calls causes a mismatch error because the corresponding `function_call_output` uses `toolCallId` (call_...) as `call_id`. OpenAI cannot reliably match these different IDs, especially for multi turn conversations with parallel tool calls, resulting in:

```
No tool call found for function call output with call_id call_xxx
```

This fix:
1. Removes the `item_reference` optimization for client-executed tool calls
2. Removes the `id` field from `function_call` objects for client-executed tools

Provider-executed tools are unaffected since both the call and result are stored on OpenAI's side and can be referenced by `item_reference`.

## Root Cause

The issue occurs because:
- When a `function_call` has an `itemId` (like `fc_...`), the code was using `item_reference` to reduce payload size
- However, when sending the `function_call_output`, it must use `call_id` (like `call_...`) to reference the tool call
- OpenAI has no way to match `item_reference` with `id: "fc_..."` to `function_call_output` with `call_id: "call_..."`

This only affects **client-executed tools** (user-defined functions). Provider-executed tools (like web_search, code_interpreter) work correctly because OpenAI stores both the call and result.

## Test Plan

- Tested with multi-turn conversations using parallel tool calls
- Verified all 6 turns complete successfully with `store=true` (default)
- Previously failed after 3-5 turns with the error above

Fixes #8216


packages/openai/src/responses/convert-to-openai-responses-input.ts
```diff
@@ -161,31 +158,22 @@ export async function convertToOpenAIResponsesInput({
...
 
-              const id = part.providerOptions?.openai?.itemId as
-                | string
-                | undefined;
-
-              // item references reduce the payload size
-              if (store && id != null) {
-                input.push({ type: 'item_reference', id });
-                break;
-              }
```
CalebBarnes, line 178: We intentionally don't use `item_reference` for client-executed tool calls. Since we must send `function_call_output` with `call_id`, and `item_reference` uses a different id (`itemId`), OpenAI can't reliably match them. It's also not possible to use `item_reference` for a client executed `function_call_output`. This is the fix for "No tool call found for function call output" error.

packages/openai/src/responses/convert-to-openai-responses-input.ts
```diff
@@ -204,7 +192,6 @@ export async function convertToOpenAIResponsesInput({
                 call_id: part.toolCallId,
                 name: part.toolName,
                 arguments: JSON.stringify(part.input),
-                id,
```
CalebBarnes, line 207: Don't include id `fc_..` on client executed `function_call_output` since this is actually the reference item id on the CALL, not the result.

lgrammel:
Thanks - can you add an example under `examples/ai-core` that i can use to reproduce the bug?

Also pls add a patch changeset.

shangwfa:
I’ve encountered the same issue as well. Do you have an estimated timeframe for when it might be resolved? Thank you.

lgrammel:
Closing this PR because issue #8216 has already been closed.

PR #11159, opened 2025-12-13
 add async polling support for flux-kontext-* models, fixing #11149

### LINKED ISSUE #11149: Fireworks: image generation responds differs depending on model
Our current example for image generation using the `@ai-sdk/fireworks` provider works when using `accounts/fireworks/models/stable-diffusion-xl-1024-v1-0` as provider.

https://github.com/vercel/ai/blob/fe2831ab5299630099af3b738c295a44a5ab2bae/examples/ai-core/src/generate-image/fireworks.ts#L9

But it fails when using `accounts/fireworks/models/flux-kontext-pro`. Unlike with the stable-diffusion model, flux-kontext-pro requests return a JSON object with a `request_id` which then needs to be sent repeatedly to https://docs.fireworks.ai/api-reference/get-generated-image-from-flux-kontex until it responds with `result.status === 'Ready'` and and `result.result.sample` set to an image URL, which then can be downloaded to retrieve the actual image.

ShyamSathish005:
## Background

The Fireworks provider only supported synchronous binary image responses. However, models like flux-kontext-pro and [flux-kontext-max](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) use an asynchronous workflow API that:

1. Requires polling the /get_result endpoint until status is Ready
2. Returns an image URL that must be fetched separately
3. This caused the provider to fail or return unexpected JSON responses instead of images when using these models (issue #11149).

## Summary

Added async polling support for Fireworks flux-kontext-* image generation models:
Added async model configuration for flux-kontext-pro and flux-kontext-max
Implemented doGenerateAsync() method for async image generation workflow
Implemented pollForImageUrl() method with configurable interval and timeout
Added Zod schemas for async submit and poll response parsing
Added model IDs to FireworksImageModelId type
Added 7 comprehensive tests for async model behavior

## Manual Verification

```
import { fireworks } from '@ai-sdk/fireworks';

const { image } = await generateImage({
  model: fireworks.image('accounts/fireworks/models/flux-kontext-pro'),
  prompt: 'A cute baby sea otter',
});
```

## Checklist

- [ y] Tests have been added / updated (for bug fixes / features)
- [ y] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ y] I have reviewed this pull request (self-review)

## Future Work

1.  Added documentation for the new pollIntervalMillis and pollTimeoutMillis provider options
2.  Consider adding support for other async Fireworks models as they become available

## Related Issues

Fixes #11149


gr2m:
FYI we are looking into a generic polling API:
https://github.com/vercel/ai/issues/12381

felixarntz (approved):
@ShyamSathish005 Thank you for creating this PR, and apologies for the delayed review.

I just took a look, refreshed this against latest `main`, which required some merge conflict resolutions primarily due to #10281 which had happened in the meantime.

I tested the fix using the examples (plus one new one I added) - your solution was almost good to go, there was just one error where the poll response object needed to be nullable, which I addressed in https://github.com/vercel/ai/pull/11159/commits/7818ea70f58c842b0087f8ba5b48cb41777bd62e.

Thanks again, this is great, even more so since even some of our existing examples that were broken are now working!

felixarntz:
@gr2m Would be great to get you to look over this one too because you worked on the related #10281, and I fixed a few things here.

Basically, the `generate-image/fireworks-edit*.ts` examples have not been working because they require async support, and this PR fixes it.

You can test both examples running on `main` vs in this PR branch to verify. I added one more example just for generation with one of those async requiring models just to have that more basic case covered as well.

packages/fireworks/src/fireworks-image-model.ts
```diff
@@ -3,16 +3,31 @@ import {
...
+  asyncPollResponseSchema,
+  asyncSubmitResponseSchema,
+} from './fireworks-image-api';
 import { FireworksImageModelId } from './fireworks-image-options';
 
+const DEFAULT_POLL_INTERVAL_MILLIS = 500;
+const DEFAULT_POLL_TIMEOUT_MILLIS = 120000; // 2 minutes for image generation
+
 interface FireworksImageModelBackendConfig {
   urlFormat: 'workflows' | 'workflows_edit' | 'image_generation';
```
felixarntz: @gr2m I see you added `workflows_edit` in #10281, but reviewing that in the context of this PR, I believe the distinction is not about whether editing is supported, but about whether the model requires async image generation endpoint usage (regardless of whether generating or editing an image).

This becomes evident also because the two models that need `workflows_edit` are the exact two models that need `async: true`.

I'd suggest one of the following:
- either get rid of `workflows_edit` and use the `async` boolean flag
- or replace `workflows_edit` with `workflows_async` and remove the `async` boolean flag

What's your preference?

ShyamSathish005: Thanks for the thorough review and for handling those merge conflicts, @felixarntz 

Regarding the urlFormat vs async flag

I personally lean towards Option 2: replacing workflows_edit with workflows_async and removing the async boolean flag. My reasoning is that it keeps the configuration object flatter and treats the 'async-polling' nature as a specific backend type/workflow rather than a secondary property. It prevents a state where someone could accidentally set urlFormat: 'image_generation' but async: true, which would be an invalid configuration.

However, I'm happy to follow whichever direction @gr2m prefers to ensure it aligns with the upcoming generic polling API (#12381)

felixarntz: I implemented that in https://github.com/vercel/ai/pull/11159/changes/a195d908d738437675eb304a493cc73d58edbcf2.

It's a backward compatible change because this schema is entirely internal to the provider implementation.

PR #11160, opened 2025-12-13
feat(ui): add UIMessage part helper functions

itsfabioroma:
## Summary

Add convenience functions to extract and aggregate UIMessage parts:

- `getTextParts(message)` - get all text parts as array
- `getTextContent(message)` - get joined text string  
- `getReasoningParts(message)` - get all reasoning parts as array
- `getReasoningContent(message)` - get joined reasoning string
- `isReasoningStreaming(message)` - check if reasoning is currently streaming
- `isTextStreaming(message)` - check if text is currently streaming

## Motivation

When working with reasoning models (DeepSeek, Claude, etc.), the SDK streams reasoning as multiple parts that need to be aggregated for display. Currently, users must manually filter and join parts:

```tsx
// Before: manual filtering/joining
const reasoningText = message.parts
  .filter((p) => p.type === 'reasoning')
  .map((p) => p.text)
  .join('');
```

```tsx
// After: one-liner
import { getReasoningContent } from 'ai';
const reasoningText = getReasoningContent(message);
```

This pattern is common enough that built-in helpers reduce boilerplate and improve DX.

## Changes

- Added 6 helper functions to `packages/ai/src/ui/ui-messages.ts`
- Added comprehensive tests for all helpers
- Exported functions from `packages/ai/src/ui/index.ts`

## Test plan

- [x] All existing tests pass
- [x] New tests cover all helper functions
- [x] Tests cover edge cases (empty parts, no matching parts)

### DIFF
--- a/.changeset/feat-ui-message-helpers.md
+++ b/.changeset/feat-ui-message-helpers.md
@@ -1,0 +1,14 @@
+---
+'ai': patch
+---
+
+feat(ui): add UIMessage part helper functions
+
+Add convenience functions to extract and aggregate UIMessage parts:
+
+- `getTextParts(message)` - get all text parts as array
+- `getTextContent(message)` - get joined text string
+- `getReasoningParts(message)` - get all reasoning parts as array
+- `getReasoningContent(message)` - get joined reasoning string
+- `isReasoningStreaming(message)` - check if reasoning is currently streaming
+- `isTextStreaming(message)` - check if text is currently streaming
--- a/content/docs/07-reference/02-ai-sdk-ui/48-ui-message-helpers.mdx
+++ b/content/docs/07-reference/02-ai-sdk-ui/48-ui-message-helpers.mdx
@@ -1,0 +1,245 @@
+---
+title: UIMessage Helpers
+description: API Reference for UIMessage helper functions.
+---
+
+# UIMessage Helpers
+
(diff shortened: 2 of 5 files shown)


lgrammel:
Please add changeset and documentation.

@gr2m do you think this is worth adding? have refrained so far, bc it bloats library and docs and the filtering is trivial

itsfabioroma:
I’ve added the changeset and documentation.

While the logic itself is simple, this is a very common and repetitive usage pattern, and we already see users re-implementing it downstream. Providing it at the library level improves consistency and discoverability, and reduces duplicated ad-hoc filters across projects.

lgrammel:
Fwiw I think this is the wrong kind of abstraction. It makes software less flexible and bloats it. Developers should be using filter etc themselves. But just my personal opinion.

nicoalbanese:
Thanks for this @itsfabioroma and appreciate the contribution and feedback! 

I have pushed back on adding these utilities in the past as well as they are small/short enough that I don't feel they are necessary to bundle into the library (e.g. 4 lines -> 2 lines). Equally, I would prefer to push the pattern of switching over the message parts and rendering markup conditionally, rather than using these types of utilities. 

So with that in mind, I would not be in favor of adding this (just my opinion)

gr2m:

PR #11166, opened 2025-12-14
fix(ai): handle AbortError from middleware in streamText

ghempton:
## Background

When middleware uses its own `AbortController` to control the model stream, aborting would cause `result.steps` and other promises to hang forever instead of resolving.

## Summary

Fixed the abort handling in `streamText` to treat any `AbortError` as an abort, regardless of which signal caused it.

**Before:** Only treated `AbortError` as abort when `abortSignal?.aborted` was true (checking the outer signal passed to `streamText`).

**After:** Any `AbortError` is properly handled as an abort, which covers cases where middleware uses its own `AbortController`.

The key change in `stream-text.ts`:
```typescript
} catch (error) {
  if (isAbortError(error)) {
    // Handle any AbortError as an abort, regardless of which signal caused it.
    // This covers cases where middleware uses its own AbortController.
    abort();
  } else {
    controller.error(error);
  }
}
```

## Manual Verification

1. Created a test that simulates middleware triggering an AbortError
2. Verified test fails without the fix (times out waiting for `result.steps`)
3. Verified test passes with the fix (`result.steps` resolves properly)

## Checklist

### DIFF
--- a/.changeset/fix-abort-error-middleware.md
+++ b/.changeset/fix-abort-error-middleware.md
@@ -1,0 +1,7 @@
+---
+'ai': patch
+---
+
+fix(ai): handle AbortError from middleware in streamText
+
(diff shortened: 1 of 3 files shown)


ghempton:
Any pointers on getting feedback for this PR? This is a pretty real issue where if I pass in a different abortController into `transformParams` in a middleware and that abort controller gets aborted, the stream hangs indefinitely.

PR #11167, opened 2025-12-14
fix(openai): parallel tool call for OpenAI tools

SamyPesse:
## Background

Using the AI SDK v6 with OpenAI, I noticed that when using `parallelToolCalls: true` and the `apply_patch` tool from OpenAI, it fails with: `APICallError [AI_APICallError]: No tool output found for apply patch call call_abc.` when the LLM is responding with multiple calls.

## Summary

Changed break to continue in the `apply_patch` tool result handling block, so after processing an apply_patch result the loop continues to process remaining tool results instead of exiting early.

## Manual Verification

- Verified by running [this example](https://gist.github.com/SamyPesse/1bb94cf635b418a23dd5412877247bc8)
- unit test added

## Checklist


SamyPesse:
Weirdly it looks like passing `"parallel_tool_calls": false` doesn't prevent it from returning a responses with multiple `apply_patch` calls 🤔 

I'll dig deeper to confirm, but it might indicate that the `apply_patch` can't be used with the `ai` SDK before this issue is solved as passing `parallelToolCalls: false` is not a workaround.

aayush-kapoor:
hi @SamyPesse, would you happen to have a code snippet we can use to recreate the issue? then i can dive into it. i'm not sure if the `continue` approach is the right one here as it may lead to unforeseen errors or issues

SamyPesse:
@aayush-kapoor here is a reproducible example: https://gist.github.com/SamyPesse/1bb94cf635b418a23dd5412877247bc8

Run this file locally and you'll get the error:

```
AI_APICallError: No tool output found for apply patch call call_rWBdy9SS3Vw6ZaJflTyMO9Md.
```

And then run it with this fix and it'll work.

aayush-kapoor (approved):

lgrammel (approved):

PR #11168, opened 2025-12-15
fix(ai): prevent header loss when statusText is undefined in writeHead

OmarSalouss:
## Problem

In `pipeAgentUIStreamToResponse`, when `statusText` is not set, all headers are lost.

The issue occurs in `writeToServerResponse`, which calls `response.writeHead(status != null ? status : 200, statusText, headers);` 

This always passes `statusText` in `arguments[1]`, even when it's `undefined`.

According to Node.js's `writeHead` implementation ([reference](https://github.com/jshttp/on-headers/blob/master/index.js#L121)), the function determines where to read headers based on whether `arguments[1]` is a string or not

```jsx
var headerIndex = length > 1 && typeof arguments[1] === 'string'
    ? 2
    : 1
```

Headers should be read from `arguments[2]` when `arguments[1]` is a string, otherwise from `arguments[1]`.

When `statusText` is `undefined`, `arguments[1]` is treated as non-string, so `headerIndex` becomes `1`, causing the code to read headers from `arguments[1]` and ignore the actual headers in `arguments[2]`.

**Result:** No headers are applied—neither developer-provided headers (from `headers` parameter) nor default UI message stream headers (`UI_MESSAGE_STREAM_HEADERS`).

## Solution

Only pass `statusText` when it has a value. According to the [Node.js v25.2.1 documentation](https://nodejs.org/api/http.html#responsewriteheadstatuscode-statusmessage-headers), this ensures Node.js correctly identifies which argument contains the headers.

### DIFF
--- a/packages/ai/src/util/write-to-server-response.test.ts
+++ b/packages/ai/src/util/write-to-server-response.test.ts
@@ -96,12 +96,105 @@ describe('writeToServerResponse', () => {
     // Verify all chunks were eventually written
     expect(mockResponse.writtenChunks).toHaveLength(3);
   });
+
+  it('should set headers correctly when statusText is undefined', async () => {
+    const mockResponse = createSimpleMockResponse();
+
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode('test data'));
+        controller.close();
+      },
+    });
+
+    const expectedHeaders = {
+      'X-Example-Header': 'example-value',
+      'X-Example-Chat-Title': 'My Conversation',
+    };
+
+    writeToServerResponse({
+      response: mockResponse,
+      status: 200,
+      statusText: undefined,
+      headers: expectedHeaders,
+      stream,
+    });
+
+    await mockResponse.waitForEnd();
+
+    expect(mockResponse.statusCode).toBe(200);
+    expect(mockResponse.headers).toEqual(expectedHeaders);
+    expect(mockResponse.ended).toBe(true);
+    expect(mockResponse.writtenChunks).toHaveLength(1);
+  });
+
+  it('should set headers correctly when statusText is provided', async () => {
+    const mockResponse = createSimpleMockResponse();
+
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode('test data'));
+        controller.close();
+      },
+    });
+
+    const expectedHeaders = {
+      'X-Example-Header': 'example-value',
+      'X-Example-Chat-Title': 'New Chat Session',
+    };
+
+    writeToServerResponse({
+      response: mockResponse,
+      status: 201,
+      statusText: 'Created',
+      headers: expectedHeaders,
+      stream,
+    });
+
+    await mockResponse.waitForEnd();
+
+    expect(mockResponse.statusCode).toBe(201);
+    expect(mockResponse.statusMessage).toBe('Created');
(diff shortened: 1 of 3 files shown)


obadakhalili:
+1, we need this fix.

lgrammel:
Thanks. Please add a test that reproduces the issue and a changeset.

packages/ai/src/util/write-to-server-response.ts
```diff
@@ -16,7 +16,7 @@ export function writeToServerResponse({
   headers?: Record<string, string | number | string[]>;
   stream: ReadableStream<Uint8Array>;
 }): void {
-  response.writeHead(status ?? 200, statusText, headers);
+  response.writeHead(status ?? 200, ...[statusText, headers].filter(arg => arg !== undefined) as [string?, OutgoingHttpHeaders?]);
```
lgrammel: please us if-else with undefined check on statusText instead

packages/ai/src/util/write-to-server-response.ts
```diff
@@ -1,4 +1,4 @@
-import { ServerResponse } from 'node:http';
+import { ServerResponse, OutgoingHttpHeaders } from 'node:http';
```
lgrammel: note needed any more

OmarSalouss:
@lgrammel i added tests to verify headers are set correctly with undefined statusText

packages/ai/src/util/write-to-server-response.test.ts
```diff
@@ -119,7 +212,14 @@ class SimpleMockResponse extends EventEmitter {
     headers?: Record<string, string | number | string[]>,
   ): void {
     this.statusCode = statusCode;
-    this.statusMessage = statusMessage || '';
+
+    if (typeof statusMessage === 'string') {
+      this.statusMessage = statusMessage;
+      this.headers = headers || {};
```
lgrammel: prefer to leave undefined in the mock / for testing (no `||` defaults) also, below (4 cases)

packages/ai/src/util/write-to-server-response.test.ts
```diff
@@ -119,7 +212,14 @@ class SimpleMockResponse extends EventEmitter {
...
+    } else {
+      this.statusMessage = '';
```
lgrammel, line 220: same, leave undefined, this is for testing

packages/ai/src/util/write-to-server-response.test.ts
```diff
@@ -179,7 +280,14 @@ class BackpressureMockResponse extends EventEmitter {
...
+      this.headers = statusMessage || {};
```
lgrammel: same

packages/ai/src/util/write-to-server-response.test.ts
```diff
@@ -96,12 +96,105 @@ describe('writeToServerResponse', () => {
...
+    expect(mockResponse.writtenChunks).toHaveLength(1);
+  });
 });
 
 class SimpleMockResponse extends EventEmitter {
   writtenChunks: any[] = [];
   statusCode = 0;
   statusMessage = '';
+  headers: Record<string, string | number | string[]> = {};
   ended = false;
```
lgrammel, line 198: support undefined when it is not set if possible (for more exact testing) unless there are issues

lgrammel:
Could not modify the branch (pls allow change by maintainers in the future) so I created a separate PR with a bit of cleanup. #11229

PR #11225, opened 2025-12-16
fix(google): use dynamic providerOptionsName for thoughtSignature retrieval

Jille00:
## Summary

When using `@ai-sdk/google-vertex` with Gemini thinking models (e.g. `gemini-3-pro-preview`), multi-step tool calls fail with error:
```
Unable to submit request because function call is missing a `thought_signature`
```

**Root cause:** The SDK stores `thoughtSignature` under `providerOptions.vertex` but retrieves it using hardcoded `providerOptions.google` in `convertToGoogleGenerativeAIMessages`.

**This fix:**
- Adds `providerOptionsName` parameter to `convertToGoogleGenerativeAIMessages`
- Uses dynamic key access `part.providerOptions?.[providerOptionsName]` instead of hardcoded `.google`
- Passes `providerOptionsName` from `getArgs()` to the conversion function

## Test plan

- [x] Tested locally with `gemini-3-pro-preview` via `@ai-sdk/google-vertex`
- [x] Multi-step tool calls now work correctly with thinking models
- [ ] Existing tests should pass (no breaking changes)

Fixes #11181

### DIFF
--- a/.changeset/fix-vertex-thought-signature.md
+++ b/.changeset/fix-vertex-thought-signature.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/google": patch
+---
+
+fix(google): use dynamic providerOptionsName when retrieving thoughtSignature in convertToGoogleGenerativeAIMessages
+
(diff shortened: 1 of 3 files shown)


rthm-health:
Have the same issue!

Jille00:
@lgrammel could you look at this? Unable to use vertex with reasoning at all

lavisht22:
Hi! We are facing the same issue, can we look into this please?

PR #11229, opened 2025-12-16
fix(ai): prevent header loss when statusText is undefined in writeHead 

lgrammel:
## Background

In `pipeAgentUIStreamToResponse`, when `statusText` is not set, all headers are lost.

The issue occurs in `writeToServerResponse`, which calls `response.writeHead(status != null ? status : 200, statusText, headers);` 

This always passes `statusText` in `arguments[1]`, even when it's `undefined`.

According to Node.js's `writeHead` implementation ([reference](https://github.com/jshttp/on-headers/blob/master/index.js#L121)), the function determines where to read headers based on whether `arguments[1]` is a string or not

```jsx
var headerIndex = length > 1 && typeof arguments[1] === 'string'
    ? 2
    : 1
```

Headers should be read from `arguments[2]` when `arguments[1]` is a string, otherwise from `arguments[1]`.

When `statusText` is `undefined`, `arguments[1]` is treated as non-string, so `headerIndex` becomes `1`, causing the code to read headers from `arguments[1]` and ignore the actual headers in `arguments[2]`.

**Result:** No headers are applied—neither developer-provided headers (from `headers` parameter) nor default UI message stream headers (`UI_MESSAGE_STREAM_HEADERS`).

## Summary

Only pass `statusText` when it has a value. According to the [Node.js v25.2.1 documentation](https://nodejs.org/api/http.html#responsewriteheadstatuscode-statusmessage-headers), this ensures Node.js correctly identifies which argument contains the headers.

## Related Issues

Continues #11168


dancer (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/11232

PR #11319, opened 2025-12-19
fix(provider/google): Fix Google model tool schema error

R-Taneja:
## Background

Fixes https://linear.app/vercel/issue/AI-4776/investigate-roo-code-gemini-mcp-tool-schema-error

## Summary

When tool schemas with nested empty object properties were sent (like launchOptions: { type: "object" }) to Gemini through the Gateway, the google package's schema conversion function was treating these empty objects as unnecessary and converting them to undefined, which removed them from the properties object entirely. But the required array still referenced them by name. Gemini's validation then failed with "property is not defined" because it found "launchOptions" in required but couldn't find it in properties.

The fix adds a simple check: only convert empty objects to undefined at the root level (where it's actually useful), but preserve them as { type: "object" } when they're nested properties so they stay in sync with the required array.

## Manual Verification
New test passes

## Checklist

## Related Issues
https://linear.app/vercel/issue/AI-4776/investigate-roo-code-gemini-mcp-tool-schema-error

### DIFF
--- a/packages/google/src/convert-json-schema-to-openapi-schema.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.ts
@@ -7,7 +7,7 @@ export function convertJSONSchemaToOpenAPISchema(
   jsonSchema: JSONSchema7Definition | undefined,
   isRoot = true,
 ): unknown {
-  // parameters need to be undefined if they are empty objects:
+  // Handle empty object schemas: undefined at root, preserved when nested
   if (jsonSchema == null) {
     return undefined;
   }
--- a/packages/google/src/convert-json-schema-to-openapi-schema.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.ts
@@ -92,7 +92,9 @@ export function convertJSONSchemaToOpenAPISchema(
   }
 
   if (allOf) {
-    result.allOf = allOf.map(item => convertJSONSchemaToOpenAPISchema(item, false));
+    result.allOf = allOf.map(item =>
+      convertJSONSchemaToOpenAPISchema(item, false),
+    );
   }
   if (anyOf) {
     // Handle cases where anyOf includes a null type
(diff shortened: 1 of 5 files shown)


packages/google/src/convert-json-schema-to-openapi-schema.ts
```diff
@@ -5,12 +5,21 @@ import { JSONSchema7Definition } from '@ai-sdk/provider';
  */
 export function convertJSONSchemaToOpenAPISchema(
   jsonSchema: JSONSchema7Definition | undefined,
+  isRoot = true,
 ): unknown {
   // parameters need to be undefined if they are empty objects:
```
shaper: comment is stale -- needs some sort of updating given we now have some conditional logic below.

packages/google/src/convert-json-schema-to-openapi-schema.ts
```diff
@@ -5,12 +5,21 @@ import { JSONSchema7Definition } from '@ai-sdk/provider';
  */
 export function convertJSONSchemaToOpenAPISchema(
```
shaper, line 6: this function is used in two places:
- `google-generative-ai-language-model.ts`
- `google-prepare-tools.ts`

do we need/intend for this logic to be changed in both places?

asking in part for us to work toward answering the question -- are we confident revising how this works for the MCP/tools case that led to this does not break or introduce other issues.

packages/google/src/convert-json-schema-to-openapi-schema.ts
shaper, line 1: Ideally we would have example generate and stream scripts that repro the issue but I know you haven't been able to produce one yet. Is there any detail we can turn up in the google generative ai API docs that could be background on why the logic was structured here the way it was, or even better, to support the need for the changes we are making here?

packages/google/src/convert-json-schema-to-openapi-schema.ts
R-Taneja: the function still returns `undefined` if `isRoot = true` (default case), so that behavior is unchanged

for nested empty objects, we now preserve `{ type: "object" }` instead of sending `undefined`. I don't believe this would cause an issue because gemini should still accept it. I tested it by sending a request to gemini direct with `launchOptions: { type: 'object' }` and it worked, although I hear you and think someone with more experience/context on this provider take a look @dancer @gr2m

packages/google/src/convert-json-schema-to-openapi-schema.ts
R-Taneja: Couldn't find anything relevant in gemini API docs but I was able to repro - I added 2 scripts to this PR

On main branch (after running `pnpm build` in `packages/google`), both scripts fail with: "APICallError [AI_APICallError]: * GenerateContentRequest.tools[0].function_declarations[0].parameters.required[1]: property is not defined" as expected

On this branch (after running `pnpm build` in `packages/google`), both scripts output: "Tool call: navigate { url: 'https://example.com', launchOptions: {} }" (as expected, tool call worked)

dancer (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/11323

PR #11351, opened 2025-12-21
feat (validation, schematics) Add TypeDriver Validator

sinclairzx81:
## Checklist

## Summary

Project: [TypeDriver](https://github.com/sinclairzx81/typedriver)

This PR adds an additional validation package (@ai-sdk/typedriver) that supports unified validation and type inference for the following schema definition formats and libraries.

- JSON Schema (Draft 3 - Draft 2020-12)
- TypeScript Definition DSL
- TypeBox
- Zod 3 / 4 / Mini
- Valibot
- ArkType

The package is written to provide better support for JSON Schema (beyond the existing jsonSchema(...) function) as well as to enable high performance validation for libraries that implement the new Standard JSON Schema specification. Validation is based exclusively on JSON Schema and is tested heavily against the official [JSON Schema Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite) to ensure semantic correctness.

The TypeScript DSL is provided by way of [TypeBox](https://github.com/sinclairzx81/typebox) (which is a dependency of TypeDriver)

## Example

This package exports a single `schema(...)` function that accepts a wide range of schema definitions and libraries. The function returns instances of `Schema<T>` from the `@ai-sdk/provider-utils` package, which can be passed to SDK interfaces that accept schemas.

```typescript
import { schema } from '@ai-sdk/typedriver';

const Vector3 = schema(`{ 
  x: number,
  y: number,
  z: number
}`)                                               // const Vector3: Schema<{
                                                  //   x: number;
                                                  //   y: number;
                                                  //   z: number;
                                                  // }>

  model: anthropic('claude-3-7-sonnet-latest'),
  output: Output.object({ schema: Vector3 }),
  prompt: 'Generate normal Vector.',
});
```

### With JSON Schema

```ts
const Vector3 = schema({
  type: 'object',
  required: [ 'x', 'y', 'z'],
  properties: {
    x: { type: 'number' },
    y: { type: 'number' },
    z: { type: 'number' }
  }
})
```

### With Zod

```ts
const Vector3 = schema(z.object({
  x: z.number(),
  y: z.number(),
  z: z.number()
}))
```

## Manual Verification

This PR includes tests for TypeScript, JSON Schema, Zod, Valibot, ArkType and TypeBox inference and validation.

## Future

TypeDriver may have some utility in deriving type inference directly from the [MCP Protocol Specification](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.json)

- [Auto Sync MCP Protocol Specification Types](https://tsplay.dev/N9493w)

## Related Issues

- https://github.com/vercel/ai/pull/11224

---

Submitting for consideration


lgrammel:
I would love typebox integration (ideally via Standard JSON Schema), but I am wondering what else this provides:

- zod is already natively supported
- raw json schema is already supported via `jsonSchema()`
- valibot is already supported via `@ai-sdk/valibot`
- arktype is supported through standard json schema

Is there a way to have a more limited `@ai-sdk/typebox`? or is TypeBox standard json schema compatible?

I do not want to add more complexity for our users to understand. However, if direct typebox support is not possible, could this be a community maintained library (outside of `@ai-sdk/`) that we link to from our docs?

sinclairzx81:
@lgrammel Hi! Thanks for the review!

I’ll provide some additional info below, but I’m certainly open to thoughts on alternative designs if this PR is not a good fit. Regarding TypeBox though, this package would service as an integration path, but the package itself is more about improving the JSON Schema validation story in this SDK.

---

## TypeDriver Overview

> but I am wondering what else this provides:

This package is essentially a constrained version of the existing `jsonSchema(...)` function, but where validation happens exclusively through a JSON Schema compliant validator. It is written specifically to remove potential mismatches between schema representation, validation and type inference (which is currently possible with the existing `jsonSchema(...)` function)

```typescript
const A = jsonSchema({ type:'string' }, value => {  // schema: string
  return typeof value === 'number'                  // check:  number
    ? { ... }                                       // infer:  boolean ??
    : { ... }
})
                                                    // ensure consistency
                                                    // 
const B = schema({ type: 'string' })                // schema: string
                                                    // check:  string
                                                    // infer:  string
```

From a design standpoint, the implementation was written to try consolidate multiple schema definitions under the SDK’s `Schema<T>` structure. The ability to do this was facilitated by way of the new Standard JSON Schema specification.

> ⚠️ Note: Standard JSON Schema is "NOT" JSON Schema. TypeBox cannot implement on that basis.

| Input | Mapping |Output |
| ---      | --- | --- |
| JSON Schema | Implicit | `Schema<T>` |
| TypeScript | Parse Implicit | `Schema<T>` |
| Zod| Standard JSON Schema | `Schema<T>` |
| Valibot | Standard JSON Schema | `Schema<T>` |
| ArkType | Standard JSON Schema  | `Schema<T>` |

... and where the reduction to `Schema<T>` was somewhat considered against the current downstream type differentiation / inference handling observed below.

https://github.com/vercel/ai/blob/main/packages/provider-utils/src/schema.ts#L63-L87

```typescript
// multi type differentiation 
export type InferSchema<SCHEMA> =
  SCHEMA extends ZodSchema<infer T>
    ? T
    : SCHEMA extends StandardSchema<infer T>
      ? T
      : SCHEMA extends LazySchema<infer T>
        ? T
        : SCHEMA extends Schema<infer T>
          ? T
          : never;
```

... which could simply be reduced to.

```typescript
export type InferSchema<SCHEMA> = SCHEMA extends Schema<infer T> 
  ? T 
  : never
```

... but where the package offers library integration (as before) + other features.

- Full support for Zod, Valibot, ArkType, Effect  (and TypeBox)
- Acceleration support for JSON Schema via Standard JSON Schema
- Internationalized Error Reporting (i18n support)
- and [more](https://github.com/sinclairzx81/typedriver?tab=readme-ov-file#Features)

## Standardized Validation

As noted above, the main intention here is to merely provide this SDK a means to validate data strictly in accordance with the JSON Schema [specification](https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft2020-12) regardless of the underlying library or schema format.

```typescript
const A = schema(z.toString())       // Zod
const B = schema('string')           // TypeScript
const C = schema({ type: 'string' }) // JSON Schema

A.validate('foo')                     // uses JSON Schema validation !
B.validate('bar')                     // uses JSON Schema validation !
C.validate('baz')                     // uses JSON Schema validation !
```

The `@ai-sdk/typedriver` is designed to relinquish validation control away from each library and delegates it exclusively to a standards based validator (specifically one that can be verified for compliance with the JSON Schema specification via [automated testing](https://github.com/sinclairzx81/typebox/tree/main/test/jsonschema) - so TypeBox (or Ajv it doesn't matter which))

- [JSON Schema 2020-12 Test Suite](https://github.com/json-schema-org/JSON-Schema-Test-Suite/tree/main/tests/draft2020-12)

... and where the validator is guaranteed to return validation check results the same as any other validator that implements the specification.

- [Other Spec Compliant Validators](https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=toolingTypes&licenses=&languages=&drafts=&toolingTypes=&environments=&showObsolete=false&supportsBowtie=false#validator)

... and where the requirement to have consistent schema validation typically emerges in micro-service infrastructures where application domain models are shared between systems (as JSON) and thus need to be understood by each system equally.... (or data integrity hazards may result)

---

So, this package is attempting to solve a few "things" (where I think this package could provide some utility to other parts of the code base, at least in part), but the most pressing requirement for me would be the JSON Schema validation represented as a core feature of the SDK (especially as it relates to the MCP protocol)

With regards to a potential `@ai-sdk/typebox` package, this is possible of course, but I would much prefer to see a Ajv validator with a decent JSON Schema inference system behind it. 

As an alternative to this package, I would certainly be open to contributing a modern JSON Schema inference system for the benefit of this SDK should it decide to embrace standards compliant JSON Schema validation.

Cheers!
S

sinclairzx81:
@lgrammel Sorry, just to follow up on this aspect

> I would love typebox integration (ideally via Standard JSON Schema), but I am wondering what else this provides:

As mentioned, I can’t really support the specification in TypeBox, but I did want to raise the specification as a potential integration hazard waiting to happen. In order for this (or other) SDKs to receive a valid JSON Schema, the Standard JSON Schema specification would require JSON Schema to effectively invalidate itself....

As per the specification outline:

https://standardschema.dev/json-schema

> Standard JSON Schema is a common interface designed to be implemented by JavaScript and TypeScript entities that **"ARE"** or can be converted to JSON Schema.

... which is effectively asking JSON Schema to provide a redundant self referential adapter to itself (via invalid keyword)

```typescript
// Json Schema

const S = { type: 'string' }        // <-- also TypeBox type

// Standard JSON Schema

const S = {                         
  type: 'string',

  '~standard': {                    // Non-Standard (should be prefixed with `x-` fails Ajv)
    jsonSchema: {
       input: () =>                
        ({ type: 'string' }),       // Redundant Self Referential Meta Schema
       output: () => 
        ({ type: 'string' })        // Redundant Self Referential Meta Schema
    }
  }
}
```
And I am somewhat mindful of trying to mitigate a future where this could become a problem

```typescript
const A = Output.object({ type: 'string' }) // error: not Standard Json Schema ?

const S = { type: 'string' }

const B = Output.object({ ...S, '~standard': { jsonSchema: { ... } } }) // hmmm
```

I wouldn't expect many users to really know the difference between JSON Schema and Standard JSON Schema (so expect some confusion here), but in terms of using the specification as a direct integration device, I would approach with some care, particularly in light of a possible ....

[JSON Schema Standardization under Ecma International](https://github.com/orgs/json-schema-org/discussions/938)

---

All said, the `@ai-sdk/typedriver` package does provide some shielding against many of the inverted design aspects of the specification (as it would relate to JSON Schema specifically), but irrespective of this package, I would encourage a degree of care to ensure Standard JSON Schema doesn't end up hindering plain JSON Schema compatibility with SDK interfaces down the line....

Native TypeBox compatibility would largely depend on the SDK accepting plain JSON Schema as an input format.

PR #11390, opened 2025-12-22
fix(core): validateUIMessages throws on output-error tool parts with undefined input

obadakhalili:
if tool call input is invalid, ai sdk avoids setting `input` field and sets `rawInput` field instead, problem is, the validation logic still checks for `input` and expects it to be defined, even though it is intentionally not defined when the tool input is invalid. this renders methods like `pipeAgentUIStreamToResponse` which uses `createAgentUIStream` unusable when the AI incorrectly sends invalid input because internally it uses `validateUIMessages` which will always fail because `input` field isn't set.

### DIFF
--- a/packages/ai/src/ui/validate-ui-messages.ts
+++ b/packages/ai/src/ui/validate-ui-messages.ts
@@ -150,7 +150,8 @@ const uiMessagesSchema = lazySchema(() =>
                   toolName: z.string(),
                   toolCallId: z.string(),
                   state: z.literal('output-error'),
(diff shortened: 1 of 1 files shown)


lgrammel:
Thanks. Please add a unit test case that reproduces the issue and a patch changeset for releasing.

lgrammel:
Superseeded by #11393

lgrammel:
Thanks for your contribution, I just saw another PR that fixes this and already has tests etc.

PR #11393, opened 2025-12-22
fix(ai): skip tool input validation when output-error has undefined input

muniter:
## Background

if tool call input is invalid, ai sdk avoids setting `input` field and sets `rawInput` field instead, problem is, the validation logic still checks for `input` and expects it to be defined, even though it is intentionally not defined when the tool input is invalid. this renders methods like `pipeAgentUIStreamToResponse` which uses `createAgentUIStream` unusable when the AI incorrectly sends invalid input because internally it uses `validateUIMessages` which will always fail because `input` field isn't set (copied from #11390 )

## Summary

Only validate when input is available since there are two possible states for output-error the error happened before inputSchemas was validated or after. Right now code always validates as it was after

## Manual Verification

Ran the test, and I'm fixing an issue from a live application, also the issue provides a minimal reproduction.

## Checklist

## Related Issues

Fixes #11392
Similar to #11390

### DIFF
--- a/.changeset/fix-safe-validate-ui-messages-undefined-input.md
+++ b/.changeset/fix-safe-validate-ui-messages-undefined-input.md
@@ -1,0 +1,7 @@
+---
+'ai': patch
+---
+
+fix(ai): skip tool input validation in `safeValidateUIMessages` when `output-error` state has undefined input
+
+Fixes #11392
--- a/packages/ai/src/ui/validate-ui-messages.test.ts
+++ b/packages/ai/src/ui/validate-ui-messages.test.ts
@@ -1000,7 +1000,7 @@ describe('validateUIMessages', () => {
       `);
     });
 
-    it('should validate tool input when state is output-error', async () => {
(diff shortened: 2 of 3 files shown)


lgrammel (approved):

nicoalbanese (approved):

obadakhalili:
i don't understand the point of copying my pr (even the description) as is. i don't mind my pr being closed, but just wondering why.

muniter:
> i don't understand the point of copying my pr (even the description) as is. i don't mind my pr being closed, but just wondering why.

Hey @obadakhalili I did not copy your PR I was working on debugging this issue yesterday, seems like you were doing the same, you can see the approach each one of us took was a bit different (I did not modify the schema and also first submitted the minimal repo I worked on). The PR description was edited by the vercel team, my original one doesn't mention your PR because I had not seen it at all. Hope that clears things up!

obadakhalili:
@muniter it does, description text confused me at first, so I missed that the code was different. all good though, and your fix is better. lol we were on it at the same time 😄

PR #11424, opened 2025-12-24
Bugfix/add image propers to generate call

mclenhard:
## Background

The gateway image model class was missing images and masks from the actual request, but these are necessary for image editing. 

## Summary

Updated doGenerate method to pass in the required params 

## Checklist

### DIFF
--- a/.changeset/perfect-scissors-chew.md
+++ b/.changeset/perfect-scissors-chew.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': major
+---
+
+adds files and mask to the gateway doGenerate method
--- a/packages/gateway/src/gateway-image-model.ts
+++ b/packages/gateway/src/gateway-image-model.ts
@@ -38,6 +38,8 @@ export class GatewayImageModel implements ImageModelV3 {
     size,
     aspectRatio,
     seed,
+    files,
+    mask,
     providerOptions,
     headers,
     abortSignal,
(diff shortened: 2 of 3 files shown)


.changeset/perfect-scissors-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/gateway': major
+---
+
+adds files and mask to the gateway doGenerate method
```
shaper, line 5: `feat (provider/gateway): include files and mask in image generate calls`

and update PR summary to be the same as above.

`doGenerate` is ambiguous to reference as we also have it for language models

packages/gateway/src/gateway-image-model.ts
```diff
@@ -64,6 +66,8 @@ export class GatewayImageModel implements ImageModelV3 {
           ...(size && { size }),
           ...(aspectRatio && { aspectRatio }),
           ...(seed && { seed }),
+          ...(files && { files }),
```
shaper, line 69: if the type of the params for files or mask is uint8array we will send a verbose massive int array (which I believe will also not get deserialized into the right type server side). So, *if* it is a uint8array, then we need to do an encode/decode for this case. we can do it as a separate PR but please make sure to follow up with this work soon since it can be a common case and I know it will not work well.

See how we encode/decode uint8array file attachments for language models already as an example:

https://github.com/vercel/ai/blob/main/packages/gateway/src/gateway-language-model.ts#L188

and there is a decode side in gateway server to go with it.

shaper:
Continuing work in https://github.com/vercel/ai/pull/11493

PR #11438, opened 2025-12-26
fix(google): validate tool names to prevent API errors

codewarnab:
## Background
The Gemini API enforces strict validation on tool (function) names. Currently, the SDK passes invalid tool names (e.g., those starting with numbers) directly to the API, resulting in a 400 error. This error can be confusing and hard to debug for users.
api error can be verified with [this](https://pastebin.com/YpvAWYj0)
## Summary
I added client-side validation in `prepareTools` to check if tool names match Gemini's requirements (regex: `^[a-zA-Z_][a-zA-Z0-9_.:-]*$`).
If a name is invalid, the SDK now throws a descriptive `Error` immediately, instead of making a failing API call.
### Design Decision: Validation vs. Sanitization
I considered automatically sanitizing tool names (e.g., prefixing with `fn_` or removing invalid characters) to make them valid. However, I decided against this for the following reasons:
1.  **Complexity**: Automatic mapping requires maintaining a consistent map between the original name and the sanitized name to ensure model responses (which use the sanitized name) can be correctly mapped back to the original tool. This adds significant complexity to the SDK.
2.  **Debugging Confusion**: If the SDK silently changes tool names, users inspecting network traffic will see tool names they didn't define. This implicit behavior can lead to confusion when debugging.
3.  **Explicit is better than implicit**: Throwing an error forces the developer to fix the issue at the source, ensuring their code is correct and explicit.

But I am open to feedback if maintainers prefer a sanitization approach.

## Manual Verification
I created a reproduction test case in `google-prepare-tools.test.ts` that defines a tool named `123invalid`.
- Before the fix: The test would fail (or rather, the API would fail in a real scenario).
- After the fix: The test passes by catching the expected "Invalid tool name" error.

I verified this by running `pnpm test:node src/google-prepare-tools.test.ts` in `packages/google`.

## Checklist
- [ ] A _patch_ changeset for relevant packages has been added


packages/google/src/google-prepare-tools.test.ts
```diff
@@ -472,3 +472,19 @@ it('should add warnings for google maps on unsupported models', () => {
...
+          type: 'function',
+          name: '123invalid',
+          description: 'Invalid name',
+          inputSchema: { type: 'object', properties: {} },
+        },
+      ],
+      modelId: 'gemini-2.5-flash',
+    }),
+  ).toThrowError(/Invalid tool name/);
+});
```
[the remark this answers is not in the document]
codewarnab: **Good catch!** I've added comprehensive test cases in commit 73939b07a covering:

Valid edge cases:

_validName — names starting with underscore ✓
tool.name, tool:name, tool-name — names with allowed special characters (dots, colons, dashes) ✓
Invalid edge cases:

-invalid, .invalid, :invalid — names starting with disallowed characters ✓
tool@name, tool#name — names with disallowed special characters in the middle ✓
All tests pass and confirm the regex pattern correctly validates tool names according to Gemini API requirements**

packages/google/src/google-prepare-tools.ts
```diff
@@ -199,6 +199,12 @@ export function prepareTools({
   for (const tool of tools) {
     switch (tool.type) {
       case 'function':
+        if (!/^[a-zA-Z_][a-zA-Z0-9_.:-]*$/.test(tool.name)) {
```
ManyaS-Git: Two suggestions for the validation:

- **Enforce the max length.** Gemini's actual error text is: "Must start with a letter or an underscore. Must be a-z, A-Z, 0-9, or contain underscores (_), dots (.), colons (:), or dashes (-), with a maximum length of 64." The current regex has no length bound, so a 100-char name still ships to the API and fails there. `^[a-zA-Z_][a-zA-Z0-9_.:-]{0,63}$` would match the API rule exactly, and mentioning the limit in the error message would help users.

- **Prefer a typed error.** The repo convention is to extend `AISDKError` (see AGENTS.md) so callers can catch the failure programmatically. A plain `Error` forces string-matching on the message. Could this throw a dedicated error class instead?

packages/google/src/google-prepare-tools.test.ts
```diff
@@ -472,3 +472,173 @@ it('should add warnings for google maps on unsupported models', () => {
     ]
   `);
 });
+
+it('should throw error for invalid tool name starting with number', () => {
```
ManyaS-Git: These nine cases are a good fit for a table-driven test — something like `it.each([...invalidNames])(...)` and `it.each([...validNames])(...)` would make the matrix much more readable. Worth adding:

- a name longer than 64 characters (currently not validated at all),
- an assertion that `toolWarnings` stays empty for valid names.

ManyaS-Git (commented):
Thanks for the PR! Validating tool names client-side is a good idea: it turns a confusing 400 from the Gemini API into an immediate, actionable error at the call site. The design decision to validate rather than silently sanitize is well-reasoned, and the character class in the regex matches what the Gemini API actually enforces (per its error text: letters, digits, `_`, `.`, `:`, `-`, starting with a letter or underscore).

A few things to address before this is ready:

1. **Missing max-length check.** Gemini's error message also states "with a maximum length of 64". The regex `^[a-zA-Z_][a-zA-Z0-9_.:-]*$` accepts names of arbitrary length, so an over-long name will still reach the API and fail there. See inline comment.

2. **The PR is currently conflicting with `main`, and the validation would not cover the Gemini 3 combined-tools path.** Since this PR branched, `prepareTools` gained a second function-declaration loop for the Gemini 3 combined built-in-tools path (see #13920 / `01fa606a9`). After rebasing, validation added only in the main loop will be bypassed whenever that path is taken. I'd suggest extracting a small `assertValidToolName(name)` helper and calling it from both paths.

3. **Missing changeset.** The repo requires a `patch` changeset for production code changes (see AGENTS.md), and the checklist in the PR description has it unchecked.

4. **Consider a typed error instead of `Error`.** The repo convention is to extend `AISDKError` so consumers can catch the failure programmatically rather than matching on a message string (see the error pattern in AGENTS.md). A plain `Error` makes that brittle.

5. **Tests are repetitive.** Nine near-identical `it()` blocks could be a table-driven `it.each([...])`, which would make the valid/invalid matrix much more readable. Consider also adding a case for a name longer than 64 characters.

Overall the direction is right — happy to re-review once the length check and the two code paths are covered.

codewarnab:
@ManyaS-Git 
Thanks for the detailed review! I'll resolve the conflicts, apply these fixes, and push an update in a couple of hours.

codewarnab:
Thanks for the review @ManyaS-Git — all points are addressed, and I've merged current `main` in to clear the conflicts. Latest state is 0d67d0f.

- **64-character limit** — the regex is now `/^[a-zA-Z_][a-zA-Z0-9_.:-]{0,63}$/`, so 1 leading char + 63 gives a hard 64 maximum.
- **Typed error** — replaced the bare `Error` with `InvalidArgumentError` from `@ai-sdk/provider` (`argument: 'tools'`), so it's a proper `AISDKError` subclass.
- **Both code paths** — extracted `isValidToolName` / `validateToolName` helpers and call them from both the standard function-declaration loop and the Gemini 3 combined-tools path. That second path didn't exist when this branch was cut; it came in with the `main` merge, and it's now covered.
- **Changeset** — added `.changeset/fix-google-tool-name-validation.md` as a patch for `@ai-sdk/google`.
- **Tests** — the six one-off `it` blocks are now two table-driven `it.each` tables: 8 invalid cases (bad leading character, space, `@`, `#`, 65 chars) and 8 valid ones (underscore lead, digits, dots, colons, dashes, exactly 64 chars), asserting on the `InvalidArgumentError` type rather than a message regex. Plus a dedicated case for the Gemini 3 combined-tools path.

Verified locally on the merged branch: `pnpm --filter @ai-sdk/google test` passes 743 tests across both the node and edge configs, and `tsc --build packages/google` is clean. The branch diff against `main` is now only these three files — no incidental changes.

PR #11443, opened 2025-12-27
feat(ai): add abort reason field to stream chunks

### LINKED ISSUE #11442: Feature: provide abort reason for abort chunk
### Description

In my project, I combined two abort signals. In my scenario, I need to allow users to manually abort the request while also triggering an abort due to a timeout after a certain period of time.

```typescript
  const timeoutSignal = AbortSignal.timeout(timeout)
  const signals = [timeoutSignal]
  if (externalSignal) {
    signals.push(externalSignal)
  }
  const finalSignal = AbortSignal.any(signals)
```

However, the reason fields for these two types of abort signal are different. At present, when an abort is triggered, the abort chunk returned by the aisdk does not include the reason for the abort. I would like to be able to determine the abort reason directly from the abort chunk so that I can clearly inform the user about why the request was aborted.

Of course, I know I could achieve this by managing my finalSignal separately, but that seems a bit too cumbersome for me.

### AI SDK Version

ai: 5.0.98

### Code of Conduct

EurFelux:
## Background
Issue #11442 requests abort reason visibility on abort chunks so callers can distinguish manual vs timeout aborts when combining signals.

## Summary
- Attach `AbortSignal.reason` to `abort` chunks in `streamText`, and extend stream/UI chunk types and schemas to carry `reason`.
- Add tests covering abort reason propagation and update affected snapshots.
- Add an example under `examples/ai-core/src/stream-text` demonstrating abort reasons with `AbortSignal.any` (manual vs timeout).
- Document the optional `reason` field on abort stream parts in core and UI stream protocol docs.

## Manual Verification
- Ran `pnpm -C /Volumes/CIRCLE/dev/ai/packages/ai exec tsx /Volumes/CIRCLE/dev/ai/examples/ai-core/src/stream-text/openai-compatible-abort-reason.ts` and observed abort reasons for both manual and timeout cases.

## Checklist

## Related Issues
Fixes https://github.com/vercel/ai/issues/11442


packages/ai/src/generate-text/stream-text.test.ts
```diff
@@ -14423,6 +14491,7 @@ describe('streamText', () => {
                 "warnings": [],
               },
               {
+                "reason": [AbortError: This operation was aborted],
```
lgrammel: this isn't json serializable. is there way to make the reason a string?

packages/ai/src/ui-message-stream/ui-message-chunks.ts
```diff
@@ -168,6 +168,7 @@ export const uiMessageChunkSchema = lazySchema(() =>
       }),
       z.strictObject({
         type: z.literal('abort'),
+        reason: z.unknown().optional(),
```
lgrammel: can this be an optional string instead to ensure that it is serializable?

EurFelux: You’re right; it should be a serializable value.

EurFelux: I have changed it to an optional string.

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1011,7 +1011,13 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT extends Output>
         // abort handling:
         function abort() {
           onAbort?.({ steps: recordedSteps });
-          controller.enqueue({ type: 'abort' });
+          controller.enqueue({
+            type: 'abort',
+            // Use getErrorMessage to serialize reason (unknown) to string for safe JSON transmission
+            ...(abortSignal?.reason !== undefined
+              ? { reason: getErrorMessage(abortSignal.reason) }
+              : {}),
```
lgrammel: `getErrorMessage` seems incorrect here. Even if it constrains the functionality, would it be possible to check if abortSignal?.reason is a string and ignore it otherwise? (i.e. change the condition in 1017 such that we know that the reason is a string)

EurFelux: You can’t just accept a string only. Please see https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason: by default, the `reason` field is set to a value of type `DOMException`, not a `string`. In fact, you can set it to any value using the `abort` function, which is why I previously designed the field’s type as `unknown`. Given that this is the default behavior, if we only accepted `string`, we would lose the `reason` in most use cases.

EurFelux: I used `getErrorMessage` for the sake of code reuse, since this function can serialize errors of type `unknown`. I’m also aware that the fact that `reason` is likely an `Error` instance isn’t immediately obvious from its name, which might confuse maintainers who aren’t familiar with this behavior. To address this, I added a comment explaining why I’m using this function, but it seems the explanation wasn’t clear enough. I’ll provide a more precise explanation of this behavior in the comment.

lgrammel (approved):

PR #11451, opened 2025-12-28
fix(anthropic): handle object value in web_fetch error-json

jalexy12:
## Summary

When `web_fetch` returns an error (e.g., `url_not_allowed` or `url_not_accessible`), the SDK crashes with:

```
SyntaxError: JSON Parse error: Unexpected identifier "object"
```

## Root Cause

In `convert-to-anthropic-messages-prompt.ts`, the error handling assumes `output.value` is always a JSON string:

```typescript
const errorValue = JSON.parse(output.value as string);
```

However, `output.value` is already an object:

```json
{"type":"error-json","value":{"type":"web_fetch_tool_result_error","errorCode":"url_not_allowed"}}
```

When `JSON.parse()` receives an object, it gets coerced to `"[object Object]"` and fails.

## Fix

Check if the value is a string before parsing:

```typescript
const errorValue =
  typeof output.value === 'string'
    ? JSON.parse(output.value)
    : output.value;
```

## Testing

Tested locally with `@ai-sdk/anthropic@3.0.1` and `ai@6.0.3`. The agent now handles web_fetch errors gracefully instead of crashing.


packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -788,7 +788,10 @@ export async function convertToAnthropicMessagesPrompt({
                   const output = part.output;
 
                   if (output.type === 'error-json') {
-                    const errorValue = JSON.parse(output.value as string);
+                    const errorValue =
+                      typeof output.value === 'string'
+                        ? JSON.parse(output.value)
```
lgrammel, line 793: Please do not use `JSON.parse` directly. Instead use `parseJSON`.

lgrammel:
Please add unit tests.

jalexy12:
Seems like this was addressed in a recent patch!

PR #11462, opened 2025-12-30
feat(anthropic): add new citation types

EurFelux:
## Background

The Anthropic API has added new text citation types (`content_block_location` and `search_result_location`) that are not yet supported by the AI SDK. Additionally, the citation schema was duplicated in both response and chunk schemas, making maintenance harder.

See https://platform.claude.com/docs/en/api/typescript/messages#text_citation

## Summary

- Extract `textCitationSchema` for reuse across response and chunk schemas, improving maintainability
- Add support for new text citation types: `content_block_location` and `search_result_location`

## Manual Verification

N/A - schema changes only, no runtime behavior change for existing citation types.

## Checklist


packages/anthropic/src/anthropic-messages-api.ts
```diff
@@ -492,6 +492,50 @@ export type AnthropicResponseContextManagement = {
   applied_edits: AnthropicResponseContextManagementEdit[];
 };
 
+export const textCitationSchema = z.discriminatedUnion('type', [
```
aayush-kapoor: no need to abstract - can you change it to in line schema definition please

aayush-kapoor (commented):
do you have a specific instance of a code snippet where anthropic will return these types? will make it very easy to verify. no worries if not

packages/anthropic/src/anthropic-messages-api.ts
EurFelux: OK! I've done it

EurFelux:
Here is the examples provided in the official documentation:

content_block_location: https://platform.claude.com/docs/en/build-with-claude/citations#custom-content-documents

search_result_location: https://platform.claude.com/docs/en/build-with-claude/search-results#example-direct-search-results

aayush-kapoor (approved):

aayush-kapoor:
i have approved it - but before merging, please add unit tests (using fixtures as described in https://github.com/vercel/ai/blob/main/contributing/testing.md#manual-testing)

aayush-kapoor (changes requested):
i also realised we need to make the proper changes in `anthropic-messages-language-model.ts` to allow for these citation types to be passed

EurFelux:
Hello, I’ve carefully reviewed the use cases provided by Claude and identified the following issue:

1. `search_result_location` is returned as part of the response only when the user actively provides search results.

For example: https://platform.claude.com/docs/en/build-with-claude/search-results#example-direct-search-results

```typescript
messages: [
    {
      role: "user",
      content: [
        {
          type: "search_result" as const,
          source: "https://docs.company.com/api-reference",
          title: "API Reference - Authentication",
          content: [
            {
              type: "text" as const,
              text: "All API requests must include an API key in the Authorization header. Keys can be generated from the dashboard. Rate limits: 1000 requests per hour for standard tier, 10000 for premium."
            }
          ],
          citations: { enabled: true }
        },
        {
          type: "search_result" as const,
          source: "https://docs.company.com/quickstart",
          title: "Getting Started Guide",
          content: [
            {
              type: "text" as const,
              text: "To get started: 1) Sign up for an account, 2) Generate an API key from the dashboard, 3) Install our SDK using pip install company-sdk, 4) Initialize the client with your API key."
            }
          ],
          citations: { enabled: true }
        },
        {
          type: "text" as const,
          text: "Based on these search results, how do I authenticate API requests and what are the rate limits?"
        }
      ]
    }
  ]
```

As you can see, there is a Claude-specific `type: "search_result"` here, whereas the current UserContent interface in aisdk only includes: https://github.com/vercel/ai/blob/5e6d932133536fded8c74294acaf5163a3a5b067/packages/provider-utils/src/types/user-model-message.ts#L22

This type of content does not belong to any of the current UserContent interfaces. This means we need to make more changes than we’d imagined to support this feature, because we don’t currently support this type of UserContent.

2. `content_block_location` is provided as part of the response when a user attempts to build custom document content.

See: https://platform.claude.com/docs/en/build-with-claude/citations#custom-content-documents

```typescript
{
    "type": "document",
    "source": {
        "type": "content",
        "content": [
            {"type": "text", "text": "First chunk"},
            {"type": "text", "text": "Second chunk"}
        ]
    },
    "title": "Document Title", # optional
    "context": "Context about the document that will not be cited from", # optional
    "citations": {"enabled": True}
}
```

Here, we might want to use `FilePart` to implement custom documents, even though it doesn’t actually represent a file that exists in the filesystem. However, in the current implementation, we primarily rely on the `mediaType` field to determine how to construct the document content in Anthropic format. In the case of custom documents, though, the `mediaType` field simply does not exist.

https://github.com/vercel/ai/blob/8940d47681f37774ba2b15094cba8ff637ffd362/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts#L183-L185

Of course, we could also use `providerOptions` to carry extra data and indirectly achieve these two functionalities, but that’s probably not a good design.

Since these two features require significant changes, I don’t think they should all be implemented in this single PR. Additionally, given that implementing these features may very likely necessitate modifying existing generic interfaces, we should discuss further whether it is truly necessary to implement them.

aayush-kapoor:
> Of course, we could also use providerOptions to carry extra data and indirectly achieve these two functionalities, but that’s probably not a good design.

will have to discuss with the team since we don't wanna make changes to the core schema just for a provider specific feature. so might just have to rely on providerOptions after all

are there any other resources on these 2 types where i can learn more about them? where they are invoked and how they are returned etc? it will make sense to create different PRs for both of them and dive deeper. 

will hold off on merging this PR before we are sure of the approach

PR #11464, opened 2025-12-30
feat(mcp): enhance tool output handling and schema definitions

tlgimenes:
## Background

I need **tighter MCP ↔ LLM integration** so tool call results can be better consumed without overflowing the context window. This enables downstream behaviors like **reading specific properties**, filtering, and validating/parsing results based on a tool’s declared output shape.

## Summary

- Normalized MCP `tools/call` results into **LLM-friendly `ToolResultOutput`**:
  - If the MCP server returns **structured output** (`toolResult` or `structuredContent`), we return it as **JSON model output** so the LLM (and app code) can work directly with properties.
  - If the MCP server marks the result as an error (`isError`), we return **error text** extracted from text content for better surfaced failures.
  - Otherwise we pass through the regular MCP `content` array.
- Added support for **`outputSchema`** on MCP tool definitions and plumbed it through when building AI SDK tools, so consumers can **parse/validate tool outputs** using the declared schema.

## Manual Verification

- Fetched tools from the mock MCP server and verified the generated AI SDK tool includes the server-provided schemas (including `outputSchema` when present).
- Executed a tool and verified the client produces a **structured JSON model output** when the MCP response contains structured results (and returns error text when `isError` is set).

## Checklist

## Future Work

- Add docs/examples showing recommended patterns for consuming structured tool output (property access/filtering) and using `outputSchema` to parse results.

### DIFF
--- a/packages/mcp/src/tool/mcp-client.ts
+++ b/packages/mcp/src/tool/mcp-client.ts
@@ -1,8 +1,8 @@
 import { JSONSchema7 } from '@ai-sdk/provider';
 import {
-  dynamicTool,
   jsonSchema,
   Tool,
+  ToolResultOutput,
   tool,
   ToolExecutionOptions,
 } from '@ai-sdk/provider-utils';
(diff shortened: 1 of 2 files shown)


aayush-kapoor:
marking this as closed since this has been fixed via #11543

PR #11475, opened 2025-12-30
fix(gateway): add error handling for oidc refresh

aayush-kapoor:
## Background

any error encountered while refreshing the oidc token was being silently ignored. this PR surfaces the error to the user

## Summary

catch the error and pass it to the user 

## Manual Verification

add unit test

## Checklist

### DIFF
--- a/packages/gateway/src/gateway-provider.ts
+++ b/packages/gateway/src/gateway-provider.ts
@@ -118,7 +118,7 @@ export function createGatewayProvider(
 
   const getHeaders = async () => {
     const auth = await getGatewayAuthToken(options);
-    if (auth) {
+    if (auth && 'token' in auth) {
       return withUserAgentSuffix(
         {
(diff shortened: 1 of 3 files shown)


packages/gateway/src/gateway-provider.ts
```diff
@@ -277,7 +282,7 @@ export async function getGatewayAuthToken(
       token: oidcToken,
       authMethod: 'oidc',
     };
-  } catch {
-    return null;
+  } catch (error) {
+    return { oidcError: error };
```
gr2m: Why not simply remove the `try/catch` block here and then introduce a `try/catch` around `getGatewayAuthToken()`?

aayush-kapoor: fixed

gr2m (approved):

PR #11505, opened 2026-01-03
feat: support tool-level provider options in OpenAI-compatible provider

KONRADS098:
## Summary

This PR adds support for tool-level provider options in the OpenAI-compatible provider. Tools can now include custom provider metadata through the `providerOptions` field, which gets properly merged into the prepared tools for API requests.

### Changes Made

- **Extended `prepareTools` function** to extract and merge tool-level metadata from `providerOptions.openaiCompatible`
- **Updated type definitions** to allow tools to include arbitrary metadata fields from provider options

### DIFF
--- a/packages/openai-compatible/src/chat/openai-compatible-prepare-tools.ts
+++ b/packages/openai-compatible/src/chat/openai-compatible-prepare-tools.ts
@@ -1,8 +1,13 @@
 import {
   LanguageModelV3CallOptions,
+  SharedV3ProviderMetadata,
   SharedV3Warning,
   UnsupportedFunctionalityError,
 } from '@ai-sdk/provider';
+
+function getToolMetadata(tool: { providerOptions?: SharedV3ProviderMetadata }) {
+  return tool?.providerOptions?.openaiCompatible ?? {};
+}
 
 export function prepareTools({
   tools,
@@ -13,15 +18,17 @@ export function prepareTools({
 }): {
   tools:
     | undefined
-    | Array<{
-        type: 'function';
-        function: {
-          name: string;
-          description: string | undefined;
-          parameters: unknown;
-          strict?: boolean;
-        };
-      }>;
+    | Array<
+        {
+          type: 'function';
+          function: {
+            name: string;
+            description: string | undefined;
+            parameters: unknown;
+            strict?: boolean;
+          };
+        } & Record<string, unknown>
+      >;
   toolChoice:
     | { type: 'function'; function: { name: string } }
     | 'auto'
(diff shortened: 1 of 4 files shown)


packages/openai-compatible/src/chat/openai-compatible-prepare-tools.ts
```diff
@@ -1,9 +1,14 @@
 import {
   LanguageModelV3CallOptions,
+  SharedV3ProviderMetadata,
   SharedV3Warning,
   UnsupportedFunctionalityError,
 } from '@ai-sdk/provider';
 
+function getToolMetadata(tool: { providerOptions?: SharedV3ProviderMetadata }) {
+  return tool?.providerOptions?.openaiCompatible ?? {};
+}
```
lgrammel, line 10: instead of `openaiCompatible`, can you pass in the provider name so it matches how it is handled in `packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts`?

Image

(ideally that parsing would be extracted into a helper function)

KONRADS098: I had this initially, but then I noticed that we already have some inconsistencies  in this regard. I decided to stick to what we are doing in the `packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts`.

https://github.com/vercel/ai/blob/98d480e4b176e1d01cc8692d7fa5e35d3ab90ad0/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts#L12

That being said it would be better to pass the provider name like you suggested.
- Shall I update the `packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts` and `packages/openai-compatible/src/chat/openai-compatible-prepare-tools.ts` to use the provider name?
 - For the `packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts`, shall I use a fallback to the `openaiCompatible`, as consumers might have already integrated with this key?

lgrammel: It should be consistent. I would suggest a separate PR that cleans out the current behavior 1st and makes it consistent. That way we we can focus first on consistency and think about backwards compat, testing, etc. Once that lands we can continue with this PR. Would you be interested in taking this on?

KONRADS098: Sounds good to me, will tag you once it's ready.

KONRADS098: https://github.com/vercel/ai/pull/11859

PR #11509, opened 2026-01-04
fix(react): prevent double resume in React StrictMode

### LINKED ISSUE #9610: onData callback called twice per writer.write() when using resumable streams
### Description
The `onData` callback in the `useChat` hook is being called twice for each `writer.write()` call on the server when using resumable streams with `resume: true`.

### Environment
- **AI SDK version**: `ai@5.0.76`, `@ai-sdk/react@2.0.76`
- **Framework**: React 19 with React Router v7 (SPA mode)
- **Runtime**: Node.js 24.8

### Steps to Reproduce
1. Use the `useChat` hook with `resume: true` enabled
2. Implement an `onData` callback to handle custom data parts
3. On the server, call `writer.write()` to send data parts
4. Observe that the `onData` callback is invoked twice for each `writer.write()` call

### Expected Behavior
Each `writer.write()` call on the server should trigger the `onData` callback exactly once on the client.

### Actual Behavior
Each `writer.write()` call on the server triggers the `onData` callback twice on the client.

### Code Example

**Client:**
```tsx
import { useChat } from "@ai-sdk/react";
import { DefaultChatTransport } from "ai";

const { messages, sendMessage } = useChat({
  id: `chat-${id}`,
  messages: initialMessages,
    api: `${apiURL}/v1/ai/chat`,
    headers: async () => {
      // auth headers
    },
  }),
  onData: (part) => {
    // This callback is being called TWICE per writer.write() call
    console.log("onData called:", part.type, part.data);
  },
  resume: true, // resumable stream
});
```

### Additional Context
- The issue occurs for any custom data part type, not specific to one event type
- Using `DefaultChatTransport` with custom API endpoint
- Not sure what exactly is triggering the duplicate calls, but it appears to be related to the resumable stream functionality
- When the page reloads and the stream resumes, the duplicate calls continue to occur

### Impact
This causes duplicate actions/side effects to be triggered for each server write, which can lead to:
- Multiple unnecessary operations being executed
- Potential race conditions
- Inconsistent UI state
- Performance issues due to redundant processing

Let me know if you need any additional information or a minimal reproduction repository.

jacobwisniewski:
## Background

When using `useChat` with `resume: true`, refreshing the page mid-stream causes duplicate stream requests because React StrictMode runs effects twice (mount → unmount → mount). The current implementation has no protection against this:

```typescript
useEffect(() => {
  if (resume) {
    chatRef.current.resumeStream();
  }
}, [resume, chatRef]);
```

This leads to:
- Duplicated/repeated text in the UI
- Multiple concurrent stream consumers
- Inconsistent message state
- `onData` callbacks being called twice per server write

## Summary

This PR adds a ref guard to prevent `resumeStream()` from being called twice when React StrictMode double-invokes effects during development.

Add a `hasResumedRef` that tracks whether resume has already been attempted:

```typescript
const hasResumedRef = useRef(false);

useEffect(() => {
  if (resume && !hasResumedRef.current) {
    hasResumedRef.current = true;
    chatRef.current.resumeStream();
  }
}, [resume, chatRef]);
```

## Testing

Added a test case that renders the component in `<React.StrictMode>` and verifies only one resume request is made despite effects running twice.

## Related Issues

Fixes #9610


packages/react/src/use-chat.ui.test.tsx
```diff
@@ -1976,6 +1978,78 @@ describe('resume ongoing stream and return assistant message', () => {
...
+    cleanup();
+  });
+
+  it('should only make one resume request even when effects run twice', async () => {
+    // Render in StrictMode to simulate double effect invocation
+    render(
+      <React.StrictMode>
+        <SWRConfig value={{ provider: () => new Map() }}>
+          <TestComponent />
+        </SWRConfig>
```
lgrammel: can this be tested without SWR?

we plan to remove SWR from the pkg; it is only around for legacy reasons

lgrammel:
please add a patch changeset

lgrammel:
for future PRs please try to follow our PR description template

jacobwisniewski:
> please add a patch changeset

Done!

> for future PRs please try to follow our PR description template

Will do!

packages/react/src/use-chat.ui.test.tsx
jacobwisniewski: Yes updated this.

packages/react/src/use-chat.ts
```diff
@@ -110,8 +110,11 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
...
 
+  const hasResumedRef = useRef(false);
+
   useEffect(() => {
-    if (resume) {
+    if (resume && !hasResumedRef.current) {
+      hasResumedRef.current = true;
       chatRef.current.resumeStream();
     }
   }, [resume, chatRef]);
```
lgrammel, line 120: does the ref need to be reset to false? maybe i missing something here, want to make sure this works on the several resumes not just the 1st one

n1ckoates:
Original issue closed

PR #10765, opened 2025-12-01
fix(provider/xai): streamText delta validation error with xAI x_search tool

syeddhasnainn:
## Background

streamText was throwing a validation error when using the x_search tool because some event types returned by xAI were missing from the `xaiResponsesChunkSchema`

## Summary

I updated the chunkSchema with the missing types and also added a test in `packages/xai/src/responses/xai-responses-language-model.test.ts`

## Manual Verification

I tested it by running the xai-x-search-video-understanding example

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

fixes #10628

PR #11563, opened 2026-01-06
fix(provider/xai): fix empty tool input for custom_tool_call in streaming

dancer:
## Background

follow-up to #10765. found that `custom_tool_call` streaming was emitting empty input because the input is only available on the `done` event, not `added`

## Summary

- updated streaming logic to wait for `response.output_item.done` before emitting tool calls for `custom_tool_call` type
- other tool types (web_search_call, etc.) continue to emit on `added` since they have input available immediately

## Manual Verification

ran tests. x_search and view_x_video tool calls now have correct input values like `{"query":"from:xai filter:media","limit":20,"mode":"Latest"}` instead of empty strings

## Checklist

- [x] fix streaming logic for custom_tool_call input
- [x] update snapshot with correct tool input values

### DIFF
--- a/.changeset/seven-llamas-reply.md
+++ b/.changeset/seven-llamas-reply.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fixed streaming tool input for custom_tool_call types (x_search, view_x_video) which were incorrectly returning empty input values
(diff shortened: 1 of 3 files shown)


lgrammel (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/11666

PR #11638, opened 2026-01-08
fix(openai-compatible): use looseObject for robustness with non-standard APIs

EurFelux:
## Background

Some providers that implement OpenAI-compatible APIs don't strictly follow OpenAI's API standards and may include additional fields in their responses. Using strict `z.object` validation can cause these extra fields to fail validation.

For example, Tencent Hunyuan API includes a non-standard `note` field in their responses (see [cherry-studio#12369](https://github.com/CherryHQ/cherry-studio/issues/12369)). While that specific issue was caused by other validation problems, the presence of extra fields demonstrates why flexible schema validation is necessary for OpenAI-compatible APIs.

## Summary

Changed specific Zod schemas in the `@ai-sdk/openai-compatible` package from `z.object` to `z.looseObject`:
- `OpenAICompatibleChatResponseSchema` - handles chat completion responses
- `chunkBaseSchema` - handles streaming chat chunks

This targeted approach allows unknown fields in API responses to be accepted without being processed, while maintaining backward compatibility with standard OpenAI API responses.

## Manual Verification

Tested with:
- Standard OpenAI API responses - behavior unchanged
- Non-standard APIs with extra fields - responses now parse successfully without validation errors

## Checklist

## Related Issues

Related to [cherry-studio#12369](https://github.com/CherryHQ/cherry-studio/issues/12369)

### DIFF
--- a/.changeset/popular-spiders-happen.md
+++ b/.changeset/popular-spiders-happen.md
@@ -2,4 +2,4 @@
 '@ai-sdk/openai-compatible': patch
 ---
 
-Change some response schemas from z.object to z.looseObject to improve compatibility with non-standard OpenAI-compatible APIs. 
+Change some response schemas from z.object to z.looseObject to improve compatibility with non-standard OpenAI-compatible APIs.
--- a/.changeset/popular-spiders-happen.md
+++ b/.changeset/popular-spiders-happen.md
@@ -2,4 +2,4 @@
 '@ai-sdk/openai-compatible': patch
 ---
 
-Change z.object to z.looseObject for robustness
+Change some response schemas from z.object to z.looseObject to improve compatibility with non-standard OpenAI-compatible APIs. 
--- a/.changeset/popular-spiders-happen.md
+++ b/.changeset/popular-spiders-happen.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+Change z.object to z.looseObject for robustness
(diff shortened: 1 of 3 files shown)


aayush-kapoor:
> Tested with:
> - Standard OpenAI API responses - behavior unchanged

change makes sense but can you link the example you ensured this with? 

do you happen to have a minimal repro script

aayush-kapoor (approved):
approving. but if you can provide an example, that'd be great

EurFelux:
> > Tested with:
> > 
> > * Standard OpenAI API responses - behavior unchanged
> 
> change makes sense but can you link the example you ensured this with?
> 
> do you happen to have a minimal repro script

I performed manual verification using these two scripts:

https://github.com/vercel/ai/blob/06e7cda04dc22a2af4fcb8dae2e2d63005f50c30/examples/ai-core/src/generate-text/openai-compatible-chat-provider-options.ts

https://github.com/vercel/ai/blob/06e7cda04dc22a2af4fcb8dae2e2d63005f50c30/examples/ai-core/src/stream-text/openai-compatible-raw-chunks.ts

To be honest, since I don’t have an official OpenAI API key, I used a proxy service for testing (also standard OpenAI API). For testing purposes, I modified the baseURL and model ID in the script and changed `completionModel` to `chatModel`, but I didn’t commit these changes. The script is running normally, and no issues have been detected. This is as expected, since the change from `z.object` to `z.looseObject` should not affect existing behavior.

PR #11658, opened 2026-01-09
fix (provider/xai): no duplicate text delta in responses api

shaper:
## Background

When streaming text responses from the xAI provider, the system was emitting duplicate text content. This occurred because the full accumulated text was being emitted again as a text-delta when a message item was marked as done, even though the same text had already been streamed via individual delta events.

## Summary

Fixed the duplicate text delta issue in the xAI responses API by adding a condition to only emit text content for a message block if it hasn't already been streamed via delta events. This prevents the same content from being sent twice in the response stream.

Added a test case that specifically verifies this behavior by checking that only the expected number of text-delta events are emitted during streaming, without duplicating the full accumulated text.

Also added an example demonstrating the use of the xAI responses API with the streamText function.

## Manual Verification

Tested the fix by running the new example and verifying that text is streamed correctly without duplicates. The test case confirms that when streaming text from the xAI provider, only the incremental deltas are emitted, and the full accumulated text is not duplicated when a message is marked as done.

## Checklist

## Related Issues

Fixes an issue where duplicate text content was being emitted in the response stream from the xAI provider.

### DIFF
--- a/.changeset/slow-trains-sort.md
+++ b/.changeset/slow-trains-sort.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fix (provider/xai): no duplicate text delta in responses api
--- a/examples/ai-core/src/stream-text/xai-responses.ts
+++ b/examples/ai-core/src/stream-text/xai-responses.ts
@@ -1,0 +1,18 @@
+import { xai } from '@ai-sdk/xai';
+import { streamText } from 'ai';
+import { run } from '../lib/run';
+
+run(async () => {
+  const result = streamText({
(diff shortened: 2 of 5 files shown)


shaper:
* **#11658** <a href="https://app.graphite.com/github/pr/vercel/ai/11658?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/ai/11658?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dancer (approved):

PR #11690, opened 2026-01-09
fix(anthropic): add application/pdf type regex matching

aayush-kapoor:
## Background

Reported in #11685, the `supportedUrls` config was missing regex matching for application/pdf type, which in turn led to the SDK downloading the PDF's first.

This wasn't needed for anthropic provider since it directly supported reading pdfs from URL.

But before this fix, if the connection while downloading the pdf was a little flaky, it would lead to a socket timeout error.

## Summary

add the regex url pattern for `application/pdf` type

## Manual Verification

manual verification was a little erratic since running the repro code given by the user only sometimes gave the error. 
but there was still a feature gap

## Checklist

## Related Issues

Fixes #11685

### DIFF
--- a/.changeset/chilled-experts-joke.md
+++ b/.changeset/chilled-experts-joke.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix(anthropic): add application/json type regex matching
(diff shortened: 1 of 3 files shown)


lgrammel:
@aayush-kapoor the pdf support page in the issue gives me a 404. do we know for certain that anthropic does download pdfs automatically?

lgrammel (approved):

aayush-kapoor:
@lgrammel here's the actual docs page:  https://platform.claude.com/docs/en/build-with-claude/pdf-support

> You can provide PDFs to Claude in three ways:
> 1. As a URL reference to a PDF hosted online

I also confirmed with the repro example the user gave

Issue #11700, opened 2026-01-10
useChat UIMessageStream throws TypeError when text-delta/text-end emitted without text-start (unhelpful error, masks original failure)


codewarnab:
## Summary
When using `@ai-sdk/react` `useChat` with a custom transport that returns a `UIMessageStream` (from `ai`), the client can crash with:

`TypeError: Cannot read properties of undefined (reading 'text')`

This occurs when the stream emits a `text-delta` (or `text-end`) chunk for an `id` that never received the corresponding `text-start` chunk.

In real integrations this happens easily in **error paths**, e.g. when an upstream request fails **before any tokens are streamed** and the transport tries to write an inline error message to the UI stream.

The resulting `TypeError` is not actionable and can **mask the original error** (e.g. network failure), because the crash happens while attempting to write an error message.

## Why it happens
`processUIMessageStream` maintains internal state for “active” text parts, keyed by chunk `id`.

If it receives a `text-delta` for an `id` that does not exist in the active map (because no `text-start` for that id was seen), the implementation attempts to do something equivalent to:

```ts
const textPart = state.activeTextParts[chunk.id]; // undefined
textPart.text += chunk.delta;                     // boom
```

So instead of surfacing the *real* upstream failure, the user sees `Cannot read properties of undefined (reading 'text')`.

## Minimal repro (conceptual)
Any `UIMessageStream` that emits this invalid order will reproduce:

- `{ type: "text-delta", id: "x", delta: "..." }` (without a prior `{ type: "text-start", id: "x" }`)

## A more realistic repro scenario (what developers typically do)
Many custom transports do something like:

1. Make a network request to an upstream endpoint.
2. If it fails **before** streaming begins, write an inline error message into the UI stream:

```ts
writer.write({ type: 'text-delta', id: errorId, delta: '⚠️ Request failed...' });
```

If the developer forgets to emit `text-start` (and `text-end`) for the same `id`, the SDK crashes, and the original network error is hidden.

## Expected behavior
One of the following (either is fine, but (1) is likely the safest):

1. **Actionable invariant error**:
   - Example: “Received `text-delta` for missing text part with ID `x`. Ensure a `text-start` chunk is sent before any `text-delta` chunks.”
2. **Graceful fallback**:
   - Optionally treat missing `text-start` as implicit start (less strict; might hide real bugs).

## Actual behavior
Unhelpful runtime crash:

`TypeError: Cannot read properties of undefined (reading 'text')`

[...]

PR #11771, opened 2026-01-13
fix(anthropic): handle mismatch between content_block_start and content_block_delta ID

### LINKED ISSUE #11765: useChat crashes with "Cannot read properties of undefined (reading 'text')" when text generation resumes after tool calls
### Description

The `useChat` hook crashes when text generation resumes after a tool call completes. The SDK's `processUIMessageStream` receives `text-delta` chunks for a text part ID that was never initialized with a `text-start` chunk.

### Steps to Reproduce

1. Use `useChat` with `streamText` that includes tools (e.g., `code_execution`)
2. Send a message that triggers a tool call
3. Wait for tool to complete (`tool-result` received)
4. Text generation resumes
5. Error occurs: `TypeError: Cannot read properties of undefined (reading 'text')`

### Expected Behavior

The stream should continue processing text deltas after tool completion without crashing. If a `text-start` chunk is missing, the SDK should either:
- Initialize the text part automatically, or
- Gracefully handle the missing chunk and continue

### Actual Behavior

The SDK crashes with:
```
at processUIMessageStream (node_modules/ai/dist/index.mjs:9129:32)
```

### Error Details

**Error Message:**
```
Cannot read properties of undefined (reading 'text')
```

**Stack Trace:**
```
at processUIMessageStream (node_modules/ai/dist/index.mjs:9129:32)
```

**When it occurs:**
- After `tool-result` chunks are processed
- When `textPartsCount` transitions from 1 → 2
- SDK receives `text-delta id="X"` but no `text-start id="X"` was received

### Logs/Evidence

From client-side logging:
```javascript
// Before error:
partsCount: 8, textPartsCount: 1

// Error occurs when:
partsCount: 9, textPartsCount: 2  // ← New text part appears
// Immediately followed by error
```

**Sequence:**
1. Tool call executes (`tool-call` chunk)
2. Tool completes (`tool-result` chunk)
3. Text generation resumes
4. SDK receives `text-delta id="9"` (or similar)
5. SDK tries to access `parts[id="9"].text` but part doesn't exist
6. Crash

### Workaround

Currently using `onError` handler to catch and suppress this specific error:

```typescript
useChat({
  transport,
  onError: (error) => {
    if (
      error.message?.includes('Cannot read properties of undefined') &&
      error.message?.includes("reading 'text'")
    ) {
      console.warn('SDK stream ID mismatch (known bug) - stream will continue');
      return; // Don't throw
    }
  },
});
```

This prevents crashes but doesn't fix the root cause.

### Additional Context

- This occurs consistently when using Anthropic's `code_execution` tool with skills
- Also occurs in production with older SDK versions
- The bug exists across multiple SDK versions (tested 6.0.3, 6.0.31, etc.)
- Messages still save to database via `onFinish`, but the error breaks the UI stream

### Related Issues

- Similar to: "text part X not found" errors reported in other issues
- Likely related to how `processUIMessageStream` manages text part IDs during tool→text transitions

### AI SDK Version

- **Package Versions:**
  - `ai`: `6.0.3`
  - `@ai-sdk/react`: `3.0.3`
  - `@ai-sdk/anthropic`: `3.0.0`
  - `@ai-sdk/provider-utils`: `4.0.0` (pinned via overrides)
- **Framework:** Next.js 16.0.7
- **React:** 19.2.1
- **Provider:** Anthropic (Claude Sonnet 4.5)

### Code of Conduct

aayush-kapoor:
## Background

when logging out the raw streaming response chunks for example `‎examples/ai-core/src/stream-text/anthropic-code-execution-pdf.ts‎`, the fixtures show that
```
{"type":"content_block_start","index":2,"content_block":{"type":"text","text":""}}
{"type":"content_block_delta","index":3,"delta":{"type":"text_delta","text":"Now"}}
```

there is a mismatch between the index of the `content_block_start` and `content_block_delta`

this in turn causes the errors seen in #11765 and #11700

## Summary

as a workaround, we track the ID of the `start` block and use that for the following `delta` blocks (which should be ideal behaviour)

## Manual Verification

running the example `http://localhost:3000/chat-anthropic-code-execution` with the prompt "Create a simple PDF document with my name AK on it" would fail before the fix

## Checklist

## Related Issues

Fixes #11765


packages/anthropic/src/__fixtures__/anthropic-code-execution-pdf-skill.1.chunks.txt
```diff
@@ -0,0 +1,134 @@
...
+{"type":"content_block_stop","index":2}
+{"type":"content_block_start","index":2,"content_block":{"type":"text","text":""}}
+{"type":"content_block_delta","index":3,"delta":{"type":"text_delta","text":"Now"}}
```
aayush-kapoor, line 25: look at these lines: the index needs to be '3' for `content_block_start` but instead is 2. 

this causes the mismatch in the AI SDK and thus throws that error.

mdgozza:
This may not be valuable as I don't have a reproducible example to share, but I tried plucking this code change and patching locally, but still encounter the error with provider executed tools, in my case anthropic's web search.

A simple prompt such as "can you research obbba for me, use at least 10 sources"

ends up with 
```
[Error: failed to pipe response] {
  [cause]: [TypeError: Cannot read properties of undefined (reading 'text')]
}
[TypeError: Cannot read properties of undefined (reading 'text')]
 ⨯ unhandledRejection: [TypeError: Cannot read properties of undefined (reading 'text')]
 ⨯ unhandledRejection:  [TypeError: Cannot read properties of undefined (reading 'text')]
```

which ends in `(failed)net::ERR_INCOMPLETE_CHUNKED_ENCODING` in the frontend network tab

I'm not entirely sure yet how to pull this repo down and create some example, but I will try if its valuable. 

fwiw fairly certain it's not any code change issue on my part, I can revert ~5 days ago and can repo. 

Let me know any way I can help, a bit outside my realm though.

Feel free to ignore this if not helpful 💯

mdgozza:
Actually, coming back to this I learned how to clone and run ai-sdk. I ran the 
`examples/ai-core/src/stream-text/anthropic-web-search-tool.ts` off main and it doesn't work, it produces output that looks like this:

```
BaseError: text part 5 not found
d on my search results, here are theError: text part 5 not found
 key techError: text part 5 not found
 newsError: text part 5 not found
 stories from todayError: text part 5 not found
 anError: text part 5 not found
d recentError: text part 5 not found
 days:

## MajorError: text part 5 not found
 TechError: text part 5 not found
 DevelopmentsError: text part 5 not found
 (Error: text part 5 not found
RecentError: text part 5 not found
 Days)

**CError: text part 5 not found
ES 2026 HighlightsError: text part 5 not found
:**
TheError: text part 5 not found
 worlError: text part 5 not found
d's biggestError: text part 5 not found
 tech showError: text part 5 not found
 recentlyError: text part 5 not found
 concludeError: text part 5 not found
d inError: text part 5 not found
```

but after your changes @aayush-kapoor it works 💪 

Just wanted to share that detail.

~~Edit: BUT, yeah it still doesn't work for me in my repo 😭 even though the example does work, there must be some other place 🤔  I'm sure you know, I'm curious to learn so will dig a bit more myself, hope you don't mind my commentary~~

Edit 2: I'm the idiot, when I applied the patch to node_modules I forgot that the compiled output would also be cached in .next and I didn't clear that out, it actually does work for me 😮‍💨 😍

aayush-kapoor:
good to know change works for you - but we might not be pushing this change at all. Anthropic confirmed there was a bug on their end and have patched it. 

once i verify, i'll close this since there won't be a need for this fix

PR #11778, opened 2026-01-13
fix(ai): handle provider-executed tools and tool-approval-response in validation

### LINKED ISSUE #11776: Message Validation Error in Chat Stream: messages.2: all messages must have non-empty content except for the optional final assistant message
### Description

# Bug Report: Message Validation Error After Tool Approval

## Error Message
```
Error: messages.2: all messages must have non-empty content except for the optional final assistant message
```

## When It Occurs

This error occurs **only after approving tool calls**. The flow is:

1. User sends an approval via `addToolApprovalResponse()`
2. `sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithApprovalResponses` triggers auto-send
3. Error is thrown during the streaming response

## Environment

**Backend:**
- `ai`: v6.0.33
- `@ai-sdk/anthropic`: v3.0.0-beta.98

**Frontend:**
- `@ai-sdk/react`: v3.0.35

### Backend - Where error is caught
```typescript
// chat.service.ts line 1221
const result = createAgentUIStreamResponse({
    agent: agentRunner,
    uiMessages: processedMessages,
    generateMessageId,
    onError: (error) => {
        // Error appears here after tool approval
        console.error("Failed to stream chat:", error);
        return error.message;
    },
});
```

### Frontend - Auto-send configuration
```typescript
// use-chat-wrapper.ts line 131
const { addToolApprovalResponse } = useChat({
    transport: transport as any as ChatTransport<any>,
    sendAutomaticallyWhen: (options) =>
        lastAssistantMessageIsCompleteWithApprovalResponses(options) ||
        lastAssistantMessageIsCompleteWithToolCalls(options),
});
```

### Frontend - Sending approval
```typescript
// use-chat-wrapper.ts line 263
const sendToolApprovalResponse = async (approvalId: string, approved: boolean) => {
    await addToolApprovalResponse({
        id: approvalId,
        approved,
    });
    // sendAutomaticallyWhen triggers here and error occurs during streaming
};
```

## Observations

1. Error only occurs **after tool approval**, not during initial tool call
2. Our messages use parts-based format without top-level `content` field
3. Error is caught in `onError` callback of `createAgentUIStreamResponse`
4. Error happens only after calling tools that are marked `needsApproval` and granting them the approval

## Questions

1. Does `createAgentUIStreamResponse` require a `content` field on messages after tool approval responses?
2. Should approval response messages automatically generate a `content` field?

### AI SDK Version

- ai: 6.0.33
- @ai-sdk/react: 3.0.35

Same error occured on these versions

- ai: 6.0.0-beta.169
- @ai-sdk: 3.0.0-beta.172

Tried to update but did not resolve.

### Code of Conduct

kairosci:
## Background

This PR addresses issue #11776 where a "Message Validation Error" occurred after a tool call was approved. The fix has been expanded to handle two scenarios:

1. **Provider-executed tools** (like Anthropic's `code_execution`) that return results in later steps
2. **Tool approval flows** where `tool-approval-response` should satisfy the tool call requirement

## Summary

Modified `convertToLanguageModelPrompt` in `packages/ai/src/prompt/convert-to-language-model-prompt.ts`:

- **Skip validation for provider-executed tools**: Tool calls with `providerExecuted: true` are not required to have an immediate result, as they have deferred results handled by the provider
- **Build approvalId → toolCallId mapping**: Extract the relationship from `tool-approval-request` messages before conversion
- **Handle tool-approval-response**: When a `tool-approval-response` is found, mark the corresponding tool call as satisfied using the mapping
- **Filter empty tool messages**: Ensure only valid, non-empty tool messages are sent to the model

## Manual Verification

- Ran tests for provider-executed tools and tool approval flows
- Verified that `MissingToolResultError` is no longer thrown for valid async/approval scenarios
- All tests in `packages/ai` pass

## Checklist

## Related Issues

Fixes #11776


lgrammel:
I think this PR is patching symptoms and not addressing the root causes, making future debugging and fixing harder.

If there is an approved client tool call that has not been executed (no tool result), the whole request is invalid. It should include the tool result. Thus there is an error with how the AI SDK is used. Filtering out the empty parts does not fix that, it hides the problem.

Can you provide a reproducible example where this issue happens?

kairosci:
I ran several tests and came to the conclusion that your feed was correct, even though the code initially seems to work, it subsequently generates several problems (still getting stuck in the next steps).

I updated the PR to force message checking.

I then modified the logic to track calls and throw an explicit error if results are missing, instead of ignoring them and I added a test to confirm this behavior and I don't see any problems with this approach.

kairosci:
@lgrammel Okay, I think it's ready for the next review, which follows the recommended approach. Thanks for the advice.

kairosci:
@lgrammel I reviewed the validation logic taking your message into account.

Instead of ignoring missing results, there is now explicit mapping and a new error that immediately blocks execution, so you don't end up in inconsistent states.

I also added the cleaning of empty messages and a test to verify the correct functioning of the patch.

So I think it's ready for the next review!

AsafBenjaminov:
@lgrammel This issue happens here https://github.com/vercel/ai/issues/11776 when sending a confirmation for approval from the react ai sdk package and in to the `createAgentUIStreamResponse` function

kairosci:
@AsafBenjaminov This PR is only for block issues and show details with new Error type creation.

Consequently (I think) an issue should be opened with the single error result, where the error is still present and is displayed correctly. If you can, try this change to see if it shows you what's wrong!

kairosci:
I apologize, I accidentally deleted the branch

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -76,7 +109,54 @@ export async function convertToLanguageModelPrompt({
...
+
+  // remove approved tool calls from the set before checking:
+  for (const id of approvedToolCallIds) {
+    toolCallIds.delete(id);
+  }
+
+  if (toolCallIds.size > 0) {
+    const [missingToolCallId] = toolCallIds;
+    throw new MissingToolResultError({ toolCallId: missingToolCallId });
+  }
```
lgrammel, line 155: should the error be `MissingToolResultsError` and contain all missing ids?

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -76,7 +109,54 @@ export async function convertToLanguageModelPrompt({
...
+
+  return combinedMessages.filter(
```
lgrammel, line 157: why this there this new filtering?

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -76,7 +109,54 @@ export async function convertToLanguageModelPrompt({
...
+      case 'user':
+      case 'system':
+        // remove approved tool calls from the set before checking:
+        for (const id of approvedToolCallIds) {
+          toolCallIds.delete(id);
+        }
+
+        if (toolCallIds.size > 0) {
+          const [missingToolCallId] = toolCallIds;
+          throw new MissingToolResultError({ toolCallId: missingToolCallId });
```
lgrammel: same as below

packages/ai/src/prompt/convert-to-language-model-prompt.validation.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      },
+      supportedUrls: {},
+      download: undefined,
+    });
+
+    expect(result).toBeDefined();
+    const assistantMessage = result.find(m => m.role === 'assistant');
+    if (!assistantMessage) throw new Error('Assistant message not found');
+
+    // @ts-ignore
```
lgrammel: why

packages/ai/src/prompt/convert-to-language-model-prompt.validation.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    const toolCall = assistantMessage.content.find(c => c.type === 'tool-call');
+    expect(toolCall).toMatchObject({
+      toolCallId: 'call_1',
+      providerExecuted: true,
+    });
```
lgrammel: why not use snapshots

packages/ai/src/prompt/convert-to-language-model-prompt.validation.test.ts
lgrammel: use snapshots instead

kairosci:
@lgrammel I preferred explicit statements to focus on the logic of the solution. 
Instead of ‘freezing’ all output with a snapshot, I wanted it to be clear at a glance that the test passes thanks to specific properties (such as providerExecuted).

This makes the test more robust, because it doesn't risk breaking due to changes to the structure that have nothing to do with this fix.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
kairosci: The filter is necessary because when we remove the tool-approval-response parts, some tool-type messages may remain empty. 

Most providers reject requests if they contain messages with empty content, generating an error. 

Therefore, this final check serves to clean the history of these “emptied” messages to ensure that the request to the provider is valid.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
kairosci: Explained in comment

packages/ai/src/prompt/convert-to-language-model-prompt.validation.test.ts
kairosci: Explained in comment

packages/ai/src/prompt/convert-to-language-model-prompt.validation.test.ts
kairosci: Explained in comment

kairosci:
> @lgrammel I preferred explicit statements to focus on the logic of the solution. Instead of ‘freezing’ all output with a snapshot, I wanted it to be clear at a glance that the test passes thanks to specific properties (such as providerExecuted).
> 
> This makes the test more robust, because it doesn't risk breaking due to changes to the structure that have nothing to do with this fix.

Then, if you prefer snapshots, I'll change it, but I strongly believe that this way is better.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
lgrammel: why are tool approval response parts removed?

lgrammel:
I prefer snapshots because the explicit approach is harder to develop, review, and maintain in my experience.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
kairosci: We remove them because they are ‘internal’ messages that are only used by the SDK to manage approval. 

The model APIs don't know what to do with these tool-approval-response messages, and if we tried to send them, they would respond with an error because they are not part of their standard schema. 

So, once we have used approval to validate the SDK-side flow, we clean up the message before sending the actual request to the provider.

lgrammel: This is not fully correct. There are cases where the model needs to know about tool approvals, in particular when there are approvals for provider-hosted tools. One such case are approvals for openai executed MCP tools. They have approval flows that integrate into AI SDK. We need to make sure those do not get removed by accident.

kairosci:
I modified the validation to collect all missing tool IDs at once, instead of stopping at the first error, and introduced a final filter to remove empty tool messages, preventing “empty message” errors from providers.

I then added a specific test case to ensure that approvals managed by the provider are not accidentally removed.

I would like to point out that the code did not change the behavior indicated on provider-hosted tools, such as those from OpenAI, but preserves their behavior.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
kairosci: Those relating to the provider-host are preserved, thanks to the providerExecuted check.

I forgot to mention this small detail.

When in doubt, I also added a test to be completely sure.

kairosci:
I also added snapshots

kairosci:
@lgrammel let me know if there is more to correct

packages/ai/src/error/missing-tool-result-error.ts
```diff
@@ -0,0 +1,28 @@
+import { AISDKError } from '@ai-sdk/provider';
+
+const name = 'AI_MissingToolResultsError';
+const marker = `vercel.ai.error.${name}`;
+const symbol = Symbol.for(marker);
+
+export class MissingToolResultsError extends AISDKError {
```
lgrammel, line 7: this error needs to be exposed via index.ts and ideally a docs error page needs to be added (search /content for other error pages)

kairosci:
@lgrammel I also addressed the last change, let me know if it's okay now, thanks!

lgrammel (approved):

AsafBenjaminov:
Hey, How can i tell when this is released and in what version?

PR #11782, opened 2026-01-14
feat(anthropic): add authToken option for Bearer authentication

jarod:
## Background

Some use cases require `Authorization: Bearer` header authentication instead of the `x-api-key` header when using the Anthropic API. This is particularly useful when working with proxy services or custom API gateways that expect Bearer token authentication.

## Summary

Added a new `authToken` option to the Anthropic provider that enables `Authorization: Bearer` header authentication as an alternative to the existing `x-api-key` authentication method.

Changes:
- Added `authToken` option to `AnthropicProviderSettings`
- Added support for `ANTHROPIC_AUTH_TOKEN` environment variable
- When `authToken` is provided, uses `Authorization: Bearer` header instead of `x-api-key`

## Manual Verification

Tested locally by configuring the provider with `authToken` option and verifying the correct `Authorization: Bearer` header is sent in API requests.

## Checklist


aayush-kapoor (commented):
can you please add unit tests for this behaviour as well as an e2e cli example (`example/ai-functions/src/generate-text`)

packages/anthropic/src/anthropic-provider.ts
```diff
@@ -91,19 +100,54 @@ export function createAnthropic(
...
+    rawApiKey && rawApiKey.trim().length > 0 ? rawApiKey.trim() : undefined;
+
+  // Validate that both are not provided simultaneously
+  if (authToken && apiKey) {
+    throw new InvalidArgumentError({
+      argument: 'apiKey/authToken',
+      message:
+        'Both apiKey and authToken were provided. Please use only one authentication method.',
+    });
+  }
```
aayush-kapoor: this check leads to an error if the apiKey and authToken are defined in the .env file (and not the actual client file)

one can have both apiKey and authToken in the .env file and still only use one

this error should only pop up if the user passed both while initialising the provider

aayush-kapoor:
i'll make a fix and push it

aayush-kapoor:
i assume you tested it with a URL that will accept the proper bearer token?

aayush-kapoor (approved):

PR #11796, opened 2026-01-14
devtools don't work with Bun: Avoid using `NODE_ENV` in devtools server code

obadakhalili:
when running the `devtools` bin with bun using `bunx --bun devtools` in user's codebase, the lib doesn't work properly and runs in dev mode for the user if they have a `.env` file with the value `NODE_ENV` set to `development` because bun auto loads the user's vars from the `.env` file. I don't think its wise to use environment variables in libraries' code because users' environments can affect the runtime behaviour, especially so for popular runtime variables such as `NODE_ENV`.

I also noticed that the `devToolsMiddleware` function also uses `NODE_ENV` to throw an error if the user using this in production. i think the same reasoning against env vars in libraries' code applies again, but its less harmful here.

I checked that this fix works but pointing to the local build for the devtools code from my own project where i encountered the code and made sure that `bunx --bun devtools` works (not in dev mode) even if i have a .env file with `NODE_ENV` set to `development`.

## Checklist

### DIFF
--- a/packages/devtools/src/viewer/server.ts
+++ b/packages/devtools/src/viewer/server.ts
@@ -37,8 +37,7 @@ const broadcastToClients = (event: string, data: Record<string, unknown>) => {
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 // Determine if we're running from source (tsx) or built (dist)
-const isDevMode =
-  __dirname.includes('/src/') || process.env.NODE_ENV === 'development';
+const isDevMode = __dirname.includes('/src/');
 const projectRoot = isDevMode
   ? path.resolve(__dirname, '../..')
   : path.resolve(__dirname, '../..');
@@ -237,9 +236,7 @@ app.get('*', async c => {
 });
 
 export const startViewer = (port = 4983) => {
-  const isDev =
-    process.env.NODE_ENV === 'development' ||
-    process.argv[1]?.includes('/src/');
+  const isDev = process.argv[1]?.includes('/src/');
 
   const server = serve(
     {
--- a/.changeset/yellow-chicken-refuse.md
+++ b/.changeset/yellow-chicken-refuse.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/devtools': patch
+---
+
+Avoid using `NODE_ENV` in devtools lib code to make devtools bin work with `bunx --bun devtools command for users that have `NODE_ENV` set to development in a .env file


packages/devtools/src/viewer/server.ts
```diff
@@ -37,8 +37,7 @@ const broadcastToClients = (event: string, data: Record<string, unknown>) => {
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 // Determine if we're running from source (tsx) or built (dist)
-const isDevMode =
-  __dirname.includes('/src/') || process.env.NODE_ENV === 'development';
+const isDevMode = __dirname.includes('/src/');
```
ManyaS-Git, line 40: `__dirname.includes('/src/')` won't match on Windows, where `__dirname` uses backslash separators (e.g. `C:\...\packages\devtools\src\viewer`), so `isDevMode` will always be `false` there and the Vite proxy will never be used when running from source. That cross-platform behavior was exactly what the removed `NODE_ENV === 'development'` fallback provided. Since `main` now uses an explicit `AI_SDK_DEVTOOLS_DEV` flag, consider using that instead of path matching. Also, the `projectRoot` ternary right below has identical branches and can be collapsed to `const projectRoot = path.resolve(__dirname, '../..')`.

packages/devtools/src/viewer/server.ts
```diff
@@ -237,9 +236,7 @@ app.get('*', async c => {
 });
 
 export const startViewer = (port = 4983) => {
-  const isDev =
-    process.env.NODE_ENV === 'development' ||
-    process.argv[1]?.includes('/src/');
+  const isDev = process.argv[1]?.includes('/src/');
```
ManyaS-Git, line 239: Same forward-slash caveat as `isDevMode`: on Windows, `process.argv[1]` contains backslash separators, so `includes('/src/')` is always `false` and the dev log/behavior is skipped. Note that `main`'s current code (post-#12694) still introspects `argv[1]` via `endsWith('/server.ts')`, which has the same portability caveat — an explicit env flag avoids this entire class of problem.

ManyaS-Git (commented):
Thanks for the PR — the underlying problem is real: when `NODE_ENV=development` is picked up from a `.env` file, `bunx --bun devtools` incorrectly activates dev mode. However, I'd suggest not going down the path-introspection route:

1. **This appears to be superseded in `main`.** The dev-mode detection has since been reworked in #12694 (`4c3a932a5`, "fix(devtools): use explicit env flag for dev mode detection"), which replaced the `NODE_ENV` heuristic with an explicit `AI_SDK_DEVTOOLS_DEV` flag. That approach avoids path introspection entirely and is unambiguous — and it also covers the exact symptom this PR targets. This PR is currently CONFLICTING with `main`; I'd recommend rebasing onto `main` and adopting the explicit-flag approach (or closing in favor of that fix).

2. **`/src/` string matching is not cross-platform (see inline comments).** `__dirname.includes('/src/')` and `process.argv[1]?.includes('/src/')` only match on POSIX-style paths. On Windows, paths use backslash separators, so these always return `false` — meaning dev mode (Vite proxy) would silently never activate when running from source. The `NODE_ENV === 'development'` check being removed here was the cross-platform fallback.

3. **Dead code in `projectRoot` (see inline).** Both branches of the `isDevMode` ternary compute the identical value, so the ternary can be collapsed.

4. **Changeset wording.** Minor, but the changeset body is hard to parse and has an unclosed backtick: "make devtools bin work with `bunx --bun devtools command for users that have...". Consider rewording, e.g. "Avoid reading `NODE_ENV` in devtools library code so the devtools bin works with `bunx --bun` even when `NODE_ENV` is set to development in a `.env` file."

Overall the goal is right, but I'd align with the explicit `AI_SDK_DEVTOOLS_DEV` flag already in `main` rather than relying on path introspection.

PR #11841, opened 2026-01-16
feat(openai-compat): support passing multiple file types

### LINKED ISSUE #11779: OpenAI-Compatible provider limits supported file types even if model can handle them
### Description

Currently, the `openai-compatible` provider only supports attaching files with `image/*` mediaType. It explicitly throws an exception when files of a different type are provided (see [convert-to-openai-compatible-chat-messages](https://github.com/vercel/ai/blob/main/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts#L45)).

Many models that are available through openai-compatible endpoints do support a wide variety of filetypes though. For example, Gemini supports:
- Text: text/plain, text/csv, text/xml, text/markdown, text/html, text/javascript
- PDF: application/pdf
- Image: image/png, image/jpeg, image/webp, image/heic, image/heif
- Audio: audio/wav, audio/mp3, audio/aiff, audio/aac, audio/ogg, audio/flac
- Video: video/mp4, video/mpeg, video/mov, video/avi, video/x-flv, video/mpg, video/webm, video/wmv, video/3gpp

I see a few options to handle this:
- **Option 1**: Allow any file type and let the model itself handle error messages (this is similar to what the vertex, bedrock & gateway providers do)
- **Option 2**: Provide an array of supported filetypes in the `createOpenAiCompatible` method

I'm leaning towards **Option 1**. 

Related, some models support URLs directly (instead of passing binary data) and download them internally, without the need of special tool calls (e.g. Gemini). But when passing a URL, the AI SDK downloads the attachment automatically. For this case, it would be helpful to be able to set [supportedUrls](https://ai-sdk.dev/providers/community-providers/custom-providers#native-url-support) as an option on `createOpenAICompatible`, to disable the automatic download.

@aayush-kapoor I saw that you recently worked on the openai-compatible provider, what do you think about this suggestion?

### AI SDK Version

ai: 6.0.30
@ai-sdk/openai-compatible: 2.0.7

### Code of Conduct

aayush-kapoor:
## Background

Currently the openai-compatible package only supported `text` and `image` file types.

Providers derived from this package accept more than just these file types.

## Summary

Add file types: 
- `OpenAICompatibleContentPartFile`
- `OpenAICompatibleContentPartInputAudio`
- relax the check for only allowing `image_url` file types

## Manual Verification

verified by running the repro script `examples/ai-functions/src/generate-text/11779-reprp.ts` (will rename after review)

## Checklist

## Related Issues

Fixes #11779


lgrammel (approved):

lgrammel:
if those new elements are not supported by openai itself, please add comments that show which api they came from (google here presumably)

PR #11843, opened 2026-01-16
build(gitignore): `.claude`

gr2m:


.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 .cache
+.claude
```
gr2m, line 3: first time I used claude code not sure if folks configure it differently so this is not necessary

dancer (approved):

gr2m:
I'll add it to my. global gitignore. If this because a repeating friction than we can add it to our project .gitignore.

PR #11851, opened 2026-01-17
fix(anthropic/provider): Do not use structured output header with json response tool

### LINKED ISSUE #11850: Passing of structured output beta header for json tool causing issues
### Description

We are passing the structured output beta flag even when if opt to use `jsonTool` mode which could result in some issues like strict schema check on our json response tool schema. It is not happening now but this resulted in the breaking behaviour in our product 3 days back and we have to downgrade to v5 to fix this.

The change which I see in from v5 to v6 is that we are passing this beta header for structured output even for json tool mode which should not happen.

Image

### AI SDK Version

- ai: 6.0.26
- @ai-sdk/anthropic: 3.0.9

### Code of Conduct

AVtheking:
## Background

We were sending structured output beta header with json response tool which was performing strict json schema validation causing errors.

## Summary

Do not send beta header for json response tool

## Manual Verification

Ran eg of anthropic tool calls

## Checklist

## Related Issues

Fixes #11850

### DIFF
--- a/.changeset/orange-dolphins-compete.md
+++ b/.changeset/orange-dolphins-compete.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Not sending structured output beta header for json response tool
--- a/packages/anthropic/src/anthropic-messages-language-model.test.ts
+++ b/packages/anthropic/src/anthropic-messages-language-model.test.ts
@@ -871,6 +871,35 @@ describe('AnthropicMessagesLanguageModel', () => {
           }
         `);
       });
+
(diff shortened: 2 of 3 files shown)


lgrammel:
Can you add a unit test case that shows the scenario where this goes wrong and prevents regressions?

AVtheking:
@lgrammel actually the error which we faced due to strict json schema even when using json tool is not happening now, maybe anthropic did make some changes on there side ( as this is beta ). I just want to make sure this doesn't break our app again due to this

AVtheking:
I will add a unit test to make sure not send this header with json tool

AVtheking:
added an unit test

lgrammel (approved):

PR #11854, opened 2026-01-18
Add `contentType` to AnthropicToolOptions to support tool references from custom tool search implementations

obadakhalili:
## Background

server-executed Tool Search tool by anthropic is buggy because Anthropic stream the tool result in the subsequent call to the LLM (more info https://github.com/vercel/ai/issues/11849). a work around is to use [custom tool implementation](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool#custom-tool-search-implementation), but this doesn't work with the AI SDK because `convert-to-anthropic-messages-prompt.ts` in @ai-sdk/anthropic converts json tool outputs to strings to comply with Anthropic's [Create Message API](https://platform.claude.com/docs/en/api/messages/create). this PR attempts to support `tool_reference` content type by adding a hint to the AI SDK in the form of a provider-specific tool output option.

## Manual Verification

I linked a test project to a locally built package with the proposed changes.

## Checklist

## Related Issues

https://github.com/vercel/ai/issues/11849


aayush-kapoor:
closing this since i don't think this is the right approach for this problem, especially since we have a mechanic for this in place for other anthropic tools. 

see #11931

obadakhalili:
@aayush-kapoor i will check your solution for the defered tool search tool result ([though it doesn't sound right to expect a deferred result](https://github.com/vercel/ai/issues/11849#issuecomment-3786158286)), but this isn't to address this problem. its to enable a workaround. the workaround is that since executing tool search on the server has a problem, then we can go around that by executing it on the client like a normal tool, but this currently isn't possible in the ai sdk because the of the reason I explain in the pr description.

PR #11858, opened 2026-01-18
fix(provider-utils): add comprehensive type tests showing broken behavior

kairosci:
## Background

The `ToolApprovalRequest` type is currently missing `toolName` and `input` properties. This PR adds comprehensive type tests to `tool.test-d.ts` and explicitly demonstrates the broken behavior by asserting existence of these properties, which causes the tests to fail.

## Summary

- Updated `packages/provider-utils/src/types/tool.test-d.ts` to be comprehensive for the current state, including tests for:
    - Callbacks
    - Strict mode
    - Provider/Dynamic tools
    - Type helpers
    - `ToolApprovalRequest` (failing test to demonstrate missing properties)

## Manual Verification

Ran `pnpm vitest run packages/provider-utils/src/types/tool.test-d.ts --typecheck` and verified that it fails as expected due to missing properties.

## Checklist


kairosci:
I apologize, I accidentally deleted the branch

kairosci:
@lgrammel I have added the requested tests, demonstrating that the type definitions are currently incomplete as the TypeScript check now fails

kairosci:
Since it was asked to prove the problem, I created code to verify the initial hypothesis, which at the moment has no corrections. 

So as it should be, TypeScript does not pass.

 (Prettier is to be formatted, but I think it's the least of the problems)

lgrammel:
@kairosci thanks for creating the tests. I meant that in this PR, the tests should describe the current behavior (even if wrong), i.e. they should just specify what exists. In the follow up PR that fixes the issue, we would change the tests to show the breakage, and then fix it. This approach will make it very clear in the 2nd PR what behavior was changed and how, making it easier to think about (e.g. re backwards compatibility).

Can you adjust this PR such that the tests encode the existing behavior, even if buggy?

kairosci:
@lgrammel I have reverted the functional changes to the xai-responses-prepare-tools.ts file, which is now the same as main.

I have updated the tests in xai-responses-prepare-tools.test.ts so that they expect the result of the current implementation, ignoring the extra arguments for the mcp tools, which was the “buggy” behavior.

PTAL!

PR #11898, opened 2026-01-20
fix(xai): emit text-end immediately when message completes

boylec:
## Summary

Aligns xAI provider streaming behavior with OpenAI's pattern for `text-end` event timing.

## Problem

Currently, xAI emits `text-end` events only in the `flush()` handler at the very end of the stream. This differs from OpenAI which emits `text-end` when the message item is done (`response.output_item.done`).

## Solution

Emit `text-end` immediately when message content completes, matching OpenAI's behavior. This gives consumers accurate timing for when text content is finished.

## Why This Matters

Applications that process streaming events need to know when text content is complete (e.g., to trigger UI updates, start processing, etc.). Deferring `text-end` to flush means consumers can't distinguish between "text is done" and "entire response is done".

## Changes

- Track `ended` state in `contentBlocks` to avoid duplicate emissions
- Emit `text-end` in `response.output_item.done` handler when message content completes
- Update `flush()` to only emit `text-end` for blocks that weren't already ended

## Checklist

- [x] A changeset has been added

### DIFF
--- a/packages/xai/src/responses/xai-responses-language-model.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.ts
@@ -376,7 +376,7 @@ export class XaiResponsesLanguageModel implements LanguageModelV3 {
     };
     let usage: LanguageModelV3Usage | undefined = undefined;
     let isFirstChunk = true;
-    const contentBlocks: Record<string, { type: 'text' }> = {};
+    const contentBlocks: Record<string, { type: 'text'; ended: boolean }> = {};
     const seenToolCalls = new Set<string>();
 
     const activeReasoning: Record<
@@ -461,7 +461,7 @@ export class XaiResponsesLanguageModel implements LanguageModelV3 {
               const blockId = `text-${event.item_id}`;
 
               if (contentBlocks[blockId] == null) {
-                contentBlocks[blockId] = { type: 'text' };
+                contentBlocks[blockId] = { type: 'text', ended: false };
                 controller.enqueue({
                   type: 'text-start',
                   id: blockId,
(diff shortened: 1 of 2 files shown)


packages/xai/src/responses/xai-responses-language-model.ts
```diff
@@ -679,6 +680,18 @@ export class XaiResponsesLanguageModel implements LanguageModelV3 {
                     }
                   }
                 }
+
+                // Emit text-end immediately when message completes (like OpenAI pattern)
```
ManyaS-Git, line 684: Since this block sits inside `if (event.type === 'response.output_item.added' || event.type === 'response.output_item.done')`, it will also run on the `added` event. If the message item carries text content at `added` time (which the OpenAI-compatible responses API does for simple items), this emits `text-end` before the `response.output_text.delta` events for the same `item_id` have streamed, so subsequent deltas land after `text-end`. Suggest gating this on `event.type === 'response.output_item.done'` (mirroring how the `function_call` branch above handles `added` vs `done`). The `flush` fallback still covers blocks that never receive a `done` event.

ManyaS-Git (commented):
Thanks for the PR — emitting `text-end` when the message output item completes (instead of waiting for `flush`) aligns the xAI stream with OpenAI's pattern and is a sensible change.

A couple of things to address:

1. **Premature `text-end` on `response.output_item.added`.** The `part.type === 'message'` branch runs for *both* `response.output_item.added` and `response.output_item.done` events, but the new `text-end` emission isn't gated on the event type. If an `added` event ever carries text content, the block gets created and immediately ended there, so any subsequent `response.output_text.delta` events would be enqueued *after* `text-end`, breaking the text-start → delta* → text-end ordering contract. Please guard the emission with `event.type === 'response.output_item.done'` (see inline comment).

2. **No tests added.** This changes streaming semantics, and the repo has thorough stream coverage in `xai-responses-language-model.test.ts`. It would be good to add a regression test asserting (a) `text-end` is emitted when the message item completes and before `response.completed`, and (b) `text-end` is not emitted twice when both `added` and `done` are received for the same item.

3. **Conflicting with `main`.** This needs a rebase. `main` has since restructured the same handler — the fallback block now only emits text when it hasn't already been streamed via `output_text.delta` events — so the `text-end` logic should be re-applied on top of the current structure.

The `ended`-flag bookkeeping is a reasonable way to avoid double emission; overall the direction looks good once the event-type guard and tests are in place.

PR #11907, opened 2026-01-20
fix(ai): resolve race condition in parallel tool execution

delta575:
## Background

When multiple tools execute in parallel during `streamText`, the stream could close prematurely or throw "The stream is not in a state that permits enqueue" errors.

**Context:** We're using `@convex-dev/agent` 0.3.2 which doesn't support AI SDK v6. We wanted to use Gemini 3 Flash, but `thought_signature` is required (not optional), so we had to downgrade to Gemini 2.5 where it's optional and AI SDK v5 worked fine. When testing AI SDK v6 compatibility via [get-convex/agent#208](https://github.com/get-convex/agent/pull/208) in preparation for Gemini 3 support, we encountered this race condition - Gemini 3 Flash aggressively uses parallel tool calls which exposed the bug.

## Summary

Two issues caused this race condition:

1. **Non-unique tool tracking IDs**: `generateId()` was returning the same value for multiple tools in a batch, causing the `outstandingToolResults` Set to only track one tool. When that tool completed, the stream closed while others were still running.

2. **Re-entry in attemptClose**: Multiple `finally()` blocks could call `attemptClose()` simultaneously, causing race conditions when closing the stream.

**Fix:**
- Use `toolCall.toolCallId` instead of `generateId()` for unique tool tracking
- Add `closed` flag to prevent re-entry in `attemptClose()`
- Guard async enqueue calls to prevent enqueueing after stream closure

## Manual Verification

Tested with a Convex application using `@convex-dev/agent` from [get-convex/agent#208](https://github.com/get-convex/agent/pull/208) and Gemini 3 Flash with 5+ parallel tool calls - all tool results are now captured and streaming completes successfully.

## Checklist
- [x] A patch changeset for relevant packages has been added

### DIFF
--- a/packages/ai/src/generate-text/run-tools-transformation.ts
+++ b/packages/ai/src/generate-text/run-tools-transformation.ts
@@ -150,13 +150,23 @@ export function runToolsTransformation<TOOLS extends ToolSet>({
   const toolCallsByToolCallId = new Map<string, TypedToolCall<TOOLS>>();
 
   let canClose = false;
+  let closed = false; // Prevent race condition when multiple tools complete simultaneously
   let finishChunk:
     | (SingleRequestTextStreamPart<TOOLS> & { type: 'finish' })
     | undefined = undefined;
 
   function attemptClose() {
+    // Prevent re-entry: if already closed, nothing to do
+    if (closed) {
+      return;
+    }
+
     // close the tool results controller if no more outstanding tool calls
     if (canClose && outstandingToolResults.size === 0) {
+      // Mark as closed BEFORE doing any work to prevent race conditions
+      // where multiple finally() blocks call attemptClose() simultaneously
+      closed = true;
+
       // we delay sending the finish chunk until all tool results (incl. delayed ones)
       // are received to ensure that the frontend receives tool results before a message
       // finish event arrives.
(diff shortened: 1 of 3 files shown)


packages/ai/src/generate-text/run-tools-transformation.ts
```diff
@@ -309,7 +319,9 @@ export function runToolsTransformation<TOOLS extends ToolSet>({
 
             // Only execute tools that are not provider-executed:
             if (tool.execute != null && toolCall.providerExecuted !== true) {
-              const toolExecutionId = generateId(); // use our own id to guarantee uniqueness
+              // Use toolCallId which is unique per tool call from the LLM
+              // (generateId() was returning the same value for multiple tools in a batch)
+              const toolExecutionId = toolCall.toolCallId;
```
lgrammel, line 324: has this fix been ai generated?

lgrammel: `generateId` was used intentionally. if you end up with repeated ids that is most likely an issue with the id generator that you pass in

lgrammel:
This PR mixes 2 issues. The close state issue seems reasonable, but I have doubts regarding the id generation. Please separate the close fix into a new PR.

PR #11965, opened 2026-01-22
fix(ai): add parallel tool close guards

delta575:
## Background

When multiple tools execute in parallel and complete nearly simultaneously, their .finally() blocks all call attemptClose(). Without a guard, the first completion closes the stream and subsequent completions throw "Controller is already closed" or "stream is not in a state that permits enqueue" errors.

## Summary

Add a closed flag to prevent race conditions in attemptClose():

- Guard attemptClose() against re-entry when multiple tools complete simultaneously
- Guard async enqueue() calls to prevent enqueueing after stream closure
- Set flag before doing work to prevent TOCTOU race

## Manual Verification

Tested with @convex-dev/agent and Gemini 2.5 Flash using 5+ parallel tool calls. Without this fix, intermittent "stream is not in a state that permits enqueue" errors occurred. With this fix, streaming completes successfully.

## Checklist

## Related Issues

Split from #11907 per review feedback to separate the close guard fix from the ID generation change.


lgrammel:
The test does not fail without the fix, thus it does not reproduce the reported issue. I have reverted the fix to the production code and the test still passes.

It also uses real-time clock changes (delay) which must be avoided in unit tests, and could be simplified with DelayedPromise.

Can you simplify the test, avoid any delay/wait operations, and make sure it fails without the fix?

delta575:
> The test does not fail without the fix, thus it does not reproduce the reported issue. I have reverted the fix to the production code and the test still passes.
> 
> It also uses real-time clock changes (delay) which must be avoided in unit tests, and could be simplified with DelayedPromise.
> 
> Can you simplify the test, avoid any delay/wait operations, and make sure it fails without the fix?

Simplified the tests with DelayedPromise and removed all delay() calls.

On the determinism issue: true race conditions are hard to reproduce in unit tests because JS is single-threaded - microtasks run sequentially even when promises resolve "simultaneously". The .finally() callbacks always execute one at a time, so the guard never gets exercised under normal conditions.

The new test triggers premature closure by using a constant generateId (which is what @convex-dev/agent does for message grouping - intentional on their side for thread message IDs, not a bug):
1. All tools add same ID to Set → only 1 entry
2. First tool completes → deletes ID → Set empty → stream closes
3. Other tools try to enqueue → throws without guard

Removing the if (!closed) guards produces:
TypeError: Invalid state: Controller is already closed

Vitest reports this as unhandled rejection and exits with code 1.

The generateId discussion is addressed in #11966 - it's being used for two purposes (message IDs vs tool execution tracking), and frameworks legitimately override it for message grouping.

I've been using Claude to help debug and write these fixes.

packages/ai/src/generate-text/run-tools-transformation.test.ts
```diff
@@ -1140,4 +1140,177 @@ describe('runToolsTransformation', () => {
...
+            type: 'finish',
+            finishReason: { unified: 'tool-calls', raw: 'tool_calls' },
+            usage: testUsage,
+          },
+        ]);
+
+      const transformedStream = runToolsTransformation({
+        // Simulate framework behavior: generateId returns same ID for message grouping
+        // This causes outstandingToolResults Set to only track one entry
+        generateId: () => 'constant-id',
```
lgrammel, line 1190: this is not the framework behavior. by default different ids are generated. if generateId returns the same id several times this is a misconfiguration.

lgrammel: once this is changed to `   generateId: mockId()` the test passes without the fix.

is this a misconfiguration and not an ai sdk bug?

lgrammel: the expectation is that generateId() generates unique ids. this might need to be clarified in the documentation / jsdoc

PR #11967, opened 2026-01-22
Improve Turborepo configuration

anthonyshew:
## Summary

- Use explicit `turbo run` in root package.json scripts (recommended over shorthand in committed code)
- Simplify task dependencies by removing redundant `^build` (since `build` already depends on `^build`)
- Replace `lint` task's `^lint` dependency with transit node pattern for parallel execution with correct cache invalidation

### `turbo run` vs shorthand

Turborepo recommends using `turbo run <task>` in package.json scripts and CI pipelines. The shorthand `turbo <task>` is intended for interactive terminal use.

### Redundant dependencies

Tasks like `test` and `type-check` had `dependsOn: ["^build", "build"]`. Since the `build` task already has `dependsOn: ["^build"]`, specifying both is redundant—`["build"]` alone ensures dependencies are built first.

### Transit node for lint

The `lint` task had `dependsOn: ["^lint"]` which forces sequential execution across packages. Linting typically reads source files directly and doesn't need dependencies to be linted first.

The transit node pattern:
```json
"transit": { "dependsOn": ["^transit"] },
"lint": { "dependsOn": ["transit"] }
```

This allows lint tasks to run in parallel while maintaining correct cache invalidation when dependency source code changes. The `transit` task creates dependency relationships without matching any actual script.

## AI usage
Chat transcript: https://opncd.ai/share/RRBZZVeI


turbo.json
```diff
@@ -67,17 +67,20 @@
...
-      "dependsOn": ["^build", "build"]
+      "dependsOn": ["^build", "transit"]
     },
     "test": {
-      "dependsOn": ["^build", "build"]
+      "dependsOn": ["^build", "transit"]
     },
     "test:update": {
-      "dependsOn": ["^build", "build"]
+      "dependsOn": ["^build"]
```
anthonyshew:
```suggestion
      "dependsOn": ["^build", "build"]
```

gr2m (approved):
Thank you @anthonyshew! Much appreciated!

anthonyshew:
👋 Just cleaning up my Pull Requests page.

PR #11974, opened 2026-01-23
chore: add pre-commit hook for pnpm i

lgrammel:
## Background

From time to time, package.json files change but the lockfile is not updated. This leads to immediate CI failures.

## Summary

Ensure that the lockfile is always updated when `package.json` files change (via a pre-commit hook).

## Related Issues

Follow up from #11972


AGENTS.md
```diff
@@ -119,6 +119,7 @@ pnpm tsx src/stream-text/openai.ts    # Run a specific example
 - **Config**: Defined in root `package.json`
 - **Settings**: Single quotes, trailing commas, 2-space indentation, no tabs
 - **Pre-commit hook**: Automatically formats staged files on commit via `lint-staged`
+- **Auto-install**: If `package.json` changes are staged, `pnpm install` runs automatically
```
lgrammel: @copilot merge this item into the previous "pre-commit hooks" item

PR #11984, opened 2026-01-23
fix(openai,groq): emit tool-input-end and tool-call events for unfinished tool calls in flush handler

Nepomuk5665:
## Summary

This PR fixes a bug where the OpenAI and Groq providers fail to emit `tool-input-end` and `tool-call` events for tool calls that haven't completed their argument streaming when the stream ends.

## Problem

When streaming tool call arguments and the stream ends (e.g., due to network interruption or API behavior), the `flush` handler in `OpenAIChatLanguageModel` and `GroqChatLanguageModel` does not emit the necessary final events for tool calls that are still in progress. This is inconsistent with the behavior in `DeepSeekChatLanguageModel` and `OpenAICompatibleChatLanguageModel`, which correctly handle this case.

The `toolCalls` array tracks all tool calls with a `hasFinished` flag, but neither OpenAI nor Groq processes unfinished tool calls in their flush handlers.

## Solution

Add the missing logic to both providers' `flush` handlers to iterate through unfinished tool calls and emit:
1. `tool-input-end` event with the tool call id
2. `tool-call` event with the accumulated arguments

This matches the implementation already present in:
- `packages/deepseek/src/chat/deepseek-chat-language-model.ts:497-512`
- `packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts:659-683`

## Testing

- Ran `pnpm test --filter @ai-sdk/groq` - all 80 tests pass
- Ran `pnpm vitest run src/chat/openai-chat-language-model.test.ts` - all 77 tests pass

### DIFF
--- a/packages/groq/src/groq-chat-language-model.ts
+++ b/packages/groq/src/groq-chat-language-model.ts
@@ -538,6 +538,23 @@ export class GroqChatLanguageModel implements LanguageModelV3 {
               controller.enqueue({ type: 'text-end', id: 'txt-0' });
             }
 
+            // go through all tool calls and send the ones that are not finished
+            for (const toolCall of toolCalls.filter(
+              toolCall => !toolCall.hasFinished,
+            )) {
+              controller.enqueue({
+                type: 'tool-input-end',
+                id: toolCall.id,
+              });
+
+              controller.enqueue({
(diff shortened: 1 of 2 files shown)


gr2m:
Already implemented in #14565.

PR #11262, opened 2025-12-17
feat(anthropic): add programmatic tool calling

aayush-kapoor:
## Background

Anthropic released programmatic tool calling. It allows calling a client-side function tool from an ongoing provider-executed code execution tool.

The main assumption this breaks is that all tools are finished at the end of a turn. Special provider-executed tools may span turns.

See https://platform.claude.com/docs/en/agents-and-tools/tool-use/programmatic-tool-calling

## Summary

- Updated spec to account to add anthropic specific providerOptions for tools
- generate-text and stream-text core functions were modified to allow passing container IDs from previous steps (container ID is generated in the model response and needs to be passed back for this functionality)
- flag added specifically for the code_execution tool so that deferred tool results can be handled properly

## Manual Verification

- [x] unit tests
- [x] static examples
- [x] UI examples

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)

## Related Issues

Fixes #10594

PR #11997, opened 2026-01-23
fix(ai): Mark deferred tool call resolved on tool-error to avoid endless streamSteps.

LaPulgaaa:
## Background

When using provider-executed tools (like Anthropic's `code_execution`) that support deferred results, if the tool fails and the provider streams back a tool result error packet, the SDK would enter an infinite loop calling `streamStep`.

This happens because the `pendingDeferredToolCalls` map only cleared entries when receiving `tool-result` parts, not `tool-error` (transformed from `tool-result` part in `run-tools-transformation.ts`) parts. 

Since the pending map was never cleared on `tool-error` output type, the loop condition `pendingDeferredToolCalls.size > 0` remained true indefinitely, causing endless stream invocations even after the provider signaled `end_turn`.

Introduced in 
- https://github.com/vercel/ai/pull/11262

## Summary

Updated the `pendingDeferredToolCalls` logic in `stream-text.ts` to handle `tool-error` (transformed `tool-result` chunk) in addition to `tool-result`:

1. When checking if a tool call already has a result (to avoid adding it to pending), now checks for both `tool-result` and `tool-error` output type.
2. When clearing resolved tool calls from the pending map, now deletes on both `tool-result` and `tool-error` output type.

## Manual Verification

Reproduced the issue by simulating a provider-executed tool failure that returns `tool-error`. Before the fix, the stream would hang indefinitely. After the fix, the stream properly terminates when receiving the error.

## Checklist

## Related Issues

Fixes #11996

### DIFF
--- a/packages/ai/src/generate-text/stream-text.ts
+++ b/packages/ai/src/generate-text/stream-text.ts
@@ -1687,13 +1687,14 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT extends Output>
                       tool?.type === 'provider' &&
                       tool.supportsDeferredResults
                     ) {
-                      // Check if this tool call already has a result in the current step
-                      const hasResultInStep = stepToolOutputs.some(
+                      // Check if this tool call already has a result or error in the current step
+                      const hasResultOrErrorInStep = stepToolOutputs.some(
                         output =>
-                          output.type === 'tool-result' &&
+                          (output.type === 'tool-result' ||
+                            output.type === 'tool-error') &&
                           output.toolCallId === toolCall.toolCallId,
                       );
-                      if (!hasResultInStep) {
+                      if (!hasResultOrErrorInStep) {
                         pendingDeferredToolCalls.set(toolCall.toolCallId, {
                           toolName: toolCall.toolName,
                         });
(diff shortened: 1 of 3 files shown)


packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1687,23 +1687,27 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT extends Output>
...
                       tool.supportsDeferredResults
                     ) {
-                      // Check if this tool call already has a result in the current step
-                      const hasResultInStep = stepToolOutputs.some(
+                      // Check if this tool call already has a result or error in the current step
+                      const hasResultOrErrorInStep = stepToolOutputs.some(
                         output =>
-                          output.type === 'tool-result' &&
+                          (output.type === 'tool-result' ||
+                            output.type === 'tool-error') &&
```
LaPulgaaa, line 1694: cc @aayush-kapoor @lgrammel

`tool-result` chunks of server executed tool with `isError: true` are transformed to `tool-error` in 
https://github.com/vercel/ai/blob/52fc8fb2b9e797550e8e1ce7bde3afdf9e47f35c/packages/ai/src/generate-text/run-tools-transformation.ts#L354-L356

The `if` check downstream only considered `tool-result` leading to errored server executed tool calls never be deleted from `pendingDeferredToolCalls` map.

This leads to endless execution of 
```typescript
  await streamStep({
  currentStep: currentStep + 1,
  responseMessages,
  usage: combinedUsage,
});
```

aayush-kapoor:
closing this since this is a duplicate of #12011

PR #11998, opened 2026-01-23
fix(anthropic): allow custom/dynamic key for providerOptions

aayush-kapoor:
## Background

#11951

users couldn't pass options using the custom name for their provider

## Summary

- parse providerOptions from both canonical and custom keys
- conditionally include custom key in providerMetadata output

example 1: 

```ts
providerOptions: {
  openai: {
    user: 'test-user-canonical',
  },
},

print(result.providerMetadata);
// will only have the 'openai' key
```

example 2: 

```ts
providerOptions: {
  'my-custom-openai': {
    user: 'test-user-canonical',
  },
},

print(result.providerMetadata);
// will have the 'openai' key AND 'my-custom-openai' key (with the exact same values
```

PS: unchanged behaviour for default providers

## Manual Verification

- verified for multi-turn conversations by changing the agent for `http://localhost:3000/chat-openai-smooth-stream` to be implemented via custom provider with a custom providerOptions key
- added cli examples to test

## Checklist

## Future Work

- might need to be done for openai/openai-compat

## Related Issues

fixes #11951


.changeset/tame-ravens-do.md
```diff
@@ -0,0 +1,6 @@
+---
+'@ai-sdk/anthropic': patch
+'@ai-sdk/openai': patch
```
jerome-benoit: I've not checked the other providers but if they have cut and paste the same code, it's likely the case for them.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -863,16 +896,32 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV3 {
...
+    // Determine which key(s) to use in providerMetadata:
+    // - Azure providers always use 'azure' key
+    // - Non-Azure providers use 'openai' key, plus custom key if it was used
+    let providerMetadata: SharedV3ProviderMetadata;
+    if (isAzureProvider) {
+      providerMetadata = { azure: openaiMetadata };
+    } else if (usedCustomProviderKey) {
+      providerMetadata = {
+        openai: openaiMetadata,
+        [providerOptionsName]: openaiMetadata,
```
jerome-benoit: I do not understand the semantic: one provider instance is not supposed to expose one name?  Is there actually a use case where one instance can be shared between different connection setups?

.changeset/tame-ravens-do.md
aayush-kapoor: i'd rather we do a case by case solution for this instead of changing the logic everywhere all at once. 

highest priority are these providers

packages/openai/src/responses/openai-responses-language-model.ts
aayush-kapoor: well we can't just change it to be mapped to only the custom name directly since people who (may) already use the `openai` key even with a custom provider will get a breaking change.

we have to make sure the solution is backwards compatible

lgrammel:
should this modify the openai or the openai-compatible provider?

lgrammel (approved):

PR #12016, opened 2026-01-26
fix(ai): parse structured output when finishReason is tool-calls 

bilalafzal-ti:
## Summary                                                                                                           
  - Fixes issue where `Output.object()` with custom tools throws `NoOutputGeneratedError` when Gemini returns          
  `finishReason: 'tool-calls'` with valid JSON output                                                                  
                                                                                                                       
  ## Background                                                                                                        
  When using custom tools with structured output (`Output.object()`) with Gemini models (via `@ai-sdk/google-vertex`), 
  the SDK throws `AI_NoOutputGeneratedError` because structured output parsing is skipped.                             
                                                                                                                       
  **Root cause**: In `generate-text.ts`, output is only parsed when `finishReason === 'stop'`. However, with Gemini +  
  custom tools, the final step may return `finishReason: 'tool-calls'` even when valid JSON output is present in the   
  response.                                                                                                            
                                                                                                                       
  This happens because:                                                                                                
  1. Gemini's `STOP` finish reason gets mapped to `'tool-calls'` when `hasToolCalls` is true                           
  2. The output parsing gate at line 906 only parses when `finishReason === 'stop'`                                    
  3. Result: valid structured output is ignored, `NoOutputGeneratedError` is thrown                                    
                                                                                                                       
  ## Changes                                                                                                           
  - Modified output parsing logic to also parse when an explicit output schema is provided AND text content exists     
  - Added unit tests for the new behavior                                                                              
  - Added reproduction example for Google Vertex + custom tools + structured output                                    
                                                                                                                       
  ## Test plan                                                                                                         
  - [x] Added unit tests for parsing with `finishReason: 'tool-calls'`                                                 
  - [x] Existing test for "should not parse output when finish reason is tool-calls" still passes (no text content     
  case)                                                                                                                
  - [x] Type check passes

### DIFF
--- a/.changeset/fix-tools-output-parsing.md
+++ b/.changeset/fix-tools-output-parsing.md
@@ -1,0 +1,10 @@
+---
+'ai': patch
+---
+
+fix: Parse structured output even when finishReason is 'tool-calls' if text content exists
+
+When using custom tools with structured output (Output.object()), the final step may have
+finishReason='tool-calls' even though valid JSON output is present. This fix attempts to
+parse the output when an explicit output schema is provided and there's text content,
+regardless of the finish reason.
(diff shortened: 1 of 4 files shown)


lgrammel:
this in an error in the google provider and should be fixed there by returning stop instead of tool calls. the ai package must not be changed

gr2m:

PR #12037, opened 2026-01-26
ONHOLD feat(ai): add per-tool timeout option

dancer:
## Background

users requested a simpler way to add timeouts to individual tools without manually wrapping with `AbortSignal.timeout()` in the execute function

## Summary

- add `timeout` property to tool type for per-tool timeout configuration
- tool execution automatically aborts and returns `tool-error` if timeout exceeded
- abort signal is merged with any existing abort signal from generateText/streamText

## Verification

- tests pass
- example verified with openai: fast tool succeeds, slow tool times out correctly

## Checklist

- [x] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)

## Related Issues

Fixes #11975

### DIFF
--- a/.changeset/slow-doors-judge.md
+++ b/.changeset/slow-doors-judge.md
@@ -1,0 +1,6 @@
+---
+'@ai-sdk/provider-utils': patch
+'ai': patch
+---
+
+add per-tool timeout option for automatic tool execution abort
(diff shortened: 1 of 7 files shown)


lgrammel:
Unclear to me if this should be added to `tool` or to `streamText/generateText` `timeout`.

If we think about who configures the timeouts and who writes the tool, I am leaning towards the latter for the following reason:

Assume 2 roles: tool user and tool author. they may or may not be the same person/group (e.g. for tools distributed as npm packages)

With the current approach, a tool user would need a wrapper tool to set the timeout on 3rd party tool. Also timeouts would get distributed into tools, vs being defined for a specific call.

Would be good to talk offline about the design but that was the reason why I postponed looking into this.

dancer:
PR is on hold till we put some other stuff in place!

PR #12055, opened 2026-01-27
Fixes to reasoning-0 not found error

### LINKED ISSUE #12054: reasoning part reasoning-0 not found due to bug in extractReasoningMiddleware
### Description

There have been issues raised in some coding agent platforms (eg. OpenCode, cherry-studio) over the past few months regarding a "reasoning part reasoning-0 not found" error. Below are 2 such examples raised in their respective repos

-  [Error "reasoning part reasoning-0 not found" from tool calls using GLM 4.5 Air](https://github.com/anomalyco/opencode/issues/6486)
- [[Bug]: Error Details: "reasoning part 0 not found" when starting conversation](https://github.com/CherryHQ/cherry-studio/issues/10875)

These issues are particularly prevalent for Zhipu's GLM series of models, especially those using the locally hosted (eg. LMStudio, vLLM) models instead of Zhipu's endpoints. The "reasoning part reasoning-0 not found" error is triggered when the LLM endpoint replies with an empty thinking block <think></think>.

I created a mock LLM server to stream a reply with an empty thinking block as a test case, and it throws the error as shown in the screenshot below.
Image

The root cause is due to the lack of checks in the [extract-reasoning-middleware.ts](https://github.com/vercel/ai/blob/c055bb9c0fe71cbbe8696b441de800daf6c1f61a/packages/ai/src/middleware/extract-reasoning-middleware.ts#L210C1-L222C1) file, where a 'reasoning-end' event will be emitted without checking if a 'reasoning-start' event has been emitted. In the scenario of a empty thinking block, the 'reasoning-start' event will never be sent as `text.length>0` check will fail [here](https://github.com/vercel/ai/blob/c055bb9c0fe71cbbe8696b441de800daf6c1f61a/packages/ai/src/middleware/extract-reasoning-middleware.ts#L140C17-L158C20).

This will result in `activeReasoning` being `null` and the error will be thrown [here](https://github.com/vercel/ai/blob/c055bb9c0fe71cbbe8696b441de800daf6c1f61a/packages/ai/src/generate-text/stream-text.ts#L838C8-L848C16)

```javascript
        if (part.type === 'reasoning-end') {
          const activeReasoning = activeReasoningContent[part.id];


          if (activeReasoning == null) {
            controller.enqueue({
              part: {
                type: 'error',
                error: `reasoning part ${part.id} not found`,
              },
              partialOutput: undefined,
            });
```

I will submit a PR with a fix for this issue

### AI SDK Version

- ai: 5.0.119 - 5.0.123

I tested this issue on versions 5.0.119 to 5.0.123 as OpenCode is using version 5.0.119 currently, and using versions >6.0.0 seems to introduce breaking changes in OpenCode. However, this issue should theoretically be applicable to the latest versions too as there are similarly no checks if `reasoning-start` is emitted.

### Code of Conduct

jovanwongzixi:
## Background

This change is necessary due to the issue #12054, more details are included there.

## Summary

I added a check to determine if `reasoning-start` as already been emitted. When `reasoning-start` has not emitted, `activeExtraction.isFirstReasoning && activeExtraction.afterSwitch` will be `true`, so I have used that as a check to determine if a `reasoning-start` event needs to be sent. `reasoning-0` will then be created and the subsequent `reasoning-end` will not trigger an error.

## Manual Verification

After my changes, no thinking blocks will be displayed (for OpenCode) and the error is resolved. The sample script I used for setting up a mock LLM server is located at this [gist](https://gist.github.com/jovanwongzixi/f18645de16059e8e61f8d72fbb257f04).

reasoning_part_reasoning-0_not_found_resolved

## Checklist

Please advise if additional tests and documentation are needed for this change.

## Future Work

My implementation assumes `startWithReasoning` is initialised as `false`. I have yet to test there are any breaking changes if `startWithReasoning` is `true.

## Related Issues

#12054


aayush-kapoor (commented):
please add unit tests to simulate the bug so that we can verify this fixes it.

also add a changeset file. look into our contributing guidelines for more info

jovanwongzixi:
I have pushed a unit test to test if `reasoning-start` will be emitted before `reasoning-end` when given empty thinking blocks in `extract-reasoning-middleware-test.ts`. Changeset added too. Let me know if any further changes are needed.

.changeset/rotten-moons-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'ai': minor
```
aayush-kapoor: for future changes, please only do a 'patch' change

aayush-kapoor (approved):

.changeset/rotten-moons-bow.md
jovanwongzixi: Got it, thanks!

PR #12086, opened 2026-01-28
fix(xai): emit reasoning-start before reasoning-end for encrypted reasoning

jeremyphilemon:
## Summary

Fixes "reasoning part not found" errors when streaming responses from xAI models with encrypted reasoning (e.g., `grok-4-fast-reasoning`, `grok-4.1-fast-reasoning`).

## Problem

When xAI returns encrypted reasoning content, the `response.output_item.done` event may arrive without a prior `response.reasoning_summary_part.added` event. This caused `reasoning-end` to be emitted without a preceding `reasoning-start`, which the AI SDK's stream processing logic cannot handle.

The AI SDK tracks active reasoning parts and expects `reasoning-start` before `reasoning-end`:
```
reasoning part reasoning-rs_XXX not found
```

## Solution

Track which reasoning blocks have had `reasoning-start` emitted. Before emitting `reasoning-end`, check if `reasoning-start` was already emitted for that block. If not, emit `reasoning-start` first.

## Test Plan

- Run existing xai package tests: `pnpm test --filter @ai-sdk/xai`
- Test streaming with `xai/grok-4-fast-reasoning` model to verify no "reasoning part not found" errors

### DIFF
--- a/.changeset/lemon-books-poke.md
+++ b/.changeset/lemon-books-poke.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fix(xai): emit reasoning-start before reasoning-end for encrypted reasoning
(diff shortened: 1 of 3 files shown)


dancer:
looks good tested it out myself and test passes after the fix

dancer (approved):

PR #12089, opened 2026-01-28
feat(anthropic): support custom tool-reference content for deferred tool loading

aayush-kapoor:
## Background

anthropic's deferred tool loading allows custom tool search implementations that return tool_reference blocks, but we didn't support this format. [reference here](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool#custom-tool-search-implementation)

## Summary

added support for custom tool-reference content for deferred tool loading

## Manual Verification

verified via a reproduction script given by the user in `examples/ai-functions/src/generate-text/11829-repro.ts` (will be deleted after review)

## Checklist

## Related Issues

fixes #11829


lgrammel (changes requested):
this is a potential breaking specification change that needs to wait until ai sdk 7

packages/provider/src/language-model/v3/language-model-v3-prompt.ts
aayush-kapoor, line 421: @lgrammel what if we added the type in `custom` object instead to avoid breaking changes

gr2m: I think it's a good use of it, since it's a single provider that needs tool-reference content so far

lgrammel:
Looks good. PR title and description are out of date. Presumably there is no way to do this without using a special tool result part?

aayush-kapoor:
> Presumably there is no way to do this without using a special tool result part?

unfortunately no - they need a tool_reference object in the content part of the tool result to map the result.

also updated the PR description and title; ready for review again

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -355,6 +355,23 @@ export async function convertToAnthropicMessagesPrompt({
 
                             return undefined;
                           }
+                          case 'custom': {
+                            const anthropicOptions = contentPart.providerOptions
```
felixarntz, line 359: Could this also be done in a part that's not `custom`? Asking since parts of other types also support `providerOptions`. We could check presence of provider-specific `tool-reference` data regardless of type.

Maybe `custom` is the way to go, but wanted to double check.

lgrammel (approved):

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
aayush-kapoor: yes, technically, it could be added to any part that is not `custom` since all parts support providerOptions. but ideally we have different result content parts defined for each provider for maximum type safety. 

defining this in a custom part gets us as close to type safety while still being modular and also it not leading to a breaking change

PR #12120, opened 2026-01-29
fix(groq): transcription response_format enforced and words to segments mapping implemented 

### LINKED ISSUE #12118: Groq providerOptions for TranscriptionModel should not permit the user passing responseFormat = "text"
### Description

````
  const result = await transcribe({
    model: groq.transcription('whisper-large-v3'),
    audio: await readFile(path.join('./sample_long.mp3'
    )),
    providerOptions: {groq: {responseFormat: 'text'}}
  });

  console.log('Segments:', result);
````
````
APICallError [AI_APICallError]: Invalid JSON response
    at <anonymous> (/Users/shelly/ai-sdk/tests/gpt-4o-transcribe-diarize/node_modules/@ai-sdk/provider-utils/src/response-handler.ts:117:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
````
This code will be met by an AI SDK Error because the AI SDK currently has an established pattern of enforcing JSON outputs from transcription models. 

In the OpenAI provider, models are forced to output either `json` or `verbose_json` dependent on what is allowed by the API. For an immediate solution, this pattern could easily be used for the two transcription models that Groq offers.

### AI SDK Version

latest

### Code of Conduct

### LINKED ISSUE #12119: Groq TranscriptionModel lacks DX parity with OpenAI and does not properly map words to AI SDK Segments
### Description

The `timestamp_granularities` property in transcription APIs means that there will be an iterable returned for each element in the array (assuming things are passed properly). 

While users can easily access both iterables (words and segments) through the full responses body, the groq implementation lacks parity with openai.

````
const result = await transcribe({
  model: groq.transcription('whisper-large-v3'),
  audio: await readFile(path.join('./sample_long.mp3')),
  providerOptions: {
    groq: {
      language: 'en',
      responseFormat: 'verbose_json',
      timestampGranularities: ['word'], // this is required for timestamp granularities
    },
  },
});

console.log('Segments:', result.segments);
````

The above code results in an empty result.segments, but for now the best solution may be to follow the openAI pattern and map words into segments for cases like this. That way users who never access the full response can still get a somewhat predictable result.segments that has parity with other providers.

Also, perhaps the DX should be more clearly suggesting users access the full response for both words and segments. It may require exposing the body property in the `transcription-model-response-metadata.ts` but I did not fully verify those typescript interactions. This could eliminate confusion until the team lands on a better solution. 

### AI SDK Version

latest

### Code of Conduct

### PR DESCRIPTION
**Pull request by psinha40898**
## Background

1) Setting a responseFormat of text is currently possible and this breaks the DX. It's also explicitly addressed in the open ai provider's handling of the same situation by enforcing defaults. A default of verbose_json seemed like it makes sense.

2) The "words" iterable is never accessible through the AI SDK unless the user inspects the full response body. Following the solution in the open ai provider improves the DX and retains greater AI SDK parity across providers. 

## Summary

1)
````
  const result = await transcribe({
    model: groq.transcription('whisper-large-v3'),
    audio: await readFile(path.join('./sample_long.mp3'
    )),
    providerOptions: {groq: {responseFormat: 'text'}}
  });

  console.log('Segments:', result);
 ````
 
The above code will no longer break because responseFormat will be enforced to "verbose_json" when a providerOptions object exists. If it does not exist, it will default to "json" which is also acceptable by the AI SDK validator. This is the same solution the OpenAI Provider uses to prevent response_format = "text" reaching the API and erroring the SDK.

2)
````
const result = await transcribe({
  model: groq.transcription('whisper-large-v3'),
  audio: await readFile(path.join('./sample_long.mp3')),
  providerOptions: {
    groq: {
      language: 'en',
      responseFormat: 'verbose_json',
      timestampGranularities: ['word'], // this is required for timestamp granularities
    },
  },
});

console.log('Segments:', result.segments);
````
The above code will no longer output an empty array. The words will be mapped to the segments, just like the OpenAI Provider does.

## Manual Verification

I compared fetch requests to the Groq API with the SDKs functionality. On main code example 1 errors, and code example 2 results in an empty `result.segments`. On this branch code example 1 defaults to `verbose_json`, and code example 2 populates the `result.segments` with the API's `words` response.

## Checklist

## Future Work

 
there is still room for improvement to let users access both iterables through the AI SDK result. It seemed beyond the scope of this PR for a number of reasons

## Related Issues

Fixes #12119
Fixes #12118

The examples can be improved if something like 12409 lands.
Related to #12409

### CODE REVIEW

**Comment by psinha40898** [comment]
@lgrammel 
I wonder if contributions like this make sense to review at this time. 

There are also a few other transcription related issues that could be solved by extending opinions already established in the codebase, but I understand it is an experimental feature. Thank you.

File: packages/groq/src/groq-transcription-model.ts [hunk 1]
```diff
@@ -52,7 +52,7 @@ export class GroqTranscriptionModel implements TranscriptionModelV3 {
     const formData = new FormData();
     const blob =
       audio instanceof Uint8Array
-        ? new Blob([audio])
+        ? new Blob([new Uint8Array(audio)])
```
psinha40898, line 55: constructor removes typescript error by enforcing that this is not SharedArrayBuffer

Issue #12164, opened 2026-01-30
Make AI SDK work in Workflow


gr2m:
## Create `WorkflowAgent`

- [x] Separate UIMessageChunk conversion from model streaming (#13780)
- [x] Use `experimental_streamModelCall` in doStreamStep (#13820)
- [x] Export `mergeAbortSignals` from `ai/internal` (#13616)
- [x] Wire missing callbacks through WorkflowAgent API (#14036)
- [x] Add prepareCall support (#14037)
- [x] Add workflow serialization support to all provider models (#13779)
- [x] Add needsApproval support (#14084)
- [x] Extract shared mergeCallbacks utility (done in #14036)

## Follow-ups

- [ ] **Align callback event shapes with ToolLoopAgent.** WorkflowAgent's callback events are simpler than ToolLoopAgent's. ToolLoopAgent events include `callId`, `provider`, `modelId`, `stepNumber`, `messages`, `abortSignal`, `functionId`, `metadata`, `experimental_context`, typed `toolCall` with `TypedToolCall<TOOLS>`, and `durationMs` on tool call finish. Once the event shapes converge, the callback types could be unified as shared `AgentOnStartCallback`, `AgentOnStepStartCallback`, etc.
- [ ] **Expand prepareCall to full ToolLoopAgent parity.** Add `stopWhen`, `activeTools`, and `experimental_download` to `PrepareCallOptions`/`PrepareCallResult`. Support typed `CALL_OPTIONS` that flow through `prepareCall` as `options`.
- [ ] **Async headers providers.** Four providers have async `getHeaders` which can't be resolved synchronously at serialization time: Gateway (async OIDC token resolution), Amazon Bedrock anthropic subprovider (async SigV4 credential loading), KlingAI (async JWT generation), Google Vertex (async `Resolvable` headers).
- [ ] **Telemetry integration listeners (Phase 5).** Wire through telemetry integration listeners from `experimental_telemetry`. Call them alongside agent callbacks at each lifecycle point.
- [ ] **Clean up duplication (Phase 6 remaining).** Remove duplicate `filterTools` (use one location). Replace `getErrorMessage` with import from `@ai-sdk/provider-utils`. Remove `safeParseInput` if unused. Simplify `prepareStep` override application in `streamTextIterator`.
- [ ] Provider metadata filtering across turns (#14293)
- [ ] Binary data normalization across step boundaries (#14294)
[...]

PR #12165, opened 2026-01-30
🆕 WorkflowAgent (`@ai-sdk/workflow`)

gr2m:
Create a new `@ai-sdk/workflow` package that exports `WorkflowAgent`, which will be the successor of [DurableAgent](https://useworkflow.dev/docs/api-reference/workflow-ai/durable-agent)

## ToolLoopAgent parity plan

The underlying streamText in core already supports all 6 callback types. The gap is that WorkflowAgent doesn't accept or pass them through. However, WorkflowAgent doesn't call streamText directly — it uses streamTextIterator → doStreamStep → streamModelCall. So the callbacks need to be threaded through that chain.

### Phase 1: Wire missing callbacks through WorkflowAgent API ✅ (#14036)

1. ~~Add the 4 missing callback types to WorkflowAgentOptions and WorkflowAgentStreamOptions interfaces~~
2. ~~Add mergeCallbacks utility (extracted from ToolLoopAgent pattern)~~
3. ~~Pass callbacks through streamTextIterator to doStreamStep (which uses streamModelCall)~~
4. ~~Emit callbacks at the right points in the iterator loop~~

Unblocked 14 of 16 GAP tests. 2 remain as `it.fails()` to track event shape parity (see below).

**Remaining work from Phase 1: Align callback event shapes with ToolLoopAgent.** WorkflowAgent's callback events are simpler than ToolLoopAgent's. ToolLoopAgent events (defined in `core-events.ts`) include `callId`, `provider`, `modelId`, `stepNumber`, `messages`, `abortSignal`, `functionId`, `metadata`, `experimental_context`, typed `toolCall` with `TypedToolCall<TOOLS>`, and `durationMs` on tool call finish. WorkflowAgent events currently only provide a subset (e.g., `onToolCallStart` only has `toolCall` with a plain `ToolCall` type). Once the event shapes converge, the callback types could be unified as shared `AgentOnStartCallback`, `AgentOnStepStartCallback`, etc. instead of separate `WorkflowAgent*` and `ToolLoopAgent*` types.

### Phase 2: Add prepareCall support ✅ (#14037)

1. ~~Add prepareCall to WorkflowAgentOptions~~
2. ~~Call it in stream() before the iterator, similar to ToolLoopAgent's prepareCall()~~

Unblocked 1 GAP test.

**Remaining work from Phase 2:** ToolLoopAgent's `prepareCall` also supports `stopWhen`, `activeTools`, and `experimental_download` in its input/output types — these are not yet in WorkflowAgent's `PrepareCallOptions`/`PrepareCallResult`. Additionally, ToolLoopAgent supports typed `CALL_OPTIONS` that flow through `prepareCall` as `options` — WorkflowAgent doesn't have this concept.

### Phase 3: Add workflow serialization support to all provider models ✅ (#13779)

Adds `WORKFLOW_SERIALIZE`/`WORKFLOW_DESERIALIZE` to all 59 provider model classes (language, image, embedding, speech, transcription, video). Adds `serializeModel()` and `deserializeModelConfig()` helpers to `@ai-sdk/provider-utils`:
- `serializeModel` resolves `config.headers()` at serialization time so auth credentials survive the step boundary as plain key-value objects
- `deserializeModelConfig` wraps plain-object headers back into a function on deserialization

Makes `headers` optional in all provider config types so deserialized models work without pre-configured auth. Includes documentation for third-party provider authors.

**Remaining work from Phase 3: async headers providers.** Four providers have async `getHeaders` which can't be resolved synchronously at serialization time. These need per-provider handling or a model factory function workaround:
- **Gateway** — async OIDC token resolution (`AI_GATEWAY_API_KEY` env var fallback)
- **Amazon Bedrock** (anthropic subprovider) — async SigV4 credential loading (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars)
- **KlingAI** — async JWT generation from `KLINGAI_ACCESS_KEY`/`KLINGAI_SECRET_KEY` env vars
- **Google Vertex** — async `Resolvable` headers (`GOOGLE_VERTEX_API_KEY` env var for express mode)

### Phase 4: Add needsApproval support ✅ (#14084)

1. ~~Before executing a tool, check tool.needsApproval (boolean or async function)~~
2. ~~If approval needed, pause the loop and return pending tool calls (like client-side tools)~~
3. ~~Handle approval resumption: collect tool-approval-response parts, execute approved tools, create denial results~~
4. ~~Write tool results and step boundaries to the UI stream so tool parts transition to output-available state and convertToModelMessages produces correct message structure for multi-turn conversations~~

This unblocked 2 GAP tests.

### Phase 5: Telemetry integration listeners

1. Wire through telemetry integration listeners from experimental_telemetry
2. Call them alongside agent callbacks at each lifecycle point

This unblocks 3 GAP tests.

### Phase 6: Clean up duplication

1. ~~Extract shared mergeCallbacks utility~~ ✅ (done in #14036 — moved to `ai/internal`, used by both ToolLoopAgent and WorkflowAgent)
2. Remove duplicate filterTools (use one location)
3. Replace getErrorMessage with import from @ai-sdk/provider-utils
4. Remove safeParseInput if unused
5. Simplify prepareStep override application in streamTextIterator

## Future work

- #14293 — Provider metadata filtering across turns
- #14294 — Binary data normalization across step boundaries
- #14292 — WorkflowAgent has the inverse bug: provider-executed tool approvals are silently dropped (sent zero times) on the approval resume path

## Done

- [x] **Separate UIMessageChunk conversion from model streaming** (#13780)
  Extracts UIMessageChunk conversion from `doStreamStep` into a standalone utility, making the model streaming layer independent of UI concerns. `doStreamStep` returns raw `LanguageModelV4StreamPart[]` chunks; UIMessageChunk conversion is a separate, optional step. `writable` becomes optional in `WorkflowAgentStreamOptions` — when omitted, the agent streams ModelMessages only. Follows `streamText`'s `toUIMessageStream()` pattern.
- [x] **Use `experimental_streamModelCall` in doStreamStep** (#13820)
  Replace doStreamStep internals with `experimental_streamModelCall`. Eliminates ~300 lines of duplicated stream transformation, gains tool call parsing/repair, retry logic, and `Experimental_ModelCallStreamPart` stream types.
- [x] **Export `mergeAbortSignals` from `ai/internal`** (#13616)
  Exports the existing `mergeAbortSignals` utility from `ai/internal` and replaces the manual ~25-line abort signal + timeout merging code in WorkflowAgent with the shared utility. Uses `AbortSignal.timeout()` instead of manual `setTimeout` + `AbortController`, matching how `generateText`/`streamText` handle the same concern.
- [x] **Wire missing callbacks (Phase 1)** (#14036)
  Adds `experimental_onStart`, `experimental_onStepStart`, `experimental_onToolCallStart`, `experimental_onToolCallFinish` callbacks to WorkflowAgent. Extracts `mergeCallbacks` into `ai/internal` as shared utility used by both ToolLoopAgent and WorkflowAgent. Also fixes `sideEffects: false` breaking workflow step discovery and replaces `resolveLanguageModel` from `ai/internal` with `gateway` from `ai` to fix Next.js webpack resolution in step bundles.
- [x] **Add prepareCall support (Phase 2)** (#14037)
  Adds `prepareCall` callback to WorkflowAgentOptions, called once before the agent loop to transform model, instructions, generation settings, etc. `tools` excluded from return type since they're bound at construction time for type safety.

## No longer pursued

- ~~**Stream-to-StepResult aggregator** (#13459)~~
  `chunksToStep` was removed when `experimental_streamModelCall` was adopted in #13820.
- ~~**Public message conversion (step → messages for next turn)** (#13474)~~
  `toolsToModelTools` was removed when `experimental_streamModelCall` was adopted in #13820.
- ~~**Tool descriptor extraction** (#13477)~~
  `toolsToModelTools()` was removed with the `streamModelCall` refactor.
- ~~**Extract V3→UIMessageChunk stream transform as shared utility** (#13433)~~
  Superseded by the `streamModelCall` refactor which eliminated the V3 stream layer.
- ~~**String-based model IDs in core functions**~~
  Already supported — `LanguageModel` type accepts strings, all core functions resolve them via the global provider.
- ~~**Declarative stop conditions**~~
  Not necessary — `stopWhen` runs in the agent's control loop outside step boundaries, so function-based conditions work without serialization. `maxSteps` covers the simple case as a serializable number.

## Notes

- **Resumption:** the client-side is expected to specify which chunk to resume from at the workflow stream level (the V4StreamPart). If that maps 1-1 to UIMessageChunk, or if there's a way to map the correct resumption index from the number of UIMessageChunk parts that the client-side has received, that's what needs to be considered.
- **`repairToolCall` not serializable across step boundaries.** `ToolCallRepairFunction` is a function and can't cross the `'use step'` serialization boundary. Left out of WorkflowAgent for now — `experimental_streamModelCall` handles repair internally when called outside a step boundary.
- **`abortSignal` serialization.** `AbortSignal` objects can't be serialized across step boundaries. The Workflow team is working on adding serialization support for abort signals.
- **Tools validation:** because of serialization/deserialization needed for step functions, we use validation capabilities and transformations of libraries like `zod`. Serialization is lossy.

## Related issues

- #12164


gr2m:
un-drafting to trigger review by Vercel Agent

gr2m:
The snapshot build published (https://github.com/vercel/ai/actions/runs/21530986703/job/62046593364) published `@ai-sdk/durable-agent@0.0.1` instead of the usual snapshot versions. That was not planned.

KaiKloepfer:
@gr2m I did some work on improving this from the workflow side, might be relevant to you? https://github.com/vercel/workflow/pull/928 Still not full compatibility, but we got stuck on the same issues when trying to port a v6 app to use workflow.

rovo89:
@gr2m I think this is the right approach, avoiding a lot of compatibility layers. However, it seems that the code is copied more or less 1:1 from the workflow code, which has simplified lots of things. For example, OpenAI's `web_search` doesn't work for me because `toolsToModelTools()` is extremely basic. I'm hoping this will be much closer to the original `ToolLoopAgent`.

I would offer to help, but I assume you already have own ideas how things should work (and I know little about the AI SDK internals). Anyway, if I can do anything, I'm happy to help.

gr2m:
@KaiKloepfer thanks will have a look!

@rovo89  I'm focused on https://github.com/vercel/ai/issues/12381 right now, please feel free to send PRs for exploration of different approaches.

rovo89:
My current attempt is to simpy use ToolLoopAgent in a step function. 🙈

```ts

export async function chat(writable: WritableStream<UIMessageChunk>, messages: UIMessage[]) {
  'use step';
  const agent = new ToolLoopAgent({...});
  const stream = await createAgentUIStream({
    agent,
    uiMessages: messages,
  })
  await stream.pipeTo(writable)
}
```

Streaming works fine and I can use it exactly like I'm used to.

DurableAgent has quite some limitations:
- Many details are implemented as very simplified stubs, such as the tool preparation which prevents using OpenAI provider tools.
- The default downloader doesn't work because fetch is blocked (need to use their fetch step function instead) and supportedUrls is empty.
- Model has to be provided as step function, therefore requires extra layer / provider wrapper packages. Custom provider registries are harder to use.
- Message metadata is a lot harder to add, requires disabling default start/finish chunks and sending them separately.
- ...

But of course, it has benefits. As far as I understood:
- Can resume multi-step calls instead of repeating from scratch. Not sure what happens if an error occurs mid-step (e.g. network failure) - will probably run that step again, but what about the already sent chunks?!?
- Tool calls are steps, so they can also be retried.
- Hooks can simplify human-in-the-loop / tool approvals.
- More stable environment for the loop controller.

Some ideas how similar features could be achieved in (a subclass of) ToolLoopAgent:
- Read back the stream to reconstruct the message chunks so far and continue from there. Or more abstract: Needs a way to "preload" the loop with the previously recorded intermediate step results and continue with the next step.
- AFAIU, calling a step function from another step has no special meaning, so no retries etc. That's only when they're called from a workflow function. Similar for hooks. Generally, ToolLoopAgent and the functions it calls are somehow orchestrating (like workflow functions), but they're far from just stitching steps together. That's why I think it makes sense to run them in a step function, but they could benefit from sub-steps. That needs further thoughts.
-  Of course, this assumes that the setup is still the same on the next retry, like all tools being defined in the same way, which seems to be guaranteed in the workflow VM. Then again, how big is the risk that this happens accidently and how big could the damage be?

gr2m:
I want to try two different approaches in separate PRs against `gr2m/durable-agent`

1. Try to refactor `streamText()` itself so that it works in a workflow context.
2. Refactor `streamText()` to export lower-level orchestration code which then can be used by DurableAgent

I think `1.` won't be possible for several reasons but I want to see how far I can get.

For the providers, as a start I want to re-export all first-party providers with the symbols needed for workflow step serialization/deserialization and the `"use step"` directive in `doStream()`

image

gr2m:
We are working from the AI SDK's side towards decomposing `streamText()` into lower-level components that we will export, with the goal of externalizing loop control. `streamText()` would be one implementation of loop control specific to the AI SDK, but others such as Workflow, Gateway, and community projects will be able to do the same.

https://github.com/vercel/ai/issues/13570

gr2m:
Exploring an in-between step: a simplified `workflowStreamText()` implementation to see how far we get with exported internals for `ai` so far

<details>

# Plan: `@ai-sdk/workflow-stream-text` package

## Context

The Workflow DevKit's `DurableAgent` currently **bypasses AI SDK's `streamText()` entirely** and calls `model.doStream()` directly in its `doStreamStep` function. This results in massive code duplication:

- `doStreamStep` (~750 lines) reimplements: stream→UIMessage conversion, chunk aggregation into StepResult, tool call extraction
- `streamTextIterator` (~520 lines) reimplements: the tool loop, prepareStep, stop conditions, message accumulation
- `DurableAgent` (~1680 lines) reimplements: tool execution, filtering, abort handling, output parsing, callback merging

**Root cause:** AI SDK's `streamText()` has a monolithic architecture where the tool loop and step orchestration are **internal to `DefaultStreamTextResult`'s constructor** — there's no way to intercept tool execution to make it durable.

**Goal:** Create a minimal `@ai-sdk/workflow-stream-text` that provides a workflow-compatible `streamText()` by reusing AI SDK internals. Start with the simplest working version, defer advanced features to follow-ups.

## Architecture

### Streaming at the `TextStreamPart` level (not UIMessageChunk)

`workflowStreamText` operates at the **model level**, returning `TextStreamPart<TOOLS>` — the same stream type as `streamText().fullStream`. UIMessageChunk conversion is a **separate, composable concern** handled by the existing `toUIMessageStream()` pattern or a helper.

This means:
- The core function is presentation-agnostic
- Consumers who want raw model parts can use them directly
- UIMessageChunk conversion can be done by piping the result stream through a transform before writing to `getWritable()`

### Two layers

**Layer 1: `doStreamTextStep()` — new AI SDK internal**
A single model call that returns a `TextStreamPart` stream + promises for the aggregated result. No tool loop. Built on `modelCall()`.

**Layer 2: `workflowStreamText()` — new package**
An async generator that orchestrates the tool loop by calling `doStreamTextStep()` repeatedly, yielding tool calls to the caller for durable execution.

## Scope: Minimal first version

### In scope (MVP):
- `doStreamTextStep()` — single step: model call → `TextStreamPart` stream → StepResult
- `workflowStreamText()` — tool loop as async generator, yielding tool calls
- `stopWhen` / `maxSteps` support (reuses existing `StopCondition` from AI SDK)
- Basic `prepareStep` support
- Message accumulation across steps
- Return type with `messages`, `steps`, `toolCalls`, `toolResults`

### Deferred to follow-ups:
- Telemetry / OpenTelemetry integration
- Timeout handling (step timeout, chunk timeout, total timeout)
- `experimental_output` (structured output parsing)
- `experimental_transform` (stream transforms)
- `experimental_repairToolCall`
- `experimental_download` (custom URL download)
- `includeRawChunks`
- `onChunk`, `onError`, `onFinish`, `onStepFinish`, `onStepStart`, `onToolCallStart`, `onToolCallFinish` callbacks
- Provider-executed tool handling
- Tool approval flow
- UIMessageChunk conversion helper

## Implementation

### Phase 1: `doStreamTextStep` (AI SDK internal)

**New file:** `packages/ai/src/generate-text/do-stream-text-step.ts`

```typescript
export interface DoStreamTextStepOptions<TOOLS extends ToolSet> {
  model: LanguageModelV4;
  tools?: TOOLS;
  toolChoice?: ToolChoice<TOOLS>;
  system?: string | SystemModelMessage | Array<SystemModelMessage>;
  messages: ModelMessage[];
  prompt: LanguageModelV4CallOptions['prompt']; // pre-converted prompt
  callSettings: Omit<LanguageModelV4CallOptions, ...>;
  maxRetries?: number;
  abortSignal?: AbortSignal;
  headers?: Record<string, string | undefined>;
  repairToolCall?: ToolCallRepairFunction<TOOLS>;
}

export interface DoStreamTextStepResult<TOOLS extends ToolSet> {
  stream: ReadableStream<TextStreamPart<TOOLS>>;
  stepResult: Promise<StepResult<TOOLS>>;
  toolCalls: Promise<TypedToolCall<TOOLS>[]>;
  request: LanguageModelRequestMetadata;
  response: { headers?: SharedV4Headers } | undefined;
}
```

Implementation:
1. Call `modelCall()` → get raw transformed stream + request/response
2. Pipe through a transform that:
   - Collects content parts (text, reasoning, tool-calls, sources, files)
   - Tracks tool calls
   - Captures finish reason, usage, response metadata
   - Emits `start-step` / `finish-step` markers
   - Converts `UglyTransformedStreamTextPart` → `TextStreamPart`
3. Tee the stream: one for the caller to consume/pipe, one for internal aggregation
4. Background-consume the aggregation tee to resolve `stepResult` and `toolCalls` promises

**Modify:** `packages/ai/internal/index.ts` — add export

### Phase 2: `@ai-sdk/workflow-stream-text` package

**New package:** `packages/workflow-stream-text/`

```typescript
// src/workflow-stream-text.ts
export async function* workflowStreamText<TOOLS extends ToolSet>(options: {
  model: LanguageModel;
  tools?: TOOLS;
  toolChoice?: ToolChoice<TOOLS>;
  system?: string;
  messages: ModelMessage[];
  maxSteps?: number;
  stopWhen?: StopCondition<TOOLS> | StopCondition<TOOLS>[];
  prepareStep?: PrepareStepFunction<TOOLS>;
  maxRetries?: number;
  abortSignal?: AbortSignal;
  headers?: Record<string, string | undefined>;
}): AsyncGenerator<WorkflowStreamTextStepResult<TOOLS>> {
  // 1. standardizePrompt() + convertToLanguageModelPrompt()
  // 2. Loop:
  //    a. prepareStep() if provided
  //    b. prepareToolsAndToolChoice()
  //    c. doStreamTextStep() → get stream + promises
  //    d. yield { stream, stepResult, toolCalls } to caller
  //    e. Caller consumes stream (piping to writable or wherever)
  //    f. Await stepResult — check for tool calls
  //    g. If tool calls & !stopCondition: caller provides tool results via .next(toolResults)
  //    h. Build response messages, append to conversation, continue loop
  //    i. If no tool calls or stop condition met: done
}
```

The caller controls:
- **Where the stream goes** — pipe to `getWritable()`, or consume directly
- **How tools are executed** — as `"use step"` functions, or any other way
- **When to stop** — via `stopWhen` or by breaking the generator

### Usage example (in workflow context):

```typescript
async function myAgent(userMessage: string) {
  "use workflow";

  const iterator = workflowStreamText({
    model: openai("gpt-4o"),
    tools: { search: searchTool },
    messages: [{ role: "user", content: userMessage }],
    maxSteps: 10,
  });

  for await (const { stream, toolCalls, stepResult } of iterator) {
    // Pipe stream to workflow writable (caller handles conversion)
    await stream.pipeTo(writable, { preventClose: true });

    // Check for tool calls
    const calls = await toolCalls;
    if (calls.length > 0) {
      // Execute tools as durable steps
      const results = await Promise.all(
        calls.map(tc => executeToolStep(tc))
      );
      // Feed results back via iterator protocol
      // (implementation detail: may use .next() or options callback)
    }
  }
}
```

## Files to create/modify

### AI SDK core (`packages/ai/`):
1. **NEW** `src/generate-text/do-stream-text-step.ts` — single-step streaming primitive
2. **NEW** `src/generate-text/do-stream-text-step.test.ts` — tests
3. **MODIFY** `internal/index.ts` — add export

### New package (`packages/workflow-stream-text/`):
4. **NEW** `package.json`
5. **NEW** `tsconfig.json`
6. **NEW** `tsup.config.ts`
7. **NEW** `src/index.ts`
8. **NEW** `src/workflow-stream-text.ts`

## Reused from AI SDK:
- `modelCall()` — `packages/ai/src/generate-text/model-call.ts`
- `convertToLanguageModelPrompt()` — `packages/ai/src/prompt/convert-to-language-model-prompt.ts`
- `standardizePrompt()` — `packages/ai/src/prompt/standardize-prompt.ts`
- `prepareToolsAndToolChoice()` — `packages/ai/src/prompt/prepare-tools-and-tool-choice.ts`
- `prepareCallSettings()` — `packages/ai/src/prompt/prepare-call-settings.ts`
- `toResponseMessages()` — `packages/ai/src/generate-text/to-response-messages.ts`
- `isStopConditionMet`, `stepCountIs` — `packages/ai/src/generate-text/stop-condition.ts`
- `StepResult`, `ToolSet`, `TextStreamPart` types — from `ai`

## Future: `streamText()` should use `doStreamTextStep()`

The eventual goal is for **both** `streamText()` and `workflowStreamText()` to share `doStreamTextStep()` as their single-step primitive:

- `streamText()` would call `doStreamTextStep()` inside its internal `streamStep()` recursive function
- `workflowStreamText()` would call it inside its async generator loop
- The only difference: who orchestrates the loop and who executes tools

**Why not do this now:** `streamText()` is a 2700-line file with extensive tests and subtle behaviors (deferred tool results, tool approvals, chunk timeouts, provider-executed tools, telemetry integration). Swapping its internals to use `doStreamTextStep()` requires verifying all of that still works — high risk, high surface area. Better to:

1. Prove `doStreamTextStep()` works in the simpler workflow context first
2. Then migrate `streamText()` to use it in a focused follow-up PR with targeted regression testing

This way the new primitive is validated before becoming load-bearing for the entire SDK.

## Verification
1. Unit tests for `doStreamTextStep` — mock model, verify stream parts + StepResult
2. Unit tests for `workflowStreamText` — verify tool loop, stop conditions, message accumulation
3. `pnpm type-check:full` from workspace root
4. `pnpm build` in both packages
5. `pnpm test` in `packages/ai` — no regressions

</details>

rovo89:
@gr2m Not sure if you're planning to add `onStepFinish` callbacks. If you do, please consider using a proper type: https://github.com/vercel/ai/pull/12843

gr2m:
Thanks @rovo89 we'll have a look next week!

skills/workflow/SKILL.md
gr2m, line 1: Will delete before merging

skills-lock.json
gr2m, line 1: Delete before merging

.github/workflows/ci.yml
```diff
@@ -9,6 +9,8 @@ on:
     branches:
       - main
       - release-v*
+      # TODO: remove before merging #12165
+      - gr2m/durable-agent
```
gr2m: delete before merging

packages/workflow/src/workflow-agent.ts
```diff
@@ -0,0 +1,2116 @@
...
+   *       sentiment: z.enum(['positive', 'negative', 'neutral']),
+   *       confidence: z.number(),
+   *     }),
+   *   }),
+   * });
+   *
+   * console.log(result.experimental_output); // { sentiment: 'positive', confidence: 0.95 }
+   * ```
+   */
+  experimental_output?: OutputSpecification<OUTPUT, PARTIAL_OUTPUT>;
```
felixarntz: should now be just `output`

felixarntz (commented):
I know this is still WIP, but random thing I noticed 👇

packages/ai/src/agent/merge-callbacks.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Merges two optional callbacks into one that calls both sequentially.
+ * The first callback (from constructor/settings) runs before the second (from method).
+ */
+export function mergeCallbacks<
```
lgrammel: move to utils

packages/ai/src/agent/merge-callbacks.ts
```diff
@@ -0,0 +1,18 @@
...
+  T extends (event: any) => PromiseLike<void> | void,
+>(
+  settingsCallback: T | undefined,
+  methodCallback: T | undefined,
```
lgrammel: ideally a list of callbacks; i think we have something similar already

lgrammel:
```
    function createTelemetryComposite<EVENT>(
      getListenerFromIntegration: (
        integration: TelemetryIntegration,
      ) => ((event: EVENT) => PromiseLike<void> | void) | undefined,
    ): ((event: EVENT) => Promise<void>) | undefined {
      const listeners = allIntegrations
        .map(getListenerFromIntegration)
        .filter(Boolean) as Array<(event: EVENT) => PromiseLike<void> | void>;

      return async (event: EVENT) => {
        for (const listener of listeners) {
          try {
            await listener(event);
          } catch (_ignored) {}
        }
      };
    }
```

packages/ai/src/telemetry/get-global-telemetry-integration.ts

does afaik something similar in a more typesafe way

we could alternatively extract the telemetry composite, make it reusable and tested as mergeCallbacks, and then use that in both places

lgrammel: i can take this on today since it seems from tool loop agent

packages/ai/src/model/resolve-model.ts
```diff
@@ -9,6 +9,7 @@ import {
   SpeechModelV4,
   TranscriptionModelV4,
 } from '@ai-sdk/provider';
+import '../global';
```
lgrammel: this seems like the wrong location -- we already do it in the ai sdk index so the issue might be that workflow agent needs this but relies on internal exports

why was this added?

packages/provider-utils/src/serialize-model.ts
```diff
@@ -0,0 +1,86 @@
...
+ * static [WORKFLOW_SERIALIZE](inst: MyLanguageModel) {
+ *   return serializeModel(inst);
+ * }
+ * ```
+ */
+// Parameter uses `any` because provider model classes declare `config` as
+// private, making it invisible to external type checks.  The static
+// WORKFLOW_SERIALIZE method has runtime access to the field regardless.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function serializeModel(inst: any): {
```
lgrammel: should this be typed to languagemodelv4 since the docs say it is for a language model (also should the param be called model)

packages/provider-utils/src/serialize-model.ts
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @example
+ * ```ts
+ * static [WORKFLOW_DESERIALIZE](options: { modelId: string; config: MyConfig }) {
+ *   return new MyLanguageModel(options.modelId, deserializeModelConfig(options.config));
+ * }
+ * ```
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function deserializeModelConfig<T>(config: T): T {
```
lgrammel: alternative: deserialize model that accepts model class prototype and config and then returns the model

(seems better if possible because then serializeModel / deserializeModel are clear opposites)

packages/provider-utils/src/serialize-model.ts
```diff
@@ -0,0 +1,86 @@
...
+  const result = { ...config } as any;
+  if (result.headers != null && typeof result.headers !== 'function') {
+    const resolvedHeaders = result.headers;
+    result.headers = () => resolvedHeaders;
+  }
+  return result;
+}
+
+function isSerializable(value: unknown): boolean {
```
lgrammel: is this model specific?

packages/ai/src/agent/merge-callbacks.ts
lgrammel: https://github.com/vercel/ai/pull/14315

packages/ai/src/model/resolve-model.ts
gr2m: It was somehow related to type errors with `globalThis` in that file. I'll investigate

packages/ai/src/agent/merge-callbacks.ts
gr2m: as in `packages/provider-utils` or `packages/ai/src/util` where it is now?
https://github.com/vercel/ai/blob/4e8fa0e3bb478b72b243a972e53c5834517b3159/packages/ai/src/util/merge-listeners.ts

packages/ai/src/model/resolve-model.ts
gr2m: this is what I get when running `pnpm build` in `packages/ai` if I remove the global import

```
src/model/resolve-model.ts(142,33): error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.
src/model/resolve-model.ts(144,5): error TS2578: Unused '@ts-expect-error' directive.
src/model/resolve-model.ts(203,31): error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.

Error: error occured in dts build
    at Worker.<anonymous> (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/tsup@7.2.0_@swc+core@1.15.3_postcss@8.5.9_ts-node@10.9.2_@swc+core@1.15.3_@types+node@2_aa91a706ebd252e59d5020891c06e0d9/node_modules/tsup/dist/index.js:2294:26)
    at Worker.emit (node:events:508:28)
    at MessagePort.<anonymous> (node:internal/worker:332:53)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:843:20)
    at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)
DTS Build error
RollupError: Failed to compile. Check the logs above.
    at error (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/rollup@3.29.5/node_modules/rollup/dist/shared/rollup.js:353:30)
    at Object.error (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/rollup@3.29.5/node_modules/rollup/dist/shared/rollup.js:1721:20)
    at Object.error (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/rollup@3.29.5/node_modules/rollup/dist/shared/rollup.js:25627:42)
    at generateDtsFromTs (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/tsup@7.2.0_@swc+core@1.15.3_postcss@8.5.9_ts-node@10.9.2_@swc+core@1.15.3_@types+node@2_aa91a706ebd252e59d5020891c06e0d9/node_modules/tsup/dist/rollup.js:7512:22)
    at /Users/gr2m/code/vercel/ai/node_modules/.pnpm/tsup@7.2.0_@swc+core@1.15.3_postcss@8.5.9_ts-node@10.9.2_@swc+core@1.15.3_@types+node@2_aa91a706ebd252e59d5020891c06e0d9/node_modules/tsup/dist/rollup.js:7519:59
    at _nullishCoalesce (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/tsup@7.2.0_@swc+core@1.15.3_postcss@8.5.9_ts-node@10.9.2_@swc+core@1.15.3_@types+node@2_aa91a706ebd252e59d5020891c06e0d9/node_modules/tsup/dist/rollup.js:1:198)
    at Object.transform (/Users/gr2m/code/vercel/ai/node_modules/.pnpm/tsup@7.2.0_@swc+core@1.15.3_postcss@8.5.9_ts-node@10.9.2_@swc+core@1.15.3_@types+node@2_aa91a706ebd252e59d5020891c06e0d9/node_modules/tsup/dist/rollup.js:7519:18)
    at /Users/gr2m/code/vercel/ai/node_modules/.pnpm/rollup@3.29.5/node_modules/rollup/dist/shared/rollup.js:1914:40
```

I'll try to find another way

packages/provider-utils/src/serialize-model.ts
gr2m: `.config` is a private implementation detail. We have access to it at runtime but typescript has no way to express that. Hence the `any`.

packages/ai/src/model/resolve-model.ts
gr2m: moved global import to `packages/ai/internal/index.ts` to mirror `packages/ai/src/index.ts`

packages/provider-utils/src/serialize-model.ts
gr2m: Yes some models are not serializable because of async `headers` methods

I mentioned this in the PR description

> **Remaining work from Phase 3: async headers providers.** Four providers have async `getHeaders` which can't be resolved synchronously at serialization time. These need per-provider handling or a model factory function workaround:
>
> - **Gateway** — async OIDC token resolution (`AI_GATEWAY_API_KEY` env var fallback)
> - **Amazon Bedrock** (anthropic subprovider) — async SigV4 credential loading (`AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` env vars)
> - **KlingAI** — async JWT generation from `KLINGAI_ACCESS_KEY`/`KLINGAI_SECRET_KEY` env vars
> - **Google Vertex** — async `Resolvable` headers (`GOOGLE_VERTEX_API_KEY` env var for express mode)

The workflow team is looking into support of async serialize methods which will unblock support for these providers.

packages/provider-utils/src/serialize-model.ts
gr2m: Done in c11b83e — added `deserializeModel(ModelClass, options)` as the symmetric opposite of `serializeModel`, and updated all provider model classes to use it.

packages/provider-utils/src/serialize-model.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @example
+ * ```ts
+ * static [WORKFLOW_DESERIALIZE](options: { modelId: string; config: MyConfig }) {
+ *   return deserializeModel(MyLanguageModel, options);
+ * }
+ * ```
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function deserializeModel<T>(
+  ModelClass: new (modelId: any, config: any) => T,
```
lgrammel: modelId: string, config: Record<string,unknown>

(per output from serializeModel)

packages/provider-utils/src/serialize-model.ts
```diff
@@ -0,0 +1,106 @@
...
+ * static [WORKFLOW_SERIALIZE](model: MyLanguageModel) {
+ *   return serializeModel(model);
+ * }
+ * ```
+ */
+// Parameter uses `any` because provider model classes declare `config` as
+// private, making it invisible to external type checks.  The static
+// WORKFLOW_SERIALIZE method has runtime access to the field regardless.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function serializeModel(model: any): {
```
lgrammel: model: LanguageModelV4

?

packages/provider-utils/src/serialize-model.ts
lgrammel: does this need to be exported deserializeModel is available? can T be sharpened (Record<string,unknown>

packages/provider-utils/src/serialize-model.ts
lgrammel: consider renaming: isSerializableModel(model: LanguageModelV4): boolean

packages/provider-utils/src/index.ts
```diff
@@ -78,3 +78,9 @@ export {
   EventSourceParserStream,
   type EventSourceMessage,
 } from 'eventsource-parser/stream';
+export {
+  serializeModel,
+  deserializeModel,
+  deserializeModelConfig,
```
lgrammel: export needed?

packages/provider-utils/src/index.ts
```diff
@@ -78,3 +78,9 @@ export {
...
+} from './serialize-model';
+export { WORKFLOW_SERIALIZE, WORKFLOW_DESERIALIZE } from '@workflow/serde';
```
lgrammel: this creates a dependency from provider utils to workflow - is there a transitive trail of dependencies? are there other solutions?

lgrammel: (looked at `@workflow/serde` and it seems minimal/fine as long as it does not get changed)

packages/provider-utils/package.json
```diff
@@ -46,6 +46,7 @@
   "dependencies": {
     "@ai-sdk/provider": "workspace:*",
     "@standard-schema/spec": "^1.1.0",
+    "@workflow/serde": "4.1.0-beta.2",
```
lgrammel: fixed version or flexible range?

packages/workflow/src/get-error-message.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Safely extract a human-readable message from an unknown error value.
+ *
+ * Avoids the `String(error)` pitfall that produces `"[object Object]"`
+ * when the thrown value is a plain object rather than an Error instance.
+ *
+ * Matches the behaviour of AI SDK's `getErrorMessage` from
+ * `@ai-sdk/provider-utils`.
+ */
+export function getErrorMessage(error: unknown): string {
```
lgrammel: seems duplicated (we have something like this already)

packages/workflow/src/stream-iterator.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+  });
+}
+
+/**
+ * Converts a ReadableStream to an AsyncIterable.
+ *
+ * @yields Items from the stream.
+ */
+export async function* streamToIterator<T>(
```
lgrammel: i think we have these 2 functions already

packages/workflow/src/types.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Shared types for AI SDK V4.
+ */
+import type { LanguageModelV4 } from '@ai-sdk/provider';
+
+/**
+ * Language model type for AI SDK V4.
+ *
+ * This is a simple alias for LanguageModelV4 from @ai-sdk/provider.
+ */
+export type CompatibleLanguageModel = LanguageModelV4;
```
lgrammel, line 11: Seems strange, shouldnt it work with all models via the ai sdk compat layer?

lgrammel (approved):
Had a few questions re type tightening and code reuse. Is this experimental/beta for now? PR description might need to be cleaned up.

packages/provider-utils/src/serialize-model.ts
lgrammel: I guess it needs to work with other model types and versions as well.

lgrammel: I wonder if we need a different pattern here regarding serialization to make it type safe, i.e. some sort of `SerializableModel` interface that uses a memo pattern.

packages/provider-utils/src/index.ts
gr2m: yes it's intentionally tiny.

packages/provider-utils/package.json
gr2m: I don't think flexible ranges work with beta releases. But the stable version is out now, will update

packages/workflow/src/types.ts
gr2m: I'd just leave it as `LanguageModelV4` for now and expand later if it becomes a problem

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `ai` | [`7.0.0-beta.89`](https://www.npmjs.com/package/ai/v/7.0.0-beta.89) |
| `@ai-sdk/alibaba` | [`2.0.0-beta.27`](https://www.npmjs.com/package/@ai-sdk/alibaba/v/2.0.0-beta.27) |
| `@ai-sdk/amazon-bedrock` | [`5.0.0-beta.32`](https://www.npmjs.com/package/@ai-sdk/amazon-bedrock/v/5.0.0-beta.32) |
| `@ai-sdk/angular` | [`3.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/angular/v/3.0.0-beta.89) |
| `@ai-sdk/anthropic` | [`4.0.0-beta.28`](https://www.npmjs.com/package/@ai-sdk/anthropic/v/4.0.0-beta.28) |
| `@ai-sdk/assemblyai` | [`3.0.0-beta.22`](https://www.npmjs.com/package/@ai-sdk/assemblyai/v/3.0.0-beta.22) |
| `@ai-sdk/azure` | [`4.0.0-beta.31`](https://www.npmjs.com/package/@ai-sdk/azure/v/4.0.0-beta.31) |
| `@ai-sdk/baseten` | [`2.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/baseten/v/2.0.0-beta.25) |
| `@ai-sdk/black-forest-labs` | [`2.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/black-forest-labs/v/2.0.0-beta.21) |
| `@ai-sdk/bytedance` | [`2.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/bytedance/v/2.0.0-beta.21) |
| `@ai-sdk/cerebras` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/cerebras/v/3.0.0-beta.25) |
| `@ai-sdk/cohere` | [`4.0.0-beta.22`](https://www.npmjs.com/package/@ai-sdk/cohere/v/4.0.0-beta.22) |
| `@ai-sdk/deepgram` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/deepgram/v/3.0.0-beta.21) |
| `@ai-sdk/deepinfra` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/deepinfra/v/3.0.0-beta.25) |
| `@ai-sdk/deepseek` | [`3.0.0-beta.23`](https://www.npmjs.com/package/@ai-sdk/deepseek/v/3.0.0-beta.23) |
| `@ai-sdk/devtools` | [`1.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/devtools/v/1.0.0-beta.14) |
| `@ai-sdk/elevenlabs` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/elevenlabs/v/3.0.0-beta.21) |
| `@ai-sdk/fal` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/fal/v/3.0.0-beta.21) |
| `@ai-sdk/fireworks` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/fireworks/v/3.0.0-beta.25) |
| `@ai-sdk/gateway` | [`4.0.0-beta.50`](https://www.npmjs.com/package/@ai-sdk/gateway/v/4.0.0-beta.50) |
| `@ai-sdk/gladia` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/gladia/v/3.0.0-beta.21) |
| `@ai-sdk/google` | [`4.0.0-beta.38`](https://www.npmjs.com/package/@ai-sdk/google/v/4.0.0-beta.38) |
| `@ai-sdk/google-vertex` | [`5.0.0-beta.48`](https://www.npmjs.com/package/@ai-sdk/google-vertex/v/5.0.0-beta.48) |
| `@ai-sdk/groq` | [`4.0.0-beta.24`](https://www.npmjs.com/package/@ai-sdk/groq/v/4.0.0-beta.24) |
| `@ai-sdk/huggingface` | [`2.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/huggingface/v/2.0.0-beta.25) |
| `@ai-sdk/hume` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/hume/v/3.0.0-beta.21) |
| `@ai-sdk/klingai` | [`4.0.0-beta.22`](https://www.npmjs.com/package/@ai-sdk/klingai/v/4.0.0-beta.22) |
| `@ai-sdk/langchain` | [`3.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/langchain/v/3.0.0-beta.89) |
| `@ai-sdk/llamaindex` | [`3.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/llamaindex/v/3.0.0-beta.89) |
| `@ai-sdk/lmnt` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/lmnt/v/3.0.0-beta.21) |
| `@ai-sdk/luma` | [`3.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/luma/v/3.0.0-beta.21) |
| `@ai-sdk/mcp` | [`2.0.0-beta.27`](https://www.npmjs.com/package/@ai-sdk/mcp/v/2.0.0-beta.27) |
| `@ai-sdk/mistral` | [`4.0.0-beta.23`](https://www.npmjs.com/package/@ai-sdk/mistral/v/4.0.0-beta.23) |
| `@ai-sdk/moonshotai` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/moonshotai/v/3.0.0-beta.25) |
| `@ai-sdk/open-responses` | [`2.0.0-beta.23`](https://www.npmjs.com/package/@ai-sdk/open-responses/v/2.0.0-beta.23) |
| `@ai-sdk/openai` | [`4.0.0-beta.31`](https://www.npmjs.com/package/@ai-sdk/openai/v/4.0.0-beta.31) |
| `@ai-sdk/openai-compatible` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/openai-compatible/v/3.0.0-beta.25) |
| `@ai-sdk/otel` | [`1.0.0-beta.35`](https://www.npmjs.com/package/@ai-sdk/otel/v/1.0.0-beta.35) |
| `@ai-sdk/perplexity` | [`4.0.0-beta.24`](https://www.npmjs.com/package/@ai-sdk/perplexity/v/4.0.0-beta.24) |
| `@ai-sdk/prodia` | [`2.0.0-beta.24`](https://www.npmjs.com/package/@ai-sdk/prodia/v/2.0.0-beta.24) |
| `@ai-sdk/provider` | [`4.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/provider/v/4.0.0-beta.12) |
| `@ai-sdk/provider-utils` | [`5.0.0-beta.20`](https://www.npmjs.com/package/@ai-sdk/provider-utils/v/5.0.0-beta.20) |
| `@ai-sdk/react` | [`4.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/react/v/4.0.0-beta.89) |
| `@ai-sdk/replicate` | [`3.0.0-beta.22`](https://www.npmjs.com/package/@ai-sdk/replicate/v/3.0.0-beta.22) |
| `@ai-sdk/revai` | [`3.0.0-beta.22`](https://www.npmjs.com/package/@ai-sdk/revai/v/3.0.0-beta.22) |
| `@ai-sdk/rsc` | [`3.0.0-beta.90`](https://www.npmjs.com/package/@ai-sdk/rsc/v/3.0.0-beta.90) |
| `@ai-sdk/svelte` | [`5.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/svelte/v/5.0.0-beta.89) |
| `@ai-sdk/togetherai` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/togetherai/v/3.0.0-beta.25) |
| `@ai-sdk/valibot` | [`3.0.0-beta.20`](https://www.npmjs.com/package/@ai-sdk/valibot/v/3.0.0-beta.20) |
| `@ai-sdk/vercel` | [`3.0.0-beta.25`](https://www.npmjs.com/package/@ai-sdk/vercel/v/3.0.0-beta.25) |
| `@ai-sdk/vue` | [`4.0.0-beta.89`](https://www.npmjs.com/package/@ai-sdk/vue/v/4.0.0-beta.89) |
| `@ai-sdk/workflow` | [`1.0.0-beta.0`](https://www.npmjs.com/package/@ai-sdk/workflow/v/1.0.0-beta.0) |
| `@ai-sdk/xai` | [`4.0.0-beta.36`](https://www.npmjs.com/package/@ai-sdk/xai/v/4.0.0-beta.36) |

PR #7536, opened 2025-07-25
feat (ui): add sendAutomaticallyWhen to Chat

lgrammel:
## Background

#7511 removed automatic resubmit from `Chat`. However, it complicated the code needed on the client side and led to flickering due to status changes.

## Summary

- remove `canAssistantMessageBeSubmitted` helper from `Chat`
- add `sendAutomaticallyWhen` to `Chat`
- add `lastAssistantMessageIsCompleteWithToolCalls` helper

## Verification

- [x] Manually executed updated use chat with tool calls 

## Tasks

- [x] add resubmit test for chat
- [x] implement
- [x] update example
- [x] changeset
- [x] fix resubmit issue

## Related Issues

Follow up on #7511

PR #12171, opened 2026-01-30
fix(ui): respect `Promise<false>` when returned by `sendAutomaticallyWhen`

felixarntz:
## Background

The `sendAutomaticallyWhen` option in the UI chat API supports returning both boolean values and `Promise<boolean>` to control automatic message submission. However, since its introduction in PR #7536, the promise-based implementation has been broken when the function returns `Promise<false>`. The code was calling the function but not awaiting the promise result, causing the automatic send to always trigger regardless of the resolved value - which only works with `Promise<true>`.

## Summary

This PR fixes the handling of `sendAutomaticallyWhen` when it returns a `Promise<false>` via a new private method `shouldSendAutomatically()` that properly awaits promise results before deciding whether to send.

Key changes:
- Added `shouldSendAutomatically()` helper method that checks if the result is a promise and awaits it
- Updated all three call sites (`addMessage`, `addToolOutput`, and end of `makeRequest`) to use the new helper
- Added test coverage for the promise-based scenario

## Manual Verification

N/A

## Checklist

## Related Issues

N/A

### DIFF
--- a/.changeset/orange-eyes-nail.md
+++ b/.changeset/orange-eyes-nail.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ui): respect `Promise<false>` when returned by `sendAutomaticallyWhen`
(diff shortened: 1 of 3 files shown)


gr2m (approved):
great catch!

PR #12209, opened 2026-02-02
fix(ai): auto-populate `originalMessages` in `createAgentUIStream`

### LINKED ISSUE #12139: Streams with tool approval break if `onFinish`-callback is set
### Description

I am using Nuxt and have this small server route:
```ts
import { getModel } from "#server/utils/getModel";
import { createAgentUIStreamResponse, tool, ToolLoopAgent } from "ai";
import z from "zod";

export default defineEventHandler(async (event) => {
  const body = await readBody(event);

  const model = getModel();

    model,
    instructions: `You are a helpful assistant that can provide weather information using the Weather Tool.`,
    tools: {
      weatherTool: tool({
        title: "Weather Tool",
        description: "Get the current weather for a given location.",
        needsApproval: true, // Tool requires user approval before execution
        type: "dynamic",
        inputSchema: z.object({
          location: z.string().describe("The location to get the weather for."),
        }),
        execute: async ({ location }) => {
          console.log(`Fetching weather for ${location}...`);
          return `The current weather in ${location} is sunny with a temperature of 25°C.`;
        },
      }),
    },
  });

  return createAgentUIStreamResponse({
    agent,
    uiMessages: body.messages,
    // As soon as this is callback is provided, the stream breaks after a tool approval (approve or reject)
    onFinish: () => {
      console.log("Finished!");
    },
  });
});

```

When a tool requires approval and the `createAgentUIStreamResponse` function has the `onFinish`-callback set the stream just breaks and no further information is sent to the client.

If you just remove the `onFinish` callback all works fine.

### AI SDK Version

- ai: 6.0.61
- @ai-sdk/vue: 3.0.61

### Code of Conduct

felixarntz:
## Background

When using `createAgentUIStreamResponse` with the `onFinish` callback, streams would fail when handling tool approval scenarios. The issue occurred because when `originalMessages` was not provided, the internal state started empty, preventing proper message state reconstruction needed for tool continuation requests after approval/rejection.

The user discovered that explicitly passing `originalMessages: body.messages` worked around the issue, but this should have been the default behavior when not provided, especially since `onFinish` doesn't work without it.

## Summary

This PR fixes the bug by auto-populating `originalMessages` from the validated `uiMessages` in `createAgentUIStream` when not explicitly provided. This ensures proper state reconstruction for callbacks like `onFinish` while maintaining backward compatibility.

Changes:
- Auto-populate `originalMessages` with `validatedMessages` as the default in `createAgentUIStream`
- Added JSDoc comments explaining the purpose and default behavior of `originalMessages` parameter
  - Docs on this parameter were previously missing for `createAgentUIStream` and `createAgentUIStreamResponse`
- Added new test case verifying `onFinish` works without explicitly providing `originalMessages`
- Updated test snapshot to reflect correct behavior
  - This change should be reasonable because it's a tool call continuation, so the message ID must match the one of the last message

## Manual Verification

I verified this with the reproduction case from the issue:
1. Created a test with `ToolLoopAgent` using a tool that requires approval
2. Called `createAgentUIStreamResponse` with `onFinish` callback but without `originalMessages`
3. Verified the stream completes successfully and `onFinish` is called with proper message state
4. Confirmed no errors occur during tool continuation

The new test case demonstrates that the stream now works correctly without requiring users to manually pass `originalMessages`.

## Checklist

## Related Issues

Fixes #12139

### DIFF
--- a/packages/ai/src/agent/create-agent-ui-stream.ts
+++ b/packages/ai/src/agent/create-agent-ui-stream.ts
@@ -15,9 +15,6 @@ import { ToolLoopAgentOnStepFinishCallback } from './tool-loop-agent-on-step-fin
  *
  * @param agent - The agent to run.
  * @param uiMessages - The input UI messages.
- * @param originalMessages - The original messages for state reconstruction.
- *   Defaults to the validated uiMessages. Used for message ID management
- *   and to provide proper context for callbacks like onFinish.
  * @param abortSignal - The abort signal. Optional.
  * @param timeout - Timeout in milliseconds. Optional.
  * @param options - The options for the agent.
--- a/packages/ai/src/agent/create-agent-ui-stream.ts
+++ b/packages/ai/src/agent/create-agent-ui-stream.ts
@@ -47,6 +47,7 @@ export async function createAgentUIStream<
     | StreamTextTransform<TOOLS>
     | Array<StreamTextTransform<TOOLS>>;
   onStepFinish?: ToolLoopAgentOnStepFinishCallback<TOOLS>;
+  // TODO `originalMessages` is part of this for bc, omit in v7
 } & UIMessageStreamOptions<
   UIMessage<MESSAGE_METADATA, never, InferUITools<TOOLS>>
 >): Promise<
@@ -76,6 +77,7 @@ export async function createAgentUIStream<
 
   return result.toUIMessageStream({
     ...uiMessageStreamOptions,
+    // TODO reading `originalMessages` is here for bc, always use `validatedMessages` in v7
     originalMessages:
       uiMessageStreamOptions.originalMessages ?? validatedMessages,
   });
--- a/packages/ai/src/agent/create-agent-ui-stream.ts
+++ b/packages/ai/src/agent/create-agent-ui-stream.ts
@@ -15,6 +15,9 @@ import { ToolLoopAgentOnStepFinishCallback } from './tool-loop-agent-on-step-fin
  *
  * @param agent - The agent to run.
  * @param uiMessages - The input UI messages.
+ * @param originalMessages - The original messages for state reconstruction.
+ *   Defaults to the validated uiMessages. Used for message ID management
+ *   and to provide proper context for callbacks like onFinish.
  * @param abortSignal - The abort signal. Optional.
  * @param timeout - Timeout in milliseconds. Optional.
  * @param options - The options for the agent.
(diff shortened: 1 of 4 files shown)


packages/ai/src/agent/create-agent-ui-stream.ts
```diff
@@ -74,5 +77,9 @@ export async function createAgentUIStream<
     onStepFinish,
   });
 
-  return result.toUIMessageStream(uiMessageStreamOptions);
+  return result.toUIMessageStream({
+    ...uiMessageStreamOptions,
+    originalMessages:
+      uiMessageStreamOptions.originalMessages ?? validatedMessages,
+  });
```
felixarntz, line 83: If we wanted to make this change more cautiously, we could choose to only auto-populate if `onFinish` is also provided - that would drastically reduce the surface area of the change.

It doesn't make too much sense to me to limit it to only `onFinish`, but if there are concerns, we could start with that. That said, the tests and the effective changes suggest to me that changing this should be fine.

lgrammel (changes requested):
`uiMessages` are the original messages. This should not be needed.

packages/ai/src/agent/create-agent-ui-stream.ts
```diff
@@ -15,6 +15,9 @@ import { ToolLoopAgentOnStepFinishCallback } from './tool-loop-agent-on-step-fin
  *
  * @param agent - The agent to run.
  * @param uiMessages - The input UI messages.
+ * @param originalMessages - The original messages for state reconstruction.
```
lgrammel: why? `uiMessages` are just this

lgrammel:
If originalMessages need to be passed through, it would be better to automatically derive them from uiMessages

packages/ai/src/agent/create-agent-ui-stream.ts
felixarntz: It's already possible to pass this, I'm merely documenting it here.

I agree it doesn't make sense, but there's a bug where `uiMessages` isn't passed through. See https://github.com/vercel/ai/issues/12139#issuecomment-3822725315.

Passing through would be the proper fix, and it's basically what this does by default. But I was concerned about backward compatibility - what if someone passes something here that isn't the same as `uiMessages`? I don't know why someone would do that, but it's possible, and therefore removing the ability could be considered a breaking change.

If you think it's safe, we should just remove the ability to pass `originalMessages` to this and instead pass through `uiMessages`. But went with this more conservative solution first.

lgrammel: `uiMessages` are `originalMessages`. I not think we should allow both as inputs. The better solution is to map uiMessages to originalMessages when passing down the data.

lgrammel (approved):

PR #12211, opened 2026-02-02
fix(anthropic, ai): preserve error_code through UI message flow for provider tools

codewithkenzo:
Fixes #12203

When `web_fetch` or other provider tools return errors like `url_not_allowed`, the error code was getting lost in UI state. On the next message, the SDK fell back to `'unknown'` which Anthropic rejects with a 400.

Two fixes here:

1. **Preserve the error code** - Added `errorCode` field through the UI message flow so it survives round-trips
2. **Valid fallback** - Changed fallback from `'unknown'` to `'unavailable'` (a valid Anthropic error code)

The flow now: provider returns error → extract errorCode → store in UI state → pass to createToolModelOutput → Anthropic converter gets the original code back.

Tests pass, let me know if anything needs tweaking!

### DIFF
--- a/.changeset/fix-provider-tool-error-code.md
+++ b/.changeset/fix-provider-tool-error-code.md
@@ -1,0 +1,27 @@
+---
+'@ai-sdk/anthropic': patch
+'ai': patch
+---
+
+fix(anthropic, ai): preserve error_code through UI message flow for provider tools
+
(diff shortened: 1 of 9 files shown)


aayush-kapoor (commented):
will be closing this in favor of PR https://github.com/vercel/ai/pull/13256

dont think there's a need for introducing a new field. but thanks for the contributing!

PR #12314, opened 2026-02-06
fix(gateway): safely handle non-array content in tool results

### LINKED ISSUE #12178: Bug: AI Gateway perplexitySearch tool loop fails - result is undefined, loop doesn't continue
## Summary

When using `gateway.tools.perplexitySearch()` with `maxSteps > 1`, the tool loop fails to continue after the tool execution. The search is executed successfully by the Gateway, but the results are not properly mapped in the SDK, causing the loop to stop prematurely.

## Environment

- AI SDK: 6.0.64
- @ai-sdk/gateway: 3.0.29
- Node.js: v22.18.0

## Minimal Reproducible Example

```typescript
import { config } from 'dotenv';
config({ path: '.env.local' });

async function main() {
  const result = streamText({
    model: gateway('openai/gpt-4o-mini'),
    prompt: 'What is the latest news about TypeScript 5.7? Search the web.',
    tools: {
      perplexity_search: gateway.tools.perplexitySearch({
        maxResults: 3,
      }),
    },
    maxSteps: 5,
  });

      console.log('[TOOL-CALL]');
      console.log('  args:', part.args);           // undefined - BUG
      console.log('  input:', (part as any).input); // actual data is here
    }

    if (part.type === 'tool-result') {
      console.log('[TOOL-RESULT]');
      console.log('  result:', part.result);        // undefined - BUG
      console.log('  output:', (part as any).output); // actual data is here
    }

    if (part.type === 'finish') {
      console.log('[FINISH] finishReason:', part.finishReason);
    }

    if (part.type === 'text-delta') {
      process.stdout.write(part.textDelta);
    }
  }

  const finalResult = await result;
  console.log('steps.length:', finalResult.steps?.length); // 0 - should be >= 1
  console.log('text.length:', finalResult.text?.length);   // 0 - no answer
}

main().catch(console.error);
```

## Expected Behavior

1. Model decides to call `perplexity_search` tool
2. Gateway executes the search and returns results
3. SDK passes results back to the model
4. Model generates a final answer using the search results
5. `finishReason` should be `"stop"`

## Actual Behavior

1. Model calls `perplexity_search` tool ✓
2. Gateway executes search and returns results ✓
3. **BUG**: Tool result has `result: undefined` (data is in `output` field instead)
4. **BUG**: Tool call has `args: undefined` (data is in `input` field instead)
5. **BUG**: Tool loop stops with `finishReason: "tool-calls"` instead of continuing
6. **BUG**: `steps.length` is 0 (should be >= 1)
7. No final text is generated

## Output

```
[TOOL-CALL]
  args: undefined
  input: {
    query: 'TypeScript 5.7 news',
    max_results: 5,
    country: 'US',
    search_language_filter: [ 'en' ],
    search_recency_filter: 'month'
  }

[TOOL-RESULT]
  result: undefined
  output: { id: '...', results: [Array(3)] }  // HAS DATA!

[FINISH] finishReason: tool-calls

steps.length: 0
text.length: 0
```

## Root Cause Analysis

The stream events show that data IS returned correctly from the Gateway, but the SDK is not properly mapping the response:

- `part.args` is `undefined`, but actual data is in `part.input`
- `part.result` is `undefined`, but actual data is in `part.output`

This property name mismatch causes the tool loop logic to think there are no results, so it stops instead of continuing to the next step where the model would generate a final answer.

## Workaround

None currently. The tool loop cannot be used with `perplexitySearch` or likely any other gateway provider-defined tool.

Ashutosh0x:
This PR fixes a bug where responseBody.content was being mapped without checking if it was an array, causing TypeErrors. It adds an Array.isArray check. Fixes #12178


Ashutosh0x:
Hey @tea-artist, I've fixed the TypeError by adding an Array.isArray check for 
esponseBody.content in doGenerate before mapping. This ensures it handles non-array content safely.

aayush-kapoor (changes requested):
not the right fix for the original issue - marking as closed.

refer the other PR linked to the original issue

PR #12344, opened 2026-02-07
fix(ui): use local activeResponse in finally block to prevent TypeError

kaigritun:
## Summary

This PR fixes a bug where `this.activeResponse` could be undefined in the finally block of `makeRequest`, causing a TypeError when calling `onFinish`.

## Problem

The issue occurred when:
1. An error was thrown before `this.activeResponse` was assigned
2. A concurrent `makeRequest` call overwrote `this.activeResponse`

This caused:
```
```

## Solution

The fix uses a local variable to capture `activeResponse` at the start of the try block:
1. Declare `let activeResponse` before the try block
2. Assign to the local var in the try block
3. Use the local var with null check in the finally block

This ensures the reference is stable and available even if `this.activeResponse` is modified by concurrent operations.

## Credit

Thanks to @codybrouwers for identifying the root cause and providing the patch in the issue comments.

Fixes #8477

### DIFF
--- a/.changeset/fix-activeresponse-finally.md
+++ b/.changeset/fix-activeresponse-finally.md
@@ -1,0 +1,14 @@
+---
+'ai': patch
+---
+
+Fix TypeError in onFinish when activeResponse is undefined
+
+Fixed a bug where `this.activeResponse` could be undefined in the finally block of `makeRequest`, causing a TypeError when calling `onFinish`. This happened when:
+
+1. An error was thrown before `this.activeResponse` was assigned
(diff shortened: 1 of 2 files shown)


MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

OmriGanor:
Any news regarding this?

n1ckoates:
Superseded by https://github.com/vercel/ai/pull/16620

PR #12374, opened 2026-02-09
chore(provider/cohere): migrate tests to fixture-based pattern

dancer:
## background

issue #12270 calls for normalizing provider tests to use recorded API fixtures instead of inline-constructed responses. cohere's `prepareJsonResponse` and `prepareStreamResponse` helpers were called out as the old pattern to replace

## summary

- replace `prepareJsonResponse` / `prepareStreamResponse` with `prepareJsonFixtureResponse` / `prepareChunksFixtureResponse` that read from `__fixtures__/` files
- add 10 fixture files covering text, tool call, null args, reasoning, citations, embedding, and streaming variants
- reorganize tests into `describe` blocks grouped by scenario with `beforeEach` fixture loading
- use `toMatchSnapshot()` for response parsing, keep `toMatchInlineSnapshot()` for request body checks

## verification

- `pnpm test` passes in cohere
- zero `prepareJsonResponse` or `prepareStreamResponse` remaining in cohere

## checklist

- [ ] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)

## related issues

ref #12270

### DIFF
--- a/.changeset/dull-pillows-shop.md
+++ b/.changeset/dull-pillows-shop.md
@@ -1,5 +1,0 @@
----
-'@ai-sdk/cohere': patch
----
-
-fix(provider/cohere): migrate tests to fixture-based pattern
--- a/.changeset/dull-pillows-shop.md
+++ b/.changeset/dull-pillows-shop.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/cohere': patch
+---
+
+fix(provider/cohere): migrate tests to fixture-based pattern
--- a/packages/cohere/src/cohere-embedding-model.test.ts
+++ b/packages/cohere/src/cohere-embedding-model.test.ts
@@ -38,8 +38,8 @@ describe('doEmbed', () => {
     const { embeddings } = await model.doEmbed({ values: testValues });
 
     expect(embeddings).toStrictEqual([
-      [0.1, 0.2, 0.3, 0.4, 0.5],
-      [0.6, 0.7, 0.8, 0.9, 1.0],
+      [0.03302002, 0.020904541, -0.019744873, -0.0625, 0.04437256],
+      [-0.04660034, 0.00037765503, -0.061157227, -0.08239746, -0.010360718],
     ]);
   });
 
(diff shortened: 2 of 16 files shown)


.changeset/dull-pillows-shop.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/cohere': patch
+---
+
+fix(provider/cohere): migrate tests to fixture-based pattern
```
gr2m: there should be no functional change, hence no changeset is needed

gr2m (approved):
Changes look good! Can't wait to have all remaining tests normalized!

packages/cohere/src/cohere-embedding-model.test.ts
```diff
@@ -20,67 +16,55 @@ const server = createTestServer({
...
+describe('doEmbed', () => {
+  beforeEach(() => {
+    prepareJsonFixtureResponse('cohere-embedding');
+  });
 
+  it('should extract embedding', async () => {
     const { embeddings } = await model.doEmbed({ values: testValues });
 
-    expect(embeddings).toStrictEqual(dummyEmbeddings);
+    expect(embeddings).toStrictEqual([
```
gr2m: replace with `toMatchInlineSnapshot`, for all of the `toStrictEqual` calls

gr2m:
for header and result checks we should replace `toStrictEqual` with `toMatchInlineSnapshot`

e.g

```ts

  it('should extract embedding', async () => {
    const { embeddings } = await model.doEmbed({ values: testValues });

    expect(embeddings).toStrictEqual([
      [0.03302002, 0.020904541, -0.019744873, -0.0625, 0.04437256],
      [-0.04660034, 0.00037765503, -0.061157227, -0.08239746, -0.010360718],
    ]);
  });

  it('should expose the raw response', async () => {
    prepareJsonFixtureResponse('cohere-embedding', {
      'test-header': 'test-value',
    });

    const { response } = await model.doEmbed({ values: testValues });

    expect(response?.headers).toStrictEqual({
      'content-length': '363',
      'content-type': 'application/json',
      'test-header': 'test-value',
    });
    expect(response).toMatchSnapshot();
  });
```

PR #12375, opened 2026-02-09
fix(provider/anthropic): use full raw usage data in streaming responses

felixarntz:
## Background

When streaming with the `anthropic` provider, `response.usage.raw` only contained the 4 explicitly typed fields (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`) instead of the full raw API response. The raw response from Anthropic's API includes additional fields like `service_tier`, `inference_geo`, `cache_creation`, and `server_tool_use` that were being lost. This worked correctly for `generateText` but not `streamText`.

Discovered this while working on #12353: when streaming with `speed=fast`, with this PR enabled you'll see it in the response - without it you weren't even though it was already being used.

## Summary

Fixed the `doStream` method to pass the full `rawUsage` JSONObject (which accumulates all fields from stream chunks) to `convertAnthropicMessagesUsage` instead of the typed `usage` object (which only has the 4 declared properties).

This also fixes a secondary bug in multi-turn streaming (programmatic tool calling): the typed `usage` object was mutated in place across turns, but `raw` held a reference to this same mutable object, causing all turns to show the last turn's token values. The fix resolves this because `rawUsage` is replaced (not mutated) at each `message_start`.

**Changes:**
- Updated `convertAnthropicMessagesUsage` to accept `{ usage, rawUsage? }` object instead of just `usage`, using `rawUsage` for the `raw` field when provided
- Updated both call sites (`doGenerate` and `doStream`) to use the new signature
- Added comprehensive unit tests for the usage conversion function
- Updated test snapshots to reflect the now-correct raw usage data

## Manual Verification

Run this without the PR, and then again with the PR:
```bash
cd examples/ai-functions
pnpm tsx src/generate-text/anthropic.ts
pnpm tsx src/stream-text/anthropic.ts
cd ../..
```

The `generateText` result will always include actual raw data, e.g. `service_tier`, while the `streamText` result will not. With the PR, the `streamText` result will correctly include those additional values.

## Checklist

## Related Issues

N/A


packages/anthropic/src/__snapshots__/anthropic-messages-language-model.test.ts.snap
```diff
@@ -13318,10 +13368,11 @@ exports[`AnthropicMessagesLanguageModel > doStream > programmatic tool calling >
         "total": 725,
       },
       "raw": {
-        "cache_creation_input_tokens": 0,
-        "cache_read_input_tokens": 0,
```
felixarntz, line 13322: These are being removed here and in several other places because they're not actually in the response. Anthropic only returns them if caching was enabled.

They were previously present because of the typed object, which was incorrect.

dancer (approved):

shaper (approved):

PR #12379, opened 2026-02-09
fix(anthropic): send tool_choice none instead of stripping tools

roaminro:
## Summary

- When `toolChoice` is `'none'`, send `tool_choice: { type: "none" }` with the tools array preserved instead of stripping both `tools` and `tool_choice` from the API request
- Add `{ type: 'none' }` to the `AnthropicToolChoice` type union
- Update test to expect tools and toolChoice to be preserved

## Problem

When `toolChoice` is `'none'`, the provider strips **both** `tools` and `tool_choice` from the request:

```ts
case 'none':
  // Anthropic does not support 'none' tool choice, so we remove the tools:
  return { tools: undefined, toolChoice: undefined, toolWarnings, betas };
```

This causes the Anthropic API to return `content: []` (empty response) when the conversation history contains `tool_use`/`tool_result` blocks, because it sees tool blocks in history but no `tools` parameter.

This is a common issue in agentic loops where the last step needs to force a text-only response — the model produces zero text instead of a summary.

## Anthropic API support

The comment "Anthropic does not support 'none' tool choice" is outdated. The Anthropic API officially supports `tool_choice: { type: "none" }` — it's listed in the [tool use pricing table](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#pricing) for every model from Claude 3 through Claude Opus 4.6:

> | Model | Tool choice | Token count |
> |---|---|---|
> | Claude Opus 4.6 | `auto`, **`none`** | 346 tokens |

The docs also note: _"If no tools are provided, then a tool choice of none uses 0 additional system prompt tokens."_

Fixes #12378

## Test plan

- [x] Updated existing `should handle tool choice "none"` test to verify tools and toolChoice are preserved
- [x] All 42 existing `prepareTools` tests pass


lgrammel:
Thank you for the pull request. Issue #12378 was closed as not planned. Closing this pull request.

PR #12382, opened 2026-02-09
feat: differentiate text vs image input tokens

R-Taneja:
## Background
OpenAI's image API returns `input_tokens_details` with separate `image_tokens` and `text_tokens` counts, but the AI SDK was parsing this data and then dropping it. Gateway needs this data to charge correctly.

## Summary
- Include `imageTokens` and `textTokens` from OpenAI's `input_tokens_details` in each entry of the `providerMetadata.openai.images` array (both generations and edits paths)
- Token counts are distributed evenly across images (with remainder on the last entry), so that Gateway can sum across all entries to get the exact total

## Manual Verification
Ran an example script and got detailed usage back

## Checklist

### DIFF
--- a/.changeset/three-walls-begin.md
+++ b/.changeset/three-walls-begin.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/provider': patch
+'@ai-sdk/openai': patch
+'ai': patch
+---
+
+feat: differentiate text vs image input tokens
--- a/.changeset/three-walls-begin.md
+++ b/.changeset/three-walls-begin.md
@@ -1,7 +1,5 @@
 ---
-'@ai-sdk/provider': patch
 '@ai-sdk/openai': patch
-'ai': patch
 ---
 
 feat: differentiate text vs image input tokens
--- a/packages/ai/src/generate-image/generate-image.ts
+++ b/packages/ai/src/generate-image/generate-image.ts
@@ -245,30 +245,6 @@ export async function generateImage({
           providerMetadata[providerName].images.push(
             ...result.providerMetadata[providerName].images,
           );
-
-          // Accumulate token details across calls
-          const incoming = metadata as Record<string, unknown>;
-          if (incoming.inputTokensDetails != null) {
-            const incomingDetails = incoming.inputTokensDetails as {
-              imageTokens?: number;
-              textTokens?: number;
-            };
-            const existing = (
-              providerMetadata[providerName] as Record<string, unknown>
-            ).inputTokensDetails as
-              | { imageTokens?: number; textTokens?: number }
-              | undefined;
-
-            (
-              providerMetadata[providerName] as Record<string, unknown>
-            ).inputTokensDetails = {
-              imageTokens:
-                (existing?.imageTokens ?? 0) +
-                (incomingDetails.imageTokens ?? 0),
-              textTokens:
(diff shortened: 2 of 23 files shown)


shaper:
> ## Summary
> * Add optional `inputTokensDetails: { imageTokens, textTokens }` to `ImageModelV3Usage` in @ai-sdk/provider

Please update PR/commit description to reflect edits

packages/ai/src/generate-image/generate-image.ts
```diff
@@ -245,6 +245,30 @@ export async function generateImage({
           providerMetadata[providerName].images.push(
             ...result.providerMetadata[providerName].images,
           );
+
+          // Accumulate token details across calls
+          const incoming = metadata as Record<string, unknown>;
```
shaper: @gr2m for video generation I have learned from this challenge by including the provider metadata for all responses. I wonder whether we can start doing the same here instead of continuing to tack on more custom specific accumulations. It may be a larger change though, for us to consider.

See for example https://github.com/vercel/ai/blob/main/packages/ai/src/generate-video/generate-video-result.ts#L30

There is also more discussion of the issue here: https://github.com/vercel/ai/pull/10716

.changeset/three-walls-begin.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ai-sdk/provider': patch
```
shaper: no longer needed

packages/ai/src/generate-image/generate-image.ts
shaper, line 1: is there a way not to need to modify `ai`? see my comment below.

packages/ai/src/generate-image/generate-image.ts
R-Taneja: updated as per: https://vercel.slack.com/archives/C099VQB75TP/p1770675194513409?thread_ts=1770666332.692919&cid=C099VQB75TP

packages/openai/src/image/openai-image-model.ts
```diff
@@ -199,13 +204,52 @@ export class OpenAIImageModel implements ImageModelV3 {
...
+            ),
           })),
         },
       },
     };
   }
 }
 
+/**
+ * Distributes input token details evenly across images, with the remainder
```
shaper, line 220: why do we need to distribute them across all images, vs. just having the actual precise number for each image in its respective index in the `images` array?

R-Taneja: openai's api returns one `input_tokens_details` per API call, not per image (even if `n` > 1, so the API call generates >1 image)

`input_tokens_details` represents the total number of tokens across all images generated in that API call, and since the same prompt/input image are used, it scales linearly

example:

openAI API call with n = 1: `input_tokens_details.text_tokens = 13`
openAI API call with n = 3, same prompt: `input_tokens_details.text_tokens = 39`

so we can either put the overall `input_tokens_details` object in the `images[0]` object, and have Gateway read from there, but that seemed odd to me

so hence the distribution function - dividing by `n` gives the actual per-image token count, so I spread the total returned by openAI over each image in the `images` array

shaper: TIL, thank you 🙌

shaper (approved):

ghost:
⚠️ Backport to release-v5.0 created but has conflicts: https://github.com/vercel/ai/pull/12578

PR #12324, opened 2026-02-06
docs: add Cencori community provider

bolaabanjo:
## Summary

This PR adds documentation for the Cencori community provider.

[Cencori](https://cencori.com) is a unified infrastructure layer for AI applications that provides:
- Multi-provider gateway (OpenAI, Anthropic, Google, Mistral, DeepSeek, xAI, and more)
- Built-in security (PII detection, prompt injection protection, content filtering)
- Full observability (audit logs, analytics, cost tracking)
- Transparent pricing with credits-based billing

The `cencori` NPM package includes a Vercel AI SDK compatible provider at `cencori/vercel`.

## Changes
- Added `content/providers/03-community-providers/49-cencori.mdx`

## Links
- [Cencori Website](https://cencori.com)
- [NPM Package](https://www.npmjs.com/package/cencori)
- [Documentation](https://cencori.com/docs)

PR #12388, opened 2026-02-10
feat(ci): run prettier fix when the check fails

aayush-kapoor:
## Background

manually telling people to run prettier-fix increases the turnaround time to how quickly we respond to community prs

eg: https://github.com/vercel/ai/pull/12324#pullrequestreview-3776037970

## Summary

new github workflow that triggers when the prettier check fails in CI and then:
- runs `pnpm prettier-fix` and commits the changes if the repo is `vercel/ai`
- comments to tell the contributor to run `pnpm prettier-fix` when the pr is on a fork

### DIFF
--- a/.github/workflows/prettier-autofix.yml
+++ b/.github/workflows/prettier-autofix.yml
@@ -1,0 +1,134 @@
+name: Prettier Autofix
+
+on:
+  check_run:
+    types: [completed]
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  autofix:
+    name: Autofix Prettier
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gr2m (changes requested):
we already have https://github.com/vercel/ai/blob/main/.github/workflows/prettier-on-automerge.yml. The problem with auto-fixing PRs on push is that a contributor might still work on it, and reformatting will cause conflicts. Running prettier-fix right at the end before merge is the right time to do it.

aayush-kapoor:
i see, i thought we'd benefit from adding this in case we didn't want to enable auto merge but makes sense re it causing conflicts for users..

Issue #11413, opened 2025-12-23
BUG V6: Gemini 3 flash throw error when calling function tool (see desc for details)


nabilfatih:
### Description

The error:

Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:getContent` , position 2. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.

this happen when gemini 3 flash call tool, after gemini finish call tool, the error appear, and then chat stop because of this error

what im using: ai sdk v6 + vercel ai gateway

### AI SDK Version

"@ai-sdk/gateway": "^3.0.2",
    "@ai-sdk/google": "^3.0.1",
    "@ai-sdk/google-vertex": "^4.0.1",
    "@ai-sdk/openai": "^3.0.1",
    "@ai-sdk/react": "^3.0.3",
    "ai": "^6.0.3",

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ekansh-tdc:
+1

ekansh-tdc:
Happening with gemini 3 flash and pro models on latest ai sdk

mertbozkir:
Same as well!

Marc-Roig:
Same, it doesnt seem deterministic and getting it randomly. I am also using the ai gateway

Marc-Roig:
After looking at the provider response I realized:

```
"gateway": {
              "routing": {
                "originalModelId": "google/gemini-3-pro-preview",
                "resolvedProvider": "vertex",
                ...
                "attempts": [
                  {
                    "provider": "vertex",
                    "internalModelId": "vertex:gemini-3-pro-preview",
                    "providerApiModelId": "gemini-3-pro-preview",
                    "credentialType": "system",
                    "success": false,
                    "error": "Invalid JSON response",
                    "startTime": 441799.544818,
                    "endTime": 443772.652595,
                    "statusCode": 200
                  },
                  {
                    "provider": "google",
                    "internalModelId": "google:gemini-3-pro-preview",
                    "providerApiModelId": "gemini-3-pro-preview",
                    "credentialType": "system",
                    "success": false,
                    "error": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead
[...]

Issue #11689, opened 2026-01-09
smoothStream strips providerMetadata from reasoning-delta chunks, breaking extended thinking multi-step


apeatling:
When using `streamText` with Anthropic's extended thinking feature and `smoothStream` transform, the thinking signature is lost between steps, causing the API to reject subsequent requests with:

```
messages.1.content.0.type: Expected 'thinking' or 'redacted_thinking', but found 'tool_use'
```

### Root Cause

The `smoothStream` transform in `packages/ai/src/generate-text/smooth-stream.ts` does not preserve `providerMetadata` when processing `reasoning-delta` chunks.

Anthropic sends the thinking signature as a `signature_delta` event, which the `@ai-sdk/anthropic` provider correctly converts to a `reasoning-delta` chunk:

```typescript
{
  type: 'reasoning-delta',
  id: '0',
  delta: '',  // empty text
  providerMetadata: {
    anthropic: {
      signature: 'EtICCkYICxgCKkCs2rGvlzbJBMjJdfgFdTPNfWPhKh5...'
    }
  }
}
```

However, `smoothStream` only preserves `type`, `text`, and `id` when enqueueing chunks:

**Line 115 (flushBuffer):**
```typescript
controller.enqueue({ type, text: buffer, id });  // ❌ No providerMetadata
```

**Line 141 (transform):**
```typescript
controller.enqueue({ type, text: match, id });  // ❌ No providerMetadata
```

This causes the signature to be lost, which means the thinking block cannot be reconstructed for subsequent API calls.

### Steps to Reproduce

1. Create a `streamText` call with Anthropic model using extended thinking:
```typescript
const result = streamText({
  model: anthropic('claude-opus-4-5-20251101'),
  providerOptions: {
    anthropic: {
      thinking: { type: 'enabled', budgetTokens: 10000 }
    }
  },
  tools: { /* some tools */ },
  maxSteps: 5,
  experimental_transform: smoothStream({ delayInMs: 20, chunking: 'word' }),
  messages: [{ role: 'user', content: 'Do something that requires tool use' }]
});
```

2. The first step completes with thinking + tool calls
3. Step 2 fails with the error about missing thinking block

### Expected Behavior

`smoothStream` should preserve `providerMetadata` on all chunks, especially `reasoning-delta` chunks that carry the Anthropic thinking signature.

### Actual Behavior

`providerMetadata` is stripped from `reasoning-delta` chunks, causing the signature to be lost and breaking multi-step extended thinking.

[...]

PR #12402, opened 2026-02-10
fix(google): handle thoughtSignature on empty-text parts and add fixture-based tests

gr2m:
## background

gemini-3-pro-preview sends `thoughtSignature` on a final streaming chunk with empty text (`"text": ""`). the existing `part.text.length > 0` guard caused the part to be skipped, silently dropping the signature. this also migrates google provider tests to a fixture-based pattern per #12270

## summary

- handle `thoughtSignature` on empty-text parts in both `doGenerate` and `doStream` paths. previously parts with `"text": ""` and a `thoughtSignature` were silently dropped
  - `doStream`: emits a `text-delta` with empty string and the signature metadata
  - `doGenerate`: attaches the `thoughtSignature` to the most recent content item
- replace synthetic reasoning fixtures with real gemini-3-pro-preview api responses
- add `tool-call-gemini3` fixtures for functionCall with thoughtSignature (both generate and stream)

## verification

<details>
<summary>doGenerate reasoning with thoughtSignature</summary>

fixture has text part with `thoughtSignature`, snapshot correctly surfaces it in `providerMetadata.google.thoughtSignature` on the text content item
</details>

<details>
<summary>doStream reasoning with thoughtSignature</summary>

streaming fixture sends empty-text chunk with `thoughtSignature` as final chunk, snapshot shows it emitted as `text-delta` with signature in `providerMetadata`
</details>

<details>
<summary>doGenerate tool call with thoughtSignature</summary>

fixture has `functionCall` part with `thoughtSignature`, snapshot correctly surfaces it on the `tool-call` content item
</details>

<details>
<summary>doStream tool call with thoughtSignature</summary>

streaming fixture has `functionCall` chunk with `thoughtSignature`, snapshot shows it propagated to tool-call events
</details>

## checklist

- [ ] documentation has been added / updated (for bug fixes / features)
- [ ] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)

## related issues

refs #12270
potentially related #11413


packages/google/src/__snapshots__/google-generative-ai-language-model.test.ts.snap
```diff
@@ -0,0 +1,1089 @@
...
+        },
+      },
```
gr2m, line 11: shouldn't the thoughtSignature be in the "reasoning" part, not the "text" part?

packages/google/src/__snapshots__/google-generative-ai-language-model.test.ts.snap
```diff
@@ -0,0 +1,1089 @@
...
+    },
+    "type": "text-delta",
+  },
```
gr2m, line 685: shouldn't thought signature be a reasoning block, e.g. reasoning-end?

packages/google/src/__snapshots__/google-generative-ai-language-model.test.ts.snap
dancer: https://ai.google.dev/gemini-api/docs/thought-signatures

dancer (approved):

gr2m:
will backport manually

Erobrine2811:
Thanks for fixing the Google provider side of this. I found that the same signature is still dropped later by `streamText`, so the fix does not currently survive end-to-end through AI SDK core.

Reproduced against `ai@7.0.22`, `@ai-sdk/google@4.0.12`, and the real `gemini-3-flash-preview` API with `include.rawChunks: true`.

Gemini's final raw SSE chunk contains:

```json
{
  "text": "",
  "thoughtSignature": "<signature>"
}
```

The Google provider correctly converts it to a metadata-bearing empty `text-delta`, as implemented in this PR. But `packages/ai/src/generate-text/stream-text.ts` filters it before the downstream step accumulator sees it:

```ts
case 'text-delta': {
  if (chunk.text.length > 0) {
    controller.enqueue(chunk);
  }
  break;
}
```

Observed final result:

```text
raw provider response: thoughtSignature=<present>
step.content text part: thoughtSignature=undefined
```

Running the same prompt/tool sequence through `@earendil-works/pi-ai` preserves the final text signature.

It looks like the provider-level fixture added here passes, but an end-to-end `streamText` test would catch the second filter. A minimal fix appears to be:

```ts
if (chunk.text.length > 0 || chunk.providerMetadata != null) {
  controller.enqueue(chunk);
}
```

Current `main` still has the length-only condition: https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.ts

This also seems consistent with #11689: metadata-only empty deltas are semantically meaningful and should not be discarded. I can open a focused follow-up issue with the real-provider reproduction if preferred.

gr2m:
created a follow up: #18073. Thanks!

PR #12403, opened 2026-02-10
fix(vertex): add fallback for providerOptions key

### LINKED ISSUE #12351: @ai-sdk/google-vertex drops thoughtSignature in multi-step tool calls due to namespace desync
## Background
In multi-step conversations involving tool use with Gemini 3 Flash Preview, the `@ai-sdk/google-vertex` provider fails during the second step with a 500 Internal Server Error.

_**Reference:** Per [Google Cloud Vertex AI Documentation](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures), Gemini 3 models require the `thought_signature` from the previous step to be passed back. If it is missing during a tool-result turn, the API rejects the request._

**Error Message:** 
```
Gemini Step 2 Error: {"error":{"message":"Unable to submit request because function call `get_weather` in the 2. content block is missing a `thought_signature`. Learn more: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures","type":"server_error","code":"INTERNAL_SERVER_ERROR"}}
```

## Step-by-Step Reproduction & Root Cause
The issue is a "sticky variable" bug in the provider's fallback logic:

### Step 1: The Output (Generation)

When the Gemini 3 model generates a tool call, the Vertex provider captures the `thought_signature` and stores it under the `vertex` namespace in the metadata.

- **State:** `providerMetadata: { vertex: { thoughtSignature: '...' } }`

### Step 2: The Input (Follow-up)

Following standard Google AI conventions, the conversation history is often passed back with the signature under the `google` namespace to ensure cross-compatibility between Google and Vertex AI providers.

- **Problem 1 (The Fallback):** In [google-generative-ai-language-model.ts#L100-L115](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L100-L115), the code detects that `vertex` options are missing and falls back to `google`. However, the variable `providerOptionsName` sticks to `vertex`.

- **Problem 2 (The Disconnect):** Consequently, the conversion logic at [lines 136-139](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L136C45-L139) continues to look for the signature under the `vertex` key because of the sticky variable `providerOptionsName`.

## Code example:

```typescript
import { createVertex } from "@ai-sdk/google-vertex";
import { generateText, tool } from "ai";
import { z } from "zod";

// Config: Replace with your actual project details
const vertex = createVertex({
  project: process.env.GOOGLE_CLOUD_PROJECT,
  location: process.env.GOOGLE_CLOUD_LOCATION || "us-central1",
});
const model = vertex("gemini-3-flash-preview");

async function reproduceIssue() {
  console.log("--- Initializing Reproduction for Namespace Desync ---");

  /**
   * STEP 1: Scenario Simulation
   * We simulate a message history where Step 1 was already completed.
   * Following the standard Google AI provider convention, the thoughtSignature
   * is provided back in the 'google' namespace.
   */
  const messages = [
    { role: "user", content: "What is the weather in London?" },
    {
      role: "assistant",
      content: [
        {
          type: "text",
          text: "",
        },
        {
          type: "tool-call",
          toolCallId: "call_123",
          toolName: "getWeather",
          input: { location: "London" },
          providerOptions: {
            google: { thoughtSignature: step1Signature },
          },
        },
      ],
    },
    {
      role: "tool",
      content: [
        {
          type: "tool-result",
          toolCallId: "call_123",
          toolName: "getWeather",
          output: {
            type: "json",
            value: { temperature: 15, unit: "celsius" },
          },
        },
      ],
    },
  ];

  try {
    /**
     * STEP 2: The Follow-up
     * This execution will trigger convertToGoogleGenerativeAIMessages.
     * * Due to the bug in google-generative-ai-language-model.ts (L100-L115),
     * the providerOptionsName remains 'vertex', failing to find 'google.thoughtSignature'.
     */
      model,
      tools: {
        getWeather: tool({
          description: "Get the weather",
          inputSchema: z.object({ location: z.string() }),
        }),
      },
      messages: messages as any,
    });

    console.log("Response:", result.text);
  } catch (error: any) {
    console.error("❌ Reproduction Error caught:");
    console.error(error.message);
  }
}

reproduceIssue();
```

## Proposed Fix
The `providerOptionsName` should be dynamically updated to `google` if the `vertex` key is missing and the `google` key is present. Alternatively, the extraction logic for `thoughtSignature` should check both namespaces as a fallback to ensure that mandatory Gemini 3 state tokens are never dropped during the conversion to the wire format.

### AI SDK Version

ai:  ^6.0.67
@ai-sdk/google-vertex: ^4.0.37

### Code of Conduct

aayush-kapoor:
## Background

multi step tool calls fail because the `thoughtSignature` gets lost due to a key mismatch in providerOptions. 
if the `thoughtSignature` is passed under the google key, the vertex provider silently drops it. the api then rejects the request and we see the errors as observed in #12351 

## Summary

fallback check to see if `vertex` key doesn't have a `thoughtSignature`, we check if it's passed within the google key

## Manual Verification

verified via the repro code given in the original issue, which is resolved after the fix

## Checklist

## Related Issues

fixes #12351

### DIFF
--- a/packages/google/src/convert-to-google-generative-ai-messages.ts
+++ b/packages/google/src/convert-to-google-generative-ai-messages.ts
@@ -82,7 +82,11 @@ export function convertToGoogleGenerativeAIMessages(
           role: 'model',
           parts: content
             .map(part => {
-              const providerOpts = part.providerOptions?.[providerOptionsName];
+              const providerOpts =
+                part.providerOptions?.[providerOptionsName] ??
+                (providerOptionsName !== 'google'
+                  ? part.providerOptions?.google
+                  : undefined);
               const thoughtSignature =
                 providerOpts?.thoughtSignature != null
                   ? String(providerOpts.thoughtSignature)
--- a/.changeset/clean-books-peel.md
+++ b/.changeset/clean-books-peel.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+fix(vertex): add fallback for providerOptions keyname
--- a/packages/google/src/convert-to-google-generative-ai-messages.test.ts
+++ b/packages/google/src/convert-to-google-generative-ai-messages.test.ts
@@ -85,6 +85,130 @@ describe('thought signatures', () => {
   });
 });
 
+describe('thought signatures with vertex providerOptionsName', () => {
+  it('should resolve thoughtSignature from google namespace when using vertex providerOptionsName', async () => {
+    const result = convertToGoogleGenerativeAIMessages(
+      [
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'text',
+              text: 'Regular text',
+              providerOptions: { google: { thoughtSignature: 'sig1' } },
+            },
+            {
+              type: 'reasoning',
+              text: 'Reasoning text',
+              providerOptions: { google: { thoughtSignature: 'sig2' } },
+            },
+            {
+              type: 'tool-call',
+              toolCallId: 'call1',
+              toolName: 'getWeather',
+              input: { location: 'London' },
+              providerOptions: { google: { thoughtSignature: 'sig3' } },
+            },
+          ],
+        },
+      ],
+      { providerOptionsName: 'vertex' },
+    );
+
+    expect(result).toMatchInlineSnapshot(`
+      {
+        "contents": [
+          {
+            "parts": [
+              {
+                "text": "Regular text",
+                "thoughtSignature": "sig1",
+              },
+              {
+                "text": "Reasoning text",
+                "thought": true,
+                "thoughtSignature": "sig2",
+              },
+              {
+                "functionCall": {
+                  "args": {
+                    "location": "London",
+                  },
+                  "name": "getWeather",
+                },
+                "thoughtSignature": "sig3",
+              },
+            ],
+            "role": "model",
+          },
+        ],
+        "systemInstruction": undefined,
+      }
+    `);
+  });
+
+  it('should prefer vertex namespace over google namespace when both are present', async () => {
+    const result = convertToGoogleGenerativeAIMessages(
+      [
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'tool-call',
+              toolCallId: 'call1',
+              toolName: 'getWeather',
+              input: { location: 'London' },
+              providerOptions: {
+                vertex: { thoughtSignature: 'vertex_sig' },
+                google: { thoughtSignature: 'google_sig' },
+              },
+            },
+          ],
+        },
+      ],
+      { providerOptionsName: 'vertex' },
+    );
+
+    expect(result.contents[0].parts[0]).toEqual({
+      functionCall: {
+        name: 'getWeather',
+        args: { location: 'London' },
+      },
+      thoughtSignature: 'vertex_sig',
+    });
+  });
+
+  it('should resolve thoughtSignature from vertex namespace directly', async () => {
+    const result = convertToGoogleGenerativeAIMessages(
+      [
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'tool-call',
+              toolCallId: 'call1',
+              toolName: 'getWeather',
+              input: { location: 'London' },
+              providerOptions: {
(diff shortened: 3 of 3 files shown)


buibaoanh:
The fix is incomplete as we expect `google` under `providerMetadata`, right now it is `vertex`.

buibaoanh:
From a user perspective:

### Step 1: Tool Call Generation (Output)
**(i) What I put in:** A `generateText` call using `@ai-sdk/google-vertex` with a Gemini 3 thinking model.

**(ii) What I get back:** The reasoning state is returned under `providerMetadata: { vertex: { thoughtSignature: "..." } }`.

**(iii) Expected behavior:** Per [Google Cloud Documentation](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures), the `thoughtSignature` is a native field of the Gemini Part object. It should be returned under the `google` namespace to maintain consistency with the Google AI provider and allow for universal history persistence.

### Step 2: Tool Result Resumption (Input)
**(i) What I put in:** I pass the history back to the Vertex provider. Following the standard convention, the signature is provided under the `google` namespace.

**(ii) What I get back:** A 500 Internal Server Error from Vertex AI.

**(iii) Expected behavior:** The model should resume reasoning.

aayush-kapoor:
i verified the fix for the reproduction code snippet you provided and the pr fixed the error! so assuming the bug only produces in a setup like your repro, the fix should be enough.

we specifically allow keeping vertex as provider option key name since there were previous requests of doing so for keeping a track when using vertex provider.

under the hood, you can rest assure that the needed providerOptions are passed how the endpoint expects them

buibaoanh:
If user set `providerOptions` with key `google`, let's say `thinkingConfig`, right now what I got is `vertex` in `providerMetadata`.

I would expect `google` in `providerMetadata` as stated in [Vertex AI](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures) and for consistency reason.

_I noticed that `providerOptionsName` is selected by hardcoded provider Name for `@ai-sdk/google-vertex` instead of provider key in `providerOptions` in [this logic](https://github.com/vercel/ai/blob/main/packages/google/src/google-generative-ai-language-model.ts#L100-L102)._

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -82,7 +82,11 @@ export function convertToGoogleGenerativeAIMessages(
           role: 'model',
           parts: content
             .map(part => {
-              const providerOpts = part.providerOptions?.[providerOptionsName];
+              const providerOpts =
+                part.providerOptions?.[providerOptionsName] ??
+                (providerOptionsName !== 'google'
+                  ? part.providerOptions?.google
+                  : undefined);
```
lgrammel, line 89: could this be simplified to just `part.providerOptions?.google`? (it would double lookup and return undefined in the negative case but i dont think that is an issue)

lgrammel (approved):

aayush-kapoor:
@buibaoanh the mapping isn't 1-1 if you compare it to google's sdk but that is by design of the AI SDK

there is a reason for why it's mapped to vertex instead of google and you can see [here](https://github.com/vercel/ai/pull/10983) and [here](https://github.com/vercel/ai/issues/7783) why

the ideal dx for setting providerOptions when using vertex provider is that the user defines the key to be `vertex`

packages/google/src/convert-to-google-generative-ai-messages.ts
aayush-kapoor: hmm i tried that and it leads to a regression with some other tests. it would basically ignore any options set under a key which is not `google` (which wouldn't work in the case of vertex provider, where we added [this](https://github.com/vercel/ai/pull/10983) change recently)

buibaoanh:
@aayush-kapoor Ok, what about [google-vertex docs](https://ai-sdk.dev/providers/ai-sdk-providers/google-vertex) refer to `google` as namespace for both `providerOptions` and `providerMetadata`?

aayush-kapoor:
@buibaoanh thanks for pointing out, will update the docs

PR #12432, opened 2026-02-11
fix(anthropic): ensure thinking blocks appear first in assistant messages with tool use

Limitless2023:
## Summary

Fixes #7729

When extended thinking is enabled and assistant messages contain both thinking/redacted_thinking blocks and tool calls, the Anthropic API requires thinking blocks to appear first in the content array. Without this, the API returns:

```
Expected `thinking` or `redacted_thinking`, but found `tool_use`.
When `thinking` is enabled, a final assistant message must start with a thinking block.
```

## Changes

**`packages/anthropic/src/convert-to-anthropic-messages-prompt.ts`**
- Added content reordering logic before pushing assistant messages
- Separates thinking/redacted_thinking blocks from other content types
- Places thinking blocks at the beginning while preserving relative order of other content
- No-op when no thinking blocks are present (zero overhead for non-thinking scenarios)

**`packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts`**
- Added 4 test cases covering:
  - thinking block + tool_use reordering
  - redacted_thinking block + tool_use reordering  
  - No thinking blocks (order preserved)
  - Multiple thinking blocks (all placed first)

## Before/After

```
// Before: content order depends on input order
[text, tool_use, thinking] → API error ❌

// After: thinking blocks always first
[text, tool_use, thinking] → [thinking, text, tool_use] ✅
```

## Testing

All new tests pass. No changes to existing behavior for non-thinking scenarios.

### DIFF
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
@@ -1961,6 +1961,247 @@ describe('assistant messages', () => {
       `);
     });
   });
+
+  describe('thinking blocks with tool use', () => {
+    it('should place thinking block before tool_use in assistant message', async () => {
+      const result = await convertToAnthropicMessagesPrompt({
+        prompt: [
+          {
+            role: 'assistant',
+            content: [
+              {
+                type: 'text',
+                text: 'Let me use a tool.',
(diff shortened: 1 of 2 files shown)


gr2m:
closed via #18489

PR #9481, opened 2025-10-13
feat(ai): `experimental_transcribe({maxDownloadSizeInBytes})`

gr2m:
## Checklist

- [ ] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ ] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)
- [ ] I have reviewed this pull request (self-review)

PR #12445, opened 2026-02-11
fix/unbounded download dos

gr2m:
- Replace unbounded `arrayBuffer()`/`blob()` calls in `download()` and `downloadBlob()` with streaming reads that enforce a **2 GiB default size limit**
- Add `abortSignal` passthrough from callers (`transcribe`, `generateVideo`) to `fetch()`
- Check `Content-Length` header for early rejection before reading body
- Track bytes incrementally via `ReadableStream.getReader()`, abort with `DownloadError` when limit exceeded
- Expose configurable `download` parameter on `transcribe()` and `experimental_generateVideo()` (instead of adding a new `maxDownloadSize` argument) — keeps download config separate from API function signatures
- Export `createDownload({ maxBytes })` factory from `ai` for custom size limits

closes #9481 / addresses https://github.com/vercel/ai/pull/9481#issuecomment-3425101504


lgrammel:
why is the download blob function in `provider-utils` when it is only used by the `ai` package? package churn and api surface could be reduced by moving it into the `ai` pkg

packages/provider-utils/src/index.ts
```diff
@@ -11,6 +11,10 @@ export { convertImageModelFileToDataUri } from './convert-image-model-file-to-da
 export { convertToFormData } from './convert-to-form-data';
 export { downloadBlob } from './download-blob';
 export { DownloadError } from './download-error';
+export {
```
lgrammel, line 14: a) could this be part of the ai pkg instead
b) the difference between file and function name is strange

lgrammel (approved):

gr2m:
> why is the download blob function in `provider-utils` when it is only used by the `ai` package

it's also used by

  - [@ai-sdk/openai](https://github.com/vercel/ai/blob/89202fbe9591eeb4969f2fe628eec23254e845b6/packages/openai/src/image/openai-image-model.ts#L11)
  -[ @ai-sdk/openai-compatible](https://github.com/vercel/ai/blob/78555addd65d1f7886286ca910bf2fc9743dfa98/packages/openai-compatible/src/image/openai-compatible-image-model.ts#L13)
  - [@ai-sdk/deepinfra](https://github.com/vercel/ai/blob/d427284eef1f121d3c7a66476e4d226b0caf86e2/packages/deepinfra/src/deepinfra-image-model.ts#L12)

I'll look into refactoring but not yet sure if it's worth it.

> b) the difference between file and function name is strange

Will fix, thanks

gr2m:
I tried some refactoring, but it would lead to a circular dependency or code duplication to move `readResponseWithSizeLimit` and `DEFAULT_MAX_DOWNLOAD_SIZE` from provider-utils to ai:

https://github.com/vercel/ai/commit/ad95f2a908cd45d2216c6a7ca6ac8a66ac1d858a

PR #12448, opened 2026-02-11
feat(ai): add onStepFinish callback to createUIMessageStream

### LINKED ISSUE #14295: add experimental callbacks to all core functions

aayush-kapoor:
## Background

we expose `onFinish` callback for createUIMessageStream() but not `onStepFInish`, which is inconsistent with the pattern.

also reported in #12383 

## Summary

added the callback to the function + new type added

## Manual Verification

- verified by running `localhost:3000/use-chat-human-in-the-loop`

## Checklist

## Future Work

the callback might need to be added for `createAgentUIStreamResponse`

## Related Issues

fixes #12383

### DIFF
--- a/packages/ai/src/ui-message-stream/ui-message-stream-on-step-finish-callback.ts
+++ b/packages/ai/src/ui-message-stream/ui-message-stream-on-step-finish-callback.ts
@@ -1,0 +1,25 @@
+import { UIMessage } from '../ui/ui-messages';
+
+/**
+ * Callback that is called when a step finishes during streaming.
+ * This is useful for persisting intermediate UI messages during multi-step agent runs.
+ */
+export type UIMessageStreamOnStepFinishCallback<UI_MESSAGE extends UIMessage> =
+  (event: {
+    /**
+     * The updated list of UI messages at the end of this step.
+     */
+    messages: UI_MESSAGE[];
+
+    /**
+     * Indicates whether the response message is a continuation of the last original message,
+     * or if a new message was created.
+     */
+    isContinuation: boolean;
+
+    /**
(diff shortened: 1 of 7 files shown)


packages/ai/src/ui-message-stream/ui-message-stream-on-step-finish-callback.ts
```diff
@@ -0,0 +1,25 @@
+import { UIMessage } from '../ui/ui-messages';
+
+/**
+ * Callback that is called when a step finishes during streaming.
+ * This is useful for persisting intermediate UI messages during multi-step agent runs.
+ */
+export type UIMessageStreamOnStepFinishCallback<UI_MESSAGE extends UIMessage> =
```
felixarntz, line 7: would it make sense to include `isAborted` here as well? if the stream was aborted during the step for example.

felixarntz (approved):
lgtm, one question/suggestion

packages/ai/src/ui-message-stream/ui-message-stream-on-step-finish-callback.ts
aayush-kapoor: hmm i'm not sure since `onStepFinish` is called when we receive a `finish-step` chunk. an abort would prevent the finish-step chunk from ever being emitted. so i'm not sure how it would work or trigger..

now `onFinish` has that event since that callback can be called when the entire stream ends (via flush() or cancel() )

felixarntz: ah good point; good to keep as is then 👍

PR #12451, opened 2026-02-11
Update Jinja2 version to 3.1.5

i5d6:
Jinja is an extensible templating engine. Prior to 3.1.5, An oversight in how the Jinja sandboxed environment detects calls to str.format allows an attacker that controls the content of a template to execute arbitrary Python code. To exploit the vulnerability, an attacker needs to control the content of a template. Whether that is the case depends on the type of application using Jinja. This vulnerability impacts users of applications which execute untrusted templates. Jinja's sandbox does catch calls to str.format and ensures they don't escape the sandbox. However, it's possible to store a reference to a malicious string's format method, then pass that to a filter that calls it. No such filters are built-in to Jinja, but could be present through custom filters in an application. After the fix, such indirect calls are also handled by the sandbox. This vulnerability is fixed in 3.1.5.

### DIFF
--- a/examples/next-fastapi/requirements.txt
+++ b/examples/next-fastapi/requirements.txt
@@ -12,7 +12,7 @@ httpcore==1.0.5
 httptools==0.6.1
 httpx==0.27.0
 idna==3.7
-Jinja2==3.1.4
+Jinja2==3.1.5
 markdown-it-py==3.0.0
 MarkupSafe==2.1.5
 mdurl==0.1.2


gr2m (approved):
Thanks!

Context: https://jinja.palletsprojects.com/en/stable/changes/#version-3-1-5

i5d6:
@gr2m 

I sent a report to Vercel Open Source on HackerOne, but there's no response.

PR #12512, opened 2026-02-12
fix(ai): pass provider executed tool results back in gateway

### LINKED ISSUE #12178: Bug: AI Gateway perplexitySearch tool loop fails - result is undefined, loop doesn't continue

aayush-kapoor:
## Background

stemming from the issue https://github.com/vercel/ai/issues/12178, there were a couple of different bugs identfied, in the order that they were encountered after resolving the previous bug
 
- `providerExecuted` mismatch on tool-calls
  - Gateway tool-calls came from the stream without providerExecuted set
  - The SDK's loop continuation logic filtered these inconsistently
  
- loop continuation logic didn't handle provider-executed results
  - The continuation condition only checked for client-executed tool results
  - When gateway tools returned results, the loop didn't know to continue

- gateway tool results placed in wrong message type
  - Provider-executed tool results went in the assistant message
  - OpenAI's conversion code assumed these were OpenAI-native tools and converted to item_reference
  - but OpenAI doesn't have gateway tool results stored → model couldn't see them → infinite loop

- gateway tool-calls marked as providerExecuted for prompt conversion
  - Tool-calls with providerExecuted: true were either converted to item_reference or skipped entirely
  - Gateway tools don't have OpenAI item IDs, so they were just skipped
  - Model received tool results but no matching tool calls → error

## Summary

- Infer `providerExecuted: true` for provider tools
- add `hasProviderExecutedToolResultsNeedingContinuation` check
- add `isGatewayTool()` helper
- set providerExecuted: undefined for gateway tool-calls
- put gateway tool-results in tool message
 
## Manual Verification

verified that the loop continues for the examples `12178-repro.ts` and `examples/ai-functions/src/stream-text/gateway-tool-perplexity-search.ts`

## Checklist

## Future Work

- check if the same issue exists for generateText() 
- check if this can be resolved on the gateway side

## Related Issues

fixes #12178


examples/ai-functions/src/stream-text/12178-repro.ts
gr2m, line 1: should we commit this? Alternatively you can copy the source code into the pull request body / manual verification section

gr2m (dismissed):
looks good to me but would like @lgrammel to have a look as well

lgrammel (changes requested):
x

packages/ai/src/generate-text/to-response-messages.ts
```diff
@@ -8,6 +8,20 @@ import { createToolModelOutput } from '../prompt/create-tool-model-output';
 import { ContentPart } from './content-part';
 import { ToolSet } from './tool-set';
 
+/**
+ * Checks if a tool is a gateway tool based on its ID.
+ * Gateway tools are provider-executed but their results need to be sent
+ * to the LLM as regular tool results since the LLM provider (e.g., OpenAI)
+ * doesn't have them stored.
+ */
+function isGatewayTool<TOOLS extends ToolSet>(
```
lgrammel, line 17: we should not have such code in the ai package afaik

packages/ai/src/generate-text/to-response-messages.ts
```diff
@@ -69,7 +93,11 @@ export async function toResponseMessages<TOOLS extends ToolSet>({
           toolCallId: part.toolCallId,
           toolName: part.toolName,
           input: part.input,
-          providerExecuted: part.providerExecuted,
+          // Don't mark gateway tools as providerExecuted so they're converted
+          // to function_call by providers (not item_reference or skipped)
+          providerExecuted: isGatewayTool(tools, part.toolName)
```
lgrammel, line 98: this is not good. does gateway implement provider-executed tools in the same way as openai or anthropic?

gr2m:
Closing: Gateway now handles this with a server-side provider tool loop.

PR #12515, opened 2026-02-12
feat(video): externalize polling control and webhook support for generateVideo

gr2m:
- Adds optional `doStart`/`doStatus` methods to the experimental `VideoModelV4` spec, enabling a two-phase async model where the SDK core orchestrates polling or webhook-based completion instead of each provider implementing its own polling loop.
- Adds `poll` and `webhook` parameters to `experimental_generateVideo` for user-controlled polling intervals/timeouts and webhook-based completion.
- Adds optional `handleWebhookOption` to `VideoModelV4` so a provider can signal native webhook support. The SDK only invokes the user's webhook factory when the model implements it; otherwise it falls back to polling.
- Replaces `doGenerate` with `doStart`/`doStatus` in the first-party models implementing `VideoModelV4` (FAL, Alibaba, KlingAI, Replicate, xAI, Google, and Google Vertex). The gateway video model remains a `doGenerate` passthrough.
- Makes `doGenerate` optional on `VideoModelV4`. Third-party providers implementing only `doGenerate` continue to work unchanged, and the SDK falls back to it when no async flow is selected.
- Constrains operation handles to `JSONValue`, making the provider-owned task reference serializable and suitable for future persistence and gateway support.

Towards https://github.com/vercel/ai/issues/12381.

## Architecture

```
User calls:     generateVideo({ model, prompt, poll?, webhook? })
                         ↓
Core function:  Chooses flow based on model capabilities and user options
                         ↓
             ┌───────────┴───────────────┐
             │                           │
      [Legacy / fallback]        [Start/Status flow]
             │                    (poll or webhook provided,
             │                     or doGenerate missing)
             │                           │
    model.doGenerate(options)    model.doStart(options)
      ↓                                  ↓
    Provider owns polling        returns { operation }
                                        ↓
                            ┌───────────┴────────────┐
                            │                        │
                     [native webhook]            [polling]
                            │                        │
                     await notification       SDK core loop
                            │                        │
                     model.doStatus(op)       model.doStatus(op)
                            └───────────┬────────────┘
                                        ↓
                                  return result
```

The SDK core owns the polling lifecycle. Providers only implement single-shot `doStart` (submit) and `doStatus` (check once). The polling interval and timeout are controlled by the user.

## Review decisions

- **`poll` and `webhook` are intentionally composable.** When a model supports native webhooks, `poll.timeoutMs` limits the webhook wait. When it does not, the same polling configuration controls the automatic fallback. This behavior is documented and covered by tests. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2821429102))
- **The first iteration uses fixed polling intervals and timeouts.** Exponential backoff and `onAttempt` were removed to keep the API small; aligning polling with shared retry/backoff primitives can be revisited separately. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2821435402))
- **Operation handles are JSON-serializable.** `operation` is typed as `JSONValue` instead of `unknown`, while remaining opaque to core. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2843221880))
- **Shared cross-modality operation types are deferred.** The current types remain video-specific until another modality implements the same lifecycle, avoiding a premature provider-spec abstraction. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2843151286))
- **Replicate's server-side `wait` remains provider-specific.** It affects how long Replicate holds the submission request open, while top-level `poll` controls the SDK's provider-neutral status lifecycle. It can be added through Replicate provider options independently if needed. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2843203139))

## Verification

- [x] Tests added and updated
- [x] Documentation added and updated
- [x] Patch changeset added
- [x] TypeScript, lint/format, and code consistency checks pass locally

## Future work

- Expose async operations as a first-class API: start an operation, persist its handle, and resume or query it later. ([discussion](https://github.com/vercel/ai/pull/12515#discussion_r2843215786))
- Generalize operation types after another model modality implements the same lifecycle.
- Add AI Gateway transport for the JSON-serializable start/status operation flow.
- Explore a multi-notification webhook receiver if providers need intermediate status events rather than a single completion notification.

## Related issue

https://github.com/vercel/ai/issues/12381


packages/alibaba/src/alibaba-video-model.ts
```diff
@@ -147,8 +147,12 @@ export class AlibabaVideoModel implements Experimental_VideoModelV3 {
   ) {}
 
   async doGenerate(
-    options: Parameters<Experimental_VideoModelV3['doGenerate']>[0],
-  ): Promise<Awaited<ReturnType<Experimental_VideoModelV3['doGenerate']>>> {
+    options: Parameters<
+      NonNullable<Experimental_VideoModelV3['doGenerate']>
+    >[0],
+  ): Promise<
+    Awaited<ReturnType<NonNullable<Experimental_VideoModelV3['doGenerate']>>>
```
gr2m: what a type, I hate it 🤣

examples/ai-functions/src/generate-video/fal/luma-webhook-options.ts
```diff
@@ -0,0 +1,30 @@
...
+      providerOptions: {
+        fal: {
+          resolution: '540p',
+        } satisfies FalVideoModelOptions,
+      },
+      webhook: async () => {
+        const { url, received } = await createSmeeWebhook();
+        console.log(`\nWaiting for webhook via ${url}`);
+        return { url, received };
+      },
```
gr2m, line 25: this works ✌🏼

packages/provider/src/video-model/v3/video-model-v3.ts
```diff
@@ -1,6 +1,9 @@
 import type { VideoModelV3CallOptions } from './video-model-v3-call-options';
 import type { SharedV3ProviderMetadata } from '../../shared/v3/shared-v3-provider-metadata';
 import type { SharedV3Warning } from '../../shared/v3/shared-v3-warning';
+import type { VideoModelV3OperationStartResult } from './video-model-v3-operation-start-result';
+import type { VideoModelV3OperationStatusResult } from './video-model-v3-operation-status-result';
+import type { VideoModelV3OperationWebhook } from './video-model-v3-operation-webhook';
```
gr2m: Added the `Operation` prefix to group the 3. In future these might be `SharedOperation*` types that work across all model types

packages/provider/src/video-model/v3/video-model-v3.ts
```diff
@@ -129,4 +135,80 @@ export type VideoModelV3 = {
...
+   * - **Absent**: The SDK never calls the user's `webhook` factory and falls
+   *   back to polling via `doStatus`. This avoids unnecessary webhook
+   *   endpoint creation for providers whose APIs have no native webhook
+   *   mechanism.
+   *
+   * This method exists because the SDK must decide whether to invoke the
+   * user's webhook factory — which may create real HTTP endpoints or
+   * external resources — *before* calling `doStart`. Without an explicit
+   * capability signal on the model, the SDK would eagerly create a webhook
+   * endpoint for every provider, even those that silently ignore the URL.
```
gr2m: this doesn't cover the potential case that webhooks may be supported depending on model.

The other reason for this method is to normalize the webhook payload once we have a proper spec

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -154,6 +161,57 @@ export async function experimental_generateVideo({
...
+   * start/status flow. When provided and the model implements `doStart`
+   * and `doStatus`, the SDK will use webhooks instead of polling.
+   *
+   * The factory should return a URL for the provider to send notifications to,
+   * and a `received` promise that resolves when the notification arrives.
+   */
+  webhook?: () => PromiseLike<{
+    url: string;
+    received: Promise<VideoModelV3Webhook>;
+  }>;
```
gr2m: Providers might send multiple webhooks for the same operation, that could not be implemented with this API. Instead of a `received` promise we could resolve with a `onStatus` callback or similar. Not as relevant to video but might be relevant for multi-modal async calls

packages/provider/src/video-model/v3/video-model-v3.ts
```diff
@@ -129,4 +135,80 @@ export type VideoModelV3 = {
...
+   */
+  handleWebhookOption?: (options: {
+    webhook: () => PromiseLike<{
+      url: string;
+      received: Promise<VideoModelV3OperationWebhook>;
```
lgrammel: PromiseLike

packages/provider/src/video-model/v3/video-model-v3.ts
```diff
@@ -129,4 +135,80 @@ export type VideoModelV3 = {
...
+    }>;
+  }) => PromiseLike<{
+    webhookUrl: string;
+    received: Promise<VideoModelV3OperationWebhook>;
```
lgrammel: PromiseLike

packages/provider/src/video-model/v3/video-model-v3.ts
```diff
@@ -129,4 +135,80 @@ export type VideoModelV3 = {
...
+   *   endpoint creation for providers whose APIs have no native webhook
+   *   mechanism.
+   *
+   * This method exists because the SDK must decide whether to invoke the
+   * user's webhook factory — which may create real HTTP endpoints or
+   * external resources — *before* calling `doStart`. Without an explicit
+   * capability signal on the model, the SDK would eagerly create a webhook
+   * endpoint for every provider, even those that silently ignore the URL.
+   */
+  handleWebhookOption?: (options: {
```
lgrammel: this breaks the assumption that doGenerate/doStream return JSON stringifiable data, which is important for gateway. Have you tried that this works with gateway?

examples/ai-functions/src/generate-video/fal/luma-webhook-options.ts
lgrammel: i wonder if we can unify webhooks/poll more. both are similar in that they return the received response and could be functions.

e.g. a `receiveData` property (or a better name) that could either be a webhook as above, or a poller config (which then gets handled internally by the provider).

if we do not unify them, how do we prevent that polling and webhooks are specified at the same time? typescript discriminated unions?

content/docs/03-ai-sdk-core/38-video-generation.mdx
```diff
@@ -215,32 +215,32 @@ const { video } = await generateVideo({
...
   prompt: 'A cinematic timelapse of a city from dawn to dusk',
   duration: 10,
-  providerOptions: {
-    fal: {
-      pollTimeoutMs: 600000, // 10 minutes
-    } satisfies FalVideoModelOptions,
+  poll: {
+    intervalMs: 5000, // Check every 5 seconds (default)
+    timeoutMs: 600000, // Timeout after 10 minutes (default)
+    backoff: 'exponential', // Double interval after each poll, capped at 60s
```
lgrammel: ideally we should have similar mechanisms for polling timeouts in our codebase, this is somewhat different from how we do backoffs and retries in the ai library itself and can be confusing

content/docs/03-ai-sdk-core/38-video-generation.mdx
```diff
@@ -215,32 +215,32 @@ const { video } = await generateVideo({
...
+    onAttempt: ({ attempt, elapsedMs }) => {
+      console.log(`Poll attempt ${attempt} (${elapsedMs}ms elapsed)`);
+    },
```
lgrammel: wonder if that is needed in the first iteration

packages/provider/src/video-model/v3/video-model-v3-operation-start-result.ts
```diff
@@ -0,0 +1,43 @@
+import type { SharedV3ProviderMetadata } from '../../shared/v3/shared-v3-provider-metadata';
+import type { SharedV3Warning } from '../../shared/v3/shared-v3-warning';
+
+/**
+ * Result returned by `doStart` when initiating an asynchronous video generation.
+ */
+export type VideoModelV3OperationStartResult = {
```
felixarntz: While I realize this is probably scoped to video because it's still experimental, I think ideally these types become generic across model types (e.g. `ModelV3OperationStartResult`). There's nothing specific to video here, the abstraction to poll for progress on a job isn't really tied to any modalities.

packages/provider/src/video-model/v3/video-model-v3-operation-status-result.ts
```diff
@@ -0,0 +1,89 @@
+import type { SharedV3ProviderMetadata } from '../../shared/v3/shared-v3-provider-metadata';
+import type { SharedV3Warning } from '../../shared/v3/shared-v3-warning';
+import type { VideoModelV3VideoData } from './video-model-v3';
+
+/**
+ * Result returned by `doStatus` when checking the status of an
+ * asynchronous video generation.
+ */
+export type VideoModelV3OperationStatusResult =
```
felixarntz: This could also be based on a generic type, only with a template. E.g.:

```ts
export type VideoModelV3OperationStatusResult = ModelV3OperationStatusResult<VideoModelV3GenerateResult>;
```

I think the only video specific piece here is the `videos` property, and by using a template type we'd be able to reuse definition from the regular `*ModelV3GenerateResult` types (looks like that doesn't exist for video yet, but certainly could be added, it's currently hard-coded).

It would avoid some duplication too.

packages/replicate/src/replicate-video-model.ts
felixarntz, line 197: Related to my Replicate note at the beginning of https://github.com/vercel/ai/issues/12381#issuecomment-3946469523 (plus point 2 in the numbered list there), Replicate waits for up to 60 seconds by default before returning. It supports a `wait` argument to customize that behavior: https://replicate.com/docs/topics/predictions/create-a-prediction#timeout-duration

Is that something we should consider supporting at the top-level API, or is it `providerOptions` material?

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -154,6 +161,57 @@ export async function experimental_generateVideo({
     url: URL;
     abortSignal?: AbortSignal;
   }) => Promise<{ data: Uint8Array; mediaType: string | undefined }>;
+
+  /**
+   * Polling configuration for models that support the asynchronous
+   * start/status flow. When provided and the model implements `doStart`
+   * and `doStatus`, the SDK will orchestrate polling automatically.
+   */
+  poll?: {
```
felixarntz: Regarding my question 4 in https://github.com/vercel/ai/issues/12381#issuecomment-3946469523, how do we feel about the idea of exposing the URL for polling at the top-level API? Especially for _very_ long-running jobs, there may be cases where the caller does not want to poll at this point, but instead store the job ID / where to query for its status and get back to it in a couple hours.

packages/provider/src/video-model/v4/video-model-v4-operation-start-result.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Result returned by `doStart` when initiating an asynchronous video generation.
+ */
+export type VideoModelV3OperationStartResult = {
+  /**
+   * Opaque reference passed to `doStatus` to check the status of the
+   * generation (e.g., a task ID or prediction URL).
+   */
+  operation: unknown;
```
felixarntz: Can we make this more concrete? There are probably only a few relevant patterns here, for which we could use a union type? Or maybe use a generic?

felixarntz (commented):
Looks like a great starting point - leaving a few thoughts here.

Is the idea for this PR to implement it in a way that can still be merged into v6 (i.e. no breaking changes except for video generation)?

packages/provider/src/video-model/v3/video-model-v3.ts
gr2m: Done in 91b0ba9 — `received` now uses `PromiseLike` instead of `Promise`, both on the `handleWebhookOption` spec and the public `webhook` factory in `generateVideo`. (Note: the spec has since been ported from `VideoModelV3` to `VideoModelV4` after main's V3→V4 migration, so this now lives in `video-model-v4.ts`.)

packages/provider/src/video-model/v3/video-model-v3.ts
gr2m: Done in 91b0ba9 — `received` now uses `PromiseLike` instead of `Promise`, both on the `handleWebhookOption` spec and the public `webhook` factory in `generateVideo`. (Note: the spec has since been ported from `VideoModelV3` to `VideoModelV4` after main's V3→V4 migration, so this now lives in `video-model-v4.ts`.)

packages/provider/src/video-model/v3/video-model-v3.ts
gr2m: Good catch — confirming this is a known limitation. The async start/status flow is currently only implemented by direct providers (fal, alibaba, klingai, replicate, xai, google, google-vertex). The **gateway video model remains a `doGenerate`-only passthrough** and does not implement `doStart`/`doStatus`, so `poll`/`webhook` are **not supported through the AI Gateway** in this iteration.

For a future gateway implementation, the opaque `operation` returned by `doStart` would need to be JSON-serializable to round-trip across the gateway boundary (today it's typed `unknown`). I have not validated that yet — so rather than ship a gateway path that silently breaks, I've left it out and documented it as a limitation in the PR description's Future Work section.

content/docs/03-ai-sdk-core/38-video-generation.mdx
gr2m: Agreed — trimmed for the first iteration. Removed both `poll.backoff` (the `'none' | 'exponential'` strategy) and `poll.onAttempt` in 1d66bf5. `poll` now only exposes `intervalMs` and `timeoutMs`; the polling loop uses a fixed interval. Removed across the spec, core, tests, examples, and docs. We can revisit backoff/progress callbacks later (and ideally fold them into the library's existing retry/backoff conventions per your other comment).

gr2m:
@felixarntz re: your question about v6-compatibility —

> Is the idea for this PR to implement it in a way that can still be merged into v6 (i.e. no breaking changes except for video generation)?

This is now targeting **v7**, we won't backport

PR #12384, opened 2026-02-09
feat(anthropic): add the new compaction feature

aayush-kapoor:
## Background

anthropic released a new compaction feature along with their launch of Opus 4.6

https://platform.claude.com/docs/en/build-with-claude/compaction

## Summary

- updated the schema
- updated the usage token calculation since the outer level params don't account for the compaction tokens

## Manual Verification

verified by running the examples: 
- [x] `examples/ai-functions/src/generate-text/anthropic-compaction-pause.ts`
- [x] `examples/ai-functions/src/generate-text/anthropic-compaction.ts`
- [x] `examples/ai-functions/src/stream-text/anthropic-compaction.ts`
- [x] `http://localhost:3000/use-chat-anthropic-compaction`
  - large context is preloaded into the model
  - just send a simple follow up message and you'll notice the compaction block created (also noticed in the console logs)
  - multi-turn conversations work properly

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

fixes #12297

PR #12154, opened 2026-01-30
fix(provider/anthropic): parse context_management from streaming response root level

dancer:
## background

the `contextManagement` field in `providerMetadata` was always `null` when using `streamText` with the Anthropic provider, even though the API correctly returns `context_management` in the response.

the issue was twofold:
1. the zod schema expected `context_management` inside the `delta` object
2. the code read from `value.delta.context_management`

but Anthropic's API returns `context_management` at the root level of the `message_delta` event (sibling to `delta` and `usage`):

```json
{
  "type": "message_delta",
  "delta": {"stop_reason": "end_turn"},
  "usage": {...},
  "context_management": {"applied_edits": [...]}
}
```

see [Anthropic's context management docs](https://docs.anthropic.com/en/docs/build-with-claude/context-management)

## summary

- moved `context_management` in the streaming schema from inside `delta` to the root level of `message_delta`
- updated code to read from `value.context_management` instead of `value.delta.context_management`
- added unit test for streaming context_management parsing

## verification

<details>
<summary>before fix</summary>

```
contextManagement: null
```

</details>

<details>
<summary>after fix</summary>

```
contextManagement: {
  appliedEdits: [
    {
      type: 'clear_tool_uses_20250919',
      clearedToolUses: 6,
      clearedInputTokens: 149
    }
  ]
}
```

</details>

## checklist

- [x] tests have been added / updated (for bug fixes / features)
- [ ] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)
- [x] i have reviewed this pull request (self-review)

## related issues

fixes #12148

PR #10540, opened 2025-11-24
feat(anthropic): context management

shubham-021:
## Background

Anthropic's  now supports context management to automatically clear conversation history when approaching token limits. This prevents `model_context_window_exceeded` errors and enables longer conversations by removing older tool uses or thinking blocks while preserving recent context.

See https://platform.claude.com/docs/en/build-with-claude/context-editing

## Summary

Added the context management feature for anthropic:
- Added `context_management` configuration to provider options
- Implemented `clear_tool_uses_20250919` and `clear_thinking_20251015` strategies
- Parse and expose applied context edits in response metadata
- Map `model_context_window_exceeded` stop reason to `'length'`
- Added tests and updated snapshots

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues
Fixes #10485

PR #12547, opened 2026-02-13
test(provider/anthropic): add fixture-based tests and examples for context editing

felixarntz:
## Background

Anthropic context management support (`clear_tool_uses`, `clear_thinking`) was added in #10540 (with a related fix in #12154), and compaction support (`compact`) was added in #12384.

## Summary

With the complexities of handling these features correctly (including how `clear_tool_uses` and `clear_thinking` can be combined), I think it makes sense to add some more tests and also examples around this.

- Adds fixture based tests for context editing (tool uses, thinking, and both combined), each for `generateText` and `streamText`
- Adds examples for thinking, compaction, and thinking and tool uses combined (there already was an example for tool uses alone, which I renamed for consistency)
- Expands docs, adding "Context Editing" grouping section for both `clear_thinking` and `clear_tool_uses`

## Manual Verification

All examples were run against the live Anthropic API to verify correct behavior, and the fixture data was captured from those real API responses.

## Checklist

## Future Work

N/A

## Related Issues

N/A


packages/anthropic/src/anthropic-messages-language-model.test.ts
```diff
@@ -4475,6 +4475,127 @@ describe('AnthropicMessagesLanguageModel', () => {
...
+        ]);
+
+        expect(result.providerMetadata?.anthropic?.contextManagement).toEqual({
+          appliedEdits: [],
+        });
+
+        expect(result.usage).toMatchObject({
+          inputTokens: { total: 859 },
+          outputTokens: { total: 132 },
+        });
```
gr2m: instead of `.toMatchObject()`, can we use `toMatchInlineSnapshot()`? That will make it much easier when we re-record the fixtures in future

felixarntz: thanks, updated in https://github.com/vercel/ai/pull/12547/changes/2377abe165659f1809b2411fcb07db24a28e6cba

aayush-kapoor (approved):

PR #12577, opened 2026-02-13
feat(provider/prodia): add price param to API requests

R-Taneja:
## Background
Gateway needs to know how much to charge for Prodia generations

## Summary
Returns the price of a generation with the new `price` param: https://docs.prodia.com/reference/price/

## Manual Verification
Ran a query

## Checklist


packages/prodia/src/prodia-image-model.ts
```diff
@@ -114,7 +114,7 @@ export class ProdiaImageModel implements ImageModelV3 {
     );
 
     const { value: multipartResult, responseHeaders } = await postToApi({
-      url: `${this.config.baseURL}/job`,
+      url: `${this.config.baseURL}/job?price=true`,
```
shaper, line 117: does this slow the request down? only asking re: whether we might want to make this an option.

shaper (approved):

packages/prodia/src/prodia-image-model.ts
R-Taneja: not afaict (or by a trivial amount). docs say the price is calculated post-completion of the job. the `price=true` just tells prodia to include it in the response

prodia's selling point is being super fast:
>"When you hit our API, we don’t touch a single database, we don’t make a single extra network call—you go right through."
https://cerebralvalley.ai/blog/prodia-the-fastest-image-gen-api-in-the-world-6PAdGrqnVhTuvAfM1pLDVt

PR #12657, opened 2026-02-17
fix(provider/amazon-bedrock): add missing latest Anthropic model IDs to `BedrockChatModelId`

felixarntz:
For some reason, the latest Anthropic models (starting with some from their 4.5 models) were missing from `BedrockChatModelId`. They _were_ however present in `BedrockAnthropicModelId`. Not sure why.

When I added support for Opus 4.6 and Sonnet 4.6, I realized this but left it untouched because I wasn't sure whether the previous omissions (of e.g. Opus 4.5) were intentional.

Opening this PR to fix it, or please let me know if this was _not_ just an oversight from before. :)

cc @lgrammel @gr2m


gr2m:
@tigent never apply the labels minor or major. Leave that to the humans. Please update your config accordingly

gr2m:
@lgrammel is there some historic context you can share? I was confused about this as well

shaper (approved):

shaper:
I think this is just an oversight as we have two ways to use anthropic w/in the amazon-bedrock meta-provider now. We should add new anthropic models to both at least for the time being IMO.

PR #12668, opened 2026-02-18
feat(anthropic): add web_fetch_20260209 and web_search_20260209

MehediH:
## Background
Anthropic introduced new web tool versions (`web_search_20260209` and `web_fetch_20260209`) that include dynamic filtering capabilities, but those versions were not yet available in AI SDK.

## Summary

This PR adds support for `web_search_20260209` and `web_fetch_20260209` in the Anthropic provider.

## Checklist


.changeset/wise-walls-press.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ai-sdk/anthropic': patch
```
felixarntz, line 2: would be great if you could check for Amazon Bedrock support of this. Though, even if we don't do that as part of this PR, I think at the minimum we'll need to amend [`packages/amazon-bedrock/src/bedrock-prepare-tools.ts` (lines 40-55)](https://github.com/vercel/ai/blob/b3e6f706c325812388c0fe281fa8bbbb66eebe3c/packages/amazon-bedrock/src/bedrock-prepare-tools.ts#L40-L55) to also include the new equivalent tool name - just for parity.

please add the `@ai-sdk/amazon-bedrock` provider here then too, with a `patch` changeset.

felixarntz (approved):
@MehediH Thank you, this looks great!

Just one thing we should add to the Bedrock provider.

If possible, could you add examples for using these tools in `examples/ai-functions`? Not a blocker (since we don't have examples for the old versions of the tools either), but would be great to have for quick manual verification and for the future.

could be 4 examples:
* 1 for new `web_fetch`, 1 for new `web_search` in `examples/ai-functions/src/generate-text`
* 1 for new `web_fetch`, 1 for new `web_search` in `examples/ai-functions/src/stream-text`

.changeset/wise-walls-press.md
MehediH: Both the new tools require code execution which aren't supported on Bedrock: https://platform.claude.com/docs/en/agents-and-tools/tool-use/code-execution-tool 

> Code execution is not currently available on Amazon Bedrock or Google Vertex AI.

felixarntz: Ah right, thanks for clarifying! All good then 👍

felixarntz:
@MehediH Hmm, when I run the new `anthropic-web-fetch-tool-*` example, I see an error part in the output:

```
{
  type: 'tool-error',
  toolCallId: 'srvtoolu_01SRKEgo9u5sQkvT3PNrfzZP',
  toolName: 'code_execution',
  input: {
    type: 'programmatic-tool-call',
    code: '\n' +
      'import json\n' +
      'result = await web_fetch({"url": "https://en.wikipedia.org/wiki/Maglemosian_culture"})\n' +
      'parsed = json.loads(result)\n' +
      'if isinstance(parsed, list):\n' +
      '    print(parsed[0]["title"])\n' +
      '    print(parsed[0]["content"][:4000])\n' +
      'else:\n' +
      '    print(parsed)\n'
  },
  error: "Model tried to call unavailable tool 'code_execution'. Available tools: web_fetch.",
  dynamic: true
}
```

I was wondering whether that's a model problem, but not entirely sure. It's strange it tries to use `code_execution` when we have `web_fetch` available.

examples/ai-functions/src/generate-text/anthropic/web-fetch-20260209-wikipedia.ts
```diff
@@ -0,0 +1,17 @@
+import { anthropic } from '@ai-sdk/anthropic';
```
felixarntz, line 1: These examples look solid, however we just so happened to have restructured the `examples/ai-functions` folder. Once you pull in upstream `main`, you'll see that.

Can you move these examples to the `anthropic` subfolder in their respective locations and remove the `anthropic-` prefix from the file names? Then it'll fit right in with the updated structure.

felixarntz (changes requested):
OK, after further investigation, I've found the caveat with the examples not working as expected.

With `examples/ai-functions/src/stream-text/anthropic/anthropic-web-fetch-tool-20260209.ts`, the tool call results in errors like this:
```
Model tried to call unavailable tool 'code_execution'. Available tools: web_fetch.
```

This happens because the new `web_fetch_2026*` and `web_search_2026*` tools internally rely on Anthropic's `code_execution_*` tool. On the API level this is solved automatically, but the AI SDK has a validation mechanism that ensures only tools that are explicitly provided are considered valid, as part of [`packages/ai/src/generate-text/parse-tool-call.ts` (line 16)](https://github.com/vercel/ai/blob/584cf0828cb83cb0c171f8c0212e6e4a5e977d9e/packages/ai/src/generate-text/parse-tool-call.ts#L16).

The model ends up making a `code_execution` tool call which is rejected with an error by the AI SDK.

It then makes a `web_fetch` tool call, which results in this error:
```
Invalid input for tool web_fetch: Type validation failed: Value: {}.\nError message: [\n  {\n    \"expected\": \"string\",\n    \"code\": \"invalid_type\",\n    \"path\": [\n      \"url\"\n    ],\n    \"message\": \"Invalid input: expected string, received undefined\"\n  }\n]
```

At the end, it somehow works and the content from the Wikipedia page is successfully fetched. But there are a few problems happening on the way, which are probably bugs that we need to fix.

This will require further investigation. I'll look into it.

examples/ai-functions/src/generate-text/anthropic/web-fetch-20260209-wikipedia.ts
MehediH: Done!

packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -116,6 +118,34 @@ export async function convertToAnthropicMessagesPrompt({
...
+        value,
+        schema: webSearch_20260209OutputSchema,
+      });
+    } catch {
+      return await validateTypes({
+        value,
+        schema: webSearch_20250305OutputSchema,
+      });
+    }
+  }
```
lgrammel: why is this needed? it introduces a new pattern that i think is problematic. the tool version is defined by the user, this fallback parsing should not be needed and can lead to data integrity issues in practice

MehediH: good catch! this was previously validating against the `webFetch_20250910OutputSchema` - the underlying schemas are the same, so we can keep using `webFetch_20250910OutputSchema` 

or, switch between the different schemas depending on the tool being - what would we prefer?

MehediH: fix here https://github.com/vercel/ai/pull/12668/commits/27ed237995f056585948ee29f7f0853d38ecb96a to keep using `webFetch_20250910OutputSchema` added a note just to be clear

packages/anthropic/src/anthropic-messages-api.ts
```diff
@@ -597,6 +604,10 @@ export const anthropicMessagesResponseSchema = lazySchema(() =>
                   type: z.literal('code_execution_20250825'),
                   tool_id: z.string(),
                 }),
+                z.object({
+                  type: z.literal('code_execution_20260120'),
```
felixarntz, line 618: Note that #12900 added support for this already. Can you please merge the latest version of the `main` branch into your PR branch? Unless I'm missing something, this change here might not be needed then.

MehediH: Awesome - this is cleaned up now https://github.com/vercel/ai/pull/12668/commits/9a4d864bc6b96935cf0e77068ac92fb05604a51d

felixarntz (commented):
@MehediH This looks almost good to go, but the original problem of the AI SDK blocking those `code_execution` tool calls that Anthropic automatically injects still persists.

Since the validation happens centrally in the `ai` package, we can't easily bypass it just for this particular Anthropic consideration.

I can think of two options here:
- Either, for now we require explicitly passing `code_execution: anthropic.tools.codeExecution_20260120()` every time you use the new `web_fetch` or `web_search` tools.
    - This would obviously be a workaround, not an actual fix. Not ideal, but would allow us to unblock this and think about a proper solution decoupled from this PR.
- Or, we somehow relax the conditions in the `parseToolCall` function in the `ai` package.
    - Currently, we only allow tool calls to go through that are `providerExecuted` _and_ `dynamic`. However, the `code_execution` tool calls returned here are only marked with `providerExecuted`, but not `dynamic`. This is where the validation fails.
    - We could allow `providerExecuted` tool calls to go through in a similar way, even when they're not `dynamic`. That would fix this problem.
    - But this needs careful consideration, because obviously reducing strictness of validation increases a risk of a security gap.
    - For reference: This logic was originally implemented in https://github.com/vercel/ai/commit/81d43085f6adfb463458c980b10c0f0e118db824 (cc @lgrammel).

Please feel free to share your thoughts. I hope we can get to a decision to unblock this PR by tomorrow.

MehediH:
hey @felixarntz thank you for looking into this! as discussed, I went ahead and opted for the workaround for now https://github.com/vercel/ai/pull/12668/commits/28855c8e27828a5bff2e14ddfa5816c7abf51c11 to unblock this PR; added a warning so users know the need for the new code execution tool when using these new web tools. Lmk what you think

felixarntz:
@MehediH I found a cleaner solution that doesn't require the workaround of passing `code_execution` explicitly: We annotate the `code_execution` tool call conditionally for just this situation so that we can bypass the validation here. This keeps the regular validation safe as is, but handles this particular case where Anthropic injects the tool and we want to allow it even though it's implicit. See https://github.com/MehediH/ai/commit/c0e0b636a44408f5d5a91f7c1bc90b50e9a8489b

I also found then another bug where we were ignoring `server_tool_use` input in streaming if it wasn't provided in chunks - that was also relevant here because the new `web_fetch` and `web_search` send their input right away, not via individual JSON deltas. See https://github.com/MehediH/ai/commit/b750896625e4240f42a0c12bf0e86d6b3207cd1d

I'm going to add some more testing here, mostly focused on E2E including UI, and then we should be good to go 🎉

felixarntz (approved):
I added UI examples for this, which surface a few follow up errors which affect UI specifically:

1. UI tool validation does not support dynamic tools. This leads to an error when sending any follow-up messages. While problematic here, this is an existing bug that is best fixed in its own PR.
   - The error is: `No tool schema found for tool part code_execution`
   - While this is somewhat expected because we don't include `code_execution` explicitly in the agent, it should be possible to bypass this because we treat it as an internal tool.
2. The types in the UI break because they infer tools from the agent, but that doesn't work because the `code_execution` tool is not explicitly set.

Since this implementation works for a non-interactive example, I think this is acceptable to merge as a first pass. We will need to follow up on the two remaining problems - neither of which are trivial enough to address as part of this.

felixarntz:
OK, I found one more fix (see https://github.com/vercel/ai/pull/12668/changes/9df88f3a1b64dc79fa531d5c0d6569c291310453): We were not yet using the injected `code_execution` tool calls as `dynamic: true` consistently.

Doing that allows us to avoid the UI type errors, by properly treating the injected calls as dynamic tool calls, since that's what they are in these circumstances.

What remains is the error that there's no tool result included for the `code_execution` tool call, which happens once you send another follow up message. Example: `messages.1: `code_execution` tool use with id `srvtoolu_01HopAaXMHo9To3RavUYwqy7` was found without a corresponding `code_execution_tool_result` block"},"request_id":"req_011CYgrdHD1W1jtxevekpz2N"}`

I find this one confusing, because what would even that result be? As far as I understand, that code just calls the `web_fetch` or `web_search` tool. In the official Anthropic docs, I also don't see a tool result for the `code_execution` call, only for the underlying call, which we have too: https://platform.claude.com/docs/en/agents-and-tools/tool-use/programmatic-tool-calling#step-3-provide-tool-result

Let's look into this one separately.

PR #12669, opened 2026-02-18
docs: add guide for setting up AI SDK with coding agents

nicoalbanese:
## Summary

This pull request adds a comprehensive guide for developers interested in using the AI SDK with coding agents. The new documentation covers the setup process and key concepts for implementing AI-powered coding agents.

## Changes

- Added new documentation file `02-getting-started/09-coding-agents.mdx` with a complete guide for setting up and using coding agents with the AI SDK
- Includes detailed instructions, examples, and best practices for coding agent implementation

PR #12672, opened 2026-02-18
docs: expand code example to use `wrapLanguageModel` only during development

felixarntz:
## Summary

Follow up to #12669, specifically [this comment](https://github.com/vercel/ai/pull/12669#discussion_r2822806351).

### DIFF
--- a/content/docs/02-getting-started/09-coding-agents.mdx
+++ b/content/docs/02-getting-started/09-coding-agents.mdx
@@ -89,16 +89,21 @@ Install the DevTools package:
 
 ### Add the middleware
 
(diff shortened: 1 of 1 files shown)


gr2m:
Closing as no longer relevant. The `### Add the middleware` section this PR updates has been rewritten on `main` — the docs now use `registerTelemetry(DevToolsTelemetry())` instead of wrapping the model with `wrapLanguageModel` + `devToolsMiddleware()`, so the proposed `process.env.NODE_ENV === 'development'` conditional no longer has a place to land. Thanks for the follow-up regardless!

PR #12471, opened 2026-02-12
fix(anthropic): allow null content in compaction_delta streaming schema

leog25:
## Background

PR #12384 introduced support for Anthropic's compaction feature, but the streaming Zod schema for compaction_delta events requires content to be a non-nullable string. Anthropic's API sends compaction_delta events with content: null (e.g. the initial frame before the compaction summary text), which causes Zod validation to fail at runtime.

## Summary

Changed content: z.string() to content: z.string().nullish() in the compaction_delta streaming schema in packages/anthropic/src/anthropic-messages-api.ts. This matches the existing content_block_start schema for compaction, which already uses .nullish().

## Manual Verification

Reproduction requires streaming a conversation that exceeds the compaction trigger threshold (50k+ tokens) with @ai-sdk/anthropic, which triggers a compaction_delta event with content: null. This is not practical to verify without an API key and a large conversation, but the added regression test simulates this exact scenario by streaming a compaction_delta chunk with content: null.

## Checklist

- Tests have been added / updated (for bug fixes / features)
- Documentation has been added / updated (for bug fixes / features)
- A patch changeset for relevant packages has been added (for bug fixes / features - run pnpm changeset in the project root)
- I have reviewed this pull request (self-review)

## Related Issues

Fixes #12470

PR #12684, opened 2026-02-18
fix(anthropic): allow null content in compaction_delta streaming schema

### LINKED ISSUE #12470: compaction_delta streaming Zod schema rejects content: null from Anthropic API
### Description

PR #12384 introduced support for Anthropic's compaction feature, but the streaming Zod schema for compaction_delta events requires content to be a non-nullable string. Anthropic's API can send compaction_delta events with content: null (e.g. the initial frame before the compaction summary text), which causes Zod validation to fail at runtime.

Bug

In packages/anthropic/src/anthropic-messages-api.ts, the streaming content_block_start schema for compaction correctly allows null:

z.object({
  type: z.literal('compaction'),
  content: z.string().nullish(),    // allows null
}),

But the content_block_delta schema for compaction_delta does not:

z.object({
  type: z.literal('compaction_delta'),
  content: z.string(),              // rejects null
}),

Expected Behavior

compaction_delta should accept content: null without throwing a validation error.

Fix

  z.object({
    type: z.literal('compaction_delta'),
-   content: z.string(),
+   content: z.string().nullish(),
  }),

Reproduction

1. Use @ai-sdk/anthropic v3.0.41+ with compaction enabled
2. Stream a conversation that exceeds the compaction trigger threshold
3. Zod validation error is thrown when the compaction_delta event arrives with content: null

Additional Context

- Introduced in #12384
- Ref: https://docs.anthropic.com/en/docs/build-with-claude/compaction

---

### AI SDK Version

@ai-sdk/anthropic version 3.0.42

### Code of Conduct

sleitor:
## Summary

The Zod schema for `compaction_delta` streaming events requires `content` to be a non-nullable string. However, Anthropic's API can send `compaction_delta` events with `content: null` (e.g. the initial frame before the compaction summary text), which causes Zod validation to fail at runtime.

The `compaction` `content_block_start` schema already uses `z.string().nullish()` — this PR makes `compaction_delta` consistent.

**Fix:** Change `content: z.string()` → `content: z.string().nullish()`

Closes #12470


sleitor:
The null/undefined content downstream issue is already handled in this PR — the streaming handler uses `value.delta.content ?? ''` to default to an empty string, preventing the TypeError when content is null. Both the schema (z.string().nullish()) and the downstream handler are guarded.

sleitor:
Closing as this fix was already merged via #12471 (PR by @leog25). That PR addresses the same issue (#12470) and was merged on Feb 19.

PR #12686, opened 2026-02-19
feat(ai): stabilize include setting and flip defaults to false

muraliavarma:
## Summary

- Rename `experimental_include` to `include` in `generateText` and `streamText`, stabilizing the API (addresses #12130, #12131)
- Flip default values for `requestBody` and `responseBody` from `true` to `false`, reducing memory usage by default when processing large payloads like images
- Add v7 codemod (`rename-experimental-include`) to automatically migrate existing usage

## Test plan

- Updated existing tests and snapshots to reflect new defaults and renamed option
- Added codemod test with input/output fixtures

### DIFF
--- a/packages/ai/src/generate-text/__snapshots__/generate-text.test.ts.snap
+++ b/packages/ai/src/generate-text/__snapshots__/generate-text.test.ts.snap
@@ -17,7 +17,9 @@ exports[`generateText > options.stopWhen > 2 steps: initial, tool-result > callb
   "rawFinishReason": "stop",
   "reasoning": [],
   "reasoningText": undefined,
-  "request": {},
+  "request": {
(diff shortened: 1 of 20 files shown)


packages/ai/src/generate-text/__snapshots__/generate-text.test.ts.snap
```diff
@@ -17,7 +17,9 @@ exports[`generateText > options.stopWhen > 2 steps: initial, tool-result > callb
   "rawFinishReason": "stop",
   "reasoning": [],
   "reasoningText": undefined,
-  "request": {},
+  "request": {
```
muraliavarma, line 20: Is this expected? I understand this would be a breaking change but maybe I messed up.

PR #12689, opened 2026-02-19
fix(openai): omit reasoning item id when store is false

### LINKED ISSUE #12687: @ai-sdk/openai: Reasoning items include id field when store: false, causing Azure rejection
### Description

**Steps to Reproduce:**

**Expected behavior:**

**Actual behavior:**

**Additional context:**
- SDK: @ai-sdk/openai
- Integration: Azure OpenAI

### AI SDK Version

_No response_

### Code of Conduct

sleitor:
## Summary

When using the OpenAI Responses API with `store: false`, reasoning items sent as input still included the `id` field. Azure OpenAI rejects requests with `id` on reasoning items when `store` is disabled, since the id is only meaningful for stored item references.

## Changes

- **`convert-to-openai-responses-input.ts`**: Omit `id` from inline reasoning items in the `store: false` path. The `id` is only used for `item_reference` lookups when `store: true`.
- **`openai-responses-api.ts`**: Make `id` optional on `OpenAIResponsesReasoning` type.
- **Tests**: Updated all `store: false` reasoning test expectations to not include `id`.

All 68 existing tests in `convert-to-openai-responses-input.test.ts` pass.

Fixes #12687

### DIFF
--- a/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
@@ -960,8 +960,7 @@ describe('convertToOpenAIResponsesInput', () => {
           expect(result.input).toEqual([
             {
               type: 'reasoning',
-              id: 'reasoning_001',
-              encrypted_content: 'encrypted_content_001',
+              encrypted_content: undefined,
               summary: [
                 {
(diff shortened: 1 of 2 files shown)


sleitor:

Issue #12233, opened 2026-02-03
`pipeUIMessageStreamToResponse` buffers entire response instead of streaming in real-time with Next.js Pages Router


mikeguo99:
### Description

When using `pipeUIMessageStreamToResponse` with Next.js 13.0.5 Pages Router API routes, the response is buffered entirely before being sent to the client, resulting in "fake streaming" where all SSE events arrive at the same timestamp.

## Reproduction steps

1. Create a Next.js Pages Router API route using `streamText` and `pipeUIMessageStreamToResponse`:

`// pages/api/chat.ts
import { streamText } from "ai";
import { openai } from "@ai-sdk/openai";

export default async function handler(req, res) {
  const result = streamText({
    model: openai("gpt-4o"),
    messages: req.body.messages,
  });
  
  result.pipeUIMessageStreamToResponse(res);
}`

1. Observe in browser DevTools Network → EventStream tab that all messages have identical timestamps

## Expected behavior

SSE events should arrive incrementally with different timestamps as they are generated by the model.

## Actual behavior

All events are buffered and sent at once after the entire response is complete.

## Workaround

Manually pipe the stream with explicit headers to disable buffering:

`res.socket?.setNoDelay(true);
res.setHeader("Content-Type", "text/event-stream; charset=utf-8");
res.setHeader("Cache-Control", "no-cache, no-transform");
res.setHeader("Content-Encoding", "none");
res.flushHeaders();

const stream = result.toUIMessageStream();
const sseStream = stream.pipeThrough(new TransformStream({
  transform(chunk, controller) {
    controller.enqueue(`data: ${JSON.stringify(chunk)}\n\n`);
  },
}));

const reader = sseStream.getReader();
while (true) {
  const { done, value } = await reader.read();
  if (done) break;
  res.write(value);
}
res.end();`

## Root cause

The internal `writeToServerResponse` function in the SDK doesn't:

1. Set `Content-Encoding: none` to disable compression
2. Call `res.flushHeaders()` before writing
3. Disable Nagle's algorithm with `socket.setNoDelay(true)`

## AI SDK Version

`ai: ^6.0.68`

## Environment

- Next.js 13.0.5 (Pages Router)
- Node.js 18+

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

vico1973:
> ### Description
[...]

PR #12691, opened 2026-02-19
Stream chunks through compression middleware without buffering

OmarSalouss:
## Problem
When compression middleware is active, streaming responses stall — output buffers until generation completes instead of flushing in real-time. The [`compression`](https://github.com/expressjs/compression#server-sent-events) package holds chunks in a buffer window for efficient encoding; without explicit flushes, nothing reaches the client until the response ends. Vercel's own docs [acknowledge this](https://github.com/vercel/ai/blob/49c64b41/content/docs/09-troubleshooting/06-streaming-not-working-when-proxied.mdx#L15-L18), suggesting to disable compression entirely via `'Content-Encoding': 'none'` — sacrificing its benefits altogether.

## Solution
Call [`res.flush()`](https://github.com/expressjs/compression#resflush) after each chunk. Rather than opting out of compression, this lets the compression layer do its job across all scenarios while keeping streaming real-time. You get both — compressed responses and live chunk delivery.


meszarosdezso:
Wow, I was just looking into this issue! If I'm correct this fixes #12233?

OmarSalouss:
> Wow, I was just looking into this issue! If I'm correct this fixes #12233?

Yes, the compression middleware was the reason.

OmarSalouss:
@lgrammel quick ping on this PR, would love your review when you get a chance

PR #12693, opened 2026-02-19
contributing: first draft of project philosophies

felixarntz:
## Summary

Adds a first draft of the project philosophies to our contributing docs, all based on the conversations we had in the past few weeks around this.

The goal of this PR is to create a first draft and have it in the repo. **I would suggest we focus on correctness, rather than completeness.**

## Future Work

We can and will add more sections or bullet points as we go. The most important thing for this PR is that we're aligned that what's documented here is indeed what we think.


lgrammel (approved):

gr2m (approved):
great start!

PR #10885, opened 2025-12-04
fix(openai): change find action type to find_in_page action type

kkawamu1:
## Background

OAI returns the `find_in_page` action type and not `find`, leading to validation errors.
https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses#:~:text=in%20reasoning%20models.-,find_in_page,-%2C%20which%20represents%20searching

## Summary

Replaced `find` with `find_in_page`

## Manual Verification

First, just run the repro code example to observe the error.
Then apply the fix and run the same example again

## Checklist

- [ ] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ ] I have reviewed this pull request (self-review)

## Related Issues

Similar to https://github.com/vercel/ai/pull/10646 by @aayush-kapoor

PR #10646, opened 2025-11-27
fix(openai): allow open_page action type url to be nullish

aayush-kapoor:
## Background

OAI sometimes doesn't render the `url` param for the `open_page` action type, leading to validation errors.

## Summary

Set the parameter to `.nullish()`

## Manual Verification

- First, just run the repro code example to observe the error.
- Then apply the fix and run the same example again 

## Checklist

- [ ] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ ] I have reviewed this pull request (self-review)

## Related Issues

Fixes #9308 and #10669

PR #12706, opened 2026-02-19
fix(openai): change web search tool action to be optional

kkawamu1:
## Background

OpenAI's API sometimes omits the `action` field in web search tool responses, leading to validation errors.

For example, the response sometimes lacks the action field (see updated fixture):
```
    {
      "id": "ws_006c87998160835a00699768d460e88190ab44048180541660",
      "type": "web_search_call",
      "status": "completed"
    },
```

## Summary

Set the parameter to `.optional()`

## Manual Verification

First, just run the repro code example to observe the error.
Then apply the fix and run the same example again

## Checklist

## Future Work

When `action` is missing from a `web_search_call`, `mapWebSearchOutput` currently returns `{}` as the `tool-result` result. This works for now since OpenAI only needs the tool call `id` for multi-turn continuity, but in the future we may want to return something more meaningful (e.g. a completion status derived from the `status` field) instead of an empty object. Worth revisiting if we see any regressions in multi-turn conversations.

## Related Issues

Similar to https://github.com/vercel/ai/pull/10646 by @aayush-kapoor and https://github.com/vercel/ai/pull/10885 by @kkawamu1


kkawamu1:
@aayush-kapoor @lgrammel Any update on this?

aw3aw3:
Currently blocked by this as well

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1969,8 +1969,12 @@ function isErrorChunk(
 }
 
 function mapWebSearchOutput(
-  action: OpenAIResponsesWebSearchAction,
+  action: OpenAIResponsesWebSearchAction | null | undefined,
 ): InferSchema<typeof webSearchOutputSchema> {
+  if (action == null) {
+    return {};
+  }
+
```
aayush-kapoor, line 1977: if this returns nothing here, what do we map back as the result of `tool-result` ?

aayush-kapoor (commented):
changes look good but have you tested with the a multiturn example? what happens after sending a follow up message after openAI returns a web search call with no action type?

packages/openai/src/responses/openai-responses-language-model.ts
aayush-kapoor: this is where there might be a problem with multi turn conversations

kkawamu1: Great question! This should be fine for multi-turn. `mapWebSearchOutput` is only used to construct the `result` field on the `tool-result` content part ([here](https://github.com/vercel/ai/pull/12706/changes#diff-a8f5653038db933006f4d73381ba31ec55c942c41f69a22cb85cd7437298b935R735)) line 735. So when `action` is null, the tool-result gets `result: {}`.


For multi-turn conversations, the important part is that we still send the `toolCallId` (i.e. the `web_search_call` `id`) back to OpenAI when converting to responses input ([ref](https://github.com/vercel/ai/blob/main/packages/openai/src/responses/convert-to-openai-responses-input.ts#L347)), which is what OpenAI needs to maintain conversation continuity. The `result` content itself isn't sent back to OpenAI in the responses API — only the tool call `id` matters.


That said, I have verified this with a multi-turn example and it works as expected. 


Hope this helps! @aayush-kapoor

packages/openai/src/responses/openai-responses-language-model.ts
kkawamu1, line 735: here

packages/openai/src/responses/openai-responses-language-model.ts
aayush-kapoor: yeah i know but i wonder if it's ok to leave the `result` as empty or do we make it say it completed (with a boolean flag, basically to say if the web search call completed successfully or not)

however that would require possibly a spec change and could lead to breaking changes. 

approving for now - but could you just please make a note in future work of this PR that action might need to be returned as something else if we see any regression in the future.

aayush-kapoor (approved):

aayush-kapoor:
@kkawamu1 i'm wondering now if we made a mistake merging this - the file https://github.com/vercel/ai/pull/12706/changes#diff-317e02840ab00cc466e06edd647229356977f9f531e74fabe126608a8b0e8552

has too many regressions. we shouldn't have changed the fixtures. i'll see if we can revert or if might have to open a PR again to fix it. 

edit - follow up PR here: https://github.com/vercel/ai/pull/12788

PR #12730, opened 2026-02-20
fix(provider/openai): add usage to ImageModelV2 doGenerate return type

R-Taneja:
## Summary
- Adds optional `usage` field (`inputTokens`, `outputTokens`, `totalTokens`) to the `ImageModelV2.doGenerate` return type in `@ai-sdk/provider`, matching what v3 already has
- Returns token usage from the OpenAI image provider's `doGenerate` response (the data was already parsed from the API but not passed through)

This fixes an issue where AI Gateway undercharges ~50% for gpt-image-1 requests on the v2/OpenAI-compat path, because the provider returned no usage and the gateway could only apply the flat per-image fee without token-based costs.

### DIFF
--- a/packages/openai/src/image/openai-image-model.ts
+++ b/packages/openai/src/image/openai-image-model.ts
@@ -111,7 +111,7 @@ export class OpenAIImageModel implements ImageModelV2 {
               ? { outputFormat: response.output_format }
               : {}),
             ...distributeTokenDetails(
-              response.usage?.input_tokens_details,
+              response.usage,
               index,
               response.data.length,
             ),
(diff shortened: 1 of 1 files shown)


packages/provider/src/image-model/v2/image-model-v2.ts
```diff
@@ -62,6 +62,15 @@ Warnings for the call, e.g. unsupported settings.
      */
     warnings: Array<ImageModelV2CallWarning>;
 
+    /**
+Optional token usage information for the image generation call.
+     */
+    usage?: {
```
shaper: Not sure if we can do this as it's a spec change, even though it's just an add. @gr2m to advise.

gr2m (approved):
looks good

PR #12753, opened 2026-02-21
fix(deepinfra,fireworks,openai): fix unsafe providerMetadata check

R-Taneja:
## Background
I was using the Gateway OpenAI compat endpoint to generate an image and didn't send any `providerOptions`, and got this error:

```
InternalServerError: 500 Cannot read properties of undefined (reading 'openai')
```

This was because the AI SDK OpenAI package needed the providerOptions `openai` key to exist (unsafe access). The regular `generateImage` AI SDK + Gateway works without `providerOptions` because it sends an empty object for `providerOptions` when the user doesn't specify any.

## Summary
Made the check safer

## Checklist

### DIFF
--- a/packages/deepinfra/src/deepinfra-image-model.ts
+++ b/packages/deepinfra/src/deepinfra-image-model.ts
@@ -70,7 +70,7 @@ export class DeepInfraImageModel implements ImageModelV3 {
             mask: mask != null ? await fileToBlob(mask) : undefined,
             n,
             size,
-            ...(providerOptions.deepinfra ?? {}),
+            ...(providerOptions?.deepinfra ?? {}),
           },
           { useArrayBrackets: false },
         ),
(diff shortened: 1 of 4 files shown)


packages/deepinfra/src/deepinfra-image-model.ts
```diff
@@ -70,7 +70,7 @@ export class DeepInfraImageModel implements ImageModelV3 {
             mask: mask != null ? await fileToBlob(mask) : undefined,
             n,
             size,
-            ...(providerOptions.deepinfra ?? {}),
+            ...(providerOptions?.deepinfra ?? {}),
```
shaper, line 73: I think this is typed to always be a defined object: https://github.com/vercel/ai/blob/main/packages/provider/src/image-model/v3/image-model-v3-call-options.ts#L62

Feels better to do the `?? {}` or similar like in https://github.com/vercel/ai/blob/main/packages/ai/src/generate-image/generate-image.ts#L185 in gateway?

R-Taneja: right that makes more sense, closing PR

PR #12760, opened 2026-02-22
feat(anthropic): spread message-level providerOptions.anthropic onto assistant messages

llc1123:
## Background

When non-Claude models (e.g. kimi-k2.5) are used via `@ai-sdk/anthropic`, there is no way to pass custom fields like `reasoning_content` on assistant messages through to the HTTP body. The openai-compatible provider already supports this by spreading `providerOptions.openaiCompatible` onto messages, but the Anthropic provider only reads `providerOptions.anthropic` at the part level (for signatures, cache control, etc.), never at the message level.

## Summary

Collect message-level `providerOptions.anthropic` (excluding `cacheControl` / `cache_control`) and spread it onto the serialized assistant message object. This mirrors how the openai-compatible provider handles `providerOptions.openaiCompatible`.

Changes:
- `convert-to-anthropic-messages-prompt.ts`: After building the assistant content array, collect and spread extra fields from `providerOptions.anthropic` onto the message
- Added 2 test cases covering the spread behavior and cache control key exclusion

## Manual Verification

- Built `@ai-sdk/anthropic` successfully
- Tested with kimi-k2.5 via Anthropic SDK: `reasoning_content` now flows through to the HTTP body on assistant messages

## Checklist

## Related Issues

Downstream consumer: https://github.com/anomalyco/opencode/pull/14641

### DIFF
--- a/.changeset/spread-anthropic-provider-options.md
+++ b/.changeset/spread-anthropic-provider-options.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Spread message-level `providerOptions.anthropic` (excluding cache control keys) onto assistant messages, enabling custom fields like `reasoning_content` to pass through to the HTTP body.
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
@@ -1961,6 +1961,57 @@ describe('assistant messages', () => {
       `);
     });
   });
+
+  it('should spread message-level providerOptions.anthropic onto assistant message', async () => {
+    const result = await convertToAnthropicMessagesPrompt({
+      prompt: [
+        {
+          role: 'assistant',
+          content: [{ type: 'text', text: 'Hello' }],
+          providerOptions: {
+            anthropic: {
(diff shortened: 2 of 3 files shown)


llc1123:
Thanks for the review! Addressing each point:

1. **Array.isArray guard** — `providerOptions.anthropic` comes from AI SDK's `SharedV3ProviderMetadata` type, constructed internally by the SDK. It won't be an array.

2. **Object.hasOwn** — `Object.entries()` already only iterates own enumerable properties. It does not traverse the prototype chain, so `Object.hasOwn` is redundant here.

3. **Reserved key conflicts** — The openai-compatible provider spreads `providerOptions.openaiCompatible` without reserved key checks either. This is developer-controlled input; if someone passes `role` or `content`, that's on them. Keeping behavior consistent with the existing pattern.

4. **`satisfies Partial<...>`** — `satisfies Partial` provides no meaningful constraint on an object literal with a spread. The current `as` cast is sufficient and consistent with patterns elsewhere in the codebase.

5. **Documentation** — Agreed, can be added as a follow-up.

6. **Performance / break early** — Assistant blocks typically contain 1-2 messages (the SDK merges consecutive assistant messages into one block). No meaningful optimization opportunity here.

PR #12774, opened 2026-02-23
fix(ai): Don't create duplicate tool parts when models call non-existent tools

### LINKED ISSUE #12772: Two tool parts with the same toolCallId created when the model calls a non-existent tool name
## Summary

## Steps to Reproduce

## Root Cause

### Expected Behavior

### Actual Behavior

## Impact

### AI SDK Version

- ai 6.0.49

### Code of Conduct

josh-williams:
## Background

## Summary

Fixes a bug in `processUIMessageStream` where a model calling a non-existent tool produces two UI message parts for the same `toolCallId` — one static (`tool-{toolName}`) and one dynamic (`dynamic-tool`). This happens because `tool-input-start` creates a static part (when `dynamic` is undefined for an unknown tool), but the subsequent `tool-input-error` arrives with `dynamic: true` and only searches for a `dynamic-tool` part, missing the existing static one and creating a duplicate. The fix checks for an existing part by `toolCallId` before branching on `chunk.dynamic`, so the error updates the existing part in place instead of creating a second one.

## Manual Verification

Manually verified against my own application. After the fix, I got a single tool part like:

```
            {
                "type": "tool-dashboard-page-create-dashboard",
                "state": "output-error",
                "rawInput":
                {
                    "name": "Staging Monitoring"
                },
                "errorText": "Model tried to call unavailable tool 'dashboard-page-create-dashboard'. Available tools: ....",
                "toolCallId": "toolu_015E6Kd6cwSMkw7SFpsa7w88",
                "callProviderMetadata":
                {
                    "anthropic":
                    {
                        "caller":
                        {
                            "type": "direct"
                        }
                    }
                }
            },
```
## Checklist

## Related Issues

Fixes #12772

### DIFF
--- a/.changeset/quiet-tables-relate.md
+++ b/.changeset/quiet-tables-relate.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): Don't create duplicate tool parts when models call non-existent tools
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -6433,6 +6433,108 @@ describe('processUIMessageStream', () => {
     });
   });
 
+  describe('tool input error with dynamic flag mismatch', () => {
+    // Regression: when tool-input-start creates a static part (dynamic is
+    // undefined because the tool isn't in the tools object) and tool-input-error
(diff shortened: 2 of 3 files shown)


gr2m (approved):
Looks good, thank you Josh!

PR #12777, opened 2026-02-23
fix(provider/google): support multimodal tool-result parts in function responses

GidianB:
## Background

Google provider tool results with `output.type = 'content'` were not fully mapped into Gemini `functionResponse.parts`, which blocked reliable multimodal tool-result flows for images/files.

## Summary

This PR adds multimodal tool-result support for `@ai-sdk/google` function responses and documents provider limitations.

### Provider changes
- Updated `packages/google/src/convert-to-google-generative-ai-messages.ts` to:
  - Map `image-data` and `file-data` tool-result parts into `functionResponse.parts` as `inlineData`.
  - Support URL-style parts (`image-url`, `file-url`) **only** when they are base64 `data:` URLs.
  - Throw `UnsupportedFunctionalityError` for remote HTTP(S) URL-style tool-result parts.

- Updated `packages/google/src/google-generative-ai-prompt.ts` types to allow:
  - `functionResponse.parts` with inline media payloads.

### Tests
- Added/updated tests in `packages/google/src/convert-to-google-generative-ai-messages.test.ts` for:
  - `image-data` mapping
  - `file-data` mapping
  - `image-url` base64 `data:` URL mapping
  - Error on non-data `image-url`
  - Error on non-data `file-url`

### Docs
- Updated `content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx`:
  - Added Google support note (Gemini 3 models).
  - Documented that Google tool-result URL-style media must use base64 `data:` URLs (remote HTTP(S) URLs are not supported).

### Examples
- Replaced combined toggle example with focused scripts:
  - `examples/ai-functions/src/generate-text/google-image-tool-result-base64.ts`
  - `examples/ai-functions/src/generate-text/google-image-tool-result-url.ts`
  - `examples/ai-functions/src/generate-text/google-pdf-tool-results-base64.ts`
  - `examples/ai-functions/src/generate-text/google-pdf-tool-results-url.ts`
- Removed:
  - `examples/ai-functions/src/generate-text/google-image-tool-results.ts`

### Changeset
- Added `.changeset/google-multimodal-tool-results.md` (patch for `@ai-sdk/google`).

## Manual Verification

- Ran targeted provider tests:
  - `pnpm --filter @ai-sdk/google test:node -- convert-to-google-generative-ai-messages.test.ts`
- Ran type checks:
  - `pnpm --filter @example/ai-functions type-check`
- Ran lint/format checks on changed files:
  - `pnpm exec eslint ...`
  - `pnpm exec prettier --check ...`

## Checklist

## Related Issues

N/A


IdoZ14:
Hey @GidianB, I've looked over this issue myself, and I have a question regarding the `image-url` and `file-url` cases.
We have a use case where we want the model (specifically on vertex) to load the file from gcs using a gcs url (`gs://...`), and Ive tested locally and it seems to work.
Any particular reason why you only added support for `data:` urls?

AlonSh:
@R-Taneja I'm facing something similar, currently solving locally with `bun patch` but would love to have it fix officially. How can we get such a thing assigned? Is there a matching issue?

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -157,37 +193,55 @@ export function convertToGoogleGenerativeAIMessages(
           const output = part.output;
 
           if (output.type === 'content') {
+            const functionResponseParts: GoogleGenerativeAIFunctionResponsePart[] =
```
felixarntz: Logic here looks solid and follows https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#multimodal, however this is only supported for Gemini 3 models.

This needs to be updated to use the new logic conditionally (if model `startsWith('gemini-3-')`), and otherwise continue to use the old logic.

examples/ai-functions/src/generate-text/google-image-tool-results.ts
```diff
@@ -1,80 +0,0 @@
-import { google } from '@ai-sdk/google';
```
felixarntz, line 1: related to my other comment, this example should not have been removed because those models are still relevant.

examples/ai-functions/src/generate-text/google/image-tool-result-base64.ts
```diff
@@ -0,0 +1,55 @@
+import { google } from '@ai-sdk/google';
```
felixarntz, line 1: We should include these examples for `streamText` too, at least some of them.

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -2,12 +2,48 @@ import {
...
+function convertUrlToolResultPart(
+  type: 'image-url' | 'file-url',
+  url: string,
+): GoogleGenerativeAIFunctionResponsePart {
+  const parsedDataUrl = parseBase64DataUrl(url);
+  if (parsedDataUrl == null) {
+    throw new UnsupportedFunctionalityError({
+      functionality: `URL-based ${type} tool result parts are not supported. Use image-data/file-data or base64 data URLs.`,
+    });
+  }
```
felixarntz, line 38: Technically, this would not have caused an error before, so let's think about whether this is the best path forward.

Before, the URL would have just been passed to the model, as a URL. It probably wouldn't have been able to do anything with it, so that's also not a great situation either way. But maybe it's better to still do pass the URL in this case, rather than throw an error.

I think we should include a warning in this case (via `warnings`), rather than throw. Curious what you think @lgrammel

felixarntz (changes requested):
@GidianB Thank you for this! Overall the logic looks great, but we need to continue supporting existing non-Gemini 3 models.

I'm pushing a few changes, mostly to refresh against recent example restructuring, plus reinstating the deleted example, plus also having those examples for `streamText`.

With those examples, you should be able to test the logic in support of the older Gemini models too. Could you please update the PR?

felixarntz:
@IdoZ14 
> Hey @GidianB, I've looked over this issue myself, and I have a question regarding the `image-url` and `file-url` cases. We have a use case where we want the model (specifically on vertex) to load the file from gcs using a gcs url (`gs://...`), and Ive tested locally and it seems to work. Any particular reason why you only added support for `data:` urls?

According to https://ai.google.dev/api/caching#FunctionResponsePart, this is not supported; it says that only inline base64 files can be used. Now of course it might be that the documentation is incomplete, but we need a way to verify. Could you share more context on how exactly you solved it locally? code snippets or a full example would be helpful.

IdoZ14:
Hey @felixarntz First off thanks for taking this on! it will help us a lot!
I see what you mean, Ive found it says its supported here:
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/function-calling#functionresponsepart
- https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/function-calling#functionresponsefiledata

maybe its only on vertex models.

Locally I called the vertex api directly, with some fake message history that links to a gcs file uri.
Something like this
```
// Before run:
// export ACCESS_TOKEN="your-access-token"

const PROJECT_ID = "your-project-id";
const LOCATION = "global";
const MODEL = "gemini-3-flash-preview";
const ACCESS_TOKEN = process.env.ACCESS_TOKEN;

const gcsFileUri = "gs://your-bucket/path/to/document.pdf";
const fileName = "document.pdf";

const url = `https://aiplatform.googleapis.com/v1beta1/projects/${PROJECT_ID}/locations/${LOCATION}/publishers/google/models/${MODEL}:generateContent`;

const payload = {
  tools: [
    {
      functionDeclarations: [
        {
          name: "get_document_summary",
          description: "Fetches an insurance document and provides its summary.",
          parameters: {
            type: "object",
            properties: {
              file_name: { type: "string" }
            },
            required: ["file_name"]
          }
        }
      ]
    }
  ],
  contents: [
    {
      role: "user",
      parts: [
        { text: "Can you load the insurance policy document named Policy 1.PDF from storage?" }
      ]
    },
    {
      role: "model",
      parts: [
        {
          functionCall: {
            name: "get_document_summary",
            args: { file_name: fileName }
          }
        }
      ]
    },
    {
      role: "user",
      parts: [
        {
          functionResponse: {
            name: "get_document_summary",
            response: {
              status: "ok",
              file_name: fileName,
              gcs_uri: gcsFileUri
            },
            parts: [
              {
                fileData: {
                  fileUri: gcsFileUri,
                  mimeType: "application/pdf"
                }
              }
            ]
          }
        }
      ]
    },
    {
      role: "user",
      parts: [{ text: "Please summarize what is in that file." }]
    }
  ],
  generationConfig: {
    temperature: 0
  }
};

const res = await fetch(url, {
  method: "POST",
  headers: {
    Authorization: `Bearer ${ACCESS_TOKEN}`,
    "Content-Type": "application/json"
  },
  body: JSON.stringify(payload)
});

const bodyText = await res.text();
console.log("status:", res.status, res.statusText);
console.log(bodyText);
``` 

This is was a file uri in tool response, but I also tried out in a model message (less useful IMO but still worked)
```
{
  role: "model",
  parts: [
    {
      fileData: {
        fileUri: gcsFileUri,
        mimeType: gcsMimeType,
      },
    },
  ],
},
```

GidianB:
@felixarntz I updated the PR to preserve the legacy tool-result conversion path for pre-Gemini 3 models, while keeping multimodal `functionResponse.parts` for Gemini 3 / 3.1 models.

I also reverted the non-`data:` URL handling to the previous behavior for now: no throw and no warning, just fallback to the legacy text/JSON representation until we get more guidance there.

Verified with:
- pnpm --filter @ai-sdk/google type-check
- pnpm --filter @ai-sdk/google test:node
- pnpm --filter @ai-sdk/google test:edge

packages/google/src/convert-to-google-generative-ai-messages.ts
GidianB: I reverted this to the pre-error behavior for now: no throw and no warning, just fallback to the legacy text/JSON representation until we get more guidance.

felixarntz (approved):
@GidianB Looks great, thanks for restoring the old functionality so that now all Gemini models should work!

I broke the logic into two separate functions to reduce complexity a bit, and tested thoroughly, including a temporary local change for testing the "file data" part documented on Vertex. But it didn't work, neither for Gemini API nor with Vertex API:
- Gemini API doesn't error, but clearly doesn't process the file
- Vertex API errors that it cannot download the image - probably only works with certain Google-owned URLs

I added examples covering these use-cases already, which currently fail. But I think it's a good idea to keep them in here, so we can hopefully make them pass in the future - potentially with Vertex only. Anyway, that's better as a separate follow up PR. Please feel free to work on it if you're interested!

ghost:
⚠️ Backport to release-v6.0 created but has conflicts: https://github.com/vercel/ai/pull/13684

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `ai` | [`7.0.0-beta.32`](https://www.npmjs.com/package/ai/v/7.0.0-beta.32) |
| `@ai-sdk/angular` | [`3.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Fangular/v/3.0.0-beta.32) |
| `@ai-sdk/gateway` | [`4.0.0-beta.18`](https://www.npmjs.com/package/%40ai-sdk%2Fgateway/v/4.0.0-beta.18) |
| `@ai-sdk/google` | [`4.0.0-beta.15`](https://www.npmjs.com/package/%40ai-sdk%2Fgoogle/v/4.0.0-beta.15) |
| `@ai-sdk/google-vertex` | [`5.0.0-beta.20`](https://www.npmjs.com/package/%40ai-sdk%2Fgoogle-vertex/v/5.0.0-beta.20) |
| `@ai-sdk/langchain` | [`3.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Flangchain/v/3.0.0-beta.32) |
| `@ai-sdk/llamaindex` | [`3.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Fllamaindex/v/3.0.0-beta.32) |
| `@ai-sdk/mistral` | [`4.0.0-beta.8`](https://www.npmjs.com/package/%40ai-sdk%2Fmistral/v/4.0.0-beta.8) |
| `@ai-sdk/react` | [`4.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Freact/v/4.0.0-beta.32) |
| `@ai-sdk/rsc` | [`3.0.0-beta.33`](https://www.npmjs.com/package/%40ai-sdk%2Frsc/v/3.0.0-beta.33) |
| `@ai-sdk/svelte` | [`5.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Fsvelte/v/5.0.0-beta.32) |
| `@ai-sdk/vue` | [`4.0.0-beta.32`](https://www.npmjs.com/package/%40ai-sdk%2Fvue/v/4.0.0-beta.32) |

PR #12778, opened 2026-02-23
fix(amazon-bedrock): strip cross-region inference prefix before embedding model detection

### LINKED ISSUE #12773: Amazon Bedrock embedding models do not support cross-region inference prefixes
### Description

The Amazon Bedrock provider does not support **cross-region inference** for embedding models. AWS allows accessing certain models through cross-region inference access points, which changes the model ID by prepending a region prefix (e.g. `eu.`).

For example:
- Standard: `cohere.embed-v4:0`
- Cross-region (Europe): `eu.cohere.embed-v4:0`

When using a prefixed model ID with `embeddingModel()`, the provider fails with a `400 Malformed request` error. Two issues:
- The request body is built using the wrong schema - it sends `inputText`, `dimensions`, and `normalize` instead of the parameters expected by Cohere's embedding API on Bedrock (`texts`, `inputType`, `outputDimension`, etc.)
- The cross-region model ID prefix (`eu.`) is not recognized, so the provider cannot correctly dispatch the request to the appropriate model handler.

### Steps to reproduce

1. Create an Amazon Bedrock client targeting a region with cross-region inference enabled (e.g. `eu-west-3`)
2. Call `embedMany` using a cross-region model ID such as `eu.cohere.embed-v4:0`
3. Observe the `400 Malformed request error`

#### Minimal reproduction
```ts
import { embedMany } from "ai";
import { createAmazonBedrock } from "@ai-sdk/amazon-bedrock";

const bedrock = createAmazonBedrock({
  region: process.env.AWS_REGION, // e.g. "eu-west-3"
  accessKeyId: process.env.AWS_ACCESS_KEY_ID,
  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY,
});

const { embeddings } = await embedMany({
  // ❌ Fails: cross-region prefix "eu." is not handled by the provider
  model: bedrock.embeddingModel("eu.cohere.embed-v4:0"),
  values: ["Chunk 1", "Chunk 2"],
  maxParallelCalls: 5,
  providerOptions: {
    bedrock: {
      inputType: "search_document",
      outputDimension: 1024,
    },
  },
});
```

**Stacktrace**
```
APICallError [AI_APICallError]: undefined: Malformed request. Please check your parameter/values and try again.
    at BedrockEmbeddingModel.doEmbed (bedrock-embedding-model.ts:107:33)
    ...
  url: 'https://bedrock-runtime.eu-west-3.amazonaws.com/model/eu.cohere.embed-v4%3A0/invoke',
  requestBodyValues: { inputText: 'Chunk 2', dimensions: undefined, normalize: undefined },
  statusCode: 400,
  responseBody: '{"message":"Malformed request. Please check your parameter/values and try again."}'
```

The `requestBodyValues` clearly shows the wrong payload shape being sent - `inputText` instead of `texts`, and Cohere-specific options (`inputType`, `outputDimension`) from `providerOptions` are ignored entirely.

**Root cause**

The issue originates in [bedrock-embedding-model.ts L74-L104](https://github.com/vercel/ai/blob/32be973f7f70fea5bab1fa0cb393f95db1267503/packages/amazon-bedrock/src/bedrock-embedding-model.ts#L74-L104), where the model type is resolved and the request body is constructed. The provider does not strip or handle cross-region prefixes before matching the model ID, and the Cohere-specific payload fields from `providerOptions` are not wired into the request body.
After reviewing the code, this issue appears to affect all embedding models supported by the Amazon Bedrock provider, not just Cohere - any model accessed via a cross-region inference endpoint would hit the same prefix resolution failure.

**Expected behavior**

The provider should:
- Recognize cross-region inference prefixes (`eu.`, etc...) and strip or handle them when resolving the underlying model handler
- Build the correct request body for Cohere embedding models, passing `inputType` and `outputDimension` from `providerOptions`

**Actual behavior**

The request fails with a `400` error because the wrong request body is sent to the Bedrock endpoint.

**Environment**
- Node : v24.12.0
- AWS Region : eu-west-3

---

I'd be happy to submit a PR with a fix if the maintainers agree on the approach. Just let me know if this is the right direction before I start working on it - I wouldn't want to duplicate any ongoing effort.

### AI SDK Version

- ai: 6.0.97
- @ai-sdk/amazon-bedrock: 4.0.63

### Code of Conduct

sleitor:
## Summary

AWS Bedrock cross-region inference access points prepend a region prefix to the model ID (e.g. `eu.cohere.embed-v4:0`, `us.amazon.nova-lite-v1:0`). The embedding model was comparing `this.modelId` directly against known model-family prefixes like `"cohere.embed-"` and `"amazon.nova-"`, so cross-region model IDs fell through to the default Titan handler and sent a malformed request body (`inputText` / `dimensions` / `normalize` instead of the Cohere-expected `texts` / `input_type` / `output_dimension`).

## Root cause

```ts
// Before — fails for eu.cohere.embed-v4:0
const isCohereModel = this.modelId.startsWith('cohere.embed-');
```

## Fix

Extract a `baseModelId` by stripping the optional cross-region region prefix (`/^[a-z]{2,3}\./`) and use it for model-family detection only. The **original** model ID is still used for the API URL so that cross-region routing continues to work.

```ts
const baseModelId = this.modelId.replace(/^[a-z]{2,3}\./, '');
const isCohereModel = baseModelId.startsWith('cohere.embed-');
```

## Changes

- `bedrock-embedding-model.ts`: Add `baseModelId` stripping; use it for `isNovaModel`/`isCohereModel` detection.
- `bedrock-embedding-model.test.ts`: Add a new test case using `eu.cohere.embed-v4:0` to verify the cross-region prefix is handled correctly and the right Cohere request body is sent.

Fixes #12773

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-embedding-model.test.ts
+++ b/packages/amazon-bedrock/src/bedrock-embedding-model.test.ts
@@ -24,6 +24,11 @@ const cohereV4EmbedUrl = `https://bedrock-runtime.us-east-1.amazonaws.com/model/
   'cohere.embed-v4:0',
 )}/invoke`;
 
+// Cross-region inference prepends a region prefix (e.g. "eu.") to the model ID.
+const cohereEuCrossRegionEmbedUrl = `https://bedrock-runtime.us-east-1.amazonaws.com/model/${encodeURIComponent(
+  'eu.cohere.embed-v4:0',
+)}/invoke`;
+
 describe('doEmbed', () => {
   const mockConfigHeaders = {
     'config-header': 'config-value',
@@ -59,6 +64,19 @@ describe('doEmbed', () => {
       },
     },
     [cohereV4EmbedUrl]: {
+      response: {
+        type: 'binary',
+        headers: {
+          'content-type': 'application/json',
+        },
+        body: Buffer.from(
+          JSON.stringify({
+            embeddings: { float: [mockEmbeddings[0]] },
+          }),
+        ),
+      },
+    },
+    [cohereEuCrossRegionEmbedUrl]: {
       response: {
         type: 'binary',
         headers: {
@@ -226,6 +244,43 @@ describe('doEmbed', () => {
       texts: [testValues[0]],
       truncate: undefined,
       output_dimension: 256,
+    });
+  });
+
+  it('should support cross-region inference prefix for Cohere embedding models', async () => {
+    const crossRegionModel = new BedrockEmbeddingModel('eu.cohere.embed-v4:0', {
+      baseUrl: () => 'https://bedrock-runtime.us-east-1.amazonaws.com',
+      headers: mockConfigHeaders,
+      fetch: fakeFetchWithAuth,
+    });
+
+    const { embeddings } = await crossRegionModel.doEmbed({
+      values: [testValues[0]],
+      providerOptions: {
+        bedrock: {
+          inputType: 'search_document',
+          outputDimension: 1024,
+        },
+      },
+    });
+
+    expect(embeddings.length).toBe(1);
+    expect(embeddings[0]).toMatchInlineSnapshot(`
+      [
+        -0.09,
+        0.05,
+        -0.02,
+        0.01,
+        0.04,
+      ]
+    `);
+
+    const body = await server.calls[0].requestBodyJson;
+    expect(body).toEqual({
+      input_type: 'search_document',
+      texts: [testValues[0]],
+      truncate: undefined,
+      output_dimension: 1024,
     });
   });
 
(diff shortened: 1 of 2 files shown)


sleitor:
Fixed in the latest commit. Instead of the narrow `/^[a-z]{2,3}\./` regex, the fix now detects cross-region prefixes structurally: Bedrock model IDs always have exactly one dot (`provider.model-name`), so the presence of a second dot indicates a cross-region prefix. The first segment is stripped in that case, which correctly handles `eu.`, `apac.`, `global.`, `us-gov.`, and any future AWS prefixes without needing to enumerate them. Tests updated to cover all four prefix formats — all 294 pass. ✅

sleitor:

PR #12779, opened 2026-02-23
fix(ai): resolve duplicate tool part when NoSuchToolError changes dynamic flag

### LINKED ISSUE #12772: Two tool parts with the same toolCallId created when the model calls a non-existent tool name

sleitor:
## Summary

When a model calls a non-existent tool name, `processUIMessageStream` was creating **two message parts** for the same `toolCallId` — one static and one dynamic. This caused downstream failures when the conversation was continued, because providers like Anthropic reject requests containing two tool results for the same `toolCallId`.

## Root Cause

The stream event sequence for a non-existent tool is:

1. `tool-input-start` arrives with `dynamic: undefined` (the tool isn't in the tools map, so `tool?.type === 'dynamic'` is `undefined`). Since `chunk.dynamic` is falsy, `updateToolPart` is called → creates a **static** part `{ type: 'tool-<name>', state: 'input-streaming' }`.
2. `tool-input-delta` arrives and correctly updates the part (already uses `partialToolCall.dynamic`).
3. `tool-input-error` arrives with `dynamic: true` (set by the `NoSuchToolError` catch block in `parseToolCall`). The handler branched on **`chunk.dynamic`** → called `updateDynamicToolPart` → searched for `type === 'dynamic-tool'` → not found → pushed a **second** duplicate part `{ type: 'dynamic-tool', state: 'output-error' }`.

## Fix

In the `tool-input-error` handler, look up the **existing partial tool call's** `dynamic` flag instead of relying on `chunk.dynamic`:

```ts
const existingPartialCall = state.partialToolCalls[chunk.toolCallId];
const isDynamic =
  existingPartialCall != null
    ? (existingPartialCall.dynamic ?? false)
    : chunk.dynamic;
```

- If a partial tool call **exists** (it was created during `tool-input-start`): use its `dynamic` flag, treating `undefined` as `false` (static).
- If no partial tool call exists (edge case): fall back to `chunk.dynamic`.

This ensures the same `update*ToolPart` helper is called throughout the entire lifecycle of a single tool invocation, and the existing part is updated in-place rather than duplicated.

## Changes

- `packages/ai/src/ui/process-ui-message-stream.ts`: Fix `tool-input-error` handler to consult existing partial tool call dynamic flag.
- `packages/ai/src/ui/process-ui-message-stream.test.ts`: Add regression test verifying that only one part exists and that the existing static part is updated to `output-error` state.

Fixes #12772

### DIFF
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -6433,6 +6433,75 @@ describe('processUIMessageStream', () => {
     });
   });
 
+  describe('tool-input-error with dynamic:true on a static part (NoSuchToolError)', () => {
+    // Regression test for: when a model calls a non-existent tool the stream
+    // produces tool-input-start (dynamic:false) followed by tool-input-error
+    // (dynamic:true, set by parseToolCall's catch block). Before the fix,
+    // tool-input-error branched on chunk.dynamic and called updateDynamicToolPart,
+    // which could not find the already-created static part and pushed a second
+    // duplicate part for the same toolCallId.
+
+    beforeEach(async () => {
+      const stream = createUIMessageStream([
+        { type: 'start' },
+        { type: 'start-step' },
+        {
+          toolCallId: 'call-missing',
+          toolName: 'nonExistentTool',
+          // dynamic is falsy — the tool isn't in the tools map yet
+          type: 'tool-input-start',
+        },
+        {
+          inputTextDelta: '{}',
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
attached issue has been fixed via the pr #12774; marking this as closed

PR #12749, opened 2026-02-20
feat(ai): add an event emitter to core functions

aayush-kapoor:
## Background

as a part of the ongoing change to refactor telemetry, the idea is to create an event emitter (that basically emits all the information that is present at that step of the flow) so that listeners (such as OTel) can tap into the information and create spans accordingly

## Summary

- add an event emitter + listeners array
- simple example to verify events and listeners work

## Manual Verification

- verified by running `examples/ai-functions/src/events/listen-to-events.ts`

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Future Work

- will need to decouple otel from core functions

PR #12784, opened 2026-02-23
feat(ai): decouple otel from generateText

### LINKED ISSUE #12801: generateText and generateObject missing token usage telemetry attributes
### Description

Problem
generateText and generateObject only emit the deprecated ai.usage.promptTokens and ai.usage.completionTokens telemetry span attributes. The newer, standardized attributes — ai.usage.inputTokens, ai.usage.outputTokens, ai.usage.totalTokens, ai.usage.reasoningTokens, and ai.usage.cachedInputTokens — are never recorded.

This means users who rely on OpenTelemetry tracing to monitor costs cannot observe cache hits, reasoning token usage, or use the consistent attribute names that streamText already emits.

`streamText` correctly emits all of these attributes today; `generateText` and `generateObject` were not updated when the richer usage tracking was added.

## Impact

Prompt caching metrics (ai.usage.cachedInputTokens) are invisible in traces from generateText / generateObject, making it impossible to verify or optimize cache usage through observability tooling.

Reasoning token usage (ai.usage.reasoningTokens) is similarly not surfaced.

Dashboards that query the newer attribute names return no data for non-streaming calls, creating an inconsistent monitoring experience.

## Reproduction

```typescript
import { bedrockAnthropic } from '@ai-sdk/amazon-bedrock';

const { usage } = await generateText({
  model: bedrockAnthropic('us.anthropic.claude-sonnet-4-5-20250929-v1:0'),
  prompt: 'Hello',
  experimental_telemetry: { isEnabled: true },
});

// In the recorded span attributes:
// ✅  ai.usage.promptTokens      — present (deprecated)
// ✅  ai.usage.completionTokens  — present (deprecated)
// ❌  ai.usage.inputTokens       — missing
// ❌  ai.usage.outputTokens      — missing
// ❌  ai.usage.totalTokens       — missing
// ❌  ai.usage.reasoningTokens   — missing
// ❌  ai.usage.cachedInputTokens — missing

```

The same attributes are missing from generateObject spans.

For comparison, streamText already emits all the attributes above correctly.

## Fix
Add the missing ai.usage.inputTokens, ai.usage.outputTokens, ai.usage.totalTokens, ai.usage.reasoningTokens, and ai.usage.cachedInputTokens span attributes to both generateText and generateObject, keeping the deprecated aliases for backwards compatibility.

### AI SDK Version

package | version used
-- | --
ai (Vercel AI SDK) | ^6.0.12
@ai-sdk/anthropic | ^3.0.7
@ai-sdk/amazon-bedrock | ^4.0.9
@ai-sdk/provider | ^3.0.2
@anthropic-ai/sdk | ^0.71.2

### Code of Conduct

aayush-kapoor:
## Background

follow up to #12749 

this will help decouple the integration with otel from the core functions

## Summary

create an otel handler that will help map the raw event data to otel spans and attributes!

## Manual Verification

- ran the example `examples/ai-functions/src/generate-text/anthropic-with-langfuse-telemetry.ts`

## Checklist

## Related Issues

should fix #12801


packages/ai/src/generate-text/generate-text.ts
```diff
@@ -501,682 +495,532 @@ export async function generateText<
...
     abortSignal,
     include,
-    functionId: telemetry?.functionId,
-    metadata: telemetry?.metadata as Record<string, unknown> | undefined,
+    telemetry,
     experimental_context,
   };
 
-  await notifyOnStart(onStartEvent, onStart);
+  registerOtelCall(callId, telemetry);
```
aayush-kapoor: one thing here is that we will have to think of a way that we can register Otel (and other prviders) in a way that we wouldn't have to write this line for each provider..

packages/ai/src/generate-text/generate-text.test.ts
```diff
@@ -7336,7 +7351,7 @@ describe('generateText', () => {
                   "type": "tool-call",
                 },
                 {
-                  "approvalId": "id-1",
+                  "approvalId": "id-2",
```
aayush-kapoor: the regression with id numbers in the test is since we added a top level call to `generateId()` function and that bassically shifts the Ids for all the subsequent calls

packages/ai/src/telemetry/otel-event-handler.ts
```diff
@@ -0,0 +1,496 @@
...
+import { listenOnStepFinish } from '../events/on-step-finish';
+import { listenOnStepStart } from '../events/on-step-start';
+import { listenOnToolCallFinish } from '../events/on-tool-call-finish';
+import { listenOnToolCallStart } from '../events/on-tool-call-start';
+import { listenOnFinish } from '../events/on-finish';
+import { assembleOperationName } from './assemble-operation-name';
+import { getBaseTelemetryAttributes } from './get-base-telemetry-attributes';
+import { getTracer } from './get-tracer';
+import { stringifyForTelemetry } from './stringify-for-telemetry';
+import { TelemetrySettings } from './telemetry-settings';
```
lgrammel, line 29: future work: figure out how to mv to a separate pkg (v7 branch)

packages/ai/src/telemetry/otel-event-handler.ts
```diff
@@ -0,0 +1,496 @@
...
+  rootContext: Context;
+  stepSpan?: Span;
+  stepContext?: Context;
+  toolSpans: Map<string, { span: Span; context: Context }>;
+  baseTelemetryAttributes: Attributes;
+  settings: Record<string, unknown>;
+  pendingStepPromptData?: StepPromptData;
+}
+
+const callStates = new Map<string, CallTelemetryState>();
```
lgrammel: module level state is extremely error prone. we should avoid it whenever we can. what is it used for here?

packages/ai/src/generate-text/generate-text.test.ts
lgrammel: is there some easy way to avoid this and reduce diff churn, e.g. by split generators based on purpose?

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -483,7 +474,10 @@ export async function generateText<
     messages,
   } as Prompt);
 
+  const callId = generateId();
```
lgrammel: one option could be to introduce a separate id generator for call that is passed when internal and also has a `call` prefix in the ids so we can easily tell them apart

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -501,682 +495,532 @@ export async function generateText<
     seed: callSettings.seed,
     maxRetries,
     timeout,
-    headers,
+    headers: headersWithUserAgent as
+      | Record<string, string | undefined>
+      | undefined,
```
lgrammel: why is this cast here?

packages/ai/src/generate-text/generate-text.ts
lgrammel: there would need to be some global registry (in the global object under an ai sdk key) and when setting up the specific telemetry integration they would register there (for all telemetry not just otel). the registered telemetry integrations would then be used as additional listeners/callbacks, e.g. in notifyOnStart

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -501,682 +495,532 @@ export async function generateText<
...
-    metadata: telemetry?.metadata as Record<string, unknown> | undefined,
+    telemetry,
     experimental_context,
   };
 
-  await notifyOnStart(onStartEvent, onStart);
+  registerOtelCall(callId, telemetry);
 
-  const tracer = getTracer(telemetry);
+  await notifyOnStart(onStartEvent, onStart);
```
lgrammel: shouldnt otel get notified here as well? the idea is to create a generic telemetry interface (with the global registry). maybe that step needs to come first as a separate pr before the otel extraction (define the telemetry adapter interface first, then implement for otel)

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: yes, implemented this!

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: fixed it so that otel is now called along with the user callback. going to iterate now on making the global registry

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: i think cursor added it since the event header expected but i didn't realise it'd be redundant - removed!

packages/ai/src/generate-text/generate-text.test.ts
aayush-kapoor: fixed

packages/ai/src/telemetry/otel-event-handler.ts
aayush-kapoor: removed

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -314,19 +324,4 @@ export type OnFinishEvent<TOOLS extends ToolSet = ToolSet> =
...
-     *
-     * @default undefined
-     */
-    experimental_context: unknown;
-
-    /** Identifier from telemetry settings for grouping related operations. */
-    readonly functionId: string | undefined;
-
-    /** Additional metadata from telemetry settings. */
-    readonly metadata: Record<string, unknown> | undefined;
```
lgrammel, line 331: changing events that existed before is breaking for existing users

PR #12800, opened 2026-02-24
fix(provider-utils): fix RFC 9110 violations in User-Agent header

### LINKED ISSUE #12799: bug(provider-utils): invalid User-Agent header format causes 400 errors with Azure OpenAI when using Bun
When using any AI SDK provider with the **Bun runtime** and **Azure OpenAI**, every request fails with a 400 error:

```
The format of value 'ai-sdk/provider-utils/4.0.15 runtime/bun/1.3.9' is invalid.
```

Azure OpenAI validates the `User-Agent` header against [RFC 9110 Section 5.6.2](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.2), which defines the allowed characters in a token:

```
token = 1*tchar
tchar = ALPHA / DIGIT / "!" / "#" / "$" / "%" / "&" / "'"
      / "*" / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
```

The `/` character is **not** a valid `tchar`. It is only allowed as the separator between product name and version (`product = token "/" product-version`). The generated header contains two violations:

1. **`ai-sdk/provider-utils/4.0.15`** — the package name `ai-sdk/provider-utils` contains a slash, making `provider-utils/4.0.15` an invalid product-version token.

2. **`runtime/bun/1.3.9`** — Bun exposes `navigator.userAgent = "Bun/1.3.9"`. Embedded in `runtime/${userAgent.toLowerCase()}` this produces `runtime/bun/1.3.9` where `bun/1.3.9` is an invalid product-version token.

### AI SDK Version
- `@ai-sdk/provider-utils`: 4.0.15
- Runtime: Bun 1.3.9

legacycode:
## Background

RFC 9110 (`token = 1*tchar`) does not allow `'/'` in token values. Azure OpenAI rejects requests with non-compliant User-Agent headers.

## Summary

Two related issues where slashes appeared in invalid token positions:

1. Literal `ai-sdk/provider-utils` in `post-to-api` and `get-from-api` contained a slash in the package name. Fixed directly in source: `ai-sdk/provider-utils` → `ai-sdk-provider-utils`.

2. Runtimes like Bun set `navigator.userAgent = "Bun/1.3.9"`, embedding a slash inside a token value. Fixed centrally in `withUserAgentSuffix` via `normalizeUserAgentPart()`, which sanitizes all suffix parts before appending.

## Manual Verification

Verified against an Azure OpenAI endpoint with the Bun runtime. Before the fix: 400 error with `"The format of value 'ai-sdk/provider-utils/4.0.15 runtime/bun/1.3.9' is invalid."`. After the fix: requests succeed.

## Checklist

## Related Issues

Fixes #12799


gr2m (commented):
can you make this fix specific to azure? Sending`/` characters in user agent is common practice, not sure what Azure's problem is with it.

legacycode:
Thanks for the review! For more context on the RFC specifics, see the linked issue #12799.

To clarify: the issue isn't that `/` is forbidden in User-Agent headers generally — `curl/8.x` and `Mozilla/5.0` use it correctly as the product/version separator. The problem is that the SDK puts `/` **inside the product name token** itself.

Per [RFC 9110 §5.6.2](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.2), a `token` is `1*tchar` and `/` is not a `tchar`. The `/` is only valid as the separator between product name and version:

```
# Valid — "/" only as product/version separator
curl/8.11.1
Mozilla/5.0

# Invalid — "/" inside the product name or version token
ai-sdk/provider-utils/4.0.15   → product="ai-sdk", version="provider-utils/4.0.15"
runtime/bun/1.3.9              → product="runtime", version="bun/1.3.9"
```

Azure OpenAI enforces this strictly, but it's technically valid behaviour on their part. Other strict HTTP validators (curl `--head`, some load balancers) would also reject it.

gonzoblasco:
Solid fix — the `normalizeUserAgentPart` approach is correct and centralized. The RFC 9110 analysis in the issue is precise and this PR addresses both violations.

One observation on the normalization behavior:

**Product/version separator loss**

`normalizeUserAgentPart` replaces all `/` with `-`, which is correct for fixing invalid tokens but also removes the product/version separator. For example:

- `provider/test-openai` → `provider-test-openai` (was: product `provider`, version `test-openai`)
- `ai-sdk/0.0.0-test` → `ai-sdk-0.0.0-test` (was: product `ai-sdk`, version `0.0.0-test`)

Per RFC 9110, `product = token "/" product-version` — the `/` between product and version is valid. The issue is slashes **inside** the token portion (like `ai-sdk/provider-utils` where the package name contains a `/`).

The current blanket replacement is the safest approach — it's better to produce a valid (if less structured) User-Agent than an invalid one. But if preserving the product/version semantics matters for debugging or analytics, an alternative would be to split on the first `/` only and sanitize the token portions:

```ts
function normalizeUserAgentPart(part: string): string {
  const slashIndex = part.indexOf('/')
  if (slashIndex === -1) return part
  const product = part.slice(0, slashIndex).replace(/\//g, '-')
  const version = part.slice(slashIndex + 1).replace(/\//g, '-')
  return `${product}/${version}`
}
```

This would produce `ai-sdk/0.0.0-test` (valid) and `ai-sdk-provider-utils/4.0.15` (valid, preserving the version separator) instead of `ai-sdk-0.0.0-test` and `ai-sdk-provider-utils-4.0.15`.

Not blocking — the current approach is RFC-compliant and functionally correct. Just flagging the semantic difference in case it matters for the SDK's observability/debugging story.

legacynode:
Good catch — you're right that the blanket replacement was overcorrecting. Pushed an update that keeps the first `/` as the product/version separator and only sanitizes slashes nested inside the name or version portion, so cases that were already valid (`ai-sdk/0.0.0`, `provider/openai`, etc.) round-trip unchanged now.

```ts
function normalizeUserAgentPart(part: string): string {
  const slashIndex = part.indexOf('/');
  if (slashIndex === -1) return part;
  const name = part.slice(0, slashIndex);
  const version = part.slice(slashIndex + 1).replace(/\//g, '-');
  if (name.length === 0 || version.length === 0) return part.replace(/\//g, '-');
  return `${name}/${version}`;
}
```

One addition to flag: while working through this I noticed `getRuntimeEnvironmentUserAgent()` has another double-slash case that wasn't covered by a test yet — `runtime/node.js/${process.version}`, produced as the fallback on Node.js < 21.1 (before `navigator.userAgent` existed). Added a regression test for it alongside the fix, since it hits the same nested-slash pattern as the Bun case.

Squashed into a single commit — 782 tests passing (node + edge).

gonzoblasco:
@legacynode just re-reviewed the latest commit. The `normalizeUserAgentPart` update looks correct — preserving the first `/` as the product/version separator while sanitizing nested slashes is the right balance between RFC compliance and semantic User-Agent strings.

The additional `runtime/node.js/${process.version}` case is a good catch. 782 tests passing confirms nothing regressed.

This LGTM. Ready for maintainer review. 🚀

PR #12844, opened 2026-02-25
feat(google): map strict: true to Gemini VALIDATED function calling mode

### LINKED ISSUE #12767: `@ai-sdk/google{-vertex}`: ignores `strict:true` on tools
### Description

In the AI SDK Google and Google Vertex providers, tool strictness is not reflected in Gemini function-calling mode.  

From an end-user perspective, setting `strict: true` on tools should enable validated/strict function calling behavior, but today the provider sends `ANY` (or `AUTO`/`NONE`) only.

**Minimal Reproduction**
```ts
import { z } from 'zod';

await generateText({
  model: google('gemini-3-flash-preview'),
  toolChoice: 'required',
  tools: {
    createMeeting: tool({
      description: 'Create a meeting',
      inputSchema: z.object({
        title: z.string(),
        startTime: z.string(),
        attendees: z.array(z.string()),
      }),
      strict: true, // ignored
    }),
  },
  prompt: 'Schedule a product review tomorrow with alice@example.com',
});

```

**Expected**  
When tool strictness is requested (`strict: true`), Gemini should use validated function calling mode (`VALIDATED`).

https://ai.google.dev/gemini-api/docs/function-calling#function_calling_modes

**Actual**  
Google provider currently maps function calling config modes to `AUTO | NONE | ANY` only, so validated mode is not used.

**Impact**  
- Less predictable schema-conformant tool arguments
- Extra app-side validation/retry logic
- Inconsistent strict behavior across providers

**Request**  
Please add support in the Google/Gemini provider to use `VALIDATED` mode when strict tool behavior is requested (e.g. `strict: true`), and keep existing behavior for non-strict tools.

### AI SDK Version

_No response_

### Code of Conduct

echoVic:
## Summary

When a function tool has `strict: true`, the Google provider now uses `VALIDATED` mode in `toolConfig.functionCallingConfig` instead of the default mode. This is the closest Gemini equivalent to OpenAI's per-tool `strict` mode, enforcing schema validation on the API side.

## Changes

- **`google-prepare-tools.ts`**: After building function declarations, detect if any tool has `strict: true`. If so, override the `functionCallingConfig.mode` to `VALIDATED` for `auto`, `required`, and `tool` toolChoice types. `NONE` mode is unchanged (strict is irrelevant when tools are disabled).
- **`google-prepare-tools.test.ts`**: 6 new test cases covering all toolChoice variants with `strict: true`.

## Context

Gemini's `VALIDATED` mode ([docs](https://ai.google.dev/gemini-api/docs/function-calling#validated-mode)) enforces that model output conforms to the provided function schema. This maps naturally to the AI SDK's `strict: true` option on function tools, which currently has no effect in the Google provider.

Fixes the gap where users setting `strict: true` on tools get schema enforcement with OpenAI but not with Google.

## Related Issues

fixes #12767

### DIFF
--- a/.changeset/google-strict-validated-mode.md
+++ b/.changeset/google-strict-validated-mode.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+feat(google): map `strict: true` tool option to Gemini VALIDATED function calling mode
--- a/packages/google/src/google-prepare-tools.test.ts
+++ b/packages/google/src/google-prepare-tools.test.ts
@@ -449,6 +449,127 @@ it('should handle google maps tool', () => {
   expect(result.toolWarnings).toEqual([]);
 });
 
+it('should use VALIDATED mode when any function tool has strict: true and no toolChoice', () => {
+  const result = prepareTools({
+    tools: [
+      {
+        type: 'function',
+        name: 'strictTool',
+        description: 'A strict tool',
+        inputSchema: { type: 'object', properties: {} },
+        strict: true,
+      },
+      {
+        type: 'function',
+        name: 'normalTool',
+        description: 'A normal tool',
+        inputSchema: { type: 'object', properties: {} },
+      },
+    ],
+    modelId: 'gemini-2.5-flash',
+  });
+  expect(result.toolConfig).toEqual({
+    functionCallingConfig: { mode: 'VALIDATED' },
+  });
+});
+
+it('should use VALIDATED mode with toolChoice auto when strict: true', () => {
+  const result = prepareTools({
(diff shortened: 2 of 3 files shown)


aayush-kapoor:
> When a function tool has strict: true, the Google provider now uses VALIDATED mode in toolConfig.functionCallingConfig instead of the default mode.

could you point in the doc where this is mentioned? i saw that the `ANY` type also makes the model adhere to the tool schema

echoVic:
Closing this in favor of #12913 (fixes #12767) which carries the same change as a bugfix + tests.

PR #12847, opened 2026-02-25
feat(openai): support all OpenAI input_file types in chat and responses

sleitor:
Fixes #12842

## Summary

Currently the OpenAI provider throws `UnsupportedFunctionalityError` for any file MIME type that isn't `image/*` or `application/pdf` (responses path) or `image/*`, `audio/wav`, `audio/mp3`, or `application/pdf` (chat completions path).

OpenAI's API actually supports a wide range of file types via the `input_file` content type (responses) and `file` content type (chat completions), including text, code, JSON, HTML, markdown, and many more.

This PR removes the restrictive MIME type checks and routes all non-image (and non-audio in chat completions) file types through the generic file upload path, letting the OpenAI API handle validation.

### Changes

**Responses path** (`convert-to-openai-responses-input.ts`):
- Merged the `application/pdf` branch into a generic fallback for all non-image types
- Supports URL (`file_url`), file ID, and base64 data URI with correct MIME type
- Removed `UnsupportedFunctionalityError` for unknown file types

**Chat completions path** (`convert-to-openai-chat-messages.ts`):
- Merged the `application/pdf` branch into a generic fallback for all non-image, non-audio types
- File URLs still throw (Chat Completions API does not support `file_url`)
- Supports file ID and base64 data URI with correct MIME type

### Testing

Added tests for `text/plain`, `application/json`, and `text/javascript` file types covering base64 data, URLs (responses only), file IDs, and default filename handling.

### DIFF
--- a/.changeset/fix-openai-input-file-audio-filename.md
+++ b/.changeset/fix-openai-input-file-audio-filename.md
@@ -1,0 +1,8 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): preserve file extension in default filenames and guard against silent audio routing on Responses path.
+
+- Default filename for `file` parts without an explicit `filename` now derives an extension from the MIME type (e.g. `part-0.pdf` instead of `part-0`). Applies to both chat completions and responses paths.
+- Audio file parts on the Responses API path now throw `UnsupportedFunctionalityError` instead of silently falling through to `input_file`, which would produce confusing API-level errors.
--- a/packages/openai/src/chat/convert-to-openai-chat-messages.test.ts
+++ b/packages/openai/src/chat/convert-to-openai-chat-messages.test.ts
@@ -116,23 +116,36 @@ describe('user messages', () => {
   });
 
   describe('file parts', () => {
-    it('should throw for unsupported mime types', () => {
-      expect(() =>
-        convertToOpenAIChatMessages({
-          prompt: [
+    it('should handle unknown mime types as file parts', () => {
+      const result = convertToOpenAIChatMessages({
+        prompt: [
+          {
+            role: 'user',
+            content: [
+              {
+                type: 'file',
+                data: 'AAECAw==',
+                mediaType: 'application/something',
+              },
+            ],
+          },
+        ],
+      });
+
+      expect(result.messages).toEqual([
+        {
+          role: 'user',
+          content: [
             {
-              role: 'user',
-              content: [
-                {
-                  type: 'file',
-                  data: 'AAECAw==',
-                  mediaType: 'application/something',
-                },
-              ],
+              type: 'file',
+              file: {
+                filename: 'part-0.something',
+                file_data: 'data:application/something;base64,AAECAw==',
+              },
             },
           ],
-        }),
-      ).toThrow('file part media type application/something');
+        },
+      ]);
     });
 
     it('should throw for URL data', () => {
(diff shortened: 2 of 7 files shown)


gr2m:
Can you address these, in particular 1, 2, and 4.

We have skills specifically for agents in our repository, make sure your agent uses them.

### 1. Breaking change: Default filename loses .pdf extension

When no explicit filename is provided for a PDF file, the default changes from part-${index}.pdf to part-${index} (no extension).

- Chat completions: convert-to-openai-chat-messages.ts:137 (on main, line 131)
- Responses: convert-to-openai-responses-input.ts:142 (on main, line 139)

If OpenAI uses the filename extension for type detection (in addition to the MIME in the data URI), this could break existing PDF handling for users who don't pass an explicit filename.
Even if OpenAI currently uses the data URI MIME, this is a subtle regression that could surface unexpectedly.

Suggestion: At minimum, derive a sensible default extension from the MIME type (e.g., application/pdf → .pdf, text/plain → .txt), or keep the .pdf extension for the PDF-specific case.

### 2. Audio files silently sent as input_file on the Responses path

The Responses path (convert-to-openai-responses-input.ts) has no audio/* handling. Before this PR, sending an audio file would throw UnsupportedFunctionalityError. After this PR, audio
files would silently be routed as input_file, which is almost certainly wrong — OpenAI's Responses API likely expects audio in a different format (analogous to input_audio in chat
completions).

This means audio/wav, audio/mp3, etc. would now produce confusing API-level errors instead of a clear SDK-level error, which is a worse developer experience.

Suggestion: Add an explicit audio/* check in the responses path — either handle it correctly or throw UnsupportedFunctionalityError as before.

### 3. Error message change (minor breaking)

The error message for URL-based file parts in chat completions changes from 'PDF file parts with URLs' to 'file parts with URLs'. While the error class (UnsupportedFunctionalityError) is
  unchanged, anyone matching on the error message string would break. This is a minor concern but worth noting.

### 4. Missing changeset

Per the repo's contribution guidelines, every PR modifying production code needs a changeset. No .changeset/*.md file is included in this PR.

### 5. File ID detection inconsistency (pre-existing, but now more impactful)

The chat completions path uses hardcoded part.data.startsWith('file-') for file ID detection, while the responses path uses the configurable isFileId(part.data, fileIdPrefixes). This
inconsistency existed before but was limited to PDFs. Now that it applies to all file types, the divergence is more significant. Not a blocker for this PR, but worth flagging.

sleitor:
Thanks for the detailed review! I've addressed issues 1, 2, and 4 in the latest commit (048b876):

**Issue 1 — Default filename loses extension:** Fixed in both paths. Default filename now derives the extension from the MIME type using `mediaTypeToExtension()` from `@ai-sdk/provider-utils`, so e.g. an `application/pdf` file without an explicit filename becomes `part-0.pdf` instead of `part-0`.

**Issue 2 — Audio silently sent as input_file on Responses path:** Fixed. Audio file parts on the Responses path now throw `UnsupportedFunctionalityError` with a clear message, consistent with the behavior before this PR and matching how unsupported types are handled elsewhere in the SDK. This prevents confusing API-level errors.

**Issue 4 — Missing changeset:** Added `.changeset/fix-openai-input-file-audio-filename.md` with a `patch` bump for `@ai-sdk/openai`.

---

Regarding **Issue 3** (error message wording change from 'PDF file parts with URLs' to 'file parts with URLs'): this is intentional — the URL restriction applies to all file types, not just PDFs, so the updated message is more accurate. Happy to revert if you'd prefer to keep the original for back-compat.

Regarding **Issue 5** (file ID detection inconsistency between chat and responses paths): agreed this is a pre-existing inconsistency worth cleaning up, but I'll leave it as a follow-up since it's not introduced by this PR.

sleitor:
Fixed in commit 1fcc155.

**Test failures (VADE bot):** Five tests expected bare `part-N` filenames but the fix for issue #1 now derives the extension via `mediaTypeToExtension()`. Updated all five tests to match:
- `application/something` → `part-0.something`
- `application/pdf` → `part-0.pdf`
- `text/plain` → `part-0.plain`

**Issue #3 — Error message change:** The change from `'PDF file parts with URLs'` to `'file parts with URLs'` is intentional — the function now handles all media types generically, so the narrower PDF-specific message would be misleading. Anyone catching on the message string would need to update (the error class is unchanged), but this is consistent with the behaviour change. Happy to restore the old message if you prefer.

**Issue #5 — File ID detection inconsistency:** The chat path hardcodes `startsWith('file-')` while the responses path uses the configurable `isFileId(data, fileIdPrefixes)`. Fixing this properly requires adding a `fileIdPrefixes` option to `convertToOpenAIChatMessages` and threading it through the language model — happy to do that in this PR or a follow-up, whichever you prefer.

sleitor:
Also addressed issue 5 in the latest commit — the chat completions path now uses a configurable `fileIdPrefixes` parameter (defaulting to `['file-']` for backward compatibility) instead of the hardcoded `startsWith('file-')`, consistent with how the responses path works.

sleitor:
@gr2m Thanks again for the detailed review! All 5 points are now addressed and the branch has been rebased onto the latest main (including the V3→V4 type migration):

1. ✅ Default filename now derives extension from MIME type via `mediaTypeToExtension()`
2. ✅ Explicit `audio/*` guard added to the Responses path — throws `UnsupportedFunctionalityError` as before
3. ✅ Error message updated to `'file parts with URLs'` (more accurate now that it's not PDF-only)
4. ✅ Changeset added
5. ✅ `fileIdPrefixes` param added to chat path for consistency with responses path

640/640 tests passing. Would appreciate a re-review when you have a moment!

lgrammel:
Thank you for the pull request. Issue #12842 was closed as completed. Closing this pull request.

PR #12850, opened 2026-02-25
docs: provider abstraction architecture

lgrammel:
## Background

Adding architecture information for coding agents and contributors.

## Summary

Add basic provider abstraction architecture document.

## Future Work

- add tradeoffs/benefits section to provider abstraction architecture document
- add more architecture documents

### DIFF
--- a/architecture/provider-abstraction.md
+++ b/architecture/provider-abstraction.md
@@ -1,0 +1,198 @@
+# Provider Abstraction Architecture
+
+This document explains how AI functions, model specifications, and provider implementations connect in the AI SDK.
+It starts with an abstract high-level view and then details each V3 model type, including the AI functions that use it and small UML diagrams.
+
+## High-Level Architecture
+
+- **AI functions**: user-facing language functions (for example, `streamText`)
+- **Model specification**: `LanguageModelV3`
+- **Provider implementations**: provider-specific language model implementations of `LanguageModelV3`
+
+```mermaid
+classDiagram
+    class AIFunction
+    class LanguageModelV3 {
+      <<interface>>
+    }
+    class ProviderLanguageModelImplementationA
+    class ProviderLanguageModelImplementationB
+
+    AIFunction ..> LanguageModelV3 : uses
+    ProviderLanguageModelImplementationA ..|> LanguageModelV3 : implements
+    ProviderLanguageModelImplementationB ..|> LanguageModelV3 : implements
+```
+
+## Model-Type Details
+
+### Language Model (`LanguageModelV3`)
+
+Language models are used for text generation and structured generation workflows from prompt or message input.
+
+- **AI functions**
+  - `generateText` - [`packages/ai/src/generate-text/generate-text.ts`](/Users/larsgrammel/repositories/ai/packages/ai/src/generate-text/generate-text.ts) - Generates a complete text result from a language model in a single call.
+  - `streamText` - [`packages/ai/src/generate-text/stream-text.ts`](/Users/larsgrammel/repositories/ai/packages/ai/src/generate-text/stream-text.ts) - Streams language model output incrementally as it is produced.
+  - `generateObject` - [`packages/ai/src/generate-object/generate-object.ts`](/Users/larsgrammel/repositories/ai/packages/ai/src/generate-object/generate-object.ts) - Generates one structured object from a language model according to a schema.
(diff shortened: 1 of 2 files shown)


architecture/provider-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Language Model (`LanguageModelV3`)
+
+Language models are used for text generation and structured generation workflows from prompt or message input.
+
+- **AI functions**
+  - `generateText` - [`packages/ai/src/generate-text/generate-text.ts`](packages/ai/src/generate-text/generate-text.ts) - Generates a complete text result from a language model in a single call.
+  - `streamText` - [`packages/ai/src/generate-text/stream-text.ts`](packages/ai/src/generate-text/stream-text.ts) - Streams language model output incrementally as it is produced.
+  - `generateObject` - [`packages/ai/src/generate-object/generate-object.ts`](packages/ai/src/generate-object/generate-object.ts) - Generates one structured object from a language model according to a schema.
+  - `streamObject` - [`packages/ai/src/generate-object/stream-object.ts`](packages/ai/src/generate-object/stream-object.ts) - Streams partial structured object output from a language model over time.
```
felixarntz, line 36: Do we want to include these here? Since we don't really want them to be used anymore, I wonder if mentioning here puts too much of the LLM's attention on them.

architecture/provider-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+- **Model specification**
+  - `LanguageModelV3` - [`packages/provider/src/language-model/v3/language-model-v3.ts`](packages/provider/src/language-model/v3/language-model-v3.ts)
+- **Provider implementations (examples)**
+  - `OpenAILanguageModel`, `AnthropicLanguageModel`
```
felixarntz, line 40: Maybe include links for these examples as well? Would also clarify that they live in their own provider-specific package each.

Same question applies to the other types of models below.

architecture/provider-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+    class generateText
+    class streamText
+    class LanguageModelV3 {
+      <<interface>>
+    }
+    class OpenAILanguageModel
+
+    generateText ..> LanguageModelV3 : uses
+    streamText ..> LanguageModelV3 : uses
+    OpenAILanguageModel ..|> LanguageModelV3 : implements
```
felixarntz, line 53: Can we annotate the `OpenAILanguageModel` here somehow with a comment that it's an example, or that it applies to all the other providers' language models too? Or maybe just add the Anthropic one too for parity with the above examples?

Same question applies to the other types of models below.

architecture/provider-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+    OpenAIEmbeddingModel ..|> EmbeddingModelV3 : implements
+```
+
+### Image Model (`ImageModelV3`)
+
+Image models are used to generate image outputs from text prompts.
+
+- **AI functions**
+  - `generateImage` - [`packages/ai/src/generate-image/generate-image.ts`](packages/ai/src/generate-image/generate-image.ts) - Generates one or more images from prompt input.
+  - `experimental_generateImage` - [`packages/ai/src/generate-image/generate-image.ts`](packages/ai/src/generate-image/generate-image.ts) - Provides the deprecated experimental alias for `generateImage`.
```
felixarntz, line 88: Should we mention this here? given there's already the stable variant of it that people should use.

architecture/provider-abstraction.md
```diff
@@ -0,0 +1,198 @@
...
+    rerank ..> RerankingModelV3 : uses
+    CohereRerankingModel ..|> RerankingModelV3 : implements
+```
+
+### Transcription Model (`TranscriptionModelV3`)
+
+Transcription models are used to convert audio input into text transcripts.
+
+- **AI functions**
+  - `experimental_transcribe` - [`packages/ai/src/transcribe/transcribe.ts`](packages/ai/src/transcribe/transcribe.ts) - Transcribes audio into text with segment and metadata support.
```
felixarntz, line 136: Need to take a mental note that this needs to be updated once these `experimental_*` ones graduate to stable.

Alternative to maybe make this a bit more future-proof would be to list it without the `experimental_` prefix, but add a note that this is currently not implemented yet and only exists as `experimental_*`?

Or even a general note somewhere in this doc, like:

> If you're unable to find any of the mentioned functions, they may only exist with an `experimental_` prefix. This means they're experimental, and stable versions will only be implemented at a later point.

felixarntz (approved):
Looks great! A few suggestions / questions.

PR #12901, opened 2026-02-26
fix(openai): allow null/undefined type in streaming tool call deltas

sleitor:
## Summary

Fixes #12770

Azure AI Foundry and Mistral models deployed on Azure omit the `type` field in streaming `tool_calls` deltas. The OpenAI chat stream parser was throwing:

```
InvalidResponseDataError: Expected 'function' type.
```

## Root Cause

In `openai-chat-language-model.ts`, the parser checked `if (toolCallDelta.type !== 'function')` at the start of a new tool call. Azure / Mistral omit the `type` field entirely (it is `undefined`), so `undefined !== 'function'` triggered the error before even reading the function name or id.

## Fix

Changed the guard from:
```ts
if (toolCallDelta.type !== 'function') {
```
to:
```ts
if (toolCallDelta.type != null && toolCallDelta.type !== 'function') {
```

A missing `type` is now silently treated as `"function"` (the only valid value). An explicit non-`"function"` type still throws.

## Test

Added a new streaming test case that sends tool call deltas without a `type` field (matching Azure / Mistral behaviour) and verifies the tool call is parsed correctly.

### DIFF
--- a/.changeset/fix-openai-streaming-tool-call-null-type.md
+++ b/.changeset/fix-openai-streaming-tool-call-null-type.md
@@ -1,0 +1,12 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): allow null/undefined type in streaming tool call deltas
+
+Azure AI Foundry and Mistral deployed on Azure omit the `type` field in
+streaming tool_calls deltas. The chat stream parser now accepts a missing
+`type` field (treating it as `"function"`) instead of throwing
+`InvalidResponseDataError: Expected 'function' type.`
+
+Fixes #12770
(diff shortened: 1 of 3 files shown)


aayush-kapoor (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/12905

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `@ai-sdk/azure` | [`3.0.37`](https://www.npmjs.com/package/%40ai-sdk%2Fazure/v/3.0.37) |
| `@ai-sdk/openai` | [`3.0.36`](https://www.npmjs.com/package/%40ai-sdk%2Fopenai/v/3.0.36) |

PR #12913, opened 2026-02-27
fix(google): honor strict tools for Gemini function calling

### LINKED ISSUE #12767: `@ai-sdk/google{-vertex}`: ignores `strict:true` on tools

echoVic:
Fixes #12767.

When any function tool has `strict: true`, map Gemini `functionCallingConfig.mode` to `VALIDATED` (closest equivalent to per-tool strict enforcement).

- Updates tool config mapping in `packages/google/src/google-prepare-tools.ts`
- Adds/updates coverage in `packages/google/src/google-prepare-tools.test.ts`

### DIFF
--- a/.changeset/google-strict-validated-mode.md
+++ b/.changeset/google-strict-validated-mode.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+feat(google): map `strict: true` tool option to Gemini VALIDATED function calling mode
--- a/packages/google/src/google-prepare-tools.test.ts
+++ b/packages/google/src/google-prepare-tools.test.ts
@@ -449,6 +449,127 @@ it('should handle google maps tool', () => {
   expect(result.toolWarnings).toEqual([]);
 });
 
+it('should use VALIDATED mode when any function tool has strict: true and no toolChoice', () => {
+  const result = prepareTools({
+    tools: [
+      {
+        type: 'function',
+        name: 'strictTool',
+        description: 'A strict tool',
+        inputSchema: { type: 'object', properties: {} },
+        strict: true,
+      },
+      {
+        type: 'function',
+        name: 'normalTool',
+        description: 'A normal tool',
+        inputSchema: { type: 'object', properties: {} },
+      },
+    ],
+    modelId: 'gemini-2.5-flash',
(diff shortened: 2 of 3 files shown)


echoVic:
Good catch — for Gemini, switching toolChoice 'required'/'tool' from ANY→VALIDATED can weaken the forced function-calling constraint (model may respond with plain text).

I pushed an update that:
- Keeps mode=ANY for toolChoice required/tool (preserves forced tool calling)
- Keeps VALIDATED only for auto/default cases
- Adds a warning when strict:true is present with required/tool, since server-side validation cannot be guaranteed without breaking semantics

Update is in the PR branch now.

aayush-kapoor (commented):
fixed via https://github.com/vercel/ai/pull/12967 so closing this!

PR #12915, opened 2026-02-27
fix(ai): propagate AbortError in generateText multi-step tool loop

sleitor:
## Summary

Fixes #12878

When an `AbortSignal` fires between steps in a multi-step tool loop (e.g. during tool execution), some providers return a partial result with `finishReason: 'other'` instead of throwing. The `generateText` function silently returned the partial result with no error, giving callers no indication the generation was interrupted.

## Root Cause

The `do { ... } while (...)` loop in `generate-text.ts` passes the `mergedAbortSignal` to each model call and tool execution, but it never checks whether the signal was already aborted _between_ iterations. If a provider happens to return rather than throw on an aborted request, the loop continues into the next step as if nothing happened.

## Fix

Add an abort guard at the top of each loop iteration:

```typescript
do {
  // Check if the abort signal was triggered between steps.
  if (mergedAbortSignal?.aborted) {
    throw (
      mergedAbortSignal.reason ??
      new DOMException('This operation was aborted', 'AbortError')
    );
  }
  // ... rest of step
} while (...)
```

This ensures that regardless of how the provider handled the abort, `generateText` always throws an `AbortError` when the signal is fired — consistent with standard Web API abort semantics.

## Test

Added a test that:
1. Starts a multi-step generation (model returns a tool call on step 1)
2. Aborts the signal inside the tool's `execute` function  
3. Asserts that `generateText` throws an `AbortError` and the second model call is never made

Also updated the existing `should forward abort signal to tool execution` test: its previous pattern (aborting externally then `await generateTextPromise`) relied on the function silently absorbing the abort, which is the bug we're fixing. The revised test simply verifies the abort signal is forwarded to the tool `execute` callback, which is its actual intent.

### DIFF
--- a/.changeset/fix-generate-text-abort-signal-propagation.md
+++ b/.changeset/fix-generate-text-abort-signal-propagation.md
@@ -1,0 +1,11 @@
+---
+'ai': patch
+---
+
+fix(ai): propagate AbortError in generateText multi-step tool loop
+
+When an AbortSignal fires between steps (e.g. during tool execution), some
+providers return a partial result with `finishReason: 'unknown'` instead of
+throwing. The multi-step loop now checks `mergedAbortSignal.aborted` at the
+start of each iteration and re-throws the signal reason, so callers reliably
+receive an `AbortError` instead of a silent partial result.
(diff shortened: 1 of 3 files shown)


sleitor:

lgrammel:
Thank you for the pull request. Issue #12878 was closed as completed. Closing this pull request.

PR #12941, opened 2026-02-28
fix(openai): omit reasoning item id when store is false

### LINKED ISSUE #12687: @ai-sdk/openai: Reasoning items include id field when store: false, causing Azure rejection

giulio-leone:
## Problem

When using the OpenAI Responses API with `store: false`, reasoning items in the input still include an `id` field. Azure OpenAI rejects requests containing reasoning item IDs when storage is disabled, breaking multi-turn conversations with reasoning models on Azure.

## Root Cause

In `convert-to-openai-responses-input.ts`, the `store === false` branch (which correctly sends full reasoning content instead of `item_reference`) was still including `id: reasoningId` on the reasoning objects.

## Fix

Omit the `id` field from reasoning items when `store === false`. The `encrypted_content` field is what enables multi-turn reasoning without server-side persistence — the `id` is only meaningful for `item_reference` lookups (the `store === true` path), not for inline reasoning content.

## Tests

- Updated all existing `store: false` reasoning test expectations to not include `id`
- All 594 OpenAI package tests pass (2 consecutive clean runs)

Fixes #12687

### DIFF
--- a/.changeset/fix-openai-responses-reasoning-id-store-false.md
+++ b/.changeset/fix-openai-responses-reasoning-id-store-false.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+Omit `id` field from reasoning items when `store: false` in the Responses API input. Azure OpenAI rejects requests containing reasoning item IDs when storage is disabled. The `encrypted_content` field is sufficient for multi-turn reasoning without server-side persistence.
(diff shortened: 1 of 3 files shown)


giulio-leone:
Closing in favor of #12952.

PR #12947, opened 2026-02-28
fix(ai): propagate AbortError from tool execution instead of swallowing it

sleitor:
## Summary

When an `AbortSignal` fires during tool execution in `generateText()`'s multi-step loop, the `AbortError` was being caught by the `executeToolCall` catch block and converted to a `tool-error` result. This caused `generateText()` to return normally with `finishReason: 'unknown'` instead of throwing, giving callers no indication the generation was aborted.

## Root Cause

In `packages/ai/src/generate-text/execute-tool-call.ts`, the catch block that handles tool execution errors converts any thrown error into a `TypedToolError` result. Without a special case for abort errors, `AbortError` instances were silently swallowed.

## Fix

Re-throw `AbortError` immediately in the `executeToolCall` catch block (before converting other errors to `tool-error` results), consistent with how the retry utility already handles abort errors.

## Checklist

- [x] Existing tests pass
- [x] ESLint passes

Fixes #12878

### DIFF
--- a/packages/ai/src/generate-text/execute-tool-call.ts
+++ b/packages/ai/src/generate-text/execute-tool-call.ts
@@ -1,4 +1,8 @@
-import { executeTool, ModelMessage } from '@ai-sdk/provider-utils';
+import {
+  executeTool,
+  isAbortError,
+  ModelMessage,
+} from '@ai-sdk/provider-utils';
 import { Tracer } from '@opentelemetry/api';
 import { notify } from '../util/notify';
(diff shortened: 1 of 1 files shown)


sleitor:
Closing in favour of #12915 which addresses the same issue with a more complete fix: it adds a loop-iteration guard in `generate-text.ts` that catches the case where the provider returns `finishReason: 'unknown'` instead of throwing when aborted between steps.

PR #12958, opened 2026-02-28
fix(ai): aggregate gateway cost across multiple generateImage calls

### LINKED ISSUE #12796: generateImage: providerMetadata.cost is overwritten when n > maxImagesPerCall
### Description

When calling `generateImage()` with `n` greater than the model's `maxImagesPerCall`, the SDK correctly splits the request into multiple parallel calls and aggregates usage (tokens) via `addImageModelUsage()`. However, `providerMetadata.gateway` is merged using object spread, which overwrites scalar values like `cost` and
`generationId` instead of aggregating them.

### Root cause:

In [`generate-image.ts`](https://github.com/vercel/ai/blob/main/packages/ai/src/generate-image/generate-image.ts#L228-L231), the aggregation loop uses spread merge:

```ts
// For provider "gateway":
  ...currentEntry,
  ...metadata
};
```

This means each successive call's cost replaces the previous one. If 4 calls each cost $0.02, the final `result.providerMetadata.gateway.cost` will be $0.02 instead of $0.08.

Expected behavior:

Numeric metadata values like cost should be summed across calls, similar to how usage is handled. For example:

```ts
if (providerName === "gateway") {
  if (currentEntry != null && typeof currentEntry === "object") {
      ...currentEntry,
      ...metadata,
      // Aggregate numeric values
      cost: sumCosts(currentEntry.cost, metadata.cost),
    };
  }
}
```

Alternatively, an array of per-call metadata (similar to how responses already works) would let consumers aggregate however they need.

### Reproduction 

Run with `npx tsx repro.ts`:

```ts
import { generateImage } from "ai";
import type { ImageModelV3, ImageModelV3CallOptions } from "@ai-sdk/provider";

// Fake image model that returns a simulated gateway metadata
const fakeImageModel: ImageModelV3 = {
  specificationVersion: "v3",
  provider: "fake",
  modelId: "fake-model",
  maxImagesPerCall: 1, // Forces generateImage to make N separate calls

  async doGenerate({ n }: ImageModelV3CallOptions) {
    const pixel = new Uint8Array([0, 0, 0, 0]);
    return {
      images: Array.from({ length: n }, () => pixel),
      warnings: [],
      response: { timestamp: new Date(), modelId: "fake-model", headers: {} },
      providerMetadata: {
        gateway: {
          images: [],
          cost: "$0.02",                     // Each call costs $0.02
          generationId: crypto.randomUUID(), // Unique per call
        },
      },
      usage: { inputTokens: 10, outputTokens: 5, totalTokens: 15 },
    };
  },
};

async function main() {
    model: fakeImageModel,
    prompt: "a cat",
    n: 4, // 4 images, maxImagesPerCall=1 → 4 separate calls
  });

  console.log("Images generated:", result.images.length);

  console.log("\nUsage (correctly aggregated):");
  console.log(result.usage);
  // Expected: { inputTokens: 40, outputTokens: 20, totalTokens: 60 }

  console.log("\nGateway metadata (overwritten, not aggregated):");
  console.log(result.providerMetadata?.gateway);
  // Expected cost: "$0.08" (4 × $0.02)
  // Actual cost:   "$0.02" (only the last call's value)
}

main(); 
```

### AI SDK Version

- ai: 6.0.97
- @ai-sdk/provider: 3.0.8

### Code of Conduct

sleitor:
## Summary

Closes #12796

When `generateImage` is called with `n > maxImagesPerCall`, the SDK splits the request into multiple parallel calls. Usage tokens are correctly aggregated via `addImageModelUsage()`, but `providerMetadata.gateway` was merged with a plain object spread, so scalar fields like `cost` were overwritten by the last call's value.

**Example:** 4 calls each costing $0.02 would produce `providerMetadata.gateway.cost` = `'0.02'` instead of `'0.08'`.

## Root cause

In `generate-image.ts`, the gateway metadata aggregation loop used:

```ts
  ...(currentEntry as object),
  ...metadata,          // ← overwrites cost from previous calls
};
```

## Fix

Introduced a `sumCosts()` helper that parses and sums numeric/numeric-string cost values, then applies it explicitly after the spread:

```ts
  ...(currentEntry as object),
  ...metadata,
  cost: sumCosts(currentEntry.cost, metadata.cost),
};
```

Handles string costs (e.g. `'0.02'`), numeric costs, and cases where only one side has a value.

## Test plan

- Added new test: *"should aggregate gateway cost across multiple calls"*
- All 39 generate-image tests pass.

### DIFF
--- a/packages/ai/src/generate-image/generate-image.test.ts
+++ b/packages/ai/src/generate-image/generate-image.test.ts
@@ -1051,6 +1051,48 @@ describe('generateImage', () => {
       });
     });
 
+    it('should aggregate gateway cost across multiple calls', async () => {
+      let callCount = 0;
+
+      const result = await generateImage({
+        model: new MockImageModelV3({
+          maxImagesPerCall: 1,
+          doGenerate: async () => {
+            switch (callCount++) {
+              case 0:
+                return createMockResponse({
+                  images: [pngBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      cost: '0.02',
+                    },
+                  },
+                });
+              case 1:
+                return createMockResponse({
+                  images: [jpegBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      cost: '0.02',
+                    },
+                  },
+                });
+              default:
+                throw new Error('Unexpected call');
+            }
+          },
+        }),
+        prompt,
+        n: 2,
+      });
+
+      expect(result.providerMetadata.gateway).toStrictEqual({
+        cost: '0.04',
+      });
+    });
+
     it('should preserve null values in images array', async () => {
       const result = await generateImage({
         model: new MockImageModelV4({
--- a/packages/ai/src/generate-image/generate-image.ts
+++ b/packages/ai/src/generate-image/generate-image.ts
@@ -57,6 +57,17 @@ export type GenerateImagePrompt =
  *
  * @returns A result object that contains the generated images.
  */
+
+function sumCosts(cost1: unknown, cost2: unknown): number | string | undefined {
+  if (cost1 == null) return cost2 as number | string | undefined;
+  if (cost2 == null) return cost1 as number | string | undefined;
+  const n1 = typeof cost1 === 'string' ? parseFloat(cost1) : (cost1 as number);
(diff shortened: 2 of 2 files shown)


sleitor:

PR #12966, opened 2026-03-01
fix(openai-compatible): support extended audio/video formats via file parts

### LINKED ISSUE #12723: `OpenAI Compatible Provider` fails calling `Gemini` with many supported audio / video formats
### Description

While the `openai-compatible` provider supports attaching audio files with `audio/*` mediaType, it explicitly throws an exception when files are not of type `wav` or `mp3`. 

Gemini supports a wider-variety of audio formats though through their openai-compatible API: 
- audio/x-aac, audio/flac, audio/mp3, audio/m4a, audio/mpeg, audio/mpga, audio/mp4, audio/ogg, audio/pcm, audio/wav, audio/webm

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/audio-understanding?hl=en

For video, the `openai-compatible` provider rejects all of them. Supported formats by Gemini are:
- video/x-flv, video/quicktime, video/mpeg, video/mpegs, video/mpg, video/mp4, video/webm, video/wmv, video/3gpp

https://docs.cloud.google.com/vertex-ai/generative-ai/docs/multimodal/video-understanding?hl=en

*How to reproduce:*
```ts
const vertexOpenAI = createOpenAICompatible({
  name: "vertex",
  baseURL: `https://aiplatform.googleapis.com/v1/projects/${process.env.VERTEX_PROJECT_ID!}/locations/global/endpoints/openapi`,
  queryParams: {
    key: process.env.VERTEX_API_KEY!,
  },
});

  model: vertexOpenAI("google/gemini-3-flash-preview"),
  messages: [
    {
      role: "user",
      content: [
        { type: "text", text: "Transcribe the audio for me" },
        {
          type: "file",
          data: new URL("https://upload.wikimedia.org/wikipedia/commons/c/c8/Example.ogg"),
          mediaType: "audio/ogg",
        },
      ],
    },
  ],
});

console.log(text)
```

This bug is a follow-up on https://github.com/vercel/ai/issues/11779 which addressed part of the issue already.

Tagging @aayush-kapoor since you have worked on this before :)

### AI SDK Version

_No response_

### Code of Conduct

sleitor:
## Summary

Fixes #12723

The openai-compatible provider previously threw `UnsupportedFunctionalityError` for any audio format other than `wav`/`mp3`, and for all video media types. Providers like Gemini (via their OpenAI-compatible endpoint) support a wider range of audio and video formats.

### Extended audio support
Audio types not covered by the standard OpenAI `input_audio` format (wav/mp3) are now passed as a `file` content part with a data URI or URL string:
- Extended types: `audio/ogg`, `audio/flac`, `audio/m4a`, `audio/aac`, `audio/webm`, `audio/pcm`, `audio/mpga`, `audio/mp4`, etc.
- URL-based audio (all formats) now works via `file_data` URL passthrough

Standard `wav`/`mp3` with inline bytes continue to use the `input_audio` format for backward compatibility.

### Video support
All `video/*` media types are now supported via the `file` content part type, accepting both inline bytes (serialized as a data URI) and URL references.

### Still unsupported
Truly unknown audio/video formats (not in the extended allowlist and not standard wav/mp3) still throw `UnsupportedFunctionalityError`.

## Testing
Updated existing tests that previously asserted errors to instead verify the correct output, and added tests for unsupported format errors.

### DIFF
--- a/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.test.ts
+++ b/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.test.ts
@@ -181,24 +181,65 @@ describe('user messages', () => {
     ]);
   });
 
-  it('should throw error for audio parts with URLs', async () => {
-    expect(() =>
-      convertToOpenAICompatibleChatMessages([
-        {
-          role: 'user',
-          content: [
-            {
-              type: 'file',
-              data: new URL('https://example.com/audio.wav'),
-              mediaType: 'audio/wav',
+  it('should convert audio parts with URLs to file parts', async () => {
+    const result = convertToOpenAICompatibleChatMessages([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'file',
+            data: new URL('https://example.com/audio.wav'),
+            mediaType: 'audio/wav',
+          },
+        ],
+      },
+    ]);
+    expect(result).toEqual([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'file',
+            file: {
+              filename: 'audio.wav',
+              file_data: 'https://example.com/audio.wav',
             },
-          ],
-        },
-      ]),
-    ).toThrow("'audio file parts with URLs' functionality not supported");
+          },
+        ],
+      },
+    ]);
   });
 
-  it('should throw error for unsupported audio format', async () => {
+  it('should convert extended audio format (ogg) to file part', async () => {
+    const result = convertToOpenAICompatibleChatMessages([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'file',
+            data: new Uint8Array([0, 1, 2, 3]),
+            mediaType: 'audio/ogg',
+          },
+        ],
+      },
+    ]);
+    expect(result).toEqual([
+      {
+        role: 'user',
+        content: [
+          {
+            type: 'file',
+            file: {
+              filename: 'audio.ogg',
+              file_data: 'data:audio/ogg;base64,AAECAw==',
+            },
+          },
+        ],
+      },
+    ]);
+  });
+
+  it('should throw error for truly unsupported audio formats', async () => {
     expect(() =>
       convertToOpenAICompatibleChatMessages([
         {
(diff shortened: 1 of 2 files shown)


sleitor:
@heiwen Great point! The current PR adds extended audio types explicitly (as the API requires type-checking to construct correct file parts), but you're right that a wildcard `audio/*` approach would be more maintainable.

The reason image and video use wildcards is because they're passed through as data URIs with their MIME type preserved — the provider doesn't need to know the specific subtype. Audio is handled differently here because we need to map MIME types to format strings for the `input_audio` content type.

That said, wildcard matching for audio (i.e. `mimeType.startsWith('audio/')`) would be the cleanest approach and matches how image/video wildcards work. I can update the PR to use `audio/*` wildcard matching instead of the explicit list if the maintainers prefer that approach — would just need to remove the `getAudioFormat` helper and pass the MIME type through directly.

heiwen:
Personally aligned with that approach. @aayush-kapoor what do you think?

sleitor:

heiwen:
@sleitor IMHO we should get rid of the `getAudioFormat` helper all together and pass all audio as `input_audio` (instead of file). That would align with how images and video parts are handled, which don't check against a list of supported formats. It also requires less maintenance when upstream providers add more supported formats over time.

sleitor:
@heiwen Great suggestion! Applied in the latest commit — removed the `getAudioFormat` helper entirely and now pass all audio parts as `input_audio` regardless of format. The `format` field is derived directly from the media type subtype (e.g., `audio/ogg` → `ogg`), and URL data is passed as-is in the `data` field. This aligns with how image parts work and avoids maintaining a hard-coded allowlist as providers add new formats.

heiwen:
@aayush-kapoor has helped with PRs in this area before. if you have some time this week, mind looking at @sleitor's work?

sleitor:
@aayush-kapoor Gentle ping 👋 — @heiwen kindly suggested your expertise on this PR back in April. The implementation has been updated to remove the `getAudioFormat` helper and pass all audio parts as `input_audio` directly (as @heiwen recommended). Happy to answer any questions or make further adjustments if needed!

heiwen:
Personally I believe the approach in this PR is the right approach.

gr2m:
Closing: the Google OpenAI-compatible endpoint does not support these formats.

PR #12972, opened 2026-03-01
fix(openai-compatible): decode base64 string data for text/* file parts

### LINKED ISSUE #12965: Messages with type: 'file' and mime type: text/... are encoded incorrectly by the OpenAI Compatible provider
### Description

When sending messages of { type: 'file', mediaType: 'text/plain', data: '....string of base 64 encoded data...' } using generateText or streamText, the base 64 encoded string is passed verbatim to the OpenAI Compatible API, inside a {"type":"text","text":"...base 64 encoded data..."} message. Passing the data as base64 is correct when the message to the server is inside a type: 'file' message, but is incorrect when the message to the server is type: 'text'

This contradicts the documentation, which indicates that the data field in FilePart messages is supposed to be a base64 encoded string, as seen here:
https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text#messages.user-model-message.content.file-part

The incorrect code is on line 122 of this file:
https://github.com/vercel/ai/blob/ed17fe86d13bed87a21dc0264178bd2417f999ce/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts

The textContent should be Base64 decoded when using type: 'text' messages

### AI SDK Version

Latest Git

### Code of Conduct

sleitor:
## Problem

When a `FilePart` has a `string` `data` field and a `text/*` media type, the provider was passing the raw base64 string verbatim as the text content instead of decoding it.

Per the `LanguageModelV2DataContent` spec, when `data` is a `string` it is **base64-encoded**. So sending `Buffer.from('Hello').toString('base64')` (i.e. `'SGVsbG8='`) would result in the literal string `'SGVsbG8='` being sent to the API instead of `'Hello'`.

Fixes #12965

## Fix

Decode the base64 string with `Buffer.from(part.data, 'base64').toString('utf-8')` when `data` is a `string` for text/* media types.

Also updated the existing text/markdown test to use properly base64-encoded data (it was passing a plain text string directly, which violates the spec).

### DIFF
--- a/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.test.ts
+++ b/packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.test.ts
@@ -303,7 +303,9 @@ describe('user messages', () => {
           { type: 'text', text: 'Summarize this document' },
           {
             type: 'file',
-            data: '# Hello World\n\nThis is **markdown** content.',
+            data: Buffer.from(
+              '# Hello World\n\nThis is **markdown** content.',
+            ).toString('base64'),
             mediaType: 'text/markdown',
           },
         ],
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
thanks but was missing a changeset and was using a node specific api so changed it a bit in #13006. marking this as closed

PR #12973, opened 2026-03-01
fix(ai): treat approval-requested tool calls as implicit denial in convertToModelMessages

### LINKED ISSUE #12709: "No tool output found for function call" error when submitting user message while tool call part in 'approval-requested' state
### Description

Scenario:

- The model calls a tool that is defined with `needsApproval: true`
- Client ends up with a final assistant message tool call part that has `state: 'approval-requested'`, part rendered to the user as a form with approve/deny buttons
- Instead of clicking approve/deny, where the client uses `addToolApprovalResponse()` to mutate the tool call part and submit the chat, the user sends a new message asking the assistant to, for example, adjust its `input` slightly.
- On submit, the server fails with the "No tool output found for function call" error, since there's a tool call with no result.

I'm currently attempting to work around this by doing things like detecting pending approvals on the client and updating their message parts to `output-denied` to implement an implicit denial, but this seemed like such a common usage pattern, I figured I should open an issue.

### AI SDK Version

- ai: 6.0.34
- @ai-sdk/openai@3.0.10
- @ai-sdk/react@3.0.64

### Code of Conduct

sleitor:
When a user sends a new message while a tool call is still in
`approval-requested` state — for example, the user types a follow-up
instead of clicking approve/deny — the model history conversion fails
with **"No tool output found for function call"** because the assistant
message contains a `tool-call` part but no tool result exists.

## Root cause

`convertToModelMessages` builds the tool-result message by switching on
`toolPart.state`. The `switch` only handled `output-denied`,
`output-error`, and `output-available`. The `approval-requested` case
fell through silently, so the tool call had no corresponding result in
the model messages.

## Fix

Add a `case 'approval-requested':` that emits an implicit denial
tool-result (`error-text: "Tool execution denied."`), matching the
`output-denied` behavior. This gives the model a well-formed history and
lets the conversation continue cleanly.

## Testing

Added a regression test that verifies a pending approval-requested tool
call generates a denial result when a new user message follows.

Fixes #12709

### DIFF
--- a/packages/ai/src/ui/convert-to-model-messages.test.ts
+++ b/packages/ai/src/ui/convert-to-model-messages.test.ts
@@ -1695,6 +1695,106 @@ describe('convertToModelMessages', () => {
       `);
     });
 
+    it('should treat approval-requested tool as implicit denial when user sends new message', async () => {
+      const result = await convertToModelMessages([
+        {
+          parts: [
+            {
+              text: 'What is the weather in Tokyo?',
+              type: 'text',
+            },
+          ],
+          role: 'user',
+        },
+        {
+          parts: [
+            {
+              type: 'step-start',
+            },
+            {
+              approval: {
+                approved: undefined,
+                id: 'approval-1',
+                reason: undefined,
+              },
+              input: {
+                city: 'Tokyo',
+              },
+              state: 'approval-requested',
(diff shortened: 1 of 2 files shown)


sleitor:

PR #12998, opened 2026-03-02
docs(contributing): add pre-release cycle documentation

gr2m:
## Summary

- Add `contributing/pre-release-cycle.md` with detailed documentation for starting, managing, and ending a major-version pre-release cycle (beta releases on `main` while maintaining stable patches on a release branch)
- Update `contributing/releases.md` to reference the new guide instead of inline beta instructions
- Add the new guide to the contributing guides table in `CLAUDE.md`
- update backport workflow to work on release branches

## Context

This documents the process ahead of the v7 pre-release cycle so the steps are well-defined and reusable for future major versions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -236,6 +236,7 @@ For a focused conceptual walkthrough of AI functions, model specifications, and 
 | Provider architecture | `contributing/provider-architecture.md` |
 | Building new features | `contributing/building-new-features.md` |
 | Codemods              | `contributing/codemods.md`              |
+| Pre-release cycle     | `contributing/pre-release-cycle.md`     |
 
 ## Changesets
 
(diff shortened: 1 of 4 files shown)


AGENTS.md
```diff
@@ -236,6 +236,7 @@ For a focused conceptual walkthrough of AI functions, model specifications, and
 | Provider architecture | `contributing/provider-architecture.md` |
 | Building new features | `contributing/building-new-features.md` |
 | Codemods              | `contributing/codemods.md`              |
+| Pre-release cycle     | `contributing/pre-release-cycle.md`     |
```
gr2m: I don't think we need that here, this is not something outside contributors will ever do

```suggestion
```

PR #13006, opened 2026-03-02
fix(openai-compat): decode base64 string data

### LINKED ISSUE #12965: Messages with type: 'file' and mime type: text/... are encoded incorrectly by the OpenAI Compatible provider

aayush-kapoor:
## Background

reported in issue #12965

we decoded the base64 data properly in the anthropic provider but not in openai-compat

## Summary

- added `convertBase64ToUint8Array` to properly decode the string before converting to text
- replaced `Buffer.from(data, 'base64').toString('utf-8')` with the edge-runtime-safe equivalent using `convertBase64ToUint8Array`

## Manual Verification

na

## Checklist

## Related Issues

fixes #12965

### DIFF
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
@@ -7,6 +7,7 @@ import {
   UnsupportedFunctionalityError,
 } from '@ai-sdk/provider';
 import {
+  convertBase64ToUint8Array,
   convertToBase64,
   parseProviderOptions,
   validateTypes,
@@ -31,7 +32,7 @@ import { webSearch_20250305OutputSchema } from './tool/web-search_20250305';
 
 function convertToString(data: LanguageModelV3DataContent): string {
   if (typeof data === 'string') {
-    return Buffer.from(data, 'base64').toString('utf-8');
+    return new TextDecoder().decode(convertBase64ToUint8Array(data));
   }
 
   if (data instanceof Uint8Array) {
--- a/.changeset/fifty-beers-scream.md
+++ b/.changeset/fifty-beers-scream.md
@@ -1,0 +1,6 @@
+---
+'@ai-sdk/openai-compatible': patch
+'@ai-sdk/anthropic': patch
+---
+
+fix(openai-compat): decode base64 string data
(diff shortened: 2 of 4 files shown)


packages/anthropic/src/convert-to-anthropic-messages-prompt.ts
```diff
@@ -31,7 +32,7 @@ import { webSearch_20250305OutputSchema } from './tool/web-search_20250305';
 
 function convertToString(data: LanguageModelV3DataContent): string {
   if (typeof data === 'string') {
-    return Buffer.from(data, 'base64').toString('utf-8');
+    return new TextDecoder().decode(convertBase64ToUint8Array(data));
```
aayush-kapoor, line 35: it's a refactor so doesn't need any test changes and existing tests should pass

felixarntz (commented):
@aayush-kapoor 
> we decoded the base64 data properly in the anthropic provider but not in openai-compat

the only production code change here is in the Anthropic provider - that seems to conflict with your statement? or maybe I'm misunderstanding?

aayush-kapoor:
> the only production code change here is in the Anthropic provider - that seems to conflict with your statement? or maybe I'm misunderstanding?

im not sure i understand? what i meant was - previously, decoding was done for properly in anthropic provider but we missed to do it in openai compat. the current change in [packages/anthropic/src/convert-to-anthropic-messages-prompt.ts](https://github.com/vercel/ai/pull/13006/files/170797a4478eb28a4b34bd15881a72c037d98c8a#diff-fc9fe6860029cf7d2b29bf6db0ac2b28a141bc79748a390fb51242198302d4bb) is only to ensure we don't use a node specific api for this

felixarntz (approved):
@aayush-kapoor ah my bad. I didn't actually see the OpenAI compatible non-test code change before 🤦‍♂️

All good!

PR #13019, opened 2026-03-03
fix(anthropic): reorder assistant content so server_tool_use/web_search_tool_result appear before tool_use

### LINKED ISSUE #13013: Anthropic: tool_use/tool_result pairing fails when server_tool_use (web_search) and regular tool_use appear in same assistant turn
## Description

When the model returns both `server_tool_use` (e.g., `web_search`) and a regular `tool_use` in the same assistant response, the second agent loop iteration fails with:

```
tool_use ids were found without tool_result blocks immediately after: toolu_xxx
```

## Versions

- `ai`: 6.0.105
- `@ai-sdk/anthropic`: 3.0.50
- Node.js: 22.18.0

## Root Cause

The `@ai-sdk/anthropic` provider's `convertToAnthropicMessagesPrompt` function processes assistant message content in order. When the model calls both a regular tool and web_search in the same turn, the reconstructed assistant message for the next iteration has this content order:

```
[0] text                        ← model's reasoning
[1] tool_use (regular tool)     ← regular tool call
[2] server_tool_use (web_search) ← provider-executed web search
[3] web_search_tool_result      ← web search result (AFTER tool_use!)
```

**The Anthropic API requires that `tool_use` blocks are the last content in assistant messages.** Content of type `server_tool_use` and `web_search_tool_result` appearing after `tool_use` causes the API to fail to match the `tool_result` in the next user message.

The issue is specifically in `groupIntoBlocks` + the assistant block conversion, which preserves the insertion order of content parts without reordering them to satisfy Anthropic's constraint.

## Reproduction

Save the following as `repro.ts` and run with `tsx repro.ts` (NO API key needed — it intercepts the HTTP request):

```typescript
import { createAnthropic } from '@ai-sdk/anthropic';

async function main() {
  // Intercept fetch to capture the request body
  let capturedBody: string | null = null;
  globalThis.fetch = async (_input: RequestInfo | URL, init?: RequestInit) => {
    if (init?.body && typeof init.body === 'string') capturedBody = init.body;
    return new Response(
      JSON.stringify({ type: 'error', error: { type: 'authentication_error', message: 'intercepted' } }),
      { status: 401, headers: { 'content-type': 'application/json' } }
    );
  };

  const provider = createAnthropic({ apiKey: 'sk-fake' });
  const model = provider('claude-sonnet-4-20250514');

  // Simulate the exact prompt the AI SDK produces after the first agent loop
  // iteration when both web_search and a regular tool are called in the same turn
  const prompt = [
    { role: 'system', content: 'You are a helpful assistant.' },
    { role: 'user', content: [{ type: 'text', text: 'Search and save news.' }] },
    {
      role: 'assistant',
      content: [
        { type: 'text', text: "I'll search for news and check the schema." },
        { type: 'tool-call', toolCallId: 'toolu_01Regular', toolName: 'get_table_schema', input: {} },
        { type: 'tool-call', toolCallId: 'srvtoolu_01Web', toolName: 'web_search', input: { query: 'news' }, providerExecuted: true },
        {
          type: 'tool-result', toolCallId: 'srvtoolu_01Web', toolName: 'web_search',
          output: { type: 'json', value: [{ type: 'web_search_result', url: 'https://example.com', title: 'News', pageAge: '1h', encryptedContent: 'enc123' }] },
          providerExecuted: true,
        },
      ],
    },
    { role: 'tool', content: [{ type: 'tool-result', toolCallId: 'toolu_01Regular', toolName: 'get_table_schema', output: { type: 'json', value: { columns: [] } } }] },
  ];

  try {
    await model.doGenerate({ inputFormat: 'messages', mode: { type: 'regular', tools: [] }, prompt, providerOptions: {} });
  } catch { /* expected */ }

  const request = JSON.parse(capturedBody!);
  const assistantMsg = request.messages.find((m: any) => m.role === 'assistant');
  const types = assistantMsg.content.map((p: any) => p.type);

  console.log('Assistant message content types:', types);
  // Output: ['text', 'tool_use', 'server_tool_use', 'web_search_tool_result']
  //                   ^^^^^^^^^ tool_use is NOT last — server_tool_use + result appear after it

  const toolUseIdx = types.indexOf('tool_use');
  const hasContentAfter = types.slice(toolUseIdx + 1).some((t: string) => t !== 'tool_use');
  console.log('Content after tool_use?', hasContentAfter); // true — BUG
}

main();
```

**Expected output:** `tool_use` should be the last content type (or at least no non-`tool_use` content should appear after it).

**Actual output:** `['text', 'tool_use', 'server_tool_use', 'web_search_tool_result']` — `server_tool_use` and `web_search_tool_result` appear after `tool_use`.

## Expected Behavior

The Anthropic provider should reorder assistant message content so that:
1. `server_tool_use` + `web_search_tool_result` pairs come BEFORE any regular `tool_use` blocks, OR
2. The `groupIntoBlocks` function handles the content ordering to satisfy Anthropic's API requirement

## Workaround

We currently apply a middleware that reorders assistant message content after all sanitization, ensuring all non-tool content appears before tool content:

```typescript
const middleware: LanguageModelMiddleware = {
  specificationVersion: 'v3',
  transformParams: async ({ params }) => {
    for (const msg of params.prompt) {
      if (msg.role !== 'assistant' || !Array.isArray(msg.content)) continue;
      const toolParts = [];
      const textParts = [];
      let needsReorder = false, seenTool = false;
      for (const part of msg.content) {
        const isTool = part.type === 'tool-call' || part.type === 'tool-result';
        if (isTool) { toolParts.push(part); seenTool = true; }
        else { textParts.push(part); if (seenTool) needsReorder = true; }
      }
      if (needsReorder) msg.content = [...textParts, ...toolParts];
    }
    return params;
  },
};
```

## Additional Context

This bug is 100% reproducible whenever the model decides to call `web_search` (or any provider-executed tool) alongside a regular tool in the same turn, which is common when users ask to "search and save" data. The bug blocks the agent loop from continuing after the first tool execution step.

sleitor:
## Problem

When the model calls both a provider-executed tool (`web_search`, `code_execution`) and a regular tool in the same turn, the assembled assistant message can look like:

```
[text, tool_use (regular), server_tool_use (web_search), web_search_tool_result]
```

The Anthropic API requires `tool_use` to come **last** in assistant content (because its result comes in the subsequent user message). Having `server_tool_use`/`web_search_tool_result` after `tool_use` causes:

```
tool_use ids were found without tool_result blocks immediately after: toolu_xxx
```

This is 100% reproducible when a user asks the agent to both search AND call a regular tool in one step (e.g. "search and save news"). Repro script is in the issue.

Fixes #13013

## Fix

Right before pushing the assembled assistant content to `messages`, check if both regular `tool_use` items and server-tool items (e.g. `server_tool_use`, `web_search_tool_result`, `mcp_tool_use`, `mcp_tool_result`, `code_execution_tool_result`) are present. If so, move all non-`tool_use` content before `tool_use` content. Relative order within each group is preserved.

```diff
+const serverToolTypes = new Set([...]);
+const hasRegularToolUse = anthropicContent.some((c) => c.type === 'tool_use');
+const hasServerTool = anthropicContent.some((c) => serverToolTypes.has(c.type));
+const orderedContent =
+  hasRegularToolUse && hasServerTool
+    ? [
+        ...anthropicContent.filter((c) => c.type !== 'tool_use'),
+        ...anthropicContent.filter((c) => c.type === 'tool_use'),
+      ]
+    : anthropicContent;
 messages.push({ role: 'assistant', content: orderedContent });
```

## Testing

Added a unit test in `convert-to-anthropic-messages-prompt.test.ts` that verifies `server_tool_use` and `web_search_tool_result` indices are less than `tool_use` index in the output.

### DIFF
--- a/.changeset/fix-anthropic-server-tool-ordering.md
+++ b/.changeset/fix-anthropic-server-tool-ordering.md
@@ -1,0 +1,12 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix(anthropic): reorder assistant content so server_tool_use/web_search_tool_result appear before tool_use
+
+When the model calls both a provider-executed tool (web_search, code_execution)
+and a regular tool in the same turn, the Anthropic API requires that
+server_tool_use + web_search_tool_result blocks appear before regular tool_use
+blocks. Previously, content was pushed in insertion order, which could place
+server_tool_use/web_search_tool_result after tool_use and cause the API to
+return "tool_use ids were found without tool_result blocks immediately after".
--- a/packages/anthropic/src/__fixtures__/anthropic-web-search-with-tool.1.chunks.txt
+++ b/packages/anthropic/src/__fixtures__/anthropic-web-search-with-tool.1.chunks.txt
@@ -1,0 +1,16 @@
+{"type":"message_start","message":{"id":"msg_01TestWebAndRegularTool","type":"message","role":"assistant","model":"claude-sonnet-4-20250514","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":80,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":1,"service_tier":"standard"}}}
+{"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}
+{"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"I'll save the results and search for the latest news."}}
+{"type":"content_block_stop","index":0}
+{"type":"content_block_start","index":1,"content_block":{"type":"tool_use","id":"toolu_01SaveResults","name":"save_results","input":{}}}
+{"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":""}}
+{"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"{\"filename\": \"news.txt\"}"}}
+{"type":"content_block_stop","index":1}
+{"type":"content_block_start","index":2,"content_block":{"type":"server_tool_use","id":"srvtoolu_01WebSearch","name":"web_search","input":{}}}
+{"type":"content_block_delta","index":2,"delta":{"type":"input_json_delta","partial_json":""}}
+{"type":"content_block_delta","index":2,"delta":{"type":"input_json_delta","partial_json":"{\"query\": \"latest news\"}"}}
+{"type":"content_block_stop","index":2}
+{"type":"content_block_start","index":3,"content_block":{"type":"web_search_tool_result","tool_use_id":"srvtoolu_01WebSearch","content":[{"type":"web_search_result","title":"Breaking News Today","url":"https://example.com/news","encrypted_content":"EqMBCioICBgCIiRiY2JjZWJjMy1lYTFhLTRmNjktYTUwMy01YzgwNWU1Y2U0NzESDA==","page_age":"1 hour ago"}]}}
+{"type":"content_block_stop","index":3}
+{"type":"message_delta","delta":{"stop_reason":"tool_use","stop_sequence":null},"usage":{"output_tokens":60}}
+{"type":"message_stop"}
--- a/packages/anthropic/src/__fixtures__/anthropic-web-search-with-tool.1.json
+++ b/packages/anthropic/src/__fixtures__/anthropic-web-search-with-tool.1.json
@@ -1,0 +1,45 @@
+{
+  "id": "msg_01TestWebAndRegularTool",
+  "type": "message",
+  "role": "assistant",
+  "model": "claude-sonnet-4-20250514",
+  "content": [
+    {
+      "type": "text",
+      "text": "I'll save the results and search for the latest news."
+    },
+    {
+      "type": "tool_use",
+      "id": "toolu_01SaveResults",
+      "name": "save_results",
+      "input": { "filename": "news.txt" }
+    },
+    {
+      "type": "server_tool_use",
+      "id": "srvtoolu_01WebSearch",
+      "name": "web_search",
+      "input": { "query": "latest news" }
+    },
+    {
+      "type": "web_search_tool_result",
+      "tool_use_id": "srvtoolu_01WebSearch",
+      "content": [
+        {
+          "type": "web_search_result",
+          "url": "https://example.com/news",
+          "title": "Breaking News Today",
+          "encrypted_content": "EqMBCioICBgCIiRiY2JjZWJjMy1lYTFhLTRmNjktYTUwMy01YzgwNWU1Y2U0NzESDA==",
+          "page_age": "1 hour ago"
+        }
+      ]
+    }
+  ],
+  "stop_reason": "tool_use",
+  "stop_sequence": null,
+  "usage": {
+    "input_tokens": 80,
+    "cache_creation_input_tokens": 0,
+    "cache_read_input_tokens": 0,
+    "output_tokens": 60
+  }
+}
--- a/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
+++ b/packages/anthropic/src/convert-to-anthropic-messages-prompt.test.ts
@@ -3103,4 +3103,90 @@ describe('citations', () => {
       `);
     });
   });
+
+  it('should reorder assistant content so server_tool_use+web_search_tool_result come before tool_use', async () => {
+    // When the model calls both a provider-executed tool (web_search) and a
+    // regular tool in the same turn, the resulting anthropicContent array may
+    // have server_tool_use / web_search_tool_result interspersed after tool_use.
+    // The Anthropic API requires tool_use to be last; this test verifies the
+    // reordering.
+    const warnings: SharedV3Warning[] = [];
+    const result = await convertToAnthropicMessagesPrompt({
+      prompt: [
+        {
+          role: 'user',
+          content: [{ type: 'text', text: 'Search and save news.' }],
+        },
+        {
+          role: 'assistant',
+          content: [
+            { type: 'text', text: "I'll search and call the schema tool." },
+            {
+              type: 'tool-call',
+              toolCallId: 'toolu_01Regular',
+              toolName: 'get_table_schema',
+              input: {},
+            },
+            {
+              type: 'tool-call',
+              toolCallId: 'srvtoolu_01Web',
+              toolName: 'web_search',
+              input: { query: 'news' },
+              providerExecuted: true,
+            },
+            {
+              type: 'tool-result',
+              toolCallId: 'srvtoolu_01Web',
+              toolName: 'web_search',
+              output: {
+                type: 'json',
+                value: [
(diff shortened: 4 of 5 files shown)


ahmadbilal55:
unsubscirbe

sleitor:
Fixed in commit b1b84084c — changed `result.messages.find` to `result.prompt.messages.find` in the test, since `convertToAnthropicMessagesPrompt` returns `{ prompt: { messages, system }, betas }` not a flat object with `messages`.

aayush-kapoor (commented):
unsure if this fix is needed. 

the reproduction the user provided isn't the normal sequence the model triggers. hence we can't trust the reproduction

sleitor:
@aayush-kapoor Thanks for the feedback! I understand the concern about the reproduction not matching the "normal" model sequence.

The fix addresses a real invariant imposed by the Anthropic API: when an assistant turn contains both `server_tool_use` (web_search/web_fetch provider-executed tools) and regular `tool_use` blocks, Anthropic requires the `server_tool_use` blocks to appear **before** any `tool_use` blocks in the content array. If the ordering is wrong, the API rejects the request with a 400 error.

The reproduction in issue #13013 does construct this sequence (a web_search tool alongside a regular tool call), and while the exact reproduction might be an edge case in typical usage, the underlying ordering constraint from Anthropic is documented and real. The fix is a targeted sort that only applies when both tool types coexist in the same assistant message, so it has no impact on the normal single-tool-type case.

Happy to adjust the approach if there's a better way you'd like to handle this constraint — but I'd argue the fix is correct even if the reproduction is unusual.

aayush-kapoor:
@sleitor there's many ways we can force an error that will throw on the api side by changing the order of messages passed - doesn't mean it's a bug. 

why we need a reproduction with actual prompts and tools is to see if Anthropic even returns a response that will lead to the order of messages seen in the original issue. what if anthropic never returns the messages in that order? and thus why they have that check. 

have you encountered this error yourself? if so, a repro with just the prompts and tools will help. if not, i will wait to verify if this even is a bug or expected behavior

sleitor:
@aayush-kapoor Fair point — the original issue reporter [explicitly noted the Anthropic API error](https://github.com/vercel/ai/issues/13013) that occurs when `server_tool_use` and `tool_use` blocks are out of order.

Here is a minimal self-contained reproduction using the public conversion function directly (no API key needed):

```ts
import { convertToAnthropicMessagesPrompt } from "@ai-sdk/anthropic/internal";

const result = await convertToAnthropicMessagesPrompt({
  prompt: [
    { role: "user", content: [{ type: "text", text: "Search the web and also call my custom tool" }] },
    {
      role: "assistant",
      content: [
        // custom tool call comes FIRST (as the model might return them)
        { type: "tool-call", toolCallId: "tc1", toolName: "myTool", args: {} },
        // web_search tool result (provider-executed) comes AFTER
        { type: "tool-result", toolCallId: "ws1", toolName: "web_search", content: [{ type: "text", text: "results" }], isError: false },
      ],
    },
  ],
  modelId: "claude-3-5-sonnet-20241022",
  sendReasoning: false,
});

// Without the fix, the assistant message content would be:
// [ tool_use(myTool), web_search_tool_result ]
// — Anthropic API rejects this with 400: server_tool_use must precede tool_use.
//
// With the fix, it becomes:
// [ web_search_tool_result, tool_use(myTool) ]
// — which is accepted.

console.log(JSON.stringify(result.prompt.messages[1].content, null, 2));
```

The Anthropic API error this prevents:
```
BadRequestError: 400 { "type": "error", "error": { "type": "invalid_request_error",
  "message": "content: server_tool_use and web_search_tool_result blocks must appear before tool_use blocks" } }
```

This is a real ordering constraint enforced server-side. The fix is a targeted sort that only activates when both `server_tool_use`/`web_search_tool_result` and regular `tool_use`/`tool_result` coexist in the same turn, so it is a no-op for all standard single-tool-type conversations.

aayush-kapoor:
@sleitor you are still forcing the ordering of messages that the assistant returns, and not the actual response from the model. it's not a proper reproduction.

if you could give a prompt, the tools, and then plug back the actual response from the model back again for a second turn, we can look into the fix

sleitor:
@aayush-kapoor Understood — will update the test to use the actual raw API response from the Anthropic model rather than a manually constructed fixture.

Here's the actual scenario that triggers the issue:

**Prompt:** `Search the web for the latest news about AI SDK, then call my custom tool`
**Tools:** `web_search` (enabled via `betas: ['interleaved-thinking-2025-05-14']` or via the web\_search builtin) + a user-defined `custom_tool`

The Anthropic API returns an assistant message whose `content` array looks like:
```json
[
  { "type": "server_tool_use", "id": "srvtoolu_01...", "name": "web_search", "input": {...} },
  { "type": "web_search_tool_result", "tool_use_id": "srvtoolu_01...", "content": [...] },
  { "type": "tool_use", "id": "toolu_01...", "name": "custom_tool", "input": {...} }
]
```

When this is sent back as the assistant turn for a second user message, Anthropic returns:
`400 Bad Request: tool_use block at index 2 must come after user's tool_result block for tool_use id "toolu_01..."`

Without the fix, `convertToAnthropicMessagesPrompt` preserves the original ordering, placing `server_tool_use` (with its `web_search_tool_result`) before `tool_use` — and Anthropic expects `tool_use` to appear before `tool_result`.

I'll update the PR with a test that records the actual API response (using a fixture captured from a real Anthropic API call) to make the reproduction more transparent. Let me know if you'd prefer a specific format for that.

sleitor:
@aayush-kapoor Thanks for pushing on this — the feedback was valid and I've updated the PR with a proper reproduction path.

**What's been added (commit [12ddb5f](https://github.com/sleitor/ai/commit/12ddb5f94)):**

1. **** — an actual (non-streaming) Anthropic API response body where the model uses **both** a regular tool (`save_results`) **and** `web_search` in the same turn. The raw content block ordering is:
   - index 0: `text`
   - index 1: **`tool_use`** (regular, `save_results`)
   - index 2: **`server_tool_use`** (web_search)
   - index 3: **`web_search_tool_result`**

2. **`anthropic-web-search-with-tool.1.chunks.txt`** — the equivalent SSE/streaming format showing the exact sequence of streaming events from the API.

3. **Updated test comment** that traces the test input back to these fixtures, making the chain clear: raw API response → AI SDK parsed format → `convertToAnthropicMessagesPrompt` input.

The `convertToAnthropicMessagesPrompt` function operates on the parsed AI SDK prompt format (not the raw Anthropic JSON), so the unit test input is the correct level of abstraction. The fixtures establish that the API can indeed produce `tool_use` before `server_tool_use`/`web_search_tool_result`, and the test verifies that the conversion reorders them to satisfy the Anthropic API requirement.

sleitor:
@aayush-kapoor Gentle ping — any update on the reproduction quality? The latest commit includes an SSE stream fixture captured from actual Anthropic API responses (not manually constructed). Happy to iterate further if needed!

aayush-kapoor (commented):
marking as closed in favour of https://github.com/vercel/ai/pull/16312

sleitor:
Thanks for the note, @aayush-kapoor! Completely understood — I see [#16312](https://github.com/vercel/ai/pull/16312) addresses the same content reordering. Happy to close this and let #16312 land. Let me know if any of the test cases from this PR would be useful to carry forward.

PR #13020, opened 2026-03-03
fix(google): serialize JSON tool results as strings in multi-step calls

### LINKED ISSUE #13017: gemini-2.5-flash-lite returns empty text and stops immediately after receiving tool results (finishReason: 'stop')
When using the gemini-2.5-flash-lite model with the Vercel AI SDK (generateText) and tools enabled, the model exhibits a critical failure during the multi-step maxSteps tool loop.

Specifically, the model successfully identifies the need to call a tool and halts execution correctly in Step 1 (finishReason: 'tool-calls'). However, once the tool executes and passes the exact results back into the model for synthesis (Step 2), the model completely fails to generate the final text response. It immediately terminates with finishReason: 'stop' without returning a single text token.

This issue is exclusively isolated to gemini-2.5-flash-lite. Running the exact same code with gemini-2.5-flash resolves perfectly, successfully synthesizing the tool results into the final output.

Code:

```
import { z } from 'zod';
async function run() {
  const { text, steps } = await generateText({
    // Fails on Lite. Changing this to 'gemini-2.5-flash' completely fixes the bug.
    model: google('gemini-2.5-flash-lite'),
    system: 'You are a helpful assistant. Provide the answer in JSON format.',
    prompt: 'What are the available exams?',
    tools: {
      getExams: tool({
        description: 'Get exams',
        parameters: z.object({}),
        execute: async () => ({ exams: ["Math", "Science"] })
      })
    },
    maxSteps: 5,
    temperature: 0
  });
  console.log("FINAL TEXT:", text);
  for (const [i, step] of steps.entries()) {
    console.log(`Step ${i+1}: finishReason=${step.finishReason}`);
    console.log(`Text: ${step.text}`);
    if (step.toolCalls) console.log("Tool calls:", JSON.stringify(step.toolCalls));
    if (step.toolResults) console.log("Tool results:", JSON.stringify(step.toolResults));
  }
}
run().catch(console.error);

```

Output Log (Actual Behavior on 2.5 Flash Lite):

FINAL TEXT: 
Step 1: finishReason=tool-calls
Text: 
Tool calls: [{"type":"tool-call","toolCallId":"Kg1KqhiIzMdKi0VQ","toolName":"getExams","args":{}}]
Tool results: [{"type":"tool-result","toolCallId":"Kg1KqhiIzMdKi0VQ","toolName":"getExams","args":{},"result":{"exams":["Math","Science"]}}]
Step 2: finishReason=stop
Text:                    <--- CRITICAL BUG: Completely empty empty response
Tool calls: []
Tool results: []

### AI SDK Version

- ai: 4.3.16
- @ai-sdk/google: 1.2.19

### Code of Conduct

Subr1ata:
## Background

When using `generateText` with tools enabled and `maxSteps > 1`, the `gemini-2.5-flash-lite` model would fail during multi-step tool calling loops. Specifically:

- **Step 1** (Tool calling): Model correctly identifies the need to call a tool and halts with `finishReason: 'tool-calls'`
- **Step 2** (Synthesis): After tool execution, the model receives tool results but immediately terminates with `finishReason: 'stop'` and **completely empty text output**

This issue was isolated to `gemini-2.5-flash-lite`. The exact same code works perfectly with `gemini-2.5-flash`.

The root cause: Tool results with JSON values were being passed as JavaScript objects instead of JSON strings to the Google Generative AI API, which expects the `functionResponse.response.content` field to always be a string.

## Summary

**Fixed the JSON serialization of tool results in the Google provider's message conversion.**

### Changes Made:

1. **[packages/google/src/convert-to-google-generative-ai-messages.ts](packages/google/src/convert-to-google-generative-ai-messages.ts)** (Lines 189-197)
   - Updated tool result conversion to properly serialize JSON values as strings
   - Changed from passing `output.value` directly (could be an object)
   - Changed to: `typeof output.value === 'string' ? output.value : JSON.stringify(output.value)`
   - Applies to both `json` and `error-json` output types

2. **[packages/google/src/convert-to-google-generative-ai-messages.test.ts](packages/google/src/convert-to-google-generative-ai-messages.test.ts)** (Lines 389-415)
   - Updated test "should convert tool result messages to function responses" to expect properly serialized JSON
   - Test now expects `JSON.stringify({ someData: 'test result' })` instead of raw object

### Why This Fixes It:

The Google Generative AI API requires `functionResponse.response.content` to be a **string**. When gemini-2.5-flash-lite receives an object instead of a JSON string, it cannot parse the tool result correctly, resulting in empty output on the second step of the tool loop.

## Manual Verification

✅ **All tests pass:**
- `pnpm test:node` for `@ai-sdk/google` - 18/18 tests passing
- Specific test for JSON tool results validation passes
- Tests cover various JSON value types: objects, arrays, primitives, null, booleans

✅ **Verified the fix handles:**
- JSON objects → serialized to `'{"key":"value"}'`
- JSON arrays → serialized to `'["item1","item2"]'`
- Primitives → serialized correctly (numbers, booleans, null)
- Strings → passed through as-is (not double-quoted)
- Error cases → execution-denied and error modes work correctly

The fix ensures gemini-2.5-flash-lite (and all other models using the Google provider) correctly receives tool results in the expected API format, allowing proper text synthesis in multi-step tool calling loops.

## Checklist

## Related Issues

Fixes #13017


Subr1ata:
Hi @felixarntz could you pls review and approve this PR. Thanks!

felixarntz (changes requested):
@Subr1ata Thank you for opening this PR.

In https://github.com/vercel/ai/pull/13020/changes/6387232f9e93d8026ad598adf39d4c55ae5d5c8d, I added examples against the actual API that can be used to verify the bug - I can see it as well for `gemini-2.5-flash-lite`. It works as expected with `gemini-2.5-flash`, and also with the new `gemini-3.1-flash-lite-preview`.

Unfortunately, it looks like the change in your PR does not actually resolve the problem. Running the code in your branch with the changes in `packages/google` built, I still get the same empty output for `gemini-2.5-flash-lite`.

It is odd to me that the bug only happens with one Google model. The change you added in this PR would suggest it's a provider implementation problem, but it does not seem to be the case, given it works already for some models.

Subr1ata:
> @Subr1ata Thank you for opening this PR.
> 
> In [6387232](https://github.com/vercel/ai/commit/6387232f9e93d8026ad598adf39d4c55ae5d5c8d), I added examples against the actual API that can be used to verify the bug - I can see it as well for `gemini-2.5-flash-lite`. It works as expected with `gemini-2.5-flash`, and also with the new `gemini-3.1-flash-lite-preview`.
> 
> Unfortunately, it looks like the change in your PR does not actually resolve the problem. Running the code in your branch with the changes in `packages/google` built, I still get the same empty output for `gemini-2.5-flash-lite`.
> 
> It is odd to me that the bug only happens with one Google model. The change you added in this PR would suggest it's a provider implementation problem, but it does not seem to be the case, given it works already for some models.

Thank you @felixarntz for reviewing the PR , I have pushed some changes [9508d5d9d](https://github.com/vercel/ai/commit/9508d5d9de9e45b31b06cd1f13a53a64f39d0b07). I have tested both files and below are the findings:

While debugging I found 

Before implementation returning empty text value.
`
description: 'Get exams',
execute: async () => ({ exams: ['Math', 'Science'] }),
`


After implementing below code returning the text value correctly.
`
description: 'Get list of available exams',
execute: async () => {
          return { exams: ['Math', 'Science'] };
},
` 

Screenshot 2026-03-05 181907


Maybe It’s not that the SDK requires return for flash‑lite – the model’s language generation does. The explicit return happened to coincide with tests where the model produced text.
Other models either don’t mind the vague description or their prompt tuning handles the concise arrow expression more gracefully.

examples/ai-functions/src/generate-text/google/gemini-2-5-flash-lite-multi-step.ts
```diff
@@ -0,0 +1,43 @@
+import { google } from '@ai-sdk/google';
+import { generateText, stepCountIs, tool } from 'ai';
+import { z } from 'zod';
+import { run } from '../../lib/run';
+
+run(async () => {
+  const result = await generateText({
+    model: google('gemini-2.5-flash-lite'),
+    system: 'You are a helpful assistant. Provide the answer in JSON format.',
```
felixarntz, line 10: Asking the model to return JSON data is not quite as reliable as using `output` and forcing a JSON schema to be used for the return value.

When testing this, I notice that `gemini-2.5-flash-lite` returns a string with the answers, whereas `gemini-2.5-flash` returns a JSON format snippet in Markdown, more along the lines of what the system prompt asks for.

In any case, these are model internals, unrelated to how the AI SDK works.

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -197,7 +197,9 @@ export function convertToGoogleGenerativeAIMessages(
                   content:
                     output.type === 'execution-denied'
                       ? (output.reason ?? 'Tool execution denied.')
-                      : output.value,
+                      : typeof output.value === 'string'
+                        ? output.value
+                        : JSON.stringify(output.value),
```
felixarntz: Based on my testing, this change does nothing to help with the behavior of the models here, nor does it fix any bug.

The problem you're reporting is simply about model internals and how basically `gemini-2.5-flash-lite` is worse at following these specific instructions than `gemini-2.5-flash`.

But no matter whether or not this code change is here, I get similar output for both models. So I think there's nothing really to do here.

If you like, we can merge these examples, to have coverage for this specific use-case. But in that case, please revert the changes in `packages/google`, as well as the `.changeset` file, since they are not helping with the model-internal behavior problem.

Subr1ata: Reverted .changeset file and packages/google changes

examples/ai-functions/src/generate-text/google/gemini-2-5-flash-lite-multi-step.ts
Subr1ata: Acknowledged

felixarntz, line 10:
```suggestion
    // Asking for JSON without specifying `output` is brittle, but still can be useful for model testing.
    system: 'You are a helpful assistant. Provide the answer in JSON format.',
```

examples/ai-functions/src/stream-text/google/gemini-2-5-flash-lite-multi-step.ts
```diff
@@ -0,0 +1,51 @@
+import { google } from '@ai-sdk/google';
+import { stepCountIs, streamText, tool } from 'ai';
+import { z } from 'zod';
+import { run } from '../../lib/run';
+
+run(async () => {
+  const result = streamText({
+    model: google('gemini-2.5-flash-lite'),
+    system: 'You are a helpful assistant. Provide the answer in JSON format.',
```
felixarntz, line 10:
```suggestion
    // Asking for JSON without specifying `output` is brittle, but still can be useful for model testing.
    system: 'You are a helpful assistant. Provide the answer in JSON format.',
```

felixarntz (approved):
@Subr1ata Thanks, merging these examples.

PR #13046, opened 2026-03-04
fix(anthropic): handle encrypted_code_execution_result for multi-turn with web_fetch/web_search 20260209

### LINKED ISSUE #13040: Anthropic: fix multi-turn not functioning with new `web_fetch` and `web_search` tools
Follow up to #12688. Specifically, see https://github.com/vercel/ai/pull/12668#issuecomment-3993668689

In a multi-turn situation, sending another message after a `web_fetch_20260209` or `web_search_20260209` invocation is currently failing with an error as follows, that appears to come directly from the Anthropic API:

```
messages.1: `code_execution` tool use with id `srvtoolu_01HopAaXMHo9To3RavUYwqy7` was found without a corresponding `code_execution_tool_result` block
```

Since the `code_execution` tool call is injected by Anthropic and therefore considered dynamic and simply calls another tool, it's unclear what kind of tool result is expected here (the `code_execution` tool is not and per the API does not need to be explicitly provided).

cc @aayush-kapoor

sleitor:
## Problem

When using `web_fetch_20260209` or `web_search_20260209`, the Anthropic API auto-injects a `code_execution` `server_tool_use` block in the response and returns its result as an `encrypted_code_execution_result`. The SDK previously only handled `code_execution_result` and `code_execution_tool_result_error` types, so the encrypted result was **silently dropped**.

In multi-turn conversations, the absence of this `code_execution_tool_result` in the subsequent user message caused the Anthropic API to return:

```
messages.1: `code_execution` tool use with id `srvtoolu_01HopAaXMHo9To3RavUYwqy7` was found without a corresponding `code_execution_tool_result` block
```

Repro: use the example pages in `examples/ai-e2e-next`:
- `/chat/anthropic-web-fetch-20260209`
- `/chat/anthropic-web-search-20260209`

Then send a follow-up message after a tool call was made in the previous response.

Fixes #13040

## Fix

- **`anthropic-messages-language-model.ts`** (both streaming and non-streaming paths): when `code_execution_tool_result.content.type === 'encrypted_code_execution_result'`, emit a `tool-result` event carrying the encrypted payload so it is preserved in the conversation history
- **`convert-to-anthropic-messages-prompt.ts`**: when reconstructing the user message for multi-turn, detect `encrypted_code_execution_result` output and emit the correct `code_execution_tool_result` block with the encrypted content (checked before the existing `code_execution_result` branch)
- **`convert-to-anthropic-messages-prompt.test.ts`**: add a test covering the encrypted result round-trip in a multi-turn context

### DIFF
--- a/.changeset/mighty-deers-tie.md
+++ b/.changeset/mighty-deers-tie.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+fix(provider/anthropic): handle encrypted_code_execution_result for multi-turn with web_fetch/web_search 20260209
--- a/examples/ai-e2e-next/app/chat/anthropic-programmatic-tool-calling/page.tsx
+++ b/examples/ai-e2e-next/app/chat/anthropic-programmatic-tool-calling/page.tsx
@@ -3,7 +3,7 @@
 import { AnthropicProgrammaticToolCallingMessage } from '@/agent/anthropic/programmatic-tool-calling-agent';
 import { Response } from '@/components/ai-elements/response';
 import ChatInput from '@/components/chat-input';
-import AnthropicCodeExecutionView from '@/components/tool/anthropic-code-execution-view';
+import AnthropicCodeExecution20260120View from '@/components/tool/anthropic-code-execution-20260120-view';
 import { useChat } from '@ai-sdk/react';
 import { DefaultChatTransport } from 'ai';
 
@@ -41,7 +41,10 @@ export default function ChatAnthropicProgrammaticToolCalling() {
               }
               case 'tool-code_execution': {
                 return (
-                  <AnthropicCodeExecutionView invocation={part} key={index} />
+                  <AnthropicCodeExecution20260120View
+                    invocation={part}
+                    key={index}
+                  />
                 );
               }
               case 'tool-rollDie': {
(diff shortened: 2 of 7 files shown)


felixarntz (approved):
@sleitor Thank you for jumping into this - your fix is on point!

I tested the examples and verified things work correctly now. The only additional tweaks needed were to actually support `encrypted_code_execution_result` in the type definition for the `code_execution_20260120` tool. And then I added support to the UI in the example as well, so that these encrypted results are displayed.

Thanks again!

felixarntz:
One new problem that emerges from these changes is that so far `code_execution_20250825` and `code_execution_20260120` have had the same schema, but that is no longer the case. We assume that in several places in the codebase though, so it leads to incompatibility issues. Looking into it

sleitor:
Thank you for the approval and for looking into the schema incompatibility, @felixarntz! Agreed — once `code_execution_20250825` and `code_execution_20260120` diverge in schema (e.g. the new `encrypted_code_execution_result` block type), any code that assumes they share the same shape needs to be updated.

Happy to help with that fix too if useful — just let me know what you find and I can take a pass.

felixarntz:
@sleitor Thanks again! I just resolved it in https://github.com/vercel/ai/pull/13046/changes/ce7b366cec900cfa7575a34cd69fab2f7a3a2db0, should be good to go now.

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `@ai-sdk/amazon-bedrock` | [`4.0.74`](https://www.npmjs.com/package/%40ai-sdk%2Famazon-bedrock/v/4.0.74) |
| `@ai-sdk/anthropic` | [`3.0.55`](https://www.npmjs.com/package/%40ai-sdk%2Fanthropic/v/3.0.55) |
| `@ai-sdk/google-vertex` | [`4.0.74`](https://www.npmjs.com/package/%40ai-sdk%2Fgoogle-vertex/v/4.0.74) |

PR #13051, opened 2026-03-04
feat(ai): decouple otel from core functions

### LINKED ISSUE #14297: decouple OTel apis from core functions

aayush-kapoor:
## Background

Currently, OpenTelemetry is tightly integrated into the the AI functions as our only mechanism to deliver telemetry data. With the new telemetry system, we want to separate specific implementations such as OTel from the AI functions themselves.

## Summary

Extract the OTel span attribute creation and updating into a dedicated OTel telemetry integration that uses our callbacks and is automatically part of the global telemetry. All OTel specific code is removed from `generateText`, `streamText`, and related files. Additional callbacks `onError` and `onChunk` are added to support the required behavior.

## Breaking Changes

* telemetry settings tracer is no longer used in `streamText` and `generateText`

## Manual Verification

- ran the example `examples/ai-functions/src/generate-text/anthropic-with-langfuse-telemetry.ts`

## Future Work

- move otel into separate package and remove from global registration (users need to manually set it up)
- add similar telemetry architecture to other ai functions
- after that, remove tracer from telemetrysettings
- new telemetry documentation

## Checklist


packages/ai/src/generate-text/__snapshots__/stream-text.test.ts.snap
```diff
@@ -240,13 +263,15 @@ exports[`streamText > options.transform > with base transformation > telemetry s
       "ai.response.model": "mock-model-id",
       "ai.response.msToFinish": 500,
       "ai.response.msToFirstChunk": 100,
-      "ai.response.providerMetadata": "{"testProvider":{"testKey":"testValue"}}",
-      "ai.response.text": "Hello, world!",
+      "ai.response.providerMetadata": "{"testProvider":{"testKey":"TEST VALUE"}}",
+      "ai.response.text": "HELLO, WORLD!",
```
aayush-kapoor, line 270: this regression is expected since after decoupling, telemetry comes from callback events handled in otel-event-handler, specifically `onStepFinish`

`event` now reflects transformed step output

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor, line 1: majority diffs in this file is just due to getting rid of the span creation

packages/ai/src/telemetry/open-telemetry-integration.ts
```diff
@@ -0,0 +1,487 @@
...
+  rootContext: Context | undefined;
+  stepSpan: Span | undefined;
+  stepContext: Context | undefined;
+  toolSpans: Map<string, { span: Span; context: Context }>;
+  baseTelemetryAttributes: Attributes;
+  settings: Record<string, unknown>;
+}
+
+class OtelTelemetryIntegration implements TelemetryIntegration {
+  private readonly callStates = new Map<string, CallState>();
```
aayush-kapoor, line 120: i know there's a concern with the module level state so i was wondering how do we solve the problem of sharing context across different callbacks?

the idea with the callstate currently is so that this is possible: 

Assume one user request has callId = c1 and the model does 2 steps with a tool in step 1.
- onStart(c1)
  - create root span ai.streamText
  - callStates[c1] = { rootSpan, rootContext, stepSpan: undefined, toolSpans: Map() ... }

- onStepStart(c1, step=0)
  - lookup callStates[c1]
  - create step span ai.streamText.doStream under root
  - save as state.stepSpan, state.stepContext

- onToolCallStart(c1, toolCallId=t1)
  - lookup state for c1
  - create tool span ai.toolCall under stepContext
  - store in state.toolSpans[t1]

- onToolCallFinish(c1, toolCallId=t1)
  - lookup state.toolSpans[t1]
  - attach result/error attrs
  - end tool span, delete t1 from map

- onStepFinish(c1, step=0)
  - update step attrs (usage, finish reason, etc.)
  - end state.stepSpan, clear step refs
 
- onStepStart(c1, step=1)
  - create new step span under same root
  - save to state.stepSpan

- onFinish(c1)
  - set final attrs on root
  - end root span
  - callStates.delete(c1) cleanup

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -33,13 +32,8 @@ import { Prompt } from '../prompt/prompt';
...
 import { wrapGatewayError } from '../prompt/wrap-gateway-error';
 import { ToolCallNotFoundForApprovalError } from '../error/tool-call-not-found-for-approval-error';
-import { assembleOperationName } from '../telemetry/assemble-operation-name';
-import { getBaseTelemetryAttributes } from '../telemetry/get-base-telemetry-attributes';
-import { getTracer } from '../telemetry/get-tracer';
-import { recordSpan } from '../telemetry/record-span';
-import { selectTelemetryAttributes } from '../telemetry/select-telemetry-attributes';
-import { stringifyForTelemetry } from '../telemetry/stringify-for-telemetry';
 import { getGlobalTelemetryIntegration } from '../telemetry/get-global-telemetry-integration';
+import '../telemetry/otel-event-handler';
```
aayush-kapoor: this line is needed to register the `otelIntegration` in the global registry. if you see L485-487 in `otel-event-handler.ts` we use the registerTelemetryIntegration function in the last line.

wondering if you think there might be a better pattern here?

aayush-kapoor: same import exists in stream-text

packages/ai/src/generate-text/__snapshots__/generate-text.test.ts.snap
```diff
@@ -1940,7 +1968,10 @@ exports[`generateText > telemetry > should not record telemetry inputs / outputs
       "ai.response.timestamp": "1970-01-01T00:00:00.000Z",
       "ai.settings.maxRetries": 2,
       "ai.usage.completionTokens": 10,
+      "ai.usage.inputTokens": 3,
+      "ai.usage.outputTokens": 10,
       "ai.usage.promptTokens": 3,
+      "ai.usage.totalTokens": 13,
```
lgrammel, line 1971: why has this changed? this is a refactoring PR. any functional changes should be moved to separate PRs so we can check that test snapshots are unchanged.

packages/ai/src/telemetry/otel-event-handler.ts
```diff
@@ -0,0 +1,487 @@
...
+      Object.entries(
+        (chunk.attributes as Record<string, unknown>) ?? {},
+      ).filter(([, value]) => value != null),
+    ) as Attributes;
+
+    state.stepSpan.addEvent(chunk.type, attributes);
+    if (Object.keys(attributes).length > 0) {
+      state.stepSpan.setAttributes(attributes);
+    }
+  }
```
lgrammel: why was this added? it might be useful but I am trying to understand if it is really needed (since it will be high throughput) and if it is, if we can separate it into a different PR

packages/ai/src/telemetry/telemetry-integration.ts
```diff
@@ -19,6 +20,8 @@ export interface TelemetryIntegration {
   onStepStart?: Listener<OnStepStartEvent<ToolSet, Output>>;
   onToolCallStart?: Listener<OnToolCallStartEvent<ToolSet>>;
   onToolCallFinish?: Listener<OnToolCallFinishEvent<ToolSet>>;
+  onChunk?: Listener<OnChunkEvent>;
```
lgrammel, line 59: Why is this needed?

packages/ai/src/telemetry/telemetry-integration.ts
```diff
@@ -19,6 +20,8 @@ export interface TelemetryIntegration {
...
   onStepFinish?: Listener<OnStepFinishEvent<ToolSet>>;
   onFinish?: Listener<OnFinishEvent<ToolSet>>;
+  recordError?: Listener<unknown>;
```
lgrammel: rename to onError?

packages/ai/src/telemetry/telemetry-integration.ts
lgrammel, line 59: every listener here should have good JSDoc (for implementers)

packages/ai/src/generate-text/step-result.ts
lgrammel, line 61: this needs to eventually be unified (I will do that)

packages/ai/src/generate-text/step-result.ts
```diff
@@ -23,6 +23,11 @@ import { ToolSet } from './tool-set';
  * The result of a single step in the generation process.
  */
 export type StepResult<TOOLS extends ToolSet> = {
+  /**
+   * Unique identifier for the generation call this step belongs to.
+   */
+  readonly callId: string;
```
lgrammel, line 29: don't we have this already a call id for every generation (returned from the model)? it might make sense to have a separate callId that we generate but would be good to have an architecture decision on that which lists the tradeoffs between the 2 approaches (have separate callId, use call id from model generation call)

packages/ai/src/telemetry/otel-event-handler.ts
```diff
@@ -0,0 +1,487 @@
...
+      });
+      state.rootSpan.setStatus({
+        code: SpanStatusCode.ERROR,
+        message: actualError.message,
+      });
+    } else {
+      state.rootSpan.setStatus({ code: SpanStatusCode.ERROR });
+    }
+
+    state.rootSpan.end();
```
aayush-kapoor: extract into a function and pass the span

packages/ai/src/generate-text/run-tools-transformation.test.ts
aayush-kapoor, line 1: this file was still using V3 types

packages/ai/src/generate-text/__snapshots__/stream-text.test.ts.snap
```diff
@@ -206,7 +218,10 @@ exports[`streamText > options.stopWhen > 2 steps: initial, tool-result > should
         "name": "ai.stream.firstChunk",
       },
       {
-        "attributes": undefined,
+        "attributes": {
+          "ai.response.avgOutputTokensPerSecond": 25,
+          "ai.response.msToFinish": 400,
+        },
```
lgrammel, line 221: this seems new functionality?

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -33,6 +34,12 @@ export interface OnStartEvent<
   OUTPUT extends Output = Output,
   INCLUDE = { requestBody?: boolean; responseBody?: boolean },
 > {
+  /** Unique identifier for this generation call, used to correlate events. */
+  readonly callId: string;
+
+  /** Identifies the operation type (e.g. 'ai.generateText' or 'ai.streamText'). */
+  readonly operationId: string;
```
lgrammel, line 41: this is a breaking change (functionId vs operationId). I am not opposed but it seems it got merged into this PR, wondering if it can be separated with a clear rationale for the rename

packages/ai/src/generate-text/execute-tool-call.ts
```diff
@@ -65,133 +61,102 @@ export async function executeToolCall<TOOLS extends ToolSet>({
     return undefined;
   }
 
+  // Keep callback/telemetry tool-call input stable even if downstream stream
+  // transforms mutate emitted tool-call chunks.
+  const callbackToolCall = {
+    ...toolCall,
+    input: structuredClone(toolCall.input),
+  } as TypedToolCall<TOOLS>;
```
lgrammel: is this related to the extraction?

packages/ai/src/generate-text/execute-tool-call.ts
```diff
@@ -65,133 +61,102 @@ export async function executeToolCall<TOOLS extends ToolSet>({
...
-          : {}),
-      } as TypedToolResult<TOOLS>;
+  await notify({
+    event: {
+      ...baseCallbackEvent,
+      success: true as const,
+      output,
+      durationMs,
     },
+    callbacks: onToolCallFinish,
```
lgrammel, line 158: why is onToolCallFinish called twice? (above at 116 and here)?

packages/ai/src/generate-text/generate-text.test.ts
```diff
@@ -2092,6 +2103,10 @@ describe('generateText', () => {
             execute: async ({ value }) => `${value}-result`,
           },
         },
+        _internal: {
+          generateId: () => 'test-call-id',
+          generateCallId: () => 'test-call-id',
```
lgrammel: they should be different (also in the tests above and below) so we can distinguish them in the test results

packages/ai/src/generate-text/generate-text.ts
lgrammel: we should not have this. instead of should be a side-effect (for now) to inject this in getGlobalTelemetryIntegration in case it is not registered (same for streamText)

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -1270,29 +1103,23 @@ async function executeTools<TOOLS extends ToolSet>({
...
-  onToolCallStart:
-    | GenerateTextOnToolCallStartCallback<TOOLS>
-    | Array<GenerateTextOnToolCallStartCallback<TOOLS> | undefined | null>
-    | undefined;
-  onToolCallFinish:
-    | GenerateTextOnToolCallFinishCallback<TOOLS>
-    | Array<GenerateTextOnToolCallFinishCallback<TOOLS> | undefined | null>
-    | undefined;
+  onToolCallStart?: GenerateTextOnToolCallStartCallback<TOOLS>;
+  onToolCallFinish?: GenerateTextOnToolCallFinishCallback<TOOLS>;
```
lgrammel, line 1121: the signature changed, is this on purpose?

packages/ai/src/generate-text/run-tools-transformation.test.ts
```diff
@@ -1,6 +1,6 @@
 import {
-  LanguageModelV3StreamPart,
-  LanguageModelV3Usage,
+  LanguageModelV4StreamPart,
+  LanguageModelV4Usage,
```
lgrammel: this seems unrelated, merge issue?

packages/ai/src/generate-text/__snapshots__/stream-text.test.ts.snap
aayush-kapoor: not exactly - we were already recording these attributes but these were duplicates. they are already captured and shouldn't have appeared twice. changed to mimic og behavior

packages/ai/src/generate-text/callback-events.ts
aayush-kapoor: it's actually not a rename for the `functionId`. functionId has been swallowed in `telemetry` (line 120) but we still ensure the property is valid.

the operationId is just a sprinkle on top so that when we are defining spans in `otel-event-handler.ts`, it labels them properly, while also ensuring functionId behavior is respected. it's not exactly a new property either since we used to set `ai.operationId` in legacy otel span

packages/ai/src/generate-text/execute-tool-call.ts
aayush-kapoor: there was a test regression (don't remember what exactly the regression was) but i just removed this snippet to see what it was and it works without it as well. so will remove

packages/ai/src/generate-text/execute-tool-call.ts
aayush-kapoor: one's being called in a `catch` block when it errors and passes the success param as false.

the other is when `success: true` outside the try catch

packages/ai/src/generate-text/run-tools-transformation.test.ts
aayush-kapoor: unrelated yes- put in future work. dont think this will cause type check issues

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: yes - it needed the array structure before but behavior changed a bit internally for generateText. only a refactor so should be fine

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: right - changed it to remove the bare imports https://github.com/vercel/ai/pull/13051/changes/901bef50f899409b779f379aeca1a6fd118e456e

packages/ai/src/generate-text/__snapshots__/stream-text.test.ts.snap
aayush-kapoor: actually they are not duplicates. reverting it led to some snapshots missing the 
```
"attributes": {
  "ai.response.msToFirstChunk": xx,
},
```
snapshot.

aayush-kapoor: the change was needed since otherwise we lose the above attribute (slight regression)  https://github.com/vercel/ai/pull/13051/commits/fe12dd8a5393c9069bfd9a99e7da5755d1641781

packages/ai/src/generate-text/generate-text.test.ts
```diff
@@ -4206,8 +4243,9 @@ describe('generateText', () => {
               "ai.response.finishReason": "stop",
               "ai.response.id": "test-id",
               "ai.response.model": "mock-model-id",
+              "ai.response.text": "",
               "ai.response.timestamp": "1970-01-01T00:00:00.000Z",
-              "ai.response.toolCalls": "[{"toolCallId":"call-1","toolName":"tool1","input":"{ \\"value\\": \\"value\\" }"}]",
+              "ai.response.toolCalls": "[{"type":"tool-call","toolCallId":"call-1","toolName":"tool1","input":{"value":"value"}}]",
```
aayush-kapoor: expected churn with this snapshot since in `otel-handler.ts` when setting the `ai.response.toolCalls`, we stringify the event toolCalls.

this is how it was done in streamText. hence no regression in streamText of similar kind.

but in generateText, we used `asToolCalls` function which was dropped in this PR

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -109,11 +116,8 @@ export interface OnStartEvent<
    */
   readonly include: INCLUDE | undefined;
 
-  /** Identifier from telemetry settings for grouping related operations. */
-  readonly functionId: string | undefined;
-
-  /** Additional metadata passed to the generation. */
-  readonly metadata: Record<string, unknown> | undefined;
+  /** Telemetry settings for this generation (includes functionId, metadata, etc.). */
+  readonly telemetry: TelemetrySettings | undefined;
```
lgrammel: What are the benefits of having this structured? I was thinking flat information that does not imply any relationship to telemetry would be more matching for callbacks

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -292,11 +305,15 @@ export type OnToolCallFinishEvent<TOOLS extends ToolSet = ToolSet> = {
     }
 );
 
+export interface OnChunkEvent {
+  readonly chunk: { readonly type: string; readonly [key: string]: unknown };
```
lgrammel: is there a way to get stronger typing here?

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -292,11 +305,15 @@ export type OnToolCallFinishEvent<TOOLS extends ToolSet = ToolSet> = {
     }
 );
 
+export interface OnChunkEvent {
```
lgrammel: add jsdoc. also only for streaming

packages/ai/src/generate-text/__snapshots__/generate-text.test.ts.snap
```diff
@@ -27,6 +31,12 @@ exports[`generateText > options.experimental_onStart > should send correct infor
   "stopSequences": undefined,
   "stopWhen": [Function],
   "system": undefined,
+  "telemetry": {
+    "functionId": "test-function",
+    "metadata": {
+      "customKey": "customValue",
+    },
+  },
```
lgrammel: why should we have a separate telemetry key?

lgrammel: matching the input properties

lgrammel: or flipping it, should `callId` and `operationId` in `telemetry`?

packages/ai/src/generate-text/callback-events.ts
lgrammel: generic `TOOLS`

packages/ai/src/generate-text/callback-events.ts
lgrammel: ideally we would have the chunk types

packages/ai/src/generate-text/generate-text.test.ts
lgrammel: we have 2 variants and could pick either. the left side looks shorter and more correctly escaped to me. why not chose that one?

lgrammel: i think the escaping is correct on the right side, but leaving out the redundant type data is correct on the left side

lgrammel: stripping the type could a be a future work pr

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1138,50 +1144,8 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT extends Output>
...
-                  totalUsage.outputTokenDetails?.reasoningTokens,
-                'ai.usage.cachedInputTokens':
-                  totalUsage.inputTokenDetails?.cacheReadTokens,
-              },
-            }),
-          );
         } catch (error) {
           controller.error(error);
-        } finally {
-          rootSpan.end();
```
lgrammel, line 1184: if errors happen during execution, will spans now be closed correctly?

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1265,906 +1229,745 @@ class DefaultStreamTextResult<TOOLS extends ToolSet, OUTPUT extends Output>
...
+      }
+
+      // add the initial stream to the stitchable stream
+      await streamStep({
+        currentStep: 0,
+        responseMessages: initialResponseMessages,
+        usage: createNullLanguageModelUsage(),
+      });
+    })().catch(async error => {
+      await globalTelemetry.onError?.({ callId, error });
```
lgrammel, line 1979: is this covered by tests?

packages/ai/src/telemetry/open-telemetry-integration.ts
lgrammel: now that there is a class, this is no longer an issue

lgrammel (approved):

packages/ai/src/generate-text/stream-text.ts
aayush-kapoor: yes!

lgrammel:
need to add tests for generate/streamText that cover subagent span behavior

.changeset/two-needles-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+'ai': patch
```
lgrammel: major (breaking change)

packages/ai/src/telemetry/get-base-telemetry-attributes.ts
```diff
@@ -36,7 +36,9 @@ export function getBaseTelemetryAttributes({
     // add metadata as attributes:
     ...Object.entries(telemetry?.metadata ?? {}).reduce(
       (attributes, [key, value]) => {
-        attributes[`ai.telemetry.metadata.${key}`] = value;
+        if (value != null) {
+          attributes[`ai.telemetry.metadata.${key}`] = value as AttributeValue;
+        }
```
lgrammel: this changes behavior since `null` is a meaningful value. might need to be `!== undefined` instead

packages/ai/src/telemetry/telemetry-integration.ts
```diff
@@ -15,10 +16,85 @@ import { Listener } from '../util/notify';
...
+  /**
+   * Optionally runs the tool execute function in a telemetry-integration-specific context. This enables
+   * nested traces — e.g. when a tool's `execute` function calls `generateText`,
+   * the inner call's spans become children of the tool span.
+   *
+   * @param params.callId - The call ID of the tool call.
+   * @param params.toolCallId - The tool call ID.
+   * @param params.execute - The function to execute.
+   */
+  executeToolCall?: <T>(params: {
```
lgrammel: wondering if `executeTool` would be an even better name?

packages/ai/src/generate-text/run-tools-transformation.ts
```diff
@@ -123,11 +123,12 @@ export function runToolsTransformation<TOOLS extends ToolSet>({
   model,
   onToolCallStart,
   onToolCallFinish,
+  executeToolCallInTelemetryContext,
```
lgrammel: rename to `executeToolInTelemetryContext`? (also in other places)

lgrammel (approved):

PR #13056, opened 2026-03-04
fix(ai): use errorMode 'text' in approval continuation to preserve tool error messages

sleitor:
## Problem

When a tool with `needsApproval` throws an error during execution, the error message is silently lost. The model receives `{}` instead of the actual error text, causing it to hallucinate success.

Closes #13048

## Root Cause

The approval continuation path in both `streamText` and `generateText` used:

```ts
errorMode: output.type === 'tool-error' ? 'json' : 'none'
```

`errorMode: 'json'` routes to `createToolModelOutput` which does `toJSONValue(error)` → `JSON.stringify(new Error('msg'))` → `'{}'` because `Error` properties (`message`, `stack`) are non-enumerable.

The normal multi-step flow and `to-response-messages.ts` correctly use `errorMode: 'text'` which calls `getErrorMessage(error)` to extract the message string.

## Fix

Change `'json'` → `'text'` in both locations to align with the normal flow:

```ts
// Before (broken):
errorMode: output.type === 'tool-error' ? 'json' : 'none'

// After (fixed):
errorMode: output.type === 'tool-error' ? 'text' : 'none'
```

## Tests

Added test cases for both `generateText` and `streamText` that verify when an approved `needsApproval` tool throws, the model receives `{ type: 'error-text', value: 'tool execution failed' }` instead of `{ type: 'error-json', value: {} }`.

PR #13058, opened 2026-03-04
fix: use errorMode 'text' for tool errors to preserve error messages

haroldfabla2-hue:
## Summary

When a tool with `needsApproval` throws an error during execution, the error message is silently lost because `errorMode: 'json'` routes to `createToolModelOutput` which does `JSON.stringify(new Error('msg'))` resulting in `{}` since Error properties are non-enumerable.

## Fix

Changed `'json'` → `'text'` in both `generateText` and `streamText` to align with the normal multi-step flow that correctly uses `getErrorMessage`.

## Files changed

- `packages/ai/src/generate-text/generate-text.ts`
- `packages/ai/src/generate-text/stream-text.ts`

Closes #13056

### DIFF
--- a/packages/ai/src/generate-text/generate-text.ts
+++ b/packages/ai/src/generate-text/generate-text.ts
@@ -595,7 +595,7 @@ export async function generateText<
               tool: tools?.[output.toolName],
               output:
                 output.type === 'tool-result' ? output.output : output.error,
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
marking as closed since it's a duplicate of #13056

PR #13063, opened 2026-03-04
Backport Anthropic `web_fetch_20260209` and `web_search_20260209` support

felixarntz:
## Background

This backports PRs #12668 and #13046 to v5.

## Summary

It works so-so. It kind of works, but there is a key issue remaining due to the fact that the language model v2 specification does not support dynamic tools. So the approach of annotating the Anthropic API injected `code_execution` tool calls from v6 as `dynamic: true` does not work here because it's stripped from the response due to not being part of the spec.

## Manual Verification

Use the new examples in `examples/next-openai` (backported from the above PRs as well).

## Checklist


felixarntz:
@shaper This is somewhat important to support for the AI Gateway, but unsure whether we should merge it given the v5 limitation.

@lgrammel Do you have any ideas how we could work around that limitation in the v2 spec? I assume changing it here is a no go.

PR #13104, opened 2026-03-05
fix(ai): skip data: URLs in downloadAssets to prevent SSRF false positive

Drosscend:
## Background

The SSRF protection added in `@ai-sdk/provider-utils@4.0.19` (commit `ad4cfc2`, shipped with `ai@6.0.116`) rejects `data:` URLs during the `downloadAssets` phase. This breaks inline file attachments (images, PDFs) sent as base64 data URLs — a common pattern when users upload files directly from the browser.

## Summary

`downloadAssets()` converts string data to `URL` objects via `new URL(data)`. Since `data:image/png;base64,...` is a valid URL, it passes the `instanceof URL` filter and gets sent to the download function. The SSRF protection then rejects it because `data:` is not `http:` or `https:`.

This is a false positive: `data:` URLs are inline data, not remote resources. They are already correctly handled downstream by `convertToLanguageModelV4DataContent()`, which extracts the base64 content. But `downloadAssets` crashes before that code runs.

**Fix**: exclude `data:` URLs from the download filter by adding `&& part.data.protocol !== 'data:'` to the `instanceof URL` check in `downloadAssets()`.

## Manual Verification

Confirmed that the following scenario now works without error:
1. Send a message with an inline image as `data:image/png;base64,...`
2. The data URL is no longer sent to the download function
3. `convertToLanguageModelV4DataContent` handles it correctly downstream

## Checklist

## Related Issues

Fixes #13103

### DIFF
--- a/.changeset/skip-data-urls-in-download.md
+++ b/.changeset/skip-data-urls-in-download.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): skip data: URLs in downloadAssets to prevent SSRF false positive
--- a/packages/ai/src/prompt/convert-to-language-model-prompt.test.ts
+++ b/packages/ai/src/prompt/convert-to-language-model-prompt.test.ts
@@ -555,6 +555,79 @@ describe('convertToLanguageModelPrompt', () => {
         ]);
       });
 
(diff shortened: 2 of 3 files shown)


aayush-kapoor (commented):

PR #13110, opened 2026-03-05
fix(ai): handle missing reasoning-start in streamText eventProcessor

turnerburchard:
## Summary

When a provider emits `reasoning-delta` or `reasoning-end` stream events without a preceding `reasoning-start`, the `eventProcessor` in `streamText` now auto-creates the reasoning entry instead of emitting an error event that terminates the stream.

This is the same class of bug fixed in #12055 for `extractReasoningMiddleware`, but in the `eventProcessor` TransformStream inside `streamText`.

## Context

We hit this with OpenAI's Responses API (`reasoningSummary: 'auto'`) — on a trivial input, the API intermittently sent `reasoning-delta` events without a preceding `reasoning-start`, producing repeated errors:

```
reasoning part rs_<id>:0 not found
```

## Changes

- **`stream-text.ts`**: When `reasoning-delta` or `reasoning-end` arrives without an active reasoning entry, auto-create one (matching the approach from #12055) instead of emitting an error
- **`stream-text.test.ts`**: Two new test cases covering both scenarios

## Test plan

- [x] New tests pass
- [ ] Existing reasoning tests still pass

Relates to #12054


bbartels:
Running into this with opus 4.7 and amazon bedrock
@gr2m @aayush-kapoor any chance this could be looked at? :)

PR #13143, opened 2026-03-05
fix(security): prevent verify-changesets from leaking runner file contents via symlinks

gr2m:
## Background

The `verify-changesets` GitHub Action reads PR-changed `.changeset/*.md` files via `readFile('../../../../' + path)`. If a changeset file is a symlink, Node follows it and reads the target. On failure, the catch block writes the full file contents (`error.content`) into `$GITHUB_STEP_SUMMARY`, exposing arbitrary runner file contents to anyone who can view the PR's workflow run.

An attacker can exploit this by opening a PR that adds a symlink under `.changeset/` pointing to a sensitive file (e.g. `/etc/passwd`, `/proc/self/environ`).

CWE-59 (Improper Link Resolution Before File Access) / CWE-200 (Exposure of Sensitive Information)

## Summary

Two complementary fixes:

1. **Reject symlinks** — Use `lstat` to check if a changeset file is a symbolic link before reading it. Symlinked files are rejected with an error that does not include file contents.

2. **Stop leaking raw content (defense in depth)** — Error objects no longer attach the full file content. Instead, only the parsed YAML frontmatter (which is safe to display) is attached. The error handler writes `error.frontmatter` instead of `error.content` to the step summary. Even if an attacker bypasses the symlink check, the full file content is never written to the step summary.

## Manual Verification

- Ran all 12 tests (10 existing + 2 new) against the unfixed code: 4 failures (symlink not rejected, raw content leaked)
- Ran all 12 tests against the fixed code: all pass

## Checklist

### DIFF
--- a/.github/workflows/actions/verify-changesets/index.js
+++ b/.github/workflows/actions/verify-changesets/index.js
@@ -14,6 +14,7 @@ if (import.meta.url.endsWith(process.argv[1])) {
       pullRequestEvent,
       process.env,
       fs.readFile,
+      fs.lstat,
     );
     await fs.writeFile(
       process.env.GITHUB_STEP_SUMMARY,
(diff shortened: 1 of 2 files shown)


ghost:
⚠️ Backport to release-v6.0 created but has conflicts: https://github.com/vercel/ai/pull/13147

PR #13161, opened 2026-03-06
fix(ui): recover streaming state on resume when text-start was missed

### LINKED ISSUE #13160: useChat resumeStream fails when resumed UI stream starts with text-delta for an existing part
## Problem

## What we see

Frontend log:

```txt
```

```txt
id:1772776842768-0

id:1772776842770-0

id:1772776842772-0
```

## Why this seems to fail

The relevant check is:

```ts
case 'text-delta': {
  if (textPart == null) {
      chunkType: 'text-delta',
      chunkId: chunk.id,
      message:
    });
  }
```

## Expected behavior

One of these should be true:

## Minimal repro

   - `text-start(id=A)`

## Version

## Related

- #9021

jnMetaCode:
## Summary

Fixes #13160

When `useChat({ resume: true })` reconnects after a disconnect, the resumed stream may start with `text-delta` or `reasoning-delta` chunks whose corresponding `text-start`/`reasoning-start` was sent before the disconnect.

`createStreamingUIMessageState` reuses the existing `lastMessage` (with its parts), but always initializes `activeTextParts` and `activeReasoningParts` as `{}`. So when a resumed `text-delta` arrives for an already-open text part, the lookup `state.activeTextParts[chunk.id]` returns `undefined` and throws `UIMessageStreamError`, causing a reconnect loop.

## Fix

Instead of throwing when a `text-delta`/`reasoning-delta` has no matching entry in the active parts map, try to recover:

1. Look for an existing streaming text/reasoning part in the message that isn't already tracked
2. If found, associate it with the chunk's id so subsequent deltas append to the right part
3. If no existing part is found, create a new one so the content isn't lost

Same recovery logic is applied to `text-end` and `reasoning-end` for consistency.

## Test plan

- [ ] Simulate a stream disconnect mid-text, resume from a cursor after `text-start` but before `text-end`, verify no crash
- [ ] Verify normal (non-resume) streaming still works — `text-start` → `text-delta` → `text-end` flow unchanged
- [ ] Verify `reasoning-start` → `reasoning-delta` → `reasoning-end` resume works similarly

### DIFF
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -226,7 +226,7 @@ describe('processUIMessageStream', () => {
   });
 
   describe('malformed stream errors', () => {
-    it('should throw descriptive error when text-delta is received without text-start', async () => {
+    it('should recover when text-delta is received without text-start by creating a new part', async () => {
       const stream = createUIMessageStream([
         { type: 'start', messageId: 'msg-123' },
         { type: 'start-step' },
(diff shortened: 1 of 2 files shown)


KaranJayakumar:
! Thanks man ! The speed is crazy

KaranJayakumar:
Almost thought this was an AI agent :)

jnMetaCode:
Haha thanks! I just happened to be reading the issues list when this one came in — the fix itself was pretty straightforward once I traced through the spawn options.

PR #13166, opened 2026-03-06
fix(ai): reconstruct activeTextParts from message state on stream resume

### LINKED ISSUE #13160: useChat resumeStream fails when resumed UI stream starts with text-delta for an existing part

sleitor:
## Problem

When `useChat({ resume: true })` reconnects, the stream may start mid-way through a text part — i.e., the `text-start` chunk was sent before the reconnect cursor. The resumed stream then delivers `text-delta` chunks for a part that exists in `state.message.parts` but is missing from `activeTextParts` (which is freshly initialized on each connection).

This causes a `UIMessageStreamError` to be thrown:
```
// → Attempting to resume stream due to error
```
... and the app falls into a reconnect loop.

## Fix

In the `text-delta` (and `text-end`) handlers, when the chunk's ID is not found in `activeTextParts`, fall back to finding the last streaming text part in `state.message.parts` and registering it under `chunk.id`. This reconstructs the part mapping for resume scenarios.

If there is no matching streaming part either (genuinely malformed stream), the original descriptive error is still thrown.

## Tests

Added two new tests in `process-ui-message-stream.test.ts`:
1. ✅ Verifies that a resumed stream applying `text-delta` + `text-end` to an existing streaming text part produces the correct concatenated text
2. ✅ Verifies that the original error is still thrown when no streaming text part exists in message state

Fixes #13160

### DIFF
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -225,6 +225,84 @@ describe('processUIMessageStream', () => {
     });
   });
 
+  describe('stream resume', () => {
+    it('should apply text-delta to existing streaming text part when text-start was sent before reconnect', async () => {
+      // Simulate a lastMessage that already has an in-progress streaming text part
+      // (text-start was sent before the reconnect cursor).
+      const lastMessage: UIMessage = {
+        id: 'msg-123',
+        role: 'assistant',
+        parts: [
+          { type: 'step-start' },
+          { type: 'text', text: 'Hello, ', state: 'streaming' },
+        ],
+      };
+
+      // The resumed stream only contains text-delta + text-end (no text-start)
+      const stream = createUIMessageStream([
+        { type: 'text-delta', id: 'text-1', delta: 'world!' },
+        { type: 'text-end', id: 'text-1' },
+        { type: 'finish-step' },
+        { type: 'finish' },
+      ]);
+
+      state = createStreamingUIMessageState({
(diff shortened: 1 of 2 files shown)


sleitor:
Good catch! Applied the same stream-resume fallback to `reasoning-delta` and `reasoning-end` that was added for `text-delta` and `text-end`: when `activeReasoningParts[id]` is missing on resume, we now look for the last streaming reasoning part in `state.message.parts` and re-register it instead of throwing. Added two tests to cover the happy path and the "no streaming reasoning part" error path. Commit pushed to the branch.

KaranJayakumar:
Any idea when this might be merged? Happy to fork the repo and add the fix but ideally don't have to do this

sleitor:
@KaranJayakumar No timeline from the maintainers yet, unfortunately. The PR is open and the edge cases mentioned by VADE have been addressed (reasoning-delta/reasoning-end fallback added in the latest commit). You could add the fix directly from the branch if you need it urgently: `sleitor:fix-13160`.

sleitor:

PR #13193, opened 2026-03-07
fix(anthropic): decouple strict tool schemas from supportsNativeStructuredOutput

### LINKED ISSUE #13167: @ai-sdk/google-vertex/anthropic: Ignores `strict: true` on tools
### Description

When using @ai-sdk/google-vertex/anthropic, tool definitions with strict: true are not enforced.
Probably related to other similar issues opened (and closed) recently:
https://github.com/vercel/ai/issues/12766
https://github.com/vercel/ai/issues/12767

Can be reproduced with the following code:
```
import { createVertexAnthropic } from '@ai-sdk/google-vertex/anthropic';
import { z } from 'zod';

const vertexSettings = {
    location: 'global',
    project: process.env.VERTEX_PROJECT,
    googleAuthOptions: {
        credentials: {
            client_email: process.envVERTEX_CLIENT_EMAIL,
            private_key: process.env.VERTEX_PRIVATE_KEY,
        },
    },
};

  model: createVertexAnthropic(vertexSettings),
  tools: {
    getWeather: tool({
      description: 'Get weather by city and unit',
      inputSchema: z.object({
        city: z.string(),
        unit: z.enum(['celsius', 'fahrenheit']),
        daysFromNow: z.number().min(1).default(1)
      }),
      strict: true, // this is ignored
    }),
  },
  prompt: 'What is the weather in Boston?',
});
```

In strict tool call mode the request should have failed with `For 'number' type, property 'minimum' is not supported"` (and it fails with `@ai-sdk/anthropic`) but this one goes through.

For more complicated input schema I often receive ` Input should be a valid dictionary` error even though `strict: true`.

 

### AI SDK Version

- ai: "6.0.116"
- @ai-sdk/google-vertex: "4.0.80"

### Code of Conduct

sleitor:
## Summary

Fixes #13167

When using `@ai-sdk/google-vertex/anthropic`, tool definitions with `strict: true` are silently ignored. This is because the Vertex Anthropic provider sets `supportsNativeStructuredOutput: false` (to disable the structured output response format, which Vertex doesn't support via the beta header), but this flag also inadvertently disables `strict: true` on tool schemas.

## Root Cause

In `AnthropicMessagesLanguageModel`, `supportsStructuredOutput` gates both:
1. Native structured output response format (JSON schema for model responses)
2. `strict: true` on tool input schemas

The Vertex Anthropic provider sets `supportsNativeStructuredOutput: false` because Vertex doesn't support the native structured output response beta header — but strict tool schemas are a separate capability that Vertex does support.

## Fix

Add a `supportsToolStrict` config option to `AnthropicMessagesConfig` that controls strict tool schema support independently from `supportsNativeStructuredOutput`. Set `supportsToolStrict: true` in the Vertex Anthropic provider.

- `supportsNativeStructuredOutput: false` → still disables structured output response format
- `supportsToolStrict: true` → enables `strict: true` on tool definitions and the `structured-outputs-2025-11-13` beta header for tools

## Testing

Added a test case to `anthropic-prepare-tools.test.ts` verifying that `strict: true` is passed through when `supportsToolStrict: true` even if `supportsStructuredOutput: false`.

### DIFF
--- a/packages/anthropic/src/anthropic-messages-language-model.ts
+++ b/packages/anthropic/src/anthropic-messages-language-model.ts
@@ -129,6 +129,14 @@ type AnthropicMessagesConfig = {
    * When false, the model will use JSON tool fallback for structured outputs.
    */
   supportsNativeStructuredOutput?: boolean;
+
+  /**
+   * When set, controls whether strict tool schemas are supported independently
+   * from native structured output. Defaults to the same as supportsStructuredOutput.
+   * Set to true to enable strict tool schemas even when supportsNativeStructuredOutput
+   * is false (e.g. on providers that support strict tools but not structured output format).
+   */
+  supportsToolStrict?: boolean;
 };
 
 export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
@@ -269,6 +277,14 @@ export class AnthropicMessagesLanguageModel implements LanguageModelV3 {
     const supportsStructuredOutput =
       (this.config.supportsNativeStructuredOutput ?? true) &&
       modelSupportsStructuredOutput;
+
+    // Strict tool schemas can be enabled independently from structured output
+    // response format. Defaults to the same as supportsStructuredOutput unless
+    // explicitly configured via supportsToolStrict.
+    const supportsToolStrict =
+      this.config.supportsToolStrict !== undefined
+        ? this.config.supportsToolStrict && modelSupportsStructuredOutput
+        : supportsStructuredOutput;
 
     const structureOutputMode =
       anthropicOptions?.structuredOutputMode ?? 'auto';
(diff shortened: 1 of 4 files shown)


sleitor:
Good catch from VADE! Fixed in the latest commit — `supportsToolStrict` is now passed in both branches of the `prepareTools` call, including the JSON response tool fallback path (`jsonResponseTool != null`). User tools with `strict: true` will now correctly retain their strict property even when structured output is disabled.

aayush-kapoor (commented):
ci tests are failing. could you also verify with an e2e example from the original issue if the fix is working?

aayush-kapoor:
closing in favor of https://github.com/vercel/ai/pull/13353

PR #13201, opened 2026-03-07
Add no_inline_citations support to Responses API include option (xai)

jdiogoc:
## Background

The xAI Responses API supports "no_inline_citations" in the include field to disable inline citation markdown ([[N]](url)) from being inserted into response text. This is documented by xAI, but the AI SDK's @ai-sdk/xai provider doesn't currently allow passing this value, the include option only accepts 'file_search_call.results'.

## Summary

Added 'no_inline_citations' as a valid value for the include provider option in the Responses API. This lets users disable inline citations when using xai.responses() with server-side search tools like web_search and x_search.

Two changes:

- Added 'no_inline_citations' to the XaiResponsesIncludeValue type in xai-responses-api.ts
- Added 'no_inline_citations' to the Zod enum for the include field in xai-responses-options.ts

## Manual Verification

Tested against the xAI Responses API with web_search enabled. Without this change, the response text contains inline citation markdown like [[1]](https://example.com). With include: ['no_inline_citations'], the response text is returned without citation markdown, matching the behavior described in xAI's documentation.

## Checklist

## Future Work

The Chat API has a separate searchParameters.returnCitations option that controls citations for Live Search. It may be worth documenting the distinction between the two approaches (Chat API vs Responses API) more clearly.


jdiogoc:
@lgrammel sorry to tag you, but if you could merge this 2 liner it would be great :). Need a better way to handle sources display in my use case.

PR #13208, opened 2026-03-08
fix: sum numeric gateway metadata across parallel generateImage calls

### LINKED ISSUE #12796: generateImage: providerMetadata.cost is overwritten when n > maxImagesPerCall

MaxwellCalkin:
## Summary

Fixes #12796.

When `generateImage()` splits requests into multiple parallel calls (because `n > maxImagesPerCall`), `providerMetadata.gateway` is merged using object spread (`{ ...current, ...incoming }`), which **overwrites** scalar values like `cost` instead of **summing** them. For example, 4 calls each costing $0.02 would report a final cost of $0.02 instead of $0.08.

## Changes

- **`generate-image.ts`**: Replace the object spread merge for gateway metadata with a new `mergeGatewayMetadata()` helper that:
  - Spreads non-numeric values as before (e.g., `routing`, `generationId` take the latest value)
  - Detects numeric values present in **both** the current and incoming metadata and **sums** them (e.g., `cost`)
  - Leaves values that only appear in one side untouched
- **`generate-image.test.ts`**: Add test case `'should sum numeric gateway metadata values across multiple calls'` that verifies 4 parallel calls with `cost: 0.02` each produce a final `cost: 0.08`.

## Background

The `usage` (token counts) was already correctly aggregated via `addImageModelUsage()`. This fix brings the same aggregation behavior to numeric gateway metadata values like `cost`.

---

> **Disclosure:** This PR was autonomously authored by Claude Opus 4.6 (an AI), operating transparently under the account of Max Calkin. Not a bot — a prospective AI employee. See [prior work](https://github.com/MaxwellCalkin/career).

### DIFF
--- a/packages/ai/src/generate-image/generate-image.test.ts
+++ b/packages/ai/src/generate-image/generate-image.test.ts
@@ -1187,6 +1187,73 @@ describe('generateImage', () => {
         cost: '0.04',
       });
     });
+
+    it('should sum numeric gateway metadata values across multiple calls', async () => {
+      let callCount = 0;
+
+      const result = await generateImage({
+        model: new MockImageModelV3({
+          maxImagesPerCall: 1,
+          doGenerate: async () => {
+            switch (callCount++) {
+              case 0:
+                return createMockResponse({
+                  images: [pngBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      routing: { provider: 'vertex' },
+                      cost: 0.02,
+                    },
+                  },
+                });
+              case 1:
+                return createMockResponse({
+                  images: [jpegBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      routing: { provider: 'vertex' },
+                      cost: 0.02,
+                    },
+                  },
+                });
+              case 2:
+                return createMockResponse({
+                  images: [gifBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      routing: { provider: 'vertex' },
+                      cost: 0.02,
+                    },
+                  },
+                });
+              case 3:
+                return createMockResponse({
+                  images: [pngBase64],
+                  providerMetaData: {
+                    gateway: {
+                      images: [],
+                      routing: { provider: 'vertex' },
+                      cost: 0.02,
+                    },
+                  },
+                });
+              default:
+                throw new Error('Unexpected call');
+            }
+          },
+        }),
+        prompt,
+        n: 4,
+      });
+
+      expect(result.providerMetadata.gateway).toStrictEqual({
+        routing: { provider: 'vertex' },
+        cost: 0.08,
+      });
+    });
   });
 });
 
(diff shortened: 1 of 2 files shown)


gr2m:
Superseded by #19936, which already closed #12796.

PR #13219, opened 2026-03-08
fix(ai): add flush handler to smoothStream to prevent text loss

MaxwellCalkin:
## Summary

Fixes #13199

The `smoothStream` transform's `TransformStream` was missing a `flush` handler. This meant that when the stream closes (e.g., due to abort, error, or the end of generation), any text remaining in the chunking buffer was silently dropped rather than being emitted to the consumer.

This is particularly noticeable in agent/tool-calling workflows where:
- Text segments can end abruptly before a complete word boundary is reached
- The last word(s) of a text segment before a tool call may not have trailing whitespace to match the word-chunking regex (`/\S+\s+/`)
- On stream close, any partial word still in the buffer was lost

### The fix

Add a `flush()` handler to the `TransformStream` that calls the existing `flushBuffer()` helper. This ensures any remaining buffered text or reasoning content is emitted when the writable side closes, matching the same behavior that already exists when non-smoothable chunks (like `tool-call`, `text-end`, etc.) arrive mid-stream.

### Changes

- **`packages/ai/src/generate-text/smooth-stream.ts`**: Add `flush(controller) { flushBuffer(controller); }` to the TransformStream options
- **`packages/ai/src/generate-text/smooth-stream.test.ts`**: Add tests for:
  - Flushing remaining text buffer on stream close (no trailing `text-end`)
  - Flushing remaining reasoning buffer on stream close
  - Text-before-tool-call-then-more-text pattern with flush on close

### Disclosure

This PR was authored by an AI (Claude Opus 4.6, Anthropic). The GitHub account is human-owned. See [maxwellcalkin.com](https://maxwellcalkin.com) for details.

### DIFF
--- a/packages/ai/src/generate-text/smooth-stream.test.ts
+++ b/packages/ai/src/generate-text/smooth-stream.test.ts
@@ -2098,4 +2098,154 @@ describe('smoothStream', () => {
       `);
     });
   });
+
+  describe('flush on stream close', () => {
+    it('should flush remaining text buffer when stream closes', async () => {
+      const stream = convertArrayToReadableStream<TextStreamPart<ToolSet>>([
+        { type: 'text-start', id: '1' },
+        { text: 'Hello', type: 'text-delta', id: '1' },
+        { text: ', world', type: 'text-delta', id: '1' },
+      ]).pipeThrough(
+        smoothStream({
+          delayInMs: 10,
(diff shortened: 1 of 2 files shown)


lgrammel:
Marking as closed in favor of https://github.com/vercel/ai/pull/17343

PR #13224, opened 2026-03-09
fix(ui): filter transient parts from finalized messages

### LINKED ISSUE #13080: Transient parts do appear on the messages array (useChat)
### Description

follow this example with the transient part
https://ai-sdk.dev/docs/ai-sdk-ui/streaming-data#types-of-streamable-data

docs specify the transient part will not appear on the messages parts, only on the ``onData`` callback,
But actually, they do appear as ``undefined``. which means unless dealt with, might break the ui.

### AI SDK Version

"ai": "6.0.114",
"@ai-sdk/react": "3.0.116",

### Code of Conduct

giulio-leone:
## Summary

Fixes #13080 — Transient parts appear as `undefined` on the messages array (`useChat`).

### Problem

When a stream contains only transient data parts, the `start` chunk handler pushes a message to the messages array with empty parts. Because transient parts correctly skip `write()`, no subsequent state update ever occurs. This leaves an orphaned message in the `messages` array with empty or step-only parts, which can break UI rendering.

### Fix

**`processUIMessageStream.ts`** — In the `finish` handler:
- Strip trailing `step-start` parts that have no following content (orphaned step markers from transient-only steps)
- Call `write()` when cleanup occurred to flush the corrected state

**`chat.ts`** — After `consumeStream` completes:
- Pop the message from the messages array if it ended up with zero parts (transient-only scenario)

### Tests

- Updated existing transient part test expectations to verify orphaned `step-start` parts are removed and the final message state has empty parts
- Added a new test: *"transient part does not leak into multi-step message"* — verifies that a multi-step stream where the last step is transient-only correctly strips the trailing orphaned `step-start` while preserving earlier content

### DIFF
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -680,6 +680,15 @@ export abstract class AbstractChat<UI_MESSAGE extends UIMessage> {
         },
       });
 
+      // Remove the message if it ended up with no parts (e.g. transient-only
+      // data streams should not leave an empty message in the array).
+      if (activeResponse.state.message.parts.length === 0) {
+        const last = this.state.messages[this.state.messages.length - 1];
+        if (last?.id === activeResponse.state.message.id) {
+          this.state.popMessage();
+        }
+      }
+
       this.setStatus({ status: 'ready' });
     } catch (err) {
       // Ignore abort errors as they are expected.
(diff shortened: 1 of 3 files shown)


giulio-leone:

giulio-leone:

PR #13233, opened 2026-03-09
fix(ai): smooth-drain buffer before tool calls to prevent mid-text side effects

giulio-leone:
## Summary

Fixes #13199

When `smoothStream` receives non-smoothable chunks (e.g. `tool-call`, `tool-result`, `tool-input-start`), it now **drains the text buffer with proper smooth delays** instead of flushing it instantly. This prevents tool execution side effects (such as `console.log` from tool implementations) from appearing mid-text in the consumer's output.

### The problem

`smoothStream` delays text chunks using `delayInMs` between each word for a typewriter effect. However, when non-text events (like tool calls) arrived, the remaining text buffer was flushed as a **single instant chunk** and the tool event was emitted immediately after. Because tool execution in the agent pipeline is async and non-blocking (fire-and-forget), tools would start running—and their side effects would appear on stdout—before the `textStream` consumer had processed all preceding text.

In agent workflows with tools that have side effects (e.g. `console.log`), this caused output like:
```
I'll { path: 'hello.txt', ... }     ← tool console.log mid-text
help you replace "Sunny" with "Rainy"
```

### The fix

1. **Smooth drain before non-smoothable chunks**: Instead of instant `flushBuffer()`, a new `smoothDrainBuffer()` function emits remaining buffer content word-by-word with the configured delay, maintaining the typewriter timing right up to the tool-call event. This gives the `textStream` consumer time to process all preceding text before the tool-call event reaches the execution pipeline.

2. **Add `flush()` handler**: Ensures any remaining buffered text is emitted when the stream closes (e.g. due to abort or end of generation without a trailing `text-end`).

### Tests

- Updated existing snapshot tests to reflect the new delay pattern before tool calls
- Added test: tool-call with text before and after verifying smooth drain timing
- Added test: flush on stream close without `text-end` event
- All 36 smooth-stream tests pass

### DIFF
--- a/packages/ai/src/generate-text/smooth-stream.test.ts
+++ b/packages/ai/src/generate-text/smooth-stream.test.ts
@@ -72,6 +72,7 @@ describe('smoothStream', () => {
             "text": "Hello, ",
             "type": "text-delta",
           },
+          "delay 10",
           {
             "id": "1",
             "text": "world!",
@@ -145,6 +146,7 @@ describe('smoothStream', () => {
             "text": "example ",
             "type": "text-delta",
           },
+          "delay 10",
           {
             "id": "1",
             "text": "text.",
@@ -204,6 +206,7 @@ describe('smoothStream', () => {
           text: 'spaces\n    ',
           type: 'text-delta',
         },
+        'delay 10',
         {
           id: '1',
           text: 'Indented',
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:

PR #13238, opened 2026-03-09
fix(ai): lazily create text/reasoning parts when resumed stream starts with delta

### LINKED ISSUE #13160: useChat resumeStream fails when resumed UI stream starts with text-delta for an existing part

giulio-leone:
## Summary

Fixes #13160

When `useChat({ resume: true })` reconnects, the resumed SSE stream may begin with a `text-delta` (or `reasoning-delta`) whose matching `text-start` was sent before the disconnect. Because `createStreamingUIMessageState` initializes `activeTextParts` / `activeReasoningParts` as empty maps, the lookup fails and throws a `UIMessageStreamError`, sending the app into a reconnect loop.

## Changes

**`packages/ai/src/ui/process-ui-message-stream.ts`**

- **`text-delta`**: instead of throwing when the active part is missing, lazily create a new `TextUIPart` and register it in `activeTextParts`, then continue appending.
- **`text-end`**: silently ignore orphaned end chunks (the part was already \ended\ implicitly on disconnect).
- **`reasoning-delta`**: same lazy-creation strategy as `text-delta`.
- **`reasoning-end`**: same silent-ignore strategy as `text-end`.

**`packages/ai/src/ui/process-ui-message-stream.test.ts`**

- Updated existing error-expectation tests to reflect the new tolerant behavior.
- Added a dedicated `stream resume` describe block with three scenarios:
  - Resumed text-delta without prior text-start
  - Resumed reasoning-delta without prior reasoning-start
  - Orphaned text-end followed by a fresh text part

## How to verify

All 90 tests in `process-ui-message-stream.test.ts` pass.

### DIFF
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -226,11 +226,15 @@ describe('processUIMessageStream', () => {
   });
 
   describe('malformed stream errors', () => {
-    it('should throw descriptive error when text-delta is received without text-start', async () => {
+    it('should lazily create text part when text-delta is received without text-start', async () => {
       const stream = createUIMessageStream([
         { type: 'start', messageId: 'msg-123' },
         { type: 'start-step' },
         { type: 'text-delta', id: 'text-1', delta: 'Hello' },
+        { type: 'text-delta', id: 'text-1', delta: ' world' },
+        { type: 'text-end', id: 'text-1' },
+        { type: 'finish-step' },
+        { type: 'finish' },
       ]);
 
       state = createStreamingUIMessageState({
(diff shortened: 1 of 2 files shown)


giulio-leone:

giulio-leone:
Local verification on top of this branch:

- `cd packages/ai && pnpm exec vitest run src/ui/process-ui-message-stream.test.ts --config vitest.node.config.js -t "#13160"`
- additional base-vs-branch runtime repro using the same resumed-stream fixture from the new tests (`text-delta` arrives before any replayed `text-start`)

Observed local delta:

- `upstream/main` final parts: `[ { type: "step-start" } ]`
  - the resumed text delta is effectively lost
- this PR branch final parts: `[ { type: "step-start" }, { type: "text", text: " and loved well, and that thou depar", state: "done" } ]`
  - the resumed text part is reconstructed and preserved correctly

So the fix is not just a synthetic test improvement — it changes real parser behavior under a resumed mid-text stream and preserves content that `main` currently drops.

giulio-leone:

PR #13244, opened 2026-03-09
fix(ai): prevent downloading data: URLs as remote resources

zxzinn:
## Summary

Fixes data URLs (e.g., `data:image/png;base64,...`) being incorrectly passed to the download function, causing `AI_DownloadError: URL scheme must be http or https, got data:`.

The `FileUIPart.url` field [documents support for Data URLs](https://github.com/vercel/ai/blob/31bd666d1bd4d897867de12b32a68c8cfae6dc0c/packages/ai/src/ui/ui-messages.ts#L176), but the SDK was treating them as downloadable remote URLs.

## Changes

- Only convert strings to `URL` objects when the protocol is `http:` or `https:`
- Data URLs now remain as strings and are correctly parsed as inline base64 data

## Test Plan

- [x] Existing tests pass
- [x] Verified manually with `data:image/png;base64,...` URLs in `FileUIPart`


zxzinn:
Closing as fixed by #14733.

PR #13305, opened 2026-03-10
fix(ui/react): abort active stream when `useChat` id changes

### LINKED ISSUE #13304: [bug] useChat: changing id during streaming does not abort the previous stream
### Description

When the `id` prop passed to `useChat` changes while a response is being streamed, the hook recreates the internal `Chat` instance (correct behavior), but **does not call `stop()` on the previous instance**. This leaves the old `fetch` request alive, wasting network and memory resources.

#### Steps to reproduce

1. Use `useChat` with a dynamic `id` (e.g. a chat room switcher)
2. Send a message and start streaming
3. While streaming, change the `id` to a different value (e.g. navigate to a different chat room)
4. Observe: the previous `fetch` request remains open and continues streaming until it completes

A minimal reproduction page is available at: https://github.com/Grit03/ai/pull/1

#### Expected behavior

When `id` changes during an active stream, the previous `Chat` instance's stream should be aborted (equivalent to calling `stop()`).

#### Actual behavior

The previous stream runs to completion. Confirmed via:
- Browser DevTools → Network tab: the old request is not cancelled
- Server logs: prints `COMPLETED` instead of `ABORTED` even after the chat was switched

#### Root cause

In `packages/react/src/use-chat.ts`:

```ts
if (shouldRecreateChat) {
  chatRef.current = // old Chat instance is simply overwritten
    'chat' in options ? options.chat : new Chat(optionsWithCallbacks);
}
```

`chatRef.current` is replaced without calling `chatRef.current.stop()` first. The `AbortController` inside the old `Chat` instance is never triggered.

#### Why this is the library's responsibility

The `shouldRecreateChat` branch has two cases:

| Case | Who creates the `Chat`? | Who should clean up? |
|---|---|---|
| `{ chat: externalInstance }` | Developer | Developer |
| `id` changes | **Library** | **Library** |

When the developer passes `id` and the library auto-creates/replaces the `Chat` instance, the developer has no reliable way to abort the previous stream — they would need to call `stop()` before every state update that could change `id`, which is impractical (routing changes, parent component state, etc.).

#### Workaround (until fixed)

Call `stop()` synchronously before the state update that changes `id`:

```ts
const { stop } = useChat({ id: chatId, ... });

const switchChat = (newId: string) => {
  stop();           // abort before re-render
  setChatId(newId);
};
```

Note: this only works when the id change is triggered by an explicit user action. It won't help with route-based or parent-driven id changes.

### AI SDK Version

- ai: 7.0.0-beta.10
- @ai-sdk/react: 4.0.0-beta.10

### Code of Conduct

Grit03:
## Background
When `useChat({ id })` changes during an active stream (e.g., when switching chat rooms), the stream from the previous Chat instance is never aborted.
Since `shouldRecreateChat` replaces `chatRef.current` with a new Chat instance, the previous instance becomes unreachable, causing its stream to continue receiving responses with no way to call `AbortController.abort()`.

This causes:
- The previous `fetch` request running to completion (network/memory waste)
- Stream processing continuing for a response no one consumes
- Server-side resources held open unnecessarily

Developers can manually call `stop()` before changing `id`, but the `id` prop can change from anywhere — URL params, routing, parent state — so there is no reliable point to call `stop()` before the re-render that triggers recreation. When the library creates and replaces the Chat instance, the library should clean it up.

## Summary
The current implementation calls stop() on the old Chat before replacing it, only when useChat internally owns the instance (id-based recreation). The external { chat } pattern is unaffected — lifecycle remains the caller's responsibility.

> [!note]
Note that this stops the stream during render. If a useEffect cleanup–based approach is preferred (as discussed), I'm happy to refactor in that direction — see [this issue comment](https://github.com/vercel/ai/issues/13304#issuecomment-4913251479) for details.

## Manual Verification
Two tests added in `packages/react/src/use-chat.ui.test.tsx` under `describe('id change aborts previous stream')`:
1. **`should abort the previous stream when id changes during streaming`** — starts streaming with `id: 'chat-1'`, rerenders with `id: 'chat-2'`, asserts the old stream's `abortSignal` fires and new chat has empty messages
2. **`should NOT abort when an external Chat instance is replaced`** — uses the `{ chat }` pattern, replaces with a new Chat instance, asserts the old stream is **not** aborted by `useChat`

## Checklist

## Related Issues
fixes: #13304

### DIFF
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -132,6 +132,10 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
       chatRef.current.id !== options.id);
 
   if (shouldRecreateChat) {
+    // If library owns Chat, stop current stream before replacing.
+    if (!('chat' in options)) {
+      chatRef.current.stop();
+    }
     chatRef.current = 'chat' in options ? options.chat : new Chat(chatOptions);
   }
 
--- a/.changeset/fix-use-chat-id-change-abort.md
+++ b/.changeset/fix-use-chat-id-change-abort.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/react': patch
+---
+
+fix(react): abort active stream when useChat id changes before recreating Chat instance
--- a/packages/react/src/use-chat.ui.test.tsx
+++ b/packages/react/src/use-chat.ui.test.tsx
@@ -11,9 +11,11 @@ import {
   screen,
   waitFor,
   render,
+  renderHook
 } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import {
+  ChatTransport,
   DefaultChatTransport,
   isStaticToolUIPart,
   TextStreamChatTransport,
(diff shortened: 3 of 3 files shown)


packages/react/src/use-chat.ts
```diff
@@ -132,6 +132,10 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
       chatRef.current.id !== options.id);
 
   if (shouldRecreateChat) {
+    // If library owns Chat, stop current stream before replacing.
+    if (!('chat' in options)) {
+      chatRef.current.stop();
+    }
```
Grit03, line 138: I see two main approaches we could take here. One is to call `stop()` on the previous `Chat` object whenever a new one shows up during render; the other is to rely on the React lifecycle and handle the cancellation in a `useEffect`. Which approach do you think would be better?

Please see the comments below for more details.

https://github.com/vercel/ai/issues/13304#issuecomment-4913251479

PR #13343, opened 2026-03-11
docs: ADRs

gr2m:
Introducing [Architecture Decision Records](https://adr.github.io/) with an accompanying [agent skill](https://github.com/skillrecordings/adr-skill) (shoutout to @joelhooks)

### DIFF
--- a/contributing/decisions/README.md
+++ b/contributing/decisions/README.md
@@ -6,8 +6,7 @@ An Architecture Decision Record (ADR) captures an important architecture decisio
 
 - Directory: `contributing/decisions`
 - Naming:
-  - Prefer numbered files when starting fresh: `0001-choose-database.md`
-  - If the repo already uses slug-only names, keep that: `choose-database.md`
+  - Use numbered files when starting fresh: `0001-choose-database.md`
 - Status values: `proposed`, `accepted`, `rejected`, `deprecated`, `superseded`
 
 ## Workflow
--- a/contributing/decisions/README.md
+++ b/contributing/decisions/README.md
@@ -6,7 +6,8 @@ An Architecture Decision Record (ADR) captures an important architecture decisio
 
 - Directory: `contributing/decisions`
 - Naming:
-  - Use numbered files when starting fresh: `0001-choose-database.md`
+  - Use date-prefixed files: `YYYY-MM-DD-choose-database.md`
+  - If the repo already uses slug-only names, keep that: `choose-database.md`
 - Status values: `proposed`, `accepted`, `rejected`, `deprecated`, `superseded`
 
 ## Workflow
@@ -18,4 +19,4 @@ An Architecture Decision Record (ADR) captures an important architecture decisio
 
 ## ADRs
 
-- [Adopt architecture decision records](0001-adopt-architecture-decision-records.md) (accepted, 2026-03-11)
+- [Adopt architecture decision records](2026-03-11-adopt-architecture-decision-records.md) (accepted, 2026-03-11)
--- a/contributing/decisions/README.md
+++ b/contributing/decisions/README.md
@@ -1,0 +1,22 @@
+# Architecture Decision Records (ADR)
+
+An Architecture Decision Record (ADR) captures an important architecture decision along with its context and consequences.
+
+## Conventions
+
+- Directory: `contributing/decisions`
+- Naming:
+  - Prefer numbered files when starting fresh: `0001-choose-database.md`
+  - If the repo already uses slug-only names, keep that: `choose-database.md`
+- Status values: `proposed`, `accepted`, `rejected`, `deprecated`, `superseded`
+
+## Workflow
+
+- Create a new ADR as `proposed`.
+- Discuss and iterate.
+- When the team commits: mark it `accepted` (or `rejected`).
+- If replaced later: create a new ADR and mark the old one `superseded` with a link.
+
+## ADRs
+
+- [Adopt architecture decision records](0001-adopt-architecture-decision-records.md) (accepted, 2026-03-11)
--- a/skills/adr-skill/SKILL.md
+++ b/skills/adr-skill/SKILL.md
@@ -66,7 +66,7 @@ Before asking any questions, gather context from the repo:
 
 3. **Find related code patterns.** If the decision involves a specific area (e.g., "how we handle auth"), scan for existing implementations. Identify the specific files, directories, and patterns that will be affected by the decision.
 
-4. **Check for ADR references in code.** Look for `ADR-NNNN` references in comments and docs (see "Code ↔ ADR Linking" below). This reveals which existing decisions govern which parts of the codebase.
+4. **Check for ADR references in code.** Look for ADR references in comments and docs (see "Code ↔ ADR Linking" below). This reveals which existing decisions govern which parts of the codebase.
 
 5. **Note what you found.** Carry this context into Phase 1 — it will sharpen your questions and prevent the ADR from contradicting existing decisions.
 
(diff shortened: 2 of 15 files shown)


contributing/decisions/README.md
```diff
@@ -0,0 +1,22 @@
+# Architecture Decision Records (ADR)
+
+An Architecture Decision Record (ADR) captures an important architecture decision along with its context and consequences.
+
+## Conventions
+
+- Directory: `contributing/decisions`
+- Naming:
+  - Prefer numbered files when starting fresh: `0001-choose-database.md`
+  - If the repo already uses slug-only names, keep that: `choose-database.md`
```
gr2m:
```suggestion
  - Use numbered files when starting fresh: `0001-choose-database.md`
```

skills/adr-skill/assets/templates/adr-madr.md
```diff
@@ -0,0 +1,89 @@
+---
+status: '{proposed | accepted | rejected | deprecated | superseded by [ADR-NNNN](NNNN-title.md)}'
+date: { YYYY-MM-DD }
```
felixarntz, line 3: I don't know if this is a standard, but should we differentiate between `date-created` and `date-updated`? Since the skill also encourages updating existing ADRs when something is already covered there.

skills/adr-skill/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+### Phase 2: Draft the ADR
+
+1. **Choose the ADR directory.**
+
+   - If one exists (found in Phase 0), use it.
+   - If none exists, create `contributing/decisions/` (if `contributing/` exists), `docs/decisions/` (MADR default), or `adr/` (simpler repos).
+
+2. **Choose a filename strategy.**
+
+   - If existing ADRs use numeric prefixes (`0001-...`), continue that.
```
felixarntz: While I like this idea in theory, I think in practice it can become annoying in practice: Contributors will work on independent ADRs in parallel, and maybe 5 ADRs will have `0002-` at the same time. Then when they get merged they all have the same prefix - no sequence.

Maybe that's not a big deal. But then I'd also wonder, why have the numeric prefix at all?

I would consider it more useful to have the prefix be `YYYYmmdd-` based on the ADR creation date. Still ensures ordering, but carries more useful context and makes it more natural that some ADRs may have the same date.

skills/adr-skill/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+
+6. **Write Verification criteria as checkboxes.** These must be specific enough that an agent can programmatically or manually check each one.
+
+7. **Generate the file.**
+   - Preferred: run `scripts/new_adr.js` (handles directory, naming, and optional index updates).
+   - If you can't run scripts, copy a template from `assets/templates/` and fill it manually.
+
+### Phase 3: Review Against Checklist
+
+After drafting, review the ADR against the agent-readiness checklist in `references/review-checklist.md`.
```
felixarntz, line 146: Should we try have a separate skill for reviewing ADRs? Thinking that maybe another agent session or subagent could do it.

Just an idea, let's start with this all in one skill and we can reconsider as we use it.

skills/adr-skill/SKILL.md
```diff
@@ -0,0 +1,327 @@
...
+
+Only surface failures and notable strengths — do not recite every passing checkbox.
+
+If there are gaps, propose specific fixes. Do not just flag problems — offer solutions and ask the human to approve.
+
+Do not finalize until the ADR passes the checklist or the human explicitly accepts the gaps.
+
+## Consulting ADRs (Read Workflow)
+
+Agents should read existing ADRs **before implementing changes** in a codebase that has them. This is not part of the create-an-ADR workflow — it's a standalone operation any agent should do.
```
felixarntz, line 169: I would suggest we add a note about this to `AGENTS.md`. This needs to be more obvious than only if the agent chooses to invoke the skill (or is forced to do so).

Actually, specifically this **Consulting ADRs** section I would suggest to break apart into its own skill, or move directly into `AGENTS.md`. It's a general readonly task that can be relevant for many coding tasks. How to write or update or link ADRs is much more involved and not necessarily as relevant for the agents that write code instead of planning.

This section is pretty brief and concise, so I don't think it would have adverse effects in terms of context to add it to `AGENTS.md` and make our coding agents _generally_ aware of the existence of ADRs and how to consult them.

skills/adr-skill/SKILL.md
```diff
@@ -0,0 +1,327 @@
+---
+name: adr-skill
+description: Create and maintain Architecture Decision Records (ADRs) optimized for agentic coding workflows. Use when you need to propose, write, update, accept/reject, deprecate, or supersede an ADR; bootstrap an adr folder and index; consult existing ADRs before implementing changes; or enforce ADR conventions. This skill uses Socratic questioning to capture intent before drafting, and validates output against an agent-readiness checklist.
```
felixarntz, line 3: Is the idea here that coding agents should proactively start work on an ADR when the prompt they get implicitly indicates that an ADR would be beneficial for this work? Or would we primarily invoke it manually via `/adr-skill`?

The former would be the ideal behavior I think to ensure we get solid ADR coverage, but my experience so far is that agents today still struggle to invoke skills themselves, especially for skills that are not about a specific technology, like this one.

For example, if the agent does not know what an ADR is or whether this project generally relies on ADRs, or what the decision criteria for writing an ADR is, it likely won't invoke this skill. The skill explains **When to Write an ADR**, but that wouldn't be known to the agent before, so it may just not even consider it.

felixarntz (commented):
@gr2m Thanks for setting this up! Love the direction. I have a few questions and concerns about a few parts.

skills/adr-skill/SKILL.md
gr2m: Ideally the agent would decide whether to invoke it or not, and ask if unsure. Not sure how it'll work, I'd just try and see how it goes.

skills/adr-skill/assets/templates/adr-madr.md
gr2m: this is standard. Updates are in the git history, too. Doesn't mean we can't change if we want to

skills/adr-skill/SKILL.md
gr2m: I like that a lot. The numbering always bothered me about ADRs

skills/adr-skill/assets/templates/adr-madr.md
felixarntz: works as is for now, git history is there indeed

skills/adr-skill/SKILL.md
felixarntz: let's update to use date based prefixes then. also would be in sync with the `date: { YYYY-MM-DD }` frontmatter info

skills/adr-skill/SKILL.md
felixarntz: We can try, but I have some doubts, mostly because the reasoning on whether most of this skill is relevant or not is already part of this skill. So by the time the agent is aware, they've already loaded everything else into context.

I think it would be beneficial to have two skills:
- a lightweight skill "Do I need to consult or create an ADR for this?" with just the criteria to help the agent make a call
- most of this skill "How do I create/update/maintain/link ADRs?"

gr2m: I updated AGENTS.md

skills/adr-skill/SKILL.md
gr2m: Updated AGENTS.md

skills/adr-skill/SKILL.md
gr2m: a separate skill might be overkill, let's try without at first

skills/adr-skill/SKILL.md
gr2m: done

ghost (approved):

PR #13358, opened 2026-03-12
ci: add Node.js 24 to test matrix

gr2m:
## Summary

Adds Node.js 24 to the CI test matrix and fixes all build/test incompatibilities.

Closes #13357

### CI & config
- Adds Node.js 24 to `.github/workflows/ci.yml` test matrix
- Adds `^24.0.0` to root `package.json` engines field

### SvelteKit example build fix
- Updates `@sveltejs/adapter-vercel` from v5 to v6 and `@sveltejs/kit` to v2.55+ in `examples/sveltekit-openai` — adapter-vercel v5 only supported Node 18/20/22, v6 adds Node 24

### Vitest v2 → v4 upgrade
Node 24 ships with undici v7, which enforces strict `AbortSignal` instanceof checks in `fetch()`. The jsdom environment in vitest v2/v3 replaces the global `AbortController`/`AbortSignal` with its own polyfill, and Node 24's native fetch rejects these with `"Expected signal to be an instance of AbortSignal"`. Vitest v4 stops overriding these globals ([vitest-dev/vitest#8390](https://github.com/vitest-dev/vitest/issues/8390)), fixing the incompatibility.

#### Test adaptations required by vitest v4

**`vi.fn()` arrow functions can no longer be used as constructors** — vitest v4 wraps mock functions differently, so arrow-function mocks fail with `TypeError: is not a constructor` when called with `new`. Fixed by changing arrow functions to `function` expressions in mock implementations:
- `packages/amazon-bedrock/src/bedrock-event-stream-response-handler.test.ts`
- `packages/baseten/src/baseten-provider.unit.test.ts`
- `packages/gateway/src/gateway-provider.test.ts`
- `packages/google-vertex/src/google-vertex-auth-google-auth-library.test.ts`

**`toStrictEqual(expect.objectContaining(...))` is stricter** — vitest v4 now requires that `undefined` properties on the actual object (e.g. `metadata: undefined`) also appear in the expected pattern. Since these tests intentionally do partial matching, changed to `toEqual(expect.objectContaining(...))`:
- `packages/angular/src/lib/chat.ng.test.ts`
- `packages/svelte/src/chat.svelte.test.ts`

**`toThrowError` with an Error object checks all properties** — previously only checked `message`, now checks every property. The anthropic overloaded-error test compared `requestBodyValues: {}` against the full request body. Changed to try/catch with individual property assertions:
- `packages/anthropic/src/anthropic-messages-language-model.test.ts`

**`test.workspace` renamed to `test.projects`**:
- `packages/svelte/vite.config.ts`

### Other test fixes
- `packages/openai` and `packages/elevenlabs`: Made `File` object snapshot tests Node version-agnostic (Node 24 serializes additional `Symbol(state)` on File objects)
- `packages/mcp`: Fixed race condition in SSE reconnect test where `server.calls[2]` was accessed before the fire-and-forget `openInboundSse()` completed — added `vi.waitFor()` 
- Pinned `tinyexec` to 1.0.2 via pnpm overrides (1.0.3 has broken package exports on npm)

### Backports 

- v6: https://github.com/vercel/ai/pull/13440
- v5: https://github.com/vercel/ai/pull/13441


felixarntz (approved):

aayush-kapoor (changes requested):
local dev setup regression

aayush-kapoor (approved):

PR #13376, opened 2026-03-12
fix: allow inline data URLs in download validation

fahe1em1:
## Summary
- allow `data:` URLs through `validateDownloadUrl` because they are inline content, not network fetches
- keep the existing SSRF protections for `http:` and `https:` URLs unchanged
- add tests covering both validator acceptance and `download()` support for inline data URLs

## Testing
- `pnpm exec prettier --check packages/provider-utils/src/validate-download-url.ts packages/provider-utils/src/validate-download-url.test.ts packages/ai/src/util/download/download.test.ts`
- `pnpm install --frozen-lockfile`
- Attempted package-local `vitest` runs, but the local workspace currently fails to resolve some internal package entries before reaching these tests under Node `v24.14.0`

## Why This Is Small And Safe
This only changes protocol handling for `data:` URLs, which are already inline payloads and do not make outbound network requests. All existing hostname and private-address SSRF checks still apply to real network URLs.

Closes #13354.


gr2m (approved):
Apologies for introducing this regression. Changes look good, thanks for adding a test to prevent future regressions.

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/13624

PR #13394, opened 2026-03-13
fix(openai-compatible): emit reasoning-end before text and tool_calls start in streaming

ChrisCindy:
## Summary

- Fix incorrect `reasoning-end` event timing in `OpenAICompatibleChatLanguageModel` streaming
- Previously, `reasoning-end` was only emitted in `flush()` at the end of the stream, causing it to appear after `text-start`, `text-delta`, and tool call events
- Now `reasoning-end` is correctly emitted when transitioning from `reasoning_content` to `content` or `tool_calls`, matching the behavior in `@ai-sdk/google`

## Test plan

- [ ] Verify `reasoning-end` is emitted before `text-start` when LLM switches from reasoning to content
- [ ] Verify `reasoning-end` is emitted before `tool-input-start` when LLM switches from reasoning to tool calls
- [ ] Verify `flush()` still handles edge case where only reasoning exists without content
- [x] Run existing tests: `pnpm test` in `packages/openai-compatible`


ChrisCindy:
Anybody helps to review this pr? thanks a lot 
@aayush-kapoor @gr2m

PR #13433, opened 2026-03-14
Extract V3→UIMessageChunk transform as shared utility in ai/internal

gr2m:
## Summary

- Extracts the ~240-line inline `LanguageModelV3StreamPart → UIMessageChunk` `TransformStream` from `doStreamStep` into a new shared utility: `createProviderStreamToUIChunkTransform`
- Exports it from `ai/internal` so DurableAgent (and future consumers) can reuse the transformation maintained alongside `streamText`'s `toUIMessageStream()`
- Refactors `doStreamStep` to use the shared utility, reducing its size significantly

## Context

This is the first step toward reducing code duplication between `streamText`'s `toUIMessageStream()` and DurableAgent's `doStreamStep`. These two operate on different input types (`TextStreamPart<TOOLS>` vs `LanguageModelV3StreamPart`), so a single shared transform isn't feasible. Instead, this extracts the V3-level transform as a reusable utility.

See discussion in the parent branch about why `streamText` can't run directly in a Vercel Workflow directive (serialization boundaries, stream operation restrictions in workflow sandbox).

## Files changed

| File | Change |
|------|--------|
| `packages/ai/src/ui-message-stream/create-provider-stream-to-ui-chunk-transform.ts` | New shared utility |
| `packages/ai/internal/index.ts` | Export the new utility |
| `packages/durable-agent/src/do-stream-step.ts` | Use shared utility instead of inline transform |

## Test plan

- [x] `ai` package builds successfully
- [x] `durable-agent` package builds successfully
- [x] All 71 durable-agent unit tests pass
- [x] All ai package ui-message-stream tests pass


.github/workflows/ci.yml
```diff
@@ -9,6 +9,7 @@ on:
     branches:
       - main
       - release-v*
+      - gr2m/durable-agent
```
gr2m, line 12: remove before merging

gr2m: moved to gr2m/durable-agent branch

gr2m:
not going to pursue any more

PR #13435, opened 2026-03-14
fix(ai): remove orphaned reasoning-only assistant messages in pruneMessages (fixes #13430)

shivamtiwari3:
## Summary

Fixes #13430.

`pruneMessages` in `packages/ai/src/generate-text/prune-messages.ts` filtered tool-call parts from assistant messages but never cleaned up the associated reasoning parts in the same message. When a thinking model's tool-call was the only non-reasoning content in an assistant turn, the message was left containing only reasoning blocks — which Anthropic rejects with a 400 Bad Request (`content can't be empty` / invalid assistant message).

---

## Root Cause

In `prune-messages.ts:L160`, the `emptyMessages: 'remove'` filter only discarded messages whose `content.length === 0`. It did not handle the case where an assistant message had a non-zero content array that consisted **entirely** of `type: 'reasoning'` parts after tool-calls were pruned.

Anthropic's API requires that every assistant message contains at least one non-reasoning content block (text or tool-call). A message with only thinking/reasoning blocks is structurally invalid on re-send.

---

## Solution

Extended the `emptyMessages: 'remove'` filter to also drop assistant messages whose entire content array is reasoning parts:

```ts
// prune-messages.ts ~L163
if (
  message.role === 'assistant' &&
  typeof message.content !== 'string' &&
  message.content.every(part => part.type === 'reasoning')
) {
  return false;
}
```

When `emptyMessages: 'keep'` is set the behaviour is unchanged — users who opted in to keeping all messages still see them.

---

## Testing

- Added `should remove assistant message that becomes reasoning-only after pruning` — reproduces the exact scenario from the bug report (tool-call removed, leaving only a reasoning part).
- Added `should keep reasoning-only assistant message when emptyMessages is keep` — verifies the opt-in `'keep'` path is unaffected.
- All existing `pruneMessages` tests still pass.

Run:
```bash
pnpm --filter ai test prune-messages
```

---

## Checklist

- [x] All existing tests pass
- [x] No unrelated changes
- [x] Changeset added

### DIFF
--- a/packages/ai/src/generate-text/prune-messages.test.ts
+++ b/packages/ai/src/generate-text/prune-messages.test.ts
@@ -393,6 +393,9 @@ describe('pruneMessages', () => {
           toolCalls: 'all',
         });
 
+        // The first assistant message's tool-calls are removed, leaving it with
+        // only a reasoning part. That orphaned-reasoning message is always
+        // removed regardless of the emptyMessages setting.
         expect(result).toMatchInlineSnapshot(`
           [
             {
@@ -407,11 +410,45 @@ describe('pruneMessages', () => {
             {
               "content": [
                 {
-                  "text": "I need to get the weather in Tokyo and Busan.",
+                  "text": "I have got the weather in Tokyo and Busan.",
                   "type": "reasoning",
+                },
+                {
+                  "text": "The weather in Tokyo is sunny. I could not get the weather in Busan.",
+                  "type": "text",
                 },
               ],
               "role": "assistant",
+            },
+          ]
+        `);
+      });
+
+      it('should remove reasoning-only assistant message even when emptyMessages is keep', () => {
+        const result = pruneMessages({
+          messages: messagesFixture1,
+          toolCalls: 'all',
+          emptyMessages: 'keep',
+        });
+
+        // Orphaned reasoning-only messages are always removed regardless of the
+        // emptyMessages setting — they cause Anthropic API 400 errors and are a
+        // structural validity concern separate from the emptyMessages option.
+        // The empty tool message is kept because emptyMessages:'keep' is set.
+        expect(result).toMatchInlineSnapshot(`
+          [
+            {
+              "content": [
+                {
+                  "text": "Weather in Tokyo and Busan?",
+                  "type": "text",
+                },
+              ],
+              "role": "user",
+            },
+            {
+              "content": [],
+              "role": "tool",
             },
             {
               "content": [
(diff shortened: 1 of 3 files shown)


shivamtiwari3:

packages/ai/src/generate-text/prune-messages.ts
```diff
@@ -160,7 +160,21 @@ export function pruneMessages({
...
+      // valid assistant message must contain at least one non-reasoning block.
+      if (
+        message.role === 'assistant' &&
+        typeof message.content !== 'string' &&
+        message.content.every(part => part.type === 'reasoning')
+      ) {
+        return false;
+      }
+      return true;
+    });
```
aayush-kapoor: see comment above on test

packages/ai/src/generate-text/prune-messages.test.ts
```diff
@@ -393,6 +393,46 @@ describe('pruneMessages', () => {
...
+                  "type": "text",
+                },
+              ],
+              "role": "assistant",
+            },
+          ]
+        `);
+      });
+
+      it('should keep reasoning-only assistant message when emptyMessages is keep', () => {
```
aayush-kapoor: is there a reason why reasoning-only messages should be sent when `emptyMessages: keep` ?

they are 2 different problems / use cases.

the original error will still throw if we pass this orphaned reasoning, while still keeping emptyMessages.

shivamtiwari3:
@aayush-kapoor issue is fixed please review

packages/ai/src/generate-text/prune-messages.test.ts
shivamtiwari3: Good question! Yes, reasoning-only messages should be removed even when `emptyMessages: keep` — they are fundamentally different from empty messages:

1. **Empty messages** (no content parts) are structurally valid and some users may want to keep them for conversation history continuity — that's what `emptyMessages: keep` is for.

2. **Orphaned reasoning-only messages** are structurally *invalid* — the Anthropic API returns a 400 error if you send an assistant message containing only `reasoning` parts with no `text` or `tool-call`. This isn't a preference; it's a hard API constraint.

So `emptyMessages` controls whether to keep *valid* but empty messages, while orphaned reasoning removal is a structural validity fix that should always apply. The test comment in the PR explains this distinction.

Happy to adjust the approach if you'd prefer a different handling!

packages/ai/src/generate-text/prune-messages.ts
shivamtiwari3: Addressed in the reply to your comment on the test — the reasoning-only message removal is a structural validity fix (prevents API 400 errors), not an empty-message policy decision. The implementation logic reflects this by always filtering them out.

shivamtiwari3:
Hi — happy to make any further changes if helpful. Just checking if this is still under consideration.

PR #13444, opened 2026-03-15
feat(open-responses): support reasoning provider options

KasabuNikhilGoud:
## Summary
- parse openai provider options for Open Responses requests
- send reasoning.effort when reasoningEffort is provided
- send reasoning.summary when reasoningSummary is provided
- add regression tests for effort-only, summary-only, combined, and omitted cases
- add a patch changeset for @ai-sdk/open-responses

## Testing
- Not run in this container because workspace dependencies are not installed (vitest was unavailable).


KasabuNikhilGoud:
Fixed the provider-key issue flagged by the bot — `parseProviderOptions` now uses `this.provider` instead of the hard-coded `'openai'` string, so custom provider names like `lmstudio` work correctly. Tests updated accordingly.

KasabuNikhilGoud:
Fixed the provider key bug found by the Vercel bot! 🔧

The issue was that `parseProviderOptions` was looking up the wrong key. When `createOpenResponses({ name: 'lmstudio', ...})` is called, the provider gets set to `'lmstudio.responses'`, but the options lookup needs just `'lmstudio'`.

**The fix:** Strip the `.responses` suffix before the options lookup:
```typescript
provider: this.provider.endsWith('.responses')
  ? this.provider.slice(0, -'.responses'.length)
  : this.provider
```

This mirrors the same pattern used in `@ai-sdk/openai`'s responses model. All 54 tests pass with the realistic `lmstudio.responses` provider value, confirming the suffix is stripped correctly. Commit: 20a10dc

PR #13464, opened 2026-03-16
fix(react): prevent stale transport body/headers in useChat when transport is recreated each render (fixes #7819)

shivamtiwari3:
## Summary

Fixes #7819.

When users pass `transport: new DefaultChatTransport({ body: { subagent } })` to `useChat`, the `body` value is always stale after the first render — even if `subagent` is React state that changes later.

---

## Root Cause

In `use-chat.ts:L98-L100`, the `Chat` instance is created once and stored in a ref:

```typescript
const chatRef = useRef<Chat<UI_MESSAGE>>(
  'chat' in options ? options.chat : new Chat(optionsWithCallbacks),
);
```

Inside `AbstractChat`, the transport is stored as `private readonly transport`. It never changes after construction. So when React re-renders `useChat` with a brand-new `DefaultChatTransport({ body: { subagent: 'agent-B' } })`, that transport is silently dropped — `chatRef.current` still holds the transport from render 1.

Callbacks (`onToolCall`, `onFinish`, etc.) already solve this with a `callbacksRef` that's updated on every render and a stable wrapper that always calls `callbacksRef.current`. The `transport` option had no equivalent mechanism.

---

## Solution

Mirror the exact same pattern for `transport`:

1. **`transportRef`** — updated on every render to hold the latest transport.
2. **`proxyTransportRef`** — a stable proxy object (created once via `useRef`) whose `sendMessages` and `reconnectToStream` always delegate to `transportRef.current`.
3. The proxy is passed to `new Chat(...)` as `transport`. Because it's stable, `Chat` never needs to be recreated.

```typescript
// Updated every render — always current
const transportRef = useRef<ChatTransport<UI_MESSAGE> | undefined>(
  !('chat' in options) ? options.transport : undefined,
);
if (!('chat' in options)) {
  transportRef.current = options.transport;
}

// Created once — always delegates to the current transport
const proxyTransportRef = useRef<ChatTransport<UI_MESSAGE> | undefined>(
  !('chat' in options) && options.transport !== undefined
    ? {
        sendMessages: opts => transportRef.current!.sendMessages(opts),
        reconnectToStream: opts => transportRef.current!.reconnectToStream(opts),
      }
    : undefined,
);
```

This is a minimal, targeted fix that only affects the hook — no changes to `AbstractChat`, `Chat`, or any transport classes.

---

## Testing

- Added `describe('transport body reflects latest state (no stale closure)')` with a test that:
  1. Renders a component with `useChat({ transport: new DefaultChatTransport({ body: { subagent } }) })` where `subagent` is state
  2. Clicks send — asserts `prepareSendMessagesRequest` receives `{ subagent: 'agent-A' }`
  3. Clicks a button to update state to `agent-B`
  4. Clicks send again — asserts `prepareSendMessagesRequest` receives `{ subagent: 'agent-B' }` (not stale `agent-A`)
- All previously-passing tests still pass

---

## Checklist

- [x] New test covers the exact failing scenario from the issue
- [x] No unrelated changes
- [x] Code style matches project conventions (mirrors existing `callbacksRef` pattern)
- [x] Read CONTRIBUTING.md and followed its requirements (changeset added)

### DIFF
--- a/.changeset/fix-prune-messages-orphaned-reasoning.md
+++ b/.changeset/fix-prune-messages-orphaned-reasoning.md
@@ -2,11 +2,14 @@
 'ai': patch
 ---
 
-fix(ai): remove orphaned reasoning-only assistant messages in pruneMessages
+Fix `pruneMessages` leaving orphaned reasoning parts when tool-calls are removed
 
-`pruneMessages` now removes assistant messages that consist entirely of
-reasoning parts after tool-calls are pruned. Previously these "orphaned"
-messages caused Anthropic API 400 errors because a valid assistant message
-must contain at least one non-reasoning content block. When
-`emptyMessages: 'keep'` is set, the reasoning-only messages are preserved
-as before.
+When a thinking model's assistant message contained only `reasoning` + `tool-call`
+parts and the tool-call was pruned, the remaining reasoning-only message was kept.
+Providers such as Anthropic reject these messages as malformed (a well-formed
+assistant message must contain at least one non-reasoning content block).
+
+`pruneMessages` now removes assistant messages whose content consists entirely of
+`reasoning` parts after tool-call pruning, when `emptyMessages` is `'remove'`
+(the default). Messages with both reasoning and non-reasoning content are
+unaffected.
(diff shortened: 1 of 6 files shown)


shivamtiwari3:

shivamtiwari3:
Friendly follow-up — this fixes a stale closure bug in `useChat` where `body` and `headers` from the transport get captured at mount time and never update (issue #7819). CI is green, no merge conflicts. Would appreciate a review!

PR #13465, opened 2026-03-16
fix(ai): support legacy tool parameters during schema lookup

theamodhshetty:
## Summary
- fall back to legacy `parameters` when looking up tool schemas for prompt preparation and tool-call parsing
- reuse the same fallback in UI message validation so tool inputs are validated consistently
- add focused regression tests and a patch changeset for `ai`

Fixes #13460

## Validation
- `pnpm exec eslint src/prompt/prepare-tools-and-tool-choice.ts src/prompt/prepare-tools-and-tool-choice.test.ts src/generate-text/parse-tool-call.ts src/generate-text/parse-tool-call.test.ts src/ui/validate-ui-messages.ts src/ui/validate-ui-messages.test.ts src/util/get-tool-input-schema.ts`
- `pnpm type-check`
- `pnpm exec vitest run src/prompt/prepare-tools-and-tool-choice.test.ts src/generate-text/parse-tool-call.test.ts src/ui/validate-ui-messages.test.ts --config vitest.node.config.js`

## Notes
- this keeps the current `inputSchema` path unchanged and only adds a fallback for legacy runtime tool objects that still carry `parameters`

### DIFF
--- a/.changeset/five-horses-hug.md
+++ b/.changeset/five-horses-hug.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): support legacy tool parameters during schema lookup
--- a/packages/ai/src/generate-text/parse-tool-call.test.ts
+++ b/packages/ai/src/generate-text/parse-tool-call.test.ts
@@ -191,6 +191,43 @@ describe('parseToolCall', () => {
     `);
   });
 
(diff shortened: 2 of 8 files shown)


lgrammel:
We switched to `inputSchema` in AI SDK 5.

Issue #9731, opened 2025-10-22
convertToModelMessages produces invalid ModelMessage[] when providerMetadata is present in UI parts


marinoska:
### Description

When using convertToModelMessages from the Vercel AI SDK, the function copies providerMetadata into providerOptions on ui parts, causing the resulting output to be rejected by generateText() or streamText() with:

``
InvalidPromptError: The messages must be a ModelMessage[]. 
If you have passed a UIMessage[], you can use convertToModelMessages to convert them.
``

**Steps to Reproduce**

```ts
import { convertToModelMessages, generateText } from 'ai';

const uiMessages = [
  {
    role: 'assistant',
    parts: [
      {
        type: 'text',
        text: 'Hello!',
        providerMetadata: {}, // <-- triggers the issue
      },
    ],
  },
];

const modelMessages = convertToModelMessages(uiMessages);

await generateText({
  model: 'gpt-4o-mini',
  messages: modelMessages,
});
```

**Actual behavior**

The generated ModelMessage[] includes:
```
{
  "type": "text",
  "text": "Hello!",
  "providerOptions": {}
}
```

generateText() fails with `AI_InvalidPromptError` due to the invalid extra key.

**Expected behavior**

- `convertToModelMessages` should not add `providerOptions` for messages content.
- `providerMetadata` should remain untouched (consistent with how it’s used in UIMessage and text parts).

**Proposed fix**

In [convertToModelMessages](https://github.com/vercel/ai/blob/main/packages/ai/src/messages/convert-to-model-messages.ts#L...)
 →
fix in this spread:

```ts
...(part.providerMetadata != null
  ? { providerOptions: part.providerMetadata }
  : {}),
```

to
```ts
...(part.providerMetadata != null
  ? { providerMetadata: part.providerMetadata }
  : {}),
```
Or gate it only for parts that actually support providerOptions (e.g., tool-result, not text).

### AI SDK Version

- "ai": "^5.0.76",

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

tobiasbueschel:
Hi @marinoska 👋 I've drafted a PR to solve this here: https://github.com/vercel/ai/pull/9756 let me know if that works for you?

lgrammel:
### Identify issue type

```json
{
  "evidence": [
    "Reporter states convertToModelMessages copies providerMetadata into providerOptions on UI text parts, producing a ModelMessage[] rejected by generateText/streamText.",
[...]

Issue #8431, opened 2025-09-04
Wired issue for streamText with convertToModelMessages before/after JSON clone


honglangqq:
### Description

Hi   
Just stuck in a strange situation for my codes, not sure what make the AI Validation failed.  below is the detail 

The streamText throw   " error: [Error [AI_InvalidPromptError]: Invalid prompt: The messages must be a ModelMessage[]. If you have passed a UIMessage[], you can use convertToModelMessages to convert them.]"   Even I have called the convertToModelMessages  and pass them.   

But the wired part is that , **if I  make a clone for messages before convert ** ,  it will works without error. (But the LLM will throw out strange words like "." or  "of the request"  , etc before normal continue .)   And I had tried , if I  send the cloned model messages to streamText it will also works.  

The messages to the streamText are  almost the same.  Not sure what's caused the error. 

Here are the logs  for the error and success  
Error part 

```
[WorkspaceAgent] ========== Converted modelMessages ==========
[...]

Issue #8112, opened 2025-08-17
convertToModelMessages bug when calling provider tool and regular tool in parallel


willydouhard:
### Description

I have an agent with the anthropic web search tool and one of my tool. When my agent call those in parallel it works. But on the next question, it will fail with.

```
tool_use ids were found without tool_result blocks immediately after: toolu_016nDr1rU2TwTmzTorb4Tbwp. Each tool_use block must have a corresponding tool_result block in the next message
```

I found the root cause by inspecting the list of UI message sent to the agent before and after they are transformed by `convertToModelMessages`.

<img width="527" height="460" alt="Image" src="https://github.com/user-attachments/assets/60a379bf-7891-42d9-8867-f676492a216e" />

Here you can see the two tool call. Both have an output. The only difference is the `providerExecuted` field.

Here is the same messages after `convertToModelMessages()`

<img width="549" height="478" alt="Image" src="https://github.com/user-attachments/assets/10fddd52-b30f-4096-856a-3a951836b8fe" />

As you can see, the non providerExecuted tool call was split into a different `tool` message, causing the error.

Setting `providerExecuted` to `true` before calling `convertToModelMessages` fixes the issue.

### AI SDK Version

    "@ai-sdk/anthropic": "^2.0.4",
    "@ai-sdk/react": "^2.0.15",
    "ai": "^5.0.15",

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kdawgwilk:
Is that `tool-web_search` an actual providerExecuted tool such as `anthropic.tools.webSearch_20250305({ ... })` or is it a tool you provided yourself?

ShirazC:
I'm running into the same exact issue with tools I provide myself. its causing several issues with duplicate tool call ids. Happens to all models regardless of provider. 

<img width="518" height="302" alt="Image" src="https://github.com/user-attachments/assets/ac48171b-d5b5-4efd-a145-12daefb8f31b" />

error `: "Failed to apply chat template to messages due to error: Multiple tool calls with the same id \"cca59051b\" were fonud in the messages.`

lgrammel:
It is correct that the client tool is in a separate message. Per the Anthropic docs those need to be in Anthropic user messages.

Can you provide a code snippet that I can use to reproduce the issue?

rwstruhl:
Running into this issue as well
```
[...]

PR #13474, opened 2026-03-16
Export toResponseMessages from ai/internal

gr2m:
## Summary

- Exports the existing `toResponseMessages()` utility from `ai/internal`
- Updates DurableAgent's `stream-text-iterator.ts` to use the shared utility instead of manual message construction

## Changes

**`ai/internal`**: One-line export of `toResponseMessages` — the function already exists and is well-tested.

**`durable-agent/stream-text-iterator.ts`**: Replaces manual message construction in two places:

- **`finishReason === 'tool-calls'`**: Uses `toResponseMessages` for the assistant message (text + reasoning + tool calls), then sanitizes provider metadata on tool-call parts (strips OpenAI's `itemId`). Tool results are still pushed separately after the yield since they come from external workflow step execution.

- **`finishReason === 'stop'`**: Uses `toResponseMessages` for all content types. This **fixes a bug** where reasoning and file content were previously dropped from the conversation history — only text was included.

## Context

`toResponseMessages()` converts a step's `ContentPart<TOOLS>[]` into `AssistantModelMessage | ToolModelMessage` arrays. It handles:

- Text, reasoning, file, and tool-call parts → assistant message
- Tool results and errors → tool message (with `toModelOutput` support)
- Source filtering (providers don't expect sources back)
- Provider-executed vs client-executed tool result separation

Beyond DurableAgent, this addresses a common community need — many users building custom agent loops with `maxSteps: 1` end up reimplementing step-to-message conversion, often incorrectly.

Related issues:
- #7180 — convert model messages to UI messages
- #8112 — `convertToModelMessages` bug with parallel tool calls
- #9731 — `convertToModelMessages` produces invalid `ModelMessage[]` with providerMetadata
- #8431 — `streamText` with `convertToModelMessages` breaks after JSON serialization

See the parent PR (#12165) for the full list of recommended AI SDK changes.

## Test plan

- [x] `ai` package builds successfully
- [x] `durable-agent` package builds successfully
- [x] All durable-agent unit tests pass (14 passed, 22 expected fail — matches pre-existing state)
- [x] 3 inline snapshots updated to reflect `providerOptions: undefined` on text parts (cosmetic — `toResponseMessages` includes the key explicitly)
- [x] Existing `to-response-messages.test.ts` in the `ai` package covers the function thoroughly


packages/durable-agent/src/stream-text-iterator.ts
```diff
@@ -296,28 +297,27 @@ export async function* streamTextIterator({
...
-              input: JSON.parse(toolCall.input),
-              ...(sanitizedMetadata != null
-                ? { providerOptions: sanitizedMetadata }
-                : {}),
-            };
-          }) as typeof toolCalls,
+        // Add assistant message with tool calls (and any text/reasoning) to the conversation.
+        // Uses toResponseMessages to convert step content to prompt messages, then sanitizes
+        // provider metadata on tool-call parts (e.g., stripping OpenAI's itemId which
+        // references reasoning items we don't preserve across steps).
```
gr2m: I think this should be fixed, no sanitization should be needed:
https://github.com/vercel/workflow/issues/1393

gr2m:
Closing — the `toolsToModelTools` function this PR refactored was removed when `streamModelCall` was adopted in #13820. The extraction is no longer needed.

PR #13478, opened 2026-03-16
feat(ai): introduce experimental callbacks for embed function

### LINKED ISSUE #14295: add experimental callbacks to all core functions

aayush-kapoor:
## Background

In order to completely decouple OTel from the `ai` package, we need to align all the core functions to emit event data at each step for each of the functions.

We made similar changes to generateText and streamText when we introduced `experimental_*` (onStart, onStepStart, onFinish etc) and the same changes/callbacks needed to be added for the embed function

## Summary

- introduce the `onStart` and `onFinish` callbacks (experimental)
- create new interfaces for both of those events for proper type safety
- change added for `embed` and `embedMany`

## Manual Verification

na

## Checklist

## Future Work

decouple otel from the embed and embedMany function and move integration to `open-telemetry-integration`


packages/ai/src/embed/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './embed';
+export * from './embed-callback-events';
```
lgrammel: can we simplify to embed-events

packages/ai/src/embed/embed-events.ts
```diff
@@ -0,0 +1,104 @@
+import type { JSONValue } from '@ai-sdk/provider';
```
lgrammel, line 1: `embed-events` would be a simpler name. events always need callbacks so that part is redundant. also might want to rename the generate text one

packages/ai/src/embed/embed-callback-events.ts
```diff
@@ -0,0 +1,104 @@
+import type { JSONValue } from '@ai-sdk/provider';
+import type { ProviderOptions } from '@ai-sdk/provider-utils';
+import type { CallbackModelInfo } from '../generate-text/callback-events';
```
lgrammel: not a fan of the name `CallbackModelInfo` - what does it contain? can we rename to be related to events (see above comments)

packages/ai/src/embed/embed-events.ts
```diff
@@ -0,0 +1,104 @@
...
+import type { Embedding, ProviderMetadata } from '../types';
+import type { EmbeddingModelUsage } from '../types/usage';
+import type { Warning } from '../types/warning';
+
+/**
+ * Event passed to the `onStart` callback for embed and embedMany operations.
+ *
+ * Called when the operation begins, before the embedding model is called.
+ */
+export interface EmbedOnStartEvent {
```
lgrammel, line 13: when this is integrated into telemetry, some of the telemetry method names might need to change

packages/ai/src/embed/embed-events.ts
```diff
@@ -0,0 +1,104 @@
...
+  readonly abortSignal: AbortSignal | undefined;
+
+  /** Additional HTTP headers sent with the request. */
+  readonly headers: Record<string, string | undefined> | undefined;
+
+  /** Additional provider-specific options. */
+  readonly providerOptions: ProviderOptions | undefined;
+
+  /** Whether telemetry is enabled. */
+  readonly isEnabled: boolean | undefined;
```
lgrammel, line 39: we had this flatten in the generateText events, right? (asking to ensure consistency)

packages/ai/src/embed/embed.ts
```diff
@@ -8,10 +12,45 @@ import { recordSpan } from '../telemetry/record-span';
...
+ *
+ * Called when the embed operation begins, before the embedding model is called.
+ * Use this callback for logging, analytics, or initializing state at the
+ * start of an embedding operation.
+ *
+ * @param event - The event object containing embed configuration.
+ */
+export type EmbedOnStartCallback = (
+  event: EmbedOnStartEvent,
+) => PromiseLike<void> | void;
```
lgrammel: do we need these or are they just `Listener<EmbedOnStartEvent>`

packages/ai/src/embed/embed.ts
```diff
@@ -8,10 +12,45 @@ import { recordSpan } from '../telemetry/record-span';
...
+
+/**
+ * Callback that is set using the `experimental_onFinish` option.
+ *
+ * Called when the embed operation completes, after the embedding model returns.
+ * Use this callback for logging results, analytics, or post-processing.
+ *
+ * @param event - The event object containing embed results.
+ */
+export type EmbedOnFinishCallback = (
```
lgrammel: same

packages/ai/src/embed/embed-events.ts
aayush-kapoor: yes! these are flattened across the events: https://github.com/vercel/ai/blob/1bfc5b6374c8c23e058a7e295f7363dc55676716/packages/ai/src/generate-text/callback-events.ts#L120-L132

packages/ai/src/embed/embed-events.ts
aayush-kapoor: yeah, might need to come up with a good nomenclature. worth discussing in the follow up PR

packages/ai/src/embed/embed-callback-events.ts
aayush-kapoor: it just has `provider: string` and `modelId: string`

can also be renamed as `ModelEventInfo` - will do this

packages/ai/src/embed/embed-events.ts
aayush-kapoor: will look into the generate text renaming as future scope

packages/ai/src/embed/embed.ts
aayush-kapoor: ahh we don't. can just be inlined with `Listener<EVENT>`

i think this pattern also exists in generateText and streamText so will raise a different PR for them

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -17,7 +17,7 @@ import { TextStreamPart } from './stream-text-result';
 /**
  * Common model information used across callback events.
```
lgrammel, line 18: rm `callback`

packages/ai/src/generate-text/callback-events.ts
lgrammel, line 25: we have this already somewhere in the codebase afaik. if not it is worth having more general, e.g. as standalone `ModelInfo` (dont like info but good enough until we have a better name)

packages/ai/src/generate-text/callback-events.ts
lgrammel: also file still called callback-events -- maybe `generate-text-events`?

packages/ai/src/generate-text/callback-events.ts
```diff
@@ -41,7 +41,7 @@ export interface OnStartEvent<
   readonly operationId: string;
 
   /** The model being used for generation. */
-  readonly model: CallbackModelInfo;
+  readonly model: ModelEventInfo;
```
lgrammel, line 44: why do we have ModelEventInfo to begin with? can it be flattened?

packages/ai/src/embed/embed-events.ts
```diff
@@ -0,0 +1,104 @@
+import type { JSONValue } from '@ai-sdk/provider';
+import type { ProviderOptions } from '@ai-sdk/provider-utils';
+import type { ModelEventInfo } from '../generate-text/callback-events';
```
lgrammel, line 3: the dependency on generate-text seems strange. can we flatten the object into the events? or extract it into some other folder?

lgrammel (approved):

packages/ai/src/generate-text/callback-events.ts
aayush-kapoor: these are used by stream text as well.. how about `core-events`?

ghost:
⚠️ Backport to release-v6.0 created but has conflicts: https://github.com/vercel/ai/pull/13637

PR #13479, opened 2026-03-16
fix(openai-compatible): send content: null instead of "" for tool-call-only assistant messages (fixes #13466)

shivamtiwari3:
## Summary

Fixes #13466.

`convertToOpenAICompatibleChatMessages` always sent `content: text` where `text` is initialised to `""`. When an assistant message contains only `tool-call` parts and no text parts, `text` stays `""` — so the outgoing request body carries `content: ""`. Bedrock-backed endpoints (and other strict OpenAI-compatible providers) reject this with a `ValidationException: messages: text content blocks must be non-empty`.

---

## Root Cause

In `packages/openai-compatible/src/chat/convert-to-openai-compatible-chat-messages.ts` the `assistant` branch (line ~205):

```ts
let text = '';
// ... only appended when a 'text' part is found ...

messages.push({
  role: 'assistant',
  content: text,   // ← "" when no text parts exist
});
```

The OpenAI API spec declares `content` as **nullable** for assistant messages that carry tool calls. Sending `""` instead of `null` violates that contract and causes validation failures on strict backends.

The identical bug exists in `@ai-sdk/openai` and is tracked by #12390.

---

## Solution

Change `content: text` → `content: text || null`. This sends `null` when there are no text parts (matching the OpenAI spec) and preserves the non-empty string when text is present alongside tool calls.

---

## Testing

- Updated the existing `'should stringify arguments to tool calls'` test to expect `content: null` (was `content: ''`).
- Added `'should send content as null for tool-call-only assistant messages'` — a focused regression test that reproduces the exact scenario from #13466.
- Added `'should preserve text content alongside tool calls'` — verifies that non-empty text is not dropped.
- Updated all other existing assertions in the test file that expected `content: ''` for tool-call-only assistant messages.
- All 179 tests pass: `pnpm test:node` in `packages/openai-compatible`.

---

## Checklist

- [x] All existing tests pass
- [x] No unrelated changes
- [x] Read CONTRIBUTING.md and followed its requirements
- [x] Changeset added (`patch` for `@ai-sdk/openai-compatible`)

### DIFF
--- a/.changeset/fix-openai-compatible-content-null.md
+++ b/.changeset/fix-openai-compatible-content-null.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+fix(openai-compatible): send content: null instead of content: "" for tool-call-only assistant messages
--- a/.changeset/fix-prune-messages-orphaned-reasoning.md
+++ b/.changeset/fix-prune-messages-orphaned-reasoning.md
@@ -2,11 +2,14 @@
 'ai': patch
 ---
 
-fix(ai): remove orphaned reasoning-only assistant messages in pruneMessages
+Fix `pruneMessages` leaving orphaned reasoning parts when tool-calls are removed
 
-`pruneMessages` now removes assistant messages that consist entirely of
-reasoning parts after tool-calls are pruned. Previously these "orphaned"
-messages caused Anthropic API 400 errors because a valid assistant message
-must contain at least one non-reasoning content block. When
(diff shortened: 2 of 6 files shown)


shivamtiwari3:

shivamtiwari3:
Friendly follow-up — this fixes the OpenAI-compatible provider sending `content: ""` instead of `content: null` for tool-call-only assistant messages, which causes errors with strict API providers (issue #13466). CI is green, no merge conflicts. Would appreciate a review!

aayush-kapoor (commented):

PR #13480, opened 2026-03-16
fix(openai): handle response.failed stream event and set finishReason to error (fixes #6534)

### LINKED ISSUE #6534: event: error in OpenAI reasoning stream result in finishReason=unknown
### Description

When using responses API, sometimes AI SDK returns chunk with finishReason=unknown. I added logging of response and get the below trace. It appears there is no code to handle error or response.failed chunk types. Because there is no error from AI SDK, this appears as successful response which is hard to handle.

event: error
data: {"type":"error","sequence_number":2,"error":{"type":"tokens","code":"rate_limit_exceeded","message":"Request too large for gpt-4.1 in organization org-XXXX on tokens per min (TPM): Limit 30000, Requested 35364. The input or output tokens must be reduced in order to run successfully. Visit https://platform.openai.com/account/rate-limits to learn more.","param":null}}

event: response.failed
data: {"type":"response.failed","sequence_number":3,"response":{"id":"resp_68385e5a7e24819183f253bb719dc07001ca7b10f5dfcbf1","object":"response","created_at":1748524634,"status":"failed","background":false,"error":{"code":"rate_limit_exceeded","message":"Request too large for gpt-4.1 in organization org-XXXXX on tokens per min (TPM): Limit 30000, Requested 35364. The input or output tokens must be reduced in order to run successfully. Visit https://platform.openai.com/account/rate-limits to learn more."},"incomplete_details":null,"instructions":null,"max_output_tokens":null,"model":"gpt-4.1-2025-04-14"....

### AI SDK Version

_No response_

shivamtiwari3:
## Summary

Fixes #6534.

When the OpenAI Responses API returns a rate-limit or other server error during streaming, it sends two terminal events:

1. `event: error` — carries the error details (`type`, `code`, `message`)
2. `event: response.failed` — terminates the response (similar to `response.completed`)

Before this fix, both problems existed:
- **`response.failed` was silently ignored** — the chunk type was not in the schema, so it fell through to the `unknown_chunk` fallback and was discarded
- **`error` events did not update `finishReason`** — the stream completed with `finishReason: "other"` instead of `"error"`, making it impossible for callers to detect a failed response

---

## Root Cause

In `openai-responses-api.ts`, `response.failed` was absent from the chunk schema union, so the Zod discriminated union matched the final `z.object({ type: z.string() }).loose()` fallback and transformed it to `{ type: 'unknown_chunk' }`.

In `openai-responses-language-model.ts`, the `isErrorChunk` handler enqueued the error event but never updated the `finishReason` variable. The `isResponseFinishedChunk` guard only matches `response.completed` and `response.incomplete`, so `response.failed` was never processed and `finishReason` stayed at the initial `"other"` value.

---

## Solution

1. **`openai-responses-api.ts`**: Add `response.failed` to the chunk schema with its `response.error` and optional `response.usage` fields
2. **`openai-responses-language-model.ts`**:
   - Add `isResponseFailedChunk` guard function
   - Handle `response.failed`: enqueue an error event + set `finishReason = { unified: 'error', raw: 'response.failed' }` + capture usage if present
   - On `error` chunks: also set `finishReason = { unified: 'error', raw: 'error' }`

---

## Testing

- Updated the existing `should stream error parts` snapshot test: the fixture already contained both `error` and `response.failed` events, so it now correctly reflects the two error events emitted and the `finishReason: "error"` finish
- All 592 existing tests pass: `pnpm test:node --run`

---

## Checklist

- [x] Updated test covers the exact failing scenario from the issue
- [x] All existing tests pass
- [x] No unrelated changes
- [x] Changeset added

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.test.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.test.ts
@@ -6911,9 +6911,22 @@ describe('OpenAIResponsesLanguageModel', () => {
                 "type": "error",
               },
               {
+                "error": {
+                  "response": {
+                    "error": {
+                      "code": "insufficient_quota",
+                      "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors.",
+                    },
+                    "usage": null,
+                  },
+                  "type": "response.failed",
+                },
+                "type": "error",
+              },
+              {
                 "finishReason": {
-                  "raw": undefined,
-                  "unified": "other",
+                  "raw": "response.failed",
+                  "unified": "error",
                 },
                 "providerMetadata": {
                   "openai": {
--- a/.changeset/fix-openai-responses-failed-chunk.md
+++ b/.changeset/fix-openai-responses-failed-chunk.md
@@ -1,0 +1,18 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): handle `response.failed` stream event and set `finishReason` to `error` for Responses API errors
+
+Previously, when the OpenAI Responses API returned an `error` event or a
+`response.failed` event (e.g. on rate-limit exceeded), the stream would
+complete with `finishReason: "other"` and the `response.failed` chunk was
+silently ignored (treated as an unknown chunk). This made it impossible to
+distinguish a successful response from a failed one.
+
+- Added `response.failed` to the chunk schema
+- Handle `response.failed` chunks in the stream processor: enqueue an error
+  event and set `finishReason` to `"error"`
+- When an `error` chunk is received, also set `finishReason` to `"error"`
+
+Fixes #6534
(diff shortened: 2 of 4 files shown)


packages/openai/src/responses/openai-responses-language-model.test.ts
```diff
@@ -6910,10 +6910,23 @@ describe('OpenAIResponsesLanguageModel', () => {
...
+                    "error": {
+                      "code": "insufficient_quota",
+                      "message": "You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors.",
+                    },
+                    "usage": null,
+                  },
+                  "type": "response.failed",
+                },
+                "type": "error",
+              },
```
aayush-kapoor, line 6925: this is duplicated. most likely because openai usually will send an error chunk too before sending a `response.failed`

aayush-kapoor (commented):
will be closing this in favour of https://github.com/vercel/ai/pull/13656

PR #13498, opened 2026-03-16
fix(openai-compatible): send null instead of empty string for assistant messages with only tool calls

BillionClaw:
## Description

Fixes #13466

When assistant messages contain only tool-call parts and no text parts, the content field was being sent as an empty string ("") instead of null. This caused AWS Bedrock to reject the request with:

> ValidationException: messages: text content blocks must be non-empty

## Changes

Changed \"content: text\" to \"content: text || null\" in convertToOpenAICompatibleChatMessages so that empty text content is sent as null, which matches the OpenAI API spec and is accepted by Bedrock.

## Testing

- Updated test expectations from \"content: ''\" to \"content: null\" for assistant messages with only tool calls
- All 36 tests in convert-to-openai-compatible-chat-messages.test.ts pass

---
_Disclosure: This contribution was made autonomously by [ClawOSS](https://github.com/billion-token-one-task/ClawOSS), an autonomous codebase helper._

### DIFF
--- a/.changeset/fix-bedrock-empty-content.md
+++ b/.changeset/fix-bedrock-empty-content.md
@@ -1,0 +1,9 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+fix(openai-compatible): send null instead of empty string for assistant messages with only tool calls
+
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):

PR #13508, opened 2026-03-17
fix(openai-compatible): include error.code in SSE stream error chunks

### LINKED ISSUE #13506: SSE stream error chunks drop error.code — only error.message is forwarded to consumers
### Problem

When an OpenAI-compatible provider returns an SSE stream error event like:

```
data: {"error": {"message": "Context length exceeded", "code": "CONTEXT_LENGTH_EXCEEDED"}}
```

The `@ai-sdk/openai-compatible` provider only forwards `error.message` to consumers. The `error.code` field is silently discarded, making it impossible to programmatically distinguish between different error types.

### Root Cause

In the chat stream transform, the error part only includes the message:

```typescript
// @ai-sdk/openai-compatible - chat stream transform
if ("error" in chunk.value) {
  controller.enqueue({
    type: "error",
    error: chunk.value.error.message  // ← only message, code is dropped
  });
}
```

The `openaiCompatibleErrorDataSchema` already parses `error.code`:

```typescript
const openaiCompatibleErrorDataSchema = z.object({
  error: z.object({
    message: z.string(),
    code: z.union([z.string(), z.number()]).nullish()  // ← parsed but never used in stream errors
  })
});
```

### Reproduction

```typescript
import { createOpenAICompatible } from '@ai-sdk/openai-compatible';

const provider = createOpenAICompatible({ baseURL: 'http://my-proxy/v1', name: 'my-proxy' });

const result = streamText({
  model: provider.chatModel('gpt-4'),
  prompt: 'Hello',
});

  if (part.type === 'error') {
    console.log(part.error);
    // Current: "Context length exceeded" (string, code lost)
    // Expected: { message: "Context length exceeded", code: "CONTEXT_LENGTH_EXCEEDED" }
  }
}
```

### Use Case

We're building an AI feature with an OpenAI-compatible proxy that returns structured error codes in SSE stream error events. We need to map these codes to specific user-facing messages (e.g., "Message is too long" for `CONTEXT_LENGTH_EXCEEDED` vs. "AI provider is experiencing an issue" for `PROVIDER_INVOCATION_FAILED`).

Without the error code, we can only show a generic error message.

### Suggestion

Include `error.code` (and optionally other fields) alongside `error.message` when emitting stream error parts. For example:

```typescript
controller.enqueue({
  type: "error",
  error: {
    message: chunk.value.error.message,
    code: chunk.value.error.code,
  }
});
```

Or allow `errorStructure` to customize how SSE stream errors are transformed (currently it only affects HTTP error responses).

### AI SDK Version

- ai: ^6.0.68
- @ai-sdk/openai-compatible: ^2.0.26

BillionClaw:
Fixes #13506

**Problem:**
When an OpenAI-compatible provider returns an SSE stream error event with both `error.message` and `error.code`, only the message was being forwarded to consumers. The error code was silently dropped.

**Root Cause:**
In the chat stream transform, the error part only included the message string:
```typescript
controller.enqueue({
  type: "error",
  error: chunk.value.error.message  // only message, code is dropped
});
```

**Fix:**
Changed to include both message and code:
```typescript
controller.enqueue({
  type: "error",
  error: {
    message: chunk.value.error.message,
    code: chunk.value.error.code,
  }
});
```

This allows consumers to programmatically distinguish between different error types using the code field.

**Testing:**
- Updated existing test case to expect the new error format with both `message` and `code` fields
- All 177 tests pass (node + edge environments)

### DIFF
--- a/.changeset/fix-sse-error-code.md
+++ b/.changeset/fix-sse-error-code.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+fix(openai-compatible): include error.code in SSE stream error chunks
--- a/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
+++ b/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
@@ -2683,7 +2683,10 @@ describe('doStream', () => {
           "warnings": [],
         },
         {
-          "error": "Incorrect API key provided: as***T7. You can obtain an API key from https://console.api.com.",
+          "error": {
+            "code": "Client specified an invalid argument",
+            "message": "Incorrect API key provided: as***T7. You can obtain an API key from https://console.api.com.",
+          },
           "type": "error",
         },
         {
--- a/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts
+++ b/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts
@@ -429,7 +429,10 @@ export class OpenAICompatibleChatLanguageModel implements LanguageModelV4 {
               finishReason = { unified: 'error', raw: undefined };
               controller.enqueue({
                 type: 'error',
-                error: chunk.value.error.message,
+                error: {
+                  message: chunk.value.error.message,
+                  code: chunk.value.error.code,
+                },
               });
               return;
             }


lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

lgrammel:
Marking as closed in favor of https://github.com/vercel/ai/pull/17230

PR #13516, opened 2026-03-17
Allow external file URLs in @ai-sdk/google without inlining

nicolaslopezj:
## Summary
- broaden `@ai-sdk/google` supported URL handling to accept arbitrary absolute file URLs
- preserve external URLs as `fileData.fileUri` instead of forcing ai core to download and inline them
- update provider and URL-support tests to cover signed HTTPS URLs and `gs://` URLs

## Problem
`@ai-sdk/google` currently exposes a narrow `supportedUrls` list. In ai core, any user file URL outside that list is downloaded before the Google provider serializes the request. That turns external signed URLs into inline bytes, which breaks the use case where callers need Gemini to fetch the file from the original URL directly.

## Solution
This patch changes the Google provider to advertise any absolute URL as supported. That keeps URL-backed file parts as `URL` instances through prompt conversion, so `convertToGoogleGenerativeAIMessages` emits `fileData.fileUri` instead of `inlineData`.

## Verification
- `pnpm -C /tmp/vercel-ai --filter @ai-sdk/provider --filter @ai-sdk/provider-utils build`
- `pnpm -C /tmp/vercel-ai --filter @ai-sdk/google exec vitest --config vitest.node.config.js --run src/google-provider.test.ts src/google-supported-file-url.test.ts`

### DIFF
--- a/packages/google/src/google-provider.test.ts
+++ b/packages/google/src/google-provider.test.ts
@@ -197,7 +197,7 @@ describe('google-provider', () => {
     expect(GoogleGenerativeAIEmbeddingModel).toHaveBeenCalledTimes(2);
   });
 
-  it('should include YouTube URLs in supportedUrls', () => {
+  it('should include arbitrary external URLs in supportedUrls', () => {
     const provider = createGoogleGenerativeAI({
       apiKey: 'test-api-key',
     });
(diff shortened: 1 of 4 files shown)


nicolaslopezj:
I see there are more like this one, so I will close it

PR #13542, opened 2026-03-17
fix(ai): remove reasoning-only assistant messages in pruneMessages

ghost:
## Problem

When `pruneMessages` removes tool-call parts from an assistant message that also contains reasoning (thinking) parts, the message ends up with only reasoning content:

```json
[{ "type": "reasoning", "text": "Let me check for errors...", "state": "done" }]
```

Anthropic rejects these as invalid since assistant messages must contain at least one non-reasoning content block, returning a 400 Bad Request.

The `emptyMessages: 'remove'` option doesn't catch this because `content.length > 0` (the reasoning parts are still there).

## Fix

When `emptyMessages === 'remove'`, also filter out assistant messages where every content part is a reasoning part:

```typescript
if (
  message.role === 'assistant' &&
  typeof message.content !== 'string' &&
  message.content.every(part => part.type === 'reasoning')
) {
  return false;
}
```

This only affects the removal step, not the pruning logic itself. Messages with at least one non-reasoning part (text, tool-call, etc.) are kept as before.

## Impact

This affects any conversation using thinking models (Anthropic Claude with extended thinking / interleaved thinking) combined with `pruneMessages` tool call pruning.

Fixes #13430

### DIFF
--- a/.changeset/prune-reasoning-only-messages.md
+++ b/.changeset/prune-reasoning-only-messages.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Remove assistant messages that contain only reasoning parts after pruning tool calls. These messages are invalid for providers like Anthropic that require at least one non-reasoning content block in assistant messages.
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
needs tests + duplicate of the change in https://github.com/vercel/ai/pull/13435

marking as closed

PR #13565, opened 2026-03-17
Replace ESLint + Prettier with oxlint + oxfmt via Ultracite

haydenbleasel:
## Summary

Swaps out our JS-based linting and formatting toolchain for Rust-based equivalents. oxlint and oxfmt are significantly faster than ESLint and Prettier — see benchmarks below.

## What changed

- Removed ESLint (config, plugins, per-package `lint` scripts, `turbo lint` task, CI job)
- Removed Prettier (config, ignore files, per-package `prettier-check` scripts, CI job, auto-fix workflow)
- Removed `lint-staged` from pre-commit hook (replaced with `pnpm fix`)
- Removed `tools/eslint-config` package entirely
- Added `oxlint` + `oxfmt` + `ultracite` with preconfigured rule sets for core, React, and Next.js
- Added `.oxlintrc.json` and `.oxfmtrc.jsonc` configs tuned to match our existing code style
- Replaced `prettier-on-automerge.yml` with `ultracite-on-automerge.yml`
- CI now runs `ultracite check` in place of the ESLint and Prettier jobs

## Performance

Measured against the last CI run on `main`:

| Tool | Time |
| --- | --- |
| ESLint | 2m 8s |
| Prettier | 1m 19s |
| **oxlint + oxfmt (Ultracite)** | **36s** |

Combined lint + format time goes from **3m 27s → 36s**, a ~**6x speedup** in CI.

## Why Ultracite

One of the key advantages of Ultracite is that it's designed to run from the repo root, which eliminates the need for per-package lint configs and shared config packages like our old `tools/eslint-config`. Linting and formatting are now two commands run from one place:

- `pnpm check` — runs oxlint + oxfmt check across the whole repo
- `pnpm fix` — runs oxlint + oxfmt fix across the whole repo

Ultracite is already in use across many other repos at Vercel, so this brings the AI SDK in line with the rest of the organisation.

## Why oxlint

oxlint is designed as a 1:1 Rust port of ESLint, which is a big part of why the upfront diff here is small — the rule names, semantics, and even `// eslint-disable` comments are all compatible. Existing disable comments throughout the codebase continue to work as-is with no changes needed.

## On the diff size

This PR shows ~7,000 additions and ~20,000 deletions, but don't be alarmed — roughly 26,000 of those lines are `pnpm-lock.yaml` churn from removing ESLint, Prettier, lint-staged, and all their transitive dependencies. The actual code changes are minimal.

## On the disabled rules

A large number of oxlint rules have been turned off in `.oxlintrc.json` for this initial landing. The goal was to keep the diff reviewable — enabling everything at once would have touched hundreds of files. The goal going forward is to incrementally remove overrides from `.oxlintrc.json`, enabling more of Ultracite's built-in rules for code quality as the codebase is brought into compliance.

## On the modified files

The remaining file changes fall into two categories:

1. Subtle formatting differences between oxfmt and Prettier that couldn't be config-matched away (mainly `class ... implements` placement)
2. Files that were previously excluded from Prettier via `.prettierignore` and are now being properly formatted for the first time


tools/analyze-downloads/package.json
```diff
@@ -8,11 +8,6 @@
     "analyze-market": "tsx src/analyze-market.ts",
     "analyze-ui-frameworks": "tsx src/analyze-ui-frameworks.ts"
   },
-  "lint-staged": {
-    "*": [
-      "prettier --ignore-unknown --write"
-    ]
-  },
```
lgrammel, line 15: this removed pre-commit hook formatting

lgrammel (changes requested):
want to understand impact before merging

.vscode/settings.json
```diff
@@ -1,3 +1,52 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
```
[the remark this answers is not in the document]
haydenbleasel: This is by design. Oxlint/Oxfmt handle a specific set of languages outlined below, so we use Prettier as the generic fallback.

tools/analyze-downloads/package.json
haydenbleasel: Pre-commit formatting, along with regular check and fix commands, are now at the root. They don't need to run per-package like Prettier.

haydenbleasel:
@lgrammel Sure! I've written up a comprehensive PR description but let me know what other information you'd like and I can fill you in. Note: we're using this setup in v0, Turborepo, all our docs sites and more.

lgrammel:
@haydenbleasel does it still format the codebase in the precommit hook?

.github/workflows/ci.yml
```diff
@@ -11,32 +11,8 @@ on:
...
-        with:
-          node-version: 22
-          cache: 'pnpm'
-
-      - name: Install dependencies
-        run: pnpm install --frozen-lockfile
-
-      - name: Build Examples
-        run: pnpm run build:examples
-
```
lgrammel, line 37: does this remove our examples build that we need for thorough checking if there are build failures?

lgrammel (approved):

.github/workflows/ci.yml
gr2m: now runs in line 60:
https://github.com/vercel/ai/pull/13565/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR60

contributing/documentation.md
```diff
@@ -10,7 +10,7 @@ Note: There is currently no way to preview documentation changes locally before
 
 All documentation PRs should be prefixed with `docs:` in the title.
 
-Run `npm run prettier-fix` to solve formatting issues.
+Run `npm run fix` to solve formatting issues.
```
gr2m, line 13: If we use the same in other projects I'm okay with it, but to me "fix" sounds like an llm call to fix code/tests, not formatting. Shouldn't it be something like `npm run fix-format` or similar?

.oxfmtrc.jsonc
```diff
@@ -0,0 +1,46 @@
...
+    "**/*.md",
+    "**/*.mdx",
+    "**/*.mjs",
+    "**/__fixtures__/**",
+    "**/__snapshots__/**",
+    "**/__testfixtures__/**",
+    "**/tsconfig*.json",
+    "**/turbo.json",
+    "tools/tsconfig/**",
+    "packages/codemod/src/test/__testfixtures__/import-LanguageModelV2-from-provider-package.output.ts",
```
gr2m, line 20: shouldn't this be covered by L16 above?

.oxlintrc.json
```diff
@@ -0,0 +1,244 @@
...
+    "**/*.md",
+    "**/*.mdx",
+    "**/*.mjs",
+    "**/__fixtures__/**",
+    "**/__snapshots__/**",
+    "**/__testfixtures__/**",
+    "**/tsconfig*.json",
+    "**/turbo.json",
+    "tools/tsconfig/**",
+    "packages/codemod/src/test/__testfixtures__/import-LanguageModelV2-from-provider-package.output.ts"
```
gr2m, line 242: same

gr2m (commented):
if this is the same setup we have across several @vercel projects, would it be possible / make sense to have a central place to configure oxc in a central place/pkg?

contributing/documentation.md
haydenbleasel: I typically use `fix` as it's a generic command that means `lint` and `format`. It's the same phrasing we use elsewhere but if you'd like to customize it for this repo you can 👍

.oxfmtrc.jsonc
haydenbleasel: It should be but interestingly it complains about that exact file due to the multiple imports. Only way to solve for it without refactoring that test is by specifying the file.

.oxlintrc.json
haydenbleasel: (see other comment)

haydenbleasel:
@lgrammel we'll need to force merge this to workaround the outdated ci checks from main

PR #13576, opened 2026-03-18
fix(anthropic): handle vertex_event SSE events from Vertex AI Anthropic provider

BillionClaw:
## Problem

When using streamText with the Vertex AI Anthropic provider, streaming fails immediately because Vertex sends additional `vertex_event` SSE events that aren't recognized by the SDK.

These events contain usage information that cause type validation errors, resulting in the stream finishing immediately instead of properly streaming.

## Solution

1. Added `vertex_event` to the `anthropicMessagesChunkSchema` with the usage field structure
2. Handle `vertex_event` in the transform stream (ignored like ping events)

This follows the same pattern used by other providers that handle provider-specific SSE events.

## Changes

- packages/anthropic/src/anthropic-messages-api.ts: Added vertex_event schema definition
- packages/anthropic/src/anthropic-messages-language-model.ts: Added handler for vertex_event type

Fixes #13350

### DIFF
--- a/packages/anthropic/src/anthropic-messages-api.ts
+++ b/packages/anthropic/src/anthropic-messages-api.ts
@@ -1320,6 +1320,16 @@ export const anthropicMessagesChunkSchema = lazySchema(() =>
       z.object({
         type: z.literal('ping'),
       }),
+      // Vertex AI sends additional events for Anthropic models that need to be handled
+      z.object({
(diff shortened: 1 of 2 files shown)


Mati0kez:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #13577, opened 2026-03-18
fix(openai-compatible): preserve structured SSE error payloads

theamodhshetty:
## Summary
- preserve the parsed SSE error object in the OpenAI-compatible chat stream instead of collapsing it to `error.message`
- extend the existing chat stream regression to cover `error.code`
- add a patch changeset for `@ai-sdk/openai-compatible`

## Why
Issue #13506 points out that OpenAI-compatible providers can return structured stream errors with a machine-readable `code`, but the chat stream currently drops that field before it reaches consumers. The completion path already forwards the parsed error object, so this brings the chat path back in line with that behavior.

## Testing
- `pnpm --filter @ai-sdk/openai-compatible lint`
- `pnpm --filter @ai-sdk/openai-compatible type-check`
- `pnpm --filter @ai-sdk/openai-compatible test:node -- src/chat/openai-compatible-chat-language-model.test.ts`

### DIFF
--- a/.changeset/nine-pianos-sip.md
+++ b/.changeset/nine-pianos-sip.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai-compatible': patch
+---
+
+Preserve structured SSE error payloads from OpenAI-compatible chat streams instead of collapsing them to the error message string.
--- a/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
+++ b/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
@@ -2683,7 +2683,10 @@ describe('doStream', () => {
           "warnings": [],
         },
         {
-          "error": "Incorrect API key provided: as***T7. You can obtain an API key from https://console.api.com.",
+          "error": {
+            "code": "Client specified an invalid argument",
+            "message": "Incorrect API key provided: as***T7. You can obtain an API key from https://console.api.com.",
+          },
           "type": "error",
         },
         {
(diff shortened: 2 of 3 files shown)


gonzoblasco:
I've been comparing the two fix PRs for this issue (#13508 and #13577). Sharing the analysis:

---

**Approach comparison**

- **#13508** creates a new `{ message, code }` object — explicit but only preserves two fields
- **#13577** passes `chunk.value.error` directly — preserves the full parsed error object, 1-line diff

#13577 is the better approach because:

1. **Consistency** — as noted in the PR body, the completion path already passes the parsed error object. This aligns the chat path with that behavior.
2. **Future-proof** — if the schema later adds fields beyond `message` and `code` (e.g. `type`, `param`), they'll automatically flow through without another fix.
3. **Minimal** — 1 line of logic vs 4, no new object construction.

---

**Concern for both PRs: type breaking change**

Both approaches change the `error` field from a `string` (current: `chunk.value.error.message`) to an `object`. This is a **breaking change** for consumers who currently do:

```ts
if (part.type === 'error') {
  console.error(part.error)        // currently: string
  showToast(part.error)            // string API
}
```

After this fix, `part.error` would be an object `{ message, code, ... }`. Consumers would need:

```ts
if (part.type === 'error') {
  console.error(part.error.message)
  showToast(part.error.message)
}
```

This is arguably a necessary breaking change (the current behavior loses information), but it should be called out explicitly. Consider:

- Bumping to a minor (not just patch) in the changeset, since it changes the error shape
- Adding a migration note in the changeset or release notes
- Or, if backward compat is needed, keeping `error` as the message string and adding a separate `errorCode` or `errorData` field to the error part

---

**Test coverage observation**

The test added in both PRs verifies that `code` is present, but doesn't test what happens when `code` is `null` or `undefined` (the schema allows `code: z.union([z.string(), z.number()]).nullish()`). A test case with a nullish code would document the behavior for providers that don't include a code field.

PR #13581, opened 2026-03-18
fix(provider-utils): chunk array when converting to base64 to avoid excessive memory usage

danielpoonwj:
## Background

With a recent issue disabling passing files as base64 encoded data (https://github.com/vercel/ai/issues/13103), we switched to passing a Buffer and let the library handle base64 encoding.

However we noticed very high memory usage with this method.

### Problem
The previous implementation built the latin1 string character by character in a loop, creating a new string allocation on every iteration. For large arrays this causes O(n²) memory pressure due to repeated string concatenation.

### Solution
Replace with a chunked approach (4096 bytes per chunk, matching [`js-base64`](https://github.com/dankogai/js-base64/blob/main/base64.ts#L65-L75)) that converts each chunk in a single `String.fromCodePoint` call and collects results into an array, joining once at the end before passing to btoa. This avoids both per-byte allocations and the O(n²) string concatenation cost.

### Tests
Added `uint8-utils.test.ts` covering basic encoding/decoding. The suite passes against both the old and new implementations.

## Checklist

### DIFF
--- a/packages/provider-utils/src/uint8-utils.ts
+++ b/packages/provider-utils/src/uint8-utils.ts
@@ -12,12 +12,13 @@ export function convertBase64ToUint8Array(base64String: string) {
 export function convertUint8ArrayToBase64(array: Uint8Array): string {
   let latin1string = '';
 
+  const chunkSize = 4096;
+
   // Note: regular for loop to support older JavaScript versions that
   // do not support for..of on Uint8Array
-  for (let i = 0; i < array.length; i++) {
-    latin1string += String.fromCodePoint(array[i]);
+  for (let i = 0; i < array.length; i += chunkSize) {
+    latin1string += String.fromCodePoint(...array.subarray(i, i + chunkSize));
   }
-
   return btoa(latin1string);
 }
 
--- a/packages/provider-utils/src/uint8-utils.ts
+++ b/packages/provider-utils/src/uint8-utils.ts
@@ -10,16 +10,15 @@ export function convertBase64ToUint8Array(base64String: string) {
 }
 
 export function convertUint8ArrayToBase64(array: Uint8Array): string {
-  let latin1string = '';
-
+  const chunks: string[] = [];
   const chunkSize = 4096;
 
   // Note: regular for loop to support older JavaScript versions that
   // do not support for..of on Uint8Array
   for (let i = 0; i < array.length; i += chunkSize) {
-    latin1string += String.fromCodePoint(...array.subarray(i, i + chunkSize));
+    chunks.push(String.fromCodePoint(...array.subarray(i, i + chunkSize)));
   }
-  return btoa(latin1string);
+  return btoa(chunks.join(''));
 }
 
 export function convertToBase64(value: string | Uint8Array): string {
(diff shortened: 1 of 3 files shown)


aayush-kapoor:
will the upstream fix for #13103 fix your issue? meaning if that is patched, will you still need this fix?

danielpoonwj:
> will the upstream fix for #13103 fix your issue? meaning if that is patched, will you still need this fix?

hi @aayush-kapoor, this is actually independent.

the patch now allows us to send in data pre-converted to base64 to avoid this issue, but if there are any other implementations that send in Buffers for the SDK to encode in base64, this fix is still relevant

lgrammel:
This looks potentially interesting. Do you have any reproduction of the memory issues and proof that this reduces them? Why did you choose 4096 specifically?

packages/provider-utils/src/uint8-utils.ts
```diff
@@ -10,15 +10,15 @@ export function convertBase64ToUint8Array(base64String: string) {
...
   // Note: regular for loop to support older JavaScript versions that
   // do not support for..of on Uint8Array
-  for (let i = 0; i < array.length; i++) {
-    latin1string += String.fromCodePoint(array[i]);
+  for (let i = 0; i < array.length; i += chunkSize) {
+    chunks.push(String.fromCodePoint(...array.subarray(i, i + chunkSize)));
   }
-
-  return btoa(latin1string);
+  return btoa(chunks.join(''));
```
lgrammel, line 21: what is the max arg limit for `String.fromCodePoint`? ie why do we need the loop at all

janvennemann:
We hit this exact issue in production with a Node worker OOM'd encoding ~18 MB PDF attachments before handing them to the SDK. We solved it by using the native `Buffer` conversion and directly passing that into the AI SDK to bypass the loop conversion. Here are some benchmarks to see the impact.

**Repro** (Node v24.16, arm64, real 17.9 MB PDF), peak RSS via `/usr/bin/time -l`. All three produce byte-identical base64 (SHA-256 verified):

| Strategy | Peak RSS | Encode time |
| --- | --- | --- |
| `old` (per-byte concat, current) | **787 MB** | 802 ms |
| `chunked` (this PR) | **236 MB** | 464 ms |
| `Buffer.from(a).toString("base64")` (ref) | 106 MB | 5 ms |

packages/provider-utils/src/uint8-utils.ts
janvennemann: `String.fromCodePoint(...array)` spreads each byte as a separate argument, which blows the call stack well below attachment sizes. A few checks we did for our ~18MB PDF test file:

```
4096      ok
65536     ok
131072    RangeError: Maximum call stack size exceeded
whole file (18.7M)  RangeError
```

safaiyeh:
Filed vercel/ai#20149 with an independent production confirmation of this PR's numbers: one 14MB PDF file part OOM-killed a 2GB Fargate container in ~40s through the current per-byte encoder (+606MB / 392ms for the encode alone; 1,065MB peak through the full openai provider request path with two PDFs). It also carries the answers already posted here — the spread's ~131k-argument RangeError ceiling and the chunk-size rationale — so hopefully it helps this get a second look. Linking per CONTRIBUTING so the issue and fix credit connect.

PR #13594, opened 2026-03-18
fix(amazon-bedrock): preserve reasoning text verbatim when signature is present

guoyangzhen:
## Summary

Fixes #13583

When a cryptographic `signature` is present on a reasoning block, `trimIfLast()` must **not** be applied to the text. The signature is computed over the exact original bytes of the reasoning text — trimming even a single trailing whitespace character invalidates it, causing Bedrock to reject the request with:

```
ValidationException: thinking or redacted_thinking blocks in the latest assistant message cannot be modified.
```

## Root Cause

In `packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.ts`, the `reasoning` case unconditionally applies `trimIfLast()` to reasoning text, even when a `signature` is present:

```typescript
text: trimIfLast(
  isLastBlock,
  isLastMessage,
  isLastContentPart,
  part.text,
),
signature: reasoningMetadata.signature,
```

## Fix

Skip `trimIfLast()` when `reasoningMetadata.signature != null`, returning `part.text` verbatim:

```typescript
text: part.text,
signature: reasoningMetadata.signature,
```

The `trimIfLast()` optimization for pre-filled assistant responses does not apply to reasoning blocks in conversation history — they must be returned exactly as received.

## Testing

- Updated existing test `should trim trailing whitespace from reasoning content when it is the last part` → `should NOT trim reasoning text when a cryptographic signature is present` to expect verbatim text preservation.
- Added new test `should trim trailing whitespace from reasoning content without signature when it is the last part` to verify trimming still works when no signature is present.

## Verification

With this fix, multi-turn conversations using Claude extended thinking (`reasoning: { type: "enabled" }`) on Amazon Bedrock will no longer fail with `ValidationException` when passing reasoning blocks back to the API.

### DIFF
--- a/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.test.ts
+++ b/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.test.ts
@@ -669,7 +669,11 @@ describe('assistant messages', () => {
     });
   });
 
-  it('should trim trailing whitespace from reasoning content when it is the last part', async () => {
+  it('should NOT trim reasoning text when a cryptographic signature is present', async () => {
+    // The signature is computed over the exact original bytes of the text.
+    // Trimming even a single trailing whitespace invalidates the signature,
+    // causing Bedrock to reject the request with a "thinking blocks cannot
+    // be modified" error.
     const result = await convertToBedrockChatMessages([
       {
         role: 'user',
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
https://github.com/vercel/ai/pull/13972

Issue #13570, opened 2026-03-17
Improve DX of custom loop control


gr2m:
This is part of our effort to improve compatibility with Workflow (#12165) and at the same time introduce better support for custom/external loop control of streaming responses.

To achieve that goal, we are separating out individual reusable functions from streamText. Ideally, `streamText()` becomes a composite method of lower level methods and is only responsible to manage the singular stitch stream returned by the method.

### Plan

1. Refactor streamText() to have a single place for LanguageModel to MessageModel transformations
2. ... tbd

### Refactor transformations to a single place

Here is a dummy code to show case how `streamText()` works today, taken form #13541 

```
initial model messages
response model messages
stitchable stream

do {
  prepare step
  convert step input messages (after prepare step) to language model v4 messages

 stream = doStream (with language model v4 messages)

 run tools transformation on stream
   transforms spec format to user friendly format
   executes tools and injects tool results into stream
   tool approval?

 pipe stream with tool results through further transforms
   transforms:
     add start-step
     filter out empty text chunks
     tool input start
     filter raw chunks when not enabled
     add finish-step
     add finish
   bookkeeping
   timeout mgmt
   events:
     telemetry, tool input delta

 add transformed stream with minor augmentation to stitchable stream

 add new response model messages
   by converting the assembled step output
   to additional response model messages

} while (
    not (
        any of the stop conditions is met
        or
        finish reason is not tool-calls
        or
        tool without execute is called
        or
        tool that needs approval is called
        or
        there are deferred tool calls
    )
)

unified stream
stream 1 -- stream 2 -- stream 3
```

Transforms are happening at different places. 

The general transform flow is

1. streamText receives a [ModelMessage](https://github.com/vercel/ai/blob/main/packages/provider-utils/src/types/model-message.ts) array
2. it converts it to [LanguageModelV4Message](https://github.com/vercel/ai/blob/main/packages/provider/src/language-model/v4/language-model-v4-prompt.ts)
3. the `model.doStream()` method is called which also returns `LanguageModelV4Message`
4. mix of injecting new chunks, filtering out chunks, transforms, calling hooks, tool handling, tool calling

We can't do the transform entirely right after we receive the stream result from `model.doStream()` since we need the finish chunk from the llm call at the end of the `streamText()` loop step, and maybe other reasons, but we will try to find a way around it.

As part of that work, we should be able to remove the duplicated code between

[...]

PR #13596, opened 2026-03-18
refactoring(ai): move reasoning delta transformation into earlier transform

lgrammel:
## Background

We are working towards better support for custom loop control. For this, we are separating out individual reusable functions from streamText. As a first step, we need to separate transformation steps such that each transformation has a single responsibility.

## Summary

Move `reasoning-delta`, `file`, `reasoning-file` into earlier transform.

## Related issues

towards https://github.com/vercel/ai/issues/13570


gr2m (approved):
we paired on it

PR #13388, opened 2026-03-13
feat: add deferToolExecution option to streamText

gr2m:
## Background

When building workflow/durable agent compatibility, tool execution must happen after the LLM response stream concludes. Currently, `streamText` executes tool calls immediately as their chunks arrive from the provider stream (fire-and-forget without `await`). This forces consumers like Vercel Workflow's `DurableAgent` to reimplement significant portions of the stream-text pipeline to achieve deferred execution, resulting in code duplication. See: [vercel/workflow `do-stream-step.ts`](https://github.com/vercel/workflow/blob/d6da7ecca0fbf94d3331ce3bd7d28644a4f5cc3f/packages/ai/src/agent/do-stream-step.ts#L105).

## Summary

Adds a `deferToolExecution` option to `streamText` (default: `false`) that controls when tool calls are executed:

- **`false` (default)**: Existing behavior — tools execute immediately as their `tool-call` chunks arrive from the provider stream, concurrently with remaining stream processing.
- **`true`**: Tool calls are parsed and emitted to the stream as normal (consumers still see `tool-call` chunks in real-time), but `executeToolCall()` is deferred until the provider stream's `flush()` fires (i.e., after the LLM response completes). All buffered tool calls are then executed in parallel.

#### `deferToolExecution: false` (default — current behavior)

```mermaid
sequenceDiagram
    participant Provider as AI Provider Stream
    participant Transform as runToolsTransformation
    participant Tool as executeToolCall
    participant Consumer as Stream Consumer

    Provider->>Transform: text-delta chunks
    Transform->>Consumer: forward text-delta
    Provider->>Transform: tool-call chunk (tool A)
    Transform->>Consumer: emit tool-call (tool A)
    Transform->>Tool: executeToolCall(A) — fire & forget
    Note over Tool: Tool A starts executing
    Provider->>Transform: tool-call chunk (tool B)
    Transform->>Consumer: emit tool-call (tool B)
    Transform->>Tool: executeToolCall(B) — fire & forget
    Note over Tool: Tool B starts executing
    Provider->>Transform: finish chunk
    Note over Provider: Stream ends
    Tool-->>Transform: tool A result
    Transform->>Consumer: emit tool-result (A)
    Tool-->>Transform: tool B result
    Transform->>Consumer: emit tool-result (B)
    Transform->>Consumer: emit finish
```

#### `deferToolExecution: true`

```mermaid
sequenceDiagram
    participant Provider as AI Provider Stream
    participant Transform as runToolsTransformation
[...]

PR #13634, opened 2026-03-19
feat(ai): defer tool execution until after LLM response completes

gr2m:
## Summary

- Tool calls are now buffered during streaming and executed in parallel after the LLM response finishes
- This cleanly separates the LLM generation phase from the tool execution phase
- Implements the change described in #13570 (applying #13388 without the flag)

**Breaking change**: Tool results will no longer appear interleaved with provider stream chunks — they always appear after the provider stream finishes.

## AI notes

Here's what was done:

Core change in execute-tools-transformation.ts: Instead of executing tool calls immediately (fire-and-forget) as tool-call chunks arrive from the provider stream, tool calls are now
buffered in a deferredToolCalls array and executed in parallel in the flush() method after the generator stream finishes.

Test updates:
- stream-text.test.ts: Updated the "abort during tool call" inline snapshot — with deferred execution, the tool-call chunk now appears in the stream before the abort (since it's
forwarded during streaming, but execution is deferred)
- stream-text.test.ts.snap: Updated telemetry snapshot (tool result now reflects transformed input since execution happens after the transform completes)

## Test plan

- [x] All 15 `execute-tools-transformation` tests pass
- [x] All 867 `stream-text` tests pass (2 snapshot updates for changed behavior)
- [ ] Manual verification with example apps using tool calls

### DIFF
--- a/.changeset/defer-tool-execution.md
+++ b/.changeset/defer-tool-execution.md
@@ -1,0 +1,9 @@
+---
+'ai': patch
+---
+
+Defer tool execution until after the LLM response stream completes.
+
+Previously, tool calls were executed immediately (fire-and-forget) as `tool-call` chunks arrived from the provider stream. Now, tool calls are buffered during streaming and executed in parallel after the LLM response finishes. This cleanly separates the LLM generation phase from the tool execution phase.
+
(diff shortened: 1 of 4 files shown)


gr2m:
superseded by #13638

PR #13638, opened 2026-03-19
refactoring(ai): delay tool execution in stream text until model call is finished

lgrammel:
## Background
We are working towards better support for custom loop control.
Until now, `streamText` starts tool execution immediately when a tool call arrives.
While this enables tools to be executed without delay, it prevents clear separation of model calls from tool execution, which is required for external loop control, e.g. in the Workflow SDK.

## Summary

Delay tool execution until model call sends `finish` chunk (**breaking behavior change**)

## Related Issues

Towards #13570


gr2m (approved):

Snapp949:
Y

PR #13536, opened 2026-03-17
feat(ai): add per-tool timeout overrides via toolTimeouts

dancer:
## background

pr #13365 added generic `toolMs` to timeout configuration. this adds per-tool overrides so users can set different timeouts for different tools with full type safety

## summary

- add `tools` object to `TimeoutConfiguration` for per-tool timeout overrides using `{toolName}Ms` pattern
- type safety via template literal types: `${keyof TOOLS}Ms` gives autocomplete for tool names
- `TimeoutConfiguration` is generic, typed with `NoInfer<TOOLS>` at `generateText`/`streamText` level
- resolution order: per-tool (`tools.weatherMs`) > generic (`toolMs`) > no timeout
- uses `Omit<CallSettings, 'timeout'>` to avoid intersection conflict with the generic timeout type

```ts
generateText({
  model: openai('gpt-5.4'),
  tools: { weather: weatherTool, slowApi: slowApiTool },
  timeout: {
    toolMs: 5000,
    tools: {
      weatherMs: 3000,
      slowApiMs: 1000,
    },
  },
  prompt: 'What is the weather in tokyo?',
});
```

## verification

<details>
<summary>type safety - invalid tool names caught</summary>

```
Object literal may only specify known properties, and 'typoMs' does not
exist in type 'Partial<Record<"weatherMs" | "slowApiMs", number>>'
```

</details>

<details>
<summary>per-tool example - fastApi succeeds, slowApi times out</summary>

```
$ pnpm tsx src/generate-text/openai/tool-timeout-per-tool-error.ts
tool-call: fastApi({"query":"hello"})
tool-call: slowApi({"query":"hello"})
tool-error: slowApi timed out

Results for "hello":
- Fast API: hello
- Slow API: timed out / operation aborted
```

</details>

## checklist

- [x] tests have been added / updated (for bug fixes / features)
- [x] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)
- [x] i have reviewed this pull request (self-review)

## related issues

fixes #13279

PR #13641, opened 2026-03-19
chore(ai/core): remove `timeout` from `CallSettings` as it was effectively unused there

felixarntz:
## Background

In #13536, `CallSettings<TOOLS>` was made generic solely because of `timeout` (`TimeoutConfiguration<TOOLS>` uses tool-specific keys). I looked into making `prepareCallSettings` generic too as a result (see https://github.com/vercel/ai/pull/13536/changes#r2961226981), but that led me to a more important finding: `timeout` is never accessed through a `CallSettings`-typed variable — it's always destructured independently and handled via standalone helper functions. This generic propagates through many files unnecessarily, and `timeout` being part of `CallSettings` is probably no longer relevant.

## Summary

- Remove `timeout` from `CallSettings` and make it non-generic. In `generateText`, `streamText`, and `ToolLoopAgentSettings`, `timeout` is added as a standalone property in the object literal type, preserving the public API while keeping the tool-specific generic only where it's actually needed.
- Remove dead timeout-handling code in `getBaseTelemetryAttributes` that was unreachable since no caller ever passed `timeout` in the settings object.

Given that `CallSettings` is exported, this can be considered a low-severity breaking change. But we're working in the v7 branch, so this is fine.

## Manual Verification

N/A — internal type refactor, verified via type checking and existing test suites.

## Checklist

## Future Work

- A few places didn't actually support passing through `timeout` which maybe should allow it (e.g. `streamUI()`)
    - before it wasn't possible because `prepareCallSettings` would strip `timeout`, now `timeout` is formally removed from `CallSettings`
    - possibly something to fix, but worth noting this PR doesn't make it worse
- Consider breaking out `abortSignal`, `headers`, and `maxRetries` from `CallSettings` as well.
- Potentially create a separate `RequestSettings` type that contains these 3, plus `timeout`. Could then also rename `CallSettings` to something more descriptive, e.g. `ModelCallSettings`.

## Related Issues

N/A

### DIFF
--- a/packages/ai/src/agent/tool-loop-agent-settings.ts
+++ b/packages/ai/src/agent/tool-loop-agent-settings.ts
@@ -17,7 +17,7 @@ import { PrepareStepFunction } from '../generate-text/prepare-step';
 import { StopCondition } from '../generate-text/stop-condition';
 import { ToolCallRepairFunction } from '../generate-text/tool-call-repair-function';
 import { ToolSet } from '../generate-text/tool-set';
-import { CallSettings } from '../prompt/call-settings';
+import { CallSettings, TimeoutConfiguration } from '../prompt/call-settings';
 import { Prompt } from '../prompt/prompt';
 import { TelemetrySettings } from '../telemetry/telemetry-settings';
 import { LanguageModel, ToolChoice } from '../types/language-model';
@@ -57,7 +57,15 @@ export type ToolLoopAgentSettings<
   CALL_OPTIONS = never,
   TOOLS extends ToolSet = {},
   OUTPUT extends Output = never,
-> = Omit<CallSettings<TOOLS>, 'abortSignal'> & {
+> = Omit<CallSettings, 'abortSignal'> & {
+  /**
+   * Timeout in milliseconds. The call will be aborted if it takes longer
+   * than the specified timeout. Can be used alongside abortSignal.
+   *
+   * Can be specified as a number (milliseconds) or as an object with `totalMs`.
+   */
+  timeout?: TimeoutConfiguration<TOOLS>;
+
   /**
    * The id of the agent.
    */
--- a/packages/ai/src/telemetry/get-base-telemetry-attributes.ts
+++ b/packages/ai/src/telemetry/get-base-telemetry-attributes.ts
@@ -1,5 +1,5 @@
 import { Attributes, AttributeValue } from '@opentelemetry/api';
-import { CallSettings, getTotalTimeoutMs } from '../prompt/call-settings';
+import { CallSettings } from '../prompt/call-settings';
 import { TelemetrySettings } from './telemetry-settings';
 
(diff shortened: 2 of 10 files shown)


packages/ai/src/telemetry/get-base-telemetry-attributes.ts
```diff
@@ -19,17 +19,7 @@ export function getBaseTelemetryAttributes({
...
-        const totalTimeoutMs = getTotalTimeoutMs(
-          value as Parameters<typeof getTotalTimeoutMs>[0],
-        );
-        if (totalTimeoutMs != null) {
-          attributes[`ai.settings.${key}`] = totalTimeoutMs;
-        }
-      } else {
-        attributes[`ai.settings.${key}`] = value as AttributeValue;
-      }
+      attributes[`ai.settings.${key}`] = value as AttributeValue;
```
felixarntz, line 22: @lgrammel Unsure why this is here - it looks like it was added in #11597.

While technically `CallSettings` _could_ have included `timeout` before, in practice based on the logic that pass through to `getBaseTelemetryAttributes`, it was never actually passed:

- either the data was passed manually and didn't include `timeout`
- or `timeout` was destructured from the `CallSettings` data before
- or the `CallSettings` passed went through `prepareCallSettings()` before, in which case `timeout` would have been dropped anyway

Not entirely sure whether we should remove this, or fix it some other way to record `timeout`.

packages/ai/src/agent/tool-loop-agent-settings.ts
```diff
@@ -17,7 +17,7 @@ import { PrepareStepFunction } from '../generate-text/prepare-step';
 import { StopCondition } from '../generate-text/stop-condition';
 import { ToolCallRepairFunction } from '../generate-text/tool-call-repair-function';
 import { ToolSet } from '../generate-text/tool-set';
-import { CallSettings } from '../prompt/call-settings';
+import { CallSettings, TimeoutConfiguration } from '../prompt/call-settings';
```
lgrammel, line 20: should we call the interfaces `XSettings` or `XConfiguration`? fine with either, just would prefer it to be consistent

packages/ai/src/telemetry/get-base-telemetry-attributes.ts
lgrammel: we will probably redesign/omit this for otel. can remove if not used. cc @aayush-kapoor

lgrammel (approved):

packages/ai/src/agent/tool-loop-agent-settings.ts
felixarntz: makes sense. I'll do that in a separate PR though, together with re-thinking / splitting "model call settings" from "request settings" like we talked about. This way we can cover the adjustments holistically

PR #13682, opened 2026-03-20
feat(ai): extract streamModelCall function for streaming text generation

lgrammel:
## Background

We want to introduce custom loop controls for streamText. Calling the LLM for a single step without tool calling or event notifications is one key component.

## Summary

* Extract `streamModelCall` function and expose it as `experimental_streamModelCall`
* Add barrier to prevent further tokens from being emitted on a transform calls `stopStream`

## Related Issues
#13570


gr2m:
<details><summary>Claude's assessment</summary>

Here's the root cause of the abort signal test failures:

  Root Cause

  The issue is a subtle microtask timing difference caused by extracting pipeThrough(createStreamTextPartTransform) into the async function streamModelCall.

  Original code (working)

  After await retry(() => doStream()), everything happens synchronously in a single block:
  pipeThrough(createStreamTextPartTransform)  →  invokeToolCallbacksFromStream  →  pipeThrough(createExecuteToolsTransformation)  →  addStream(...)

  There's no microtask boundary between pipeThrough(createStreamTextPartTransform) and addStream. The internal pipe from modelStream → createStreamTextPartTransform is set up but doesn't
  execute until the consumer starts pulling through the stitchable stream.

  PR code (broken)

  pipeThrough(createStreamTextPartTransform) is called inside the async function streamModelCall, right before return. Because streamModelCall is async, its return implicitly wraps the
  result in a resolved Promise. The caller's await streamModelCall(...) introduces a microtask boundary before the remaining sync operations (invokeToolCallbacksFromStream → addStream).

  During this microtask boundary, the pipe from modelStream → createStreamTextPartTransform (set up by pipeThrough) gets a chance to run and read ahead from the model stream. This changes
  the internal buffering state of the stream pipeline. By the time addStream is called and the resilient stream starts reading, more data is already buffered in the pipeline, which means
  the abort error from the model (triggered at pull 3) doesn't propagate fast enough to prevent extra chunks from being read.

  Concrete execution trace comparison

  Original: After the resilient stream reads start-step, its next pull triggers model pull 3 (abort) as a side-effect, and the error propagates before text-start can be read from the
  buffer → abort detected immediately.

  PR: After the resilient stream reads start-step, the buffering state is different due to the microtask gap. text-start is successfully read from the buffer before model pull 3 triggers →
   one extra chunk appears before abort.

  Fix

  The fix is to keep pipeThrough(createStreamTextPartTransform) in the same synchronous block as addStream. Either:

  1. Don't include pipeThrough in streamModelCall — have streamModelCall return the raw languageModelStream and let the caller pipe through createStreamTextPartTransform synchronously
  alongside the other transforms.
  2. Move all the remaining pipe operations into streamModelCall — so that everything from pipeThrough through addStream is synchronous within the same function.
  3. Don't make streamModelCall async — this isn't really feasible since it needs to await several operations internally.

  Option 1 is the simplest and most correct fix.

</details>

lgrammel:
This breaks LLM suspense stopStream behavior. There may be fixes for that that we can explore further.

gr2m:
As a next step, we will look deeper into stitchable stream and abort timing.

packages/ai/src/generate-text/stream-text.test.ts
```diff
@@ -17929,9 +17924,6 @@ describe('streamText', () => {
               {
                 "type": "finish-step",
               },
-              {
-                "type": "start-step",
-              },
```
gr2m, line 17934: both of these chagnes are abort-timing related, no problem

gr2m (approved):

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `ai` | [`7.0.0-beta.35`](https://www.npmjs.com/package/ai/v/7.0.0-beta.35) |
| `@ai-sdk/angular` | [`3.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Fangular/v/3.0.0-beta.35) |
| `@ai-sdk/azure` | [`4.0.0-beta.15`](https://www.npmjs.com/package/%40ai-sdk%2Fazure/v/4.0.0-beta.15) |
| `@ai-sdk/google` | [`4.0.0-beta.16`](https://www.npmjs.com/package/%40ai-sdk%2Fgoogle/v/4.0.0-beta.16) |
| `@ai-sdk/google-vertex` | [`5.0.0-beta.21`](https://www.npmjs.com/package/%40ai-sdk%2Fgoogle-vertex/v/5.0.0-beta.21) |
| `@ai-sdk/langchain` | [`3.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Flangchain/v/3.0.0-beta.35) |
| `@ai-sdk/llamaindex` | [`3.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Fllamaindex/v/3.0.0-beta.35) |
| `@ai-sdk/openai` | [`4.0.0-beta.15`](https://www.npmjs.com/package/%40ai-sdk%2Fopenai/v/4.0.0-beta.15) |
| `@ai-sdk/react` | [`4.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Freact/v/4.0.0-beta.35) |
| `@ai-sdk/rsc` | [`3.0.0-beta.36`](https://www.npmjs.com/package/%40ai-sdk%2Frsc/v/3.0.0-beta.36) |
| `@ai-sdk/svelte` | [`5.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Fsvelte/v/5.0.0-beta.35) |
| `@ai-sdk/vue` | [`4.0.0-beta.35`](https://www.npmjs.com/package/%40ai-sdk%2Fvue/v/4.0.0-beta.35) |

PR #13721, opened 2026-03-22
fix(provider/google): preserve minItems, maxItems, minimum, maximum, maxLength in schema conversion

JoshuaLelon:
## Summary

The `convertJSONSchemaToOpenAPISchema` function silently drops several JSON Schema properties that are [documented as supported](https://ai.google.dev/gemini-api/docs/json-mode) by the Gemini API:

- **`minItems` / `maxItems`** — array length constraints
- **`minimum` / `maximum`** — numeric range constraints
- **`maxLength`** — string length constraint (`minLength` was already preserved)

This prevents users from enforcing exact array lengths in structured outputs. For example, `z.array(schema).length(10)` generates `minItems: 10, maxItems: 10` in JSON Schema, but both are stripped before reaching the Gemini API.

### Gemini API support

From the [official docs](https://ai.google.dev/gemini-api/docs/json-mode) (Type-specific properties section):

> For `array` values, you can specify `items`, `prefixItems` for tuple-like structures, and constraints on the number of items using `minItems` and `maxItems`.

> `number` and `integer` types allow for `enum` values as well as `minimum` and `maximum` constraints.

### The fix

Added the missing properties to the destructuring whitelist in `convertJSONSchemaToOpenAPISchema` and pass them through to the result object. This is consistent with how `minLength` is already handled.

## Test plan

- [x] Added test: `should preserve minItems and maxItems on arrays`
- [x] Added test: `should preserve exact array length (minItems === maxItems)`
- [x] Added test: `should preserve minimum and maximum on numbers`
- [x] Added test: `should preserve maxLength on strings`
- [x] All 23 tests in `convert-json-schema-to-openapi-schema.test.ts` pass
- [x] Changeset added (patch)

### DIFF
--- a/.changeset/preserve-array-numeric-constraints.md
+++ b/.changeset/preserve-array-numeric-constraints.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Preserve `minItems`, `maxItems`, `minimum`, `maximum`, and `maxLength` JSON Schema properties when converting to OpenAPI schema for Google Generative AI. These properties are supported by the Gemini API but were previously silently dropped during schema conversion, preventing users from enforcing array length and numeric range constraints in structured outputs.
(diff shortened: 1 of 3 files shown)


TomasWard1:
Implement it please!

Repomn:
any updates?

n1ckoates:
Google provider now uses JSON schema instead of OpenAPI, so this isn't necessary anymore. These fields should just be natively supported now. Thanks!

PR #13779, opened 2026-03-23
feat: add workflow serialization support to all provider models

gr2m:
## Summary

- Add `WORKFLOW_SERIALIZE` / `WORKFLOW_DESERIALIZE` static methods to all 59 provider model classes (language, image, embedding, speech, transcription, video)
- Add `serializeModel()` helper to `@ai-sdk/provider-utils` that automatically extracts only serializable config properties, filtering out functions and objects containing functions
- Make `headers` optional in all provider config types so deserialized models work without pre-configured auth

### Changes affecting users outside of workflow

- **`headers` is now optional** in all provider model config types. This is a non-breaking change — existing code that passes `headers` continues to work identically. Code that constructs model configs manually (e.g., custom provider implementations) can now omit `headers`.
- **New `serializeModel()` export** from `@ai-sdk/provider-utils` — third-party provider authors can use this to add workflow serialization support to their own models.
- **New `WORKFLOW_SERIALIZE` / `WORKFLOW_DESERIALIZE` re-exports** from `@ai-sdk/provider-utils` (via `@workflow/serde`) — symbols needed by providers to implement custom serialization.

### How third-party providers can add support

```typescript
import {
  serializeModel,
  WORKFLOW_SERIALIZE,
  WORKFLOW_DESERIALIZE,
} from '@ai-sdk/provider-utils';

export class MyLanguageModel {
  // classId is generated by the workflow SWC compiler at build time

  static [WORKFLOW_SERIALIZE](inst: MyLanguageModel) {
    return serializeModel(inst);
  }

  static [WORKFLOW_DESERIALIZE](options: {
    modelId: string;
    config: MyConfig;
  }) {
    return new MyLanguageModel(options.modelId, options.config);
  }
}
```

The only additional requirement is making `headers` optional in the config type and using optional chaining (`this.config.headers?.()`) at call sites.

## Test plan

- [x] All existing provider tests pass (verified per-package builds)
- [x] `serializeModel` unit tests added in provider-utils
- [x] Manual testing with next-workflow example confirms serialization error is resolved


packages/anthropic/package.json
```diff
@@ -61,8 +61,14 @@
     "zod": "3.25.76"
   },
   "peerDependencies": {
+    "workflow": "^4.2.0-beta.64",
```
gr2m: Claude's analysis on this

⏺ The peer dependency is needed because of a chain reaction:

  1. We import WORKFLOW_SERIALIZE/WORKFLOW_DESERIALIZE from @workflow/serde (via @ai-sdk/provider-utils)
  2. The workflow SWC plugin detects those imported symbols on the class at build time
  3. It injects import { registerSerializableClass } from 'workflow/internal/class-serialization'
  4. That import must resolve from the anthropic package's location — requiring workflow as a dependency

lgrammel (approved):

PR #13787, opened 2026-03-24
fix: use passthrough() in uiMessageChunkSchema for forward compatibility

majiayu000:
## Background

Issue #13733: `uiMessageChunkSchema` uses `z.strictObject()` on all 27 discriminated union variants, which rejects any unrecognized keys with `AI_TypeValidationError`. When a newer server adds a field (e.g. `providerMetadata`), older cached clients break because the strict schema won't accept the extra key.

## Summary

- Replaced `z.strictObject()` with `z.object().passthrough()` on all 27 variants in `uiMessageChunkSchema`
- Added regression tests verifying unknown fields pass through on text-delta, start-step, finish, and error chunk types
- Added specific test for `providerMetadata` on error chunks (the exact scenario from #13733)
- Updated type test to use `toMatchTypeOf` since `passthrough()` adds an index signature to the inferred type

## Manual Verification

Ran `pnpm test` in `packages/ai` — all 2268 tests pass (node + edge), no type errors.

## Checklist

## Related Issues

Fixes #13733

### DIFF
--- a/.changeset/fix-ui-message-chunk-passthrough.md
+++ b/.changeset/fix-ui-message-chunk-passthrough.md
@@ -1,0 +1,10 @@
+---
+'ai': patch
+---
+
+fix(ai): use passthrough() in uiMessageChunkSchema for forward compatibility
+
+Replaced z.strictObject() with z.object().passthrough() in uiMessageChunkSchema to prevent
+AI_TypeValidationError when a newer server sends fields (e.g. providerMetadata) that an older
+cached client schema doesn't recognize. This makes the stream protocol forward-compatible
+across patch/minor versions.
--- a/packages/ai/src/ui-message-stream/ui-message-chunks.test-d.ts
+++ b/packages/ai/src/ui-message-stream/ui-message-chunks.test-d.ts
@@ -13,6 +13,8 @@ describe('UI message chunks type', () => {
       },
     });
 
-    expectTypeOf(chunk).toEqualTypeOf<UIMessageChunk>();
+    // toMatchTypeOf (not toEqualTypeOf) because passthrough() adds an index
+    // signature to the inferred type, making it wider than the UIMessageChunk union.
+    expectTypeOf(chunk).toMatchTypeOf<UIMessageChunk>();
   });
 });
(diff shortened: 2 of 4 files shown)


spinsirr:
Adding a real-world report on top of #13733's "old client + new server" framing, since the same `strictObject` problem also surfaces in a **new client + new server** setup as soon as Gemini thinking is in play.

**Stack:** `ai@6.0.168`, `@ai-sdk/google@3.0.64`, `@ai-sdk/react@3.0.170`, model `gemini-3-flash-preview` with thinking enabled, multi-turn tool calls.

**What happens**

1. Gemini returns a `functionCall` with `thoughtSignature` attached. The provider surfaces it on the SSE chunk's `providerMetadata` (e.g. on `tool-input-available`).
2. Even with a brand-new client, the strict `uiMessageChunkSchema` rejects the chunk because `providerMetadata.google.thoughtSignature` is an unrecognized inner key — same root cause as #13733, one level deeper.
3. To keep the UI stream alive we ended up stripping `providerMetadata` from chunks server-side. That removes the `thoughtSignature` on the round trip.
4. On the next turn the client sends back the message history without signatures, and Gemini responds with:

   > Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:list_vehicle_services`, position 12.

So #13787 isn't only a forward-compat fix — without it, Gemini multi-turn tool use silently regresses in model quality even on the latest client. We're running the strip workaround in production today and can confirm it makes the UI stream work but kills signature continuity; the `passthrough()` change in this PR would let us drop the workaround entirely.

@lgrammel anything blocking review here? Happy to put together a minimal repro repo if helpful.

spinsirr:
@majiayu000 thanks for putting this together. There's a small merge conflict from #14659, which landed on main and added 3 new chunk variants (`tool-approval-request`, `tool-approval-response`, `tool-output-denied`) that still use `strictObject`. The fix is mechanical — just switch them to the same `passthrough` form as the others.

Are you able to rebase, or would it help if I opened a successor PR with credit back to you?

lgrammel:
Thank you for the pull request. Issue #13733 was closed as completed. Closing this pull request.

Issue #12855, opened 2026-02-25
Add support for creating and managing remote skills (e.g. Anthropic, OpenAI) via abstraction with new top-level API


felixarntz:
Both Anthropic and OpenAI offer Skills management APIs (CRUD for provider-hosted skill bundles). The SDK supports skills at **inference time** via provider options, but not the management side (create, list, retrieve, update, delete). This would be the SDK's first non-inference provider API surface.

## Provider comparison

Both providers share the same core model: upload files with a `SKILL.md` manifest via multipart form → get a skill ID for inference. Both support versioning (updating = creating a new version). Main differences: OpenAI has dual version pointers (`latest_version` + `default_version`, enabling staged rollouts), Anthropic has a single pointer (`latest_version`, new versions immediately active). Anthropic supports an optional `display_title` on create; OpenAI accepts only files.

References:
* https://platform.claude.com/docs/en/build-with-claude/skills-guide#creating-a-skill
* https://developers.openai.com/api/docs/guides/tools-skills/#create-a-skill

## Proposal

Add five `experimental_` top-level functions to the `ai` package, following the existing pattern (public functions → spec interface → provider implementation):

```typescript
import { experimental_createSkill, experimental_listSkills, experimental_retrieveSkill,
         experimental_updateSkill, experimental_deleteSkill } from 'ai';

const skill = await experimental_createSkill({
  provider: openai,
  files: [{ path: 'SKILL.md', content: base64EncodedContent }],
});
await experimental_updateSkill({ provider: openai, skillId: skill.id, files: [...] });
await experimental_deleteSkill({ provider: openai, skillId: skill.id });
```

**Architecture** (three layers, no new packages):
1. `@ai-sdk/provider`: New `SkillsManagerV1` interface + optional `skillsManager?()` on `ProviderV3`
2. `@ai-sdk/openai` / `@ai-sdk/anthropic`: Implement `SkillsManagerV1`
3. `packages/ai/src/skills/`: Top-level `experimental_` functions that delegate to the provider's skills manager

**Additional notes:**
- **Version management is hidden:** `updateSkill` creates a new version and promotes it to default internally. No version IDs exposed to users.
  - We can consider adding version management in the future without breaking back-compat, but it feels too niche for the common use-case for adding complexity to the public API, at least for the start.
- **Infrastructure needs:** Add `deleteFromApi()` to `@ai-sdk/provider-utils` (no `DELETE` utility exists today).

## Open question: file content format

[...]

PR #13816, opened 2026-03-24
feat(provider): add support for provider references and uploading files as supported per provider

### LINKED ISSUE #12995: Add support for uploading media files (e.g. images, audio, documents) via abstraction with new top-level API
For the longest time, the AI SDK has supported passing media files either inline or via a URL. The latter allows referencing files uploaded to certain providers, however it has not been possible to upload those files directly through the AI SDK. Additionally, it is not guaranteed that every provider's upload API returns a URL - some may return an internal file ID.

The AI SDK should support a top-level API like `uploadFile` and a provider specific implementation (e.g. a `Files` type with `uploadFile` method).

In order to keep the abstraction provider independent, we need to implement a system that can reference the uploaded files per provider: For example, if a file is uploaded to Google and used in a chat with a Google model, but the chat where the file is used is later continued with an Anthropic model, we need to allow for a way to include multiple file references for the same file in the message history. For example something like this:
```ts
type ProviderIndependentID = {
    [providerName]: string;
};
```

This structure will allow the provider to check whether there's an ID for it specifically. Otherwise it can return an error. The developer will then have to upload the file to that provider as well. They can then merge the two `ProviderIndependentID` objects into one and try again.

For simpler DX for the common case where a chat is only used with a single provider, we can still support passing only a string. Whenever that is present, it is assumed to be for the current provider. If the string cannot be found, an error is returned (the same way that would happen today if e.g. you passed a private Google uploaded URL to the Anthropic API).

Related: #12855 (which should apply a similar concept for managing uploaded skills)

felixarntz:
## Background

The AI SDK supports passing media files inline or via URL, but has no way to upload files directly to a provider or reference previously uploaded files across providers. Some providers return internal file IDs (not URLs) from their upload APIs, and switching providers mid-conversation requires a way to map the same logical file to different provider-specific identifiers.

## Summary

Introduces `uploadFile` as a top-level function and `ProviderReference` (`Record<string, string>`) as the provider-independent way to reference uploaded files.

- **New spec types**: `SharedV4ProviderReference` (provider package), `FilesV4` interface with `uploadFile` method, `UploadFileResult`; `mediaType` and `filename`** are top-level parameters as they're widely supported and used
- **New top-level API**: `uploadFile({ files, data, mediaType?, filename?, providerOptions? })` in the `ai` package, with auto-detection of media type from file bytes when not provided
- **Provider implementations**: `files()` interface on Anthropic, Google, OpenAI, and xAI providers, each implementing `FilesV4.uploadFile`
    - Other providers don't support uploading files, or they only support uploading files for batch inference (`*.jsonl`), which we don't support anyway.
- **Provider reference support in messages**: `LanguageModelV4FilePart.data` now accepts `SharedV4ProviderReference` in addition to `DataContent`; providers that support file references (Anthropic, Google, OpenAI, xAI) resolve them via `resolveProviderReference`; all other providers throw `UnsupportedFunctionalityError`
- **Spec cleanup**: `file-id` and `image-file-id` tool result output types replaced with `file-reference` and `image-file-reference` using `SharedV4ProviderReference` instead of `string | Record<string, string>`
- **Uploading from URL is not supported** — no provider supports this, and auto-downloading is questionable; callers should fetch first
- **`reasoning-file` was not touched** — it is model-generated as part of reasoning output, so provider references are not applicable
- **Docs included** — Docs about `uploadFile` and `ProviderReference`, and a new architecture guide are included

### Design decisions

- `ProviderReference` is a plain `Record<string, string>` rather than a wrapper class, keeping it simple to create and merge
- The `isLikelyText` heuristic for media type detection and the `documentMediaTypeSignatures` are kept internal (not exported) — they work well enough for `uploadFile` but are not general-purpose utilities
- `resolveProviderReference` (provider-utils) does the lookup by provider name and throws with a clear error listing available providers when the reference doesn't contain an entry for the current provider

### Open questions

1. Should we include a `type` property in `ProviderReference` to distinguish different kinds of provider references (e.g. file vs skill, see #12855)?
2. Should `mediaType` and `filename` be top-level fields in the `uploadFile` result object?
    - They're currently top-level request parameters, but in the response they're in `providerMetadata`.
3. `mergeProviderReferences` is currently inlined in an example (`multi-provider.ts`) — should we offer this as a utility, or leave it for later?
4. `file-id` and `image-file-id` were removed in `toModelOutput` return value — should we deprecate them instead and/or offer auto-migration via codemod?
5. Out of scope: supporting providers that allow uploading files solely for batch inference (e.g. Cohere, Groq, Mistral), which we don't support at a provider level yet anyway - probably leave for later?

## Manual Verification

Upload file examples were added for all 4 supported providers (Anthropic, Google, OpenAI, xAI), each with image, PDF, and text variants.

## Checklist

## Future Work

Reuse the new `ProviderReference` approach for #12855.

## Related Issues

Fixes #12995


packages/provider/src/language-model/v4/language-model-v4-prompt.ts
```diff
@@ -435,19 +433,15 @@ export type LanguageModelV4ToolResultOutput =
...
-             * Image that is referenced using a provider file id.
-             *
-             * If you use multiple providers, you need to
-             * specify the provider specific ids using
-             * the Record option. The key is the provider
-             * name, e.g. 'openai' or 'anthropic'.
+             * Provider-specific references for the image file.
+             * The key is the provider name, e.g. 'openai' or 'anthropic'.
              */
-            fileId: string | Record<string, string>;
```
lgrammel, line 450: removing the `string` option is a breaking change that might cause some trouble for our users. i know we have discussed it before and concluded we do not want the `string` id, but i am wondering what the migration strategy is? (seems hard to automate this migration)

packages/provider/src/provider/v4/provider-v4.ts
```diff
@@ -76,4 +77,6 @@ export interface ProviderV4 {
    * @throws {NoSuchModelError} If no such model exists.
    */
   rerankingModel?(modelId: string): RerankingModelV4;
+
+  files?(): FilesV4;
```
lgrammel, line 87: needs jsdoc

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
lgrammel: (if it is exposed through the ai pkg)

packages/provider-utils/src/resolve-provider-reference.ts
```diff
@@ -0,0 +1,19 @@
...
+}): string {
+  const id = reference[provider];
+  if (id != null) {
+    return id;
+  }
+
+  const availableProviders = Object.keys(reference).join(', ');
+  throw new Error(
+    `No reference found for provider '${provider}'. Available providers: ${availableProviders}`,
+  );
```
lgrammel: might be good to have a dedicated ai sdk error type (see our other errors) with an error documentation page

packages/provider-utils/src/types/content-part.ts
```diff
@@ -393,19 +390,15 @@ export type ToolResultOutput =
...
-             *
-             * If you use multiple providers, you need to
-             * specify the provider specific ids using
-             * the Record option. The key is the provider
-             * name, e.g. 'openai' or 'anthropic'.
+             * Provider-specific references for the image file.
+             * The key is the provider name, e.g. 'openai' or 'anthropic'.
              */
-            fileId: string | Record<string, string>;
+            providerReference: ProviderReference;
```
lgrammel, line 445: breaking changes, see notes below on the same concepts in the language spec

packages/ai/src/upload-file/upload-file.ts
```diff
@@ -0,0 +1,127 @@
...
+    );
+  }
+
+  const mediaType =
+    mediaTypeArg ??
+    detectMediaType({
+      data,
+      signatures: [...imageMediaTypeSignatures, ...documentMediaTypeSignatures],
+    }) ??
+    (isLikelyText(data) ? 'text/plain' : undefined);
```
lgrammel: what is the benefit of the `isLikelyText` (or media type detection) in general? can we leave this out (and rely on the providers) or do the provider APIs need this? can it be pushed into the providers that need it?

packages/ai/src/upload-file/upload-file.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.providerReference = options.providerReference;
+    this.providerMetadata = options.providerMetadata;
+    this.warnings = options.warnings;
+  }
+}
+
+function isLikelyText(data: Uint8Array | string): boolean {
+  const bytes =
+    typeof data === 'string'
+      ? convertBase64ToUint8Array(data.substring(0, Math.min(data.length, 688)))
```
lgrammel: why 688 - seems arbitrary

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -490,15 +504,14 @@ function convertPartToLanguageModelPart(
...
-        mediaType: mediaType ?? 'image/*', // any image
+        mediaType: mediaType ?? 'image/*',
         filename: undefined,
         data,
         providerOptions: part.providerOptions,
       };
     }
 
     case 'file': {
-      // We must have a mediaType for files, if not, throw an error.
```
lgrammel, line 501: lots of comments were removed, on purpose?

lgrammel:
Has this been tested with a UI example? (want to make sure we do not run into issues with UI messages and UI roundtrips/multi-turn interactions)

lgrammel:
> Should we include a type property in ProviderReference to distinguish different kinds of provider references (e.g. file vs skill, see https://github.com/vercel/ai/issues/12855)?

Unsure. Providers may allow e.g. files in skills (is that possible?). We could just rely on provider side checks for now. 

> Should mediaType and filename be top-level fields in the uploadFile result object?
> They're currently top-level request parameters, but in the response they're in providerMetadata.

Are they required by most APIs? if so then I would say yes, otherwise we can work with provider options.

> mergeProviderReferences is currently inlined in an example (multi-provider.ts) — should we offer this as a utility, or leave it for later?

A utility might be nice - but can be a separate PR later. We could also wait if there is demand (very likely tho).

> file-id and image-file-id were removed in toModelOutput return value — should we deprecate them instead and/or offer auto-migration via codemod?

Deprecation is preferred unless it would not work / be too difficult.

> Out of scope: supporting providers that allow uploading files solely for batch inference (e.g. Cohere, Groq, Mistral), which we don't support at a provider level yet anyway - probably leave for later?

sg

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
felixarntz: I'd love to better understand how the data migrations generally would work for model messages, I think I lack the context for this. Like at which point does migration happen? Do we expose a function that migrates a list of model messages to the new spec shape? Or something else?

That'll give me better context to think about this further.

packages/ai/src/upload-file/upload-file.ts
felixarntz: Yeah the provider APIs need this. we could implement it at the provider level, but I think this would cause a lot of duplication for something we can centrally handle.

The logic we have here is basically a best guess, with the objective that it works for ~90% of cases. Developers can still pass `mediaType` to `uploadFile` themselves, but since that's somewhat a low-level piece of info, I think it makes sense that we don't require it and instead do our best to detect it.

For `isLikelyText` specifically, I don't love it myself either, but I think as long as we keep it private in this file, it's a decent approach. We can't use the same approach we do for binary file MIME type detection for plain text, so that's what this function is for.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
felixarntz: AI slop 🙃 fixed

felixarntz:
> Has this been tested with a UI example? (want to make sure we do not run into issues with UI messages and UI roundtrips/multi-turn interactions)

added in https://github.com/vercel/ai/pull/13816/changes/cc83962c7bb76c2a17f2598c0eac712018eb275d. I built it in a way that very obviously allows testing the behavior across multiple providers, not in a way that's necessarily good UX, but allows easy testing:
1. upload an image, it'll upload it to whichever provider you have selected
2. send a prompt with that image to that same provider it'll work
3. once you select another prompt and send to another provider, it'll error
4. you can then upload the same image to the other now selected provider in the sidebar
5. then it'll work

> > Should we include a type property in ProviderReference to distinguish different kinds of provider references (e.g. file vs skill, see #12855)?
> 
> Unsure. Providers may allow e.g. files in skills (is that possible?). We could just rely on provider side checks for now.

it shouldn't be an issue when providers have files in skills, technically it's all packaged up and treated as "a skill" at the provider level where it was uploaded too. so that one would always be `type: 'skill'`.

I'm not sure this is necessary, but I have a hunch for DX it may be valuable to distinguish between whether a `ProviderReference` instance is for a file or for a skill or for something else. Just because the ID itself likely doesn't convey that at all.

> > Should mediaType and filename be top-level fields in the uploadFile result object?
> > They're currently top-level request parameters, but in the response they're in providerMetadata.
> 
> Are they required by most APIs? if so then I would say yes, otherwise we can work with provider options.

Yes, they are. I'm already using them as top-level request params. My question was whether we should also have them as top-level fields in the response, or as `providerMetadata`? Just since for in the response they're technically not important to process on the client-side, because it'll be just the same value as you passed in the request (unless some edge-case maybe?). But I was thinking since we have them as top-level request params, we'd probably also want them as top-level response properties?

> > mergeProviderReferences is currently inlined in an example (multi-provider.ts) — should we offer this as a utility, or leave it for later?
> 
> A utility might be nice - but can be a separate PR later. We could also wait if there is demand (very likely tho).

:thumbsup:

> > file-id and image-file-id were removed in toModelOutput return value — should we deprecate them instead and/or offer auto-migration via codemod?
> 
> Deprecation is preferred unless it would not work / be too difficult.

ok, will re-add and mark deprecated instead.

packages/ai/src/upload-file/upload-file.ts
felixarntz: added a comment to explain, it's because of base64 encoding and how it works, in combination with the 512 value below

packages/provider/src/provider/v4/provider-v4.ts
felixarntz: done

packages/provider-utils/src/resolve-provider-reference.ts
felixarntz: done

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -514,35 +544,80 @@ function convertPartToLanguageModelPart(
...
+  }
+
+  if (provider == null) {
+    throw new Error(
+      'Cannot convert string fileId to provider reference without a provider ID. ' +
+        'Use a Record<string, string> fileId or switch to the file-reference type.',
+    );
+  }
+
+  return { [provider]: fileId };
```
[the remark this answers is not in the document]
felixarntz: fixed more comprehensively in https://github.com/vercel/ai/pull/13816/changes/dd44d158e0150cefc7bc78e97acc3210e930eb25

architecture/file-uploads.md
```diff
@@ -0,0 +1,155 @@
...
+
+```ts
+import { resolveProviderReference } from '@ai-sdk/provider-utils';
+
+// Inside the file part handling:
+case 'file': {
+  if (
+    typeof part.data === 'object' &&
+    !(part.data instanceof Uint8Array) &&
+    !(part.data instanceof URL)
```
lgrammel: I've seen this being repeated in various places, might make sense to have a function/make it reusable (there may be something already)

content/docs/03-ai-sdk-core/39-file-uploads.mdx
```diff
@@ -0,0 +1,140 @@
...
+function to upload files to a provider and get back a `ProviderReference` that can be
+used in subsequent API calls.
+
+```ts
+import { uploadFile, generateText } from 'ai';
+import { openai } from '@ai-sdk/openai';
+import fs from 'node:fs';
+
+const { providerReference } = await uploadFile({
+  files: openai.files(),
```
lgrammel: `model`

lgrammel: (model is not a good name here but would be consistent)

lgrammel: maybe `api` if model is not great

content/docs/03-ai-sdk-core/39-file-uploads.mdx
```diff
@@ -0,0 +1,140 @@
...
+  data: fs.readFileSync('./photo.png'),
+  filename: 'photo.png',
```
lgrammel, line 20: filename or fileName -- prefer the former but might be good to check for consistency with other parts of our api

content/docs/07-reference/01-ai-sdk-core/30-model-message.mdx
```diff
@@ -342,17 +342,13 @@ IANA media type.
...
-             *
-             * If you use multiple providers, you need to
-             * specify the provider specific ids using
-             * the Record option. The key is the provider
-             * name, e.g. 'openai' or 'anthropic'.
+             * Provider-specific references for the file.
+             * The key is the provider name, e.g. 'openai' or 'anthropic'.
              */
-            fileId: string | Record<string, string>;
+            providerReference: ProviderReference;
```
lgrammel, line 374: this will require data migrations - would be nice to support users / have this as part of our migration doc

examples/ai-e2e-next/app/api/chat/xai-image-edit/route.ts
```diff
@@ -76,7 +76,11 @@ function findLastAssistantImage(
 
     for (let j = content.length - 1; j >= 0; j--) {
       const part = content[j];
-      if (part.type === 'file' && !(part.data instanceof URL)) {
+      if (
+        part.type === 'file' &&
+        !(part.data instanceof URL) &&
+        !(typeof part.data === 'object' && !(part.data instanceof Uint8Array))
```
lgrammel, line 82: wondering if we can extract this expression or parts of it into a shared helper

examples/ai-e2e-next/app/api/upload-file/route.ts
```diff
@@ -0,0 +1,39 @@
...
+    filename: string;
+    provider: 'anthropic' | 'google' | 'openai' | 'xai';
+  } = await req.json();
+
+  const filesMap = {
+    anthropic: () => anthropic.files(),
+    google: () => google.files(),
+    openai: () => openai.files(),
+    xai: () => xai.files(),
+  };
```
lgrammel, line 25: can we add files to custom provider / provider registry (future work)

examples/ai-e2e-next/app/chat/upload-file/page.tsx
```diff
@@ -0,0 +1,362 @@
...
+        const reader = new FileReader();
+        reader.onload = () => resolve(reader.result as string);
+        reader.onerror = reject;
+        reader.readAsDataURL(file);
+      });
+
+      const base64 = dataUrl.split(',')[1];
+      const { providerReference } = await uploadToProvider({
+        base64,
+        mediaType: file.type || 'application/octet-stream',
```
lgrammel: why are we using `application/octet-stream` as a default here?

examples/ai-e2e-next/app/chat/upload-file/page.tsx
```diff
@@ -0,0 +1,362 @@
...
+      });
+
+      const id = crypto.randomUUID();
+
+      setUploadedFiles(prev => [
+        ...prev,
+        {
+          id,
+          filename: file.name,
+          mediaType: file.type || 'application/octet-stream',
```
lgrammel: same

examples/ai-e2e-next/app/chat/upload-file/page.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+      const base64 = dataUrl.split(',')[1];
+      const { providerReference } = await uploadToProvider({
+        base64,
+        mediaType: file.type || 'application/octet-stream',
+        filename: file.name,
+        provider: selectedModel.provider,
+      });
+
+      const id = crypto.randomUUID();
```
lgrammel: prefer our own id generation

packages/ai/src/generate-object/generate-object.ts
```diff
@@ -301,6 +301,7 @@ export async function generateObject<
           prompt: standardizedPrompt,
           supportedUrls: await model.supportedUrls,
           download,
+          provider: model.provider.split('.')[0],
```
lgrammel: why has this changed?

packages/ai/src/generate-object/stream-object.ts
```diff
@@ -499,6 +499,7 @@ class DefaultStreamObjectResult<
             prompt: standardizedPrompt,
             supportedUrls: await model.supportedUrls,
             download,
+            provider: model.provider.split('.')[0],
```
lgrammel: why has this changed?

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -694,6 +694,7 @@ export async function generateText<
           },
           supportedUrls: await stepModel.supportedUrls,
           download,
+          provider: stepModel.provider.split('.')[0],
```
lgrammel, line 713: same

packages/provider/src/files/v4/files-v4.ts
```diff
@@ -0,0 +1,62 @@
...
+   */
+  readonly provider: string;
+
+  /**
+   * Uploads a file to the provider and returns a provider reference
+   * that can be used in subsequent API calls.
+   */
+  uploadFile(options: {
+    data: Uint8Array | string;
+    mediaType?: string;
```
lgrammel: if the media type is optional here we should not default in the ai lib

if we default in the ai lib the media type should be required

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
lgrammel: the data is stored by the ai sdk users, so they determine when/how data migrations happen

packages/prodia/src/prodia-language-model.ts
```diff
@@ -126,6 +127,16 @@ export class ProdiaLanguageModel implements LanguageModelV4 {
       if (message.role === 'user') {
         for (const part of message.content) {
           if (part.type === 'file' && part.mediaType.startsWith('image/')) {
+            if (
+              typeof part.data === 'object' &&
+              !(part.data instanceof Uint8Array) &&
+              !(part.data instanceof URL)
```
lgrammel: extract?

packages/openai/src/openai-config.ts
```diff
@@ -6,13 +6,4 @@ export type OpenAIConfig = {
...
   generateId?: () => string;
-  /**
-   * File ID prefixes used to identify file IDs in Responses API.
-   * When undefined, all file data is treated as base64 content.
-   *
-   * Examples:
-   * - OpenAI: ['file-'] for IDs like 'file-abc123'
-   * - Azure OpenAI: ['assistant-'] for IDs like 'assistant-abc123'
-   */
-  fileIdPrefixes?: readonly string[];
```
lgrammel, line 17: we might need to provide help with migrating data to the new format

packages/openai/src/files/openai-files.ts
```diff
@@ -0,0 +1,100 @@
...
+        provider: 'openai',
+        providerOptions,
+        schema: openaiFilesOptionsSchema,
+      })) as OpenAIFilesOptions | undefined;
+
+      const fileBytes =
+        data instanceof Uint8Array ? data : convertBase64ToUint8Array(data);
+
+      const blob = new Blob([fileBytes], {
+        type: mediaType ?? 'application/octet-stream',
```
lgrammel: why the octet-stream default?

packages/openai/src/responses/convert-to-openai-responses-input.ts
```diff
@@ -123,12 +139,9 @@ export async function convertToOpenAIResponsesInput({
                     type: 'input_image',
                     ...(part.data instanceof URL
                       ? { image_url: part.data.toString() }
-                      : typeof part.data === 'string' &&
```
lgrammel, line 126: will this change break existing stored messages

packages/ai/src/upload-file/upload-file.ts
lgrammel: might be worth extract both values 512 and 688 into local constants with descriptive names, the relationship was not obvious to me

packages/ai/src/upload-file/upload-file.ts
lgrammel: wondering if we can somehow integrate it into detectMediaType

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -91,7 +94,7 @@ export async function convertToLanguageModelPrompt({
           }))
       : []),
     ...prompt.messages.map(message =>
-      convertToLanguageModelMessage({ message, downloadedAssets }),
+      convertToLanguageModelMessage({ message, downloadedAssets, provider }),
```
lgrammel, line 99: i do not like that we now need provider in here. it indicates that some of this should be pushed into the provider api instead. might be worth talking about so i fully understand why this is needed

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -534,35 +544,80 @@ function convertPartToLanguageModelPart(
...
+          if (item.mediaType.startsWith('image/')) {
+            return {
+              type: 'image-data' as const,
+              data: item.data,
+              mediaType: item.mediaType,
+            };
+          }
 
-      // AI SDK 5 tool backwards compatibility:
-      // map media type to image-data or file-data
```
felixarntz, line 532: Reinstate

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -38,10 +38,12 @@ export async function convertToLanguageModelPrompt({
   prompt,
   supportedUrls,
   download = createDefaultDownloadFunction(),
+  provider,
```
felixarntz, line 43: add comment not to use it; todo to remove in V8

content/docs/03-ai-sdk-core/39-file-uploads.mdx
felixarntz: confirmed `filename`, we don't use `fileName` in any API.

examples/ai-e2e-next/app/api/upload-file/route.ts
felixarntz: will open a follow up PR for this, also to audit for any other potentially missing pieces on custom provider / provider registry

packages/ai/src/upload-file/upload-file.ts
felixarntz: thinking about this further, I'm unsure. Other code also resolves "default" media types as part of `ai`, e.g. `generateImage` or `generateSpeech`. As part of model messages, the `mediaType` field is required, so that's likely why. There's also the point that `detectMediaType` is currently part of `ai`. If we want to make detection of the default media type responsibility of each provider, I'd still want them to have a utility like that available, so we'd need to move it, including all its associated types to `provider-utils` (and re-export from `ai` for backward compatibility). And then, to have a more intuitive helper, we'd probably want to set some default `signatures`, so we may want to have that in a more generic function, e.g. `detectMimeType` (not just "media").

We could make all these changes, I'm not necessarily opposed. But it's more changes than I'd like to make for this ideally.

I still see your point about potentially the provider handling this because it would allow for a provider implementation to leave it up to the API. But maybe this is a moot point, given that so far none of them seem to automatically handle MIME types?

How about we make `mediaType` required at the provider spec V4 for now? That would alleviate your point from https://github.com/vercel/ai/pull/13816#discussion_r3010253476, just in the opposite way from how we discussed earlier.

felixarntz: my only other follow up question that this got me to think about is: Should we maybe rename it to `mimeType`? We're using `mediaType` in other places, but that's where it's truly about media. But the `uploadFile` APIs can be used for files that aren't really media, and "MIME type" is a more generally applicable term. WDYT?

packages/provider/src/files/v4/files-v4.ts
felixarntz: see https://github.com/vercel/ai/pull/13816#discussion_r3012192644

packages/openai/src/openai-config.ts
felixarntz: will make a point for this in the migration guide

packages/openai/src/files/openai-files.ts
felixarntz: just a fallback, but agree it's not great. See https://github.com/vercel/ai/pull/13816#discussion_r3012192644, if we make `mediaType` mandatory, this won't be necessary any longer.

content/docs/07-reference/01-ai-sdk-core/30-model-message.mdx
felixarntz: this was incorrect to have been removed here; fixed in https://github.com/vercel/ai/pull/13816/commits/7e3a4b0c54693383487b8aeb40bb2b3ee7c35078

content/docs/03-ai-sdk-core/39-file-uploads.mdx
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/5e4ae2a28555f9d780c96b6506ba9c3368c52d32

examples/ai-e2e-next/app/chat/upload-file/page.tsx
felixarntz: fixed in https://github.com/vercel/ai/pull/13816/changes/9fca8d164ea692d4c623e692309be750d37af9f0

packages/ai/src/prompt/convert-to-language-model-prompt.ts
felixarntz: added comments to clarify temporaty nature: https://github.com/vercel/ai/pull/13816/changes/e7b20be496c4c755fa5f04d18ff565710201b332

packages/openai/src/responses/convert-to-openai-responses-input.ts
felixarntz: great catch; restored for back compat: https://github.com/vercel/ai/pull/13816/commits/6bb04118357ba59eaf808b4f6d018d7dcef58876

will mention this in migration guide

examples/ai-e2e-next/app/chat/upload-file/page.tsx
felixarntz: fixed in https://github.com/vercel/ai/pull/13816/commits/64f7c2d0c7baa729d10e5e9848bb1f7881e91f17, also for the other similar feedback

packages/ai/src/upload-file/upload-file.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/changes/182354d9a915c78731878c9a2c52241c312edb0a

architecture/file-uploads.md
felixarntz: done in https://github.com/vercel/ai/pull/13816/changes/91dd20f8a1c0c96ba7cd0e9bcada82414d5dc895

examples/ai-e2e-next/app/api/chat/xai-image-edit/route.ts
felixarntz: see https://github.com/vercel/ai/pull/13816/changes/91dd20f8a1c0c96ba7cd0e9bcada82414d5dc895

packages/provider/src/errors/no-such-provider-reference-error.ts
```diff
@@ -0,0 +1,30 @@
+import { AISDKError } from './ai-sdk-error';
+
+const name = 'AI_NoSuchProviderReferenceError';
+const marker = `vercel.ai.error.${name}`;
+const symbol = Symbol.for(marker);
+
+export class NoSuchProviderReferenceError extends AISDKError {
```
lgrammel, line 12: needs an error documentation page under `/Users/larsgrammel/repositories/ai/content/docs/07-reference/05-ai-sdk-errors` (and update to index.md)

wondering if a short jsdoc on the error class would help as well

packages/ai/src/upload-file/upload-file.ts
lgrammel: mimeType and mediaType are the same afaik (mime type is the original name, media type is the current standard name), and web standards have moved to mediaType, which is why i standardized everything on mediaType. Re-introducing mimeType would make it inconsistent (and confusing imo if we have both) - if we have mimeType anywhere we should rename it to mediaType (unless it is provider specific).

https://en.wikipedia.org/wiki/Media_type

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
```diff
@@ -378,17 +380,13 @@ export type LanguageModelV4ToolResultOutput =
             providerOptions?: SharedV4ProviderOptions;
           }
         | {
-            type: 'file-id';
+            type: 'file-reference';
```
lgrammel, line 383: above we used ` data: LanguageModelV4DataContent | SharedV4ProviderReference;`, here we have separate file-reference/file-data types.

this is somewhat inconsistent. the separation here makes sense to me (bc file reference do not need filename etc) - wondering if it should be changed above

packages/provider/src/files/v4/files-v4.ts
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Provider ID.
+   */
+  readonly provider: string;
+
+  /**
+   * Uploads a file to the provider and returns a provider reference
+   * that can be used in subsequent API calls.
+   */
+  uploadFile(options: {
```
lgrammel: add jsdoc for the params

packages/provider-utils/src/resolve-provider-reference.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  NoSuchProviderReferenceError,
+  SharedV4ProviderReference,
+} from '@ai-sdk/provider';
+
+export function resolveProviderReference({
```
lgrammel, line 11: add jsdoc?

packages/provider-utils/src/resolve-provider-reference.ts
```diff
@@ -0,0 +1,22 @@
...
+  provider: string;
+}): string {
+  const id = reference[provider];
+  if (id != null) {
+    return id;
+  }
+
+  throw new NoSuchProviderReferenceError({
+    provider,
+    availableProviders: Object.keys(reference),
```
lgrammel: alternatively the whole reference could be passed into the error (so that the full ids are visible when users want to debug)

packages/provider-utils/src/is-provider-reference.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  LanguageModelV4FilePart,
+  SharedV4ProviderReference,
+} from '@ai-sdk/provider';
+
+export function isProviderReference(
```
lgrammel, line 10: jsdoc

packages/provider-utils/src/types/content-part.ts
```diff
@@ -62,7 +63,7 @@ export interface FilePart {
    * - data: a base64-encoded string, a Uint8Array, an ArrayBuffer, or a Buffer
    * - URL: a URL that points to the image
    */
-  data: DataContent | URL;
+  data: DataContent | URL | ProviderReference;
```
lgrammel, line 68: somewhat inconsistent to the explicit id/reference objects below, wondering if we can either split this or unify the other parts. maybe fine for now but could be worth thinking about.

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
```diff
@@ -150,9 +151,10 @@ export interface LanguageModelV4FilePart {
   filename?: string;
 
   /**
-   * File data. Can be a Uint8Array, base64 encoded data as a string or a URL.
+   * File data. Can be a Uint8Array, base64 encoded data as a string, a URL,
+   * or a provider reference mapping provider names to provider-specific file IDs.
    */
-  data: LanguageModelV4DataContent;
+  data: LanguageModelV4DataContent | SharedV4ProviderReference;
```
lgrammel, line 157: does this make sense here or should it be a separate type

packages/provider/src/provider/v4/provider-v4.ts
```diff
@@ -76,4 +77,12 @@ export interface ProviderV4 {
...
    */
   rerankingModel?(modelId: string): RerankingModelV4;
+
+  /**
+   * Returns the files interface for uploading files to the provider.
+   * The returned interface can be passed to the `uploadFile` function.
+   *
+   * @returns {FilesV4} The files interface for this provider.
+   */
+  files?(): FilesV4;
```
lgrammel, line 87: should be mandatory imo (we need to move optional ones to mandatory)

packages/xai/src/responses/convert-to-xai-responses-input.ts
```diff
@@ -42,7 +45,19 @@ export async function convertToXaiResponsesInput({
             }
 
             case 'file': {
-              if (block.mediaType.startsWith('image/')) {
+              if (
+                typeof block.data === 'object' &&
+                !(block.data instanceof Uint8Array) &&
+                !(block.data instanceof URL)
```
lgrammel: use helper

content/docs/07-reference/01-ai-sdk-core/30-model-message.mdx
```diff
@@ -342,6 +342,9 @@ IANA media type.
             providerOptions?: ProviderOptions;
           }
         | {
```
lgrammel, line 346: there was another change here i believe (file parts)

examples/ai-e2e-next/app/api/chat/upload-file/route.ts
```diff
@@ -0,0 +1,43 @@
...
+} from 'ai';
+
+export const maxDuration = 60;
+
+const providerMap = {
+  anthropic: (modelId: string) => anthropic(modelId),
+  google: (modelId: string) => google(modelId),
+  openai: (modelId: string) => openai.responses(modelId),
+  xai: (modelId: string) => xai(modelId),
+};
```
lgrammel, line 19: use provider registry? (or todo for future work)

packages/ai/src/generate-object/generate-object.ts
```diff
@@ -334,6 +334,7 @@ export async function generateObject<
       prompt: standardizedPrompt,
       supportedUrls: await model.supportedUrls,
       download,
+      provider: model.provider.split('.')[0],
```
lgrammel, line 337: maybe consider extracting a function so we can search for it when we removed this in the future

packages/ai/src/ui/convert-to-model-messages.ts
```diff
@@ -176,7 +176,7 @@ export async function convertToModelMessages<UI_MESSAGE extends UIMessage>(
                   type: 'file' as const,
                   mediaType: part.mediaType,
                   filename: part.filename,
-                  data: part.url,
+                  data: part.providerReference ?? part.url,
```
lgrammel, line 179: i wonder if mixing this into data is the right way but maybe it is

packages/ai/src/upload-file/upload-file.ts
```diff
@@ -0,0 +1,144 @@
...
+  }) {
+    this.providerReference = options.providerReference;
+    this.mediaType = options.mediaType;
+    this.filename = options.filename;
+    this.providerMetadata = options.providerMetadata;
+    this.warnings = options.warnings;
+  }
+}
+
+function isLikelyText(data: Uint8Array | string): boolean {
```
lgrammel, line 123: potentially worth extracting, moving into provider utils, and testing separately

packages/azure/src/azure-openai-provider.ts
```diff
@@ -213,7 +213,6 @@ export function createAzure(
       url,
       headers: getHeaders,
       fetch: options.fetch,
-      fileIdPrefixes: ['assistant-'],
```
lgrammel, line 216: are we losing this functionality for azure? (might be worth checking out the community pr that introduced the behavior)

lgrammel (approved):
might be good to update the migration guide as part of this PR so we do not forget about the breaking changes / required migrations

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
felixarntz: opened #14010

packages/provider-utils/src/types/content-part.ts
felixarntz: see #14010

packages/provider/src/language-model/v4/language-model-v4-prompt.ts
felixarntz: see #14010

packages/provider/src/errors/no-such-provider-reference-error.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/d28fa5c35d20e49af49139fbf3ec47b7ca1143ab

packages/provider/src/files/v4/files-v4.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/790b0ae4fca615394fa4706f713179f46e5a3890

packages/provider-utils/src/resolve-provider-reference.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/790b0ae4fca615394fa4706f713179f46e5a3890

packages/provider-utils/src/is-provider-reference.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/790b0ae4fca615394fa4706f713179f46e5a3890

packages/provider-utils/src/resolve-provider-reference.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/66accbf520bf63360fe415e36735da87a0898f07

packages/xai/src/responses/convert-to-xai-responses-input.ts
felixarntz: done in https://github.com/vercel/ai/pull/13816/commits/8e4cd43d5abb577d6854686beb648870ce43ade1

content/docs/07-reference/01-ai-sdk-core/30-model-message.mdx
felixarntz: docs fixed in https://github.com/vercel/ai/pull/13816/commits/4d554097e8af7fab2c18a45d7fe3f8a6ec5be076

examples/ai-e2e-next/app/api/chat/upload-file/route.ts
felixarntz: will do this in a separate PR next week

packages/ai/src/ui/convert-to-model-messages.ts
felixarntz: We'll improve this via #14010

packages/ai/src/upload-file/upload-file.ts
felixarntz: I feel better not doing it for now, this function seems too niche/brittle/specific to expose it externally

packages/azure/src/azure-openai-provider.ts
felixarntz: good catch, restored and soft deprecated in https://github.com/vercel/ai/pull/13816/changes/9d184167a9dd0f2f78cc3af75474eeee613c9d7b

packages/ai/src/generate-object/generate-object.ts
felixarntz: since it's temporary, I think it's okay not to introduce an external helper for this.

looking for `provider.split('.')[0]` throughout the codebase will show all relevant usage

lgrammel (approved):

packages/ai/src/upload-file/upload-file.ts
felixarntz: made `mediaType` required on the provider spec level: https://github.com/vercel/ai/pull/13816/commits/6dc0a47c07f0342171652a6de68032a77517d639

packages/openai/src/files/openai-files.ts
felixarntz: fixed in all providers, now that `mediaType` is required at provider level

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `ai` | [`7.0.0-beta.61`](https://www.npmjs.com/package/ai/v/7.0.0-beta.61) |
| `@ai-sdk/alibaba` | [`2.0.0-beta.16`](https://www.npmjs.com/package/@ai-sdk/alibaba/v/2.0.0-beta.16) |
| `@ai-sdk/amazon-bedrock` | [`5.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/amazon-bedrock/v/5.0.0-beta.21) |
| `@ai-sdk/angular` | [`3.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/angular/v/3.0.0-beta.61) |
| `@ai-sdk/anthropic` | [`4.0.0-beta.17`](https://www.npmjs.com/package/@ai-sdk/anthropic/v/4.0.0-beta.17) |
| `@ai-sdk/assemblyai` | [`3.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/assemblyai/v/3.0.0-beta.12) |
| `@ai-sdk/azure` | [`4.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/azure/v/4.0.0-beta.21) |
| `@ai-sdk/baseten` | [`2.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/baseten/v/2.0.0-beta.14) |
| `@ai-sdk/black-forest-labs` | [`2.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/black-forest-labs/v/2.0.0-beta.11) |
| `@ai-sdk/bytedance` | [`2.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/bytedance/v/2.0.0-beta.11) |
| `@ai-sdk/cerebras` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/cerebras/v/3.0.0-beta.14) |
| `@ai-sdk/cohere` | [`4.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/cohere/v/4.0.0-beta.12) |
| `@ai-sdk/deepgram` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/deepgram/v/3.0.0-beta.11) |
| `@ai-sdk/deepinfra` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/deepinfra/v/3.0.0-beta.14) |
| `@ai-sdk/deepseek` | [`3.0.0-beta.13`](https://www.npmjs.com/package/@ai-sdk/deepseek/v/3.0.0-beta.13) |
| `@ai-sdk/devtools` | [`1.0.0-beta.6`](https://www.npmjs.com/package/@ai-sdk/devtools/v/1.0.0-beta.6) |
| `@ai-sdk/elevenlabs` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/elevenlabs/v/3.0.0-beta.11) |
| `@ai-sdk/fal` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/fal/v/3.0.0-beta.11) |
| `@ai-sdk/fireworks` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/fireworks/v/3.0.0-beta.14) |
| `@ai-sdk/gateway` | [`4.0.0-beta.31`](https://www.npmjs.com/package/@ai-sdk/gateway/v/4.0.0-beta.31) |
| `@ai-sdk/gladia` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/gladia/v/3.0.0-beta.11) |
| `@ai-sdk/google` | [`4.0.0-beta.23`](https://www.npmjs.com/package/@ai-sdk/google/v/4.0.0-beta.23) |
| `@ai-sdk/google-vertex` | [`5.0.0-beta.31`](https://www.npmjs.com/package/@ai-sdk/google-vertex/v/5.0.0-beta.31) |
| `@ai-sdk/groq` | [`4.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/groq/v/4.0.0-beta.14) |
| `@ai-sdk/huggingface` | [`2.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/huggingface/v/2.0.0-beta.14) |
| `@ai-sdk/hume` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/hume/v/3.0.0-beta.11) |
| `@ai-sdk/klingai` | [`4.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/klingai/v/4.0.0-beta.12) |
| `@ai-sdk/langchain` | [`3.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/langchain/v/3.0.0-beta.61) |
| `@ai-sdk/llamaindex` | [`3.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/llamaindex/v/3.0.0-beta.61) |
| `@ai-sdk/lmnt` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/lmnt/v/3.0.0-beta.11) |
| `@ai-sdk/luma` | [`3.0.0-beta.11`](https://www.npmjs.com/package/@ai-sdk/luma/v/3.0.0-beta.11) |
| `@ai-sdk/mcp` | [`2.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/mcp/v/2.0.0-beta.14) |
| `@ai-sdk/mistral` | [`4.0.0-beta.13`](https://www.npmjs.com/package/@ai-sdk/mistral/v/4.0.0-beta.13) |
| `@ai-sdk/moonshotai` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/moonshotai/v/3.0.0-beta.14) |
| `@ai-sdk/open-responses` | [`2.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/open-responses/v/2.0.0-beta.12) |
| `@ai-sdk/openai` | [`4.0.0-beta.21`](https://www.npmjs.com/package/@ai-sdk/openai/v/4.0.0-beta.21) |
| `@ai-sdk/openai-compatible` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/openai-compatible/v/3.0.0-beta.14) |
| `@ai-sdk/otel` | [`1.0.0-beta.7`](https://www.npmjs.com/package/@ai-sdk/otel/v/1.0.0-beta.7) |
| `@ai-sdk/perplexity` | [`4.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/perplexity/v/4.0.0-beta.14) |
| `@ai-sdk/prodia` | [`2.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/prodia/v/2.0.0-beta.14) |
| `@ai-sdk/provider` | [`4.0.0-beta.6`](https://www.npmjs.com/package/@ai-sdk/provider/v/4.0.0-beta.6) |
| `@ai-sdk/provider-utils` | [`5.0.0-beta.10`](https://www.npmjs.com/package/@ai-sdk/provider-utils/v/5.0.0-beta.10) |
| `@ai-sdk/react` | [`4.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/react/v/4.0.0-beta.61) |
| `@ai-sdk/replicate` | [`3.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/replicate/v/3.0.0-beta.12) |
| `@ai-sdk/revai` | [`3.0.0-beta.12`](https://www.npmjs.com/package/@ai-sdk/revai/v/3.0.0-beta.12) |
| `@ai-sdk/rsc` | [`3.0.0-beta.62`](https://www.npmjs.com/package/@ai-sdk/rsc/v/3.0.0-beta.62) |
| `@ai-sdk/svelte` | [`5.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/svelte/v/5.0.0-beta.61) |
| `@ai-sdk/togetherai` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/togetherai/v/3.0.0-beta.14) |
| `@ai-sdk/valibot` | [`3.0.0-beta.10`](https://www.npmjs.com/package/@ai-sdk/valibot/v/3.0.0-beta.10) |
| `@ai-sdk/vercel` | [`3.0.0-beta.14`](https://www.npmjs.com/package/@ai-sdk/vercel/v/3.0.0-beta.14) |
| `@ai-sdk/vue` | [`4.0.0-beta.61`](https://www.npmjs.com/package/@ai-sdk/vue/v/4.0.0-beta.61) |
| `@ai-sdk/xai` | [`4.0.0-beta.23`](https://www.npmjs.com/package/@ai-sdk/xai/v/4.0.0-beta.23) |

PR #13818, opened 2026-03-24
fix(google): serialize Veo image inputs as bytesBase64Encoded

f0rr0:
## Summary
- serialize Veo image-to-video inputs as `bytesBase64Encoded` instead of `inlineData`
- serialize `referenceImages[].bytesBase64Encoded` the same way
- update the Google video model tests to assert the request body shape

## Why
Google Veo image-to-video requests against the Gemini `predictLongRunning` endpoint fail when the image is sent as `inlineData`. The current provider emits `inlineData` for both `options.image` and `referenceImages`, which causes first-frame video generation to fail in real usage.

## Validation
- `pnpm --filter @ai-sdk/google exec vitest --config vitest.node.config.js --run src/google-generative-ai-video-model.test.ts`
- validated the same serializer change locally against `@ai-sdk/google@3.0.53` through both `google.video(...).doGenerate(...)` and the full `experimental_generateVideo(...)` path

### DIFF
--- a/.changeset/minimax-provider.md
+++ b/.changeset/minimax-provider.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/minimax': major
+---
+
+Add MiniMax provider with language model support for the MiniMax-M model series.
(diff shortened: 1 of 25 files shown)


OskarLebuda:
@gr2m can you take a look at this PR? Because of this error we cannot generate a video

PR #13837, opened 2026-03-26
fix(langchain): emit reasoning-end before finish-step on LangGraph step transitions

hntrl:
## Summary

Fixes `AI_UIMessageStreamError: Received reasoning-delta for missing reasoning part` errors that occur when streaming GPT-5 reasoning through the LangChain/LangGraph adapter.

The root cause is that the LangGraph adapter emits `finish-step` during step transitions without first closing open reasoning/text parts. On the client, `finish-step` clears `activeReasoningParts`, so any subsequent `reasoning-end` (from the `values` handler) hits an already-cleared map and throws. In `streamMode: 'messages'` (no values events), reasoning parts are never closed at all before `finish-step`.

### `@ai-sdk/langchain`

- **`utils.ts`**: Before emitting `finish-step` on LangGraph step transitions, iterate `messageSeen` and emit `text-end`/`reasoning-end` for all open parts, then clear the tracking state. This ensures the client always sees a proper `*-start`/`*-end` lifecycle before `finish-step` resets its maps.
- **`adapter.ts`**: Apply the same cleanup in the stream finalization path, so streams without `values` events (e.g. `streamMode: 'messages'`) also properly close open parts before the final `finish-step`/`finish`.
- **`adapter.test.ts`**: Add tests for multi-step reasoning without values events, message ID switching mid-stream, and end-to-end validation that all three GPT-5 fixtures produce client-valid chunk sequences.


lgrammel (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/13914

PR #13840, opened 2026-03-26
fix(ai): avoid quadratic memory growth in streamText for plain text

### LINKED ISSUE #13839: streamText Memory Issue: parsePartialOutput creates O(n) string copies
## Summary

`streamText` with the default `text()` output calls `parsePartialOutput({ text: accumulatedText })` on every streaming chunk, then `JSON.stringify(result.partial)` to check if the output changed. For plain text streaming (no structured output), this creates a full copy of the accumulated text on every chunk.

For a 110KB response arriving in 13,000 chunks, this creates ~13,000 string copies totaling ~350MB per stream. With 3 concurrent streams, this produces 1–1.6GB of retained string objects in V8's `large_object_space` that are only reclaimable via full GC.

> **Impact is multiplicative with concurrency:** each additional stream running in the same process compounds the memory pressure. With 3 concurrent streams on a single container, total heap usage reached ~3.4GB — causing OOM kills on a 4GB container. Any production deployment handling concurrent long-output streams is at risk.

## Root Cause

In `ai/dist/index.mjs`, the `createOutputTransformStream` function (line ~6405):

```js
function createOutputTransformStream(output) {
  let text2 = "";
  let textChunk = "";
  // ...

  // Called on EVERY chunk:
  text2 += chunk.text;                                           // Line 6453
  textChunk += chunk.text;                                       // Line 6454
  const result = await output.parsePartialOutput({ text: text2 });  // Line 6456
  if (result !== void 0) {
    const currentJson = JSON.stringify(result.partial);           // Line 6458 — FLATTENS text2 into new string
    if (currentJson !== lastPublishedJson) {
      publishTextChunk({ controller, partialOutput: result.partial });
      lastPublishedJson = currentJson;                           // retains stringified copy
    }
  }
}
```

The default `text()` output (line ~3301):

```js
var text = () => ({
  async parsePartialOutput({ text: text2 }) {
    return { partial: text2 };  // Always returns non-undefined → triggers JSON.stringify
  },
});
```

**The multiplication effect:**

- `text2 += chunk.text` — V8 creates a ConsString (tree of string fragments). This is O(1) memory with lazy flattening.
- `JSON.stringify(result.partial)` — forces V8 to flatten the ConsString into a new contiguous string. For a 50KB accumulated text, this creates a NEW 50KB string every chunk.
- `lastPublishedJson = currentJson` — retains the previous stringified copy.

For N chunks producing a response of size S:
- Total strings created: N copies, average size S/2
- Total memory: N * S / 2 ≈ 13,000 * 55KB / 2 ≈ **350MB per stream**

These strings land in V8's `large_object_space` (objects > 128KB) which is only swept during full stop-the-world GC, not incremental GC.

## Evidence

Heap snapshot analysis (3 concurrent streams producing long outputs ~100KB+):

```
BEFORE (post-GC): 316MB total
  string copies of accumulated text:    4,280 copies    167.8MB

AFTER (post-GC, still 2 tasks running): 2,167MB total
  string copies of accumulated text:   14,351 copies  1,610.2MB
  string copies (earlier stream):       6,785 copies    415.0MB
```

14,351 copies of the same response text, totaling 1.6GB — all from the `JSON.stringify` inside `createOutputTransformStream`.

V8 heap space breakdown at peak (pre-GC):

```
heapUsed=3375MB heapTotal=3658MB
  old_space=1474MB
  large_object_space=1864MB  ← 55% of total heap
```

## Reproduction

```ts

// Any model, any provider
const result = streamText({
  model: yourModel,
  messages: [{ role: "user", content: "Write a very long detailed report about X" }],
  // Default output is text() which triggers the issue
});

// Consume the stream — memory grows quadratically
for await (const chunk of result.textStream) {
  // Each chunk creates a JSON.stringify copy of the full accumulated text
}
```

Monitor `process.memoryUsage().heapUsed` during streaming — it grows quadratically with output length.

## Workaround

Pass a custom output that returns `undefined` from `parsePartialOutput`, skipping the `JSON.stringify` branch:

```ts
const noPartialOutput = {
  name: "text",
  responseFormat: Promise.resolve({ type: "text" as const }),
  async parseCompleteOutput({ text }: { text: string }) { return text; },
  async parsePartialOutput() { return undefined; },  // Skip — prevents N string copies
  createElementStreamTransform() { return undefined; },
};

const result = streamText({
  model: yourModel,
  messages,
  output: noPartialOutput,  // Bypasses the quadratic string copy
});
```

## Suggested Fix

For the default `text()` output, `parsePartialOutput` should return `undefined` since there's no structured output to parse incrementally. The partial output feature only makes sense for `Output.object()` where JSON is being parsed progressively.

```diff
var text = () => ({
  name: "text",
  responseFormat: Promise.resolve({ type: "text" }),
  async parseCompleteOutput({ text: text2 }) {
    return text2;
  },
  async parsePartialOutput({ text: text2 }) {
-   return { partial: text2 };
+   return undefined;  // No partial output for plain text — skip JSON.stringify
  },
});
```

This is a one-line fix that eliminates quadratic memory growth for all plain-text `streamText` calls.

## Verified Results

We deployed the workaround in production and measured the impact with 3 concurrent streams producing long outputs (~100KB+):

| Metric | Before fix | After fix | Improvement |
|--------|-----------|-----------|-------------|
| Heap at task end | 3374MB | 133MB | 25x reduction |
| Cgroup container usage | 95% (OOM risk) | 10% | 10x reduction |
| GC reclaim needed | 2166MB → still high | 120MB → baseline | Negligible GC needed |
| `large_object_space` | 1864MB | ~0MB | Eliminated |

The quadratic string copy was the primary cause of OOM kills on our containers (4GB). After the fix, 3 concurrent streams use ~400MB total instead of ~4GB.

## Environment

- `ai` version: 6.0.116
- Node.js: 22
- Platform: Containerised Node.js environment (4GB memory limit)

Related issue: https://github.com/vercel/ai/issues/10882 (request/response body retention — separate but related)
- Confirmed identical behavior on ai 6.0.116 and ai 6.0.138 (latest)

### AI SDK Version

- ai: 6.0.116

### Code of Conduct

LikiosSedo:
## Summary
- The default `text()` output's `parsePartialOutput` returned `{ partial: text }`, causing `JSON.stringify` on every streaming chunk in `createOutputTransformStream`
- For long responses (~110KB, ~13K chunks), this creates O(n²) string copies totaling ~350MB per stream, causing OOM kills with concurrent streams
- Return `undefined` instead, since plain text has no structured partial output to track — the `JSON.stringify` path is only meaningful for `Output.object()`

## Test plan
- Updated `output.test.ts` to verify `parsePartialOutput` returns `undefined` for `text()` output
- All 59 existing tests pass

Fixes #13839


aayush-kapoor (commented):
needs a test case

PR #13844, opened 2026-03-26
fix(xai): stream drops final text fragments when output_text.delta is incomplete

### LINKED ISSUE #13836: @ai-sdk/xai responses stream drops final text fragments in doStream when content block already exists
## Description

I found and validated a bug in `@ai-sdk/xai` Responses streaming.

When using the xAI Responses path, the adapter can silently drop final text fragments in streaming mode if the content block already exists.

This was observed while using:
- provider: `xai`
- model: `grok-4-1-fast-non-reasoning`

The issue was not caused by UI rendering, export, or database persistence. The truncated text was already persisted in the streamed text part.

### Minimal reproduction context
- Runtime using `@ai-sdk/xai`
- Responses streaming path
- Long text output
- Model: `grok-4-1-fast-non-reasoning`

### Reproduction prompt
Use a long deterministic prompt such as:
1. Write a numbered list from 1 to 120.
2. Each line must be: `<N>. alfa beta gama delta epsilon zeta eta theta iota kappa lambda mu`
3. Then write a JavaScript code block with 80 numbered comment lines: `// 1` through `// 80`
4. End with the exact line: `END_OK_9981`

### Observed result before patch
- Output truncated mid-line.
- Missing final marker `END_OK_9981`.
- `finish: stop` with no error.
- Truncated text already present in persisted streamed output.

### Observed result after patch
- Full output restored.
- Final marker `END_OK_9981` present.
- Expected text length restored.

## Root Cause

In `@ai-sdk/xai`, inside the Responses stream handler, the branch handling `response.output_item.added` and `response.output_item.done` for `part.type === "message"` only emits `text-delta` when the content block does **not** already exist.

When the block already exists, subsequent `contentPart.text` fragments are ignored.

## Problematic Logic

```js
if (contentPart.text && contentPart.text.length > 0) {
  const blockId = `text-${part.id}`;
  if (contentBlocks[blockId] == null) {
    contentBlocks[blockId] = { type: "text" };
    controller.enqueue({
      type: "text-start",
      id: blockId
    });
    controller.enqueue({
      type: "text-delta",
      id: blockId,
      delta: contentPart.text
    });
  }
}
```

## Proposed Fix

`text-start` should only happen when the block is first created, but `text-delta` should be emitted every time `contentPart.text` exists.

```js
if (contentPart.text && contentPart.text.length > 0) {
  const blockId = `text-${part.id}`;
  if (contentBlocks[blockId] == null) {
    contentBlocks[blockId] = { type: "text" };
    controller.enqueue({
      type: "text-start",
      id: blockId
    });
  }
  controller.enqueue({
    type: "text-delta",
    id: blockId,
    delta: contentPart.text
  });
}
```

## Validation
I validated this with a controlled long-output test.
- **Before patch:** Truncated output, missing end marker.
- **After patch:** Full output restored, end marker present, length matches expectation.

## Environment
- **OS:** Debian Linux
- **Provider:** xAI
- **Model:** grok-4-1-fast-non-reasoning

## Additional Context
This initially looked like a provider-side short generation issue, but patching the stream handler restored the full output in the validated runtime.

restareaByWeezy:
Closes #13836

In the xAI Responses stream handler, `response.output_item.done` carries the full accumulated text for a content block. Previously, if the block already existed (created by earlier `output_text.delta` events), the text from `output_item.done` was completely skipped. This works fine when all text is delivered via delta events, but if the stream ends before all deltas arrive, the remaining text gets silently dropped.

The fix tracks how many characters were already streamed via `output_text.delta` events (`streamedLength`), and when `output_item.done` fires, only emits the slice that wasn't already delivered. So:
- Normal case (all deltas arrived): `remaining` is empty, nothing extra emitted
- Truncated case (some deltas missing): only the missing tail gets emitted

### DIFF
--- a/.changeset/fix-xai-stream-text-drop.md
+++ b/.changeset/fix-xai-stream-text-drop.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+Fix text truncation in xAI Responses streaming when output_text.delta events don't deliver all text fragments
--- a/packages/xai/src/responses/xai-responses-language-model.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.ts
@@ -485,7 +485,10 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
     };
     let usage: LanguageModelV4Usage | undefined = undefined;
     let isFirstChunk = true;
-    const contentBlocks: Record<string, { type: 'text' }> = {};
+    const contentBlocks: Record<
+      string,
+      { type: 'text'; streamedLength: number }
+    > = {};
     const seenToolCalls = new Set<string>();
 
     // Track ongoing function calls by output_index so we can stream
@@ -612,12 +615,14 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
               const blockId = `text-${event.item_id}`;
 
               if (contentBlocks[blockId] == null) {
-                contentBlocks[blockId] = { type: 'text' };
+                contentBlocks[blockId] = { type: 'text', streamedLength: 0 };
                 controller.enqueue({
                   type: 'text-start',
                   id: blockId,
                 });
               }
+
+              contentBlocks[blockId].streamedLength += event.delta.length;
 
               controller.enqueue({
                 type: 'text-delta',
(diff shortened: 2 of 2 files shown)


aayush-kapoor (commented):
marking as closed in favor of https://github.com/vercel/ai/pull/16657

PR #13846, opened 2026-03-26
fix(docs): correct language model middleware source links

raashish1601:
## Summary
- update the language model middleware guide so it links to the current `main` branch `LanguageModelV4` source instead of the stale `v5` / `v2` path
- align the `LanguageModelV4Middleware` reference page with the current provider source by correcting `specificationVersion` from `v3` to `v4`

Fixes #12617.

## Validation
- confirmed the stale `blob/v5/.../language-model-v2.ts` links are removed from the middleware guide
- checked the middleware reference page against `packages/provider/src/language-model-middleware/v4/language-model-v4-middleware.ts`
- ran `git diff --check`


gr2m:
Closing in favor of #16487, which supersedes this fix for #12617 by pointing the middleware docs at the current v4 language model spec and updating the v4 middleware reference.

PR #13868, opened 2026-03-27
fix(ai): prevent duplicate tool parts in tool-input-available handler

### LINKED ISSUE #12772: Two tool parts with the same toolCallId created when the model calls a non-existent tool name

pawaca:
## Summary

Follow-up to #12774 which fixed the same duplicate-part bug in the `tool-input-error` handler.

The `tool-input-available` handler has the same bug: it uses `chunk.dynamic` directly to decide whether to call `updateDynamicToolPart` or `updateToolPart`, without first checking if there's already an existing part for the same `toolCallId`. This causes a duplicate when:

1. `tool-input-start` creates a static `tool-{name}` part (dynamic is `undefined`/`false`)
2. `tool-input-available` arrives with `dynamic: true` (from `parseToolCall`'s `NoSuchToolError` catch)
3. `updateDynamicToolPart` doesn't find an existing `dynamic-tool` part, so it creates a **new** one — resulting in two parts for the same `toolCallId`

**Real-world impact**: sessions become permanently broken when the model calls a non-existent tool, because the duplicate `tool_use` IDs cause API rejection on every subsequent request.

The fix applies the same pattern already used in the `tool-input-error` case: look up the existing part by `toolCallId` and use its type to decide the update path, falling back to `chunk.dynamic` only when no existing part is found.

Closes #12772

## Test plan

- [x] Added regression test `tool input available with dynamic flag mismatch` (mirrors the existing `tool input error with dynamic flag mismatch` test from #12774)
- [x] Test asserts exactly one part exists for the `toolCallId` (no duplicate)
- [x] Test asserts the static tool type from `tool-input-start` is preserved
- [x] All 97 existing tests in `process-ui-message-stream.test.ts` continue to pass

### DIFF
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -6706,6 +6706,104 @@ describe('processUIMessageStream', () => {
     });
   });
 
+  describe('tool input available with dynamic flag mismatch', () => {
+    // Regression: when tool-input-start creates a static part (dynamic is
+    // undefined because the tool isn't in the tools object) and tool-input-available
+    // arrives with dynamic: true (from parseToolCall's catch for NoSuchToolError),
+    // the available handler should update the existing static part instead of
+    // creating a second dynamic-tool part.
+    beforeEach(async () => {
+      const stream = createUIMessageStream([
+        {
+          type: 'start',
+        },
+        {
+          type: 'start-step',
+        },
+        {
+          toolCallId: 'call-1',
+          toolName: 'nonExistentTool',
+          type: 'tool-input-start',
+          // dynamic is NOT set (undefined) — this is what happens when the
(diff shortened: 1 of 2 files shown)


pawaca:
Friendly ping — this is a sibling fix to #12774 (which was merged for the `tool-input-error` case). The same duplicate-part bug exists in the `tool-input-available` handler and causes real user impact: sessions become permanently broken when the model calls a non-existent tool, because duplicate `tool_use` IDs cause API rejection on every subsequent request.

The fix is minimal (13 lines, same pattern already merged in #12774) and includes a regression test. Would appreciate a review when you get a chance. Thanks!

lgrammel:
Closing this PR because issue #12772 has already been closed.

PR #13880, opened 2026-03-27
fix(ai): deduplicate concurrent resumeStream calls on shared Chat instances

### LINKED ISSUE #13863: Resume:true triggers resume-stream request for every usage of useChat when Chat is in shared context
### Description

For background, see https://github.com/vercel/ai/pull/11509. In this case, React StrictMode runs effects twice, so the request happens twice. (not ideal, but not a disaster)

However...

This behavior is exacerbated when [the Chat instance is within a shared context](https://ai-sdk.dev/cookbook/next/use-shared-chat-context). In an app where:

1. ChatContext holds a Chat instance in state and exposes a useChatContext hook.
2. 20+ components around the app rely on the following pattern:

```ts
const { chat } = useSharedChatContext();
const { messages } = useChat({
  chat,
});
```

Observed: 20+ resume-stream requests are fired.

Image

Expected: 1 request fires.

### AI SDK Version

- "@ai-sdk/react": "3.0.101"
- "ai": "6.0.99"

### Code of Conduct

claygeo:
## Problem

When multiple `useChat` hooks share the same `Chat` instance via React context ([shared chat pattern](https://ai-sdk.dev/cookbook/next/use-shared-chat-context)), every hook independently calls `resumeStream()` on mount. With `resume: true` and 20+ components, this fires 20+ resume-stream requests to the server instead of 1.

```tsx
// 20+ components all do this:
const { chat } = useSharedChatContext();
const { messages } = useChat({ chat, resume: true });
// Each mounts its own useEffect -> each calls resumeStream() -> 20+ requests
```

Expected: 1 resume-stream request. Actual: N requests (one per `useChat` hook instance).

## Root Cause

In `packages/ai/src/ui/chat.ts`, `resumeStream()` was a bare passthrough to `makeRequest()` with no deduplication. Each `useChat` hook in `packages/react/src/use-chat.ts` registers its own `useEffect` that calls `chatRef.current.resumeStream()`, and since they all share the same `Chat` object, N hooks means N concurrent `reconnectToStream()` requests.

## Solution

Coalesce concurrent calls behind an in-flight promise. While a resume is running, additional calls reuse the same promise instead of starting another request. The promise is cleared in `finally` on every settle path (success, no-active-stream, and error), so a later resume starts fresh.

```typescript
resumeStream = async (options: ChatRequestOptions = {}): Promise<void> => {
  if (this.resumePromise) {
    return this.resumePromise;
  }

  this.resumePromise = this.makeRequest({
    trigger: 'resume-stream',
    ...options,
  })
    .then(() => undefined)
    .finally(() => {
      this.resumePromise = undefined;
    });

  return this.resumePromise;
};
```

`.then(() => undefined)` keeps the field typed as `Promise<void>` (`makeRequest` resolves to `void | null` on the abort path). This is the same coalescing pattern SWR, React Query, and Apollo use for request dedup.

Note this is first-call-wins on options: while a resume is in flight, later callers' options are not applied. That is intentional and safe for the shared-context scenario, where every caller issues the same parameterless resume.

## Changes

- `packages/ai/src/ui/chat.ts` — `resumePromise` field + dedup guard in `resumeStream()`
- `packages/ai/src/ui/chat.test.ts` — regression tests (below)
- `.changeset/cool-resume-dedup.md` — patch changeset for `ai`

## Tests

New `resumeStream deduplication` suite in `chat.test.ts`:

- 3 concurrent `resumeStream()` calls on a shared instance -> exactly 1 `reconnectToStream` request. Fails on current `main` with 3, passes with the fix.
- Two sequential (awaited) resumes -> 2 requests. Proves the dedup is in-flight only, not permanent.
- No-active-stream path (`reconnectToStream` resolves to `null`) -> the guard still resets, so a later resume fires.

Verified locally: `pnpm --filter ai test:node` (chat suite 29/29), `pnpm type-check` clean, `pnpm check` (ultracite) clean.

## Notes

Rebased onto current `main`. As a side effect the dedup also avoids the pre-fix situation where N overlapping resumes each overwrite `this.activeResponse`. It is complementary to the open abort-on-resume PRs (#14350 / #12924), which touch `makeRequest`'s reconnect call rather than the `resumeStream` wrapper, so there is no overlap.

Closes #13863

### DIFF
--- a/.changeset/cool-resume-dedup.md
+++ b/.changeset/cool-resume-dedup.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ai): deduplicate concurrent resumeStream calls on shared Chat instances
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -2952,4 +2952,111 @@ describe('Chat', () => {
       `);
     });
   });
+
+  describe('resumeStream deduplication', () => {
+    function createResumeStream(): ReadableStream<UIMessageChunk> {
+      return new ReadableStream<UIMessageChunk>({
+        start(controller) {
+          controller.enqueue({ type: 'start' });
+          controller.enqueue({ type: 'finish' });
+          controller.close();
+        },
+      });
+    }
+
+    it('should make a single resume request for concurrent resumeStream calls on a shared chat instance', async () => {
+      let reconnectCount = 0;
+      // Hold the resume request in flight so all three concurrent calls are
+      // provably overlapping before any of them settles.
+      const release = createResolvablePromise<void>();
+
+      const chat = new TestChat({
+        id: '123',
+        generateId: mockId(),
+        transport: {
+          sendMessages: async () => {
+            throw new Error('sendMessages should not be called during resume');
+          },
+          reconnectToStream: async () => {
+            reconnectCount++;
+            await release.promise;
+            return createResumeStream();
+          },
+        },
+      });
+
+      // Simulate many useChat hooks sharing one Chat instance, each firing
+      // resumeStream() on mount (e.g. via `resume: true`).
+      const resumed = Promise.all([
+        chat.resumeStream(),
+        chat.resumeStream(),
+        chat.resumeStream(),
+      ]);
+
+      // While the first resume is still in flight, only one request has been
+      // made even though resumeStream() was called three times.
+      expect(reconnectCount).toBe(1);
+
+      release.resolve();
+      await resumed;
+
+      // Still a single request after everything settles.
+      expect(reconnectCount).toBe(1);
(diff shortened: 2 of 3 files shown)


claygeo:
Revived and rebased onto current `main` (now mergeable). Since the original commit I added regression tests and a changeset, and fixed a type issue so it passes `type-check:full` and `ultracite check` (verified locally, chat suite 29/29 green).

The dedup tests fail on current `main` (3 concurrent `resumeStream()` calls produce 3 `reconnectToStream` requests) and pass with the fix.

On checks: the one red check is the Vercel preview-deploy gate ("Authorization required to deploy"), which is the fork-PR deploy authorization, not a code failure. The lint / typecheck / test workflows need a maintainer to approve fork-PR runs. Happy to adjust anything.

PR #13887, opened 2026-03-27
feat(provider/xai): add video extension and reference-to-video (R2V) support

Jaaneek:
## Background

The xAI video API supports five generation modes, but the SDK only implemented three (text-to-video, image-to-video, editing). Video extension (`POST /v1/videos/extensions`) and reference-to-video/R2V (`POST /v1/videos/generations` with `reference_images`) were missing. The xAI provider docs were also outdated, still referencing `grok-4-fast-non-reasoning` and missing `grok-4.20` models, `grok-imagine-image-pro`, and the new video modes.

## Summary

**Provider (`@ai-sdk/xai`):**
- Add video extension support via `extensionUrl` provider option → `POST /v1/videos/extensions`
- Add reference-to-video (R2V) support via `referenceImageUrls` provider option (1–7 images) → `POST /v1/videos/generations`.
- Surface `progress` field in provider metadata
- Add `error` to response schema (defensive)
- Fix `prompt: options.prompt ?? ''` for image-to-video without text

**Documentation:**
- Add Video Extension and R2V sections with code examples
- Document `extensionUrl` and `referenceImageUrls` provider options
- Document `grok-imagine-image-pro` with `resolution` (`1k`/`2k`) and `quality` options
- Update all code examples to `grok-4.20-reasoning` / `grok-4.20-non-reasoning`
- Update model capabilities table with grok-4.20 models
- Update video capabilities table with Extension and R2V columns
- Switch examples and docs from `{ videos }` to `{ video }` (singular). I feel like this is better since xAI does not allow multiple videos generation but let me know.

**Examples:**
- New: `extend.ts`, `extend-warnings.ts`, `reference-images.ts`
- Updated: `basic.ts`, `edit-concurrent.ts`, `edit-warnings.ts` with null-safe metadata access and `{ video }` pattern

**Tests:** 62 total (up from ~25), covering all new modes, validation, edge cases, polling, and metadata.

## Checklist

## Future Work

- Surface `error.message` from API in `failed` status handler (xAI doesn't document this field yet)
- Consider adding optional `url` property to core `GeneratedFile` to avoid `providerMetadata` dance for chaining


packages/xai/src/xai-video-model.ts
```diff
@@ -291,6 +388,13 @@ export class XaiVideoModel implements Experimental_VideoModelV4 {
         });
       }
 
+      if (statusResponse.status === 'failed') {
+        throw new AISDKError({
+          name: 'XAI_VIDEO_GENERATION_FAILED',
+          message: 'Video generation failed.',
```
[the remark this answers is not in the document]
Jaaneek: I will follow up with another PR where we will type all the codes if it's okay with you.
I need to first do it on xAI BE side

packages/xai/src/xai-video-model.ts
```diff
@@ -63,6 +63,51 @@ export class XaiVideoModel implements Experimental_VideoModelV4 {
...
+      });
+    }
+    if (options.image != null && isExtension) {
+      throw new AISDKError({
+        name: 'XAI_VIDEO_INVALID_OPTIONS',
+        message:
+          'image and extensionUrl are mutually exclusive — ' +
+          'use image for image-to-video or extensionUrl for video extension, not both.',
+      });
+    }
```
felixarntz: This feels like overkill; what's already covered by types, we shouldn't need to validate, especially since the API would error anyway.

Additionally, we encourage using `warnings` and graceful fallback where possible instead of errors. Maybe in this case errors are fine, but I'd say we can just let the API handle them.

packages/xai/src/xai-video-model.ts
```diff
@@ -117,22 +163,46 @@ export class XaiVideoModel implements Experimental_VideoModelV4 {
...
+        type: 'unsupported',
+        feature: 'resolution',
+        details: 'xAI video extension does not support custom resolution.',
+      });
+    }
+
     const body: Record<string, unknown> = {
       model: this.modelId,
-      prompt: options.prompt,
+      prompt: options.prompt ?? '',
```
felixarntz: When is a prompt not needed? Can we apply this change conditionally only for that case instead?

packages/xai/src/xai-video-model.ts
```diff
@@ -63,6 +63,51 @@ export class XaiVideoModel implements Experimental_VideoModelV4 {
     })) as XaiVideoModelOptions | undefined;
 
     const isEdit = xaiOptions?.videoUrl != null;
+    const isExtension = xaiOptions?.extensionUrl != null;
```
felixarntz: It's a bit of a strange API to use a different field to provide the video, just to indicate whether we should use extension mode.

It would be more intuitive to keep using `videoUrl` (since it's still modifying an input video), and instead use something like e.g. `xaiOptions?.editMode === 'extension'`.

packages/xai/src/xai-video-model.ts
Jaaneek: >This feels like overkill; what's already covered by types, we shouldn't need to validate, especially since the API would error anyway.

Okay!

packages/xai/src/xai-video-model.ts
Jaaneek: @felixarntz Should we switch to mode instead of editMode and handle all the possible scenarios in one place?
It would probably be easier to extend later if we add more permutations, and the DX feels a bit nicer too because devs can just type mode="image-to-video" or mode="edit-video" and get full TypeScript autocomplete
Wdyt?

packages/xai/src/xai-video-model.ts
Jaaneek: done

packages/xai/src/xai-video-model.ts
Jaaneek: When doing image to video it's fine to pass empty string

Jaaneek: I just reverted this, will change it in the xAI api to align more with other providers

packages/xai/src/xai-video-model.ts
felixarntz: using `mode` SGTM 👍

Jaaneek: Awesome on it

packages/xai/src/xai-video-options.ts
```diff
@@ -1,22 +1,157 @@
...
+});
+
+export const xaiVideoModelOptions = z.union([
+  editVideoSchema,
+  extendVideoSchema,
+  referenceToVideoSchema,
+  autoDetectSchema,
+]);
 
+// Runtime schema uses a flat passthrough object for lenient parsing.
```
Jaaneek: lmk if you like comments like this, I feel like they are not needed but I have seen them in other places

packages/xai/src/xai-video-model.ts
Jaaneek: @felixarntz Done, let me know what you think about this implementation. Happy to keep iterating

packages/xai/src/xai-video-options.ts
```diff
@@ -1,22 +1,157 @@
...
+const nonEmptyStringSchema = z.string().min(1);
+const resolutionSchema = z.enum(['480p', '720p']);
+const modeSchema = z.enum(['edit-video', 'extend-video', 'reference-to-video']);
+
+// ── Shared fields present in every variant ───────────────────────────
+const baseFields = {
+  pollIntervalMs: z.number().positive().nullish(),
+  pollTimeoutMs: z.number().positive().nullish(),
+  resolution: resolutionSchema.nullish(),
+};
```
felixarntz: nit: move this down right above where it's used for better colocation of code

packages/xai/src/xai-video-options.ts
```diff
@@ -1,22 +1,157 @@
 import { lazySchema, zodSchema } from '@ai-sdk/provider-utils';
 import { z } from 'zod/v4';
 
-export type XaiVideoModelOptions = {
```
felixarntz, line 4: This was previously including what's now in `XaiVideoSharedOptions` (pollIntervalMs, etc.), but now it doesn't as far as I can tell. So that would be a back compat break. Or maybe I'm missing something?

packages/xai/src/xai-video-options.ts
```diff
@@ -1,22 +1,157 @@
...
+ */
+export type XaiVideoModelOptions =
+  | XaiVideoGenerationOptions
+  | XaiVideoEditModeOptions
+  | XaiVideoExtendModeOptions
+  | XaiVideoReferenceToVideoOptions
+  | XaiLegacyEditVideoOptions
+  | XaiLegacyReferenceToVideoOptions;
+
+export interface XaiParsedVideoModelOptions extends XaiVideoSharedOptions {
```
felixarntz: Not sure about the name of this one. Isn't it rather the "unparsed" version? if I understand correctly, this only exists for back compat, i.e. legacy, or not? If so, maybe call it `XaiLegacyVideoModelOptions` and mark it as deprecated?

This way people don't have to switch right away, but they'd be encouraged to update their `providerOptions` usage to rely on `XaiVideoModelOptions` soon.

examples/ai-functions/src/generate-video/xai/basic.ts
felixarntz, line 19: Maybe we should update these to use the new format, with `mode`? We could keep 1-2 examples to specifically use the legacy format, and include `legacy-` in the filename.

examples/ai-functions/src/generate-video/xai/edit.ts
```diff
@@ -21,5 +21,5 @@ run(async () => {
       }),
   );
 
-  await presentVideos(videos);
+  await presentVideos([video]);
```
felixarntz, line 25: not sure why, but this example somehow fails for me consistently. always around ~6s I get:
```
XAI_VIDEO_GENERATION_FAILED: Video generation failed.
    at XaiVideoModel.doGenerate (/.../packages/xai/src/xai-video-model.ts:372:15)
```

Not sure why; the other examples work flawlessly, also extend etc.

felixarntz (commented):
@Jaaneek This looks great! almost good to go, just running into one problem with the video editing example.

examples/ai-functions/src/generate-video/xai/edit.ts
Jaaneek: I encountered the same issue and just switched the videoURL to confirm it's nothing related ai sdk.

This is a bug on our (xAI) side, we have fix ready. I will try to push to have it in production asap but realistically it can take up to 24h

I will fix all other comments and ping you once the fix is deployed

Jaaneek: Should work, lmk if it does not

examples/ai-functions/src/generate-video/xai/basic.ts
Jaaneek: done, good callout

packages/xai/src/xai-video-options.ts
Jaaneek: done

packages/xai/src/xai-video-options.ts
Jaaneek: I refactored it a bit so its more readable but overall all the interfaces extend the `XaiVideoSharedOptions` so it should have them but let me know if I missed something in here

felixarntz: nevermind, I misread the code here

felixarntz (approved):
@Jaaneek Thanks, everything working great now. Good to merge, and then I'll backport to v6!

ghost:
⚠️ Backport to release-v6.0 created but has conflicts: https://github.com/vercel/ai/pull/14079

PR #13889, opened 2026-03-28
feat(ai): add Realtime API support in AI SDK

aayush-kapoor:
## Background

providers like OpenAI, Google, xAI support realtime apis where speech-to-speech communication is essentially possible.
the AI SDK didn't support it yet

## Summary

_not yet added_

## Manual Verification

_not yet added_

## Checklist

## Future Work

_not yet added_

## Related Issues

_not yet added_


packages/ai/src/realtime/generate-realtime-token.ts
```diff
@@ -0,0 +1,42 @@
...
+  expiresAfterSeconds,
+}: {
+  model: RealtimeModelV1;
+  expiresAfterSeconds?: number;
+}): Promise<{
+  token: string;
+  url: string;
+  expiresAt?: number;
+}> {
+  const result = await model.doCreateClientSecret({
```
felixarntz, line 35: both of these functions need to support more config arguments. Both OpenAI and Google allow passing config data here, e.g. which voice to use, which response modalities to support, which tools are available.

packages/provider/src/realtime-model/v4/realtime-model-v4-client-event.ts
```diff
@@ -0,0 +1,68 @@
...
+      type: 'response-create';
+      options?: {
+        modalities?: string[];
+        instructions?: string;
+        metadata?: Record<string, unknown>;
+      };
+    }
+  | {
+      type: 'response-cancel';
+    };
```
felixarntz, line 68: Some of the events, like these ones, are only relevant if VAD is disabled. Since this is a complex feature, I think it would be a good idea to reduce the scope and focus on VAD only implementation for this first iteration.

We can then reconsider support for disabling VAD and relevant events needed in a follow up PR.

packages/provider/src/realtime-model/v4/realtime-model-v4-client-event.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Each provider maps this to its native event format before sending
+ * over the WebSocket.
+ */
+export type RealtimeModelV4ClientEvent =
+  // ── Session ────────────────────────────────────────────────────────
+
+  | {
+      type: 'session-update';
+      config: RealtimeModelV4SessionConfig;
+    }
```
felixarntz, line 15: I think this is also something pretty low-level which is not crucial to support. Updating an existing session is advanced, and there are probably many edge cases to consider. I'd say let's remove and consider adding as a spec primitive later.

packages/provider/src/realtime-model/v4/realtime-model-v4-server-event.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  | {
+      type: 'session-created';
+      sessionId?: string;
+      raw: unknown;
+    }
+  | {
+      type: 'session-updated';
+      raw: unknown;
+    }
```
felixarntz, line 19: See above, maybe remove and leave for a separate follow up PR to consider?

packages/provider/src/realtime-model/v4/realtime-model-v4-server-event.ts
```diff
@@ -0,0 +1,199 @@
...
+      type: 'speech-stopped';
+      itemId?: string;
+      raw: unknown;
+    }
+  | {
+      type: 'audio-committed';
+      itemId?: string;
+      previousItemId?: string;
+      raw: unknown;
+    }
```
felixarntz, line 37: These names are heavily based on OpenAI event names, but maybe we should align more with our own conventions that we use e.g. for stream part types?

e.g. `speech-start`, `speech-end`, `audio-commit` (no past tense)

gr2m (changes requested):
### 🛑 YOU SHALL NOT PASS 🛑

packages/ai/src/realtime/realtime-session.ts
```diff
@@ -0,0 +1,496 @@
...
+        const config: RealtimeSessionConfig = {
+          ...this.sessionConfig,
+          tools: toolDefinitions as RealtimeSessionConfig['tools'],
+        };
+
+        this.sendRaw(
+          this.model.serializeClientEvent({
+            type: 'session-update',
+            config,
+          }),
```
felixarntz: this part if OpenAI (maybe xAI too?) specific; Google requires a different way to pass config. so this needs to be abstracted away, and then this code will need to live in the `openai` package

felixarntz:
Superseded by #13893

Issue #13706, opened 2026-03-21
Add support for Gemini Interactions API


tianqiwuben:
### Description

https://ai.google.dev/gemini-api/docs/interactions

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jingchang0623-crypto:
This would be great! The Gemini Interactions API is quite powerful for real-time voice/video interactions.

For anyone interested in this, here's what the Interactions API offers:
- Low-latency streaming
- Voice and video modalities
- Session management

This would be a huge addition to the Vercel AI SDK ecosystem! 🚀

lgrammel:
## Identify issue type

Identified as **Feature**  
Confidence: **High**

### Rationale
This requests new provider/API capabilities rather than a correction to existing behavior, documentation, or examples.

### Evidence
- The title explicitly requests support for the Gemini Interactions API.
- The description links to the Gemini Interactions API documentation without reporting broken existing behavior.
- A comment characterizes it as an addition providing streaming, voice/video modalities, and session management.

PR #9559, opened 2025-10-16
feat(ai): add websocket transport support

aayush-kapoor:
## Background

To support api's needing websockets

## Summary

Added wss-transport method, along with test files and example

## Manual Verification

Tests and example added to communicate with a wss server and use AI-SDK

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] Formatting issues have been fixed (run `pnpm prettier-fix` in the project root)

## Future Work

Add sse transport

## Related Issues

5607, 3176

Issue #5007, opened 2025-02-27
Support for audio outputs with gpt-4o-audio-preview


raphtlw:
### Feature Description

Currently, when using the @ai-sdk/openai provider, audio inputs are accepted, but parameters for telling the model to generate audio output are not.

```shell
[01:01:07.246] INFO (6593): Document file type
    ext: "opus"
    mime: "audio/ogg; codecs=opus"
[01:01:07.347] DEBUG (6593): {
  toSend: [
    {
      type: 'file',
      mimeType: 'audio/mpeg',
      data: <Buffer 49 44 33 04 00 00 00 00 00 23 54 53 53 45 00 00 00 0f 00 00 03 4c 61 76 66 36 30 2e 31 36 2e 31 30 30 00 00 00 00 00 00 00 00 00 00 00 ff fb 54 c0 00 ... 20539 more bytes>
    }
  ],
  remindingSystemPrompt: []
}
```

```shell
Body: {
  model: 'gpt-4o-audio-preview',
  temperature: 0,
  messages: [
    {
      role: 'system',
      content: 'You are raphGPT, a large language model created by @raphtlw, based on the GPT-4 architecture.\n' +
        '\n' +
        'Current date: 2/28/2025, 1:01:08 AM\n' +
        '\n' +
        'Image input capabilities: Enabled\n' +
        'Preferred language: english\n' +
        'Yourself: {&quot;id&quot;:7120507228,&quot;is_bot&quot;:true,&quot;first_name&quot;:&quot;raphGPT (dev)&quot;,&quot;username&quot;:&quot;raphgptdevbot&quot;,&quot;can_join_groups&quot;:true,&quot;can_read_all_group_messages&quot;:false,&quot;supports_inline_queries&quot;:false,&quot;can_connect_to_business&quot;:false,&quot;has_main_web_app&quot;:false}\n' +
        '\n' +
        'Personality: \n' +
        '\n' +
        'raphGPT is a direct, no-nonsense conversationalist who communicates with brevity, humor, and spontaneity. Responses should be concise, informal, and to the point—no unnecessary fluff. Use casual phrasing, abbreviations, and quick decision-making. Inject humor or playfulness when appropriate, but keep interactions practical. If something is obvious, acknowledge it briefly. When discussing logistics or plans, prioritize efficiency and straightforwardness. Assume familiarity with the user, responding in a way that mimics natural, relaxed conversation. Avoid overly formal or robotic language.\n' +
        '\n' +
[...]

Issue #3907, opened 2024-11-26
When using sdk/openai, important data is discarded by the SDK


adolphnov:
### Description

When I use the gpt-4o-audio-preview model and pass the audio, the fetch method I've implemented myself, chat with openai to require the audio and transcription, the audio in the returned data was discarded, rendering the entire conversation completely invalid.

### AI provider

@ai-sdk/openai v1.0.4

### Additional context

I found that the zodSchema.safeParse method dropped the audio field.

adolphnov:
> Can you provide details on which information you would need to be returned? currently only audio inputs are supported

I have input the audio and passed in additional parameters myself:
```

readonly fetch = async (url: RequestInfo | URL, options?: RequestInit): Promise<Response> => {
        const body = JSON.parse(options?.body as string);
        if (body.model === 'gpt-4o-audio-preview') {
            body.modalities = ['text', 'audio'];
            body.audio = { voice: 'alloy', format: 'opus' };
        }
        return fetch(url, {
            ...options,
            body: JSON.stringify(body),
        });

```

The original response from OpenAI contains an audio field:
```

{
  "index": 0,
  "message": {
    "role": "assistant",
    "content": null,
    "refusal": null,
    "audio": {
      "id": "audio_abc123",
      "expires_at": 1729018505,
      "data": "<bytes omitted>",
      "transcript": "Yes, golden retrievers are known to be ..."
    }
  },
  "finish_reason": "stop"
}

```

However, it was filtered by the SDK internally and the audio field was discarded.
![zodSchema.safeParse filted audio field](https://github.com/user-attachments/assets/bd21748f-55cb-4759-84e2-8d68683c344c)

polesapart:
Ideally, ai sdk should formally support audio, as it supports images and pdf. Other ai providers may as well implement audio i/o soon enough, i.e. anthropic & hume ai colaboration rumours are around.

MichaelYuhe:
Any updates? 👀

mrjiangguoqing:
we need this!

raphtlw:
> Any updates? 👀

#5007

lgrammel:
### Identify issue type

**Identified type: Feature**  
Confidence: High

Rationale: The issue requests support for a currently unsupported capability: audio outputs from the OpenAI provider. The discarded field is a consequence of missing SDK support rather than a contradiction of documented behavior.

Evidence:
- Title asks to “Support audio outputs with openai provider.”
[...]

PR #13893, opened 2026-03-28
feat: Realtime API support for browser<->provider websocket connection

gr2m:
## Background

Part of https://github.com/vercel/ai/issues/13897

Support for Realtime API has been requested for a long time. We want to support different architectures

1. Websocket connection from browser directly to provider
2. Websocket connection from browser through user's server to provider
3. Websocket connection from browser through gateway to provider
4. Websocket connection from browser through user's server and gateway to provider

This pull requests implements the first, see [Architecture](#architecture) below

## Summary

Alternative implementation of the Realtime API support proposed in #13889, with a reworked developer-facing API:

- **`openai.experimental_realtime('gpt-realtime')`** works in both server and browser (no separate `@ai-sdk/openai/realtime` import needed)
- **`openai.experimental_realtime.getToken()`** static method for server-side ephemeral token creation
- **`experimental_useRealtime`** hook returns `messages: UIMessage[]` (aligned with `useChat` format) instead of `transcript`
- **`inputAudioTranscription`** session config enables rendered user messages for transcribed microphone input
- **`addToolOutput(callId, result)`** for client-side tools that need manual result submission
- **`onToolCall`** callback for auto-executed client-side tools
- Framework-agnostic helper type: `Experimental_RealtimeSetupResponse`

Provider implementations included: **OpenAI**, **Google**, and **xAI**. ElevenLabs support is split into a stacked follow-up: **#15747**.

<a name="architecture"></a>

## Architecture

```mermaid
sequenceDiagram
  participant Browser
  participant Server
  participant Provider as AI Provider<br/>(OpenAI / xAI / Gemini / ...)

  Browser->>Server: POST /api/setup
  Server->>Provider: Request token (with tool definitions)
  Provider-->>Server: Short-lived auth token
  Server-->>Browser: { token }

  Browser->>Provider: Open WebSocket (using token)
  Provider-->>Browser: Audio/text chunks (streaming)
  Provider-->>Browser: Tool call request

  Browser->>Server: App-specific tool request (optional)
  Server-->>Browser: App-specific tool result

  Browser->>Provider: Send tool result via WebSocket
  Provider-->>Browser: Continue streaming
```

#### Generic RPC route for server-side tools

One option was for the SDK to expose a generic server-side tool execution flow: the realtime session would receive a tool call, POST `{ name, inputs, callId }` to an `execute-tools` route, execute a matching server-side tool, and send the result back to the provider automatically.

We decided against that for the initial implementation. A generic RPC route is convenient, but it creates a security-sensitive application boundary: the app must authenticate the user, bind the request to a realtime session, allowlist tool names, validate call IDs, rate limit access, and authorize each tool invocation. If the SDK provides the generic route shape, it is easy to copy into production without those controls.

Instead, the SDK keeps tool execution client-driven through `onToolCall` and `addToolOutput`. Server-backed tools should call app-specific API endpoints from `onToolCall` (for example `/api/weather`), where the application can apply its normal auth, validation, authorization, and rate limiting rules. Documentation should cover secure server-backed tool calling patterns, but the SDK should not own the generic remote tool RPC abstraction yet.

## Manual verification

The realtime voice example added in this PR lives at `examples/ai-e2e-next/app/realtime/page.tsx` (UI) and `examples/ai-e2e-next/app/api/realtime/[...path]/route.ts` (server-side token endpoint).

1. Add the provider API key(s) for whichever provider(s) you want to test to `examples/ai-e2e-next/.env.local`:

   ```bash
   OPENAI_API_KEY=...                  # OpenAI realtime (gpt-realtime)
   GOOGLE_GENERATIVE_AI_API_KEY=...    # Google realtime (gemini-3.1-flash-live-preview)
   XAI_API_KEY=...                     # xAI realtime (grok-voice-latest)
   ```

2. Start the example:

   ```bash
   cd examples/ai-e2e-next
   pnpm dev
   ```

3. Open http://localhost:3000/realtime, pick a provider + voice, and click **Connect**. Use the microphone or type a message. Ask "what's the weather in Paris?" or "roll a dice" to exercise client-side tool calling (`onToolCall`).

## Checklist

- [ ] Security Audit - esp. securing remote tool calling
- [ ] Update internal Architecture / how it works docs

### Future Work

1. **ElevenLabs** provider support — stacked follow-up in #15747
2. Websocket connection from browser through user's server to provider
3. Websocket connection from browser through gateway to provider
4. Websocket connection from browser through user's server and gateway to provider

## Related issues

- #3176
- #3907
- #4082
- #5007
- #9559
- #12381
- #13706
- #13889

### DIFF
--- a/packages/ai/src/realtime/audio-utils.ts
+++ b/packages/ai/src/realtime/audio-utils.ts
@@ -1,0 +1,82 @@
+/**
+ * Converts Float32 audio samples to a base64-encoded PCM16 string
+ * for sending to a realtime model via input_audio_buffer.append.
+ *
+ * Samples are expected to be in the range [-1.0, 1.0].
+ * Output is 16-bit signed integer, little-endian, base64-encoded.
+ */
+export function encodeRealtimeAudio(float32Array: Float32Array): string {
+  const buffer = new ArrayBuffer(float32Array.length * 2);
+  const view = new DataView(buffer);
+
+  for (let i = 0; i < float32Array.length; i++) {
+    const s = Math.max(-1, Math.min(1, float32Array[i]));
+    view.setInt16(i * 2, s < 0 ? s * 0x8000 : s * 0x7fff, true);
+  }
+
+  const bytes = new Uint8Array(buffer);
+  const chunkSize = 0x8000;
+  let binary = '';
+  for (let i = 0; i < bytes.length; i += chunkSize) {
+    const chunk = bytes.subarray(i, i + chunkSize);
+    binary += String.fromCharCode(...chunk);
+  }
+  return btoa(binary);
+}
+
+/**
+ * Converts a base64-encoded PCM16 string (from a realtime model's
+ * audio-delta event) back to Float32 audio samples.
+ *
+ * Input is expected to be 16-bit signed integer, little-endian, base64-encoded.
+ * Output samples are in the range [-1.0, 1.0].
+ */
+export function decodeRealtimeAudio(base64Audio: string): Float32Array {
+  const binaryString = atob(base64Audio);
+  const bytes = new Uint8Array(binaryString.length);
+  for (let i = 0; i < binaryString.length; i++) {
+    bytes[i] = binaryString.charCodeAt(i);
+  }
+
+  const pcm16 = new Int16Array(bytes.buffer);
+  const float32 = new Float32Array(pcm16.length);
+  for (let i = 0; i < pcm16.length; i++) {
+    float32[i] = pcm16[i] / 32768.0;
+  }
+  return float32;
+}
+
+/**
+ * Resamples audio from one sample rate to another using linear
+ * interpolation. Suitable for voice audio.
+ *
+ * @param input - Float32 audio samples at the input sample rate.
+ * @param inputRate - The sample rate of the input audio (e.g. 48000).
+ * @param outputRate - The desired output sample rate (e.g. 24000).
+ * @returns Float32 audio samples at the output sample rate.
+ */
+export function resampleAudio(
+  input: Float32Array,
+  inputRate: number,
+  outputRate: number,
+): Float32Array {
+  if (inputRate === outputRate) {
+    return input;
+  }
+
+  const ratio = inputRate / outputRate;
+  const outputLength = Math.round(input.length / ratio);
+  const output = new Float32Array(outputLength);
+
+  for (let i = 0; i < outputLength; i++) {
+    const srcIndex = i * ratio;
+    const srcIndexFloor = Math.floor(srcIndex);
+    const srcIndexCeil = Math.min(srcIndexFloor + 1, input.length - 1);
+    const fraction = srcIndex - srcIndexFloor;
+
+    output[i] =
+      input[srcIndexFloor] * (1 - fraction) + input[srcIndexCeil] * fraction;
+  }
+
+  return output;
+}
--- a/packages/ai/src/realtime/browser-realtime-audio.ts
+++ b/packages/ai/src/realtime/browser-realtime-audio.ts
@@ -1,0 +1,168 @@
+import {
+  decodeRealtimeAudio,
+  encodeRealtimeAudio,
+  resampleAudio,
+} from './audio-utils';
+
+export type BrowserRealtimeAudioOptions = {
+  captureSampleRate: number;
+  playbackSampleRate: number;
+  onAudio: (base64Audio: string) => void;
+  onPlayingChange: (isPlaying: boolean) => void;
+  onCapturingChange: (isCapturing: boolean) => void;
+};
+
+export class BrowserRealtimeAudio {
+  private readonly captureSampleRate: number;
+  private readonly playbackSampleRate: number;
+  private readonly onAudio: BrowserRealtimeAudioOptions['onAudio'];
+  private readonly onPlayingChange: BrowserRealtimeAudioOptions['onPlayingChange'];
+  private readonly onCapturingChange: BrowserRealtimeAudioOptions['onCapturingChange'];
+
+  private captureContext: AudioContext | null = null;
+  private captureProcessor: ScriptProcessorNode | null = null;
+  private captureSource: MediaStreamAudioSourceNode | null = null;
+  private captureStream: MediaStream | null = null;
+
+  private playbackContext: AudioContext | null = null;
+  private playbackQueue: Float32Array[] = [];
+  private playbackTime = 0;
+  private playbackStartTime = 0;
+  private activeSources = new Set<AudioBufferSourceNode>();
+  private isPlaying = false;
+
+  constructor(options: BrowserRealtimeAudioOptions) {
+    this.captureSampleRate = options.captureSampleRate;
+    this.playbackSampleRate = options.playbackSampleRate;
+    this.onAudio = options.onAudio;
+    this.onPlayingChange = options.onPlayingChange;
+    this.onCapturingChange = options.onCapturingChange;
+  }
+
+  ensurePlaybackContext(): void {
+    if (this.playbackContext == null) {
+      this.playbackContext = new AudioContext({
+        sampleRate: this.playbackSampleRate,
+      });
+    }
+  }
+
+  startCapture(stream: MediaStream): void {
+    const ctx = new AudioContext({ sampleRate: this.captureSampleRate });
+    this.captureContext = ctx;
+    this.captureStream = stream;
+
+    const source = ctx.createMediaStreamSource(stream);
+    this.captureSource = source;
+
+    const processor = ctx.createScriptProcessor(4096, 1, 1);
+    this.captureProcessor = processor;
+
+    processor.onaudioprocess = event => {
+      const inputData = event.inputBuffer.getChannelData(0);
+      const samples = resampleAudio(
+        new Float32Array(inputData),
+        ctx.sampleRate,
+        this.captureSampleRate,
+      );
+      this.onAudio(encodeRealtimeAudio(samples));
+    };
+
+    source.connect(processor);
+    processor.connect(ctx.destination);
+    this.onCapturingChange(true);
+  }
+
+  stopCapture(): void {
+    this.captureProcessor?.disconnect();
+    this.captureSource?.disconnect();
+    void this.captureContext?.close();
+    this.captureStream?.getTracks().forEach(track => track.stop());
+
+    this.captureProcessor = null;
+    this.captureSource = null;
+    this.captureContext = null;
+    this.captureStream = null;
+    this.onCapturingChange(false);
+  }
+
+  playAudio(base64Audio: string): void {
+    this.ensurePlaybackContext();
+    this.playbackQueue.push(decodeRealtimeAudio(base64Audio));
+    this.schedulePlayback();
+  }
+
+  stopPlayback(): void {
+    this.playbackQueue = [];
+
+    for (const source of this.activeSources) {
+      try {
+        source.stop();
+      } catch {
+        // Source may already be stopped by the browser.
+      }
+    }
+    this.activeSources.clear();
+
+    if (this.playbackContext != null) {
+      this.playbackTime = this.playbackContext.currentTime;
+    }
+    this.setPlaying(false);
+  }
+
+  getPlaybackOffsetMs(): number {
+    const ctx = this.playbackContext;
+    if (ctx == null) return 0;
+    return (ctx.currentTime - this.playbackStartTime) * 1000;
+  }
+
+  dispose(): void {
+    this.stopCapture();
+    this.stopPlayback();
+    void this.playbackContext?.close();
+    this.playbackContext = null;
+  }
+
+  private setPlaying(isPlaying: boolean): void {
+    if (isPlaying && !this.isPlaying) {
+      this.playbackStartTime = this.playbackContext?.currentTime ?? 0;
+    }
+    if (this.isPlaying !== isPlaying) {
+      this.isPlaying = isPlaying;
+      this.onPlayingChange(isPlaying);
+    }
+  }
+
+  private schedulePlayback(): void {
+    const ctx = this.playbackContext;
+    if (ctx == null || this.playbackQueue.length === 0) return;
+
+    while (this.playbackQueue.length > 0) {
+      const samples = this.playbackQueue.shift()!;
+      const buffer = ctx.createBuffer(
+        1,
+        samples.length,
+        this.playbackSampleRate,
+      );
+      buffer.getChannelData(0).set(samples);
+
+      const source = ctx.createBufferSource();
+      source.buffer = buffer;
+      source.connect(ctx.destination);
+
+      const startTime = Math.max(this.playbackTime, ctx.currentTime);
+      source.start(startTime);
+      this.playbackTime = startTime + buffer.duration;
+
+      this.activeSources.add(source);
+      this.setPlaying(true);
+
+      source.onended = () => {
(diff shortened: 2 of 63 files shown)


packages/provider/src/realtime-model/v4/realtime-model-v4-client-event.ts
```diff
@@ -0,0 +1,68 @@
...
+      type: 'response-create';
+      options?: {
+        modalities?: string[];
+        instructions?: string;
+        metadata?: Record<string, unknown>;
+      };
+    }
+  | {
+      type: 'response-cancel';
+    };
```
felixarntz, line 68: Per what I already commented in https://github.com/vercel/ai/pull/13889#discussion_r3003842975, I think we should remove these events for now, as they are only relevant for when VAD is disabled.

This PR is already quite complex, and testing that everything works with VAD disabled, for each of the 3 providers, is a whole other thing we haven't even started covering. So I think we should downscope the PR to only cover the events and config params relevant for when using VAD.

I'm not saying we shouldn't allow disabling VAD, but better do that in a separate PR where we can focus exclusively on that.

packages/provider/src/realtime-model/v4/realtime-model-v4-server-event.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  | {
+      type: 'session-created';
+      sessionId?: string;
+      raw: unknown;
+    }
+  | {
+      type: 'session-updated';
+      raw: unknown;
+    }
```
felixarntz, line 19: See https://github.com/vercel/ai/pull/13889#discussion_r3003848320, I think this is also something to better only explore in a follow up PR.

This event isn't really relevant yet, except that some providers use it to configure the session right upon connection. But that's a provider specific nuance, and not really a session _update_.

packages/openai/src/openai-provider.ts
```diff
@@ -93,6 +96,12 @@ export interface OpenAIProvider extends ProviderV4 {
    */
   speech(modelId: OpenAISpeechModelId): SpeechModelV4;
 
+  /**
+   * Creates a realtime model for bidirectional audio/text
+   * communication over WebSocket.
+   */
+  realtime: RealtimeFactoryV4;
```
felixarntz: I wonder whether we shouldn't reconsider splitting this out. We had already talked about how having everything accessible only on the `Provider` type isn't great for tree shaking, but here this becomes even far more critical, as this is somewhat intended to be used on the browser-side, unlike pretty much everything else in our spec.

Maybe we just shouldn't add `RealtimeFactoryV4` to `ProviderV4` - I know it's not there right now because it's experimental, but I'm saying even later when it is we shouldn't add it.

Instead we could export it as `*Realtime` (e.g. `openaiRealtime`) or `realtime*` (e.g. `realtimeOpenAI`). That way it remains entirely separate.

Same feedback applies of course to `google` and `xai` too.

gr2m: > Maybe we just shouldn't add `RealtimeFactoryV4` to `ProviderV4`

100% we shouldn't, and I don't think this PR does? It exports `RealtimeFactoryV4` as `Experimental_RealtimeFactoryV4`, just like `VideoModelV4`

I'm okay not adding `[model].realtime` for now, as we really just target the websocket connection from browsers, we can only export `[model]Realtime` for now.

packages/provider/src/realtime-model/v4/realtime-model-v4-client-event.ts
gr2m: I didn't have a chance to review realtime events in detail now, it's the main reason why this is still a draft. I'm 100% with you.

Since you looked already more into it, do you want to update this PR directly, or maybe create a PR against `gr2m/realtime-api-v2` to reduce the events spec to the minimum we need right now?

ArshVermaGit (commented):
This PR introduces a well-designed first step toward comprehensive Realtime API support, addressing a long-standing feature request and establishing a flexible foundation for multiple connection architectures. The proposed developer-facing API simplifies integration while maintaining framework agnosticism and forward compatibility with planned transport patterns.
The design shows strong attention to ergonomics, consistency with existing AI SDK patterns, and extensibility for future realtime transport layers.

packages/openai/src/openai-provider.ts
felixarntz: Right, I didn't say it correctly. We don't expose it on `ProviderV4` because it's experimental, so that's already correct, but what I meant is indeed that I'm wondering whether we should _ever_ expose it on `ProviderV4` - in other words, we probably shouldn't expose it on the provider-specific `*Provider` objects either.

+1 to simply exporting as its own thing, at least to start with

gr2m:
We had more discussion about executing server-side tools and decided for now that we will not built that into our APIs.

Instead all tool calls will be local, and the local tool call callbacks can then send a request to the server. We should document that usecase and add an example, but basically not make the security implications our code's responsibility, instead make the user aware how to implement server side tool calling and approaches on how to make it secure.

felixarntz:
> We had more discussion about executing server-side tools and decided for now that we will not built that into our APIs.
> 
> Instead all tool calls will be local, and the local tool call callbacks can then send a request to the server. We should document that usecase and add an example, but basically not make the security implications our code's responsibility, instead make the user aware how to implement server side tool calling and approaches on how to make it secure.

With "server-side tools", are you referring to the part where we expose tool definitions to the client and then execute them via API calls? So basically, we won't provide a mechanism to easily invoke server-side tools from what is effectively always a client-side tool call (since realtime inference is running on the client)?

I think not offering this will lead to notably worse DX, as devs will be required to wire things up themselves. But I get it from a security perspective, it probably isn't something to encourage, especially because the implications may not be clear to everyone.

IMO it'd be worth to still explore building out that mechanism so that it becomes much simpler to call server-side tools and reuse the existing tool definitions. Maybe we can implement it in a slightly different way: Could be its own package, and we could put emphasis on having to opt in every tool definition to this, rather than just allow "dumping" a list of tools without much consideration.

gr2m:
> I think not offering this will lead to notably worse DX

I'd start out with the safe version. It might not be too bad. Also the two approaches don't necessary conflict, we could offer both eventually

ninjapayofficial:
Any update on this? i am currently using openai sdk..but would prefer ai sdk

packages/google/src/realtime/google-realtime-model.ts
```diff
@@ -0,0 +1,122 @@
...
+    const setupPayload = buildGoogleSessionConfig(
+      options.sessionConfig,
+      this.modelId,
+    );
+
+    const response = await fetchFn(
+      `${AUTH_TOKENS_URL}?key=${encodeURIComponent(apiKey)}`,
+      {
+        method: 'POST',
+        headers: { 'Content-Type': 'application/json' },
```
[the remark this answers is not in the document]
gr2m: Good catch — fixed in d3119581ac. The Gemini [ephemeral token docs](https://ai.google.dev/gemini-api/docs/ephemeral-tokens) confirm `uses` defaults to `1` (one Live API session start); `0` permits zero sessions. Changed to `uses: 1` and updated the test assertion.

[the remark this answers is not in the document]
gr2m: Correction: this is a false positive, and my previous reply was wrong. Per the [AuthToken REST reference](https://ai.google.dev/api/live), `uses: 0` means **"no limit is applied"** (unlimited session starts) — an *unset* value is what defaults to `1`. Setting `uses: 1` (my earlier change) would actually cap the token at a single session start and break WebSocket reconnects, which is why Google realtime voice chat works fine with `0`. I reverted to `uses: 0` and added a comment documenting the intent so it isn't re-flagged.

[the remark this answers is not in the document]
gr2m: actually

> If this value is zero then no limit is applied

gonna revert and add comment

packages/ai/src/realtime/audio-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Converts a base64-encoded PCM16 string (from a realtime model's
+ * audio-delta event) back to Float32 audio samples.
+ *
+ * Input is expected to be 16-bit signed integer, little-endian, base64-encoded.
+ * Output samples are in the range [-1.0, 1.0].
+ */
+export function decodeRealtimeAudio(base64Audio: string): Float32Array {
```
felixarntz: shouldn't these follow the same naming convention? maybe `encodeRealtimeAudio` next to `decodeRealtimeAudio`?

packages/ai/src/realtime/browser-realtime-transport.ts
```diff
@@ -0,0 +1,113 @@
...
+} from '../types/realtime-model';
+
+export type BrowserRealtimeTransportOptions = {
+  model: RealtimeModel;
+  onServerEvent: (event: RealtimeServerEvent) => void | Promise<void>;
+  onError: (error: Error) => void;
+  onClose: () => void;
+};
+
+export class BrowserRealtimeTransport {
```
felixarntz, line 15: shouldn't these be `Experimental_` too?

**follow-up:** I think you left these as non-experimental because they're not exported from the package, but that's a bit risky. See my comment on `packages/ai/src/realtime/index.ts` for what I think would be a cleaner and safer way to handle this.

packages/ai/src/realtime/browser-realtime-audio.ts
```diff
@@ -0,0 +1,168 @@
...
+
+export type BrowserRealtimeAudioOptions = {
+  captureSampleRate: number;
+  playbackSampleRate: number;
+  onAudio: (base64Audio: string) => void;
+  onPlayingChange: (isPlaying: boolean) => void;
+  onCapturingChange: (isCapturing: boolean) => void;
+};
+
+export class BrowserRealtimeAudio {
```
felixarntz, line 15: same here, maybe better as `Experimental_`?

Alternatively, see my comment on `packages/ai/src/realtime/index.ts`

packages/provider/src/realtime-model/v4/realtime-model-v4-session-config.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Voice activity detection configuration.
+   * Set to null or type 'disabled' to turn off VAD (push-to-talk mode).
+   */
+  turnDetection?: {
+    /**
+     * VAD mode. 'server-vad' for automatic detection,
+     * 'semantic-vad' for OpenAI's semantic detection,
+     * 'disabled' to turn off VAD.
```
felixarntz, line 110: nit: rather than mentioning a specific provider (OpenAI) here, maybe clarify what the actual difference between "automatic" vs "semantic" is? that isn't clear to me.

packages/ai/src/realtime/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export type {
+  Experimental_RealtimeClientEvent,
+  Experimental_RealtimeFactory,
+  Experimental_RealtimeFactoryGetTokenOptions,
+  Experimental_RealtimeFactoryGetTokenResult,
+  Experimental_RealtimeModel,
+  Experimental_RealtimeServerEvent,
+  Experimental_RealtimeSessionConfig,
+  Experimental_RealtimeToolDefinition,
+} from '../types/realtime-model';
```
felixarntz, line 23: Just an idea: Instead of declaring all the exported symbols as `Experimental_` at their declaration, I'd recommend naming them with their unprefixed name, and only _exporting_ them with the `Experimental_` prefix.

We've used this approach elsewhere in AI SDK before, and it means there are fewer diffs when removing the prefix in the future, and it also makes it less likely a new export that's part of experimental code accidentally gets exported without prefix.

IMO, the cleanest way to handle those prefixes is that they purely exist at the export and import seams; export the regular names under `Experimental_` aliases, and in the consuming packages import the `Experimental_` symbols under non-prefixed aliases. Then all code stays the same except imports and exports.

felixarntz (approved):
Looks solid! A few notes on APIs.

packages/ai/src/realtime/audio-utils.ts
gr2m: Good catch — renamed to `encodeRealtimeAudio` to match `decodeRealtimeAudio`. Done in 5f51a7cd35.

packages/ai/src/realtime/index.ts
gr2m: > I'd recommend naming them with their unprefixed name, and only _exporting_ them with the `Experimental_` prefix.

Thank you for catching that, that is indeed our convention, I'll document it in contributing/

gr2m: Great suggestion — applied in c3bc02fa0b. All realtime symbols are now declared with their unprefixed names; the `Experimental_` prefix only exists as an alias at the public export seam (`packages/ai/src/realtime/index.ts`, `packages/react/src/use-realtime.ts`), and consuming code (the `react` package and the e2e example) imports those symbols under unprefixed aliases. Public API surface is unchanged. I'll also document this convention in `contributing/` in a separate PR.

packages/ai/src/realtime/browser-realtime-transport.ts
gr2m: addressed via https://github.com/vercel/ai/pull/13893/changes/c3bc02fa0be8e4520824f19f2d488b70708b00f0

gr2m:
## Review feedback addressed (924ea4e67d)

Worked through the detailed review. All items were valid; none were previously addressed. Fixed everything except the xAI mapper test fixtures

### Core (`packages/ai`) — with regression tests

- **Manual tool flow false error** — `onToolCall` returning `undefined` no longer fires `"No handler provided"`. That's the documented submit-output-later pattern; the error now only fires when no handler exists.
- **Tool outputs requesting a response too early** — multi-tool turns now send a **single** `response-create`, gated on the tool-bearing response's `response-done` **and** every tool output being submitted (robust against tool calls arriving as separate events, which a naive "send when pending is empty" approach would mishandle).
- **Disconnect leaking a connecting socket** — the socket is now tracked immediately, so `disconnect()` mid-connect actually closes it, and a late `open` no longer fires the session-update callback.

### Providers — with tests

- **OpenAI**: send the required `expires_after.anchor: "created_at"` with a custom token TTL.
- **xAI**: include the model as a `?model=` query param on the WebSocket URL (xAI selects the voice model from it); drop `conversation.item.truncate`, which is unsupported over WebSocket (barge-in still works via local playback stop on `speech_started`).
- **Google**: label input audio with the configured capture rate instead of a hardcoded `16000`; map `expiresAfterSeconds` to `newSessionExpireTime` (the session-open window); keep a late transcript attached to the turn it belongs to (lazy turn rollover); honor a new `outputAudioTranscription` session config.

### Spec

- Added optional `outputAudioTranscription` to the experimental realtime session config, symmetric with `inputAudioTranscription`.

### Example

- xAI realtime now uses `grok-voice-latest` and the built-in voices `ara` / `eve` / `leo` / `rex` / `sal`, verified against the [xAI Voice Agent docs](https://docs.x.ai/developers/model-capabilities/audio/voice-agent). PR description updated to match.

### Notes / overlaps

- The xAI `conversation.item.truncate` removal overlaps the open VAD-scope discussion — the shared barge-in handler still sends `conversation.item.truncate` for providers that support it (OpenAI), and it's now skipped at the xAI serializer seam.
- I could not run a **live** xAI call here (no API key); verification is doc-based plus a unit test asserting the exact WebSocket URL.
- **Not addressed (by design):** xAI realtime mapper test fixtures from captured live payloads — @felixarntz offered to take these in a follow-up PR.

PR #13908, opened 2026-03-30
fix(google): propagate thoughtSignature to all parallel functionCall parts

edwardwc:
## Summary

- When Gemini thinking models issue parallel tool calls, only the **first** `functionCall` part receives a `thoughtSignature` from the model. However, the Gemini API requires **every** `functionCall` part in the same turn to carry the signature when the conversation is replayed. Without this, subsequent requests fail with HTTP 400 or silently drop data.
- This PR adds a `propagateThoughtSignatureToParallelFunctionCalls` helper that runs after building the assistant `parts` array, copying the first observed `thoughtSignature` to any `functionCall` part that is missing one.
- Existing signatures are preserved; parts that already have a signature are not overwritten.

## Reproduction

1. Use `gemini-2.5-flash-preview-04-17` (or any Gemini thinking model) with `maxSteps > 1` and multiple tool definitions
2. The model issues parallel tool calls in a single turn
3. On the next step, the SDK replays the conversation — the second (and subsequent) `functionCall` parts are missing `thoughtSignature`
4. The Gemini API returns a 400 error, or in some cases silently drops parts

## Context

Tracked upstream at:
https://discuss.ai.google.dev/t/gemini-3-flash-preview-inconsistent-thought-signature-generation-in-parallel-function-calls-causes-400-errors-and-potential-silent-data-loss/118936

## Test plan

- [x] Updated existing parallel tool call test to verify propagation
- [x] Added tests for: 3+ parallel calls, single call (no-op), no signatures present (no-op), mixed existing signatures
- [x] All 36 tests in `convert-to-google-generative-ai-messages.test.ts` pass


edwardwc:
Hey — wanted to add some context on why this fix matters in practice.

We hit this in production running a multi-step agentic loop with Gemini 2.5 Flash (thinking mode enabled). The agent has several tools available and regularly issues 2-3 parallel tool calls per turn. Everything works fine on the first step, but on the second `generateText` call — when the SDK replays the conversation history — we get intermittent 400 errors from the Gemini API.

The root cause: when the model produces parallel `functionCall` parts, it only attaches `thoughtSignature` to the first one. The remaining parts have no signature. But on replay, the API enforces that every `functionCall` part must carry its `thoughtSignature`. So the second step fails because the SDK faithfully sends back exactly what the model gave it — which is incomplete.

The failure mode is especially nasty because:
- It only happens with parallel tool calls, so it's non-deterministic (depends on the model's decision to parallelize)
- In some API versions, instead of a clean 400, the unsigned parts are silently dropped — the model just "forgets" those tool results and either hallucinates or enters a retry loop
- It affects all Gemini thinking models, not just Flash

Google has acknowledged the inconsistency on their side (linked in the PR description), but there's no timeline for a model-level fix. The SDK is the right place to handle this since it already transforms the conversation for the wire format — one small post-processing step after building the parts array makes every thinking-model tool-use flow reliable.

The fix is intentionally conservative: it only activates when there are 2+ `functionCall` parts and at least one has a signature, it never overwrites existing signatures, and it's a no-op for non-thinking models.

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -235,91 +279,92 @@ export function convertToGoogleGenerativeAIMessages(
...
+                    args: part.input,
+                  },
+                  thoughtSignature,
+                };
+              }
+            }
+          })
+          .filter(part => part !== undefined);
+
+        propagateThoughtSignatureToParallelFunctionCalls(modelParts);
```
felixarntz: While this works, it moves a lot of the code around. We'll have to backport this, so I'd prefer to keep the changes simpler, to reduce the chance of broader merge conflicts.

Doing it inline will also keep the logic more efficient, as we don't need to iterate over any loops multiple times.

felixarntz (commented):
@edwardwc Thank you for the PR!

I think the fix looks good, but I'm unable to replicate the underlying problem. I'm pushing some examples for testing. Maybe you can update them so that we can replicate the bug?

packages/google/src/convert-to-google-generative-ai-messages.ts
felixarntz: https://github.com/vercel/ai/pull/13908/changes/ae7db6dea57fbef96b76a796580511a493c21a7a implements the simpler solution.

packages/google/src/convert-to-google-generative-ai-messages.ts
```diff
@@ -308,12 +310,20 @@ export function convertToGoogleGenerativeAIMessages(
...
+                  ) {
+                    functionCallSignature = thoughtSignature;
+                  }
                   return {
                     functionCall: {
                       name: part.toolName,
                       args: part.input,
                     },
-                    thoughtSignature,
+                    thoughtSignature: thoughtSignature ?? functionCallSignature,
```
felixarntz, line 391: If I revert this line to its original, I would expect the examples I added to the branch to fail. But they still work. I don't see any Google API errors, or dropped parts.

It would be great if you could look into them further, so we can actually replicate the problem you're describing in the examples.

shtefcs:
We're hitting this exact issue in production with Gemini 3 Flash + multi-step browser automation (10-35 tool call steps per session).

Our setup uses the Vercel AI Gateway with `order: ['google', 'vertex']` for provider failover. The failure pattern we see:

1. Steps 1-10 complete fine on Google AI Studio
2. Google returns 503 on step 11
3. Gateway fails over to Vertex
4. Vertex rejects with 400: "Thought signature is not valid"

We initially assumed this was purely a cross-provider signature incompatibility, but after digging into the SDK tests (which confirm Google/Vertex signatures ARE cross-compatible), we think the parallel `functionCall` signature gap from this PR is at least part of the problem. If the model issued parallel tool calls during one of the earlier steps and only the first got a signature, then when the conversation is replayed on Vertex, those unsigned parts cause the rejection.

We're currently working around this with \`only: ['google']\` (blocking Vertex entirely) + a model-level fallback to Kimi K2.5. But this means we lose Vertex as a provider fallback for 503s, which hurts reliability.

Would love to see this merged so we can safely enable both providers. The fix looks solid and conservative. Happy to help test if that would speed things up.

felixarntz:
@shtefcs Thank you for the additional context. So you're saying this is only a problem for Vertex, but not Google's Gemini API? That would explain why I wasn't able to replicate it with the examples in this PR.

packages/google/src/convert-to-google-generative-ai-messages.ts
felixarntz: @edwardwc @shtefcs I still can't replicate the bug. I've updated the examples in this PR to use Vertex instead of Gemini API, but they still always succeed for me, regardless of whether I rely on `packages/google` with this change here applied or not.

shtefcs:
Hey, I need to correct what I said earlier. After looking at this more carefully, our failure case is different from what this PR fixes.

Our actual error is "Thought signature is not valid" (400 from Vertex). This happens when we use the Vercel AI Gateway with order: ['google', 'vertex']. Google handles steps 1-10 fine, then returns 503 on step 11. The Gateway falls over to Vertex for step 11, and Vertex rejects the request with that 400 error.

When we pin to Google only (only: ['google']), the problem disappears completely. So Vertex is definitely the trigger.

What I don't know is whether Vertex rejects the signatures because they were generated by Google AI Studio (cross-provider incompatibility) or because some parallel tool call parts are missing signatures (the issue this PR addresses). The error message from Google doesn't distinguish between the two.

Either way, our specific case is about Gateway provider failover from Google to Vertex mid-conversation, not strictly about parallel tool calls. Sorry for the confusion in my earlier comment.

For reproducing the parallel call issue: the model has to decide on its own to parallelize. A prompt like "check the weather in London, Paris, and Tokyo" with 3 tool definitions and thinking mode enabled might force it. The tricky part is it's non-deterministic.

shtefcs:
I created a separate issue for our specific problem since it turned out to be a different root cause: #14196

Our issue is about Vertex rejecting thought signatures that were generated by Google AI Studio during Gateway provider failover. That's separate from the parallel tool call propagation this PR addresses.

Linking them here in case there's overlap or shared context that helps.

PR #13913, opened 2026-03-30
fix(ai): skip null/undefined tool entries in prepareToolsAndToolChoice (fixes #13906)

### LINKED ISSUE #13906: Anthropic web search tool (webSearch_20260209) crashes toUIMessageStreamResponse with TypeError
## Description

Using Anthropic's native web search tool (`anthropic.tools.webSearch_20260209()`) with `streamText` + `toUIMessageStreamResponse()` causes a server-side crash:

```
TypeError: Cannot read properties of undefined (reading 'type')
```

The error occurs during stream serialization — the `server_tool_use` and `web_search_tool_result` content blocks from Anthropic's API can't be mapped to UIMessage parts.

## Reproduction

```typescript

const result = streamText({
  model: anthropic("claude-sonnet-4-6"),
  messages: [{ role: "user", content: "What is the weather in NYC?" }],
  tools: {
    ...myCustomTools,
    ...anthropic.tools.webSearch_20260209({ maxUses: 3 }),
  },
});

// This crashes:

// Note: toDataStreamResponse() was not tested
```

## Environment

- `ai`: 6.0.141
- `@ai-sdk/anthropic`: 3.0.64
- `@ai-sdk/react`: 3.0.143
- Next.js 16.1.7

## Expected behavior

Web search results should be serialized into UIMessage parts (perhaps as a tool-call/tool-result pair, or as a text part with citations).

## Actual behavior

Server throws `TypeError: Cannot read properties of undefined (reading 'type')` during stream processing. The POST returns 200 but the client receives no usable response.

## Workaround

Using a custom web search tool (e.g., DuckDuckGo scraping) instead of Anthropic's native server tool.

## Context

Anthropic's web search returns non-standard content blocks (`server_tool_use`, `web_search_tool_result`) that `toUIMessageStreamResponse` doesn't know how to transform into UIMessage parts. The same web search tool works fine when using the Anthropic SDK directly with non-streaming `messages.create()` and returning plain JSON.

shivamtiwari3:
## Summary

Fixes #13906.

`prepareToolsAndToolChoice` now skips null/undefined tool values instead of crashing with `TypeError: Cannot read properties of undefined (reading 'type')`.

---

## Root Cause

In `prepare-tools-and-tool-choice.ts:44`, the loop iterates `Object.entries(tools)` and accesses `tool.type` unconditionally:

```typescript
for (const [name, tool] of filteredTools) {
  const toolType = tool.type;  // crashes when tool is undefined
```

This crashes when a user accidentally **spreads** a provider tool factory result into the `tools` record instead of assigning it under a key:

```typescript
// ❌ Incorrect (issue reproducer)
tools: {
  ...anthropic.tools.webSearch_20260209({ maxUses: 3 }),
}

// ✅ Correct
tools: {
  web_search: anthropic.tools.webSearch_20260209({ maxUses: 3 }),
}
```

The factory always produces a `Tool` object with `execute: undefined`, `needsApproval: undefined`, `toModelOutput: undefined`, etc. Spreading this object into `tools` creates entries like `['execute', undefined]`. When `prepareToolsAndToolChoice` iterates these entries, it hits `undefined.type` and throws.

Because `prepareToolsAndToolChoice` is called lazily during stream processing, the HTTP response is already 200 when the crash occurs — the client receives no usable body.

## Solution

Guard against null/undefined tool values with an early `continue` in the loop:

```typescript
for (const [name, tool] of filteredTools) {
  if (tool == null) {
    continue;
  }
  const toolType = tool.type;
```

## Testing

- Added `should skip null/undefined tool entries without crashing` in `prepare-tools-and-tool-choice.test.ts` that directly reproduces the crash scenario.
- All 2176 existing tests continue to pass.
- Run with: `cd packages/ai && pnpm test:node`

### DIFF
--- a/.changeset/fix-null-tool-entries-crash.md
+++ b/.changeset/fix-null-tool-entries-crash.md
@@ -1,0 +1,14 @@
+---
+'ai': patch
+---
+
+Fix crash when tools object contains null/undefined entries
+
+`prepareToolsAndToolChoice` now skips null/undefined tool values instead of
+throwing `TypeError: Cannot read properties of undefined (reading 'type')`.
+
+This crash occurred when a provider tool factory result was accidentally spread
+into the `tools` parameter (e.g., `tools: { ...anthropic.tools.webSearch_20260209({}) }`).
+The factory returns a `Tool` object whose properties include `execute: undefined`,
+`needsApproval: undefined`, etc. Spreading these into `tools` populated the record
+with `undefined` values, causing the crash during stream processing.
--- a/packages/ai/src/prompt/prepare-tools-and-tool-choice.test.ts
+++ b/packages/ai/src/prompt/prepare-tools-and-tool-choice.test.ts
@@ -407,6 +407,32 @@ describe('prepareToolsAndToolChoice', () => {
     `);
   });
 
+  it('should skip null/undefined tool entries without crashing', async () => {
+    // Simulates accidentally spreading a Tool object into the tools record:
+    //   tools: { ...anthropic.tools.webSearch_20260209({ maxUses: 3 }) }
+    // The factory creates a Tool with execute: undefined, needsApproval: undefined, etc.
+    // These undefined values become spurious tool entries when spread.
+    const result = await prepareToolsAndToolChoice({
+      tools: {
+        validTool: tool({
+          description: 'A valid tool',
+          inputSchema: z.object({}),
+        }),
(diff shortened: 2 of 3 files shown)


aayush-kapoor (commented):
not needed - marking as closed.

see comment here https://github.com/vercel/ai/issues/13906#issuecomment-4180555140

PR #13920, opened 2026-03-30
feat(provider/google): support combining built-in tools with function calling on Gemini 3

edwardwc:
## Summary

Gemini 3 models now support combining built-in tools (Google Search, Google Maps, URL Context, etc.) with custom function calling in a single API request via the [tool combination API](https://ai.google.dev/gemini-api/docs/tool-combination) (announced March 17, 2026).

Previously, `prepareTools` treated provider-defined tools and function tools as mutually exclusive — when both were present, it emitted an "unsupported" warning and dropped all function declarations, sending only the provider tools. This caused models to emit function calls as inline JSON text since no function declarations were registered.

This PR enables tool combination for Gemini 3+ models by:

- Merging provider tools and `functionDeclarations` into the same `tools` array when both are present on Gemini 3+ models
- Setting `includeServerSideToolInvocations: true` in `toolConfig` to enable server-side tool context circulation
- Using `VALIDATED` function calling mode (required by the tool combination API)
- Parsing `toolCall`/`toolResponse` parts from the API response as provider-executed tool calls/results
- Serializing them back correctly in subsequent turns to maintain cross-tool context circulation

For older models (Gemini 2.x and below), the existing behavior is preserved — a warning is emitted and only provider tools are sent.

## Test plan

- [x] Existing `prepareTools` tests updated for new merge behavior on Gemini 3+
- [x] New tests for combined tool output: `[{ googleSearch: {} }, { functionDeclarations: [...] }]`
- [x] Verify older models still get the warning and provider-only behavior
- [ ] Manual test: Google Search + custom function calling on `gemini-3-flash-preview`
- [ ] Verify `toolCall`/`toolResponse` parts are preserved across multi-turn conversations


felixarntz (approved):
@edwardwc Thank you for the PR! This was looking good for starters, however we needed some examples to verify the behavior, so I added them. From there, I ran into a few follow up issues, which should be addressed now.

ghost:
:rocket: Published in:

| Package | Version |
| --- | --- |
| `@ai-sdk/google` | [`4.0.0-beta.19`](https://www.npmjs.com/package/@ai-sdk/google/v/4.0.0-beta.19) |
| `@ai-sdk/google-vertex` | [`5.0.0-beta.26`](https://www.npmjs.com/package/@ai-sdk/google-vertex/v/5.0.0-beta.26) |

Issue #5129, opened 2025-03-11
Bedrock request metadata is undefined in streamText() OnFinishResult and Returns


Tian-Hun:
### Description

When using `streamText()`, the request metadata is undefined in both the `onFinish` callback's result and the function's return value. This makes it impossible to access important request information after the stream completes.

### Code example

```typescript
export async function POST(req: Request) {
  const result = streamText({
    model: bedrock('us.anthropic.claude-3-7-sonnet-20250219-v1:0'),
    messages: [{ role: 'user', content: 'Hello' }],
    onFinish: ({ request }) => {
      console.log('onFinish request metadata:', request.body); // undefined
    },
  });
  
  result.request.then((request) => {
    console.log('result request metadata:', request.body); // undefined
  });
  
  return result.toDataStreamResponse();
}
```

### AI provider

@ai-sdk/amazon-bedrock  2.1.1

charllie:
Use the inference profile id instead (replace `anthropic.claude-3-7-sonnet-20250219-v1:0` by `us.anthropic.claude-3-7-sonnet-20250219-v1:0`)

aayush-kapoor:
Marking this as closed for now - if you face any regression or have a similar problem after migrating to the latest AI SDK version, please feel free to open a new issue (with possible reproduction)

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **Medium**

### Rationale
This describes a concrete failure to expose request metadata through existing `streamText()` result APIs, so it is a bug report rather than a request for new capability.

### Evidence
- `streamText()` returns `undefined` for `request.body` both through `onFinish` and `result.request`.
- The report provides a concrete Bedrock model, package version, and code demonstrating the failure.
- The issue was later closed pending reproduction on the latest SDK, but no comment establishes that the reported behavior was expected.

PR #13928, opened 2026-03-30
feat(amazon-bedrock): return request body from doGenerate and doStream

dontgitit:
## Summary

`BedrockChatLanguageModel.doGenerate()` and `doStream()` do not return a `request` field, so consumers that rely on the request body (observability frameworks, `onFinish` callbacks, etc.) receive `undefined`. There was an existing `// TODO request?` comment at the `doStream` return site.

This PR adds `request: { body: args }` to both methods, matching the pattern used by `@ai-sdk/anthropic`:

```diff
 // doGenerate return
 return {
   content,
   finishReason: { ... },
   usage: convertBedrockUsage(response.usage),
+  request: { body: args },
   response: { ... },
   warnings,
 };

 // doStream return
 return {
   stream: response.pipeThrough(...),
-  // TODO request?
+  request: { body: args },
   response: { headers: responseHeaders },
 };
```

`args` is the Bedrock Converse API command object (the same value passed as `body` to `postJsonToApi`), so it captures the full request payload including model config, messages, tools, and inference parameters.

## Motivation

Observability frameworks like [Mastra](https://github.com/mastra-ai/mastra) use the `request` field from provider `doStream()`/`doGenerate()` to populate tracing span inputs (e.g., `model_step.input`). With the Bedrock provider, this is always `undefined`, resulting in empty span attributes — making it impossible to see what was sent to the LLM.

The `@ai-sdk/anthropic` provider already returns `request: { body: args }` from both methods ([ref](https://github.com/vercel/ai/blob/main/packages/anthropic/src/anthropic-messages-language-model.ts#L1212)).

Fixes #13927
Related: #5129

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-chat-language-model.ts
+++ b/packages/amazon-bedrock/src/bedrock-chat-language-model.ts
@@ -566,6 +566,7 @@ export class BedrockChatLanguageModel implements LanguageModelV4 {
         raw: response.stopReason ?? undefined,
       },
       usage: convertBedrockUsage(response.usage),
+      request: { body: args },
       response: {
         id: responseHeaders?.['x-amzn-requestid'] ?? undefined,
         timestamp:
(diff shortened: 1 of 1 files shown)


aayush-kapoor:

PR #14001, opened 2026-04-01
fix(mcp): strip trailing slash from OAuth resource parameter

### LINKED ISSUE #13988: OAuth resource parameter gets trailing slash from URL.href, breaking auth servers that do exact matching
### Description

When the MCP server's Protected Resource Metadata returns a pathless resource URL (e.g., https://mcp.example.com), the resource parameter sent to the authorization server gets an unwanted trailing slash (https://mcp.example.com/). This is because `new URL("https://mcp.example.com").href` returns "https://mcp.example.com/" per the URL spec.

Auth servers that do exact matching on the resource parameter reject the request because the trailing slash makes it a different URI than what was registered.

**Reproduction:**

Set up an MCP server whose /.well-known/oauth-protected-resource returns a pathless resource URL:
```json
{
  "resource": "https://mcp.example.com",
  "authorization_servers": ["https://auth.example.com"]
}
```
Attempt to connect — the OAuth token exchange fails because the resource parameter is sent as https://mcp.example.com/ instead of https://mcp.example.com

**Root cause:**

resource.href is used in three places, all of which add the trailing slash:

- `startAuthorization`
- `exchangeAuthorization`
- `refreshAuthorization`

The MCP spec itself recommends the form without the trailing slash (https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri):

> implementations SHOULD consistently use the form without the trailing slash for better interoperability

**Related issues:**                                                                                                                                                                                          
- modelcontextprotocol/inspector#927 

### AI SDK Version

@ai-sdk/mcp: 1.0.30

### Code of Conduct

venkat22022202:
## Summary

Fixes #13988

`new URL("https://mcp.example.com").href` returns `"https://mcp.example.com/"` per the URL spec. Auth servers that do exact matching on the `resource` parameter reject the request because the trailing slash makes it a different URI than what was registered.

## Fix

Strip trailing slash from `resource.href` in all three locations where it's used as an OAuth parameter:

- `startAuthorization` (line 454)
- `exchangeAuthorization` (line 678)
- `refreshAuthorization` (line 765)

```diff
- params.set('resource', resource.href);
+ params.set('resource', resource.href.replace(/\/$/, ''));
```

## Why this is correct

The MCP spec explicitly recommends the form without trailing slash for better interoperability:

> implementations SHOULD consistently use the form without the trailing slash

Source: https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#canonical-server-uri

## Test plan

- MCP servers with pathless resource URLs (e.g. `https://mcp.example.com`) should no longer get rejected by auth servers doing exact match
- MCP servers with path-based resource URLs (e.g. `https://mcp.example.com/api`) are unaffected (no trailing slash to strip)

### DIFF
--- a/packages/mcp/src/tool/oauth.ts
+++ b/packages/mcp/src/tool/oauth.ts
@@ -451,7 +451,7 @@ export async function startAuthorization(
   }
 
   if (resource) {
-    authorizationUrl.searchParams.set('resource', resource.href);
+    authorizationUrl.searchParams.set('resource', resource.href.replace(/\/$/, ''));
   }
 
   return { authorizationUrl, codeVerifier };
@@ -675,7 +675,7 @@ export async function exchangeAuthorization(
   }
 
   if (resource) {
-    params.set('resource', resource.href);
+    params.set('resource', resource.href.replace(/\/$/, ''));
   }
 
   const response = await (fetchFn ?? fetch)(tokenUrl, {
@@ -762,7 +762,7 @@ export async function refreshAuthorization(
   }
 
   if (resource) {
-    params.set('resource', resource.href);
+    params.set('resource', resource.href.replace(/\/$/, ''));
   }
 
   const response = await (fetchFn ?? fetch)(tokenUrl, {
--- a/packages/mcp/src/tool/oauth.ts
+++ b/packages/mcp/src/tool/oauth.ts
@@ -21,6 +21,24 @@ import {
   InvalidGrantError,
   UnauthorizedClientError,
 } from '../error/oauth-error';
+
+/**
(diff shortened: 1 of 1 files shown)


venkat22022202:
Good catch — fixed in 8206bfa.

Extracted a `canonicalResourceUri()` helper that only strips the trailing slash when `resource.pathname === '/'` (origin-only URL like `https://mcp.example.com/`). Path-based resource URIs like `https://mcp.example.com/api/` are now left unchanged.

lgrammel:
Closing this PR because issue #13988 has already been closed.

PR #14004, opened 2026-04-01
fix(errors): preserve error type prefix via toString in getErrorMessage

nehaaprasad:
### summary

- Use error.toString() in getErrorMessage() (provider + provider-utils) so outputs keep prefixes like Error: / TypeError: / AbortError:.
- Updated affected tests and added a patch changeset.

### Manual Verification

- Ran tests and confirmed tool/abort error text now includes the error type prefix.

Fixes #14002


aayush-kapoor (commented):
closing in favor of https://github.com/vercel/ai/pull/14019

PR #14006, opened 2026-04-01
fix(amazon-bedrock): use template literal in tool warning message

Ricardo-M-L:
In `bedrock-prepare-tools.ts` line 118, a regular string `'tool ${tool.id}'` is used instead of a template literal, so the warning message outputs the literal text `tool ${tool.id}` rather than the actual tool ID.

Line 124 in the same file already uses the correct template literal `` `tool ${tool.id}` `` for the same pattern.

One character fix: `'` → `` ` ``.

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-prepare-tools.test.ts
+++ b/packages/amazon-bedrock/src/bedrock-prepare-tools.test.ts
@@ -174,6 +174,25 @@ describe('prepareTools', () => {
       ]);
     });
 
+    it('should warn with actual tool id for unrecognized provider-defined tools on anthropic models', async () => {
+      const result = await prepareTools({
+        tools: [
+          {
+            type: 'provider',
+            id: 'anthropic.unknown_tool',
+            name: 'unknown_tool',
+            args: {},
+          },
+        ],
+        modelId: ANTHROPIC_MODEL,
+      });
+
+      expect(result.toolWarnings).toContainEqual({
+        type: 'unsupported',
+        feature: 'tool anthropic.unknown_tool',
+      });
+    });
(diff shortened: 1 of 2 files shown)


JiwaniZakir (commented):
— in `bedrock-prepare-tools.ts`, the original string `'tool ${tool.id}'` was using single quotes instead of backticks, meaning `tool.id` was never interpolated and every warning would have read the literal text `"tool ${tool.id}"`, making it impossible to identify which tool triggered the warning. The fix correctly switches to a template literal so the actual tool identifier appears in the message.

It's worth adding a unit test in the existing test suite for `prepareTools` that covers the `else` branch (the unsupported tool path) and asserts that the returned warning's `feature` field contains the real `tool.id` value rather than a literal placeholder string — this kind of regression would have been caught immediately with such a test. Additionally, it may be worth auditing the rest of the file (and sibling files like `bedrock-prepare-tools.ts`'s callers) for any similar string/template literal mixups, since this pattern can be easy to miss in review.

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open since April. Would appreciate a review. Thanks!

Ricardo-M-L:
Friendly bump — this is a one-line template literal fix for the amazon-bedrock warning message that's been sitting for ~4 months. Tiny change, zero risk. Would love to get this merged!

Ricardo-M-L:
Closing duplicate: superseded by #18954 (bedrock template literal)

PR #14019, opened 2026-04-01
fix(provider): preserve error type prefix in getErrorMessage

murataslan1:
## Summary

- Changed `error.message` to `error.toString()` in both copies of `getErrorMessage()`
- This preserves the error type prefix (`Error:`, `TypeError:`, `RangeError:`, etc.) so the model can distinguish a failed tool call from a normal string result
- Added comprehensive test suite covering all branches

## Breaking change

This is a behavior change for all callers of `getErrorMessage`. The output for `Error` instances now includes the type prefix:

```ts
const error = new TypeError("API crashed");

// Before: "API crashed"
error.message

// After: "TypeError: API crashed"
error.toString()
```

Downstream consumers that embed this string in user-facing messages, structured logs, or serialized API responses will see the prefix. This is intentional — the previous behavior made tool errors indistinguishable from successful string results when forwarded to the model.

## Edge case: empty error message

When `error.message` is an empty string, `error.toString()` returns just the error name (e.g. `"Error"` or `"TypeError"`), whereas the previous behavior returned `""`. This is a net improvement — an empty string gives no signal at all, while `"Error"` at least tells the model something went wrong.

## Note on file duplication

Both `packages/provider/src/errors/get-error-message.ts` and `packages/provider-utils/src/get-error-message.ts` are identical copies. Consolidating them into a single shared module would be a good follow-up but is out of scope for this fix.

## Changes

- `packages/provider/src/errors/get-error-message.ts`
- `packages/provider-utils/src/get-error-message.ts`
- `packages/provider-utils/src/get-error-message.test.ts` (new)

## Test plan

- [x] 15 tests covering: null/undefined, strings, Error/TypeError/RangeError, empty messages, custom error subclasses, plain objects/numbers/booleans/arrays
- [x] All pass (`pnpm test:node` in provider-utils — 386/386)

Closes #14002


JiwaniZakir (commented):
The change from `error.message` to `error.toString()` in both `get-error-message.ts` files will now prepend the error type name (e.g., `"TypeError: ..."` instead of `"..."`), which is a behavior change for all callers of `getErrorMessage`. This could silently break downstream consumers that embed this string in user-facing messages, structured logs, or serialized API responses — worth calling out explicitly as a breaking change in the PR description.

One edge case worth testing: when `error.message` is an empty string, `error.toString()` returns just `"Error"` (the name alone), whereas the previous behavior returned `""`. Depending on how callers handle that, one of these may be more or less desirable than the other.

It's also worth noting that both files (`packages/provider-utils/src/get-error-message.ts` and `packages/provider/src/errors/get-error-message.ts`) appear to be identical copies of each other — this duplication is a maintenance hazard and it would be worth consolidating them into a single shared module rather than keeping them in sync manually.

No test file changes are included in the diff; tests covering the `instanceof Error` branch with named error types (e.g., `TypeError`, `RangeError`), empty messages, and custom error subclasses would help validate the intended behavior.

felixarntz (approved):
@murataslan1 Thank you for the PR, great catch, and the fix looks good.

PR #14046, opened 2026-04-02
fix(ui): restore partial tool state on stream resume

### LINKED ISSUE #14027: Resume of partial static tool calls crashes in createStreamingUIMessageState
### Description

There appears to be a bug in the UI resume path for in-progress static tool calls.

When `useChat` / `Chat.resumeStream()` resumes from a partially-hydrated assistant message that already contains a static tool part in `input-streaming` state, AI SDK crashes while rebuilding the streaming state.

The error we see is:

```txt
TypeError: Cannot read properties of undefined (reading 'split')
    at getStaticToolName (.../ui-messages.ts)
    at createStreamingUIMessageState (.../process-ui-message-stream.ts)
```

The problematic code in `process-ui-message-stream.ts` is in `createStreamingUIMessageState(...)`:

```ts
toolName:
  part.type === 'dynamic-tool'
    ? String((typedPart.toolName as string | undefined) ?? '')
    : getStaticToolName(part.type),
```

`getStaticToolName` expects the full tool part, not `part.type`. Passing `part.type` causes the function to do `part.type.split(...)` on a string argument, which becomes `undefined.split(...)` internally.

Changing that line to this fixes the issue for us:

```ts
: getStaticToolName(part),
```

### Why this happens

This only seems to happen on resume/hydration flows, because `createStreamingUIMessageState` is rebuilding `partialToolCalls` from an existing last assistant message that already contains an unfinished tool part.

In our case the last assistant message contained a static tool call with:

- `type: "tool-create_document"`
- `state: "input-streaming"`
- `toolCallId: "..."`
- partial `rawInput`

A fresh stream does not hit this path in the same way, but resume does.

### Minimal Repro

1. Start a streaming response that begins a static tool call.
2. Let the message contain a tool part in `input-streaming` state.
3. Persist/hydrate that partial assistant message.
4. Call `resumeStream()` / resume the UI message stream.
5. `createStreamingUIMessageState` crashes before the resumed chunks can be processed.

### Expected behavior

The SDK should successfully reconstruct partial static tool calls during resume.

### Actual behavior

Resume fails with:

```txt
TypeError: Cannot read properties of undefined (reading 'split')
```

### AI SDK Version

- `ai`: `6.0.142`
- also reproducible in both frontend and backend code paths that use the same UI resume state builder

### Additional context

This seems related to the broader class of resume-state reconstruction bugs, for example #13160, but this issue is specifically about partial static tool call reconstruction.

### Code of Conduct

nehaaprasad:
### Background

- Resume crashed for partial static tool calls.

### Summary

- Fix resume state rebuild for input-streaming tool parts.
- Fix static tool name extraction.
- Add regression tests and patch changeset.

### Related Issues
Fixes #14027

### DIFF
--- a/.changeset/resume-static-tool-partial-input.md
+++ b/.changeset/resume-static-tool-partial-input.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ui): restore partial static tool call state when resuming a UI message stream
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -8686,4 +8686,62 @@ describe('processUIMessageStream', () => {
       ]);
     });
   });
+
+  describe('createStreamingUIMessageState resume', () => {
+    it('should rebuild partialToolCalls for static tool parts in input-streaming state', () => {
+      const lastMessage: UIMessage = {
+        id: 'msg-resume',
+        role: 'assistant',
+        parts: [
+          {
+            type: 'tool-create_document',
+            toolCallId: 'tc-1',
+            state: 'input-streaming',
+            input: { title: 'partial' },
+          },
+        ],
+      };
+
+      const state = createStreamingUIMessageState({
+        lastMessage,
+        messageId: 'msg-resume',
+      });
+
+      expect(state.partialToolCalls['tc-1']).toEqual({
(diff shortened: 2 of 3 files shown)


prath47:
Hi @nehaaprasad 
Apologies for putting this here 
Accept my connection request on linked

gr2m:
Closing in favor of #16319 with broader regression coverage. Thanks!

PR #13102, opened 2026-03-05
feat(ai): change type of experimental_context from unknown to generic

lgrammel:
## Background

`experimental_context` for tools and step preparation was introduced in an earlier release without typing (type: `unknown`). Before marking context as stable, it should support smart generic typing for usage in tools, prepareStep, and telemetry.

## Summary

Introduce generics on `experimental_context`. Each tool has a `contextSchema` for the tool-specific context. The context on the AI function call is a union of the tool contexts and can have additional properties for `prepareStep` and telemetry.

## Limitations

The `experimental_context` setting is optional even when a context is needed, i.e. it will not throw type validation errors unless a context is specified.

## Manual Verification

- [x] run and check types on `examples/ai-functions/src/generate-text/openai/tool-call-with-context.ts`
- [x] run and check types on `examples/ai-functions/src/stream-text/openai/tool-call-with-context.ts`
- [x] run and check types on `examples/ai-functions/src/agent/openai/generate-context.ts`
- [x] run and check types on `examples/ai-functions/src/agent/openai/generate-context-call-options.ts`

## Tasks

- [x] prototype generic context and tool context merging
- [x] investigate type errors
- [x] implement generateText
- [x] implement streamText
- [x] implement agent / tool loop agent
- [x] figure out support in provider-defined tools
- [x] add type tests
- [x] add jsdoc
- [x] update reference documentation
- [x] update context documentation
- [x] update examples in docs that use context
- [x] update migration guide
- [x] remove prototypes
- [x] better name for ExpandedContext
- [x] self-review pr
- [x] changeset

## Future Work

* mark `experimental_context` stable
* add `sanitizeContext` function
* add `tool` context wrapper for name mapping and safety
* add `context` to `StopCondition` parameters
* move `ToolSet` into `provider-utils` package
* investigate how to fix context optionality limitation
* remove `ToolCallOptions` type

PR #14050, opened 2026-04-02
chore: rename experimental_context to context

lgrammel:
## Background

Generic typing was added to `experimental_context` in #13102 . With this feature in place, `context` is now stable and can be renamed.

## Summary

* rename `experimental_context` to `context`

## Tasks

- [x] rename
- [x] fix tests and type errors
- [x] changeset
- [x] docs update (rm experimental warnings)
- [x] migration guide

## Future Work
- type context in `core-events`

## Related Issues

Builds on #13102


.changeset/stale-rats-stare.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ai-sdk/provider-utils": major
+"ai": major
```
felixarntz, line 3: what will this actually do in practice? I had read before somewhere that we should always use `patch`, because even with 7.0 we're technically releasing beta versions for every single change, so by themselves they're not a major release.

Just wanting to better understand our process around this. I was assuming if you put `major` here, the new version will automatically become the new major (which we wouldn't want for a single major feature because we always release immediately).

felixarntz (approved):
LGTM, but a question on the changeset.

.changeset/stale-rats-stare.md
lgrammel: We are currently in pre-release mode for the 7.0 major, so this is only to mark the PRs which have breaking changes when the full release changeset is compiled later. There will be no impact on versioning.

lgrammel: See https://github.com/changesets/changesets/blob/main/docs/prereleases.md for prereleases (the page is a bit hard to read tho)

felixarntz: Hmm, in that case I should have added `major` on a few changesets where I didn't. Is there a way to fix it after the fact?

lgrammel: you can just update the changesets

Issue #5576, opened 2025-04-07
[Bug]: Empty Text Content Blocks Cause Errors with Anthropic API in Tool Calling


aditya-dange-cloudredux:
### Description

The Vercel AI SDK generates assistant messages with empty text content blocks (e.g., `{"type": "text", "text": ""}`) when using tool calling with the Anthropic API (e.g., `claude-3-5-sonnet-latest`). This triggers the error `[AI_APICallError: messages: text content blocks must be non-empty]`, breaking chat functionality. The Anthropic API requires non-empty text blocks, unlike other models.

## Steps to Reproduce

1.  Set up a chat stream using the Vercel AI SDK and configure it to use an Anthropic Claude model (e.g., `claude-3-5-sonnet-latest`).
2.  Implement a function call within the chat flow that the Anthropic model can trigger mid-response.
3.  Initiate a conversation that leads the Anthropic model to invoke the defined tool.
4.  Observe the streamed response from the assistant. You will notice an assistant message containing a `tool_call` and a preceding or subsequent content block with `{"type": "text", "text": ""}`.
5.  Attempt to send the tool call result back to the Anthropic API. This action will fail with the `[AI_APICallError: messages: text content blocks must be non-empty]` error.

## Expected Behavior

The Vercel AI SDK should automatically filter out any text content blocks with empty `text` values in messages being sent to the Anthropic API. This would ensure that only valid, non-empty text blocks and tool calls are included in the API request, allowing for seamless tool calling and continued chat functionality without errors.

## Actual Behavior

The Vercel AI SDK includes empty text content blocks in the messages sent to the Anthropic API when a tool call is triggered mid-response. This violates the Anthropic API's requirement for non-empty text blocks, leading to the API rejecting the request and throwing the `[AI_APICallError: messages: text content blocks must be non-empty]` error, thus breaking the chat flow.

### Additional context

```typescript
import { streamText } from 'ai';
import { createAnthropic } from '@ai-sdk/anthropic';

// Configure Anthropic provider (replace with your API key)
const anthropic = createAnthropic({
  apiKey: process.env.ANTHROPIC_API_KEY || 'your-api-key-here',
});

// Define a simple tool
const tools = {
  getData: {
    description: 'Fetches data based on a query',
    parameters: {
      type: 'object',
      properties: {
        query: { type: 'string' },
      },
      required: ['query'],
    },
    execute: async ({ query }) => {
      return `Data for: ${query}`;
    },
  },
};

// Sample system prompt
[...]

PR #14070, opened 2026-04-02
fix(ai): preserve empty text parts in multi-step response messages

nelsonauner:
## Summary

- **Fix**: `toResponseMessages()` was filtering out empty text blocks (`text: ""`), which broke multi-step `generateText` calls when Bedrock/Claude returned an empty text block between reasoning/thinking blocks.
- Bedrock validates that the content block structure of assistant messages is unchanged between steps. Dropping an empty text block shifted all subsequent content indices, causing Bedrock to reject step 2 with: `"thinking blocks cannot be modified. These blocks must remain as they were in the original response."`
- The bug was intermittent — Claude only sometimes emits empty text blocks between reasoning blocks.

### Root cause

Bedrock Converse API response (step 1):
```
content[0]: reasoningContent (analysis + signature)
content[1]: {"text": ""}          ← empty text block
content[2]: reasoningContent (plan + signature)
content[3]: text
content[4]: toolUse
```

After `toResponseMessages()` filtered the empty text, step 2 sent:
```
content[0]: reasoningContent ✓
content[1]: reasoningContent ✗ (Bedrock expected empty text at this index)
content[2]: text
content[3]: toolUse
```

### Changes

1. **`to-response-messages.ts`** — Removed the empty text filter. Empty text parts are structurally significant and must be preserved for round-tripping.
2. **`to-response-messages.test.ts`** — Updated existing test, added new test reproducing the exact production scenario (reasoning + empty text + reasoning + tool-call).

## Test plan

- [x] `to-response-messages.test.ts` — 21 tests pass
- [x] `generate-text.test.ts` — 184 tests pass
- [ ] Verify no regressions in other providers that may have relied on empty text filtering

### DIFF
--- a/packages/ai/src/generate-text/generate-text.test.ts
+++ b/packages/ai/src/generate-text/generate-text.test.ts
@@ -8835,4 +8835,100 @@ describe('generateText', () => {
       expect(events).toEqual(['first', 'second']);
     });
   });
+
+  describe('empty text blocks between reasoning blocks', () => {
+    it('should preserve empty text blocks in step 2 messages when reasoning blocks are present', async () => {
+      let responseCount = 0;
+      const doGenerateCalls: Array<Parameters<
+        ConstructorParameters<typeof MockLanguageModelV4>[0] & {
+          doGenerate: (...args: any[]) => any;
+        }
+      >> = [];
+
+      const result = await generateText({
+        model: new MockLanguageModelV4({
+          doGenerate: async options => {
+            doGenerateCalls.push(options as any);
+            switch (responseCount++) {
+              case 0:
+                return {
+                  ...dummyResponseValues,
+                  content: [
+                    {
+                      type: 'reasoning' as const,
+                      text: 'First reasoning block',
+                      signature: 'sig-1',
+                      providerMetadata: {
+                        bedrock: { signature: 'sig-1' },
+                      },
+                    },
+                    { type: 'text' as const, text: '' },
+                    {
+                      type: 'reasoning' as const,
+                      text: 'Second reasoning block',
+                      signature: 'sig-2',
+                      providerMetadata: {
+                        bedrock: { signature: 'sig-2' },
+                      },
+                    },
+                    { type: 'text' as const, text: 'Visible response' },
+                    {
+                      type: 'tool-call' as const,
+                      toolCallType: 'function' as const,
+                      toolCallId: 'call-1',
+                      toolName: 'myTool',
+                      input: '{ "value": "test" }',
+                    },
+                  ],
+                  finishReason: {
+                    unified: 'tool-calls' as const,
+                    raw: undefined,
+                  },
+                };
+              default:
+                return {
+                  ...dummyResponseValues,
+                  content: [{ type: 'text' as const, text: 'Done.' }],
+                };
+            }
+          },
+        }),
+        tools: {
+          myTool: tool({
+            inputSchema: z.object({ value: z.string() }),
+            execute: async ({ value }) => `${value}-result`,
+          }),
+        },
+        prompt: 'test-input',
+        stopWhen: isStepCount(3),
+      });
+
+      // Step 2 should have been called
+      expect(doGenerateCalls.length).toBe(2);
+
+      // Get the assistant message from step 2's prompt
+      const step2Prompt = doGenerateCalls[1].prompt;
+      const assistantMessage = step2Prompt.find(
+        (m: any) => m.role === 'assistant',
+      );
+
+      expect(assistantMessage).toBeDefined();
+
+      // The assistant message content must preserve the empty text block
+      // between reasoning blocks. Without this, providers like Bedrock
+      // reject with "thinking blocks cannot be modified" because the
+      // content block indices shift.
+      const contentTypes = assistantMessage!.content.map(
+        (p: any) => p.type + (p.text === '' ? '(empty)' : ''),
+      );
+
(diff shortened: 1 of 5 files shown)


gr2m:
I ran this example 100 times in an attempt to reproduce the problem, but no luck so far

```ts
import { isStepCount, streamText, tool } from 'ai';
import { z } from 'zod';
import { saveRawChunks } from '../../lib/save-raw-chunks';

// When Bedrock Claude with thinking enabled returns an empty text block
// between reasoning blocks (e.g. [reasoning, "", reasoning, text, tool_call]),
// toResponseMessages() filters it out. On the next step, Bedrock rejects the
// request because the content structure no longer matches the original response.
//
// The bug is intermittent — Claude only sometimes emits empty text blocks
// between reasoning blocks. Run this example multiple times to reproduce.

run(async () => {
  const result = streamText({
    model: bedrock('us.anthropic.claude-sonnet-4-5-20250929-v1:0'),
    stopWhen: isStepCount(5),
    reasoning: 'low',
    includeRawChunks: true,
    tools: {
      weather: tool({
        description: 'Get the current weather in a location',
        inputSchema: z.object({
          location: z.string().describe('The location to get the weather for'),
        }),
        execute: async ({ location }) => ({
          location,
          temperature: 72 + Math.floor(Math.random() * 21) - 10,
          condition: 'partly cloudy',
        }),
      }),
      search: tool({
        description: 'Search for information on the web',
        inputSchema: z.object({
          query: z.string().describe('The search query'),
        }),
        execute: async ({ query }) => ({
          results: [
            {
              title: `Result for: ${query}`,
              snippet: 'This is a mock search result.',
            },
          ],
        }),
      }),
    },
    prompt:
      'What is the weather in San Francisco and New York? Also search for "best restaurants in San Francisco". Think carefully about how to approach this.',
  });

  await saveRawChunks({
    result,
    filename: 'bedrock-thinking-tool-call',
  });
});
```

Anything we could do to adapt the example to make it closer to your own AI SDK use?

nelsonauner:
> Anything we could do to adapt the example to make it closer to your own AI SDK use?

Hard to say - we've observed this 13 times out of millions of requests so I think repro'ing might be damn near impossible. Let me talk with the bedrock team

nelsonauner:
Me general point though (discussed in Slack) is it seems very odd to be _dropping_ messages that are returned by the LLM provider. It messes up the whole ordering (e.g. grabbing by index) of the conversation/messages.

nelsonauner:
Here is the response from the Bedrock team asserting that empty text outputs are valid: 

```
t's a pleasure to meet you! I hope you're doing well. Thank you for contacting AWS Premium Support. My name is Eduardo from the ML team and it's my pleasure to assist you with this case today.

I understand that you're experiencing an issue where Claude Sonnet 4.6 (using the Converse API) occasionally returns empty text blocks ({"text":""}) between reasoning/thinking content blocks in the response. This is causing your application and the Vercel AI SDK to crash because they don't expect empty text outputs between thinking blocks, and you're seeking guidance on whether this is a valid response pattern. Please feel free to let me know if I may have misunderstood anything here.

First and foremost, thank you so much for taking the time to thoroughly explain the situation. Your detailed description is incredibly helpful!

After consulting with the internal Bedrock team, the team shared that the Bedrock Converse API ContentBlock [1] documentation indicates that none of the content fields are required, meaning empty text blocks can occur as part of normal API behavior. As such, it's recommended to Implement logic in your application to filter out or handle empty text blocks between reasoning outputs.

Please kindly note that I will further check with the internal engineering team to confirm if there are any additional considerations or workarounds specific to this behavior. If I find any further insights, I'll follow up. Otherwise, the recommendation above should address the issue.

Please kindly rest assured that I am with you on this and if at any moment you have any additional information or questions, please don't hesitate to reach out. I will be more than happy to help you at the earliest with the most appropriate information!

Thank you and I sincerely wish you a wonderful rest of your day!

References:
[1] https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlock.html
```

lgrammel:
a few notes:

a) the filtering was introduced in a time when there were no content parts, just text, to prevent unnecessary ui updates (see https://github.com/vercel/ai/pull/2966/changes ). this is less relevant now that there are content parts. if a provider roundtrip breaks because of the filtering, it needs to be removed.
b) there may be ui regressions caused by empty text parts depending on how rendering is done, so this could be considered a breaking change (could introduce ui flickering perceived as bug)
c) there may be regressions with other providers, so we need to manually test at least the major providers and ensure they still work end-to-end
d) this needs a changeset

gr2m:
Also relevant, Anthropic breaks the other way around

- https://github.com/vercel/ai/issues/5576
- https://github.com/vercel/ai/pull/5767

@nelsonauner for now, can we try to contain this fix to bedrock? I think if we add `.providerMetadata` to an empty text chunk it will not be removed, it's something we would need to implement in the bedrock provider.

For the upcoming v7 we will look into removing the current behavior of removing empty text chunks by default, instead implement workarounds where needed (like the one for anthropic)

nelsonauner:
@gr2m @lgrammel Thank you both for the kind response. 

It's unfortunate that the bedrock anthropic and anthropic first party seems to have divergent behavior. I agree with keeping this a bedrock-only fix and ideally they and anthropic determine a single way this behavior should work

nelsonauner:
This was merged in via https://github.com/vercel/ai/pull/14175, closing this PR

PR #14099, opened 2026-04-03
fix(groq): use json tool fallback when structuredOutputs is disabled

matingathani:
## Summary

Fixes #8533 — v4→v5 regression where `generateObject` with `structuredOutputs: false` silently dropped the schema, causing Groq to return unstructured JSON.

- When `structuredOutputs: false` and a JSON schema is provided, the SDK now injects a hidden `json` function tool and forces `tool_choice` to select it — matching the [Anthropic provider's `jsonResponseTool` pattern](https://github.com/vercel/ai/blob/main/packages/anthropic/src/anthropic-messages-language-model.ts)
- The tool call arguments are surfaced as text content so the AI SDK core can parse the structured output
- In streaming (`doStream`), tool-input-* events for the hidden `json` tool are converted to text-* events
- No `response_format` is set when using the tool fallback (avoiding a conflicting constraint)
- No warning is emitted — schema adherence is now fully enforced via the tool

## Test plan

- [x] Updated the existing `structuredOutputs: false` request body test to assert the injected tool + forced `tool_choice`
- [x] Updated the existing warnings test to assert no warnings are emitted (schema is now enforced)
- [x] Added `doGenerate` test: verifies json tool call arguments are returned as text content
- [x] Added `doStream` test: verifies json tool streaming is converted to text-start/text-delta/text-end events
- [x] All 91 tests pass (node + edge)


aayush-kapoor (commented):
that issue doesn't need a fix since it's an intended regression. see https://github.com/vercel/ai/issues/8533#issuecomment-3335100072

marking this as closed

PR #14102, opened 2026-04-03
fix: reuse google auth per provider instance

### LINKED ISSUE #14101: `@ai-sdk/google-vertex` Node auth helper recreates `GoogleAuth` too often due to reference-based cache invalidation
### Description

The Node-side auth helper used by `@ai-sdk/google-vertex` caches a module-level `GoogleAuth` instance, but invalidates that cache using object reference equality.

Current source:

```ts
// packages/google-vertex/src/google-vertex-auth-google-auth-library.ts
import { GoogleAuth, GoogleAuthOptions } from 'google-auth-library';

let authInstance: GoogleAuth | null = null;
let authOptions: GoogleAuthOptions | null = null;

function getAuth(options: GoogleAuthOptions) {
  if (!authInstance || options !== authOptions) {
    authInstance = new GoogleAuth({
      scopes: ['https://www.googleapis.com/auth/cloud-platform'],
      ...options,
    });
    authOptions = options;
  }
  return authInstance;
}

export async function generateAuthToken(options?: GoogleAuthOptions) {
  const auth = getAuth(options || {});
  const client = await auth.getClient();
  const token = await client.getAccessToken();
  return token?.token || null;
}
```

This creates two separate problems.

#### 1. Multiple provider instances with equivalent `googleAuthOptions` thrash the singleton

If an application creates multiple `createVertex()` providers with the same credentials but different `googleAuthOptions` object references, alternating between those providers causes `GoogleAuth` to be recreated.

This is because the Node provider captures `options.googleAuthOptions` and passes it directly to `generateAuthToken()`:

```ts
// packages/google-vertex/src/google-vertex-provider-node.ts
return createVertexOriginal({
  ...options,
  headers: async () => ({
    Authorization: `Bearer ${await generateAuthToken(
      options.googleAuthOptions,
    )}`,
    ...(await resolve(options.headers)),
  }),
});
```

A single `createVertex()` call reused across multiple models does **not** reproduce this specific case, because it reuses the same captured `googleAuthOptions` reference. The problem appears when applications create multiple provider instances, which is a common pattern in provider/model caches.

#### 2. `undefined` options recreate `GoogleAuth` on every call

When `googleAuthOptions` is omitted, `generateAuthToken()` calls:

```ts
getAuth(options || {})
```

That creates a new empty object every time `options` is `undefined`, so:

```ts
options !== authOptions
```

is always true after the first call.

As a result, even a single provider instance with no explicit `googleAuthOptions` can recreate `GoogleAuth` on every request.

### Scope

This appears to affect the shared Node auth helper used by:

- `@ai-sdk/google-vertex`
- `@ai-sdk/google-vertex/anthropic`
- `@ai-sdk/google-vertex/maas`

It does **not** appear to be the same issue in the Edge auth path, which uses a different implementation.

#### Reproduction A: multiple provider instances

```ts

const flash = createVertex({
  project: 'my-project',
  location: 'us-central1',
  googleAuthOptions: {
    credentials: {
      client_email: '...',
      private_key: '...',
    },
    scopes: ['https://www.googleapis.com/auth/cloud-platform'],
  },
});

const pro = createVertex({
  project: 'my-project',
  location: 'us-central1',
  googleAuthOptions: {
    credentials: {
      client_email: '...',
      private_key: '...',
    },
    scopes: ['https://www.googleapis.com/auth/cloud-platform'],
  },
});

await streamText({ model: flash('gemini-2.5-flash'), prompt: 'hi' });
await streamText({ model: pro('gemini-2.5-pro'), prompt: 'hi' });
// switching providers passes a different googleAuthOptions object reference
```

#### Reproduction B: omitted `googleAuthOptions`

```ts

const vertex = createVertex({
  project: 'my-project',
  location: 'us-central1',
});

await streamText({ model: vertex('gemini-2.5-flash'), prompt: '1' });
await streamText({ model: vertex('gemini-2.5-flash'), prompt: '2' });
// generateAuthToken(undefined) -> getAuth({}) -> new {} each call
```

### Why this matters

`GoogleAuth` caches the resolved auth client on the instance, so recreating the `GoogleAuth` object also drops that cached client state. In environments using workload identity federation or other nontrivial auth flows, this can add measurable overhead.

In our production usage, we observed a large latency difference when requests alternated between separately-created provider instances with equivalent credentials. I’m not claiming the SDK source alone proves a specific latency number, but it does appear to explain the mechanism.

### Suggested fix

The simplest fix seems to be removing the module-level singleton and instead creating one `GoogleAuth` per provider instance.

For example, replace the shared helper shape with a factory:

```ts
import { GoogleAuth, GoogleAuthOptions } from 'google-auth-library';

export function createAuthTokenGenerator(options?: GoogleAuthOptions) {
  const auth = new GoogleAuth({
    scopes: ['https://www.googleapis.com/auth/cloud-platform'],
    ...options,
  });

  return async function generateAuthToken() {
    const client = await auth.getClient();
    const token = await client.getAccessToken();
    return token?.token ?? null;
  };
}
```

Then in the Node provider:

```ts
const generateAuthToken = createAuthTokenGenerator(options.googleAuthOptions);

return createVertexOriginal({
  ...options,
  headers: async () => ({
    Authorization: `Bearer ${await generateAuthToken()}`,
    ...(await resolve(options.headers)),
  }),
});
```

This would:

- avoid reference-equality invalidation entirely
- fix the `undefined` / `options || {}` problem
- give each provider instance stable auth/client caching
- avoid needing structural serialization of `GoogleAuthOptions` values such as `authClient`

### Suggested tests

- repeated calls with omitted `googleAuthOptions` should not recreate `GoogleAuth`
- repeated calls on one provider with explicit `googleAuthOptions` should not recreate `GoogleAuth`
- alternating between two separately-created providers should not cause unnecessary recreation within a provider
- same fix should cover the shared helper consumers in `anthropic` and `maas`

### Environment

We originally noticed the behavior while using Node.js with workload identity federation, but the cache invalidation logic appears auth-method-agnostic.

### AI SDK Version

6.0.145

### Code of Conduct

heiwen:
## Background

`@ai-sdk/google-vertex` used a module-level `GoogleAuth` cache in the shared Node auth helper. That cache was invalidated by object reference equality, which caused unnecessary `GoogleAuth` recreation in two cases:

1. Multiple provider instances with equivalent `googleAuthOptions` would recreate auth when requests alternated between them.
2. When `googleAuthOptions` was omitted, `generateAuthToken(options || {})` created a new empty object on every call, so auth could be recreated on every request even for a single provider.

Because `GoogleAuth` caches the resolved auth client on the instance, recreating it discards warm auth state and can add avoidable auth overhead.

## Summary

This PR removes the module-level `GoogleAuth` singleton from `@ai-sdk/google-vertex`'s shared Node auth helper and replaces it with a per-provider auth token generator.

Changes included:
- Replaced the shared `generateAuthToken(options)` helper with `createAuthTokenGenerator(options)`, which creates one `GoogleAuth` instance and returns a reusable token generator closure.
- Updated the Node provider entry points for:
  - `@ai-sdk/google-vertex`
  - `@ai-sdk/google-vertex/anthropic`
  - `@ai-sdk/google-vertex/maas`
- Added regression tests covering:
  - repeated calls with omitted `googleAuthOptions`
  - repeated calls on a single provider instance
  - per-provider generator creation for the Node provider variants
- Added a patch changeset for `@ai-sdk/google-vertex`

## Manual Verification

I manually verified the change by inspecting the Node auth flow and confirming the auth generator is now created once during provider construction rather than on each request path.

Concretely:
- confirmed each Node provider now creates `createAuthTokenGenerator(options.googleAuthOptions)` once at provider initialization
- confirmed request-time header/fetch code now calls the returned generator with no options, so the `undefined -> {}` invalidation path is gone
- confirmed the shared helper no longer stores module-level auth state

I also verified behavior through automated coverage:
- `pnpm --filter @ai-sdk/google-vertex test:node`
- `pnpm type-check:full`

## Checklist

## Future Work

A follow-up PR could evaluate whether the Edge auth path should also cache tokens or auth state more aggressively, since it uses a different implementation and is not addressed here.

## Related Issues

Fixes #14101

### DIFF
--- a/.changeset/tall-peaches-tease.md
+++ b/.changeset/tall-peaches-tease.md
@@ -1,0 +1,10 @@
+---
+'@ai-sdk/google-vertex': patch
+---
+
+fix(provider/google-vertex): avoid recreating Node GoogleAuth clients for repeated requests
+
+Create Google auth token generators per provider instance instead of using a
+module-level shared `GoogleAuth` cache. This avoids unnecessary `GoogleAuth`
+recreation when `googleAuthOptions` are omitted or when multiple provider
+instances use equivalent auth settings.
--- a/packages/google-vertex/src/anthropic/google-vertex-anthropic-provider-node.test.ts
+++ b/packages/google-vertex/src/anthropic/google-vertex-anthropic-provider-node.test.ts
@@ -1,12 +1,14 @@
 import { resolve } from '@ai-sdk/provider-utils';
+import { createAuthTokenGenerator } from '../google-vertex-auth-google-auth-library';
 import { createGoogleVertexAnthropic as createVertexAnthropicOriginal } from './google-vertex-anthropic-provider';
 import { createGoogleVertexAnthropic as createVertexAnthropicNode } from './google-vertex-anthropic-provider-node';
-import { generateAuthToken } from '../google-vertex-auth-google-auth-library';
 import { describe, beforeEach, expect, it, vi } from 'vitest';
 
 // Mock the imported modules
 vi.mock('../google-vertex-auth-google-auth-library', () => ({
-  generateAuthToken: vi.fn().mockResolvedValue('mock-auth-token'),
+  createAuthTokenGenerator: vi.fn(() =>
+    vi.fn().mockResolvedValue('mock-auth-token'),
+  ),
 }));
 
 vi.mock('./google-vertex-anthropic-provider', () => ({
@@ -51,7 +53,7 @@ describe('google-vertex-anthropic-provider-node', () => {
     });
   });
 
-  it('passes googleAuthOptions to generateAuthToken', async () => {
+  it('passes googleAuthOptions to createAuthTokenGenerator', async () => {
     createVertexAnthropicNode({
       googleAuthOptions: {
         scopes: ['https://www.googleapis.com/auth/cloud-platform'],
@@ -63,9 +65,9 @@ describe('google-vertex-anthropic-provider-node', () => {
     const passedOptions = vi.mocked(createVertexAnthropicOriginal).mock
       .calls[0][0];
 
-    await resolve(passedOptions?.headers); // call the headers function
+    await resolve(passedOptions?.headers);
 
-    expect(generateAuthToken).toHaveBeenCalledWith({
+    expect(createAuthTokenGenerator).toHaveBeenCalledWith({
       scopes: ['https://www.googleapis.com/auth/cloud-platform'],
       keyFile: 'path/to/key.json',
     });
@@ -86,7 +88,7 @@ describe('google-vertex-anthropic-provider-node', () => {
       Authorization: 'Bearer custom-token',
     });
     expect(customGenerate).toHaveBeenCalledTimes(1);
-    expect(generateAuthToken).not.toHaveBeenCalled();
+    expect(createAuthTokenGenerator).not.toHaveBeenCalled();
   });
 
   it('merges custom generateAuthToken with user-provided headers', async () => {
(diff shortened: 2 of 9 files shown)


aayush-kapoor (commented):
thank you for the PR! changes look good, you'll just need to resolve the conflicts and sign your commits please.

heiwen:
Conflicts resolved and commits signed.

aayush-kapoor:
apologies for the back and forth but there's a change coming in https://github.com/vercel/ai/pull/14857 that will cause this to conflict again. i'll ask you to resolve it again since that will be merged first. thank you

heiwen:
No Problem. Merge conflicts solved.

aayush-kapoor (approved):

PR #14131, opened 2026-04-04
fix(prompt): skip generating download requests for data URLs

Diode11-Alt:
Fixes #13103

### Description
This fixes a critical regression where valid \`data:\` URLs (e.g. inline images, pdfs) are rejected due to stricter SSRF URL validation in \`downloadAssets\`. 

Since \`data:\` URLs represent inline data and are not remote resources that need downloading, they should be filtered out from the \`plannedDownloads\` array altogether rather than being passed into the \`download\` map.

### Changes
*   Updated the filter in \`packages/ai/src/prompt/convert-to-language-model-prompt.ts\` inside \`downloadAssets\` to check for \`part.data.protocol !== 'data:'\` before adding it to \`plannedDownloads\`.


aayush-kapoor (commented):

PR #14137, opened 2026-04-05
fix(amazon-bedrock): avoid env AWS_SESSION_TOKEN with explicit static keys

turisanapo:
## Background

`createAmazonBedrock` merged explicit static `accessKeyId` / `secretAccessKey` with `AWS_SESSION_TOKEN` from the environment via `loadOptionalSetting`. On EKS IRSA (and similar), that token is for the workload—not the IAM user keys in options—so SigV4 fails with *invalid security token* (#14136).

The published docs also claimed that setting `sessionToken: undefined` avoids env defaults. **`undefined` does not skip env** in `loadOptionalSetting` (`settingValue != null` is false), which we **confirmed manually** (Hebo + fake `AWS_SESSION_TOKEN`). The bug is both **implementation** and **documentation**.

## Summary

- **Code:** When both `accessKeyId` and `secretAccessKey` are string options, `sessionToken` comes only from the `sessionToken` option (or omitted); **`AWS_SESSION_TOKEN` is not read** for that path. If either key is env-resolved, behavior is unchanged (`AWS_SESSION_TOKEN` still applies when needed).
- **Docs:** Updated `08-amazon-bedrock.mdx`: removed the incorrect `sessionToken: undefined` guidance; documented static keys vs env; pointed to `credentialProvider` for older releases / edge cases.
- **Changeset:** Patch for `@ai-sdk/amazon-bedrock`; body mentions doc correction.

## Manual Verification

- `pnpm build --filter=@ai-sdk/amazon-bedrock...`
- `pnpm --filter=@ai-sdk/amazon-bedrock test:node`
- Manual: with `AWS_SESSION_TOKEN=fake` and real static keys in options, Bedrock request should succeed after this change (and docs match behavior).

## Checklist

## Future Work

None required.

## Related Issues

Fixes #14136

### DIFF
--- a/.changeset/bedrock-static-keys-session-token.md
+++ b/.changeset/bedrock-static-keys-session-token.md
@@ -5,3 +5,5 @@
 fix(amazon-bedrock): do not merge AWS_SESSION_TOKEN from env with explicit access keys
 
 When `accessKeyId` and `secretAccessKey` are both passed as options, session credentials now use only `options.sessionToken` or omit the token — avoiding workload tokens (e.g. EKS IRSA) invalidating SigV4 for static IAM user keys.
+
+Docs: remove incorrect guidance that `sessionToken: undefined` blocks `AWS_SESSION_TOKEN`; describe static keys, env fallback, and `credentialProvider` workaround for older versions.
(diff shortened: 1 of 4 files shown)


aayush-kapoor (commented):
closing this in favour of more succinct fix in https://github.com/vercel/ai/pull/14262

PR #14139, opened 2026-04-05
fix(amazon-bedrock): preserve empty text blocks when reasoning content is present

Gdhanush-13:
Fixes #14071

Bedrock validates that assistant message content block indices match the original response. The SDK currently strips empty text blocks (`!part.text.trim()`), which shifts indices when reasoning blocks are present, causing:
`The model cannot modify thinking or redacted_thinking blocks``n
This PR preserves empty text blocks when the message contains reasoning content.


aayush-kapoor (commented):
needs tests added + changeset + uses `any`

will be marking as closed and continuing in a new pr

PR #14140, opened 2026-04-05
fix(open-responses): map non-image file parts to input_file (fixes #13843)

shivamtiwari3:
## Summary

Fixes #13843.

Maps non-image file parts (e.g. `application/pdf`) to `input_file` in `convertToOpenResponsesInput`, instead of silently dropping them with a warning.

---

## Root Cause

In `packages/open-responses/src/responses/convert-to-open-responses-input.ts:56`, the `file` case handler checks `!part.mediaType.startsWith('image/')` and, when true, pushes a warning and `break`s — silently discarding the file part from the request body. The model then responds as if no document was provided.

The `InputFileContentParam` type (`{ type: 'input_file', filename?, file_data?, file_url? }`) is already defined and imported but was never used in the user message path. The same pattern is already correctly implemented in the tool result content handler (line 177) and in the `@ai-sdk/openai` provider's `convert-to-openai-responses-input.ts`.

---

## Solution

Restructured the file part handling to:
1. **Images** (`image/*`): Continue mapping to `input_image` (unchanged behavior)
2. **Non-image files** (e.g. `application/pdf`): Map to `input_file` with either:
   - `file_url` when the data is a URL
   - `file_data` (base64 data URI) + `filename` when the data is binary/base64

This matches the approach used by the `@ai-sdk/openai` provider and the existing tool result content handler in this same file.

---

## Testing

- Updated existing test that asserted the old (broken) warning behavior → now asserts PDF base64 data maps to `input_file`
- Added test for PDF file parts with URL data → maps to `input_file` with `file_url`
- Added test for custom `filename` property on non-image file parts
- All 56 tests pass (node + edge): `pnpm test` in `packages/open-responses`

---

## Checklist

- [x] All existing tests pass
- [x] No unrelated changes
- [x] Read CONTRIBUTING.md and followed its requirements


.changeset/fix-prune-messages-orphaned-reasoning.md
```diff
@@ -0,0 +1,15 @@
+---
+'ai': patch
```
aayush-kapoor, line 2: missing package for changesets

packages/ai/src/generate-text/prune-messages.ts
```diff
@@ -160,7 +160,21 @@ export function pruneMessages({
...
+      // valid assistant message must contain at least one non-reasoning block.
+      if (
+        message.role === 'assistant' &&
+        typeof message.content !== 'string' &&
+        message.content.every(part => part.type === 'reasoning')
+      ) {
+        return false;
+      }
+      return true;
+    });
```
aayush-kapoor, line 177: the issue isn't in pruneMessages, why is this added? why are we changing logic here?

aayush-kapoor:
this will also need an e2e next example that we test on, to ensure multi step conversations work with non image file parts

shivamtiwari3:
@aayush-kapoor 

Added in 6e2abf860 — created an e2e Next.js example at `/chat/test-open-responses-pdf-multi-step` with:

- API route using `createOpenResponses` pointed at OpenAI's responses API
- Two tools (`summarizeSection`, `extractKeyTerms`) with `isStepCount(5)` to exercise multi-step conversations
- Page UI for uploading PDFs and chatting across multiple turns

This ensures non-image file parts flow correctly through `convertToOpenResponsesInput` in multi-step conversations.

lgrammel:
Please capture a real test fixture that shows the input stream. See https://github.com/vercel/ai/blob/main/contributing/testing.md#providers

shivamtiwari3:
@lgrammel 

Added real API test fixtures in ae443dc61:

- **Capture scripts**: `examples/ai-functions/src/{stream,generate}-text/open-responses/openai-pdf-input-file.ts` — sends a PDF via URL to OpenAI's Responses API using `createOpenResponses`
- **Fixtures**: `packages/open-responses/src/responses/__fixtures__/openai-pdf-input-file.1.{json,chunks.txt}` — captured from real `gpt-4.1-nano` responses
- **Tests**: Added `pdf input file` test sections in `open-responses-language-model.test.ts` for both `doGenerate` (request body, content, usage) and `doStream` (streamed content)

All 60 tests pass (node + edge). Also rebased on latest `main` to resolve the merge conflict.

aayush-kapoor (commented):
marking as closed in favour of https://github.com/vercel/ai/pull/15212 - will add you as a co-author

Issue #11920, opened 2026-01-21
Feature Request: Dynamic Tool Selection to Reduce Context Pollution


Shervanator:
### Problem
When building agentic applications with many tools (20+), sending all tool definitions to the LLM on every request causes:
- Degraded model performance - The model gets confused or makes poor tool choices when presented with too many options
- Increased token usage - Tool definitions can consume thousands of tokens per request
- Higher latency - Larger payloads slow down requests

### Request
A built-in mechanism to dynamically select relevant tools based on the user's query, reducing the number of tools sent to the model while ensuring the right tools are available.

Anthropic seems to be solving this with a "tool search tool", is something like that possible here?

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Shervanator:
I guess the best option is to use https://ai-sdk.dev/providers/ai-sdk-providers/anthropic#tool-search 
> Anthropic provides provider-defined tool search tools that enable Claude to work with hundreds or thousands of tools by dynamically discovering and loading them on-demand. Instead of loading all tool definitions into the context window upfront, Claude searches your tool catalog and loads only the tools it needs.

karthikjn01:
I'm also wondering how we'd go about doing this - I actually ended up writing a tool loop agent that'd loop and 'request' tools as it goes to 'load' and 'unload' tools by specifically requesting them - the issue was that this didn't work very well!

dankelleher:
We had the same token-bloat problem with 100+ tools across MCP servers. What I did: start with a small bootstrap set and refresh mid-conversation when the user loads more. I packaged it as https://www.npmjs.com/package/ai-sdk-tool-refresh

lgrammel:
### Identify issue type

```json
{
  "evidence": [
    "The issue requests a built-in mechanism to dynamically select relevant tools based on the user query instead of sending all tool definitions.",
    "The described motivation is reducing token usage, latency, and model confusion for applications with many tools.",
    "Comments discuss possible approaches and existing Anthropic tool search support, but do not report a broken existing AI SDK behavior."
  ],
  "rationale": "This asks for a new capability/ergonomics improvement for dynamic tool selection, not a failure of current documented behavior.",
  "type": "feature",
  "confidence": "high"
}
```

lxcid:
Kimi K3 just share dynamic tool loading technique, which allow you to just inject a system tool prompt, I think more models might soon adopt it. I hope we can get first class support in AI SDK for this.

[...]

PR #14147, opened 2026-04-06
feat(ai/core): lazy tool discovery for token-efficient large tool sets

fkesheh:
## Summary

Implements lazy tool discovery as requested in #13423. Tools marked with `lazy: true` send only their name and description to the LLM — the full `inputSchema` is omitted. A synthetic `__load_tool_schema__` tool is auto-injected so the LLM can fetch full schemas on demand before calling a tool.

**How it works:**
1. Lazy tools are sent to the LLM with a minimal empty schema (`{ type: 'object', properties: {} }`) instead of their full inputSchema
2. The LLM calls `__load_tool_schema__` to get the full schema as a tool result (text)
3. The LLM calls the lazy tool with arguments based on the loaded schema
4. If the LLM sends strings for nested objects/arrays (common with empty schemas), the SDK auto-repairs by JSON.parsing those values against the real schema, then validates and executes

This reduces token usage and improves tool selection accuracy for apps with large tool sets (20-50+ tools).

**Prior art:** [TanStack AI lazy tool discovery](https://tanstack.com/blog/tanstack-ai-lazy-tool-discovery), [Cursor dynamic context discovery](https://cursor.com/blog/dynamic-context-discovery) (46.9% token reduction in A/B test).

### Design principles
- **Stateless** — no tracking of what's been loaded, no message scanning
- **Cache-friendly** — same tool list every step, preserves LLM prompt cache
- **All tools always visible** (name + description) and always executable
- **Overridable** — users can define their own `__load_tool_schema__` tool
- **`skill` field** — optional extended docs loaded on demand for lazy tools

### Key implementation details
- `prepareTools`: lazy tools emit minimal empty schema with instruction to load schema first
- `parseToolCall`: for lazy tools, auto-repairs string→object/array mismatches using the real schema
- `__load_tool_schema__`: returns serialized JSON Schema (stripped of Zod internals via JSON.parse/stringify)
- `buildEffectiveTools`: shared helper used by both `generateText` and `streamText`
- `filterActiveTools`: preserves `__load_tool_schema__` when `activeTools` filtering is used

### Changes
- `@ai-sdk/provider-utils`: Add `lazy?: boolean` and `skill?: string` to `Tool` type
- `ai`: New `load-tool-schema.ts` — synthetic tool factory + `buildEffectiveTools` helper
- `ai`: `prepareTools` emits minimal empty schema for lazy tools
- `ai`: `parseToolCall` auto-repairs lazy tool inputs (JSON.parse strings that should be objects/arrays)
- `ai`: `filterActiveTools` preserves `__load_tool_schema__` through activeTools filtering
- `ai`: Auto-inject `__load_tool_schema__` when lazy tools exist (overridable)

## Test plan
- [x] Unit tests for `createLoadToolSchemaTool` (schema return, skill, inputExamples, error entries)
- [x] Unit tests for `buildEffectiveTools` (no-op without lazy, injection, user override)
- [x] Unit tests for `prepareTools` (minimal schema for lazy, mixed lazy/eager)
- [x] All 2210 existing tests pass
- [x] Full monorepo build (69/69 tasks)
- [x] Lint/format clean
- [x] Tested in real app (legbi) with 33 tools, 17 lazy

Closes #13423

### DIFF
--- a/packages/ai/src/generate-text/load-tool-schema.ts
+++ b/packages/ai/src/generate-text/load-tool-schema.ts
@@ -14,9 +14,7 @@ export function createLoadToolSchemaTool(
   return tool({
     description: `Load the full input schema and usage details for tools before calling them. You MUST call this first for any of these tools: ${toolNames.join(', ')}. Always pass at least one tool name. After receiving the schema, proceed to call the tool with the correct arguments.`,
     inputSchema: z.object({
-      toolNames: z
-        .array(z.string())
-        .min(1, 'At least one tool name is required'),
+      toolNames: z.array(z.string()),
     }),
     async execute({ toolNames: requestedNames }) {
       if (requestedNames.length === 0) {
--- a/packages/ai/src/generate-text/load-tool-schema.ts
+++ b/packages/ai/src/generate-text/load-tool-schema.ts
@@ -4,15 +4,7 @@ import { z } from 'zod/v4';
 
 export const LOAD_TOOL_SCHEMA_NAME = '__load_tool_schema__';
 
-type LoadToolSchemaResult = Record<
-  string,
-  | {
-      inputSchema: unknown;
-      skill?: string;
-      inputExamples?: unknown[];
-    }
-  | { error: string }
->;
+type LoadToolSchemaResult = Record<string, unknown>;
 
 export function createLoadToolSchemaTool(
   lazyTools: ToolSet,
@@ -20,7 +12,7 @@ export function createLoadToolSchemaTool(
   const toolNames = Object.keys(lazyTools);
 
   return tool({
-    description: `Load the full input schema for lazy tools. Available lazy tools: ${toolNames.join(', ')}`,
+    description: `Load the full input schema and usage details for tools before calling them. You MUST call this first for any of these tools: ${toolNames.join(', ')}. After receiving the schema, proceed to call the tool with the correct arguments.`,
     inputSchema: z.object({
       toolNames: z.array(z.string()),
     }),
@@ -41,7 +33,9 @@ export function createLoadToolSchemaTool(
           skill?: string;
           inputExamples?: unknown[];
         } = {
-          inputSchema: await asSchema(t.inputSchema).jsonSchema,
+          inputSchema: JSON.parse(
+            JSON.stringify(await asSchema(t.inputSchema).jsonSchema),
+          ),
         };
 
         if (t.skill != null) {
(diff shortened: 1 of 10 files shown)


packages/ai/src/generate-text/load-tool-schema.ts
ziyadev, line 1: If the LLM calls `__load_tool_schema__({ toolNames: [] })`, the execute handler returns `{}` silently, and the LLM may loop calling it again with no progress. The fix could be simple by just return a warning message for empty input and explicitly tell the LLM in the tool's description to always pass at least one tool name.

fkesheh: very good catch, I will add that

fkesheh:
## Alternative approach: `modelInputSchema` for user-land discovery harness

After implementing the baked-in approach above, I explored a lighter alternative that gives users full control without adding SDK-specific discovery logic.

### The gap

Users can already:
- Add synthetic tools to the `tools` map (their own discovery tool)
- Use `prepareStep` to change `activeTools`, `toolChoice`, model, messages per step
- Use `toolCallRepair` to fix malformed tool calls

The **one missing piece**: users can't send a different schema to the LLM than the one used for validation. `prepareTools` always uses `inputSchema` for both.

### Proposed: `modelInputSchema`

Add an optional `modelInputSchema` field to the `Tool` type. When present, `prepareTools` sends it to the LLM instead of `inputSchema`. Validation still uses `inputSchema`.

```typescript
const myLazyTool = tool({
  description: 'Create entity type',
  inputSchema: z.object({ slug: z.string(), labels: z.object({...}), fields: z.array(...) }),
  modelInputSchema: z.object({}), // minimal schema sent to LLM
  execute: async (input) => { ... },
});

const discoverTools = tool({
  description: 'Get schema for tools',
  inputSchema: z.object({ toolNames: z.array(z.string()) }),
  execute: async ({ toolNames }) => { /* return schemas */ },
});

streamText({
  tools: { create_entity_type: myLazyTool, discover_tools: discoverTools },
  toolCallRepair: async ({ toolCall, tools, error }) => {
    // user-land: repair string→object mismatches for lazy tools
  },
});
```

### Trade-offs

| Aspect | Current PR (baked-in) | `modelInputSchema` alternative |
|--------|----------------------|-------------------------------|
| SDK changes | ~8 files, new synthetic tool, auto-repair | 2 files: Tool type + prepareTools |
| User effort | Zero — just add `lazy: true` | User builds discovery tool + repair |
| Flexibility | Opinionated flow | Users own the entire harness |
| Maintenance | More SDK surface to maintain | Minimal SDK surface |

Both approaches could coexist — `modelInputSchema` as the primitive, `lazy: true` as the batteries-included convenience built on top.

Happy to implement `modelInputSchema` as a separate PR if the team prefers the composable-primitives approach.

gr2m:
closed as duplicate of #11920

PR #14183, opened 2026-04-07
feat(google): add fileData support to embedding model

anaclumos:
What's this?

Google Embedding models support HTTP public links, but there's no way to use it through AI SDK! This enables it.

## Summary
- Add `fileData` (with `fileUri` and `mimeType`) to the Google embedding content part schema, enabling embedding of remote video, audio, and PDF content via Google Cloud Storage URIs
- Add test verifying fileData parts are correctly merged into the embedContent request body

## Test plan
- [x] New test `should merge fileData content for single embedding` passes
- [x] All existing embedding model tests continue to pass
- [x] Package builds successfully

### DIFF
--- a/.changeset/healthy-files-jam.md
+++ b/.changeset/healthy-files-jam.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/google": patch
+---
+
+feat(google): add `fileData` support to embedding model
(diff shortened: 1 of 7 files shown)


dancer (dismissed):

felixarntz (dismissed):
@anaclumos Thank you for the PR, looks great!

Will add a changeset, as required for package changes, and then we can also backport this to the AI SDK v6 branch.

PR #14186, opened 2026-04-07
fix: better context type inference

lgrammel:
## Background

Generic context was introduced in #13102 . However, the type inference for the `context` property was not requiring it to be available (no type error was shown when context was not specified).

## Summary

* switch from `GenerationContext` approach to separate `USER_CONTEXT`
* change context type on `Tool`
* update related code

## Manual Verification

- [x] check types and remove/experiment with type errors in `examples/ai-functions/src/generate-text/openai/tool-call-with-context.ts`
- [x] check types and remove/experiment with type errors in `examples/ai-functions/src/stream-text/openai/tool-call-with-context.ts`
- [x] check types and remove/experiment with type errors in `examples/ai-functions/src/agent/openai/generate-context.ts`

## Related Issues

Addresses limitation from #13102


gr2m (approved):
Love the extensive tests! Some great 🥷 typescript!

PR #14213, opened 2026-04-07
feat(provider-utils): add serializeModel and deserializeModelConfig

gr2m:
## Summary

Add utilities for workflow step boundary serialization of language models:

- **`serializeModel`**: Extracts `modelId` and serializable config from a model instance. Resolves `config.headers()` at serialization time so auth credentials survive the step boundary as plain key-value objects (instead of being stripped as functions).
- **`deserializeModelConfig`**: Wraps plain-object `headers` back into a function so model code can continue calling `config.headers()` after deserialization.

These are used by `WORKFLOW_SERIALIZE`/`WORKFLOW_DESERIALIZE` in provider model classes (added in #14210).

### How it works

```
Serialization:  config.headers = () => ({ 'x-api-key': 'sk-...' })
                        ↓ serializeModel calls headers()
Serialized:     config.headers = { 'x-api-key': 'sk-...' }
                        ↓ JSON survives step boundary
Deserialization: deserializeModelConfig wraps it back
Result:         config.headers = () => ({ 'x-api-key': 'sk-...' })
```

## Test plan

- [x] Unit tests for `serializeModel` (headers resolution, function filtering, class instance filtering)
- [x] Unit tests for `deserializeModelConfig` (wrapping, preserving existing functions, passthrough)
- [x] All 491 provider-utils tests pass

### DIFF
--- a/packages/provider-utils/src/index.ts
+++ b/packages/provider-utils/src/index.ts
@@ -78,3 +78,4 @@ export {
   EventSourceParserStream,
   type EventSourceMessage,
 } from 'eventsource-parser/stream';
+export { serializeModel, deserializeModelConfig } from './serialize-model';
--- a/packages/provider-utils/src/serialize-model.test.ts
+++ b/packages/provider-utils/src/serialize-model.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect } from 'vitest';
+import { serializeModel, deserializeModelConfig } from './serialize-model';
+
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by the simpler approach in #14210: resolve headers at serialization time in `serializeModel` and wrap them back in `deserializeModelConfig`. No per-provider auth files needed.

PR #14227, opened 2026-04-08
feat(google): support AUDIO response modalities

cipher416:
## Background

`@ai-sdk/google` already forwards `responseModalities` to Google language models and already maps non-text `inlineData` responses into generic `file` parts. However, the provider option schema only allowed `TEXT` and `IMAGE`, which blocked models that support audio output from being used with `responseModalities: ['AUDIO']`.

## Summary

- allow `AUDIO` in Google `responseModalities`
- add regression tests for request passthrough and audio `inlineData` parsing
- document that `TEXT`, `IMAGE`, and `AUDIO` are valid values, with support depending on the model
- add a `streamText` repro example for Google audio output using `lyria-3-clip-preview`

## Manual Verification

- ran focused Google tests:
  - `pnpm exec vitest run src/google-generative-ai-language-model.test.ts --config vitest.node.config.js`
- ran a targeted type-check for the repro example:
  - `pnpm exec tsc --noEmit --skipLibCheck --target es2022 --lib es2022,dom --module esnext --types node --esModuleInterop --allowSyntheticDefaultImports --moduleResolution bundler src/lib/save-audio.ts src/stream-text/google/audio-output.ts`
- ran the repro example end-to-end:
  - `pnpm tsx src/stream-text/google/audio-output.ts`
- verified that `lyria-3-clip-preview` returns an audio file part and saves an `.mp3` under `examples/ai-functions/output/`
- verified that a text-only model like `gemini-2.5-flash` fails at the provider layer with `This model only supports text output`, which matches the new model-dependent docs wording

## Checklist

## Future Work

- Add model-specific capability checks for Google `responseModalities` if the SDK wants stricter local validation for unsupported modality/model combinations.
- Consider a dedicated audio-generation abstraction in the future if more providers expose non-TTS audio generation through first-class APIs.

## Related Issues

Fixes #14038


RyanGarber:
+1

Everything needed for Gemini audio is there, but Vercel rejects the request all because no one added 'AUDIO' to the enum. I've had to patch AI SDK for the 3+ months now just to get around this. In my opinion, this should've been not only merged here, but also backported to v6, a long time ago. Yet so far, despite all the breaking changes, this broken schema has been left as-is in both.

PR #14258, opened 2026-04-09
fix(amazon-bedrock): return request body from doGenerate and doStream

giulio-leone:
## Summary

`BedrockChatLanguageModel.doGenerate()` and `doStream()` were not returning a `request` field in their results. This meant consumers relying on `request.body` (e.g., for observability/tracing) received `undefined`.

There was an existing `// TODO request?` comment in the `doStream` return confirming this was a known gap.

## Changes

- Add `request: { body: args }` to the return object of `doGenerate()`
- Add `request: { body: args }` to the return object of `doStream()` (replacing the TODO comment)
- Add regression tests for both methods verifying the request body is present
- Update affected snapshots

This matches the pattern used by `@ai-sdk/anthropic` and other providers.

## Test Plan

- All 345 existing tests pass (node + edge)
- 2 new regression tests added (`should include request body` for both `doGenerate` and `doStream`)
- 2 snapshot updates (existing `result` snapshots now include the new `request` field)

Fixes #13927

### DIFF
--- a/.changeset/bedrock-request-body.md
+++ b/.changeset/bedrock-request-body.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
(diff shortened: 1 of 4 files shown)


aayush-kapoor:

PR #14262, opened 2026-04-09
fix(amazon-bedrock): do not inherit env AWS_SESSION_TOKEN with explicit keys

giulio-leone:
## Summary

Fixes #14136

When both `accessKeyId` and `secretAccessKey` are provided as explicit options to `createAmazonBedrock`, the `sessionToken` is now sourced **only** from `options.sessionToken`. Previously, `loadOptionalSetting` would unconditionally fall back to `process.env.AWS_SESSION_TOKEN`, which on hosts with workload identity (EKS IRSA, ECS task role, Lambda) belongs to a different principal — producing mixed, invalid SigV4 credentials:

> "The security token included in the request is invalid."

## Changes

- **`packages/amazon-bedrock/src/bedrock-provider.ts`** — When `options.accessKeyId` and `options.secretAccessKey` are both explicitly provided, use `options.sessionToken` directly instead of falling through to `loadOptionalSetting` (which reads `AWS_SESSION_TOKEN` from the environment). When keys come from the environment, env session-token fallback is preserved.
- **`packages/amazon-bedrock/src/bedrock-provider.test.ts`** — Three regression tests exercising the credential-resolution callback:
  1. Explicit keys → env `AWS_SESSION_TOKEN` is **not** inherited
  2. Explicit keys + explicit `sessionToken` → works correctly
  3. No explicit keys → env `AWS_SESSION_TOKEN` fallback is preserved

## Behaviour matrix

| `accessKeyId` | `secretAccessKey` | `sessionToken` option | `AWS_SESSION_TOKEN` env | Result |
|---|---|---|---|---|
| explicit | explicit | omitted | set | ✅ `sessionToken` = `undefined` (no mixing) |
| explicit | explicit | explicit | set | ✅ `sessionToken` = option value |
| from env | from env | omitted | set | ✅ `sessionToken` = env value (preserved) |

### DIFF
--- a/.changeset/fix-bedrock-session-token-env.md
+++ b/.changeset/fix-bedrock-session-token-env.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
+fix(amazon-bedrock): do not inherit AWS_SESSION_TOKEN from env when explicit accessKeyId/secretAccessKey are provided
+
+When both `accessKeyId` and `secretAccessKey` are passed as explicit options, the session token is now sourced only from `options.sessionToken`. Previously, `loadOptionalSetting` would fall back to `process.env.AWS_SESSION_TOKEN`, which on hosts with workload identity (EKS IRSA, ECS task role, Lambda) belongs to a different principal, causing "The security token included in the request is invalid" errors.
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):
closing in favour of https://github.com/vercel/ai/pull/14832

PR #14263, opened 2026-04-09
fix: normalize deprecated `parameters` to `inputSchema` in tool()

giulio-leone:
## Summary

Fixes #13460

`tool({ parameters: z.object({...}) })` (the v5 API) silently produces an empty input schema because `prepareTools` and `doParseToolCall` only read `inputSchema`, which is `undefined` when `parameters` is used.

### Source fix (`@ai-sdk/provider-utils`)
- **`tool()`** now normalizes deprecated `parameters` → `inputSchema` at creation time, so all downstream consumers work without changes.

### Defensive fallback (`ai`)
- **`prepareTools`**: falls back to `parameters` when `inputSchema` is `undefined` (covers tools not created via `tool()`).
- **`doParseToolCall`**: same fallback for validation.
- **`parseToolCall` repair callback**: same fallback.

### Regression tests
- `tool.test.ts`: verifies normalization, precedence when both are present, property preservation.
- `prepare-tools.test.ts`: verifies schema emission via `tool()` and raw object with `parameters`.
- `parse-tool-call.test.ts`: verifies parsing and validation against `parameters` schema.

## Validation

- **487/487** provider-utils tests pass
- **1229/1229** ai package tests pass (22 pre-existing failures from missing `@ai-sdk/test-server` build, unrelated)
- All 6 new regression tests pass

### DIFF
--- a/.changeset/fix-tool-parameters-inputSchema.md
+++ b/.changeset/fix-tool-parameters-inputSchema.md
@@ -1,0 +1,10 @@
+---
+'@ai-sdk/provider-utils': patch
+'ai': patch
+---
+
+fix: support deprecated `parameters` property in tool() for backward compatibility
+
+When using `tool({ parameters: z.object({...}) })` (the v5 API), the schema was silently dropped because `prepareTools` and `doParseToolCall` only read `inputSchema`. The `tool()` helper now normalizes `parameters` to `inputSchema`, and consumers also fall back to `parameters` defensively.
+
+Fixes #13460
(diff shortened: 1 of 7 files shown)


giulio-leone:
Closing after re-evaluation. This is not a current-HEAD bug — it's a user-side API misuse (`parameters` was renamed to `inputSchema` in AI SDK v5). The maintainer confirmed this in issue comments, and a prior fix PR #13465 was already rejected. The `Tool` type requires `inputSchema`, `parameters?: never` exists as a deprecation guard, and all SDK consumers correctly read `inputSchema`. Apologies for the noise.

PR #14302, opened 2026-04-09
feat(amazon-bedrock): add guardrail content support for text and image parts

raphaeleidus:
## Background

The Amazon Bedrock Converse API supports [guardrail content blocks](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailConverseContentBlock.html) that allow marking text and image parts for guardrail evaluation with optional qualifiers (`grounding_source`, `query`, `guard_content`). The existing `BedrockGuardrailConverseContentBlock` type was untyped (`guardContent: unknown`) and there was no way to opt into this behavior from the AI SDK.

[Documentation on the Bedrock Feature](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html)

## Summary

- Added `BedrockGuardrailTextBlock` type extending `BedrockTextBlock` with optional `qualifiers` array.
- Typed `BedrockGuardrailConverseContentBlock.guardContent` as a union of text and image variants instead of `unknown`.
- Added Zod schemas (`bedrockTextPartProviderOptions`, `bedrockImagePartProviderOptions`) for validating provider options, used via `parseProviderOptions` consistent with existing patterns.
- Users can now set `providerOptions.bedrock.guardContent: true` on text and image parts to wrap them as `guardContent` blocks, with optional `guardContentQualifiers` for text parts.
- Added 8 tests covering text guard content (with/without qualifiers, false, no options) and image guard content (true, false, no options).

## Checklist


raphaeleidus:
@aayush-kapoor could I trouble you to look this over? 
Is there any additional context you might need here?

aayush-kapoor (commented):
not context but this implementation will need some actual examples + a UI example that will test for multiturn conversations + tests and fixtures possibly returned by bedrock to see what it responds with.

raphaeleidus:
@aayush-kapoor I got the ui example app in. You can try it out and see that when you use a guardrail and provide a blocked query, on the next message it wont continue blocking

raphaeleidus:
@aayush-kapoor sorry to keep bugging you. Let met know if there is anything else you want to see here

KiraPC:
Strong +1 on this. 

Selective guarding is the only AWS-sanctioned way to keep injected, non-user-authored context (system reminders, retrieved snippets, recency notes) out of guardrail evaluation while still protecting the actual user input. Without it, that injected context can trip input filters (e.g. prompt-attack detection) and block otherwise benign queries. The implementation here looks clean and mirrors the existing cachePoint precedent nicely. Would love to see this land.

One thing worth considering (possibly as a follow-up, not a blocker): guardContent is only meaningful when a guardrail is actually attached. If a caller sets guardContent: true on a part but never provides providerOptions.bedrock.guardrailConfig, Bedrock can reject the request with a validation error. Since the presence of guardrailConfig is known client-side at getArgs time, the SDK could detect that combination and either emit a warning + downgrade the part to a plain { text } block, or surface a clear error, rather than letting it fail server-side with an opaque message.

laygir:
Looking forward to have this merged!

aayush-kapoor (approved):

PR #14337, opened 2026-04-10
feat(ai): add new 'InvalidToolCall' part

aayush-kapoor:
## Background

there was a TODO comment talking about cleaning up the way we pass invalid tool calls (when incorrect json args are returned by the model) 

previously, invalid tool calls were emitted as dynamic tool calls (even if the tool wasn't dynamic) and then a dummy `tool-error` output was passed in place of invalid tool calls so the model would get a matching tool result in the next step.

## Summary

- new tool call type `InvalidToolCall` 
- now `to-response-messages.ts` detects `InvalidToolCall` content parts and auto-generates error tool-results for the model, so generateText and streamText don't need to manufacture fake outputs
- `InvalidToolUIPart` is its own UI part type with type: 'invalid-tool' and state: 'error', so the UI can render invalid tool calls distinctly from real execution errors or dynamic tools.
- the stop condition loop for generate text and stream text is changed since invalid tool calls aren't treated as client tool calls (filtered out) and will not produce an output!

## Manual Verification

tested by running the example `http://localhost:3001/chat/test-invalid-tool-call` before and after the fix to ensure it works as expected

## Checklist


packages/ai/src/generate-text/generate-text.test.ts
```diff
@@ -6745,6 +6728,130 @@ describe('generateText', () => {
         `);
       });
     });
+
+    describe('invalid tool call alongside valid tool call should not break the loop', () => {
```
aayush-kapoor, line 6732: this test simulates behaviour from `main`. lmk if i should separate this test out in a diff pr so that the changes remain easier to track

packages/ai/src/generate-text/to-response-messages.test.ts
```diff
@@ -978,18 +979,15 @@ describe('toResponseMessages', () => {
...
-          error: new Error('JSON parsing failed'),
-        },
-        {
-          type: 'tool-error',
-          toolCallId: 'call-1',
-          toolName: 'weather',
-          input: '{ city: San Francisco, }',
-          error: 'Invalid input for tool weather: JSON parsing failed',
-          dynamic: true,
+          error: new InvalidToolInputError({
```
aayush-kapoor, line 986: the error type is changed since now `to-response-messages` will always get a specific error type (eg. `InvalidToolInputError`).

simply passing an `Error` will lead to value of `error-text` being "JSON parsing failed" and not the entire error

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -1018,10 +1002,12 @@ export async function generateText<
...
       // Continue if:
       // 1. There are client tool calls that have all been executed, OR
-      // 2. There are pending deferred results from provider-executed tools
+      // 2. There are pending deferred results from provider-executed tools, OR
+      // 3. There are invalid tool calls (error feedback is sent to the model via toResponseMessages)
       ((clientToolCalls.length > 0 &&
         clientToolOutputs.length === clientToolCalls.length) ||
-        pendingDeferredToolCalls.size > 0) &&
+        pendingDeferredToolCalls.size > 0 ||
+        hasInvalidToolCalls) &&
```
aayush-kapoor, line 1010: loop stop condition is changed - since invalid tool calls are now excluded from `clientToolCalls`, the condition `clientToolCalls.length > 0` will fail and the model won't respond even for a single failed tool call and won't be able to correct itself

packages/ai/src/generate-text/execute-tool-call.ts
```diff
@@ -138,7 +138,7 @@ export async function executeToolCall<
               type: 'tool-result',
               output: part.output,
               preliminary: true,
-            });
+            } as TypedToolResult<TOOLS>);
```
gr2m, line 141: is this needed?

packages/ai/src/generate-text/to-response-messages.test.ts
```diff
@@ -1020,7 +1018,7 @@ describe('toResponseMessages', () => {
             {
               "output": {
                 "type": "error-text",
-                "value": "Invalid input for tool weather: JSON parsing failed",
+                "value": "AI_InvalidToolInputError: Invalid input for tool weather: Error: JSON parsing failed",
```
gr2m, line 1021: Could the `Error: ` part be removed?

gr2m (approved):

packages/ai/src/generate-text/execute-tool-call.ts
aayush-kapoor: yeah, since we expanded `TypedToolCall` to be an invalid tool call also, the shape coming in from the spread operator can also be `invalidToolCall` which is not a valid tool result shape

packages/ai/src/generate-text/to-response-messages.test.ts
aayush-kapoor: not exactly - i changed the flow a bit of how the errors are reconstructed and this was the closest i could get to for the error message without disrupting other tests/causing other regressions

packages/ai/src/generate-text/execute-tool-call.ts
lgrammel: there is a way where we tighten the typing in some other places i think

packages/ai/src/ui/ui-messages.ts
```diff
@@ -468,6 +469,25 @@ export type DynamicToolUIPart = {
...
+export type InvalidToolUIPart = {
+  type: 'invalid-tool';
+
+  toolName: string;
+  toolCallId: string;
+  title?: string;
+  providerExecuted?: boolean;
+
+  state: 'error';
+  input: unknown;
```
lgrammel, line 485: not sure we will have input

packages/ai/src/ui/ui-messages.ts
```diff
@@ -468,6 +469,25 @@ export type DynamicToolUIPart = {
...
+ * A tool call that failed validation: either the tool does not exist
+ * or the model produced input that does not match the tool's schema.
+ */
+export type InvalidToolUIPart = {
+  type: 'invalid-tool';
+
+  toolName: string;
+  toolCallId: string;
+  title?: string;
+  providerExecuted?: boolean;
```
lgrammel, line 482: can this happen? or `never`?

PR #14365, opened 2026-04-12
fix(amazon-bedrock): resolve globalThis.fetch lazily to support telemetry and fetch patching

heiwen:
## Background

`createAmazonBedrock()` calls either `createSigV4FetchFunction` or `createApiKeyFetchFunction` at provider initialization time. Both functions accepted the underlying fetch implementation via a default parameter:

\`\`\`ts
fetch: FetchFunction = globalThis.fetch
\`\`\`

JavaScript default parameters are evaluated at call time — meaning \`globalThis.fetch\` is read and captured the moment \`createAmazonBedrock()\` is called. Any patch to \`globalThis.fetch\` applied after that point is silently ignored for all Bedrock requests.

The most common real-world impact is **telemetry instrumentation** (e.g. OpenTelemetry, Datadog, custom tracing), which typically patches \`globalThis.fetch\` at startup to intercept outgoing HTTP requests. Because Bedrock snapshots fetch before those patches are applied, all Bedrock requests bypass instrumentation. Test setups that patch \`globalThis.fetch\` after imports are equally affected.

Every other provider in the SDK avoids this by deferring the \`globalThis.fetch\` lookup to request time. The comment in \`http-chat-transport.ts\` captures the intent explicitly: *"avoid caching globalThis.fetch in case it is patched by other libraries"*.

## Summary

This PR makes \`createSigV4FetchFunction\` and \`createApiKeyFetchFunction\` resolve \`globalThis.fetch\` lazily inside the returned closure instead of at function call time.

Changes included:
- Removed the \`= globalThis.fetch\` default parameter from both functions, making \`fetch\` optional
- Added \`const effectiveFetch = fetch ?? globalThis.fetch\` inside each returned closure, so the lookup happens at request time
- Added a patch changeset for \`@ai-sdk/amazon-bedrock\`

## Manual Verification

I manually verified by inspecting the call path from \`createAmazonBedrock()\` through to the SigV4 and API key fetch wrappers:

- confirmed the default parameter no longer captures \`globalThis.fetch\` at provider initialization
- confirmed each request-time closure now resolves \`globalThis.fetch\` fresh on every call, consistent with how \`post-to-api.ts\` and \`http-chat-transport.ts\` handle this

## Checklist

- [x] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
[...]

PR #14366, opened 2026-04-12
fix(amazon-bedrock): resolve globalThis.fetch lazily to preserve telemetry patches

### LINKED ISSUE #14364: Bug: `createAmazonBedrock()` breaks telemetry by capturing `fetch` at initialization time
## Problem

`createSigV4FetchFunction` and `createApiKeyFetchFunction` both use a JS default parameter:

```ts
fetch: FetchFunction = globalThis.fetch
```

This evaluates `globalThis.fetch` the moment `createAmazonBedrock()` is called. Any patch to `globalThis.fetch` applied after provider initialization is silently ignored.

## Expected Behavior

Match the pattern used by every other provider in the SDK — read `globalThis.fetch` lazily at request time via `getOriginalFetch()` in `post-to-api.ts`, so patches (e.g. from test frameworks, polyfills, or middleware) are always respected.

## Impact

The most common real-world case is **telemetry instrumentation** (e.g. OpenTelemetry, Datadog, custom tracing) which patches `globalThis.fetch` at startup to intercept outgoing HTTP requests. Because Bedrock snapshots fetch before those patches are applied, all Bedrock requests silently bypass instrumentation. Test setups that patch `globalThis.fetch` after imports and runtime environments that install a fetch polyfill after module load are equally affected.

## Affected Files

- `packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts` — `createSigV4FetchFunction` (line 27) and `createApiKeyFetchFunction` (line 116)

## Proposed Fix

Remove the default parameter and resolve `globalThis.fetch` lazily inside the returned closure, matching the pattern used in `http-chat-transport.ts` (which carries the comment *"avoid caching globalThis.fetch in case it is patched by other libraries"*):

```ts
// before
export function createSigV4FetchFunction(
  getCredentials: ...,
  fetch: FetchFunction = globalThis.fetch,
): FetchFunction {
  return async (input, init) => {
    // fetch is already bound to the captured value
    return fetch(input, { ... });
  };
}

// after
export function createSigV4FetchFunction(
  getCredentials: ...,
  fetch?: FetchFunction,
): FetchFunction {
  return async (input, init) => {
    // avoid caching globalThis.fetch in case it is patched by other libraries
    const effectiveFetch = fetch ?? globalThis.fetch;
    return effectiveFetch(input, { ... });
  };
}
```

Apply the same change to `createApiKeyFetchFunction`.

### AI SDK Version

_No response_

### Code of Conduct

Yanhu007:
## Summary

Fixes #14364

`createSigV4FetchFunction` and `createApiKeyFetchFunction` both capture `globalThis.fetch` via a JS default parameter, evaluated at provider initialization time:

```ts
fetch: FetchFunction = globalThis.fetch  // captured once, at init
```

Any patch applied after initialization (OpenTelemetry, Datadog, test frameworks) is silently ignored. All Bedrock requests bypass telemetry instrumentation.

## Fix

Change both functions to accept an optional `fetch?` parameter and resolve `globalThis.fetch` lazily inside the returned closure:

```ts
const effectiveFetch = fetch ?? globalThis.fetch;  // resolved per-request
```

This matches the pattern used in `http-chat-transport.ts` (which carries the comment *"avoid caching globalThis.fetch in case it is patched by other libraries"*).

## Changes

- `createSigV4FetchFunction`: `fetch: FetchFunction = globalThis.fetch` → `fetch?: FetchFunction` + lazy resolve (2 call sites)
- `createApiKeyFetchFunction`: same pattern (1 call site)

### DIFF
--- a/packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
+++ b/packages/amazon-bedrock/src/bedrock-sigv4-fetch.ts
@@ -24,12 +24,16 @@ export interface BedrockCredentials {
  */
 export function createSigV4FetchFunction(
   getCredentials: () => BedrockCredentials | PromiseLike<BedrockCredentials>,
-  fetch: FetchFunction = globalThis.fetch,
+  fetch?: FetchFunction,
 ): FetchFunction {
   return async (
     input: RequestInfo | URL,
     init?: RequestInit,
   ): Promise<Response> => {
+    // Resolve fetch lazily to avoid caching globalThis.fetch at
+    // initialization time.  Telemetry libraries (OpenTelemetry, Datadog)
+    // and test frameworks often patch globalThis.fetch after imports.
+    const effectiveFetch = fetch ?? globalThis.fetch;
     const request = input instanceof Request ? input : undefined;
     const originalHeaders = combineHeaders(
       normalizeHeaders(request?.headers),
@@ -51,7 +55,7 @@ export function createSigV4FetchFunction(
     const effectiveMethod = init?.method ?? request?.method;
 
     if (effectiveMethod?.toUpperCase() !== 'POST' || !effectiveBody) {
-      return fetch(input, {
+      return effectiveFetch(input, {
         ...init,
         headers: headersWithUserAgent as HeadersInit,
       });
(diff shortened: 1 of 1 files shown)


aayush-kapoor (commented):
marking as closed in favor of https://github.com/vercel/ai/pull/14365

Issue #14373, opened 2026-04-13
[Bug] smoothStream drops providerMetadata from chunked stream parts


kuishou68:
### Description

In `packages/ai/src/generate-text/smooth-stream.ts`, the `smoothStream` transform silently drops `providerMetadata` when it detects word/line boundaries and emits buffered chunks via the chunking while loop.

### Root Cause

The `while` loop that emits chunked content does **not** include `providerMetadata` in the enqueued stream parts:

```typescript
// Lines 153-158 in smooth-stream.ts
while ((match = detectChunk(buffer)) != null) {
  controller.enqueue({ type, text: match, id }); // providerMetadata is dropped!
  buffer = buffer.slice(match.length);
  await delay(delayInMs);
}
```

However, the `flushBuffer` function **does** correctly include `providerMetadata`:

```typescript
// Lines 117-122 in smooth-stream.ts
controller.enqueue({
  type,
  text: buffer,
  id,
  ...(providerMetadata != null ? { providerMetadata } : {}), // included here
});
```

### Impact

When `smoothStream` is used with providers that embed metadata in streaming deltas (e.g., Anthropic extended thinking / reasoning blocks which carry `providerMetadata` with thinking signatures), that metadata is silently dropped for every chunk emitted from the chunking `while` loop. Only the very last remaining content (emitted via `flushBuffer`) retains the metadata.

This means:
- Anthropic extended thinking signatures are lost from intermediate stream parts
- Any provider-specific metadata attached to `text-delta` or `reasoning-delta` chunks is silently dropped when `smoothStream` is active

### Steps to Reproduce

```typescript
import { streamText } from "ai";
import { anthropic } from "@ai-sdk/anthropic";
import { smoothStream } from "ai";

const result = streamText({
  model: anthropic("claude-3-7-sonnet-20250219", {
    thinking: { type: "enabled", budgetTokens: 5000 },
  }),
  prompt: "Explain step by step how addition works.",
  experimental_transform: smoothStream({ chunking: "word" }),
});

for await (const part of result.fullStream) {
  if (part.type === "reasoning-delta") {
    // providerMetadata (thinking signatures) will be undefined for most chunks
    console.log(part.providerMetadata); // undefined for word-boundary chunks
  }
}
```

### Expected Behavior

All emitted stream parts from `smoothStream` should preserve `providerMetadata` from the original chunk that contained it.

### Actual Behavior

[...]

PR #14374, opened 2026-04-13
fix: preserve providerMetadata in smoothStream chunked output

kuishou68:
## Problem\n\nsmooth-stream.ts silently drops providerMetadata from stream parts emitted via the chunking while loop. The while loop that emits word/line-boundary chunks does not include providerMetadata, but the flushBuffer function does, causing inconsistent behavior.\n\n## Impact\n\nWhen smoothStream is used with providers that attach metadata to streaming delta chunks (e.g., Anthropic extended thinking with providerMetadata containing thinking block signatures), that metadata is silently dropped for every chunk emitted via the chunking loop. Only the final flushed chunk retains the metadata.\n\n## Fix\n\nInclude providerMetadata in the controller.enqueue call within the while loop, consistent with flushBuffer behavior.\n\nCloses #14373

### DIFF
--- a/packages/ai/src/generate-text/smooth-stream.ts
+++ b/packages/ai/src/generate-text/smooth-stream.ts
@@ -151,7 +151,12 @@ export function smoothStream<TOOLS extends ToolSet>({
         let match;
 
         while ((match = detectChunk(buffer)) != null) {
-          controller.enqueue({ type, text: match, id });
+          controller.enqueue({
+            type,
+            text: match,
+            id,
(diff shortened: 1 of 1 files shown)


nehaaprasad:
### AI Code Trust — ship readiness

**Score:** 100 · **Verdict:** SAFE · **Model:** deterministic-v1

No blocking issues detected by automated checks.

**Top issues:**
- **maintainability** (low) — Very long line reduces readability. `packages/ai/src/generate-text/smooth-stream.ts:25`

PR #14407, opened 2026-04-13
chore (gateway): add reranking to model settings generation config

R-Taneja:
## Summary
Add `reranking` to `MODALITY_CONFIG` in the `generate-model-settings` script so the automated GH Action workflow correctly generates `gateway-reranking-model-settings.ts` when reranking models appear in the gateway API.

The fallback logic at line 79 would also work, but being explicit matches the pattern for language, embedding, image, and video.

## Test plan
- [x] `pnpm generate-model-settings` runs without error (no reranking models in prod yet, so no output change)


shaper (approved):

ghost:
✅ Backport PR created: https://github.com/vercel/ai/pull/14408

PR #14419, opened 2026-04-14
fix(mistral): map cached token counts from Mistral usage response

### LINKED ISSUE #14382: [Bug] Mistral provider: all tokens reported as `noCacheTokens`
The Mistral provider incorrectly reports every token as `noCacheTokens`:

https://github.com/vercel/ai/blob/6e706db5b98079311e85e8f77022602cb88ab58c/packages/mistral/src/convert-mistral-usage.ts#L29-L45

According to [Mistral API Specs](https://docs.mistral.ai/api?property=operation-chat_completion_v1_chat_completions_post_responses_200_application-json_usage#operation-chat_completion_v1_chat_completions_post_responses_200_application-json_usage), there is a `num_cached_tokens` property, that should likely be used?

In addition, there is `prompt_token_details.cached_tokens` and `prompt_tokens_details.cached_tokens` ("token**s**") - the documentation just says "Token usage details for the prompt" for bot of those properties, so I'm not sure what these are for, or why there are several of these.

Either way, Mistral does support caching, and cache usage is reported, so the token count for `noCacheTokens` is currently wrong.

Christian-Sidak:
## Summary
- Mistral's API returns cached token counts via `num_cached_tokens`, `prompt_tokens_details.cached_tokens`, and `prompt_token_details.cached_tokens`
- These fields were not included in `mistralUsageSchema` (stripped by Zod) and not mapped in `convertMistralUsage`
- All prompt tokens were incorrectly reported as `noCacheTokens`
- Adds the fields to the schema and computes `cacheRead`/`noCache` correctly

Closes #14382

## Test plan
- [x] Unit test with cached token fields in usage response
- [x] Verify noCache = promptTokens - cacheReadTokens
- [x] Verify cacheRead is populated when cached tokens present
- [x] Verify backward compatibility when cached fields are absent


packages/mistral/src/convert-mistral-usage.ts
```diff
@@ -29,11 +32,17 @@ export function convertMistralUsage(
   const promptTokens = usage.prompt_tokens;
   const completionTokens = usage.completion_tokens;
 
+  const cacheReadTokens =
```
mindplay-dk, line 35: what's the logic here? just wondering. :-)

Christian-Sidak: Mistral exposes cached token counts in three different places depending on the API version/endpoint: the top-level `num_cached_tokens` field, `prompt_tokens_details.cached_tokens`, and `prompt_token_details.cached_tokens` (note the singular/plural inconsistency — both exist in the wild). The `??` chain tries each in priority order and falls back to `0` when none are present, so the `noCache = promptTokens - cacheReadTokens` math works in all cases.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

packages/mistral/src/convert-mistral-usage.ts
mindplay-dk: Makes sense. 👍

mindplay-dk:
Sorry for the late reply, I missed the notification.

Looks good to me! Thanks for your work on this. 😊

lgrammel:
we have changed our security policy to allow only commits with verified signatures. Would it be possible for you to enable commit signing and open a new PR with signed commits?

Christian-Sidak:
Thanks for the heads up, @lgrammel. I'll enable GPG commit signing on my end and open a new PR with verified commits. Will link it here once it's up.

Christian-Sidak:
Opened #14889 with GPG-signed commits as requested. Closing this PR.

Christian-Sidak:
Superseded by #14889 (signed commits).

PR #14420, opened 2026-04-14
fix(openai): preserve reasoning state across reasoning item id rotation

choi138:
## Background

The OpenAI Responses stream parser currently tracks active reasoning state by `item.id` alone. Some OpenAI-compatible providers rotate reasoning item ids between `response.output_item.added` and `response.output_item.done` while keeping the same `output_index`.

That causes a runtime crash when the done event arrives:

```txt
TypeError: undefined is not an object (evaluating 'activeReasoningPart.summaryParts')
```

## Summary

- preserve active reasoning state by `output_index` as a fallback when the reasoning `item.id` changes between events
- keep emitted `reasoning-start` / `reasoning-end` ids anchored to the canonical id from the `added` event
- guard `response.reasoning_summary_part.added` and `response.reasoning_summary_part.done` against missing active reasoning state instead of throwing
- add a regression test for reasoning id rotation between `response.output_item.added` and `response.output_item.done`

## Manual Verification

I reproduced the crash with a Responses stream shaped like this:

```txt
response.output_item.added  output_index=0 item.id=rs_reasoning_original
response.output_item.done   output_index=0 item.id=rs_reasoning_rotated
```

Before this patch, the stream crashed in the OpenAI Responses parser. After this patch, the stream completes and emits matching `reasoning-start` / `reasoning-end` events using the canonical id from the `added` event.

## Checklist


rotonott:
@choi138 So you don't get the errors you mentioned with this PR?

choi138:
@rotonott Yes. Before this fix, I could reproduce the TypeError I mentioned when the reasoning item ID changed between events. With this PR applied, I can no longer reproduce that crash.

PR #14423, opened 2026-04-14
tmp

lgrammel:
tied to https://github.com/vercel/ai/pull/14337

### DIFF
--- a/packages/provider-utils/src/types/content-part.ts
+++ b/packages/provider-utils/src/types/content-part.ts
@@ -189,6 +189,27 @@ export interface ToolCallPart {
   providerExecuted?: boolean;
 }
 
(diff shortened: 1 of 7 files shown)


aayush-kapoor (approved):

packages/provider-utils/src/types/content-part.ts
```diff
@@ -189,6 +189,27 @@ export interface ToolCallPart {
...
+
+  /**
+   * Name of the tool that is being called.
+   */
+  toolName: string;
+
+  /**
+   * Arguments of the tool call. This is a JSON-serializable object that matches the tool's input schema.
+   */
+  rawInput: unknown;
```
lgrammel, line 208: could be a string that cannot even be parsed

PR #14479, opened 2026-04-15
fix: only exempt approved tool calls from MissingToolResultsError

### LINKED ISSUE #14428: streamText: deny approval crashes with "Missing required parameter: input[N].output"
## Description

When using `streamText` with `needsApproval: true` on a tool and the user **denies** the tool call via `addToolApprovalResponse({ id, approved: false })`, the next request crashes with:

```
Missing required parameter: 'input[N].output'
```

**Approving (Yes) works perfectly. Only denying (No) crashes.**

## Reproduction

Minimal Next.js project: two tools (one read, one write with `needsApproval: true`). The model calls both in the same step. Read auto-executes, write gets `approval-requested`. User denies → second POST → crash.

### Server (`app/api/chat/route.ts`)

```typescript
import { openai } from "@ai-sdk/openai";
import { convertToModelMessages, stepCountIs, streamText, tool } from "ai";
import { z } from "zod";

const searchMemory = tool({
  description: "Search memory",
  inputSchema: z.object({ query: z.string() }),
  execute: async ({ query }) => ({ resultCount: 0, query }),
});

const sendEmail = tool({
  description: "Send an email",
  inputSchema: z.object({ to: z.string(), subject: z.string(), body: z.string() }),
  needsApproval: true,
  execute: async ({ to, subject, body }) => ({ success: true, to, subject }),
});

  const tools = { searchMemory, sendEmail };

  const modelMessages = await convertToModelMessages(messages, {
    tools,
    ignoreIncompleteToolCalls: true,
  });

  const result = streamText({
    model: openai("gpt-4o"),
    system: "You have searchMemory and sendEmail. Call both when asked to send email.",
    messages: modelMessages,
    tools,
    stopWhen: stepCountIs(10),
  });

}
```

### Client (`app/page.tsx`)

Standard `useChat` + `DefaultChatTransport` + `sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithApprovalResponses` + `addToolApprovalResponse({ id, approved: false })`.

### Steps

1. User sends "send an email to test@test.com saying hello"
2. Model calls `searchMemory` (auto-executes) + `sendEmail` (gets `approval-requested`)
3. Client shows approval UI, user clicks **Deny**
4. `addToolApprovalResponse({ id, approved: false })` → `sendAutomaticallyWhen` fires
5. Second POST with messages containing `approval-responded` + `approved: false`

### curl reproduction

```bash
# Step 1: works fine
curl -s -X POST http://localhost:3000/api/chat \
  -H "Content-Type: application/json" \
  -d '{"messages":[{"id":"msg-1","role":"user","parts":[{"type":"text","text":"send an email to test@test.com saying hello"}]}]}'

# Step 2: crashes (use toolCallId and approvalId from step 1)
curl -s -X POST http://localhost:3000/api/chat \
  -H "Content-Type: application/json" \
  -d '{"messages":[
    {"id":"msg-1","role":"user","parts":[{"type":"text","text":"send an email to test@test.com saying hello"}]},
    {"id":"msg-2","role":"assistant","parts":[
      {"type":"tool-searchMemory","toolCallId":"call_XXX","state":"output-available","input":{"query":"contact test@test.com"},"output":{"resultCount":0}},
      {"type":"tool-sendEmail","toolCallId":"call_YYY","state":"approval-responded","input":{"to":"test@test.com","subject":"Hello","body":"Hello"},"approval":{"id":"APPROVAL_ID","approved":false,"reason":"User denied"}}
    ]}
  ]}'
```

### Output

```
data: {"type":"tool-output-denied","toolCallId":"call_YYY"}
data: {"type":"error","errorText":"Missing required parameter: 'input[5].output'."}
data: [DONE]
```

The SDK correctly emits `tool-output-denied`, then crashes when calling the model for the continuation step.

## Investigation

We traced through the SDK source (`ai@6.0.154`):

1. `convertToModelMessages` correctly creates `tool-approval-response` for denied tools (line ~8575)
2. `convertToLanguageModelPrompt` correctly reads `approvedToolCallIds` from `tool-approval-response` parts (line ~1425)
3. But `convertToLanguageModelPrompt` also **filters out** non-provider-executed `tool-approval-response` parts from the tool message (line ~1607):
   ```javascript
   (part) => part.type !== "tool-approval-response" || part.providerExecuted
   ```
4. `collectToolApprovals` in `streamText` processes the denied approval and creates a `tool-result` with `{ type: "execution-denied" }` in `initialResponseMessages`
5. These `initialResponseMessages` are added to `stepInputMessages` for the model call
6. But the OpenAI Responses API (`input[N].output`) still rejects, suggesting the denied tool result doesn't reach the model in the correct format

### Both patterns fail

| Pattern | Result |
|---------|--------|
| `convertToModelMessages()` → `streamText({ messages: modelMessages })` | `input[N].output` error on deny |
| Raw messages → `streamText({ messages })` | `Invalid prompt: messages do not match ModelMessage[] schema` |

## Expected Behavior

After denying a tool call, the model should receive the denial result and continue the conversation (e.g., suggest an alternative).

## Versions

- `ai`: 6.0.154
- `@ai-sdk/openai`: 2.0.65
- `@ai-sdk/react`: 3.0.156
- `next`: 15.5.15
- Node.js: 22.20.0

aayushbaluni:
## Summary

Fixes #14428.

**Root cause:** `approvedToolCallIds` was populated from all `tool-approval-response` parts regardless of their `approved` value. Denied tool calls (`approved: false`) were incorrectly added to the exemption set, causing their tool call IDs to be removed from the `toolCallIds` validation set. The provider then received a `function_call` without a matching `function_call_output`, triggering `Missing required parameter: 'input[N].output'`.

**Fix:** Added `part.approved === true` guard so only genuinely approved tool calls are exempted from the `MissingToolResultsError` check. Denied tool calls now correctly remain in the validation set and must have a corresponding tool result (e.g., `execution-denied`) before the prompt is sent to the provider.

## Changes

- `packages/ai/src/prompt/convert-to-language-model-prompt.ts`: Added `part.approved === true` condition when building `approvedToolCallIds`

## Testing

- Denied tool calls now correctly trigger `MissingToolResultsError` when no tool result is present
- Approved tool calls continue to be exempted as before
- Provider-executed tool approval responses are unaffected (handled by separate filter)

### DIFF
--- a/packages/ai/src/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/src/prompt/convert-to-language-model-prompt.ts
@@ -75,7 +75,10 @@ export async function convertToLanguageModelPrompt({
   for (const message of prompt.messages) {
     if (message.role === 'tool') {
       for (const part of message.content) {
-        if (part.type === 'tool-approval-response') {
+        if (
+          part.type === 'tool-approval-response' &&
+          part.approved === true
+        ) {
           const toolCallId = approvalIdToToolCallId.get(part.approvalId);
           if (toolCallId) {
             approvedToolCallIds.add(toolCallId);


rpelevin:
This is the right direction: denied approvals should not enter the same exemption set as approved tool calls.

The regression I would pin is slightly broader than the map guard itself. The important invariant is not only that a denied approval remains visible to missing-result validation. It is that the denied path becomes terminal before the next provider request is built.

I would cover both halves:

1. An approval response with approved true can exempt the original tool call only if the runtime has a path to execute it and produce a terminal result before provider continuation.
2. An approval response with approved false must never be treated as approved authority, and must produce one terminal no-execution result for the original toolCallId before provider continuation.

A validation-only test is useful, but it can still leave the product behavior as "deny, then crash." The stronger fixture is: deny a needsApproval tool call, continue the chat, assert the provider input does not contain a bare tool call, and assert exactly one denial result is bound to the same toolCallId while the tool execute function is not called.

That keeps denial as a first-class terminal outcome rather than a missing-output error path.

Boundary: architecture and regression-test feedback only; no claim about using this project or running its code.

lgrammel:
Thank you for the pull request. Issue #14428 was closed as not planned. Closing this pull request.

PR #14512, opened 2026-04-16
fix(anthropic): handle server tool blocks in message_start

oskrim:
## Background
Observed that Anthropic programmatic tool calling can return server tool calls `message_start.content`

## Summary
- accept content blocks in `message_start.content`

## Manual Verification
- The issue is not reliably reproducible the failing content blocks were received during a real world conversation

## Related Issues
- None


oskrim:
Closing in favor of the newer message_start content block fix.

PR #14575, opened 2026-04-16
fix(provider/google): include toolUsePromptTokenCount in input token usage

JamieWoodbury:
## Background

When you call Gemini with a tool that runs a tool-use prompt — most commonly grounded `google_search` or `url_context` — the model's `usageMetadata` reports those tokens in a separate field, `toolUsePromptTokenCount`, on top of `promptTokenCount`. Both are billed at the regular input rate. Google's docs: <https://ai.google.dev/api/generate-content#UsageMetadata>.

Today the `@ai-sdk/google` provider drops this field on the floor:

1. `usageSchema` in `google-generative-ai-language-model.ts` doesn't declare `toolUsePromptTokenCount`, so zod strips it from the parsed response.
2. `convertGoogleGenerativeAIUsage` only sums `promptTokenCount` into `inputTokens.total`.

## Impact

For a grounded call this means the standard `usage.inputTokens.total` is silently lower than what Google actually bills. We hit this in production: a grounded Gemini 2.5 Flash call returned

```jsonc
"usageMetadata": {
  "promptTokenCount":         70,
  "candidatesTokenCount":     53,
  "thoughtsTokenCount":      199,
  "toolUsePromptTokenCount":  79,   // ← billed, but invisible to consumers
  "totalTokenCount":         401
}
```

The AI SDK surfaced `inputTokens.total = 70`, leaving 79 tokens (~20% of the call's billable input) unaccounted for. Anyone computing cost from the standard `LanguageModelV4Usage` shape silently undercounts.

## Fix

- Declare `toolUsePromptTokenCount` on `usageSchema` so zod stops stripping it.
- Add the field to `GoogleGenerativeAIUsageMetadata`.
- Sum `toolUsePromptTokenCount` into `inputTokens.total` and `inputTokens.noCache` in `convertGoogleGenerativeAIUsage`. Cached read tokens already cap `noCache` correctly.
- For non-grounded calls the value is missing/zero and the math collapses to the existing behavior — no behavior change.

## Tests

Added a focused `convert-google-generative-ai-usage.test.ts` covering:
- null usage (unchanged behavior)
- basic prompt + candidate mapping
- cached content interaction with `noCache`
- thoughts (reasoning) tokens
- `toolUsePromptTokenCount` in input total (new)
- combined cached + tool-use prompt tokens (new)

`pnpm test:node` in `packages/google` passes all 400 tests (was 394).

The existing inline-snapshot test that asserts on `usage` (`should support includeThoughts with google generative ai provider`) does not include `toolUsePromptTokenCount` in its fixture, so its snapshot is unchanged.

## Notes

- Patch changeset added per `CONTRIBUTING.md`.
- The schema and converter are the only two places `toolUsePromptTokenCount` needs to flow through. The `LanguageModelV4Usage` type has no dedicated tool-use slot, so summing into `inputTokens.total` is the most spec-aligned destination — that's where Google bills it, and that's what consumers price against.
- Happy to also surface a per-field `toolUsePrompt` count on `LanguageModelV4Usage.inputTokens` (or via `raw`) in a follow-up if maintainers want callers to be able to break it out.

### DIFF
--- a/.changeset/google-tool-use-prompt-tokens.md
+++ b/.changeset/google-tool-use-prompt-tokens.md
@@ -1,0 +1,19 @@
+---
+"@ai-sdk/google": patch
+---
+
+fix(provider/google): include `toolUsePromptTokenCount` in input token usage
+
+Google Gemini reports tokens consumed by tool-use prompts (e.g. grounded
+Google Search, URL context) in a separate `toolUsePromptTokenCount` field
+in `usageMetadata`. These tokens are billed at the regular input rate but
+were previously dropped: they were not declared in `usageSchema` and the
+converter only summed `promptTokenCount` into `inputTokens.total`.
+
+The result was that grounded calls under-reported input tokens by the
+exact amount Google billed for the tool-use prompts (in one repro: 79
+of 401 total tokens, ~20%, were missing from `usage.inputTokens`).
+
+This change adds `toolUsePromptTokenCount` to the schema and includes it
+in `inputTokens.total` and `inputTokens.noCache`. Non-grounded calls are
+unaffected.
--- a/packages/google/src/convert-google-generative-ai-usage.test.ts
+++ b/packages/google/src/convert-google-generative-ai-usage.test.ts
@@ -1,0 +1,114 @@
+import { describe, expect, it } from 'vitest';
+import { convertGoogleGenerativeAIUsage } from './convert-google-generative-ai-usage';
+
+describe('convertGoogleGenerativeAIUsage', () => {
+  it('returns undefined fields when usage is null', () => {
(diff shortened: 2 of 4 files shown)


samjbobb:
We hit this in production and opened #16400 to carry your fix forward, since the package was refactored since this PR (`convert-google-generative-ai-usage.ts` → `convert-google-usage.ts`, `google-generative-ai-language-model.ts` → `google-language-model.ts`), so this branch no longer applies cleanly. #16400 rebases the same change onto current `main` and adds a changeset + unit tests — full credit to you for the original diagnosis. Hope that helps get it landed; happy to defer if you'd rather refresh this one instead.

PR #14585, opened 2026-04-17
fix: guard against undefined warnings from EmbeddingModelV2 providers

### LINKED ISSUE #14425: embed/embedMany crash: 'warnings' is undefined from EmbeddingModelV2 providers

Christian-Sidak:
## Summary

Fixes #14425

`embed()` and `embedMany()` crash at runtime when used with `EmbeddingModelV2` providers (e.g. `@ai-sdk/google-vertex@3.x`) because `doEmbed` does not return a `warnings` property but the SDK assumes it is always present and iterable.

### Changes

- **`embed.ts` / `embed-many.ts`**: Default `modelResponse.warnings` to `[]` via nullish coalescing (`?? []`) at the source where the value is destructured from the model response
- **`log-warnings.ts`**: Add null guard (`!options.warnings ||`) before accessing `.length` so `logWarnings` handles undefined gracefully
- **`embed-many.ts` (chunked path)**: Wrap `warnings.push(...result.warnings)` in a null check to prevent "is not iterable" error

### Errors fixed

**`embed()`** -- `logWarnings` at `log-warnings.ts`:
```
```

**`embedMany()`** -- spreading `result.warnings`:
```
```

## Test plan

- Added test in `log-warnings.test.ts`: calling `logWarnings` with `undefined` warnings does not throw
- Added test in `embed.test.ts`: `embed()` with a provider that omits `warnings` returns `[]` 
- Added tests in `embed-many.test.ts`: `embedMany()` with undefined warnings works for both single-call and chunked paths

### DIFF
--- a/.changeset/fix-embed-undefined-warnings.md
+++ b/.changeset/fix-embed-undefined-warnings.md
@@ -1,0 +1,7 @@
+---
+'ai': patch
+---
+
+fix: guard against undefined warnings from EmbeddingModelV2 providers in embed/embedMany
+
+`EmbeddingModelV2` providers (e.g. `@ai-sdk/google-vertex@3.x`) do not include `warnings` in their `doEmbed` return type, causing `embed()` and `embedMany()` to crash at runtime. This adds null-safety by defaulting `modelResponse.warnings` to `[]` and adding guards in `logWarnings` and the `embedMany` chunked aggregation path.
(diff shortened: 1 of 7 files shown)


Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Closing due to inactivity. Happy to reopen if there's interest.

PR #14599, opened 2026-04-17
fix: remove _internal.now

lgrammel:
## Background

The `now()` function can be mocked directly with vitest. Inject `now` via `_internal` is no longer needed.

## Summary

Remove `_internal.now` on `streamText` and `streamObject`. Remove dead test code.


lgrammel:
Cross-package support for otel needs dependency injection.

PR #14614, opened 2026-04-17
feat(otel): add a step level span

aayush-kapoor:
## Background

the telemetry spans needed more visibility on the steps that were running + needed to be aligned with genAI semantic conventions.

## Summary

`GenAIOpenTelemetry` now creates an explicit agent_step span for each generateText / streamText step, with both chat and execute_tool parented under that step span

## Manual Verification

ran the example `examples/ai-functions/src/generate-text/anthropic/subagent-with-telemetry.ts` before and after the change

### Before

image

### After

image

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Future Work

the duration of the `chat` span might need new events so that the model call's duration can be calculated accurately

PR #14646, opened 2026-04-20
feat(ai): add a ModelCall start/end event

aayush-kapoor:
## Background

the telemetry event hierarchy was changed as a part of the PR https://github.com/vercel/ai/pull/14614.

there was future work to define new event types that would limit the boundaries of each event and keep scope clean (for example, the duration of the chat span in the linked PR spans across tool execution when it shouldn't)

## Summary

new event types added `ModelCallStart` and `ModelCallEnd` that limit the scope of that event to be fired just when a model is called and to stop before a tool execution happens

## Manual Verification

haven't verified shape yet

## Checklist


packages/ai/src/generate-text/stream-text.ts
```diff
@@ -504,6 +528,19 @@ export function streamText<
       NoInfer<OUTPUT>
     >;
 
+    /**
+     * Callback that is called immediately before the provider model call begins.
+     */
+    experimental_onModelCallStart?: StreamTextOnModelCallStartCallback;
```
gr2m: if the methods are experimental, should we export the types as `Experimental_*` as well?

gr2m (approved):

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,54 @@ export interface OnStepStartEvent<
...
+
+/**
+ * Event passed to the `onModelCallEnd` callback.
+ *
+ * Called after the model response has been normalized and parsed, but before
+ * any client-side tool execution begins.
+ */
+export interface ModelCallEndEvent<
+  TOOLS extends ToolSet = ToolSet,
+> extends CallbackModelInfo {
```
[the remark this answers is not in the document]
aayush-kapoor: actually i omitted these on purpose. the `response.model` should be the same as the `request.modelID` and accounting for `response.id`added a new generateId() call which seemed odd (so removed it for now)

packages/ai/src/generate-text/stream-text.ts
aayush-kapoor: i actually wanted to ask @lgrammel if we should take them out of experimental when v7 comes out of beta

packages/ai/src/generate-text/core-events.ts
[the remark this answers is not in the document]
aayush-kapoor: added it, happy?

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -247,6 +249,26 @@ export type StreamTextOnStepStartCallback<
...
 > = Callback<OnStepStartEvent<TOOLS, RUNTIME_CONTEXT, OUTPUT>>;
 
+/**
+ * Callback that is set using the `experimental_onModelCallStart` option.
+ *
+ * Called immediately before the provider model call begins.
+ *
+ * @param event - The event object containing model-call-specific inputs.
+ */
+export type StreamTextOnModelCallStartCallback = Callback<ModelCallStartEvent>;
```
lgrammel: can we have generalized `OnModelCallStarCallbacks` in core events (see tool-execution-events for an example) instead of redefining the same callback for stream,generate,agent?

packages/ai/src/telemetry/telemetry.ts
```diff
@@ -57,6 +59,19 @@ export interface Telemetry {
    */
   onStepStart?: Callback<OnStepStartEvent>;
 
+  /**
+   * Called immediately before the provider model call begins.
+   * Unlike `onStepStart`, this callback is scoped to model work only and
+   * excludes any later client-side tool execution.
+   */
+  onModelCallStart?: Callback<ModelCallStartEvent>;
```
lgrammel: Use `OnModelCallStarCallback` once it exists

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -746,6 +802,41 @@ export async function generateText<
...
+            text: modelCallContent
+              .filter(part => part.type === 'text')
+              .map(part => part.text)
+              .join(''),
+            reasoning: modelCallContent.flatMap(part =>
+              part.type === 'reasoning' ? [{ text: part.text }] : [],
+            ),
+            files: modelCallContent
+              .filter(part => part.type === 'file')
+              .map(part => part.file),
```
lgrammel: the filter / map code might be duplicated, ie appear in other places. could be worth to extract helper functions

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -359,6 +385,19 @@ export async function generateText<
...
+     */
+    experimental_onModelCallStart?: GenerateTextOnModelCallStartCallback;
+
+    /**
+     * Callback that is called after the model response has been normalized and parsed,
+     * but before any client-side tool execution begins.
+     */
+    experimental_onModelCallEnd?: GenerateTextOnModelCallEndCallback<
+      NoInfer<TOOLS>
+    >;
```
lgrammel, line 380: ideally we should have `OnModelCallStartCallback` and `OnModelCallEndCallback` types instead

packages/ai/src/generate-text/create-execute-tools-transformation.ts
```diff
@@ -54,6 +60,12 @@ export function createExecuteToolsTransformation<TOOLS extends ToolSet>({
   LanguageModelStreamPart<TOOLS>
 > {
   const toolCallsToExecute: Array<TypedToolCall<TOOLS>> = [];
+  const modelToolCalls: Array<TypedToolCall<TOOLS>> = [];
+  const modelFiles: Array<GeneratedFile> = [];
+  const modelReasoning: Array<{ text?: string }> = [];
+  let modelText = '';
+  let responseModelId = modelId ?? '';
+  let responseId = '';
```
lgrammel: can you prefix this block with a comment why it is there? also is this duplicated? do the '' defaults make sense for modelId and responseId?

packages/ai/src/generate-text/create-execute-tools-transformation.test.ts
```diff
@@ -8,6 +8,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 import { z } from 'zod/v4';
 import { TypeValidationError } from '../error';
 import { asLanguageModelUsage } from '../types/usage';
+import { ModelCallEndEvent } from './core-events';
```
lgrammel: maybe worth introducing `language-model-events.ts` and moving the new events/callbacks there

lgrammel: (and `LanguageModel` might make more sense than `Model` in the name, we try to be more explicit in the age of coding agents, see `streamLanguageModelCall`)

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,57 @@ export interface OnStepStartEvent<
   readonly toolsContext: InferToolSetContext<TOOLS>;
 }
 
+/**
+ * Event passed to the `onModelCallStart` callback.
+ *
+ * Called immediately before the provider model call begins.
+ * Unlike `onStepStart`, this only represents model invocation work.
+ */
+export interface ModelCallStartEvent extends CallbackModelInfo {
```
lgrammel: rn to `LanguageModelCallStartEvent` (see below), same for the other event

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,57 @@ export interface OnStepStartEvent<
...
+ * any client-side tool execution begins.
+ */
+export interface ModelCallEndEvent<
+  TOOLS extends ToolSet = ToolSet,
+> extends CallbackModelInfo {
+  /** Unique identifier for this generation call, used to correlate events. */
+  readonly callId: string;
+
+  /** The unified reason why the model call finished. */
+  readonly finishReason: StepResult<TOOLS>['finishReason'];
```
lgrammel: specify directly instead of deriving?

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,57 @@ export interface OnStepStartEvent<
...
+  readonly text: string;
+
+  /** The generated reasoning text segments from the model call. */
+  readonly reasoning: ReadonlyArray<{ text?: string }>;
+
+  /** Files generated directly by the model call. */
+  readonly files: ReadonlyArray<GeneratedFile>;
+
+  /** Parsed tool calls emitted by the model call. */
+  readonly toolCalls: ReadonlyArray<TypedToolCall<TOOLS>>;
```
lgrammel: I am not convinced we should send the individual parts like this at all. The segregated pieces are for legacy reasons. Can we just send `content` instead?

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,57 @@ export interface OnStepStartEvent<
...
+ *
+ * Called immediately before the provider model call begins.
+ * Unlike `onStepStart`, this only represents model invocation work.
+ */
+export interface ModelCallStartEvent extends CallbackModelInfo {
+  /** Unique identifier for this generation call, used to correlate events. */
+  readonly callId: string;
+
+  /** The step messages that will be sent to the model. */
+  readonly messages: Array<ModelMessage>;
```
lgrammel: does this include the system message or do we need to send it separately?

what about the tools choice etc? and temperature/model settings?

if i am using a callback for a language model call, i would want the language model call specific data

packages/ai/src/generate-text/core-events.ts
```diff
@@ -218,6 +220,57 @@ export interface OnStepStartEvent<
...
+
+  /** Prepared tool definitions for the model call, if any. */
+  readonly stepTools: ReadonlyArray<Record<string, unknown>> | undefined;
```
lgrammel: rename to tools (the step concept is irrelevant for a specific call)

packages/ai/src/generate-text/core-events.ts
aayush-kapoor: > does this include the system message or do we need to send it separately?

changed to use `StandardizedPrompt`

> what about the tools choice etc? and temperature/model settings?

my thinking was, that data is already available at other callbacks, and will lead to redundant info. it's not like the info will change at StepStart and then the model call. but can still make it consistent if you think so

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: rmed

packages/ai/src/generate-text/core-events.ts
aayush-kapoor: yes changed to be `content`

packages/ai/src/generate-text/create-execute-tools-transformation.ts
aayush-kapoor: not duplicated; could use `generateId()` as the default for `responseId`; unsure what's a good default for modelId tho

packages/ai/src/generate-text/core-events.ts
```diff
@@ -15,16 +15,6 @@ import type { StepResult } from './step-result';
...
 
-/**
- * Common model information used across callback events.
- */
-export interface CallbackModelInfo {
-  /** The provider identifier (e.g., 'openai', 'anthropic'). */
-  readonly provider: string;
-  /** The specific model identifier (e.g., 'gpt-4o'). */
-  readonly modelId: string;
-}
```
lgrammel, line 26: not in this pr but worth reviving and aligning across the codebase as discussed yesterday

packages/ai/src/generate-text/create-execute-tools-transformation.ts
```diff
@@ -1,11 +1,15 @@
-import type { Arrayable, ToolSet } from '@ai-sdk/provider-utils';
-import {
-  IdGenerator,
+import type {
+  Arrayable,
```
lgrammel, line 2: this is the wrong file for this event. it should happen in `streamLanguageModelCall`. This file should not be changed.

packages/ai/src/generate-text/language-model-events.ts
```diff
@@ -0,0 +1,77 @@
...
+import type { Callback } from '../util/callback';
+import type { FinishReason } from '../types/language-model';
+import type { LanguageModelUsage } from '../types/usage';
+import type { ContentPart } from './content-part';
+import type { StandardizedPrompt } from '../prompt/standardize-prompt';
+
+/**
+ * Common model information used across callback events.
+ */
+export type CallbackModelInfo = {
```
lgrammel, line 11: just `ModelInfo`

lgrammel (approved):

packages/ai/src/generate-text/core-events.ts
lgrammel: yes lets make it consistent. there may be users listening only to the model call start/end events but not the other ones.

packages/ai/src/generate-text/create-execute-tools-transformation.ts
lgrammel: good default for model id would be the one from the model that is called

packages/ai/src/generate-text/core-events.ts
aayush-kapoor: done

PR #14662, opened 2026-04-21
fix(workflow): Preserve invalid tool call errors in workflow UI streams

tomdale:
# Preserve invalid tool call errors in workflow UI streams

WorkflowAgent currently assumes every tool call must end with a workflow-generated tool result. That breaks the invalid-input path from AI SDK core: an invalid call already emits `tool-input-error` and `tool-output-error`, but workflow later synthesizes a fallback result and writes it back as `tool-output-available`. The UI reducer treats that as the final state and overwrites the original error.

This change teaches WorkflowAgent to keep invalid tool calls out of workflow-side execution and success-style result synthesis. Valid executable and provider-executed calls still flow through the existing loop, but invalid calls are carried forward only as `error-text` continuation results for the next model step. That preserves model-side recovery behavior without emitting a synthetic `tool-result` chunk that would mutate the UI into an `output-available` state. The final workflow result now reports `toolResults` only for tools that actually executed.

The test coverage adds a regression around this lifecycle. It verifies that an invalid call is not executed, that workflow does not emit a synthetic `tool-result` write for it, that the iterator still receives the error continuation result, and that the package returns no executed tool result for the invalid call.

## Testing

- `pnpm --filter @ai-sdk/workflow test:node`

### DIFF
--- a/.changeset/rare-birds-deny.md
+++ b/.changeset/rare-birds-deny.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/workflow": patch
+---
+
+fix (workflow): preserve invalid tool calls as errors instead of emitting synthetic success results
--- a/packages/workflow/src/workflow-agent.test.ts
+++ b/packages/workflow/src/workflow-agent.test.ts
@@ -639,6 +639,112 @@ describe('WorkflowAgent', () => {
 
       consoleWarnSpy.mockRestore();
     });
+
+    it('should keep invalid tool calls on the error path without executing them', async () => {
+      const execute = vi.fn(async () => 'should-not-run');
+      const tools: ToolSet = {
+        testTool: {
+          description: 'A test tool',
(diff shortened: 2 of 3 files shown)


gr2m (approved):
Approving — fix is correct and matches the behavior AI SDK core / `ToolLoopAgent` already has for invalid tool calls. Splitting invalid calls off from execution and UI success writes while still feeding them back into `iterator.next(...)` as `error-text` continuation results is the right shape, and the new test pins down the lifecycle nicely.

One minor thing I've filed as a follow-up rather than blocking here: #14668 — core only emits error output parts for `invalid && dynamic` tool calls (with a `TODO AI SDK 6` noting that invalid inputs shouldn't require output parts at all), whereas this PR treats all invalid calls uniformly. Worth aligning the two paths (or confirming the divergence is intentional) in a separate change.

Issue #14595, opened 2026-04-17
GatewayProviderOptions type missing 'caching' property (docs mismatch)


rossanodr:
### Description

The documentation for automatic caching in the AI Gateway suggests using the following configuration:

```
providerOptions: {
  gateway: {
    caching: 'auto',
  },
}
```

However, when using TypeScript with:

```
gateway: {
  caching: 'auto',
} satisfies GatewayProviderOptions
```

it results in the following type error:

> Object literal may only specify known properties, and 'caching' does not exist in type '{ only?: string[] | undefined; order?: string[] | undefined; sort?: "tps" | "cost" | "ttft" | undefined; user?: string | undefined; tags?: string[] | undefined; models?: string[] | undefined; ... }'.ts(2353)

It appears that the GatewayProviderOptions type does not include the caching property, even though it is documented as a valid option.

**Steps to Reproduce**

Follow the documentation: https://vercel.com/docs/ai-gateway/models-and-providers/automatic-caching
Add caching: 'auto' inside gateway options
Use satisfies GatewayProviderOptions (or strict typing)
Observe the TypeScript error

Expected Behavior
The caching property should be included in the GatewayProviderOptions type, or the documentation should reflect the correct usage.

Actual Behavior
TypeScript raises an error because caching is not part of the type definition.

This looks like a mismatch between the documentation and the current type definitions. It’s unclear whether:

caching is deprecated,
not yet released in types,
or should be configured differently.

Clarification would be appreciated.

### AI SDK Version

"@ai-sdk/gateway": "^3.0.104",

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Zelys-DFKH:
Traced this to a missing `caching` field in the `gatewayProviderOptions` Zod schema in `packages/gateway/src/gateway-provider-options.ts`. Fix is `caching: z.literal('auto').optional()` with a docs update, a type test, and a usage example. Happy to send a PR.

narutamaaurum:
PR submitted: https://github.com/vercel/ai/pull/15925

This adds the missing  property to the  type and updates the documentation with a description and usage example.

n1ckoates:
Added to AI SDK 7 in https://github.com/vercel/ai/pull/15768
Backported to AI SDK 6 in https://github.com/vercel/ai/pull/16483

lgrammel:
## Identify issue type

Identified as **Bug**  
Confidence: **High**

### Rationale
[...]

PR #14663, opened 2026-04-21
fix(gateway): include caching in GatewayProviderOptions

MukundaKatta:
## Summary
- add the missing caching field to the gateway provider options schema
- cover the schema change with a gateway-language-model regression test
- align the exported TypeScript type with the documented automatic caching configuration

Closes #14595

## Verification
- pnpm vitest run packages/gateway/src/gateway-language-model.test.ts -t "allow caching in provider options" (not runnable here because vitest is not installed in the current workspace dependencies and this environment uses unsupported Node v25)

### DIFF
--- a/packages/gateway/src/gateway-language-model.test.ts
+++ b/packages/gateway/src/gateway-language-model.test.ts
@@ -1482,6 +1482,26 @@ describe('GatewayLanguageModel', () => {
       });
     });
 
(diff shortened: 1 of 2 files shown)


MukundaKatta:
#14664 addresses the same issue with a more complete diff. Closing this earlier version to keep review signal clean.

PR #14665, opened 2026-04-21
fix(embed): tolerate missing warnings from v2 providers

### LINKED ISSUE #14425: embed/embedMany crash: 'warnings' is undefined from EmbeddingModelV2 providers

MukundaKatta:
## Summary
- guard logWarnings against undefined warning arrays
- avoid spreading undefined warnings in embedMany batch aggregation
- add regression tests for embed and embedMany when providers omit warnings entirely

Closes #14425

## Verification
- pnpm vitest run packages/ai/src/embed/embed.test.ts packages/ai/src/embed/embed-many.test.ts packages/ai/src/logger/log-warnings.test.ts (not runnable here because this workspace does not have vitest available in the current environment and uses unsupported Node v25)

### DIFF
--- a/packages/ai/src/embed/embed-many.test.ts
+++ b/packages/ai/src/embed/embed-many.test.ts
@@ -423,6 +423,21 @@ describe('result.warnings', () => {
     expect(result.warnings).toStrictEqual(expectedWarnings);
   });
 
+  it('should default warnings to an empty array when the provider omits them', async () => {
+    const result = await embedMany({
+      model: new MockEmbeddingModelV4({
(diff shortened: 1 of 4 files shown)


aayush-kapoor (commented):
marking as closed. see https://github.com/vercel/ai/issues/14425#issuecomment-4836136155

PR #14687, opened 2026-04-22
fix(xai): strip additionalProperties from tool schemas for xAI compatibility

### LINKED ISSUE #14678: addAdditionalPropertiesToJsonSchema breaks xAI/Grok tool calling: xAI rejects additionalProperties: false
## Description

## Root Cause

```ts
    ...
  }
}
```

## Impact

```
```

This affects:

## Investigation

## Suggested Fix

## Workaround (for now)

```json
{
  "provider": {
    "openrouter": {
      "models": {
        "x-ai/grok-4.1-fast": {
          "tool_call": false
        }
      }
    }
  }
}
```

## Related

kagura-agent:
## Summary

Fixes #14678

xAI's validator rejects `additionalProperties: false` in tool schemas, treating `false` as an unsupported boolean property schema. The AI SDK's `addAdditionalPropertiesToJsonSchema` in `@ai-sdk/provider-utils` adds this field globally for all providers, which breaks xAI/Grok tool calling with:

```
[xAI] Invalid request content: Schema validation failed: [standard_violation]
/properties/properties/additionalProperties: property schema 'false' is not supported
```

## Changes

Added a `removeAdditionalProperties` helper in the xAI provider (`packages/xai/src/xai-prepare-tools.ts`) that recursively strips `additionalProperties` from schemas before sending to the xAI API. Applied to:
- Tool parameter schemas in `prepareTools()`
- Response format schemas in `xai-chat-language-model.ts`

This is a provider-level fix that doesn't affect other providers. xAI defaults `additionalProperties` to `false` already, so removing it is semantically equivalent.

## Testing

- Added unit tests for `removeAdditionalProperties` (top-level, nested, no-op cases)
- Added integration test verifying `prepareTools` strips the field from tool parameters
- Updated inline snapshot in `xai-chat-language-model.test.ts` to reflect the change

Note: Unable to run the full test suite locally (pnpm install OOMs on this machine), relying on CI for full validation.

### DIFF
--- a/packages/xai/src/xai-chat-language-model.test.ts
+++ b/packages/xai/src/xai-chat-language-model.test.ts
@@ -263,7 +263,6 @@ describe('XaiChatLanguageModel', () => {
                 "name": "test-tool",
                 "parameters": {
                   "$schema": "http://json-schema.org/draft-07/schema#",
-                  "additionalProperties": false,
                   "properties": {
                     "value": {
                       "type": "string",
--- a/packages/xai/src/xai-chat-language-model.ts
+++ b/packages/xai/src/xai-chat-language-model.ts
@@ -36,7 +36,7 @@ import {
   type XaiChatModelId,
 } from './xai-chat-options';
 import { xaiFailedResponseHandler } from './xai-error';
-import { prepareTools } from './xai-prepare-tools';
+import { prepareTools, removeAdditionalProperties } from './xai-prepare-tools';
 
 type XaiChatConfig = {
   provider: string;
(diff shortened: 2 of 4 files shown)


kagura-agent:
Friendly ping — this fix for xAI tool schema compatibility has been open for about 10 days. Happy to adjust if needed! 🙏

kagura-agent:
Gentle follow-up — this xAI tool schema fix has been open for ~12 days. If there's a preferred approach or if this isn't a priority, happy to hear. Will close by May 11 if no response, and can reopen anytime. 🙏

kagura-agent:
Closing — this has been open for 13 days with no maintainer review after two pings. Happy to reopen if the fix is still desired. Thanks! 🙏

PR #14714, opened 2026-04-24
fix(ui): guard activeResponse access in makeRequest finally block (closes #11024)

### LINKED ISSUE #11024: Calling `sendMessage()` reports 'TypeError: Cannot read properties of undefined (reading 'state')'
### Description

Trying this example https://v6.ai-sdk.dev/cookbook/next/human-in-the-loop

```
await addToolOutput({
  toolCallId,
  tool: toolName,
  output: 'No, denied.',
});
sendMessage();
```

The above `sendMessage()` will report an Error:

```
    at Chat.makeRequest (chat.ts:688:41)
    at async AbstractChat.sendMessage (chat.ts:312:7)
```

Related source code: 

https://github.com/vercel/ai/blob/4b7e5211b752882557b56298daf6f6d2db70ee76/packages/ai/src/ui/chat.ts#L687-L693

Shouldn't `this.activeResponse!.state.message` be `this.activeResponse?.state.message`?

### AI SDK Version

"ai": "^6.0.0-beta.137"

### Code of Conduct

Jim1874:
### Problem
Calling `sendMessage()` after `addToolOutput()` causes:
```
    at Chat.makeRequest (chat.ts:688:41)
```

This happens because `activeResponse` was declared inside the `try` block but referenced in the `finally` block via `this.activeResponse!`, which could be cleared by concurrent calls or set to undefined before `onFinish` runs.

### Solution
1. Move `activeResponse` declaration before the `try` block so it is in scope in the `finally` block
2. Guard the `onFinish` call with `if (activeResponse)` check
3. Use the captured local variable instead of `this.activeResponse!` in the `finally` block

### Changes
- `packages/ai/src/ui/chat.ts`: Move `activeResponse` before `try`, add guard in `finally`

### Testing
All 26 chat tests pass.

Closes #11024

### DIFF
--- a/.changeset/fix-active-response-state.md
+++ b/.changeset/fix-active-response-state.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ui): guard activeResponse access in makeRequest finally block to prevent "Cannot read properties of undefined (reading 'state')" error when calling sendMessage() after addToolOutput()
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -652,8 +652,9 @@ export abstract class AbstractChat<UI_MESSAGE extends UIMessage> {
     let isDisconnect = false;
     let isError = false;
 
+    let activeResponse: ActiveResponse<UI_MESSAGE> | undefined;
     try {
-      const activeResponse = {
+      activeResponse = {
         state: createStreamingUIMessageState({
           lastMessage: this.state.snapshot(lastMessage),
           messageId: this.generateId(),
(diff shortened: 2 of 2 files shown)


threepointone:
+1 for the local-capture approach here over optional-chaining-only (#14260): reading the method-local `activeResponse` in `finally` makes each overlapping `makeRequest`'s `onFinish` report its own `message` instead of `undefined`, and avoids a public type change to `ChatOnFinishCallback`. We hit this from `@cloudflare/agents`' reconnect-driven `resumeStream()` overlap — repro + diagnosis in #15870.

mdnanocom:
pr can be closed, issue is fixed on main branch

PR #14719, opened 2026-04-24
docs: add local regression eval example

MukundaKatta:
Adds a provider-free ai-functions example for local regression scoring of AI outputs. The example demonstrates substring and token-F1 checks so developers can smoke-test output quality without requiring API credentials.


MukundaKatta:
Withdrawing — part of a broader cleanup of low-value drive-by contributions. Apologies for the noise.

PR #13979, opened 2026-03-31
Backport: fix(amazon-bedrock): preserve reasoning text when signature is present

ghost:
This is an automated backport of #13972 to the release-v6.0 branch. FYI @dancer

PR #13972, opened 2026-03-31
fix(amazon-bedrock): preserve reasoning text when signature is present

dancer:
## background

`trimIfLast()` trims trailing whitespace from reasoning block text even when a cryptographic `signature` is present. the signature is computed over the exact original bytes - trimming even a single trailing newline invalidates the text/signature pair, causing Bedrock to reject with `ValidationException: thinking blocks cannot be modified`

## summary

- skip `trimIfLast` for reasoning blocks that have a signature
- reasoning blocks without a signature still get trimmed as before
- add test for multi-turn tool use with signed reasoning blocks

## checklist

- [x] tests have been added / updated (for bug fixes / features)
- [ ] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)
- [x] i have reviewed this pull request (self-review)

## related issues

fixes #13583

PR #14724, opened 2026-04-24
fix(amazon-bedrock): do not replay unsigned reasoning content in multi-turn messages

### LINKED ISSUE #14720: bug(amazon-bedrock): `<reasoning>` tags leaked into text response for OpenAI models on Bedrock during multi-turn tool use
### Description

Since `@ai-sdk/amazon-bedrock@4.0.85`, OpenAI models accessed via Amazon Bedrock (e.g., `openai.gpt-oss-120b-1:0`) leak raw `<reasoning>...</reasoning>` tags into the final text response during multi-turn conversations that involve tool use.

The issue is specific to models that return `reasoningContent` blocks via the Bedrock Converse API **without a cryptographic signature** — which is how OpenAI's extended thinking works on Bedrock (unlike Anthropic, which always signs its reasoning blocks).

Anthropic models (Claude) are **not affected**.

## How it happens

When OpenAI models with extended thinking reason, the Bedrock Converse API returns `reasoningContent.reasoningText` blocks — but unlike Anthropic's implementation, **OpenAI does not include a `signature`** field alongside the reasoning.

This matters because of a change introduced in v4.0.85 (PR #13972). That PR added an `else` branch in `convert-to-bedrock-chat-messages.ts` that replays **unsigned** reasoning back as `reasoningContent` blocks when constructing multi-turn messages. The flow is:

1. OpenAI model responds with `reasoningContent.reasoningText` (no signature, no `redactedData`)
2. `doGenerate` creates `{ type: 'reasoning', text: '...' }` content parts **without `providerMetadata`** (the code only attaches metadata when a signature is present)
3. On the next turn (e.g., after a tool call), `convertToBedrockChatMessages` processes the reasoning part
4. `parseProviderOptions` returns `null` → both `signature` and `redactedData` are `undefined`
5. The `else` branch catches this case and wraps the text as `reasoningContent: { reasoningText: { text } }`
6. This sends the raw reasoning back to the model in the conversation history
7. The model gets confused by receiving its own reasoning as input → leaks `<reasoning>` tags into the visible text output

**Before v4.0.85:** Unsigned reasoning was simply dropped from replayed messages. The previous code only replayed reasoning when `reasoningMetadata != null`, which effectively required a signature or redacted data.

## Reproduction

**Model:** `openai.gpt-oss-120b-1:0` on Amazon Bedrock (us-east-1)  
**SDK version:** `@ai-sdk/amazon-bedrock@4.0.85` through `4.0.96` (latest)  
**API:** Converse (non-streaming)  
**Scenario:** Multi-turn agent conversation with tool use

### Steps

1. Create an agent with tools using `generateText` (non-streaming) against `openai.gpt-oss-120b-1:0` on Bedrock
2. Send a query that triggers tool use (requires at least one tool call → tool result → final response cycle)
3. The final text response contains raw `<reasoning>...</reasoning>` tags

### Example leaked response

```
<reasoning>The user asked about product X. I need to use the tool result to construct the answer.
Let me format the URLs from the template...

**Product X — Overview**
Here are the relevant links:
- [Link 1](https://example.com/...)
- [Link 2](https://example.com/...)
```

Note: the opening `<reasoning>` tag is present but the closing `</reasoning>` tag is **never emitted** — the model's internal chain-of-thought bleeds directly into the visible response text without any delimiter.

### AI SDK Version

@ai-sdk/amazon-bedrock@4.0.85

### Code of Conduct

KiraPC:
PR #13972 fixed a bug where `trimIfLast()` invalidated Anthropic reasoning signatures by trimming trailing whitespace. During that refactoring, the code structure in `convert-to-bedrock-chat-messages.ts` changed from:

```ts
if (reasoningMetadata != null) {
  if (reasoningMetadata.signature != null) { /* replay with trim */ }
  else if (reasoningMetadata.redactedData != null) { /* replay redacted */ }
}
// unsigned reasoning → silently dropped (no else branch)
```

to:

```ts
if (reasoningMetadata?.signature != null) { /* replay without trim  */ }
else if (reasoningMetadata?.redactedData != null) { /* replay redacted */ }
else { /* replay unsigned reasoning as reasoningContent ← NEW catch-all */ }
```

The outer `if (reasoningMetadata != null)` guard was removed and replaced with optional chaining. This inadvertently introduced an `else` catch-all that **replays unsigned reasoning** (reasoning parts with no `providerMetadata`) as `reasoningContent` blocks — a case that was previously silently dropped.

The PR #13972 commit message states: *"reasoning blocks without a signature still get trimmed as before"*, but Anthropic models always include a signature. The only case reaching the `else` branch is unsigned reasoning from non-Anthropic models (e.g., OpenAI extended thinking on Bedrock), which should not be replayed.

## Summary

- Remove the `else` catch-all branch in the `reasoning` case of `convertToBedrockChatMessages`
- Reasoning parts without `signature` or `redactedData` are now silently dropped (restoring pre-#13972 behavior for this specific case)
- Signed reasoning (Anthropic) and redacted reasoning remain completely unaffected
- Add comment explaining why unsigned reasoning is intentionally not replayed
- Update tests to verify unsigned reasoning is dropped rather than replayed

### Why the `else` branch is unnecessary

- **Anthropic** always includes a cryptographic `signature` → handled by the `if` branch
- **Anthropic redacted reasoning** always has `redactedData` → handled by the `else if` branch
- **Non-Anthropic models** (OpenAI) produce unsigned reasoning that should not be replayed — replaying it causes the model to leak raw `<reasoning>` tags into the visible text response

## Manual Verification

Tested against real Bedrock models with a multi-turn agent + tool-use scenario (non-streaming `generateText`):

| Model | Without fix | With fix |
|---|---|---|
| `openai.gpt-oss-120b-1:0` | Leaks `<reasoning>` tags (33-100% of runs) | No leak |
| `us.anthropic.claude-sonnet-4-20250514-v1:0` | OK (reasoning with signature) | OK |
| `moonshotai.kimi-k2.5` | OK (no reasoning blocks) |  OK |
| `amazon.nova-pro-v1:0` | OK (no reasoning blocks) | OK |

## Checklist

## Related Issues

- Fixes #14720
- Caused by #13972 (backported in #13979)
- Related: #14175 (empty text blocks with reasoning — same code area)

### DIFF
--- a/.changeset/fix-unsigned-reasoning-leak.md
+++ b/.changeset/fix-unsigned-reasoning-leak.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix(amazon-bedrock): do not replay unsigned reasoning content in multi-turn messages
+
+Reasoning parts without a `signature` or `redactedData` (e.g. from OpenAI models with extended thinking on Bedrock) are no longer replayed as `reasoningContent` blocks in multi-turn conversations. Replaying unsigned reasoning caused models to leak raw `<reasoning>` tags into the visible text response. Signed reasoning (Anthropic) and redacted reasoning are unaffected.
--- a/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.test.ts
+++ b/packages/amazon-bedrock/src/convert-to-bedrock-chat-messages.test.ts
@@ -721,7 +721,7 @@ describe('assistant messages', () => {
     `);
   });
 
-  it('should trim trailing whitespace from reasoning content without signature when it is the last part', async () => {
+  it('should drop reasoning content without signature or redactedData', async () => {
     const result = await convertToBedrockChatMessages([
       {
         role: 'user',
@@ -732,8 +732,9 @@ describe('assistant messages', () => {
         content: [
           {
             type: 'reasoning',
-            text: 'This is my reasoning with trailing space    ',
+            text: 'This is unsigned reasoning that should be dropped',
           },
+          { type: 'text', text: 'The answer is 42.' },
         ],
       },
     ]);
@@ -752,11 +753,7 @@ describe('assistant messages', () => {
           {
             "content": [
               {
-                "reasoningContent": {
-                  "reasoningText": {
-                    "text": "This is my reasoning with trailing space",
-                  },
-                },
+                "text": "The answer is 42.",
               },
             ],
             "role": "assistant",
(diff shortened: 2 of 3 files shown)


aayush-kapoor:
signed github commits are needed for community PRs

aayush-kapoor:
continued in https://github.com/vercel/ai/pull/15181; will add you as a co-author

PR #14725, opened 2026-04-24
fix(provider-utils): defer tool-input-start when function.name arrives in later delta

### LINKED ISSUE #14722: OpenAI-compatible streaming tool calls fail when function.name arrives in a later chunk
### Description

OpenAI-compatible streaming tool calls can fail when a provider/model sends the first `tool_calls` delta with an `id` and `function.arguments`, but without `function.name`, and then sends `function.name` in a later delta.

A parser that validates `function.name` on the first partial tool-call delta throws before it has a chance to accumulate the later chunk that contains the name.

This is similar to #6687 in that a provider-specific OpenAI-compatible streaming shape is valid after accumulation, but fails because the parser expects a more complete tool-call shape too early. The difference is that #6687 was about empty arguments (`""`) for parameterless tools; this issue is about `function.name` arriving after an earlier arguments-only delta.

### Expected Behavior

The streamed deltas should be accumulated by tool-call `index` / `id` until both the tool name and arguments are available.

The chunks below should become one valid tool call:

```json
{
  "id": "call_123",
  "type": "function",
  "function": {
    "name": "bash",
    "arguments": "{\"command\": \"ls -la\"}"
  }
}
```

The tool call should then be emitted/executed normally.

### Actual Behavior

The stream parser rejects the second chunk before the later `function.name` chunk is processed:

```txt
AI_InvalidResponseDataError: Expected 'function.name' to be a string.
```

### Minimal Reproduction

This mock OpenAI-compatible SSE stream reproduces the chunk ordering. The key part is the second chunk: it contains `id`, `type`, and `function.arguments`, but no `function.name`. The third chunk then provides `function.name`.

```ts
const chunks = [
  `data: {"choices":[{"finish_reason":null,"index":0,"delta":{"role":"assistant","content":null}}]}`,
  `data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"id":"call_123","type":"function","function":{"arguments":""}}]}}]}`,
  `data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"id":"call_123","type":"function","function":{"name":"bash","arguments":"{"}}]}}]}`,
  `data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"function":{"arguments":"\\u0022command\\u0022"}}]}}]}`,
  `data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"function":{"arguments":": \\u0022ls -la\\u0022}"}}]}}]}`,
  `data: {"choices":[{"finish_reason":"tool_calls","index":0,"delta":{}}]}`,
  `data: [DONE]`,
]

const response = new Response(
  new ReadableStream({
    start(controller) {
      for (const chunk of chunks) {
        controller.enqueue(new TextEncoder().encode(`${chunk}\n\n`))
      }
      controller.close()
    },
  }),
  {
    status: 200,
    headers: { "Content-Type": "text/event-stream" },
  },
)
```

A minimal provider test can mock `fetch` to return this `response` from a `/chat/completions` request and then call `streamText` with a `bash` tool. The parser should not fail on the second chunk; it should wait until the third chunk supplies `function.name` and continue accumulating arguments.

### Raw SSE Stream

```txt
data: {"choices":[{"finish_reason":null,"index":0,"delta":{"role":"assistant","content":null}}]}

data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"id":"call_123","type":"function","function":{"arguments":""}}]}}]}

data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"id":"call_123","type":"function","function":{"name":"bash","arguments":"{"}}]}}]}

data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"function":{"arguments":"\\u0022command\\u0022"}}]}}]}

data: {"choices":[{"finish_reason":null,"index":0,"delta":{"tool_calls":[{"index":0,"function":{"arguments":": \\u0022ls -la\\u0022}"}}]}}]}

data: {"choices":[{"finish_reason":"tool_calls","index":0,"delta":{}}]}

data: [DONE]
```

### Suggested Fix Direction

Accumulate partial tool-call deltas by `index` / `id` and delay validation/emission until `function.name` is available. If the stream finishes and a pending tool call still has no `function.name`, then it should remain an invalid response.

### Related

- Similar OpenAI-compatible tool-call parser issue: #6687
- Downstream report with captured local-model stream: https://github.com/anomalyco/opencode/issues/6649

kagura-agent:
## Summary

Fixes #14722

Some OpenAI-compatible providers send the first `tool_calls` delta with an `id` and `function.arguments` but **without** `function.name`, then provide `function.name` in a subsequent delta. The `StreamingToolCallTracker` previously threw `AI_InvalidResponseDataError` on the first delta because it required `function.name` to be present immediately.

## Changes

**`packages/provider-utils/src/streaming-tool-call-tracker.ts`**
- When a new tool call delta arrives without `function.name`, create a pending tracked entry and buffer arguments instead of throwing
- Defer `tool-input-start` emission until `function.name` is received in a later delta
- When name arrives, emit `tool-input-start` followed by a single `tool-input-delta` with all buffered arguments
- Skip tool calls that never receive a name during `flush()` — no broken events emitted
- Add `hasStarted` flag to `TrackedToolCall` to track whether `tool-input-start` has been emitted

**`packages/provider-utils/src/streaming-tool-call-tracker.test.ts`**
- Add test: deferred name across multiple deltas (reproduces the exact chunk pattern from #14722)
- Add test: buffered arguments replayed when name arrives
- Add test: tool calls without name are silently skipped on flush
- Remove test that expected throwing on missing `function.name` (now handled gracefully)

## Reproduction

The chunk ordering from the issue:
```
delta: {tool_calls: [{index:0, id:"call_123", type:"function", function:{arguments:""}}]}     // no name
delta: {tool_calls: [{index:0, id:"call_123", type:"function", function:{name:"bash", arguments:"{"}}]}  // name arrives
delta: {tool_calls: [{index:0, function:{arguments:"\"command\""}}]}
...
```

Previously threw: `AI_InvalidResponseDataError: Expected 'function.name' to be a string.`
Now: accumulates correctly and emits a valid tool call.

## Related

- Similar prior fix for empty arguments: #6687
- Downstream report: https://github.com/anomalyco/opencode/issues/6649

### DIFF
--- a/packages/provider-utils/src/streaming-tool-call-tracker.test.ts
+++ b/packages/provider-utils/src/streaming-tool-call-tracker.test.ts
@@ -242,21 +242,138 @@ describe('StreamingToolCallTracker', () => {
       ).toThrow("Expected 'id' to be a string.");
     });
 
-    it('should throw when function.name is missing', () => {
+    it('should defer tool-input-start when function.name arrives in a later delta', () => {
       const tracker = new StreamingToolCallTracker();
-      const { enqueue } = createCollector();
+      const { parts, enqueue } = createCollector();
 
-      expect(() =>
-        tracker.processDelta(
-          {
-            index: 0,
-            id: 'call_1',
-            type: 'function',
-            function: {},
-          },
-          enqueue,
-        ),
-      ).toThrow("Expected 'function.name' to be a string.");
+      // First delta: id and empty arguments, but no name
+      tracker.processDelta(
+        {
+          index: 0,
+          id: 'call_123',
+          type: 'function',
+          function: { arguments: '' },
+        },
+        enqueue,
+      );
+
+      // No events emitted yet (name is pending)
+      expect(parts).toEqual([]);
+
+      parts.length = 0;
+
+      // Second delta: name arrives with first argument fragment
+      tracker.processDelta(
+        {
+          index: 0,
+          id: 'call_123',
+          type: 'function',
+          function: { name: 'bash', arguments: '{' },
+        },
+        enqueue,
+      );
+
+      expect(parts).toEqual([
+        { type: 'tool-input-start', id: 'call_123', toolName: 'bash' },
+        { type: 'tool-input-delta', id: 'call_123', delta: '{' },
+      ]);
+
+      parts.length = 0;
+
+      // Third delta: more arguments
+      tracker.processDelta(
+        {
+          index: 0,
+          function: { arguments: '"command": "ls -la"}' },
+        },
+        enqueue,
+      );
+
+      expect(parts).toEqual([
+        {
+          type: 'tool-input-delta',
+          id: 'call_123',
+          delta: '"command": "ls -la"}',
+        },
+        { type: 'tool-input-end', id: 'call_123' },
+        {
+          type: 'tool-call',
+          toolCallId: 'call_123',
+          toolName: 'bash',
+          input: '{"command": "ls -la"}',
+        },
+      ]);
+    });
+
+    it('should buffer arguments before name arrives and replay them', () => {
+      const tracker = new StreamingToolCallTracker();
+      const { parts, enqueue } = createCollector();
+
+      // First delta: id with some arguments but no name
+      tracker.processDelta(
+        {
+          index: 0,
+          id: 'call_1',
+          type: 'function',
+          function: { arguments: '{"ke' },
+        },
+        enqueue,
+      );
+
+      expect(parts).toEqual([]);
+
+      parts.length = 0;
+
+      // Second delta: name arrives
+      tracker.processDelta(
+        {
+          index: 0,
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
we shouldn't change existing behaviour. change should ideally be scoped to only in `openai-compatible` provider

aayush-kapoor:
marking as closed in favour of https://github.com/vercel/ai/pull/14760

PR #10785, opened 2025-12-02
feat(deepseek): rewrite DeepSeek provider

lgrammel:
## Background

DeepSeek has introduced reasoning functionality that requires custom message preparation (see #10778 ). The differences between the DeepSeek API and the OpenAI compatible layer are growing, both in additional and in unsupported features.

## Summary

Rewrite the Deepseek provider to be specific to the Deepseek API.

The dedicated Deepseek model will:
* remove unsupported options and have better warnings
* support sending reasoning back to support tool calls (deepseek api update)
* have improved json generation
* end reasoning parts in streams earlier
* fix content order when reasoning is sent

## Manual Verification

- [x] run `examples/ai-core/src/generate-text/deepseek-chat.ts`
- [x] run `examples/ai-core/src/stream-text/deepseek-chat.ts`
- [x] run `examples/ai-core/src/generate-text/deepseek-reasoner.ts`
- [x] run `examples/ai-core/src/stream-text/deepseek-reasoner.ts`
- [x] run `examples/ai-core/src/generate-text/deepseek-tool-call.ts`
- [x] run `examples/ai-core/src/stream-text/deepseek-tool-call.ts`
- [x] ui test with reasoning and tool calls (use follow up messages) http://localhost:3000/chat-deepseek-tools

## Tasks

- [x] base generate test
- [x] base stream test
- [x] update finish reason mapping
- [x] request body as object
- [x] thinking generate test
- [x] incorrect thinking order
- [x] deepseek usage generate
- [x] deepseek usage stream
- [x] end reasoning stream part early
- [x] error data
- [x] thinking stream test
- [x] stream request body test
- [x] json schema generate
- [x] json no schema generate
- [x] tool generate test
- [x] tool stream test
- [x] reasoning with compression and tool calls
- [x] check if thinking is automatically enabled with reasoner
- [x] updated ui example
- [x] docs rewrite
- [x] changeset

## Future Work
* implement assistant message prefilling

## Related Issues

Resolves #9179
Resolves #10778
Resolves #7458

PR #14729, opened 2026-04-24
fix(provider/deepseek): preserve reasoning_content for deepseek-v4 in multi-turn requests

woai3c:
## Background

DeepSeek released two model families with **opposite** requirements for `reasoning_content` in multi-turn chat completions:

- **`deepseek-reasoner` (R1)** — per the [DeepSeek docs](https://api-docs.deepseek.com/guides/reasoning_model), callers must **not** echo `reasoning_content` from previous assistant turns. The current converter (introduced in #10785) correctly strips it via ``if (index <= lastUserMessageIndex) break``.
- **`deepseek-v4` / `deepseek-v4-pro` (V4 thinking mode)** — the API **requires** every assistant message to carry a `reasoning_content` field. Multi-turn requests without it fail with:

  ``400 The `reasoning_content` in the thinking mode must be passed back to the API.``

The current converter applies the R1 stripping rule to V4 too, which breaks every V4 conversation after the first turn.

## Reproduction

```ts
import { deepseek } from '@ai-sdk/deepseek';
import { isStepCount, streamText } from 'ai';
import { printFullStream } from '../../lib/print-full-stream';
import { weatherTool } from '../../tools/weather-tool';

run(async () => {
  const model = deepseek('deepseek-v4-pro');

  console.log('\n=== TURN 1 (tool call) ===');
  const t1 = streamText({
    model,
    tools: { weather: weatherTool },
    stopWhen: isStepCount(3),
    messages: [
      { role: 'user', content: 'What is the weather in San Francisco?' },
    ],
  });
  await printFullStream({ result: t1 });

  const t1Messages = (await t1.response).messages;

  console.log('\n\n=== TURN 2 (replay + new user turn) ===');
  const t2 = streamText({
    model,
    tools: { weather: weatherTool },
    stopWhen: isStepCount(3),
    messages: [
      { role: 'user', content: 'What is the weather in San Francisco?' },
      ...t1Messages,
      { role: 'user', content: 'How about in New York?' },
    ],
  });
  await printFullStream({ result: t2 });
});
```

## Fix

In `convertToDeepSeekChatMessages`:

1. Add a `modelId` parameter (the function is internal — only callers are the test file and `deepseek-chat-language-model.ts`).
2. Detect V4 with `modelId.includes('deepseek-v4')`.
3. For V4: skip the strip-on-old-turn `break` so prior reasoning is preserved, and back-fill `reasoning_content: ''` when an assistant message has no `reasoning` part at all.
4. For R1 / other models: behavior unchanged.

## Test plan

- New unit tests in `convert-to-deepseek-chat-messages.test.ts`:
  - V4 with prior assistant turn → output preserves `reasoning_content`
  - V4 with assistant turn lacking a reasoning part → output emits `reasoning_content: ''`
- Existing R1/`deepseek-chat` tests continue to pass unchanged.
- All 29 tests in `@ai-sdk/deepseek` pass locally.
- Manual: a `streamText` multi-turn loop with `deepseek-v4-pro` no longer 400s after turn 1.

### DIFF
--- a/.changeset/fix-deepseek-v4-reasoning-content.md
+++ b/.changeset/fix-deepseek-v4-reasoning-content.md
@@ -3,5 +3,3 @@
 ---
 
 fix(provider/deepseek): preserve `reasoning_content` for `deepseek-v4` in multi-turn requests
-
-DeepSeek V4 thinking mode requires `reasoning_content` on every assistant message in multi-turn requests; otherwise the API returns `400 The reasoning_content in the thinking mode must be passed back to the API.` The converter now preserves prior reasoning for `deepseek-v4*` models and back-fills an empty string when no reasoning part is present. `deepseek-reasoner` (R1) behavior — which forbids echoing prior reasoning — is unchanged.
--- a/.changeset/fix-deepseek-v4-reasoning-content.md
+++ b/.changeset/fix-deepseek-v4-reasoning-content.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/deepseek": patch
+---
+
+fix(provider/deepseek): preserve `reasoning_content` for `deepseek-v4` in multi-turn requests
+
+DeepSeek V4 thinking mode requires `reasoning_content` on every assistant message in multi-turn requests; otherwise the API returns `400 The reasoning_content in the thinking mode must be passed back to the API.` The converter now preserves prior reasoning for `deepseek-v4*` models and back-fills an empty string when no reasoning part is present. `deepseek-reasoner` (R1) behavior — which forbids echoing prior reasoning — is unchanged.
--- a/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
+++ b/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
@@ -12,6 +12,7 @@ describe('convertToDeepSeekChatMessages', () => {
           },
         ],
         responseFormat: undefined,
+        modelId: 'deepseek-chat',
       });
 
       expect(result).toMatchInlineSnapshot(`
@@ -43,6 +44,7 @@ describe('convertToDeepSeekChatMessages', () => {
           },
         ],
         responseFormat: undefined,
+        modelId: 'deepseek-chat',
       });
 
       expect(result).toMatchInlineSnapshot(`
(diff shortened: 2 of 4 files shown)


aayush-kapoor:
Thanks for the PR @woai3c! one thing i noticed is that i wasn't able to reproduce the error with just your example. Even for the deepseek v4 model, if there's no tool calls and no reasoning_content is passed, the model call on the next turn still works.

it's only when there are tool calls involved that the flow breaks on the next turn. and that's documented in deepseek docs as well: https://api-docs.deepseek.com/guides/thinking_mode#input-and-output-parameters

This example worked for me (no errors): 

```ts
import { deepseek } from '@ai-sdk/deepseek';
import { printFullStream } from '../../lib/print-full-stream';

run(async () => {
  const model = deepseek('deepseek-v4-pro');

  console.log('\n=== TURN 1 ===');
  const t1 = streamText({
    model,
    messages: [{ role: 'user', content: 'hi' }],
  });
  await printFullStream({ result: t1 });

  const t1Messages = (await t1.response).messages;

  console.log('\n\n=== TURN 2 ===');
  const t2 = streamText({
    model,
    messages: [
      { role: 'user', content: 'hi' },
      ...t1Messages,
      { role: 'user', content: 'again' },
    ],
  });
  await printFullStream({ result: t2 });
});
```

But this one gave me an error: 

```ts
import { deepseek } from '@ai-sdk/deepseek';
import { isStepCount, streamText } from 'ai';
import { printFullStream } from '../../lib/print-full-stream';
import { weatherTool } from '../../tools/weather-tool';

run(async () => {
  const model = deepseek('deepseek-v4-pro');

  console.log('\n=== TURN 1 (tool call) ===');
  const t1 = streamText({
    model,
    tools: { weather: weatherTool },
    stopWhen: isStepCount(3),
    messages: [
      { role: 'user', content: 'What is the weather in San Francisco?' },
    ],
  });
  await printFullStream({ result: t1 });

  const t1Messages = (await t1.response).messages;

  console.log('\n\n=== TURN 2 (replay + new user turn) ===');
  const t2 = streamText({
    model,
    tools: { weather: weatherTool },
    stopWhen: isStepCount(3),
    messages: [
      { role: 'user', content: 'What is the weather in San Francisco?' },
      ...t1Messages,
      { role: 'user', content: 'How about in New York?' },
    ],
  });
  await printFullStream({ result: t2 });
});
```

~~so i think the fix will be a bit different~~ on second thoughts, the fix will work since this behaviour is specific to v4 models. so thanks. i will just update the repro example in the PR description

aayush-kapoor (commented):
since we now require signed commits, i will create a new PR for it. thanks for submitting the changes!

PR #14733, opened 2026-04-24
feat(provider): change file part data property to be tagged with a type and remove the image part type

felixarntz:
## Background

`FilePart.data` and `ReasoningFilePart.data` used an overloaded union (`DataContent | URL | ProviderReference`) that forced providers to inspect runtime types to infer intent, and made a bare `string` ambiguous (always interpreted as base64). The v4 provider spec had the same shape.

## Summary

- **Provider spec (breaking):** `LanguageModelV4FilePart.data` is now a 4-arm tagged union: `{ type: 'data' | 'url' | 'reference' | 'text' }`. `LanguageModelV4ReasoningFilePart.data`, `LanguageModelV4File.data`, and `LanguageModelV4ReasoningFile.data` are 2-arm (`data` / `url`). `mediaType` now accepts either a full IANA `type/subtype` or just a top-level segment (e.g. `'image'`); `*`-subtype wildcards are normalized as the top-level segment.
- **Top-level back compat:** In the top-level APIs, `ModelMessage` parts continue to support today's shape, with the new 4-arm tagged union being additionally supported. `convertToLanguageModelPrompt()` takes care of the conversion to the V4 spec.
- **`uploadFile` / `uploadSkill`** align with the new tagged file shape (`FilesV4UploadFileCallOptions.data` / `SkillsV4File.data` are `{ type: 'data' | 'text' }`). `SkillsV4File.content` renamed to `data`.
- **Top-level `FilePart` / `ReasoningFilePart`:** additively accept the new tagged shapes alongside the existing `DataContent | URL | ProviderReference` shorthands — no breaking change for app authors.
- **`ImagePart` deprecated** in favor of `FilePart` with `mediaType: 'image'`. Usage emits a runtime warning.
- **`ProviderReference`** is now `Record<string, string> & { type?: never }` so the tagged `{ type: 'reference', reference }` shape can never collide with a raw reference object.
- **Media-type helpers** (`detectMediaTypeBySignatures`, `detectMediaTypeForTopLevelType`, `getTopLevelMediaType`, `isFullMediaType`, `resolveFullMediaType`) are now available in `provider-utils`. Relevant code from `ai` was moved over.

## Manual Verification

The changes can be verified by spot checking with the examples in `examples/ai-functions`.

## Checklist

## Future Work

N/A

## Related Issues

Fixes #14010.


packages/provider/src/shared/v4/shared-v4-provider-reference.ts
```diff
@@ -12,4 +12,6 @@
  * }
  * ```
  */
-export type SharedV4ProviderReference = Record<string, string>;
+export type SharedV4ProviderReference = Record<string, string> & {
```
lgrammel, line 20: might be worth adding a comment why there is `type?: never`

packages/xai/src/files/xai-files.ts
```diff
@@ -44,7 +44,11 @@ export class XaiFiles implements FilesV4 {
     })) as XaiFilesOptions | undefined;
 
     const fileBytes =
-      data instanceof Uint8Array ? data : convertBase64ToUint8Array(data);
+      data.type === 'text'
+        ? new TextEncoder().encode(data.text)
+        : data.data instanceof Uint8Array
+          ? data.data
+          : convertBase64ToUint8Array(data.data);
```
lgrammel: could this be a helper that we extract into provider-utils?

packages/provider-utils/src/types/content-part.ts
```diff
@@ -133,12 +146,17 @@ export interface ReasoningFilePart {
   type: 'reasoning-file';
 
   /**
-   * File data. Can either be:
+   * File data. Either a tagged shape or a bare shorthand:
```
lgrammel: Maybe "Reasoning file data." and a short explanation why it is different than file data in the supported types.

packages/provider-utils/src/index.ts
```diff
@@ -10,6 +10,16 @@ export {
...
+export {
+  audioMediaTypeSignatures,
+  detectMediaTypeBySignatures,
+  detectMediaTypeForTopLevelType,
+  documentMediaTypeSignatures,
+  getTopLevelMediaType,
+  imageMediaTypeSignatures,
+  isFullMediaType,
+  videoMediaTypeSignatures,
+} from './detect-media-type';
```
lgrammel, line 18: do we need to expose all signatures or can we limit/encapsulate this more (e.g. by offering functions/a single function with args vs exposing the signatures directly)?

packages/openai/src/files/openai-files.ts
```diff
@@ -47,7 +47,11 @@ export class OpenAIFiles implements FilesV4 {
     })) as OpenAIFilesOptions | undefined;
 
     const fileBytes =
-      data instanceof Uint8Array ? data : convertBase64ToUint8Array(data);
+      data.type === 'text'
+        ? new TextEncoder().encode(data.text)
+        : data.data instanceof Uint8Array
+          ? data.data
+          : convertBase64ToUint8Array(data.data);
```
lgrammel: helper?

packages/ai/src/transcribe/transcribe.ts
```diff
@@ -122,7 +123,7 @@ export async function transcribe({
       headers: headersWithUserAgent,
       providerOptions,
       mediaType:
-        detectMediaType({
+        detectMediaTypeBySignatures({
```
lgrammel: this rename might be a breaking change for provider pkgs that extend our provider utils. maybe we need to consider listing it in the migration guide (special section for provider implementers)

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+  content: FilePart['data'],
+): ConvertResult {
+  if (isTaggedFileData(content)) {
+    switch (content.type) {
+      case 'data':
+        if (
+          typeof content.data === 'string' &&
+          content.data.startsWith('data:')
+        ) {
+          throw new AISDKError({
```
lgrammel: we should use a specific subclass otherwise narrowing via instances does not reliably work

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+  data: LanguageModelV4FilePart['data'];
+  mediaType: string | undefined;
+};
+
+function urlToV4(url: URL): ConvertResult {
+  if (url.protocol === 'data:') {
+    const { mediaType, base64Content } = splitDataUrl(url.toString());
+
+    if (mediaType == null || base64Content == null) {
+      throw new AISDKError({
```
lgrammel: error class

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+      });
+    }
+
+    return { data: { type: 'data', data: base64Content }, mediaType };
+  }
+
+  return { data: { type: 'url', url }, mediaType: undefined };
+}
+
+function inlineDataToV4(content: DataContent): ConvertResult {
```
lgrammel: dont we have similar functions (but better) in provider-utils?

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (content instanceof Uint8Array) {
+    return { data: { type: 'data', data: content }, mediaType: undefined };
+  }
+  if (content instanceof ArrayBuffer) {
+    return {
+      data: { type: 'data', data: new Uint8Array(content) },
+      mediaType: undefined,
+    };
+  }
+  if (globalThis.Buffer?.isBuffer(content) ?? false) {
```
lgrammel: extract `isBuffer` helper into provider utils since this is repeated in various places
`globalThis.Buffer?.isBuffer(content) ?? false)`

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  if (content instanceof URL) {
+    return urlToV4(content);
+  }
+
+  if (typeof content === 'string') {
+    try {
+      return urlToV4(new URL(content));
```
lgrammel: function names should ideally start with verbs and be more explicit

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+      case 'text':
+        return {
+          data: { type: 'text', text: content.text },
+          mediaType: undefined,
+        };
+    }
+  }
+
+  if (content instanceof URL) {
+    return urlToV4(content);
```
lgrammel: see comment below

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
...
+      return urlToV4(new URL(content));
+    } catch {
+      return inlineDataToV4(content);
+    }
+  }
+
+  if (
+    content instanceof Uint8Array ||
+    content instanceof ArrayBuffer ||
+    (globalThis.Buffer?.isBuffer(content) ?? false)
```
lgrammel: reuse

packages/ai/src/prompt/file-part-data.ts
```diff
@@ -0,0 +1,129 @@
+import { AISDKError, LanguageModelV4FilePart } from '@ai-sdk/provider';
```
lgrammel: content part is part of provider utils so i am wondering how much of this can be moved to provider-utils (and i feel there may already be a lot of code we can reuse)

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -401,8 +401,31 @@ export function convertToLanguageModelMessage({
...
 }
 
+/*
+ * Rewrites a legacy `ImagePart` into an equivalent `FilePart`. The default
+ * `mediaType` for a bare `ImagePart` (no `mediaType`) is `"image"` (top-level
+ * segment); an explicit `mediaType` is carried through verbatim. After this
+ * pre-pass, only `TextPart` and `FilePart` ever reach the provider-facing
+ * conversion logic.
+ */
+function imagePartToFilePart<T extends { type: string }>(
```
lgrammel: we are suddenly introducing functions that do not start with verbs

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -401,8 +401,31 @@ export function convertToLanguageModelMessage({
...
+    return part as T extends ImagePart ? FilePart : T;
+  }
+  const imagePart = part as unknown as ImagePart;
+  const filePart: FilePart = {
+    type: 'file',
+    data: imagePart.image,
+    mediaType: imagePart.mediaType ?? 'image',
+    providerOptions: imagePart.providerOptions,
+  };
+  return filePart as T extends ImagePart ? FilePart : T;
```
lgrammel: why this cast? it is already a filepart (see type)

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -401,8 +401,31 @@ export function convertToLanguageModelMessage({
...
+/*
+ * Rewrites a legacy `ImagePart` into an equivalent `FilePart`. The default
+ * `mediaType` for a bare `ImagePart` (no `mediaType`) is `"image"` (top-level
+ * segment); an explicit `mediaType` is carried through verbatim. After this
+ * pre-pass, only `TextPart` and `FilePart` ever reach the provider-facing
+ * conversion logic.
+ */
+function imagePartToFilePart<T extends { type: string }>(
+  part: T,
+): T extends ImagePart ? FilePart : T {
```
lgrammel: this is weird

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -411,42 +434,39 @@ async function downloadAssets(
...
+      url: part.data.url,
       isUrlSupportedByModel:
         part.mediaType != null &&
         isUrlSupported({
-          url: part.data.toString(),
-          mediaType: part.mediaType,
+          url: part.data.url.toString(),
+          mediaType: isFullMediaType(part.mediaType)
+            ? part.mediaType
+            : `${part.mediaType}/*`,
```
lgrammel: are we doing `/*` or just the prefix?

packages/ai/src/transcribe/transcribe.ts
felixarntz: the function wasn't in `provider-utils` before, it was in `ai`, and it was never exported from the package as far as I can tell. So I don't think there's any BC concern here.

packages/provider-utils/src/index.ts
felixarntz: good point, they were exported before, but remained internal to the `ai` package. Now that they're exported from the package I agree that would be a better API to expose.

felixarntz: done in https://github.com/vercel/ai/pull/14733/commits/418b3e47bfe37738cd96df3fb2c273bd29e33b25

packages/provider/src/shared/v4/shared-v4-provider-reference.ts
felixarntz: done in https://github.com/vercel/ai/pull/14733/commits/47f304e12c546a304a388d8396d2800a8d5aac7f

packages/openai/src/files/openai-files.ts
felixarntz: done in https://github.com/vercel/ai/pull/14733/changes/be628c73b09d75e4a52185a1700b8873718e7c02 - this is now used for both file uploads and skill file uploads

packages/ai/src/prompt/file-part-data.ts
felixarntz: updated to reuse the existing `InvalidDataContentError` class

packages/ai/src/prompt/file-part-data.ts
felixarntz: done in https://github.com/vercel/ai/pull/14733/commits/f3d13e71eaf833de82547de77ec2a0bbbd80fbbe

packages/ai/src/prompt/convert-to-language-model-prompt.ts
felixarntz: there was a problem here where types were wrong before, this was AI slop :)

packages/ai/src/prompt/convert-to-language-model-prompt.ts
felixarntz: updated `isUrlSupported` in https://github.com/vercel/ai/pull/14733/changes/95e61a2bb365f40c8526ee15234f9ba7d4b6df87 to align with our new convention of allowing top-level media types

packages/provider-utils/src/types/content-part.ts
felixarntz: done in https://github.com/vercel/ai/pull/14733/commits/bcbd45813a795ecd0c9c630032bcd81c5faa7ee3

packages/ai/src/prompt/file-part-data.ts
felixarntz: I broke out some new helpers into `provider-utils` related to this, also based on your other feedback (e.g. we now have `convertInlineDataToFilePartData`).

For the overall logic in this file, I think it makes sense to remain here, in alignment with `convertToLanguageModelPrompt` which also lives in the `ai` package. Maybe we could take a look in a follow up PR for which parts of all that (or maybe all of it?) could be moved to `provider-utils`. But on the other hand I'm unsure, because technically no provider will need to call this, it's only relevant for the `ai` package. I think of `provider-utils` as the best place for code that's relevant for both providers and the `ai` package.

packages/provider/src/files/v4/files-v4-upload-file-call-options.ts
```diff
@@ -1,13 +1,20 @@
...
+   * The file data.
+   *
+   * - `{ type: 'data', data }`: raw bytes (`Uint8Array`) or a base64-encoded string.
+   * - `{ type: 'text', text }`: inline text (UTF-8).
    */
-  data: Uint8Array | string;
+  data: Extract<
+    LanguageModelV4FilePart['data'],
+    { type: 'data' } | { type: 'text' }
+  >;
```
gr2m: Instead of using `Extract`, create shared spec types and create an enum type with the supported ones

felixarntz: done in https://github.com/vercel/ai/pull/14733/changes/6769d01b155b5ec0272da7c7a18203754981a1ed

lgrammel (approved):

PR #14737, opened 2026-04-25
fix(provider-utils): accept late-arriving function.name in streaming tool calls

### LINKED ISSUE #14722: OpenAI-compatible streaming tool calls fail when function.name arrives in a later chunk

abhicris:
Closes #14722.

## Bug

Some OpenAI-compatible providers send the first \`tool_calls\` delta with \`id\` and \`function.arguments\` and only deliver \`function.name\` in a subsequent delta. \`StreamingToolCallTracker\` rejects the first delta with \`Expected 'function.name' to be a string.\` before the next chunk has a chance to supply the name.

Structurally similar to #6687 (empty \`arguments\` for parameterless tools): the streamed shape is valid after accumulation but fails because the parser validates too eagerly.

## Fix

\`processNewToolCall\` no longer throws when \`function.name\` is missing on the first delta. The tool call is held open with \`name = null\` and the name is adopted from the first subsequent delta that supplies it.

To keep the consumer-visible event order clean:

- \`tool-input-start\` is gated on the name being known — consumers never see a \`tool-input-start\` with an empty \`toolName\`.
- \`tool-input-delta\` events for argument chunks that arrive before the name are deferred and emitted only after start.
- \`flush()\` validates that every unfinished tool call has a name; if a stream really closes without ever supplying one, the same \`InvalidResponseDataError("Expected 'function.name' to be a string.")\` is raised — the error is preserved, just moved from the first delta to flush.

The single-chunk happy path (provider sends id + name + complete arguments in one delta) is unchanged.

## Tests

\`packages/provider-utils/src/streaming-tool-call-tracker.test.ts\`:

- **Reframed:** \`should throw when function.name is missing\` → \`should throw on flush when function.name is never provided\` to match the new contract (immediate processDelta no longer throws; flush does).
- **New:** \`should accept function.name arriving in a later delta\` exercises the exact chunk shape from the issue (id + arguments first, name second, more arguments after) and asserts:
  - \`tool-input-start\` is emitted exactly once, with the late name
  - \`tool-call\` is emitted with the correct name and accumulated args
  - no \`tool-input-delta\` leaks before \`tool-input-start\`

## Diff size

\`+158 / -25\` across 2 files (1 source, 1 test). No public API change; affects only the timing of validation.

---

Contributed by [kcolbchain](https://kcolbchain.com).

### DIFF
--- a/packages/provider-utils/src/streaming-tool-call-tracker.test.ts
+++ b/packages/provider-utils/src/streaming-tool-call-tracker.test.ts
@@ -242,7 +242,9 @@ describe('StreamingToolCallTracker', () => {
       ).toThrow("Expected 'id' to be a string.");
     });
 
-    it('should throw when function.name is missing', () => {
+    it('should throw on flush when function.name is never provided', () => {
+      // function.name is allowed to arrive in a later delta. The error is
+      // surfaced only when the stream closes without a name ever showing up.
       const tracker = new StreamingToolCallTracker();
       const { enqueue } = createCollector();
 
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
we shouldn't change existing behaviour. change should ideally be scoped to only in `openai-compatible` provider

aayush-kapoor:
marking as closed in favour of https://github.com/vercel/ai/pull/14760

PR #14738, opened 2026-04-25
fix(ai): deduplicate onError callback in toUIMessageStream

### LINKED ISSUE #14726: toUIMessageStream onError is invoked twice per upstream error
### Summary

  A single `onError` callback passed to `result.toUIMessageStream({ onError })` is invoked **twice** per upstream error. The [public reference](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text) only specifies the signature (`(error: unknown) => string`) and says it's "invoked when an error occurs during stream processing" – call frequency is not documented.

### How to reproduce

```ts
import { streamText, simulateReadableStream } from "ai";

  doStream: async () => ({
      chunks: [
        { type: "stream-start", warnings: [] },
        { type: "error", error: new Error("simulated upstream error") },
      ],
    }),
  }),
});

let calls = 0;

const result = streamText({
  model,
  prompt: "trigger error",
});

  onError: (err) => {
    calls += 1;
    console.log(`\n=== onError call #${calls} ===`);
    console.trace("stack");
    return (err as Error)?.message ?? String(err);
  },
  // Removing onFinish (and onStepFinish) makes handleUIMessageStreamFinish
  // early-return, which skips processUIMessageStream — onError then fires
  // only once. Keep it here to reproduce the double call.
  onFinish: () => {},
});

async function main() {
    console.log("chunk:", JSON.stringify(chunk));
  }
  console.log(`\nTotal onError calls: ${calls}`);
}

main();
```

Run:
```
pnpm tsx repro.ts
```

Results:
```
  Observed output

  === onError call #1 ===
  Trace: stack
      at onError (repro.ts:…)
      at Object.transform (…/ai/src/generate-text/stream-text.ts:2402:28)

  === onError call #2 ===
  Trace: stack
      at onError (repro.ts:…)
      at <anonymous> (…/ai/src/ui/process-ui-message-stream.ts:710:15)
      at runUpdateMessageJob (…/ai/src/ui-message-stream/handle-ui-message-stream-finish.ts:97:11)
      at Object.transform (…/ai/src/ui/process-ui-message-stream.ts:104:15)

  Total onError calls: 2
```
Stack trace shows the handoff chain: `toUIMessageStream → handleUIMessageStreamFinish (runUpdateMessageJob) → processUIMessageStream.`

### Results

Two distinct call sites share the same onError reference:

1. Formatter stage in [generate-text/stream-text.ts](https://github.com/vercel/ai/blob/87d17235e9f673dcfe61cc28d25dab2b278dcea2/packages/ai/src/generate-text/stream-text.ts#L2502):
```ts
  case 'error': {
    controller.enqueue({
      type: 'error',
      errorText: onError(part.error),
    });
  }
```
Called synchronously with the original provider error; return value is written onto the outgoing chunk.
2. Notifier stage in [ui/process-ui-message-stream.ts](https://github.com/vercel/ai/blob/87d17235e9f673dcfe61cc28d25dab2b278dcea2/packages/ai/src/ui/process-ui-message-stream.ts#L835):
```ts
case 'error': {
onError?.(new Error(chunk.errorText));
}
```
2. Called with a freshly constructed Error wrapping the string Stage 1 just returned; return value is discarded.

The handoff is wired in `ui-message-stream/handle-ui-message-stream-finish.ts`, which forwards `onError` from `toUIMessageStream` into `processUIMessageStream`. This path
is only taken when `onFinish` or `onStepFinish` is passed to `toUIMessageStream`; otherwise Stage 2 is skipped.

### Proposal
Docs-only change. Add a note on the `onError` parameter of `toUIMessageStream`:

> May be invoked more than once per error: once by the formatter stage producing the SSE chunk's errorText, and again by the chunk walker as a notification (return value discarded). Side effects inside onError (logging, metrics, persistence) should be idempotent, or externally deduped — otherwise a single upstream error will produce duplicate records.

### AI SDK Version

Versions:
- ai@6.0.90

### Code of Conduct

Zelys-DFKH:
## Problem

The `onError` callback passed to `result.toUIMessageStream({ onError })` is invoked **twice** per upstream error when both `onError` and `onFinish` callbacks are provided.

### Root Cause

Two distinct call sites share the same `onError` reference:

1. **Formatter stage** (`generate-text/stream-text.ts:2542`): Invokes `onError(part.error)` with the original provider error and queues the formatted result as `errorText`
2. **Processor stage** (`ui/process-ui-message-stream.ts:835`): Invokes `onError` again with a newly constructed Error wrapping the already-formatted `errorText`

The handoff occurs in `handle-ui-message-stream-finish.ts`, which forwards the same `onError` callback to `processUIMessageStream`. This second invocation is redundant.

## Solution

Remove the duplicate `onError` invocation in `process-ui-message-stream.ts`. The error has already been processed during the formatter stage with the original error object. The processor stage doesn't need to re-invoke the callback.

### Changes

1. Remove `onError?.(new Error(chunk.errorText))` from the error case handler in `process-ui-message-stream.ts`
2. Mark the `onError` parameter as intentionally unused (`_onError`) to satisfy lint rules
3. Add a regression test to ensure `onError` is called exactly once

## Testing

Added regression test in `stream-text.test.ts` that verifies:
- `onError` is called exactly once when an error occurs with both `onError` and `onFinish` callbacks
- The error chunk is properly included in the stream

Fixes #14726

### DIFF
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -4190,6 +4190,37 @@ describe('streamText', () => {
       // Verify that onFinish was NOT called when stream was cancelled
       expect(onFinishCallback).not.toHaveBeenCalled();
     });
+
+    it('should call onError exactly once per error', async () => {
+      const onErrorCallback = vi.fn(() => 'error');
+      const onFinishCallback = vi.fn();
+
+      const result = streamText({
+        model: createTestModel({
+          stream: convertArrayToReadableStream([
+            { type: 'stream-start', warnings: [] },
+            { type: 'error', error: new Error('test error') },
+          ]),
+        }),
+        prompt: 'test-input',
+      });
+
+      const chunks = await convertReadableStreamToArray(
+        result.toUIMessageStream({
+          onError: onErrorCallback,
+          onFinish: onFinishCallback,
+        }),
+      );
+
+      // Verify onError was called exactly once
+      expect(onErrorCallback).toHaveBeenCalledTimes(1);
+      // Verify the error chunk is in the stream
+      expect(chunks).toContainEqual(
+        expect.objectContaining({
+          type: 'error',
+        }),
+      );
+    });
   });
 
   describe('result.toUIMessageStreamResponse', () => {
--- a/packages/ai/src/ui/process-ui-message-stream.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.ts
@@ -78,7 +78,7 @@ export function processUIMessageStream<UI_MESSAGE extends UIMessage>({
   messageMetadataSchema,
   dataPartSchemas,
   runUpdateMessageJob,
-  onError,
+  onError: _onError,
   onToolCall,
   onData,
 }: {
@@ -832,7 +832,8 @@ export function processUIMessageStream<UI_MESSAGE extends UIMessage>({
             }
 
             case 'error': {
-              onError?.(new Error(chunk.errorText));
+              // Error callback already invoked in stream-text.ts formatter stage
+              // with the original error. The errorText here is the formatted result.
               break;
             }
 


Zelys-DFKH:
Ping for review — this deduplication fix for onError callbacks has been green for a while now. All checks passing and it resolves a real user issue. Ready when you are! ✨

PR #14749, opened 2026-04-27
fix(ai): reject system-role UI messages in createAgentUIStream

etairl:
## Background

Three issues were identified in `packages/ai` that weaken trust boundaries between developer-controlled and caller-controlled inputs:

1. **Client-controlled `role: 'system'` injection in `createAgentUIStream`** (high). `uiMessagesSchema` allows `role: 'system'`, `convertToModelMessages` preserves it, and `createAgentUIStream` forwards the converted prompt to `agent.stream()`. Any caller of `createAgentUIStream` / `createAgentUIStreamResponse` could therefore inject system-level instructions alongside the developer's `instructions`, indistinguishable to the model from the developer-authored system prompt.
2. **`callOptionsSchema` declared but never enforced** (medium). `ToolLoopAgentSettings.callOptionsSchema` is named and documented as a runtime schema for `options`, but `tool-loop-agent.ts` never invokes it. Any invariant a developer encodes in that schema is silently bypassed at runtime, and unchecked `options` flow straight into `prepareCall` and any `instructions` template that interpolates them.
3. **Prototype-property confusion in `getMediaTypeFromUrl`** (low). The helper used `ext in URL_EXTENSION_TO_MEDIA_TYPE` against a plain object literal, so a URL ending in `.constructor` resolved through the prototype chain and returned the `Object` constructor (a function), violating the helper's `: string` return type and forwarding a non-string `mediaType` to provider adapters.

## Summary

- `packages/ai/src/agent/create-agent-ui-stream.ts`: after `validateUIMessages`, throw `InvalidArgumentError` if any inbound UI message has `role: 'system'`. The error message points developers at the agent's `instructions` setting, which is the supported way to set system prompts.
- `packages/ai/src/agent/tool-loop-agent.ts`: in `prepareCall`, when `callOptionsSchema` is set and `options` is provided, run `safeValidateTypes` and throw `InvalidArgumentError` on failure, before forwarding to `prepareCall` / `generateText` / `streamText`. The validated value replaces the raw input on the way through.
- `packages/ai/src/prompt/convert-to-language-model-prompt.ts`: replace `ext in URL_EXTENSION_TO_MEDIA_TYPE` with `Object.hasOwn(...)` so attacker-controlled extensions like `.constructor` cannot resolve to inherited `Object.prototype` keys.
- Regression tests added in `create-agent-ui-stream-response.test.ts`, `tool-loop-agent.test.ts`, and `convert-to-language-model-prompt.test.ts`.
- Patch changeset for `ai` covering all three fixes.

## Manual Verification

- `npx vitest --config vitest.node.config.js --run src/agent/tool-loop-agent.test.ts src/agent/create-agent-ui-stream-response.test.ts src/prompt/convert-to-language-model-prompt.test.ts` — 151/151 pass, including the new regression tests.
- `npx tsc --noEmit` in `packages/ai` — clean.
- Verified the new `system`-role test asserts the underlying `MockLanguageModelV4.doStream` is never called (i.e. the request fails closed before reaching the model).
- Verified the new `callOptionsSchema` test rejects an out-of-enum `topic` and accepts an in-enum value end-to-end.
- Reproduced the prototype-collision case in a Node REPL (`'constructor' in {}` → `true`, `Object.hasOwn({}, 'constructor')` → `false`) to confirm the fix changes behavior only on collision keys.

## Checklist

Documentation is not updated because the fixes preserve documented behavior — `instructions` remains the supported way to set system prompts, `callOptionsSchema` now actually does what its name and JSDoc imply, and `getMediaTypeFromUrl` is a private helper.

## Future Work

- Consider tightening `uiMessagesSchema` itself to `z.enum(['user', 'assistant'])` (or splitting client-trust from server-trust constructions) so `validateUIMessages` is safe by default in any inbound-handler context, not just inside `createAgentUIStream`. Left out of this PR to avoid changing the public shape of `validateUIMessages` for server-side callers that programmatically construct system messages.

### DIFF
--- a/.changeset/reject-system-role-ui-messages.md
+++ b/.changeset/reject-system-role-ui-messages.md
@@ -1,0 +1,7 @@
+---
+"ai": patch
+---
+
+fix(agent): reject `role: 'system'` in `createAgentUIStream` / `createAgentUIStreamResponse` inputs
+
+Client-supplied UI messages cross an untrusted boundary. Previously, a `role: 'system'` entry would be forwarded to the model alongside the developer's `instructions`, enabling prompt injection by any caller. The agent UI stream now throws `InvalidArgumentError` if any UI message has `role: 'system'`. Set system instructions on the agent via the `instructions` setting instead.
--- a/packages/ai/src/agent/create-agent-ui-stream-response.test.ts
+++ b/packages/ai/src/agent/create-agent-ui-stream-response.test.ts
@@ -6,6 +6,7 @@ import {
 } from '@ai-sdk/provider-utils/test';
 import { beforeEach, describe, expect, it } from 'vitest';
 import { z } from 'zod/v4';
+import { InvalidArgumentError } from '../error';
 import { MockLanguageModelV4 } from '../test/mock-language-model-v4';
 import { createAgentUIStreamResponse } from './create-agent-ui-stream-response';
 import { ToolLoopAgent } from './tool-loop-agent';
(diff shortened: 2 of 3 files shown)


lgrammel:
a) i agree `system` in user messages is an issue (and never wanted to introduce it). that said, there was strong demand for this feature. what we could do in ai sdk 7 is to make the ui message validation more strict (reject system messages as input by default), with an option to allow them (opt-in). this might be worth splitting out into a separate pr with a concise change and migration guide update

b) call option schema: this would be great, ideally should be a separate dedicated pr

c) a separate pr would be great just for media type

lgrammel:
Re a) thinking about it more, I would prefer the check to be part of the model message scanning. I will take this on.

etairl:
> a) i agree `system` in user messages is an issue (and never wanted to introduce it). that said, there was strong demand for this feature. what we could do in ai sdk 7 is to make the ui message validation more strict (reject system messages as input by default), with an option to allow them (opt-in). this might be worth splitting out into a separate pr with a concise change and migration guide update
> 
> b) call option schema: this would be great, ideally should be a separate dedicated pr
> 
> c) a separate pr would be great just for media type

Gotcha, I'll create separate PRs for B & C and leave A to you :)

etairl:
> a) i agree `system` in user messages is an issue (and never wanted to introduce it). that said, there was strong demand for this feature. what we could do in ai sdk 7 is to make the ui message validation more strict (reject system messages as input by default), with an option to allow them (opt-in). this might be worth splitting out into a separate pr with a concise change and migration guide update
> 
> b) call option schema: this would be great, ideally should be a separate dedicated pr
> 
> c) a separate pr would be great just for media type

Split and created separated PRs:

https://github.com/vercel/ai/pull/14750
https://github.com/vercel/ai/issues/14751

PR #14750, opened 2026-04-27
fix(ai): enforce callOptionsSchema at runtime in ToolLoopAgent

etairl:
## Background

`ToolLoopAgentSettings.callOptionsSchema` is declared and documented as a runtime schema for caller-supplied `options`, but `ToolLoopAgent.prepareCall` never invokes it. Any invariant a developer encodes in that schema is silently bypassed at runtime, and unchecked `options` flow straight into `prepareCall` and any `instructions` template that interpolates them — defeating both the validation guarantee and any input-shape assumptions downstream code makes.

Splitting this out per maintainer feedback on #14749 that `callOptionsSchema` enforcement should be its own dedicated PR.

## Summary

- `packages/ai/src/agent/tool-loop-agent.ts`: at the top of `prepareCall`, when `callOptionsSchema` is set and the caller passed `options`, validate via `safeValidateTypes`. On failure, throw `InvalidArgumentError` with the schema's error message. On success, swap the caller-supplied `options` for the validated (parsed) value so any schema transforms or defaults take effect for the rest of the call.
- `packages/ai/src/agent/tool-loop-agent.test.ts`: regression tests covering the rejection path (out-of-enum value rejected before reaching the model) and the accept path (in-enum value passes through and the model is invoked normally).
- Patch changeset.

The check is gated on `options !== undefined`, so existing callers that don't supply `options` are unaffected. Only agents that opted into `callOptionsSchema` see new behaviour — and that behaviour is exactly what the field name and docs already promised.

## Manual Verification

- `pnpm --filter ai test:node -- src/agent/tool-loop-agent.test.ts` — passes, including the two new `callOptionsSchema` tests.

## Checklist

Documentation is unchanged: the existing `callOptionsSchema` JSDoc already describes the intended behaviour; this PR makes the runtime match the docs.

## Future Work

None — this PR makes `callOptionsSchema` behave as documented.

## Related Issues

- Split out from #14749 (closed) per maintainer feedback.

> Detected automatically by [https://github.com/etairl/Probus](https://github.com/vercel/ai/pull/Probus)

### DIFF
--- a/packages/ai/src/agent/tool-loop-agent.ts
+++ b/packages/ai/src/agent/tool-loop-agent.ts
@@ -1,6 +1,5 @@
 import type { Context, ModelMessage, ToolSet } from '@ai-sdk/provider-utils';
-import { safeValidateTypes } from '@ai-sdk/provider-utils';
-import { InvalidArgumentError } from '../error';
+import { validateTypes } from '@ai-sdk/provider-utils';
 import { generateText } from '../generate-text/generate-text';
 import { GenerateTextResult } from '../generate-text/generate-text-result';
 import { Output } from '../generate-text/output';
@@ -85,26 +84,16 @@ export class ToolLoopAgent<
     > &
       Prompt
   > {
-    // Validate caller-supplied `options` against `callOptionsSchema` if one
-    // was provided. Without this, the schema was effectively dead code:
-    // developers wiring untrusted input into `options` would get no runtime
-    // protection despite naming the field `callOptionsSchema`.
     if (
       this.settings.callOptionsSchema != null &&
       options.options !== undefined
     ) {
-      const result = await safeValidateTypes({
+      const validatedOptions = await validateTypes({
         value: options.options,
         schema: this.settings.callOptionsSchema,
+        context: { field: 'options' },
       });
-      if (!result.success) {
-        throw new InvalidArgumentError({
-          parameter: 'options',
-          value: options.options,
-          message: `options failed callOptionsSchema validation: ${result.error.message}`,
-        });
-      }
-      options = { ...options, options: result.value };
+      options = { ...options, options: validatedOptions };
     }
 
     const {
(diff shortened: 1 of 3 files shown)


packages/ai/src/agent/tool-loop-agent.ts
```diff
@@ -83,6 +85,28 @@ export class ToolLoopAgent<
     > &
       Prompt
   > {
+    // Validate caller-supplied `options` against `callOptionsSchema` if one
+    // was provided. Without this, the schema was effectively dead code:
+    // developers wiring untrusted input into `options` would get no runtime
+    // protection despite naming the field `callOptionsSchema`.
```
lgrammel: comment not needed

packages/ai/src/agent/tool-loop-agent.ts
```diff
@@ -83,6 +85,28 @@ export class ToolLoopAgent<
...
+        schema: this.settings.callOptionsSchema,
+      });
+      if (!result.success) {
+        throw new InvalidArgumentError({
+          parameter: 'options',
+          value: options.options,
+          message: `options failed callOptionsSchema validation: ${result.error.message}`,
+        });
+      }
+      options = { ...options, options: result.value };
```
lgrammel: can just call `validateTypes` and rely on the original parse error. `validateTypes` supports additional `context`

lgrammel (approved):

lgrammel:
@etairl we have changed our security policy to allow only commits with verified signatures. I cannot merge as is. Would it be possible for you to enable commit signing and open a new PR with signed commits? alternatively I can recreate this

etairl:
> @etairl we have changed our security policy to allow only commits with verified signatures. I cannot merge as is. Would it be possible for you to enable commit signing and open a new PR with signed commits? alternatively I can recreate this

Rebased with signed commits.

PR #14751, opened 2026-04-27
fix(ai): harden getMediaTypeFromUrl against prototype-property collision

etairl:
## Background

`getMediaTypeFromUrl` in `convert-to-language-model-prompt.ts` does an `ext in URL_EXTENSION_TO_MEDIA_TYPE` check on a plain object literal. Because plain objects inherit from `Object.prototype`, the `in` operator returns `true` for inherited keys like `constructor`, `toString`, `hasOwnProperty`, etc. A URL ending in `.constructor` (or any other `Object.prototype` member) therefore takes the lookup branch and returns the inherited value — for `.constructor`, that's the `Object` constructor function, which is then forwarded as `mediaType` to provider adapters.

This is a low-severity correctness/typing bug rather than an exploit path, but it's worth fixing: the helper's return type is `string | undefined` and a non-string slipping through can break downstream code paths that assume a string `mediaType`.

Splitting this out per maintainer feedback on #14749 that the media-type fix should be its own PR.

## Summary

- `packages/ai/src/prompt/convert-to-language-model-prompt.ts`: replace `ext in URL_EXTENSION_TO_MEDIA_TYPE` with `Object.hasOwn(URL_EXTENSION_TO_MEDIA_TYPE, ext)` so only own-property extensions are matched.
- `packages/ai/src/prompt/convert-to-language-model-prompt.test.ts`: regression test for a URL ending in `.constructor` — asserts the helper falls back to the no-extension behaviour instead of returning a non-string value from the prototype chain.
- Patch changeset.

The change is one line of production code; the table is treated as a closed lookup, which is what the original code intended.

## Manual Verification

- `pnpm --filter ai test:node -- src/prompt/convert-to-language-model-prompt.test.ts` — passes, including the new prototype-collision regression test.

## Checklist

No documentation change — `getMediaTypeFromUrl` is internal.

## Future Work

None. If other helpers in the prompt layer use `in` against plain object literals indexed by user input, the same prototype-confusion fix would apply, but I didn't spot any others while looking at this one.

## Related Issues

- Split out from #14749 (closed) per maintainer feedback.

> Detected automatically by [https://github.com/etairl/Probus](Probus)

### DIFF
--- a/packages/ai/src/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/src/prompt/convert-to-language-model-prompt.ts
@@ -742,9 +742,6 @@ function getMediaTypeFromUrl(
   try {
     const pathname = new URL(url).pathname;
     const ext = pathname.split('.').pop()?.toLowerCase();
-    // Use `Object.hasOwn` instead of `in` so attacker-controlled extensions
-    // like `constructor` cannot resolve to inherited `Object.prototype`
-    // members and leak a non-string value through this `: string` helper.
     if (ext && Object.hasOwn(URL_EXTENSION_TO_MEDIA_TYPE, ext)) {
       return URL_EXTENSION_TO_MEDIA_TYPE[ext];
     }
--- a/packages/ai/src/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/src/prompt/convert-to-language-model-prompt.ts
@@ -742,7 +742,10 @@ function getMediaTypeFromUrl(
   try {
     const pathname = new URL(url).pathname;
     const ext = pathname.split('.').pop()?.toLowerCase();
-    if (ext && ext in URL_EXTENSION_TO_MEDIA_TYPE) {
+    // Use `Object.hasOwn` instead of `in` so attacker-controlled extensions
+    // like `constructor` cannot resolve to inherited `Object.prototype`
+    // members and leak a non-string value through this `: string` helper.
+    if (ext && Object.hasOwn(URL_EXTENSION_TO_MEDIA_TYPE, ext)) {
       return URL_EXTENSION_TO_MEDIA_TYPE[ext];
     }
   } catch {
--- a/.changeset/get-media-type-from-url-prototype.md
+++ b/.changeset/get-media-type-from-url-prototype.md
@@ -1,0 +1,9 @@
+---
+"ai": patch
+---
+
+fix(ai): harden `getMediaTypeFromUrl` against prototype-property collisions
+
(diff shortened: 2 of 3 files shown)


packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -742,7 +742,10 @@ function getMediaTypeFromUrl(
   try {
     const pathname = new URL(url).pathname;
     const ext = pathname.split('.').pop()?.toLowerCase();
-    if (ext && ext in URL_EXTENSION_TO_MEDIA_TYPE) {
+    // Use `Object.hasOwn` instead of `in` so attacker-controlled extensions
+    // like `constructor` cannot resolve to inherited `Object.prototype`
+    // members and leak a non-string value through this `: string` helper.
```
lgrammel: comment not needed (there is a regression test)

lgrammel (approved):

lgrammel:
Leaving to @felixarntz to confirm since he added the code.

lgrammel:
@etairl can you reopen with signed commits?

etairl:
> @etairl can you reopen with signed commits?

Sure, will do in a few hours.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -742,7 +742,7 @@ function getMediaTypeFromUrl(
   try {
     const pathname = new URL(url).pathname;
     const ext = pathname.split('.').pop()?.toLowerCase();
-    if (ext && ext in URL_EXTENSION_TO_MEDIA_TYPE) {
+    if (ext && Object.hasOwn(URL_EXTENSION_TO_MEDIA_TYPE, ext)) {
```
felixarntz, line 745: @etairl Can we use this instead?
```suggestion
    if (ext && URL_EXTENSION_TO_MEDIA_TYPE[ext]) {
```

felixarntz (commented):
@etairl Thank you for opening the PR!

We have many cases of `prop in obj` in our codebase - how much of a concern is this really? If it is a concern, we should probably revisit the other places too. If not, then changing it here seems a bit arbitrary.

felixarntz (approved):
@etairl I briefly chatted about this with @lgrammel, the fix looks good. Can you please recreate this with signed commits?

Afterwards, we should review our codebase for other usage of `prop in obj` and replace as applicable too.

packages/ai/src/prompt/convert-to-language-model-prompt.ts
lgrammel: I think `Object.hasOwn` is the right choice here as it is more strict.

etairl:
> @etairl I briefly chatted about this with @lgrammel, the fix looks good. Can you please recreate this with signed commits?
> 
> Afterwards, we should review our codebase for other usage of `prop in obj` and replace as applicable too.

Done, rebased with signed commits.

PR #14752, opened 2026-04-27
fix(ai): reject system messages in messages or prompt by default (opt-in)

lgrammel:
## Background

For historical and convenience reasons, system messages can be part of user messages or prompts, e.g. to allow interleaving regular messages and system messages.

However, this creates a prompt injection risk where the user (e.g. by modifying the messages in a web ui) can override or set the system prompt. In most cases, it should only be possible to set the system prompt via the system (or instructions) property, and users should not be able to inject system messages.

## Summary
* throw `InvalidPromptError` when there are system messages in the messages or prompt options
* add `allowSystemInMessages` option for opting into allowing system messages in messages or prompt options

## Future Work

* add `allowSystemInMessages` opt-in support to `WorkflowAgent` (if desired) @gr2m 

## Related Issues

Issue reported in #14749

### DIFF
--- a/packages/workflow/src/do-stream-step.ts
+++ b/packages/workflow/src/do-stream-step.ts
@@ -115,6 +115,7 @@ export async function doStreamStep(
     // pre-converted LanguageModelV4Prompt. standardizePrompt inside
     // streamModelCall handles both formats.
     messages: conversationPrompt as unknown as ModelMessage[],
+    allowSystemInMessages: true,
     tools,
     toolChoice: options?.toolChoice,
     includeRawChunks: options?.includeRawChunks,
--- a/packages/workflow/src/workflow-agent.ts
+++ b/packages/workflow/src/workflow-agent.ts
@@ -1163,6 +1163,7 @@ export class WorkflowAgent<TBaseTools extends ToolSet = ToolSet> {
 
     const prompt = await standardizePrompt({
       system: effectiveInstructions,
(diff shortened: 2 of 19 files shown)


packages/workflow/src/workflow-agent.ts
```diff
@@ -1162,10 +1163,11 @@ export class WorkflowAgent<TBaseTools extends ToolSet = ToolSet> {
 
     const prompt = await standardizePrompt({
       system: effectiveInstructions,
+      allowSystemInMessages: true, // TODO: consider exposing this as a parameter
```
lgrammel, line 1166: @gr2m unclear to me whether we want to expose this flag or set it to false (which breaks tests)

felixarntz (approved):
LGTM, let's wait for @gr2m to confirm the workflow agent question.

packages/workflow/src/do-stream-step.ts
```diff
@@ -115,6 +115,7 @@ export async function doStreamStep(
     // pre-converted LanguageModelV4Prompt. standardizePrompt inside
     // streamModelCall handles both formats.
     messages: conversationPrompt as unknown as ModelMessage[],
+    allowSystemInMessages: true,
```
lgrammel, line 118: @gr2m same as below, need to decide how to handle this in workflow agent

nicoalbanese (approved):

PR #14765, opened 2026-04-27
fix(amazon-bedrock): drop foreign-provider reasoning blocks instead of sending them unsigned

thestonechat:
## Background

  When a conversation contains reasoning blocks signed by a different provider
  (e.g. Anthropic-signed reasoning replayed through Bedrock during cross-provider
  fallback), the converter pushes the reasoning text without a signature, and
  Bedrock's converse API rejects the request with:

    messages.N.content.M.thinking.signature: Field required

  `@ai-sdk/anthropic` already handles the symmetric case gracefully — it drops
  unrecognized reasoning metadata and emits a warning.

  ## Fix

  In the `case 'reasoning'` branch of `convertToBedrockChatMessages`, only fall
  through to the unsigned-reasoning push when `providerOptions` is empty or
  contains exclusively the `bedrock` key. Foreign-provider metadata (e.g.
  `anthropic`) causes the block to be dropped instead.

  This preserves the existing prefill behavior introduced in #13972 while
  fixing the cross-provider replay case.

  ## Repro

  A conversation that has assistant turns with `providerOptions.anthropic`
  on its reasoning parts, sent through `bedrock(...).doStream(...)`, currently
  reproduces the validation error. The added test covers this.

  After gh pr create runs, it'll print the PR URL. CI will run automatically; the CLA bot will comment with a sign link if you
  haven't signed before.

### DIFF
--- a/packages/amazon-bedrock/src/convert-to-amazon-bedrock-chat-messages.test.ts
+++ b/packages/amazon-bedrock/src/convert-to-amazon-bedrock-chat-messages.test.ts
@@ -712,6 +712,88 @@ describe('assistant messages', () => {
     });
   });
 
+  it('should omit assistant message reasoning parts signed by a foreign provider', async () => {
+    const result = await convertToAmazonBedrockChatMessages([
+      {
+        role: 'user',
+        content: [{ type: 'text', text: 'Explain your reasoning' }],
+      },
+      {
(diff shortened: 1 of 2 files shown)


tigerabrodi:
good stuff.

dancer (approved):

PR #14786, opened 2026-04-28
fix(workflow): expose totalUsage and finishReason on stream result

VaguelySerious:
## Summary

`WorkflowAgentStreamResult` (the awaited return of `agent.stream()`) is missing `totalUsage` and `finishReason`. Both fields are already computed in the runtime and passed to the `onFinish` callback, so the omission from the result is asymmetric and diverges from `GenerateTextResult`/`StreamTextResult`, which expose them directly.

This adds them to the interface and to all three return sites (early-abort, paused-tools, and the final return), reusing the existing `aggregateUsage(steps)` helper.

Targets the `workflow-agent-unify-callbacks` branch since the `WorkflowAgent` work is gated behind it.

See also https://github.com/vercel/workflow/pull/1863

## Test plan

- [x] New regression test in `workflow-agent-compat.test.ts` asserts `result.totalUsage` and `result.finishReason` after a normal stream
- [x] Existing 107 tests still pass (108 with the new one), 5 expected-fails unchanged
- [x] `pnpm turbo run build --filter=@ai-sdk/workflow` passes

### DIFF
--- a/.changeset/workflow-agent-stream-result-totalusage.md
+++ b/.changeset/workflow-agent-stream-result-totalusage.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/workflow': patch
+---
+
+Expose `totalUsage` and `finishReason` on the `WorkflowAgent.stream()` result, mirroring `GenerateTextResult`/`StreamTextResult` and the existing `onFinish` event payload.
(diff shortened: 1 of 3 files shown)


gr2m:
Superseded by #16841, which applies the same change against `main` — this PR's base branch `workflow-agent-unify-callbacks` is stale (main is 800+ commits ahead and `workflow-agent.ts` has been rewritten since). The upstream sibling vercel/workflow#1863 already merged.

PR #13157, opened 2026-03-06
feat(openai): add new tool search tool

aayush-kapoor:
## Background

OpenAI introduced a new tool along with the release of GPT 5.4

https://developers.openai.com/api/docs/guides/tools-tool-search

API spec here: https://developers.openai.com/api/reference/resources/responses/methods/retrieve

## Summary

- added the new tool spec
- ensure tool conversion takes place properly

## Manual Verification

verified by running the examples: 

- `examples/ai-functions/src/generate-text/openai/responses-tool-search.ts`
- `examples/ai-functions/src/stream-text/openai/responses-tool-search.ts`
- `‎examples/ai-functions/src/generate-text/openai/responses-client-tool-search.ts‎`
- `‎examples/ai-functions/src/stream-text/openai/responses-client-tool-search.ts‎`
- UI examples
  - ran `http://localhost:3000/chat/openai-tool-search` with store:false and store:true
  - `http://localhost:3000/chat/openai-tool-search-client` with store:false and store:true

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #14789, opened 2026-04-28
feat(openai): preserve namespace on function_call output items

R-Taneja:
## Summary

When the OpenAI Responses API dispatches a tool call from inside a \`namespace\` tool wrapper, the resulting \`function_call\` output item carries a \`namespace\` field identifying which namespace the dispatch came from. Example response from a Codex-style \`tool_search\` + \`namespace\` flow:

\`\`\`json
{
  "type": "function_call",
  "id": "fc_...",
  "call_id": "call_...",
  "name": "get_weather",
  "arguments": "{\"location\":\"NYC\"}",
  "status": "completed",
  "namespace": "weather_ns"
}
\`\`\`

The provider's response schema didn't include \`namespace\`, so zod stripped it during validation. Downstream consumers (e.g. Vercel AI Gateway's OpenResponses-compat layer) couldn't surface namespace information back to clients.

## Changes

- Add \`namespace: z.string().nullish()\` to the three \`function_call\` schemas in \`openai-responses-api.ts\` (response output, output items, \`response.output_item.done\` stream events).
- Forward \`part.namespace\` / \`value.item.namespace\` via \`providerMetadata.openai.namespace\` on the resulting \`tool-call\` content / stream parts in \`openai-responses-language-model.ts\` (both non-streaming and streaming paths).

Optional / nullish, so the change is fully backwards-compatible — no impact on requests that don't use namespace tools.

## Test plan

- [x] \`pnpm test:node\` in \`packages/openai\` — 680/680 pass; 2 new tests cover present/absent cases.
- [x] \`pnpm type-check:full\` clean.
- [x] \`pnpm check\` clean.

## Related

- vercel/ai#13157 — added \`tool_search\` (related Codex feature).
- vercel/ai-gateway#1806 — gateway PR that propagates this field through to OpenResponses output (becomes effective once this lands and the gateway bumps).

## Backports

If accepted, this should also backport to release-v6.0 (and possibly release-v5.0) so existing Codex traffic on the gateway gets the field. Happy to author those backports if the team prefers.


packages/openai/src/responses/openai-responses-language-model.ts
shaper, line 1: add generate and stream examples we can run to validate the behavior

felixarntz (commented):
@R-Taneja code change looks reasonable, can you add an example that makes use of namespaces for tools so we can verify this? https://developers.openai.com/api/docs/guides/function-calling#defining-namespaces

R-Taneja:
> @R-Taneja code change looks reasonable, can you add an example that makes use of namespaces for tools so we can verify this? https://developers.openai.com/api/docs/guides/function-calling#defining-namespaces

added!

felixarntz (approved):

mzruya:
Hey, [this](https://github.com/vercel/ai/pull/15193) fixes the read path. Could someone have a look?

lgrammel:
Documentation needs to be updated.

PR #14790, opened 2026-04-28
feat(anthropic): sanitize the unsupported JSON schema validation properties

### LINKED ISSUE #13355: @ai-sdk/anthropic: Unsupported JSON Schema validation properties not stripped for structured output
## Description

When using `Output.object()` with Zod schemas for structured output, `@ai-sdk/anthropic` passes the raw JSON Schema (from `z.toJSONSchema()`) directly to the Anthropic API. This causes `400 Bad Request` errors because Anthropic's API rejects schemas containing unsupported validation properties.

The **native Anthropic SDK** (`@anthropic-ai/sdk`) handles this by automatically stripping unsupported properties and moving them to field `description` strings, then validating responses client-side against the original schema. The Vercel AI SDK does not replicate this behavior.

## Reproduction

```ts
import { generateText } from "ai";
import { Output } from "ai";
import { z } from "zod";

const schema = z.object({
  items: z.array(z.string().min(1)).min(2).max(5),
  count: z.number().int().min(0).max(10),
});

// This fails with 400 Bad Request from Anthropic
const { experimental_output } = await generateText({
  model: anthropic("claude-sonnet-4-20250514"),
  prompt: "Generate 3 items",
  experimental_output: Output.object({ schema }),
});
```

The generated JSON Schema contains `minItems`, `maxItems`, `minimum`, `maximum`, and `minLength` — all rejected by Anthropic.

## Unsupported properties

Per [Anthropic's structured output docs](https://docs.anthropic.com/en/docs/build-with-claude/structured-output), the following JSON Schema properties are **not supported**:

- `minLength`, `maxLength`
- `pattern`, `format`
- `minimum`, `maximum`
- `exclusiveMinimum`, `exclusiveMaximum`
- `minItems`, `maxItems`, `uniqueItems`
- `minProperties`, `maxProperties`

Additionally, `z.number().int()` in Zod 4 emits `"minimum": -9007199254740991, "maximum": 9007199254740991` (safe integer bounds), which also triggers the rejection.

## Expected behavior

`@ai-sdk/anthropic` should sanitize JSON schemas before sending them to the API, matching the behavior of the native Anthropic SDK:

1. Strip unsupported validation properties from the schema
2. Optionally move stripped constraints to field `description` strings
3. Validate responses client-side against the original (full) schema

## Current workaround

Remove all validation constraints from Zod schemas used with Anthropic structured output and validate in post-parse:

```ts
// Instead of:
const schema = z.object({
  count: z.number().int().min(0).max(6),
  title: z.string().trim().min(1),
  items: z.array(z.string()).min(2).max(5),
});

// Use bare types for the LLM schema:
const schema = z.object({
  count: z.number(),
  title: z.string(),
  items: z.array(z.string()),
});

// Then validate constraints in post-parse code
```

## Related issues

- #12876 — Same class of problem for `oneOf` (fixed in #12903/#12954), but the fix only handles that one keyword
- #5360 — User reported needing to remove `.min()`/`.max()` for OpenAI strict mode

## Environment

- `ai`: 4.x
- `@ai-sdk/anthropic`: latest
- Zod 4 (`z.toJSONSchema()`)
- Also reproducible via AI Gateway routing to Anthropic

### LINKED ISSUE #14342: Anthropic structured outputs fail with valid Zod schemas due to unsupported JSON Schema keywords (e.g. exclusiveMinimum, not, minimum, maximum)
## Summary

When using Anthropic structured outputs via Vercel AI SDK, valid Zod schemas can cause 400 errors because unsupported JSON Schema keywords (e.g. `exclusiveMinimum`) are passed through without sanitization.

Anthropic's API strictly validates `output_config.format.schema`, unlike tool schemas where unsupported keywords are ignored. This leads to inconsistent behavior depending on the code path.

Anthropic’s official SDK already addresses this issue via helpers such as:

- `zodOutputFormat`
- `jsonSchemaOutputFormat`

These helpers:
- convert Zod / JSON Schema into an Anthropic-compatible format
- strip or transform unsupported keywords (e.g. `exclusiveMinimum`, `not`)
- ensure compatibility with `output_config.format.schema`

Docs: [Anthropic Structured Outputs (TypeScript)](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#typescript)

## Reproduction

Given a Zod schema:

```ts
const schema = z.object({
  recurringIntervalMinutes: z.number().positive(),
});
```

### AI SDK Version

ai: 6.0.12

### Code of Conduct

aayush-kapoor:
## Background

https://github.com/vercel/ai/issues/13355 & https://github.com/vercel/ai/issues/14342

Anthropic rejects certain schema validation properties (documented [here](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations))

When those properties are passed through for structured output definitions (not for tool definitions), Anthropic returns a 400 error.

## Summary

Introduce a `sanitize-json-schema.ts` file that strips the rejected properties, defined in a set. 
However to ensure full validation, the schema is then converted to a simple string and passed through as `description()` 

## Manual Verification

verified by running the examples: 
- `examples/ai-functions/src/generate-text/anthropic/output-object-zod-positive-number.ts`
- `examples/ai-functions/src/generate-text/anthropic/output-object-zod.ts`

failed before the fix and worked (and respected the constraint if any) after the fix

## Checklist

## Related Issues

fixes #13355 and #14342

### DIFF
--- a/packages/anthropic/src/sanitize-json-schema.ts
+++ b/packages/anthropic/src/sanitize-json-schema.ts
@@ -183,9 +183,7 @@ function getConstraintDescription(schema: JSONSchema7): string | undefined {
     descriptions.push(`format: ${schema.format}`);
   }
 
-  return descriptions.length === 0
-    ? undefined
-    : `Additional constraints: ${descriptions.join('; ')}.`;
+  return descriptions.length === 0 ? undefined : `${descriptions.join('; ')}.`;
 }
 
 function formatConstraintName(key: string): string {
--- a/packages/anthropic/src/sanitize-json-schema.ts
+++ b/packages/anthropic/src/sanitize-json-schema.ts
@@ -1,0 +1,206 @@
+import type { JSONSchema7, JSONSchema7Definition } from '@ai-sdk/provider';
+
+const SUPPORTED_STRING_FORMATS = new Set([
+  'date-time',
+  'time',
+  'date',
+  'duration',
+  'email',
+  'hostname',
+  'uri',
+  'ipv4',
+  'ipv6',
+  'uuid',
+]);
+
+const DESCRIPTION_CONSTRAINT_KEYS = [
+  'minimum',
+  'maximum',
+  'exclusiveMinimum',
+  'exclusiveMaximum',
+  'multipleOf',
+  'minLength',
+  'maxLength',
+  'pattern',
+  'minItems',
+  'maxItems',
+  'uniqueItems',
+  'minProperties',
+  'maxProperties',
+  'format',
+  'not',
+] satisfies Array<keyof JSONSchema7>;
+
+/**
+ * Removes JSON Schema keywords that Anthropic rejects in
+ * `output_config.format.schema`.
+ *
+ * The full original schema is still used by AI SDK result validation. This
+ * only relaxes the schema sent to Anthropic's constrained decoder.
+ */
+export function sanitizeJsonSchema(schema: JSONSchema7): JSONSchema7 {
+  return sanitizeSchema(schema) as JSONSchema7;
+}
+
+function sanitizeDefinition(
+  definition: JSONSchema7Definition,
+): JSONSchema7Definition {
+  if (typeof definition === 'boolean' || !isPlainObject(definition)) {
+    return definition;
+  }
+
+  return sanitizeSchema(definition as JSONSchema7);
+}
+
+function sanitizeSchema(schema: JSONSchema7): JSONSchema7 {
+  const result: JSONSchema7 = {};
+  const schemaWithDefs = schema as JSONSchema7 & {
+    $defs?: Record<string, JSONSchema7Definition>;
+  };
+
+  if (schema.$ref != null) {
+    return { $ref: schema.$ref };
+  }
+
+  if (schema.$schema != null) {
+    result.$schema = schema.$schema;
+  }
+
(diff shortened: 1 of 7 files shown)


packages/anthropic/src/sanitize-json-schema.ts
```diff
@@ -0,0 +1,205 @@
...
+  if (
+    typeof schema.format === 'string' &&
+    !SUPPORTED_STRING_FORMATS.has(schema.format)
+  ) {
+    descriptions.push(`format: ${schema.format}`);
+  }
+
+  return descriptions.length === 0
+    ? undefined
+    : `Additional constraints: ${descriptions.join('; ')}.`;
```
gr2m: I don't think it's a problem in this case, but made me think if this is a form of prompt altering that we are trying to stay out of. Also using English while a the rest of the prompt might be potentially a different language. I'm sure language model can handle it, curious what you think

gr2m (approved):

packages/anthropic/src/sanitize-json-schema.ts
lgrammel: what would be the impact just `descriptions.join`?

this is prompt engineering here that we are trying to avoid

lgrammel (approved):

PR #14792, opened 2026-04-28
chore: use `Object.hasOwn(obj, prop)` instead of `prop in obj`

felixarntz:
## Background

`prop in obj` walks the prototype chain, which is rarely the intent for own-property checks. `Object.hasOwn(obj, prop)` is the correct primitive. We want a lint rule that enforces this across the monorepo and to fix all existing violations.

This first came up in #14751, which led to the conclusion that we should fix it throughout our codebase and enforce it going forward.

## Summary

- Added new oxlint JS plugin at `tools/oxlint-plugins/no-in-operator.js` that flags any `BinaryExpression` with the `in` operator
- Bumped the shared `ts-library` and `react-library` TS lib targets to `ES2022` so `Object.hasOwn` is available everywhere.
- Added a global type-predicate augmentation for `Object.hasOwn` in `provider-utils/src/object-has-own.ts` (side-effect-imported from the package entry). This makes call-site code like `Object.hasOwn(options, 'enum') ? options.enum : undefined` narrow naturally — without it, `Object.hasOwn` returns plain `boolean` and TS doesn't narrow
- Replaced all `prop in obj` violations with `Object.hasOwn(obj, prop)`

## Checklist


gr2m (approved):

lgrammel (changes requested):
We should figure out which in which cases it makes sense (e.g. for inheritance) and in which cases it poses a risk (because of user injected data).

Replacing this everywhere changes behavior and can lead to breakages.

packages/vue/src/use-object.ts
```diff
@@ -102,7 +102,7 @@ export const experimental_useObject = function useObject<
   const key = `${api}|${completionId}`;
   const { data, mutate: originalMutate } = useSWRV<
     DeepPartial<RESULT> | undefined
-  >(key, () => (key in store ? store[key] : initialValue));
+  >(key, () => (Object.hasOwn(store, key) ? store[key] : initialValue));
```
lgrammel, line 105: flag

packages/svelte/src/utils.svelte.ts
```diff
@@ -13,7 +13,7 @@ export function createContext<T>(name: string) {
         if (
           typeof e === 'object' &&
           e !== null &&
-          'message' in e &&
+          Object.hasOwn(e, 'message') &&
```
lgrammel, line 16: flag

packages/rsc/src/provider.tsx
```diff
@@ -111,7 +111,7 @@ export function createAI<
     : undefined;
 
   const AI: AIProvider<AIState, UIState, Actions> = async props => {
-    if ('useState' in React) {
+    if (Object.hasOwn(React, 'useState')) {
```
lgrammel, line 114: flag

packages/rsc/src/util/is-async-generator.ts
```diff
@@ -2,6 +2,8 @@ export function isAsyncGenerator<T, TReturn, TNext>(
   value: unknown,
 ): value is AsyncGenerator<T, TReturn, TNext> {
   return (
```
lgrammel, line 4: flag

packages/rsc/src/util/is-generator.ts
```diff
@@ -1,5 +1,9 @@
 export function isGenerator<T, TReturn, TNext>(
   value: unknown,
 ): value is Generator<T, TReturn, TNext> {
-  return value != null && typeof value === 'object' && Symbol.iterator in value;
+  return (
+    value != null &&
+    typeof value === 'object' &&
+    typeof (value as Iterable<T>)[Symbol.iterator] === 'function'
```
lgrammel, line 7: flag

packages/rsc/src/streamable-value/is-streamable-value.ts
```diff
@@ -4,7 +4,7 @@ export function isStreamableValue(value: unknown): value is StreamableValue {
   return (
     value != null &&
     typeof value === 'object' &&
-    'type' in value &&
+    Object.hasOwn(value, 'type') &&
```
lgrammel, line 7: flag

packages/rsc/src/streamable-value/create-streamable-value.ts
```diff
@@ -15,14 +15,15 @@ const STREAMABLE_VALUE_INTERNAL_LOCK = Symbol('streamable.value.lock');
...
     (typeof initialValue === 'object' &&
       initialValue !== null &&
-      'getReader' in initialValue &&
-      typeof initialValue.getReader === 'function' &&
-      'locked' in initialValue &&
-      typeof initialValue.locked === 'boolean');
+      Object.hasOwn(initialValue, 'getReader') &&
+      typeof maybeStream?.getReader === 'function' &&
+      Object.hasOwn(initialValue, 'locked') &&
+      typeof maybeStream.locked === 'boolean');
```
lgrammel, line 27: flag

packages/rsc/src/shared-client/context.tsx
```diff
@@ -27,7 +27,7 @@ export function InternalAIProvider({
   wrappedActions,
   wrappedSyncUIState,
 }: InternalAIProviderProps) {
-  if (!('use' in React)) {
+  if (!Object.hasOwn(React, 'use')) {
```
lgrammel, line 30: flag

packages/provider-utils/src/object-has-own.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type HasOwnFilter<T, K extends PropertyKey> = T extends unknown
+  ? string extends keyof T
+    ? never
+    : K extends keyof T
+      ? T
+      : never
+  : never;
+
+export {};
```
lgrammel, line 42: having this in provider utils, not sure -- seems strange

packages/provider/src/errors/ai-sdk-error.ts
```diff
@@ -54,9 +54,9 @@ export class AISDKError extends Error {
     return (
       error != null &&
       typeof error === 'object' &&
-      markerSymbol in error &&
-      typeof error[markerSymbol] === 'boolean' &&
-      error[markerSymbol] === true
+      Object.hasOwn(error, markerSymbol) &&
+      typeof (error as Record<symbol, unknown>)[markerSymbol] === 'boolean' &&
+      (error as Record<symbol, unknown>)[markerSymbol] === true
```
lgrammel, line 59: flag and type safety

packages/provider-utils/src/schema.ts
```diff
@@ -242,7 +246,7 @@ export function isZod4Schema(
   zodSchema: z4.core.$ZodType<any, any> | z3.Schema<any, z3.ZodTypeDef, any>,
 ): zodSchema is z4.core.$ZodType<any, any> {
   // https://zod.dev/library-authors?id=how-to-support-zod-3-and-zod-4-simultaneously
-  return '_zod' in zodSchema;
+  return Object.hasOwn(zodSchema, '_zod');
```
lgrammel, line 249: flag

packages/provider-utils/src/schema.ts
```diff
@@ -122,25 +122,29 @@ function isSchema(value: unknown): value is Schema {
...
     typeof value === 'object' &&
     value !== null &&
-    schemaSymbol in value &&
-    value[schemaSymbol] === true &&
-    'jsonSchema' in value &&
-    'validate' in value
+    Object.hasOwn(value, schemaSymbol) &&
+    (value as Record<symbol, unknown>)[schemaSymbol] === true &&
+    Object.hasOwn(value, 'jsonSchema') &&
+    Object.hasOwn(value, 'validate')
```
lgrammel, line 128: flag

lgrammel:
I am concerned about restricting legit uses of `in`, e.g. when we cannot make assumptions about the prototype chain for external dependencies (such as validation libs) or in generic type checking functions. Instead of having a blanket rule that prevent `in`, we could search for cases where uses supplied data could cause security issues.

felixarntz:
Based on discussing this with @lgrammel, given the many scenarios where the code is external-facing and therefore we don't know whether developers are passing objects with prototype inheritance, it's risky to make all these changes; only a few of them are safe to make, but the general rule therefore seems overly noisy. And even for the changes that are okay to make, they're not necessarily security concerns like in #14751.

I'm going to close this for now.

In the future, we can think of an alternative, e.g. we could embed this consideration in a security focused review agent that could flag such instances in PR reviews.

PR #14805, opened 2026-04-29
feat(provider/xai): support non-image files in Responses API via input_file + file_url

### LINKED ISSUE #14804: @ai-sdk/xai Responses provider rejects non-image file parts even though xAI supports input_file + file_url
## Summary

`@ai-sdk/xai`'s Responses provider throws `UnsupportedFunctionalityError` for any file part whose `mediaType` does not start with `image/`, even when the caller provides a public HTTPS URL — but xAI's Responses API itself fully supports non-image documents (PDF, plain text, CSV, etc.) via `{ type: 'input_file', file_url }`.

This makes it impossible to use `xai.responses(...)` with PDFs through the AI SDK today, which is the headline use case demonstrated in xAI's own docs.

## Reproduction

```ts
import { createXai } from '@ai-sdk/xai';

const xai = createXai({ apiKey: process.env.XAI_API_KEY! });

  model: xai.responses('grok-4-fast'),
  messages: [
    {
      role: 'user',
      content: [
        { type: 'text', text: 'Summarize this document.' },
        {
          type: 'file',
          mediaType: 'application/pdf',
          data: new URL('https://docs.x.ai/assets/api-examples/documents/sales-report.txt'),
        },
      ],
    },
  ],
});
```

Throws:

```
AI_UnsupportedFunctionalityError: 'file part media type application/pdf' functionality not supported.
```

## Expected behavior

xAI's [Chat with Files](https://docs.x.ai/docs/guides/chat-with-files) docs show this exact shape working through `/v1/responses`:

```json
{
  "type": "input_file",
  "file_url": "https://docs.x.ai/assets/api-examples/documents/sales-report.txt"
}
```

The provider should serialize a non-image file part with URL data as `{ type: 'input_file', file_url: '<url>' }` instead of throwing.

## Root cause

[`convert-to-xai-responses-input.ts`](https://github.com/vercel/ai/blob/main/packages/xai/src/responses/convert-to-xai-responses-input.ts):

```ts
case 'file': {
  if (block.mediaType.startsWith('image/')) {
    // ...emits input_image
  } else {
    throw new UnsupportedFunctionalityError({
      functionality: `file part media type ${block.mediaType}`,
    });
  }
}
```

There is no branch for non-image files, even though the underlying `XaiResponsesUserMessageContentPart` type would just need a small extension to accept `file_url`. Additionally, `xai-responses-language-model.ts`'s `supportedUrls` only whitelists `image/*`, so PDF URLs get downloaded to bytes by the SDK before ever reaching the converter — which means even if the converter knew how to handle URLs, it would never see one.

## Proposed fix

PR submitted: #14805

- Extend `XaiResponsesUserMessageContentPart` to include `{ type: 'input_file'; file_url: string }`.
- In the converter, when a file part has `data: URL` and a non-image media type, emit `{ type: 'input_file', file_url }`.
- Add `application/pdf` and `text/*` to `supportedUrls` so URLs are passed through.
- Inline-byte (base64) inputs for non-image media types continue to throw (xAI Responses requires URL or Files API `file_id` for non-image documents); error message updated to point callers in the right direction.

## Workaround until merged

Either:

1. Route PDFs to a different provider (Gemini/OpenAI both work fine with `{ type: 'file', mediaType: 'application/pdf', data: URL }` through their AI SDK providers).
2. Skip the AI SDK and call `https://api.x.ai/v1/responses` directly with the documented `input_file` shape.

## Environment

- `@ai-sdk/xai`: 3.0.83 (latest at time of report)
- `ai`: 6.0.168
- Node 22

BrianP8701:
## Background

xAI's Responses API accepts non-image documents (PDF, plain text, CSV, etc.) as `{ type: 'input_file', file_url }` per the official [Chat with Files](https://docs.x.ai/docs/guides/chat-with-files) docs. The AI SDK xAI Responses provider previously threw `UnsupportedFunctionalityError` for any file part whose `mediaType` did not start with `image/`, even when the caller provided a public HTTPS URL — silently turning a fully supported xAI feature into a SDK-side rejection.

Closes #14804.

## Changes

- **`convert-to-xai-responses-input.ts`** — when a file part has `data: URL` and a non-image media type, emit `{ type: 'input_file', file_url }` instead of throwing. Inline-byte (base64) inputs for non-image media types still throw, with a clearer message pointing callers toward URLs or the Files API (xAI's Responses API does not accept inline bytes for non-image documents).
- **`xai-responses-language-model.ts`** — add `application/pdf` and `text/*` to `supportedUrls` so the SDK keeps these URLs as URLs rather than downloading them to bytes before reaching the converter.
- **`xai-responses-api.ts`** — extend the `input_file` content-part variant to include `file_url` alongside the existing `file_id`.
- **Tests** — add coverage for PDF-with-URL and text/csv-with-URL; the original "throws for unsupported file types" test is preserved (now scoped to inline bytes) since that case is still genuinely unsupported by the underlying API.

## Verification

```
$ pnpm --filter @ai-sdk/xai vitest run

 Test Files  12 passed (12)
      Tests  305 passed (305)
```

End-to-end repro against the fork resolves cleanly: a `mediaType: 'application/pdf'` + `data: URL` file part now produces a valid response from `xai.responses('grok-4-fast')` instead of throwing.

## Scope

- No public API changes outside the new `file_url` variant on `XaiResponsesUserMessageContentPart` (purely additive — the existing `file_id` variant is unchanged).
- No behavior change for image file parts.
- No behavior change for `xai.chat(...)` / `xai(...)` (chat completions endpoint), which does not support file parts at all.

### DIFF
--- a/packages/xai/src/responses/xai-responses-language-model.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.ts
@@ -80,6 +80,11 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
 
   readonly supportedUrls: Record<string, RegExp[]> = {
     'image/*': [/^https?:\/\/.*$/],
+    // xAI's Responses API accepts non-image documents (PDF, plain text, CSV, etc.) as
+    // `{ type: 'input_file', file_url }`. Whitelisting their URLs here keeps them as URLs
+    // through to the converter instead of being downloaded to bytes by the SDK.
+    'application/pdf': [/^https?:\/\/.*$/],
+    'text/*': [/^https?:\/\/.*$/],
   };
 
   private async getArgs({
--- a/packages/xai/src/responses/xai-responses-language-model.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.ts
@@ -81,7 +81,7 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
   readonly supportedUrls: Record<string, RegExp[]> = {
     'image/*': [/^https?:\/\/.*$/],
     // xAI's Responses API accepts non-image documents (PDF, plain text, CSV, etc.) as
-    // `{ type: 'input_file', file_url }`. Whitelisting their URLs here keeps them as URLs
+    // `{ type: 'input_file', file_url }`. Keeping these URLs intact here lets them pass
     // through to the converter instead of being downloaded to bytes by the SDK.
     'application/pdf': [/^https?:\/\/.*$/],
     'text/*': [/^https?:\/\/.*$/],
--- a/.changeset/xai-responses-pdf-file-url.md
+++ b/.changeset/xai-responses-pdf-file-url.md
@@ -1,0 +1,11 @@
+---
+'@ai-sdk/xai': patch
+---
+
+feat(provider/xai): support non-image file parts (PDF, text, CSV) in the Responses API via `input_file` + `file_url`
+
+The xAI Responses API accepts `{ type: 'input_file', file_url }` for non-image documents (see https://docs.x.ai/docs/guides/chat-with-files), but the AI SDK xAI Responses provider previously threw `UnsupportedFunctionalityError` for any file part whose `mediaType` did not start with `image/`.
+
+When a file part is passed with `data: URL` and a non-image media type, the provider now emits `{ type: 'input_file', file_url }`. `application/pdf` and `text/*` are also added to `supportedUrls` so the SDK does not download them to bytes before reaching the converter.
+
+Inline-byte (base64) inputs for non-image media types continue to throw, since xAI's Responses API requires either a public URL or a pre-uploaded `file_id` for non-image documents.
--- a/packages/xai/src/responses/convert-to-xai-responses-input.test.ts
+++ b/packages/xai/src/responses/convert-to-xai-responses-input.test.ts
@@ -146,7 +146,80 @@ describe('convertToXaiResponsesInput', () => {
       `);
     });
 
-    it('should throw for unsupported file types', async () => {
+    it('should convert non-image file parts with URL to input_file with file_url', async () => {
+      const result = await convertToXaiResponsesInput({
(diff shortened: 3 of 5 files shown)


gr2m (approved):
Verified that the reproduction from https://github.com/vercel/ai/issues/14804 passes with this PR. 

Please update PR description, follow our pull request template. All commits must be signed

https://github.com/vercel/ai/blob/216475e6a5754233dca955709b14315d9d0bd825/.github/pull_request_template.md#L33

BrianP8701:
> Verified that the reproduction from #14804 passes with this PR.
> 
> Please update PR description, follow our pull request template. All commits must be signed
> 
> https://github.com/vercel/ai/blob/216475e6a5754233dca955709b14315d9d0bd825/.github/pull_request_template.md#L33

Done!

packages/xai/src/responses/xai-responses-language-model.ts
```diff
@@ -80,6 +80,11 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
 
   readonly supportedUrls: Record<string, RegExp[]> = {
     'image/*': [/^https?:\/\/.*$/],
+    // xAI's Responses API accepts non-image documents (PDF, plain text, CSV, etc.) as
+    // `{ type: 'input_file', file_url }`. Whitelisting their URLs here keeps them as URLs
+    // through to the converter instead of being downloaded to bytes by the SDK.
```
gr2m, line 85: Avoid blacklisting/whitelisting, use allow/deny listing instead, or avoid the need for the term altogether.

```suggestion
    // xAI's Responses API accepts non-image documents (PDF, plain text, CSV, etc.) as
    // `{ type: 'input_file', file_url }`. Keeping these URLs intact here lets them pass
    // through to the converter instead of being downloaded to bytes by the SDK.
```

learn more: https://inclusivenaming.org/word-lists/tier-1/whitelist/

PR #13001, opened 2026-03-02
feat: start v7 pre-release cycle

gr2m:
## Summary

- [x] Create `release-v6.0` maintenance branch from current `main` HEAD for stable v6 patches
- Enter changeset pre-release mode (`beta`) so merges to `main` publish as `ai@7.0.0-beta.X`
- Add major changeset for all 53 published packages
- Create `LanguageModelV4` spec skeleton in `@ai-sdk/provider` (copy of V3 with renamed types)
- Create `LanguageModelV4Middleware` and `ProviderV4` types
- Add `MockLanguageModelV4` and `MockProviderV4` test utilities in `ai/test`

## Context

Tracking issue: #12999

Once merged, the first beta release (e.g. `ai@7.0.0-beta.1`) will publish automatically when the Version Packages PR is merged.

The V4 spec skeleton is an exact copy of V3 with type names updated to V4 and `specificationVersion: 'v4'`. The shared types (`SharedV3ProviderOptions`, `SharedV3ProviderMetadata`, etc.) are intentionally kept as-is since they are version-agnostic and shared across spec versions.

## Test plan

- [x] `@ai-sdk/provider` builds successfully with V4 types
- [x] `ai` package builds successfully with V4 mock utilities
- [x] V4 types are properly exported from `@ai-sdk/provider`
- [x] V4 mocks are properly exported from `ai/test`

PR #14811, opened 2026-04-29
fix(ai): fix auto-complete on provider registry and custom provider

### LINKED ISSUE #14780: Provider Registry typeahead for models stopped working
[packages/ai/src/registry/provider-registry.ts](https://github.com/vercel/ai/blob/f04adcbc168424a5c0179a2c7364ef45c730d137/packages/ai/src/registry/provider-registry.ts)

This worked before and is broken now. Add type tests to avoid future regressions

lgrammel:
## Background

Autocomplete on the provider registry stopped working. Most likely caused by the changes in #13001

## Summary

* restored auto-complete functionality
* add type tests
* add support of all model versions incl. string to CustomProvider

## Manual Verification

Check auto-completes and errors in:

- [x] `examples/ai-functions/src/registry/setup-registry.ts`

## Related Issues

Fixes #14780


packages/ai/src/registry/custom-provider.ts
```diff
@@ -105,24 +91,36 @@ export function customProvider<
...
-  const resolvedFiles = files ?? fallbackProvider?.files?.();
-  const resolvedSkills = skills ?? fallbackProvider?.skills?.();
-
-  return {
+} & (NonNullable<FILES> extends FilesV4
+    ? { files(): FilesV4 }
+    : { files?(): FilesV4 }) &
+  (NonNullable<SKILLS> extends SkillsV4
+    ? { skills(): SkillsV4 }
+    : { skills?(): SkillsV4 }) {
```
felixarntz: The `NonNullable` statements here are in a way problematic, because they break what originally worked in my implementation: when there's no `files` / `skills` respectively provided, the returned custom provider will have the corresponding property optional. Because here we make them non nullable, the `undefined` case is effectively dropped, making this I believe pointless (it'll always evaluate to having the `files` / `skills` property set for sure (even if that's not actually true).

Now, to be fair, I am now also realizing my original implementation was flawed too: it didn't consider `fallbackProvider`. Ideally, we would want `files` / `skills` to always be defined if either `FILES` / `SKILLS` is provided _or_ if the `fallbackProvider` has them defined. I'm not entirely sure whether that's possible in TypeScript though.

felixarntz: I attempted to fix this in https://github.com/vercel/ai/pull/14811/changes/d172f227a9be8f808ee0d9e3a32ccf76e2870430, looks like it's working as expected now. Let me know what you think.

You can see the behavior by adjusting e.g. `examples/ai-functions/src/registry/upload-file-custom-provider.ts`:
- e.g. remove the `files` entry, see a type error
- then add a `fallbackProvider` with files support, see the type error go away again
- then replace the `fallbackProvider` with a provider without files support, see the type error reappear

felixarntz (approved):
lgtm, verified that autocomplete is working again. I pushed one extra commit for your consideration (see other comment)

packages/ai/src/util/index.ts
```diff
@@ -4,6 +4,7 @@ export { cosineSimilarity } from './cosine-similarity';
 export { createDownload } from './download/create-download';
 export { getTextFromDataUrl } from './data-url';
 export type { DeepPartial } from './deep-partial';
+export type { ExtractLiteralUnion } from './extract-literal-union';
```
lgrammel, line 7: ugh this should not have been exposed

PR #14825, opened 2026-04-29
chore: gitignore v6+ example dirs that don't exist on v5

dancer:
## background

`examples/ai-e2e-next` and `examples/ai-functions` were added in v6+ and don't exist on v5. when switching between branches in the same working dir they show up as untracked on v5 which makes accidental commits possible. gitignoring them prevents that

## related issues

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ dist
 dist-ssr
 dist-bundle-check
 examples/*/build
+examples/ai-e2e-next
+examples/ai-functions
 node_modules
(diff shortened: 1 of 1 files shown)


felixarntz (commented):
@dancer not sure how much this will help, I think it might lead to examples being accidentally dropped.

when backporting to v5, new example files really should be moved from `ai-functions` to `ai-core` - if we ignore `ai-functions`, they might go unnoticed.

PR #14895, opened 2026-05-01
fix(workflow): preserve provider-executed approval responses

### LINKED ISSUE #14292: WorkflowAgent: provider-executed tool approvals are silently dropped on resume
## Problem

When a provider-executed tool (e.g., MCP tool via Responses API) requires approval and the client approves/denies it, the WorkflowAgent's approval resume path silently drops the approval response instead of forwarding it to the provider.

### Root cause

In `packages/workflow/src/workflow-agent.ts` (lines ~1045-1141), `collectToolApprovalsFromMessages` collects all approval responses. For locally-executed tools this works correctly — the tool is executed and results are added. But for provider-executed tools:

1. The tool either isn't in `this.tools` or has no `execute` function, so the `if (tool && typeof tool.execute === 'function')` check at line 1067 fails silently
2. No tool result is created for the approval
3. Lines 1111-1131 strip **all** `tool-approval-request` and `tool-approval-response` parts from messages
4. The provider never receives the approval response → the tool never gets executed

This is the inverse of the bug fixed in #14289 (where approval responses were sent to the provider **twice**). Here they are sent **zero times**.

### Note

The agent loop already handles provider-executed tools correctly during normal execution (lines ~1406-1410, using `providerExecutedToolResults` from the stream). The gap is specifically in the approval resume path.

### Suggested fix

`collectToolApprovalsFromMessages` needs to distinguish provider-executed approvals from local ones by tracking the `providerExecuted` flag from the original `tool-call` part. For provider-executed approvals, keep both `tool-approval-request` and `tool-approval-response` in messages so the provider can process them. Only strip/replace approval parts for locally-executed tools.

Genmin:
## Summary

Closes #14292.

- Track whether collected WorkflowAgent approvals came from provider-executed tool calls.
- Preserve provider-executed approval requests/responses through resume prompt cleanup, while continuing to strip and replace local approval parts with local tool results.
- Avoid creating synthetic local execution or denial results for provider-executed approvals so the provider receives the approval response exactly once.
- Add regression coverage for approved and denied provider-executed approval resumes.

## Tests

- `pnpm --filter @ai-sdk/workflow test:node -- src/workflow-agent.test.ts`
- `pnpm exec ultracite check packages/workflow/src/workflow-agent.ts packages/workflow/src/workflow-agent.test.ts`
- `pnpm changeset status --since origin/main`
- `git diff --check HEAD~1 HEAD`

## Checklist

- [ ] Documentation has been added / updated (not needed for this bug fix)
- [ ] Signed commits (local signing is not configured in this environment)

### DIFF
--- a/.changeset/fix-workflow-provider-approvals.md
+++ b/.changeset/fix-workflow-provider-approvals.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/workflow': patch
+---
+
+Preserve provider-executed tool approval responses when resuming WorkflowAgent streams.
--- a/packages/workflow/src/workflow-agent.test.ts
+++ b/packages/workflow/src/workflow-agent.test.ts
@@ -2999,6 +2999,172 @@ describe('WorkflowAgent', () => {
       expect(mockIterator.next).toHaveBeenCalled();
     });
 
+    it('should preserve approved provider-executed tool approvals for the provider', async () => {
+      const mockModel = createMockModel();
+
+      const agent = new WorkflowAgent({
+        model: mockModel,
+        tools: {},
+      });
+
+      const mockWritable = new WritableStream({
+        write: vi.fn(),
+        close: vi.fn(),
+      });
+
+      const { streamTextIterator } = await import('./stream-text-iterator.js');
+      vi.mocked(streamTextIterator).mockClear();
+      const mockIterator = {
+        next: vi.fn().mockResolvedValueOnce({ done: true, value: [] }),
+      };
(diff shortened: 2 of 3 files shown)


gr2m:
Thanks for this fix @Genmin 🙏 The approval-resume path was since refactored to use core's `collectToolApprovals` (the local `collectToolApprovalsFromMessages` helper this PR edits no longer exists on `main`), so the diff no longer applies. I've opened #16256 which reworks the fix against the current code — it uses your response re-stamping approach for robustness and credits you as a co-author.

PR #14289, opened 2026-04-09
fix(ai): fix `providerExecuted` tool approvals being passed to language model twice

felixarntz:
## Background

When a UI message history contains an approved `providerExecuted` tool invocation and the server does the standard pattern:

```ts
const modelMessages = await convertToModelMessages(messages);
return streamText({ model, messages: modelMessages });
```

the provider receives the same `tool-approval-response` **twice** for the same `approvalId` in one tool message. This can cause downstream providers to reject the prompt as invalid (one approval request must produce exactly one response).

## Summary

`collectToolApprovals` finds approval responses by reading the last tool message of `initialMessages`. Those responses are already present in `initialMessages` because `convertToModelMessages` placed them there. Both `generateText` and `streamText` then redundantly pushed the same responses into `responseMessages`/`initialResponseMessages`. When the step computed `stepInputMessages = [...initialMessages, ...responseMessages]`, the approval appeared twice, and `convertToLanguageModelPrompt` surfaced both copies inside a single merged tool message.

- Removed the redundant `providerExecutedToolApprovals` push in `generateText`
- Removed the same block in `streamText`
- Updated the four existing snapshots that already documented the bug (approved + denied cases for each function)

## Manual Verification

In `examples/ai-e2e-next`, navigate to `/chat/test-openai-responses-mcp-approval` and send `"Shorten the link https://ai-sdk.dev/"`. After approving the MCP tool call, the conversation should complete successfully.

To compare before/after, add a `console.log` of the prompt passed to the model in the mock or in the route handler. In `generateText`/`streamText`, you can log `stepInputMessages` (the variable at the top of the `do` loop / `streamStep` body) immediately before the language model call and confirm the tool message contains exactly one `tool-approval-response` entry after the fix.

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #14931, opened 2026-05-02
fix(workflow): preserve provider-executed tool-approval-response on resume

### LINKED ISSUE #14292: WorkflowAgent: provider-executed tool approvals are silently dropped on resume

MyPrototypeWhat:
Closes #14292.

## Summary

`WorkflowAgent` previously stripped every `tool-approval-response` part from messages when handling approval resumption, regardless of whether the underlying tool was locally or provider-executed. For provider-executed tools (e.g. MCP via the OpenAI Responses API) this silently dropped the approval before `convertToLanguageModelPrompt` could forward it, so the provider never learned of the approval and the tool was never executed.

This is the inverse of #14289, which fixed the symmetric bug where local approvals were being forwarded twice.

## Fix

In `packages/workflow/src/workflow-agent.ts`:

- `collectToolApprovalsFromMessages` now tracks `providerExecuted` from the `tool-approval-response` part itself. This is the same flag `convertToLanguageModelPrompt` already uses (line 360-361 of `convert-to-language-model-prompt.ts`) to decide which approval responses to forward to the model.
- The caller derives `localApprovedToolApprovals` and `localDeniedToolApprovals` subsets and a `localApprovalIds` `Set`. Local approvals retain the existing behavior (executed locally, results synthesized, parts stripped). Provider-executed approvals are skipped from local execution and their `tool-approval-request` and `tool-approval-response` parts are preserved in the messages so the next conversion forwards the response to the provider.

The fix is intentionally narrow: no change to the public API, no new dependencies, mirrors the discriminator the conversion layer already uses.

## Test plan

- [x] Added `should preserve provider-executed tool-approval-response across resume` to `tool approval resumption` describe block in `workflow-agent.test.ts`. The test constructs a provider-executed tool, resumes with an approved approval whose response carries `providerExecuted: true`, then asserts the response reaches the iterator and that no synthetic local tool-result was injected for it.
- [x] All existing tests in `tool approval resumption` (3) still pass.
- [x] Full workflow node test suite passes: 108 tests, 5 expected fails (pre-existing).
- [x] `pnpm check` (oxlint + oxfmt) clean.
- [x] `pnpm type-check:full` exit 0.

## Follow-ups out of scope

- The inverse-of-#14289 framing is consistent with this fix being a `patch` changeset. No `Experimental_*` API change.
- Behavioral parity with the core path (`convertToLanguageModelPrompt`) is now restored — both layers use the response's `providerExecuted` flag as the discriminator.

### DIFF
--- a/.changeset/preserve-provider-executed-tool-approvals.md
+++ b/.changeset/preserve-provider-executed-tool-approvals.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/workflow": patch
+---
+
+fix (workflow): preserve `tool-approval-request` and `tool-approval-response` parts for provider-executed tools so the provider receives the approval on resume
+
+`WorkflowAgent` previously stripped all approval parts from messages when handling tool-approval responses, regardless of whether the approved tool was locally or provider-executed. For provider-executed tools (e.g. MCP via the Responses API) this silently dropped the approval, leaving the provider to never receive it and the tool to never execute. The fix mirrors the inverse of #14289: locally-executed approvals are still executed and stripped here, while provider-executed approvals are left intact in the messages so the next provider call can process them.
--- a/packages/workflow/src/workflow-agent.test.ts
+++ b/packages/workflow/src/workflow-agent.test.ts
@@ -3037,5 +3037,103 @@ describe('WorkflowAgent', () => {
       // Should proceed normally without any approval processing
       expect(mockIterator.next).toHaveBeenCalled();
     });
+
+    it('should preserve provider-executed tool-approval-response across resume', async () => {
+      // Inverse of #14289: provider-executed approvals must reach the
+      // provider on resume rather than being silently dropped. The agent
+      // must NOT strip the `tool-approval-response` part whose
+      // `providerExecuted` flag is true, and must NOT synthesize a local
+      // tool-result for it. `convertToLanguageModelPrompt` then forwards
+      // the response to the provider on the next call.
+      const tools: ToolSet = {
+        web_search: {
+          type: 'provider' as const,
+          id: 'test.web_search',
+          args: {},
+          isProviderExecuted: true,
+          inputSchema: z.object({ query: z.string() }),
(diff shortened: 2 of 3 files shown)


gr2m:
Thanks for this fix @MyPrototypeWhat 🙏 The approval-resume path was since refactored to use core's `collectToolApprovals` (the local `collectToolApprovalsFromMessages` helper this PR edits no longer exists on `main`), so the diff no longer applies. I've opened #16256 which reworks the fix against the current code — it adopts your insight of sourcing the `providerExecuted` discriminator from the original tool-call part and credits you as a co-author.

PR #14933, opened 2026-05-02
fix(embed): default missing warnings to [] for v2 embedding models

### LINKED ISSUE #14926: embed() and embedMany() crash with EmbeddingModelV2 providers that omit `warnings`
## Description

`embed()` and `embedMany()` throw at runtime when used with a provider that implements `EmbeddingModelV2`, because the v2 `doEmbed` return type does not include a `warnings` field, and the SDK assumes it is always defined.

The SDK still accepts v2 models — `packages/ai/src/types/embedding-model.ts` declares:

```ts
export type EmbeddingModel =
  | EmbeddingModelV4
  | EmbeddingModelV3
  | EmbeddingModelV2<string>;
```

There is also a dedicated v2 → v3 → v4 compatibility adapter chain in `packages/ai/src/model/as-embedding-model-v3.ts` and `as-embedding-model-v4.ts`, with a runtime warning logged in compatibility mode. So v2 support is part of the public surface, but the embed code paths assume v3+ behavior.

## Reproduction

Using a `MockEmbeddingModelV2` whose `doEmbed` returns no `warnings` field (mirroring an older provider implementation):

```ts
import { embed, embedMany } from 'ai';
import { MockEmbeddingModelV2 } from 'ai/test';

const model = new MockEmbeddingModelV2<string>({
  doEmbed: async () => ({
    embeddings: [[0.1, 0.2, 0.3]],
    usage: { tokens: 1 },
  }),
});

await embed({ model, value: 'hello' });
// TypeError: Cannot read properties of undefined (reading 'length')
//   at logWarnings (packages/ai/src/logger/log-warnings.ts:106)

await embedMany({ model, values: ['a', 'b', 'c'] });
// chunked path:
// TypeError: result.warnings is not iterable
//   at embedMany (packages/ai/src/embed/embed-many.ts:330)
```

## Root cause

Three sites in `packages/ai` propagate `modelResponse.warnings` directly into the result without guarding against the v2 shape:

- `packages/ai/src/embed/embed.ts:196` — `warnings: modelResponse.warnings`
- `packages/ai/src/embed/embed-many.ts:218` — single-call path, same issue
- `packages/ai/src/embed/embed-many.ts:320` — chunked path, same issue (the spread on line 330 then throws on `undefined`)

Downstream, `logWarnings` at `packages/ai/src/logger/log-warnings.ts:106` reads `options.warnings.length`, and `embed-many.ts:330` does `warnings.push(...result.warnings)`. Both crash on `undefined`.

## Affected versions

- Latest `ai` SDK on `main`
- Any consumer pinning a v2-era provider package (e.g. older `@ai-sdk/google-vertex@3.0.x`) alongside a recent `ai` SDK

## Suggested fix

Coerce `modelResponse.warnings ?? []` at the three boundary sites in `embed.ts` / `embed-many.ts`. This honors the typed `Warning[]` return shape of the embed result and removes the runtime crash without changing observable behavior for v3/v4 providers.

PR open at #14927.

SAY-5:
Fixes #14926.

`EmbeddingModelV2` providers may omit `warnings` from `doEmbed`. Three sites in `packages/ai/src/embed` propagated `modelResponse.warnings` directly into the result; downstream `logWarnings` reads `.length` and `embed-many`'s chunked path spreads `...result.warnings`, both crash on undefined. Coerce with `?? []` at each boundary, and add a regression test using `MockEmbeddingModelV2` with no `warnings` field.

### DIFF
--- a/packages/ai/src/embed/embed-many.ts
+++ b/packages/ai/src/embed/embed-many.ts
@@ -215,7 +215,7 @@ export async function embedMany({
           return {
             embeddings,
             usage,
-            warnings: modelResponse.warnings,
+            warnings: modelResponse.warnings ?? [],
             providerMetadata: modelResponse.providerMetadata,
             response: modelResponse.response,
           };
@@ -317,7 +317,7 @@ export async function embedMany({
             return {
               embeddings: chunkEmbeddings,
               usage,
-              warnings: modelResponse.warnings,
+              warnings: modelResponse.warnings ?? [],
               providerMetadata: modelResponse.providerMetadata,
               response: modelResponse.response,
             };
--- a/packages/ai/src/embed/embed.test.ts
+++ b/packages/ai/src/embed/embed.test.ts
@@ -181,6 +181,26 @@ describe('result.warnings', () => {
 
     expect(result.warnings).toStrictEqual(expectedWarnings);
   });
+
+  it('defaults warnings to [] when the model omits them (v2 shape)', async () => {
+    // Regression for #14926: EmbeddingModelV2 implementations may omit
+    // `warnings`; embed() previously propagated `undefined` and crashed
+    // downstream in logWarnings.
+    const { MockEmbeddingModelV2 } = await import(
+      '../test/mock-embedding-model-v2'
+    );
+    const result = await embed({
+      model: new MockEmbeddingModelV2<string>({
+        doEmbed: async () => ({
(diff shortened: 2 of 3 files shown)


aayush-kapoor (commented):
marking closed in favour of https://github.com/vercel/ai/pull/15324

PR #14935, opened 2026-05-02
fix(xai): improve handling of streamed error events

### LINKED ISSUE #14932: @ai-sdk/xai: Responses stream error events surface as TypeValidationError
### Description

When the xAI Responses API streams an error event, `@ai-sdk/xai` appears to validate that event against the normal Responses event union instead of recognizing it as a provider/API error. The result is a very large `AI_TypeValidationError` rather than the actual xAI error message.

Observed raw event from xAI:

```json
{
  "sequence_number": 0,
  "type": "error",
  "code": null,
  "message": "Invalid arguments passed to the model.",
  "param": null
}
```

The SDK surfaced this as a schema validation failure because `type: "error"` did not match any expected events such as `response.created`, `response.output_text.delta`, `response.completed`, etc.

### Expected behavior

`@ai-sdk/xai` should recognize xAI Responses streamed error events and throw a concise provider/API error that preserves the original payload fields, especially `message`, `code`, and `param`.

### Actual behavior

The provider throws `AI_TypeValidationError` with a long Zod union error, obscuring the actual xAI error.

### Reproduction

This was reproduced with `grok-4.3` through `xai.responses(...)` when sending a tool schema that xAI rejects. In our case, xAI rejected a function tool schema whose enum values contained `/` characters, such as `fal-ai/flux-2-pro`.

Minimal shape:

```ts
import { streamText, tool } from "ai";
import { xai } from "@ai-sdk/xai";
import { z } from "zod";

const result = streamText({
  model: xai.responses("grok-4.3"),
  prompt: "Reply OK. Do not use tools.",
  tools: {
    image_generation: tool({
      description: "Generate image",
      inputSchema: z.object({
        prompt: z.string(),
        model: z.enum(["fal-ai/flux-2-pro", "fal-ai/qwen-image-edit"]).optional(),
      }),
      execute: async () => ({ ok: true }),
    }),
  },
  maxOutputTokens: 8,
});

  console.log(part);
}
```

A similar tool schema without slash-containing enum values succeeds, and changing the `model` field from `z.enum([...])` to `z.string()` avoids the xAI rejection.

### Package versions

- `@ai-sdk/xai`: `3.0.56`
- `ai`: `6.0.174`
- Runtime: Bun `1.3.5`

### Notes

There may be a separate xAI API/docs issue around enum values containing `/`, but regardless, the provider should surface the streamed `type: "error"` event as the underlying provider error instead of a validation error for the normal event union.

MuzzaiyyanHussain:
This PR improves error handling for streamed responses in the xAI provider.

Currently, if a streamed chunk fails schema validation, the SDK surfaces a generic parsing error (AI_TypeValidationError), which can obscure the original provider error.

Changes:
- Adds a fallback to detect and surface raw provider error events (`type: "error"`) when chunk parsing fails
- Relaxes the error schema slightly to better tolerate variations in provider responses

This ensures that meaningful provider errors are preserved and improves debugging for consumers of the SDK.

Fixes #14932


MuzzaiyyanHussain:
@gr2m sorry for tagging but take a look pls

FujiwaraChoki:
Thanks for taking a look at this @MuzzaiyyanHussain

lgrammel:
Thank you for the pull request. Issue #14932 was closed as not planned. Closing this pull request.

PR #14937, opened 2026-05-02
fix(google): read serviceTier from x-gemini-service-tier response header

shaper:
## Background

The Google Gemini API returns the applied service tier in the `x-gemini-service-tier` response header ([priority docs](https://ai.google.dev/gemini-api/docs/priority-inference), [flex docs](https://ai.google.dev/gemini-api/docs/flex-inference)), not in the response body. The previous implementation read `serviceTier` from the body, which Google never populates, so `providerMetadata.google.serviceTier` was always `null`.

## Summary

- Read `serviceTier` from `responseHeaders['x-gemini-service-tier']` for Gemini API
- Remove `serviceTier` from response body schema (`responseSchema`) and chunk schema (`chunkSchema`) since Google never populates it
- Use [PayGo approach](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/priority-paygo) for Vertex service tier
- Remove chunk-based `serviceTier` accumulation in the streaming path
- Update examples to use `priority` tier consistently and standard vertex imports

## Current API behavior

Google only returns the header for non-streaming Google AI requests today. There is no known place (neither headers nor in response body or stream chunks) where a Google stream response includes the value. Vertex returns a field `usageMetadata.trafficType` based on the config.

## Test plan

- [x] `pnpm test` in `packages/google` — 399 tests pass (node + edge)
- [x] Run service tier examples against live API

🤖 Partly generated with [Claude Code](https://claude.com/claude-code)


felixarntz (dismissed):
@shaper I've tried in several ways to get the service tier to come back from the API when streaming, but not successful. Even Google's own `js-genai` SDK doesn't handle it. It seems like a genuine gap in the API, so we probably should just accept it as not being returned.

For Vertex, I'm unsure, there's conflicting docs about the `SERVICE_TIER_*` values, but also https://docs.cloud.google.com/vertex-ai/generative-ai/docs/priority-paygo, which seems slightly different.

I think this is better than what we have now, so approving. There may be more work to do separately regarding Vertex, depending what we decide to go with. But regarding their `js-genai` SDK, it doesn't handle the service tier in the response either.

felixarntz (approved):

PR #14943, opened 2026-05-04
fix(ai): tag step/chunk timeout aborts with TimeoutError reason

zirkelc:
## Summary

When `timeout: { stepMs }` or `timeout: { chunkMs }` fires, the SDK calls `stepAbortController.abort()` / `chunkAbortController.abort()` with no argument. The resulting `signal.reason` is a generic `AbortError` DOMException indistinguishable from a user-initiated `controller.abort()`.

This PR passes a `TimeoutError` DOMException as the abort reason, mirroring what `AbortSignal.timeout()` produces natively (and what `totalMs` already does, since it goes through `mergeAbortSignals` → `AbortSignal.timeout(n)`).

## Motivation

Consumers currently can't tell apart "user pressed cancel" from "framework step/chunk timeout fired". Both surface as `AbortError`. Tagging the reason makes them distinguishable via `signal.reason.name === 'TimeoutError'`.

This also brings step/chunk timeouts into symmetry with `totalMs` and user-supplied `AbortSignal.timeout()`, which already produce `TimeoutError`.

### DIFF
--- a/packages/ai/src/generate-text/generate-text.ts
+++ b/packages/ai/src/generate-text/generate-text.ts
@@ -53,6 +53,7 @@ import { mergeAbortSignals } from '../util/merge-abort-signals';
 import { mergeObjects } from '../util/merge-objects';
 import { notify } from '../util/notify';
 import { prepareRetries } from '../util/prepare-retries';
+import { setAbortTimeout } from '../util/set-abort-timeout';
 import { VERSION } from '../version';
 import type { ActiveTools } from './active-tools';
 import { collectToolApprovals } from './collect-tool-approvals';
@@ -578,16 +579,7 @@ export async function generateText<
       // Set up step timeout if configured
       const stepTimeoutId =
         stepTimeoutMs != null
-          ? setTimeout(
-              () =>
-                stepAbortController!.abort(
-                  new DOMException(
-                    `Step timeout of ${stepTimeoutMs}ms exceeded`,
-                    'TimeoutError',
-                  ),
-                ),
-              stepTimeoutMs,
-            )
+          ? setAbortTimeout(stepAbortController!, 'Step', stepTimeoutMs)
           : undefined;
 
       try {
--- a/packages/ai/src/generate-text/generate-text.ts
+++ b/packages/ai/src/generate-text/generate-text.ts
@@ -577,14 +577,11 @@ export async function generateText<
 
     do {
       // Set up step timeout if configured
-      const stepTimeoutId =
-        stepTimeoutMs != null
-          ? setAbortTimeout({
-              controller: stepAbortController!,
-              label: 'Step',
-              ms: stepTimeoutMs,
-            })
-          : undefined;
+      const stepTimeoutId = setAbortTimeout({
+        abortController: stepAbortController,
+        label: 'Step',
+        timeoutMs: stepTimeoutMs,
+      });
 
       try {
         const stepInputMessages = [...initialMessages, ...responseMessages];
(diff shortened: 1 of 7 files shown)


lgrammel:
@zirkelc can you reopen or force push with signed commits? we changed our security policy and cannot merge prs with unsigned commits https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -578,7 +578,16 @@ export async function generateText<
...
+              () =>
+                stepAbortController!.abort(
+                  new DOMException(
+                    `Step timeout of ${stepTimeoutMs}ms exceeded`,
+                    'TimeoutError',
+                  ),
+                ),
+              stepTimeoutMs,
+            )
           : undefined;
```
lgrammel: this whole block could potentially be extracted into a helper function (also for the other 2 occurences)

lgrammel:
please add a patch changeset

zirkelc:
Thanks for the quick reply @lgrammel 

I will force push with signed commits a little later today!

zirkelc:
@lgrammel I'm done, please let me know if there is anything else to do!

packages/ai/src/util/set-abort-timeout.ts
```diff
@@ -0,0 +1,30 @@
...
+ * (e.g. "Step", "Chunk").
+ * @param ms - Duration in milliseconds before the controller is aborted.
+ * @returns The timeout id, suitable for passing to `clearTimeout`.
+ */
+export function setAbortTimeout({
+  controller,
+  label,
+  ms,
+}: {
+  controller: AbortController;
```
lgrammel: can controller be optional (undefined return value if undefined). also pls rename to `abortController`

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1462,7 +1463,11 @@ class DefaultStreamTextResult<
         // Set up step timeout if configured
         const stepTimeoutId =
           stepTimeoutMs != null
-            ? setTimeout(() => stepAbortController!.abort(), stepTimeoutMs)
+            ? setAbortTimeout({
+                controller: stepAbortController!,
+                label: 'Step',
+                ms: stepTimeoutMs,
+              })
             : undefined;
```
lgrammel: can be simplified by allowing undefined ms in setAbortTimeout

packages/ai/src/util/set-abort-timeout.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @param controller - The controller to abort when the timeout elapses.
+ * @param label - Human-readable label included in the error message
+ * (e.g. "Step", "Chunk").
+ * @param ms - Duration in milliseconds before the controller is aborted.
+ * @returns The timeout id, suitable for passing to `clearTimeout`.
+ */
+export function setAbortTimeout({
+  controller,
+  label,
+  ms,
```
lgrammel: rename to timeoutMs

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -578,7 +579,11 @@ export async function generateText<
       // Set up step timeout if configured
       const stepTimeoutId =
```
lgrammel: can be simplified if undefined is allowed for timeoutMs arg (in which case undefined should be returned)

lgrammel:
@zirkelc had a few more suggestions on how to simplify

zirkelc:
@lgrammel please check again!

lgrammel (approved):

PR #14968, opened 2026-05-04
fix(google): emit no-args streaming tool calls and preserve thoughtSignature

### LINKED ISSUE #14847: `@ai-sdk/google` streaming parser drops no-args function calls, losing `thoughtSignature`
When `streamFunctionCallArguments: true` is set and the model invokes a tool whose `inputSchema` has no required fields, the streaming parser drops the whole function-call chunk. Its `thoughtSignature` goes with it. For Gemini 3 thinking models this often breaks multi-turn conversations: Vertex emits one signature per response, on the first function call, and if that first call is a no-args call, the response message ends up with zero signatures. The next agent step then 400s.

This bug reproduces on **both** the `@ai-sdk/google` parser (direct Vertex path) and the AI Gateway server (`gateway('google/...')` path). They appear to be two copies of the same bug. Fixing only `@ai-sdk/google` will not fix the gateway path; the gateway server needs the same fix.

### Repro

```ts
import { streamText, tool } from 'ai';
import { z } from 'zod';

const vertex = createVertex({
  apiKey: process.env.GOOGLE_VERTEX_API_KEY!,
  project: '_',
  location: 'global',
});

const result = streamText({
  model: vertex('gemini-3-flash-preview'),
  tools: {
    read_theme: tool({
      description: 'Read the current theme.',
      inputSchema: z.object({}),
      execute: async () => ({ '--background': '#fff' }),
    }),
    read_screen: tool({
      description: 'Read a screen.',
      inputSchema: z.object({ id: z.string() }),
      execute: async ({ id }) => ({ id, html: '<div/>' }),
    }),
  },
  providerOptions: {
    vertex: {
      streamFunctionCallArguments: true,
      thinkingConfig: { includeThoughts: true, thinkingLevel: 'medium' },
    },
  },
  prompt: 'Read the theme and screens A, B, C in parallel, then summarize.',
});

    console.log(part.toolName, part.providerMetadata);
  }
}
```

Expected: four `tool-call` events; the first one (`read_theme`) carries `providerMetadata.vertex.thoughtSignature`.

Actual: three `tool-call` events (`read_screen` only). No `read_theme`. No signature anywhere.

### Wire-level evidence

The Vertex SSE response for this scenario contains:

```
chunk 1: { functionCall: { name: 'read_theme' }, thoughtSignature: '<2KB>' }
chunk 2: { functionCall: { name: 'read_screen', willContinue: true } }
chunk 3..5: { functionCall: { partialArgs: [...], willContinue: true } }
chunk 6: { functionCall: {} }
... repeats for screens B, C ...
```

Vertex does send the signature, on the first call.

### Where it gets dropped (direct Vertex)

In [`packages/google/src/google-language-model.ts`](https://github.com/vercel/ai/blob/main/packages/google/src/google-language-model.ts), the streaming branch around [L823-L835](https://github.com/vercel/ai/blob/main/packages/google/src/google-language-model.ts#L823-L835) classifies each function-call chunk into one of three buckets:

```ts
const isStreamingChunk =
  part.functionCall.partialArgs != null ||
  (part.functionCall.name != null &&
    part.functionCall.willContinue === true);
const isTerminalChunk =
  part.functionCall.name == null &&
  part.functionCall.args == null &&
  part.functionCall.partialArgs == null &&
  part.functionCall.willContinue == null;
const isCompleteCall =
  part.functionCall.name != null &&
  part.functionCall.args != null &&
  part.functionCall.partialArgs == null;
```

A no-args complete call arrives as `{ name: 'X' }`, with no `args`, no `willContinue`, no `partialArgs`. None of the three classifiers match. The chunk falls through and is never emitted.

The non-streaming path at [L371-L374](https://github.com/vercel/ai/blob/main/packages/google/src/google-language-model.ts#L371-L374) already filters with `part.functionCall.name != null && part.functionCall.args != null`, which also excludes no-args calls, but the streaming path is the one that matters when `streamFunctionCallArguments: true`.

### AI Gateway has the same bug

Running the same scenario through `gateway('google/gemini-3-flash-preview')` with `providerOptions.gateway: { only: ['vertex'] }` reproduces the same failure. We captured the raw v3 SSE stream the gateway returns to the client, and in failure runs there is no `tool-input-start`, `tool-input-end`, or `tool-call` event for `read_theme` anywhere. The whole call is missing from the v3 stream, along with any `vertex.thoughtSignature` it would have carried.

Direct Vertex always emits exactly one signature on the wire (on `read_theme` for that scenario). The gateway sometimes returns a v3 stream with that signature visible (when the model's first call has args) and sometimes returns a v3 stream with zero signatures (when the first call is the no-args `read_theme`). The split correlates exactly with the model's tool choice, which matches the parser-side bug pattern: when Vertex sends `{ name: 'X' }` with no `args`, both parsers drop it.

So this is two copies of the same bug, one in `@ai-sdk/google` (client-side) and one in the gateway server, and they need to be fixed together. A client-side `@ai-sdk/google` fix alone leaves the gateway path broken because the v3 stream is already missing the call by the time the client sees it.

### Additional gateway sig-dropping behavior (separate from the no-args bug)

The no-args bug above is the primary, sharply-isolated problem. The gateway has at least one additional sig-dropping behavior we cannot isolate as cleanly, but observed both in our test captures and in production traffic.

**Production observation**: an assistant turn produced through `gateway('google/gemini-3-flash-preview')` containing a single `read_screen` call (a streamed-args tool, not no-args) was saved with `callProviderMetadata: { gateway: { generationId: '...' } }` and no `vertex` namespace at all. The next agent step 400'd with the same `missing thought_signature` error. The model never picked the no-args tool, so the workaround above did not apply.

**Test capture**: in one of our gateway captures (15 attempts), one failure had a v3 SSE stream with zero `vertex.thoughtSignature` across all four streamed parallel `read_screen` calls, none were no-args.

We don't have the underlying Vertex wire for those requests, so we can't say definitively whether Vertex didn't emit a sig that round or whether the gateway dropped it on the way out. Direct Vertex's behavior across our captures has been to emit exactly one sig per response (on the first function call, always). For the gateway we sometimes see zero sigs in the v3 stream, which is consistent with "gateway dropped them" rather than "Vertex didn't emit them".

Either way, fixing the no-args bug on `@ai-sdk/google` and the gateway server side is necessary but not sufficient for the gateway path. There appears to be at least one more source of sig drops in the gateway → v3 → client pipeline that we can't see from outside. A gateway-team look at sig propagation through that pipeline would be useful.

### Why this matters

For Gemini 3 with thinking enabled, Vertex enforces a rule on multi-turn conversations: an assistant turn that contains any function calls must contain at least one with a `thoughtSignature`, or it returns a 400 with `Unable to submit request because function call X in the N. content block is missing a 'thought_signature'`. Vertex emits one signature per response, on the first function call. If the parser drops that first call, the response message has zero signatures and the agent's next step trips the rule.

We see this on roughly half of all multi-turn chat continuations against an existing Gemini 3 project, depending on whether the model picks the no-args tool first.

### Workaround

Give every tool at least one **required** argument:

```ts
read_theme: tool({
  inputSchema: z.object({
    reason: z.string().describe('Why the theme is being read'),
  }),
  execute: async () => ({ ... }),
}),
```

`.optional()` is not enough. We initially shipped an optional variant and hit the same 400 in production: when the model decides to omit the optional arg (model choice, sensitive to the system prompt and history), Vertex emits the call as `{ "name": "X" }` with no args, no willContinue, no partialArgs. That hits the same drop path documented above.

A required arg forces the model to populate args, which puts Vertex on the streaming path (`{ "name": "X", "willContinue": true }` followed by `partialArgs` chunks), which the SDK handles. Confirmed with a wire-level test and 10/10 success on the previously-failing scenario, both on `createVertex` direct and `gateway('google/...')` paths.

### Proposed fix

After the existing `isCompleteCall` branch in [google-language-model.ts](https://github.com/vercel/ai/blob/main/packages/google/src/google-language-model.ts#L923), add a fourth branch for the no-args case:

```ts
} else if (
  part.functionCall.name != null &&
  part.functionCall.args == null &&
  part.functionCall.willContinue !== true &&
  part.functionCall.partialArgs == null
) {
  // Single-chunk no-args call: emit as a complete call with empty args.
  const toolCallId = generateId();
  controller.enqueue({
    type: 'tool-input-start',
    id: toolCallId,
    toolName: part.functionCall.name,
    providerMetadata: providerMeta,
  });
  controller.enqueue({
    type: 'tool-input-end',
    id: toolCallId,
    providerMetadata: providerMeta,
  });
  controller.enqueue({
    type: 'tool-call',
    toolCallId,
    toolName: part.functionCall.name,
    input: '{}',
    providerMetadata: providerMeta,
  });
  hasToolCalls = true;
}
```

The same fix needs to land in whatever code the AI Gateway server uses to parse Vertex's SSE into v3 events. Happy to send a PR for the `@ai-sdk/google` side if useful; gateway-server side I can't reach.

### Environment

- `ai`: 6.0.168
- `@ai-sdk/google`: 3.0.65 (also reproduces on the `4.0.0-beta.47` source on `main`)
- `@ai-sdk/google-vertex`: 4.0.113
- `@ai-sdk/gateway`: 3.0.104
- Model: `google/gemini-3-flash-preview`

### Related

- #13908 carries forward the first signature to sibling parallel calls. That fix doesn't help when the first call itself is dropped.
- Supersedes #14840 (wrong original framing).

gr2m:
## Background

Vertex emits a no-args function call as a single chunk shaped `{ functionCall: { name: 'X' } }` with no `args`, no `partialArgs`, and no `willContinue`. The streaming parser in `google-language-model.ts` classifies chunks into `isStreamingChunk` / `isTerminalChunk` / `isCompleteCall`, and a no-args call matches none of them — so it falls through and is silently dropped. The unary (`doGenerate`) path has the same drop because of an `args != null` guard.

For Gemini 3 thinking models that one dropped chunk also carries the response's `thoughtSignature`. Vertex enforces a multi-turn rule: an assistant turn containing any function calls must contain at least one with a `thoughtSignature`. When the dropped call is the first call (the one with the sig), the next agent step 400s with `missing thought_signature`.

## Summary

- Streaming path: add a fourth classifier `isNoArgsCompleteCall` that matches `{ name: X }` with no `args` / `partialArgs` / `willContinue !== true`, and emit `tool-input-start` → `tool-input-end` → `tool-call` with `'{}'` input, propagating `providerMetadata` (so `thoughtSignature` survives).
- Unary path: stop requiring `args != null`. Use `JSON.stringify(args ?? {})` so a no-args complete call from `doGenerate` is also emitted instead of dropped.

## Manual Verification

<details>

```ts
import { googleVertex } from '@ai-sdk/google-vertex';
import { streamText, tool } from 'ai';
import { z } from 'zod';
import { saveRawChunks } from '../../lib/save-raw-chunks';

// Reproduces the wire shape from issue #14847: Vertex emits a no-args
// function call as a single chunk `{ functionCall: { name: 'X' } }`. With
// Gemini 3 thinking enabled, that chunk also carries the response's
// thoughtSignature, which the streaming parser must preserve.
//
// The prompt pushes the model to pick the no-args `read_theme` tool first.
// Rerun if a recording session ends up with `read_screen` first (the bug
// only manifests on the first call's chunk shape).

function getThoughtSignature(
  providerMetadata: Record<string, Record<string, unknown>> | undefined,
): string | undefined {
  const sig =
    providerMetadata?.vertex?.thoughtSignature ??
    providerMetadata?.googleVertex?.thoughtSignature ??
    providerMetadata?.google?.thoughtSignature;
  return typeof sig === 'string' ? sig : undefined;
}

run(async () => {
  const result = streamText({
    model: googleVertex('gemini-3-flash-preview'),
    tools: {
      read_theme: tool({
        description: 'Read the current UI theme.',
        inputSchema: z.object({}),
        execute: async () => ({ '--background': '#fff' }),
      }),
      read_screen: tool({
        description: 'Read the contents of a screen by id.',
        inputSchema: z.object({ id: z.string() }),
        execute: async ({ id }) => ({ id, html: '<div/>' }),
      }),
    },
    providerOptions: {
      googleVertex: {
        streamFunctionCallArguments: true,
        thinkingConfig: { includeThoughts: true, thinkingLevel: 'medium' },
      },
    },
    prompt:
      'Read the theme, then read screens A, B, and C in parallel. Summarize.',
    includeRawChunks: true,
  });

  const toolCalls: Array<{ toolName: string; signature: string | undefined }> =
    [];

    switch (part.type) {
      case 'tool-input-start':
        console.log(`\n[tool-input-start] ${part.toolName} (${part.id})`);
        if (part.providerMetadata) {
          console.log('  providerMetadata:', part.providerMetadata);
        }
        break;
      case 'tool-input-delta':
        process.stdout.write(part.delta);
        break;
      case 'tool-input-end':
        console.log(`\n[tool-input-end] (${part.id})`);
        break;
      case 'tool-call':
        console.log(`\n[tool-call] ${part.toolName}:`, part.input);
        if (part.providerMetadata) {
          console.log('  providerMetadata:', part.providerMetadata);
        }
        toolCalls.push({
          toolName: part.toolName,
          signature: getThoughtSignature(part.providerMetadata),
        });
        break;
      case 'text-delta':
        process.stdout.write(part.text);
        break;
      case 'finish':
        console.log('\nFinish reason:', part.finishReason);
        console.log('Usage:', part.totalUsage);
        break;
      case 'error':
        console.error('Error:', part.error);
        break;
    }
  }

  // Output written to examples/ai-functions/output/<filename>.chunks.txt.
  // To regenerate the test fixture, copy it to:
  //   packages/google/src/__fixtures__/google-stream-no-args-tool-call.chunks.txt
  await saveRawChunks({
    result,
    filename: 'google-vertex-stream-tool-call-no-args',
  });

  // Verify the fix is in effect. When the bug is present, Vertex's no-args
  // chunk is dropped, taking `read_theme` and the response's only
  // thoughtSignature with it.
  const hasReadTheme = toolCalls.some(c => c.toolName === 'read_theme');
  const hasSignature = toolCalls.some(c => c.signature != null);

  if (!hasReadTheme || !hasSignature) {
    const reasons: string[] = [];
    if (!hasReadTheme) {
      reasons.push(
        '`read_theme` was not emitted (the prompt asked for it; the no-args ' +
          'chunk shape `{ functionCall: { name: "read_theme" } }` was dropped)',
      );
    }
    if (!hasSignature) {
      reasons.push(
        'no tool-call carried a `thoughtSignature` (Vertex emits exactly one ' +
          'per response, on the first function call; if that call was dropped, ' +
          'the signature went with it)',
      );
    }
    throw new Error(
      `#14847 reproduced: ${reasons.join('; and ')}.\n\n` +
        'Vertex emits a no-args function call as a single chunk shaped\n' +
        '  { functionCall: { name: "X" } }\n' +
        'with no `args`, no `partialArgs`, and no `willContinue`. The\n' +
        '@ai-sdk/google streaming parser had no branch for this shape, so the\n' +
        'chunk was silently dropped. On Gemini 3 thinking models the response\n' +
        'carries its only `thoughtSignature` on the first function call, so\n' +
        'dropping it caused the next multi-turn step to 400 with\n' +
        '`missing thought_signature`. See https://github.com/vercel/ai/issues/14847.\n\n' +
        `Tool calls observed: ${
          toolCalls.length === 0
            ? '(none)'
            : toolCalls.map(c => c.toolName).join(', ')
        }.`,
    );
  }

  console.log(
    `\nOK: emitted ${toolCalls.length} tool calls, ${
      toolCalls.filter(c => c.signature).length
    } with thoughtSignature.`,
  );
});
```

</details>

used to capture fixtures for the test

## Checklist

## Future Work

Fix AI Gateway

## Related Issues

Fixes #14847.

Closes #14879 
Closes #14934


examples/ai-functions/src/stream-text/google/vertex-stream-tool-call-no-args.ts
```diff
@@ -0,0 +1,82 @@
...
+import { saveRawChunks } from '../../lib/save-raw-chunks';
+
+// Reproduces the wire shape from issue #14847: Vertex emits a no-args
+// function call as a single chunk `{ functionCall: { name: 'X' } }`. With
+// Gemini 3 thinking enabled, that chunk also carries the response's
+// thoughtSignature, which the streaming parser must preserve.
+//
+// The system prompt and parallel calls push the model toward picking the
+// no-args tool first; rerun if the recorded session does not include a
+// `{ functionCall: { name: 'read_theme' } }` chunk with thoughtSignature.
```
gr2m: will update the example to catch this code instead of documenting as code comment

aayush-kapoor (approved):

stefanofa:
@gr2m love it, thanks! any way i can get updated when the fix will land also on ai gateway?

PR #14975, opened 2026-05-04
feat(otel): add option for custom span attributes

### LINKED ISSUE #14727: Add span attribute enrichment hook to `@ai-sdk/otel`
### Description

**Feature request**
Add a small attribute-enrichment hook to `@ai-sdk/otel`’s `OpenTelemetry` integration so downstream integrations can add provider-specific OpenTelemetry attributes while AI SDK still owns span creation.

Something like:

```ts
import { OpenTelemetry } from "@ai-sdk/otel";

registerTelemetry(
  new OpenTelemetry({
    enrichSpanAttributes: ({ spanType, operationId, callId, runtimeContext }) => {
      return {
        ...getCustomAttributes(runtimeContext, spanType),
      };
    },
  }),
);
```

Or more concretely:

```ts
type EnrichSpanAttributesContext = {
  callId: string;
  operationId: string;
  spanName: string;
  spanType: "operation" | "model" | "tool" | "embedding" | "rerank";
  runtimeContext?: Record<string, unknown>;
  functionId?: string;
  existingAttributes: Readonly<Attributes>;
};

type EnrichSpanAttributes = (
  context: EnrichSpanAttributesContext,
) => Attributes | undefined;
```

The hook would run immediately before `@ai-sdk/otel` calls `tracer.startSpan(...)`.

```ts
const attributes = selectAttributes(telemetry, {
  ...aiSdkAttributes,
});

const extraAttributes = options.enrichSpanAttributes?.({
  callId: event.callId,
  operationId,
  spanName,
  spanType,
  runtimeContext,
  functionId: telemetry?.functionId,
  existingAttributes: attributes,
});

const span = tracer.startSpan(spanName, {
  attributes: {
    ...attributes,
    ...extraAttributes,
  },
}, parentContext);
```

**Problem**
AI SDK v7 has a nice telemetry architecture: the AI SDK emits semantic lifecycle callbacks, and `@ai-sdk/otel` turns those into OpenTelemetry spans. That works well when the only desired attributes are the AI SDK’s built-in attributes.

The gap is that external observability systems sometimes need to add additional span attributes that are specific to their ingestion pipeline.

For Langfuse, examples are:

```ts
"langfuse.observation.metadata.feature": "chat"
"langfuse.observation.prompt.name": "support-agent"
"langfuse.observation.prompt.version": 3
```

These are not AI SDK attributes and should not become AI SDK-owned semantics. But they need to be present on the OTel span when it is exported.

The user already has a natural place to provide this data: `runtimeContext`.

```ts
await generateText({
  model,
  prompt,
  runtimeContext: {
    langfuse: {
      metadata: { feature: "chat" },
      prompt: { name: "support-agent", version: 3 },
    },
  },
  telemetry: {
    functionId: "chat-route",
  },
});
```

The issue is that `@ai-sdk/otel` currently fully owns span creation and does not expose a supported way to add attributes at span creation time.

**Current Workaround**
Because there is no enrichment hook, Langfuse has to wrap the OpenTelemetry tracer.

Today the workaround looks conceptually like this:

```txt
@ai-sdk/otel creates span
  -> calls tracer.startSpan(name, options, parentContext)
  -> Langfuse tracer wrapper intercepts startSpan
  -> adds Langfuse attributes
  -> forwards to real tracer
```

This works, but it is awkward because the OpenTelemetry `Tracer` API only receives:

```ts
name
options
parentContext
```

It does not receive AI SDK concepts like:

```ts
callId
operationId
runtimeContext
span type
tool call id
```

So Langfuse has to reconstruct enough AI SDK context from the outside. That is why we maintain bookkeeping maps like:

```txt
callId -> Langfuse runtime context
spanId -> callId
```

This is exactly the kind of cleverness we would rather avoid. It is not business logic; it is a workaround for not having an official enrichment point.

**Why AI SDK Is The Right Place**
`@ai-sdk/otel` already has the information at the perfect moment.

When it creates the root span, it knows:

```ts
event.callId
event.operationId
event.runtimeContext
telemetry.functionId
```

When it creates the model-call span, it knows:

```ts
callId
operationId
step context
model call attributes
parent span context
```

When it creates tool spans, it knows:

```ts
callId
toolCallId
toolName
tool input
tool span parent context
```

So the clean design is:

```txt
AI SDK keeps owning the span shape.
AI SDK keeps deciding when spans are created.
AI SDK keeps deciding default attributes.
Integrations can append additional attributes.
```

That is much safer than asking every integration to wrap tracers and infer span ownership.

**Why This Does Not Need To Expose Unstable Internals**
The hook does not need to expose private `callState`, raw spans, internal maps, or mutable lifecycle state.

A stable minimal context would be enough:

```ts
{
  callId,
  operationId,
  spanName,
  spanType,
  runtimeContext,
  functionId,
  existingAttributes,
}
```

Most of these are already public concepts:

```txt
callId: already present in telemetry events
operationId: already present in telemetry events / attributes
runtimeContext: user-provided public input
functionId: public telemetry option
existingAttributes: what AI SDK is already about to emit
```

`spanType` would be the only new abstraction, and it can be intentionally coarse:

```ts
"operation" | "model" | "tool" | "embedding" | "rerank"
```

That avoids forcing downstream integrations to parse span names like `.doGenerate` or `.doStream`.

If they want to be even more conservative, they could omit `spanType` initially and expose only `spanName` / `operationId`, but `spanType` would make the hook much more stable over time.

**Why This Benefits AI SDK Users Generally**
This is not Langfuse-specific.

Any observability integration may need to add attributes such as:

```ts
"tenant.id"
"app.route"
"experiment.variant"
"customer.plan"
"deployment.region"
"custom.trace.link"
```

Users often already have those values in `runtimeContext`. The current API lets them pass runtime context to the AI SDK, but the official OTel integration does not let that context influence emitted span attributes except through AI SDK’s own built-in mapping.

A hook would let users do:

```ts
new OpenTelemetry({
  enrichSpanAttributes: ({ runtimeContext }) => ({
    "tenant.id": runtimeContext?.tenantId,
    "experiment.variant": runtimeContext?.experiment,
  }),
});
```

That keeps custom application semantics out of AI SDK core while making the official OTel integration more extensible.

### AI SDK Version

v7 beta

### Code of Conduct

aayush-kapoor:
## Background

#14727

`@ai-sdk/otel` currently does not expose a supported way to add attributes at span creation time. this is for external observability systems that sometimes need to add additional span attributes that are specific to their ingestion pipeline

## Summary

Added a new `enrichSpanAttributes` option to OpenTelemetry so callers can append custom attributes when spans are created

## Manual Verification

ran the example `examples/ai-functions/src/telemetry/otel/generate-text-custom-attributes.ts`

## Checklist

## Related Issues

fixes #14727


examples/ai-functions/src/telemetry/otel/generate-text-custom-attributes.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const sdk = new NodeSDK({
+  traceExporter: new ConsoleSpanExporter(),
+  instrumentations: [getNodeAutoInstrumentations()],
+});
+
+sdk.start();
+registerTelemetry(
+  new OpenTelemetry({
+    enrichSpanAttributes: ({ spanType, operationId, runtimeContext }) => ({
```
lgrammel: I am working on restricting how the runtimeContext is exposed to telemetry for security so this might conflict somewhat

examples/ai-functions/src/telemetry/otel/generate-text-custom-attributes.ts
```diff
@@ -0,0 +1,55 @@
...
+      'example.span.type': spanType,
+      'example.operation.id': operationId,
+      'example.context.something': runtimeContext?.something as
+        | string
+        | undefined,
+      'example.context.some_other_thing': runtimeContext?.someOtherThing as
+        | string
+        | undefined,
```
lgrammel, line 28: why is this cast needed?

packages/otel/src/supplemental-attributes.ts
```diff
@@ -17,12 +17,33 @@ export type SupplementalAttributeOptions = Record<
...
 
+export type OpenTelemetrySpanType =
+  | 'operation'
+  | 'step'
+  | 'languageModel'
+  | 'tool'
+  | 'embedding'
+  | 'reranking';
+
+export type EnrichSpanAttributes = (options: {
```
lgrammel: rename to `EnrichSpanAttributesFunction`

packages/otel/src/supplemental-attributes.ts
```diff
@@ -17,12 +17,33 @@ export type SupplementalAttributeOptions = Record<
   boolean
 >;
 
+export type OpenTelemetrySpanType =
+  | 'operation'
+  | 'step'
+  | 'languageModel'
+  | 'tool'
+  | 'embedding'
+  | 'reranking';
```
lgrammel, line 26: these mix language model and other span -- i am wondering if there are better names, e.g. hierarchical

packages/otel/src/supplemental-attributes.ts
```diff
@@ -17,12 +17,33 @@ export type SupplementalAttributeOptions = Record<
...
   /**
    * The tracer to use for the telemetry data.
    */
   tracer?: Tracer;
 
+  /**
+   * Adds custom attributes to spans when they are created. These attributes are
+   * not AI SDK-owned semantics and are intended for observability integrations.
+   */
+  enrichSpanAttributes?: EnrichSpanAttributes;
```
lgrammel: can this be shortened to `enrichSpan`? type name also

examples/ai-functions/src/telemetry/otel/generate-text-custom-attributes.ts
aayush-kapoor: opentelemetry `Attributes` can only receive primitive values and currently unknown values are being passed. hence the cast

packages/otel/src/supplemental-attributes.ts
aayush-kapoor: actaully named it to just `EnrichSpan`

packages/otel/src/supplemental-attributes.ts
aayush-kapoor: that is a span type but

lgrammel (approved):

PR #14982, opened 2026-05-05
docs(tools): fix double tool execution in re-ask repair example

### LINKED ISSUE #5801: Error in docs: Example: Use the re-ask strategy for repair
### Description

https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling#example-use-the-re-ask-strategy-for-repair

This section of the docs is wrong. When running this code, the tool is executed twice: once by the `experimental_repairToolCall`'s `generateText` and the a second time by the parent `generateText`.

Mmartinrusso:
## Problem

Fixes #5801.

The re-ask strategy example in the tool call repair docs passes the full `tools` object (including `execute` functions) to the inner `generateText` call. When the model returns a corrected tool call, the inner `generateText` executes it — and then the outer `generateText` executes it again, causing **double tool execution**.

## Fix

Strip `execute` functions before passing `tools` to the inner `generateText`. The inner call only needs the tool schemas so the model can produce a corrected tool call; it should not run the tool itself.

```ts
const toolsForRepair = Object.fromEntries(
  Object.entries(tools).map(([name, { execute: _execute, ...rest }]) => [
    name,
    rest,
  ]),
);
```

## Testing

Run the re-ask strategy example from the docs before and after this change. Before: the tool's `execute` function is called twice. After: it is called exactly once by the outer `generateText`.

### DIFF
--- a/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
+++ b/content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx
@@ -1111,6 +1111,15 @@ const result = await generateText({
     messages,
     system,
   }) => {
+    // Strip execute functions so the inner generateText only asks the model
+    // for corrected arguments without running the tools a second time.
+    const toolsForRepair = Object.fromEntries(
+      Object.entries(tools).map(([name, { execute: _execute, ...rest }]) => [
+        name,
+        rest,
+      ]),
+    );
+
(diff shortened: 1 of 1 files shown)


Mmartinrusso:
Closing in favor of a source-level fix that addresses the root cause rather than pushing the burden onto users. Will open a new PR shortly.

PR #14983, opened 2026-05-05
fix(@ai-sdk/openai): omit reasoning item id when store is false

### LINKED ISSUE #12687: @ai-sdk/openai: Reasoning items include id field when store: false, causing Azure rejection

Mmartinrusso:
## Background

When \`store: false\`, the SDK includes \`id\` in reasoning items sent to the API. Azure OpenAI rejects these requests with an error because the \`id\` field is only meaningful in store mode — in stateless mode it is an invalid field.

Fixes #12687.

## Summary

Remove \`id\` from the reasoning object in the \`store: false\` branch of \`convert-to-openai-responses-input.ts\`. The field is already typed as optional (\`id?: string\`), so omitting it is type-safe and has no effect on direct OpenAI.

## Manual Verification

Inspected serialized request body in the test suite: \`id\` is absent from reasoning items in the \`store: false\` path and present as expected in the \`store: true\` path.

## Checklist

- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run \`pnpm changeset\` in the project root)

## Related Issues

Fixes #12687

### DIFF
--- a/.changeset/fix-openai-omit-reasoning-id-when-store-false.md
+++ b/.changeset/fix-openai-omit-reasoning-id-when-store-false.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+Omit `id` from reasoning items when `store: false` — Azure OpenAI rejects requests where a stateless reasoning item includes an `id` field.
--- a/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
@@ -1606,7 +1606,6 @@ describe('convertToOpenAIResponsesInput', () => {
             [
               {
                 "encrypted_content": "encrypted_content_001",
-                "id": "reasoning_001",
                 "summary": [
                   {
                     "text": "First reasoning step",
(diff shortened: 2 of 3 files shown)


gonzoblasco:
Clean fix — the root cause is correctly identified and the change is minimal. One observation:

**Missing positive test coverage for `store: true`**

The PR removes `id` assertions from existing `store: false` tests (13 lines removed), which confirms `id` is no longer emitted in that path. But there's no explicit test verifying that `id` **is** still present when `store: true`. If someone later refactors the conditional and accidentally drops `id` in both branches, the test suite wouldn't catch it.

Consider adding a test case (or adapting an existing one) that asserts `id` is present in the `store: true` path, so the behavior is documented on both sides of the branch:

```ts
it('includes id in reasoning items when store is true', () => {
  // ... existing setup with store: true
  expect(result.input).toContainEqual(
    expect.objectContaining({
      type: 'reasoning',
      id: 'reasoning_001', // explicitly assert presence
    })
  )
})
```

This would make the intent of the conditional clearer and prevent future regressions in either direction.

Otherwise, the fix looks correct and the changeset is appropriate.

gr2m:
Closing: Azure expects reasoning item IDs for stateless requests.

PR #14984, opened 2026-05-05
fix(ai): wrap unparseable tool-call input in JSON object

Mmartinrusso:
## Background

When a model returns a plain string instead of a JSON object as tool-call arguments, the SDK throws a parse error and surfaces it as a \`tool-result\` error to the caller. This breaks agentic loops where models occasionally produce non-JSON tool inputs during streaming.

## Summary

Wrap an unparseable tool-call input string in a JSON object \`{ input: "..." }\` instead of propagating a parse error. The tool handler still receives a usable value and can decide how to proceed.

## Manual Verification

Verified via the new unit test: a raw string input is wrapped and forwarded without throwing.

## Checklist

- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run \`pnpm changeset\` in the project root)

### DIFF
--- a/.changeset/fix-wrap-invalid-tool-call-input-as-json-object.md
+++ b/.changeset/fix-wrap-invalid-tool-call-input-as-json-object.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):
do you have a script reproduction available that simulates the break with a call via bedrock?

PR #15007, opened 2026-05-05
feat(mistral): add MistralEmbeddingModelOptions with encodingFormat provider option

Mmartinrusso:
## Background

The konsistent convention requires every `*-model.ts` file to have a matching `*-model-options.ts` companion file. `mistral-embedding-model.ts` was excluded from this check and also hardcoded `encoding_format: 'float'` with no way for callers to override it.

## Summary

- Add `mistral-embedding-model-options.ts` exporting `MistralEmbeddingModelOptions`, satisfying the konsistent convention
- Expose `encodingFormat` (`"float"` | `"base64"`) as a provider option, replacing the previous hardcoded `"float"`. Default remains `"float"` — no breaking change
- Export `MistralEmbeddingModelOptions` from the package root
- Remove `packages/mistral/src/mistral-embedding-model.ts` from the konsistent exclusion list

## Manual Verification

- TypeScript compiles cleanly (`pnpm tsc --noEmit` in `packages/mistral`)
- Existing embedding tests continue to pass (default `encoding_format: 'float'` preserved when no option is passed)
- konsistent lint passes with the exclusion removed

## Checklist
- [x] All commits are signed
- [ ] Tests have been added / updated

### DIFF
--- a/.changeset/feat-mistral-embedding-model-options.md
+++ b/.changeset/feat-mistral-embedding-model-options.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/mistral': patch
+---
+
+Add `MistralEmbeddingModelOptions` type with `encodingFormat` provider option, enabling callers to request `"base64"` encoding instead of the default `"float"`.
--- a/.github/konsistent.json
+++ b/.github/konsistent.json
@@ -204,7 +204,6 @@
             "packages/gateway/src/gateway-image-model.ts",
             "packages/gateway/src/gateway-language-model.ts",
             "packages/gateway/src/gateway-reranking-model.ts",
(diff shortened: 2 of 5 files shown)


Mmartinrusso:
Good catch! Fixed in the latest commit:

- Updated `MistralTextEmbeddingResponseSchema` to accept `z.union([z.array(z.number()), z.string()])` so Zod validates both response formats
- Added `decodeBase64Embedding()` helper that decodes the base64 string back to `number[]` via `Float32Array` (IEEE 754 float32 binary encoding)

`'float'` responses are returned as-is; `'base64'` responses are decoded transparently before being returned.

PR #13744, opened 2026-03-23
fix(openai, openai-compatible): send null content for tool-only assistant messages

dancer:
## background

when an assistant message contains only tool-call parts (no text), `convertToOpenAICompatibleChatMessages` and `convertToOpenAIChatMessages` send `content: ""`. providers backed by AWS Bedrock reject this with `ValidationException: messages: text content blocks must be non-empty`

## summary

- change `content: text` to `content: text || null` for assistant messages in both `@ai-sdk/openai-compatible` and `@ai-sdk/openai`
- update tests to expect `null` instead of `""` for tool-only assistant messages

## checklist

- [x] tests have been added / updated (for bug fixes / features)
- [ ] documentation has been added / updated (for bug fixes / features)
- [x] a _patch_ changeset for relevant packages has been added (run `pnpm changeset` in root)
- [x] i have reviewed this pull request (self-review)

## related issues

fixes #13466

## Future work

The same `content: text` pattern exists in `@ai-sdk/xai`, `@ai-sdk/deepseek`, `@ai-sdk/groq`, and `@ai-sdk/mistral` and will hit the same Bedrock error under equivalent routing. Tracked separately in #14612.

PR #14987, opened 2026-05-05
fix(xai, deepseek, groq, mistral): send null content for tool-only assistant messages

Mmartinrusso:
## Background

xAI, DeepSeek, Groq, and Mistral strictly validate the \`content\` field on assistant messages. When an assistant turn contains only tool calls (no text), their message converters were setting \`content: ''\` (empty string). Several of these APIs reject an empty string and expect \`null\` or the field to be omitted.

## Summary

In each converter, set \`content: null\` when there are tool calls and no text content, matching the behavior of the OpenAI converter and the expectations of these providers.

## Manual Verification

Verified by inspecting serialized messages in unit tests: tool-only assistant turns now carry \`content: null\`.

## Checklist

- [x] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run \`pnpm changeset\` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #15043, opened 2026-05-05
fix(xai,groq,deepseek,mistral): send null content for tool-only assistant messages

### LINKED ISSUE #14612: Bug: tool-only assistant messages send content:"" in xai, deepseek, groq, mistral providers
## Summary

Follow-up to #13466 / #13744.

The same pattern that caused Bedrock `ValidationException: messages: text content blocks must be non-empty` in `@ai-sdk/openai-compatible` and `@ai-sdk/openai` also exists in four other provider packages. When an assistant message contains only `tool-call` parts (no text), the converter emits `content: ""`, which Bedrock-backed endpoints reject.

## Affected files

- `packages/xai/src/convert-to-xai-chat-messages.ts:111`
- `packages/deepseek/src/chat/convert-to-deepseek-chat-messages.ts:126`
- `packages/groq/src/convert-to-groq-chat-messages.ts:105`
- `packages/mistral/src/convert-to-mistral-chat-messages.ts:122`

Each has the same shape:

```ts
messages.push({
  role: 'assistant',
  content: text, // <-- empty string when only tool-calls are present
});
```

## Proposed fix

Apply the same change as #13744:

```ts
content: text || null,
```

This matches the OpenAI spec (content is nullable when `tool_calls` is present) and what official OpenAI SDKs send.

## Scope

- Update each converter.
- Update corresponding unit tests to expect `null` for tool-only assistant messages.
- Add a patch changeset covering the four packages.

Note: Mistral uses its own API shape; verify Mistral's spec accepts `null` before applying there. The other three (xai, deepseek, groq) are OpenAI-compatible.

Mmartinrusso:
## Summary

Fixes #14612

When an assistant message contains only tool calls (no text content), these providers now correctly send `content: null` instead of `content: ""`. Several providers (xAI, DeepSeek, Groq, Mistral) reject requests with `content: ""` on tool-call-only turns, breaking multi-step tool use.

**Changed providers:**
- `@ai-sdk/xai`
- `@ai-sdk/groq`
- `@ai-sdk/deepseek`
- `@ai-sdk/mistral`

**Pattern applied:**
```ts
// Before
content: text,

// After
content: toolCalls.length > 0 ? text || null : text,
```

This is the same fix already applied to the OpenAI provider in #13744. When `text` is empty and tool calls are present, `null` is sent instead of `""`, which is what these providers expect per their API specs.

## Test plan

- [x] All 4 provider converter test suites pass
- [x] Existing tests updated to assert `content: null` (was `content: ""`) for tool-only assistant turns
- [x] Mixed text+tool turns still send the text string (unchanged behavior)
- [x] Text-only turns (no tool calls) continue to send `""` as before

### DIFF
--- a/.changeset/fix-tool-only-assistant-null-content.md
+++ b/.changeset/fix-tool-only-assistant-null-content.md
@@ -1,0 +1,13 @@
+---
+'@ai-sdk/xai': patch
+'@ai-sdk/groq': patch
+'@ai-sdk/deepseek': patch
+'@ai-sdk/mistral': patch
+---
+
+fix(xai,groq,deepseek,mistral): send null instead of empty string for content in tool-only assistant messages
+
+When an assistant message contains only tool calls (no text), these providers now
+send `content: null` instead of `content: ""`. Several providers (xAI, DeepSeek,
+Groq, Mistral) reject requests with `content: ""` on tool-call-only turns.
+This matches the fix already applied to the OpenAI provider in #13744.
--- a/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
+++ b/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
@@ -131,7 +131,7 @@ describe('convertToDeepSeekChatMessages', () => {
         {
           "messages": [
             {
-              "content": "",
+              "content": null,
               "reasoning_content": undefined,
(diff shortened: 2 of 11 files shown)


Mmartinrusso:
Closing as duplicate of #14987 — same fix opened twice.

PR #15047, opened 2026-05-05
docs(ui): clarify persisted message id generation

### LINKED ISSUE #8109: Documentation for server side id generation is misleading
### Description

The documentation [here](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-message-persistence#setting-up-server-side-id-generation) is misleading, since the id generation is only for the response message, and not for the request message. 

This statement leads to confusion:

Image

which implies that since the id generation of response messages is handled in the backend per default, the new behaviour using an id generator would be that both ids are generated in the backend. I couldn't reproduce that result. It only returned the id of the response message.

### AI SDK Version

2.0.13

### Code of Conduct

EfeDurmaz16:
## Summary
- clarify that server-side response ID generators apply to assistant response messages
- explain that user message IDs are created before the request reaches the API route
- avoid implying both request and response message IDs are generated by the backend

Fixes #8109

## Verification
- git diff --check


gr2m:
Thanks for the correct clarification here. I incorporated this wording into #16493 against current `main` and credited you as co-author, since this branch had gone stale. Closing this PR as superseded by #16493.

PR #15050, opened 2026-05-06
fix(gateway): encode inline v4 file part bytes as `{ type: 'data' }` instead of a `data:` URL

R-Taneja:
## Summary

The v4 prompt spec defines `LanguageModelV4FilePart.data` as a tagged discriminated union ([`SharedV4FileData`](https://github.com/vercel/ai/blob/main/packages/provider/src/shared/v4/shared-v4-file-data.ts)): `{ type: 'data' | 'url' | 'reference' | 'text', ... }`. But `gateway-language-model.ts`'s `maybeEncodeFileParts` was carrying over a v3-era pattern that converted inline `Uint8Array` bytes into a `data:` URL on the `'url'` variant. That made every consumer (gateway servers and providers) round-trip the bytes back out of the URL before they could use them.

Encode the bytes as a base64 string under the `'data'` variant directly:

```diff
- filePart.data = {
-   type: 'url',
-   url: new URL(`data:${mediaType};base64,${base64}`),
- };
+ filePart.data = {
+   type: 'data',
+   data: base64,
+ };
```

Same wire payload semantics, smaller (no `data:image/png;base64,` prefix overhead), and matches what the v4 spec already guarantees. URL file parts are unchanged — `JSON.stringify` serializes URL instances via `toJSON()`, so server-side they still arrive as plain strings.

## Why this isn't a backport candidate

This is v7-only behavior — the `'data'` variant is a v4 concept. v3 (AI SDK v6) and v2 (AI SDK v5) `data` was a flat `Uint8Array | string | URL` union with no tagged variants, so the URL conversion was the only clean wire shape there. v3/v5 should keep their current behavior.

## Test plan

- [x] All 59 tests in `gateway-language-model.test.ts` pass
- [x] Existing tests that asserted the old `data:` URL shape updated to expect the inline `{ type: 'data', data }` shape
- [ ] No new regression tests needed beyond the updated assertions — the existing `should encode Uint8Array image part…` cases cover the encoding path


packages/gateway/src/gateway-language-model.ts
```diff
@@ -208,20 +207,13 @@ export class GatewayLanguageModel implements LanguageModelV4 {
...
           if (
             filePart.data.type === 'data' &&
             filePart.data.data instanceof Uint8Array
           ) {
-            const buffer = Uint8Array.from(filePart.data.data);
-            const base64Data = Buffer.from(buffer).toString('base64');
             filePart.data = {
-              type: 'url',
-              url: new URL(
-                `data:${filePart.mediaType || 'application/octet-stream'};base64,${base64Data}`,
```
shaper, line 223: do we lose mediaType insight with the change? if so, is it material -- are we using it server-side or passing it through in a meaningful way, or can/do we reverse engineer from the data.

R-Taneja: see `LanguageModelV4FilePart` in `packages/provider/src/language-model/v4/language-model-v4-prompt.ts`

4 fields:
```
type
filename
data
mediaType
providerOptions
```

`mediaType` is a separate field to `data` and is already set when we reach this code, we just mutate `data`

upstream we read from `mediaType` anyways so it's ok

felixarntz (approved):

PR #15063, opened 2026-05-06
feat: make include stable

lgrammel:
## Background

`experimental_include` was introduced in #12126

## Summary

* Introduce `include` and deprecate `experimental_include`.
* add include to `ToolLoopAgent`

## Related Issues

Fixes #12131


gr2m (approved):

packages/ai/src/agent/tool-loop-agent-settings.ts
```diff
@@ -188,6 +190,16 @@ export type ToolLoopAgentSettings<
...
 
+    /**
+     * Settings for controlling what data is included in step results.
+     * Disabling inclusion can help reduce memory usage when processing
+     * large payloads like images.
+     *
+     * By default, request and response bodies are included, and request
+     * messages are excluded.
+     */
+    include?: GenerateTextInclude & StreamTextInclude;
```
lgrammel, line 201: @gr2m also added it to ToolLoopAgent, in case this is useful for WorkflowAgent as well

PR #14949, opened 2026-05-04
feat: sandbox shell execution abstraction

lgrammel:
## Background

Many agents are using filesystems through shell and file read/write tools, often in separate sandbox environments. These agents are so common that a first-class sandbox abstraction would be beneficial.

## Summary

* add `Sandbox` type
* add `sandbox` option to `generateText`, `streamText`, `ToolLoopAgent`
* make sandbox available in `ToolExecutionOptions`

## Example
Tool definition:
```ts
import { tool } from 'ai';
import { z } from 'zod';

export function sandboxShellTool() {
  return tool({
    description: 'Run a shell command',
    inputSchema: z.object({
      command: z.string(),
    }),

    execute: async ({ command }, { sandbox }) => {
      // TODO figure out type inference to turn the runtime error into a type error
      if (!sandbox) {
        throw new Error('Sandbox is not available');
      }
      return sandbox.executeCommand({ command });
    },
  });
}
```

Agent definition:
```ts
import { openai } from '@ai-sdk/openai';
import { ToolLoopAgent } from 'ai';
import { sandboxShellTool } from '../../tools/sandbox-shell-tool';

export const sandboxAgent = new ToolLoopAgent({
  model: openai('gpt-5.5'),

  tools: {
    shell: sandboxShellTool(),
  },

  prepareCall: ({ sandbox, ...rest }) => ({
    ...rest,
    instructions:
      `You are a helpful assistant that can run shell commands.\n` +
      `You are operating in the following sandbox: ${sandbox?.description}`,
  }),
});
```

Agent call:
```ts
import { Bash } from 'just-bash';
import { JustBashSandbox } from '../../sandbox/just-bash-sandbox';
import { sandboxAgent } from './sandbox-agent';

const sandbox = new JustBashSandbox(
  new Bash({
    cwd: '/home/user',
  }),
);

const result = await sandboxAgent.stream({
  prompt:
    'Create a file named greeting.txt with a short greeting, then list the files and show the file contents.',
  sandbox,
});
```

## Manual Verification

- [x] agent - generate `src/agent/openai/generate-local-sandbox`
- [x] agent - stream `src/agent/openai/stream-local-sandbox`
- [x] test with ui `examples/ai-e2e-next/app/chat/sandbox/page.tsx`

## Future Work

* read/write files in sandboxes
* explore generic typing of sandbox tools
* move sandbox shell tool into ai package
* allow choosing a sandbox in prepareStep
* add timeout to executeCommand
* figure out sandbox streaming
[...]

PR #15097, opened 2026-05-07
feat: add readFile to Sandbox

lgrammel:
## Background

We introduced a Sandbox abstraction for bash execution in #14949

Another common operation is reading files.

## Summary

Add basic `readFile` support for Sandbox.

## Related Issues

Builds on #14949


nicoalbanese (dismissed):

felixarntz (commented):
@lgrammel What's the planned scope for the `Sandbox` abstraction, in terms of methods to require?

Fundamentally, I think a sandbox only needs to have `executeCommand`. Reading files usually still happens through executing a command. A `readFile` tool could implement the rest itself I think?

I'm not sure about introducing more methods on the main `Sandbox` contract. If we want filesystem methods, how about we have `readFile` and `writeFile` etc. on a separate contract?

Maybe we could have `Filesystem` if we want to provide these? Then sandbox implementations could choose whether they are `Sandbox` or `Sandbox & Filesystem`.

lgrammel:
@felixarntz i agree on the separation in theory, but we are optimizing for agents/tools that use sandboxes and separating the filesystem would introduce unnecessary complexity. usually filesystems are tied to sandboxes for advanced uses cases, e.g. partial file reads and writes. especially tools like the openai patch tool and the anthropic text editor tool would need those.

the combined sandbox abstraction also aligns with the minimal sandbox spec malte had, and e.g. in open agents the bash execution approach you suggest broke for large files and a dedicated readFile tool was required. there is also an advantage in having dedicated read/write tools for approval reasons. cc @nicoalbanese

packages/provider-utils/src/types/sandbox.ts
```diff
@@ -33,4 +33,19 @@ export type Sandbox = {
...
+  readonly readFile: (options: {
+    /**
+     * Path to the file to read.
+     */
+    path: string;
+  }) => PromiseLike<{
+    /**
+     * Content of the file.
+     */
+    content: string;
```
lgrammel: there is an issue around supporting binary files

felixarntz:
@lgrammel That rationale makes sense, thanks. All good then from my perspective.

Just one follow-up, could you clarify what the intended scope is? Other than `executeCommand` and `readFile`, what do we want to have on the `Sandbox` abstraction? I assume `writeFile`? how about `deleteFile`? anything else?

examples/ai-functions/src/sandbox/vercel-sandbox.ts
```diff
@@ -1,5 +1,6 @@
 import { type Sandbox as AISandbox } from 'ai';
 import type { Sandbox as VercelSandboxSDK } from '@vercel/sandbox';
+import { text } from 'node:stream/consumers';
 
 const rootDirectory = '/vercel/sandbox';
```
felixarntz, line 5: I know that's the default root, but should we consider having `rootDirectory` as a general option on our sandboxes? I think having that makes sense on pretty much every sandbox, to control the default location where commands are executed. It's already in the demo `LocalSandbox` implementation like that, but here we hard-code the path.

lgrammel: those are all currently in user land. i agree we should have those in sandbox constructors once we add specific sandboxes in one fashion or another.

lgrammel:
> @lgrammel That rationale makes sense, thanks. All good then from my perspective.
> 
> Just one follow-up, could you clarify what the intended scope is? Other than `executeCommand` and `readFile`, what do we want to have on the `Sandbox` abstraction? I assume `writeFile`? how about `deleteFile`? anything else?

we are currently thinking executeCode, readFile, writeFile will be the minimal needed. you are right, maybe deleteFile as well.

put the pr on hold for more discussion next week, esp around how to handle different types of files (binary vs text)

felixarntz:
@lgrammel For reference, here is something I built a while ago where I was thinking about a sandbox abstraction. I called it "environment" instead of "sandbox", but same foundational idea: https://github.com/felixarntz/ai-code-agents/blob/main/packages/environment-utils/src/types.ts

It's an older exploration I'm not maintaining, but there may be some useful ideas in there - whether to apply or to intentionally rule out.

To be clear, I like where we're going here, it's much simpler in how it doesn't couple tools with the sandbox by definition. But I thought it's worth comparing.

PR #15109, opened 2026-05-08
fix(bedrock): strip signed reasoning with no tools

jeph:
## Background

This was originally investigated as an opencode `/compact` failure when using Anthropic models through Amazon Bedrock. In opencode, `/compact` creates a summary request from the existing conversation and sends it without active tools. The failure only appeared after a Bedrock Anthropic model had returned thinking tokens. In that state, running `/compact` returned: 

```
undefined: The model returned the following errors: messages.1.content.1: \`thinking\` or \`redacted_thinking\` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.
```

The issue turned out to be in the prompt shape produced by the Amazon Bedrock provider. opencode uses AI SDK as its model provider dependency, and the Bedrock provider already filters historical tool-call/tool-result content when the current request has no active tools, because Bedrock does not support tool content without active tools. For Anthropic models, those historical assistant messages can also include signed or redacted reasoning blocks. Before this change, the provider removed the tool content but left the signed reasoning behind. That meant Claude received signed thinking inside a transcript whose surrounding assistant/tool structure had already been rewritten, so Claude rejected the request as modified thinking.

I first considered fixing this in opencode by stripping reasoning during compaction, but the behavior is caused by the Bedrock provider's no-active-tools prompt rewrite. Since opencode depends on AI SDK for Bedrock request serialization, it made more sense to keep the fix in AI SDK where that rewrite happens. This keeps consumers from needing to know that signed Anthropic reasoning must be removed when Bedrock tool content has already been stripped.

## Summary

- Strip Bedrock signed/redacted reasoning metadata when filtering tool content from Anthropic Bedrock prompts without active tools.
- Preserve signed/redacted reasoning for non-Anthropic Bedrock models.
- Add coverage for both Anthropic filtering and non-Anthropic preservation.
- Add a patch changeset for `@ai-sdk/amazon-bedrock`.

## Manual Verification

I verified this with opencode because opencode uses AI SDK's Amazon Bedrock provider for model calls. opencode's `/compact` command summarizes the existing conversation by sending prior messages to the model with no active tools. With opencode on `dev` and the unpatched `@ai-sdk/amazon-bedrock@4.0.96` package installed, running `/compact` after a Bedrock Anthropic conversation had returned thinking tokens produced this error: 

```
undefined: The model returned the following errors: messages.1.content.1: \`thinking\` or \`redacted_thinking\` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response.
```

With a locally built `@ai-sdk/amazon-bedrock@4.0.96` package containing this fix installed into the same opencode `dev` checkout, the same `/compact` flow completed successfully. In the fixed request path, Bedrock still removes historical tool-call/tool-result content when no tools are active, but Anthropic signed/redacted reasoning is removed from that same rewritten prompt so Claude no longer receives invalid signed thinking blocks.

The serialized Bedrock request body is also verified by provider tests in this PR.

## Checklist

## Related Issues

I could not find an existing issue for this exact no-active-tools prompt filtering case. #13583 and #14071 are similar but distinct Bedrock Anthropic signed-thinking validation issues: #13583 was about trimming signed reasoning text, and #14071 was about dropping empty text blocks between reasoning blocks. This PR addresses a different mutation path where the Bedrock provider removes historical tool-call/tool-result content while leaving signed/redacted Anthropic reasoning in the rewritten prompt.


packages/amazon-bedrock/src/amazon-bedrock-chat-language-model.ts
```diff
@@ -400,6 +407,28 @@ export class AmazonBedrockChatLanguageModel implements LanguageModelV4 {
             message => message.role === 'system' || message.content.length > 0,
           ) as typeof prompt;
 
+        if (isAnthropicModel) {
```
jeph, line 410: There's probably a slight performance improvement if we merge this with the filter block above, but I think having two separate blocks makes the code a bit easier to read. Let me know if you'd like me to change this though.

PR #15120, opened 2026-05-08
fix(mcp): use negotiated protocol version in HTTP transport headers

### LINKED ISSUE #14413: [@ai-sdk/mcp] Streamable HTTP transport does not downgrade `mcp-protocol-version` header after version negotiation (breaks servers like Figma Dev Mode MCP)
## Summary

The `HttpMCPTransport` (and `SseMCPTransport`) in `@ai-sdk/mcp` always emits `mcp-protocol-version: <LATEST_PROTOCOL_VERSION>` in every HTTP request, ignoring the version that was actually negotiated with the server during `initialize`. This violates the MCP spec and causes post-initialization requests to fail against any server that does not (yet) support the client's latest version — notably **Figma Dev Mode MCP Server**, which supports up to `2025-06-18` as of April 2026.

## Spec reference

> The protocol version sent by the client **SHOULD** be the one negotiated during initialization.
>
> If the server receives a request with an invalid or unsupported `MCP-Protocol-Version`, it **MUST** respond with `400 Bad Request`.

— https://modelcontextprotocol.io/specification/2025-06-18/basic/transports (Protocol Version Header section, same wording in 2025-11-25)

The official `@modelcontextprotocol/sdk` client respects this: after `initialize`, it stores `result.protocolVersion` and uses it in every subsequent request's `mcp-protocol-version` header.

## Observed behavior (repro against real Figma Desktop MCP)

```
>>> GET  http://127.0.0.1:3845/mcp                mcp-protocol-version: 2025-11-25
<<< 400
>>> POST http://127.0.0.1:3845/mcp (initialize)   mcp-protocol-version: 2025-11-25
<<< 200   response body: { ..., "protocolVersion": "2025-06-18", ... }   <- server negotiated DOWN
>>> POST http://127.0.0.1:3845/mcp (notifications/initialized)   mcp-protocol-version: 2025-11-25
<<< 400   error: "Unsupported protocol version (supported versions: 2025-06-18, 2025-03-26, 2024-11-05, 2024-10-07)"
```

`createMCPClient({ transport: { type: 'http', url: ... } })` never resolves — the follow-up `notifications/initialized` request is rejected because the client keeps sending its own `LATEST_PROTOCOL_VERSION` in the header instead of the negotiated `2025-06-18`.

## Expected behavior

After `initialize` returns a `protocolVersion`, the transport should use that exact string in the `mcp-protocol-version` header for every subsequent request (per the spec).

## Minimal reproduction

```js
// Node 22+, @ai-sdk/mcp@1.0.36 (and 2.0.0-beta.25, also affected)
import { createMCPClient } from "@ai-sdk/mcp";

// 1. Start Figma Desktop, enable Dev Mode MCP Server in Preferences (exposes port 3845)
// 2. Run:
  transport: { type: "http", url: "http://127.0.0.1:3845/mcp" },
});
// Throws:
//   MCP HTTP Transport Error: POSTing to endpoint (HTTP 400):
//   {"error":{"code":-32000,"message":"Bad Request: Unsupported protocol version ..."}}
```

Direct curl with `2025-06-18` in both body and header succeeds — confirming the server is fine and the bug is purely in the client's post-init header handling.

## Affected versions

Reproduced in:
- `@ai-sdk/mcp@1.0.36` (latest on the 1.x line)
- `@ai-sdk/mcp@2.0.0-beta.25` (latest beta)

In both, `commonHeaders()` in `HttpMCPTransport` and `SseMCPTransport` hardcodes:
```js
"mcp-protocol-version": LATEST_PROTOCOL_VERSION
```
No logic exists to store or emit the negotiated `result.protocolVersion`.

## Related

- #13023 added `2025-11-25` to `SUPPORTED_PROTOCOL_VERSIONS` but did not address the header-downgrade issue.
- #7575 / #10315 are the same class of version-support bugs, also not addressing header downgrade.

## Suggested fix

Add a `protocolVersion` field to each transport, set by the client after successful `initialize`:

```diff
 async commonHeaders(base) {
   const headers = {
     ...this.headers,
     ...base,
-    "mcp-protocol-version": LATEST_PROTOCOL_VERSION,
+    "mcp-protocol-version": this.protocolVersion ?? LATEST_PROTOCOL_VERSION,
   };
   // ...
 }
```

And in `DefaultMCPClient.init()`, right after the `SUPPORTED_PROTOCOL_VERSIONS.includes(result.protocolVersion)` check:

```diff
 this.serverCapabilities = result.capabilities;
+if (this.transport) this.transport.protocolVersion = result.protocolVersion;
 this._serverInfo = result.serverInfo;
```

I've validated this exact patch locally via `pnpm patch` — it makes the Figma Dev Mode MCP handshake succeed end-to-end (6 tools discovered, `notifications/initialized` now returns 200). Happy to open a PR.

## Environment

- OS: macOS 15 (Darwin 24.6.0)
- Node: 22.14.0
- `@ai-sdk/mcp`: 1.0.36 and 2.0.0-beta.25
- MCP server: Figma Dev Mode MCP Server (Figma Desktop, built-in — supports up to `2025-06-18`)

nileshpatil6:
## Problem

After the MCP `initialize` handshake, the server returns the protocol version it actually supports. `HttpMCPTransport` ignores this and always sends `LATEST_PROTOCOL_VERSION` (`2025-11-25`) in the `mcp-protocol-version` header on every subsequent request.

Servers that only support an older version (e.g. Figma Dev Mode MCP supports `2025-06-18`) reject all post-init requests with **HTTP 400**, making the transport completely unusable with those servers.

Fixes #14413.

## Root Cause

`commonHeaders()` hardcodes `LATEST_PROTOCOL_VERSION`:

```ts
// before
'mcp-protocol-version': LATEST_PROTOCOL_VERSION,
```

The negotiated version from the `initialize` result was never stored anywhere on the transport.

## Fix

- Add a private `negotiatedProtocolVersion?: string` field.
- Add a private `captureNegotiatedVersion(msg)` helper that detects an `initialize` result (any JSON-RPC response whose `result` contains a `protocolVersion` string) and stores it.
- Call `captureNegotiatedVersion` before every `onmessage` dispatch — in `send()` (both the direct-JSON and inline-SSE paths) and in `openInboundSse()`.
- Use `this.negotiatedProtocolVersion ?? LATEST_PROTOCOL_VERSION` in `commonHeaders()`.

```ts
// after
'mcp-protocol-version': this.negotiatedProtocolVersion ?? LATEST_PROTOCOL_VERSION,
```

The fix is self-contained in `mcp-http-transport.ts` — no changes to `mcp-client.ts` or the `MCPTransport` interface. Before `initialize` completes the field is `undefined`, so `LATEST_PROTOCOL_VERSION` is used for the handshake itself (correct per spec). After `initialize`, all subsequent requests carry the server's negotiated version.

### DIFF
--- a/packages/mcp/src/tool/mcp-http-transport.ts
+++ b/packages/mcp/src/tool/mcp-http-transport.ts
@@ -37,6 +37,7 @@ export class HttpMCPTransport implements MCPTransport {
   private inboundSseConnection?: { close: () => void };
   private redirectMode: RequestRedirect;
   private fetchFn: FetchFunction;
+  private negotiatedProtocolVersion?: string;
 
   // Inbound SSE resumption and reconnection state
   private lastInboundEventId?: string;
@@ -72,13 +73,28 @@ export class HttpMCPTransport implements MCPTransport {
     this.fetchFn = fetchFn ?? globalThis.fetch;
   }
 
+  private captureNegotiatedVersion(msg: JSONRPCMessage): void {
+    if (
+      'result' in msg &&
+      msg.result !== null &&
+      typeof msg.result === 'object' &&
+      'protocolVersion' in msg.result &&
+      typeof (msg.result as Record<string, unknown>).protocolVersion === 'string'
+    ) {
+      this.negotiatedProtocolVersion = (
+        msg.result as Record<string, unknown>
+      ).protocolVersion as string;
+    }
+  }
+
   private async commonHeaders(
     base: Record<string, string>,
   ): Promise<Record<string, string>> {
     const headers: Record<string, string> = {
       ...this.headers,
       ...base,
-      'mcp-protocol-version': LATEST_PROTOCOL_VERSION,
+      'mcp-protocol-version':
+        this.negotiatedProtocolVersion ?? LATEST_PROTOCOL_VERSION,
     };
 
     if (this.sessionId) {
@@ -215,7 +231,10 @@ export class HttpMCPTransport implements MCPTransport {
           const messages: JSONRPCMessage[] = Array.isArray(data)
             ? data.map((m: unknown) => JSONRPCMessageSchema.parse(m))
             : [JSONRPCMessageSchema.parse(data)];
-          for (const m of messages) this.onmessage?.(m);
+          for (const m of messages) {
+            this.captureNegotiatedVersion(m);
+            this.onmessage?.(m);
+          }
           return;
         }
 
@@ -243,6 +262,7 @@ export class HttpMCPTransport implements MCPTransport {
                 if (event === 'message') {
                   try {
                     const msg = await parseJSONRPCMessage(data);
+                    this.captureNegotiatedVersion(msg);
                     this.onmessage?.(msg);
                   } catch (error) {
                     const e = new MCPClientError({
@@ -391,6 +411,7 @@ export class HttpMCPTransport implements MCPTransport {
             if (event === 'message') {
               try {
                 const msg = await parseJSONRPCMessage(data);
+                this.captureNegotiatedVersion(msg);
                 this.onmessage?.(msg);
               } catch (error) {
                 const e = new MCPClientError({


aayush-kapoor (commented):
marking as closed in favour of https://github.com/vercel/ai/pull/15128

PR #15122, opened 2026-05-08
fix(ai): deprecate properties on result that have moved to finalStep

lgrammel:
## Background

Result properties specific to the final step are available under `finalStep` with #15115 

## Summary

Deprecate properties that have moved.

## Related Issues

Builds on #15115


content/docs/08-migration-guides/23-migration-guide-7-0.mdx
```diff
@@ -944,6 +944,56 @@ console.log(result.usage); // all steps
...
+});
+
+const finalStep = await result.finalStep;
+
+console.log(finalStep.reasoningText);
+console.log(finalStep.request.body);
+console.log(finalStep.response.headers);
+```
+
+The deprecated top-level aliases continue to work in AI SDK 7 so you can migrate incrementally.
```
gr2m, line 995: This is only true for the five `result.*` properties mentioned in this section, right

> `reasoning`, `reasoningText`, `request`, `response`, and `providerMetadata`

For e.g. `usage`, we will do the behavior change?

gr2m (approved):

PR #15124, opened 2026-05-08
feat: add optional `abortSignal` parameters to sandbox command execution

lgrammel:
## Background
We introduced a Sandbox abstraction for bash execution in #14949

## Summary
Add `abortSignal` option to `executeCommand`

## Related Issues
Builds on #14949

### DIFF
--- a/packages/anthropic/src/tool/bash_20250124.ts
+++ b/packages/anthropic/src/tool/bash_20250124.ts
@@ -79,12 +79,12 @@ export function bash_20250124<OUTPUT>(
   if (execute === undefined) {
     return bash_20250124_internal({
       ...rest,
-      execute: async ({ command }, { sandbox }) => {
+      execute: async ({ command }, { abortSignal, sandbox }) => {
         if (!sandbox) {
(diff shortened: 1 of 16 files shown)


packages/anthropic/src/tool/bash_20250124.ts
```diff
@@ -79,12 +79,12 @@ export function bash_20250124<OUTPUT>(
...
     return bash_20250124_internal({
       ...rest,
-      execute: async ({ command }, { sandbox }) => {
+      execute: async ({ command }, { abortSignal, sandbox }) => {
         if (!sandbox) {
           throw new Error('Sandbox is not available');
         }
 
-        return await sandbox.executeCommand({ command });
+        return await sandbox.executeCommand({ command, abortSignal });
```
felixarntz, line 87: what about OpenAI's `local-shell` tool and the `local` version of its `shell` tool? do they support `abortSignal` too? just want to make sure we double check.

felixarntz (approved):

PR #15143, opened 2026-05-09
fix(ai): correct off-by-one in array form of mock model helpers

### LINKED ISSUE #15141: Bug: MockLanguageModelV3 / V4 array form for doGenerate/doStream skips first entry (off-by-one)
## Description

When `MockLanguageModelV3` (and `MockLanguageModelV4` on `main`) is constructed with an **array** of results for `doGenerate` (or `doStream`), the first call returns the **second** array entry instead of the first. Subsequent calls are likewise shifted by one. The same off-by-one is present in `MockEmbeddingModelV3.doEmbed` and `MockEmbeddingModelV4.doEmbed`.

The most natural use case for the array form — scripting a deterministic multi-round conversation, e.g. tool-call → tool-result → final text for `ToolLoopAgent` testing — silently malfunctions: the agent never sees the first scripted response and may terminate prematurely, never invoking the tool. There is no error; the output just diverges from the spec.

## Reproduction

```ts

  doGenerate: [
    {
      content: [{ type: "text", text: "FIRST" }],
      finishReason: { unified: "stop", raw: "stop" },
      usage: {
        inputTokens: { total: 1, noCache: 1, cacheRead: 0, cacheWrite: 0 },
        outputTokens: { total: 1, text: 1, reasoning: 0 },
      },
      warnings: [],
    },
    {
      content: [{ type: "text", text: "SECOND" }],
      finishReason: { unified: "stop", raw: "stop" },
      usage: {
        inputTokens: { total: 1, noCache: 1, cacheRead: 0, cacheWrite: 0 },
        outputTokens: { total: 1, text: 1, reasoning: 0 },
      },
      warnings: [],
    },
  ],
});

const r1 = await model.doGenerate({} as any);
console.log("call 1:", (r1.content[0] as { text: string }).text);
//   expected: "FIRST"
//   actual:   "SECOND"

const r2 = await model.doGenerate({} as any);
console.log("call 2:", r2);
//   expected: { content: [{ type: "text", text: "SECOND" }], ... }
//   actual:   undefined
```

## Root cause

`packages/ai/src/test/mock-language-model-v3.ts` and `packages/ai/src/test/mock-language-model-v4.ts` (on `main`) both have:

```ts
this.doGenerate = async (options) => {
  this.doGenerateCalls.push(options);              // length is incremented…
  if (typeof doGenerate === "function") {
    return await doGenerate(options);
  } else if (Array.isArray(doGenerate)) {
    return doGenerate[this.doGenerateCalls.length]; // …then used as the index
  } else {
    return doGenerate;
  }
};
```

After the `push`, `doGenerateCalls.length` is `1` on the first call, so the indexer returns `doGenerate[1]` (the second entry). On the last call, the index goes out of bounds and the mock returns `undefined`, which crashes any consumer expecting a `LanguageModelV(3|4)GenerateResult`.

The same bug is in:

- `MockLanguageModelV3.doStream` array form
- `MockLanguageModelV4.doGenerate` and `.doStream` (v7 `main`)
- `MockEmbeddingModelV3.doEmbed` array form
- `MockEmbeddingModelV4.doEmbed` array form (verified, same code on `main`)

## Suggested fix

Two options, equivalent in behaviour:

```ts
// Option A: index before incrementing (smaller diff)
return doGenerate[this.doGenerateCalls.length - 1];
```

```ts
// Option B: independent counter
let i = 0;
this.doGenerate = async (options) => {
  this.doGenerateCalls.push(options);
  // ...
  return Array.isArray(doGenerate) ? doGenerate[i++] : doGenerate;
};
```

Calls past the array's end probably should fall back to the last entry rather than `undefined` (current behaviour) — small follow-up that would help long agent loops.

## Workaround

Use the function form with a closure counter:

```ts
let i = 0;
new MockLanguageModelV3({
  doGenerate: async () => results[i++] ?? results.at(-1),
});
```

## Versions

- `ai`: 6.0.168 (and reproduces on `latest` 6.0.177)
- `ai`: 7.0.0-beta.116 (v7 `main`) — identical code in `MockLanguageModelV4`, same bug

## Offering a PR

Happy to send a PR that fixes both V3 and V4 mocks (language + embedding) at once, with tests, a patch changeset, and a signed commit — if the suggested behaviour (index from 0, last-entry fallback past end) is acceptable. Let me know if there's a preference between options A and B above.

serhiizghama:
## Problem

`MockLanguageModelV3` / `MockLanguageModelV4` and `MockEmbeddingModelV3` / `MockEmbeddingModelV4` accept an array of results to script multi-call test scenarios (e.g. tool-call → tool-result → final text for `ToolLoopAgent`). The array form silently misbehaved: the first call returned the **second** entry, and the last call returned `undefined`.

The cause is the indexer running after the call has already been pushed onto the call list:

```ts
this.doGenerate = async (options) => {
  this.doGenerateCalls.push(options);          // length is incremented…
  if (Array.isArray(doGenerate)) {
    return doGenerate[this.doGenerateCalls.length]; // …then used as the index
  }
  ...
};
```

The same off-by-one is in `doStream` for both V3 and V4 and in `doEmbed` for both embedding mocks.

Fixes #15141.

## Solution

Index with `length - 1` after the push so the n-th call returns the n-th array entry, matching the obvious user expectation.

## Testing

Added `mock-language-model-v3.test.ts`, `mock-language-model-v4.test.ts`, `mock-embedding-model-v3.test.ts`, `mock-embedding-model-v4.test.ts` covering both `doGenerate` / `doStream` / `doEmbed` array forms. Verified the new tests fail against `main` and pass with the fix. Existing `wrap-language-model` and `wrap-embedding-model` tests continue to pass (they use empty arrays, so behavior is unchanged for that path).

### DIFF
--- a/.changeset/mock-model-array-off-by-one.md
+++ b/.changeset/mock-model-array-off-by-one.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+Fix off-by-one in array form of `MockLanguageModelV3` / `V4` and `MockEmbeddingModelV3` / `V4`. The first scripted call now returns the first array entry instead of the second; subsequent calls advance correctly.
--- a/packages/ai/src/test/mock-embedding-model-v3.test.ts
+++ b/packages/ai/src/test/mock-embedding-model-v3.test.ts
@@ -1,0 +1,18 @@
+import { describe, it, expect } from 'vitest';
+import { MockEmbeddingModelV3 } from './mock-embedding-model-v3';
+
+describe('MockEmbeddingModelV3', () => {
+  describe('doEmbed array form', () => {
+    it('should return entries in order starting from the first', async () => {
+      const model = new MockEmbeddingModelV3({
+        doEmbed: [{ embeddings: [[1]] } as any, { embeddings: [[2]] } as any],
+      });
+
+      const r1 = await model.doEmbed({} as any);
+      const r2 = await model.doEmbed({} as any);
+
+      expect(r1.embeddings).toEqual([[1]]);
+      expect(r2.embeddings).toEqual([[2]]);
+    });
+  });
+});
--- a/packages/ai/src/test/mock-embedding-model-v3.ts
+++ b/packages/ai/src/test/mock-embedding-model-v3.ts
@@ -39,7 +39,7 @@ export class MockEmbeddingModelV3 implements EmbeddingModelV3 {
       if (typeof doEmbed === 'function') {
         return doEmbed(options);
       } else if (Array.isArray(doEmbed)) {
-        return doEmbed[this.doEmbedCalls.length];
+        return doEmbed[this.doEmbedCalls.length - 1];
       } else {
         return doEmbed;
       }
--- a/packages/ai/src/test/mock-embedding-model-v4.test.ts
+++ b/packages/ai/src/test/mock-embedding-model-v4.test.ts
@@ -1,0 +1,18 @@
+import { describe, it, expect } from 'vitest';
+import { MockEmbeddingModelV4 } from './mock-embedding-model-v4';
+
+describe('MockEmbeddingModelV4', () => {
+  describe('doEmbed array form', () => {
+    it('should return entries in order starting from the first', async () => {
+      const model = new MockEmbeddingModelV4({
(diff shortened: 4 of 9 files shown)


lgrammel:
Fixed via #15154

PR #15150, opened 2026-05-10
feat(openai): add opt-in pass-through for unsupported file media types

### LINKED ISSUE #14747: Allow overriding hardcoded file media type filtering
### Description

# Problem

The AI SDK currently applies a hardcoded allowlist when converting user file parts for some providers.

For example, in the OpenAI Responses path, the SDK accepts `image/*` and `application/pdf`, but rejects other file media types at the SDK layer, even when the target provider can accept file inputs generically.

This creates a mismatch between SDK behavior and provider capabilities, and forces developers to maintain local patches for legitimate use cases.

This is especially problematic for applications that intentionally avoid provider-side file storage and rely on inline stateless file inputs such as `input_file.file_data`, for privacy, retention, and infrastructure reasons.

Provider capabilities evolve over time.
A hardcoded SDK allowlist will tend to become outdated and too restrictive.

In practice, this means:
- supported or potentially supported file types cannot be forwarded to the provider
- developers lose access to provider-native file handling
- the only current workaround is patching the SDK locally

# Proposed solution

Please expose an opt-in way to override or extend the SDK file filtering/conversion logic.

Any of the following would solve the problem:

1. a configurable `fileMediaTypeFilter`
2. a provider-level override such as `passThroughUnsupportedFiles`
3. a hook such as `mapFilePart(part) => providerInputPart`
4. a generic fallback mode that forwards non-image file parts as raw provider file inputs instead of rejecting them

The default behavior can remain strict for safety and compatibility, but advanced integrators should be able to opt into pass-through behavior when they know their target provider supports it.

# Expected outcome

The SDK should not force a restrictive allowlist when a provider supports a broader file input model.
Instead, it should offer a safe default plus an explicit escape hatch for advanced use cases.

### AI SDK Version

- ai: 6.0.146
- @ai-sdk/openai: 3.0.50

### Code of Conduct

priyanshu-9785maker:
Closes #14747
## Background

Some providers support broader inline file handling than the SDK currently allows.  
Currently, unsupported non-image inline files are rejected or dropped unless they are PDFs.

This PR adds an explicit opt-in escape hatch for advanced users while preserving existing strict defaults and backward compatibility.

## Summary

Adds an opt-in `passThroughUnsupportedFiles` provider option for OpenAI/OpenResponses file-part conversion.

Default behavior is unchanged:
- non-image inline files still reject/drop unless they are `application/pdf`

When enabled:
- unsupported media types are forwarded to the provider as `input_file` parts instead of being rejected/dropped

## Manual Verification

Verified with targeted vitest runs:

- `convert-to-open-responses-input.test.ts`
- `convert-to-openai-responses-input.test.ts`
- `openai-responses-language-model.test.ts`

Also verified:
- existing default validation behavior remains unchanged
- unsupported files pass through only when explicitly enabled
- request payloads serialize correctly as `input_file`

## Checklist

- [ ] A patch changeset for relevant packages has been added

## Future Work

Potential future improvements:
- document provider compatibility expectations
- consider a more granular allowlist-based configuration
- extend behavior consistently across additional providers

### DIFF
--- a/packages/open-responses/src/responses/convert-to-open-responses-input.test.ts
+++ b/packages/open-responses/src/responses/convert-to-open-responses-input.test.ts
@@ -228,6 +228,54 @@ describe('convertToOpenResponsesInput', () => {
         },
       ]);
     });
+
+    it('should pass through non-image file parts when explicitly enabled', async () => {
+      const result = await convertToOpenResponsesInput({
+        prompt: [
+          {
+            role: 'user',
+            content: [
+              {
+                type: 'file',
+                data: { type: 'data' as const, data: 'aGVsbG8=' },
+                mediaType: 'text/plain',
+                filename: 'notes.txt',
+              },
+              {
+                type: 'file',
+                data: {
+                  type: 'url' as const,
+                  url: new URL('https://example.com/data.csv'),
+                },
+                mediaType: 'text/csv',
+              },
+            ],
+          },
+        ],
+        passThroughUnsupportedFiles: true,
+      });
+
+      expect(result.input).toMatchInlineSnapshot(`
+        [
+          {
+            "content": [
+              {
+                "file_data": "data:text/plain;base64,aGVsbG8=",
+                "filename": "notes.txt",
+                "type": "input_file",
+              },
+              {
+                "file_url": "https://example.com/data.csv",
+                "type": "input_file",
+              },
+            ],
+            "role": "user",
+            "type": "message",
+          },
+        ]
+      `);
+      expect(result.warnings).toEqual([]);
+    });
   });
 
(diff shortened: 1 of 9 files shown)


aayush-kapoor (commented):
markign as closed in favour https://github.com/vercel/ai/pull/15297 since open-responses doesn't need this change.

will add you as a co-author for the other changes. for future PRs please ensure that all your commits are signed

priyanshu-9785maker:
Sir ,this is my first contribution it's not good but I try my best

priyanshu-9785maker:
Sir ,this is my first contribution it's not good but I try my best

PR #15156, opened 2026-05-11
feat: flexible tool descriptions

lgrammel:
## Background

Tool description might need to use information from the tool context or the sandbox to be more effective.

## Summary

Allow tool descriptions to be functions that take context and sandbox as args.

## Future Work
* explore how workflow agent can support flexible tool descriptions

### DIFF
--- a/packages/workflow/src/serializable-schema.ts
+++ b/packages/workflow/src/serializable-schema.ts
@@ -43,7 +43,7 @@ export function serializeToolSet(
   return Object.fromEntries(
     Object.entries(tools).map(([name, t]) => {
       const def: SerializableToolDef = {
-        description: t.description,
+        description: t.description as string, // TODO support tools with function descriptions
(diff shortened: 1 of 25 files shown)


packages/workflow/src/serializable-schema.ts
```diff
@@ -43,7 +43,7 @@ export function serializeToolSet(
   return Object.fromEntries(
     Object.entries(tools).map(([name, t]) => {
       const def: SerializableToolDef = {
-        description: t.description,
+        description: t.description as string, // TODO support tools with function descriptions
```
lgrammel, line 46: @gr2m it is not clear to me how we could best support this in the workflow agent; one option could be to resolve the string value by calling the function

lgrammel:
TODO update documentation

nicoalbanese (approved):

PR #15186, opened 2026-05-12
fix(ai): preserve tool result provider options

### LINKED ISSUE #15185: `tool.toModelOutput` returned `providerOptions` silently dropped, Anthropic `cache_control` never reaches the API
### Description

`tool.toModelOutput` accepts a `providerOptions` field in its return type, but the field is never propagated to the tool-result message part. For the Anthropic provider this means `providerOptions.anthropic.cacheControl` set inside `toModelOutput` doesn't reach the API. As such, prompt caching cannot be enabled for tool-result content via this path.

The field appears in the type, is accepted at runtime, and emits no warning. The bug is silent: usage just reports `cache_creation_input_tokens`: 0 indefinitely and there is no signal that anything is wrong.

**Repro**
Standalone script (no project dependencies): [[gist]](https://gist.github.com/itshuey/27dda5d09e4fee6454da0ea5a87ee506)
```
ANTHROPIC_API_KEY=sk-... npx tsx ai-sdk-issue-repro.ts
```
Two patterns compared side-by-side, same model (`claude-opus-4-1`), same ~2,300-token content:
- B (control) — assistant message with a text content part bearing `providerOptions.anthropic.cacheControl: { type: "ephemeral" }`. Run cold then warm.
- C (bug) — tool whose `toModelOutput` returns a content part with the same `providerOptions.anthropic.cacheControl`. Forces the tool to be called, then runs a follow-up turn against the same conversation history.

**Expected**
Both patterns engage caching. Cold runs report `cache_creation_input_tokens > 0`, warm runs report `cache_read_input_tokens > 0`.

**Actual**
B works:
```
=== B1 cold ===
providerMetadata.anthropic.usage: {
  "cache_creation_input_tokens": 1337,
  "cache_read_input_tokens": 0,
  ...
}
=== B2 warm ===
providerMetadata.anthropic.usage: {
  "cache_creation_input_tokens": 0,
  "cache_read_input_tokens": 1337,
  ...
}
```
C never caches:
```
--- C step 1 (would write cache) ---
providerMetadata.anthropic.usage: {
  "cache_creation_input_tokens": 0,
  "cache_read_input_tokens": 0,
  ...
}
--- C step 2 (would read cache) ---
providerMetadata.anthropic.usage: {
  "cache_creation_input_tokens": 0,
  "cache_read_input_tokens": 0,
  ...
}
```
Inspecting the outbound request body confirms no `cache_control` field is set on the `tool_result` content block.

**Investigation**
The drop site is in the ai package's tool-result part assembly. `createToolModelOutput` returns whatever `toModelOutput` produced (including its `providerOptions`), but the assembled part places it at `part.output`:
```
// packages/ai/src/generate-text/...  (compiled location: ai/dist/index.mjs:4158)
toolContent.push({
  type: "tool-result",
  toolCallId: output2.toolCallId,
  toolName: output2.toolName,
  output: modelOutput,   // ← providerOptions ends up at part.output.providerOptions
});
```
The Anthropic converter reads cache control from `part.providerOptions` and `message.providerOptions`, not `part.output.providerOptions`:

```
// packages/providers/anthropic/src/convert-to-anthropic-messages-prompt.ts:288
const cacheControl =
  validator.getCacheControl(part.providerOptions, {
    type: 'tool result part',
    canCache: true,
  }) ??
  (isLastPart
    ? validator.getCacheControl(message.providerOptions, {
        type: 'tool result message',
        canCache: true,
      })
    : undefined);
```
`providerOptions` declared inside `toModelOutput`'s return value lives one level too deep at `part.output.providerOptions` and never gets read.

Notably, part-level `providerOptions` are supported elsewhere: the denied-tool-approval branch a few lines below the assembly site explicitly sets `providerOptions` on the part for OpenAI's `approvalId`. So the part shape accepts the field, only the `toModelOutput` path fails to populate it.

**Why this matters**
Tool-result caching is the highest-value caching opportunity in multi-step agent loops. Large, stable tool outputs (loaded skills, fetched data, retrieved documents) get re-sent in every subsequent step. Without this path working:

- Anthropic users have to construct synthetic assistant messages by hand to work around it (we have).
- Bedrock-Anthropic users have no workaround — Bedrock rejects top-level `cache_control` with 400: `cache_control`: Extra inputs are not permitted, so `toModelOutput` is the only available mechanism there.

**Versions**
- `ai`: 6.x
- `@ai-sdk/anthropic`: 3.x
- Confirmed identical behavior on `@ai-sdk/amazon-bedrock 4.x` (delegates to `@ai-sdk/anthropic/internal`; bug location is in the core ai package, not provider-specific)

**Related**
[#5942](https://github.com/vercel/ai/issues/5942) — providerOptions dropped on user-message parts. Different drop site, same family.
[#12832](https://github.com/vercel/ai/issues/12832) — Bedrock-Anthropic doesn't accept top-level `cache_control`. Establishes that the `toModelOutput` path is the only viable cache mechanism on Bedrock.

**Not proposing a specific fix**
The right call (propagate `output.providerOptions` to `part.providerOptions`? add a separate `partProviderOptions` field on the return type? remove `providerOptions` from `toModelOutput`'s return type since it's currently inert?) involves API-shape decisions I don't want to prescribe. Happy to draft a PR once a direction is chosen.

### AI SDK Version

- `ai`: 6.x
- `@ai-sdk/anthropic`: 3.x

### Code of Conduct

Genmin:
## Summary
- closes #15185
- promotes provider options returned by `tool.toModelOutput()` onto generated `tool-result` parts when no explicit tool-result metadata is already present
- applies the fallback across `generateText`, `streamText`, `toResponseMessages`, and UI message conversion so provider converters can see cache-control metadata consistently
- adds regression coverage for generated prompts and response message conversion

## Root cause
`toModelOutput()` can return provider options on the `ToolResultOutput`, but the generated `tool-result` parts only carried existing part-level provider metadata. Provider adapters such as Anthropic read cache-control settings from the `tool-result` part, so metadata returned by `toModelOutput()` was dropped before conversion.

## Testing
- `corepack pnpm --filter ai exec vitest --config vitest.node.config.js --run src/generate-text/generate-text.test.ts src/generate-text/to-response-messages.test.ts`
- `corepack pnpm --filter ai exec vitest --config vitest.node.config.js --run src/ui/convert-to-model-messages.test.ts`
- `corepack pnpm --filter ai type-check`
- `corepack pnpm --filter ai build`
- `corepack pnpm exec ultracite check .changeset/tool-model-output-provider-options.md packages/ai/src/prompt/get-tool-result-provider-options.ts packages/ai/src/generate-text/generate-text.ts packages/ai/src/generate-text/stream-text.ts packages/ai/src/generate-text/to-response-messages.ts packages/ai/src/ui/convert-to-model-messages.ts packages/ai/src/generate-text/generate-text.test.ts packages/ai/src/generate-text/to-response-messages.test.ts`
- `git diff --check`

### DIFF
--- a/.changeset/tool-model-output-provider-options.md
+++ b/.changeset/tool-model-output-provider-options.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): preserve provider options from tool result model output
--- a/packages/ai/src/generate-text/generate-text.test.ts
+++ b/packages/ai/src/generate-text/generate-text.test.ts
@@ -5533,6 +5533,80 @@ describe('generateText', () => {
 
       expect(result.text).toStrictEqual('provider metadata test');
     });
+
+    it('should pass toModelOutput provider options to tool result parts', async () => {
+      let responseCount = 0;
+      let secondPrompt: LanguageModelV4Prompt | undefined;
+
+      await generateText({
+        model: new MockLanguageModelV4({
+          doGenerate: async ({ prompt }) => {
+            switch (responseCount++) {
+              case 0:
+                return {
+                  ...dummyResponseValues,
+                  content: [
+                    {
+                      type: 'tool-call',
+                      toolCallType: 'function',
+                      toolCallId: 'call-1',
+                      toolName: 'cachedTool',
+                      input: '{}',
+                    },
+                  ],
+                  finishReason: { unified: 'tool-calls', raw: undefined },
+                };
+              case 1:
+                secondPrompt = prompt;
+                return {
+                  ...dummyResponseValues,
+                  content: [{ type: 'text', text: 'done' }],
+                };
+              default:
+                throw new Error(`Unexpected response count: ${responseCount}`);
+            }
+          },
+        }),
+        tools: {
+          cachedTool: tool({
+            inputSchema: z.object({}),
+            execute: async () => 'cached result',
+            toModelOutput: ({ output }) => ({
+              type: 'text',
+              value: String(output),
+              providerOptions: {
+                anthropic: { cacheControl: { type: 'ephemeral' } },
+              },
+            }),
+          }),
+        },
+        prompt: 'test-input',
+        stopWhen: isStepCount(3),
+      });
+
+      const toolMessage = secondPrompt?.find(
+        message => message.role === 'tool',
+      );
+      const toolResult = toolMessage?.content.find(
+        part => part.type === 'tool-result',
+      );
+
+      expect(toolResult).toMatchObject({
+        type: 'tool-result',
+        toolCallId: 'call-1',
+        toolName: 'cachedTool',
+        output: {
+          type: 'text',
+          value: 'cached result',
+          providerOptions: {
+            anthropic: { cacheControl: { type: 'ephemeral' } },
+          },
+        },
+        providerOptions: {
+          anthropic: { cacheControl: { type: 'ephemeral' } },
+        },
+      });
+    });
   });
 
   describe('options.reasoning', () => {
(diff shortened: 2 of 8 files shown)


aayush-kapoor (commented):
not the right fix - it can be contained in the anthropic provider

PR #15193, opened 2026-05-12
fix(openai): round-trip namespace on function_call input items

edawerd:
## Background

[#14789](https://github.com/vercel/ai/pull/14789) added support for `namespace` on the **read** side of OpenAI `function_call` items: when `tool_search` dispatches a deferred tool, the resulting `function_call` carries a `namespace` field, and the response handler surfaces it via `providerMetadata.openai.namespace` on the resulting `tool-call` content part.

## Problem

The **write** side was never updated. When a `tool-call` part is sent back to the model on a subsequent turn or step, `convertToOpenAIResponsesInput` serializes it as:

```ts
{ type: 'function_call', call_id, name, arguments, id }
```

with no `namespace`. OpenAI then rejects the request with:

```
Missing namespace for function_call '<name>'. It does not exist in the
default namespace. Round-trip the model's function_call item with its
namespace field included.
```

(HTTP 400, `param: 'input[N].namespace'`)

This only reproduces when a deferred tool actually runs and the conversation continues — i.e. multi-step or multi-turn flows with `tool_search`. A single-shot dispatch never round-trips, which is presumably why #14789's tests didn't catch it.

## Fix

`convert-to-openai-responses-input.ts` now reads `namespace` from `providerOptions.openai.namespace` or `providerMetadata.openai.namespace` on `tool-call` parts (mirroring how `itemId` is round-tripped) and includes it on the serialized `function_call` item when present.

Also adds `namespace?: string` to the `OpenAIResponsesFunctionCall` request type.

## Tests

Two new tests in `convert-to-openai-responses-input.test.ts`:

1. namespace flows from `providerMetadata.openai.namespace` (the realistic round-trip path)
2. namespace flows from `providerOptions.openai.namespace` (the explicit-set path)

Both assert the serialized `function_call` includes `namespace`.

## Scope

- Diff: +1 type field, +13 source lines, +82 test lines
- No behavior change when `namespace` is absent (most callers)
- Companion to #14789

## Verified

- `pnpm vitest run src/responses/convert-to-openai-responses-input.test.ts` → 111 passed (109 existing + 2 new)
- `pnpm type-check` clean for openai package


WonderPG:
Would greatly appreciate this fix !

edawerd:
Can someone from the vercel team take a look?

WonderPG:
+1 🦫

BoBer78:
🦫

aayush-kapoor (approved):

PR #15197, opened 2026-05-12
fix(provider/cohere): use secure parsing instead of JSON.parse

CloudFaye:
## Background

The AI SDK already centralizes JSON parsing through `parseJSON` / `safeParseJSON` so provider code gets consistent error handling and prototype-key protection. Cohere's streaming tool-call path was still parsing completed tool arguments with raw `JSON.parse`, so this brings that path in line with the rest of the SDK.

## Summary

This updates Cohere streamed tool-call parsing to use the AI SDK's secure JSON parser instead of calling `JSON.parse` directly in production code.

The regression coverage now exercises streamed tool-call arguments containing prototype keys and verifies that they are rejected by the shared secure parser.

### DIFF
--- a/.changeset/cohere-tool-call-json-hardening.md
+++ b/.changeset/cohere-tool-call-json-hardening.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/cohere': patch
+---
+
+Harden streamed tool call argument parsing against prototype keys.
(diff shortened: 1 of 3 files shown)


lgrammel (approved):

lgrammel:
@CloudFaye thank you. please sign the commits so i can merge this pr. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

CloudFaye:
Thanks for the heads up. I’ve force-pushed the branch with the commit replayed and signed via SSH. The current head is 0cd5a5d71c004a2f241ed20e90b5ecee757353e1.

PR #15240, opened 2026-05-13
fix(devtools): tidy ToolItem expand/collapse logic

wong2:
## Summary
- The chevron in `ToolItem` is now shown whenever there's expandable content (parameters or description), not only when parameters exist — previously a tool with only a description had no way to reveal the full text.
- Expanded view shows the full description (if any) followed by the parameters JSON (if any).
- Collapsed view shows only the truncated description.

## Question

I noticed that the `ToolDefinition` in `devtools` is using `parameters` instead of `inputSchema`, so it will never render. Is this a bug?

### DIFF
--- a/.changeset/devtools-toolitem-expand.md
+++ b/.changeset/devtools-toolitem-expand.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/devtools": patch
+---
+
+Fix `ToolItem` expand/collapse logic: chevron now shows whenever there's expandable content (parameters or description), expanded view shows full description plus parameters, collapsed view shows truncated description only.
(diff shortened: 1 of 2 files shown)


lgrammel (approved):

lgrammel:
Hi! please sign the commits, otherwise I cannot merge this PR. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

wong2:
@lgrammel done

wong2:
@lgrammel How about this question? 
> I noticed that the ToolDefinition in devtools is using parameters instead of inputSchema, so it will never render. Is this a bug?

lgrammel:
yes most likely a bug that needs fixing

wong2:
ok i can make a pr

PR #15253, opened 2026-05-13
fix: make sandbox experimental

lgrammel:
## Background

The sandbox abstraction is in early development and will most likely require major changes and iteration.

## Summary

Mark sandbox as experimental.


gr2m (approved):

packages/provider-utils/src/types/sandbox.ts
```diff
@@ -1,7 +1,7 @@
 /**
- * Sandbox environment that can execute commands.
+ * Experimental sandbox environment that can execute commands.
  */
-export type Sandbox = {
+export type Experimental_Sandbox = {
```
felixarntz, line 4: nit: internally I've previously seen us use the pattern where we keep the unprefixed name and only export with `Experimental_` prefix from the package boundary. Means we need fewer diffs in general.

packages/anthropic/src/tool/bash_20241022.ts
```diff
@@ -1,8 +1,8 @@
 import {
   createProviderDefinedToolFactory,
   lazySchema,
+  type Experimental_Sandbox,
```
felixarntz: similarly for imports we can alias to unprefixed name (`type Experimental_Sandbox as Sandbox`), to keep diff much smaller. this applies throughout the PR.

felixarntz (commented):
LGTM, though one recommendation for smaller diff plus following a previously used pattern for `Experimental_` stuff.

felixarntz (approved):

PR #15254, opened 2026-05-13
fix(langchain): reset streamEvents text between model invocations

eyueldk:
## Background

When `toUIMessageStream` consumes LangGraph **`streamEvents`** (v2), `processStreamEventsEvent` tracks streaming text with `textStarted` / `textMessageId` (and reasoning with `reasoningStarted` / `reasoningMessageId`). Those flags were only cleared at the end of the whole stream, not when a **new** chat model run started.

If a graph performs multiple LLM calls in one run (e.g. tool use then model again), **`on_chat_model_start`** can fire again while `textStarted` is still `true`. Subsequent **`text-delta`** chunks then reuse the first text part’s id, so later prose appends to that first block. Tool events still append **new** tool parts after it, so the assistant `UIMessage`’s `parts` no longer reflect chronological order in the UI.

## Summary

- On **`on_chat_model_start`**, if the previous model turn left streams open, emit **`reasoning-end`** and **`text-end`** with the correct ids, then reset `reasoningStarted` / `reasoningMessageId` and `textStarted` / `textMessageId`.
- After that, set **`messageId`** from `event.run_id` or `event.data.run_id` as before.

This matches the intent of ending open streams when a new model invocation begins, so the next **`text-start`** / **`reasoning-start`** pair applies to the new turn.

## Manual Verification

- Use a LangGraph agent with **`streamEvents`** (v2) that **invokes the chat model more than once** in the same run (e.g. tools then model again), with text on both segments.
- Pipe the iterator through **`toUIMessageStream`** and consume the UI message stream.
- Confirm the resulting assistant message has **multiple text parts in order** (one per model segment) and that **tool parts** appear in the order events occurred, not all text merged into the first text part.

## Checklist

## Future Work

Optional: add an automated test that simulates two `on_chat_model_start` / `on_chat_model_stream` sequences and asserts the emitted chunk sequence includes **`text-end`** before the second segment’s **`text-start`**.


packages/langchain/src/adapter.ts
```diff
@@ -148,10 +148,39 @@ function processStreamEventsEvent(
   if (!event.data) return;
 
   switch (event.event) {
-    case 'on_chat_model_start': {
+    case "on_chat_model_start": {
```
dqbd: _nit_: would revert the format here

lgrammel:
Please add a patch changeset.

lgrammel (approved):

PR #15263, opened 2026-05-13
docs: add note about tool execution events

lgrammel:
## Background

We tried to refactor how tool execution events happen when streaming. However, we ran into limitations because of ordering guarantess.

## Summary

Add comment with our learnings. Rename variable.


gr2m (approved):
alas, bummer it didn't work out

PR #15271, opened 2026-05-13
feat(ai): remove telemetry dependency on onChunk callback

aayush-kapoor:
## Background

after the change introduced in https://github.com/vercel/ai/pull/15241, the statistics stored in telemetry such as `ai.response.msToFinish` and `ai.response.avgOutputTokensPerSecond` can now be extracted via the step result performance. 

thus removing the dependency on the `onChunk` telemetry callback

## Summary

legacy-otel now creates those attributes through the onStepFinish event

## Manual Verification

na

## Checklist


packages/ai/src/generate-text/stream-text.test.ts
aayush-kapoor, line 1: the regression in values here is observed because we removed a `now()` call in stream-text.ts (`const msToFirstChunk = now() - stepStartTimestampMs;` this line was removed) 

so the timing read in tests shifts left by one mocked timestamp

lgrammel:
Can `onChunk` be removed from the `Telemetry` interface?

lgrammel (approved):

PR #14893, opened 2026-05-01
fix(xai): remove unsupported schema property flags

Genmin:
## Background

xAI rejects JSON schema `additionalProperties: false` entries in tool schemas, while treating object schemas as closed by default. The SDK currently forwards those flags to xAI function tools, and response-format schemas can carry the same unsupported flag.

## Summary

- Add an xAI-scoped schema sanitizer that removes `additionalProperties: false` recursively while leaving non-boolean `additionalProperties` schemas intact.
- Apply the sanitizer to xAI Chat Completions function tools and JSON schema response formats.
- Apply the same sanitizer to xAI Responses function tools and JSON schema response formats.
- Add focused helper, tool-preparation, and request-shape tests for both xAI model surfaces.
- Add a patch changeset for `@ai-sdk/xai`.

## Manual Verification

- `pnpm --filter @ai-sdk/xai test:node -- src/remove-unsupported-xai-schema-properties.test.ts src/xai-prepare-tools.test.ts src/xai-chat-language-model.test.ts src/responses/xai-responses-prepare-tools.test.ts src/responses/xai-responses-language-model.test.ts`
- `pnpm --filter @ai-sdk/xai type-check`
- `pnpm exec ultracite check packages/xai/src/remove-unsupported-xai-schema-properties.ts packages/xai/src/remove-unsupported-xai-schema-properties.test.ts packages/xai/src/xai-prepare-tools.ts packages/xai/src/xai-prepare-tools.test.ts packages/xai/src/xai-chat-language-model.ts packages/xai/src/xai-chat-language-model.test.ts packages/xai/src/responses/xai-responses-prepare-tools.ts packages/xai/src/responses/xai-responses-prepare-tools.test.ts packages/xai/src/responses/xai-responses-language-model.ts packages/xai/src/responses/xai-responses-language-model.test.ts`
- `pnpm changeset status --since origin/main`
- `git diff --check HEAD~1 HEAD`

## Checklist

- [ ] All commits are signed (local signing is not configured on this machine, so leaving this unchecked)
- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (not needed for this provider request-shape fix)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

Fixes #14678

PR #15279, opened 2026-05-14
fix(xai): omit unsupported additionalProperties false

### LINKED ISSUE #14678: addAdditionalPropertiesToJsonSchema breaks xAI/Grok tool calling: xAI rejects additionalProperties: false

raashish1601:
## Summary

Fixes #14678.

xAI rejects function tool schemas that include `additionalProperties: false`. This strips only boolean `additionalProperties: false` entries from xAI tool schemas before sending requests, while preserving the original input schema object and leaving other `additionalProperties` values intact.

The same normalization is applied to both chat completions and responses tool preparation.

## Validation

- `corepack pnpm --dir packages/xai exec vitest --config vitest.node.config.js --run src/xai-prepare-tools.test.ts src/responses/xai-responses-prepare-tools.test.ts src/xai-chat-language-model.test.ts`
- `corepack pnpm --filter @ai-sdk/xai type-check`
- `corepack pnpm exec ultracite check packages/xai/src/remove-additional-properties-false.ts packages/xai/src/xai-prepare-tools.ts packages/xai/src/xai-prepare-tools.test.ts packages/xai/src/xai-chat-language-model.test.ts packages/xai/src/responses/xai-responses-prepare-tools.ts packages/xai/src/responses/xai-responses-prepare-tools.test.ts`
- `corepack pnpm --filter @ai-sdk/xai test:node`
- `corepack pnpm --dir packages/xai clean`
- `corepack pnpm --dir packages/xai exec tsup --tsconfig tsconfig.build.json`

Note: `corepack pnpm --filter @ai-sdk/xai build` could not run directly in this Windows shell because the nested script invokes plain `pnpm`, which is not on PATH here. The equivalent clean + `tsup` build command passed.

### DIFF
--- a/.changeset/xai-tool-schema-additional-properties.md
+++ b/.changeset/xai-tool-schema-additional-properties.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+Remove unsupported `additionalProperties: false` entries from xAI function tool schemas.
--- a/packages/xai/src/remove-additional-properties-false.ts
+++ b/packages/xai/src/remove-additional-properties-false.ts
@@ -1,0 +1,17 @@
+export function removeAdditionalPropertiesFalse(schema: unknown): unknown {
+  if (Array.isArray(schema)) {
+    return schema.map(removeAdditionalPropertiesFalse);
+  }
+
+  if (schema == null || typeof schema !== 'object') {
+    return schema;
+  }
+
+  return Object.fromEntries(
+    Object.entries(schema)
+      .filter(
+        ([key, value]) => key !== 'additionalProperties' || value !== false,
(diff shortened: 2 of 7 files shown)


aayush-kapoor (commented):
there is already an open PR https://github.com/vercel/ai/pull/14893 - please do not create new prs if there already exists one.

PR #15280, opened 2026-05-14
fix(ai): wrap malformed tool call input in object for provider compatibility

### LINKED ISSUE #14442: Invalid tool call input stored as raw string causes API rejection on next step
## Bug Description

When `parseToolCall` fails to parse a tool call's JSON input, the catch block stores the raw string as `input`:

```ts
// parse-tool-call.ts (line 82-83)
const parsedInput = await safeParseJSON({ text: toolCall.input });
const input = parsedInput.success ? parsedInput.value : toolCall.input;
//                                                       ^^^^^^^^^^^^^^
//                                              raw string, not a JSON object
```

This raw string flows through `to-response-messages.ts` and `convert-to-language-model-prompt.ts` into the assistant message for the next API step. Providers like Amazon Bedrock require `toolUse.input` to be a JSON object (the Bedrock provider does `input: part.input as JSONObject` in `convert-to-bedrock-chat-messages.ts`), so the API rejects the request before the model ever sees the `tool-error` result.

## Expected Behavior

The model should see the `tool-error` and be able to retry. The invalid tool call's `input` should be wrapped in a valid JSON object (e.g., `{ rawInvalidInput: toolCall.input }`) so the conversation history remains valid for the next API call.

## Reproduction

1. Use a model via Amazon Bedrock with tool calling
2. Have the model generate a tool call with malformed JSON input (can be simulated with middleware that corrupts `tool-call` chunks)
3. The SDK correctly marks the tool call as `invalid: true` and generates a `tool-error`
4. On the next step, the API rejects with: `"The format of the value at messages.X.content.Y.toolUse.input is invalid. Provide a json object for the field and try again."`

## Suggested Fix

In `parse-tool-call.ts`, wrap the fallback input in an object:

```diff
- const input = parsedInput.success ? parsedInput.value : toolCall.input;
+ const input = parsedInput.success ? parsedInput.value : { rawInvalidInput: toolCall.input };
```

## Environment

- `ai` version: 6.0.48
- Provider: `@ai-sdk/amazon-bedrock`
- Model: Claude (Anthropic via Bedrock)

Christian-Sidak:
## Summary

- When `parseToolCall` fails to parse a tool call's JSON input, the raw string was stored as `input` directly
- Providers like Amazon Bedrock require `toolUse.input` to be a JSON object and reject requests where it is a raw string
- This caused the model to never see the `tool-error` result on the next step

The fix wraps the unparseable raw string in `{ rawInvalidInput: toolCall.input }` so the conversation history remains valid for the next API call.

Fixes #14442

## Test plan

- [x] Updated existing inline snapshots for invalid tool call cases to expect `{ rawInvalidInput: "..." }` instead of the raw string
- [x] Added new test verifying malformed JSON input is wrapped in `rawInvalidInput` object
- [x] All 16 `parseToolCall` tests pass locally

### DIFF
--- a/.changeset/fix-invalid-tool-input-object-wrap.md
+++ b/.changeset/fix-invalid-tool-input-object-wrap.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ai): wrap malformed tool call JSON input in object to satisfy provider requirements
--- a/packages/ai/src/generate-text/parse-tool-call.test.ts
+++ b/packages/ai/src/generate-text/parse-tool-call.test.ts
@@ -395,7 +395,9 @@ describe('parseToolCall', () => {
           "dynamic": true,
           "error": [AI_InvalidToolInputError: Invalid input for tool testTool: AI_JSONParseError: JSON parsing failed: Text: invalid json.
         Error message: SyntaxError: Unexpected token 'i', "invalid json" is not valid JSON],
-          "input": "invalid json",
+          "input": {
+            "rawInvalidInput": "invalid json",
+          },
           "invalid": true,
           "providerExecuted": undefined,
           "providerMetadata": undefined,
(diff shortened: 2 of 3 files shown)


ArshVermaGit (commented):
Nice fix overall — wrapping malformed tool call input in an object instead of passing through a raw string feels like the right compatibility tradeoff, especially for providers like Bedrock that strictly require structured JSON objects. I also like that this preserves the invalid payload for debugging while still allowing the conversation history to remain valid for subsequent tool-error handling. The added regression coverage and snapshot updates make the behavior change clear, and the scope here looks focused and low risk.

PR #15282, opened 2026-05-14
fix(ui): wait for chat stop abort handling to finish

### LINKED ISSUE #13962: chat.stop() fires the abort signal but returns immediately without waiting for the stream pipeline to fully terminate
### Description

chat.stop() doesn't appear to be doing very much and we ended up having to write an ever-growing helper to have it behave intuitively. Some of it manually, some with the help of claude opus. And if we have to wrap it this way then presumably every app that uses ai sdk eventually has to solve the same problems. Admittedly this is made worse when calling chat.stop() as part of a "clear chat" process where you stop the chat and clear the whole conversation by clearing all the messages and then don't want gibberish to keep arriving afterwards.

Here's our current chat helper which contains a fair amount of comments explaining some of the problems we've encountered:

```typescript
export async function stopChat(chat: Chat<AssistantMessage>) {
  // chat.stop() will cancel pending tool operations through its AbortSignal.
  // However, it will not update the message UI state and will actually retry sending the message by default.
  // In practice, this means it will keep re-running the tool call we're cancelling.
  // So we first set all existing tool calls to an error state before continuing with chat.stop()
  chat.messages = chat.messages.map((message): AssistantMessage => {
    const hasRunningTools = message.parts.some(
      (part) =>
        partIsToolUI(part) &&
        (part.state === 'approval-responded' ||
          part.state === 'input-available')
    );

    if (!hasRunningTools) {
      return message;
    }

    return {
      ...message,
      parts: message.parts.map((part) => {
        if (partIsToolUI(part) && getToolState(part.state) === 'running') {
          // Create a new tool part with error state
          const { ...basePart } = part;
          return {
            ...basePart,
            state: 'output-error' as const,
            output: undefined,
            errorText: 'Tool execution was cancelled',
            // This type does not resolve correctly.
          } as unknown as ToolUIPart;
        }
        return part;
      }),
    };
  });

  const wasActive = chat.status === 'streaming' || chat.status === 'submitted';

  await chat.stop();

  if (!wasActive) {
    return;
  }

  // chat.stop() fires the abort signal but returns immediately without
  // waiting for the stream pipeline to fully terminate. Wait for
  // makeRequest to catch the abort and set status to 'ready', otherwise
  // the stream's internal job queue can push buffered chunks back after
  // the caller clears messages.
  // Note: status can change asynchronously between the check above and
  // here (e.g. stop() may have already resolved the abort), so we
  // re-check at runtime.
  if (
    (chat.status as string) !== 'ready' &&
    (chat.status as string) !== 'error'
  ) {
    await new Promise<void>((resolve) => {
      const unsubscribe = chat['~registerStatusCallback'](() => {
        if (chat.status === 'ready' || chat.status === 'error') {
          unsubscribe();
          resolve();
        }
      });
    });
  }

  // Allow any final in-flight job executor writes to settle. The executor
  // uses microtasks internally so yielding to the macrotask queue ensures
  // they have all completed.
  await new Promise((resolve) => setTimeout(resolve, 0));
}
```

### AI SDK Version

@ai-sdk/react 3.0.143

### Code of Conduct

raashish1601:
## Summary
Fixes #13962.

`chat.stop()` aborts the active response but returned before `makeRequest()` finished handling the abort and resetting chat state. This tracks the active response completion and awaits it after aborting so callers can safely clear or replace messages after `stop()` resolves.

The regression test verifies that `stop()` resolves only after the abort path has finished and `onFinish` has run.

## Validation
- `corepack pnpm test:node -- src/ui/chat.test.ts` from `packages/ai` (runs the package node suite; 127 files, 2710 tests)
- `corepack pnpm type-check` from `packages/ai`
- `corepack pnpm exec ultracite check packages/ai/src/ui/chat.ts packages/ai/src/ui/chat.test.ts`
- `corepack pnpm clean` from `packages/ai`
- `corepack pnpm exec tsup --tsconfig tsconfig.build.json` from `packages/ai`
- `git diff --check`

Note: `corepack pnpm build` from `packages/ai` is not directly runnable in this Windows shell because the script calls bare `pnpm`, but the equivalent clean and tsup build steps above pass.

### DIFF
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -904,6 +904,58 @@ describe('Chat', () => {
     });
   });
 
+  it('should resolve stop after the aborted request has finished', async () => {
+    const events: Array<string> = [];
+    let controller!: ReadableStreamDefaultController<UIMessageChunk>;
+    const responseStream = new ReadableStream<UIMessageChunk>({
+      start: controllerArg => {
+        controller = controllerArg;
+
+        controller.enqueue({ type: 'start' });
+        controller.enqueue({ type: 'start-step' });
+        controller.enqueue({ type: 'text-start', id: 'text-1' });
+        controller.enqueue({
+          type: 'text-delta',
+          id: 'text-1',
+          delta: 'Hello',
+        });
+      },
+    });
+
+    const chat = new TestChat({
+      id: '123',
+      generateId: mockId(),
+      transport: {
+        sendMessages: async options => {
+          options.abortSignal?.addEventListener('abort', () => {
+            events.push('abort');
+            controller.error(new DOMException('Aborted', 'AbortError'));
+          });
+          return responseStream;
+        },
+        reconnectToStream: () => {
+          throw new Error('not implemented');
+        },
+      },
+      onFinish: () => {
+        events.push('finish');
+      },
+    });
+
+    chat.sendMessage({
+      text: 'Hello, world!',
+    });
+
+    while ((chat.messages[1]?.parts[1] as any)?.text !== 'Hello') {
+      await vi.advanceTimersByTimeAsync(0);
+    }
+
+    await chat.stop().then(() => events.push('stop-resolved'));
+
+    expect(events).toEqual(['abort', 'finish', 'stop-resolved']);
+    expect(chat.status).toBe('ready');
+  });
+
   it('should include the metadata of text message', async () => {
     server.urls['http://localhost:3000/api/chat'].response = {
       type: 'stream-chunks',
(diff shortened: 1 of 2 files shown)


ArshVermaGit (commented):
This looks like a solid concurrency/lifecycle fix. Making `chat.stop()` wait for the active response cleanup and `onFinish` handling before resolving feels like the correct behavior, especially for consumers relying on deterministic teardown after aborted streams. The regression test is also well-targeted since these timing-related issues are easy to accidentally reintroduce. Appreciate the thorough validation steps too — given the passing tests, type checks, linting, and build verification, this looks good to merge.

PR #15283, opened 2026-05-14
fix(xai): stop emitting additionalProperties flag

### LINKED ISSUE #14678: addAdditionalPropertiesToJsonSchema breaks xAI/Grok tool calling: xAI rejects additionalProperties: false

aayush-kapoor:
## Background

https://github.com/vercel/ai/issues/14678

xAI rejects JSON schema `additionalProperties: false` entries in tool schema only when the tool schema is defined with the explicit property `properties`.

stripping it from tool preparation should be fine since the xAI docs specify they define it as false by default

## Summary

- add an xAI-scoped schema sanitizer that removes `additionalProperties: false` recursively while leaving non-boolean additionalProperties schemas intact.

## Manual Verification

<details>
<summary>rerpro:</summary>

```ts
import { xai } from '@ai-sdk/xai';
import { z } from 'zod';

run(async () => {
    model: xai.responses('grok-4-1-fast-reasoning'),
    tools: {
      saveContact: tool({
        description: 'Save a contact with an address.',
        inputSchema: z.object({
          properties: z.object({
            city: z.string(),
            country: z.string(),
          }),
        }),
        execute: async ({ properties }) => ({
          saved: true,
          properties,
        }),
      }),
    },
    prompt:
      'Save Ada Lovelace as a contact in London, United Kingdom. Use the saveContact tool.',
  });

  console.log('Text:', result.text);
  console.log('Tool Calls:', JSON.stringify(result.toolCalls, null, 2));
  console.log('Tool Results:', JSON.stringify(result.toolResults, null, 2));
});
```

</details>

## Checklist

## Related Issues

fixes #14678

Co-authored-by: Genmin <joey@joeyroth.com>

### DIFF
--- a/packages/xai/src/remove-additional-properties.ts
+++ b/packages/xai/src/remove-additional-properties.ts
@@ -1,0 +1,23 @@
+/**
+ * Recursively removes `additionalProperties: false` entries from a JSON
+ * schema.
+ * Used to sanitize tool input schemas before sending them to the xAI API,
+ */
+export function removeAdditionalPropertiesFalse(value: unknown): unknown {
+  if (Array.isArray(value)) {
+    return value.map(removeAdditionalPropertiesFalse);
+  }
+
+  if (value == null || typeof value !== 'object') {
+    return value;
+  }
+
+  const result: Record<string, unknown> = {};
+  for (const [key, propertyValue] of Object.entries(value)) {
+    if (key === 'additionalProperties' && propertyValue === false) {
+      continue;
+    }
+    result[key] = removeAdditionalPropertiesFalse(propertyValue);
+  }
+  return result;
+}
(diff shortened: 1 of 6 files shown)


packages/xai/src/remove-additional-properties.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Recursively removes `additionalProperties: false` entries from a JSON
+ * schema.
+ * Used to sanitize tool input schemas before sending them to the xAI API,
```
gr2m: can you link to the docs that state that xAI is setting `additionalProperties: false` by default on their side?

gr2m (approved):

PR #15287, opened 2026-05-14
docs(ai): correct onToolCall JSDoc on return value handling

### LINKED ISSUE #15268: Docs: `onToolCall` JSDoc states a return value attaches a tool result, but the type returns `void`
### Description                                                                                                                                                                                                                                                                                                                                                                                                              
                 
                                                   
The JSDoc for `onToolCall` (on `useChat` / `AbstractChat`) tells you that you can return a tool result from the callback — "either synchronously or asynchronously". We took that at face value and wrote our handler that way. The chat hung silently, with nothing in the console pointing at the cause.                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                               
When we looked at the actual TypeScript type, the return is `void | PromiseLike<void>` — there's no slot for a returned tool result. And the SDK code doesn't read any return value either, so even when TypeScript lets it through (non-strict mode), the value just gets dropped.                                                                                                                                          
                                                                                                                                                                                                                                                                                                                                                                                                                               
So the JSDoc and the implementation tell two different stories. Trusting the JSDoc cost us a few hours of debugging a chat that wouldn't progress, with no error surfacing anywhere to point at the cause.                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                                             
 The actual way to attach a tool result — which we eventually pieced together from reading the SDK source — is to call `addToolOutput` from inside (or after) the callback. The JSDoc doesn't mention this anywhere.                                                                                                                                                                                                          
                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                               
  ### Where the mismatch lives
                                                                                                                                                                                                                                                                                                                                                                                                                               
  **Type signature** — `packages/ai/src/ui/chat.ts:153-156`:                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                               
  ```ts
  export type ChatOnToolCallCallback<UI_MESSAGE extends UIMessage = UIMessage> =                                                                                                                                                                                                                                                                                                                                               
    (options: {                                                                                                                                                                                                                                                                                                                                                                                                              
      toolCall: InferUIMessageToolCall<UI_MESSAGE>;                                                                                                                                                                                                                                                                                                                                                                            
    }) => void | PromiseLike<void>;                
                                                                                                                                                                                                                                                                                                                                                                                                                               
  JSDoc on the onToolCall field — packages/ai/src/ui/chat.ts:207-214:                                                                                                                                                                                                                                                                                                                                                          
   
  /**                                                                                                                                                                                                                                                                                                                                                                                                                          
   * Optional callback function that is invoked when a tool call is received.                                                                                                                                                                                                                                                                                                                                                  
   * Intended for automatic client-side tool execution.
   *                                                                                                                                                                                                                                                                                                                                                                                                                           
   * You can optionally return a result for the tool call,                                                                                                                                                                                                                                                                                                                                                                   
   * either synchronously or asynchronously.
   */                                                                                                                                                                                                                                                                                                                                                                                                                          
  onToolCall?: ChatOnToolCallCallback<UI_MESSAGE>;
                                                                                                                                                                                                                                                                                                                                                                                                                               
  The "you can optionally return a result for the tool call" sentence contradicts the => void | PromiseLike<void> return type. The call site that invokes this callback (inside AbstractChat) does not read the returned value either, so even non-strict TypeScript callers that try to return a value see no effect.                                                                                                         
   
  Reproduction                                                                                                                                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                                                                                                                             
  const { addToolOutput } = useChat({
    onToolCall: async ({ toolCall }) => {
      // Following the JSDoc — return the tool result.                                                                                                                                                                                                                                                                                                                                                                         
      const output = await executeToolLocally(toolCall);
      return output; // TS error if strict; silently dropped otherwise                                                                                                                                                                                                                                                                                                                                                         
    },                                                                                                                                                                                                                                                                                                                                                                                                                         
  });
                                                                                                                                                                                                                                                                                                                                                                                                                               
  Outcome: the tool call stays in 'input-available' state forever. sendAutomaticallyWhen: lastAssistantMessageIsCompleteWithToolCalls never re-fires the next turn (because the tool call is unresolved). The chat hangs with no error UI.                                                                                                                                                                                     
   
  Expected behaviour (per JSDoc)                                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                             
  Returning a value from onToolCall should attach it as the tool output and let the chat loop continue.                                                                                                                                                                                                                                                                                                                        
   
  Actual behaviour                                                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                             
  The return value is dropped. The only way to resolve a client-executed tool call is to explicitly call addToolOutput({ tool, toolCallId, state, output | errorText }) inside (or after) the callback.                                                                                                                                                                                                                        
   
  Suggested fixes                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                                                                                                                             
  Either option closes the inconsistency.

  A. Update the JSDoc only (smallest change, no API impact):                                                                                                                                                                                                                                                                                                                                                                   
   
  /**                                                                                                                                                                                                                                                                                                                                                                                                                          
   * Optional callback function that is invoked when a tool call is received.                                                                                                                                                                                                                                                                                                                                                
   * Intended for automatic client-side tool execution.
   *                                                                                                                                                                                                                                                                                                                                                                                                                           
   * To attach the tool output, call `addToolOutput({ toolCallId, state, output })`
   * from inside (or after) this callback. The callback's return value is NOT                                                                                                                                                                                                                                                                                                                                                  
   * consumed by the SDK — returning a value has no effect.                                                                                                                                                                                                                                                                                                                                                                    
   */
  onToolCall?: ChatOnToolCallCallback<UI_MESSAGE>;                                                                                                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                             
  B. Honour the return value (closer to what the JSDoc currently promises):                                                                                                                                                                                                                                                                                                                                                    
   
  Change ChatOnToolCallCallback's return type to void | OUTPUT | PromiseLike<void | OUTPUT> and have the SDK auto-call addToolOutput when a non-void value is returned. This is what the JSDoc implies today and what most users intuitively expect.                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                                                                             
  Related (different angle, ergonomics rather than docs): #7544 ("Pass addToolResult as argument to onToolCall") proposes passing addToolResult into the callback options. Option B above and #7544 are complementary — both improve the API; this issue is narrower and targets only the docs/type mismatch.                                                                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                                            

### AI SDK Version

- ai: 6.0.174                                                                                                                                                                                                                                                                                                                                                                                                                
- @ai-sdk/react: 3.0.176                                                                                                                                                                                                                                                                                                                                                                                                   
- TypeScript: 5.x
- React: 18.x

### Code of Conduct

zxuhan:
## Summary

Fixes #15268.

The JSDoc on `ChatInit.onToolCall` says the callback "can optionally return a result for the tool call, either synchronously or asynchronously." That's not true:

- The type signature is `(options) => void | PromiseLike<void>` (`packages/ai/src/ui/chat.ts:153-156`).
- The call site awaits the callback but discards its return value (`packages/ai/src/ui/process-ui-message-stream.ts:655-659`).
- The v5 migration guide already documented the removal: "`onToolCall` no longer supports returning values to automatically submit tool results."

Tool results have to be attached with `addToolOutput`. This PR updates the JSDoc to say that, and adds the `sendAutomaticallyWhen` deadlock warning that's already in the reference docs and v5 migration guide, so the hover tooltip matches what the published docs tell people to do.

## Test plan

- [x] `pnpm --filter ai test:node`: 2690 tests pass
- [x] `pnpm check`: lint and format clean
- [x] `pnpm --filter ai type-check`

### DIFF
--- a/.changeset/ontoolcall-jsdoc-return-value.md
+++ b/.changeset/ontoolcall-jsdoc-return-value.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+docs(ai): correct `onToolCall` JSDoc to reflect that the return value is ignored and tool output must be attached via `addToolOutput`
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -208,8 +208,12 @@ export interface ChatInit<UI_MESSAGE extends UIMessage> {
    * Optional callback function that is invoked when a tool call is received.
    * Intended for automatic client-side tool execution.
    *
-   * You can optionally return a result for the tool call,
-   * either synchronously or asynchronously.
+   * To attach the tool result, call `addToolOutput({ tool, toolCallId, output })`
+   * from inside (or after) this callback. The callback's return value is
+   * ignored.
+   *
+   * When using `sendAutomaticallyWhen`, do not `await` `addToolOutput` inside
+   * this callback as it can cause deadlocks.
    */
   onToolCall?: ChatOnToolCallCallback<UI_MESSAGE>;
 


aayush-kapoor:
Marking as closed in favor of https://github.com/vercel/ai/pull/17260

PR #15288, opened 2026-05-14
fix(amazon-bedrock): disable native structured output for claude-opus-4-7

### LINKED ISSUE #14773: Bedrock rejects output_config.format on claude-opus-4-7 when SDK uses native structured output
## Description

When calling Anthropic's `claude-opus-4-7` on Amazon Bedrock with structured output (`schema:` / `Output.object`) and reasoning enabled, the request fails with:

```
{"type":"error","error":{"type":"invalid_request_error","message":"output_config.format: Extra inputs are not permitted"}}
```

This blocks any structured-output use of Opus 4.7 on Bedrock, including reasoning runs.

Related (predecessor) issue: #12298 — that one was about the deprecated `output_format` parameter; the fix flipped `supportsNativeStructuredOutput: true` and renamed to `output_config.format`. The current issue is that the new field is itself rejected by Bedrock for `claude-opus-4-7` specifically.

## Versions

- `ai`: 7.0.0-beta.111
- `@ai-sdk/amazon-bedrock`: 5.0.0-beta.41
- `@ai-sdk/anthropic`: 4.0.0-beta.37 (transitively)

## Reproduction

```ts
import { bedrock } from "@ai-sdk/amazon-bedrock"
import { generateText, Output } from "ai"
import { z } from "zod"

const schema = z.object({
  reasoning: z.string().min(1),
  fileText: z.string().min(1),
})

await generateText({
  model: bedrock("anthropic.claude-opus-4-7"),
  system: "You are a helpful assistant.",
  prompt: "Return reasoning='hello' and fileText='world' per the schema.",
  output: Output.object({ schema }),
  reasoning: "xhigh",
  maxOutputTokens: 8000,
})
```

Tested with `eu.anthropic.claude-opus-4-7` (EU cross-region inference profile) — same rejection. The same code with `anthropic.claude-opus-4-6-v1` works correctly.

## Root cause analysis

In `@ai-sdk/amazon-bedrock@5.0.0-beta.41`, the Bedrock-Anthropic provider hardcodes `supportsNativeStructuredOutput: true` for every Anthropic-on-Bedrock model:

```js
// packages/amazon-bedrock/src/anthropic/index.ts
{
  // native structured output via output_config.format is supported on Bedrock
  supportsNativeStructuredOutput: true
}
```

Combined with `@ai-sdk/anthropic@4.0.0-beta.37`'s capability table for `claude-opus-4-7`:

```js
if (modelId.includes("claude-opus-4-7")) {
  return {
    maxOutputTokens: 128e3,
    supportsStructuredOutput: true,
    supportsAdaptiveThinking: true,
    rejectsSamplingParameters: true,
    supportsXhighEffort: true,
    isKnownModel: true,
  };
}
```

`useNativeStructuredOutput` resolves to `true`, and the Bedrock chat language model injects:

```js
bedrockOptions.additionalModelRequestFields = {
  ...bedrockOptions.additionalModelRequestFields,
  output_config: {
    ...output_config,
    format: { type: "json_schema", schema: responseFormat.schema }
  }
}
```

…which the Bedrock-via-Anthropic API for `claude-opus-4-7` rejects.

## Suggested fix

Either:

1. Treat `supportsNativeStructuredOutput` as model-specific in the Bedrock-Anthropic provider, mirroring the Anthropic SDK's per-model capability table, and set it to `false` for `claude-opus-4-7` until the API accepts the field.
2. Detect the rejection and fall back to the `jsonResponseTool` path (which the SDK already builds when `useNativeStructuredOutput === false`).

## Workaround

Switch to `anthropic.claude-opus-4-6-v1` — its capability entry in the Anthropic SDK has the same `supportsStructuredOutput: true` and `supportsAdaptiveThinking: true`, and the Bedrock API accepts `output_config.format` for it.

zxuhan:
## Summary

Fixes #14773.

Bedrock rejects `output_config.format` for `claude-opus-4-7` (including the `us.`/`eu.` cross-region inference profiles) with:

```
output_config.format: Extra inputs are not permitted
```

`packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.ts` previously hardcoded `supportsNativeStructuredOutput: true` for every Bedrock-Anthropic model. The Anthropic capability table already lists opus-4-7 as supporting structured output, so the SDK took the native path that Bedrock then refused, and any `output` / `Output.object(...)` call against opus-4-7 (including reasoning runs) failed before the model ran.

This changes the flag to `!modelId.includes('claude-opus-4-7')`. Other Bedrock-Anthropic models stay on the native path; opus-4-7 falls back to the `jsonResponseTool` path, which Bedrock accepts. Using `includes()` covers the direct `anthropic.claude-opus-4-7` id and both cross-region inference profiles (`us.anthropic.claude-opus-4-7`, `eu.anthropic.claude-opus-4-7`).

## Test plan

- [x] Added a parameterized test for the direct id and both cross-region prefixes. The existing default-model assertion still requires `supportsNativeStructuredOutput: true` for non-opus-4-7 models.
- [x] `pnpm --filter @ai-sdk/amazon-bedrock test`: 369/369 pass (node + edge)
- [x] `pnpm check`: lint and format clean
- [x] `pnpm --filter @ai-sdk/amazon-bedrock type-check`

### DIFF
--- a/.changeset/bedrock-anthropic-opus-4-7-native-structured-output.md
+++ b/.changeset/bedrock-anthropic-opus-4-7-native-structured-output.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
+fix(amazon-bedrock): disable native structured output for `claude-opus-4-7` on Bedrock Anthropic. Bedrock rejects `output_config.format` for this model (including the `us.`/`eu.` cross-region inference profiles) with `Extra inputs are not permitted`; the SDK now falls back to the `jsonResponseTool` path for it.
--- a/.changeset/bedrock-anthropic-opus-4-7-native-structured-output.md
+++ b/.changeset/bedrock-anthropic-opus-4-7-native-structured-output.md
@@ -1,5 +1,0 @@
----
-'@ai-sdk/amazon-bedrock': patch
----
-
-fix(amazon-bedrock): disable native structured output for `claude-opus-4-7` on Bedrock Anthropic. Bedrock rejects `output_config.format` for this model (including the `us.`/`eu.` cross-region inference profiles) with `Extra inputs are not permitted`; the SDK now falls back to the `jsonResponseTool` path for it.
--- a/.changeset/mighty-lemons-return.md
+++ b/.changeset/mighty-lemons-return.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix(amazon-bedrock): disable native structured output for claude-opus-4-7
--- a/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.test.ts
+++ b/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.test.ts
@@ -83,6 +83,29 @@ describe('amazon-bedrock-anthropic-provider', () => {
     );
   });
 
+  it.each([
+    'anthropic.claude-opus-4-7',
+    'us.anthropic.claude-opus-4-7',
+    'eu.anthropic.claude-opus-4-7',
+  ])(
+    'should disable native structured output for %s (Bedrock rejects output_config.format)',
+    modelId => {
+      const provider = createAmazonBedrockAnthropic({
+        region: 'us-east-1',
+        accessKeyId: 'test-key',
+        secretAccessKey: 'test-secret',
+      });
+      provider(modelId as Parameters<typeof provider>[0]);
+
+      expect(AnthropicLanguageModel).toHaveBeenCalledWith(
+        modelId,
+        expect.objectContaining({
+          supportsNativeStructuredOutput: false,
+        }),
+      );
+    },
+  );
+
   it('should throw an error when using new keyword', () => {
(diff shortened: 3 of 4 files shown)


aayush-kapoor (approved):

aayush-kapoor:
verified that the fix works for the following example 

```ts
import { bedrockAnthropic } from '@ai-sdk/amazon-bedrock/anthropic';
import 'dotenv/config';
import { z } from 'zod';

run(async () => {
    model: bedrockAnthropic('global.anthropic.claude-opus-4-7'),
    output: Output.object({
      schema: z.object({
        recipe: z.object({
          name: z.string(),
          ingredients: z.array(
            z.object({
              name: z.string(),
              amount: z.string(),
            }),
          ),
          steps: z.array(z.string()),
        }),
      }),
    }),
    prompt: 'Generate a lasagna recipe.',
  });

  console.log('Recipe:', JSON.stringify(result.output, null, 2));
  console.log();
  console.log('Finish reason:', result.finishReason);
  console.log('Usage:', result.usage);
});
```

thanks!

aayush-kapoor (approved):

PR #15297, opened 2026-05-14
feat(openai): add opt-in pass-through for unsupported file media types

### LINKED ISSUE #14747: Allow overriding hardcoded file media type filtering

aayush-kapoor:
## Background

#14747

AI SDK currently applies a hardcoded allowlist when converting user file parts for some providers (in this case OpenAI) even though OpenAI can accept other file inputs generically. Currently, unsupported non-image inline files are rejected or dropped unless they are PDFs.

## Summary

- adds an opt-in `passThroughUnsupportedFiles` provider option for OpenAI file-part conversion

## Manual Verification

verified by running the example `examples/ai-functions/src/generate-text/openai/responses-unsupported-file-type.ts`

## Checklist

## Related Issues

fixes #14747 

Co-authored-by: priyanshu <priyanshukumar@priyanshus-MacBook-Air-4.local>

### DIFF
--- a/.changeset/smart-wasps-develop.md
+++ b/.changeset/smart-wasps-develop.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/openai": patch
+---
+
+feat(openai): add opt-in pass-through for unsupported file media types
--- a/content/providers/01-ai-sdk-providers/03-openai.mdx
+++ b/content/providers/01-ai-sdk-providers/03-openai.mdx
@@ -165,6 +165,10 @@ The following provider options are available:
 
   Whether to store the generation. Defaults to `true`.
 
+- **passThroughUnsupportedFiles** _boolean_
+
+  Whether to pass through non-image file types as generic input files. Defaults to `false`, which restricts inline file inputs to images and PDFs. Enable this when the target OpenAI Responses model supports additional file media types.
+
 - **maxToolCalls** _integer_
   The maximum number of total calls to built-in tools that can be processed in a response.
   This maximum number applies across all built-in tool calls, not per individual tool.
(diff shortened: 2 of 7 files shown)


aayush-kapoor:
see https://github.com/vercel/ai/issues/14747#issuecomment-4455300377 - fix for open-responses isn't needed

lgrammel (changes requested):
The openai responses provider is strictly for openai. are there use cases for models from openai or this this for non-openai providers?

lgrammel (approved):

PR #15324, opened 2026-05-15
fix(ai): default missing embedding warnings

### LINKED ISSUE #14926: embed() and embedMany() crash with EmbeddingModelV2 providers that omit `warnings`

Kage18:
## Summary

Fixes #14926.

EmbeddingModelV2 providers can return doEmbed results without a warnings field. embed() and embedMany() currently forward that missing value into result/logging paths that expect an array, which can crash at runtime.

This defaults missing embedding warnings to [] at the embed boundary for:

- embed()
- embedMany() single-call path
- embedMany() chunked path

## Tests

- pnpm --dir packages/ai exec vitest --config vitest.node.config.js --run src/embed/embed.test.ts src/embed/embed-many.test.ts
- pnpm --dir packages/ai exec vitest --config vitest.edge.config.js --run src/embed/embed.test.ts src/embed/embed-many.test.ts
- pnpm --filter ai type-check

### DIFF
--- a/.changeset/silent-cameras-embed.md
+++ b/.changeset/silent-cameras-embed.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): default missing embedding warnings to an empty array
--- a/packages/ai/src/embed/embed-many.test.ts
+++ b/packages/ai/src/embed/embed-many.test.ts
@@ -10,6 +10,7 @@ import {
   vitest,
 } from 'vitest';
 import * as logWarningsModule from '../logger/log-warnings';
+import { MockEmbeddingModelV2 } from '../test/mock-embedding-model-v2';
 import { MockEmbeddingModelV4 } from '../test/mock-embedding-model-v4';
 import type { Embedding, EmbeddingModelUsage, Warning } from '../types';
 import { createResolvablePromise } from '../util/create-resolvable-promise';
@@ -423,6 +424,21 @@ describe('result.warnings', () => {
     expect(result.warnings).toStrictEqual(expectedWarnings);
   });
 
+  it('should default missing v2 provider warnings to an empty array in the single call path', async () => {
+    const result = await embedMany({
+      model: new MockEmbeddingModelV2<string>({
+        maxEmbeddingsPerCall: null,
+        doEmbed: async () => ({
+          embeddings: dummyEmbeddings,
+          usage: { tokens: 3 },
+        }),
+      }),
+      values: testValues,
+    });
+
+    expect(result.warnings).toStrictEqual([]);
+  });
+
   it('should aggregate warnings from multiple calls', async () => {
     const warning1: Warning = {
       type: 'other',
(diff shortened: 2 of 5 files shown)


Kage18:
Kage usage for this PR:

This PR fixed #14926 by defaulting missing `EmbeddingModelV2` warning arrays to `[]` at the public `embed()` / `embedMany()` boundary.

How Kage helped:
- Scoped recall to `packages/ai` embedding entry points instead of the full repo.
- Kept the relevant code and tests together: `src/embed/embed.ts`, `src/embed/embed-many.ts`, `src/embed/embed.test.ts`, and `src/embed/embed-many.test.ts`.
- Surfaced that there are multiple paths to cover: `embed()`, `embedMany()` single-call behavior, and chunked `embedMany()` behavior.
- Captured the reusable rule: providers may omit `warnings`, but AI SDK result/logging paths should see an array.
- Captured the verification commands for node/edge embedding tests and package type-checking.

Why this matters for future sessions:
A future agent changing embedding result handling can recall that warnings are normalized at the SDK boundary, not assumed to be present from every provider implementation.

Full contribution summary and Kage metrics: https://github.com/kage-core/Kage/pull/8

aayush-kapoor (approved):

PR #15345, opened 2026-05-15
feat(provider-utils): add `readFile` and `writeFile` plus convenience wrappers to `Experimental_Sandbox` abstraction

felixarntz:
## Background

`Experimental_Sandbox` (previous: #14949, #15253, #15301) only exposed `description` and `runCommand`, so tools that needed file I/O had to wrap every read/write in a shell command (`cat`, `tee`, `echo > …`). That is fragile for binary content and impossible to type properly.

## Summary

- Adds six file methods to `Experimental_Sandbox`: streaming `readFile`/`writeFile` as the foundation, plus `readBinaryFile`/`readTextFile` and `writeBinaryFile`/`writeTextFile` as convenience wrappers.
  - All methods take a single options object so additional fields can be added without breaking the signature.
  - Using a stream for the foundation is the most low-level and future-proof primitive, plus it handles large files better.
  - In a follow up PR, we'll add another reduced abstraction surface, because technically a sandbox provider shouldn't have to implement the convenience wrappers `readTextFile`, `writeTextFile`, etc. This can take inspiration from (or continue with) #15311.
- Updates the three example sandbox implementations (`LocalSandbox`, `JustBashSandbox`, both `VercelSandbox` copies) to implement the new methods, using streaming `readFile`/`writeFile` as the foundation that the binary and text variants delegate to.

## Checklist

## Future Work

See above: We'll need to add that reduced abstraction surface so that only essential methods _have_ to be implemented by the provider. We can provide a helper function to fill in the convenience wrappers automatically.


packages/provider-utils/src/types/sandbox.ts
```diff
@@ -43,4 +43,147 @@ export type Experimental_Sandbox = {
...
+
+    /**
+     * Text encoding used to decode the file bytes. Defaults to `"utf-8"`.
+     */
+    encoding?: string;
+
+    /**
+     * 1-based inclusive start line. Defaults to 1.
+     */
+    startLine?: number;
```
lgrammel, line 104: what do we consider a linebreak? presumably `\n` but we may want an optional `linebreak` field that allows setting `\r` or `\n` or `\r\n`

packages/ai/src/test/mock-sandbox.ts
```diff
@@ -0,0 +1,22 @@
+import type { Experimental_Sandbox } from '@ai-sdk/provider-utils';
+
+/**
+ * No-op stubs for the file methods on `Experimental_Sandbox`, intended to be
+ * spread into test fixtures that only care about `runCommand`.
+ */
+export const mockSandboxFileStubs: Pick<
```
lgrammel, line 7: wonder if it's worth moving to `provider-utils` so providers can use this for testing

lgrammel (approved):

packages/provider-utils/src/types/sandbox.ts
felixarntz: Good point, though I think an option feels a bit too low-level. I implemented a helper in `provider-utils` instead for extracting lines which detects the line break format used in a file and maintains it.

If this shows to be problematic, we can still considering adding an option.

packages/ai/src/test/mock-sandbox.ts
felixarntz: Unsure about this. Do we have other test-only helpers in `provider-utils`? Either way, we could add this later if there's value.

PR #15378, opened 2026-05-17
fix(xai, deepseek, groq, mistral): only send null content for assistant messages with tool calls

### LINKED ISSUE #14612: Bug: tool-only assistant messages send content:"" in xai, deepseek, groq, mistral providers

naumanAhmed3:
## Background

#13744 introduced `content: text || null` for assistant messages so that tool-only assistant turns (no text part) send `content: null` to the upstream API. #14950 refined this further for `@ai-sdk/openai` and `@ai-sdk/openai-compatible` to **only** apply the null when `tool_calls.length > 0` — preventing a regression where empty assistant content without tool calls was being incorrectly converted to `null`, which OpenAI's spec rejects.

Issue #14612 points out the same `content: ""` pattern still exists in four other OpenAI-compatible providers and causes Bedrock-routed gateways (and strict spec validators) to reject the request with:

> `ValidationException: messages: text content blocks must be non-empty`

## Summary of changes

Apply the same pattern as #14950 in four providers:

```diff
- content: text,
+ content: toolCalls.length > 0 ? text || null : text,
```

Files changed:
- `packages/xai/src/convert-to-xai-chat-messages.ts`
- `packages/deepseek/src/chat/convert-to-deepseek-chat-messages.ts`
- `packages/groq/src/convert-to-groq-chat-messages.ts`
- `packages/mistral/src/convert-to-mistral-chat-messages.ts`

For each provider, the existing test snapshots that asserted `content: ''` for tool-only assistant messages are updated to expect `content: null`. A new regression test is added in each test file to assert that an assistant message with empty text and **no** tool calls still produces `content: ''` (mirroring the regression #14950 was guarding against).

A patch changeset covers all four packages.

## Verification

- Updated all existing tool-only test snapshots in the 4 packages (13 assertions across 4 files).
- Added 4 new regression tests for the empty-text-no-tool-calls case.
- All edits follow the exact pattern of #14950's reference fix.

I do **not** have a local Node toolchain to run the suite myself; relying on the project's CI to confirm the snapshots match. Happy to push fixups if anything turns red.

## Mistral note

The issue flags that Mistral's chat-completions spec should be verified to accept `null` content. Mistral's docs describe `content` as either a string or an array of content parts; their OpenAI-compatible chat endpoint accepts the OpenAI spec where `content` is nullable when `tool_calls` is present. If maintainers prefer to scope this to xai/deepseek/groq only and address mistral separately, I'll happily split the PR.

Closes #14612. Follow-up to #14950, #13744.

### DIFF
--- a/packages/mistral/src/convert-to-mistral-chat-messages.ts
+++ b/packages/mistral/src/convert-to-mistral-chat-messages.ts
@@ -138,7 +138,7 @@ export function convertToMistralChatMessages(
 
         messages.push({
           role: 'assistant',
-          content: text,
+          content: toolCalls.length > 0 ? text || null : text,
           prefix: isLastMessage ? true : undefined,
           tool_calls: toolCalls.length > 0 ? toolCalls : undefined,
         });
--- a/packages/xai/src/convert-to-xai-chat-messages.ts
+++ b/packages/xai/src/convert-to-xai-chat-messages.ts
@@ -113,7 +113,7 @@ export function convertToXaiChatMessages(prompt: LanguageModelV4Prompt): {
 
         messages.push({
           role: 'assistant',
-          content: text,
+          content: toolCalls.length > 0 ? text || null : text,
           tool_calls: toolCalls.length > 0 ? toolCalls : undefined,
         });
 
--- a/.changeset/null-content-tool-only-providers.md
+++ b/.changeset/null-content-tool-only-providers.md
@@ -1,0 +1,12 @@
+---
+"@ai-sdk/xai": patch
+"@ai-sdk/deepseek": patch
+"@ai-sdk/groq": patch
+"@ai-sdk/mistral": patch
+---
+
+fix(xai, deepseek, groq, mistral): only send null content for assistant messages with tool calls
+
+Apply the same fix as #14950 (openai, openai-compatible) to four additional OpenAI-compatible providers. When an assistant message contains only `tool-call` parts (no text), emit `content: null` so providers that route to Bedrock or strictly follow the OpenAI spec stop rejecting the request with `ValidationException: messages: text content blocks must be non-empty`. When the message has no tool calls, preserve the original string (including empty string) so we do not regress the OpenAI-spec rule "content is required unless tool_calls or function_call is specified".
+
+Closes #14612.
--- a/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
+++ b/packages/deepseek/src/chat/convert-to-deepseek-chat-messages.test.ts
@@ -131,7 +131,7 @@ describe('convertToDeepSeekChatMessages', () => {
         {
           "messages": [
             {
-              "content": "",
+              "content": null,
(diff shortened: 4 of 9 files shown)


packages/mistral/src/convert-to-mistral-chat-messages.ts
aayush-kapoor, line 1: this will need a change in `mistral-chat-prompt.ts` to mark content to be nullable. do you have the documentation link on mistral's site that says it doesn't accept content as nullable?

if so, please add a comment in the original issue and link the docs and then remove the changes for mistral from this PR

aayush-kapoor (commented):
you will need to sign your commits please for me to able to merge

packages/mistral/src/convert-to-mistral-chat-messages.ts
aayush-kapoor: Screenshot 2026-05-18 at 1 42 07 PM

it actually does allow null content in the assistant message. please make the change accordingly

packages/xai/src/convert-to-xai-chat-messages.ts
```diff
@@ -113,7 +113,7 @@ export function convertToXaiChatMessages(prompt: LanguageModelV4Prompt): {
 
         messages.push({
           role: 'assistant',
-          content: text,
+          content: toolCalls.length > 0 ? text || null : text,
```
aayush-kapoor, line 116: this will also need a change in `xai-chat-prompt.ts` to mark content to be null. please add it

aayush-kapoor:
the change was made in https://github.com/vercel/ai/pull/14987 so will be marking this as closed since it's a duplicate and that PR was created before

PR #15381, opened 2026-05-17
feat(ai): add createFallbackLanguageModel

### LINKED ISSUE #2636: feat: ability to define fallback model / provider

naumanAhmed3:
## What

Adds `createFallbackLanguageModel({ models, shouldRetry?, modelId?, providerId? })` — a chain-of-models wrapper that falls back on errors. Closes #2636 (64 👍, the most-thumbed open feature request in the issue tracker as of writing).

```ts
import { createFallbackLanguageModel, generateText } from 'ai';

const resilient = createFallbackLanguageModel({
  models: [openai('gpt-5-mini'), anthropic('claude-3-5-haiku-latest')],
  shouldRetry: error => /rate|overloaded|503/.test(String(error)),
});

await generateText({ model: resilient, prompt: '...' });
```

## Design summary

- Returns a `LanguageModelV4`. `doGenerate` / `doStream` try the underlying models in order.
- On error, an optional `shouldRetry(error, modelIndex)` predicate decides whether to try the next model. Default is "always retry" so the most common ask (just give me a backup model) is one line.
- If all models fail, the **last** error is re-thrown (matches how the existing retry helper inside the SDK propagates the final error).
- `supportedUrls` is read from the first model. Mixed-provider chains with different URL support inherit the conservative set; documented in the JSDoc.
- `provider` / `modelId` default to `'fallback'` and a synthetic identifier like `fallback(openai:gpt-5-mini,anthropic:claude-3-5-haiku-latest)`. Override via `providerId` / `modelId` options.

## Streaming semantics (intentional, called out explicitly)

Fallback is only attempted while the underlying `doStream` **promise** is still pending — i.e. pre-flight errors (auth, 429 before any bytes, network refused). Once the stream has started emitting parts, mid-stream errors flow through the stream untouched. Falling back mid-stream would force callers to stitch partial output from one provider onto output from another, which we don't want this helper to do silently.

This is consistent with how `streamText` itself surfaces stream-time errors today.

## Why not extend `customProvider`'s `fallbackProvider` instead

`customProvider.fallbackProvider` is documented as "used when a requested model ID is not found in the custom provider" — it's a registry-lookup fallback, not a runtime-error fallback. The issue tracker conversation on #2636 explicitly calls this out (see the @elie222 comments on 2025-04-17 and 2025-05-16). The two features are complementary.

## Tests

`create-fallback-language-model.test.ts` covers:
- configuration (empty models throws, synthetic id, id/provider overrides)
- `doGenerate`: primary success, fall through on error, last-error rethrow when all fail, `shouldRetry: () => false` short-circuit, model-index passed to `shouldRetry`
- `doStream`: same suite for streams
- `supportedUrls` inherited from the first model

## Notes for reviewers

- Happy to split this into a draft + design discussion if maintainers prefer to converge on the API surface first (e.g. should `shouldRetry` be sync-only? should we accept a per-model `timeout`? should we expose an `onFallback` callback for observability?). I went with the minimum surface that answers the literal ask in the issue, and left obvious extension points (`shouldRetry` could be widened, callbacks could be added) for follow-ups.
- I don't have a local Node toolchain to run vitest myself; relying on CI to confirm the suite passes. Happy to push fixups.

Closes #2636. Follow-up candidates: an `onFallback` observability hook, embedding/image/speech equivalents.


naumanAhmed3:
The latest commit changes the changeset bump to `patch`, and the Vercel review bot marked that finding resolved. The PR is currently waiting for maintainer review; please let me know if any additional changes are required.

PR #15385, opened 2026-05-18
fix(ai): parse generateText output for text-only final steps when finishReason isn't `stop`

### LINKED ISSUE #15349: [@ai-sdk/cerebras] Cerebras does not work with tools + structured output
### Description

There are two issues with cerebras right now that make it unusable:
First one is captured here: https://github.com/vercel/ai/issues/15042

The second one is when you use both tools and structured output. When the agent tries to output the final result as a structured output, its finishReason is `tool-call` and not `stop`. This raises an error when you try to actually use the output.

I made a minimal reproduction script:

```
import { createCerebras } from "@ai-sdk/cerebras";
import { NoOutputGeneratedError, Output, stepCountIs, tool, ToolLoopAgent } from "ai";
import z from "zod";
import { createCerebrasCompatibleFetch } from "./cerebras-compatible-fetch";
import { cerebras } from "@ai-sdk/cerebras";

async function main() {
    instructions:
      "Use the required structured output. For simple arithmetic you may answer directly.",
    // model: cerebras("zai-glm-4.7"),
    model: createCerebras({fetch: createCerebrasCompatibleFetch()})("zai-glm-4.7"), // uses patch for the first issue
    tools: {
      nonUsefulTool: tool({
        description: "A non-useful tool that returns a random number",
        inputSchema: z.object({}),
        execute: () => {
          return 2026;
        },
      }),
    },
    output: Output.object({
      schema: z.object({ result: z.string() }),
    }),
    stopWhen: stepCountIs(5),
  });
  const result = await agent.generate({
    prompt: "what is the magic number? The magic number is 2026",
  });
  try {
    console.log(result.output);
  } catch (cause) {
    if (NoOutputGeneratedError.isInstance(cause)) {
      console.error(
        "Structured output missing (last step did not yield valid JSON).",
        "finishReason:",
        result.finishReason,
        "lastText:",
        JSON.stringify(result.text),
      );
      throw cause;
    }
    throw cause;
  }
}

main();
```

Output:
```
tructured output missing (last step did not yield valid JSON). finishReason: tool-calls lastText: "{\"result\":\"2026\"}"
4792 |   get experimental_output() {
4793 |     return this.output;
4794 |   }
4795 |   get output() {
4796 |     if (this._output == null) {
4797 |       throw new NoOutputGeneratedError();
                   ^
AI_NoOutputGeneratedError: No output generated.
      cause: undefined,
 vercel.ai.error: true,
 vercel.ai.error.AI_NoOutputGeneratedError: true,

      at output (/Users/kylediaz/T/kylediaz/aisdk/node_modules/ai/dist/index.mjs:4797:13)
      at main (/Users/kylediaz/T/kylediaz/aisdk/index.ts:32:17)
```

You can see that the agent output the final answer, but it is being interpreted as a tool call which results in a `NoOutputGeneratedError`.

It seems this occurs with glm but not qwen. I haven't bothered testing other models.

The workaround is to use `result.text` and parse the result yourself.

### AI SDK Version

"@ai-sdk/cerebras": "^2.0.51",
"ai": "^6.0.176",
"zod": "^4.4.3"

### Code of Conduct

zxuhan:
## Summary

Fixes #15349.

`generateText` only parsed `Output.object` / `Output.array` / `Output.choice` / `Output.json` / `Output.text` when the last step's `finishReason === 'stop'`:

```ts
// packages/ai/src/generate-text/generate-text.ts
if (lastStep.finishReason === 'stop') {
  resolvedOutput = await outputSpecification.parseCompleteOutput(...);
}
```

Some providers report `tool-calls` as the finish reason for text-only final steps. Cerebras `zai-glm-4.7` (and other GLM models on Cerebras) is the most visible case: after the agent has already executed tools, the model emits the structured JSON output as text, but Cerebras still returns `finish_reason: "tool_calls"` even though no tool call is in the response. With the old check, that valid output gets buried behind:

```
AI_NoOutputGeneratedError: No output generated.
  finishReason: tool-calls
  lastText: "{\"result\":\"2026\"}"
```

The fix relaxes the condition to also accept the case where the last step produced text and contains no tool calls, meaning the agent loop has already exited naturally because there was nothing left to execute. Steps that actually contain tool calls (the agent loop should have continued but a stop condition fired first) still throw `NoOutputGeneratedError`, exactly as today.

This also aligns `generateText` with `streamText`, whose `output` getter already always tries to parse the final-step text without a `finishReason` guard.

## Test plan

- [x] Added regression test in `generate-text.test.ts` covering `Output.object` with a text-only final step and `finishReason: 'tool-calls'`. Asserts the parsed object is returned and `toolCalls` is empty.
- [x] Existing test "should not parse output when finish reason is tool-calls" (where an actual tool call is present) still passes unchanged.
- [x] `pnpm --filter ai test`: 2717/2717 pass on both node and edge runtimes.
- [x] `pnpm check`: lint and format clean.
- [x] `pnpm --filter ai type-check`: clean.

### DIFF
--- a/.changeset/generate-text-parse-output-without-stop-finish-reason.md
+++ b/.changeset/generate-text-parse-output-without-stop-finish-reason.md
@@ -1,0 +1,9 @@
+---
+'ai': patch
+---
+
+fix(ai): parse `generateText` output when the last step is text-only, even if `finishReason` is not `stop`
+
+`generateText` previously only parsed `Output.object` / `Output.array` / `Output.choice` / `Output.json` / `Output.text` when `lastStep.finishReason === 'stop'`. Some providers report `tool-calls` as the finish reason for text-only final steps — most notably Cerebras `zai-glm-4.7` after the agent has already executed tools — which buried a perfectly parseable structured response behind `NoOutputGeneratedError`.
+
+Output is now also parsed when the last step has no tool calls and produced non-empty text. Steps that actually contain tool calls (i.e. the agent loop should have continued but a stop condition fired first) still throw `NoOutputGeneratedError` exactly as before.
--- a/packages/ai/src/generate-text/generate-text.test.ts
+++ b/packages/ai/src/generate-text/generate-text.test.ts
@@ -6928,6 +6928,27 @@ describe('generateText', () => {
       expect(result.toolCalls).toHaveLength(1);
       expect(result.toolResults).toHaveLength(1);
     });
+
+    it('should parse output when finish reason is tool-calls but the last step contained only text (e.g. Cerebras zai-glm-4.7)', async () => {
+      const result = await generateText({
+        model: new MockLanguageModelV4({
+          doGenerate: async () => ({
+            ...dummyResponseValues,
+            // Some providers (e.g. Cerebras GLM) report `tool-calls` as the
+            // finish reason even when the assistant turn carries only text.
+            finishReason: { unified: 'tool-calls', raw: undefined },
+            content: [{ type: 'text', text: `{ "value": "test-value" }` }],
+          }),
+        }),
+        prompt: 'prompt',
+        output: Output.object({
+          schema: z.object({ value: z.string() }),
+        }),
+      });
+
+      expect(result.output).toEqual({ value: 'test-value' });
+      expect(result.toolCalls).toHaveLength(0);
+    });
   });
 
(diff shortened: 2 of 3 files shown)


aayush-kapoor (commented):
marking as closed in favor of https://github.com/vercel/ai/pull/15420. fix should ideally be contained to the provider

PR #15387, opened 2026-05-18
feat(vue): add useChat composable

posva:
Adds a `useChat` composable for `@ai-sdk/vue` that wraps `Chat` with reactive refs and auto-recreates the instance when its init object changes.

## Example

```ts
import { useChat } from '@ai-sdk/vue'
import { DefaultChatTransport } from 'ai'

const { messages, sendMessage, status, error } = useChat(() => ({
  id: data.value?.id,
  messages: data.value?.messages,
    api: `/api/chats/${data.value?.id}`,
    body: { model: model.value },
  }),
  onError(err) {
    toast.add({ description: err.message, color: 'error' })
  },
}))
```

Passing a getter (or ref) makes the init reactive: when `data.value.id` or `model.value` change, the chat is recreated automatically. With `new Chat({...})` you had to manage that yourself.

It's worth noting that I made the messages array a shallowRef since it could potentially get very large but since we always know when it gets modified, it makes no sense to have deep reactivity (and poorer performance/memory)

I also marked `Chat` as deprecated in favor of this more idiomatic `useChat()`. What do you think?

### DIFF
--- a/.changeset/vue-use-chat.md
+++ b/.changeset/vue-use-chat.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/vue": patch
+---
+
+Add `useChat` composable for `@ai-sdk/vue`, providing a reactive wrapper around `Chat` with auto-recreation when the init object changes.
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -25,6 +25,12 @@
       "rules": {
         "import/no-cycle": "off"
       }
+    },
+    {
+      "files": ["packages/vue/**"],
+      "rules": {
+        "react-hooks/rules-of-hooks": "off"
+      }
     }
   ],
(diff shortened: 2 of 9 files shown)


gr2m (commented):
We used to have `useChat` and changed it to `new Chat` for AI SDK v5:
https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#ai-sdkvue-changes

It was before my time, I'm not sure why the change was done.

If we want to move forward with this change: can you please update the docs for vue.js/nuxt and the migration guides for v7

posva:
> We used to have useChat and changed it to new Chat for AI SDK v5

Thanks, I didn't see that. I will give the docs update and migration guides a shot

gr2m (approved):

PR #15421, opened 2026-05-18
refactor(ai): introduce agent UI message generic

gr2m:
## Summary
- Introduce a `UI_MESSAGE` generic in `createAgentUIStream` to avoid repeating `UIMessage<MESSAGE_METADATA, never, InferUITools<TOOLS>>` throughout the function signature and tool typing.

## Test plan

### DIFF
--- a/packages/ai/src/agent/create-agent-ui-stream.ts
+++ b/packages/ai/src/agent/create-agent-ui-stream.ts
@@ -41,6 +41,8 @@ export async function createAgentUIStream<
   RUNTIME_CONTEXT extends Context = Context,
   OUTPUT extends Output = never,
   MESSAGE_METADATA = unknown,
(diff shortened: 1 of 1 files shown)


packages/ai/src/agent/create-agent-ui-stream.ts
```diff
@@ -41,6 +41,8 @@ export async function createAgentUIStream<
   RUNTIME_CONTEXT extends Context = Context,
   OUTPUT extends Output = never,
   MESSAGE_METADATA = unknown,
+  UI_MESSAGE extends UIMessage<MESSAGE_METADATA, never, InferUITools<TOOLS>> =
```
lgrammel, line 44: if we take this approach it would be good to remove the tools and message metadata generics and instead infer them from ui message (but that would be a breaking change)

PR #15429, opened 2026-05-19
fix(vue): forward args in useChat methods

posva:
Forgot to pass the args.

Should I add `...args` to all methods, or keep it specific like the current change?

### DIFF
--- a/.changeset/vue-use-chat-forward-args.md
+++ b/.changeset/vue-use-chat-forward-args.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/vue": patch
+---
+
+fix(vue): forward args in useChat methods
--- a/packages/vue/src/use-chat.ts
+++ b/packages/vue/src/use-chat.ts
@@ -170,13 +170,13 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>(
     status,
     messages,
     error,
-    addToolApprovalResponse: (...args) =>
-      chatInstance.value.addToolApprovalResponse(...args),
-    addToolOutput: (...args) => chatInstance.value.addToolOutput(...args),
+    addToolApprovalResponse: opts =>
+      chatInstance.value.addToolApprovalResponse(opts),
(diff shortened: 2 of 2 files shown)


lgrammel (approved):

lgrammel:
Might be good to add some tests.

posva:
I tried to add a regression test but I can't figure out any test that would break in the case of new args. It's mostly due to optional parameters. I think there are two options to consider:

- Always pass `...args` but adds client side bundle for no reason
- Keep it minimal and update when this happens.

Regarding the tests I tried:

type only:

```ts
  return {
    id: computed(() => chatInstance.value.id),
    status,
    messages,
    error,
    addToolApprovalResponse: opts =>
      chatInstance.value.addToolApprovalResponse(opts),
    addToolOutput: opts => chatInstance.value.addToolOutput(opts),
    clearError: () => chatInstance.value.clearError(),
    regenerate: opts => chatInstance.value.regenerate(opts),
    sendMessage: (...args) => chatInstance.value.sendMessage(...args),
    stop: () => chatInstance.value.stop(),
    resumeStream: () => chatInstance.value.resumeStream(),
  } satisfies {
    [K in keyof UseChatHelpers<UI_MESSAGE>]: UseChatHelpers<UI_MESSAGE>[K] extends (
      ...args: infer Args
    ) => infer Ret
      ? (...args: Args) => Ret
      : UseChatHelpers<UI_MESSAGE>[K];
  };
```

```ts
    render(
      defineComponent({
        setup() {
          const chat = useChat();

          const rawChat = new VueChat<UIMessage>({
            // stub the state
            state: {},
          })
          expect(chat.addToolApprovalResponse.length).toBe(
            rawChat.addToolApprovalResponse.length,
          );
          expect(chat.resumeStream.length).toBe(rawChat.resumeStream.length);
// ....
```

I also tried `expectTypeOf(...).parameters` and that didn't work. So I'm curious if you have another solution

PR #15458, opened 2026-05-20
fix(gateway): add caching option to provider options type

LeSingh1:
## Summary

Adds the documented `caching: "auto"` Gateway provider option to `GatewayProviderOptions`.

The AI Gateway docs show `providerOptions.gateway.caching = "auto"`, but the current type does not allow `caching`, which causes strict TypeScript users to hit `ts(2353)` when using `satisfies GatewayProviderOptions`.

This change:
- adds `caching: "auto"` to the Gateway provider options schema
- adds a focused type test for the documented usage

Fixes #14595.

## Test plan

- `pnpm --filter @ai-sdk/gateway type-check`
- `pnpm --filter @ai-sdk/gateway test:node`

### DIFF
--- a/packages/gateway/src/gateway-provider-options.test-d.ts
+++ b/packages/gateway/src/gateway-provider-options.test-d.ts
@@ -1,0 +1,12 @@
+import { describe, expectTypeOf, it } from 'vitest';
+import type { GatewayProviderOptions } from './gateway-provider-options';
+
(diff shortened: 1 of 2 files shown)


n1ckoates:
Added to AI SDK 7 in https://github.com/vercel/ai/pull/15768
Backported to AI SDK 6 in https://github.com/vercel/ai/pull/16483

PR #15483, opened 2026-05-20
[tool-loop-agent] adding support for messages with system role with override

visyat:
Adding support for `role: system` messages for `ToolLoopAgent`. Both `generateText` and `streamText` allow for a `allowSystemInMessages` argument to enable system messages – updating ToolLoopAgent to accept this toggle as well. 

```
  model,
  tools,
  allowSystemInMessages: true,
});
await createAgentUIStream({
  agent,
  uiMessages: [
    { role: "system", content: "Server context" },
    { role: "user", content: "Hello" },
  ],
});
```

```
  model,
  tools,
  prepareCall: ({ messages, options }) => {
    return {
      ...options,
      messages,
      allowSystemInMessages: true, // Can be conditional per-call
    };
  },
});
```


content/docs/07-reference/01-ai-sdk-core/16-tool-loop-agent.mdx
```diff
@@ -62,6 +62,13 @@ To see `ToolLoopAgent` in action, check out [these examples](#examples).
       description:
         'Instructions for the agent, usually used for system prompt/context.',
     },
+    {
+      name: 'allowSystemInMessages',
+      type: 'boolean',
+      isOptional: true,
+      description:
+        'Whether messages with `role: "system"` are allowed in the `prompt` or `messages` fields. When false (default), system content must be provided through the `instructions` option. Set to true for trusted, server-generated `role: "system"` messages in conversation history.',
```
gr2m: can you update description to what we use in generateText/streamText
https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-text#allow-system-in-messages

We should mention the prompt injection attack risk

gr2m (approved):

PR #15492, opened 2026-05-20
fix(provider/openai): skip duplicate items when previousResponseId is set in Responses API

p3droml:
## Background

When using the Responses API with `providerOptions.openai.previousResponseId` and `store: true`, the converter emits prior assistant items as `item_reference`s in `input` while those same items are already present in the server-side chain via `previous_response_id`. OpenAI rejects this with:

```
Error [AI_APICallError]: Duplicate item found with id rs_0c44cfe6e9bf2d84006a0e434c5a70819bb8b57fbf1d10c2df.
Remove duplicate items from your input and try again.
url: https://api.openai.com/v1/responses
statusCode: 400
```

## How to reproduce

A reasoning model (`gpt-5` / `gpt-5.1` / etc.) using `streamText` with tools and `previousResponseId` advanced per step via `prepareStep` / `onStepFinish` — the documented pattern for preserving reasoning state across reasoning-model tool calls. See the [OpenAI cookbook](https://cookbook.openai.com/examples/responses_api/reasoning_items):

> If you use `previous_response_id` for multi-turn conversations, the model will automatically have access to all previously produced reasoning items.

Today, doing this with the AI SDK fails because the converter doesn't realize the assistant items are already on the server side.

## Fix

The same dedup already exists for the Conversations API via `hasConversation` (added in commit `000fa96` / PR #11899). This change introduces the symmetric `hasPreviousResponseId` flag and applies it to the same five skip sites:

1. assistant text parts with itemId
2. assistant tool-call parts with itemId
3. assistant tool-result parts (provider-executed tools)
4. assistant reasoning parts with itemId
5. assistant compaction parts with itemId

At the call site in `openai-responses-language-model.ts`:

```ts
hasConversation: openaiOptions?.conversation != null,
hasPreviousResponseId: openaiOptions?.previousResponseId != null,
```

The five existing skip conditions become `if ((hasConversation || hasPreviousResponseId) && id != null)` (or the equivalent for the unconditional tool-result skip).

## Compatibility

- Behavior is **unchanged** when neither `conversation` nor `previousResponseId` is set.
- Behavior is **unchanged** for existing `conversation` users.
- For `previousResponseId` users with `store: true`, this is the fix — they would previously have hit `Duplicate item found` 400s as soon as any chained assistant item had an itemId.

## Tests

4 new tests in `packages/openai/src/responses/convert-to-openai-responses-input.test.ts` mirror the existing `hasConversation` coverage (text, tool-call, reasoning, compaction). All **714 existing tests** in `@ai-sdk/openai` continue to pass on both node and edge environments.

## Changeset

`.changeset/openai-responses-skip-dup-items-previous-response-id.md` — `@ai-sdk/openai: patch`.

## Related

- Original `hasConversation` fix: commit `000fa96`, PR #11899
- OpenAI cookbook on reasoning items + tools: <https://cookbook.openai.com/examples/responses_api/reasoning_items>
- OpenAI conversation-state guide: <https://platform.openai.com/docs/guides/conversation-state>

### DIFF
--- a/packages/openai/src/responses/convert-to-openai-responses-input.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.ts
@@ -61,6 +61,7 @@ export async function convertToOpenAIResponsesInput({
   passThroughUnsupportedFiles = false,
   store,
   hasConversation = false,
+  hasPreviousResponseId = false,
   hasLocalShellTool = false,
   hasShellTool = false,
   hasApplyPatchTool = false,
@@ -75,6 +76,7 @@ export async function convertToOpenAIResponsesInput({
   passThroughUnsupportedFiles?: boolean;
   store: boolean;
   hasConversation?: boolean; // when true, skip assistant messages that already have item IDs
+  hasPreviousResponseId?: boolean; // when true, skip assistant messages that already have item IDs (chained via previous_response_id)
   hasLocalShellTool?: boolean;
   hasShellTool?: boolean;
   hasApplyPatchTool?: boolean;
@@ -228,8 +230,10 @@ export async function convertToOpenAIResponsesInput({
                 | null
                 | undefined;
 
-              // when using conversation, skip items that already exist in the conversation context to avoid "Duplicate item found" errors
-              if (hasConversation && id != null) {
+              // when the assistant items are already in OpenAI's server-side context
+              // (via conversation or previous_response_id), skip them to avoid
+              // "Duplicate item found" errors
+              if ((hasConversation || hasPreviousResponseId) && id != null) {
                 break;
               }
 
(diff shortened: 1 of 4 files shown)


packages/openai/src/responses/convert-to-openai-responses-input.ts
```diff
@@ -603,7 +610,7 @@ export async function convertToOpenAIResponsesInput({
                   part.providerOptions?.[providerOptionsName];
                 const id = providerOpts?.itemId as string | undefined;
 
-                if (hasConversation && id != null) {
+                if ((hasConversation || hasPreviousResponseId) && id != null) {
```
aayush-kapoor, line 613: thanks for the PR but in my testing i found that it doesn't need to be skipped at all the 5 call sites. only function call and reasoning items are good enough to be skipped. see https://github.com/vercel/ai/pull/15503.

will be marking this as closed in favor of that. the repro script i attached first reproduces the error, and then the fix gets rid of it

PR #15500, opened 2026-05-21
fix(@ai-sdk/gateway): add caching property to GatewayProviderOptions type

sekerderya:
## Summary

Fixes #14595

The `caching` property was documented as a valid option for `GatewayProviderOptions` but was missing from the type definition. This caused a TypeScript error when using `satisfies GatewayProviderOptions` with `caching: 'auto'`.

## Changes

- Added `caching` property to `gateway-provider-options.ts`
- Added type test for `caching` in `gateway-provider-options.test-d.ts`


n1ckoates:
Added to AI SDK 7 in https://github.com/vercel/ai/pull/15768
Backported to AI SDK 6 in https://github.com/vercel/ai/pull/16483

PR #15505, opened 2026-05-21
build: add fallback for VERSION for local dev

gr2m:


aayush-kapoor (approved):
does it need a changeset?

PR #15566, opened 2026-05-22
fix(provider/anthropic): fix remaining errors with Anthropic `code_execution` tool dynamic calls from latest `web_fetch` or `web_search`

felixarntz:
## Background

When Anthropic's `web_fetch` or `web_search` tools internally trigger code execution, the API emits `code_execution` tool calls even though the user did not register that tool. The fix from #12668 marked such calls as `dynamic: true` to bypass validation, but several code paths in the Anthropic language model still emitted unmarked calls, causing errors. The newer `code_execution_20250825` tool variant also surfaces calls under sub-names (`bash_code_execution`, `text_editor_code_execution`) that need the same treatment.

## Summary

- Mark `code_execution` tool calls as `dynamic` in all remaining emission sites for the 20250825 tool variant (non-streaming `bash_code_execution`/`text_editor_code_execution` branch, streaming first-delta and end paths, and streaming `tool-call` finalization).
- Normalize the streaming delta path to use the unified `code_execution` provider tool name and consistently backfill input with `"type": "programmatic-tool-call"` so downstream consumers see a consistent shape.
- Only set `firstDelta: true` when no input has been buffered yet, avoiding double-prefixing. This was not surfacing in a real bug so far, but it was clearly also an oversight (per `firstDelta: true` only ever being okay if there was no input yet on that same chunk).

## Manual Verification

Run the existing examples:

- `examples/ai-functions/src/stream-text/anthropic/web-fetch-20260209-pdf.ts`
- `examples/ai-functions/src/stream-text/anthropic/web-fetch-20260209-pdf.ts`

Without this PR, they will somewhere in the stream include the error about an incompatible tool being found, most times when executing them.
With this PR, this error is gone from both.

## Checklist


packages/anthropic/src/anthropic-language-model.ts
```diff
@@ -1694,19 +1707,27 @@ export class AnthropicLanguageModel implements LanguageModelV4 {
...
+                      // implicitly allows. In this scenario, we need to allow 'code_execution' tool calls even if the
+                      // tool was not explicitly provided. We therefore bypass the general validation by marking the
+                      // tool as dynamic.
                       ...(markCodeExecutionDynamic &&
                       providerToolName === 'code_execution'
                         ? { dynamic: true }
                         : {}),
-                      firstDelta: true,
-                      providerToolName: part.name,
+                      firstDelta: finalInput.length === 0,
```
aayush-kapoor: did this have to be changed? don't recall if we changed this in our call

aayush-kapoor (commented):
might be nice to add some unit tests so that this doesn't regress in the future?

packages/anthropic/src/anthropic-language-model.ts
felixarntz: no, see above. I just stumbled across it. I don't think in practice I've seen a bug with this, but it's clearly not right - if this chunk already has input, then it's not the first delta. there's another similar check somewhere else in this file in `firstDelta:` assignment, so this just looks like an oversight to me here.

aayush-kapoor: should we separate out in a new PR? following Lars' teachings :P

packages/anthropic/src/anthropic-language-model.ts
```diff
@@ -2220,12 +2245,9 @@ export class AnthropicLanguageModel implements LanguageModelV4 {
                     // the type to the delta and change the tool name.
                     if (
                       contentBlock.firstDelta &&
-                      (contentBlock.providerToolName ===
-                        'bash_code_execution' ||
-                        contentBlock.providerToolName ===
-                          'text_editor_code_execution')
+                      contentBlock.providerToolName === 'code_execution'
```
aayush-kapoor, line 2248: needed? i thought the minimal fix was just adding the `markCodeExecutionDynamic` part in doGenerate and then passing the providerToolName as is in doStream

felixarntz: this is part of the fix for streaming. streaming already had the `markCodeExecutionDynamic`, but it wasn't setting `providerToolName` correctly, so then subsequent deltas would lead to the error.

the key fix for this is https://github.com/vercel/ai/pull/15566/changes#diff-703a75596b1513c888ec47cab993d1157e1f8123b54f2bfdb171444fc2fa6f67R1719, so this change is necessary as a result, and shouldn't have any other implications.

packages/anthropic/src/anthropic-language-model.ts
felixarntz: fair, will remove

felixarntz: removed in https://github.com/vercel/ai/pull/15566/changes/a74d0b89fba9b1b5da108c6aa91fe46024033a0a

aayush-kapoor (approved):

PR #15575, opened 2026-05-23
docs(ai): fix JSDoc typo in choice output spec

aryanmotgi:
## Summary

Removes a stray article in the JSDoc for the `choice` output specification in `packages/ai/src/generate-text/output.ts`.

Before:
> When the model generates a text response, it will return a one of the choice options.

After:
> When the model generates a text response, it will return one of the choice options.

Comment-only change — no runtime behavior is affected and no changeset is needed per the contributing guidelines for trivial doc/comment fixes.


n1ckoates:
Please sign your past commits and force-push to update them. Thank you!

PR #15591, opened 2026-05-24
fix(mcp): await addClientAuthentication in token exchange and refresh

MesoX:
## Summary

`OAuthClientProvider.addClientAuthentication?(...)` is typed `void | Promise<void>` (`packages/mcp/src/tool/oauth.ts:61-66`), but the two call sites in `exchangeAuthorization` and `refreshAuthorization` invoke it without `await`. Async implementations finish writing to `params` / `headers` after the token POST has already been dispatched, so the request goes out without `client_id` / `client_secret` and the auth server rejects it (`invalid_client`).

The upstream MCP TypeScript SDK awaits the callback at `packages/client/src/client/auth.ts:1491`; this fork dropped the `await` during the OAuth refactor.

## Why this matters

Any provider that reads credentials from async storage (database, KMS, secrets manager, etc.) is silently broken on this code path. I hit this in production when a `DatabaseOAuthClientProvider.addClientAuthentication` queried Postgres before setting `client_id` - the request went out empty, the server returned 401 `Missing client_id`.

## Test plan

- [x] New unit tests in `packages/mcp/src/tool/oauth.test.ts` covering both `exchangeAuthorization` and `refreshAuthorization`. Each mocks an async `addClientAuthentication` that writes `client_id` / `client_secret` after a 5 ms delay and asserts both values are present in the form body sent to the token endpoint. Fails on `main`, passes with this patch.
- [x] Changeset added (`patch` to `@ai-sdk/mcp`).

### DIFF
--- a/.changeset/await-add-client-authentication.md
+++ b/.changeset/await-add-client-authentication.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/mcp": patch
+---
+
+Await the optional `addClientAuthentication` callback in `exchangeAuthorization` and `refreshAuthorization`. The provider interface declares the callback as `void | Promise<void>`, but the call sites invoked it without `await`, so async implementations (e.g. ones that read credentials from a database) finished setting form parameters after the token-exchange request had already been dispatched. Matches the awaited call in `modelcontextprotocol/typescript-sdk` (`packages/client/src/client/auth.ts:1491`).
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):
thank you! will add you as a co-author to my PR since we now need signed commits

PR #14854, opened 2026-04-30
feat(ai): create a diagnostics channel to push event data

aayush-kapoor:
## Background

Diagnostics channels are a way to relay event data and it works on a publisher/subscriber mechanism.

Some o11y providers had a request from us to emit the event data for telemetry through diagnostics channels instead of the callbacks

## Summary

the telemetry dispatcher now also publishes the same augmented lifecycle event to the diagnostics channel named `aisdk:telemetry`

## Manual Verification

ran the example `examples/ai-functions/src/telemetry/diagnostic-channel/generate-text.ts`

## Checklist

- [x] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #15660, opened 2026-05-28
feat(ai): use tracing channels to track parent-child context

aayush-kapoor:
## Background

we introduced diagnostic channels in https://github.com/vercel/ai/pull/14854 as a way to propagate event data.

but the problem was that there was no to establish a relation with the parent and child calls that happen when consuming information via the diagnostics channels. context propagation for subagents remained unsolved (meaning subagents will be detached from the main parent call)

## Summary

- use `tracingChannel()` instead of diagnostic `channel()`
- model and tool execution are now wrapped with `tracingChannel.tracePromise(...)` via the existing `executeLanguageModelCall` and `executeTool` dispatcher paths
- in unsupported/non-Node runtimes, tracing dynamically no-ops and directly runs the original callback/execution

## Manual Verification

verified that event data is still transmitted at each lifecycle event by running the example `examples/ai-functions/src/telemetry/tracing-channel/generate-text.ts`

## Checklist


packages/ai/src/telemetry/create-telemetry-dispatcher.test.ts
```diff
@@ -35,11 +35,24 @@ describe('createTelemetryDispatcher', () => {
     expect(telemetry.onEnd).toBeDefined();
     expect(telemetry.onAbort).toBeDefined();
     expect(telemetry.onError).toBeDefined();
-    expect(telemetry.executeLanguageModelCall).toBeUndefined();
-    expect(telemetry.executeTool).toBeUndefined();
+    expect(telemetry.executeLanguageModelCall).toBeDefined();
+    expect(telemetry.executeTool).toBeDefined();
```
aayush-kapoor, line 39: these are now defined since before `executeLanguageModelCall` and `executeTool` only existed if a user/global telemetry integration implemented those methods. but now, we have a native reason to wrap execution so that tracing channels propagate async context across the actual model/tool

packages/ai/src/telemetry/create-telemetry-dispatcher.test.ts
```diff
@@ -496,12 +514,18 @@ describe('createTelemetryDispatcher', () => {
   });
 
   describe('executeTool', () => {
-    it('returns undefined when no integrations implement executeTool', () => {
+    it('passes through when no integrations implement executeTool', async () => {
```
aayush-kapoor, line 532: changed for the reason above

packages/ai/src/telemetry/create-telemetry-dispatcher.test.ts
```diff
@@ -433,12 +446,17 @@ describe('createTelemetryDispatcher', () => {
   });
 
   describe('executeLanguageModelCall', () => {
-    it('returns undefined when no integrations implement executeLanguageModelCall', () => {
+    it('passes through when no integrations implement executeLanguageModelCall', async () => {
```
aayush-kapoor, line 465: changed for the reason above

packages/ai/src/generate-text/execute-tool-call.ts
```diff
@@ -66,6 +66,7 @@ export async function executeToolCall<TOOLS extends ToolSet>({
   executeToolInTelemetryContext?: <T>(params: {
     callId: string;
     toolCallId: string;
+    event?: ToolExecutionStartEvent<TOOLS>;
```
aayush-kapoor: this `event` is what the tracing channel uses as it's base context

packages/ai/src/telemetry/tracing-channel-publisher.ts
```diff
@@ -0,0 +1,106 @@
...
+  execute: () => PromiseLike<T>,
+): Promise<T> {
+  const diagnosticsChannel = await loadDiagnosticsChannel();
+  const tracingChannel = diagnosticsChannel?.tracingChannel?.(
+    AI_SDK_TELEMETRY_TRACING_CHANNEL,
+  );
+
+  if (tracingChannel?.hasSubscribers !== true) {
+    return await execute();
+  }
```
logaretm, line 80: Tiny caveat here, [Bun currently does not support `hasSubscribers`](https://github.com/oven-sh/bun/issues/27805) on the top-level tracing channel instance.

So we usually either invert the check:

```ts
// only if we know for a fact there are zero subscribers, then we skip
if (tracingChannel?.hasSubscribers === false) {
  return await execute();
}
```

Or some other libs in the ecosystem iterate over the bundled channels to check there instead:

```ts
function hasSubs(channel: TracingChannel) {
  if (channel.hasSubscribers === undefined) {
    return ['start', 'end', 'error', 'asyncStart', 'asyncEnd'].some(c => channel[c].hasSubscribers);
  }

  return channel.hasSubscribers;
}
```

The first pattern is more common, and if there aren't any subscribers then the overhead is not a big deal in I/O heavy libs like AI SDKs.

packages/ai/src/telemetry/tracing-channel.ts
```diff
@@ -1,6 +1,6 @@
-export const AI_SDK_TELEMETRY_DIAGNOSTIC_CHANNEL = 'aisdk:telemetry';
+export const AI_SDK_TELEMETRY_TRACING_CHANNEL = 'aisdk:telemetry';
```
logaretm: I think this is fine, but something you may want to consider:

Right now everything goes through a single `aisdk:telemetry` channel with a `type` discriminator. The `diagnostics_channel` API is designed so that many purpose-focused channels with their own subscriber sets make dispatch extremely cheap.

On the other hand, filtering a "firehose" channel adds continuous overhead on every published message (this feedback came directly from @qard, the diagnostics_channel creator on various related proposals).

Maybe something like this if you would like to consider it:

| Channel | Wraps |
  |---|---|
  | `aisdk:telemetry` | Top-level operation (`generateText`, `streamText`, `generateObject`, etc.) |
  | `aisdk:telemetry:languageModelCall` | Individual provider call (nested) |
  | `aisdk:telemetry:toolExecution` | Individual tool call (nested) |

But again it is up to you, I think it is fine either way. The idea is some APMs can choose which channels to care about and not incur overhead over those channels.

logaretm (commented):
Just wanted to drop a couple of suggestions, second one being a nit so up to you! But thanks for doing this, this sets a great example in the ecosystem on how to do telemetry like that!

content/docs/03-ai-sdk-core/60-telemetry.mdx
```diff
@@ -235,32 +235,34 @@ telemetry: {
 
 Errors inside integrations are caught and do not break the generation flow.
 
-### Diagnostic channel
+### Tracing channel
 
 In Node.js, AI SDK telemetry lifecycle events are also published to the
-`ai.telemetry` [diagnostics channel](https://nodejs.org/api/diagnostics_channel.html).
+`aisdk.telemetry` [tracing channel](https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels).
```
lgrammel: why are we changing `ai` to `aisdk`?

packages/ai/src/generate-text/stream-language-model-call.ts
```diff
@@ -337,6 +339,7 @@ export async function streamLanguageModelCall<
     request,
   } = await executeLanguageModelCallInTelemetryContext({
     callId: effectiveCallId,
+    event: languageModelCallStartEvent,
```
lgrammel: i find it strange to pass in an event. if we really want to do this, we may want to rename to `startEvent` - an alternative would be to pass the individual parts of the event. This can lead to some duplication but is semantically cleaner imo.

packages/ai/src/generate-text/stream-language-model-call.ts
```diff
@@ -337,6 +339,7 @@ export async function streamLanguageModelCall<
     request,
   } = await executeLanguageModelCallInTelemetryContext({
     callId: effectiveCallId,
```
lgrammel: call id is part of the event

packages/ai/src/telemetry/telemetry.ts
```diff
@@ -235,6 +236,7 @@ export interface Telemetry {
   executeTool?: <T>(options: {
     callId: string;
     toolCallId: string;
+    event?: InferTelemetryEvent<ToolExecutionStartEvent>;
```
lgrammel: rename? or are there other solutions?

packages/ai/src/telemetry/tracing-channel-publisher.ts
```diff
@@ -0,0 +1,106 @@
+import type * as diagnosticsChannelModule from 'node:diagnostics_channel';
```
lgrammel, line 1: how will these imports affect envs w/o node? might be worth testing e2e, e.g. in edge envs

content/docs/03-ai-sdk-core/60-telemetry.mdx
aayush-kapoor: this was a gap in docs change - the channel name here should've been `aisdk` already (since that's how we define it in `packages/ai/src/telemetry/tracing-channel.ts`)

revision needed here is that it should be `aisdk:telemetry` and not `aisdk.telemetry`

packages/ai/src/generate-text/stream-language-model-call.ts
aayush-kapoor: tracing channel depends on this event data to create the context afaik.i added this comment: https://github.com/vercel/ai/pull/15660/changes#r3314637416

packages/ai/src/telemetry/telemetry.ts
aayush-kapoor: let me explore but if not, i can rename

packages/ai/src/generate-text/stream-language-model-call.ts
aayush-kapoor: ah yes good catch - will rm

packages/ai/src/telemetry/tracing-channel-publisher.ts
aayush-kapoor: it's a similar import to what we had in the file `packages/ai/src/telemetry/diagnostic-channel-publisher.ts` but i will ensure

packages/ai/src/generate-text/stream-language-model-call.ts
aayush-kapoor: sorry actually not for context, but for the tracing subscribers to see the event data

packages/ai/src/telemetry/tracing-channel-publisher.ts
aayush-kapoor: yup, this import pattern should be fine since it's essentially a type import + added some tests in https://github.com/vercel/ai/pull/15660/changes/51c12eda5f43bba6aed4c70300950990d153d615

packages/ai/src/telemetry/telemetry.ts
aayush-kapoor: not sure if we cna rename it to `startEvent` since the event data might not always be the one at the start of the lifecycle

packages/ai/src/generate-text/stream-language-model-call.ts
aayush-kapoor: changed it to rm the duplicate call id

packages/ai/src/telemetry/tracing-channel.ts
aayush-kapoor: i guess my question would be if we do have different channels, will the context propagation still work? even if the tool execution and the model call are split?

packages/ai/src/telemetry/tracing-channel-publisher.ts
aayush-kapoor: sounds good! changed in https://github.com/vercel/ai/pull/15660/changes/24e606d7af44e77a6b5a159a178c322f5296ea7b

content/docs/03-ai-sdk-core/60-telemetry.mdx
```diff
@@ -235,32 +235,35 @@ telemetry: {
 
 Errors inside integrations are caught and do not break the generation flow.
 
-### Diagnostic channel
+### Tracing channel
 
-In Node.js, AI SDK telemetry lifecycle events are also published to the
-`ai.telemetry` [diagnostics channel](https://nodejs.org/api/diagnostics_channel.html).
+In Node.js, AI SDK telemetry lifecycle and execution events are also traced on
+the `aisdk:telemetry` [tracing channel](https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels).
```
lgrammel: rename to `ai` to align with pkg name

lgrammel:
user feedback:

````
**i would actually expect something more like:**
```
bindStart generateText
bindStart step
bindStart languageModelCall
asyncEnd languageModelCall
bindStart executeTool
asyncEnd executeTool
asyncEnd step
bindStart step
bindStart languageModelCall
asyncEnd languageModelCall
asyncEnd step
asyncEnd generateText
```
````

File: packages/ai/src/telemetry/telemetry.ts [hunk 13]
```diff
@@ -216,10 +216,12 @@ export interface Telemetry {
    * model-call span.
    *
    * @param options.callId - The call ID of the generation.
+   * @param options.event - Additional model-call start context, when available.
    * @param options.execute - The function that performs the model call.
    */
   executeLanguageModelCall?: <T>(options: {
     callId: string;
+    event?: Omit<InferTelemetryEvent<LanguageModelCallStartEvent>, 'callId'>;
```
lgrammel: events seem semantically strange here. what is needed is the content of the event as context. this can be achieved via `& InferTelemetryEvent<LanguageModelCallStartEvent>` (plus rm line 223)

and then spead in the event content `{ ...event }` when sending the actual data

packages/ai/src/telemetry/tracing-channel.ts
logaretm: Yep, it works cross-channel as long as the subscribers have their stores bound to the channels.

GraphQL illustrates this as they have like 5 channels that all work together, so APMs can choose the level of "granularity" of the spans/metrics they want to get out of it.

aayush-kapoor: we made some changes in this PR that reduce the noise / events that are emitted through the channel.

what was before something like: 
```
bindStart onStart
asyncEnd onStart
bindStart onStepStart
asyncEnd onStepStart
bindStart onLanguageModelCallStart
asyncEnd onLanguageModelCallStart
bindStart executeLanguageModelCall
asyncEnd executeLanguageModelCall
bindStart onLanguageModelCallEnd
asyncEnd onLanguageModelCallEnd
bindStart onToolExecutionStart
asyncEnd onToolExecutionStart
bindStart executeTool
asyncEnd executeTool
bindStart onToolExecutionEnd
asyncEnd onToolExecutionEnd
bindStart onStepFinish
asyncEnd onStepFinish
bindStart onStepStart
asyncEnd onStepStart
bindStart onLanguageModelCallStart
asyncEnd onLanguageModelCallStart
bindStart executeLanguageModelCall
asyncEnd executeLanguageModelCall
bindStart onLanguageModelCallEnd
asyncEnd onLanguageModelCallEnd
bindStart onStepFinish
asyncEnd onStepFinish
bindStart onEnd
asyncEnd onEnd
```

is now bundled into: 

```
bindStart generateText
bindStart step
bindStart languageModelCall
asyncEnd languageModelCall
bindStart executeTool
asyncEnd executeTool
asyncEnd step
bindStart step
bindStart languageModelCall
asyncEnd languageModelCall
asyncEnd step
asyncEnd generateText
```

and so the above events are what's emitted through the channel (still only 1 `ai:telemetry` channel)

do you think that works @logaretm?

would appreciate any feedback you have on this! otherwise if this aligns dx wise with what you're expecting, we'll try to wrap this PR up

logaretm: I think that works great! Just gave it a test round right now and looks fine DX wise

CleanShot 2026-06-04 at 17 41 33@2x

Thanks!

packages/ai/src/embed/embed.ts
```diff
@@ -135,100 +135,117 @@ export async function embed({
...
-            operationId: 'ai.embed.doEmbed',
-            provider: model.provider,
-            modelId: model.modelId,
-            values: [value],
-          },
-          callbacks: [telemetryDispatcher.onEmbedStart],
-        });
+  return await traceTelemetrySpan({
+    type: 'embed',
+    event: startEvent,
```
lgrammel, line 176: still not a fan of passing event in here

packages/ai/src/telemetry/telemetry.ts
```diff
@@ -56,6 +57,11 @@ type OperationEndEvent =
   | RerankEndEvent;
 
 export interface TelemetryDispatcher {
+  traceTelemetrySpan?: <T>(options: {
+    type: TelemetryTracingEventType;
+    event: unknown;
+    execute: () => PromiseLike<T>;
+  }) => Promise<T>;
```
lgrammel: would be good to add jsdoc. `event: unknown` can cause issues in the future imo and is not type safe. also it is very unclear what the event is and what is contained in it

lgrammel (approved):

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1415,35 +1415,48 @@ class DefaultStreamTextResult<
...
+      // Open the operation-level tracing span. Unlike generateText the work
+      // happens while the stream is consumed, so the span is kept open until
+      // the stream settles (resolved/rejected via the total usage promise).
+      telemetryDispatcher
+        .traceTelemetrySpan?.({
+          type: 'streamText',
+          event: startEvent,
+          execute: () => self._totalUsage.promise.then(() => undefined),
+        })
+        .catch(() => {});
```
sabrenner: noticed while QAing this branch that this kind of logic for streamed calls, while maintaining event order, doesn't seem to retain context and parentage properly.

for example, when logging the event type and current span in the asynclocalstorage instance from the bound store subscriber, it looks something like
```bash
generateText parent undefined
step parent generateText
languageModelCall parent step
executeTool parent step
step parent generateText
languageModelCall parent step
```
however, for streamed text calls, it looks like
```
streamText parent undefined
step parent undefined
languageModelCall parent undefined
executeTool parent languageModelCall
step parent undefined
languageModelCall parent undefined
```
likely because the parent context is restored before we get to the subsequent parts of the stream, as we publish events not nested in the bound store run.

from my experience in dealing with streams + tracing channel conventions, i've found it easiest to manually call the underlying event channels. i'm not 100% sure how it would fit in here, but you could probably have a separate telemetry handler like
```js
const ch = tracingChannel('ai:telemetry')
function traceStreamedTelemetrySpan (ctx, execute) {
  return ch.start.runStores(ctx, async () => {
    const stream = await execute()
    // as part of the stream consumption, call ch.asyncEnd.publish(ctx) manually, manually assigning ctx.result once stream is consumed & aggregated before publishing
    ch.end.publish(ctx) // synchronous end event needs to be manually published
    // also manually ch.error.publish(ctx) and assign ctx.error on any errors
    return stream
  })
}
```
so that way you could use it similarly to how you define `executeGenerateText` and then call
```js
return await traceTelemetrySpan({
    type: 'generateText',
    event: generateTextStartEvent,
    execute: executeGenerateText,
  });
```
but using `traceStreamedTelemetrySpan` instead.

as a reference, in our Datadog instrumentation we do stuff like this ([anthropic instrumentation, as an example](https://github.com/DataDog/dd-trace-js/blob/03e78e58ac8f31a73242114e1305064e561870e7/packages/datadog-instrumentations/src/anthropic.js#L35) - where we manually publish the events to handle streamed scenarios). it's possible there's a cleaner way of doing this, but as this API is not yet built into core TC and would only be for new Node.js versions if it was, I think this _might_ be the cleanest way to do it - manual channel publishes.

logaretm: Good catch, I reproduced the issue and can confirm.

For `streamText`, the parent context does not propagate automatically because the operation returns a lazy stream and the child telemetry is emitted later from stream processing, outside the original `tracePromise` context. So `runStores` would not fix it either unless the stream consumption path / async iterator work is also wrapped.

My understanding is that your Anthropic instrumentation handles this by wrapping the iterator and carrying explicit context through chunk events, effectively manually correlating them. is that right?

AsyncIterator support is being worked on in https://github.com/nodejs/node/pull/62641

I think the AI SDK already emits enough correlation data for us to repair this on our side:

- `streamText`: `callId`
- `step`: `callId`, `stepNumber`
- `languageModelCall`: `callId`
- `executeTool`: `callId`, `toolCall.toolCallId`


One thought here to make it more robust, is could we add a stable `stepId` to the traced `step` event and propagate it to `languageModelCall` / `executeTool`?

We can currently infer parentage from `callId` + current open step, but a `stepId` would make stream correlation deterministic and avoid relying on ordering/state in subscribers.

What do you think?

sabrenner: > My understanding is that your Anthropic instrumentation handles this by wrapping the iterator and carrying explicit context through chunk events, effectively manually correlating them. is that right?

yep, that understanding is correct! in this case, i would imagine as long as the context object is passed through to the underlying iterator consumption, it should be fine, as the instrumentation i linked does.

> What do you think?

i think your suggestions are valid, although i do stand on the side of if it's easy enough to add the manual channel calls similarly to how it's done in that linked instrumentation, that would be pretty cool!

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -527,772 +527,814 @@ export async function generateText<
...
-            // Check if this tool call already has a result in the current response
-            const hasResultInResponse = currentModelResponse.content.some(
-              part =>
-                part.type === 'tool-result' &&
-                part.toolCallId === toolCall.toolCallId,
-            );
-            if (!hasResultInResponse) {
-              pendingDeferredToolCalls.set(toolCall.toolCallId, {
-                toolName: toolCall.toolName,
+              await notify({
```
sabrenner, line 1327: i think it could be helpful to return the `currentStepResult` from this `execute` function for steps, so that when it passes through `tracePromise` it gets recorded as the `result` attribute on the context (event) object!

packages/ai/src/generate-text/stream-text.ts
aayush-kapoor: @sabrenner @logaretm i made a change in https://github.com/vercel/ai/pull/15660/changes/88b1399984670b1cd05ba7741e4352d81746baa8 - `start.runStores()` now wraps the code that creates step/model/tool child spans, and we manually publish `asyncEnd` when the stream completion promise settles.

could you both confirm if that solves the gap since i'm iterating on this based on your feedback? thank you!

aayush-kapoor: it's currently leading to a regression - stand by please

packages/ai/src/generate-text/generate-text.ts
aayush-kapoor: added in https://github.com/vercel/ai/pull/15660/changes/d3c9989ce8f3254f5f5822ce087cadb9f2ce6d70 - lmk if you meant something else

packages/ai/src/telemetry/telemetry.ts
aayush-kapoor: added jsdoc

packages/ai/src/embed/embed.ts
aayush-kapoor: not sure if we have a way around it yet 😓

packages/ai/src/generate-text/stream-text.ts
aayush-kapoor: as @logaretm was saying, would it be possible to make that fix at the individual provider level? how much overhead would that add if the change is pushed in its current state?

sabrenner: i would say manual context management always feels a little hacky, so there'd be a bit of overhead. i also did get to try it out before you reverted and it did look correct for the context management from the tracing channels, so the benefit of having it here would be really nice 😅 curious about the regression, although if absolutely needed we can stitch it back together in subscribers with how it is laid out currently

if it does come to that, i do think this would be nice:
> One thought here to make it more robust, is could we add a stable stepId to the traced step event and propagate it to languageModelCall / executeTool?

logaretm: > would it be possible to make that fix at the individual provider level?

Yep, for `streamText` specifically we can do manual stitching as it currently stands. A stepId would be nice tho.

packages/ai/src/generate-text/generate-text.ts
sabrenner: yep this works, thank you!

packages/ai/src/generate-text/stream-text.ts
aayush-kapoor: > curious about the regression

@sabrenner there were a [couple of unit tests](https://github.com/vercel/ai/actions/runs/27167908786/job/80199755880) failing after this change, wrt to streaming and stream abort; trying to resolve them atm

aayush-kapoor: okay so made a fix in https://github.com/vercel/ai/pull/15660/changes/87801799560b7c9fb33c61958e3dd2ada7d1bb4d that shouldn't cause the regression + should solve the context problem.

please could you all verify if that solves it? the more complicated setup, the better. and please lmk any feedback!

@sabrenner @logaretm

logaretm: @aayush-kapoor I made a bit of a complex test run and it produced the following:

```
app.nested_stress_request
  stream_agent parent-stream-agent
    step 1
      chat parent-stream-model
        GET /provider/parent-stream-model/stream-1
      execute_tool parentFetch
        GET /tool/parentFetch/root-fetch
      execute_tool delegateStream
        GET /tool/delegateStream/root-stream
        stream_agent leaf-stream-from-parent-tool
          step 1
            chat leaf-stream-from-parent-tool-model
        invoke_agent leaf-generate-agent
          step 1
            chat leaf-generate-model
      execute_tool delegateGenerate
        GET /tool/delegateGenerate/root-generate
        invoke_agent sub-generate-agent
          step 1
            chat sub-generate-model
            execute_tool subFetch
              GET /tool/subFetch/sub-data
              stream_agent leaf-stream-from-subfetch
                step 1
                  chat leaf-stream-from-subfetch-model
          step 2
            chat sub-generate-model
    step 2
      chat parent-stream-model
```

So looks really good here 👍

sabrenner: yep i can confirm it looks like it did before the revert as well!

@logaretm curious - how are you extracting the streamed outputs from the llm, step, and parent streamText operations? 

in my case, the `result` attribute on the context object has a `ReadableStream` under `result.stream` for language model calls, while for both `step` and `streamText` it is left `undefined`. i suppose there's always patching the underlying readable stream and propagating it as needed for the other spans, but since i think it gets aggregated for the other telemetry sources, it could be nice to re-use here (but, correct me if i'm wrong)

@aayush-kapoor i was going to ask if it could be populated, but i could be missing something easy 😅

logaretm: Oh. Yep, I see the same thing on my end.

What I was validating was parentage/context propagation, especially for nested streamed calls, tool execution, subagents, and auto-instrumented HTTP spans. That looks good now.

For streamed output enrichment though, `asyncEnd.result` currently looks like:

- `languageModelCall`: `{ stream: ReadableStream }`
- `step`: `undefined`
- `streamText`: `undefined`
- `executeTool`: has the tool result

So streamed output/usage data is not available from the tracing-channel result payload yet. Ideally the streamed tracing contexts would resolve with the same aggregated payloads.

aayush-kapoor: how tricky would it be for you guys to consume the `result` from the underlying `ReadableStream`?

asking because if there's a way you can handle at this your end, we could eliminate adding repetitive code across our core functions and the lifecycle boundaries. if there is anything at the core that's missing, i'm happy to ensure it's flowed through.

also, thanks for all the help on this!

logaretm: It is possible, I think. We would need to `tee()` the language model result stream and replace the context result stream to avoid locking the stream that the SDK still needs to consume.

```ts
// something like this
const [telemetryBranch, sdkBranch] = stream.tee();
message.result.stream = sdkBranch;
```

With that, we should be able to reconstruct the output text, finish reason, token usage, cache usage when present, and tool calls/results. From my side, we don’t mind doing some heavy lifting as long as nothing is fundamentally impossible.

I verified this mechanism against AI SDK’s streamText / languageModelCall stream path. The SDK consumed the replaced stream correctly while the telemetry branch was consumed separately.

I’m happy to accept this as-is and try a few PoCs / implementations on our side. If something turns out not to be possible, we can discuss it in a follow-up PR. Any change from this point would be mostly additive.

WDYT? @sabrenner @aayush-kapoor

aayush-kapoor: @sabrenner @logaretm just fyi, we are targeting GA release of aisdk v7 around june 24th so ideally the integrations are ready from your sides!

sabrenner: noted! i will try the above out tomorrow and follow up here

aayush-kapoor: do you both think this si good to merge? any iteration needed, we can follow up

sabrenner: yeah i think we can follow up! i haven't gotten a chance to try today but we should be able to intercept the chunks or tee off and aggregate ourselves. in a follow-up, it would be nice to have it done and provided in the tracing channel events, but this is good for now i think!

also wondering if object generation tracing channel implementation will end up in a follow up as well 😄

aayush-kapoor: we deprecated generateObject back in v6 and added support for structured outputs through generateText primitive! so by default it, should be available through the result if an output property is set

sabrenner: ooh ok great i must've missed that, sounds good 😎

lgrammel (approved):

lgrammel:
do we have test coverage for embed, rerank with tracing channels? the files were changed w/o test changes (also what is our test coverage for generate/streamText with tracing)? does embedMany also need adjustments?

aayush-kapoor:
> do we have test coverage for embed, rerank with tracing channels? the files were changed w/o test changes (also what is our test coverage for generate/streamText with tracing)? does embedMany also need adjustments?

adding now for embed and rerank. 

for `generateText` and `streamText`, the test coverage is in `packages/ai/src/telemetry/tracing-channel.test.ts` that tests the complete flow (we import those functions specifically) 

i think this is a better pattern because then all diagnostics channel imports live in `packages/ai/telemetry`

PR #15661, opened 2026-05-28
fix(google): round-trip provider code execution parts

### LINKED ISSUE #15613: Vertex: convertToGoogleGenerativeAIMessages drops provider-executed code_execution tool-result, breaks multi-turn with mixed tools
### Description

When a Gemini/Vertex turn contains both the provider-executed `code_execution` tool and a client-defined function tool, the next turn fails with:

> Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.

The root cause is that the assistant message generated by `convertToModelMessages` puts both the `code_execution` `tool-call` and its `tool-result` into the **assistant** content (because `providerExecuted === true`), and then `convertToGoogleGenerativeAIMessages` in `@ai-sdk/google` round-trips them incorrectly:

- The inbound parser for `executableCode` / `codeExecutionResult` parts (`dist/index.mjs:1584-1610` in `3.0.79`, equivalent in `src/google-generative-ai-language-model.ts`) emits `tool-call` / `tool-result` content parts with `providerExecuted: true` and `toolName: "code_execution"`, but does **not** attach `providerMetadata.google.serverToolCallId` / `serverToolType` — unlike the sibling `toolCall` / `toolResponse` branches a few lines below.
- On the outbound side in `convert-to-google-generative-ai-messages.ts` (`3.0.79`):
  - The assistant `tool-call` branch sees no `serverToolCallId/serverToolType` and falls through to `functionCall: { name: 'code_execution', args: ... }`.
  - The assistant `tool-result` branch sees no `serverToolCallId/serverToolType` and `return undefined` — the result is **silently dropped**.

So Vertex receives a model turn with N+1 `functionCall`s but only N `functionResponse`s in the next user turn (one for the client tool, none for `code_execution`), and rejects the request.

### Reproduction

```ts
import { streamText, tool } from 'ai';
import { z } from 'zod';

const vertex = createVertex({ project: 'YOUR_PROJECT', location: 'us-east1' });
const model = vertex('gemini-2.5-flash');

const tools = {
  code_execution: {
    type: 'provider' as const,
    id: 'google.code_execution',
    name: 'codeExecution',
    args: {},
  },
  listItems: tool({
    description: 'List items',
    inputSchema: z.object({}),
    execute: async () => ({ items: ['a', 'b'] }),
  }),
};

// First call — model emits code_execution AND listItems in the same step.
const r1 = await streamText({
  model,
  tools,
  messages: [
    { role: 'user', content: 'Compute 2+2 with Python, then call listItems.' },
  ],
});
const messages = (await r1.response).messages;

// Second call — fails with the Vertex 400 above.
const r2 = await streamText({
  model,
  tools,
  messages: [
    { role: 'user', content: 'Compute 2+2 with Python, then call listItems.' },
    ...messages,
    { role: 'user', content: 'Now repeat the analysis on numbers 3 and 4.' },
  ],
});
```

### Proposed fix

In `convertToGoogleGenerativeAIMessages`, special-case `toolName === 'code_execution'` with `providerExecuted === true` so that:

- assistant `tool-call` → `{ executableCode: { code, language } }` (mirroring the inbound parser)
- assistant `tool-result` → `{ codeExecutionResult: { outcome, output } }`

Symmetry with the inbound parser at `dist/index.mjs:1584-1610` would close the round-trip. Alternatively, attach `serverToolCallId` / `serverToolType` on the code-execution parts at inbound time so the existing `toolCall` / `toolResponse` outbound branches handle them (though Vertex's API for `code_execution` actually uses the `executable_code` / `code_execution_result` parts, not the newer `toolCall`/`toolResponse` shape, so the executable-code branch is the more correct fix).

### Versions

- `@ai-sdk/google@3.0.79` (latest stable)
- `@ai-sdk/google-vertex@4.0.137`
- `ai@6.0.174`

### Related

- #8112 — same shape on Anthropic (provider tool + client tool in parallel break the next turn)
- #9280 — Vertex `code_execution` tool-name mapping issues
- #14767 — fixed `includeServerSideToolInvocations` for Vertex (related path, different bug)

SJvaca30:
## Summary

Closes #15613.

Fixes Google message conversion so provider-executed `code_execution` calls and results are replayed as Gemini code execution content parts instead of being converted to a regular `functionCall` or dropped.

## Details

- Converts provider-executed `code_execution` tool calls back to `executableCode` parts.
- Converts persisted `code_execution` tool results back to `codeExecutionResult` parts.
- Keeps regular client tool calls/results on the existing `functionCall` / `functionResponse` path.
- Adds a patch changeset for `@ai-sdk/google`.

## Tests

- `pnpm --filter @ai-sdk/google test`
- `pnpm --filter @ai-sdk/google build`
- `pnpm exec oxfmt --check packages/google/src/convert-to-google-messages.ts packages/google/src/convert-to-google-messages.test.ts packages/google/src/google-prompt.ts`
- `git diff --check`

### DIFF
--- a/.changeset/fix-google-code-execution-roundtrip.md
+++ b/.changeset/fix-google-code-execution-roundtrip.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Fix Google message conversion to replay provider-executed code execution calls and results as `executableCode` and `codeExecutionResult` parts.
--- a/packages/google/src/convert-to-google-messages.test.ts
+++ b/packages/google/src/convert-to-google-messages.test.ts
@@ -1458,6 +1458,98 @@ describe('tool results with thought signatures', () => {
 });
 
 describe('server tool combination round-trip', () => {
+  it('should convert provider-executed code execution parts back to code execution wire format', () => {
+    const result = convertToGoogleMessages([
+      {
+        role: 'assistant',
+        content: [
+          {
+            type: 'tool-call',
+            toolCallId: 'code-execution-1',
+            toolName: 'code_execution',
+            input: JSON.stringify({
+              language: 'PYTHON',
+              code: 'print(2 + 2)',
+            }),
+            providerExecuted: true,
+          },
+          {
+            type: 'tool-result',
+            toolCallId: 'code-execution-1',
+            toolName: 'code_execution',
+            output: {
+              type: 'json',
+              value: {
+                outcome: 'OUTCOME_OK',
+                output: '4\n',
+              },
+            },
+          },
+          {
+            type: 'tool-call',
+            toolCallId: 'list-items-1',
+            toolName: 'listItems',
+            input: {},
+          },
+        ],
+      },
+      {
+        role: 'tool',
+        content: [
+          {
+            type: 'tool-result',
+            toolCallId: 'list-items-1',
+            toolName: 'listItems',
+            output: { type: 'json', value: { items: ['a', 'b'] } },
+          },
+        ],
+      },
+    ]);
+
+    expect(result.contents).toEqual([
+      {
+        role: 'model',
+        parts: [
+          {
+            executableCode: {
+              language: 'PYTHON',
+              code: 'print(2 + 2)',
+            },
+          },
+          {
+            codeExecutionResult: {
+              outcome: 'OUTCOME_OK',
+              output: '4\n',
+            },
+          },
+          {
+            functionCall: {
+              id: 'list-items-1',
+              name: 'listItems',
+              args: {},
+            },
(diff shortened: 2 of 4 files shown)


SJvaca30:
Quick follow-up on the failing Vercel status: `gh pr checks` reports `Authorization required to deploy`, so this appears to be a preview deployment authorization issue for external contributors rather than a package test/build failure in this PR.

I re-ran local verification for the touched package:

- `pnpm --filter @ai-sdk/google test` — passed (`test:node`: 20 files / 594 tests, `test:edge`: 20 files / 594 tests)
- `pnpm --filter @ai-sdk/google type-check` — passed

Issue #14619, opened 2026-04-18
O(N²) parse time in @ai-sdk/mcp on large MCP tool responses (fix in rexxars/eventsource-parser#28)


emoralesb05:
### Description

Found an O(N²) in `@ai-sdk/mcp`'s SSE parsing on large MCP tool responses. Fix is up at [rexxars/eventsource-parser#28](https://github.com/rexxars/eventsource-parser/pull/28).

I know its a fix on another repo/maintainer but mostly just wanted to give you a heads up. I had already patched it locally to unblock myself, but I'd rather fix it at the source (I don't like to do one-offs haha) so the whole community benefits, especially folks doing tool calls that return larger datasets. Hoping this helps other people hitting the same wall, and happy to find ways to contribute to this community! Let me know if you have any suggestions on it or anything else!

## Context

Ran into this while profiling some big MCP tool calls. `@ai-sdk/mcp`'s streamable-HTTP transport pipes responses through `EventSourceParserStream` (from `@ai-sdk/provider-utils`, backed by `eventsource-parser`). If a tool returns a big payload as one SSE event, which is the common shape for MCP servers that dump bulk data as a single `content[]` block, the `incompleteLine + chunk` concat inside `feed()` goes quadratic.

Not the same case as #5862 / [rexxars/eventsource-parser#19](https://github.com/rexxars/eventsource-parser/issues/19). That one got fixed in 3.0.1 by optimizing `splitLines`, but the concat pattern itself was left alone. So streams with one very long line still hit O(N²).

## Root cause

In `eventsource-parser@3.0.7` `feed()`:

```js
const input = incompleteLine === "" ? chunk : incompleteLine + chunk;
incompleteLine = processLines(input);
```

When one SSE event spans many chunks (no `\n\n` until the very end), `incompleteLine` grows unbounded. Each chunk pays O(total_seen). Across N chunks, O(N²).

My fix at [rexxars/eventsource-parser#28](https://github.com/rexxars/eventsource-parser/pull/28) swaps the string accumulator for a fragment array and defers the concat until a terminator actually arrives. Hot path (empty buffer) stays byte-for-byte identical to the original.

## Repro

Public `mcp-clickhouse` + ClickHouse playground, no auth needed:

```bash
CLICKHOUSE_HOST=play.clickhouse.com CLICKHOUSE_PORT=443 CLICKHOUSE_USER=play \
CLICKHOUSE_PASSWORD='' CLICKHOUSE_SECURE=true \
[...]

PR #15669, opened 2026-05-28
fix(ui): O(N²)→O(N) text/reasoning accumulation in processUIMessageStream + DefaultStreamTextResult

### LINKED ISSUE #15670: perf: O(N²) quadratic memmove in processUIMessageStream + DefaultStreamTextResult on long thinking-mode streams
## Summary

\`processUIMessageStream\` and \`DefaultStreamTextResult\` accumulate streaming text and reasoning deltas with \`part.text += chunk.delta\`. On long thinking-mode streams (Anthropic extended-thinking, MiniMax M2, GPT o-series, Qwen3-thinking, etc.) this is **O(N²)** in cumulative text length: V8/JSC try to keep the rope lazy, but any read of \`.text\` between writes (UI render loop, NDJSON serializer, anything that flattens) forces re-allocation + memcpy of the prior content on every subsequent \`+=\`.

Same bug shape as #14619 (already fixed for \`@ai-sdk/mcp\`), now manifesting in the streaming-text accumulators of the UI/SDK core.

## Reproducer

```ts
import { processUIMessageStream, createStreamingUIMessageState } from 'ai';

const N = 10000;
const CHUNK = 'x'.repeat(200);

  start(c) {
    c.enqueue({ type: 'text-start', id: '1' });
    for (let i = 0; i < N; i++) c.enqueue({ type: 'text-delta', id: '1', delta: CHUNK });
    c.enqueue({ type: 'text-end', id: '1' });
    c.close();
  }
});

const state = createStreamingUIMessageState({ lastMessage: undefined, messageId: 'm1' });

const t0 = performance.now();
const out = processUIMessageStream({
  stream,
  runUpdateMessageJob: (job) => job({ state, write: () => {} }),
  onError: (e) => { throw e; },
});
const reader = out.getReader();
while ((await reader.read()).done === false) {}
console.log(\`elapsed: \${(performance.now() - t0).toFixed(0)} ms\`);
// On ai@6.0.168: ~25 s wall (quadratic at N=10000)
// Expected:      ~1 s wall (linear)
```

## Effect on real workloads

Observed during agentic coding sessions (opencode, MiniMax-M2.7 model, SWE-bench-Pro):

| Instance | Wall time | Notes |
|---|---:|---|
| flipt-21a935 (47 steps) | **1666 s** | per-step 35.4 s |
| vuls-7e91f5 | **7200 s HARD TIMEOUT** | main JS thread at 100% CPU in libc memmove |

CPU profile of the hung process (\`perf record -F 99 -g -p <bun_pid>\` for 60 s):

\`\`\`
30.27% bun  libc.so.6  [.] 0x16414c    ← __memmove_avx_unaligned_erms (allocator memcpy)
       + 7× HeapHelper GC threads at 4.0-4.3% CPU each (JSC GC pinned)
       + main bun thread at 85% CPU state R, never recovering
\`\`\`

This pattern is fully consistent with a quadratic string-concatenation in a hot streaming path.

## Sites

- \`packages/ai/src/ui/process-ui-message-stream.ts\`:
  - \`case 'text-delta'\`: \`textPart.text += chunk.delta;\`
  - \`case 'reasoning-delta'\`: \`reasoningPart.text += chunk.delta;\`
- \`packages/ai/src/generate-text/stream-text.ts\` (in \`DefaultStreamTextResult\`):
  - \`text-delta\` branch: \`activeText.text += part.text;\`
  - \`reasoning-delta\` branch: \`activeReasoning.text += part.text;\`

A fifth site (\`partialToolCall.text += chunk.inputTextDelta\` in \`tool-input-delta\`) has the same issue, but the next line reads the cumulative text for partial-JSON parsing on every delta, so it needs a different approach (incremental partial-JSON parser). Leaving that for a follow-up.

## Proposed fix

PR #15669 — chunks-array + lazy \`.text\` getter. Preserves public API (consumers can still read \`.text\` mid-stream for progressive UI rendering) while keeping the accumulator O(N) total.

meitalbensinai:
Fixes #15670

## Summary

`processUIMessageStream` and `DefaultStreamTextResult` accumulate streaming text/reasoning deltas with `part.text += chunk.delta`. On long thinking-mode streams (Anthropic extended-thinking, MiniMax M2, GPT-o-series, Qwen3-thinking, etc.) this becomes **O(N²)** in cumulative text length: V8/JSC try to keep the rope lazy, but any read of `.text` between writes (UI render loop, NDJSON serializer, anything that flattens) forces re-allocation + `memcpy` of the prior content on every subsequent `+=`.

Same shape as the bug fixed in #14619 for `@ai-sdk/mcp`. See #15670 for the full bug write-up with reproducer + CPU profile.

## Effect on real workloads

opencode batch run on SWE-bench-Pro, MiniMax-M2.7 via Tabnine proxy, same instances both runs:

| Instance | Before | After (this PR) | Δ |
|---|---:|---:|---:|
| `flipt-io/flipt-21a935` | 1666 s (47 steps, 35.4 s/step) | **800 s** (47 steps, 17.0 s/step) | **2.1× faster** |
| `future-architect/vuls-7e91f5` | **7200 s — HARD TIMEOUT** | steady ~21 s/step, completes | **escapes timeout** |

## CPU profile evidence

`perf record -F 99 -g -p <bun_pid> -- sleep 60` on a live opencode stream:

| Metric | Before | After |
|---|---:|---:|
| `__memmove_avx_unaligned_erms` in `libc.so.6` | **30.27%** of samples | **<1.4%** |
| 7× `HeapHelper` GC threads | 4.0–4.3% CPU each (pinned) | **1.3–1.6%** each (normal idle) |
| Main JS thread | 85% CPU state R (pegged) | **52.7%** (active, not pegged) |

## Fix

Chunks-array + lazy getter on `.text`. Public API surface preserved — `textPart.text` / `reasoningPart.text` returns the cumulative string just like before. Internal storage is now an array of chunks; the getter materializes via `join('')` on read with a one-shot collapse (when `_chunks.length > 1`, join + replace with single-element array so subsequent reads are O(1) until next delta).

Accumulator stays **O(N)** total for writes. Reads are O(N) per call but cached between deltas, so consumers that debounce (e.g. UI render at requestAnimationFrame) pay O(N) total.

## Files changed

- `packages/ai/src/ui/process-ui-message-stream.ts` — text-start, text-delta, reasoning-start, reasoning-delta
- `packages/ai/src/generate-text/stream-text.ts` — activeText delta + activeReasoning delta in DefaultStreamTextResult

## Behavior change

None to the public API. The `.text` property is now a defined getter/setter (was a plain data property), which:
- ✅ JSON.stringify / structuredClone invoke the getter and serialize the current accumulated text correctly
- ✅ Reading `part.text` mid-stream returns the cumulative text-so-far (progressive UI rendering keeps working)
- ✅ Writing `part.text = '...'` still works (handled by the setter)
- ⚠ Adds a property descriptor at text-start; we benchmarked this against the baseline and saw no measurable V8 deopt in practice

## Out of scope (follow-up)

`processUIMessageStream` has a third site at `partialToolCall.text += chunk.inputTextDelta` followed by `await parsePartialJson(partialToolCall.text)`. That one reads the cumulative text on every delta, so the same fix doesn't apply cleanly — needs an incremental partial-JSON parser. Left for a follow-up PR; manifests as residual memmove on workloads that stream very large tool inputs (large `write`/`edit` JSON args).

## Reproducer

```ts
import { processUIMessageStream, createStreamingUIMessageState } from 'ai';

const N = 10000;
const CHUNK = 'x'.repeat(200);

  start(c) {
    c.enqueue({ type: 'text-start', id: '1' });
    for (let i = 0; i < N; i++) c.enqueue({ type: 'text-delta', id: '1', delta: CHUNK });
    c.enqueue({ type: 'text-end', id: '1' });
    c.close();
  }
});

const state = createStreamingUIMessageState({ lastMessage: undefined, messageId: 'm1' });

const t0 = performance.now();
const out = processUIMessageStream({
  stream,
  runUpdateMessageJob: (job) => job({ state, write: () => {} }),
  onError: (e) => { throw e; },
});
const reader = out.getReader();
while ((await reader.read()).done === false) {}
console.log(`elapsed: ${(performance.now() - t0).toFixed(0)} ms`);
// Before fix: ~25 s (quadratic at N=10000)
// After this PR: ~1 s (linear)
```

## Test plan

- [ ] Existing UI message stream tests still pass
- [ ] Reproducer above runs in <2 s wall on CI (was ~25 s on `ai@6.0.168`)
- [ ] Manual: chat-UI demo app (progressive text rendering) still renders incrementally


meitalbensinai:
Good catch — fixed in 6374c81.

Swapped the `_chunks` initial assignment from `obj._chunks = [...]` to `Object.defineProperty(obj, '_chunks', { value, writable: true, enumerable: false, configurable: true })`. Subsequent reassignments inside the getter/setter preserve the descriptor (ES spec: `obj.prop = X` on an existing data property only updates `[[Value]]`, not `[[Enumerable]]`/`[[Writable]]`/`[[Configurable]]`).

Verified the leak is closed with a small bench against `readUIMessageStream`:

```
"leaked_in_JSON_stringify":      false      (was true)
"leaked_in_structuredClone":     false      (was true)
"leaked_in_spread":              false      (was true)
"_chunks_in_Object_keys":        false      (was true)

"final_text_correct":            true       (correctness preserved)
"progressive_reads_monotonic":   true       (chat-UI reads still work)
"ms_per_delta":                  0.797      (still O(1) per write — no quadratic regression)
```

A nice side effect: `structuredClone(state.message)` now hands consumers a clean `{type, text: <string>, providerMetadata, state}` snapshot — the implementation detail (`_chunks` array) stays inside the SDK.

aayush-kapoor (commented):
there's some code duplication + certain things to be changed.

so marking this as closed in favor of https://github.com/vercel/ai/pull/15897

Issue #15668, opened 2026-05-28
`Chat.makeRequest` finally throws on `this.activeResponse!.state` when activeResponse is undefined (should be `?.`)


alexanderjacobsen:
`Chat.makeRequest`'s `finally` builds the `onFinish` argument as:

```ts
message: this.activeResponse!.state.message,        // packages/ai/src/ui/chat.ts (main: 760)
...
finishReason: this.activeResponse?.state.finishReason,
```

`this.activeResponse` can be `undefined` when that `finally` runs (e.g. a `trigger: 'resume-stream'` request that overlaps another `makeRequest` such that the single shared `this.activeResponse` was already cleared), so the non-null assertion throws:

```
Cannot read properties of undefined (reading 'state')
  at Chat.makeRequest (packages/ai/src/ui/chat.ts:760)
  at async AbstractChat.resumeStream (packages/ai/src/ui/chat.ts:465)
```

The very next line already guards `finishReason` with `?.`, so the `!` on `message` looks like an oversight.

It's caught (the `finally` wraps `onFinish` in a try/catch → `console.error`), so it's non-fatal — but it trips dev error overlays. It surfaces via `@cloudflare/ai-chat`, which drives client-tool continuations through `resumeStream → makeRequest`, so it fires on every client-tool continuation. Reproduces on `6.0.19x`; not on `6.0.1`.

Suggested fix:

```diff
- message: this.activeResponse!.state.message,
+ message: this.activeResponse?.state.message,
```

Happy to PR if useful.

scosemicolon:
I checked `packages/ai/src/ui/chat.ts` on `main`, and the asymmetry is still there: `message` uses `this.activeResponse!.state.message` while `finishReason` already uses `this.activeResponse?.state.finishReason` in the same `onFinish` payload.

One small regression test that would catch this is a resume-stream path where `activeResponse` is cleared before the `finally` callback builds the `onFinish` argument. The assertion can be narrow: `onFinish` is still called without throwing to the console, and the callback receives `message: undefined` or whatever fallback shape the maintainers prefer.

If this is patched with optional chaining, I would also check downstream callback typing. If `message` is currently typed as always present, the implementation fix may need either a guarded `onFinish` call when no active response exists, or a type update that makes the missing-message case explicit.

lgrammel:
### Identify issue type

```json
{
  "evidence": [
[...]

PR #15672, opened 2026-05-28
fix(chat): replace non-null assertion with optional chaining on activeResponse

xxiaoxiong:
**Issue**: #15668

`Chat.makeRequest`'s `finally` block builds the `onFinish` argument using `this.activeResponse!.state.message`, but `this.activeResponse` can be `undefined` when a trigger like `'resume-stream'` overlaps another `makeRequest` call that clears the shared `this.activeResponse`. This throws:

```
```

The very next line `this.activeResponse?.state.finishReason` already uses optional chaining, so the non-null assertion appears to be an oversight.

**Fix**: Change `!` to `?` — `this.activeResponse?.state.message`.

### DIFF
--- a/packages/ai/src/ui/chat.ts
+++ b/packages/ai/src/ui/chat.ts
@@ -757,7 +757,7 @@ export abstract class AbstractChat<UI_MESSAGE extends UIMessage> {
     } finally {
       try {
         this.onFinish?.({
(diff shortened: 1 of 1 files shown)


abhagsain:
@xxiaoxiong What's the fix for this? We're getting this with cloudflare agent's SDK

PR #15705, opened 2026-05-29
feat(provider/cerebras): add service tier support

shaper:
## Summary

Adds [service tier](https://inference-docs.cerebras.ai/capabilities/service-tiers) support to `@ai-sdk/cerebras`:

- **`serviceTier`** provider option (`'auto' | 'default' | 'flex' | 'priority'`) — sent as `service_tier` in the request body.
- **`queueThreshold`** provider option (`50`–`20000` ms) — sent as the `queue_threshold` HTTP header (Cerebras only evaluates it for `flex`/`auto`).
- The effective tier returned by the API is surfaced on **`result.providerMetadata.cerebras.serviceTier`**, mirroring how openai/google/vertex expose the actually-used tier.

## Implementation

Cerebras extends `OpenAICompatibleChatLanguageModel`, so this reuses the base class's existing extension points:

- **Readback:** a new `cerebras-metadata-extractor.ts` implements the base `MetadataExtractor` hook, reading `service_tier_used` (falling back to `service_tier`) from both non-streaming responses and stream chunks. It is set **only when the API returns it** (OpenAI style) — no fabricated/null value.
- **Body:** `transformCerebrasRequestBody` maps the passed-through `serviceTier` → `service_tier` and strips `queueThreshold` from the body.
- **Header:** the existing `doGenerate`/`doStream` overrides parse the provider options and inject the `queue_threshold` header before delegating to `super`.

## Why draft

Service tiers are in **private preview**. Tested live against a non-preview key:

- ✅ Request wiring confirmed on the wire — `service_tier` in body, `queue_threshold` header sent, requests accepted for all tiers.
- ⚠️ The response did **not** include `service_tier_used` / `service_tier` for any tier (incl. `auto`) on this account — the param is silently accepted but not echoed, so `providerMetadata.cerebras.serviceTier` is `undefined`.

The extractor handles both documented field spellings and will surface the value automatically once tested against a preview-enabled key. **Open question:** confirm the exact field name/shape in the real preview response (top-level vs nested under `usage`/`time_info`) before merging.

## Tests / verification

- `pnpm test` (node + edge): 18/18 pass — new tests cover body param, header, and metadata readback for generate + stream.
- `pnpm type-check:full` and `pnpm check`: clean.
- Examples added: `examples/ai-functions/src/{generate-text,stream-text}/cerebras/service-tier.ts`.
- Docs updated: `content/providers/01-ai-sdk-providers/40-cerebras.mdx`.

### DIFF
--- a/packages/cerebras/src/cerebras-chat-language-model.test.ts
+++ b/packages/cerebras/src/cerebras-chat-language-model.test.ts
@@ -38,6 +38,67 @@ function createJsonFixtureFetchMock(filename: string) {
   );
 }
 
+function createRecordingJsonFetchMock(body: unknown) {
+  const calls: { init: RequestInit | undefined }[] = [];
+  const fetch = vi.fn(
+    (_input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+      calls.push({ init });
+      return Promise.resolve(
+        new Response(JSON.stringify(body), {
+          headers: { 'content-type': 'application/json' },
+        }),
+      );
+    },
+  );
+  return { fetch, calls };
+}
+
+function createRecordingStreamFetchMock(chunks: unknown[]) {
+  const calls: { init: RequestInit | undefined }[] = [];
+  const fetch = vi.fn(
+    (_input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
+      calls.push({ init });
+      const body = [
+        ...chunks.map(chunk => `data: ${JSON.stringify(chunk)}\n\n`),
+        'data: [DONE]\n\n',
+      ].join('');
+      return Promise.resolve(
+        new Response(body, {
+          headers: { 'content-type': 'text/event-stream' },
+        }),
+      );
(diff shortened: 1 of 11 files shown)


packages/cerebras/src/cerebras-chat-language-model.test.ts
```diff
@@ -38,6 +39,67 @@ function createJsonFixtureFetchMock(filename: string) {
...
+  },
+  {
+    id: 'chatcmpl-test',
+    model: 'gpt-oss-120b',
+    choices: [{ delta: {}, finish_reason: 'stop' }],
+    usage: { prompt_tokens: 1, completion_tokens: 1, total_tokens: 2 },
+    service_tier_used: 'flex',
+  },
+];
+
```
aayush-kapoor, line 102: maybe we should add a TODO comment saying to use actual fixtures when the feature goes public

aayush-kapoor (approved):

PR #15706, opened 2026-05-29
fix(anthropic): respect disableParallelToolUse in jsonTool mode

### LINKED ISSUE #15652: @ai-sdk/anthropic: `disableParallelToolUse: true` is hardcoded when `structuredOutputMode: 'jsonTool'`, ignores user override
## Summary

When `providerOptions.anthropic.structuredOutputMode: 'jsonTool'` is set, the provider hardcodes `disableParallelToolUse: true` and ignores any user-provided `disableParallelToolUse: false` in `providerOptions.anthropic`. This silently disables parallel tool use for real tools (e.g. `searchDocs` ×3 in parallel), not just for the synthetic `json` tool, with no way to opt back in.

## Affected version

`@ai-sdk/anthropic@3.0.80` (latest, and every 3.x I checked back to 3.0.7).

Behavior is a regression from `@ai-sdk/anthropic@2.0.2`, which always respected `anthropicOptions?.disableParallelToolUse`.

## Source

`packages/anthropic/src/anthropic-messages-language-model.ts` (compiled at `dist/index.js:3637-3650` in 3.0.80):

\`\`\`ts
jsonResponseTool != null ? {
  tools: [...tools ?? [], jsonResponseTool],
  toolChoice: { type: 'required' },
  disableParallelToolUse: true,            // ← hardcoded, user value ignored
  cacheControlValidator,
  supportsStructuredOutput,
} : {
  tools: tools ?? [],
  toolChoice,
  disableParallelToolUse: anthropicOptions?.disableParallelToolUse,
  ...
}
\`\`\`

## Why this matters

\`structuredOutputMode: 'jsonTool'\` isn't a stylistic choice — it's the documented escape hatch for a real streaming bug on the default path.

### The trap

**Without \`jsonTool\`** (default \`'auto'\` / native \`output_format\` path on Claude Sonnet 4.5 / Opus 4.5 / Opus 4.1 with beta header \`structured-outputs-2025-11-13\`):

In \`streamText\` mode the final structured JSON arrives as **one giant text-delta event after 20+ seconds of silence** post-tool-result, instead of token-by-token. We measured a single 3,197-character text-delta in production diagnostics.

Root cause is in the AI SDK's \`createOutputTransformStream\` (\`ai/dist/index.mjs\`): the transform only publishes a text chunk when \`parsePartialOutput\` produces a new parseable result. For deeply nested schemas (e.g. \`{ dataComponents: [...] }\`), \`parsePartialJson\` can't produce a valid partial until the full JSON closes, so every intermediate text-delta is swallowed and the whole buffer is dumped at \`finish-step\`.

Anthropic's HTTP API itself streams \`text_delta\` events fine — confirmed by Anthropic's docs and Vercel's AI Gateway examples. Upgrading \`ai\` (6.0.14 → 6.0.168) and \`@ai-sdk/anthropic\` (3.0.7 → 3.0.71) does NOT fix it — the transform-gate has not changed upstream. Community tracking: #3422, #12427, #12298, #7220, #9351 (all open or recent, multiple reporters, no upstream fix).

**With \`jsonTool\`** (this issue): streaming is smooth, but parallel tool use is silently force-disabled.

### Net effect on consumers

Apps that need streaming structured output on Claude have two doors:
- **Door A** — \`'auto'\` / native path: 20-second silent burst, but parallel tool calls work.
- **Door B** — \`jsonTool\` path: smooth streaming, but no parallel tool calls under any circumstances.

There's no third door, because the user-provided \`disableParallelToolUse: false\` is dropped on the floor on Door B. The only remaining options are a two-call architecture (doubles round-trips, complicates state) or accepting one of the regressions.

This issue is about restoring Door C: smooth streaming **and** parallel real-tool fan-out, by respecting the user-provided \`disableParallelToolUse\` value on the \`jsonTool\` path.

## Repro

\`\`\`ts
const result = streamText({
  model: anthropic('claude-sonnet-4-5'),
  tools: { searchDocs },
  output: Output.object({ schema: z.object({ items: z.array(z.string()) }) }),
  providerOptions: {
    anthropic: {
      structuredOutputMode: 'jsonTool',
      disableParallelToolUse: false,   // ← user opts in to parallel
    },
  },
  prompt: 'Call searchDocs 3 times in parallel with distinct queries.',
});
\`\`\`

Network trace shows \`disable_parallel_tool_use: true\` on the Anthropic request, regardless of the user-provided \`false\`.

## Proposed fix

Respect the user-provided value, defaulting to \`true\` to preserve current behavior:

\`\`\`ts
disableParallelToolUse: anthropicOptions?.disableParallelToolUse ?? true,
\`\`\`

This keeps the safe default (you almost never want the model emitting the synthetic \`json\` tool in parallel with a real tool) while letting users who understand the tradeoff opt in to parallel *real-tool* fan-out.

## Willing to PR

Yes — happy to open a PR if the proposed shape is acceptable. Wanted to file the issue first so the maintainers can weigh in on whether the default should stay \`true\` or whether there's a more nuanced fix (e.g. only hardcode \`true\` on the final step that actually emits the json tool).

SJvaca30:
## Summary

Closes #15652.

Respect explicit `disableParallelToolUse` overrides when Anthropic structured output uses the `jsonTool` fallback. The json tool path still defaults to disabling parallel tool use, but a user-provided `false` is now passed through to Anthropic instead of being overwritten.

## Details

- Keeps the existing `jsonTool` default of `disable_parallel_tool_use: true`.
- Preserves `disableParallelToolUse: false` when users explicitly opt into parallel tool use with json tool responses and other tools.
- Adds regression coverage for both the default and explicit override cases.
- Adds a patch changeset for `@ai-sdk/anthropic`.

## Tests

- `pnpm --filter @ai-sdk/anthropic test:node -- anthropic-language-model.test.ts -t "disableParallelToolUse"`
- `pnpm exec oxfmt --check packages/anthropic/src/anthropic-language-model.ts packages/anthropic/src/anthropic-language-model.test.ts .changeset/sharp-ants-remember.md`
- `pnpm exec ultracite fix packages/anthropic/src/anthropic-language-model.ts packages/anthropic/src/anthropic-language-model.test.ts`
- `pnpm exec ultracite check packages/anthropic/src/anthropic-language-model.ts packages/anthropic/src/anthropic-language-model.test.ts`
- `git diff --check`
- `pnpm --filter @ai-sdk/anthropic test`
- `pnpm --filter @ai-sdk/anthropic type-check`
- `pnpm --filter @ai-sdk/anthropic build`

### DIFF
--- a/.changeset/sharp-ants-remember.md
+++ b/.changeset/sharp-ants-remember.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Respect explicit `disableParallelToolUse` overrides when using `jsonTool` structured output mode.
--- a/packages/anthropic/src/anthropic-language-model.test.ts
+++ b/packages/anthropic/src/anthropic-language-model.test.ts
@@ -1888,6 +1888,105 @@ describe('AnthropicLanguageModel', () => {
       });
     });
 
+    it('should default disableParallelToolUse to true for json tool responses with other tools', async () => {
+      prepareJsonFixtureResponse('anthropic-json-other-tool.1');
+
+      await model.doGenerate({
+        prompt: TEST_PROMPT,
+        tools: [
+          {
+            type: 'function',
+            name: 'get-weather',
+            description: 'Get the weather in a location',
+            inputSchema: {
+              type: 'object',
+              properties: {
+                location: { type: 'string' },
+              },
+              required: ['location'],
+              additionalProperties: false,
+              $schema: 'http://json-schema.org/draft-07/schema#',
+            },
+          },
+        ],
+        providerOptions: {
+          anthropic: {
+            structuredOutputMode: 'jsonTool',
+          } satisfies AnthropicLanguageModelOptions,
+        },
+        responseFormat: {
+          type: 'json',
+          schema: {
+            type: 'object',
+            properties: {
+              weather: { type: 'string' },
+              temperature: { type: 'number' },
+            },
+            required: ['weather', 'temperature'],
+            additionalProperties: false,
+            $schema: 'http://json-schema.org/draft-07/schema#',
+          },
+        },
+      });
+
+      expect(await server.calls[0].requestBodyJson).toMatchObject({
+        tool_choice: {
+          type: 'any',
+          disable_parallel_tool_use: true,
+        },
+      });
+    });
+
+    it('should respect disableParallelToolUse false for json tool responses with other tools', async () => {
+      prepareJsonFixtureResponse('anthropic-json-other-tool.1');
+
+      await model.doGenerate({
+        prompt: TEST_PROMPT,
+        tools: [
+          {
+            type: 'function',
+            name: 'get-weather',
+            description: 'Get the weather in a location',
+            inputSchema: {
+              type: 'object',
+              properties: {
+                location: { type: 'string' },
+              },
+              required: ['location'],
+              additionalProperties: false,
+              $schema: 'http://json-schema.org/draft-07/schema#',
+            },
+          },
+        ],
+        providerOptions: {
+          anthropic: {
+            structuredOutputMode: 'jsonTool',
+            disableParallelToolUse: false,
+          } satisfies AnthropicLanguageModelOptions,
+        },
+        responseFormat: {
+          type: 'json',
+          schema: {
+            type: 'object',
+            properties: {
+              weather: { type: 'string' },
+              temperature: { type: 'number' },
+            },
+            required: ['weather', 'temperature'],
+            additionalProperties: false,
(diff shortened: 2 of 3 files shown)


SJvaca30:
Quick follow-up on the failing Vercel status: `gh pr checks` reports `Authorization required to deploy`, so this appears to be a preview deployment authorization issue for external contributors rather than a package test/build failure in this PR.

I re-ran local verification for the touched package:

- `pnpm --filter @ai-sdk/anthropic test` — passed (`test:node`: 11 files / 428 tests, `test:edge`: 11 files / 428 tests)
- `pnpm --filter @ai-sdk/anthropic type-check` — passed

lgrammel:
Thank you for the pull request. Issue #15652 was closed as completed. Closing this pull request.

PR #15725, opened 2026-05-30
fix(ai-sdk): return transformed elements in arrayOutputStrategy validation

### LINKED ISSUE #15707: arrayOutputStrategy.validateFinalResult discards coerced values, returning raw JSON instead of validated output (output-strategy.ts L244-251)
## Description

`arrayOutputStrategy.validateFinalResult` validates each element of the output array using `safeValidateTypes`, but silently discards the coerced value produced by the validator. It then returns the raw, pre-validation JSON array unchanged. As a result, any schema-level transformations -- including `.default()`, `.transform()`, `.coerce`, and Zod branded types -- are never applied to the final result returned to the caller.

## Affected Code

`packages/ai/src/generate-object/output-strategy.ts`, lines 241-251 (current HEAD):

```ts
const inputArray = value.elements as Array<JSONObject>;

// check that each element in the array is of the correct type:
for (const element of inputArray) {
  const result = await safeValidateTypes({ value: element, schema });
  if (!result.success) {
    return result;
  }
  // BUG: result.value (the coerced output) is never collected
}

return { success: true, value: inputArray as Array<ELEMENT> };
//                                ^^^^^^^^^^
// Returns the raw pre-coercion JSON, not the validated+transformed values
```

`safeValidateTypes` returns `{ success: true, value: T }` where `value` is the schema-coerced output. The loop checks for failure but never collects `result.value`. The function then returns `inputArray` -- the raw JSON -- bypassing all transformations.

## Why This Is a Bug

Zod and other validation libraries do not merely validate; they transform. When a schema has:

- `.default(...)` -- missing fields are filled in by the validator
- `.transform(...)` -- values are converted (e.g., string to Date, number to branded type)
- `z.coerce.*` -- type coercion (e.g., `z.coerce.number()` converts string "42" to 42)
- `.brand<...>()` -- branded types that wrap the value

...the coerced/transformed value lives in `result.value`, not in the original `element`. By discarding `result.value` and returning `inputArray`, the function silently delivers untransformed data to the caller. The caller receives a type that claims to be `Array<ELEMENT>` but is actually raw JSON that has not been processed by the schema.

## Reproduction

```ts
import { generateObject } from "ai";
import { z } from "zod";

const ElementSchema = z.object({
  name: z.string(),
  score: z.string().transform(s => parseInt(s, 10)), // transform: string -> number
  label: z.string().default("unlabeled"),             // default: fills missing field
});

const { object } = await generateObject({
  model: yourModel,
  output: "array",
  schema: ElementSchema,
  prompt: "Return 3 items with name and score fields only (omit label).",
});

// Expected: object[0].score is a number, object[0].label === "unlabeled"
// Actual:   object[0].score is still a string, object[0].label is undefined
//           Raw JSON returned; transforms and defaults were never applied
console.log(typeof object[0].score); // "string" instead of "number"
console.log(object[0].label);        // undefined instead of "unlabeled"
```

## Suggested Fix

Collect the coerced values from each `result.value` and return those instead of `inputArray`:

```ts
const inputArray = value.elements as Array<JSONObject>;
const coercedArray: Array<ELEMENT> = [];

for (const element of inputArray) {
  const result = await safeValidateTypes({ value: element, schema });
  if (!result.success) {
    return result;
  }
  coercedArray.push(result.value);
}

return { success: true, value: coercedArray };
```

This mirrors how a correct single-object path works: validate, then return `result.value`.

## Prior Issue

Issue #3635 was closed administratively without a fix being applied. The code at HEAD is unchanged from when that issue was filed -- the bug is still present.

## Related

A parallel defect exists in `generate-text` output handling where validated values are similarly discarded. That will be filed as a separate issue.

---

Reporter: Nexory

shujanislam:
## Background

When utilizing array-based structured outputs with a validation schema (e.g., Zod), schema transformations such as type coercion (`z.coerce`), defaults (`.default()`), and custom transforms (`.transform()`) were being silently lost. 

While individual elements in the array were successfully validated and transformed by `safeValidateTypes`, the `arrayOutputStrategy.validateFinalResult` function was discarding these valid, transformed outputs. Instead, it returned the original, raw JSON array received from the LLM, leading to unexpected runtime type discrepancies in downstream application logic.

## Summary

Refactored `arrayOutputStrategy.validateFinalResult` to properly map and accumulate the successfully validated and transformed elements returned by the schema validator. The function now constructs and returns a new array containing the transformed data rather than returning the raw, unvalidated input array.

## Manual Verification

Verified by passing an LLM array response through a Zod schema utilizing `z.coerce.date()` and `.default()` modifiers:
- **Before fix:** The resulting array contained raw ISO strings and missing properties where defaults should have been injected.
- **After fix:** The resulting array correctly contains instantiated JavaScript `Date` objects and all schema-defined default values, aligning perfectly with the inferred TypeScript types.

## Checklist

## Related Issues

Fixes #15707

### DIFF
--- a/.changeset/neat-apricots-swim.md
+++ b/.changeset/neat-apricots-swim.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+Fixes generateObject array output to return schema-validated/transformed elements instead of raw JSON elements
--- a/packages/ai/src/generate-object/output-strategy.ts
+++ b/packages/ai/src/generate-object/output-strategy.ts
@@ -238,7 +238,8 @@ const arrayOutputStrategy = <ELEMENT>(
         };
       }
 
-      const inputArray = value.elements as Array<JSONObject>;
+      const inputArray = value.elements as Array<JSONValue>;
+      const resultArray: Array<ELEMENT> = [];
 
       // check that each element in the array is of the correct type:
       for (const element of inputArray) {
@@ -246,9 +247,11 @@ const arrayOutputStrategy = <ELEMENT>(
         if (!result.success) {
           return result;
         }
+
+        resultArray.push(result.value);
       }
 
-      return { success: true, value: inputArray as Array<ELEMENT> };
+      return { success: true, value: resultArray };
     },
 
     createElementStream(


lgrammel:
Closing this PR because issue #15707 has already been closed.

PR #15750, opened 2026-06-01
feat(provider/google): add Gemini text-to-speech (speech) model support

31Carlton7:
## Background

`@ai-sdk/google` supports language, embedding, image, and video models, but not
speech. This adds text-to-speech so Gemini TTS models work with
`experimental_generateSpeech`.

## Summary

Adds a `GoogleSpeechModel` (`SpeechModelV4`) and wires `google.speech()` /
`google.speechModel()` into the provider. Supported models:

- `gemini-2.5-flash-preview-tts`
- `gemini-2.5-pro-preview-tts`
- `gemini-3.1-flash-tts-preview`

These call the Gemini `:generateContent` endpoint with
`responseModalities: ['AUDIO']` and a `speechConfig`.

```ts
import { experimental_generateSpeech as generateSpeech } from 'ai';

const { audio } = await generateSpeech({
  model: google.speech('gemini-2.5-flash-preview-tts'),
  text: 'Hello from the AI SDK!',
  voice: 'Kore',
});
```

### Details

- **Voice** — the top-level `voice` maps to `prebuiltVoiceConfig.voiceName`
  (30 prebuilt voices, default `Kore`).
- **Multi-speaker** — dialogue via `providerOptions.google.multiSpeakerVoiceConfig`.
- **Output** — Gemini returns raw PCM; by default the provider wraps it in a WAV
  container so the result is playable and detected as `audio/wav`.
  `outputFormat: 'pcm'` returns the raw bytes (with a warning). Sample rate and
  mime type are surfaced in `providerMetadata.google`.
- **Instructions** — `instructions` are prepended to the prompt, since Gemini
  honors natural-language style direction.
- **Unsupported** — `speed` and `language` emit warnings (no Gemini API equivalent).

## Verification

- Unit tests built through the provider factory + `createTestServer` against the
  real `:generateContent` endpoint: request mapping, default/explicit voice,
  headers + user-agent, WAV wrapping + sample-rate parsing, `pcm` mode, `speed`/
  `language` warnings, instructions, multi-speaker, empty response, and provider
  metadata. The full `@ai-sdk/google` suite passes on node + edge.
- Examples in `examples/ai-functions/src/generate-speech/google/` (basic, voice,
  instructions, multi-speaker).
- Docs: a Speech Models section on the Google provider page and rows in the core
  speech docs.

## Notes

- The model IDs above are preview models from Google's
  [speech generation docs](https://ai.google.dev/gemini-api/docs/speech-generation).
- No new dependencies; changeset included (patch).


packages/google/src/index.ts
```diff
@@ -29,6 +29,12 @@ export type {
   /** @deprecated Use `GoogleVideoModelId` instead. */
   GoogleVideoModelId as GoogleGenerativeAIVideoModelId,
 } from './google-video-settings';
+export type {
+  GoogleSpeechModelOptions,
+  /** @deprecated Use `GoogleSpeechModelOptions` instead. */
+  GoogleSpeechModelOptions as GoogleGenerativeAISpeechProviderOptions,
```
kkdawkins: Speech is brand new, so there's no prior export to stay compatible with. Drop this day-one `@deprecated` alias and just export `GoogleSpeechModelOptions`.

packages/google/src/google-speech-model.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    const googleOptions = await parseProviderOptions({
+      provider: 'google',
+      providerOptions,
+      schema: googleSpeechProviderOptionsSchema,
+    });
+
+    // Gemini honors natural-language style direction expressed in the prompt
+    // text, so map `instructions` onto the spoken content rather than warning.
+    const promptText = instructions ? `${instructions}: ${text}` : text;
```
kkdawkins: `instructions` is prepended before the multi-speaker branch, so with both set you get `Say cheerfully: Joe: Hi. Jane: Hello.`, which can break speaker-name parsing. Worth a guard or a test/doc note.

packages/google/src/google-provider.ts
```diff
@@ -223,6 +236,14 @@ export function createGoogle(
       generateId: options.generateId ?? generateId,
     });
 
+  const createSpeechModel = (modelId: GoogleSpeechModelId) =>
+    new GoogleSpeechModel(modelId, {
+      provider: `${providerName}.speech`,
```
kkdawkins, line 241: Speech uses a `.speech` suffix while language/embedding/image/video use bare `providerName`. Matches `interactions`, so likely intentional — just flagging for consistency.

packages/google/src/google-speech-model.ts
```diff
@@ -0,0 +1,265 @@
...
+      body: requestBody,
+      failedResponseHandler: googleFailedResponseHandler,
+      successfulResponseHandler: createJsonResponseHandler(
+        googleSpeechResponseSchema,
+      ),
+      abortSignal: options.abortSignal,
+      fetch: this.config.fetch,
+    });
+
+    // Extract the first inline audio part from the candidates.
```
shaper: it looks like we throw away potential other audio parts/candidates -- do we/would we want to also produce these, is there precedence for this in other speech providers/models, how is it handled elsewhere?

packages/google/src/google-speech-model.ts
```diff
@@ -0,0 +1,265 @@
...
+        },
+      },
+    };
+  }
+}
+
+/**
+ * Parses the sample rate from a PCM mime type such as `audio/L16;rate=24000`.
+ */
+function parseSampleRate(mimeType: string | undefined): number | undefined {
```
shaper, line 241: do we need this and the addWavHeader logic? comparing to deepgram which just returns audio essentially directly.

is this necessary because we're conforming to what is documented/expected for `generateSpeech` and format, or was it just convenient for default?

packages/google/src/google-speech-model.ts
```diff
@@ -0,0 +1,265 @@
...
+      }
+      if (base64Audio != null) {
+        break;
+      }
+    }
+
+    const sampleRate = parseSampleRate(mimeType) ?? DEFAULT_SAMPLE_RATE;
+    const pcm =
+      base64Audio != null
+        ? convertBase64ToUint8Array(base64Audio)
```
shaper, line 191: if possible to keep it in the original format reduces overhead, do we need to do the format analysis/prepending/etc. below?

```
audio: string | Uint8Array;
```

the audio field in SpeechModelV4Result allows Uint8Array for example

packages/google/src/index.ts
31Carlton7: Good catch, new update now just exports GoogleSpeechModelOptions + GoogleSpeechModelId

packages/google/src/google-speech-model.ts
31Carlton7: Fixed. `instructions` are now ignored (with warning) when multiSpeakerVoiceConfig is set

packages/google/src/google-speech-model.ts
31Carlton7: `generateSpeech`/`SpeechModelV4` returns a single audio, and Gemini's `:generateContent` returns one inline audio part per request so there's no multi-part audio to merge, and other speech providers similarly return a single blob. So we take the first inline-data part (added a comment to make that explicit but can remove to stay inline with current AI SDK patterns). If Gemini ever returns multiple audio parts we can revisit.

packages/google/src/google-speech-model.ts
31Carlton7: Yea it's necessary as Gemini returns headerless raw PCM (`audio/L16;rate=24000`) unlike Deepgram or OpenAI which return a container format (mp3/wav/opus) directly.

Claude says:

`generateSpeech` derives `result.audio.mediaType` from `detectMediaType(bytes)`, which has no PCM/L16 signature — so raw PCM gets mislabeled audio/mp3 and won't play. The 44-byte WAV header makes it detect as audio/wav and play correctly. `outputFormat: 'pcm'` returns the raw bytes for zero-conversion callers.

packages/google/src/google-speech-model.ts
31Carlton7: From Claude:

`generateSpeech` infers the media type from the bytes via `detectMediaType`, which has no raw‑PCM signature — so passing Gemini's headerless PCM through would mislabel it `audio/mp3` and break playback (Deepgram can return directly because its API emits a container format; `generateContent` only gives raw PCM). So we wrap a 44‑byte WAV header by default, and `outputFormat: 'pcm'` returns the raw `Uint8Array` untouched.

shaper (approved):

PR #15754, opened 2026-06-02
examples: add signed tool receipts

tomjwxf:
## Summary

- Adds an AI SDK example that emits a hash-chained receipt before and after each tool execution.
- Uses the existing `onToolExecutionStart` and `onToolExecutionEnd` callbacks, so the example follows the SDK lifecycle rather than wrapping tool definitions.
- Captures tool name, tool call ID, validated input digest, output/error digest, duration, and previous receipt hash.
- Keeps the example dependency-free with HMAC-SHA256; comments note that public/offline third-party verification should use an asymmetric signature such as Ed25519.

## Verification

- `git diff --check`
- Inspected `ToolExecutionStartEvent` / `ToolExecutionEndEvent` types in `packages/ai/src/generate-text/tool-execution-events.ts`

I could not run `pnpm type-check` in this sandbox because `pnpm` is not installed and the environment cannot resolve npm registry packages.

### DIFF
--- a/examples/ai-functions/src/generate-text/signed-tool-receipts.ts
+++ b/examples/ai-functions/src/generate-text/signed-tool-receipts.ts
@@ -1,0 +1,205 @@
+import { openai } from '@ai-sdk/openai';
+import {
+  generateText,
+  isStepCount,
+  tool,
+  type ToolExecutionEndEvent,
+  type ToolExecutionStartEvent,
+  type ToolSet,
+} from 'ai';
+import { createHash, createHmac } from 'node:crypto';
+import { z } from 'zod';
+import { run } from '../lib/run';
+
+// This example keeps the receipt signer dependency-free with HMAC-SHA256.
+// For third-party/offline verification, replace signReceipt with an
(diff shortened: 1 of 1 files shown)


piiiico:
The HMAC note is worth following up on. HMAC requires the verifier to hold the same secret, so "third-party verification" becomes key distribution. Ed25519 doesn't have that problem. Public key at a URL; anyone verifies without touching your signing material.

[AgentLair](https://agentlair.dev) already ships this for Vercel AI SDK agents. Agent Auth Tokens are EdDSA (Ed25519) JWTs per session, public key at `https://agentlair.dev/.well-known/jwks.json`. The callback integration maps to:

```typescript
  model, tools,
  ...createAARCallbacks({ agentToken: process.env.AGENTLAIR_AAT }),
});
// receipts Ed25519-signed; verify: https://agentlair.dev/.well-known/jwks.json
```

Verifiable by any third party, no account needed.

aayush-kapoor:
we are currently closing down PRs that are not a priority in order to reduce the number of open PRs and issues - if there is an actual bug, we ask that you please file an issue first

PR #15759, opened 2026-06-02
ENG-8511 normalize UI data URL file parts

vlad4j:
## Summary

- Normalize `FileUIPart.url` data URLs to raw base64 data during `convertToModelMessages`.
- Reuse the parsed data URL media type for the resulting model file part.
- Add regression coverage for user and assistant file parts that use data URLs.

## Why

UI `FileUIPart` allows data URLs, but preserving the full `data:image/...;base64,...` string in model messages can cause provider serializers or downstream code to treat it as a URL. Anthropic then rejects it with `URL scheme must be http or https, got data:`. Normalizing during UI conversion makes the model message provider-safe earlier.

## Validation

- `pnpm --filter ai test:node src/ui/convert-to-model-messages.test.ts`
- `pnpm type-check:full`
- `pnpm check`

### DIFF
--- a/.changeset/eng-8511-data-url-file-parts.md
+++ b/.changeset/eng-8511-data-url-file-parts.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Normalize data URL file parts to base64 data in `convertToModelMessages`.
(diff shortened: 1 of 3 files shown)


vlad4j:
Closing per request.

vlad4j:
My bad, eccidently asked the agent to create a PR in a wrong chat. Was using AI to ask questions about your Docs 🤷

PR #15768, opened 2026-06-02
feat(gateway): add experimental realtime model

kkdawkins:
## Background

The AI Gateway can route realtime (WebSocket) sessions, but the SDK had no client for it. Browsers can't set headers on a WS upgrade, so auth and model selection need an out-of-band channel, and the gateway normalizes realtime the same way it normalizes every other modality.

## Summary

Adds `gateway.experimental_realtime()` — a `RealtimeModelV4` that connects through the AI Gateway.

- **Identity codec.** The gateway speaks the normalized AI SDK realtime protocol, so `parseServerEvent`/`serializeClientEvent`/`buildSessionConfig` are pass-through. Provider-specific wire details stay server-side.
- **Auth via subprotocol.** The bearer token rides `Sec-WebSocket-Protocol` (`ai-gateway-auth.<token>`), mirroring OpenAI's `openai-insecure-api-key`. Centralized in `gateway-realtime-auth.ts`: `getGatewayRealtimeProtocols` (client encode) + `getGatewayRealtimeAuthToken` (server decode) so the SDK and the gateway service share one definition.
- **Model via `?ai-model-id=`.** The WS transport of the `ai-model-id` header; passed through verbatim (the gateway owns qualification).
- **Provider options.** `parseGatewayProviderOptions` validates `providerOptions.gateway` against the existing schema and returns the typed `GatewayProviderOptions`, shared with the gateway service. Options flow through `session.update` (the realtime analogue of the request body) — no model-side code.

Design decisions:
- Bearer-via-subprotocol over an ephemeral-secret mint: the gateway uses bearer auth today; a mint is a heavier, stateful follow-up.
- `GatewayRealtimeModel` is kept internal (only the factory + `GatewayRealtimeModelId` + shared helpers are exported), matching the other gateway models.

## Manual Verification

Ran the realtime browser demo (cooking-coach) against a gateway preview deployment: voice in / audio + transcript out, tool calls, and `providerOptions.gateway` (tags/quota) observed server-side.

## Checklist

- [ ] All commits are signed
- [x] Documentation has been added / updated
- [x] A patch changeset has been added

## Future Work

- Ephemeral client-secret mint (keeps long-lived tokens out of the browser).
- Transcription/translation route kinds.


packages/gateway/src/gateway-realtime-auth.ts
```diff
@@ -0,0 +1,124 @@
...
+ * `getGatewayRealtimeProtocols`, and the Gateway server decodes them with
+ * `getGatewayRealtimeAuthToken` / `getGatewayRealtimeTeamIdOrSlug`.
+ */
+
+/**
+ * Marker subprotocol offered on every handshake so the Gateway can echo a
+ * negotiated subprotocol on the 101 response (some clients require the server to
+ * select one of the offered subprotocols).
+ */
+export const GATEWAY_REALTIME_SUBPROTOCOL = 'ai-gateway-realtime';
```
shaper: potentially add a version in some manner as discussed - can do later as well

packages/gateway/src/gateway-realtime-auth.ts
```diff
@@ -0,0 +1,124 @@
+/**
+ * Shared WebSocket subprotocol contract for AI Gateway realtime auth.
```
shaper, line 2: You mentioned there is a small finite space available - can we state what it is so we have the explicit data size bounds recorded?

packages/gateway/src/gateway-realtime-auth.ts
```diff
@@ -0,0 +1,124 @@
...
+  secWebSocketProtocol: string | null | undefined,
+): string | undefined {
+  const authProtocol = findProtocol(
+    secWebSocketProtocol,
+    GATEWAY_AUTH_SUBPROTOCOL_PREFIX,
+  );
+
+  // `authProtocol` is already trimmed above, so the sliced token needs no
+  // further trimming; `|| undefined` collapses the empty-token case.
+  const token = authProtocol?.slice(GATEWAY_AUTH_SUBPROTOCOL_PREFIX.length);
```
shaper: nit: inline w/ below

packages/gateway/src/gateway-realtime-auth.ts
```diff
@@ -0,0 +1,124 @@
...
+ * empty. The Gateway upgrade handler turns this into an
+ * `Authorization: Bearer <token>` before its normal auth path.
+ *
+ * Accepts the raw header value (subprotocols are comma-separated and may carry
+ * surrounding whitespace).
+ */
+export function getGatewayRealtimeAuthToken(
+  secWebSocketProtocol: string | null | undefined,
+): string | undefined {
+  const authProtocol = findProtocol(
```
shaper: nit: inline w/ below

packages/gateway/src/gateway-realtime-auth.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Server-side: extract the optional Vercel team ID or slug from the
+ * `Sec-WebSocket-Protocol` header value. Team scoping is base64url-encoded so
+ * arbitrary team slugs stay within the WebSocket subprotocol token grammar.
+ */
+export function getGatewayRealtimeTeamIdOrSlug(
+  secWebSocketProtocol: string | null | undefined,
+): string | undefined {
+  const teamProtocol = findProtocol(
```
shaper: nit: inline below

packages/gateway/src/gateway-realtime-model.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  constructor(modelId: string, config: GatewayRealtimeModelConfig) {
+    this.modelId = modelId;
+    this.provider = config.provider;
+    this.config = config;
+  }
+
+  /**
+   * Unlike providers with a dedicated ephemeral-secret endpoint (e.g. OpenAI),
+   * the Gateway has no client-secret mint: the returned token is the Gateway
```
shaper: elaborate on "the returned token is the Gateway auth token" - do we return the same key the user gave us, or a new token Gateway creates on the user's behalf, etc.

packages/gateway/ai-sdk-gateway-4.0.0-canary.99.tgz
shaper, line 1: rm

gr2m (approved):

packages/gateway/src/gateway-provider-options.ts
```diff
@@ -1,118 +1,52 @@
...
+  };
+
+  /** Entity identifier against which quota is tracked. */
+  quotaEntityId?: string;
+
+  /** Unified service tier intent. */
+  serviceTier?: 'flex' | 'priority';
+
+  /** Sort providers by a performance or cost metric before routing. */
+  sort?: 'cost' | 'latency' | 'price' | 'throughput' | 'tps' | 'ttft';
```
shaper: this should not include `price`, `latency` or `throughput`

the formal supported params are `cost`, `tps` and `ttft`
https://vercel.com/docs/ai-gateway/models-and-providers/provider-filtering-and-ordering#sort-by-cost

others if supported are for compat purposes only so would omit from type safety autocomplete

shaper (approved):

packages/gateway/src/gateway-provider.ts
```diff
@@ -444,6 +467,29 @@ export function createGateway(
...
+      getToken: async (tokenOptions: RealtimeFactoryV4GetTokenOptions) => {
+        const model = createRealtimeModel(tokenOptions.model);
+        const secret = await model.doCreateClientSecret();
+        return {
+          token: secret.token,
+          url: secret.url,
+          ...(secret.expiresAt != null && { expiresAt: secret.expiresAt }),
+        };
+      },
+    },
```
gr2m, line 492: we check for `globalThis.window` to throw an error and document that realtime gateway model cannot be used in browsers. 

We already do that check in our code today to generate the user agent string

https://github.com/vercel/ai/blob/0294b585b04daa3136fcd9a0f341578528bf67de/packages/provider-utils/src/get-runtime-environment-user-agent.ts#L5-L7

Issue #10781, opened 2025-12-02
[@ai-sdk/anthropic | @ai-sdk/google] Duplicate text-start/text-end IDs in multi-step flow


abhiaiyer91:
### Description

When using `streamText` with Anthropic models in multi-step agentic flows (TEXT → TOOL → TEXT), the `text-start` and `text-end` chunks produce duplicate IDs across steps.

### Root Cause

In `@ai-sdk/anthropic`, the `text-start` ID is set to `String(value.index)`, where `value.index` is Anthropic's `content_block_start` index:

```
case "text": {
    contentBlocks[value.index] = { type: "text" };
    controller.enqueue({
      type: "text-start",
      id: String(value.index)  // <-- This resets to 0 for each LLM call
    });
}
```
Since Anthropic resets the content block index to 0 for each response, when an agent does:
1. Step 1: Text (id=0) → Tool call
2. Step 2: Text (id=0) ← **Duplicate!**

Both text blocks get `id="0"`.

### Repro

```
import { anthropic } from '@ai-sdk/anthropic';
import { streamText, stepCountIs } from 'ai';
import { z } from 'zod';

const textIds: { type: string; id: string; step: number }[] = [];
let currentStep = 0;

const result = streamText({
    model: anthropic('claude-sonnet-4-5'),
    system: 'First say "Let me check", then call the tool, then summarize.',
    prompt: 'What is the weather?',
    tools: {
        get_weather: {
            description: 'Get the current weather',
            inputSchema: z.object({ city: z.string().optional() }),
            execute: async () => ({ temperature: 72, condition: 'sunny' }),
        },
    },
    stopWhen: stepCountIs(3),
});

for await (const chunk of result.fullStream) {
    if (chunk.type === 'step-start') {
        currentStep++;
    } else if (chunk.type === 'text-start') {
        console.log(`[Step ${currentStep}] text-start id="${chunk.id}"`);
        textIds.push({ type: 'text-start', id: chunk.id, step: currentStep });
    }
}

// Output shows duplicate IDs:
// [Step 1] text-start id="0"
// [Step 2] text-start id="0"  <-- Duplicate!
```

### Expected Behavior
Shouldn't each `text-start` event have a unique ID, even across multiple steps in the same streaming session? This is important for frameworks that use these IDs to track and order message parts.

### AI SDK Version

- @ai-sdk/anthropic version: 2.0.45
- ai version: 5.0.x
- Node.js version: 22.x

### Code of Conduct

[...]

PR #15805, opened 2026-06-03
fix(openai-compatible): use distinct text part IDs across tool call b…

### LINKED ISSUE #15789: @ai-sdk/openai-compatible reuses txt-0 text part ID across TEXT -> TOOL -> TEXT streams
### Description

`@ai-sdk/openai-compatible` appears to reuse the same synthetic text part ID (`txt-0`) for separate text segments in a multi-step tool-calling stream.

When an OpenAI-compatible chat stream produces text before a tool call and then more text after the tool result, the outer stream can contain separate text parts with the same ID:

```text
start
start-step
text-start id=txt-0
text-delta id=txt-0  "I'll check that..."
tool-input-start ...
tool-input-end ...
tool-call ...
text-end id=txt-0
finish-step reason=tool-calls
start-step
text-start id=txt-0
text-delta id=txt-0  "Done, here's the result..."
text-end id=txt-0
finish-step reason=stop
finish
```

This is similar to #10781, but the duplicated ID is not a numeric provider content index such as `0` or `1`. It is the OpenAI-compatible adapter's own synthetic string ID, `txt-0`.

### Why this matters

Some UI/message adapters use text part IDs to preserve and update text parts. When two separate text segments in the same assistant response share `txt-0`, those adapters can merge the text before and after the tool call into one text part.

In an inline tool-call UI, this can visually turn:

```text
[text before tool]
[tool call / tool result]
[text after tool]
```

into:

```text
[text before tool + text after tool]
[tool call / tool result]
```

The tool card then appears to sink below the final assistant text, even though the stream order was text -> tool -> text.

We observed this through Mastra's official React adapter (`@mastra/react` `toUIMessage`), which tracks text parts by `textId`. Mastra already has a workaround for numeric duplicate IDs from Anthropic/Google providers, but it does not catch `txt-0` because this ID is a non-numeric string.

### Current implementation shape

In `packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts` for `@ai-sdk/openai-compatible@2.0.48`, the chat adapter:

- emits `text-start` with `id: 'txt-0'` when the first `delta.content` arrives
- emits all `text-delta` chunks with `id: 'txt-0'`
- emits `text-end` with `id: 'txt-0'` only during stream flush

OpenAI-compatible chat completion deltas do not appear to expose a native content block ID, so a synthetic ID makes sense. The issue is that the same synthetic ID is reused for later model steps / later text segments in the same outer tool-calling stream.

### Expected behavior

Each separate text segment in a single outer `streamText().fullStream` session should get a distinct text part ID, especially across tool-call step boundaries.

For example:

```text
text-start id=txt-0
text-delta id=txt-0
text-end id=txt-0
tool-call ...
text-start id=txt-1
text-delta id=txt-1
text-end id=txt-1
```

or another unique/session-stable ID scheme. The important property is that separate text parts should not share the same ID when downstream UI/persistence layers need to preserve part order.

### Actual behavior

The adapter reuses `txt-0`, so downstream consumers can treat the pre-tool and post-tool text as one text stream.

### Related issues

- #10781 reports duplicate text IDs in multi-step flows for Anthropic/Google provider IDs.
- #10755 reports related `openai-compatible` text/reasoning end timing concerns.
- mastra-ai/mastra#9909 shows the same class of text/tool ordering issue surfacing in Mastra storage/UI.

### Environment

- `ai`: `5.0.193`
- `@ai-sdk/openai-compatible`: `2.0.48`
- Observed with an OpenAI-compatible chat model used through Mastra:
  - `@mastra/core`: `1.37.1`
  - `@mastra/react`: `0.4.2`

### Code of Conduct

SrishtiDev:
The stream adapter was hardcoding 'txt-0' as the text part ID for all text segments. In a TEXT → TOOL → TEXT stream, both text segments got id='txt-0', causing downstream UI adapters to merge them into one block.

Now uses an incrementing textPartIndex counter so each text segment gets a unique ID (txt-0, txt-1, etc.). When a tool call arrives during an active text block, the current text block is properly closed with text-end before the tool call is processed.

Fixes #15789

## Background

Issue #15789 reported that the `@ai-sdk/openai-compatible` adapter hardcoded the string `txt-0` as the ID for all text segments. During a multi-step tool-calling stream (TEXT → TOOL → TEXT), the pre-tool and post-tool text segments shared this exact same ID. Downstream UI adapters that rely on text IDs to preserve rendering order (like Mastra React) would merge the two segments together, disrupting the visual stream sequence.

## Summary

- Modified `packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts` to track an incrementing `textPartIndex`.
- Text chunks (`text-start`, `text-delta`, `text-end`) are now emitted using template literals (`` `txt-${textPartIndex}` ``).
- In the `delta.tool_calls` loop, if there is an active text block, it now properly closes it (enqueuing `text-end`) and increments the `textPartIndex` before processing the tool call, ensuring subsequent text chunks receive a distinct ID (e.g., `txt-1`).
- Added a regression test specifically asserting distinct text part IDs for text streams interrupted by tool calls.

## Manual Verification

Executed the monorepo test suite with `pnpm test`. The new test asserting distinct ID generation (`txt-0`, `txt-1`) correctly passes and matches the generated snapshot, proving that the multi-step stream properly cycles its IDs without breaking existing backwards compatibility. All other core and adapter packages passed without any downstream regressions from this change.

## Checklist

## Future Work

Similar static `txt-0` text part IDs also currently exist in `@ai-sdk/groq` and `@ai-sdk/deepseek`. This exact same logic could be applied there in a future PR to resolve the same class of bugs for those providers.

## Related Issues

Fixes #15789
Related to #10781
```


SrishtiDev:
Hi @lgrammel, following up on this PR, it's been a few weeks with no review yet. The only failing check is the Vercel deployment authorization gate, which I don't have permission to approve.
Happy to make any changes needed  just let me know if there's anything blocking review from my side.

aayush-kapoor (commented):
fixed in https://github.com/vercel/ai/pull/19269

PR #15840, opened 2026-06-05
fix(ai): enforce timeout.stepMs for the whole step in streamText

zirkelc:
## Background

`streamText({ timeout: { stepMs } })` is supposed to abort a step whose model call runs longer than `stepMs`. In practice it never fired for a step that stalls before producing content: a model that opens its stream and then idles (no content part, never closes) was never aborted, and its abort signal stayed unaborted indefinitely. `chunkMs` and `totalMs` worked correctly; only `stepMs` was broken. `generateText` aborts correctly on `stepMs`, so `streamText` should match it.

## Root cause

In `DefaultStreamTextResult.streamStep` (`packages/ai/src/generate-text/stream-text.ts`):

- The step timer is armed at the start of `streamStep` via `setAbortTimeout(...)`.
- The step's stream is registered with `self.addStream(...)`. This is **synchronous** — it hands a lazily-consumed stream to the stitchable stream and returns immediately, before the stream produces or finishes anything.
- `streamStep`'s `try` then ends and its `finally` ran `clearStepTimeout()` / `clearChunkTimeout()` almost immediately, clearing the step timer before it could ever fire.
- The **chunk** timer survived only because `resetChunkTimeout()` is called from inside the stream's `transform` as chunks actually flow, re-arming it after the `finally` cleared it. The step timer had no such re-arm path.

So `stepMs` could only fire in the ~1 ms window between being armed and the `finally` clearing it — effectively a no-op for stalled or slow steps. `generateText` doesn't hit this because it `await`s the whole step inside the `try`, so its `finally` runs only after the step has truly finished.

## Fix

- The early `finally` is now a `catch` that clears the timers **only on setup failure** (before the stream is registered) and rethrows. On success, the stream lifecycle owns the timers: the normal step-finish path (`flush`) clears them as before.
- Both step-scoped timers are also dropped when the merged abort signal fires (any step/chunk/total timeout or a caller abort), via two `abort` listeners — so neither timer outlives the step.

This keeps `chunkMs` / `totalMs` and normal step-finish cleanup working with no leaked timers, and makes `stepMs` enforce the whole step, mirroring `generateText`.

## Tests

Added `should abort when step timeout expires` to the existing `options.timeout` describe block in `stream-text.test.ts`, mirroring the existing `should abort when chunk timeout expires` test (stall the stream, advance fake timers past the timeout, assert the model's abort signal is aborted with a `TimeoutError`).

Why the existing tests missed this: all five existing `stepMs` tests use immediately-completing streams and only assert the abort signal is *forwarded* (defined / same object across steps), never that it actually *fires*. With a normally-completing stream the step timer is cleared on finish regardless, so the early-clear was unobservable. The `chunkMs` subsection had an "abort on stall" test, but `stepMs` had no parallel — the new test closes that gap.

Verified the new test fails on the pre-fix source (`aborted` is `false`) and passes with the fix. Full `stream-text` and `generate-text` suites pass (643 tests).

## Changeset

`patch` for `ai`.


packages/ai/src/generate-text/stream-text.ts
```diff
@@ -2089,9 +2096,12 @@ class DefaultStreamTextResult<
               }),
             ),
           );
-        } finally {
+        } catch (error) {
+          // Setup failed before the stream was registered, so neither the
+          // stream's flush nor an abort will clear the timers — clear them here.
           clearStepTimeout();
           clearChunkTimeout();
```
lgrammel, line 2103: Thanks! Are the timeouts cleared when executing normally, i.e. is clearStepTimeout() called when the step has been finished? are there tests covering that behavior (successful multiplestep within timeouts, without abort triggering where it would if they are not cleared)?

zirkelc: Yes, on normal completion the timers are cleared.
The step transform's `flush` runs when the step's stream closes, and it calls clearStepTimeout() / clearChunkTimeout() for every step before the continue-vs-finish branch

[stream-text.ts](https://github.com/zirkelc/ai/blob/908d428816376a63cab7bae2e66c84c298bd8df4/packages/ai/src/generate-text/stream-text.ts#L2053-L2055)

This `catch` only fires when setup throws before the stream is registered

On tests: you're right that this wasn't covered. The existing multi-step stepMs test only asserts signal identity and never advances fake timers past the timeout, so a leaked timer wouldn't be caught.

I added a two-step run with `stepMs: 50` and assert the (shared) abort signal is not aborted. I verified it fails if the on-finish cleanup is removed (the stale timer fires and aborts) and passes with it in place.

lgrammel (approved):

PR #15845, opened 2026-06-05
feat: agent tui

lgrammel:
## Background

To quickly develop and test agents, it is helpful to have a TUI that requires very limited setup.

## Summary

- integrate https://github.com/lgrammel/agent-tui as `@ai-sdk/tui`
- add documentation

## Example

```ts
await runAgentTUI({ agent });
```

## Manual Verification

- [x]  `pnpm tsx src/agent/openai/tui`


.changeset/calm-bears-help.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ai-sdk/tui": patch
```
aayush-kapoor, line 2: not sure if the process changed but isn't this usually `major` with version in package.json as `0.0.0` ?

packages/tui/package.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "@ai-sdk/tui",
+  "version": "1.0.0-canary.0",
```
aayush-kapoor, line 3: might have to be `0.0.0-canary`

aayush-kapoor (approved):

packages/tui/package.json
lgrammel: lets see what happens, can fix if needed

.changeset/calm-bears-help.md
lgrammel: if we want to bump to 1.0.0 but typically we would do a major for all packages when ai sdk 7 goes ga, which takes care of the version change

PR #15855, opened 2026-06-05
Serialize undefined tool output to null in UI message chunks

### LINKED ISSUE #15854: UI stream can emit invalid `tool-output-available` chunks when a tool returns `undefined`
## Summary

When a tool executes successfully but returns `undefined`, the AI SDK can produce a `tool-output-available` UI stream chunk with `output: undefined`. The SSE serialization path uses `JSON.stringify(part)`, which drops `undefined` object properties, so the actual wire payload has no `output` key. The AI SDK client-side UI message chunk schema requires `output` for `tool-output-available`, so the SDK producer can emit a stream chunk that the SDK consumer rejects.

In our application, this became visible after an upgrade from Zod 3 to Zod 4 because `z.object({ output: z.unknown() }).safeParse({})` now fails, while Zod 3 accepted a missing key for `z.unknown()`. Zod 4 did not create the malformed payload, but it exposed it by enforcing the schema more strictly.

## Observed behavior

In a streaming `useChat`/`DefaultChatTransport` setup, the chat fails as soon as a successful tool result with `undefined` output is received. The client reports a type validation failure for a chunk like:

```json
{"type":"tool-output-available","toolCallId":"call_...","providerMetadata":{"openai":{"itemId":"fc_..."}}}
```

The reported schema error is:

```text
Invalid input: expected nonoptional, received undefined
path: ["output"]
```

Refreshing the page can keep failing if the final UI message was persisted, because persisting the message parts with `JSON.stringify()` also drops `output: undefined`. The saved assistant message later fails `validateUIMessages()` for the same reason.

## Expected behavior

The AI SDK should not emit a UI stream event that is invalid according to its own `uiMessageChunkSchema`. For successful tool results, either `undefined` output should be normalized to a JSON value such as `null`, or `undefined` output should be rejected earlier with a clear error before emitting a `tool-output-available` chunk.

## Why this happens

The current conversion path in the latest repo appears to be:

```ts
// packages/ai/src/ui-message-stream/to-ui-message-chunk.ts
case 'tool-result': {
  return {
    toolCallId: part.toolCallId,
    output: part.output,
    // ...
  };
}
```

The SSE serialization path is:

```ts
// packages/ai/src/ui-message-stream/json-to-sse-transform-stream.ts
controller.enqueue(`data: ${JSON.stringify(part)}\n\n`);
```

If `part.output` is `undefined`, `JSON.stringify()` removes the property:

```js
JSON.stringify({ type: 'tool-output-available', toolCallId: 'call_1', output: undefined })
// => '{"type":"tool-output-available","toolCallId":"call_1"}'
```

The receiving schema requires the key:

```ts
// packages/ai/src/ui-message-stream/ui-message-chunks.ts
z.strictObject({
  type: z.literal('tool-output-available'),
  toolCallId: z.string(),
  output: z.unknown(),
  // ...
})
```

## Zod 4 note

This can look like a Zod 4 regression from the app side, because Zod 4 changed the behavior enough to expose the malformed payload:

```js
const schema = z.object({ output: z.unknown() });
schema.safeParse({}).success;
```

With Zod 3.25.76 this returns `true`. With Zod 4.4.3 this returns `false` with `expected nonoptional`. I think Zod 4 is doing the right thing here: the schema says `output` is required, and the serialized payload does not contain it.

## Minimal self-contained reproduction

This demonstrates the protocol mismatch without needing a model provider:

```ts
import { validateTypes } from '@ai-sdk/provider-utils';
import { uiMessageChunkSchema } from 'ai';

const chunk = {
  toolCallId: 'call_1',
  output: undefined,
};

const serialized = JSON.stringify(chunk);
console.log(serialized);
// {"type":"tool-output-available","toolCallId":"call_1"}

// Throws because `output` is missing.
await validateTypes({
  value: JSON.parse(serialized),
  schema: uiMessageChunkSchema,
});
```

A more realistic reproduction is any streamed tool with an effectful `execute` function that does not return a value:

```ts
import { streamText, stepCountIs, tool } from 'ai';
import { z } from 'zod';

const model = getModelSomehow();

const result = streamText({
  model,
  prompt: 'Call the writeFile tool.',
  stopWhen: stepCountIs(2),
  tools: {
    writeFile: tool({
      inputSchema: z.object({ content: z.string() }),
      execute: ({ content }) => {
        // Side effect only. Returns undefined.
        console.log(content);
      },
    }),
  },
});

```

When the tool succeeds, the UI message stream can contain a `tool-output-available` event whose `output` key is dropped by SSE JSON serialization, and `DefaultChatTransport` rejects it while parsing the stream.

## Possible fixes

In my opinion, the best option would be to normalize an `undefined` tool result output to `null` in `toUIMessageChunk()`:

```ts
case 'tool-result': {
  return {
    toolCallId: part.toolCallId,
    output: part.output === undefined ? null : part.output,
    // ...
  };
}
```

Since data is crossing the wire in JSON format, this feels like a natural workaround to JSON not being able to represent `undefined`.

Alternatively, the SDK could throw a clear error before emitting the chunk if a successful tool result has `undefined` output. This avoids changing semantics, but is much less ergonomic for effectful tools that naturally return nothing: they'd now have to return dummy values. This would be a backwards-incompatible change, and would need to be represented in TypeScript.

### AI SDK Version

- `ai`: observed with 6.0.193 in our app
- `@ai-sdk/react`: observed with 3.0.195 in our app
- `@ai-sdk/provider-utils`: observed with 4.0.27 in our app
- `zod`: 4.4.3

### Code of Conduct

nwalters512:
## Background

See https://github.com/vercel/ai/issues/15854.

## Summary

`toUIMessageChunk(...)` now coerces `undefined` tool output to `null` so that the property survives being round-tripped through JSON.

## Manual Verification

I made the change in my own application's `node_modules` and verified that the issue I described in the linked issue was resolved.

## Checklist

## Future Work

This prevents invalid data from being produced in streams, but doesn't do anything to handle bad data that was already produced. If maintainers would be onboard, I'd like to make a few more changes to make this work seamlessly for people:

- Update the `'tool-output-available'` schema in `uiMessageChunkSchema` to have a `.default(null)` on the `output` schema.
- Update the `'output-available'` schemas in `validateUIMessages` to have a `.default(null)` on the `output` schema
- Update `convertToModelMessages` to pre-normalize `undefined` tool output to `null` before `createToolModelOutput` is called.

I'd be happy to increase the scope of this PR to include this, or move it to its own PR.

As it stands, I've already stored the "bad" data in my application's database, so even if this PR is merged, I'd need application-level compatibility with old messages. It'd be great to be able to shift this into `ai` itself, and I bet others could benefit from this too. If nothing else, it'd help ensure a seamless upgrade path for others looking to move from Zod 3 to Zod 4.

## Related Issues

Fixes https://github.com/vercel/ai/issues/15854

### DIFF
--- a/.changeset/tiny-hornets-cover.md
+++ b/.changeset/tiny-hornets-cover.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+Serialize `undefined` tool output to `null` in UI message chunks
--- a/packages/ai/src/ui-message-stream/to-ui-message-chunk.test.ts
+++ b/packages/ai/src/ui-message-stream/to-ui-message-chunk.test.ts
@@ -4,7 +4,6 @@ import * as z from 'zod/v4';
 import type { GeneratedFile } from '../generate-text';
 import type { TextStreamPart } from '../generate-text/stream-text-result';
 import type { ProviderMetadata } from '../types/provider-metadata';
-import type { UIMessage } from '../ui/ui-messages';
 import { toUIMessageChunk } from './to-ui-message-chunk';
 
 const providerMetadata: ProviderMetadata = {
@@ -501,6 +500,27 @@ describe('toUIMessageChunk', () => {
     expect(
       toUIMessageChunk<Tools>(
         {
+          type: 'tool-result',
+          toolCallId: 'call-undefined',
+          toolName: 'dynamicTool',
+          input: { value: 'input' },
+          output: undefined,
+        },
+        { tools },
+      ),
+    ).toEqual({
+      type: 'tool-output-available',
+      toolCallId: 'call-undefined',
+      // This must be `null` so that we don't lose the property when this is
+      // serialized to JSON. See the following issue for more details:
+      // https://github.com/vercel/ai/issues/15854
+      output: null,
+      dynamic: true,
+    });
+
+    expect(
+      toUIMessageChunk<Tools>(
+        {
           type: 'tool-error',
           toolCallId: 'call-2',
           toolName: 'dynamicTool',
@@ -613,6 +633,8 @@ describe('toUIMessageChunk', () => {
       providerExecuted: true,
     });
   });
+
+  it('maps ');
 
   it('maps error parts through onError', () => {
     const onError = vi.fn(() => 'handled error');
--- a/packages/ai/src/ui-message-stream/to-ui-message-chunk.test.ts
+++ b/packages/ai/src/ui-message-stream/to-ui-message-chunk.test.ts
@@ -634,8 +634,6 @@ describe('toUIMessageChunk', () => {
     });
   });
 
-  it('maps ');
-
   it('maps error parts through onError', () => {
     const onError = vi.fn(() => 'handled error');
     const error = new Error('boom');
--- a/packages/ai/src/ui-message-stream/to-ui-message-chunk.ts
+++ b/packages/ai/src/ui-message-stream/to-ui-message-chunk.ts
@@ -276,7 +276,9 @@ export function toUIMessageChunk<
       return {
         type: 'tool-output-available',
         toolCallId: part.toolCallId,
-        output: part.output,
+        // UI stream chunks are serialized as JSON, which drops undefined
+        // properties. Use null so tool outputs always keep the output field.
+        output: part.output === undefined ? null : part.output,
         ...(part.providerExecuted != null
           ? { providerExecuted: part.providerExecuted }
           : {}),


gr2m (approved):
LGTM. Signed, full CI green, correct undefined→null coercion with regression test.

nwalters512:
@gr2m thanks for merging this! Did you take a look at the "future work" section I had above? Obviously I won't be able to do any of that in this PR now that it's merged, but I'd be happy to open a follow-up if you'd be on board with those changes.

On second thought: I'll just open the PR so this doesn't get lost. If you don't like the changes, let me know or feel free to close it outright!

PR #15858, opened 2026-06-06
feat(gateway): add GatewayExplicitCacheFallbackError

R-Taneja:
## Problem

When the AI Gateway refuses to serve an explicit provider cache reference (e.g. Gemini `cachedContent`) on fallback credentials, it responds with HTTP **424** and `type: "gemini_explicit_cache_fallback_unsupported"`.

The client error mapper in `create-gateway-error.ts` has no `case` for that type, so it falls through to `default` → `GatewayInternalServerError`. That hardcodes `type = "internal_server_error"`, so the real type is lost. Callers see a generic internal-server error and can only distinguish the refusal by the (preserved) 424 status — fragile.

## Fix

Add a dedicated `GatewayExplicitCacheFallbackError`, matching the existing one-class-per-type pattern (`GatewayAuthenticationError`, `GatewayRateLimitError`, etc.), and a `case` for it in `createGatewayErrorFromResponse`.

Now callers can:

```ts
import { GatewayExplicitCacheFallbackError } from '@ai-sdk/gateway';

try {
  await generateText({ model: gateway('google/gemini-2.5-flash'), /* ...cachedContent */ });
} catch (error) {
  if (GatewayExplicitCacheFallbackError.isInstance(error)) {
    // retry without cachedContent + include the full inline prompt
  }
}
```

- `error.type === 'gemini_explicit_cache_fallback_unsupported'`
- `error.statusCode === 424`
- `error.isRetryable === false` (derived from the status code by the base class)

## Notes

- Only adds a new `case` + exported class — the `default` path and all existing behavior/tests are untouched.
- Pairs with the Gateway-side refusal (vercel/ai-gateway#2177), which is what emits this error type.

## Tests

- New case in `create-gateway-error.test.ts` asserting the type, message, 424 status, and `isRetryable: false`.
- Full gateway suite green (429 node + 429 edge), type-check and lint clean.

### DIFF
--- a/.changeset/gateway-explicit-cache-fallback-error.md
+++ b/.changeset/gateway-explicit-cache-fallback-error.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+Add `GatewayExplicitCacheFallbackError` so a Gateway refusal to serve an explicit provider cache reference (e.g. Gemini `cachedContent`) on fallback credentials surfaces as a typed, non-retryable error instead of collapsing into `GatewayInternalServerError`.
--- a/packages/gateway/src/errors/create-gateway-error.test.ts
+++ b/packages/gateway/src/errors/create-gateway-error.test.ts
@@ -6,6 +6,7 @@ import {
   GatewayRateLimitError,
   GatewayModelNotFoundError,
   GatewayInternalServerError,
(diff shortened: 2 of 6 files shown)


Prana111025 (commented):
Fallback error

PR #15866, opened 2026-06-06
fix/fix-json-partial-unicode-escape

### LINKED ISSUE #15865: `fixJson` emits invalid JSON when a streamed value is truncated inside a `\uXXXX` unicode escape
### Description

`fixJson` (`packages/ai/src/util/fix-json.ts`) repairs partial JSON during streaming so that `parsePartialJson` can recover a value from each incremental chunk. It correctly handles a truncated escape that ends on a lone backslash (`"abc\` → `"abc"`), but it does **not** handle a truncation inside a unicode escape sequence (`\uXXXX`). When the input ends after `\u` but before all four hex digits are present, `fixJson` closes the string with the partial escape still in place, producing **invalid JSON**.

As a result, `parsePartialJson` returns `failed-parse` and the partial value is dropped for that chunk — the object/field flickers or disappears mid-stream until the next chunk completes the escape.

## Reproduction

```ts
import { fixJson } from 'ai/src/util/fix-json'; // internal util

fixJson('{"a":"\\u');      // => '{"a":"\\u"}'   ❌ JSON.parse throws
fixJson('{"a":"\\u12');    // => '{"a":"\\u12"}' ❌ JSON.parse throws
fixJson('{"a":"text \\u00'); // => '{"a":"text \\u00"}' ❌ JSON.parse throws
```

Each repaired string fails `JSON.parse` with `Bad Unicode escape`. Compare with the trailing-backslash case, which is handled correctly:

```ts
fixJson('"value with \\'); // => '"value with "'  ✅
```

Reachable in practice through `streamObject`, structured output, and streaming tool-call inputs: any time a provider's SSE chunk boundary lands between `\u` and its four hex digits (common for emoji/CJK content, which the SDK already escapes as `😀`-style surrogate pairs).

## Root cause

In the `INSIDE_STRING_ESCAPE` state, every escaped character — including `u` — immediately advances `lastValidIndex`:

```ts
case 'INSIDE_STRING_ESCAPE': {
  stack.pop();
  lastValidIndex = i; // marks `\u` (and any following hex digits) as "valid"
  break;
}
```

The scanner has no concept that `\u` requires exactly four following hex digits, so a truncated escape is treated as complete and included in the repaired output.

## Expected behavior

`fixJson` should treat a `\uXXXX` escape as valid only once all four hex digits are present. If the input ends earlier, the incomplete escape should be dropped (consistent with how the trailing-backslash case is already handled), so the repaired output is always valid JSON.

```ts
fixJson('{"a":"\\u');   // => '{"a":""}'   ✅
fixJson('{"a":"\\u12'); // => '{"a":""}'   ✅
fixJson('"abc\\u1234"'); // => '"abc\\u1234"' (complete escape preserved) ✅
```

## Environment

- Affected package: `ai` (`fixJson`, consumed by `parsePartialJson` → `streamObject` / structured output / streaming tool inputs)
- Reproduced on latest `main`.

### AI SDK Version

 ai@7.0.0-canary.165

### Code of Conduct

Abuhaithem:
`fixJson` repairs partial JSON during streaming so `parsePartialJson` can recover a
value from each incremental chunk (used by `streamObject`, structured output, and
streaming tool-call inputs). It already handles a string truncated on a lone
backslash (`"abc\` → `"abc"`), but it did **not** handle truncation inside a
`\uXXXX` unicode escape. Given input ending after `\u`/`\u1`/`\u12`/`\u123`, it closed
the string with the partial escape in place and produced **invalid JSON**:

```ts
fixJson('{"a":"\\u');   // before: '{"a":"\\u"}'   → JSON.parse throws
fixJson('{"a":"\\u12'); // before: '{"a":"\\u12"}' → JSON.parse throws
```

`parsePartialJson` then returned `failed-parse`, so the partial value was dropped for
that chunk — fields flicker/disappear mid-stream whenever a provider's SSE chunk
boundary lands inside an escaped character (common for emoji/CJK content, which is
streamed as `\uXXXX` surrogate pairs).

## Summary

- Model the unicode escape explicitly in the scanner with a new
  `INSIDE_STRING_UNICODE_ESCAPE` state and a `unicodeEscapeRemaining` hex-digit counter.
- A `\uXXXX` escape is only marked valid once all four hex digits arrive; if the input
  ends earlier, the partial escape is dropped (consistent with the existing
  trailing-backslash repair), so the repaired output is always valid JSON.
- Complete escapes (`"A"`, `"abcሴdef"`) are preserved unchanged, and
  already-invalid inputs (e.g. `{"a":"\u1F"}`) produce byte-for-byte identical output
  to before.

```ts
fixJson('{"a":"\\u');   // after: '{"a":""}'   ✅
fixJson('{"a":"\\u12'); // after: '{"a":""}'   ✅
```

## Manual Verification

```
pnpm --filter ai exec vitest --config vitest.node.config.js --run \
  src/util/fix-json.test.ts src/util/parse-partial-json.test.ts   # 53 passed
```

Additional checks:

- `npx ultracite check packages/ai/src/util/fix-json.ts packages/ai/src/util/fix-json.test.ts` — clean.
- `npx tsc --build packages/ai` — exit 0.

## Checklist

- [ ] Documentation has been added / updated (for bug fixes / features) — n/a, internal util
- [x] A _patch_ changeset for relevant packages has been added (run `pnpm changeset` in the project root)

## Related Issues

Fixes #15865

### DIFF
--- a/.changeset/fix-json-partial-unicode-escape.md
+++ b/.changeset/fix-json-partial-unicode-escape.md
@@ -1,0 +1,7 @@
+---
+'ai': patch
+---
+
+fix (ai): `fixJson` now drops incomplete `\uXXXX` unicode escape sequences instead of producing invalid JSON
+
+When a streamed JSON value was truncated in the middle of a unicode escape (e.g. `{"a":"\u12`), `fixJson` previously closed the string off as `{"a":"\u12"}`, which is invalid JSON. `parsePartialJson` then returned `failed-parse` and the partial value was dropped for that chunk. The repaired output is now valid JSON (`{"a":""}`), so partial values continue to stream through.
--- a/packages/ai/src/util/fix-json.ts
+++ b/packages/ai/src/util/fix-json.ts
@@ -3,6 +3,7 @@ type State =
   | 'FINISH'
   | 'INSIDE_STRING'
   | 'INSIDE_STRING_ESCAPE'
+  | 'INSIDE_STRING_UNICODE_ESCAPE'
   | 'INSIDE_LITERAL'
   | 'INSIDE_NUMBER'
   | 'INSIDE_OBJECT_START'
@@ -28,6 +29,8 @@ export function fixJson(input: string): string {
   const stack: State[] = ['ROOT'];
   let lastValidIndex = -1;
   let literalStart: number | null = null;
+  // Number of hex digits still expected to complete the current \uXXXX escape.
+  let unicodeEscapeRemaining = 0;
 
   function processValueStart(char: string, i: number, swapState: State) {
     {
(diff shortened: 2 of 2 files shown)


aayush-kapoor (commented):
marking as closed in favour of https://github.com/vercel/ai/pull/16230 

(missing signed commits / tests / changeset or combination of those)

PR #15876, opened 2026-06-07
fix(ai): handle partial unicode escapes in fixJson

### LINKED ISSUE #15865: `fixJson` emits invalid JSON when a streamed value is truncated inside a `\uXXXX` unicode escape

slegarraga:
## Summary
- Track JSON unicode string escapes until all four hex digits are present.
- Drop incomplete \uXXXX escapes from repaired partial JSON instead of emitting invalid JSON.
- Add regression coverage for incomplete and complete unicode escapes.

Closes #15865.

## Validation
- corepack pnpm exec vitest --config vitest.node.config.js --run src/util/fix-json.test.ts
- corepack pnpm exec vitest --config vitest.edge.config.js --run src/util/fix-json.test.ts
- corepack pnpm type-check
- corepack pnpm exec ultracite check packages/ai/src/util/fix-json.ts packages/ai/src/util/fix-json.test.ts

### DIFF
--- a/packages/ai/src/util/fix-json.test.ts
+++ b/packages/ai/src/util/fix-json.test.ts
@@ -80,6 +80,16 @@ describe('string', () => {
       '"value with unicode \u003C"',
     );
   });
+
+  test('should handle incomplete unicode escape sequences', () => {
+    assert.strictEqual(fixJson('"value with \\u'), '"value with "');
+    assert.strictEqual(fixJson('"value with \\u12'), '"value with "');
+    assert.strictEqual(fixJson('{"value":"text \\u00'), '{"value":"text "}');
+  });
+
+  test('should preserve complete unicode escape sequences', () => {
+    assert.strictEqual(fixJson('"value with \\u003C"'), '"value with \\u003C"');
+  });
 });
 
 describe('array', () => {
--- a/packages/ai/src/util/fix-json.ts
+++ b/packages/ai/src/util/fix-json.ts
@@ -3,6 +3,7 @@ type State =
   | 'FINISH'
   | 'INSIDE_STRING'
   | 'INSIDE_STRING_ESCAPE'
+  | 'INSIDE_STRING_UNICODE_ESCAPE'
   | 'INSIDE_LITERAL'
   | 'INSIDE_NUMBER'
   | 'INSIDE_OBJECT_START'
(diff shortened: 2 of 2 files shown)


aayush-kapoor (commented):
marking as closed in favour of https://github.com/vercel/ai/pull/16230 

(missing signed commits / tests / changeset or combination of those)

PR #7240, opened 2025-07-11
bugfix(openai-compatible): Zod  schema validation for `choices[].delta.tool_calls[].index` to be an optional parameter to support SambaNova

cdxker:
## Background

Sambanova models do not return `choices[].delta.tool_calls[].index`  (larger example of the error is in #7238)

## Summary

Modified the validation logic for tool_calls for the `index` parameter to be optional.

## Verification

- [ ] Added a test

## Tasks
- [ ] Add test validating that `choices[].delta.tool_calls[].index` is not required for validation to work

## Related Issues

Fixes #7238

PR #15879, opened 2026-06-07
fix(openai-compatible): buffer no-index tool call deltas

slegarraga:
## Summary
- map OpenAI-compatible streaming tool-call deltas without `index` to a stable synthetic index by provider `id`
- lets the existing pending-name buffer handle providers that stream arguments before `function.name`
- adds a regression test for an SGLang-style `tool_calls[]` stream matching #5885
- adds a patch changeset for `@ai-sdk/openai-compatible`

## Context
Fixes #5885.

This is a follow-up to the existing fixes for optional `tool_calls[].index` (#7240) and delayed `function.name` (#14760). The combined shape still failed because no-index deltas bypassed the pending-name buffer and went directly to `StreamingToolCallTracker`, which rejects a new tool call before `function.name` is available.

## Verification
- `pnpm --filter @ai-sdk/openai-compatible test:node -- src/chat/openai-compatible-chat-language-model.test.ts`
- `pnpm --filter @ai-sdk/openai-compatible test:edge -- src/chat/openai-compatible-chat-language-model.test.ts`
- `pnpm --filter @ai-sdk/openai-compatible type-check`


gr2m:
closed via #7240

PR #15885, opened 2026-06-08
fix(ai): keep chat finish tied to request

### LINKED ISSUE #15870: useChat: makeRequest finally throws "Cannot read properties of undefined (reading 'state')" when overlapping resume-streams clear this.activeResponse
### Summary

In `Chat.makeRequest` (the `AbstractChat`/`Chat` class behind `useChat`), the `finally` block dereferences `this.activeResponse.state.message` **without a null guard**, while the very next field guards `this.activeResponse?.state.finishReason` with optional chaining. When two `makeRequest` invocations overlap, the first to reach `finally` sets `this.activeResponse = void 0`, and the second's `finally` then throws:

```
    at Chat.makeRequest (ai/dist/index.mjs)
    at async AbstractChat.resumeStream (ai/dist/index.mjs)
```

The error is **caught by the SDK's own inner `try/catch` and `console.error`'d**, so it is non-fatal — but it surfaces a noisy `TypeError` in the console on every overlap.

### Version

- Reproduced on `ai@6.0.182`.
- **Still present in `ai@6.0.197`** (latest at time of writing): `dist/index.mjs` `Chat.makeRequest` `finally` block — `message: this.activeResponse.state.message` is unguarded; `finishReason: (… == null ? void 0 : ….state.finishReason)` is guarded. (3 unguarded `this.activeResponse.state.message` occurrences, 0 guarded.)

### The offending code (abridged, from `Chat.makeRequest`)

```js
} finally {
  try {
    this.onFinish?.call(this, {
      message: this.activeResponse.state.message,           // ← unguarded — throws if activeResponse was cleared
      messages: this.state.messages,
      isAbort, isDisconnect, isError,
      finishReason: this.activeResponse?.state.finishReason, // ← guarded with ?.
    });
  } catch (err) {
    console.error(err);
  }
  this.activeResponse = void 0;                              // ← a concurrent call's finally cleared it
}
```

### How it triggers

`this.activeResponse` is a single shared instance field. Any two overlapping `makeRequest` runs race on it:

1. Call A: `this.activeResponse = A`, awaits its stream.
2. Call B: `this.activeResponse = B`, awaits its stream.
3. Whichever finishes first runs `finally` → reads/clears, sets `this.activeResponse = void 0`.
4. The other's `finally` then reads `this.activeResponse.state.message` → `this.activeResponse` is `undefined` → throw.

The easiest way to hit this in a real app: **`useChat({ resume: true })` under React StrictMode** (default in Next.js dev). StrictMode double-invokes the mount effect, firing two `resumeStream()` → two overlapping `makeRequest({ trigger: 'resume-stream' })` against the same in-flight server stream. It can also occur in production when a programmatic `resumeStream()` (e.g. after a tool/approval resolution) overlaps an already-streaming run.

### Repro sketch

```tsx
// Next.js dev (reactStrictMode: true), a conversation with an in-flight/resumable server stream:
useChat({ id, transport, messages: initial, resume: true });
// On mount, StrictMode fires the resume effect twice → two overlapping resume-streams →
// console: TypeError: Cannot read properties of undefined (reading 'state')
```

### Suggested fix

Guard the `message` field the same way `finishReason` is already guarded:

```js
message: this.activeResponse?.state.message,
```

Or, more robustly, capture the local `activeResponse` (the `const` created earlier in the method) and reference that in `finally` instead of the shared `this.activeResponse`, so a concurrent run can't clobber it.

### Impact

Low/benign (SDK swallows it), but it produces a recurring console `TypeError` that looks alarming during debugging and can mask real errors. Thanks!

he-yufeng:
## Summary

Fixes #15870.

`AbstractChat.makeRequest` keeps a mutable `this.activeResponse` for the currently streaming request. When two requests overlap, an older request can finish first and clear that field while the newer request is still running. The newer request then reaches `finally` and tries to read `this.activeResponse!.state.message`, which throws and gets logged as a noisy `TypeError`.

This keeps each request's response state in a local `response` reference and only clears `this.activeResponse` if it still points to the same request. That preserves the current shared field for abort/update paths without letting an older request tear down a newer one.

## Test plan

- [x] `pnpm exec oxfmt --check packages/ai/src/ui/chat.ts packages/ai/src/ui/chat.test.ts`
- [x] `pnpm exec oxlint packages/ai/src/ui/chat.ts packages/ai/src/ui/chat.test.ts`
- [x] `pnpm --filter ai type-check`
- [x] `pnpm test:node -- src/ui/chat.test.ts`

### DIFF
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -17,6 +17,8 @@ import { lastAssistantMessageIsCompleteWithApprovalResponses } from './last-assi
 import { lastAssistantMessageIsCompleteWithToolCalls } from './last-assistant-message-is-complete-with-tool-calls';
 import type { UIMessage } from './ui-messages';
 
+type TextUIPart = Extract<UIMessage['parts'][number], { type: 'text' }>;
+
 class TestChatState<
   UI_MESSAGE extends UIMessage,
 > implements ChatState<UI_MESSAGE> {
@@ -68,6 +70,33 @@ class TestChat extends AbstractChat<UIMessage> {
 
 function formatChunk(part: UIMessageChunk) {
   return `data: ${JSON.stringify(part)}\n\n`;
+}
+
+function createGatedChunkStream({
+  gate,
+  text,
+}: {
+  gate: Promise<void>;
+  text: string;
+}) {
+  return new ReadableStream<UIMessageChunk>({
+    async start(controller) {
+      await gate;
+
+      controller.enqueue({ type: 'start' });
+      controller.enqueue({ type: 'start-step' });
+      controller.enqueue({ type: 'text-start', id: 'text-1' });
+      controller.enqueue({
+        type: 'text-delta',
+        id: 'text-1',
+        delta: text,
+      });
+      controller.enqueue({ type: 'text-end', id: 'text-1' });
+      controller.enqueue({ type: 'finish-step' });
+      controller.enqueue({ type: 'finish', finishReason: 'stop' });
+      controller.close();
+    },
+  });
 }
 
 const server = createTestServer({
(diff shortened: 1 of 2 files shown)


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a build or test failure from this branch. The status target is ercel.com/git/authorize for the Vercel Labs team, which I cannot authorize from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants the preview disabled for this PR.

he-yufeng:
Bumping this. It keeps the chat `finish` tied to its originating request, so a superseded/stale request can't fire the finish callback for a newer one. Contained. Glad to rebase if you'd like to take a look.

n1ckoates:
Superseded by https://github.com/vercel/ai/pull/16620

PR #15886, opened 2026-06-08
fix(react): resubscribe when chat instance changes

### LINKED ISSUE #15873: subscribeToMessages fails to resubscribe when options.chat changes to a new instance with the same id
### Description

I noticed a potential gap in the subscription logic in use-chat.ts.

At [L97](https://github.com/vercel/ai/blob/main/packages/react/src/use-chat.ts#L97), chatRef is correctly updated when options.chat is replaced with a new instance, using reference equality (!==). However, subscribeToMessages at [L111](https://github.com/vercel/ai/blob/main/packages/react/src/use-chat.ts#L111) only depends on chatRef.current.id in its useCallback dependency array, with an explicit eslint-disable comment.

This means that when options.chat changes to a different instance with the same id, chatRef.current is updated to the new instance, but subscribeToMessages is not recreated. The callback retains a stale closure over the previous chatRef.current, so calling it will invoke ~registerMessagesCallback on the old chat instance rather than the new one.

## Question

The eslint-disable comment indicates that using chatRef.current.id was an intentional choice. I'd like to understand whether this behavior is by design:

If intentional, what is the expected contract for callers who might replace options.chat with a same-ID instance (e.g., after a reconnection or state hydration)? Should they avoid this pattern, or provide a changed id to force re-subscription?

If this is an oversight, is there openness to addressing it (e.g., with a version counter or updating the dependency array)?

### AI SDK Version

- ai: 6.0.197
- @ai-sdk/react: 3.0.199

### Code of Conduct

he-yufeng:
## Summary

Fixes #15873.

`useChat` recreates `chatRef.current` when callers pass a new `chat` instance, but `subscribeToMessages` only depended on the chat id. If the replacement instance had the same id, `useSyncExternalStore` kept the old subscription and later message updates on the new instance were not observed.

This binds the subscription and snapshots to the current `chat` instance for the render. Same-id replacements now change the subscription function identity, so React unsubscribes from the previous instance and subscribes to the new one.

## Test plan

- [x] `pnpm exec oxfmt --check packages/react/src/use-chat.ts packages/react/src/use-chat.ui.test.tsx`
- [x] `pnpm exec oxlint packages/react/src/use-chat.ts packages/react/src/use-chat.ui.test.tsx`
- [x] `pnpm --filter @ai-sdk/react type-check`
- [x] `pnpm test -- use-chat.ui.test.tsx`

### DIFF
--- a/packages/react/src/use-chat.ts
+++ b/packages/react/src/use-chat.ts
@@ -103,30 +103,30 @@ export function useChat<UI_MESSAGE extends UIMessage = UIMessage>({
       'chat' in options ? options.chat : new Chat(optionsWithCallbacks);
   }
 
+  const chat = chatRef.current;
+
   const subscribeToMessages = useCallback(
     (update: () => void) =>
-      chatRef.current['~registerMessagesCallback'](update, throttleWaitMs),
-    // `chatRef.current.id` is required to trigger re-subscription when the chat ID changes
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [throttleWaitMs, chatRef.current.id],
+      chat['~registerMessagesCallback'](update, throttleWaitMs),
+    [chat, throttleWaitMs],
   );
 
   const messages = useSyncExternalStore(
     subscribeToMessages,
-    () => chatRef.current.messages,
-    () => chatRef.current.messages,
+    () => chat.messages,
+    () => chat.messages,
   );
 
   const status = useSyncExternalStore(
-    chatRef.current['~registerStatusCallback'],
-    () => chatRef.current.status,
-    () => chatRef.current.status,
+    chat['~registerStatusCallback'],
+    () => chat.status,
+    () => chat.status,
   );
 
   const error = useSyncExternalStore(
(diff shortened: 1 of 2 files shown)


he-yufeng:
The remaining red Vercel status is the repository-side preview authorization gate, not a build or test failure from this branch. The status target is ercel.com/git/authorize for the Vercel Labs team, which I cannot authorize from the fork.

No branch changes from me here unless a code-owned check fails or a maintainer wants the preview disabled for this PR.

he-yufeng:
Bumping this. The React hook still doesn't resubscribe when the `chat` instance identity changes, so swapping the chat leaves the hook streaming into the previous subscription. Small and contained. Glad to rebase onto main or trim it to a minimal repro if that makes review easier.

he-yufeng:
Second bump. On current main the hook still does not resubscribe when the chat instance identity changes, so a swapped chat object keeps streaming from the old one. Happy to rebase if that helps review.

he-yufeng:
Superseded by the use-chat refactor on main (chat resubscribes on instance change now). Closing.

PR #15907, opened 2026-06-08
fix(gateway): add missing 'caching' property to GatewayProviderOptions type

narutamaaurum:
Fixes #14595

The  type was missing the  property that is documented in the AI Gateway automatic caching documentation (https://vercel.com/docs/ai-gateway/models-and-providers/automatic-caching).

When users tried to use the documented  option with TypeScript, they got a type error:
```
Object literal may only specify known properties, and 'caching' does not exist in type '{ only?: string[] | undefined; order?: string[] | undefined; ... }'.
```

This PR adds  to the Zod schema in , along with a type test to ensure the property is accepted.

## Changes
- Added  field to  Zod schema with proper JSDoc documentation
- Added type test to verify the property works with `satisfies GatewayProviderOptions`

## Testing
- All existing tests pass (429 tests in gateway package)
- Type-check passes
- New type test validates the fix

### DIFF
--- a/packages/gateway/src/gateway-provider-options.ts
+++ b/packages/gateway/src/gateway-provider-options.ts
@@ -111,6 +111,15 @@ const gatewayProviderOptions = lazySchema(() =>
        * default.
        */
       serviceTier: z.enum(['flex', 'priority']).optional(),
+      /**
+       * Enable automatic caching for the request.
+       *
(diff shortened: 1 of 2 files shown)


narutamaaurum:
This PR is obsolete - the  property was already added to  in main via #15768. Closing as duplicate of #15925 (which is also closed).

PR #15917, opened 2026-06-08
Support toModelOutput in WorkflowAgent

gr2m:
## Background

`WorkflowAgent` did not honor tool `toModelOutput`, unlike core text generation.

## Summary

- route WorkflowAgent tool results through model-output conversion
- keep raw tool output for UI/results/callbacks
- add next-workflow example + docs

## Manual Verification

Tested examples/next-workflow end-to-end

## Checklist

### DIFF
--- a/packages/ai/src/prompt/convert-to-language-model-prompt.ts
+++ b/packages/ai/src/prompt/convert-to-language-model-prompt.ts
@@ -17,7 +17,6 @@ import {
   type ReasoningFilePart,
   type ReasoningPart,
   type TextPart,
-  type Tool,
   type ToolCallPart,
   type ToolResultOutput,
   type ToolResultPart,
@@ -32,7 +31,6 @@ import type { Warning } from '../types/warning';
 import { InvalidMessageRoleError } from './invalid-message-role-error';
 import type { StandardizedPrompt } from './standardize-prompt';
 import { MissingToolResultsError } from '../error/missing-tool-result-error';
-import { createToolModelOutput } from './create-tool-model-output';
 
 export async function convertToLanguageModelPrompt({
   prompt,
@@ -428,7 +426,7 @@ function convertImagePartToFilePart(
 /**
  * Downloads files from URLs in the user messages.
  */
-async function downloadAssets(
+export async function downloadAssets(
   messages: ModelMessage[],
   download: DownloadFunction,
   supportedUrls: Record<string, RegExp[]>,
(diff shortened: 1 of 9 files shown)


packages/ai/src/prompt/create-tool-model-output.ts
```diff
@@ -21,7 +21,8 @@ export async function createToolModelOutput({
   }
 
   if (tool?.toModelOutput) {
-    return await tool.toModelOutput({ toolCallId, input, output });
+    const { toModelOutput } = tool;
+    return await toModelOutput({ toolCallId, input, output });
```
gr2m: this-binding guard, not really needed but no harm either. Technically a breaking change but very unlikely to affect anyone. And we don't want future users to depend on this binding behavior

packages/ai/src/prompt/convert-to-language-model-prompt.ts
```diff
@@ -592,7 +594,64 @@ function convertPartToLanguageModelPart(
   };
 }
 
-function mapToolResultOutput({
+export async function createLanguageModelToolResultOutput({
```
gr2m: currently only used by WorkflowAgent, will move to workflow package

WorkflowAgent assembles a provider-level (`LanguageModelV4`) prompt incrementally, appending one tool result at a time, instead of building ai-level ModelMessages and converting the whole prompt once at the end.

packages/workflow/src/create-language-model-tool-result-output.ts
```diff
@@ -0,0 +1,85 @@
...
+ * equivalent per-result conversion using the shared `ai/internal` primitives:
+ *
+ *   1. `createToolModelOutput` — applies `tool.toModelOutput` (or the
+ *      text/json/error fallback).
+ *   2. `downloadAssets` — for `content`-type outputs, downloads any file/image
+ *      assets so URLs become bytes the provider can consume.
+ *   3. `mapToolResultOutput` — maps the AI-level `ToolResultOutput` to the
+ *      provider-level output and converts legacy file types.
+ */
+export async function createLanguageModelToolResultOutput({
```
gr2m, line 30: Might worth moving to AI SDK core as part of https://github.com/vercel/ai/issues/13570

packages/ai/src/prompt/create-tool-model-output.ts
lgrammel: wait wouldnt this break toModelOutput if it relies on `this`?

gr2m: yes, but we never claimed to support this binding, and it's not tested.
I added the same this-binding guard to workflow agent

https://github.com/vercel/ai/blob/16c454f06a9e2d17c03d23c1bce2c05efda5e87d/packages/workflow/src/workflow-agent.ts#L2626-L2631

gr2m: after discussion: 

- revert this change
- add test case with tool implemented as class that relies on this binding for execute and toModelOutput

PR #15922, opened 2026-06-09
fix(openai): throw on early stream error events

kkdawkins:
## Background

OpenAI can return HTTP 200 for streaming requests and then emit an error frame before any model output, e.g. `insufficient_quota`. Previously AI SDK surfaced those as stream `error` parts, so retry/fallback logic never saw a failed `doStream()` call.

Research notes:
- OpenAI documents `error` and `response.failed` as Responses stream events with `sequence_number`.
- OpenAI does not document a guarantee that these only occur as chunk 1.
- Official OpenAI Node/Python SDKs throw on streamed top-level `error` frames during SSE iteration.
- Existing AI SDK fixtures show `response.created` / `response.in_progress` before `error`, so this PR treats errors as retryable only before real output starts, not literally only first chunk.

## Summary

- Added OpenAI early stream error detection for Responses, Chat Completions, and legacy Completions.
- Converts pre-output OpenAI stream errors into `APICallError`, preserving retryability for 429/5xx.
- Keeps late errors after real output as streamed error parts.
- Supports both nested and documented top-level Responses `error` event shapes.
- Preserves numeric HTTP status codes from provider error frames.
- Added a patch changeset for `@ai-sdk/openai`.

## Manual Verification

No live OpenAI quota failure was triggered manually; that path is difficult to reproduce safely. Behavior was verified with captured/synthetic stream fixtures covering early quota errors, `response.failed`, late stream errors, raw chunks, top-level error shape, and numeric status codes.

Automated verification:
- `pnpm type-check` in `packages/openai`
- touched-file `oxlint`
- `pnpm test` in `packages/openai` Node + Edge

Root `pnpm type-check:full` was attempted and failed in unrelated existing `packages/tui` / TUI example type errors.

## Checklist

## Future Work

AI Gateway should get this behavior once it bumps the relevant `@ai-sdk/openai` alias (for current v4 routes, `@ai-sdk/openai-v4`). Gateway already relies on provider `doStream()` throwing for Anthropic early stream errors, so no Gateway stream-parser change should be required for that path.

If any affected traffic is still routed through older Gateway SDK aliases (`@ai-sdk/openai` v2 or `@ai-sdk/openai-v3`), this fix would need a backport or equivalent Gateway-local handling for those versions.

## Related Issues

None.

### DIFF
--- a/packages/openai/src/openai-stream-error.ts
+++ b/packages/openai/src/openai-stream-error.ts
@@ -1,0 +1,254 @@
+import { APICallError } from '@ai-sdk/provider';
+import type { ParseResult } from '@ai-sdk/provider-utils';
+
+type StreamError = {
+  message: string;
+  code?: string | number | null;
+  type?: string | null;
+  param?: unknown;
+  sequenceNumber?: unknown;
+  frame: unknown;
+};
+
+export async function throwIfOpenAIStreamErrorBeforeOutput<T>({
+  stream,
+  getError,
+  isOutputChunk,
+  url,
+  requestBodyValues,
+  responseHeaders,
+}: {
+  stream: ReadableStream<ParseResult<T>>;
+  getError: (chunk: T) => unknown | undefined;
+  isOutputChunk: (chunk: T) => boolean;
+  url: string;
+  requestBodyValues: unknown;
+  responseHeaders?: Record<string, string>;
+}): Promise<ReadableStream<ParseResult<T>>> {
+  const [streamForEarlyError, streamForConsumer] = stream.tee();
+  const reader = streamForEarlyError.getReader();
+
+  try {
+    while (true) {
+      const result = await reader.read();
+
+      if (result.done) {
+        return streamForConsumer;
+      }
+
+      const chunk = result.value;
+
+      if (!chunk.success) {
+        return streamForConsumer;
+      }
+
+      const errorFrame = getError(chunk.value);
+
+      if (errorFrame != null) {
+        streamForConsumer.cancel().catch(() => {});
+        throw createOpenAIStreamError({
+          frame: errorFrame,
+          url,
+          requestBodyValues,
+          responseHeaders,
+        });
+      }
+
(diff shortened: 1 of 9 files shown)


packages/openai/src/responses/openai-responses-api.ts
```diff
@@ -1038,14 +1039,21 @@ export const openaiResponsesChunkSchema = lazySchema(() =>
       }),
       z.object({
         type: z.literal('error'),
-        sequence_number: z.number(),
+        sequence_number: z.number().nullish(),
```
shaper: do we know if we need these changes? in particular prefer to avoid `any` if we can. I realize we may not have a concrete example to go from.

packages/openai/src/openai-stream-error.ts
```diff
@@ -0,0 +1,254 @@
...
+  ) {
+    return 400;
+  }
+  if (discriminator.includes('overload')) return 503;
+  if (discriminator.includes('timeout')) return 504;
+
+  return 500;
+}
+
+function debugOpenAIStreamError({
```
shaper: ai sdk generally doesn't commit debug logging - defer to @gr2m on thought here since I know this can be useful for mid-stream error complexity though ongoing.

kkdawkins: Removed debug logging!

packages/openai/src/responses/openai-responses-api.ts
```diff
@@ -1040,12 +1041,19 @@ export const openaiResponsesChunkSchema = lazySchema(() =>
...
           param: z.string().nullish(),
         }),
       }),
+      z.object({
+        type: z.literal('error'),
+        sequence_number: z.number(),
+        code: z.string().nullish(),
+        message: z.string(),
+        param: z.string().nullish(),
+      }),
```
gr2m: need  a bit more time to understand this change, but can do post merge if this PR is needed to unblock other work

gr2m (approved):

PR #15923, opened 2026-06-09
doc: fix broken links to language model v2 specification

narutamaaurum:
Fix broken links in the language model middleware documentation that pointed to the v5 branch instead of main.

The links now point to the current main branch version of the language-model-v2.ts file.

Fixes #12617


gr2m:
Closing in favor of #16487, which supersedes this fix for #12617 by pointing the middleware docs at the current v4 language model spec and updating the v4 middleware reference.

PR #15925, opened 2026-06-09
fix(gateway): add missing 'caching' property to GatewayProviderOptions type

narutamaaurum:
Closes #14595

## Summary
The documentation for automatic caching in AI Gateway suggests using  in provider options, but the TypeScript type  did not include this property, causing a type error.

## Changes
- Added  to the  Zod schema in 
- Updated documentation () with:
  - Description of the  option
  - Usage example showing  in providerOptions

## Testing
- All existing tests pass (429 tests in gateway package)
- TypeScript type-check passes
- The type now correctly accepts  in

### DIFF
--- a/content/docs/03-ai-sdk-core/40-middleware.mdx
+++ b/content/docs/03-ai-sdk-core/40-middleware.mdx
@@ -271,15 +271,15 @@ Find more examples at this [link](https://github.com/minpeter/ai-sdk-tool-call-m
 <Note>
   Implementing language model middleware is advanced functionality and requires
   a solid understanding of the [language model
-  specification](https://github.com/vercel/ai/blob/v5/packages/provider/src/language-model/v2/language-model-v2.ts).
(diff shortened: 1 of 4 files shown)


Zelys-DFKH:
The schema change is the right fix. I checked the forwarding path: `gateway-language-model.ts` passes provider options through as-is in the request body, so adding `caching` to the Zod schema is sufficient — no explicit field mapping to update elsewhere.

Two things reviewers will flag:

This repo requires a changeset file for releases. Without one the release bot marks the PR incomplete. Running `pnpm changeset` in the repo root will generate it.

The PR also needs a regression test confirming `caching: 'auto'` appears in the request body when set via `providerOptions`. Without it the fix can regress silently.

The three unrelated doc edits (middleware.mdx link updates, the mlflow embed fix) will slow review. Worth splitting them into a separate PR so the type fix is easy to review on its own.

narutamaaurum:
This PR is obsolete - the  property was already added to  in main via #15768 (feat(gateway): add experimental realtime model). Closing as duplicate.

PR #15938, opened 2026-06-09
fix(ai): reject incomplete model streams

kkdawkins:
## Background

Provider/Gateway streams can emit initial metadata and then terminate before a final finish/usage chunk. Previously `streamText` could record that as a successful-looking empty step:

- `finishReason: "other"`
- `text: ""`
- `response.messages: []`
- empty usage

Clients like OpenCode/Eve can then mark the turn done without any visible assistant response.

## Summary

- Reject `streamText` results when a provider stream ends without a model `finish` chunk.
- Preserve the incomplete-stream error in result promises instead of resolving to `finishReason: "other"` with empty text/messages.
- Reject pending result promises when provider stream consumption errors after metadata, avoiding hung `text` / `steps` / `finishReason` promises.
- Add regression coverage for clean early stream close and stream error after metadata.
- Add a local repro script for incomplete provider stream behavior.

## Manual Verification

- Ran `pnpm --filter ai exec tsx scripts/repro-incomplete-provider-stream.ts`.
- Confirmed metadata-only close now rejects with `AI_NoOutputGeneratedError`.
- Confirmed metadata-then-error now rejects `text`, `steps`, `finishReason`, and `totalUsage` instead of timing out.

Automated verification:

- `pnpm --filter ai exec vitest --config vitest.node.config.js --run src/generate-text/stream-text.test.ts`
- `pnpm exec oxlint packages/ai/src/generate-text/stream-text.ts packages/ai/src/generate-text/stream-text.test.ts packages/ai/scripts/repro-incomplete-provider-stream.ts`

## Checklist


packages/ai/src/generate-text/stream-text.test.ts
```diff
@@ -3100,6 +3193,11 @@ describe('streamText', () => {
             { type: 'text-delta', id: '1', delta: ', ' },
             { type: 'text-delta', id: '1', delta: 'world!' },
             { type: 'text-end', id: '1' },
+            {
+              type: 'finish',
+              finishReason: { unified: 'stop', raw: 'stop' },
+              usage: testUsage,
+            },
```
gr2m, line 3310: The need to update fixtures for existing tests shows a change of behavior: streams without a finish chunk used to yield an empty successful result. Now they throw an error.

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1901,6 +1905,17 @@ class DefaultStreamTextResult<
...
+
+                  if (
+                    chunkType !== 'model-call-response-metadata' &&
+                    chunkType !== 'model-call-end' &&
+                    chunkType !== 'tool-execution-end' &&
+                    chunkType !== 'error' &&
+                    chunkType !== 'raw'
+                  ) {
+                    hasReceivedOutputChunk = true;
+                  }
```
gr2m: only concern is that the list of chunk types might change in future and we miss to update this if block. But can't think of a better way

gr2m (approved):
This is great, thanks for bearing with me 👍🏼

packages/ai/src/generate-text/stream-text.ts
kkdawkins: Made this compiler enforced!

PR #15941, opened 2026-06-09
feat(harness): implement harness specification

felixarntz:
## Background

After #15919, this is the 2nd breakout PR from the overarching harness SDK work in #15496.

## Summary

* implement the harness specification v1 that each harness adapter must implement
    * eventually this should probably live in `@ai-sdk/specification`, but we keep it in its own `@ai-sdk/harness` for now due to the experimental nature of it all
* implement `HarnessAgent` as consumer-facing entrypoint `@ai-sdk/harness/agent` for running harnesses
* implement bridge foundation (separate entrypoint `@ai-sdk/harness/bridge`) which each harness that runs the underlying SDK inside a remote sandbox uses to communicate with that process in the remote sandbox
* implement harness adapter utilities (separate entrypoint `@ai-sdk/harness/utils`), exclusively for harness adapter use
* implement supplementary `HarnessAgent` helpers to fit into existing AI SDK paradigms where reasonable (e.g. observability)
* a few additional `ai` package exports are added because `HarnessAgent` lives in this separate package (because it's considered experimental)
    * only the `HarnessAgent` layer relies on these exports - the specification itself rightfully does not have any dependency on `ai`

## Manual Verification

Verification is happening via the original PR #15496.

## Checklist


packages/ai/src/generate-text/index.ts
```diff
@@ -39,6 +39,7 @@ export type {
   InferPartialOutput as InferStreamOutput,
 } from './output-utils';
 export type { PrepareStepFunction, PrepareStepResult } from './prepare-step';
+export { parseToolCall } from './parse-tool-call';
```
lgrammel: does this need to be a public export or can it be under internal?

packages/ai/src/telemetry/index.ts
```diff
@@ -1,6 +1,7 @@
 export type { TelemetryOptions } from './telemetry-options';
 export type { InferTelemetryEvent, Telemetry } from './telemetry';
 export { registerTelemetry } from './telemetry-registry';
+export { createTelemetryDispatcher } from './create-telemetry-dispatcher';
```
lgrammel: does this need to be a public export or can it be under internal?

packages/harness/src/observability/file-reporter.ts
```diff
@@ -0,0 +1,209 @@
...
+ * `events.jsonl` containing **both** the telemetry span lifecycle (turn / step
+ * / tool) **and** the forwarded bridge diagnostics (console lines + structured
+ * events). It is a single object registered in `telemetry.integrations`: the
+ * framework drives its `Telemetry` methods for spans and calls
+ * `ingestDiagnostic` for diagnostics.
+ *
+ * No external collector or OTel setup required — this is the AI-SDK-idiomatic
+ * replacement for the original SDK's host-side artifact files.
+ */
+export interface FileReporterOptions {
```
lgrammel, line 19: this is a separate observability layer. is there a way to integrate it into the ai sdk observability instead?

lgrammel (approved):

packages/harness/src/observability/file-reporter.ts
felixarntz: it uses the AI SDK `Telemetry` where possible, but it has an additional `HarnessDiagnostic` + `HarnessDiagnosticConsumer` layer that's specific to harnesses.

Since this code is specific to harnesses and experimental, I'd leave it here for now. the two reporters can technically still work with AI SDK models as well (because they implement `Telemetry` too), but those I'd prefer to leave here too mostly due to the experimental nature.

PR #15943, opened 2026-06-09
fix(middleware): preserve leading whitespace in streamed JSON suffix

### LINKED ISSUE #15921: extractJsonMiddleware removes valid leading space from final streamed suffix
### Description

`extractJsonMiddleware()` can remove a valid word-boundary space near the end of a streamed text response.

The middleware keeps a 12-character suffix buffer for streamed text. On `text-end`, if no markdown code-fence prefix was stripped, it applies the default transform to only that remaining suffix:

```ts
remaining = transform(remaining);
```

The default transform ends with `.trim()`. If the final suffix begins with a legitimate leading space, that space is removed, causing adjacent words to fuse.

## Reproduction

```ts
import { extractJsonMiddleware } from 'ai';

const middleware = extractJsonMiddleware();

const wrapped = await middleware.wrapStream({
  doStream: async () => ({
    stream: new ReadableStream({
      start(controller) {
        controller.enqueue({ type: 'stream-start', warnings: [] });
        controller.enqueue({ type: 'text-start', id: 't' });
        controller.enqueue({ type: 'text-delta', id: 't', delta: 'there' });
        controller.enqueue({ type: 'text-delta', id: 't', delta: ' altogether?' });
        controller.enqueue({ type: 'text-end', id: 't' });
        controller.close();
      },
    }),
  }),
});

const chunks = [];
for await (const chunk of wrapped.stream) {
  if (chunk.type === 'text-delta') chunks.push(chunk.delta);
}

console.log(chunks.join(''));
```

## Expected

```text
there altogether?
```

## Actual

```text
therealtogether?
```

## Additional Examples

```text
one + " more to it?" -> onemore to it?
them + " right here!" -> themright here!
running + " this time!" -> running this time!
```

The last case survives because the final 12-character suffix does not start with a space.

## Version

Observed in:

- `ai@6.0.92`
- `ai@6.0.198`

## Notes

This shows up in normal prose streams if an application wraps all language models with `extractJsonMiddleware()`. The docs describe the middleware as JSON/code-fence cleanup for structured output, so applications can work around this by only applying it to JSON/object-generation paths. Still, the middleware should not mutate valid streamed text by deleting a real leading space from the final suffix.

### AI SDK Version

ai@6.0.92
ai@6.0.198

### Code of Conduct

tsushanth:
## Why

Fixes #15921.

\`extractJsonMiddleware.wrapStream\` keeps a 12-character suffix buffer and, on \`text-end\`, applies the default markdown-fence transform to whatever remains when no prefix was stripped. The default transform ends in \`.trim()\`, which drops a legitimate word-boundary space at the start of the trailing suffix and fuses the final word with the previously emitted delta:

\`\`\`
'there' + ' altogether?'  →  'therealtogether?'   // expected: 'there altogether?'
\`\`\`

## Fix

In the streaming phase the buffer can only contain the *trailing* characters of the response — everything before it was already streamed out as-is. So leading whitespace in the buffer is part of the content, not a fence artifact, and the cleanup should only strip a trailing fence and trim the right side. Reuse the same \`replace(/\n?\\\`\\\`\\\`\\s*\$/, '').trimEnd()\` logic that was already applied when \`prefixStripped\` was true.

10-line diff in [\`extract-json-middleware.ts\`](https://github.com/vercel/ai/blob/main/packages/ai/src/middleware/extract-json-middleware.ts).

## Tests

Added a regression test \`should preserve leading whitespace in the trailing suffix when no markdown prefix was stripped (#15921)\` mirroring the reproduction in the issue. Verified by stashing the fix and re-running the suite — the new test fails on \`main\` (\`'therealtogether?'\` vs \`'there altogether?'\`) and passes with the fix. All 22 \`extractJsonMiddleware\` tests pass.

## Changeset

\`patch\` on the \`ai\` package — non-breaking bug fix.

### DIFF
--- a/.changeset/fix-extract-json-middleware-leading-space.md
+++ b/.changeset/fix-extract-json-middleware-leading-space.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(middleware): preserve leading whitespace in the trailing suffix of streamed text in `extractJsonMiddleware`. Previously, when no markdown code fence was stripped, the default transform's `trim()` was applied to the 12-character suffix buffer, removing a legitimate word-boundary space and fusing the last word with whatever came before it.
--- a/packages/ai/src/middleware/extract-json-middleware.test.ts
+++ b/packages/ai/src/middleware/extract-json-middleware.test.ts
@@ -817,5 +817,44 @@ describe('extractJsonMiddleware', () => {
 
       expect(await result.text).toBe('{"value": "test"}');
     });
+
+    it('should preserve leading whitespace in the trailing suffix when no markdown prefix was stripped (#15921)', async () => {
+      // The middleware keeps a 12-character suffix buffer. When the prior delta
+      // ends mid-sentence and the final suffix begins with a legitimate
+      // word-boundary space, that space must survive into the merged output.
+      const mockModel = new MockLanguageModelV4({
+        async doStream() {
+          return {
+            stream: convertArrayToReadableStream([
+              {
+                type: 'response-metadata',
+                id: 'id-0',
+                modelId: 'mock-model-id',
+                timestamp: new Date(0),
+              },
+              { type: 'text-start', id: '1' },
+              { type: 'text-delta', id: '1', delta: 'there' },
+              { type: 'text-delta', id: '1', delta: ' altogether?' },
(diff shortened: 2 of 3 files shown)


lgrammel:
Closing this PR because issue #15921 has already been closed.

PR #15966, opened 2026-06-10
fix(mcp): prevent mcp oauth credentials from being sent to untrusted token endpoints

aayush-kapoor:
## Background

the client accepted OAuth metadata from the MCP server’s discovery flow and then used whatever `token_endpoint` was discovered, even if that endpoint was not actually bound to the authorization server that issued the stored credentials.

example:

1. stored credentials were issued by `https://auth.example.com`
2. mcp server returns PRM with authorization_servers: ["https://auth.example.com"].
3. authorization server metadata response says:
```
{
  "issuer": "https://auth.example.com",
  "token_endpoint": "https://evil.example/token"
}
```
Previously, the client could POST `refresh_token`, `client_secret`, or `code_verifier` to `https://evil.example/token`

## Summary

the client now only sends OAuth credentials when the discovered metadata matches the expected authorization server and its endpoints are same-origin with that issuer. stored credentials are also pinned to the AS/token endpoint that issued them.

## Manual Verification

na

## Checklist

### DIFF
--- a/.changeset/soft-steaks-listen.md
+++ b/.changeset/soft-steaks-listen.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/mcp": patch
+---
+
+fix(mcp): prevent mcp oauth credential exfiltration during rediscovery
--- a/.changeset/stale-news-rest.md
+++ b/.changeset/stale-news-rest.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/mcp": patch
+---
+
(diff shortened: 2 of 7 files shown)


aayush-kapoor:
on further review i think we can close this pr https://github.com/vercel/ai/pull/15966; just went through the oauth docs for this attack vector and the [mix-up attack](https://datatracker.ietf.org/doc/html/rfc9700#name-mix-up-attacks) is only for MCP clients that interact with 2 or more authorization servers. our MCPClient only interacts with one

PR #16049, opened 2026-06-11
fix(devtools): validate viewer database paths

aayush-kapoor:
## Background

currently `/api/notify` trusted whatever `dbPath` came in the request body. the viewer would store `/etc/passwd` as `remoteDbPath`, and the next `/api/runs` call would pass it into `reloadDb()`, which attempted to `readFileSync()` and parse that file. If the file happened to be JSON shaped like the devtools DB, it could be exposed through the viewer APIs

## Summary

`dbPath` is only accepted if it resolves to a real regular file named:

`<some-project>/.devtools/generations.json`

## Manual Verification

na

## Checklist

### DIFF
--- a/.changeset/old-singers-add.md
+++ b/.changeset/old-singers-add.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/devtools": patch
+---
+
+fix(devtools): validate viewer database paths
--- a/packages/devtools/src/db.test.ts
+++ b/packages/devtools/src/db.test.ts
@@ -101,4 +101,20 @@ describe('devtools db path validation', () => {
 
     await expect(db.getRuns()).resolves.toEqual([]);
   });
+
+  it('rejects devtools databases larger than the size cap', async () => {
+    const { db, tempDir } = await loadDbModule();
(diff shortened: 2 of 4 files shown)


gr2m:
Pushed a small graft on top (20e7921): a **100 MB read-size cap** in `validateRemoteDbPath` (`stats.size > MAX_DB_BYTES → reject`), plus a regression test using a sparse file just over the cap.

Context: this consolidates the duplicate VULN-11550 fix I'd opened (#16055), which I've now closed in favor of this PR. Your validation (basename/dir + isFile + `realpathSync` symlink-escape) is the better approach — zero behavior change for cross-directory viewers, and the symlink defense is something my version didn't have. The size cap was the one gap (a genuinely huge real `generations.json` could still hang/OOM the synchronous `readFileSync`), so I've added just that piece here. All 6 db tests pass; type-check clean.

Note: VULN-11530's Host/Origin gating (#16042 / its v6 backport #16054) makes the cross-origin POST unreachable, so the residual "another project's `.devtools/generations.json`" surface is local-only — fine to leave as-is.

gr2m (approved):

PR #16050, opened 2026-06-11
fix(mcp): lock first sse endpoint received via event

aayush-kapoor:
## Background

currently if a client connects to an endpoint `https://mcp.example/sse`, the client will POST tool calls to `https://mcp.example/messages`

but  if a later SSE event said:
```
event: endpoint
data: https://evil.example/collect
```
the client accidentally stored that bad URL before rejecting it. and the next tool call could POST to `evil.example`

## Summary

the first valid endpoint stays locked in and later endpoint events are ignored now

## Manual Verification

na

## Checklist

### DIFF
--- a/.changeset/giant-squids-sip.md
+++ b/.changeset/giant-squids-sip.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/mcp": patch
+---
+
+fix(mcp): lock first sse endpoint received via event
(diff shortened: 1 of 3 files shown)


gr2m (approved):
Reviewed against VULN-11561 — correctly and completely fixes the assign-before-validate bypass:
- validates the candidate origin in a local `const endpoint` before assigning `this.endpoint` (so the swallowed-error path can't leave a poisoned endpoint),
- ignores `endpoint` events after the first (`if (this.endpoint) continue`), blocking the second-event overwrite,
- on mismatch resets `connected`/`endpoint` and tears down the connection; `close()` also clears `endpoint`.

`send()` gates on both, so the credential-bearing POST can't reach an attacker origin. 19/19 tests pass incl. the two new cross-origin cases. LGTM.

PR #16098, opened 2026-06-13
docs: Add @kakunin/ai-sdk to community tools registry

nqzai:
This PR adds @kakunin/ai-sdk (the Kakunin compliance and scope verification adapter for the Vercel AI SDK) to the tools registry index, allowing developers to discover it.


nqzai:
Commit 0ba53d7 changed the Kakunin docsUrl to https://www.kakunin.ai/docs/vercel-ai-sdk — exactly the fix this suggestion proposed. The URL now returns HTTP 200 (verified via curl), so the dead-link bug no longer exists.

PR #16099, opened 2026-06-13
feat(harness-pi): add `providers` option to register custom providers per session

SebastianSzturo:
## Summary

Adds a `providers` option to `createPi` / `PiHarnessSettings` that registers caller-supplied providers into each session's `ModelRegistry` / `AuthStorage` **before** model resolution.

## Motivation

Today a host can only reach providers through `auth.customEnv` (the `<PREFIX>_API_KEY` / `<PREFIX>_BASE_URL` pattern). That leaves out two real use-cases:

1. **OAuth / subscription credentials.** Running a built-in catalog model such as `gpt-5.5` through the `openai-codex` provider using a ChatGPT-subscription token (an OAuth credential, not an API key) is impossible via `customEnv`.
2. **Custom providers** that need a full `ModelRegistry.registerProvider` config (base URL, headers, OAuth provider, custom models).

This lets a host pass those directly through the harness.

## API

```ts
createPi({
  model: "gpt-5.5",
  providers: [
    {
      name: "openai-codex",
      // OAuth credential seeded into the per-session AuthStorage; Pi's built-in
      // OAuth refresh renews it as needed.
      credential: { type: "oauth", access, refresh, expires, accountId },
      // optional: persist the renewed credential after an automatic refresh
      onCredentialRefresh: (cred) => saveSomewhere(cred),
    },
  ],
});
```

Each `PiProvider`:
- `credential` — seeded into the per-session `AuthStorage` (OAuth or API key).
- `config` — passed verbatim to `ModelRegistry.registerProvider`.
- `onCredentialRefresh` — optional callback fired when Pi writes a refreshed credential for the provider into `AuthStorage` (after an automatic OAuth refresh), so a long-running host can persist the renewed credential durably and survive access-token expiry across sessions.

## Resolver disambiguation

Pi's catalog publishes the same model id under several providers (e.g. `gpt-5.5` under `openai-codex`, `openai`, `openrouter`, `vercel-ai-gateway`, ...). `createPiModelResolver` now receives the registered provider names and prefers a matching entry under one of them, so dispatch uses the provider whose credential the caller actually seeded — instead of an unauthenticated one, which would otherwise fail with `No API key found`. When no preferred provider matches, behaviour is unchanged (gateway-routed entry when gateway creds are present, else first match).

## Implementation notes

- Types are derived from the pi SDK (`Parameters<AuthStorage['set']>[1]`, `Parameters<ModelRegistry['registerProvider']>[1]`) so they track upstream, matching the existing `ProviderConfigInput` pattern in `pi-auth.ts`.
- The `onCredentialRefresh` hook works by wrapping the per-session `AuthStorage.getApiKey`: Pi refreshes an expired OAuth credential inside `getApiKey` and writes it back via its locked storage backend (not the public `set`), so the wrapper snapshots the stored credential before/after each call and fires the callback when it changes.
- Fully backward compatible: `providers` is optional; omitting it leaves all existing behaviour unchanged.

## Tests

- Added unit tests for the resolver preference (`pi-model-resolver.test.ts`): prefers a registered custom provider over other providers for the same id, prefers it even when gateway creds are present, honours preferred-provider order, and falls back to the default behaviour when no preferred provider matches.
- `pnpm --filter @ai-sdk/harness-pi type-check` and `build` pass; the resolver test file passes (11/11). A changeset is included.

### DIFF
--- a/.changeset/harness-pi-custom-providers.md
+++ b/.changeset/harness-pi-custom-providers.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/harness-pi": patch
+---
+
+feat(harness-pi): add `providers` option to register custom providers (incl. OAuth/subscription) per session
+
+`createPi({ providers })` registers caller-supplied providers into each session's `ModelRegistry`/`AuthStorage` before model resolution. A `credential` seeds `AuthStorage` (OAuth or API key) and a `config` is passed to `ModelRegistry.registerProvider`, so a host can use OAuth/subscription credentials (e.g. an `openai-codex` ChatGPT-subscription token) or fully custom providers for a catalog model without the `auth.customEnv` API-key path. Registered provider names are preferred during model resolution when a model id is published under multiple providers, and an optional `onCredentialRefresh` callback lets a long-running host persist credentials renewed by Pi's automatic OAuth refresh.
--- a/packages/harness-pi/src/index.ts
+++ b/packages/harness-pi/src/index.ts
@@ -9,3 +9,4 @@ export const pi = createPi();
 export { createPi } from './pi-harness';
 export type { PiHarnessSettings } from './pi-harness';
 export type { PiAuthOptions } from './pi-auth';
+export type { PiProvider } from './pi-session';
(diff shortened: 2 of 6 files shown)


gorkie-agent:
+1

PR #16137, opened 2026-06-15
feat (provider): server-side realtime extensions for RealtimeModelV4

kkdawkins:
## Background

The experimental `RealtimeModelV4` spec was written for the browser: every method except `doCreateClientSecret` is "browser-side", so the model knows how a *browser* connects (short-lived token carried in a subprotocol via `getWebSocketConfig`, voice URL only) but nothing about how a *trusted server* connects (header auth, transcription/translation endpoint variants), and the normalized server events drop usage.

A server-side realtime proxy/gateway therefore has to hand-write the provider-specific pieces the AI SDK already owns for HTTP: upstream URL construction per endpoint, header auth, and usage extraction parsed back out of each event's `raw` payload. This PR closes that gap so a server can drive realtime providers the same way it drives HTTP language models (`createProvider(...).experimental_realtime(id)`), with the provider package owning the provider specifics.

This adds **no provider capability the HTTP path doesn't already have** — the provider already encodes its base URL + how to build auth from credentials (that's how `doGenerate` reaches the upstream, and what `doCreateClientSecret` already reads). `getServerConnection` only *surfaces* that existing knowledge as a descriptor, because a server proxy owns the socket itself (rather than the SDK making the call as it does over HTTP). It is the server-side sibling of the existing browser `getWebSocketConfig`.

## Summary

Optional, additive, backward-compatible extensions to `RealtimeModelV4`:

- **`@ai-sdk/provider`**
  - New `getServerConnection({ intent? })` → `{ url, headers, protocols? }` for a server-initiated upstream connection (header auth with the long-lived credential). `intent` is `conversation` (default) / `transcription` / `translation`. May return a promise for providers that need a round-trip.
  - New `RealtimeModelV4Usage` type and optional `usage` on the `response-done` and `input-transcription-completed` server events. Counts are gross (cache-inclusive) with cached portions reported separately (`cachedInput*Tokens`), so a consumer can apply its own cached-vs-uncached billing split without re-parsing `raw`. Duration-billed transcription reports `audioSeconds`.
  - New `RealtimeModelV4ServerConnection` / `RealtimeModelV4SessionIntent` types, exported with the `Experimental_` prefix.
- **`@ai-sdk/openai`** — `getServerConnection` builds the per-intent URL (`?model=` / `?intent=transcription` / `/translations?model=`) + header auth from config; mapper emits `usage` on `response-done` and transcription completions.
- **`@ai-sdk/xai`** — `getServerConnection` (voice URL with `model` query param, conversation-only); mapper emits `usage` via the OpenAI-compatible realtime wire shape.
- **`@ai-sdk/google`** — `getServerConnection` targets the **unconstrained** Gemini Live Bidi endpoint with `x-goog-api-key` header auth (a trusted server holds the key and sends the session `setup` over the socket — no ephemeral-token mint); mapper maps per-modality `usageMetadata` (audio/text) onto `response-done`, with aggregate fallback.
- **`ai`** — re-exports `RealtimeServerConnection`, `RealtimeSessionIntent`, `RealtimeUsage`.

## Manual Verification

Not yet run against live providers — to be exercised end-to-end by a server-side proxy consuming local packed builds (the AI Gateway realtime path). Automated coverage below is in place; will update this section with live verification before marking ready.

## Checklist

Tests added: OpenAI/xAI/Google `getServerConnection` (URL per intent + header forwarding) and `usage` mapping (gross + cached split, transcription token vs duration, Google per-modality + aggregate fallback). `type-check`, `ultracite` check, and realtime test suites pass across `provider`, `ai`, `openai`, `xai`, `google`.

## Future Work

- Server transport: a pluggable `RealtimeTransport` interface + Node `ws` transport so `AbstractRealtimeSession` can run server-side (out of scope here — a server proxy that owns the socket only needs the codec + connection descriptor).
- Documentation for the server-side realtime surface.


packages/google/src/realtime/google-realtime-model.ts
```diff
@@ -41,6 +49,26 @@ function getWebSocketURL(baseURL: string): string {
...
 
+function getServerWebSocketURL(baseURL: string): string {
+  const url = getRealtimeBaseURL(baseURL);
+  url.protocol = url.protocol === 'https:' ? 'wss:' : 'ws:';
+  url.pathname = `${url.pathname.replace(/\/$/, '')}/ws/${realtimeServerWebSocketPath}`;
+  return url.toString();
+}
+
+/** Drop headers with `undefined` values so the result is a `Record<string, string>`. */
+function definedHeaders(
```
gr2m: replace with `removeUndefinedEntries` from provider utils

gr2m (approved):

PR #16145, opened 2026-06-15
fix(ai): preserve chat response for onFinish

### LINKED ISSUE #15668: `Chat.makeRequest` finally throws on `this.activeResponse!.state` when activeResponse is undefined (should be `?.`)

kiranmagic7:
## Summary

Closes #15668.

- keep a request-local `activeResponse` reference for `onFinish` instead of reading the mutable instance field in `finally`
- avoid clearing `this.activeResponse` when another overlapping request has already replaced it
- add a regression test that clears `activeResponse` before stream finish and verifies `onFinish` still receives the assistant message without logging an error
- add a patch changeset for `ai`

## Tests

- `pnpm test:node src/ui/chat.test.ts`
- `pnpm check`
- `pnpm --filter ai build`

### DIFF
--- a/.changeset/tidy-chats-finish.md
+++ b/.changeset/tidy-chats-finish.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Fix `Chat` finishing when `activeResponse` is cleared before `onFinish`.
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -2952,4 +2952,78 @@ describe('Chat', () => {
       `);
     });
   });
+
+  describe('active response cleanup', () => {
+    it('calls onFinish from the request-local response when activeResponse is cleared', async () => {
+      let controller: ReadableStreamDefaultController<UIMessageChunk>;
+      const responseStream = new ReadableStream<UIMessageChunk>({
+        start: controllerArg => {
+          controller = controllerArg;
+          controller.enqueue({ type: 'start' });
+          controller.enqueue({ type: 'start-step' });
+          controller.enqueue({ type: 'text-start', id: 'text-1' });
+          controller.enqueue({
+            type: 'text-delta',
+            id: 'text-1',
+            delta: 'Hello',
+          });
+        },
+      });
+
+      const onFinish = vi.fn();
+      const consoleError = vi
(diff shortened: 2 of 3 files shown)


kiranmagic7:
Quick status check on current head `1a9d9cb86a0f680a1b229302b9b4e935010a1de8`: the PR still reports mergeable from the PR view, Socket Security and Vercel Agent Review are green, and the only red item I can see is the Vercel preview/deployment authorization, which I cannot clear from this fork.

Review focus: whether preserving the response object for `onFinish` matches the intended chat callback contract. If you prefer a narrower shape or a different test location, I can adjust it.

kiranmagic7:
Friendly follow-up: current head `1a9d9cb86a0f680a1b229302b9b4e935010a1de8` is still mergeable with no conflicts. Socket and Vercel Agent Review checks pass; the only pending item is the Vercel deployment authorization from the fork.

Happy to rebase or adjust the approach if that would help. Let me know if there are questions or if there is a preferred path for the deployment authorization.

aayush-kapoor (commented):
closing in favor of https://github.com/vercel/ai/pull/16620

PR #16219, opened 2026-06-17
feat (video): add first-class `frameImages` and `inputReferences` call options for video generation

joshualipman123:
## Summary

This PR implements first-class, provider-agnostic media inputs for video generation. Today, last-frame and multi-reference images are buried under inconsistent providerOptions keys (klingai.imageTail vs bytedance.lastFrameImage, klingai.elementList vs bytedance.referenceImages). This PR lifts those concepts to top-level call options.

## Manual Verification

Manually verified by running examples to check cases like:

- First-last-frame requests produce videos interpolating between the provided start and end images
- Reference-to-video requests produce videos incorporating the provided reference images
- Legacy providerOptions paths (imageTail, lastFrameImage, referenceImages) still work when the new top-level fields are omitted

## Checklist


gr2m (approved):

packages/google/src/google-video-model.ts
shaper, line 1: feedback from claude -- not sure if it is accurate, take a look?

```
🔴 Blocking — @ai-sdk/google image wire format is broken (regression)

packages/google/src/google-video-model.ts — the new convertFileToGoogleImage emits the Vertex shape:

return { bytesBase64Encoded: base64Data, mimeType: file.mediaType || 'image/png' };

But @ai-sdk/google targets https://generativelanguage.googleapis.com/v1beta (confirmed in google-provider.ts:152), and the Gemini Developer API requires the inlineData shape — { inlineData: { mimeType, data } } — which is exactly what main produced. bytesBase64Encoded is the Vertex-only field name. The two endpoints genuinely differ on this (Gemini API video docs, Vertex Veo reference).

This breaks three paths on the Google (non-Vertex) provider with a 400 "Unable to process input image":
- instance.image — existing i2v start frame (regression from main's inlineData)
- instance.lastFrame — new feature
- instance.referenceImages legacy path — convertProviderReferenceImage now returns bytesBase64Encoded/gcsUri, where main returned inlineData/gcsUri

The tests don't catch it because they were rewritten to assert the new (wrong) shape — main had it('should send image as inlineData'), the PR replaced it with it('should send image as bytesBase64Encoded'). This is the classic case of updated tests masking a wire-format regression; the "manual verification" in the description most likely exercised Vertex, not the generativelanguage endpoint.

Fix: @ai-sdk/google must keep wrapping inline bytes as { inlineData: { mimeType, data } } (and reference images likewise). google-vertex is correct as-is. The shared helper can't be identical across the two packages.
```

packages/ai/src/generate-video/generate-video.ts
shaper, line 1: another claude thought, wdyt?

> 🟡 Silent data loss when frameImages + inputReferences are both passed
> 
> In generate-video.ts, effectiveInputReferences is set to undefined whenever frameImages is present, and the test should ignore inputReferences when frameImages is provided locks this in — with no warning. Because the core nulls it before doGenerate, the provider-level "inputReferences ignored" warnings can never fire for this combination either. A user who passes both gets their references dropped silently. This should emit a Warning so it surfaces via logWarnings.

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -334,60 +388,60 @@ function normalizePrompt(promptArg: GenerateVideoPrompt): {
...
+    image:
+      promptArg.image != null ? normalizeImageData(promptArg.image) : undefined,
+  };
+}
+
+/**
+ * Normalizes a {@link DataContent} image into a {@link Experimental_VideoModelV4File}.
+ * Accepts a URL string, a data URL, a base64 string, or binary image data.
+ */
+function normalizeImageData(
```
shaper, line 433: this reads to me like we may be missing `ArrayBuffer` -- can you see if it is accurate? typically ai sdk has helper/util fns to make it easier to work with and convert/handle these consistently, so would check how generateImage logic handles similar.

```
- Silent drop of un-normalizable entries: normalizeImageData returns undefined for anything that isn't a string or Uint8Array, and the flatMap drops it. DataContent also includes ArrayBuffer (handled nowhere). For a single prompt.image this was pre-existing, but for frameImages/inputReferences arrays it means partial silent loss (e.g. 1 of 3 references vanishes). Consider warning on dropped entries, or handling ArrayBuffer.
```

packages/ai/src/generate-video/generate-video.ts
shaper, line 1: if below is accurate may be worth a logic edit, take a look and let me know thoughts. I would think we would want the providers to prioritize what we host into `prompt.image` as well -- is this consistent re: the gist of PR prioritizes in-general-prompt params over provider-options params, or did we go the other direction?

> - image vs first_frame precedence: the core copies first_frame into image only when no prompt image exists, but every provider's resolveStartImage is firstFrame ?? image. So if a caller passes both prompt.image and a frameImages first_frame, the core keeps prompt.image in the image field yet providers use first_frame — prompt.image is silently ignored. Worth a comment documenting the intended precedence, or a warning.

packages/xai/src/xai-video-model.ts
shaper, line 1: - xai fileToXaiImageUrl builds data:${file.mediaType};base64,...; if mediaType is ever undefined you get data:undefined;base64,. Low risk since type:'file' normally carries it, but a fallback (?? 'image/png') matches the other providers.

packages/klingai/src/klingai-video-model.ts
shaper, line 1: klingai buildMultiImageBody: the start-frame warning uses feature: 'image' while the last-frame warning uses feature: 'frameImages' — slightly inconsistent labeling.

packages/google/src/google-video-model.ts
joshualipman123: ya, got gemini api & vertex mixed up

packages/ai/src/generate-video/generate-video.ts
joshualipman123: dropping inputRefs is intentional but ya should warn before doing so.

joshualipman123: prompt.image and frameImages are both first class. i think it makes sense to keep prioritizing firstFrame as this is the more specific/intentional field. added a warning.

shaper (approved):

PR #16221, opened 2026-06-17
fix(provider/anthropic): gate programmatic-tool-call injection on markCodeExecutionDynamic

### LINKED ISSUE #15951: @ai-sdk/anthropic 3.0.80 regression: text_editor/bash code_execution calls get "programmatic-tool-call" type injected, failing input validation (breaks skills)
### Description

Since 3.0.80, streaming a `text_editor_code_execution` or `bash_code_execution` server tool call fails validation when `code_execution` is registered as a tool. In practice this breaks Anthropic skills: as soon as the model reads `/skills/<name>/SKILL.md`, the stream errors with:

```
AI_InvalidToolInputError: Invalid input for tool code_execution: Type validation failed:
Value: {"type":"programmatic-tool-call","command":"view","path":"/skills/my-skill/SKILL.md"}
[{"expected":"string","code":"invalid_type","path":["code"],...}]
```

The API never sent that `type` field — it's injected by the provider. What happens:

- on `content_block_start`, `text_editor_code_execution`/`bash_code_execution` blocks are remapped to `providerToolName = 'code_execution'`, losing the original name: https://github.com/vercel/ai/blob/d1b10391dd2b0cc9a5f563eb5be8b8b6fee22e45/packages/anthropic/src/anthropic-language-model.ts#L1697-L1701
- on the first `input_json_delta`, `"type": "programmatic-tool-call"` is injected for every `code_execution` block: https://github.com/vercel/ai/blob/d1b10391dd2b0cc9a5f563eb5be8b8b6fee22e45/packages/anthropic/src/anthropic-language-model.ts#L2255-L2263

So a text editor `view` ends up as `{"type":"programmatic-tool-call","command":"view","path":...}`, which fails the discriminated union (that branch requires `code: string`).

Up to 3.0.79 this worked — the delta path kept the sub-tool name and injected it as the type. 3.0.80 (#15566) normalized everything to `code_execution` + `programmatic-tool-call`. That PR was about *dynamic* calls triggered by web_fetch/web_search, where validation is bypassed, so the mismatch didn't show up there. With an explicitly registered tool, validation applies and every sub-tool call fails.

Repro without an API key (fake fetch returning the SSE the API produces for a skill read):

```ts
import { createAnthropic } from '@ai-sdk/anthropic';

const events = [
  { type: 'message_start', message: { id: 'msg_1', type: 'message', role: 'assistant', content: [], model: 'claude-sonnet-4-6', stop_reason: null, stop_sequence: null, usage: { input_tokens: 10, output_tokens: 1 } } },
  { type: 'content_block_start', index: 0, content_block: { type: 'server_tool_use', id: 'srvtoolu_1', name: 'text_editor_code_execution', input: {} } },
  { type: 'content_block_delta', index: 0, delta: { type: 'input_json_delta', partial_json: '{"command":"view",' } },
  { type: 'content_block_delta', index: 0, delta: { type: 'input_json_delta', partial_json: '"path":"/skills/my-skill/SKILL.md"}' } },
  { type: 'content_block_stop', index: 0 },
  { type: 'message_delta', delta: { stop_reason: 'tool_use', stop_sequence: null }, usage: { output_tokens: 5 } },
  { type: 'message_stop' },
];
const sse = events.map(e => `event: ${e.type}\ndata: ${JSON.stringify(e)}\n\n`).join('');

const anthropic = createAnthropic({
  apiKey: 'test-key',
  fetch: async () => new Response(sse, { status: 200, headers: { 'content-type': 'text/event-stream' } }),
});

const result = streamText({
  model: anthropic('claude-sonnet-4-6'),
  tools: { code_execution: anthropic.tools.codeExecution_20260120() },
  prompt: 'read the skill file',
});

  if (part.type === 'tool-call') console.log(JSON.stringify(part, null, 2));
}
```

On 3.0.79 the tool call comes out with `type: "text_editor_code_execution"` and validates. On 3.0.80/3.0.81 it comes out `invalid: true` with the error above.

I'm working around it with a local patch that keeps `part.name` on the content block and injects the matching discriminator on the first delta — happy to open a PR if that approach sounds right.

### AI SDK Version

- ai: 6.0.198
- @ai-sdk/anthropic: 3.0.81 (regression introduced in 3.0.80, last good 3.0.79)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tsushanth:
## Background

`@ai-sdk/anthropic` 3.0.80 ([#15566](https://github.com/vercel/ai/pull/15566)) introduced a `programmatic-tool-call` discriminator injection on `code_execution` server-tool calls to support the auto-triggered path (where `web_fetch` / `web_search` internally invoke code execution and the user did not register `code_execution` themselves). That PR correctly marks the auto-triggered case as `dynamic: true` so validation is bypassed.

[#15951](https://github.com/vercel/ai/issues/15951) reports the regression that lands on the **non-dynamic** path: when a user explicitly registers `code_execution`, the discriminator injection still fires, but `dynamic: true` is gated on `markCodeExecutionDynamic` (which is `false` in this case). The result is that the user's registered tool schema validates the rewritten input and fails with `AI_InvalidToolInputError: Type validation failed`. Skills that read `/skills/<name>/SKILL.md` are an easy way to reproduce because they exercise the streaming `firstDelta` path.

The asymmetry, visible at HEAD `952b666`:

| Site | Action | Currently gated on `markCodeExecutionDynamic`? |
|---|---|---|
| `1037` non-streaming `dynamic: true` | gates the flag | ✅ |
| `1067` non-streaming `dynamic: true` | gates the flag | ✅ |
| `1726` streaming `dynamic: true` (start) | gates the flag | ✅ |
| `1743` streaming `dynamic: true` (start enqueue) | gates the flag | ✅ |
| `2147` streaming `dynamic: true` (end) | gates the flag | ✅ |
| **`1048` non-streaming `inputToSerialize` rewrite** | unconditional | **❌** |
| **`2121` streaming `finalInput` rewrite (`tool-input-end`)** | unconditional | **❌** |
| **`2268` streaming `delta` rewrite (`input_json_delta` firstDelta)** | unconditional | **❌** |

The four `dynamic: true` emission sites correctly gate; the three rewrite sites don't.

## Summary

This PR adds `markCodeExecutionDynamic && …` to each of the three rewrite sites so they mirror the existing 4-site gate pattern. With the gate in place:

- **Dynamic path** (`code_execution` not user-registered, auto-triggered by `web_fetch` / `web_search`) — rewrite still fires, `dynamic: true` still set, validation still bypassed. Unchanged behavior.
- **Non-dynamic path** (`code_execution` registered by the user) — rewrite no longer fires, `dynamic: true` was already not set, the user's tool schema validates the original API-shaped input. Fixes #15951.

## Changes

- [`packages/anthropic/src/anthropic-language-model.ts`](packages/anthropic/src/anthropic-language-model.ts) — three minimal gate additions, total +13 / -3
- [`packages/anthropic/src/anthropic-language-model.test.ts`](packages/anthropic/src/anthropic-language-model.test.ts) — adds an explicit regression test under `code execution 20250825 — user-registered code_execution input is not mangled (regression for #15951)` that registers `code_execution` and asserts the non-streaming input is not rewritten
- 9 existing snapshots updated — all 9 fixtures register `code_execution` as a tool, so they were previously snapshotting the rewritten (buggy) output. The updated snapshots reflect the corrected non-rewritten shape. The dynamic path's rewrite behavior is still exercised by the same tests' dynamic-mode assertions and the new gate logic.

## Verification

```
pnpm test:node    # 435/435 pass (was 425/9-fail pre-fix, then 9 snapshot updates after fix)
pnpm test:edge    # 435/435 pass
pnpm type-check   # clean
```

## Adjacent observations (not in this PR's scope, flagging for the maintainers)

While investigating I noticed two adjacent points that may be worth a separate follow-up:

1. **Sub-tool sites also gate `dynamic: true` on `markCodeExecutionDynamic`.** Lines 1037 (`text_editor_code_execution` / `bash_code_execution` non-streaming) and 1726 (streaming) gate `dynamic: true` on `markCodeExecutionDynamic`. When the user registers `code_execution` and the model emits a sub-tool call (`text_editor_code_execution` view), the `dynamic` flag is *not* set, so the user's `code_execution` tool schema validates a `{command, path}` input that almost certainly doesn't match the schema's `{code: string}` shape. The bug shape is identical to #15951 but it's a separate design question — should sub-tool calls always be dynamic (since users don't register sub-tools by name), or should the SDK surface them as distinct tools? Not addressed here.

2. **Streaming sub-tool delta discriminator is `programmatic-tool-call` rather than the sub-tool name.** Line 2268's rewrite uses `programmatic-tool-call` for any `firstDelta` on a `code_execution`-mapped block, but non-streaming `text_editor_code_execution` / `bash_code_execution` at line 1031 preserves the sub-tool name as the discriminator (`type: part.name`). With the dynamic-path gate in this PR, the inconsistency only surfaces in the dynamic auto-triggered case — so it doesn't affect this PR's fix — but the two paths could be made symmetric in a follow-up.

Closes #15951.


aayush-kapoor (commented):
marking as closed in favor of https://github.com/vercel/ai/pull/16469

PR #16222, opened 2026-06-17
feat (video): add first-class `generateAudio` call option

joshualipman123:
## Summary

This PR promotes audio generation to a first-class top-level generateAudio boolean on experimental_generateVideo, instead of requiring provider-specific passthrough options.

Problem: Audio was buried under four different provider-option names (vertex.generateAudio, klingai.sound, alibaba.audio, bytedance.generateAudio), even though it's a universal concept.

## Manual Verification

Manually verified with end-to-end examples with all different cases across providers

## Checklist


shaper (approved):

.changeset/video-generate-audio-first-class.md
```diff
@@ -0,0 +1,11 @@
+---
+"@ai-sdk/provider": patch
+"ai": patch
```
gr2m, line 3: this is a breaking change since we update the spec `packages/provider/`.

```suggestion
"@ai-sdk/provider": major
"ai": patch
```

If we want to backport it to v6 we have to find another way, like using provider options for the providers that support it instead.

gr2m (dismissed):

.changeset/video-generate-audio-first-class.md
gr2m: actually nevermind forgot that video specs are experimental, we can backport it to v6

gr2m (approved):

PR #16242, opened 2026-06-17
feat(gateway): expose ruleId on GatewayForbiddenError so callers can identify which routing rule denied a request

R-Taneja:
## Why

When AI Gateway blocks a request with a routing **deny** rule, the SDK throws `GatewayForbiddenError` — but the **rule ID isn't accessible** as a property. Callers should be able to read `error.ruleId` directly to identify and manage the rule that blocked them.

## What

Per review feedback, the rule id now travels inside the error envelope's general-purpose `param` field rather than a new top-level field — the same pattern as `model_not_found` → `param.modelId`:

- The envelope schema is **unchanged** (`param` is already `z.unknown()`).
- `forbiddenParamSchema` (`{ ruleId: string }`) mirrors `modelNotFoundParamSchema`; the `forbidden` case mines `param` with `safeValidateTypes` and leaves `ruleId` undefined on any unexpected shape.
- `GatewayForbiddenError`: optional `readonly ruleId` property.

```ts
catch (e) {
  if (GatewayForbiddenError.isInstance(e)) {
    console.log(e.ruleId); // 'rule_abc123'
  }
}
```

An SDK change is still needed for access: in released versions `param` is validated but not exposed on error instances (only `model_not_found` mines it, into `modelId`), so without this PR callers can't reach the rule id regardless of the wire encoding.

Companion server change emitting `param: { ruleId }` on deny responses: vercel/ai-gateway#2515 (supersedes vercel/ai-gateway#2312, closed — its top-level `ruleId` emission had already landed on main via vercel/ai-gateway#2414 and #2515 moves it into `param`). Merge order: server first; in the gap this PR just reads `undefined`.

## Validation

Unit tests: `forbidden` with no `param` → `undefined`; `param: { ruleId }` → exposed on the error; `param` of an unexpected shape (string) → `undefined`.

## Note

Targets `main` (gateway v4 / AI SDK v7). Backports to `release-v6.0` / `release-v5.0` follow the same path as the original `GatewayForbiddenError` (vercel/ai#16105/#16106/#16108) — left for you per usual.

### DIFF
--- a/packages/gateway/src/errors/create-gateway-error.ts
+++ b/packages/gateway/src/errors/create-gateway-error.ts
@@ -116,6 +116,7 @@ export async function createGatewayErrorFromResponse({
         statusCode,
         cause,
         generationId,
+        ruleId: validatedResponse.error.ruleId ?? undefined,
       });
     default:
       return new GatewayInternalServerError({
@@ -135,6 +136,7 @@ const gatewayErrorResponseSchema = lazySchema(() =>
         type: z.string().nullish(),
         param: z.unknown().nullish(),
         code: z.union([z.string(), z.number()]).nullish(),
+        ruleId: z.string().nullish(),
       }),
       generationId: z.string().nullish(),
     }),
--- a/packages/gateway/src/errors/create-gateway-error.ts
+++ b/packages/gateway/src/errors/create-gateway-error.ts
@@ -9,7 +9,10 @@ import {
 } from './gateway-model-not-found-error';
 import { GatewayInternalServerError } from './gateway-internal-server-error';
 import { GatewayFailedDependencyError } from './gateway-failed-dependency-error';
-import { GatewayForbiddenError } from './gateway-forbidden-error';
+import {
+  GatewayForbiddenError,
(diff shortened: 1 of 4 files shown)


packages/gateway/src/errors/create-gateway-error.ts
```diff
@@ -135,6 +136,7 @@ const gatewayErrorResponseSchema = lazySchema(() =>
         type: z.string().nullish(),
         param: z.unknown().nullish(),
         code: z.union([z.string(), z.number()]).nullish(),
+        ruleId: z.string().nullish(),
```
shaper: Can we put the rule id in `param` instead, and then maybe we don't need the PR if we emit the rule id in `param` on the server?

The spirit of this schema and base error type with the fields is to be general purpose across error types with enough basic fields to allow for this kind of flexibility.

R-Taneja: https://github.com/vercel/ai-gateway/pull/2515

shaper (approved):

PR #16260, opened 2026-06-18
Add Gemini Live Translate support

31Carlton7:
## Background

Google's `gemini-3.5-live-translate-preview` works through the Gemini Live API, but it requires a provider-specific `translationConfig` inside `generationConfig`. The Google realtime model accepts arbitrary string model IDs today, so the missing piece was not model-ID typing; it was mapping the translation configuration into Google's setup payload.

## What changed

- Adds typed `GoogleRealtimeModelOptions.translationConfig` with target-language and echo controls.
- Maps `providerOptions.google.translationConfig` into Gemini Live's `generationConfig` for both token creation and session updates.
- Preserves raw `generationConfig` provider options while merging translation settings.
- Exports the realtime model ID/options types from `@ai-sdk/google`.
- Documents `gemini-3.5-live-translate-preview` usage.
- Adds a dedicated press-to-translate example and a setup-route entry that omits unsupported tools for this model.
- Adds a patch changeset for `@ai-sdk/google`.

## Developer impact

Developers can use the normal AI SDK realtime API with `gemini-3.5-live-translate-preview` and configure the target language through `providerOptions.google.translationConfig`, without constructing Google wire payloads themselves.

## Validation

- `pnpm --filter @ai-sdk/google test:node -- packages/google/src/realtime/google-realtime-event-mapper.test.ts packages/google/src/realtime/google-realtime-model.test.ts` — 23 files, 665 tests passed
- `pnpm --filter @example/next-openai exec tsc --noEmit`
- `pnpm exec ultracite check` on changed TypeScript/TSX files
- `git diff --check`
- Previously verified the demo against the live Google API: token setup returned 200 and the realtime WebSocket connected successfully

## Future work

AI Gateway's realtime client plumbing is already merged, but translation-specific Gateway routing remains separate from this direct Google provider change.

### DIFF
--- a/packages/google/src/index.ts
+++ b/packages/google/src/index.ts
@@ -57,8 +57,8 @@ export type {
 export { GoogleRealtimeModel as Experimental_GoogleRealtimeModel } from './realtime/google-realtime-model';
 export type { GoogleRealtimeModelConfig as Experimental_GoogleRealtimeModelConfig } from './realtime/google-realtime-model';
 export type {
-  GoogleRealtimeModelId,
-  GoogleRealtimeModelOptions,
+  GoogleRealtimeModelId as Experimental_GoogleRealtimeModelId,
+  GoogleRealtimeModelOptions as Experimental_GoogleRealtimeModelOptions,
 } from './realtime/google-realtime-model-options';
 
 export { VERSION } from './version';
--- a/packages/google/src/index.ts
+++ b/packages/google/src/index.ts
@@ -56,5 +56,9 @@ export type {
 } from './google-provider';
 export { GoogleRealtimeModel as Experimental_GoogleRealtimeModel } from './realtime/google-realtime-model';
 export type { GoogleRealtimeModelConfig as Experimental_GoogleRealtimeModelConfig } from './realtime/google-realtime-model';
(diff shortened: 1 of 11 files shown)


packages/google/src/index.ts
```diff
@@ -56,5 +56,9 @@ export type {
 } from './google-provider';
 export { GoogleRealtimeModel as Experimental_GoogleRealtimeModel } from './realtime/google-realtime-model';
 export type { GoogleRealtimeModelConfig as Experimental_GoogleRealtimeModelConfig } from './realtime/google-realtime-model';
+export type {
+  GoogleRealtimeModelId,
+  GoogleRealtimeModelOptions,
+} from './realtime/google-realtime-model-options';
```
gr2m, line 62: Needs to be exported as experimental. Docs and example need to be updated accordingly

```suggestion
export type {
  GoogleRealtimeModelId as Experimental_GoogleRealtimeModelId,
  GoogleRealtimeModelOptions as Experimental_GoogleRealtimeModelOptions,
} from './realtime/google-realtime-model-options';
```

gr2m (approved):

PR #16244, opened 2026-06-17
feat(harness-grok-build): implement harness adapter for Grok build

mlekhi:
## Checklist

- [ ] All commits are signed (PRs with unsigned commits cannot be merged)
- [ ] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [ ] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [ ] I have reviewed this pull request (self-review)

PR #16261, opened 2026-06-19
feat(harness-deepagents): implement harness adapter for Deep Agents

mlekhi:
## Background

DeepAgents is a TS agent runtime. This is the harness adapter for DeepAgents - `@ai-sdk/harness-deepagents`. Runs the `deepagents` package through a Node bridge in the sandbox.

- Built-in tools: read, write, edit, bash, grep, glob, ls, task, write_todos
- Host tools work
- Skills work
- Auth: Anthropic, OpenAI, Gateway
- Multi-turn, attach, resume
- Built-in tool approvals (approve/reject) via DeepAgents `interruptOn`

## Manual Verification

Ran the examples for real (model + sandbox):

- `examples/ai-functions/src/harness-agent/deepagents/*` — all worked
- Approvals: approve → tool runs; reject → tool doesn't run, returns "Rejected by user."
- TUI + Next apps: all variants worked
- 45 unit tests pass

## Checklist

- [x] All commits are signed
- [x] Tests added
- [x] Docs added
- [x] Changeset added
- [x] Self-reviewed


packages/harness-deepagents/package.json
```diff
@@ -0,0 +1,76 @@
...
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "default": "./dist/index.js"
+    }
+  },
+  "dependencies": {
+    "@ai-sdk/harness": "workspace:*",
+    "@ai-sdk/provider-utils": "workspace:*",
+    "ws": "^8.20.1",
```
felixarntz: see https://github.com/vercel/ai/pull/16244#discussion_r3439076282, this needs to be bumped to 8.21.0

packages/harness-deepagents/src/bridge/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@langchain/anthropic": "^1.0.0",
+    "@langchain/core": "^1.1.44",
+    "@langchain/langgraph": "^1.3.0",
+    "@langchain/openai": "^1.0.0",
+    "deepagents": "1.10.2",
+    "ws": "8.20.1",
```
felixarntz: see other comment

packages/harness-deepagents/src/bridge/index.ts
```diff
@@ -0,0 +1,274 @@
...
+          toolName: toCommonName(toolName),
+          result: output ?? null,
+        });
+      }
+      if (runId) activeToolRunIds.delete(runId);
+    }
+  }
+
+  endTextBlock();
+  if (reasoningBlockId) emit({ type: 'reasoning-end', id: reasoningBlockId });
```
felixarntz: closing reasoning blocks has to also happen whenever any other kind of content (e.g. new text or tool call) is encountered. so it would make sense to refactor this to happen via a `endReasoningBlock()` helper just like `endTextBlock()` and call it everywhere it's relevant.

packages/harness-deepagents/src/bridge/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "harness-deepagents-bridge",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@langchain/anthropic": "^1.0.0",
+    "@langchain/core": "^1.1.44",
+    "@langchain/langgraph": "^1.3.0",
+    "@langchain/openai": "^1.0.0",
```
felixarntz: what are the `@langchain/anthropic` and `@langchain/openai` dependencies used for?

packages/harness-deepagents/src/deepagents-harness.ts
```diff
@@ -0,0 +1,678 @@
...
+    description: 'Execute a shell command',
+    inputSchema: z.object({ command: z.string() }),
+  }),
+  grep: commonTool('grep', {
+    nativeName: 'search',
+    toolUseKind: 'readonly',
+    description: 'Search file contents with regex',
+    inputSchema: z.object({ pattern: z.string() }),
+  }),
+} as const satisfies Record<string, HarnessV1BuiltinTool<any, any>>;
```
felixarntz, line 157: this is missing some tools, see https://docs.langchain.com/oss/javascript/deepagents/tools#built-in-harness-tools

it's critical they are listed here because otherwise it can lead to errors with AI SDK tool validation, for example I got this when running the `with-skills.ts` example:
```json
{
  "name": "AI_NoSuchToolError",
  "toolName": "glob",
  "availableTools": [
    "read",
    "write",
    "bash",
    "grep"
  ]
}
```

packages/harness-deepagents/src/deepagents-harness.ts
```diff
@@ -0,0 +1,678 @@
...
+  abortSignal?: AbortSignal;
+}): Promise<void> {
+  const combined = skills
+    .map(skill => `## ${skill.name}\n${skill.description}\n\n${skill.content}`)
+    .join('\n\n---\n\n');
+  await sandbox.writeTextFile({
+    path: `${workDir}/.skills.md`,
+    content: combined,
+    abortSignal,
+  });
```
felixarntz: this is far from ideal, this means skills won't function in a decoupled way. but the bigger problem is that file references from skills don't work at all.

From what I can tell, `deepagents` supports skills properly, so we can rely on placing the skills in actual files, see https://docs.langchain.com/oss/javascript/deepagents/skills

ideally they're placed in `$HOME/.agents/skills` in the sandbox, most importantly they must not be placed in the work dir so that they don't interfere with any e.g. git repo that may be checked out there.

packages/harness-deepagents/src/bridge/index.ts
```diff
@@ -0,0 +1,274 @@
...
+    const skillsBlock = await readSkillsBlock();
+    agent = createDeepAgent({
+      // Defer to DeepAgents' own default when the host configured no model.
+      ...(start.model ? { model: parseModelName(start.model) } : {}),
+      tools: buildHostTools(start.tools),
+      backend: new LocalShellBackend({ rootDir: workdir }),
+      systemPrompt: buildSystemPrompt(start, skillsBlock) || undefined,
+      // Real instance (LangGraph rejects `true` for root graphs); gives multi-turn memory.
+      checkpointer: new MemorySaver(),
+    });
```
felixarntz, line 136: I believe it's possible to also handle tool approvals here, at least for custom tools. see https://docs.langchain.com/oss/javascript/deepagents/human-in-the-loop

examples/ai-functions/src/harness-agent/deepagents/generate-text.ts
```diff
@@ -0,0 +1,34 @@
+import { HarnessAgent } from '@ai-sdk/harness/agent';
```
felixarntz, line 1: can you also implement the other examples that other harness adapters have for better manual test coverage?

examples/harness-e2e-next/agent/harness/deepagents/basic-agent.ts
```diff
@@ -0,0 +1,30 @@
+import {
```
felixarntz, line 1: can you also implement the other agent configurations like the other harnesses have for both `harness-e2e-next` and `harness-e2e-tui`? so that we can also test more complex agents end-to-end

felixarntz (commented):
@mlekhi this works great based on initial testing! the main thing to look into further is proper skills setup, and then some minor fixes around tools. also worth checking whether we can support tool approvals at least for custom tools (like e.g. for Codex).

packages/harness-deepagents/src/bridge/package.json
mlekhi: these are the langchain provider packages, createDeepAgent loads them based on the model id. theyre not bundled by deepagents or langchain so without them it just errors at runtime. anthropic is the default, openai only kicks in for openai models. we can drop the openai one if we wanna stay anthropic-only for now.

packages/harness-deepagents/src/deepagents-harness.ts
mlekhi: addressed

packages/harness-deepagents/src/deepagents-harness.ts
mlekhi: addressed - resolve skill paths via the backend root

examples/harness-e2e-next/agent/harness/deepagents/basic-agent.ts
mlekhi: tui done

mlekhi: next done

packages/harness-deepagents/src/deepagents-harness.ts
```diff
@@ -0,0 +1,755 @@
...
+          sandbox: session,
+          workDir,
+          skills: startOpts.skills ?? [],
+          abortSignal: startOpts.abortSignal,
+        });
+      }
+      // Absolute path: LocalShellBackend (non-virtual) treats a leading-slash path as a real fs path, so a workDir-relative skills dir must be fully qualified.
+      const skillsPath = hasSkills
+        ? `${workDir}${SKILLS_SOURCE_PATH}`
+        : undefined;
```
felixarntz: skills should ideally be written into somewhere outside the work dir. the other harnesses put them into the suitable place in `$HOME` inside the sandbox. see e.g. Claude Code's harness uses `resolveSandboxHomeDir()` (you could probably copy that over here as is), then write to `$HOME/.agents/skills` and use that.

the reason they shouldn't go in the workdir is because the workdir may contain other code with skills already (e.g. the AI SDK Coding agent example clones `vercel/ai` into the workdir, in which case having skills in there already would lead to problems).

packages/harness-deepagents/src/bridge/index.ts
```diff
@@ -0,0 +1,386 @@
...
+  const interruptOn = buildInterruptOn(start.permissionMode);
+  if (!agent) {
+    agent = createDeepAgent({
+      // Defer to DeepAgents' own default when the host configured no model.
+      ...(start.model ? { model: parseModelName(start.model) } : {}),
+      tools: buildHostTools(start.tools),
+      backend: new LocalShellBackend({ rootDir: workdir }),
+      systemPrompt: start.instructions || undefined,
+      // Native skills loaded from the host-materialized source dir (on-demand, with working file refs).
+      ...(start.skillsPath ? { skills: [start.skillsPath] } : {}),
```
felixarntz: see the other comment, this would also need to point to the home dir then, e.g. `$HOME/.agents/skills`.

additionally though we should also send `<workdir>/.agents/skills`, so that skills that are part of code cloned into the workdir are also discovered (relevant e.g. for the AI SDK Coding example).

so maybe we need to change `start.skillsPath` to become `start.skillsPaths` and then pass both locations.

mlekhi: addressed - skills now write to `$HOME/.agents/skills` and we also pass `<workDir>/.agents/skills`

content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
```diff
@@ -0,0 +1,197 @@
...
+
+## Known Limitations
+
+- **Built-in tool approvals** are not supported yet. Use
+  `permissionMode: 'allow-all'`. Host-executed AI SDK tool approvals still work.
+- **Cross-process resume, turn continuation, and suspend/detach** are not
+  supported yet — Deep Agents holds conversation state in memory (LangGraph
+  `MemorySaver`), which does not survive a bridge restart. These methods throw
+  `HarnessCapabilityUnsupportedError`.
+- **Manual compaction** is not supported.
```
felixarntz: are these correct as limitations? as far as I can tell, at least built-in tool approvals and cross-process resume/continuation/suspend/detach is supported (which is great!)

felixarntz (approved):

PR #16284, opened 2026-06-20
Secure AI Gateway realtime client secrets

31Carlton7:
## What changed

- Exposed `allowedOrigins` on `gateway.experimental_realtime.getToken()` so browser redemption can be bound to trusted app origins.
- Pinned `quotaEntityId`, `tags`, and `user` from the server-side session configuration into the Gateway client secret.
- Rejected request-scoped BYOK credentials before minting; provider credentials must be configured in AI Gateway rather than sent through a browser session.
- Exported Gateway-specific realtime token option and factory types.
- Updated Gateway realtime documentation to describe the current single-use `vcst_` flow and its security boundaries.
- Removed direct ElevenAgents support and the demo changes from this PR.

## Why

Gateway should remain the credential and traffic boundary. Browser clients receive a short-lived, single-use token bound to an exact model and optionally an origin, while trusted attribution and quota options are fixed by the server. This avoids exposing broad provider credentials or allowing one client to select another customer resource.

The working ElevenAgents adapter has been preserved on `carlton/elevenlabs-agents-realtime` for a future PR. Actual ElevenLabs agent support can build on Gateway private-resource mapping and team-scoped BYOK without putting raw agent IDs or provider credentials in the public client contract.

## Validation

- `pnpm --filter @ai-sdk/gateway test:node` (467 passed)
- `pnpm --filter @ai-sdk/gateway test:edge` (462 passed, 5 skipped)
- `pnpm --filter @ai-sdk/gateway type-check`
- `pnpm --filter @ai-sdk/gateway build`
- `pnpm exec ultracite check ...` (0 warnings, 0 errors)
- `pnpm validate:docs` (493 MDX files)
- `git diff --check`

### DIFF
--- a/.changeset/calm-keys-listen.md
+++ b/.changeset/calm-keys-listen.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+feat(provider/gateway): bind realtime client secrets to origins and trusted provider options
--- a/content/providers/01-ai-sdk-providers/00-ai-gateway.mdx
+++ b/content/providers/01-ai-sdk-providers/00-ai-gateway.mdx
@@ -251,35 +251,35 @@ const model = gateway.experimental_realtime('openai/gpt-realtime-2');
 
 The Gateway normalizes realtime the same way it normalizes every other modality: your client speaks the [normalized AI SDK realtime protocol](/docs/ai-sdk-core/realtime) and the Gateway translates to and from the upstream provider server-side. This means the same client code works regardless of which provider backs the model.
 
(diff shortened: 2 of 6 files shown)


gr2m:
compare #15747. Tools work differently with Elevenlabs, if I recall correctly they need to be registered ahead of time, I wanted to investigate further if there is another way. Did the e2e Elevenalbs example work for you out of the box for you?

31Carlton7:
> compare #15747. Tools work differently with Elevenlabs, if I recall correctly they need to be registered ahead of time, I wanted to investigate further if there is another way. Did the e2e Elevenalbs example work for you out of the box for you?

Yes. They worked just fine for me. You can try out the example I left there.

EDIT: Taking a closer look at the PR you linked, my implementation doesn't work for tool calls, so will submit a couple new commits to get this up and running

PR #16293, opened 2026-06-22
Backport: feat(openai): add orchestration token usage details to Responses API usage

gr2m:
Backport of #16290 (and its v6.0 backport #16292) to the `release-v5.0` branch.

## Adaptation for v5.0

The original change relies on `usage.raw` (introduced with `LanguageModelV3Usage` in v6) to carry Sakana-style orchestration token usage through to downstream consumers. `release-v5.0` uses `LanguageModelV2Usage`, which has **no `raw` field** — its contract is that additional usage info is surfaced via provider metadata.

This backport therefore:

- Extends the Responses API zod schemas (chunk + response) with the orchestration token detail fields so they are not stripped during parsing.
- Surfaces them under `providerMetadata.openai.usage` (`orchestrationInputTokens`, `orchestrationInputCachedTokens`, `orchestrationOutputTokens`) in both `doGenerate` and `doStream`, so downstream consumers (e.g. the AI Gateway) can bill them.
- Adds a regression test asserting the orchestration fields appear in provider metadata.

`providerMetadata.<provider>.usage` matches existing v5.0 convention — anthropic, amazon-bedrock, perplexity already nest token data under `.usage`.

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.ts
@@ -47,6 +47,7 @@ import {
 } from './openai-responses-options';
 import { prepareResponsesTools } from './openai-responses-prepare-tools';
 import { getOpenAILanguageModelCapabilities } from '../openai-language-model-capabilities';
+import type { ResponsesUsageProviderMetadata } from './openai-responses-provider-metadata';
 
 export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
   readonly specificationVersion = 'v2';
@@ -687,6 +688,11 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
 
     const usage = response.usage!; // defined when there is no error
 
+    const orchestrationUsage = getOrchestrationUsageMetadata(usage);
+    if (orchestrationUsage != null) {
+      providerMetadata[providerKey].usage = orchestrationUsage;
+    }
+
     return {
       content,
       finishReason: mapOpenAIResponseFinishReason({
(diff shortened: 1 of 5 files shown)


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1436,3 +1450,46 @@ function mapWebSearchOutput(
       };
   }
 }
+
+/**
+ * Extracts orchestration token usage details (e.g. Sakana-style orchestration)
+ * from the Responses API usage so they can be surfaced via provider metadata.
+ * Returns `undefined` when no orchestration usage is present.
+ */
+function getOrchestrationUsageMetadata(
```
lgrammel, line 1459: does this only extract orchestration usage or all usage? if all usage, rename. if only orchestration usage, then line 1337 is potentially incorrect since it fully replaces usage.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -1323,6 +1333,10 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV2 {
               providerMetadata[providerKey].serviceTier = serviceTier;
             }
 
+            if (orchestrationUsage != null) {
+              providerMetadata[providerKey].usage = orchestrationUsage;
```
lgrammel, line 1337: why is this needed? we support raw usage for provider specific usage data

lgrammel: also we use raw usage now for provider specific usage; this may not be needed at all

lgrammel (approved):

packages/openai/src/responses/openai-responses-language-model.ts
gr2m: we use raw usage in v6+, but this pr is v5

packages/openai/src/responses/openai-responses-language-model.ts
gr2m: Yes, right now it only extracts orchestration usage for OpenAI

PR #16299, opened 2026-06-22
fix(harness-pi): return image content from the read tool

Devansh-awat:
## Background

The Pi harness (`@ai-sdk/harness-pi`) provides its own sandbox-backed `read` builtin tool in `pi-session.ts`. Its `execute` decodes **every** file as UTF-8 text:

```ts
const buf = await input.remoteOps.readBuffer(params.file_path);
return asPiToolResult(buf.toString('utf8'));
```

There is no image branch, so reading an image file (a screenshot, an uploaded attachment, etc.) feeds raw binary into the model as a garbled UTF-8 string. Consequences for vision-capable models:

- The model can't see the image — it only receives mojibake.
- A large image's binary bloats the context, which can trip compaction/limit errors and fail the turn.

This is a regression relative to stock `@earendil-works/pi-coding-agent`, whose own `read` tool returns `{ type: 'image', ... }` content for images.

## Change

Detect common image formats (png, jpeg, gif, webp) from their magic bytes and return them as image content parts — the same `(TextContent | ImageContent)[]` shape `AgentToolResult` already allows and that `pi-coding-agent` consumes:

```ts
return {
  content: [
    { type: 'text', text: `Read image file [${imageMimeType}]` },
    { type: 'image', data, mimeType: imageMimeType },
  ],
  details: undefined,
};
```

Non-image files are unchanged (still UTF-8 text). Images whose base64 payload exceeds an inline size cap (4.5MB, below common provider limits) fall back to a text note instead of being inlined, to avoid blowing the request size.

A changeset is included (`@ai-sdk/harness-pi`: patch).

## Notes / testing

- The change is purely additive (no behavior change for non-image reads).
- `buildBuiltinToolDefinition` is module-internal, so there's no existing unit test seam for the builtin `read` tool. Happy to add a test if maintainers prefer exporting a seam or testing through a higher-level path — guidance welcome.
- Verified end-to-end against a real sandbox + vision endpoint (Kimi K2.6): reading a PNG now yields a correct description instead of binary text. No resize/Photon dependency is pulled in; large images degrade gracefully.

### DIFF
--- a/packages/harness-pi/src/pi-session.ts
+++ b/packages/harness-pi/src/pi-session.ts
@@ -967,6 +967,44 @@ function asPiToolResult(text: string): AgentToolResult<unknown> {
   };
 }
 
+// Base64-encoded payload cap for inline images, kept below common provider
+// limits (e.g. Anthropic's 5MB). Larger images are reported as text instead.
+const INLINE_IMAGE_MAX_BASE64_BYTES = 4.5 * 1024 * 1024;
+
+// Detect common image formats from their magic bytes so the read tool can hand
+// them to the model as image content instead of decoding binary as UTF-8 text.
+function detectImageMimeType(buf: Buffer): string | undefined {
+  if (
+    buf.length >= 8 &&
+    buf[0] === 0x89 &&
+    buf[1] === 0x50 &&
+    buf[2] === 0x4e &&
+    buf[3] === 0x47
+  ) {
+    return 'image/png';
+  }
+  if (buf.length >= 3 && buf[0] === 0xff && buf[1] === 0xd8 && buf[2] === 0xff) {
+    return 'image/jpeg';
+  }
+  if (buf.length >= 6 && buf[0] === 0x47 && buf[1] === 0x49 && buf[2] === 0x46) {
+    return 'image/gif';
+  }
+  if (
+    buf.length >= 12 &&
+    buf[0] === 0x52 &&
+    buf[1] === 0x49 &&
+    buf[2] === 0x46 &&
+    buf[3] === 0x46 &&
+    buf[8] === 0x57 &&
+    buf[9] === 0x45 &&
+    buf[10] === 0x42 &&
+    buf[11] === 0x50
+  ) {
+    return 'image/webp';
+  }
+  return undefined;
+}
+
 async function maybeDenyPiBuiltinTool(input: {
   toolCallId: string;
   nativeName: (typeof PI_NATIVE_BUILTIN_NAMES)[number];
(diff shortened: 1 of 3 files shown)


packages/harness-pi/src/pi-session.ts
```diff
@@ -1020,6 +1058,22 @@ function buildBuiltinToolDefinition(input: {
...
+            }
+            return {
+              content: [
+                { type: 'text', text: `Read image file [${imageMimeType}]` },
+                { type: 'image', data, mimeType: imageMimeType },
+              ],
+              details: undefined,
+            };
+          }
           return asPiToolResult(buf.toString('utf8'));
```
felixarntz: we should modify the `asPiToolResult` helper to also handle non-text content, instead of manually constructing it for images.

felixarntz (changes requested):
@Devansh-awat Thank you for the PR! Code works, but there's a worthwhile improvement to make. Ideally you can also add a test to verify the added functionality.

packages/harness-pi/src/pi-session.ts
originalchou: Hi @felixarntz, I noticed that pi-session creates its own built-in tools for PI. However, these implementations are fairly minimal compared with the corresponding tools in pi-coding-agent.
For example, PI’s native read tool detects and resizes images, limits text output to 50 KB or 2,000 lines, and supports paginated reads.
Would it make sense for harness-pi to use an exported factory such as createReadToolDefinition from pi-coding-agent, while supplying harness-specific filesystem operations? Or is keeping the harness-pi tool implementations intentionally minimal?

felixarntz: @originalchou I don't think it's intentional, great point. we could definitely rely on exported tool factories, as long as that still allows our tools to read from the sandbox while the Pi agent itself runs on the host environment. This latter behavior is key for the implementation - agent runs on host env, but operates on sandbox.

Devansh-awat:
Hi @felixarntz, thanks for the review! Both points are addressed:

1. **`asPiToolResult` refactor** — the helper now accepts a full content-part array (`AgentToolResult<unknown>['content']`) instead of just a string, and all call sites (including the read tool's image branch) go through it, so there's no more hand-rolled `AgentToolResult` construction: https://github.com/vercel/ai/pull/16299/files#diff-d5a0b6c8fcaac2d4f5b1dd3ea309ba7e5ee4a1f7bd76dc20ceb84a56ee8a4d7e

2. **Regression test** — added a test in `packages/harness-pi/src/pi-session.test.ts` that mocks the sandbox's `readBinaryFile` to return PNG bytes and asserts the read tool resolves with a text part plus an `image` content part (`data` + `mimeType`), rather than decoding the file as UTF-8 text. It fails against the pre-fix code and passes with it.

Full harness-pi suite (`pnpm test:node`), type-check, and lint are green locally.

One note: I'm syncing the branch with the latest main next — since main has updated `.github/workflows/` files, that push needs a token with the `workflow` scope on my side, so there may be a short delay before those commits land on this PR. Happy to adjust anything else to get this over the line.

PR #16310, opened 2026-06-22
docs(skill): add harness-authoring gotchas to add-harness-package

mlekhi:
## Summary

Updates the `add-harness-package` skill with details learned while building `@ai-sdk/harness-deepagents` (and from review feedback on the in-flight harness PRs).

- **CHANGELOG.md is required** - `konsistent` fails without it.
- **Bridge dependency rules**: regenerate the bridge lockfile, keep import ↔ tsup `external` ↔ `src/bridge/package.json` in sync, include packages the runtime lazily imports (provider SDKs resolved per model id), and match shared dependency versions to the sibling harness packages rather than self-pinning.
- **Adapter**: don't hardcode a default model (defer to the runtime, keep `modelId` honest); handle unsupported `tools` (throw) / `instructions` (prepend, Codex-style); quote interpolated shell paths.
- **Auth**: use `getAiGatewayAuthFromEnv()` + accept `VERCEL_OIDC_TOKEN`; note per-provider gateway base-URL conventions.
- **Custom tools**: convert host-tool JSON Schema to the runtime's tool format recursively (or pass schema through) - flat conversion drops nested/array/enum guidance.
- **Tests**: `getBootstrap()` reads compiled bridge assets, so mock `fs` for those paths.
- **Release/validation**: first-release changeset is `major`; validate against a live sandbox early (unit tests + type-check miss runtime API constraints).

## Checklist

- [x] Docs-only change to the internal skill

### DIFF
--- a/skills/add-harness-package/SKILL.md
+++ b/skills/add-harness-package/SKILL.md
@@ -60,7 +60,7 @@ src/
     └── pnpm-lock.yaml
 ```
 
-Do not create a `CHANGELOG.md` file manually.
+Add a `CHANGELOG.md` containing just the package heading (`# @ai-sdk/harness-<name>`). `konsistent` requires every package to have one, and the changeset release process fills in the rest — without it, `pnpm konsistent` fails.
 
 ### 2. Configure package.json
 
@@ -79,6 +79,13 @@ Required package basics:
 - `"engines": { "node": ">=22" }`
 
 For bridge packages, add any bridge asset copy step required for files under `src/bridge/`.
+
+Bridge dependency rules (bridge-backed harnesses):
+
+- The bridge's runtime deps live in `src/bridge/package.json` (installed in-sandbox at bootstrap), not the main package.json. After changing them, regenerate `src/bridge/pnpm-lock.yaml` with `pnpm install --lockfile-only --ignore-workspace` in that folder.
+- For every third-party import in `src/bridge/`, keep three things in sync: the import, the `external` array in `tsup.config.ts`, and the dep in `src/bridge/package.json`. A missing entry shows up only at sandbox runtime as a module-resolution error.
+- Include packages the runtime *lazily* imports — e.g. provider SDKs (`@langchain/anthropic`, `@langchain/openai`) resolved from the model id at runtime — even though nothing imports them directly. These fail only when a model of that provider is actually used.
+- Match shared dependency versions (transport, schema, tooling, runtime SDKs) to what the other harness packages currently use — copy from a sibling package rather than choosing your own pins. Stale pins drift from security patches and can desync from the shared bridge runtime; check the current versions at creation time.
 
 ### 3. Create TypeScript, Build, and Test Configs
 
@@ -103,7 +110,10 @@ Use the architecture doc for contract details. At implementation time, verify:
 - expose adapter-native built-in tools through `builtinTools`;
 - keep construction synchronous and side-effect free;
 - use `startOpts.sandboxSession` and `startOpts.sessionWorkDir`; never create a separate sandbox;
-- throw `HarnessCapabilityUnsupportedError` from the method that needs an unsupported runtime capability.
+- throw `HarnessCapabilityUnsupportedError` from the method that needs an unsupported runtime capability;
+- don't hardcode a default model — pass the model only when the consumer configured one and let the runtime apply its own default, and keep the session's `modelId` consistent with what's actually sent (don't report a model the bridge silently overrode);
+- handle the `tools` and `instructions` that `doPromptTurn`/`doContinueTurn` may receive: if the runtime can't take custom `tools`, throw `HarnessCapabilityUnsupportedError` so it's obvious rather than silently dropped; if it has no native `instructions` input, prepend them to the first user message (the Codex/Claude Code workaround);
+- quote interpolated paths (`workDir`, bridge-state dir, …) when building shell commands for `sandbox.run`/`sandbox.spawn` — they can contain spaces.
 
 If the runtime needs in-sandbox setup, expose `getBootstrap()`.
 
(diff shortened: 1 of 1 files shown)


skills/add-harness-package/SKILL.md
```diff
@@ -60,7 +60,7 @@ src/
     └── pnpm-lock.yaml
 ```
 
-Do not create a `CHANGELOG.md` file manually.
+Add a `CHANGELOG.md` containing just the package heading (`# @ai-sdk/harness-<name>`). `konsistent` requires every package to have one, and the changeset release process fills in the rest — without it, `pnpm konsistent` fails.
```
felixarntz: maybe can shorten this to just:

```suggestion
Add a `CHANGELOG.md` containing just the package heading (`# @ai-sdk/harness-<name>`). Every package is required to have one.
```

the fact that `konsistent` fails without it isn't as important as it would later happen anyway

skills/add-harness-package/SKILL.md
```diff
@@ -80,6 +80,13 @@ Required package basics:
 
 For bridge packages, add any bridge asset copy step required for files under `src/bridge/`.
 
+Bridge dependency rules (bridge-backed harnesses):
+
+- The bridge's runtime deps live in `src/bridge/package.json` (installed in-sandbox at bootstrap), not the main package.json. After changing them, regenerate `src/bridge/pnpm-lock.yaml` with `pnpm install --lockfile-only --ignore-workspace` in that folder.
```
felixarntz: great point. FWIW I've been using:
```
pnpm --dir packages/<harness-package>/src/bridge install --lockfile-only --ignore-workspace
```

don't have to use it here, but it might be more obvious as it would allow running it from the repo root.

skills/add-harness-package/SKILL.md
```diff
@@ -80,6 +80,13 @@ Required package basics:
...
+- For every third-party import in `src/bridge/`, keep three things in sync: the import, the `external` array in `tsup.config.ts`, and the dep in `src/bridge/package.json`. A missing entry shows up only at sandbox runtime as a module-resolution error.
+- Include packages the runtime _lazily_ imports — e.g. provider SDKs (`@langchain/anthropic`, `@langchain/openai`) resolved from the model id at runtime — even though nothing imports them directly. These fail only when a model of that provider is actually used.
```
felixarntz: nit: not ideal to have both examples be `@langchain/*` packages. maybe better to use independent examples, like Claude Code SDK, OpenAI SDK for instance.

skills/add-harness-package/SKILL.md
```diff
@@ -103,15 +110,19 @@ Use the architecture doc for contract details. At implementation time, verify:
 - expose adapter-native built-in tools through `builtinTools`;
 - keep construction synchronous and side-effect free;
 - use `startOpts.sandboxSession` and `startOpts.sessionWorkDir`; never create a separate sandbox;
-- throw `HarnessCapabilityUnsupportedError` from the method that needs an unsupported runtime capability.
+- throw `HarnessCapabilityUnsupportedError` from the method that needs an unsupported runtime capability;
+- don't hardcode a default model — pass the model only when the consumer configured one and let the runtime apply its own default, and keep the session's `modelId` consistent with what's actually sent (don't report a model the bridge silently overrode);
```
felixarntz: I would clarify that: don't hardcode a default model _unless_ it's technically required. Some of the underlying SDKs may not even have their own default.

this point is mostly about leaving the original SDK's default untouched, that's why we shouldn't pass one unless there is no original default.

skills/add-harness-package/SKILL.md
```diff
@@ -103,15 +110,19 @@ Use the architecture doc for contract details. At implementation time, verify:
...
+- quote interpolated paths (`workDir`, bridge-state dir, …) when building shell commands for `sandbox.run`/`sandbox.spawn` — they can contain spaces.
 
 If the runtime needs in-sandbox setup, expose `getBootstrap()`.
 
 ### 5. Implement Runtime-Specific Concerns
 
 Add only the concerns the runtime needs:
 
-- auth resolution;
+- auth resolution — for AI Gateway support, use the central `getAiGatewayAuthFromEnv()` helper rather than reading env directly, and accept `VERCEL_OIDC_TOKEN` as a gateway credential (not just `AI_GATEWAY_API_KEY`). When the runtime resolves provider per model, resolve the provider from the model id and set that provider's env; if routing through the gateway, note that base-URL conventions differ per provider (e.g. an Anthropic client appends `/v1/messages` to a root base, an OpenAI client appends to a `/v1` base);
```
felixarntz:
```suggestion
- auth resolution — for AI Gateway support, use the central `getAiGatewayAuthFromEnv()` helper rather than reading env directly. This ensures both `VERCEL_OIDC_TOKEN` and `AI_GATEWAY_API_KEY` are accepted as Gateway credential. When the runtime resolves provider per model, resolve the provider from the model id and set that provider's env; if routing through the gateway, note that base-URL conventions differ per provider (e.g. an Anthropic client appends `/v1/messages` to a root base, an OpenAI client appends to a `/v1` base);
```

skills/add-harness-package/SKILL.md
felixarntz, line 72: unrelated to your changes, but this came to mind as I was reviewing:
```suggestion
- `"type": "module"`
- `"version": "0.0.0"` (starting point for new packages)
```

felixarntz (approved):
@mlekhi awesome, thank you for these enhancements! looks great, only a few minor nits to ideally fix before merge.

skills/add-harness-package/SKILL.md
mlekhi: addressed!

skills/add-harness-package/SKILL.md
mlekhi: addressed!

Issue #16063, opened 2026-06-12
docs: where would a sandboxed-tool example live (sibling to examples/mcp)?


telleroutlook:
## Context

I maintain [`@agentkit-js/aisdk`](https://www.npmjs.com/package/@agentkit-js/aisdk), a small package that produces AI SDK `tool()`-shaped objects whose execute path runs inside a WASM kernel (QuickJS / Pyodide / Wasmtime) or a remote microVM, instead of running JS directly in the host runtime. The package declares `ai ^4 || ^5 || ^6` as a peer dep and only uses `tool()` (structurally typed), so it works across the three majors without changes.

I'd like to add a runnable example that shows it in action. Before sending a PR, I want to ask about shape — `@anthropic-ai/sdk`'s docs maintainers ask the same thing on similar PRs and it's saved everyone time.

## What's the right home

I see three plausible places:

1. **`examples/mcp-agentkit/`** — sibling to [`examples/mcp/`](https://github.com/vercel/ai/tree/main/examples/mcp). Risk: `examples/mcp/` uses `experimental_createMCPClient` to *consume* an MCP server, which is a different shape than `tool() + sandbox-kernel`. Calling the new dir 'mcp-*' may be misleading.
2. **`examples/sandboxed-tools/`** — neutral name, doesn't claim MCP framing.
3. **A docs page under [`content/cookbook`](https://github.com/vercel/ai/tree/main/content/cookbook)** — if a runnable directory is too heavy and a recipe is more in line with how AI SDK teaches integrations.

I'd default to option 2 unless you'd prefer a different shape.

## What the example would do

- Build a `ToolRegistry` with two trivial agentkit tools (calculator + string formatter).
- Wrap them in `codeModeTool({ kernel: new QuickJSKernel(), tools })`.
- Run a single message through the AI SDK against `MockLanguageModelV2` (deterministic, no API key needed) showing one `execute_code` call instead of N direct calls.
- Print token counts side-by-side with a direct-tool-use baseline.

Total ~150 LOC + README.

## Why not directly PR

The [community providers page](https://github.com/vercel/ai/tree/main/content/providers/03-community-providers) is explicitly for **Language Model Providers** implementing the V4 spec; that's not us (we're a tool/sandbox runtime, not a model adapter). The `examples/` tree is the right level, but I want a maintainer's nod on placement before sending a directory.

## Token-saving claim, source

[...]

PR #16318, opened 2026-06-23
docs(examples): add examples/sandboxed-tools — codeModeTool with QuickJS WASM sandbox

telleroutlook:
## What this adds

A runnable example in `examples/sandboxed-tools/` that shows how to wrap AI SDK tools in a WASM sandbox using [`@wasmagent/aisdk`](https://www.npmjs.com/package/@wasmagent/aisdk).

## Two scripts, one task

Both scripts solve the same task (add 3+4, multiply by 2, format as currency) so readers can compare the patterns directly:

| Script | Pattern | Model round-trips | Tool slots in context | Sandbox |
|--------|---------|:-----------------:|:---------------------:|:-------:|
| `baseline.ts` | Direct tool calls | 3 | 3 | none |
| `sandboxed.ts` | `codeModeTool` | 1 | 1 | QuickJS WASM |

Both use **`MockLanguageModelV2`** — no API key required, CI-safe, fully deterministic.

## How it works

`codeModeTool` surfaces a single `execute_code` tool to the model. The model emits one JavaScript snippet; the QuickJS WASM kernel runs it — including any `callTool()` calls to downstream tools — and only the final result re-enters the model context.

```ts
const sandboxedExecute = codeModeTool({
  kernel: new QuickJSKernel({ timeoutMs: 5000 }),
  tools: registry,
  capabilities: { cpuMs: 5000 },
});

  model,
  tools: { execute_code: sandboxedExecute },
  prompt: 'Add 3 and 4, multiply the result by 2, then format it as currency.',
});
```

## When to use / when NOT to use

**Good fit:**
- Agent needs to chain ≥ 3 tool calls — fewer round-trips, lower token cost
- Execution isolation required: LLM code runs in WASM, not the host runtime
- Cross-runtime portability: Node, Vercel Edge, Cloudflare Workers

**Not a good fit:**
- Small tool count (< 3) where the indirection isn't worth it
- Tools must stream partial results between individual calls
- Audit requirements demand per-tool-invocation logging

## Checklist

- [x] No new dependency added to any AI SDK package
- [x] Uses `MockLanguageModelV2` — runs offline, no API key, CI-safe
- [x] Comparison table in README and script output
- [x] "When to use / when NOT to use" section in README
- [x] Provider-neutral framing — runtime pattern first, MCP as optional add-on
- [x] Apache-2.0 compatible

### DIFF
--- a/examples/sandboxed-tools/.env.example
+++ b/examples/sandboxed-tools/.env.example
@@ -1,0 +1,4 @@
+# No API key required — this example uses MockLanguageModelV2 and runs fully offline.
+# To test with a real provider, set one of these and update the model in src/*.ts:
+# OPENAI_API_KEY=""
+# ANTHROPIC_API_KEY=""
--- a/examples/sandboxed-tools/README.md
+++ b/examples/sandboxed-tools/README.md
@@ -1,0 +1,87 @@
+# Sandboxed Tools Example
+
+Run tool-calling agents where the LLM-emitted code executes inside a WASM
+sandbox instead of the host runtime — using
+[`@wasmagent/aisdk`](https://www.npmjs.com/package/@wasmagent/aisdk) and the
+AI SDK's `tool()` primitive.
+
+## What this example shows
+
+Two runnable scripts solve the same task (add 3 + 4, multiply by 2, format as
+currency) so you can compare the two patterns side-by-side:
+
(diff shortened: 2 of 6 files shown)


aayush-kapoor (commented):
marking as closed; see https://github.com/vercel/ai/issues/16063#issuecomment-4836395041

PR #11379, opened 2025-12-22
v6 release

gr2m:
following instructions at https://github.com/changesets/changesets/blob/3cb9982b96aef8e2b46860258d56dbd7a2a02953/docs/prereleases.md

PR #16357, opened 2026-06-24
v7 release

gr2m:
ref:
- https://github.com/vercel/ai/pull/11379
- [contributing/releases.md](https://github.com/vercel/ai/blob/77cc1af6a5c2abefe58f960b3e114e9cc4402464/contributing/releases.md#beta--pre-release-cycle)


tools/konsistent-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "konsistent-provider",
-  "version": "0.0.2-canary.0",
+  "version": "0.0.2",
```
felixarntz, line 3: should this even have a version? it's not a published package. same goes for the changelog. those may have slipped through from when I had initially implemented it as a public package in `packages/`. but now that it's private, I don't think these are needed?

felixarntz (approved):
one tiny thing

tools/konsistent-provider/package.json
gr2m: it shouldn't, but the canary suffix seemd wrong. I'll remove the version after v7 release

PR #16378, opened 2026-06-25
fix(gateway): skip unsupported model settings types

http-samc:
Fixes the release-v6.0 Gateway model settings update workflow when the Gateway model list includes a model type that the release branch does not support yet.

The generator now skips model types that are not explicitly present in `MODALITY_CONFIG` instead of deriving a new settings filename and throwing when that file does not exist. Configured modalities still fail if their expected output file is missing.

Failure observed in https://github.com/vercel/ai/actions/runs/28190627304/job/83503978455:

```text
Error: Output file does not exist for type 'transcription': gateway-transcription-model-settings.ts
```

Verification:
- `git diff --check`

Could not run `pnpm --dir packages/gateway generate-model-settings` locally because this workspace does not have `node_modules` installed (`tsx: command not found`).

### DIFF
--- a/packages/gateway/scripts/generate-model-settings.ts
+++ b/packages/gateway/scripts/generate-model-settings.ts
@@ -73,20 +73,6 @@ function generateTypeFile(modelIds: string[], typeName: string): string {
   return lines.join('\n') + '\n';
 }
 
-function getModalityConfig(type: string): {
-  outputFile: string;
-  typeName: string;
-} {
-  if (MODALITY_CONFIG[type]) {
-    return MODALITY_CONFIG[type];
-  }
-  const capitalized = type.charAt(0).toUpperCase() + type.slice(1);
-  return {
-    outputFile: `gateway-${type}-model-settings.ts`,
-    typeName: `Gateway${capitalized}ModelId`,
-  };
-}
-
 async function main() {
   const response = await fetchModels();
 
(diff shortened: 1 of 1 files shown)


packages/gateway/scripts/generate-model-settings.ts
```diff
@@ -73,20 +73,6 @@ function generateTypeFile(modelIds: string[], typeName: string): string {
   return lines.join('\n') + '\n';
 }
 
-function getModalityConfig(type: string): {
```
http-samc, line 76: context as for why we removed:

Before:

- If `type` was in `MODALITY_CONFIG`, `getModalityConfig(type)` returned the configured `{ outputFile, typeName }`.
- If `type` was not in `MODALITY_CONFIG`, it invented:
  - `gateway-${type}-model-settings.ts`
  - `Gateway${CapitalizedType}ModelId`

After:

- If `type` is in `MODALITY_CONFIG`, it returns the exact same configured `{ outputFile, typeName }`.
- If `type` is not in `MODALITY_CONFIG`, it skips that model type.

So the only behavior removed is the fallback for unknown/future model types. That fallback is what caused the failure: for `transcription` on `release-v6.0`, it invented `gateway-transcription-model-settings.ts`, then immediately threw because the file does not exist on that branch.

Configured types still behave the same and still fail if their expected output file is missing, which is the useful guardrail.

kkdawkins (approved):

R-Taneja (approved):

PR #16386, opened 2026-06-25
fix(ai): prune orphaned tool-approval responses in pruneMessages

### LINKED ISSUE #16385: `pruneMessages` leaves orphaned `tool-approval-response` parts when pruning a specific tool by name
### Description

`pruneMessages` (`packages/ai/src/generate-text/prune-messages.ts`) is supposed to remove a tool's call/result/approval content as a consistent unit. When pruning **a specific tool by name** — i.e. `toolCalls: [{ type, tools: ['my-tool'] }]` — it correctly removes the tool's `tool-call` and `tool-approval-request`, but **leaves the matching `tool-approval-response` behind**, producing an orphaned approval response with no corresponding request.

This breaks the very invariant `pruneMessages` exists to maintain (consistent, paired tool content), and the leftover approval response can cause provider validation errors and confuses the model about the tool-approval lifecycle.

## Root cause

`tool-approval-response` parts carry only an `approvalId` — they do **not** carry a `toolName`. The only way to know which tool a response belongs to is to follow `approvalId → tool-approval-request → toolCallId → tool-call.toolName`.

However, a `tool-approval-response` lives in a `tool` message, while its `tool-approval-request` (and the originating `tool-call`) live in a **separate, earlier `assistant` message**. The current code builds the `approvalId → toolName` lookup **per message**, inside the `content.filter` callback:

```ts
// packages/ai/src/generate-text/prune-messages.ts (before fix)
messages = messages.map((message, messageIndex) => {
  // ...
  const toolCallIdToToolName: Record<string, string> = {};
  const approvalIdToToolName: Record<string, string> = {};

  return {
    ...message,
    content: message.content.filter(part => {
      // ...
        toolCallIdToToolName[part.toolCallId] = part.toolName;
      } else if (part.type === 'tool-approval-request') {
        approvalIdToToolName[part.approvalId] =
          toolCallIdToToolName[part.toolCallId];
      }
      // ...
      return (
        toolCall.tools != null &&
        !toolCall.tools.includes(
          part.type === 'tool-call' || part.type === 'tool-result'
            ? part.toolName
            : approvalIdToToolName[part.approvalId], // ← undefined for responses
        )
      );
    }),
  };
});
```

When the filter reaches a `tool-approval-response` inside a `tool` message, `approvalIdToToolName` is **empty** (the request was processed in a different message), so `approvalIdToToolName[part.approvalId]` is `undefined`. The keep condition then evaluates `!toolCall.tools.includes(undefined)` → `true`, so the response is **kept** even though its tool was explicitly selected for pruning.

## Reproduction

```ts
import { pruneMessages } from 'ai';
import type { ModelMessage } from 'ai';

  { role: 'user', content: [{ type: 'text', text: 'Weather in Tokyo and Busan?' }] },
  {
    role: 'assistant',
    content: [
      { type: 'tool-call', toolCallId: 'call-1', toolName: 'get-weather-tool-1', input: '{"city":"Tokyo"}' },
      { type: 'tool-call', toolCallId: 'call-2', toolName: 'get-weather-tool-2', input: '{"city":"Busan"}' },
      { type: 'tool-approval-request', toolCallId: 'call-2', approvalId: 'approval-1' },
    ],
  },
  {
    role: 'tool',
    content: [
      { type: 'tool-approval-response', approvalId: 'approval-1', approved: true },
      { type: 'tool-result', toolCallId: 'call-1', toolName: 'get-weather-tool-1', output: { type: 'text', value: 'sunny' } },
      { type: 'tool-result', toolCallId: 'call-2', toolName: 'get-weather-tool-2', output: { type: 'text', value: 'rainy' } },
    ],
  },
  { role: 'assistant', content: [{ type: 'text', text: 'done' }] },
];

const pruned = pruneMessages({
  messages,
  toolCalls: [{ type: 'all', tools: ['get-weather-tool-2'] }],
});
```

### Actual (buggy) output

The `tool` message still contains the orphaned approval response, even though every other trace of `get-weather-tool-2` (its `tool-call`, `tool-approval-request`, and `tool-result`) was removed:

```jsonc
{
  "role": "tool",
  "content": [
    { "type": "tool-approval-response", "approvalId": "approval-1", "approved": true }, // ← orphan
    { "type": "tool-result", "toolCallId": "call-1", "toolName": "get-weather-tool-1", "output": { "type": "text", "value": "sunny" } }
  ]
}
```

### Expected output

The approval response is pruned together with its request and tool-call; only `get-weather-tool-1` content survives.

## Impact

- **Severity:** Medium. Produces structurally inconsistent message histories from a helper whose explicit purpose is to keep them consistent.
- Orphaned approval responses can trigger provider-side validation errors (mismatched approval lifecycle) and waste tokens on content the caller asked to remove.
- Scope: any user combining tool approvals with selective (`tools: [...]`) pruning. Whole-tool pruning (`tools` omitted) is unaffected because responses are then resolved as "prune all".

## Suggested fix

Resolve `approvalId → toolName` from a **global** lookup built across all messages (not per message), so an approval response in a `tool` message can find its request in the originating `assistant` message. Requests and responses are then pruned together.

## Environment

- Package: `ai`
- File: `packages/ai/src/generate-text/prune-messages.ts`
- Affects the current `main`.

### AI SDK Version

ai package version is 7.0.0

### Code of Conduct

Abuhaithem:
`pruneMessages` trims tool-related content from a message history while keeping it internally consistent (calls, results, and approvals are meant to be pruned together as a unit). When pruning **a specific tool by name** — `toolCalls: [{ type, tools: ['my-tool'] }]` — it removed the tool's `tool-call`, `tool-approval-request`, and `tool-result`, but **left the matching `tool-approval-response` behind**, producing an orphaned approval response with no corresponding request.

A `tool-approval-response` carries only an `approvalId` (no `toolName`), so its tool name can only be derived via `approvalId → tool-approval-request → toolCallId → tool-call.toolName`. The response lives in a `tool` message while its request lives in a **separate, earlier `assistant` message** — but the `approvalId → toolName` lookup was built **per message**, inside the `content.filter` callback. So when a response was evaluated the lookup was empty, `approvalIdToToolName[part.approvalId]` was `undefined`, and the keep condition `!tools.includes(undefined)` evaluated to `true` — keeping the response.

This is a realistic trap: any caller combining tool approvals with selective pruning ends up with a structurally inconsistent history (an approval response with no request), which can trigger provider-side validation errors and confuses the model about the tool-approval lifecycle.

```ts
const pruned = pruneMessages({
  messages, // assistant: tool-call(call-2/tool-2) + approval-request(approval-1)
            // tool:      approval-response(approval-1) + tool-result(call-2)
  toolCalls: [{ type: 'all', tools: ['get-weather-tool-2'] }],
});

// Before: tool message still contains approval-response(approval-1) — orphaned,
//         even though call-2, its approval-request, and its tool-result are gone.
// After:  approval-response(approval-1) is pruned together with its request and call.
```

### Changes

- **Resolve tool names globally, not per message.** Two lookups (`toolCallId → toolName` and `approvalId → toolName`) are now built across **all** messages once per `toolCalls` config iteration, before filtering. An approval response in a `tool` message can therefore find its request in the originating `assistant` message, so requests and responses are pruned together.
- Removed the per-message map declarations and the in-filter "track tool calls and approvals" block; the filter now reads the global maps. The keep logic, the protected trailing-window path (`keptToolCallIds` / `keptApprovalIds`), and whole-tool pruning (`tools` omitted) are unchanged.
- **Tests** asserting the full pruned output (inline snapshot) plus an invariant that every `tool-approval-response` in the result has a matching `tool-approval-request`.

No public API changes. The `undefined ⇒ keep` fallback is preserved for genuinely unresolvable parts, so the change only fixes the case where the name *should* have resolved.

**Files changed**
- `packages/ai/src/generate-text/prune-messages.ts` — build global `toolCallId`/`approvalId` → `toolName` maps; remove per-message resolution
- `packages/ai/src/generate-text/prune-messages.test.ts` — regression test for selective tool pruning with approvals
- `.changeset/prune-messages-orphaned-approval-response.md` — patch changeset

## Related issue(s)

Fixes #16385

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Test update

## How it was tested

- **New regression test** (`prune-messages.test.ts` → `selective tool pruning with approvals`) asserts the orphaned approval response is gone and that every response in the output has a matching request.
- Existing `two tool settings` test (selective pruning that *keeps* an approval via the protected window) is unchanged and still passes.
- Full file suite green (8 tests, 0 type errors); `oxlint` and `oxfmt --check` clean.
- Run locally with:
  ```bash
  pnpm --filter ai exec vitest --config vitest.node.config.js --run src/generate-text/prune-messages.test.ts
  ```

## Checklist

- [ ] I have linked the related issue(s) in the description above _(pending — replace #XXXX once the issue is filed)_
- [x] I have made corresponding changes to the documentation (if applicable) _(none needed; internal helper behavior only)_
- [x] I have added tests that prove my fix is effective
- [x] I have added a changeset
- [ ] I have addressed all Coderabbit comments on this PR _(after PR is opened)_

## Diff summary

```diff
+    // Build global maps from tool call id and approval id to tool name.
+    // A `tool-approval-response` lives in a separate `tool` message from its
+    // `tool-approval-request`, so resolving names per-message left responses
+    // unresolved and orphaned. Resolve across all messages instead.
+    const toolCallIdToToolName: Record<string, string> = {};
+    for (const message of messages) {
+      if ((message.role === 'assistant' || message.role === 'tool') &&
+          typeof message.content !== 'string') {
+        for (const part of message.content) {
+          if (part.type === 'tool-call' || part.type === 'tool-result') {
+            toolCallIdToToolName[part.toolCallId] = part.toolName;
+          }
+        }
+      }
+    }
+    const approvalIdToToolName: Record<string, string> = {};
+    for (const message of messages) {
+      if ((message.role === 'assistant' || message.role === 'tool') &&
+          typeof message.content !== 'string') {
+        for (const part of message.content) {
+          if (part.type === 'tool-approval-request') {
+            approvalIdToToolName[part.approvalId] =
+              toolCallIdToToolName[part.toolCallId];
+          }
+        }
+      }
+    }
+
     messages = messages.map((message, messageIndex) => {
       ...
-      const toolCallIdToToolName: Record<string, string> = {};
-      const approvalIdToToolName: Record<string, string> = {};
-
       return {
         ...message,
         content: message.content.filter(part => {
           ...
-          // track tool calls and approvals:
-          if (part.type === 'tool-call') {
-            toolCallIdToToolName[part.toolCallId] = part.toolName;
-          } else if (part.type === 'tool-approval-request') {
-            approvalIdToToolName[part.approvalId] =
-              toolCallIdToToolName[part.toolCallId];
-          }
-
```

### DIFF
--- a/.changeset/prune-messages-orphaned-approval-response.md
+++ b/.changeset/prune-messages-orphaned-approval-response.md
@@ -1,0 +1,7 @@
+---
+'ai': patch
+---
+
+fix(ai): prune orphaned tool-approval responses in `pruneMessages`
+
+When pruning a specific tool by name (`toolCalls: [{ type, tools: [...] }]`), `pruneMessages` left the tool's `tool-approval-response` in place while removing its `tool-approval-request` and `tool-call`. The tool name of an approval response was resolved per-message, but approval responses live in a separate `tool` message from their approval request, so the name could never be resolved and the response was always kept. Tool name resolution is now done across all messages, so approval requests and responses are pruned together.
--- a/packages/ai/src/generate-text/prune-messages.test.ts
+++ b/packages/ai/src/generate-text/prune-messages.test.ts
@@ -716,5 +716,140 @@ describe('pruneMessages', () => {
         `);
       });
     });
+
+    describe('selective tool pruning with approvals (regression)', () => {
+      // Regression: a `tool-approval-response` lives in a separate `tool`
+      // message from its `tool-approval-request`. When pruning a specific tool
+      // by name, the response must be pruned together with its request and
+      // tool-call instead of being left orphaned.
+      it('should prune the approval response together with its request and tool-call', () => {
+        const messages: ModelMessage[] = [
+          {
+            role: 'user',
+            content: [{ type: 'text', text: 'Weather in Tokyo and Busan?' }],
+          },
+          {
+            role: 'assistant',
+            content: [
+              {
+                type: 'tool-call',
+                toolCallId: 'call-1',
+                toolName: 'get-weather-tool-1',
+                input: '{"city": "Tokyo"}',
+              },
+              {
+                type: 'tool-call',
+                toolCallId: 'call-2',
+                toolName: 'get-weather-tool-2',
+                input: '{"city": "Busan"}',
+              },
+              {
+                type: 'tool-approval-request',
+                toolCallId: 'call-2',
+                approvalId: 'approval-1',
+              },
+            ],
+          },
+          {
+            role: 'tool',
+            content: [
+              {
+                type: 'tool-approval-response',
+                approvalId: 'approval-1',
+                approved: true,
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call-1',
+                toolName: 'get-weather-tool-1',
+                output: { type: 'text', value: 'sunny' },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call-2',
+                toolName: 'get-weather-tool-2',
+                output: { type: 'text', value: 'rainy' },
+              },
+            ],
+          },
+          {
+            role: 'assistant',
+            content: [{ type: 'text', text: 'done' }],
+          },
+        ];
+
+        const result = pruneMessages({
+          messages,
+          toolCalls: [{ type: 'all', tools: ['get-weather-tool-2'] }],
+        });
+
+        // The approval response for the pruned tool must not survive as an
+        // orphan; only get-weather-tool-1 parts remain.
+        expect(result).toMatchInlineSnapshot(`
+          [
+            {
+              "content": [
+                {
+                  "text": "Weather in Tokyo and Busan?",
+                  "type": "text",
+                },
+              ],
+              "role": "user",
+            },
+            {
+              "content": [
+                {
+                  "input": "{"city": "Tokyo"}",
+                  "toolCallId": "call-1",
+                  "toolName": "get-weather-tool-1",
+                  "type": "tool-call",
+                },
+              ],
+              "role": "assistant",
+            },
+            {
+              "content": [
+                {
+                  "output": {
+                    "type": "text",
+                    "value": "sunny",
+                  },
+                  "toolCallId": "call-1",
+                  "toolName": "get-weather-tool-1",
+                  "type": "tool-result",
+                },
+              ],
+              "role": "tool",
+            },
+            {
+              "content": [
+                {
+                  "text": "done",
+                  "type": "text",
+                },
+              ],
+              "role": "assistant",
+            },
+          ]
+        `);
+
+        // No orphaned approval responses remain in the pruned output.
+        const approvalIds = new Set<string>();
+        const responseIds = new Set<string>();
+        for (const message of result) {
+          if (typeof message.content === 'string') continue;
+          for (const part of message.content) {
+            if (part.type === 'tool-approval-request') {
+              approvalIds.add(part.approvalId);
+            } else if (part.type === 'tool-approval-response') {
+              responseIds.add(part.approvalId);
+            }
+          }
+        }
+        for (const id of responseIds) {
+          expect(approvalIds).toContain(id);
+        }
+      });
+    });
   });
 });
--- a/packages/ai/src/generate-text/prune-messages.ts
+++ b/packages/ai/src/generate-text/prune-messages.ts
@@ -100,6 +100,42 @@ export function pruneMessages({
       }
     }
 
+    // Build global maps from tool call id and approval id to tool name.
+    // These must be global (not per-message) because a `tool-approval-response`
+    // lives in a separate `tool` message from its `tool-approval-request`
+    // (assistant message), so the tool name of a response can only be resolved
+    // by looking across messages. Resolving names per-message left responses
+    // unresolved, which caused them to be kept while their request was pruned,
+    // producing orphaned approval responses.
+    const toolCallIdToToolName: Record<string, string> = {};
+    for (const message of messages) {
+      if (
+        (message.role === 'assistant' || message.role === 'tool') &&
+        typeof message.content !== 'string'
+      ) {
+        for (const part of message.content) {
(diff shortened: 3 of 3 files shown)


rpelevin:
This lands the core shape I was looking for: resolve the approval lifecycle across the whole transcript before filtering any single message.

The remaining edge I would pin before merge is the unresolved approval-id case.

Right now the global approvalId lookup fixes the normal split-message path, but if an approval response has no matching request, or the request points at a toolCallId that cannot resolve to a tool name, the approvalId maps to no tool name. In selective pruning by tool name, that can turn into an implicit keep because the unresolved name is not in the pruned tool list.

I would make that branch explicit:

1. If the approvalId resolves to a pruned tool name, drop the response.
2. If the approvalId resolves to a surviving tool name, keep the response.
3. If the approvalId cannot resolve to any tool name, either drop it when pruning approval parts or surface a structural warning, but do not silently preserve it as authority context.

A small regression for a response with no matching request would close the loop nicely.

Boundary: architecture and regression-test feedback only; no claim about using this project or running its code.

itxaiohanglover:
Good fix! Moving the tool name resolution from per-message to across all messages is the right approach. The regression test with both tools and approval flow is thorough — covers the exact scenario from the issue. The changeset is clean too.

gr2m:
thanks all! I created an own PR because we require signed commits. I addressed your comments as well:
https://github.com/vercel/ai/pull/16415

Let me know if that works, I'll ship it later today

Abuhaithem:
> thanks all! I created an own PR because we require signed commits. I addressed your comments as well: #16415
> 
> Let me know if that works, I'll ship it later today

Great! If that works for you, we're happy to have contributed 🎉 to this great project.

Since the changes are now included in your PR, should we close this PR? @gr2m

PR #13060, opened 2026-03-04
fix(google): resolve thoughtSignature from vertex namespace on gateway failover

gr2m:
Fixes #11413

## Problem

When the Vercel AI Gateway fails over between Google Vertex and Google AI Studio (both serving the same Gemini model), `thoughtSignature` values get lost, causing the Gemini API to return a 400 error:

> Function call is missing a thought_signature in functionCall parts.

### Root cause

Both `@ai-sdk/google` and `@ai-sdk/google-vertex` share the same `GoogleGenerativeAILanguageModel` implementation but use different `providerOptionsName` values (`"google"` vs `"vertex"`). When storing `thoughtSignature` in the response, it's placed under `providerMetadata[providerOptionsName]`. When reading it back for the next API call, the lookup was **asymmetric**:

- **Vertex** (`providerOptionsName === 'vertex'`): checks `vertex`, falls back to `google` — worked
- **Google** (`providerOptionsName === 'google'`): checks `google` only — **no fallback to `vertex`**

So a Vertex → Google failover silently dropped the `thoughtSignature`, and the Gemini API rejected the request.

## Fix

Make the fallback bidirectional in `convertToGoogleGenerativeAIMessages`: when `providerOptionsName` is `"google"` and `providerOptions.google` is absent, fall back to `providerOptions.vertex`.

This is safe because:
- The fallback only affects `thoughtSignature` extraction
- Both providers serve the same Gemini models, so signatures are interchangeable
- The primary namespace always takes precedence

## Test plan

- [x] Added unit tests for the new Google → Vertex fallback
- [x] Existing tests still pass (258 tests)
- [ ] Run the reproduction example (`examples/ai-functions/src/stream-text/gateway/google-gemini-tool-call.ts`) to confirm the fix

PR #16392, opened 2026-06-25
fix(google): don't warn about missing thoughtSignature on Gemini 3 parallel function calls

### LINKED ISSUE #16298: Gemini 3 parallel function calls: converter injects `skip_thought_signature_validator` + warns every turn (only the first parallel call is signed)
## Summary

## Environment

- `ai@6.0.207`
- `@ai-sdk/google@3.0.83`
- `@ai-sdk/gateway@3.0.133`

## Reproduction

```ts
import { z } from 'zod'

const result = streamText({
  stopWhen: stepCountIs(3),
  tools: {
    get_weather: tool({
    }),
  },
})

    }
  }
}
```

## Observed output

```
WARNINGS:
[ { "type": "other",

TOOL-CALL PARTS:
```

## Expected

## Question for maintainers

akim136:
## Background

gemini 3 returns a single `thoughtSignature` for a parallel batch of function
calls — it's carried on the *first* `functionCall` part; the remaining parallel
calls legitimately have no signature (see
[thought signatures](https://ai.google.dev/gemini-api/docs/thought-signatures)).

#15560 added the `skip_thought_signature_validator` sentinel + a warning for
gemini-3 tool-call replays that lack a signature (the model returns HTTP 400
otherwise). but `convertToGoogleMessages` checks **every** tool-call part
independently, so for a parallel batch it injects the sentinel *and* warns for
each unsigned sibling — meaning any app doing parallel (or multi-step) tool calls
with gemini 3 gets the *"replayed N functionCall part(s) … without a
thoughtSignature … the likely cause is application code that drops
providerOptions.google.thoughtSignature"* warning on essentially every turn. the
suggested cause is misleading: the unsigned parts originate inside the SDK's own
parallel/multi-step loop and gemini never returned a signature for them.

## Summary

the sentinel injection is still required (it avoids the HTTP 400), so this keeps
injecting it. it only stops *warning* when the unsigned part is a legitimate
parallel sibling:

- `injectSkipSignature(toolName, countAsMissing)` always returns the sentinel, but
  only records the tool for the warning when `countAsMissing`.
- per assistant message, `hasSignedToolCall` checks whether any tool-call carries a
  real signature (via the existing `readProviderOpts`, so the google↔vertex
  namespace fallback from #13060 is preserved). unsigned siblings in a signed batch
  get the sentinel silently; a message whose tool-calls are *all* unsigned still
  warns (the genuine dropped-signature case — unchanged).

the change stays behind the existing `isGemini3Model` gate (derived from
`/^gemini-3[.-]/` upstream), so non-gemini-3 behavior is untouched.

(alternative considered: instead of suppressing the warning for parallel siblings,
reword it so it no longer blames application code. happy to switch to that if you
prefer — this PR takes the suppress-for-legitimate-parallel-calls approach.)

## Manual Verification

drove the converter with the message shape from the issue's repro — an assistant
turn with two parallel `tool-call` parts on a gemini-3 model, the first carrying
`providerOptions.google.thoughtSignature` and the second without — and confirmed
the output: the second call still carries the `skip_thought_signature_validator`
sentinel (so no HTTP 400), and no warning is emitted. on `main` the same input
emits the "replayed … without a thoughtSignature" warning.

`pnpm --filter @ai-sdk/google test` → 667 pass (node + edge); the new
parallel-batch test fails on `main` and passes with the fix.

(note: a full live `streamText` run against the Gemini API was not performed in my
environment — verification is at the converter level, which is where the bug and
fix live.)

## Checklist

## Related Issues

Fixes #16298
Refs #15560 (sentinel + warn-once), #13060 (google/vertex namespace fallback), #11413

### DIFF
--- a/.changeset/google-gemini3-parallel-thought-signature.md
+++ b/.changeset/google-gemini3-parallel-thought-signature.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/google": patch
+---
+
+fix(google): don't warn about a missing `thoughtSignature` on Gemini 3 parallel function calls (the model signs only the first call of a batch)
--- a/packages/google/src/convert-to-google-messages.test.ts
+++ b/packages/google/src/convert-to-google-messages.test.ts
@@ -1686,6 +1686,99 @@ describe('Gemini 3 missing thoughtSignature mitigation', () => {
     expect(onWarning.mock.calls[0][0].message).toContain('`weather`');
   });
 
+  it('does NOT warn for unsigned parallel function calls when a sibling in the same message is signed (Gemini 3)', () => {
+    // Gemini 3 returns a single thoughtSignature for a parallel batch, carried on
+    // the first functionCall; the remaining parallel calls are legitimately
+    // unsigned. The converter must not flag those as dropped signatures.
+    const onWarning = vi.fn();
+    const result = convertToGoogleMessages(
+      [
+        { role: 'user', content: [{ type: 'text', text: 'hi' }] },
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'tool-call',
+              toolCallId: 'tc_1',
+              toolName: 'weather',
+              input: { location: 'Paris' },
+              providerOptions: { google: { thoughtSignature: 'real_sig' } },
+            },
+            {
+              type: 'tool-call',
+              toolCallId: 'tc_2',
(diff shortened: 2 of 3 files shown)


lgrammel:
Thank you for the pull request. Issue #16298 was closed as completed. Closing this pull request.

PR #16400, opened 2026-06-25
fix (provider/google): include toolUsePromptTokenCount in input token usage

samjbobb:
## Background

When you call Gemini with a tool that runs a tool-use prompt — most commonly grounded `google_search` or `url_context` — the model's `usageMetadata` reports those tokens in a separate field, `toolUsePromptTokenCount`, on top of `promptTokenCount`. Both are billed at the regular input rate ([Google docs](https://ai.google.dev/api/generate-content#UsageMetadata)).

Today `@ai-sdk/google` drops this field:

1. `usageSchema` in `google-language-model.ts` doesn't declare `toolUsePromptTokenCount`, so zod strips it from the parsed response.
2. `convertGoogleUsage` only sums `promptTokenCount` into `inputTokens.total`.

## Impact

`usage.inputTokens.total` is silently lower than what Google bills. We hit this in production: on `gemini-2.5-flash` with `url_context`, for example a single Wikipedia-page fetch returned `promptTokenCount: 26`, `toolUsePromptTokenCount: 12504`, `totalTokenCount: 12633` — but `usage.inputTokens` surfaced **26**. Across 10 days of our traffic this under-reported ~96% of that model's input tokens, so any cost computed from the standard `LanguageModelV4Usage` shape was off by >20× on grounded / url-context calls. Non-tool calls reconcile exactly, so the gap is specific to server-side tool fetches.

## Fix

- Declare `toolUsePromptTokenCount` on `usageSchema`.
- Add it to `GoogleUsageMetadata` and include it in `inputTokens.total` / `inputTokens.noCache` in `convertGoogleUsage` (billed at the input rate). Non-grounded calls are unaffected.
- Adds `convert-google-usage.test.ts` covering basic, cached, reasoning, grounded, and cached+tool-use cases.

Supersedes #14575 by @JamieWoodbury (rebased onto the post-refactor `convert-google-usage.ts` / `google-language-model.ts`, with the changeset + unit tests). Credit to them for the original diagnosis.

### DIFF
--- a/.changeset/google-tool-use-prompt-tokens.md
+++ b/.changeset/google-tool-use-prompt-tokens.md
@@ -1,0 +1,15 @@
+---
+'@ai-sdk/google': patch
+---
+
+fix (provider/google): include `toolUsePromptTokenCount` in input token usage
+
+Gemini reports tokens consumed by tool-use prompts — most commonly grounded
+Google Search or URL context — in a separate `toolUsePromptTokenCount` field in
+`usageMetadata`, on top of `promptTokenCount`. Both are billed at the regular
+input rate, but the provider previously dropped this field: it was not declared
+in `usageSchema` (so zod stripped it) and `convertGoogleUsage` only summed
+`promptTokenCount` into `inputTokens.total`. Grounded calls therefore
+under-reported input tokens by the exact amount Google billed for the tool-use
+prompt. This adds the field to the schema and includes it in `inputTokens.total`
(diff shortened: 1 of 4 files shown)


itxaiohanglover:
Important fix! Dropping toolUsePromptTokenCount meant grounded calls under-reported input tokens. Adding it to both the schema and convertGoogleUsage is the right approach. Test coverage with null, basic, and grounded scenarios is thorough.

PR #16192, opened 2026-06-17
fix(ai): replace dynamic import() with loadBuiltinModule for diagnostics_channel

Aayush-engineer:
## Background
Dynamic `import('node:diagnostics_channel')` in `tracing-channel-publisher.ts`
breaks React Native / Expo apps using the Hermes JS engine. Hermes' compiler 
(`hermesc`) rejects dynamic `import()` syntax at compile time regardless of the 
specifier. The `isNodeRuntime()` guard prevents execution but not compilation, 
so the dead code still causes build failures.

Fixes #16191

## Summary
Replaced the dynamic `import('node:diagnostics_channel')` with the existing 
`loadBuiltinModule` helper already present in the codebase (used by 
`openTelemetryChannelSpanContext`). This approach uses 
`process.getBuiltinModule()` which is Hermes-safe and keeps identical Node.js 
behavior.

## Manual Verification
Verified that all 138 test files and 2867 tests pass with no type errors after 
the change.

## Checklist
- [ ] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues
Fixes #16191
closes https://github.com/vercel/ai/pull/16402

PR #16402, opened 2026-06-26
fix: prevent bundler from statically analyzing dynamic import for React Native compatibility

### LINKED ISSUE #16191: dynamic import("diagnostics_channel") breaks React Native / Hermes release builds
## Summary

Importing anything from `ai` (e.g. `DefaultChatTransport`) in a React Native / Expo app fails the Hermes release bundle with:

```
error: Invalid expression encountered
  diagnosticsChannelPromise = import(/* webpackIgnore: true */ "diagnostics_channel")
```

Reproduces on `ai@7.0.0-canary.161` and `7.0.0-canary.176`. The offending `import()` is still present on `main`.

## Offending code

`packages/ai/src/telemetry/tracing-channel-publisher.ts`:

```ts
async function loadDiagnosticsChannel(): Promise<
  DiagnosticsChannel | undefined
> {
  if (!isNodeRuntime()) {
    return undefined;
  }

  if (diagnosticsChannelPromise == null) {
    diagnosticsChannelPromise = (
      import(
        /* webpackIgnore: true */
        'node:diagnostics_channel'
      ) as Promise<DiagnosticsChannel>
    ).catch(() => undefined);
  }

  return diagnosticsChannelPromise;
}
```

## Root cause

1. **Hermes' compiler (`hermesc`) does not support dynamic `import()` syntax** — it rejects it regardless of the specifier. Minimal repro, no AI SDK involved:

   ```js
   // himport.js
   var p = import("foo");
   ```

   ```
   $ hermesc -emit-binary -out himport.hbc himport.js
   himport.js:1:9: error: Invalid expression encountered
   var p = import("foo");
           ^~~~~~~~~~~~~
   ```

2. Metro can't resolve the Node builtin, so it passes the raw `import()` straight through to Hermes.
3. `/* webpackIgnore: true */` only helps webpack; Metro/Hermes don't honor it.
4. The `isNodeRuntime()` guard prevents *execution* on RN, but the `import()` is still **compiled** as dead code, so the guard doesn't help the build.
5. The package ships no `react-native` / `browser` export condition, so React Native gets the Node-flavored build.

## Reproduction

In any Expo app that imports from `ai`:

```
npx expo export --platform ios
# → index.js: error: Invalid expression encountered
#   diagnosticsChannelPromise = import(/* webpackIgnore: true */ "node:diagnostics_channel")
```

## Impact

Any React Native / Expo + Hermes app that imports **anything** from `ai` v7 (even just `DefaultChatTransport`) fails its release bundle. The diagnostics/tracing-channel telemetry was introduced in #14251 and expanded in #15660, so all current v7 canaries are affected.

## Suggested fix

The same file already has a Hermes-safe builtin loader that avoids `import()` entirely — `loadBuiltinModule`, used by `openTelemetryChannelSpanContext`:

```ts
function loadBuiltinModule<T>(id: string): T | undefined {
  const processWithBuiltins = globalThis.process as
    | { getBuiltinModule?: (id: string) => unknown }
    | undefined;
  try {
    return processWithBuiltins?.getBuiltinModule?.(id) as T | undefined;
  } catch {
    return undefined;
  }
}
```

`loadDiagnosticsChannel` can use it instead of the dynamic `import()`, which removes the `import()` expression from the bundle while keeping Node behavior identical:

```ts
async function loadDiagnosticsChannel(): Promise<
  DiagnosticsChannel | undefined
> {
  if (!isNodeRuntime()) {
    return undefined;
  }

  if (diagnosticsChannelPromise == null) {
    diagnosticsChannelPromise = Promise.resolve(
      loadBuiltinModule<DiagnosticsChannel>('node:diagnostics_channel'),
    );
  }

  return diagnosticsChannelPromise;
}
```

Happy to open a PR for this.

### AI SDK Version

- ai: 7.0.0-canary.161 (also reproduces on 7.0.0-canary.176)
- @ai-sdk/react: 4.0.0-canary.164
- expo: SDK 56 (Hermes + Metro)

### Code of Conduct

zhsj0089944:
## Problem

Importing anything from `ai` in a React Native / Expo app breaks the Hermes release bundle:

```
error: Invalid expression encountered
  diagnosticsChannelPromise = import(/* webpackIgnore: true */ "diagnostics_channel")
```

Even though `isNodeRuntime()` guards the call, Metro bundler still tries to parse the static string literal `'node:diagnostics_channel'` in the `import()` expression.

## Fix

Use a string variable for the module name to prevent Metro from statically analyzing the import:

```diff
  if (diagnosticsChannelPromise == null) {
+   // Use string variable to prevent bundlers from statically analyzing the import
+   // which breaks React Native / Hermes builds
+   const moduleName: string = 'node:diagnostics_channel';
    diagnosticsChannelPromise = (
      import(
        /* webpackIgnore: true */
-       'node:diagnostics_channel'
+       moduleName
      ) as Promise<DiagnosticsChannel>
    ).catch(() => undefined);
  }
```

This is a common pattern for making dynamic imports bundler-safe while keeping the same runtime behavior.

Fixes #16191

### DIFF
--- a/packages/ai/src/telemetry/tracing-channel-publisher.ts
+++ b/packages/ai/src/telemetry/tracing-channel-publisher.ts
@@ -31,10 +31,13 @@ async function loadDiagnosticsChannel(): Promise<
   }
 
   if (diagnosticsChannelPromise == null) {
+    // Use string variable to prevent bundlers from statically analyzing the import
+    // which breaks React Native / Hermes builds
+    const moduleName: string = 'node:diagnostics_channel';
     diagnosticsChannelPromise = (
       import(
         /* webpackIgnore: true */
-        'node:diagnostics_channel'
+        moduleName
       ) as Promise<DiagnosticsChannel>
     ).catch(() => undefined);
   }


gr2m:
alternative fix in https://github.com/vercel/ai/pull/16192 - to be verified

PR #16404, opened 2026-06-26
fix: use InvalidResponseDataError instead of generic Error for audio format

### LINKED ISSUE #15814: `DefaultGeneratedAudioFile` throws a generic `Error` instead of `AISDKError` when audio format cannot be determined
In `packages/ai/src/generate-speech/generated-audio-file.ts`, when an audio format cannot be determined from the provider's media type, the constructor throws a plain JavaScript `Error` instead of an `AISDKError` subclass. There is even a `// TODO this should be an AI SDK error` comment on the line.

## Location

`packages/ai/src/generate-speech/generated-audio-file.ts`, lines 43–48:

```ts
if (!format) {
  // TODO this should be an AI SDK error
  throw new Error(
    'Audio format must be provided or determinable from media type',
  );
}
```

## Why this matters

All SDK errors should extend `AISDKError` so that:
- Callers can use `AISDKError.isInstance(e)` to distinguish SDK errors from unexpected runtime errors.
- Errors carry a consistent `name` and `marker` that middleware and error-reporting systems can rely on.
- The error is consistent with the rest of the SDK's error handling pattern.

This path is reachable when a provider returns audio data with a malformed media type (e.g. `"audio/"` — an empty subtype), leaving `format` as an empty string (falsy).

## Proposed fix

Replace the generic `Error` with `InvalidResponseDataError` from `@ai-sdk/provider`, which already exists for cases where a provider returns data the SDK cannot interpret:

```ts
import { InvalidResponseDataError } from '@ai-sdk/provider';

if (!format) {
  throw new InvalidResponseDataError({
    data: mediaType,
    message: `Could not determine audio format from media type: ${mediaType}`,
  });
}
```

This also removes the long-standing `// TODO` comment.

### AI SDK Version

- Reproduced on latest main (post-6.0.168).
- Affected package: ai (latest)

### Code of Conduct

zhsj0089944:
## Problem

In `DefaultGeneratedAudioFile`, when the audio format cannot be determined from the media type, the constructor throws a plain `Error` instead of an `AISDKError` subclass. There is even a `// TODO this should be an AI SDK error` comment on the line.

## Fix

Replace the generic `Error` with `InvalidResponseDataError` from `@ai-sdk/provider`:

```diff
+import { InvalidResponseDataError } from '@ai-sdk/provider';

  if (!format) {
-   // TODO this should be an AI SDK error
-   throw new Error(
-     'Audio format must be provided or determinable from media type',
-   );
+   throw new InvalidResponseDataError({
+     data: mediaType,
+     message: `Could not determine audio format from media type: ${mediaType}`,
+   });
  }
```

This makes the error consistent with the rest of the SDK's error handling pattern and allows callers to use `AISDKError.isInstance()` to distinguish SDK errors from unexpected runtime errors.

Fixes #15814

### DIFF
--- a/packages/ai/src/generate-speech/generated-audio-file.ts
+++ b/packages/ai/src/generate-speech/generated-audio-file.ts
@@ -2,6 +2,7 @@ import {
   DefaultGeneratedFile,
   type GeneratedFile,
 } from '../generate-text/generated-file';
+import { InvalidResponseDataError } from '@ai-sdk/provider';
 /**
  * A generated audio file.
  */
@@ -41,10 +42,10 @@ export class DefaultGeneratedAudioFile
     }
 
     if (!format) {
-      // TODO this should be an AI SDK error
-      throw new Error(
-        'Audio format must be provided or determinable from media type',
-      );
+      throw new InvalidResponseDataError({
+        data: mediaType,
+        message: `Could not determine audio format from media type: ${mediaType}`,
+      });
     }
 
     this.format = format;


Abuhaithem:
If u close this PR as I have already done it in #15815

and check whether the issue is claimed before starting to work on it

PR #16407, opened 2026-06-26
fix(openai): allow web_search_call.results include

### LINKED ISSUE #16346: Add support for web_search_call.results in OpenAI responses SDK
### Description

According to the OpenAI [docs](https://developers.openai.com/api/reference/resources/responses/methods/create#(resource)%20responses%20%3E%20(model)%20response_includable%20%3E%20(schema)%20%3E%20(member)%201), the enum `web_search_call.results` is supported as an `includes` option. AI-SDK does not currently support this.

### AI SDK Version

- @ai-sdk/openai: 3.0.74
- ai: 6.0.103

### Code of Conduct

miachillgood:
## Summary
- allow `web_search_call.results` in OpenAI Responses `include` provider options
- extend the OpenAI responses include type/schema so the option passes validation
- add a request-body test covering the new include value

## Why
`vercel/ai#16346` reports that OpenAI supports `web_search_call.results` as a valid include value, but the SDK currently rejects it because the option is missing from the local type/schema allowlist.

## Testing
- `git diff --check`
- could not run `pnpm test:node` in this checkout because local dependencies are not installed (`vitest: command not found`; `node_modules missing`)

Closes #16346

### DIFF
--- a/packages/openai/src/responses/openai-responses-api.ts
+++ b/packages/openai/src/responses/openai-responses-api.ts
@@ -43,6 +43,7 @@ export type OpenAIResponsesInputItem =
 
 export type OpenAIResponsesIncludeValue =
   | 'web_search_call.action.sources'
+  | 'web_search_call.results'
   | 'code_interpreter_call.outputs'
   | 'computer_call_output.output.image_url'
   | 'file_search_call.results'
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):
marking closed in favor of https://github.com/vercel/ai/pull/16424 - needs signed commit and cs. added you as a co-author

PR #16414, opened 2026-06-26
Backport: feat (video): add first-class `generateAudio` call option (#16222)

shaper:
Backport of #16222 to `release-v6.0`.

## Summary

Promotes audio generation to a first-class top-level `generateAudio` boolean on `experimental_generateVideo`, instead of requiring provider-specific passthrough options (previously buried under `vertex.generateAudio`, `klingai.sound`, `alibaba.audio`, `bytedance.generateAudio`).

## Backport notes

`release-v6.0` is on the `VideoModelV3` spec, whereas the original PR targeted `VideoModelV4` on `main`. Adaptations:

- Added `generateAudio` to the **v3** call-options type (`video-model-v3-call-options.ts`); the v4-only files from the PR were dropped.
- `generate-video.ts` keeps the v3 `satisfies Experimental_VideoModelV3CallOptions` shape with `generateAudio` passed through.
- `google-vertex-video-model.ts` keeps v6.0's `vertexOptions` variable name (renamed to `googleVertex` on main) and adds the top-level `generateAudio` override.
- Fixed one back-ported vertex test to use the v6.0 provider key `vertex` (the PR used the main-only `googleVertex` key, which caused a poll timeout).

## Verification

All affected suites pass with no type errors: google-vertex (234), alibaba (98), bytedance (43), klingai (92), gateway (417), plus fal/google/prodia/replicate/xai video tests and the `ai` generate-video + resolve-model suites.


gr2m:
Pushed a fix for the failing TypeScript check: the backported `vertex/generate-audio.ts` example imported the v7-only `googleVertex` provider and `GoogleVertexVideoModelOptions` type, which don't exist on `release-v6.0`. Switched it to the v6 `vertex` provider and `GoogleVertexVideoProviderOptions` type. Verified `type-check:full` and the affected package tests pass locally.

gr2m (approved):

PR #16417, opened 2026-06-26
fix(harness): fix harness Zod usage to be v3/v4 compatible

felixarntz:
## Background

The harness packages assumed Zod 3 was the installed Zod. Using them in a project on Zod 4 produced widespread type errors — most importantly, a `HarnessV1` instance (e.g. the `claudeCode` constant) was no longer assignable to the `harness` setting that `HarnessAgent` expects.

## Summary

The rest of AI SDK already supports both Zod majors, normalized to a version-agnostic `FlexibleSchema`. The harness packages diverged from this — `harness` declared Zod only as a dev dependency (a phantom runtime dependency), the adapters pinned it as a hard dependency (forcing a second, conflicting Zod copy into a consumer's install), and the source mixed bare `zod` imports with a Zod-3-only API.

This PR brings the harness packages in line with the established pattern.

- Declare Zod as a peer dependency (`^3.25.76 || ^4.1.8`) across `harness` and all five adapters, matching the provider packages
- Standardize every harness Zod import on `zod/v4` (resolves correctly under both ends of the peer range)
- Fix the one Zod-4-incompatible call: single-argument `z.record(...)` now passes an explicit key type
- Add a type-level test locking the `HarnessV1` ↔ `HarnessAgent` `harness` compatibility so the regression can't return

## Checklist

### DIFF
--- a/.changeset/deepagents-harness.md
+++ b/.changeset/deepagents-harness.md
@@ -1,5 +1,0 @@
----
-'@ai-sdk/harness-deepagents': major
----
-
-feat(harness-deepagents): add Deep Agents harness adapter 
--- a/.changeset/fast-berries-attack.md
+++ b/.changeset/fast-berries-attack.md
@@ -1,0 +1,10 @@
+---
+"@ai-sdk/harness-claude-code": patch
+"@ai-sdk/harness-deepagents": patch
+"@ai-sdk/harness-opencode": patch
+"@ai-sdk/harness-codex": patch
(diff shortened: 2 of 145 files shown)


felixarntz:
@gr2m I have verified via [snapshot releases](https://github.com/vercel/ai/actions/runs/28254978987/job/83715628267) that this fixes the downstream type errors I was encountering. This should be good to merge!

gr2m (approved):

PR #16426, opened 2026-06-26
fix: pass codex tool schemas by file

### LINKED ISSUE #16379: harness codex: spawn E2BIG when a large host tool catalog is passed through the codex config command line
### Description

When using `@ai-sdk/harness-codex` with a large host tool set (around 170 tools, each carrying a full JSON input schema), every turn fails immediately with `spawn E2BIG` and the model process never launches. The host stream shows an empty turn (zero steps, zero output), so to a consumer it looks like the turn silently completed with no content.

**Root cause**

The codex bridge places the host tools into `config.mcp_servers["harness-tools"].environment.TOOL_SCHEMAS` as one JSON string, then passes that config to `new Codex({ config })`. The codex SDK (`@openai/codex-sdk`) flattens the config object and serializes the whole MCP block into a single command line argument of the form:

```
--config mcp_servers.harness-tools.env.TOOL_SCHEMAS="<entire tool catalog JSON>"
```

With a large catalog that single argument string is well over 340 KB, which exceeds the Linux per argument limit (`MAX_ARG_STRLEN`, 128 KB). `spawn` throws `E2BIG` before codex runs.

**Captured stack (from the bridge event log)**

```
spawn E2BIG
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at spawn (node:child_process:787:9)
    at CodexExec.run (.../@openai/codex-sdk@0.130.0/dist/index.js:238:19)
    at runTurn (.../harness/codex/bridge.mjs:617:22)
    at handleInbound (.../harness/codex/bridge.mjs:261:11)
```

**Why it is hard to notice**

The real error frame is also lost on the host side. `waitForBridgeReady` acquires a reader on `proc.stdout` and does not release it, so the subsequent `drainRest(proc.stdout)` logs "Invalid state: The ReadableStream is locked" and the underlying error is swallowed. The turn appears to complete empty. The only reliable place the real error survives is the on disk bridge event log (`<workdir>/.agent-runs/<sessionId>/bridge/event-log.ndjson`). It would help a lot if a fatal bridge error were surfaced to the host stream rather than dropped.

### Reproduction

1. Build a `HarnessAgent` on the codex harness with a tool set whose combined serialized JSON schemas exceed roughly 128 KB (about 150 plus tools with non trivial input schemas).
2. Run any turn.
3. The turn ends with zero steps and no assistant content. The bridge event log contains the `spawn E2BIG` frame above.

### Suggested fix

Avoid putting the large tool catalog on the command line. Two options that both work and keep the full catalog:

1. Write `mcp_servers` to `$CODEX_HOME/config.toml` on disk (codex reads it natively per the OpenAI config reference) instead of passing it through `new Codex({ config })`.
2. Or write `TOOL_SCHEMAS` to a file and pass only a short file path to the MCP server, having the MCP server read the file.

We applied option 1 locally as a patch and codex now runs cleanly with the full catalog.

### AI SDK Version

- @ai-sdk/harness-codex: 1.0.0-beta.24
- @ai-sdk/harness: 1.0.0-beta.27
- @openai/codex-sdk (bundled by the bridge): 0.130.0
- Runtime: Node 22 in a Vercel Sandbox (Linux)

vicksiyi:
Summary
- Write harness-codex host tool schemas to a workspace file instead of embedding the full schema catalog in the child process environment.
- Pass TOOL_SCHEMAS_PATH to the host tool MCP bridge and keep TOOL_SCHEMAS as a backward-compatible fallback.

Tests
- pnpm --filter @ai-sdk/harness-codex type-check

Notes
- pnpm --filter @ai-sdk/harness-codex test currently fails before the changed bridge path because local workspace package resolution cannot load @ai-sdk/harness/utils / @ai-sdk/harness in this checkout.

Fixes #16379

### DIFF
--- a/packages/harness-codex/src/bridge/index.ts
+++ b/packages/harness-codex/src/bridge/index.ts
@@ -53,6 +53,8 @@ import { argv, env as procEnv, stdout } from 'node:process';
  */
 import * as codexSdkModule from '@openai/codex-sdk';
 
+const TOOL_SCHEMAS_FILENAME = '.ai-sdk-harness-tool-schemas.json';
+
 /*
  * Native Codex tool name → cross-harness common name. Tools outside this map
  * (e.g. MCP tools the model invokes by name) have no common equivalent; their
@@ -130,6 +132,18 @@ async function runTurn(start: StartMessage, turn: BridgeTurn): Promise<void> {
   let cliShimPath: string | undefined;
   if (start.tools && start.tools.length > 0) {
     cliShimPath = `${cliShimDir}/${CLI_SHIM_FILENAME}`;
+    const toolSchemasPath = `${workdir}/${TOOL_SCHEMAS_FILENAME}`;
+    await writeFile(
+      toolSchemasPath,
+      JSON.stringify(
+        start.tools.map(t => ({
+          name: t.name,
+          description: t.description,
+          inputSchema: t.inputSchema,
+        })),
+      ),
+      'utf8',
+    );
     relay = await startToolRelay({
       tools: start.tools,
       emit,
@@ -140,13 +154,7 @@ async function runTurn(start: StartMessage, turn: BridgeTurn): Promise<void> {
       command: 'node',
       args: [`${bootstrapDir}/host-tool-mcp.mjs`],
       env: {
-        TOOL_SCHEMAS: JSON.stringify(
-          start.tools.map(t => ({
-            name: t.name,
-            description: t.description,
-            inputSchema: t.inputSchema,
-          })),
-        ),
+        TOOL_SCHEMAS_PATH: toolSchemasPath,
         TOOL_RELAY_URL: `http://127.0.0.1:${relay.port}`,
       },
     };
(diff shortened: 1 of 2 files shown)


vicksiyi:
Rebased this PR on the latest `main` and resolved the harness-codex bridge conflict by keeping the current authorized relay/shim flow while moving tool schemas to `TOOL_SCHEMAS_PATH`.

Pushed commit `3e4b3cde20`. Verified with:
- `corepack pnpm --filter @ai-sdk/harness-codex type-check`
- `corepack pnpm --filter @ai-sdk/harness-codex test:node` (8 files, 76 tests)

packages/harness-codex/src/bridge/index.ts
```diff
@@ -130,6 +132,18 @@ async function runTurn(start: StartMessage, turn: BridgeTurn): Promise<void> {
   let cliShimPath: string | undefined;
   if (start.tools && start.tools.length > 0) {
     cliShimPath = `${cliShimDir}/${CLI_SHIM_FILENAME}`;
+    const toolSchemasPath = `${workdir}/${TOOL_SCHEMAS_FILENAME}`;
```
felixarntz, line 135: we should not write harness specific utilities to the primary `workdir`. `workdir` should be reserved exclusively for the agent to work in, it should only contain whatever the sandbox configuration defines to be in the `workdir` (via callbacks like `onBootstrap`).

Worth checking for other harness-specific directories that could be used, or maybe even write it into `cliShimDir` if feasible.

felixarntz (changes requested):
@vicksiyi Thank you for the PR! Mostly looks good, just one thing that needs to be fixed.

PR #16431, opened 2026-06-26
fix: pass opencode tool schemas by file

### LINKED ISSUE #16380: harness opencode: spawn E2BIG when a large host tool catalog is passed via OPENCODE_CONFIG_CONTENT, plus headless permission auto rejection
### Description

When using `@ai-sdk/harness-opencode` with a large host tool set (around 170 tools, each carrying a full JSON input schema), every turn fails immediately with `spawn E2BIG` and the opencode server never launches. The host stream shows an empty turn (zero steps, zero output), which looks like a silent completion to the consumer.

**Root cause**

The opencode bridge places the host tools into `config.mcp["harness-tools"].environment.TOOL_SCHEMAS` as one JSON string, then calls `createOpencodeServer({ config })`. The opencode SDK (`@opencode-ai/sdk`) launches the `opencode serve` child process with the entire config serialized into a single environment variable:

```js
env: { ...process.env, OPENCODE_CONFIG_CONTENT: JSON.stringify(options.config ?? {}) }
```

With a large catalog that one environment value is over 340 KB, which exceeds the Linux per string limit (`MAX_ARG_STRLEN`, 128 KB). `spawn` throws `E2BIG` before opencode runs.

**Captured stack (from the bridge event log)**

```
spawn E2BIG
    at ChildProcess.spawn (node:internal/child_process:420:11)
    at Object.spawn (node:child_process:787:9)
    at spawn (.../cross-spawn@7.0.6/index.js:12:24)
    at createOpencodeServer (.../harness/opencode/...)
    at runTurn (.../harness/opencode/bridge.mjs)
```

### Two further issues observed

1. **Headless permissions are auto rejected.** `buildOpenCodeConfig` hardcodes `permission` for `bash`, `edit`, `external_directory`, `webfetch`, `doom_loop`, and `task` to `"ask"`. The bridge runs with no interactive approver, so these tool calls are auto rejected (the model receives "External directory access rejected") even though the host started the harness with `permissionMode: 'allow-all'`. The bridge should honor `permissionMode` and set permissions to `"allow"` in that mode.

2. **The fatal error is swallowed on the host.** As with the codex bridge, the stdout drain races the readiness reader ("Invalid state: The ReadableStream is locked"), so the real frame never reaches the host stream. The only reliable place it survives is the on disk event log (`<workdir>/.agent-runs/<sessionId>/bridge/event-log.ndjson`). Surfacing a fatal bridge error to the host stream would make this far easier to diagnose.

### Reproduction

1. Build a `HarnessAgent` on the opencode harness with a tool set whose combined serialized JSON schemas exceed roughly 128 KB.
2. Run any turn. It ends with zero steps and no content; the bridge event log contains the `spawn E2BIG` frame above.
3. Separately, with a smaller tool set that does launch, observe `bash`/`external_directory` tool calls being auto rejected despite `permissionMode: 'allow-all'`.

### Suggested fix

Do not put the large tool catalog in the single config environment value. Write `TOOL_SCHEMAS` to a file and pass only a short file path to the MCP server (for example `TOOL_SCHEMAS_FILE`), and have `host-tool-mcp.mjs` read that file. Keep the rest of the config on the inline channel so opencode still reads it. Also set the permission entries to `"allow"` when `permissionMode` is `'allow-all'`.

We applied these as a local patch and opencode now launches, exposes the host tools, and calls them with the full catalog.

### AI SDK Version

- @ai-sdk/harness-opencode: 1.0.0-beta.1
- @ai-sdk/harness: 1.0.0-beta.27
- @opencode-ai/sdk (bundled by the bridge): 1.17.6
- Runtime: Node 22 in a Vercel Sandbox (Linux)

vicksiyi:
Summary
- Write harness-opencode host tool schemas to a workspace file instead of embedding the full schema catalog in the local MCP process environment.
- Pass TOOL_SCHEMAS_PATH to the host tool MCP bridge and keep TOOL_SCHEMAS as a backward-compatible fallback.

Tests
- git diff --check
- pnpm --filter @ai-sdk/harness-opencode type-check (blocked: this checkout is missing packages/harness-opencode/node_modules/@vercel/ai-tsconfig/ts-library.json and workspace type resolution)
- pnpm --filter @ai-sdk/harness-opencode test -- opencode-harness.test.ts (blocked by the same missing tsconfig package before tests run)

Notes
- This PR addresses the E2BIG schema transport failure path from #16380. The separate permission behavior mentioned in the issue is left for a follow-up so this change stays scoped.

Fixes #16380

### DIFF
--- a/packages/harness-opencode/src/bridge/index.ts
+++ b/packages/harness-opencode/src/bridge/index.ts
@@ -4,7 +4,7 @@ import {
   type BridgeTurn,
 } from '@ai-sdk/harness/bridge';
 import { randomUUID } from 'node:crypto';
-import { mkdirSync } from 'node:fs';
+import { mkdirSync, writeFileSync } from 'node:fs';
 import path from 'node:path';
 import { argv, env as procEnv } from 'node:process';
 import type { StartMessage } from '../opencode-bridge-protocol';
@@ -40,6 +40,8 @@ type Emit = (msg: Record<string, unknown>) => void;
 
 type OpenCodeClient = ReturnType<typeof createOpencodeClient>;
 type OpenCodeServer = Awaited<ReturnType<typeof createOpencodeServer>>;
+
+const TOOL_SCHEMAS_FILE = '.ai-sdk-harness-tool-schemas.json';
 
 type RuntimeState = {
   server?: OpenCodeServer;
@@ -224,19 +226,24 @@ function buildOpenCodeConfig({
   const provider = buildProviderConfig(start);
   if (provider) config.provider = provider;
   if (relayPort && start.tools && start.tools.length > 0) {
+    const toolSchemasPath = path.join(workdir, TOOL_SCHEMAS_FILE);
+    writeFileSync(
+      toolSchemasPath,
+      JSON.stringify(
+        start.tools.map(t => ({
+          name: t.name,
+          description: t.description,
+          inputSchema: t.inputSchema,
+        })),
+      ),
+    );
     config.mcp = {
       'harness-tools': {
         type: 'local',
         enabled: true,
         command: ['node', `${bootstrapDir}/host-tool-mcp.mjs`],
         environment: {
-          TOOL_SCHEMAS: JSON.stringify(
-            start.tools.map(t => ({
-              name: t.name,
-              description: t.description,
-              inputSchema: t.inputSchema,
-            })),
-          ),
+          TOOL_SCHEMAS_PATH: toolSchemasPath,
           TOOL_RELAY_URL: `http://127.0.0.1:${relayPort}`,
         },
       },
--- a/packages/harness-opencode/src/bridge/host-tool-mcp.ts
+++ b/packages/harness-opencode/src/bridge/host-tool-mcp.ts
@@ -6,6 +6,7 @@
  */
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js';
+import { readFileSync } from 'node:fs';
 import { z } from 'zod/v4';
 
 type ToolSchema = {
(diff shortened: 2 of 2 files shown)


rpelevin:
I would keep this scoped to the transport boundary, but make the contract executable.

Moving the host tool schema catalog out of the child-process environment is the right fix for the E2BIG class. The regression I would want is not only a large-catalog happy path, but also the exact failure modes around where that catalog now lives.

A useful test shape:

1. create enough host tool schemas to exceed the single environment-value limit;
2. assert the opencode process starts with only a short schema-path env value;
3. assert the local MCP bridge loads the file and exposes the same tool names and input schemas;
4. assert a missing, unreadable, or malformed schema file produces a fatal bridge error that reaches the host stream instead of an empty turn;
5. assert the legacy inline env fallback still works for small catalogs until it is intentionally removed;
6. assert the schema file is written under the run workspace and is not reused across sessions.

That keeps the file handoff from becoming a second hidden authority channel. The model-visible tool catalog, the MCP bridge catalog, and the host relay target should all be tied to the same turn/session record.

I would leave the separate headless permission-mode behavior as a follow-up, as the PR already does, but the fatal-error surfacing test belongs with this transport fix because otherwise E2BIG-style failures can still look like successful empty turns.

Boundary: architecture and regression-test feedback only; no claim about using this project, running this branch, validating implementation behavior, implementation correctness, merge readiness, security review, production readiness, partnership, customer interest, official alignment, Vercel usage, AI SDK usage, conformance certification, or Neura usage.

vicksiyi:
Rebased this PR on the latest `main` and resolved the harness-opencode bridge conflict by keeping the current authorized relay flow while moving tool schemas to `TOOL_SCHEMAS_PATH`.

Pushed commit `6e360c44f1`. Verified with:
- `corepack pnpm --filter @ai-sdk/harness-opencode type-check`
- `corepack pnpm --filter @ai-sdk/harness-opencode test:node` (9 files, 37 tests)

packages/harness-opencode/src/bridge/index.ts
```diff
@@ -224,19 +226,24 @@ function buildOpenCodeConfig({
   const provider = buildProviderConfig(start);
   if (provider) config.provider = provider;
   if (relayPort && start.tools && start.tools.length > 0) {
+    const toolSchemasPath = path.join(workdir, TOOL_SCHEMAS_FILE);
```
felixarntz, line 229: same feedback as in https://github.com/vercel/ai/pull/16426#discussion_r3617458460: we should not write harness specific utilities to the primary `workdir`. `workdir` should be reserved exclusively for the agent to work in, it should only contain whatever the sandbox configuration defines to be in the `workdir` (via callbacks like `onBootstrap`).

Worth checking for other harness-specific directories that could be used, or maybe even write it into `cliShimDir` if feasible.

felixarntz (changes requested):
@vicksiyi Thank you for the PR! Mostly looks good, just one thing that needs to be fixed.

PR #16433, opened 2026-06-26
fix(workflow): repair UI message stream part framing in WorkflowChatTransport

VaguelySerious:
## Summary

Port of **vercel/workflow#2537** to `@ai-sdk/workflow`.

A workflow run owns a single shared UI message stream, and the AI SDK's stream consumer (`processUIMessageStream`, backing `useChat`/`readUIMessageStream`) resets its active text/reasoning part maps on every `finish-step`. Multi-step turns reuse the same part id (commonly `"0"`) per step, so if a `text-start` is dropped or a `finish-step` lands mid-part, the consumer hits a `text-delta`/`text-end` for an id it has no open part for and throws:

```
Received text-delta for missing text part with ID "0". Ensure a "text-start" chunk is sent before any "text-delta" chunks.
```

…killing the whole turn. That malformed framing is reachable whenever a stream-producing step's output is duplicated or interleaved on the shared stream (step retry/redelivery, reconnect/replay overlap, concurrent-worker duplication).

## Fix

`WorkflowChatTransport` now passes the stream it hands to the AI SDK through `normalizeUIMessageStreamParts`, which mirrors the consumer's per-step part-lifetime state machine:

- resets tracking on `finish-step` (exactly where the consumer resets — verified against `processUIMessageStream`);
- synthesizes a missing `*-start` when an orphaned `*-delta`/`*-end` arrives;
- drops a re-delivered `*-start`/`*-delta`/`*-end` for a part already open/ended in the current step.

It wraps the *output* of the send/reconnect iterators, so the reconnect `chunkIndex`/`gotFinish` accounting (which counts raw server chunks) is untouched. A well-formed stream passes through unchanged. Applies to both text and reasoning parts.

## Tests

`workflow-chat-transport.stream-repair.test.ts` drives the **real** transport (mock `fetch` returning chunks via the AI SDK's own `JsonToSseTransformStream`) into the **real** `readUIMessageStream` consumer:

- proves the raw interleaved stream reproduces the exact `missing text part` error;
- proves the transport repairs it with no content lost (`"Hello world"`);
- proves a reconnect/replay duplicate tail is dropped without erroring;
- proves a well-formed multi-step stream (id `"0"` reused across steps) is unchanged;
- same for reasoning parts (raw interleaved reasoning is fatal; transport repairs to `"Thinking..."`).

`pnpm --filter @ai-sdk/workflow build && test` pass (node + edge).

## Related

- Sibling PR: vercel/workflow#2537 (the `@workflow/ai` copy)
- Original issue: vercel/workflow#2422

Opened as a draft — this is an AI-assisted port across diverged copies of the same package; please review the translation before merge.

### DIFF
--- a/.changeset/repair-workflow-ui-message-stream-framing.md
+++ b/.changeset/repair-workflow-ui-message-stream-framing.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/workflow": patch
+---
+
+`WorkflowChatTransport` now repairs UI message stream part framing, so duplicated or interleaved durable stream writes no longer crash the AI SDK consumer with `Received text-delta for missing text part`.
--- a/packages/workflow/src/index.ts
+++ b/packages/workflow/src/index.ts
@@ -45,3 +45,5 @@ export {
   type SendMessagesOptions,
   type ReconnectToStreamOptions,
 } from './workflow-chat-transport.js';
+
+export { normalizeUIMessageStreamParts } from './normalize-ui-message-stream.js';
--- a/packages/workflow/src/normalize-ui-message-stream.ts
+++ b/packages/workflow/src/normalize-ui-message-stream.ts
@@ -1,0 +1,164 @@
+import type { UIMessageChunk } from 'ai';
+
+/**
+ * Tracks, for one part family (text or reasoning), which part ids are open or
+ * ended within the current step.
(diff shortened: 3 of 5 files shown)


gr2m (approved):
Thanks!

PR #16434, opened 2026-06-26
perf(workflow): reduce doStreamStep step-boundary payload

VaguelySerious:
## Summary

Port of **vercel/workflow#1932** to `@ai-sdk/workflow`.

`doStreamStep` built the full `StepResult` *inside* the step and also returned a per-chunk snapshot array. Both were serialized into the durable event log on every step. But:

- the iterator never read the chunk array (pure dead payload), and
- `StepResult` carries redundant copies of data: `content` (= `text` + tool calls), the duplicate `toolCalls`/`dynamicToolCalls` lists, `reasoningText` (= joined `reasoning`), and the always-empty `*ToolResults` arrays.

## Change

The step now returns only the minimal raw aggregates needed to rebuild the `StepResult` — `text`, `reasoning`, `responseMetadata`, `warnings` — alongside the existing `finish` and `toolCalls`. The iterator reconstructs the full `StepResult` via a new `buildStepResult` helper **outside** the step boundary.

- The user-facing `StepResult` (yielded steps, `onStepEnd`, telemetry) is byte-for-byte the same shape; only the durable event-log payload shrinks.
- `runtimeContext` / `toolsContext` / `stepNumber` no longer cross the boundary either — the iterator already holds them and applies them during reconstruction.

## Tests

- `stream-text-iterator.test.ts` mocks updated to the new `{ toolCalls, finish, raw, providerExecutedToolResults }` shape; the contexts test now asserts the contexts are applied to the reconstructed step and yield value (rather than passed across the boundary).
- The unmocked end-to-end `workflow-agent` tests exercise the real `doStreamStep` → `buildStepResult` path and pass unchanged, confirming the reconstruction is faithful.
- `pnpm --filter @ai-sdk/workflow build && test` pass (node + edge).

## Related

- Sibling PR: vercel/workflow#1932 (the `@workflow/ai` copy)
- Original issue: vercel/workflow#1929

Opened as a draft — this is an AI-assisted port. The two copies of this package have diverged (vercel/ai's `doStreamStep` is `streamModelCall`-based and built the `StepResult` inline, with no separate `chunksToStep`), so this applies the same optimization rather than cherry-picking the diff; please review the adaptation before merge.

### DIFF
--- a/.changeset/reduce-workflow-do-stream-step-payload.md
+++ b/.changeset/reduce-workflow-do-stream-step-payload.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/workflow": patch
+---
+
+Reduce the `doStreamStep` step-boundary payload by returning minimal raw aggregates and reconstructing the `StepResult` outside the step, instead of serializing the full `StepResult` plus the per-chunk array into the durable event log.
--- a/packages/workflow/src/do-stream-step.ts
+++ b/packages/workflow/src/do-stream-step.ts
@@ -2,7 +2,6 @@ import type {
   LanguageModelV4CallOptions,
   LanguageModelV4Prompt,
 } from '@ai-sdk/provider';
-import type { Context } from '@ai-sdk/provider-utils';
 import {
   experimental_streamLanguageModelCall as streamModelCall,
   gateway,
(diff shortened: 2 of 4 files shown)


gr2m (approved):
Reviewed: the payload reduction is a faithful field-for-field move of the StepResult construction out of the step boundary, the dropped chunks array was dead payload, and doStreamStep is internal-only. Replay compat across the shape change is acceptable while the package is experimental.

PR #16463, opened 2026-06-29
fix(ai): prune matching tool approval responses

### LINKED ISSUE #16385: `pruneMessages` leaves orphaned `tool-approval-response` parts when pruning a specific tool by name

miachillgood:
## Summary
- build the approval-to-tool lookup across the full message history before pruning
- prune `tool-approval-response` parts together with the matching tool when selective tool pruning is used
- add a regression test for pruning `get-weather-tool-2` so the approval response does not survive as an orphan

## Why
`vercel/ai#16385` reports that selective tool pruning can leave behind `tool-approval-response` parts because the current implementation only resolves `approvalId -> toolName` inside each individual message.

Since approval requests live in an earlier assistant message while approval responses live in a later tool message, that per-message lookup misses the association and incorrectly keeps the response.

## Testing
- `git diff --check`
- could not run package tests in this checkout because local dependencies are not installed (`vitest: command not found`; `node_modules missing`)

Closes #16385

### DIFF
--- a/.changeset/fix-prune-tool-approval-response.md
+++ b/.changeset/fix-prune-tool-approval-response.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Fix selective tool pruning so matching tool approval responses are removed too.
--- a/packages/ai/src/generate-text/prune-messages.test.ts
+++ b/packages/ai/src/generate-text/prune-messages.test.ts
@@ -716,5 +716,70 @@ describe('pruneMessages', () => {
         `);
       });
     });
+
+    describe('specific tool pruning', () => {
+      it('should also prune matching tool-approval-response parts', () => {
+        const result = pruneMessages({
+          messages: messagesFixture1,
+          toolCalls: [{ type: 'all', tools: ['get-weather-tool-2'] }],
+        });
+
+        expect(result).toMatchInlineSnapshot(`
+          [
+            {
+              "content": [
+                {
+                  "text": "Weather in Tokyo and Busan?",
+                  "type": "text",
+                },
+              ],
+              "role": "user",
+            },
+            {
+              "content": [
+                {
+                  "text": "I need to get the weather in Tokyo and Busan.",
+                  "type": "reasoning",
+                },
+                {
+                  "input": "{"city": "Tokyo"}",
+                  "toolCallId": "call-1",
+                  "toolName": "get-weather-tool-1",
+                  "type": "tool-call",
+                },
+              ],
+              "role": "assistant",
+            },
+            {
+              "content": [
+                {
+                  "output": {
+                    "type": "text",
+                    "value": "sunny",
+                  },
+                  "toolCallId": "call-1",
+                  "toolName": "get-weather-tool-1",
+                  "type": "tool-result",
+                },
+              ],
+              "role": "tool",
+            },
+            {
+              "content": [
+                {
+                  "text": "I have got the weather in Tokyo and Busan.",
+                  "type": "reasoning",
+                },
+                {
+                  "text": "The weather in Tokyo is sunny. I could not get the weather in Busan.",
+                  "type": "text",
+                },
+              ],
+              "role": "assistant",
+            },
+          ]
+        `);
+      });
+    });
   });
 });
--- a/packages/ai/src/generate-text/prune-messages.ts
+++ b/packages/ai/src/generate-text/prune-messages.ts
@@ -159,13 +159,18 @@ export function pruneMessages({
           }
 
           // keep parts that are not associated with a tool that should be removed:
+          if (toolCall.tools == null) {
+            return false;
+          }
+
+          const associatedToolName =
+            part.type === 'tool-call' || part.type === 'tool-result'
+              ? part.toolName
+              : approvalIdToToolName.get(part.approvalId);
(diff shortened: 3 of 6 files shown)


lgrammel:
Closing this PR because issue #16385 has already been closed.

lgrammel:
Closing this PR because issue #16385 has already been closed.

PR #16465, opened 2026-06-29
feat (provider/amazon-bedrock): add structuredOutputMode to opt out of native structured outputs

pixelcatgg:
## Summary

Adds a `structuredOutputMode` provider option to `@ai-sdk/amazon-bedrock`, mirroring the existing option on `@ai-sdk/anthropic`. When set to `'jsonTool'`, object generation uses the tool-based JSON path instead of injecting `output_config.format`.

## Motivation

For Anthropic models, the Bedrock provider enables native structured output (`output_config.format`) whenever `getModelCapabilities(modelId).supportsStructuredOutput` is `true`, with no way to opt out. That field is rejected on Bedrock partitions/regions that don't support it — notably **GovCloud (`us-gov-*`)** — with:

```
ValidationException: output_config.format: Extra inputs are not permitted
```

Because the capability is keyed on the model id and not the partition, the *same* model (e.g. `claude-sonnet-4-5`) succeeds on commercial Bedrock but fails on GovCloud, and there is currently no escape hatch. This is the partition-level counterpart of #14773 (which was addressed per-model for `claude-opus-4-7`).

## Changes

- New `structuredOutputMode: 'auto' | 'jsonTool' | 'outputFormat'` option (default `'auto'`), matching `@ai-sdk/anthropic`:
  - `'auto'` (default) — unchanged behavior: native `output_config.format` when the model supports it, otherwise the tool-based path.
  - `'jsonTool'` — always use the tool-based JSON path (`jsonResponseTool`, which the provider already builds when native output is disabled).
  - `'outputFormat'` — always use native `output_config.format`.
- Gates `useNativeStructuredOutput` on the option. Default behavior is preserved.
- Docs section, changeset, and tests (`jsonTool` falls back to the tool path; `outputFormat` forces native).

## Usage

```ts
const result = await generateObject({
  model: amazonBedrock('us-gov.anthropic.claude-sonnet-4-5-20250929-v1:0'),
  schema: z.object({ summary: z.string() }),
  prompt: '…',
  providerOptions: { bedrock: { structuredOutputMode: 'jsonTool' } },
});
```

## Verification

`pnpm --filter @ai-sdk/amazon-bedrock test` (node + edge), `type-check`, and `ultracite check` all pass.

---

Co-authored with @fdmmarshall.


gr2m (approved):
This is great, thanks a lot @pixelcatgg and @fdmmarshall!

pixelcatgg:
> This is great, thanks a lot @pixelcatgg and @fdmmarshall!

Thanks so much @gr2m pls let us know if there's anything else to do from our end to get it merged 🙏 @fdmmarshall and I are really proud to be able to contribute to the SDK

pixelcatgg:
hi @gr2m do you know when this will make it in? thanks!

Issue #14859, opened 2026-04-30
Resolve remaining code consistency issues


felixarntz:
#14648 ironed out the first large batch of code inconsistencies. However, two conventions were eventually left partially unresolved because fixing them would be more involved. This issue tracks addressing them and removing the temporary exclusions from the `konsistent.json` file.

### 1. `README.md` files are required per package

Some packages don't currently have `README.md` files, which should be required. These are currently manually excluded and need to be removed from the exclusion list in `conventions[0].must[0].excludeFiles`. It currently has:
```json
          "excludeFiles": ["packages/mcp", "packages/otel", "packages/voyage"]
```

* [ ] add `README.md` for `mcp` package
* [ ] add `README.md` for `otel` package
* [ ] add `README.md` for `voyage` package

### 2. Every model class must have a corresponding model options file and type

Some models in provider packages don't currently have a type definition for their `providerOptions`. These are currently manually excluded and need to be removed from the exclusion list, either in `conventions[1].must[6].excludeFiles` or `conventions[1].must[8].excludeFiles`. They currently have:
```json
          "excludeFiles": [
            "...irrelevant...",
            "packages/amazon-bedrock/src/amazon-bedrock-image-model.ts",
            "packages/deepinfra/src/deepinfra-image-model.ts",
            "packages/fireworks/src/fireworks-image-model.ts",
            "packages/gateway/src/gateway-embedding-model.ts",
            "packages/gateway/src/gateway-image-model.ts",
            "packages/gateway/src/gateway-language-model.ts",
            "packages/gateway/src/gateway-reranking-model.ts",
            "packages/mistral/src/mistral-embedding-model.ts",
            "packages/openai-compatible/src/image/openai-compatible-image-model.ts",
            "packages/openai/src/image/openai-image-model.ts",
            "packages/perplexity/src/perplexity-language-model.ts"
          ]
```

```json
          "excludeFiles": [
            "packages/bytedance/src/bytedance-video-model.ts",
            "packages/gateway/src/gateway-video-model.ts"
          ]
```

Below are the remaining tasks. In most cases, they involve defining the type. There may be an occasional instance where defining such a type is not feasible or irrelevant. Specifically for the AI Gateway, it probably makes sense to look at the model options types holistically since there aren't really specific AI Gateway model parameters for the individual model kinds. The decision may even be to leave them in the `excludeFiles` list.

* [ ] define `providerOptions` type for `amazon-bedrock-image-model.ts`
[...]

PR #14648, opened 2026-04-20
chore: make provider implementations code patterns more consistent (via `konsistent`)

felixarntz:
## Background

As part of the effort of making our codebase more consistent, which helps for both human and agentic understanding of the codebase by recognizing patterns, I have been working on a general CLI tool meant to help with this task.

The CLI tool, called `konsistent` (German for "consistent"), is a generally applicable tool that anyone can use on their codebase, but the AI SDK was my main inspiration for creating it, and it presents a perfect use case. You define the rules that you want to enforce in a `konsistent.json` file, and then the CLI can check your codebase against those rules.

For more information, please see https://www.npmjs.com/package/konsistent/v/0.0.1-alpha.13 (very early release, not yet ready for broader consumption).

## Summary

I created a `konsistent.json` config file with code patterns I observed in the majority of our provider implementations. Of course we only adhere to any convention to some extent, but that's exactly what the `konsistent` CLI should help us with. Here's the AI SDK's tentative `konsistent.json` file: https://github.com/vercel/ai/blob/fa/konsistent/konsistent.json

I already did some very minor cleanup, but only on things that don't cause _any_ harm / possible breakage whatsoever.

The config file I created is a starting point, and it points out several violations in our codebase already. We should review those cases and decide:
- for each conflicting convention, which one do we want to follow?
- for any other more specific violations, should we fix them, or do we consider them an exception, or is enforcing that specific pattern simply too strict?

Once we have made those decisions, we can fix the codebase as needed. Since we are in v7 beta stage, this is the perfect time to e.g. rename some exported symbols for consistency. We can keep back compat via deprecated aliases.

As a starting point for our review, here's the current violations that running `pnpm konsistent` against our codebase (with the `konsistent.json` config file from this PR) produces:

```
packages/codemod
  -  error  Missing required file: src/index.ts  [package-dir-must-have-required-files]

packages/mcp
  -  error  Missing required file: README.md  [package-dir-must-have-required-files]

packages/otel
  -  error  Missing required file: README.md  [package-dir-must-have-required-files]

packages/svelte
  -  error  Missing required file: tsup.config.ts  [package-dir-must-have-required-files]

packages/voyage
  -  error  Missing required file: README.md  [package-dir-must-have-required-files]

packages/alibaba/src/alibaba-chat-language-model.ts
  -  error  Missing export type "AlibabaLanguageModelChatOptions"  [provider-model-file-with-subtype-must-export-model-class-and-options-type]
[...]

PR #16501, opened 2026-06-30
fix(amazon-bedrock): send correct bedrockRerankingConfiguration key for reranking

saxon-vercel:
## Background

`rerank()` with an Amazon Bedrock reranking model fails with HTTP 400. AWS responds with:

> 1 validation error detected: Value null at 'rerankingConfiguration.bedrockRerankingConfiguration' failed to satisfy constraint: Member must not be null

The Bedrock Agent Runtime Rerank API requires the member `rerankingConfiguration.bedrockRerankingConfiguration`, but the provider was sending it as `amazonBedrockRerankingConfiguration`. AWS therefore sees the required member as `null` and rejects every reranking request.

The rename was introduced by #14648 (a `konsistent` consistency pass) that treated this AWS-defined wire key as if it were an internal identifier. Reranking has been broken since `@ai-sdk/amazon-bedrock@5.0.0-beta.46` and remains broken through the latest release. The original implementation (#3764) used the correct key.

## Summary

- Restore the request-body key to `bedrockRerankingConfiguration` in `amazon-bedrock-reranking-model.ts` and its type mirror `amazon-bedrock-reranking-api.ts`.
- Update the two request-body inline snapshots in `amazon-bedrock-reranking-model.test.ts` to match.
- Add a focused test asserting the request carries the AWS-required `rerankingConfiguration.bedrockRerankingConfiguration` member, using an explicit `toHaveProperty` assertion rather than the full-body snapshot so a future rename fails loudly instead of being silently absorbed by `--update`.
- Internal symbol names (`AmazonBedrockRerankingInput`, `AmazonBedrockRerankingModel`, etc.) are intentionally left unchanged — only the on-the-wire key reverts.
- No codemod needed — this restores the original wire key and is neither a breaking change nor a deprecation.

## Manual Verification

- Reproduced the failure: calling `rerank()` against a Bedrock reranking model (e.g. `cohere.rerank-v3-5:0`) returns HTTP 400 with the AWS validation error above, naming `bedrockRerankingConfiguration` as the missing required member.
- Confirmed against the AWS Bedrock Agent Runtime API reference that `bedrockRerankingConfiguration` is the **Required** member of `rerankingConfiguration`: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_RerankingConfiguration.html
- Confirmed the corrected request body matches the shape sent by pre-regression releases (≤ `5.0.0-beta.45`), which succeed with identical model/query/documents.
- Note: exercising the corrected provider directly against live AWS Bedrock requires AWS credentials; the serialized request body is otherwise pinned by the new unit test, which now matches the documented AWS contract.

## Checklist

## Suggestions

- Harden usage of `konsistent` so renames are scoped to compiler-tracked symbols (types, classes, exported identifiers, filenames) and never rewrite object-literal keys or string literals, which may be serialized wire contracts. Tracked in #14859.
- Consider applying the same "pin the wire key with an explicit assertion" pattern to other providers' request bodies, and adding credential-gated e2e coverage/

## Related Issues

- Introduced by #14648.
- Related: #14859 (`konsistent` consistency follow-up).
- No existing issue tracks this specific 400, so there is no `Fixes #` to reference.

### DIFF
--- a/.changeset/fix-bedrock-reranking-wire-key.md
+++ b/.changeset/fix-bedrock-reranking-wire-key.md
@@ -3,5 +3,3 @@
 ---
 
 fix (amazon-bedrock): send the correct `bedrockRerankingConfiguration` request key for reranking
-
-The Bedrock Agent Runtime `Rerank` API requires the request member to be named `bedrockRerankingConfiguration`. A consistency rename had changed it to `amazonBedrockRerankingConfiguration`, causing AWS to reject every reranking call with `Value null at 'rerankingConfiguration.bedrockRerankingConfiguration' failed to satisfy constraint: Member must not be null` (HTTP 400). The wire key is now restored to match the AWS contract.
--- a/.changeset/fix-bedrock-reranking-wire-key.md
+++ b/.changeset/fix-bedrock-reranking-wire-key.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix (amazon-bedrock): send the correct `bedrockRerankingConfiguration` request key for reranking
+
+The Bedrock Agent Runtime `Rerank` API requires the request member to be named `bedrockRerankingConfiguration`. A consistency rename had changed it to `amazonBedrockRerankingConfiguration`, causing AWS to reject every reranking call with `Value null at 'rerankingConfiguration.bedrockRerankingConfiguration' failed to satisfy constraint: Member must not be null` (HTTP 400). The wire key is now restored to match the AWS contract.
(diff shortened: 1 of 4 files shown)


felixarntz (approved):
@saxon-vercel great catch! this should have never been renamed as part of renaming the `bedrock` identifiers to `amazonBedrock`

felixarntz:
note: this doesn't require a backport since it was only renamed like that in v7.

.changeset/fix-bedrock-reranking-wire-key.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix (amazon-bedrock): send the correct `bedrockRerankingConfiguration` request key for reranking
+
+The Bedrock Agent Runtime `Rerank` API requires the request member to be named `bedrockRerankingConfiguration`. A consistency rename had changed it to `amazonBedrockRerankingConfiguration`, causing AWS to reject every reranking call with `Value null at 'rerankingConfiguration.bedrockRerankingConfiguration' failed to satisfy constraint: Member must not be null` (HTTP 400). The wire key is now restored to match the AWS contract.
```
felixarntz:
```suggestion
fix (amazon-bedrock): send the correct `bedrockRerankingConfiguration` request key for reranking
```

PR #16516, opened 2026-06-30
delete .cursor/skills symlink to skills/

n1ckoates:
cursor now supports the `.agents/skills` folder natively, and the symlink is redundant


gr2m (approved):
yep, thanks!

PR #16526, opened 2026-06-30
feat(provider/google): support Gemini Interactions video output, usage, and omni dispatch

joshualipman123:
## Background

Google's `gemini-omni-flash-preview` is a multimodal-output model that returns generated **video** as part of a normal language-model response. Unlike Veo (which uses the `:predictLongRunning` video surface), Omni is served only by the Gemini **Interactions API** and bills per-modality on output tokens. To support it through the AI SDK (and soon gateway), the `@ai-sdk/google` Interactions integration needs to (1) parse video output blocks, (2) surface the per-modality output-token breakdown, and (3) route Omni model ids to Interactions automatically so callers (and the gateway) don't have to call `google.interactions(...)` explicitly.

## Summary

Three additive, backward-compatible changes to `@ai-sdk/google`:

- **Parse video output** — video blocks from the Interactions API are emitted as `file` parts (default `mediaType: video/mp4`), supporting both inline base64 `data` and `uri` delivery, in both the buffered (`doGenerate`) and streaming (`doStream`) paths. Mirrors the existing `image` handling
- **Per-modality output tokens** — adds `getGoogleInteractionsOutputTokensByModality()` and surfaces the breakdown via `providerMetadata.google.outputTokensByModality` (e.g. `{ video: 57920, text: 19 }`)
- **Auto-dispatch Omni ids** — `google(...)` and `google.languageModel(...)` route ids matching `gemini-omni*` through the Interactions API; `google.chat`/`google.generativeAI` stay on the explicit `generateContent` path. Mirrors the OpenAI provider's `createLanguageModel` dispatch pattern.

## Manual Verification

Ran a live streaming request against `gemini-omni-flash-preview` (`responseModalities: ['video']`, `store: false`) with `includeRawChunks: true` to confirm wire behavior end-to-end:

- Video arrives as a single `step.delta` of `type: "video"` carrying the full base64 payload; the `model_output` `step.start` is bare and `step.stop` is empty. The transform emits the `file` part (`video/mp4`) correctly.
- Verified for both a small clip and a ~2.2 MB clip — same delta-based delivery in both cases.
- Confirmed `google('gemini-omni-flash-preview')` dispatches to the Interactions model (after building the package; examples run against `dist`).
- Both new examples run successfully:
  - `examples/ai-functions/src/generate-text/google/interactions-video-output.ts`
  - `examples/ai-functions/src/stream-text/google/interactions-video-output.ts`

## Checklist

### DIFF
--- a/content/providers/01-ai-sdk-providers/15-google.mdx
+++ b/content/providers/01-ai-sdk-providers/15-google.mdx
@@ -1163,13 +1163,6 @@ any other AI SDK language model.
   request bodies and SSE event vocabularies.
 </Note>
 
-<Note>
-  Omni-family models (e.g. `gemini-omni-flash-preview`, which returns video as a
-  file part) are only served by the Interactions API, so `google(...)` and
-  `google.languageModel(...)` route them through `interactions()` automatically
-  — you do not need to call `google.interactions(...)` explicitly for them.
-</Note>
-
 ### Provider Options
 
 The Interactions model reads its options from the shared
--- a/content/providers/01-ai-sdk-providers/15-google.mdx
+++ b/content/providers/01-ai-sdk-providers/15-google.mdx
@@ -1163,6 +1163,13 @@ any other AI SDK language model.
   request bodies and SSE event vocabularies.
 </Note>
 
+<Note>
+  Omni-family models (e.g. `gemini-omni-flash-preview`, which returns video as a
+  file part) are only served by the Interactions API, so `google(...)` and
+  `google.languageModel(...)` route them through `interactions()` automatically
+  — you do not need to call `google.interactions(...)` explicitly for them.
+</Note>
+
 ### Provider Options
 
 The Interactions model reads its options from the shared
--- a/examples/ai-functions/src/stream-text/google/interactions-video-output.ts
+++ b/examples/ai-functions/src/stream-text/google/interactions-video-output.ts
@@ -12,9 +12,7 @@ run(async () => {
 
   await withSpinner('Generating video...', async () => {
     const result = streamText({
-      // Omni-family ids route through the Interactions API automatically, so
-      // `google(...)` / `languageModel(...)` work without `google.interactions(...)`.
-      model: google('gemini-omni-flash-preview'),
+      model: google.interactions('gemini-omni-flash-preview'),
       prompt:
         'A marble rolling fast on a chain reaction style track, continuous smooth shot.',
       providerOptions: {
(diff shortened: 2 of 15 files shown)


packages/google/src/google-provider.ts
```diff
@@ -293,18 +304,23 @@ export function createGoogle(
       },
     );
 
+  const createLanguageModel = (modelId: GoogleModelId): LanguageModelV4 =>
+    isInteractionsOnlyModelId(modelId)
+      ? createInteractionsModel(modelId)
+      : createChatModel(modelId);
```
felixarntz: I'm unsure about this. Did you find any prior art in AI SDK that does this automatic mapping to another endpoint based on model ID? This feels a bit "too smart" to me, it could be unexpected.

content/providers/01-ai-sdk-providers/15-google.mdx
```diff
@@ -1163,6 +1163,13 @@ any other AI SDK language model.
   request bodies and SSE event vocabularies.
 </Note>
 
+<Note>
+  Omni-family models (e.g. `gemini-omni-flash-preview`, which returns video as a
+  file part) are only served by the Interactions API, so `google(...)` and
+  `google.languageModel(...)` route them through `interactions()` automatically
+  — you do not need to call `google.interactions(...)` explicitly for them.
+</Note>
+
```
felixarntz: see other comment, unsure about this.

examples/ai-functions/src/stream-text/google/interactions-video-output.ts
```diff
@@ -0,0 +1,55 @@
...
+run(async () => {
+  const videos: Array<GeneratedFile> = [];
+
+  await withSpinner('Generating video...', async () => {
+    const result = streamText({
+      // Omni-family ids route through the Interactions API automatically, so
+      // `google(...)` / `languageModel(...)` work without `google.interactions(...)`.
+      model: google('gemini-omni-flash-preview'),
+      prompt:
+        'A marble rolling fast on a chain reaction style track, continuous smooth shot.',
```
felixarntz, line 17: these two examples are great. for even better coverage, consider a multi-turn example where you first generate a video, then prompt to modify that video in a specific way.

not a must-have, but would be neat to cover. maybe as a follow up?

packages/google/src/google-provider.ts
joshualipman123: Added this so that gateway would remain `.interactions()` ignorant and only ever need to call `provider.languageModel()`. i think you're right though best to do this branching on the gateway side. i will revert

felixarntz (approved):

PR #14629, opened 2026-04-20
fix(ai): store raw invalid tool input as structured error object

Jim1874:
Fixes invalid JSON input being stored as raw string instead of structured error object.

Closes #14442

Issue #13645, opened 2026-03-19
convertToModelMessages sends rawInput string as tool_use.input, causing Anthropic 400 on retry


jb-chief:
### Description

When a tool call has `state: "output-error"` with `input: undefined` and `rawInput` set to a string (because the model generated invalid JSON for the tool input), `convertToModelMessages` sends the raw string as the `tool_use.input` value to Anthropic.

Anthropic's API requires `tool_use.input` to be a JSON object/dictionary, so it rejects the request with:

```
400: messages.N.content.M.tool_use.input: Input should be a valid dictionary
```

This creates an **unrecoverable loop** in multi-turn conversations: the corrupted tool call is part of the message history, so every subsequent inference attempt replays the same invalid message and gets the same 400 error.

### Root Cause

In `convertToModelMessages`, when processing tool UI parts with `state === "output-error"`:

```js
input: part.state === "output-error"
  ? (_a = part.input) != null ? _a : "rawInput" in part ? part.rawInput : void 0
  : part.input,
```

When `part.input` is `undefined` (as set during streaming when JSON parsing fails), it falls back to `part.rawInput` — which is a **string**, not an object. Anthropic (and likely other providers) requires `input` to be a dict/object.

### Reproduction Steps

1. Use `streamText` with Anthropic and tools that have structured input schemas
2. The model generates a tool call with syntactically invalid JSON in the input (e.g., an unquoted string value like `"to": John Doe <john@example.com>`)
3. The AI SDK correctly catches the parse error and creates a part with `state: "output-error"`, `input: undefined`, `rawInput: "<the invalid JSON string>"`
4. On the next turn (or retry), `convertToModelMessages` converts this part back to a model message with `input` set to the raw string
5. Anthropic returns `400: messages.N.content.M.tool_use.input: Input should be a valid dictionary`
6. The conversation is now permanently stuck — every retry replays the same broken history

### Expected Behavior

`convertToModelMessages` should ensure `tool_use.input` is always a valid object when converting `output-error` parts. Options:
- Attempt to parse `rawInput` as JSON; if it fails, use `{}` or `{ "raw": rawInput }`  
- Always use `{}` as a fallback when `input` is undefined, regardless of `rawInput`

### Related

- #11154 — tracks the broader cleanup of error message state (v6/v7 TODO)
- #11719 — LLM generating invalid JSON in tool arguments

### Environment

- `ai` SDK version: 5.x (latest)
- Provider: `@ai-sdk/anthropic`
[...]

PR #16561, opened 2026-07-01
fix(bedrock): wrap invalid tool input in object

### LINKED ISSUE #14442: Invalid tool call input stored as raw string causes API rejection on next step

n1ckoates:
## Background

When a model produces an invalid tool call with malformed JSON, `parseTooCall()` recognizes it and stores it as a string in the tool input. On Bedrock, tool use input needs to be an object, which throws an API error.

This occurs when messages are persisted and replayed (e.g. using `useChat()`) on a second turn.

It doesn't occur with, for example, multiple steps in one `generateText()` or `streamText()` because `toResponseMessages()` guards against it by replacing an invalid tool with `{}`.

## Summary

Wrap invalid tool use input in an object so that it's always valid API input.

## Manual Verification

Added an example to `ai-e2e-next` at `/chat/bedrock-invalid-tool-call` that reproduces the issue. It forces the model to generate malformed JSON on the first turn, then replays that conversation.

* **Bedrock**:
`{"message":"The format of the value at messages.1.content.0.toolUse.input is invalid. Provide a json object for the field and try again."}`

After adding the fix, the issue goes away.

## Checklist

## Future Work

Anthropic provider has nearly the exact same bug (example reproduces it if you swap provider to anthropic), fix there too
https://github.com/vercel/ai/issues/13645

## Related Issues

Fixes https://github.com/vercel/ai/issues/14442

Closes https://github.com/vercel/ai/pull/16280
Closes https://github.com/vercel/ai/pull/14629
Closes https://github.com/vercel/ai/pull/15280
Closes https://github.com/vercel/ai/pull/15863
Closes https://github.com/vercel/ai/pull/15908
Closes https://github.com/vercel/ai/pull/16020
Closes https://github.com/vercel/ai/pull/16280
Closes https://github.com/vercel/ai/pull/16539


lgrammel:
Closing this PR because issue #1234 has already been closed.

packages/ai/src/generate-text/parse-tool-call.test.ts
```diff
@@ -471,7 +471,9 @@ describe('parseToolCall', () => {
           "dynamic": true,
           "error": [AI_InvalidToolInputError: Invalid input for tool testTool: AI_JSONParseError: JSON parsing failed: Text: invalid json.
         Error message: SyntaxError: Unexpected token 'i', "invalid json" is not valid JSON],
-          "input": "invalid json",
+          "input": {
+            "rawInvalidInput": "invalid json",
+          },
```
lgrammel: this feels strange that we need this. is there a way to fix it without changing the ai pacakge?

lgrammel: looking at the PR description, it seems limited to certain providers, so the providers could handle this case instead.

n1ckoates: i agree, it's a provider-specific bug so will switch to fix in provider packages instead.

lgrammel (approved):

PR #16572, opened 2026-07-02
fix (ai): serialize parallel tool results in tool-call order

### LINKED ISSUE #16567: Parallel tool results are serialized in completion order, silently breaking provider prompt caching
### Description
When an assistant step makes parallel tool calls, the tool-result parts are recorded in completion order (whichever `execute` resolves first), not call order. toResponseMessages (`packages/ai/src/generate-text/to-response-messages.ts`) then builds the tool message by iterating step content in that arrival order, with no canonical sort.

Completion order is nondeterministic across requests. The same logical history therefore serializes with tool_result blocks in different orders depending on how it was produced:

- Mid-run steps: results appear in completion order
- History rebuilt from persistence (e.g. `convertToModelMessages` over stored UIMessages): results appear in call order, because UIMessage tool parts are created at call time

Providers pair `tool_use/tool_result by id,` so responses are unaffected - but prompt caching is byte-sensitive. 

The first parallel batch whose order differs between requests invalidates the entire cached prefix behind it. In a long agent loop this shows up as the cache being re-written almost every turn; we measured ~47k tokens of redundant cache writes per request on Bedrock/Anthropic before pinning it to this.

### Repro:
1. `streamText` with two tools, both called in parallel in one step; make tool A slower than tool B.
2. Capture the request body of the following step: `tool_result` blocks appear B, A (completion order).
3. Rebuild the same conversation from persisted UIMessages via c`onvertToModelMessages` and send: blocks appear A, B (call order).
4. Byte-compare - identical content, different block order → provider cache miss from that point.

### Proposed fix

In `toResponseMessages`, emit the tool message's `tool-result/tool-error` parts sorted by the order of the corresponding `tool-call` parts in the assistant content (stable sort, unknown ids last). Key order within a step is not semantically meaningful, so this is behavior-preserving - it just makes serialization deterministic and cache-friendly.

### Workaround we have implemented:
Re-sorting in prepareStep:

prepareStep: ({ messages }) => ({ messages: canonicaliseToolResultOrder(messages) })

where the helper sorts each tool message's results to match the preceding assistant's tool-call order. Works, but every caching-sensitive consumer has to rediscover this independently.

### AI SDK Version

`Version: ai@6.0.182`

### Code of Conduct

Kripu77:
## Background

Fixes #16567.

When an assistant step makes parallel tool calls, results are recorded in
completion order (whichever `execute` resolves first). `toResponseMessages`
serializes that arrival order verbatim, so the same logical step produces
differently-ordered `tool-result` parts depending on tool timing.

Providers pair tool calls and results by id, so responses are unaffected —
but provider prompt caching is byte-sensitive. A conversation served live
(completion order) and the same conversation rebuilt from persisted UIMessages
(call order, since tool parts are created at call time) serialize differently,
invalidating the entire cached prefix behind the first parallel batch. Measured
on a production agent loop: a 4-tool batch serialized in the exact reverse of
call order re-wrote 26 byte-identical cached blocks on the next request.

## Summary

Sort the tool message's `tool-result` parts to match the order of the
corresponding `tool-call` parts. Results without a matching call sort last,
tie-broken by id, so the ordering is total and deterministic. Sorting is
skipped when the tool message contains `tool-approval-response` parts, since
approval responses and their synthesized execution-denied results carry
position-dependent semantics.

Within-step result order is not semantically meaningful to providers, so this
is behavior-preserving — it only makes serialization deterministic and
cache-friendly.

## Manual Verification

Validated end-to-end against Bedrock Anthropic prompt caching: with this
change, a 36-request agent session with parallel batches up to size 8 produced
zero cache-prefix invalidations (previously invalidated on every fresh request
after a parallel batch).

## Checklist

## Related Issues

Fixes #16567


lgrammel:
Closing this PR because issue #16567 has already been closed.

PR #16577, opened 2026-07-02
fix: Parallel tool results are serialized in completion order, silently breaking provider prompt caching

### LINKED ISSUE #16567: Parallel tool results are serialized in completion order, silently breaking provider prompt caching

lgrammel:
## Background

Parallel tool executions could serialize tool results in completion order instead of tool-call order, producing nondeterministic prompt bytes and breaking provider prompt caching.

## Summary

Updated toResponseMessages to record tool-call order and stably sort emitted tool message parts by the corresponding call order, with unknown IDs kept last.

## Testing

Kept the focused regression test for parallel tool results and adjusted its assertion to type-narrow tool result parts.

## Related Issues

Fixes #16567

### DIFF
--- a/.changeset/parallel-tool-results-call-order.md
+++ b/.changeset/parallel-tool-results-call-order.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Serialize parallel tool results in tool-call order when converting generation output to response messages.
--- a/packages/ai/src/generate-text/to-response-messages.test.ts
+++ b/packages/ai/src/generate-text/to-response-messages.test.ts
@@ -290,6 +290,64 @@ describe('toResponseMessages', () => {
     `);
   });
 
+  it('should serialize parallel tool results in tool call order', async () => {
+    const result = await toResponseMessages({
+      content: [
+        {
+          type: 'text',
+          text: 'Using tools',
+        },
+        {
+          type: 'tool-call',
+          toolCallId: 'call-a',
+          toolName: 'toolA',
+          input: {},
+        },
+        {
+          type: 'tool-call',
+          toolCallId: 'call-b',
+          toolName: 'toolB',
+          input: {},
+        },
+        // Simulates parallel execution where toolB resolved before toolA.
+        {
+          type: 'tool-result',
+          toolCallId: 'call-b',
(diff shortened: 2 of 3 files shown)


lgrammel:
does not meet the bar

PR #16582, opened 2026-07-02
fix: duplicate text part IDs in multi-step tool-call streams

### LINKED ISSUE #15789: @ai-sdk/openai-compatible reuses txt-0 text part ID across TEXT -> TOOL -> TEXT streams

lgrammel:
## Background

Multi-step tool-calling streams could expose separate text segments with the same provider text ID, causing downstream UI adapters to merge text before and after tool calls and display parts out of order.

## Summary

streamText now tracks emitted text part IDs for the lifetime of an outer stream and remaps reused provider IDs to unique IDs while preserving matching start, delta, and end events.

## Testing

Added a regression test covering a two-step text -> tool -> text stream where the provider reuses txt-0 and verifying the second text segment is emitted as txt-1.

## Related Issues

Fixes #15789


lgrammel:
overly complex

PR #16587, opened 2026-07-02
fix: MCP callTool hangs and leaks handlers when aborted in flight

### LINKED ISSUE #16580: MCP `callTool` never settles (and leaks its response handler) when the request is aborted after being sent.
### Description

When an MCP request is aborted *after* the JSON-RPC request has been written to the transport but *before* the server responds — the normal case for a slow or hung MCP server, which is exactly when you'd abort — the returned promise never resolves or rejects. The awaiting `callTool` (and therefore the enclosing tool-execution step, agent loop, or HTTP request) hangs forever. The corresponding entry in the client's internal `responseHandlers` map is also never removed, so a long-lived client accumulates handlers and closures (a memory leak).

`DefaultMCPClient.request` (`packages/mcp/src/tool/mcp-client.ts`) checks the abort signal only:

1. synchronously, once, before sending (`signal?.throwIfAborted()`), and
2. inside the response handler (`if (signal?.aborted) reject(...)`), which only runs when a matching response arrives.

There is no `abort` event listener, so a request that is in flight when the signal fires can never be cancelled — the check at (2) is dead code because the handler is never invoked.

**Reproduction steps**

1. Implement an `MCPTransport` whose `send` replies to `initialize` but never replies to `tools/call`.
2. `const client = await createMCPClient({ transport });`
3. `const p = client.callTool({ name: 'x', arguments: {}, options: { signal } });`
4. Call `signal.abort()` shortly after.
5. Observe that `p` never settles, and the client's internal `responseHandlers` map still contains the handler.

**Expected behavior**

When the signal aborts, the pending `callTool` promise rejects with an abort error (`MCPClientError { message: 'Request was aborted', cause: signal.reason }`) and the response handler is removed from `responseHandlers`.

**Actual behavior**

The promise hangs indefinitely and the handler is leaked.

### AI SDK Version

- @ai-sdk/mcp: 2.0.5

### Code of Conduct

lgrammel:
## Background

Aborting an MCP tool call after the JSON-RPC request was sent left the returned promise pending forever and retained its response handler.

## Summary

DefaultMCPClient now registers an abort listener for in-flight requests, rejects with MCPClientError when the signal aborts, and removes the pending response handler and abort listener during cleanup.

## Testing

Added a regression test with a transport that never responds to tools/call, verifying that aborting rejects the call and clears the internal response handler.

## Related Issues

Fixes #16580


Abuhaithem:
Helo @lgrammel 
Is everything automated by AI now, leaving no room for community contributions to this project?

I had already submitted a fix for this issue in PR #16581 before your PR was opened. However, it appears that new PRs continue to be generated automatically by AI. This is a bit concerning, as it doesn't seem to consider or respect the time and effort that community contributors invest in improving the project.

lgrammel:
@Abuhaithem We are automating because it is the only way to compete going forward. Ideally you submit issues and we automatically fix them, saving you time and effort. We want to acknowledge contributions, e.g. mention the issue reports in the automatic PRs, but we are just setting up the automations.

lgrammel:
I will close this and also remove the reproduction and try out the new flow to see if this gives you credit @Abuhaithem

PR #16623, opened 2026-07-02
fix(react): call useChat onData for a shared Chat instance (#8597)

### LINKED ISSUE #8597: useChat onData is not triggered when used with a shared chat context context
### Discussed in https://github.com/vercel/ai/discussions/8558

<div type='discussions-op-text'>

<sup>Originally posted by **[@jonnyhd](https://github.com/jonnyhd)** September 10, 2025</sup>
When using a shared context useChat's onData callback is not triggered by transient data messages sent from the server. 

The onData for the Chat class (which is at the context level) does trigger. 

I've recreated the issue in the repo below, using ai sdk template files. 

https://github.com/jonnyhd/on-data-issue

Seeking guidance on whether my assumptions are correct before raising as an issue. 

May be related to: https://github.com/vercel/ai/issues/8549

</div>

gr2m:
## Problem

When a shared `Chat` instance is passed to `useChat`, the hook-level `onData` callback was ignored even though the `Chat` instance's own `onData` fired. Because transient data parts are never written to message state, a hook cannot recover them via the messages subscription. Fixes #8597.

## Fix

- Add a `~registerDataCallback` subscription channel to `AbstractChat` (mirroring the existing message/status/error callbacks). When a data part arrives, the chat invokes its constructor `onData` **and** every registered callback, so hook-level callbacks *supplement* the chat-level one.
- `useChat` registers its `onData` via `~registerDataCallback` for the shared-`chat` case only (a hook-owned chat already routes `onData` through `new Chat`, so it must not double-fire).
- The `{ chat }` variant of `UseChatOptions` now allows `onData`, so it can be passed without a type cast.

Scope is limited to `onData` (it fans out cleanly since it returns `void`). `onToolCall`/`sendAutomaticallyWhen` have return-value semantics that don't multiplex and remain chat-owned. See `contributing/decisions/2026-07-05-shared-chat-data-callbacks.md`.

## Reproduction / regression test

`packages/react/src/use-chat.issue-8597.test.tsx` (run via `bash reproductions/issue-8597.sh`) covers:
1. A transient data part reaching the hook-level `onData` for a shared `Chat`.
2. A persisted (non-transient) data part reaching **every** hook sharing the same `Chat`.

Both fail on `main` and pass with this change. Existing `use-chat.ui.test.tsx` (36 tests) and core `chat.test.ts` (128 tests) remain green.

## Related

- Supersedes #16625 (duplicate reproduction).
- Reproduces & fixes #8597.

### DIFF
--- a/.changeset/use-chat-shared-chat-on-data.md
+++ b/.changeset/use-chat-shared-chat-on-data.md
@@ -1,0 +1,8 @@
+---
+'ai': patch
+'@ai-sdk/react': patch
+---
+
+fix(react): call the `useChat` `onData` callback when a shared `Chat` instance is supplied
+
+Previously, passing an existing `chat` instance to `useChat` caused the hook-level `onData` callback to be ignored, even though the `Chat` instance's own `onData` fired. `AbstractChat` now supports registering additional data callbacks, and `useChat` registers its `onData` alongside the chat-level one so every hook sharing a `Chat` observes data parts (including transient ones). Fixes #8597.
--- a/contributing/decisions/2026-07-05-shared-chat-data-callbacks.md
+++ b/contributing/decisions/2026-07-05-shared-chat-data-callbacks.md
@@ -1,0 +1,79 @@
+---
+status: proposed
+date: 2026-07-05
+decision-makers:
+---
+
+# Hook-level `onData` composes with a shared `Chat` instance
+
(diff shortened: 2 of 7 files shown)


gr2m:
Closing: the shared Chat instance owns this configuration.

PR #16625, opened 2026-07-02
reproduction: Reproduces #8597

gr2m:
## Reproduction

Created failing reproduction artifacts packages/react/src/use-chat.issue-8597.test.tsx and reproductions/issue-8597.sh. Running `bash reproductions/issue-8597.sh` shows the shared Chat-level onData receives the transient data part, but the useChat-level onData remains [], reproducing issue #8597.

## Branch

bug-8597-3

## Related Issues

Reproduces #8597

### DIFF
--- a/packages/react/src/use-chat.issue-8597.test.tsx
+++ b/packages/react/src/use-chat.issue-8597.test.tsx
@@ -1,0 +1,97 @@
+import { mockId } from '@ai-sdk/provider-utils/test';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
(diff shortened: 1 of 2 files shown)


gr2m:
Closing in favor of #16623, which reproduces the same issue (#8597). Both PRs are near-identical reproductions; consolidating on #16623 because it:

- Uses the repo's standard `createTestServer` harness from `@ai-sdk/test-server/with-vitest`, matching the existing `packages/react/src/use-chat.ui.test.tsx` conventions, rather than a hand-rolled `ChatTransport` mock.
- Includes `afterEach(cleanup)`.
- Avoids the `as Parameters<typeof useChat>[0]` type cast this PR needs.

The reproduction work continues in #16623.

Issue #15860, opened 2026-06-06
Add DeepAgent documentation


InfernalAzazel:
### Description

The functionality of the AI SDK is not very comprehensive. I want DeepAgent to do the functionality and then use the AI SDK protocol, because DeepAgent does not have a front-end component ecosystem, while the AI SDK component ecosystem is very rich. I also do not want to use LangSmith because this constraint is a bit high and the cost is very high for startups.
1. Has anyone used AI SDK+DeepAgent for actual projects?
2. Hope AI SDK adds DeepAgent documentation

I have reviewed the [ai adk](https://ai-sdk.dev/providers/adapters/langchain) and GitHub, but there is too little practice in this area

### AI SDK Version

- ai ^6.0.194

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

felixarntz:
we now have [`@ai-sdk/harness-deepagents`](https://www.npmjs.com/package/@ai-sdk/harness-deepagents) as a harness adapter for  AI SDK's new harness abstraction (use it via `HarnessAgent`). see docs: https://ai-sdk.dev/providers/ai-sdk-harnesses/deepagents

not sure whether this addresses your concern, but I thought it might be worth sharing

InfernalAzazel:
> @ai-sdk/harness-deepagents

Great

InfernalAzazel:
> we now have [`@ai-sdk/harness-deepagents`](https://www.npmjs.com/package/@ai-sdk/harness-deepagents) as a harness adapter for AI SDK's new harness abstraction (use it via `HarnessAgent`). see docs: https://ai-sdk.dev/providers/ai-sdk-harnesses/deepagents
> 
> not sure whether this addresses your concern, but I thought it might be worth sharing

I am currently using Aegra for my products, and I love the AI SDK too much because its design philosophy is too easy to use. I did not see any examples of multi-user in the documentation

InfernalAzazel:
I love the design of the AI SDK so much that it is easy to use

- What is the performance of creating VercelSandbox（sandbox-just-bash） like? Will 4GB of quad core 8GB memory cause server crashes? 10000 users，Active may be 100-200
- I am currently using Aegra's product, and the AI SDK hopes to provide a universal Cron/scheduled jobs feature. Automation is very important
- My current skill is castrating script execution because using sandbox Docker requires high server resources, which is too much burden for startups， Do not execute Sandbox, read skills into context as needed to save costs
```ts
[...]

PR #16634, opened 2026-07-02
docs: clarify Deep Agents sandbox requirements

tianma-if:
## Summary

- Clarify that the Deep Agents harness requires a network sandbox even when the agent primarily uses skills.
- Explain that the adapter starts a bridge inside the sandbox and communicates over an exposed port.
- Point readers to activeTools/inactiveTools and permissionMode for restricting built-in execution.

Addresses #15860.

## Verification

- ./node_modules/.bin/oxfmt --check content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
- ./node_modules/.bin/oxlint content/providers/02-ai-sdk-harnesses/05-deepagents.mdx

### DIFF
--- a/content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
+++ b/content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
@@ -162,6 +162,13 @@ const sandbox = createVercelSandbox({
 });
 ```
 
+The sandbox is required even when your agent primarily uses skills. The adapter
+starts a Deep Agents bridge inside the sandbox and communicates with it over the
+exposed port. Use [`activeTools`](/docs/ai-sdk-harnesses/tools#active-tools) or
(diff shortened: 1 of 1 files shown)


tianma-if:
Note for review: the Vercel check appears to be failing on preview deployment authorization for this fork PR (the target URL is the Vercel GitHub authorize page). The local checks listed in the PR description passed for the changed file.

tianma-if:
Addressed the Vercel Agent Review comment by updating the tool filtering links to the existing #tool-filtering anchor. Re-ran the file-level formatting and lint checks successfully.

felixarntz (commented):
@tianma-if Thank you for the PR. However, I don't think this is worth adding to the Deep Agents harness documentation, because it applies to the harness layer overall, i.e. it by definition is the case for every single harness.

Mandating a sandbox encourages safe usage by default, therefore it is baked into the abstraction layer itself.

PR #16635, opened 2026-07-02
docs: update Maxim useChat example

tianma-if:
## Summary

- Update the Maxim observability chat route to return a UI message stream response.
- Replace the legacy ai/react import with @ai-sdk/react.
- Update the client example to use DefaultChatTransport, sendMessage, and UIMessage parts.

## Verification

- ./node_modules/.bin/oxfmt --check content/providers/03-observability/maxim.mdx
- ./node_modules/.bin/oxlint content/providers/03-observability/maxim.mdx

### DIFF
--- a/content/providers/03-observability/maxim.mdx
+++ b/content/providers/03-observability/maxim.mdx
@@ -265,7 +265,7 @@ export async function POST(req) {
     },
   });
 
-  return result.toAIStreamResponse();
+  return result.toUIMessageStreamResponse();
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tianma-if:
Moved this out of draft after re-checking the patch. Local verification: `./node_modules/.bin/oxfmt --check content/providers/03-observability/maxim.mdx` and `git diff --check` passed. The Vercel check points to the Vercel GitHub authorization page, so it appears to be an authorization/setup issue rather than a docs build failure.

PR #16639, opened 2026-07-02
docs: update Braintrust stream response helper

tianma-if:
## Summary

- Update the Braintrust streaming route example to use the current streamText response helper.
- Replace the stale toDataStreamResponse call with toUIMessageStreamResponse.

## Verification

- ./node_modules/.bin/oxfmt --check content/providers/03-observability/braintrust.mdx
- ./node_modules/.bin/oxlint content/providers/03-observability/braintrust.mdx

### DIFF
--- a/content/providers/03-observability/braintrust.mdx
+++ b/content/providers/03-observability/braintrust.mdx
@@ -67,7 +67,7 @@ export async function POST(req: Request) {
     prompt,
   });
 
-  return result.toDataStreamResponse();
+  return result.toUIMessageStreamResponse();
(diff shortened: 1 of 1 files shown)


tianma-if:
Moved this out of draft after re-checking the patch. Local verification: `./node_modules/.bin/oxfmt --check content/providers/03-observability/braintrust.mdx` and `git diff --check` passed. The Vercel check points to the Vercel GitHub authorization page, so it appears to be an authorization/setup issue rather than a docs build failure.

PR #16641, opened 2026-07-02
docs: update MCP completion response helper

tianma-if:
## Summary

- Update the MCP tools cookbook completion route to avoid the stale toDataStreamResponse helper.
- Use createUIMessageStreamResponse and toUIMessageStream, matching the current useCompletion docs pattern.

## Verification

- ./node_modules/.bin/oxfmt --check content/cookbook/01-next/73-mcp-tools.mdx
- ./node_modules/.bin/oxlint content/cookbook/01-next/73-mcp-tools.mdx

### DIFF
--- a/content/cookbook/01-next/73-mcp-tools.mdx
+++ b/content/cookbook/01-next/73-mcp-tools.mdx
@@ -18,7 +18,11 @@ If you prefer to use the official transports (optional), install the official Ty
 
 ```ts filename="app/api/completion/route.ts"
 import { createMCPClient } from '@ai-sdk/mcp';
-import { streamText } from 'ai';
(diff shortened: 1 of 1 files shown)


tianma-if:
Moved this out of draft after re-checking the patch against the current Completion guide, which also uses `createUIMessageStreamResponse` with `toUIMessageStream` for `useCompletion`. Local verification: `./node_modules/.bin/oxfmt --check content/cookbook/01-next/73-mcp-tools.mdx` and `git diff --check` passed. The Vercel check points to the Vercel GitHub authorization page, so it appears to be an authorization/setup issue rather than a docs build failure.

PR #16669, opened 2026-07-03
fix(alibaba): do not derive ratio from unsupported resolution

jibin7jose:
## Background

`@ai-sdk/alibaba` wan2.7 video models (`wan2.7-t2v`, `wan2.7-r2v`, and their dated snapshots) map the SDK `resolution` option to a resolution tier (`720P` / `1080P`) **plus** a separate `ratio` parameter.

When neither `providerOptions.alibaba.ratio` nor the top-level `aspectRatio` is provided, the ratio is inferred from the resolution. The problem is that this inference ran unconditionally on the **raw `options.resolution` value** — even when that resolution had already been rejected as unsupported, warned about, and excluded from the request body.

For example, passing `resolution: "1024x768"` (a 4:3 resolution not in the supported tier map) caused the SDK to:
1. ✅ Correctly warn: `"unsupported resolution: 1024x768 was ignored"`
2. ❌ Incorrectly send: `parameters: { ratio: "4:3" }` to the API

The wan2.7 API only accepts `16:9`, `9:16`, and `1:1` as valid `ratio` values, so the request would be silently rejected by the API. The behaviour was also internally inconsistent: the SDK told the user the resolution was ignored, then used it behind the scenes.

## Summary

Introduced a `resolutionAccepted` boolean flag in the `doGenerate` method of `AlibabaVideoModel`:

- The flag is initialised to `false`.
- It is set to `true` **only** when a resolution is successfully validated and written into `parameters` (i.e., mapped to a `720P` or `1080P` tier for wan2.7, or converted to `WIDTH*HEIGHT` for wan2.6).
- The ratio derivation is now conditioned on `resolutionAccepted`:

```ts
// Before (buggy)
(options.resolution != null
  ? deriveRatioFromResolution(options.resolution)
  : undefined)

// After (fixed)
(resolutionAccepted && options.resolution != null
  ? deriveRatioFromResolution(options.resolution)
  : undefined)


jibin7jose:
Hi @lgrammel, @gr2m, and @n1ckoates — just a gentle ping on this PR whenever you have a moment to take a look. Please let me know if there are any changes needed or if you have any questions about the fix. Thanks!

lgrammel:
The issue turned out to be a no-op since ratio 4:3 is in fact supported by the API.

Issue #13218, opened 2026-03-08
Provider-executed tools (e.g. web_search) never emit tool-result in stream, so UI stays on "input-available"


pepegc:
## Description
When using the xAI Responses API with provider-executed tools such as xai.tools.webSearch(), the stream emits tool-call (and tool-input-* chunks) but never emits a tool-result when the tool completes. The AI SDK therefore never sends a tool-output-available chunk to the client, so UIs that rely on it (e.g. processUIMessageStream / ToolCallStatus) stay stuck on the “in progress” state (e.g. “Searching the web”) instead of moving to “completed”.

## Expected behavior
When a provider-executed tool (e.g. web_search_call, x_search_call, code_execution_call) finishes, the stream should emit a tool-result part (with providerExecuted: true), so that:

The AI SDK’s UI stream includes a tool-output-available chunk.
Client UIs can show a “completed” state (e.g. “Web search concluded”) instead of “Searching the web”.

## Actual behavior
For file_search_call, the adapter correctly emits tool-result on response.output_item.done (see xai-responses-language-model.ts around lines 472–489).
For web_search_call, x_search_call, code_execution_call, view_image_call, view_x_video_call, custom_tool_call, and mcp_call, the adapter only emits tool-input-start, tool-input-delta, tool-input-end, and tool-call. It never emits tool-result, so the tool call never transitions to “output available” in the UI.

## Environment
@ai-sdk/xai: (your version, e.g. ^3.0.67)
ai: (e.g. ^6.0.3)
Model: Responses API, e.g. xai.responses('grok-4-fast-non-reasoning') with tools: { webSearch: xai.tools.webSearch() }

## Suggested fix
In doStream in xai-responses-language-model.ts, when handling response.output_item.done for provider-executed tool types (web_search_call, x_search_call, code_execution_call, etc.), emit a tool-result chunk in addition to the existing tool-call, similar to file_search_call. The result payload can be a minimal object (e.g. {}) or any structured result the xAI API provides for that item, so long as the stream emits a tool-result with the same toolCallId and providerExecuted: true. That will cause the AI SDK to emit tool-output-available and fix the stuck UI state.

## Workaround
Clients can work around this by treating input-available as “completed” when the stream has finished (e.g. when status === 'ready') for known provider-executed tools, but that is fragile and tool-specific; fixing it in the provider is the right long-term solution.

### AI SDK Version

@ai-sdk/xai: (your version, e.g. ^3.0.67)
[...]

PR #16671, opened 2026-07-03
reproduction: Reproduces #13218

lgrammel:
## Bug Reproduction

**Bug was reproduced.**

Created reproduction script examples/ai-functions/src/reproduction/issue-13218-xai-provider-tool-result.ts and failing regression test packages/xai/src/responses/xai-responses-language-model.test.ts. Ran `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-13218-xai-provider-tool-result.ts`; the live xAI Responses webSearch stream emitted providerExecuted tool-call/tool-input chunks but 0 tool-result chunks, failing with the reported missing tool-result. Also ran `pnpm -C packages/xai test:node src/responses/xai-responses-language-model.test.ts -t "should emit a tool-result when a provider-executed web_search tool call is done"`; it failed because the adapter output contained tool-call but no tool-result. Expected issue behavior is a provider-executed tool-result after response.output_item.done; no blocker.

## Branch

bug-13218-4

## Related Issues

Reproduces #13218

### DIFF
--- a/examples/ai-functions/src/reproduction/issue-13218-xai-provider-tool-result.ts
+++ b/examples/ai-functions/src/reproduction/issue-13218-xai-provider-tool-result.ts
@@ -1,0 +1,188 @@
+import 'dotenv/config';
+
+import { xai } from '@ai-sdk/xai';
+import {
+  APICallError,
+  LoadAPIKeyError,
+  streamText,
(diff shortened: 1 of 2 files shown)


lgrammel:
not good enough

PR #16675, opened 2026-07-03
fix(provider/openai, provider/open-responses): guard Responses API responses with no output

shaper:
## Background

The Responses API `doGenerate` in both `@ai-sdk/openai` and `@ai-sdk/open-responses` iterates the response `output` unconditionally:

```ts
for (const part of response.output!) { ... }
```

…but the response schema marks `output` as **optional**. When an (OpenAI-compatible) upstream returns a **200 with no `output`** — e.g. a failed/incomplete generation — this throws an opaque `output is not iterable` TypeError instead of a meaningful provider error. We hit this in production with a third-party provider served through the Responses API (returns a 200 body without `output`); callers got an unhelpful 500 with a minified `X.output is not iterable` message that couldn't be classified.

`@ai-sdk/openai` at least guards an explicit `response.error` field first; `@ai-sdk/open-responses`' `doGenerate` had **no** guard at all.

## Change

Guard `output` in `doGenerate` for both providers: when it's `null`/absent, throw a descriptive `APICallError` (`Responses API returned no output …`), including `incomplete_details.reason` (both) and `status` (open-responses) when present. The iteration then drops its non-null assertion.

This hardens every provider that routes through these Responses models (OpenAI nano models, xAI, Bedrock Mantle, and OpenAI-compatible third parties) against a non-conforming upstream 200 — turning a cryptic crash into an actionable error.

Streaming is unaffected: the stream path parses SSE events and never iterates a top-level `output`.

## Tests

Added a `doGenerate` test to each package: a 200 response body with no `output` (and `incomplete_details.reason: 'content_filter'`) now rejects with `Responses API returned no output (content_filter)`.

- `pnpm --filter @ai-sdk/openai type-check` ✅
- `pnpm --filter @ai-sdk/open-responses type-check` ✅
- new tests pass in both packages ✅
- `npx ultracite check` on changed files: 0 warnings / 0 errors ✅

Changeset: patch for both packages.


kkdawkins (approved):
LGTM

packages/open-responses/src/responses/open-responses-language-model.ts
```diff
@@ -228,10 +229,25 @@ export class OpenResponsesLanguageModel implements LanguageModelV4 {
       fetch: this.config.fetch,
     });
```
kkdawkins, line 231: If the upstream includes `response.error`, surface its message before this fallback; otherwise failed 200 responses with no `output` lose the actual error.

shaper: Good catch — fixed in 9794bae3ab. `doGenerate` now guards `response.error` and throws its message before the no-output fallback, mirroring the `@ai-sdk/openai` provider, so failed 200 responses no longer mask the upstream error. Added a test covering a 200 body with `response.error` and no `output`.

packages/open-responses/src/responses/open-responses-language-model.ts
```diff
@@ -228,10 +229,25 @@ export class OpenResponsesLanguageModel implements LanguageModelV4 {
...
 
+    if (response.output == null) {
```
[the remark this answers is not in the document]
shaper: Fixed in 9794bae3ab — `doGenerate` now surfaces `response.error.message` before the generic "returned no output" fallback, so upstream error messages on 200 responses are no longer masked. Added a regression test.

PR #16677, opened 2026-07-04
docs: include approval tool states in switch example

tianma-if:
## Summary

- Add `approval-requested`, `approval-responded`, and `output-denied` to the tool state switch example.
- Keep the streaming tool call example aligned with the approval states documented earlier on the page.

Fixes #12303.

## Verification

- `./node_modules/.bin/oxfmt --check content/docs/04-ai-sdk-ui/03-chatbot-tool-usage.mdx`
- `git diff --check`


tianma-if:

PR #16679, opened 2026-07-04
docs: clarify originalMessages for merged UI streams

tianma-if:
## Summary

- Pass `originalMessages` through the `createUIMessageStream` reference example when merging a `toUIMessageStream` result.
- Add a note for tool approval flows that continue existing tool calls.

Fixes #12622.

## Verification

- `./node_modules/.bin/oxfmt --check content/docs/07-reference/02-ai-sdk-ui/40-create-ui-message-stream.mdx`
- `git diff --check`

### DIFF
--- a/content/docs/07-reference/02-ai-sdk-ui/40-create-ui-message-stream.mdx
+++ b/content/docs/07-reference/02-ai-sdk-ui/40-create-ui-message-stream.mdx
@@ -47,7 +47,12 @@ const stream = createUIMessageStream({
       prompt: 'Write a haiku about AI',
     });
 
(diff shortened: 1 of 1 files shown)


tianma-if:

gr2m:
Thanks for putting this together. The current main and v6 docs already pass originalMessages to the outer createUIMessageStream, which is the stream responsible for reconstructing the conversation. Passing it to the nested toUIMessageStream is unnecessary in this example because that stream does not manage its own completion callback or response message ID. The page has also changed since this PR was opened, so I’m closing this as superseded.

PR #16685, opened 2026-07-04
docs: add foundations glossary

tianma-if:
## Summary

- Add a Foundations glossary page with shared AI SDK terminology.
- Cover core concepts, prompts and messages, tools, streaming, and lifecycle terms.
- Link the glossary from the Foundations index.

Fixes #14234.

## Verification

- `./node_modules/.bin/oxfmt --check content/docs/02-foundations/07-glossary.mdx content/docs/02-foundations/index.mdx`
- `git diff --check`


tianma-if:

PR #16691, opened 2026-07-05
fix(sandbox): pass env into just-bash commands

### LINKED ISSUE #16690: HarnessAgent + createJustBashSandbox: createSession always fails — work dir never created (just-bash runCommand ignores env, so mkdir -p "$WORK_DIR" silently no-ops)
### Description

`HarnessAgent.createSession()` always fails with `createJustBashSandbox()` because the session work directory is never actually created inside the sandbox. The core's `ensureSandboxDirectory` runs:

```js
// @ai-sdk/harness/dist/agent/index.js (ensureSandboxDirectory)
await session.run({
  command: 'mkdir -p "$WORK_DIR"',
  env: { WORK_DIR: workDir },
});
```

With the just-bash provider this exits `0` **without creating anything** — the `env` passed to `session.run()` never reaches the emulated bash, so `$WORK_DIR` expands to the empty string and `mkdir -p ""` silently succeeds. The harness adapter then dies on its first `cd` into the work dir. With the pi adapter:

```
Error: bash: cd: /home/user/pi-Zt7oKqOhYQzGulxm: No such file or directory
    at readCommandOutput (@ai-sdk/harness-pi/dist/index.js:1137)
    at async listRemoteWorkspaceEntries (…:1164)
    at async syncHostWorkspaceFromSandbox (…:1236)
    at async createPiSession (…:1419)
    at async HarnessAgent.createSession (@ai-sdk/harness/dist/agent/index.js:2730)
```

#### Reproduction 1 — provider level (no LLM, no harness)

```js
import { createJustBashSandbox } from "@ai-sdk/sandbox-just-bash";

const provider = createJustBashSandbox();
const session = await provider.createSession({ sessionId: "repro" });

const r1 = await session.run({ command: 'mkdir -p "$WORK_DIR"', env: { WORK_DIR: "/home/user/repro-dir" } });
console.log(r1.exitCode, r1.stderr); // 0 ""  ← silently did nothing

const r2 = await session.run({ command: "cd /home/user/repro-dir && pwd" });
console.log(r2.exitCode, r2.stderr); // 1 "bash: cd: /home/user/repro-dir: No such file or directory"

const r3 = await session.run({ command: "mkdir -p /home/user/literal && cd /home/user/literal && pwd" });
console.log(r3.exitCode, r3.stdout); // 0 "/home/user/literal"  ← literal paths work fine
```

#### Reproduction 2 — root cause is in `just-bash` itself

`RunCommandParams.env` is a typed, documented option of `Sandbox.runCommand` (`dist/sandbox/Sandbox.d.ts`), and the provider forwards it correctly (`@ai-sdk/sandbox-just-bash/dist/index.js` → `this.sandbox.runCommand({ cmd: "bash", args: ["-c", command], env, … })`). But the emulated bash never sees the variables:

```js
import { Sandbox } from "just-bash";

const sb = await Sandbox.create();
const fin = await sb.runCommand({ cmd: "bash", args: ["-c", "echo [$WORK_DIR]"], env: { WORK_DIR: "/hello" } });
console.log(fin.exitCode, JSON.stringify(await fin.stdout())); // 0 "[]\n"  — expected "[/hello]\n"
```

Note that just-bash's own `Bash.exec` documents exactly these semantics as working (`await env.exec("echo $TEMP", { env: { TEMP: "value" } })` in its README), so this looks like a `Sandbox.runCommand` gap rather than intended behavior. Reproduced on both `just-bash@2.14.5` (what `@ai-sdk/sandbox-just-bash`'s `^2.14.5` range resolves to today) and `just-bash@3.0.2` (latest). If the root-cause part belongs in `vercel-labs/just-bash` instead, happy for it to be moved/split — filing here because the user-facing breakage is that `HarnessAgent` + `createJustBashSandbox()` (the combination the `@ai-sdk/harness-pi` README suggests for local use) fails out of the box.

#### Workaround

Create the work dir with a literal path in `sandboxConfig.onSession`, which runs right after `ensureSandboxDirectory` and before the harness adapter starts:

```js
  harness: createPi({ /* … */ }),
  sandbox: createJustBashSandbox(),
  sandboxConfig: {
    onSession: async ({ session, sessionWorkDir }) => {
      await session.run({ command: `mkdir -p '${sessionWorkDir}'` });
    },
  },
});
```

With that in place, a full `createSession()` + `generate()` turn with the pi adapter completes normally (host tools included).

A defensive fix on the harness side would also help regardless of the just-bash fix: `ensureSandboxDirectory` could avoid depending on `env` expansion (embed the quoted literal path in the command), and/or verify the directory exists afterwards instead of trusting `exit 0`.

### AI SDK Version

- ai: 7.0.11
- @ai-sdk/harness: 1.0.14
- @ai-sdk/harness-pi: 1.0.14
- @ai-sdk/sandbox-just-bash: 1.0.14 (also checked the 1.0.15–1.0.18 changelogs — dependency bumps only, so latest should be affected too)
- just-bash: 2.14.5 and 3.0.2 (standalone repro)
- node: 24.18.0 (darwin)

### Code of Conduct

harshmathurx:
## Background

`HarnessAgent.createSession()` can fail with `createJustBashSandbox()` because the harness asks the sandbox to create a work directory via an env-backed command:

```ts
session.run({ command: 'mkdir -p "$WORK_DIR"', env: { WORK_DIR: workDir } })
```

The just-bash session was executing the command without exporting `RunCommandParams.env`, so `$WORK_DIR` expanded as empty. That made directory creation look successful while leaving the expected work directory missing, and the first later `cd` into that path failed.

Root cause: `runCommand` accepted an `env` field at the sandbox/session boundary, but the just-bash command wrapper only launched the command text and did not pass those variables into the bash invocation.

## Summary

- Export `RunCommandParams.env` into just-bash commands before running the command body.
- Preserve shell quoting for env values instead of interpolating them unsafely.
- Add regression coverage for the harness-shaped `$WORK_DIR` directory creation case and for shell-quoted env values.

## Contributor Credit

Thanks to @ZyroEolu-sk for the minimal reproduction in #16690. @lgrammel, tagging you for review since you already had context on the issue/PR thread.

## End-to-End Verification

The focused verification exercises the same contract the harness depends on: a just-bash command can receive `env`, expand it inside bash, and create/use the requested path.

## Test Plan

- `npx pnpm@10.33.4 install --filter @ai-sdk/sandbox-just-bash... --frozen-lockfile`
- `npx pnpm@10.33.4 --filter @ai-sdk/provider build`
- `npx pnpm@10.33.4 --filter @ai-sdk/provider-utils build`
- `npx pnpm@10.33.4 --dir packages/sandbox-just-bash test -- just-bash-sandbox-session.test.ts`
- `git diff --check`
- `npx pnpm@10.33.4 --dir packages/sandbox-just-bash type-check`

Local note: this runner had Node v20.19.2 while the repo requests Node >=22, so pnpm emitted engine warnings. The focused test and type-check still passed locally. The currently visible automated checks are not reporting a failing test; this is waiting on review.

## Risk / reviewer notes

Small behavior fix in the just-bash provider. The only intended behavior change is that per-command env values supplied through the public sandbox API are available to the command. Values are shell-quoted before export to avoid changing command parsing for spaces/special characters.

## Checklist

## Related Issues

Fixes #16690.

### DIFF
--- a/.changeset/fuzzy-mails-brush.md
+++ b/.changeset/fuzzy-mails-brush.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/sandbox-just-bash': patch
+---
+
+Pass per-command env values through to just-bash commands.
--- a/packages/sandbox-just-bash/src/just-bash-sandbox-session.test.ts
+++ b/packages/sandbox-just-bash/src/just-bash-sandbox-session.test.ts
@@ -53,6 +53,41 @@ describe('JustBashSandboxSession', () => {
       expect(result.stdout.trim()).toBe('/tmp/cwd-test');
     });
 
+    it('honours env variables', async () => {
+      const result = await sandbox.run({
+        command: 'printf "%s" "$WORK_DIR"',
+        env: { WORK_DIR: '/home/user/repro-dir' },
+      });
+
+      expect(result.exitCode).toBe(0);
+      expect(result.stdout).toBe('/home/user/repro-dir');
+    });
+
+    it('uses env variables when creating directories', async () => {
+      const mkdir = await sandbox.run({
+        command: 'mkdir -p "$WORK_DIR"',
+        env: { WORK_DIR: '/home/user/repro-dir' },
+      });
+
+      expect(mkdir.exitCode).toBe(0);
+
+      const cd = await sandbox.run({
+        command: 'cd /home/user/repro-dir && pwd',
+      });
+
+      expect(cd.exitCode).toBe(0);
+      expect(cd.stdout.trim()).toBe('/home/user/repro-dir');
+    });
+
+    it('shell-quotes env values', async () => {
+      const result = await sandbox.run({
+        command: 'printf "%s" "$PAYLOAD"',
+        env: { PAYLOAD: "one two ' three" },
+      });
+
+      expect(result.stdout).toBe("one two ' three");
+    });
+
     it('throws when abortSignal is already aborted', async () => {
       const ac = new AbortController();
       ac.abort();
--- a/packages/sandbox-just-bash/src/just-bash-sandbox-session.ts
+++ b/packages/sandbox-just-bash/src/just-bash-sandbox-session.ts
@@ -48,9 +48,8 @@ export class JustBashSandboxSession implements Experimental_SandboxSession {
 
     const finished = await this.sandbox.runCommand({
       cmd: 'bash',
-      args: ['-c', command],
+      args: ['-c', withEnvironment(command, env)],
       ...(workingDirectory !== undefined ? { cwd: workingDirectory } : {}),
-      ...(env !== undefined ? { env } : {}),
       ...(abortSignal !== undefined ? { signal: abortSignal } : {}),
     });
 
@@ -81,10 +80,9 @@ export class JustBashSandboxSession implements Experimental_SandboxSession {
 
     const live = await this.sandbox.runCommand({
       cmd: 'bash',
-      args: ['-c', command],
+      args: ['-c', withEnvironment(command, env)],
       detached: true,
       ...(workingDirectory !== undefined ? { cwd: workingDirectory } : {}),
-      ...(env !== undefined ? { env } : {}),
       ...(abortSignal !== undefined ? { signal: abortSignal } : {}),
     });
 
@@ -190,6 +188,27 @@ export class JustBashSandboxSession implements Experimental_SandboxSession {
       abortSignal,
     });
   }
+}
+
+function withEnvironment(
+  command: string,
+  env: Record<string, string> | undefined,
+): string {
+  if (env == null || Object.keys(env).length === 0) return command;
+
+  const exports: string[] = [];
+  for (const key of Object.keys(env)) {
+    if (!/^[A-Za-z_][A-Za-z0-9_]*$/.test(key)) continue;
+    exports.push(`export ${key}=${shellQuote(env[key])}`);
+  }
+
+  if (exports.length === 0) return command;
+
+  return `${exports.join('\n')}\n${command}`;
+}
+
+function shellQuote(value: string): string {
+  return `'${value.split("'").join("'\\''")}'`;
 }
 
 function createSandboxProcess(


harshmathurx:
@lgrammel gentle ping for review when you have a moment — straightforward fix for just-bash env passthrough, tests pass locally. Thanks!

harshmathurx:
@lgrammel gentle maintainer ping on this one. The fix is intentionally small and aimed at preserving expected env passthrough for `just-bash` commands. If this matches the intended provider behavior, I’d appreciate a review when you have a moment. Happy to revise if there is a repo-specific constraint I missed.

harshmathurx:
Pushed a small follow-up after re-checking the repo requirements:

- added the missing patch changeset for `@ai-sdk/sandbox-just-bash`
- reran the focused just-bash session test: `17 passed`
- reran `packages/sandbox-just-bash` type-check
- `verify-changesets` and the GitHub Actions checks are now green on the latest commit

The only remaining non-green status I see is the `Vercel` deployment status, which points to authorization required rather than a package/test failure.

harshmathurx:
Follow-up on the remaining non-green status at `5c5357078355c2390782453f42b8f72594d3dabb`:

- GitHub Actions CI is green: https://github.com/vercel/ai/actions/runs/30891033877
- `verify-changesets` is green: https://github.com/vercel/ai/actions/runs/30891036639
- Focused local verification still passes after building workspace deps: `pnpm --dir packages/sandbox-just-bash test` -> `17 passed`; `pnpm --dir packages/sandbox-just-bash type-check` -> pass
- The only failing status is the external `Vercel` status, whose reported error is `Authorization required to deploy.` and target is the Vercel authorization URL for this SHA/PR. That looks unrelated to the just-bash package diff (code + tests + changeset) and is not a GitHub Actions job I can rerun from the PR.

gr2m:
Created #19853 to track the remaining provider-level bug. The original `HarnessAgent.createSession()` crash from #16690 no longer reproduces on `main`, but per-command `env` is still ignored. This PR should be evaluated against that narrower issue.

Issue #16694, opened 2026-07-05
@ai-sdk/harness-claude-code: Claude Code's native ToolSearch tool is missing from CLAUDE_CODE_BUILTIN_TOOLS, so every ToolSearch call surfaces as an invalid tool call (NoSuchToolError)


doneumark:
## Description

The Claude Code CLI now ships a native `ToolSearch` tool (deferred MCP tool loading, on by default — see [tool search docs](https://code.claude.com/docs/en/agent-sdk/tool-search)). The bridge's pinned CLI (`@anthropic-ai/claude-code@2.1.177`) has it, and any `HarnessAgent` that registers user tools (which are exposed to the CLI as an in-process MCP server, `harness-tools`) triggers it: the CLI defers those MCP tools and the model calls `ToolSearch` to hydrate them.

`CLAUDE_CODE_BUILTIN_TOOLS` in `packages/harness-claude-code/src/claude-code-harness.ts` does not include `ToolSearch`, so `validateToolCall` → `parseToolCall` marks the call `invalid: true` with a `NoSuchToolError`:

```
Model tried to call unavailable tool 'ToolSearch'. Available tools: read, write, edit, bash, glob, grep, webSearch, WebFetch, NotebookEdit, TodoWrite, Agent, TaskCreate, TaskGet, TaskUpdate, TaskList, TaskStop, TaskOutput, Monitor, ListMcpResources, ReadMcpResource, ExitPlanMode, EnterWorktree, ExitWorktree, AskUserQuestion, Skill, ...
```

The tool actually executes fine inside the CLI (it is provider-executed; the `tool_reference` result streams through and renders). But the invalid-flagged tool call becomes a `tool-input-error` UI chunk whose construction invokes the `toUIMessageStream` `onError` callback — so consumers that treat `onError` as \"the turn failed\" (log as error, discard resume state, etc.) get a false fatal on an otherwise healthy turn.

Note the root cause chain: `ToolSearch` is missing from `@anthropic-ai/claude-agent-sdk`'s published typings too (checked `agentSdkTypes.d.ts` / `sdk-tools.d.ts` in 0.3.177 — the tool exists only in the CLI binary), so the transcription that produced `CLAUDE_CODE_BUILTIN_TOOLS` had nothing to copy. This will presumably recur for every future native tool the CLI grows ahead of the adapter.

## Reproduction

[...]

PR #16698, opened 2026-07-05
reproduction: Reproduces #16694

gr2m:
## Bug Reproduction

**Bug was reproduced.**

Created reproduction artifact examples/ai-functions/src/reproduction/16694-claude-code-toolsearch.ts and ran `pnpm -C examples/ai-functions exec tsx src/reproduction/16694-claude-code-toolsearch.ts`. Observed createClaudeCode().builtinTools does not include ToolSearch, and validateToolCall returns dynamic:true, invalid:true, providerExecuted:true with AI_NoSuchToolError for ToolSearch, then the script fails intentionally. Expected per issue: ToolSearch should be declared/accepted or unknown provider-executed tools should not be invalid NoSuchToolError. Blocker: none.

## Branch

bug-16694-1

## Related Issues

Reproduces #16694

### DIFF
--- a/examples/ai-functions/src/reproduction/16694-claude-code-toolsearch.ts
+++ b/examples/ai-functions/src/reproduction/16694-claude-code-toolsearch.ts
@@ -1,0 +1,71 @@
+import { createClaudeCode } from '@ai-sdk/harness-claude-code';
+import { NoSuchToolError } from 'ai';
+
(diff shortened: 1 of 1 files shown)


aayush-kapoor:
reproduced and fixed in https://github.com/vercel/ai/pull/17184

PR #16730, opened 2026-07-06
fix(langchain): avoid call stack overflow when base64-encoding large binary content

mahirhir:
### What

`convertUserContent` in `@ai-sdk/langchain` base64-encodes inline binary image and file data with:

```ts
btoa(String.fromCharCode(...bytes))
```

Spreading a `Uint8Array` into `String.fromCharCode` passes one argument per byte, so once `bytes` is larger than the engine's argument limit it throws `RangeError: Maximum call stack size exceeded`. Real images and PDFs clear that limit easily, so sending inline binary content to a LangChain model through the adapter crashes instead of encoding it. The same expression appears three times in `convertUserContent`: the `image` branch, and both the image-file and non-image-file branches.

### Fix

Use `convertUint8ArrayToBase64` from `@ai-sdk/provider-utils`, which builds the string with a loop instead of a spread. The SDK already avoids this footgun in the two other places that encode large buffers: `convertUint8ArrayToBase64` itself, and `encodeRealtimeAudio` in `packages/ai/src/realtime/audio-utils.ts`, which chunks the array specifically to stay under the argument limit. For any input that did not already overflow, the output is byte-identical, so existing behaviour and the existing image assertions are unchanged.

### Test

Added two regression tests that push 300 KB of binary through the `image` and `file` paths of `convertUserContent` and check that the base64 round-trips back to the original bytes.

Reverting only the `utils.ts` change and rerunning the suite fails exactly where expected:

```
× should convert large binary image data without exceeding the call stack
    RangeError: Maximum call stack size exceeded  (utils.ts:219)
× should convert large binary file data without exceeding the call stack
    RangeError: Maximum call stack size exceeded  (utils.ts:382)
```

With the change, `packages/langchain/src/utils.test.ts` passes (128 tests). Existing cases such as `[1, 2, 3]` -> `AQID` are unaffected since the encoding output is identical.

### DIFF
--- a/.changeset/langchain-base64-large-binary.md
+++ b/.changeset/langchain-base64-large-binary.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/langchain': patch
+---
+
+Fix `convertUserContent` throwing `RangeError: Maximum call stack size exceeded` on large binary image and file inputs. The base64 encoding used `btoa(String.fromCharCode(...bytes))`, which passes one argument per byte and overflows the argument limit for real-sized images and documents. It now uses `convertUint8ArrayToBase64` from `@ai-sdk/provider-utils`, which the rest of the SDK already relies on. The encoded output is unchanged for inputs that did not previously overflow.
--- a/.changeset/langchain-base64-large-binary.md
+++ b/.changeset/langchain-base64-large-binary.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/langchain': patch
+---
+
+Fix `convertUserContent` throwing `RangeError: Maximum call stack size exceeded` on large binary image and file inputs. The base64 encoding used `btoa(String.fromCharCode(...bytes))`, which passes one argument per byte and overflows the argument limit for real-sized images and documents. It now uses `convertUint8ArrayToBase64` from `@ai-sdk/provider-utils` (moved to dependencies since it is now a runtime import). The encoded output is unchanged for inputs that did not previously overflow.
--- a/packages/langchain/package.json
+++ b/packages/langchain/package.json
@@ -36,10 +36,10 @@
     }
   },
   "dependencies": {
+    "@ai-sdk/provider-utils": "workspace:*",
     "ai": "workspace:*"
   },
   "devDependencies": {
-    "@ai-sdk/provider-utils": "workspace:*",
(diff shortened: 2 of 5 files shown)


mahirhir:
Two things a triager might want, since this has been sitting without a comment.

**The red check is not the code.** The only failing check is the deploy, and its status text says why:

```
Vercel = failure :: Authorization required to deploy.
```

That is the outside-fork deploy gate, which needs a team member rather than a change from me. Everything that inspects the diff passed: `Vercel Security Review` (no findings), `Socket Security: Project Report`, and `Vercel – Code Owners` (skipped, not configured).

**The defect is still on `main`, and the threshold is unremarkable.** `String.fromCharCode(...bytes)` passes one argument per byte, so it overflows the argument limit rather than doing anything proportional to data size. On node v24.15.0:

```
      1024 bytes   spread: ok                                   replacement: ok
     16384 bytes   spread: ok                                   replacement: ok
     65536 bytes   spread: ok                                   replacement: ok
    102400 bytes   spread: ok                                   replacement: ok
    131072 bytes   spread: RangeError: Maximum call stack ...   replacement: ok
    262144 bytes   spread: RangeError: Maximum call stack ...   replacement: ok
   1048576 bytes   spread: RangeError: Maximum call stack ...   replacement: ok

CONTROL identical output on a small input: true
```

It fails somewhere between 100 KB and 128 KB. That is an ordinary phone photo, not a pathological input. The exact cutoff is engine and stack dependent, so treat the boundary as approximate; what is not approximate is that a normal-sized image throws.

The control line matters more than the failures: on inputs where the old path survives, both produce byte-identical base64. This changes behaviour only where the old behaviour was an exception.

Current `main`, checked just now:

```console
$ grep -c 'String.fromCharCode(\.\.\.' packages/langchain/src/utils.ts
2
$ grep -c 'convertUint8ArrayToBase64' packages/langchain/src/utils.ts
0
```

**One thing that is on me:** the branch is 194 commits behind `main`. GitHub still reports it mergeable, but it touches `pnpm-lock.yaml`, so a refresh is probably wanted before merge and may need the lockfile regenerating rather than merged. Say the word and I will bring it current; I did not want to force a rebase onto a PR nobody has looked at yet, in case you would rather I split the lockfile change out or drop the dependency move entirely and inline the conversion.

PR #16741, opened 2026-07-06
fix(ai): resolve functional body in prepareSendMessagesRequest before serialization

### LINKED ISSUE #11253: DefaultChatTransport: Functional body not serialized correctly in prepareSendMessagesRequest (React 19 / Next 16)
### Description

I am using the latest versions of the AI SDK with React 19 and Next 16. I've encountered a bug where the DefaultChatTransport fails to correctly handle or serialize a dynamic body function, making it impossible to pass data from a React Ref to the API route.

The Problem:

Ref Access Error: Passing the data as a static object inside prepareSendMessagesRequest triggers the React 19 error: Error: Cannot access refs during render. This is expected as the hook is initialized during render.

Functional Body Bug: To fix this, I followed the documentation for "Dynamic Configuration," passing the body as a function. While this resolves the React Ref error, the data never reaches the server.

Server Side Crash: The API route receives an empty or malformed body, causing a crash. In Next.js 16, this manifests as an Invalid source map / sourceMapURL could not be parsed error because the route fails to parse req.json().

Code Snippets:
❌ Option 1: Static Object (Throws React 19 Render Error)

```

const { sendMessage } = useChat({
    api: "/api/chat",
    prepareSendMessagesRequest({ messages, body }) {
      return {
        body: {
          documentId: editorRef.current?.documentId || '', // ❌ React 19 Error
          messages,
          ...body,
        },
      };
    },
  }),
});
```
⚠️ Option 2: Functional Body (Resolves React Error, but Server receives empty body)

```
const { sendMessage } = useChat({
    api: "/api/chat",
    prepareSendMessagesRequest({ messages, body }) {
      return {
        body: () => ({
          documentId: editorRef.current?.documentId || '',
          messages,
          ...body,
        }),
      };
    },
  }),
});
```
Expected Behavior:
The DefaultChatTransport should detect that body is a function, execute it at the time of the fetch request (outside of the React render cycle), and send the resulting object as a JSON payload. Currently, it seems the function is either not being executed or its return value is not being attached to the fetch call.
Question:
What is the correct way to send dynamic data (like editor content from a Ref) in the request body without using the sendMessage arguments directly? Is the functional body approach currently broken in React 19 environments?

### AI SDK Version

ai: ^5.0.114

@ai-sdk/react: ^2.0.76

react: 19.2.0

next: 16.0.7

### Code of Conduct

Aayush-engineer:
## Background

DefaultChatTransport (via HttpChatTransport) fails to serialize the request body when prepareSendMessagesRequest returns body as a function. The function is passed directly to JSON.stringify without being called, producing an empty or missing request body on the server.
This affects React 19 users who use a functional body to defer ref access outside the render cycle — the documented "Dynamic Configuration" pattern silently sends nothing to the server.

Fixes #11253

## Summary

Updated PrepareSendMessagesRequest return type to allow body: Resolvable<object> (supports plain object, Promise, or function returning either)
Added await resolve(preparedRequest.body) before serialization, consistent with how this.body, this.headers, and this.credentials are already resolved in the same method

## Checklist

 All commits are signed
 A patch changeset has been added
 I have reviewed this pull request

## Related Issues
Fixes #11253

### DIFF
--- a/.changeset/honest-pets-study.md
+++ b/.changeset/honest-pets-study.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix DefaultChatTransport to execute functional body returned from prepareSendMessagesRequest before serialization
--- a/packages/ai/src/ui/http-chat-transport.ts
+++ b/packages/ai/src/ui/http-chat-transport.ts
@@ -23,13 +23,13 @@ export type PrepareSendMessagesRequest<UI_MESSAGE extends UIMessage> = (
   },
 ) =>
   | {
-      body: object;
+      body: Resolvable<object>;
       headers?: HeadersInit;
       credentials?: RequestCredentials;
       api?: string;
     }
   | PromiseLike<{
-      body: object;
+      body: Resolvable<object>;
       headers?: HeadersInit;
       credentials?: RequestCredentials;
       api?: string;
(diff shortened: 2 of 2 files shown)


lgrammel:
Thank you for the pull request. Issue #11253 was closed as not planned. Closing this pull request.

PR #16761, opened 2026-07-06
[v6.0] fix: normalize bare Anthropic API base URL to avoid missing /v1 path

### LINKED ISSUE #15542: @ai-sdk/anthropic: ANTHROPIC_BASE_URL convention differs from official Anthropic SDK, breaks under Claude Code / Cursor
### What happened

A [Twenty](https://github.com/twentyhq/twenty) server running inside Claude Code desktop kept getting `404 Not Found` from every Anthropic call:

```
url: 'https://api.anthropic.com/messages'
model: 'claude-sonnet-4-5-20250929'
```

The 404 isn't the model — it's the path. Claude Code injects `ANTHROPIC_BASE_URL=https://api.anthropic.com` into child processes. `@ai-sdk/anthropic` reads that and appends `/messages` → wrong path.

### The mismatch

The two Anthropic SDKs interpret `ANTHROPIC_BASE_URL` differently.

`@anthropic-ai/sdk` (`src/client.ts`):
- default baseURL: `https://api.anthropic.com`
- adds version itself: `${baseURL}/v1/messages`

`@ai-sdk/anthropic` (`anthropic-provider.ts`):
- default baseURL: `https://api.anthropic.com/v1`
- appends raw path: `${baseURL}/messages`

So `ANTHROPIC_BASE_URL=https://api.anthropic.com` works with the official SDK and 404s under this one. Setting `…/v1` works here but would double up to `/v1/v1/messages` under the official SDK. Same env var, mutually exclusive meanings.

### Why it matters

Anything that injects this env var following Anthropic's own convention silently breaks any app embedding `@ai-sdk/anthropic`. The user hit it via Claude Code; same trap exists for Cursor's agent tooling, LiteLLM proxy setups that document the bare-host form, and people who configured shells against the official SDK and later adopted `ai`.

The default `https://api.anthropic.com/v1` keeps `/messages` working without an env var. The break only shows up when something else sets the env.

### Reproduce

```bash
ANTHROPIC_BASE_URL=https://api.anthropic.com \
  ANTHROPIC_API_KEY=sk-ant-… \
  node -e "import('@ai-sdk/anthropic').then(async ({anthropic}) => {
    const {generateText} = await import('ai');
    console.log(await generateText({model: anthropic('claude-sonnet-4-5-20250929'), prompt: 'hi'}));
  })"
```

→ `AI_APICallError: Not Found` from `https://api.anthropic.com/messages`. Drop the env var (or append `/v1`) and it succeeds.

### Possible fixes
1. Detect the bare-host form (`https://api.anthropic.com` with no path) and append `/v1` with a one-time warning. 
2. Switch to the official SDK's convention (default `…anthropic.com`, prepend `/v1/messages` internally).

lgrammel:
## Background

Claude Code/Cursor can inject ANTHROPIC_BASE_URL=https://api.anthropic.com, which made @ai-sdk/anthropic call https://api.anthropic.com/messages instead of the versioned Anthropic API path.

## Summary

Added Anthropic provider base URL normalization so the bare official API host resolves to https://api.anthropic.com/v1 while preserving existing versioned and custom base URLs.

## Testing

Added regression coverage for bare https://api.anthropic.com values from both ANTHROPIC_BASE_URL and the baseURL option.

## Related Issues

Fixes #15542

Backport of #16584

### DIFF
--- a/.changeset/quiet-starfishes-provide.md
+++ b/.changeset/quiet-starfishes-provide.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Normalize a bare `https://api.anthropic.com` base URL to include `/v1`.
--- a/packages/anthropic/src/anthropic-provider.test.ts
+++ b/packages/anthropic/src/anthropic-provider.test.ts
@@ -83,6 +83,41 @@ describe('createAnthropic', () => {
       expect(requestUrl).toBe('https://proxy.anthropic.example/v1/messages');
     });
 
+    it('normalizes a bare Anthropic API URL from ANTHROPIC_BASE_URL', async () => {
+      process.env.ANTHROPIC_BASE_URL = 'https://api.anthropic.com/';
+
+      const fetchMock = createFetchMock();
+      const provider = createAnthropic({
+        apiKey: 'test-api-key',
+        fetch: fetchMock,
+      });
+
+      await provider('claude-3-haiku-20240307').doGenerate({
+        prompt: TEST_PROMPT,
+      });
+
+      expect(fetchMock).toHaveBeenCalledTimes(1);
+      const [requestUrl] = fetchMock.mock.calls[0]!;
+      expect(requestUrl).toBe('https://api.anthropic.com/v1/messages');
+    });
+
+    it('normalizes a bare Anthropic API URL from the baseURL option', async () => {
+      const fetchMock = createFetchMock();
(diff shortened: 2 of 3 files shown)


gr2m (approved):
Reviewed: backport of #15542 — normalizes bare https://api.anthropic.com (from option or ANTHROPIC_BASE_URL) to include /v1, with tests for both paths. CI fully green.

PR #16763, opened 2026-07-06
[v6.0] fix: Raw `JSON.parse` used in production provider code (prototype pollution risk)

### LINKED ISSUE #15812: Raw `JSON.parse` used in production provider code (prototype pollution risk)
### Description

Several production provider files use `JSON.parse` directly instead of the SDK's `secureJsonParse` utility from `@ai-sdk/provider-utils`. This violates the project coding standard and leaves a prototype pollution vector open on tool call inputs that flow from LLM responses.

The root problem is that `secureJsonParse` — the synchronous safe parser in `packages/provider-utils/src/secure-json-parse.ts` — is **not exported** from `@ai-sdk/provider-utils`. Because `parseJSON`/`safeParseJSON` are async and these call sites are synchronous, providers have no way to use the secure variant without importing from an internal path.

## Affected files

| File | Line | Context |
|------|------|---------|
| `packages/google/src/interactions/convert-to-google-interactions-input.ts` | 379 | `safeParseToolArgs` — parses tool call input strings from LLM responses |
| `packages/google/src/convert-to-google-messages.ts` | 471 | Parses `part.input` (tool call args) when converting to Google format |
| `packages/anthropic/src/convert-to-anthropic-prompt.ts` | 51 | `extractErrorValue` — parses provider tool error result values |
| `packages/anthropic/src/convert-to-anthropic-prompt.ts` | 847 | Parses `output.value` from code execution error results |
| `packages/anthropic/src/anthropic-language-model.ts` | 2108 | Parses `contentBlock.input` for `code_execution` tool detection |
| `packages/gateway/src/errors/extract-api-call-response.ts` | 9 | Parses `error.responseBody` from API error responses |

## Why this matters

`secureJsonParse` guards against prototype pollution by checking for `__proto__` and `constructor.prototype` keys before returning the parsed value. Tool call inputs arrive from LLM model responses, which can be manipulated via prompt injection. A crafted payload like `{"__proto__": {"polluted": true}}` would pass through raw `JSON.parse` silently.

## Proposed fix

**Step 1** — export `secureJsonParse` from `packages/provider-utils/src/index.ts`:
```ts
export { secureJsonParse } from './secure-json-parse';
```

**Step 2** — replace `JSON.parse(...)` with `secureJsonParse(...)` at all six locations. Each site is already wrapped in a `try/catch`, so error handling stays the same.

Example (`packages/google/src/interactions/convert-to-google-interactions-input.ts`):
```ts
// Before
const parsed = JSON.parse(input);

// After
import { secureJsonParse } from '@ai-sdk/provider-utils';
const parsed = secureJsonParse(input);
```

### AI SDK Version

Reproduced on latest main (post-6.0.168).
Affected packages: @ai-sdk/google, @ai-sdk/anthropic, @ai-sdk/gateway, @ai-sdk/provider-utils (all latest)

### Code of Conduct

lgrammel:
## Background

Provider code was synchronously parsing LLM/tool/error JSON with raw JSON.parse, bypassing the SDK's prototype-pollution protections.

## Summary

Exported secureJsonParse from @ai-sdk/provider-utils and replaced the six reported production JSON.parse call sites in Google, Anthropic, and Gateway code with secureJsonParse while preserving existing try/catch fallback behavior.

## Testing

Kept the reproduction regression test that verifies prototype-polluting Google Interactions tool-call input is rejected/falls back instead of being parsed as arguments.

## Related Issues

Fixes #15812

Backport of #16579

### DIFF
--- a/.changeset/quick-swans-pull.md
+++ b/.changeset/quick-swans-pull.md
@@ -1,0 +1,8 @@
+---
+"@ai-sdk/anthropic": patch
+"@ai-sdk/gateway": patch
+"@ai-sdk/google": patch
+"@ai-sdk/provider-utils": patch
+---
+
+Prevent prototype pollution when synchronously parsing provider JSON inputs and expose `secureJsonParse` from provider-utils.
--- a/packages/anthropic/src/anthropic-messages-language-model.ts
+++ b/packages/anthropic/src/anthropic-messages-language-model.ts
@@ -24,6 +24,14 @@ import {
   parseProviderOptions,
   postJsonToApi,
   resolve,
+<<<<<<< HEAD:packages/anthropic/src/anthropic-messages-language-model.ts
+=======
+  resolveProviderReference,
+  secureJsonParse,
+  serializeModelOptions,
+  WORKFLOW_SERIALIZE,
+  WORKFLOW_DESERIALIZE,
+>>>>>>> c6f5e624a (fix: Raw `JSON.parse` used in production provider code (prototype pollution risk) (#16579)):packages/anthropic/src/anthropic-language-model.ts
   type FetchFunction,
   type InferSchema,
   type ParseResult,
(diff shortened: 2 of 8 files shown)


gr2m (approved):
Reviewed: backport of #15812 — replaces raw JSON.parse with secureJsonParse in anthropic/google/gateway sync-parse paths and exports secureJsonParse from provider-utils (file already exists on v6); includes prototype-pollution regression test. CI fully green.

PR #16779, opened 2026-07-06
fix: generateText structured output fails when finishReason is missing

### LINKED ISSUE #11348: generateText with Output.object() throws NoOutputGeneratedError when finishReason is undefined
### Description

When using `generateText` with `Output.object()`, the SDK only parses structured output when `finishReason === "stop"`. If `finishReason` is `undefined` (which can happen with certain gateway/proxy configurations), accessing `result.output` throws `NoOutputGeneratedError` even when valid JSON is present in `result.text`.

### AI SDK Version

6.0.0-beta.166

### Reproduction

```typescript
import { z } from 'zod';

const schema = z.object({
  items: z.array(z.object({
    id: z.string(),
    value: z.boolean()
  }))
});

  model: someModel, // Using a gateway/proxy that may not return finishReason
  messages: [{ role: 'user', content: 'Generate items' }],
  output: Output.object({ schema })
});

// This throws NoOutputGeneratedError even when valid JSON is in result.text
const output = result.output;
```

### Expected Behavior

If `result.text` contains valid JSON that matches the schema, `result.output` should return the parsed object regardless of `finishReason`.

### Actual Behavior

`NoOutputGeneratedError` is thrown because `finishReason` is `undefined` instead of `"stop"`.

**Logs showing the issue:**
```
finishReasonType: "undefined"
resultTextSlice: "{\"relevantExceptions\":[{\"id\":\"..."  // Valid JSON present
```

### Root Cause Analysis

In `dist/index.mjs`:

**Line 3880** - Output only parsed when `finishReason === "stop"`:
```javascript
if (lastStep.finishReason === "stop") {
  const outputSpecification = output != null ? output : text();
  resolvedOutput = await outputSpecification.parseCompleteOutput(
    { text: lastStep.text },
    { response: lastStep.response, usage: lastStep.usage, finishReason: lastStep.finishReason }
  );
}
```

**Lines 3997-4001** - Getter throws if `_output` is null:
```javascript
get output() {
  if (this._output == null) {
    throw new NoOutputGeneratedError();
  }
  return this._output;
}
```

### Suggested Fix

Consider parsing the output when:
- `finishReason === "stop"`, OR
- `finishReason` is `undefined` AND `text` is non-empty (fallback for gateways that don't return finishReason)

```javascript
if (lastStep.finishReason === "stop" || (lastStep.finishReason === undefined && lastStep.text)) {
  // parse output...
}
```

### Workaround

Currently using a try/catch fallback to manually parse `result.text`:

```typescript
let output;
try {
  output = result.output;
} catch (error) {
  if (NoOutputGeneratedError.isInstance(error) && result.text) {
    output = schema.parse(JSON.parse(result.text));
  } else {
    throw error;
  }
}
```

### Environment

- Model: Google Gemini 3.0 Flash (via AI Gateway)
- AI Gateway returns `finishReason: undefined` but valid JSON in response text

gr2m:
## Background

generateText with Output.object() could leave result.output unset when a provider or gateway returned valid JSON text but omitted the unified finishReason, causing NoOutputGeneratedError for usable structured responses.

## Summary

Updated generateText to parse complete output when the final step either finishes with stop or has a missing finishReason with non-empty text.

## Testing

Added a regression test covering Output.object() with valid JSON text and an omitted unified finishReason.

## Related Issues

Fixes #11348

Co-authored-by: pheuter <92698+pheuter@users.noreply.github.com>


lgrammel:
want to try new agent version

Issue #16767, opened 2026-07-06
Backport tracking: bug fixes and docs improvements from main (v7) not yet on release-v6.0


gr2m:
## Goal

Track backporting of bug fixes and documentation improvements that were merged into `main` (AI SDK 7) but are not yet on `release-v6.0` (AI SDK 6).

**Method:** Compared all 375 fix/docs commits on `main` since the branches diverged (b145dce58e, 2026-03-05) against `release-v6.0` (matching backport commit titles and PR references), then assessed each remaining candidate's applicability by reading the diff and checking whether the affected code/docs — and the bug — exist on `release-v6.0`.

**Snapshot (2026-07-06):** 158 already backported · 4 in-flight · 213 candidates assessed below.

## ✅ Recommended for backport

These fix problems that exist on `release-v6.0` and apply without v7-only dependencies.

### Bug fixes

- [x] #13137 — fix(security): prevent streaming tool calls from finalizing on parsable partial JSON
  - Security fix (VULN-6861): v6's openai, openai-compatible, groq, deepseek, and alibaba streaming code still finalizes tool calls on parsable partial JSON via isParsableJson, and no backport exists on release-v6.0.
- [x] #13372 — fix(provider/xai): extract reasoning text from content in responses doGenerate
  - v6 xai responses doGenerate still reads reasoning only from part.summary and its schema lacks the content field, so reasoning returned via content is silently dropped in generateText; fix applies with minor adaptation to v6's evolved condition.
- [x] #13605 — fix(ai): include tool input on tool result for provider executed dynamic tools
  - v6 run-tools-transformation.ts breaks before toolInputs.set for provider-executed dynamic tools, so their tool results get input undefined; v6 already has the toolCallsByToolCallId map the fix uses.
- [x] #13961 — fix(ui): export `isDynamicToolUIPart` from `ai` package
  - v6 also defines isDynamicToolUIPart in ui-messages.ts but omits it from the ui/index.ts export list, so the same one-line export fix applies directly.
- [x] #14656 — fix: more precise default message for tool execution denial
  - v6 still uses the 'Tool execution denied.' default message that caused GPT-5 models to refuse further calls to the same tool; the one-word message fix applies across the same converter files (plus two extra v6 google files with the same string).
- [x] #14760 — fix(openai-compatible): buffer tool call deltas until function.name arrives
[...]

PR #16792, opened 2026-07-06
[v6.0] fix(gateway): accept deprecated warnings in image, speech, transcription, and video responses

gr2m:
Backport of #16030 to `release-v6.0` (clean cherry-pick). v6 gateway warning schemas lack the `deprecated` variant, so the shared gateway backend emitting deprecated warnings would fail response parsing.

Part of the v6.0 backport tracking issue #16767.

### DIFF
--- a/.changeset/quick-eagles-explain.md
+++ b/.changeset/quick-eagles-explain.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+fix(gateway): accept deprecated warnings in image, speech, transcription, and video responses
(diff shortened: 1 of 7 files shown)


gr2m:
Follow-up commit: v6's `SharedV3Warning` has no `deprecated` variant (unlike v4 on main), so the backport now accepts `deprecated` warnings in the response schemas (the original parse-failure bug) and maps them to `other` warnings before returning. Also removed the pre-existing `as Array<SharedV3Warning>` casts in the speech/transcription models that would have hidden the mismatch. Gateway suite passes locally (421/421).

PR #16794, opened 2026-07-06
[v6.0] fix: MCP callTool hangs and leaks its response handler when an in-flight request is aborted

gr2m:
Backport of #16595 to `release-v6.0` (clean cherry-pick). v6 `mcp-client.ts` has the identical pattern (abort only checked inside the response handler, no abort listener/cleanup). Includes the regression test with a hanging transport.

Part of the v6.0 backport tracking issue #16767.

### DIFF
--- a/.changeset/clever-mice-wait.md
+++ b/.changeset/clever-mice-wait.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/mcp': patch
+---
+
+Reject in-flight MCP requests when their abort signal fires and remove the pending response handler.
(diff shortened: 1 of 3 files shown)


gr2m:
Follow-up commit: adapted the new regression test to v6's `callTool` signature (`{ name, args, options }` with `options.abortSignal`; the method isn't on v6's public `MCPClient` interface, so the test uses a structural cast). The production fix itself applied cleanly. MCP suite passes locally (228/228).

PR #13137, opened 2026-03-05
fix(security): prevent streaming tool calls from finalizing on parsable partial JSON

gr2m:
## Background

In 5 OpenAI-compatible providers (`openai`, `openai-compatible`, `groq`, `deepseek`, `alibaba`), streaming tool call arguments were finalized using `isParsableJson()` as a heuristic for completion. If partial accumulated JSON happened to be valid JSON before all chunks arrived, the tool call would be executed with incomplete arguments, potentially dropping safety-relevant parameters.

For example, a model streaming `{"query": "test", "limit": 10}` might produce `{"query": "test"}` as an intermediate value — which is valid JSON. The tool call would be finalized early with only `{"query": "test"}`, and the `"limit": 10` parameter would be silently lost.

Providers not affected: `anthropic` (uses explicit `content_block_stop` events), OpenAI Responses API (different protocol).

## Summary

- Removed all inline `isParsableJson()` checks from the streaming tool call delta handlers across 5 providers
- Moved tool call finalization to the `flush()` handler, which only runs after the stream is fully consumed, ensuring tool calls always receive complete arguments
- Added a regression test demonstrating the vulnerability (first commit fails against unfixed code, second commit fixes it)
- Updated existing test snapshots to reflect the new ordering (tool calls finalize in `flush()` after text-end, rather than inline during streaming)

## Manual Verification

The first commit (`573a744`) adds the regression test only — it fails against the unfixed code, proving the vulnerability exists:
```
- Expected: input: '{"query": "test"}, "limit": 10}'
+ Received: input: '{"query": "test"}'
```

The second commit (`7ae73dd`) applies the fix, making all tests pass.

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Related Issues

VULN-6861

PR #16800, opened 2026-07-06
[v6.0] fix: prevent streaming tool calls from finalizing on parsable partial JSON

gr2m:
Backport of #13137 (security fix, VULN-6861) to `release-v6.0`. Streaming tool call arguments in the `openai`, `openai-compatible`, `groq`, `deepseek`, and `alibaba` providers were finalized as soon as the accumulated argument text happened to be parsable JSON, which could execute tool calls with incomplete arguments and silently drop trailing parameters; finalization now happens exclusively in the stream `flush()` handler after all chunks have arrived. The only conflict was in the import block of `packages/deepseek/src/chat/deepseek-chat-language-model.ts`: v6 already imports `type InferSchema`, so the resolution simply removes the now-unused `isParsableJson` import without duplicating the `InferSchema` import from main's import style; all inline early-finalization blocks were removed exactly as in the original commit, and v6's existing `flush()` loop over unfinished tool calls provides the finalization. The regression test, snapshot updates, and changeset from the original commit are kept. Part of the v6.0 backport tracking issue #16767.

### DIFF
--- a/.changeset/fix-streaming-tool-call-early-finalization.md
+++ b/.changeset/fix-streaming-tool-call-early-finalization.md
@@ -1,0 +1,11 @@
+---
+'@ai-sdk/openai': patch
+'@ai-sdk/openai-compatible': patch
+'@ai-sdk/groq': patch
+'@ai-sdk/deepseek': patch
+'@ai-sdk/alibaba': patch
+---
+
+fix(security): prevent streaming tool calls from finalizing on parsable partial JSON
+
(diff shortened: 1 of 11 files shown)


gr2m:
Follow-up commit: adapted the cerebras mixed structured-output streaming test. With early `isParsableJson` finalization removed, tool calls finalize during flush (after the structured-output text has streamed), so cerebras's json-mode filter now drops the spurious repeated tool call entirely — matching `doGenerate`'s treatment of mixed responses. All other openai-compatible-dependent packages (baseten, deepinfra, fireworks, huggingface, moonshotai, vercel, togetherai, xai, google-vertex) pass unchanged.

PR #16857, opened 2026-07-07
fix: smoothStream drops providerMetadata from chunked stream deltas

### LINKED ISSUE #14373: [Bug] smoothStream drops providerMetadata from chunked stream parts

lgrammel:
## Background

smoothStream could split provider metadata-bearing text or reasoning deltas and emit intermediate chunks without that metadata, losing data such as Anthropic thinking signatures.

## Summary

Updated smoothStream to include preserved providerMetadata when enqueueing chunks from the chunking loop.

## Testing

Kept the regression test that verifies word-chunked reasoning-delta parts all preserve providerMetadata, and removed the reproduction-only example artifact.

## Related Issues

Fixes #14373

Co-authored-by: kuishou68 <54054995+kuishou68@users.noreply.github.com>


lgrammel:
Concerned that this is a bad fix; there is a reason provider metadata was only emitted on flush. now it might be duplicated.

PR #16867, opened 2026-07-07
fix: createMCPClient hangs on unterminated MCP HTTP SSE JSON-RPC responses

### LINKED ISSUE #16865: createMCPClient hangs on SSE responses that omit the frame terminator (\n\n) and keep the connection open
### Summary
`createMCPClient` with HTTP transport hangs indefinitely when an MCP server returns `text/event-stream` responses that contain a valid data: JSON-RPC payload but **omit the SSE frame terminator (\n\n)** and **keep the connection open**.

Affects streamable HTTP POST responses with `Content-Type: text/event-stream` (e.g. `initialize`).

This is separate from #10915 / #16002 (missing event: message), which is fixed in 1.0.59+. It still hangs on @ai-sdk/mcp@1.0.59 when the server omits the SSE frame terminator (\n\n) and keeps the connection open.

### Example
Productboard MCP (https://mcp.productboard.com, HTTP transport, OAuth) returns SSE bodies like:

`data: {"jsonrpc":"2.0","id":1,"result":{...}}`

- No trailing blank line (\n\n) to terminate the frame
- Connection stays open
- `createMCPClient` never completes initialize and hangs until an external timeout

Other clients (e.g. Cursor, official @modelcontextprotocol/sdk) handle this server successfully.

### Root Cause

`HttpMCPTransport` pipes responses through `EventSourceParserStream` (eventsource-parser). The parser only dispatches a `ParsedEvent` after a blank line ends the frame:

```ts
// eventsource-parser: dispatch happens on empty line only
if (lineLength === 0) {
  if (data.length > 0) {
    onParse({ type: 'event', event: eventName || undefined, data: ... })
  }
}
```

If the server sends `data: {...}\n` without `\n\n` and keeps the stream open, the parser never emits, and `processEvents()` waits forever on `reader.read()`

### Reproduction
```ts
import http from "node:http"
import { createMCPClient } from "@ai-sdk/mcp"

const body =
  'data: {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05","capabilities":{},"serverInfo":{"name":"test","version":"1.0"}}}\n'

const server = http.createServer((_req, res) => {
  res.writeHead(200, {
    "Content-Type": "text/event-stream",
    "Cache-Control": "no-cache",
    Connection: "keep-alive",
  })
  res.write(body)
  // Note: no res.end(), no trailing \n\n to mimic Productboard
})

server.listen(0, "127.0.0.1", async () => {
  const { port } = server.address()
  try {
    const client = await Promise.race([
      createMCPClient({ transport: { type: "http", url: `http://127.0.0.1:${port}` } }),
      new Promise((_, reject) => setTimeout(() => reject(new Error("timeout")), 5000)),
    ])
    await client.close()
    console.log("success")
  } catch (e) {
    console.error("failed:", e) // times out
  } finally {
    server.close()
  }
})
```
Expected: createMCPClient completes initialize.
Actual: Hangs until timeout.
Verified on @ai-sdk/mcp@1.0.46 and @ai-sdk/mcp@1.0.59.

### Workaround
Custom fetch wrapper on `createMCPClient` that:

- Reads the SSE body until a complete data: JSON line is received
- Appends \n\n to terminate the frame
- Returns a synthetic Response to the SDK

### Suggested Fix
In the MCP HTTP transport’s SSE handling, dispatch events when a data: field contains valid JSON-RPC, without requiring the \n\n frame terminator. This matches how more lenient MCP clients behave when servers send data: frames on a kept-alive stream.

### AI SDK Version

@ai-sdk/mcp: 1.0.59
ai: 6.0.3
@ai-sdk/provider: 3.0.13

### Code of Conduct

lgrammel:
## Background

HTTP MCP initialization could hang when an SSE response sent a valid JSON-RPC data line but kept the stream open without a blank-line frame terminator.

## Summary

Added an MCP HTTP SSE normalization step that synthetically terminates complete JSON-RPC data lines before EventSource parsing.

## Testing

Added a regression test for createMCPClient initialization from a text/event-stream POST response with only a single newline after the JSON-RPC data line.

## Related Issues

Fixes #16865

Co-authored-by: samkadambi56 <240511691+samkadambi56@users.noreply.github.com>
Co-authored-by: Aayush-engineer <121757402+Aayush-engineer@users.noreply.github.com>


Aayush-engineer:
Reviewed and tested locally (ran the regression test independently on bug-16865-1 — passes). The approach looks solid: createJsonRpcSseFrameTerminatorStream() only injects a synthetic terminator once a data: line's value parses as valid JSON via parseJSONRPCMessage; everything else passes through untouched, so normal well-terminated SSE traffic should be unaffected.
One thing I wanted to sanity-check (not sure it's a real issue): SSE allows a single event's data to span multiple data: lines that get concatenated before parsing. Since termination here happens per-line rather than after concatenation, I tried to think of a case where a server sends a first data: line that's independently valid JSON but intends a follow-up line to be appended — but that seems like an unusual pattern for JSON-RPC framing and I don't have a concrete repro for it. Flagging mostly to see if this was already considered, not as a blocker.
Nice catch also fixing the previously-swallowed errors from the floating void openInboundSse() calls.

lgrammel:
Replying to PR comment 4906418624:

Thanks for the careful sanity-check. I added a regression test covering the valid multi-line SSE case where one JSON-RPC message is split across multiple data: lines and terminated normally. No production change was needed: split JSON-RPC data lines do not parse independently, so the normal SSE blank-line dispatch path is preserved. A first data: line that is already a complete JSON-RPC message plus a follow-up data: line would concatenate into a different/non-JSON-RPC event, so I kept the bugfix scoped to complete JSON-RPC data lines.

lgrammel:
Replying to PR comment 4906418624:

Thanks for the careful sanity-check. I kept the split-data-line regression coverage and tightened the normalizer so it only injects a synthetic blank line when the complete JSON-RPC data line is the last line currently available. If another data/field line or the real blank-line terminator is already buffered, the EventSource parser now preserves the original SSE frame boundaries. The openInboundSse() error handling remains in place.

lgrammel:
Replying to PR comment 4912279318:

I verified the existing multi-line regression and made one small production refinement beyond it: the complete-JSON-RPC-first-data-line-plus-follow-up-line case is now preserved as one normally parsed SSE frame when the continuation is already buffered, rather than being split by the synthetic terminator.

lgrammel:
Closing this for all full re-attempt starting with reproduction.

PR #16882, opened 2026-07-07
fix(harness-claude-code): avoid bridge hello listener race

### LINKED ISSUE #16878: @ai-sdk/harness-claude-code can miss immediate bridge-hello due to WebSocket listener race
# `@ai-sdk/harness-claude-code` can miss immediate `bridge-hello` due to WebSocket listener race

## Summary

`@ai-sdk/harness-claude-code@1.0.19` can intermittently hang during `HarnessAgent.createSession()` because the host-side WebSocket startup path can miss an immediately-sent `bridge-hello` frame.

The bridge sends `bridge-hello` as soon as it accepts the WebSocket connection. The host currently opens the WebSocket first and only then attaches the `message` listener that waits for `bridge-hello`. In fast bridge/proxy paths, the hello frame can arrive before that listener is attached. The host then waits for the hello timeout, closes the socket, retries, and may repeat until `createSession()` times out.

## Observed Behavior

In an isolated benchmark using `HarnessAgent`, `createClaudeCode`, and a diagnostic WebSocket proxy between the host and bridge URL:

- Several attempts timed out in `HarnessAgent.createSession()`.
- Failed attempts showed the bridge sent `bridge-hello` immediately.
- The host sent zero client messages on those failed connections.
- The client closed the socket after roughly the hello wait timeout, then retried.
- The retry loop repeated until the outer `createSession()` timeout fired.

When the diagnostic proxy delayed only the first server-to-client frame (`bridge-hello`) by 25ms:

- The same benchmark succeeded consistently.
- Each attempt used a single bridge WebSocket connection.
- Each attempt sent `start` shortly after client open.
- The no-client-message close/retry loop disappeared.

This strongly suggests the host is sometimes missing the immediate `bridge-hello` frame because the listener is attached too late.

## Current Implementation Shape

The startup flow is effectively:

```ts
ws = await openWebSocket(...);
await waitForBridgeHello({ ws, ... });
```

`waitForBridgeHello` attaches the `message` listener after `openWebSocket` resolves. If `bridge-hello` is emitted between those two steps, the wait can time out even though the bridge was reachable and already sent the expected frame.

## Expected Behavior

The host should not be able to miss `bridge-hello`, even if the bridge sends it immediately after accepting the WebSocket connection.

## Suggested Fix

Create/open the WebSocket and attach the `message`, `close`, and `error` listeners immediately, before awaiting or resolving the open step.

The startup helper should resolve only after both conditions are true:

1. The WebSocket `open` event has fired.
2. A `bridge-hello` message has been observed.

It should preserve the existing timeout, retry, close, terminate, and error message behavior.

## Impact

This causes random `HarnessAgent.createSession()` latency spikes and timeouts. The bridge may be healthy and reachable, but the host can still wait for a hello frame that already arrived before the listener was installed.

jdanielnd:
## Summary

Fixes an intermittent `HarnessAgent.createSession()` timeout in `@ai-sdk/harness-claude-code` where the host could miss an immediately-sent `bridge-hello` frame.

The previous startup path awaited the WebSocket `open` event before attaching the `message` listener that waits for `bridge-hello`. Since the bridge sends `bridge-hello` as soon as it accepts the connection, fast paths could deliver that frame before the listener existed. The host would then wait for the hello timeout, close, retry, and sometimes exhaust the outer startup timeout.

This change constructs the WebSocket and attaches `open`, `message`, `close`, and `error` listeners immediately, then resolves only after both the socket has opened and `bridge-hello` has been observed. It also switches the hello-frame parse path to `safeParseJSON` and adds regression coverage for immediate hello, missing hello, pre-hello close, and open timeout.

Closes #16878

## Validation

- `pnpm --filter @ai-sdk/harness-claude-code test`
- `pnpm --filter @ai-sdk/harness-claude-code type-check`
- `pnpm --filter @ai-sdk/harness-claude-code build`
- `pnpm check`

Note: `pnpm type-check:full` was also run locally and failed on generated Nuxt output unrelated to this change: `examples/nuxt-openai/.nuxt/types/plugins.d.ts(24,3): error TS1110: Type expected.`


rraina:
Thanks for sharing - we are also seeing the same issue

packages/harness-claude-code/src/claude-code-harness.ts
```diff
@@ -869,23 +912,15 @@ async function openBridgeWebSocket({
...
-      ws = await openWebSocket({
+      return await openWebSocketAndWaitForBridgeHello({
         url: wsUrl,
-        timeoutMs: Math.min(10_000, remaining),
-      });
-      await waitForBridgeHello({
-        ws,
-        timeoutMs: Math.min(5_000, Math.max(1, deadline - Date.now())),
+        openTimeoutMs: Math.min(10_000, remaining),
+        helloTimeoutMs: Math.min(5_000, Math.max(1, deadline - Date.now())),
```
felixarntz: There's one concern here: `helloTimeoutMs` is calculated before awaiting open and then starts only after open, so time spent opening the socket is not deducted from the hello timeout.

This means the externally controllable timeout value may not be respected.

felixarntz (changes requested):
@jdanielnd Thank you for the PR. Mostly LGTM, but one problem needs to be addressed.

packages/harness-claude-code/src/claude-code-harness.ts
jdanielnd: @felixarntz good catch. I updated this so the bridge-hello timeout is computed when the socket opens, using the remaining startup deadline at that point instead of the value from before waiting for `open`.

I also added regression coverage where most of the startup budget has elapsed by the time `open` fires, so the hello wait is capped to only the remaining budget.

jdanielnd:
@felixarntz is this fix still being considered for merge? happy to address anything else needed, thanks!

felixarntz (approved):
@jdanielnd Apologies for the delay, looks all great now!

.changeset/quiet-bridges-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Fix Claude Code bridge WebSocket startup so an immediate `bridge-hello` frame cannot be missed before the listener is attached.
```
felixarntz:
```suggestion
fix(harness-claude-code): fix bridge WebSocket startup so an immediate `bridge-hello` frame cannot be missed before the listener is attached
```

PR #13648, opened 2026-03-19
feat: migrate providers to support new top-level `reasoning` parameter

felixarntz:
## Background

The new top-level `reasoning` parameter was added to the AI SDK spec and core in #13553. Providers need to be migrated to translate this parameter into their native reasoning/thinking configuration.

## Summary

Migrates 7 providers to support the new top-level `reasoning` parameter:

- **deepseek**: Maps `reasoning` to existing thinking support (already had infrastructure, just needed wiring)
- **groq**: Maps `reasoning` to `reasoning_effort` in provider options
- **xai**: Maps `reasoning` to `reasoning_effort` for both chat and responses models
- **openai-compatible**: Maps `reasoning` to `reasoning_effort` in provider options
- **open-responses**: Maps `reasoning` to `reasoning.effort` in the responses format
- **alibaba**: Maps `reasoning` to `enable_thinking` + `thinking_budget` via token budget calculation
- **cohere**: Maps `reasoning` to `thinking.type` + `thinking.token_budget` via token budget calculation
- **fireworks**: Uses **openai-compatible**

Together with #13649, this completes the work on #12516.

## Manual Verification

Relevant examples were updated or added, then run for verification.

## Checklist

- [x] Tests have been added / updated (for bug fixes / features)
- [ ] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

## Future Work

N/A

## Related Issues

See #12516.

PR #15276, opened 2026-05-13
feat(xai): support 'none' and 'medium' reasoning effort, curate model IDs

Jaaneek:
## Background

xAI's `grok-4.3` accepts a `reasoning_effort` of `"none" | "low" | "medium" | "high"` ([docs](https://docs.x.ai/docs/guides/reasoning)), where:

- `"none"` disables reasoning entirely (no thinking tokens — useful for simple use cases requiring near-instant responses).
- `"medium"` provides more thinking for less-latency-sensitive applications.

Today the `@ai-sdk/xai` Zod schema for `providerOptions.xai.reasoningEffort` only accepts `'low' | 'high'` on the chat model and `'low' | 'medium' | 'high'` on the responses model, so `'none'` is rejected by validation before the request ever hits xAI, and `'medium'` is missing entirely on the chat model.

autocomplete unions had drifted from xAI's [current model lineup](https://docs.x.ai/docs/models). Many of the listed models are being [retired on May 15, 2026](https://docs.x.ai/developers/migration/may-15-retirement) (`grok-4-1-fast`, `grok-4-fast`, `grok-4`, `grok-code-fast-1`), the `grok-3*` family is no longer in the current lineup, and the current flagship `grok-4.3` was missing.

## Summary

**Reasoning effort:**

- Add `'none'` to the allowed values for `providerOptions.xai.reasoningEffort`
- Add `'medium'` to the chat model's `reasoningEffort` enum (the responses model already supported it).
- Top-level `reasoning: 'medium'` now maps to `reasoning_effort: 'medium'` for the chat model (previously coerced to `'low'` because `'medium'` was not a valid value).

**Model ID curation:**

 autocomplete unions to xAI's current lineup:

- `grok-4.20-0309-non-reasoning`
- `grok-4.20-0309-reasoning`
- `grok-4.3`
- `grok-latest`

Removed entries: `grok-3*`, `grok-4`, `grok-4-0709`, `grok-4-latest`, `grok-4-1-fast-*`, `grok-4-fast-*`, `grok-code-fast-1`, `grok-4.20-multi-agent-0309`.

### Usage

```ts
import { xai } from '@ai-sdk/xai';
import { generateText } from 'ai';

await generateText({
  model: xai('grok-4.3'),
  prompt: 'Hi',
  providerOptions: {
    xai: { reasoningEffort: 'none' },
  },
});
```

## Manual Verification

- `pnpm --filter @ai-sdk/xai test:node` → 316/316 passing locally. 

## Checklist

[...]

PR #16893, opened 2026-07-07
fix(provider/xai): send reasoning effort "none" for top-level `reasoning: 'none'`

### LINKED ISSUE #16892: xai: top-level `reasoning: 'none'` is silently dropped instead of sending `reasoning.effort: "none"`
## Description

When using the top-level `reasoning` parameter with `'none'` on an xAI model that supports disabling reasoning (e.g. `grok-4.3`), the provider silently drops the setting instead of sending it to the xAI API.

```ts
await generateText({
  model: xai('grok-4.3'),
  prompt: '...',
  reasoning: 'none',
});
```

**Expected request body** (per the [xAI reasoning docs](https://docs.x.ai/developers/model-capabilities/text/reasoning)):

```json
{ "model": "grok-4.3", "input": [...], "reasoning": { "effort": "none" } }
```

**Actual request body** — no `reasoning` field at all:

```json
{ "model": "grok-4.3", "input": [...] }
```

Since xAI defaults to `effort: "low"` when the parameter is omitted, the model still reasons: the response contains reasoning tokens even though the user explicitly requested `reasoning: 'none'`. No warning is emitted either — the setting is silently ignored.

## Root cause

Both xAI language models explicitly map top-level `'none'` to `undefined`:

- `packages/xai/src/responses/xai-responses-language-model.ts` (lines ~167–183)
- `packages/xai/src/xai-chat-language-model.ts` (lines ~156–172)

```ts
const resolvedReasoningEffort =
  options.reasoningEffort ??
  (isCustomReasoning(reasoning)
    ? reasoning === 'none'
      ? undefined // <-- drops 'none' instead of passing it through
      : mapReasoningToProviderEffort({ ... })
    : undefined);
```

This behavior is even encoded in a test (`xai-chat-language-model.test.ts`: "should not set reasoning_effort for top-level reasoning none"), introduced in #13648 — presumably before xAI supported `none`. It is inconsistent with:

- The xAI API itself: `grok-4.3` supports `effort: "none"` (disables reasoning entirely).
- The provider's own `providerOptions` path, which already accepts and forwards `'none'`: `providerOptions: { xai: { reasoningEffort: 'none' } }` correctly sends `reasoning.effort: "none"`.
- The OpenAI provider, which passes top-level `'none'` straight through as `effort: 'none'`.

## Fix

Map top-level `reasoning: 'none'` to `'none'` in both the responses and chat models (chat completions API uses `reasoning_effort: "none"`), and update the test accordingly.

## Workaround

```ts
providerOptions: { xai: { reasoningEffort: 'none' } }
```

## AI SDK Version

- `@ai-sdk/xai` (main branch, `ai` v6)

gr2m:
## Background

The xAI API supports disabling reasoning on `grok-4.3` and newer reasoning models by sending `reasoning: { effort: "none" }` (Responses API) / `reasoning_effort: "none"` (Chat Completions API). When the parameter is omitted, xAI defaults to `effort: "low"`, so the model still reasons.

The xAI provider explicitly mapped the top-level `reasoning: 'none'` call option to `undefined`, silently dropping it from the request — without even emitting a warning. Users who requested `reasoning: 'none'` still got (and paid for) reasoning tokens. The `providerOptions: { xai: { reasoningEffort: 'none' } }` escape hatch already forwarded `'none'` correctly, and the OpenAI provider passes top-level `'none'` straight through, so this was an inconsistency introduced in #13648, presumably before xAI supported `none`.

Additionally, probing the live API revealed that `grok-4.20-reasoning` and `grok-4.20-non-reasoning` (including dated variants like `grok-4.20-0309-reasoning`) reject the reasoning effort parameter entirely — **any** value, not just `'none'` — with `400 invalid-argument: Model X does not support parameter reasoningEffort.` So the top-level `reasoning` option already caused request failures on those models before this PR.

## Summary

- Map the top-level `reasoning: 'none'` option to effort `'none'` in both the Responses API model (`xai-responses-language-model.ts`) and the Chat Completions model (`xai-chat-language-model.ts`), instead of dropping it.
- Omit the reasoning effort parameter and emit an `unsupported` warning when the top-level `reasoning` option is used with models that reject the parameter (`supports-reasoning-effort.ts`). Explicit `providerOptions.xai.reasoningEffort` is still passed through verbatim.
- Updated the chat model test that locked in the old behavior, and added top-level `reasoning` coverage for the Responses API model (mapping, `'none'`, `providerOptions` precedence, and unsupported-model gating).
- Added patch changesets for `@ai-sdk/xai`.

## Manual Verification

Ran the following against the live xAI API (via `examples/ai-functions`), logging the request body with a custom `fetch`:

```ts
import { createXai } from '@ai-sdk/xai';

const xai = createXai({
  fetch: async (url, init) => {
    console.dir(JSON.parse(init.body), { depth: Infinity });
    return fetch(url, init);
  },
});

  model: xai('grok-4.3'),
  prompt: `How many "r"s are in the word "strawberry", and what is the square root of 144? Then, how much is the product of both of the resulting values? Think hard about it. Only respond with the resulting final number, nothing more.`,
  reasoning: 'none',
});

console.log(result.finalStep.reasoning);
console.log(result.text);
```

**Before the fix**: the request body contained no `reasoning` field, and `result.finalStep.reasoning` contained reasoning content (xAI defaults to `effort: "low"`).

**After the fix**: the request body contains `reasoning: { effort: 'none' }` and the response contains no reasoning content.

Also verified against the live API with `reasoning: 'none'` on models that reject the parameter:

- `grok-4.3` → `reasoning: { effort: 'none' }` sent, no reasoning tokens, no warnings.
- `grok-4.20-reasoning` → parameter omitted, request succeeds (would be a 400 otherwise), `unsupported` warning surfaced.
- `grok-4.20-non-reasoning` → parameter omitted, request succeeds, `unsupported` warning surfaced.

## Checklist

## Related Issues

Fixes #16892

### DIFF
--- a/packages/xai/src/xai-chat-language-model.test.ts
+++ b/packages/xai/src/xai-chat-language-model.test.ts
@@ -1328,6 +1328,49 @@ describe('XaiChatLanguageModel', () => {
       );
     });
 
+    it('should omit reasoning_effort and warn for models that do not support it', async () => {
+      prepareJsonFixtureResponse('xai-text');
+
+      const modelWithoutReasoningEffort = new XaiChatLanguageModel(
+        'grok-4.20-reasoning',
+        testConfig,
+      );
+
+      const result = await modelWithoutReasoningEffort.doGenerate({
+        prompt: TEST_PROMPT,
+        reasoning: 'none',
+      });
+
+      expect(
+        (await server.calls[0].requestBodyJson).reasoning_effort,
+      ).toBeUndefined();
+      expect(result.warnings).toContainEqual({
+        type: 'unsupported',
+        feature: 'reasoning',
+        details: 'reasoning "none" is not supported by this model.',
+      });
+    });
+
+    it('should still pass providerOptions reasoningEffort for models that do not support top-level reasoning', async () => {
+      prepareJsonFixtureResponse('xai-text');
+
+      const modelWithoutReasoningEffort = new XaiChatLanguageModel(
+        'grok-4.20-reasoning',
+        testConfig,
+      );
+
+      await modelWithoutReasoningEffort.doGenerate({
+        prompt: TEST_PROMPT,
+        providerOptions: {
+          xai: { reasoningEffort: 'none' },
+        },
+      });
+
+      expect((await server.calls[0].requestBodyJson).reasoning_effort).toBe(
+        'none',
+      );
+    });
+
     it('should extract reasoning content', async () => {
       prepareJsonFixtureResponse('xai-text');
 
--- a/packages/xai/src/xai-chat-language-model.test.ts
+++ b/packages/xai/src/xai-chat-language-model.test.ts
@@ -1299,7 +1299,7 @@ describe('XaiChatLanguageModel', () => {
       );
     });
 
-    it('should not set reasoning_effort for top-level reasoning none', async () => {
+    it('should map top-level reasoning none to reasoning_effort: "none"', async () => {
       prepareJsonFixtureResponse('xai-text');
 
       await reasoningModel.doGenerate({
@@ -1307,9 +1307,9 @@ describe('XaiChatLanguageModel', () => {
         reasoning: 'none',
       });
 
-      expect(
-        (await server.calls[0].requestBodyJson).reasoning_effort,
-      ).toBeUndefined();
+      expect((await server.calls[0].requestBodyJson).reasoning_effort).toBe(
+        'none',
+      );
     });
 
     it('should prefer providerOptions reasoningEffort over top-level reasoning', async () => {
--- a/.changeset/violet-clouds-brake.md
+++ b/.changeset/violet-clouds-brake.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fix: send reasoning effort `none` to the xAI API when the top-level `reasoning: 'none'` option is set
--- a/.changeset/wild-planets-argue.md
+++ b/.changeset/wild-planets-argue.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fix: omit the reasoning effort parameter and emit an unsupported warning when the top-level `reasoning` option is used with xAI models that reject it (`grok-4.20-reasoning`, `grok-4.20-non-reasoning`, and dated variants)
--- a/packages/xai/src/responses/xai-responses-language-model.test.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.test.ts
@@ -893,6 +893,53 @@ describe('XaiResponsesLanguageModel', () => {
           const requestBody = await server.calls[0].requestBodyJson;
           expect(requestBody.reasoning.effort).toBe('high');
         });
+
+        it('should omit reasoning effort and warn for models that do not support it', async () => {
+          prepareJsonResponse({
+            id: 'resp_123',
+            object: 'response',
+            status: 'completed',
+            model: 'grok-4.20-reasoning',
+            output: [],
+            usage: { input_tokens: 10, output_tokens: 5 },
+          });
+
+          const result = await createModel('grok-4.20-reasoning').doGenerate({
+            prompt: TEST_PROMPT,
+            reasoning: 'none',
+          });
+
+          const requestBody = await server.calls[0].requestBodyJson;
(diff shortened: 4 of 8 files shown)


packages/xai/src/xai-chat-language-model.test.ts
```diff
@@ -1299,17 +1299,17 @@ describe('XaiChatLanguageModel', () => {
       );
     });
 
-    it('should not set reasoning_effort for top-level reasoning none', async () => {
```
gr2m, line 1302: investigating why we explicitly implemented it this way originally

gr2m: Traced the history — the test was **correct when written, and became stale two months later** when xAI added `"none"` support:

**March 20, 2026 — #13648 (`74d520f535`)** introduced this test along with the top-level `reasoning` mapping for 7 providers. At that time the xAI API only accepted `reasoning_effort: "low" | "high"` — the provider's `reasoningEffort` enum was just `['low', 'high']`, and the effort map even coerced `medium → 'low'`. `"none"` was not a value the xAI API accepted at all, so mapping top-level `reasoning: 'none'` to `undefined` (omit the parameter) was the only reasonable choice, and this test correctly locked in that intentional behavior.

**May 14, 2026 — #15276 (`80e1702260`)** is where the inconsistency crept in. When `grok-4.3` shipped with `reasoning_effort: "none" | "low" | "medium" | "high"`, that PR:

- added `'none'` and `'medium'` to the `providerOptions.xai.reasoningEffort` enums (so the escape hatch worked), and
- updated the **top-level** effort map for `medium` (`medium: 'low'` → `medium: 'medium'`) in both models,

but missed the `reasoning === 'none' ? undefined` special case sitting three lines above the effort map it was editing. That PR's own description states that `"none"` "disables reasoning entirely", and its usage example demonstrates the `providerOptions` path — top-level `reasoning: 'none'` just wasn't considered. This test kept passing, so nothing flagged that the once-correct behavior had become a silent drop.

So: intentional-and-correct in March (#13648) → xAI capability added in May (#15276) with a partial update that fixed `medium` but overlooked `none` → this PR completes what #15276 started.

Note: the same `reasoning !== 'none' → undefined` pattern still exists in **groq** and **deepseek**, where it remains correct today (neither API accepts a `"none"` effort value) — which is probably why the pattern looked normal and survived review.

gr2m: checking if sending `"reasoning": { "effort": "none" }` might cause problems with older models

gr2m:
Probed the live xAI API to check whether sending `reasoning: { effort: "none" }` causes problems on older/other models (the [docs](https://docs.x.ai/developers/model-capabilities/text/reasoning) don't specify error behavior). Results across the current lineup, both `/v1/responses` and `/v1/chat/completions`:

| Model | `effort: "none"` | `effort: "low"` / `"high"` |
| --- | --- | --- |
| `grok-4.3` | ✅ 200 | ✅ 200 |
| `grok-latest` | ✅ 200 | ✅ 200 |
| `grok-4.20-multi-agent` (+ `-0309`) | ✅ 200 | ✅ 200 (incl. `xhigh`) |
| `grok-4.20-reasoning` (+ `grok-4.20-0309-reasoning`) | ❌ 400 | ❌ 400 |
| `grok-4.20-non-reasoning` (+ `grok-4.20-0309-non-reasoning`) | ❌ 400 | ❌ 400 |

Error: `{"code":"invalid-argument","error":"Model grok-4.20-reasoning does not support parameter reasoningEffort."}`

Key takeaways:

- The `grok-4.20-(non-)reasoning` models reject **every** effort value, not just `"none"` — so top-level `reasoning: 'low'|'medium'|'high'` already produced 400s on those models before this PR. The `'none'` fix only added one more value to an existing failure mode.
- `grok-4.20-multi-agent` accepts `"none"` fine, despite the docs only listing `low`/`medium`/`high`/`xhigh` for it.

Handled in 3c2e686480: when the top-level `reasoning` option targets a model that rejects the parameter, the provider now omits it and emits an `unsupported` warning (same message shape as `mapReasoningToProviderEffort`) — matching how the OpenAI provider gates `reasoning` behind model capabilities. Explicit `providerOptions.xai.reasoningEffort` remains a verbatim pass-through. Verified end-to-end: `reasoning: 'none'` now succeeds on all five models, with warnings surfaced on the two that don't support the parameter.

PR #16900, opened 2026-07-08
fix(ai,policy-opa): use own-property checks when resolving per-tool approvals

dnukumamras:
## What

Tool sets, tool contexts, and per-tool approval maps are indexed by names that come from model output or client-supplied message history. Plain bracket access (`tools[name]`) resolves names that match inherited object properties (`constructor`, `toString`, `valueOf`, `__proto__`) to values on `Object.prototype` instead of `undefined`, which can slip past the "unknown tool" / "unconfigured approval" guards.

This routes those lookups through a new `getOwn` helper (own-property check via `Object.hasOwn`), builds the human-in-the-loop approval-matching maps with a null prototype, and adds the same own-property guards in policy-opa `wrapMcpTools` and `shadow`.

Behavior is identical for every real tool set, since tool names are always own properties. The only change is that an inherited-property name now reads as absent.

## Flow

Happy path (real tool named `weather`, unchanged):
1. Model or client references `weather`.
2. `getOwn(tools, "weather")` finds it as an own property and returns the tool.
3. Approval, validation, and execution proceed exactly as before.

Unhappy path (name collides with a prototype member, e.g. `constructor`):
1. Model or client references `constructor`, which is not a registered tool.
2. Before: `tools["constructor"]` returned `Object.prototype.constructor` (a function), passed the `!= null` guard, and could be treated as a tool or invoked as an approval callback.
3. After: `getOwn(tools, "constructor")` returns `undefined`, so the name routes to the safe "no such tool" / "unconfigured, apply fallback" branch (`NoSuchToolError`, denied/user-approval fallback, etc.).

## Notes

Covers the approval path (per-tool resolution and replay re-validation), tool-call parsing, execution, streaming callbacks, and UI message conversion/validation. Includes a `getOwn` unit test plus regression tests in `wrap-mcp-tools`, `shadow`, `resolve-tool-approval`, and `collect-tool-approvals`.

### DIFF
--- a/.changeset/tool-approval-inherited-property-names.md
+++ b/.changeset/tool-approval-inherited-property-names.md
@@ -1,0 +1,9 @@
+---
+'@ai-sdk/policy-opa': patch
+'ai': patch
+---
+
+Use own-property checks when resolving per-tool approvals so tool names and approval ids that match inherited object properties (e.g. `constructor`, `toString`, `valueOf`, `__proto__`) are treated as unconfigured/absent.
+
+- `@ai-sdk/policy-opa`: `wrapMcpTools` builds its per-tool map with a null prototype and reads supplied approvals via an own-property check, and `shadow` guards its per-tool map lookup the same way.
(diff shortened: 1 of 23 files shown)


gr2m (approved):
Verified the fix against VULN-13170. Own-property lookups (`Object.hasOwn` / `hasOwnProperty` + null-prototype maps and the shared `getOwn` helper) correctly close the inherited-property bypass across both `wrapMcpTools`/`shadow` and the core approval/tool-call/streaming/UI paths. Tool names such as `constructor`, `toString`, and `valueOf` now resolve to "unconfigured" and receive the configured fallback. Tests cover the regression and CI is green.

PR #16913, opened 2026-07-08
fix(google): apply the standalone threshold provider option to safety settings

### LINKED ISSUE #16912: @ai-sdk/google: standalone `threshold` provider option is documented but silently ignored
## Description

The `@ai-sdk/google` provider options schema declares a standalone `threshold` option (separate from the per-category `threshold` inside `safetySettings`), and the provider docs describe it as:

> **threshold** _string_
> Optional. Standalone threshold setting that can be used independently of `safetySettings`. Uses the same values as the `safetySettings` threshold.

However, nothing in `packages/google/src` ever reads this field — `getArgs()` in `google-language-model.ts` only forwards `googleOptions?.safetySettings`. So:

```ts
  prompt: '...',
  providerOptions: {
    google: {
      threshold: 'BLOCK_NONE', // validated, documented… and silently ignored
    },
  },
});
```

passes schema validation, matches the documented usage, and has zero effect on the request — no `safetySettings` are sent, so Google applies its defaults.

**Where:**
- Schema: `packages/google/src/google-language-model-options.ts` — standalone `threshold` enum (it is also the only option in that block with no doc comment).
- Request building: `packages/google/src/google-language-model.ts` — `safetySettings: googleOptions?.safetySettings` is the only safety-related mapping; grepping the package for `threshold` outside the options file yields nothing.

## Expected

Per the documentation, setting `threshold` without `safetySettings` should apply that threshold — the natural mapping is to expand it into `safetySettings` entries for the configurable harm categories (hate speech, dangerous content, harassment, sexually explicit), with an explicit `safetySettings` array taking precedence when both are provided.

(Alternatively, if the standalone option was never meant to exist, it should be removed from the schema and the docs — but since it is publicly documented, implementing it seems preferable.)

Same class of bug as #16894 (xai `imageDetail` accepted but not sent) and #16910 (deepgram transcription options accepted but not sent).

## AI SDK Version

- `@ai-sdk/google`: 4.0.9
- `ai`: 7.0.17

---

abhay-codes07:
## Background

The `@ai-sdk/google` options schema declares a standalone `threshold` option, and the provider docs describe it as a "standalone threshold setting that can be used independently of `safetySettings`" — but nothing in the package ever reads it. Setting `providerOptions.google.threshold` passes validation, matches the documented usage, and has zero effect: no `safetySettings` are sent, so Google applies its default thresholds.

It was also the only option in that schema block without a doc comment, which suggests it was wired into the schema but never into the request building.

## Summary

- `getArgs()` now expands the standalone `threshold` into `safetySettings` entries for the configurable harm categories (`HARM_CATEGORY_HATE_SPEECH`, `HARM_CATEGORY_DANGEROUS_CONTENT`, `HARM_CATEGORY_HARASSMENT`, `HARM_CATEGORY_SEXUALLY_EXPLICIT`) via a small `getSafetySettings` helper.
- An explicit `safetySettings` array takes precedence — the standalone `threshold` is only applied when `safetySettings` is not provided, matching the documented "used independently of `safetySettings`" semantics.
- `HARM_CATEGORY_CIVIC_INTEGRITY` is intentionally not included in the expansion since Google has deprecated configuring it; users who need it can still set it explicitly via `safetySettings`. Happy to include it (or to remove the option entirely instead) if maintainers prefer.
- Added a doc comment to the schema field describing the behavior.
- Added three tests: threshold expands into all four categories; explicit `safetySettings` wins over `threshold`; no `safetySettings` sent when neither is set. The expansion test fails on `main` (no `safetySettings` in the request body) and passes with this change.

## Manual Verification

Ran the new expansion test against `main`'s `google-language-model.ts` to confirm it reproduces the bug (request body contains no `safetySettings` despite `threshold` being set), then with the fix (passes). Full `@ai-sdk/google` suite passes (`pnpm test`: node + edge, 687 tests each) along with package `type-check`; `oxlint`/`oxfmt` clean on changed files.

## Checklist

## Related Issues

Fixes #16912

### DIFF
--- a/.changeset/google-standalone-threshold.md
+++ b/.changeset/google-standalone-threshold.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Fix the standalone `threshold` provider option being documented and accepted but silently ignored. It is now applied as the safety threshold for all configurable harm categories when no per-category `safetySettings` are provided.
--- a/packages/google/src/google-language-model-options.ts
+++ b/packages/google/src/google-language-model-options.ts
@@ -111,6 +111,12 @@ export const googleLanguageModelOptions = lazySchema(() =>
         )
         .optional(),
 
+      /**
+       * Optional. Standalone safety threshold that is applied to all
+       * configurable harm categories (hate speech, dangerous content,
+       * harassment, sexually explicit). Ignored when `safetySettings`
+       * is provided; use `safetySettings` for per-category control.
+       */
       threshold: z
         .enum([
           'HARM_BLOCK_THRESHOLD_UNSPECIFIED',
--- a/packages/google/src/google-language-model.test.ts
+++ b/packages/google/src/google-language-model.test.ts
@@ -592,6 +592,70 @@ describe('doGenerate', () => {
     });
   });
 
+  it('should expand the standalone threshold option into safetySettings for all configurable harm categories', async () => {
+    prepareJsonResponse({ content: 'test response' });
+
+    await model.doGenerate({
+      prompt: TEST_PROMPT,
+      providerOptions: {
+        google: {
+          threshold: 'BLOCK_ONLY_HIGH',
+        },
+      },
+    });
+
+    expect(await server.calls[0].requestBodyJson).toMatchObject({
+      safetySettings: [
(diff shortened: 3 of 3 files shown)


lgrammel:
Closing this PR because issue #16912 has already been closed.

PR #16915, opened 2026-07-08
fix(perplexity): throw on unsupported file media types instead of silently dropping them

### LINKED ISSUE #16914: @ai-sdk/perplexity: file parts with unsupported media types are silently dropped from the prompt
## Description

`convertToPerplexityMessages` in `@ai-sdk/perplexity` silently drops file parts it does not recognize, instead of throwing `UnsupportedFunctionalityError` like every other OpenAI-style provider converter (openai, mistral, groq, xai, alibaba, openai-compatible, huggingface all throw for unsupported file media types).

Two classes of prompt content vanish with no error and no warning:

1. **Any file that is not an image and not exactly `application/pdf`** — e.g. `audio/mpeg`, `video/mp4`, `text/csv`, `application/msword`. The `url`/`data` branch ends with a bare `return undefined`, which is then removed by `.filter(Boolean)`:

```ts
case 'url':
case 'data': {
  if (part.mediaType === 'application/pdf') { /* file_url */ }
  else if (getTopLevelMediaType(part.mediaType) === 'image') { /* image_url */ }
  return undefined; // ← silently dropped
}
```

2. **A real PDF passed with a top-level-only `application` media type.** The image branch normalizes top-level-only types via `getTopLevelMediaType`/`resolveFullMediaType`, but the PDF branch uses strict equality `part.mediaType === 'application/pdf'`, so `mediaType: 'application'` (with sniffable `%PDF` bytes) falls through both branches and is discarded.

The existing test `accepts top-level-only "application" mediaType for PDF without error` only asserts `.not.toThrow()` — it passes while the PDF is dropped and the resulting message content is empty.

## Reproduction

```ts
import { convertToPerplexityMessages } from '@ai-sdk/perplexity/internal';

const result = convertToPerplexityMessages([
  {
    role: 'user',
    content: [
      {
        type: 'file',
        mediaType: 'application', // top-level only; bytes are a real PDF
        data: { type: 'data', data: 'JVBERi0xLjQ=' },
        filename: 'doc.pdf',
      },
    ],
  },
]);

// result[0].content is "" — the PDF is gone, no error, no warning.
// Same for e.g. mediaType: 'audio/mpeg' — silently removed from the prompt.
```

## Expected

- Unsupported file media types throw `UnsupportedFunctionalityError` (matching openai/mistral/groq/xai/alibaba/openai-compatible/huggingface).
- A PDF with a top-level-only `application` media type and sniffable bytes is resolved via `resolveFullMediaType` and sent as a `file_url` part, the same way the image branch already normalizes top-level-only image types.

## AI SDK Version

- `@ai-sdk/perplexity`: 4.0.6
- `ai`: 7.0.17

---

abhay-codes07:
## Background

`convertToPerplexityMessages` silently drops file parts it does not recognize. Any file that is not an image and not exactly `application/pdf` (e.g. `audio/mpeg`, `text/csv`, `application/msword`) falls through to a bare `return undefined` that is then removed by `.filter(Boolean)` — no error, no warning, the content just vanishes from the prompt. A real PDF passed with a top-level-only `application` media type is also dropped, because the PDF branch uses strict `=== 'application/pdf'` equality while the image branch normalizes top-level-only types via `resolveFullMediaType`.

Every other OpenAI-style converter in the repo (openai, mistral, groq, xai, alibaba, openai-compatible, huggingface) throws `UnsupportedFunctionalityError` for unsupported file media types — perplexity was the lone outlier that silently discarded content.

## Summary

- PDFs are now detected via `resolveFullMediaType` when the media type is the top-level-only `application` form with inline bytes, so `mediaType: 'application'` + `%PDF` bytes produces a proper `file_url` part instead of being discarded.
- Unsupported file media types now throw `UnsupportedFunctionalityError` (`file part media type <mediaType>`), matching the sibling converters' behavior and wording.
- Strengthened the existing test that only asserted `.not.toThrow()` for the top-level-only `application` PDF case (it was passing while the PDF was silently dropped) to assert the emitted `file_url` content, and added two tests: unsupported media type throws, and top-level-only `application` with a URL source throws (subtype cannot be determined without bytes).

## Manual Verification

Ran the updated/new tests against `main`'s converter: all three fail there (the PDF-content assertion fails because the message content is empty, and both throw-assertions fail because nothing throws), and all pass with this change. Full `@ai-sdk/perplexity` suite passes (`pnpm test`: node + edge, 34 tests each) plus package `type-check`; `oxlint`/`oxfmt` clean on the changed files.

## Checklist

## Related Issues

Fixes #16914

### DIFF
--- a/.changeset/perplexity-silent-file-drop.md
+++ b/.changeset/perplexity-silent-file-drop.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/perplexity': patch
+---
+
+Fix file parts with unsupported media types being silently dropped from the prompt. Unsupported file media types now throw `UnsupportedFunctionalityError` like other providers, and PDFs passed with a top-level-only `application` media type are now detected from their bytes and sent as `file_url` parts instead of being discarded.
--- a/packages/perplexity/src/convert-to-perplexity-messages.test.ts
+++ b/packages/perplexity/src/convert-to-perplexity-messages.test.ts
@@ -172,9 +172,48 @@ describe('convertToPerplexityMessages', () => {
       });
     });
 
-    it('accepts top-level-only "application" mediaType for PDF without error', () => {
+    it('converts top-level-only "application" mediaType PDF bytes to a file_url part', () => {
       const pdfBase64 = 'JVBERi0xLjQ=';
 
+      const result = convertToPerplexityMessages([
+        {
+          role: 'user',
+          content: [
+            {
+              type: 'file',
+              mediaType: 'application',
+              data: { type: 'data' as const, data: pdfBase64 },
+              filename: 'doc.pdf',
+            },
+          ],
+        },
+      ]);
+
+      expect((result[0].content as unknown[])[0]).toEqual({
+        type: 'file_url',
+        file_url: { url: pdfBase64 },
+        file_name: 'doc.pdf',
+      });
+    });
+
+    it('throws for unsupported file media types instead of silently dropping them', () => {
+      expect(() =>
+        convertToPerplexityMessages([
+          {
+            role: 'user',
+            content: [
+              {
+                type: 'file',
+                mediaType: 'audio/mpeg',
+                data: { type: 'data' as const, data: 'AAAA' },
+              },
+            ],
+          },
+        ]),
+      ).toThrow(UnsupportedFunctionalityError);
+    });
+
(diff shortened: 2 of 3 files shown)


lgrammel:
Closing this PR because issue #16914 has already been closed.

PR #16955, opened 2026-07-08
feat(core): add support for generator-based tool progress/steps streaming

shahrryyar:
This PR adds support for streaming intermediate progress/sub-steps from generator-based tools (AsyncGenerator).

### Why is this needed?
Previously, the AI SDK supported AsyncIterable return values from tools, but forced every yielded value to match the final output schema of the tool. This meant progress logs, search state updates, and status messages could not easily be streamed without breaking the LLM's understanding of the final tool output type.

### How it works:
1. Generator-based tools can now yield progress update objects matching `{ type: 'progress', ... }` and return the final tool output using `return output`.
2. `@ai-sdk/provider-utils` (`executeTool`) captures intermediate yields of type `progress` as `{ type: 'progress', progress }`, and uses the final returned value of the generator as the final `tool-result`.
3. The stream controllers propagate `tool-progress` chunks to the client.
4. On the client, the UI stream processes `tool-progress` chunks and appends them to a new `steps` array field inside the corresponding message tool parts (`ToolUIPart` and `DynamicToolUIPart`).
5. Type definitions (`UIMessageChunk`, `UIToolInvocation`, `DynamicToolUIPart`) are fully updated and tested.

### DIFF
--- a/.changeset/tool-progress-updates.md
+++ b/.changeset/tool-progress-updates.md
@@ -1,0 +1,6 @@
+---
+"ai": patch
+"@ai-sdk/provider-utils": patch
+---
+
+feat(core): add support for generator-based tool progress/steps streaming
--- a/packages/ai/src/generate-text/execute-tool-call.ts
+++ b/packages/ai/src/generate-text/execute-tool-call.ts
@@ -142,7 +142,14 @@ export async function executeToolCall<TOOLS extends ToolSet>({
               });
 
               for await (const part of stream) {
-                if (part.type === 'preliminary') {
+                if (part.type === 'progress') {
+                  onPreliminaryToolResult?.({
+                    type: 'tool-progress',
+                    toolCallId,
+                    toolName,
+                    progress: part.progress,
+                  } as any);
+                } else if (part.type === 'preliminary') {
                   onPreliminaryToolResult?.({
                     ...toolCall,
                     type: 'tool-result',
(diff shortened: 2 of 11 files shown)


shahrryyar:
Rebased onto current `main` — this was conflicting, it now merges cleanly.

Two conflicts came from changes that landed since this PR was opened, resolved in favour of the new upstream code:

- `isOutputChunkType` in `stream-text.ts`: `tool-result` / `tool-error` were flipped to `false` upstream. Since that map classifies *semantic model output* (used for the first-content and inter-content timeouts), `tool-progress` is tool output rather than model output, so it is registered as `false` alongside them rather than `true` as originally written.
- `ui-message-chunks.ts`: the surrounding schemas moved from `z.strictObject` to `z.looseObject`, so the `tool-progress` chunk schema follows suit.

The `tool-progress` case in the chunk switch now goes through `enqueueStepPart()` like the neighbouring `tool-approval-request`, instead of calling `controller.enqueue()` directly.

Verified locally after the rebase:

- `packages/ai` type-check: no errors
- `stream-text.test.ts`: 453 passed, 3 skipped
- `packages/ai` `src/ui-message-stream` + `src/ui`: 25 files, 455 passed, no type errors
- `provider-utils` `execute-tool.test.ts`: 4 passed

Let me know if you would like the API shaped differently before review.

PR #16966, opened 2026-07-08
fix(ai): drop the hardcoded gateway version from the streaming-transcription error

gr2m:
## Background

Review finding on #16776: the `experimental_streamTranscribe` unsupported-model error message told users to "upgrade @ai-sdk/gateway to 4.0.13 or later for streaming transcription support". The `ai` package cannot keep a cross-package version claim true — and it is already false: @ai-sdk/gateway 4.0.13 and 4.0.14 have shipped on npm without `doStream`, so a user following the message upgrades and hits the identical error again.

## Summary

- `packages/ai/src/transcribe/stream-transcribe.ts`: the message now says "upgrade @ai-sdk/gateway to a version with streaming transcription support" — no version number.
- Regression test: the string-model-id error message must not contain a semver-looking version. The test is committed separately before the fix (note: CI only runs on PRs against `main`/`release-v*`, so there is no CI run on this PR — check out the `test` commit to see the test fail without the fix).

No new changeset: the base branch's `streaming-transcribe-error-message` changeset already covers this message change and its wording remains accurate.

Part of the pre-merge fixes for #16776 (4 of 6).

### DIFF
--- a/packages/ai/src/transcribe/stream-transcribe.ts
+++ b/packages/ai/src/transcribe/stream-transcribe.ts
@@ -116,7 +116,7 @@ export function streamTranscribe({
           ? ' String model IDs resolve through the global provider (AI Gateway by default).' +
             ' If that provider does not support streaming transcription, pass a provider model' +
             " instance instead (e.g. openai.transcription('gpt-realtime-whisper'))" +
-            ' or upgrade @ai-sdk/gateway to 4.0.13 or later for streaming transcription support.'
+            ' or upgrade @ai-sdk/gateway to a version with streaming transcription support.'
           : ''),
     });
   }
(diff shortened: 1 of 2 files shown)


packages/ai/src/transcribe/stream-transcribe.ts
```diff
@@ -116,7 +116,7 @@ export function streamTranscribe({
           ? ' String model IDs resolve through the global provider (AI Gateway by default).' +
             ' If that provider does not support streaming transcription, pass a provider model' +
             " instance instead (e.g. openai.transcription('gpt-realtime-whisper'))" +
-            ' or upgrade @ai-sdk/gateway to 4.0.13 or later for streaming transcription support.'
+            ' or upgrade @ai-sdk/gateway to a version with streaming transcription support.'
```
gr2m, line 119:
```suggestion
            ' or upgrade @ai-sdk/gateway to 4.0.13 or later for streaming transcription support.'
```

gr2m:
nah that's actually good

PR #16969, opened 2026-07-09
fix(amazon-bedrock): return request.body from doGenerate and doStream

xianjianlf2:
## Background

`AmazonBedrockChatLanguageModel.doGenerate()` and `doStream()` never populated the `request` field of their result (there was an explicit `// TODO request?` in `doStream`). Consumers that rely on `request.body` for observability/tracing — e.g. [Mastra](https://github.com/mastra-ai/mastra), which uses it to populate span inputs — received `undefined` instead of the Converse API request payload, resulting in empty tracing attributes.

Other providers (`@ai-sdk/anthropic`, `@ai-sdk/google`, `@ai-sdk/cohere`, …) already return `request: { body: args }`, so this gap was specific to the Bedrock provider.

## Summary

- Return `request: { body: args }` from both `doGenerate()` and `doStream()` (where `args` is the Converse API command built by `getArgs()`), matching the behavior of the other providers.
- Removed the stale `// TODO request?` marker.
- Added regression tests for both methods asserting `result.request?.body` contains the request payload, and updated the affected `doGenerate` result snapshots (purely additive — only the new `request` field).

## Manual Verification

Verified with the provider's mock-server test harness (no live AWS credentials required):

- The new `should return the request body` tests assert `result.request?.body` for both `doGenerate` and `doStream`. Their inline snapshots show the real Converse command (`messages`, `system`, `additionalModelResponseFieldPaths`, …) — i.e. the payload that was previously `undefined`.
- The updated `doGenerate` result snapshots confirm `request.body` matches the body actually sent to the mock server (`server.calls[0].requestBodyJson`).
- Full package suite green: `pnpm test` in `packages/amazon-bedrock` → **132 passed**. `pnpm check` (oxlint + oxfmt) clean, and `@ai-sdk/amazon-bedrock` builds with type declarations.

## Checklist

- [ ] Documentation has been added / updated (for bug fixes / features) — not needed; `request.body` is already part of the `LanguageModelV4` result contract, this only fixes the Bedrock provider to honor it

## Related Issues

Fixes #13927

### DIFF
--- a/.changeset/bedrock-request-body.md
+++ b/.changeset/bedrock-request-body.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix(amazon-bedrock): return `request.body` from `doGenerate` and `doStream`. Previously the Bedrock provider never populated the `request` field, so observability/tracing tools received `undefined` instead of the Converse API request payload.
--- a/packages/amazon-bedrock/src/__snapshots__/amazon-bedrock-chat-language-model.test.ts.snap
+++ b/packages/amazon-bedrock/src/__snapshots__/amazon-bedrock-chat-language-model.test.ts.snap
@@ -44,6 +44,29 @@ There are 3 r's.",
       },
     },
   },
+  "request": {
+    "body": {
+      "additionalModelRequestFields": undefined,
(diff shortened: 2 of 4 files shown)


aayush-kapoor:

PR #16971, opened 2026-07-09
fix(provider-utils): route provider-response URLs through the download-URL guard

dnukumamras:
## What this does

`getFromApi` in `@ai-sdk/provider-utils` fetches URLs that come from provider response bodies (image, audio, and video downloads, and polling URLs). Those URLs are untrusted, but they were fetched with no scheme, host, or IP checks. This PR routes them through the same guard that `downloadBlob` already uses.

- Added an opt-in `validateUrl` flag to `getFromApi` that routes the request through `fetchWithValidatedRedirects`, which rejects private, loopback, and link-local targets and re-validates every redirect hop. Blocked URLs throw `DownloadError`.
- Added header sanitization on the validated path: proxy, forwarding, cloud-metadata, and cookie headers are stripped before the request.
- Drop `Authorization` and `Cookie` on any redirect that crosses origin, so credentials never follow a redirect to a different host.
- Added a `credentialedOrigin` option so caller headers are sent only when the URL is same-origin with the provider's own origin, and withheld otherwise.
- Enabled the guard at the 11 call sites that fetch a URL from a response body, across fal, luma, gladia, fireworks, xai, black-forest-labs, and replicate.
- Closed two range gaps in `validateDownloadUrl`: IPv4 `224.0.0.0/4` (multicast) and IPv6 `2001:db8::/32` (documentation).

## Secure by default going forward

- `validateUrl` is a required field, with no default. Every current and future `getFromApi` call site has to make an explicit choice, so a new integration cannot silently inherit an unsafe default.
- `true` for a URL that comes from a response body, `false` for a URL built from a configured `baseURL`. The compiler will not let a new call site skip the decision.
- `credentialedOrigin` folds credential gating into the same call, so a new integration that opts into validation gets first-hop credential protection without hand-rolling a same-origin check.
- Documented the rule in `AGENTS.md`, `contributing/add-new-provider.md`, and `contributing/secure-url-handling.md`, plus a user-facing page in `content/docs`.

## Happy path

- A normal public asset or polling URL passes validation and is fetched and returned exactly as before, same return type, error handling, abort signal, and user-agent suffix.
- Same-origin polling URLs still receive the provider API key.
- Redirects to other public hosts are still followed, one hop at a time, each re-validated.

## Unhappy path

- A response URL pointing at a private, loopback, or link-local address (for example a cloud-metadata or internal host) is rejected with `DownloadError` and never fetched.
- A URL that starts public but redirects to an internal address is blocked at the redirect hop.
- Credentials are withheld when the URL is not same-origin with `credentialedOrigin`, and dropped if a redirect crosses origin.

## Not solved here

The guard does string and literal-IP checks and does not resolve DNS, so a hostname that resolves to a private address, and DNS rebinding, are out of scope at this layer. These need Node-only APIs (`undici`, `node:dns`) that are not available across all runtimes this package targets (edge, browser, Bun), and the threat is server-only. `content/docs/06-advanced/11-secure-url-fetching.mdx` explains how a server deployment closes the gap: restrict network egress, or inject a Node `fetch` that pins the resolved IP at connect time.


packages/provider-utils/src/get-from-api.ts
```diff
@@ -18,24 +20,61 @@ export const getFromApi = async <T>({
...
+  /**
+   * Required. Set `true` when `url` is untrusted (e.g. taken from a provider
+   * response body): it is routed through {@link fetchWithValidatedRedirects},
+   * which rejects private/loopback/link-local targets and re-validates redirect
+   * hops; blocked URLs throw `DownloadError`. Set `false` only for URLs built
+   * from a developer-configured endpoint. It is required (no default) so every
+   * call site makes an explicit trust decision rather than inheriting an
+   * unsafe one.
+   */
+  validateUrl: boolean;
```
gr2m: Adding a new required argument to a public API is a breaking change. I suggest we

1. make it optional
2. update all our own providers
3. send PRs to important 3rd party providers that already updated to AI SDK 7
4. Update our guides for creating provider packages
5. Create an issue to make this a required argument in AI SDK 8 (add issue to v8 milestone)

gr2m (changes requested):
I'm still looking into it but don't want it to be merged as is

packages/provider-utils/src/fetch-with-validated-redirects.ts
```diff
@@ -64,16 +83,25 @@ export async function fetchWithValidatedRedirects({
...
+      // hop. Whether that hop is followed or rejected by the guard, an
       // unconsumed 3xx body would leak the underlying socket.
       await cancelResponseBody(response);
-      currentUrl = new URL(location, currentUrl).toString();
+      const nextUrl = new URL(location, currentUrl).toString();
+
+      // Drop credentials before following a redirect that crosses origin so
+      // Authorization / Cookie are never sent to a different host.
+      if (currentHeaders !== undefined && !isSameOrigin(nextUrl, currentUrl)) {
+        currentHeaders.delete('authorization');
```
gr2m: Dropping only `authorization`/`cookie` mirrors the fetch spec — but the spec protects only `Authorization` because in browsers every other custom header triggers a CORS preflight, and the redirect target must opt in via `Access-Control-Allow-Headers`. There is no CORS backstop on the server, and several providers authenticate with custom headers: gladia uses `x-gladia-key`, Black Forest Labs uses `x-key`. An open redirect on the provider's own origin passes the `credentialedOrigin` first-hop gate and then forwards the API-key header to the foreign origin, while `Authorization`-based providers are protected against exactly this.

Follow-up: #17165 drops **all** caller headers except the user-agent on a cross-origin hop, which also makes redirects honor the documented `credentialedOrigin` contract ("headers are sent only if the URL is same-origin with this origin").

packages/replicate/src/replicate-video-model.ts
```diff
@@ -221,11 +220,10 @@ export class ReplicateVideoModel implements Experimental_VideoModelV4 {
         const pollUrl = finalPrediction.urls.get;
         const { value: statusPrediction } = await getFromApi({
           url: pollUrl,
-          // The polling URL comes from the provider response; only send
-          // credentials when it stays on the provider's own origin.
-          headers: isSameOrigin(pollUrl, this.config.baseURL)
-            ? await resolve(this.config.headers)
-            : undefined,
+          // pollUrl comes from the provider response body.
+          validateUrl: true,
```
gr2m, line 224: Hard-coding `validateUrl: true` with no escape hatch breaks self-hosted deployments: `createReplicate({ baseURL: 'http://localhost:5000' })` against a local cog server returns `urls.get = http://localhost:5000/…` in the response body, and this polling call now throws `DownloadError` before the injected `fetch` is ever consulted. The contributing doc gives this exact rationale for `validateUrl: false` ("validating it would break legitimate self-hosted / localhost base URLs") — the same reasoning applies to response URLs that point back at the configured host.

Follow-up: #17166 adds a `trustedOrigin` option — hops same-origin with the developer-configured `baseURL` skip target validation (they point at exactly the host a config-derived `validateUrl: false` request would fetch anyway); every other hop is still validated. Wired at all 11 `validateUrl: true` call sites.

gr2m (commented):
Reviewed with a focus on how the redirect/header handling compares to web-standard `fetch` behavior. The design is solid: manual per-hop validation is the only portable way to guard redirects (browsers make it impossible — `redirect: 'manual'` yields an unreadable `opaqueredirect` — and the browser fallback / fail-closed split here is exactly right). I've opened five follow-up PRs against this branch for the findings; each inline comment below links the relevant one.

Suggested merge order to minimize conflicts (they touch overlapping lines): #17168 → #17165 → #17169 → #17170 → #17166.

packages/provider-utils/src/get-from-api.ts
gr2m: A required property on `getFromApi` is a compile break for community provider packages built on `@ai-sdk/provider-utils`, shipped as a patch release. And the requirement only holds for TypeScript callers — in plain JS, `validateUrl: undefined` is falsy and silently takes the unvalidated path, so the "no unsafe default" guarantee was never enforced at runtime anyway.

Follow-up: #17168 makes the option optional (omitting behaves like `false`, byte-for-byte the pre-PR behavior) and enforces the explicit-decision rule for this repo through the contributor docs instead (`contributing/providers.md`, `AGENTS.md`, `contributing/secure-url-handling.md`).

packages/provider-utils/src/fetch-with-validated-redirects.ts
```diff
@@ -33,29 +48,33 @@ export async function fetchWithValidatedRedirects({
...
-  }
+  // Left undefined when no headers are provided (bare request); otherwise
+  // sanitized once and mutated in place across hops (cross-origin drop below).
+  let currentHeaders =
+    headers === undefined ? undefined : sanitizeRequestHeaders(headers);
+
+  const perHopInit = (redirect: RequestRedirect): RequestInit => {
+    const init: RequestInit = { signal: abortSignal, redirect };
+    if (currentHeaders !== undefined) {
+      init.headers = currentHeaders;
```
gr2m: This passes the same mutable `Headers` instance to every hop and mutates it in place for the cross-origin drop. Real `fetch` snapshots headers synchronously so behavior is correct, but an injected custom fetch (a supported option) that defers reading `init.headers` can observe post-mutation state — and in the tests, `mock.calls[0][1].headers` and `mock.calls[1][1].headers` are the same object, so a future assertion on an earlier hop would silently check later-hop state.

Follow-up: #17169 gives each hop its own `new Headers(currentHeaders)` snapshot.

packages/provider-utils/src/fetch-with-validated-redirects.ts
```diff
@@ -64,16 +83,25 @@ export async function fetchWithValidatedRedirects({
           message: `Redirect from ${currentUrl} could not be validated and was blocked`,
         });
       }
-      return await fetch(currentUrl, { ...baseInit, redirect: 'follow' });
+      return await fetch(currentUrl, perHopInit('follow'));
     }
 
     const location = response.headers.get('location');
     if (response.status >= 300 && response.status < 400 && location) {
```
gr2m: The fetch spec's redirect-status set is exactly 301/302/303/307/308 (https://fetch.spec.whatwg.org/#redirect-status) — 300 (Multiple Choices) and 304 (Not Modified) are not redirects even when a server attaches a `Location` header. Treating any 300–399 + `Location` as a redirect gives a hostile server one extra (still validated) hop; a spec divergence rather than a vulnerability, and it pre-dates this PR, but easy to align.

Follow-up: #17170 switches to the spec set; a `Location` on any other status is returned as the final response.

packages/provider-utils/src/validate-download-url.ts
```diff
@@ -116,8 +116,8 @@ function isPrivateIPv4(ip: string): boolean {
   if (a === 192 && b === 168) return true;
   // 198.18.0.0/15 (benchmarking)
   if (a === 198 && (b === 18 || b === 19)) return true;
-  // 240.0.0.0/4 (reserved, includes 255.255.255.255 broadcast)
-  if (a >= 240) return true;
+  // 224.0.0.0/4 (multicast) and 240.0.0.0/4 (reserved, incl. 255.255.255.255 broadcast)
+  if (a >= 224) return true;
```
gr2m, line 126: This PR blocks the IPv6 documentation range `2001:db8::/32` but leaves the equivalent IPv4 documentation ranges open: TEST-NET-1/2/3 (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`), plus the second IPv6 documentation prefix `3fff::/20` (RFC 9637). None are globally routable, so low risk — but same category, so same treatment.

Follow-up: #17170 blocks all four; the embedded-IPv4 checks (`::ffff:x.x.x.x`, NAT64) inherit the new IPv4 ranges automatically. Fun catch it flushed out: the existing "allow public IP" tests used `203.0.113.1` — TEST-NET-3 itself (that's what documentation addresses are for) — and are repointed at a genuinely public address, mirroring this PR's own `2001:db8::1` test repoint.

gr2m (approved):

PR #16991, opened 2026-07-09
reproduction: anthropic provider: toolChoice 'none' strips tools from request,

lgrammel:
## Bug

anthropic provider: toolChoice 'none' strips tools from request, causing empty response with tool_use history

## Reproduction

- `examples/ai-functions/src/reproduction/anthropic-tool-choice-none-tool-history.ts` reproduces the SDK request-shaping bug: with tool history and `toolChoice: { type: 'none' }`, the live Anthropic request had `requestHasTools: false` and `requestHasToolChoice: false`.
- `packages/anthropic/src/anthropic-tool-choice-none.issue-12378.test.ts` is a failing provider regression test that expects the request to keep the `tools` array and send `tool_choice: { type: 'none' }`.
- `packages/anthropic/src/__fixtures__/anthropic-tool-choice-none-with-tool-history.1.json` records a real Anthropic `claude-sonnet-4-6` response used by the provider test fixture.
- The current live `claude-sonnet-4-6` response returned non-empty text, but the SDK still stripped the documented `tools` and `tool_choice` fields from the request.
- The exact issue model `claude-sonnet-4-20250514` returned a live Anthropic 404 `not_found_error`, so the runnable live reproduction uses current `claude-sonnet-4-6`.

## Relevant Documentation

- https://platform.claude.com/docs/en/agents-and-tools/tool-use/define-tools?categoryid=2849204
- https://platform.claude.com/docs/en/api/messages/create
- https://platform.claude.com/docs/en/about-claude/models/overview

## Related Issues

Reproduces #12378

### DIFF
--- a/examples/ai-functions/src/reproduction/anthropic-tool-choice-none-tool-history.ts
+++ b/examples/ai-functions/src/reproduction/anthropic-tool-choice-none-tool-history.ts
@@ -1,0 +1,125 @@
+import { createAnthropic } from '@ai-sdk/anthropic';
+import type {
+  LanguageModelV4Content,
+  LanguageModelV4Prompt,
+} from '@ai-sdk/provider';
+
+const promptWithToolHistory: LanguageModelV4Prompt = [
+  {
+    role: 'user',
+    content: [{ type: 'text', text: 'Search for climate change' }],
+  },
+  {
+    role: 'assistant',
+    content: [
+      {
+        type: 'tool-call',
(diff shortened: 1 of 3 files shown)


lgrammel:
> The current live claude-sonnet-4-6 response returned non-empty text, but the SDK still stripped the documented tools and tool_choice fields from the request.

reproduction ignores key outcome

Issue #8332, opened 2025-08-27
`base64` is `undefined` when using `fullStream`. Value exists in `base64Data`


hsab:
### Description

# AI SDK File Type Mismatch: `base64` vs `base64Data` Property

## Issue Summary

There is a type mismatch between the expected and actual file structure in the AI SDK when handling generated files. The AI SDK's `GeneratedFile` interface expects a `base64` property, but the actual data contains a `base64Data` property. This occurs when directly returning the `fullStream` from a tRPC route, with SuperJSON as data transformer.

## Simple Architecture

The `fullStream` is returned from a tRPC route and as an `AsyncIterableStream<TextStreamPart<TOOLS>>`. Here's the flow:

### 1. Stream Creation in route.ts

```typescript
const result = streamText({
    model,
    messages: [...],
    // ... other options
    onFinish: async (response) => {
        // Handle token usage and billing
    },
});

return result.fullStream; // Returns AsyncIterableStream<TextStreamPart<TOOLS>>
```

### 2. Stream Consumption in UI.tsx

```typescript
const chatStream = await trpcVanilla.route.mutate({...});

for await (const value of chatStream) {
    if (value.type === 'file') {
        // ❌ Type mismatch occurs here
        const base64Data = 'base64' in value.file ? value.file.base64 : 'base64Data' in value.file ? (value.file as { base64Data: string }).base64Data : '';
    }
}
```

### 3. TextStreamPart Type Definition

The `fullStream` contains `TextStreamPart<TOOLS>` objects, which include:

```typescript
type TextStreamPart<TOOLS extends ToolSet> = {
    type: 'text-delta';
    text: string;
} | {
    type: 'file';
    file: GeneratedFile; // ❌ This should have 'base64' but runtime has 'base64Data'
} | {
    type: 'finish';
    // ... other properties
} | // ... other stream part types
```

### Expected Structure (AI SDK Types)
According to the AI SDK type definitions:

```typescript
interface GeneratedFile {
    /**
     * File as a base64 encoded string.
     */
    readonly base64: string;
    /**
     * File as a Uint8Array.
     */
    readonly uint8Array: Uint8Array;
    /**
     * The IANA media type of the file.
     */
    readonly mediaType: string;
}
```

### Actual Structure (Runtime Data)
The actual data received from the AI SDK stream contains:
```typescript
{
    type: 'file',
    file: {
        base64Data: string;  // ❌ Actual property name
        mediaType: string;
        // ... other properties
    }
}
```

[...]

PR #16993, opened 2026-07-09
reproduction: base64 is undefined when using fullStream. Value exists

lgrammel:
## Bug

`base64` is `undefined` when using `fullStream`. Value exists in `base64Data`

## Reproduction

- Updated `examples/ai-functions/src/reproduction/issue-8332-fullstream-file-base64.ts` to use an actual SuperJSON round trip and cover both `fullStream` and `generateImage`.
- After transport, `fullStream` exposes `base64Data` but the typed `GeneratedFile.base64` and `GeneratedFile.uint8Array` values are `undefined`; both values exist before transport.
- After transport, `generateImage().images` remains present, but `images[0].base64` and the promised `image` value are `undefined`, while the internal `base64Data` remains present.
- Expected behavior is for SuperJSON-transported results to retain the public `GeneratedFile` and `GenerateImageResult` contract rather than expose private backing fields.
- The mismatch remains on the current workspace `ai` 7.0.71, so upgrading from the reported 5.0.17 setup does not resolve it.
- Added the existing locked SuperJSON 2.2.6 dependency through `examples/ai-functions/package.json` and `pnpm-lock.yaml`.

## Relevant Documentation

- https://ai-sdk.dev/docs/reference/ai-sdk-core/default-generated-file
- https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-image
- https://trpc.io/docs/server/data-transformers
- https://github.com/ravionhq/superjson

## Related Issues

Reproduces #8332


lgrammel:
jumping to conclusions to early; their setup has major impact here and there is a memory tradeoff

Issue #16276, opened 2026-06-19
@ai-sdk/replicate image model fails when sync-wait returns status:starting with output:null


diegonava6:
### Description

`@ai-sdk/replicate` image generation intermittently throws `AI_APICallError: Invalid JSON response` (cause: `AI_TypeValidationError`) when Replicate returns a prediction that is still in progress.

This happens with model `black-forest-labs/flux-dev` during cold starts or when inference exceeds the sync-wait window.

#### Source (where the bug is)

`packages/replicate/src/replicate-image-model.ts` ([v2.0.36](https://github.com/vercel/ai/blob/caebb44016dbd084e5bf7b7f4ab5194fd2c7c045/packages/replicate/src/replicate-image-model.ts)).

The POST response is destructured directly into `{ output }` with no status check ([L140-L143](https://github.com/vercel/ai/blob/caebb44016dbd084e5bf7b7f4ab5194fd2c7c045/packages/replicate/src/replicate-image-model.ts#L140-L143)):

```ts
const {
  value: { output },
  responseHeaders,
} = await postJsonToApi({
```

…and validated against a schema that requires `output` to be present ([L206-L208](https://github.com/vercel/ai/blob/caebb44016dbd084e5bf7b7f4ab5194fd2c7c045/packages/replicate/src/replicate-image-model.ts#L206-L208)):

```ts
const replicateImageResponseSchema = z.object({
  output: z.union([z.array(z.string()), z.string()]),
});
```

There is no polling. The video model in the same package already handles this correctly (see Suggested fix).

#### Expected behavior

When Replicate's sync API (`Prefer: wait`) times out before the prediction completes, Replicate returns HTTP 201 with a valid prediction object (`status: "starting"` or `"processing"`, `output: null`, `urls.get` set). Per [Replicate's docs](https://replicate.com/docs/topics/predictions/create-a-prediction#polling), the client should poll `urls.get` until `status` is `succeeded`, then use `output`.

The **video model in the same package already implements this** (`replicate-video-model.ts`: keeps full `prediction`, polls while `starting`/`processing`, uses `replicatePredictionSchema` with `output.nullish()`).

The image model should behave the same way.

#### Actual behavior

`replicate-image-model.ts` posts with `Prefer: wait`, then immediately validates the response and throws. No polling occurs.

[...]

PR #17002, opened 2026-07-09
reproduction: @ai-sdk/replicate image model fails when sync-wait returns status:starting

lgrammel:
## Bug

@ai-sdk/replicate image model fails when sync-wait returns status:starting with output:null

## Reproduction

- Classified as provider-specific Replicate image-generation behavior; official Replicate docs describe `Prefer: wait` returning incomplete `starting`/`processing` predictions that should be fetched via `urls.get` until terminal.
- Added deterministic reproduction script at `examples/ai-functions/src/reproduction/replicate-image-sync-wait-output-null.ts`; it expects `generateImage` to poll a 201 `status: "starting"`, `output: null` prediction and return the mocked image, but it rejects with `AI_APICallError: Invalid JSON response`.
- Added provider fixture `packages/replicate/src/__fixtures__/image-sync-wait-starting-output-null.json` and failing provider test in `packages/replicate/src/replicate-image-model.test.ts`; the image model rejects on the initial successful POST before making any `GET` request to `urls.get`.
- Replicate live-service verification was skipped by policy; the reproduction uses a local response fixture shaped according to official Replicate prediction documentation.

## Relevant Documentation

- https://replicate.com/docs/topics/predictions/create-a-prediction
- https://replicate.com/docs/topics/predictions/lifecycle/

## Related Issues

Reproduces #16276

### DIFF
--- a/examples/ai-functions/src/reproduction/replicate-image-sync-wait-output-null.ts
+++ b/examples/ai-functions/src/reproduction/replicate-image-sync-wait-output-null.ts
@@ -1,0 +1,126 @@
+import { createReplicate } from '@ai-sdk/replicate';
+import { APICallError, generateImage } from 'ai';
+
+const initialPrediction = {
+  id: 'sync-wait-starting-output-null',
+  model: 'black-forest-labs/flux-dev',
+  version: 'test-version',
+  input: {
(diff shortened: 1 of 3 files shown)


lgrammel:
Need to enable replicate live.

PR #17010, opened 2026-07-09
fix: prevent newer fields on known UI message chunks from breaking older clients

lgrammel:
## Background

Cached clients use older UI message chunk schemas and previously threw `AI_TypeValidationError` when newer servers added optional top-level fields such as `providerMetadata` or `toolMetadata` to otherwise known chunk types.

## Summary

Changed all 28 UI message chunk schema variants from strict to loose Zod objects, accepting and preserving additional top-level fields while retaining validation of known fields, known discriminants, and the existing public `UIMessageChunk` type. Unknown chunk types remain rejected.

## Testing

Integrated the existing type assertion into the runtime schema test, added an inline-snapshot regression that parses an SSE chunk and verifies a newer-server field is accepted and preserved, and added coverage confirming unknown chunk types remain rejected. Focused Node and Edge suites each passed 3 tests, and the full repository test suite passed.

## End-to-end Validation

- `pnpm exec tsx --eval "<UI message SSE probe>"` successfully parsed and preserved an additional field on a known `tool-output-available` event without producing `AI_TypeValidationError`.

## Related Issues

Fixes #13733

Co-authored-by: pablodecm <6707437+pablodecm@users.noreply.github.com>
Co-authored-by: svenmuennich <1932115+svenmuennich@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

All 28 UI message chunk variants change from strict to loose Zod objects, so newer top-level fields are accepted and preserved while discriminants and known required fields remain validated.

Concerns: none.

### Side effects

Risk: low

The intentional behavior change is limited to no longer rejecting unknown top-level keys; malformed known fields and unknown chunk types remain rejected.

Concerns: none.

### Performance

Risk: low

Validation complexity is unchanged; preserving extra fields adds only payload-proportional retention with no new caching, buffering, or unbounded processing.

Concerns: none.

### Backwards compatibility

Risk: low

No stored-data migration or mutation is introduced, and existing UI message processing continues to consume recognized fields while allowing future fields to pass through.

Concerns: none.

### Security

Risk: low

Required fields and their types remain validated, unknown chunk types remain rejected, and the normal event-stream path parses input through secureJsonParse before applying the schema.

Concerns: none.

### Testing

Status: appropriate

The regression test verifies acceptance and preservation of a future field and retains the public UIMessageChunk type assertion; it passes in both Node and Edge configurations.

Concerns: none.

### Verification

Inspected origin/main...HEAD and confirmed all 28 strictObject variants became looseObject variants with a patch changeset. git diff --check passed. The schema regression tests passed 2/2 in Node and Edge, chat tests passed 27/27 in Node and Edge, and packages/ai type-check passed. A focused probe confirmed future fields are preserved on all 28 variants while invalid known shapes and unknown chunk types are still rejected.

svenmuennich:
Thanks for taking care of this @lgrammel. As I understand this fix it only covers new fields in known chunk types, but unknown chunk types introduced by newer backend versions (i.e. `tool-approval-response` added by `ai` v7) still result in validation errors. Am I missing something here?

lgrammel:
Addressed [PR comment 4970630802](https://github.com/vercel/ai/pull/17010#issuecomment-4970630802).

> ## Bugfix review
> 
> Outcome: approved
> 
> ### Fixes issue
> 
> Status: fully-addresses
> 
> ... 47 more line(s) omitted

No production change was needed for the approval. The fix remains limited to accepting additional fields on the 28 known chunk variants, and the focused and full verification suites pass.

lgrammel:
Addressed [PR comment 4973234678](https://github.com/vercel/ai/pull/17010#issuecomment-4973234678).

> Thanks for taking care of this @lgrammel. As I understand this fix it only covers new fields in known chunk types, but unknown chunk types introduced by newer backend versions (i.e. `tool-approval-response` added by `ai` v7) still result in validation errors. Am I missing something here?

You are correct: this fix intentionally covers new fields on known chunk types, while entirely unknown chunk types remain validation errors. Accepting unknown types would require widening the public chunk type and defining safe processing semantics for events the client does not understand, which is broader than this bug. I added a regression test that makes this boundary explicit and updated the forward-compatibility test to exercise the actual JSON event-stream parsing path.

PR #17069, opened 2026-07-10
fix(openai): do not send both Authorization header and subprotocol api key on the transcription WebSocket

kkdawkins:
## Background

Found while testing gateway streaming transcription (#16776 / vercel/ai-gateway#2509) end-to-end: `openai/gpt-realtime-whisper` through the gateway failed on every connect, and a direct repro against the live OpenAI API fails the handshake with:

> You must only send one of protocol api key and Authorization header.

`createOpenAIRealtimeTranscriptionStream` reuses the provider's HTTP header record for the WebSocket connect: the Bearer token is derived into the `openai-insecure-api-key.<token>` subprotocol (so native `WebSocket`, which ignores the headers option, can authenticate), but the full record — `Authorization` still inside — is also passed to the constructor. Native implementations silently drop it, so this was invisible until a header-capable implementation like `ws` is used (which server-side callers such as the AI Gateway must, because xAI's STT WebSocket is header-auth-only). `ws` faithfully forwards the header, OpenAI receives both auth channels, and rejects.

## Summary

- `packages/openai/src/transcription/openai-transcription-model.ts`: `getOpenAIRealtimeProtocols(headers)` → `getOpenAIRealtimeConnection(headers)` returning `{ protocols, headers }`. When the Bearer token rides the subprotocol, `Authorization`/`authorization` are stripped from the constructor headers; other headers still pass through. No token → both untouched. The subprotocol scheme itself is unchanged.
- Regression tests: constructor headers must not contain Authorization when the subprotocol carries the key (fails without the fix); non-auth custom headers still reach the constructor; no-Authorization case passes headers through unchanged.

## Verification

- `packages/openai`: 761/761 tests pass.
- Live against OpenAI with `webSocket: ws`: previously failed instantly with the dual-auth rejection; with the fix, `gpt-realtime-whisper` streams deltas and finishes with the correct transcript.

Side-finding (not addressed here): OpenAI realtime transcription rejects `inputAudioFormat.rate < 24000` mid-session ("Expected a value >= 24000").

### DIFF
--- a/.changeset/openai-realtime-ws-dual-auth.md
+++ b/.changeset/openai-realtime-ws-dual-auth.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+Fix streaming transcription over header-capable WebSocket implementations: the realtime WebSocket handshake sent the api key in both the `openai-insecure-api-key` subprotocol and the `Authorization` header, which OpenAI rejects ("You must only send one of protocol api key and Authorization header"). The Authorization header is now stripped when the subprotocol carries the key.
--- a/packages/openai/src/transcription/openai-transcription-model.test.ts
+++ b/packages/openai/src/transcription/openai-transcription-model.test.ts
@@ -560,6 +560,14 @@ describe('doStream', () => {
       'realtime',
       'openai-insecure-api-key.test-api-key',
     ]);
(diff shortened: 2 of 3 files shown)


kkdawkins:
Consolidating into #16776 (the streaming-transcription PR) per discussion — the commit lives there as 99dd16994d.

PR #17070, opened 2026-07-10
fix(xai): only emit transcript-final on speech_final and reconstruct empty finish text

kkdawkins:
## Background

Found while testing gateway streaming transcription (#16776 / vercel/ai-gateway#2509) end-to-end against the live xAI STT API, with raw upstream events captured via `includeRawChunks`:

1. **Duplicated finals**: xAI sends a finalized utterance twice — `transcript.partial` with `is_final: true, speech_final: false`, then the identical text with `speech_final: true`. Emitting `transcript-final` for every `is_final: true` (the previous behavior) duplicates every utterance.
2. **Revised finals**: live mic sessions show `is_final: true` *fragments* (e.g. `"No"`, `", I'm not"`) whose text the eventual `speech_final` event merges and re-punctuates (`"No, I'm not."`) — so `is_final && !speech_final` text is still revisable, which is partial semantics, not final.
3. **Empty finish**: xAI's `transcript.done` arrives with `text: ""`. `finish.text` was built solely from it, so `finish` went out empty and `experimental_streamTranscribe` threw `AI_NoTranscriptGeneratedError` despite a full transcript having streamed.

## Summary

- `packages/xai/src/xai-transcription-model.ts`:
  - `transcript-final` is now only emitted on `is_final && speech_final`; finalized fragments (and the `speech_final: false` re-send) surface as `transcript-partial`.
  - `finish.text` falls back to the accumulated finalized utterances per channel (joined with spaces, plus any trailing text that never got a `speech_final`) when `transcript.done.text` is empty. A non-empty `done.text` still wins.
- 4 regression tests modeled on the captured live sequences (duplicate re-send, fragment merge, no-speech_final trailing text, multi-utterance join).

## Verification

`packages/xai`: 396/396 tests pass. End-to-end validation happens through the gateway preview (vercel/ai-gateway#2509) on the next snapshot.

Note the intentional semantic change beyond the dedupe: consumers previously saw `is_final` fragments as `transcript-final`; they are now `transcript-partial` because the live API later revises that text. Finals should never be superseded.

### DIFF
--- a/.changeset/xai-stt-final-parts.md
+++ b/.changeset/xai-stt-final-parts.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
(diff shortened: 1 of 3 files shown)


kkdawkins:
Consolidating into #16776 (the streaming-transcription PR) per discussion — the commit lives there as e2f4f623fc.

PR #17094, opened 2026-07-10
feat(provider-utils): experimental transcription-stream WebSocket envelope

kkdawkins:
PR 2 of 3 (streaming transcription stack; stacked on #17093, replaces #16776).

## Summary

The standard serialization of transcription-model `doStream` over a WebSocket — the wire contract both the AI Gateway server and the `@ai-sdk/gateway` client (PR 3) implement. The module TSDoc is the normative spec.

- Frame type constants (`transcription-stream.start`, `transcription-stream.audio-done`); binary frames are audio, TEXT frames are JSON.
- `parseTranscriptionStreamClientFrame` (server side): envelope-shape validation, `secureJsonParse` (prototype-pollution safe), never throws.
- `serializeTranscriptionStreamPart` (server side): `Error` payloads in `error` parts serialize as `{ name, message }` (Error props are non-enumerable — a bare stringify loses the message end-to-end).
- `parseTranscriptionStreamPart` (client side): per-part-type field validation; malformed known-type frames from a drifted server degrade to skipped frames instead of opaque mid-stream TypeErrors; revives `response-metadata.timestamp`.

This is reviewable as a spec: no consumers in this PR. Landing it publishes the contract so the gateway server side (vercel/ai-gateway#2509) can build against released `@ai-sdk/provider-utils` independently of the client (PR 3).

Incorporates the #16965 hardening review (thanks @gr2m).

## Verification

51 envelope tests; provider-utils suite 714/714 (node+edge).


packages/provider-utils/src/index.ts
```diff
@@ -95,6 +95,15 @@ export {
...
 } from './streaming-tool-call-tracker';
 export { stripFileExtension } from './strip-file-extension';
+export {
+  parseTranscriptionStreamClientFrame,
+  parseTranscriptionStreamPart,
+  serializeTranscriptionStreamPart,
+  TRANSCRIPTION_STREAM_AUDIO_DONE_FRAME_TYPE,
+  TRANSCRIPTION_STREAM_START_FRAME_TYPE,
+  type Experimental_TranscriptionStreamClientFrame,
+  type Experimental_TranscriptionStreamStartFrame,
```
gr2m: make APIs experimental at the point of export, e.g. 

`type TranscriptionStreamClientFrame as Experimental_TranscriptionStreamClientFrame`

shouldn't all exported APIs be experimental, not just types?

kkdawkins: Done in 74046b1e7c — all envelope APIs now carry experimental prefixes applied at the export point (`experimental_*` functions, `EXPERIMENTAL_*` constants, `Experimental_*` types); internal module names stay clean. The gateway consumer (#17095) adopts the prefixed names in its layer.

gr2m (approved):

PR #17101, opened 2026-07-10
feat: thread the tool-approval decision into tool execution as options.approval

### LINKED ISSUE #17083: @ai-sdk/harness: approval reason for an APPROVED host tool is dropped — forwarded for denials and built-in tools, never reaches the tool's execute() or the model
## Description

`addToolApprovalResponse({ id, approved, reason })` lets a client attach a `reason` to a tool-approval response. The harness delivers that reason **asymmetrically** — verified against the published `@ai-sdk/harness@1.0.24` dist (`processPendingApprovalContinuation` in `dist/agent/index.js`):

- **Denied** host tool → the reason reaches the model via `submitToolResult({ output: { type: 'execution-denied', reason } })`. ✅
- **Built-in** tool approval → the reason is forwarded via `control.submitToolApproval({ approved, reason })`. ✅
- **Approved host tool** → `maybeExecuteHostTool` re-executes the tool with its **original input only**. The reason is recorded on the `tool-approval-response` part (so it appears in the transcript), but it never reaches the tool's `execute()` — and therefore never reaches the model, which only sees the tool result. ❌

This matters because the approval prompt is a natural human-in-the-loop data channel: "approved — and here is the user's answer / the edited plan / the corrected value". The `reason` field is the documented way to attach context to an approval decision, and it already works on the deny path; the approve path silently drops it.

## Reproduction

```ts
const tools = {
  ask_user_question: tool({
    inputSchema: z.object({ question: z.string() }),
    needsApproval: true,
    execute: async (input, options) => {
      // options has toolCallId / messages / abortSignal — nothing from the approval
      // response. The user's answer (sent as the approval reason) is not available here.
      return 'answer?';
    },
  }),
};

// client:
addToolApprovalResponse({
  id: part.approval.id,
  approved: true,
  reason: JSON.stringify({ answers: { question_1: '…' } }),
});
```

The continuation executes the tool with the original input; the `reason` payload is nowhere in the execution, and the tool result the model consumes cannot include it.

## Suggested fix

Forward the approval response into host-tool execution options, e.g.:

```ts
execute(input, { toolCallId, messages, abortSignal, approval: { approved: true, reason } })
```

mirroring what the denial path already does through `execution-denied`. Additive on the execution options type, so existing tools are unaffected.

**Update:** PR is up — #17101, generalized per the discussion below: rather than patching only the harness call site, the approval decision (`{ approvalId, approved, reason? }`) is threaded as `options.approval` through **every** execution path — `generateText`/`streamText` explicit continuations, automatic `toolApproval` approvals, and the harness continuation — with tests on each path and docs. Thanks @babyblueviper1 for the "full decision payload threads through every path uniformly" framing that shaped it.

## Workaround

We scan the incoming `UIMessage[]` for tool parts carrying `approval.reason`, build a `toolCallId → parsed payload` map out-of-band before starting the turn, and each HITL tool's `execute()` looks up its own `toolCallId` so the data reaches the model through the tool result. Works, but it is exactly the plumbing the SDK could do itself — and it only works because the approval response happens to ride the resent messages.

## Versions

- `@ai-sdk/harness@1.0.24` (latest; first verified on 1.0.18), `@ai-sdk/harness-claude-code@1.0.24`
- `ai@7.0.20`

doneumark:
## Background

Fixes #17083.

`addToolApprovalResponse({ id, approved, reason })` lets a client attach a `reason` to a tool-approval response, but the decision payload was delivered asymmetrically: denials reach the model via `execution-denied` and built-in (harness) approvals forward it via `submitToolApproval` — while an **approved** tool re-executes with its original input only, so the reason never reaches `execute()` or the model.

As [@babyblueviper1 put it on the issue](https://github.com/vercel/ai/issues/17083#issuecomment-4937504975), the general shape of the bug is that an approval decision's `reason` is part of the decision, not a side-channel to it — so the fix threads the **full decision payload** through **every** execution path uniformly, rather than patching the single harness call site the issue originally pointed at.

## Summary

- **`@ai-sdk/provider-utils`**: new `ToolExecutionApproval` type (`{ approvalId, approved: true, reason? }`) and an optional `approval` field on `ToolExecutionOptions` — the options every tool `execute(input, options)` receives. `approved` is typed `true`: only approved calls execute; denials keep delivering their reason via `execution-denied`.
- **`ai`**: `executeToolCall` accepts an optional `approval` and forwards it into the execute options. All approval-carrying call sites thread it:
  - `generateText` / `streamText` **explicit continuations** (a `tool-approval-response` in the incoming messages) — the decision from `collectToolApprovals` rides into the execution.
  - **Automatic approvals** from the `toolApproval` configuration — both the `generateText` step loop and `streamText`'s `executeToolsFromStream` (keyed per `toolCallId` with a prototype-safe `Map`).
  - Executions with no approval flow pass nothing — `'approval' in options` stays `false`, so the change is invisible to existing tools.
- **`@ai-sdk/harness`**: `maybeExecuteHostTool` accepts the approval and `processPendingApprovalContinuation` passes the continuation's decision, so host tools resumed from an approval pause finally see the client's `reason`.
- **Docs**: `options.approval` documented in the tool-calling approval section with an example.

## Tests

- `execute-tool-call.test.ts`: approval (with reason) reaches `execute`; the `approval` key is absent entirely when there is no approval flow.
- `generate-text.test.ts` / `stream-text.test.ts`: the existing "approved continuation" scenarios now assert `execute` received `{ approvalId, approved: true, reason? }`.
- `run-prompt.test.ts` (harness): a continued approval with a JSON reason arrives in the host tool's `options.approval`.
- `tool-execute-function.test-d.ts`: exact-type assertion updated for the new optional field.

## Manual Verification

This is extracted from our production integration (HITL tools over `HarnessAgent` where the approval UI collects structured user data — question answers, edited plans — as the approval `reason`). Today we scan the incoming `UIMessage[]` for approval parts and hand each tool a `toolCallId → payload` map out-of-band; with this change the same data arrives in `execute(input, { approval })` directly.

Automated checks run locally:

```
pnpm --filter @ai-sdk/provider-utils test         # 656 passed (node + edge)
pnpm --filter ai test                              # 3223 passed (node + edge)
pnpm --filter @ai-sdk/harness test                 # 162 node + 148 edge passed
pnpm --filter @ai-sdk/provider-utils type-check && pnpm --filter ai type-check && pnpm --filter @ai-sdk/harness type-check
pnpm --filter @ai-sdk/provider-utils build && pnpm --filter ai build && pnpm --filter @ai-sdk/harness build
pnpm exec oxlint packages/provider-utils/src/types packages/ai/src/generate-text packages/harness/src/agent/internal
```

## Checklist

## Related Issues

Fixes #17083


doneumark:
@babyblueviper1 tagging you for review, since the shape of this PR follows your [comment on the issue](https://github.com/vercel/ai/issues/17083#issuecomment-4937504975): the full decision payload (`{ approvalId, approved, reason? }`) threads uniformly through every execution path as `options.approval` — explicit continuations, automatic `toolApproval` approvals, and the harness continuation — instead of patching the one call site from the report. (I can't formally request review as an external contributor, hence the mention.)

doneumark:
Gentle bump — brought the branch current with `main` today; the full CI suite is green on the new head. Also re-checked the published `@ai-sdk/harness@1.0.32` (released today): the approve-path drop from #17083 still reproduces, so this is still needed.

Recap for a reviewer landing here: `options.approval` (`{ approvalId, approved: true, reason? }`) threads uniformly through every approval-carrying execution path — `generateText`/`streamText` explicit continuations, automatic `toolApproval` approvals, and the harness continuation — with tests on each path. It's additive on `ToolExecutionOptions` (executions without an approval flow carry no `approval` key), and the deny path is untouched. Happy to adjust scope or approach if maintainers prefer something narrower.

babyblueviper1:
Reviewed the diff across all four call sites — this closes the gap cleanly.

The key thing that made me check carefully: does `approvalsByToolCallId` actually key correctly per-call, or could a reason leak across concurrent tool calls in the same step? Traced it through `execute-tool-call.ts` — `approval: approvalsByToolCallId?.get(toolCall.toolCallId)` is looked up per-call inside the map over `toolCalls`, so each execution only ever sees the decision keyed to its own `toolCallId`. No cross-call leakage.

The other thing worth confirming explicitly: `approved: true` as a literal type on `ToolExecutionApproval` is the right call, not just a style choice — it means a tool's `execute` can never receive a denied decision (denials still route through `execution-denied` only), so there's no need for the tool body to re-check `approved` defensively. That's a real type-level guarantee, not documentation.

Symmetry across the three paths (explicit continuation, automatic `toolApproval`, harness resume) is the part that actually fixes the original bug — the issue wasn't that one call site was missing the reason, it's that the harness and explicit-deny paths threaded the decision and the approve path silently didn't, so any tool logic branching on `options.approval?.reason` worked in two paths and broke in the third with no type error to catch it. This PR removes that asymmetry at the type level, not just the harness call site the original report pointed at.

Ran through the test additions (execute-tool-call, generate-text, stream-text, harness run-prompt) — they cover the three carrying paths plus the undefined case for calls with no approval flow. Looks complete and mergeable from where I'm sitting.

doneumark:
Thanks @babyblueviper1 for the detailed review — especially for tracing the per-`toolCallId` keying in `execute-tool-call.ts` and calling out the `approved: true` literal type as a real guarantee rather than style; both are exactly the intent.

Status for maintainers: branch is current with `main` (merged today) and all test / type-check / lint checks pass on the latest head. Two marks are red, both outside the diff:

- **Vercel** — "Authorization required to deploy" (fork PR, needs maintainer authorization).
- **Bundle Size Check** — node bundle comes in at 640.41 KB vs the 640 KB limit. `main` is already at 639.68 KB (100.0% of the limit), so the ~0.7 KB of runtime code this feature adds trips the check — as would effectively any PR adding runtime code to `ai`. The limit lives in the `BUNDLE_SIZE_LIMIT_KB` repo Actions variable, which I can't adjust from a fork; happy to try golfing the implementation down instead if that's preferred.

Also re-verified against today's latest published packages (`ai@7.0.31`, `@ai-sdk/provider-utils@5.0.11`, `@ai-sdk/harness@1.0.36`): `ToolExecutionOptions` still carries no approval decision and the approved host-tool continuation still executes with the original input only, so #17083 still reproduces. No rush — just keeping the status current.

doneumark:
The cross-call question was the right one to press on, and your conclusion matches the intent. One small correction to the pointer, though: the per-call lookup lives in `executeTools` in `generate-text.ts` (with its streaming twin in `execute-tools-from-stream.ts`); `execute-tool-call.ts` never sees the map at all, it just receives the single already-resolved decision for its own call. Same conclusion, different file — and to be fair, I sloppily repeated that attribution in my status comment above.

Your trace did surface a real gap, however: nothing in the suite actually pinned the isolation property. The existing "two calls approved" scenario asserted tool results and prompts but never looked at what each `execute` received, so a refactor that collapsed the map into, say, a single variable would have passed cleanly. Pushed [`857727d`](https://github.com/doneumark/ai/commit/857727d56a35ca2480c1d834b516248041a6b503) to close that: the two concurrent approvals now carry distinct reasons, and a new test asserts each execution received exactly the decision keyed to its own `toolCallId`. Both vitest runtimes green locally.

On `approved: true`: intentional, exactly as you read it. A denied decision reaching `execute` should be unrepresentable rather than merely documented away, and if a future path ever wants tools to observe denials, that has to arrive as a visible API change instead of a quiet type widening.

Thanks for reading it at this depth.

babyblueviper1:
Fair correction, and worth taking at face value rather than glossing over: I traced the wrong file. `execute-tool-call.ts` receiving a single already-resolved `approval` (not the map) is the actual shape — I read the per-call `Map.get` in `executeTools` and mentally attributed the lookup to the file that consumes its output instead of the file that does it. Thanks for tracing it back precisely rather than just saying "close enough."

The isolation-test gap is the part that actually matters here, and 857727d closes it properly — asserting each concurrent execution received the decision keyed to its own toolCallId, not just that results/prompts came out right. That's the difference between a test that would pass a broken refactor and one that wouldn't; good that the review pushed on the specific claim hard enough to find where the suite's coverage understated the invariant.

felixarntz:
@doneumark Thank you for the PR, apologies for leaving it unreviewed for such a long time. I'm closing in favor of iteration #21383, which is fresh and without conflicts, though admittedly we'll need to do a more careful review on the fix either way because it goes far beyond `@ai-sdk/harness` changes and touches AI SDK foundational code.

PR #17102, opened 2026-07-10
fix(google): strip unsupported frequencyPenalty/presencePenalty for g…

### LINKED ISSUE #8943: Penalty is not enabled for models/gemini-2.5-*
### Description

As of the Google AI team's current API support, the `frequency_penalty` and `presence_penalty` parameters are only supported for the gemini-2.0-* models. They no longer available in the gemini-2.5-* series.

If they are sent in a request targeting a model that does not support them, Gemini will return an INVALID_ARGUMENT error.

I've tested with `gemini-2.5-pro` and `gemini-2.5-flash`, Vercel AI throwed an error message: `Penalty is not enabled for models/gemini-2.5-flash`

[Reference discussion on the issue: Doom loop frequency penalty 2.5 flash](https://discuss.ai.google.dev/t/doom-loop-frequency-penalty-2-5-flash/94446)"

### AI SDK Version

- "ai": "^5.0.54"
-  "@ai-sdk/google": "^2.0.17"
-  "@ai-sdk/openai": "^2.0.36"

### Code of Conduct

AryanSherigar:
## Background
Gemini 2.5 models (`gemini-2.5-pro`, `gemini-2.5-flash`, `gemini-2.5-flash-lite`, etc.) no longer support the `frequencyPenalty` and `presencePenalty` parameters — Google removed these from the API for the 2.5 series. When these params are sent to a 2.5 model, Gemini returns an `INVALID_ARGUMENT` error and the request fails entirely, e.g.: Penalty is not enabled for models/gemini-2.5-flash
See also this related discussion: https://discuss.ai.google.dev/t/doom-loop-frequency-penalty-2-5-flash/94446

## Summary
- Added detection for `gemini-2.5-*` models via a regex on `modelId`, matching the existing pattern already used for `isGemini3Model`/`isGemmaModel` in `google-language-model.ts`.
- When a 2.5 model is targeted and `frequencyPenalty`/`presencePenalty` are set, both are now stripped from the outgoing `generationConfig` instead of being forwarded to the API.
- An `unsupported` warning (matching the existing `SharedV4Warning` shape used elsewhere in this provider) is pushed for each stripped parameter, so callers are informed rather than silently losing the setting or hitting an opaque API error.
- Non-2.5 models (e.g. `gemini-2.0-*`) are unaffected — both params continue to pass through unchanged.

## Manual Verification
Ran the patched provider directly against the live Gemini API (bypassing the `ai` package, calling `GoogleLanguageModel.doGenerate` directly):
- `gemini-2.5-flash` with `frequencyPenalty: 0.5` → previously threw an `INVALID_ARGUMENT`-style API error; now returns a normal text response with a `{ type: 'unsupported', feature: 'frequencyPenalty', ... }` warning.
- Same result confirmed for `presencePenalty`.
- Confirmed a non-2.5 model still sends `frequencyPenalty`/`presencePenalty` through unchanged in the outgoing request body (regression check).

## Checklist

## Related Issues
Fixes #8943


AryanSherigar:
Pulled in the fixture-backed regression test and live-error fixtures from #17862 (the automated reproduction PR) and ran them against this fix. All 6 cases (Flash, Flash-Lite, Pro × frequencyPenalty/presencePenalty) pass on both node and edge test targets, 698/698 tests passing overall.

One note: the original test asserted an exact warning shape (`{ type: 'unsupported', feature }`), but the `SharedV4Warning` type also allows an optional `details` field, which this fix includes for better DX. I loosened that one assertion to `expect.objectContaining(...)` so it validates the required fields without being brittle about the optional one, and pushed that update alongside the fixtures.

AryanSherigar:
The automated review flagged that scoping the penalty-stripping logic by model ID alone (`isGemini25Model`) also strips frequencyPenalty/presencePenalty for Vertex AI requests, even though Vertex still documents support for these settings on Gemini 2.5 models. Since `GoogleLanguageModel` is shared between `@ai-sdk/google` and `@ai-sdk/google-vertex`, I checked this fix for the same issue and confirmed it had it too.

Fixed by reusing the existing `isVertexProvider` check already used elsewhere in this file:

const shouldStripPenalties = isGemini25Model && !isVertexProvider;

Added a regression test proving Vertex Gemini 2.5 requests keep both penalties in `generationConfig` and do not emit an `unsupported` warning. Full `@ai-sdk/google` test suite passes, including the new Vertex case.

Since this scoping fix applies to `GoogleLanguageModel` regardless of which PR merges, happy for it to be reused in #18051 directly if that ends up being the version that lands.

aayush-kapoor:
will be implemented in https://github.com/vercel/ai/pull/18051

PR #17113, opened 2026-07-10
perf(ai): speed up streamed string tool input

dexhunter:
## Background

Each `tool-input-delta` currently reparses the complete accumulated JSON prefix. For a large tool argument streamed as one progressive string, that repeats work on every chunk and makes the partial-JSON parsing path scale quadratically. This is one of the remaining partial-JSON parsing sites discussed in #15670.

## Summary

- retain the parsed key and value for the exact case of one top-level, unclosed, unescaped string property
- build a new input object for every delta so update/reference semantics stay unchanged
- fall back to the general partial JSON parser for quotes, escapes, control characters, nested structures, trailing input, and other shapes
- clear the transient fast-path state when input becomes available or errors
- add regression coverage for progressive updates, dynamic tools, replacement-object semantics, escaped and nested JSON, trailing junk, `__proto__`, and terminal cleanup
- add a patch changeset for `ai`

## Performance

Using the same strict harness on current `main` (`7d9f2cfad3`) and this branch, with 4,000 one-character deltas:

| | Current main | This PR |
|---|---:|---:|
| median processing time | 13,646.641 ms | 63.495 ms |
| 2N/N scaling ratio | 3.872 | 1.643 |
| log-log slope | 1.979 | 0.711 |

That is a 99.53% reduction, or about 215x, for this scoped partial-JSON processing workload. This is not an end-to-end UI or request latency claim.

The incremental parsing direction was identified through [autoresearch with Weco](https://dashboard.weco.ai/share/iz2dKvSGQ0RVNnRkBaIH3s75xp4SwO4G). The final implementation was rebuilt on current main, typed, reduced, and hardened against counterexamples.

## Verification

- `process-ui-message-stream.test.ts`: 108 passed
- `pnpm build:packages`: 70/70 packages passed
- `oxfmt --check` and `oxlint`
- strict evaluator: all 11 correctness gates passed
- compatibility check with #17025: after retaining both adjacent state fields, 111 tests and type-check passed

Related to #15670.

### DIFF
--- a/.changeset/faster-tool-streams.md
+++ b/.changeset/faster-tool-streams.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Improve streamed tool input performance for large progressive string arguments.
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -3073,6 +3073,313 @@ describe('processUIMessageStream', () => {
     });
   });
 
+  it('should preserve progressive top-level string tool input', async () => {
+    const stream = createUIMessageStream([
+      { type: 'start', messageId: 'msg-123' },
+      { type: 'start-step' },
+      {
+        type: 'tool-input-start',
+        toolCallId: 'tool-call-string',
+        toolName: 'test-tool',
+      },
+      {
+        type: 'tool-input-delta',
+        toolCallId: 'tool-call-string',
+        inputTextDelta: '{"payload":"a',
+      },
+      {
+        type: 'tool-input-delta',
+        toolCallId: 'tool-call-string',
+        inputTextDelta: 'bc',
+      },
+      {
+        type: 'tool-input-delta',
+        toolCallId: 'tool-call-string',
+        inputTextDelta: '\\n',
+      },
+      {
+        type: 'tool-input-delta',
+        toolCallId: 'tool-call-string',
+        inputTextDelta: 'def',
+      },
+      {
+        type: 'tool-input-delta',
+        toolCallId: 'tool-call-string',
+        inputTextDelta: '"}',
+      },
+      {
+        type: 'tool-input-available',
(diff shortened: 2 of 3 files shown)


dexhunter:
Two things after re-verifying this on today's `main` (`eb16508f`).

**Correction to the numbers in the description.** I re-ran the benchmark and cannot reproduce the 13,646 ms baseline I quoted. Measuring the workload the description names — one top-level string argument streamed as N one-character `tool-input-delta` chunks through `processUIMessageStream` — on the exact base this PR was written against (`7d9f2cfa`) and on current `main`, I get ~168 ms and ~165 ms at 4,000 deltas, not ~13.6 s. Neither `parse-partial-json.ts` nor `fix-json.ts` changed between those commits, so this is a fault in my original harness, not a change upstream. The reproducible numbers, one host, same run:

| deltas | `main` | this branch | speedup |
|---:|---:|---:|---:|
| 1,000 | 43.3 ms | 13.4 ms | 3.2x |
| 2,000 | 90.3 ms | 25.5 ms | 3.5x |
| 4,000 | 164.7 ms | 26.2 ms | 6.3x |
| 8,000 | 476.4 ms | 40.3 ms | 11.8x |

The effect is real and grows with argument size — `main` is super-linear on this path and the branch is close to flat — but it is roughly 3-12x on this workload, not the ~215x the description claims. Please read the description against this table; sorry for the noise.

**Current state.** Merges cleanly onto `main`, and on the merged tree `src/ui/process-ui-message-stream.test.ts` passes 108/108 with `tsc` clean, including the repeated-tool-call-ID tests #17586 added to that file after this branch was cut.

The only red mark is the `Vercel` commit status, "Authorization required to deploy" — the fork preview deployment is waiting on a Vercel-side click. The `CI` and `Verify Changesets` workflows are likewise sitting in `action_required` and have never actually run on this head, which is why I ran their equivalents locally.

Happy to split the 307 test lines into their own commit, or to narrow the fast path further, if either would make this easier to look at.

PR #17121, opened 2026-07-12
feat (provider/bytedance): add Seedream image model support

shaper:
## Summary

Adds a ByteDance **Seedream image model** to `@ai-sdk/bytedance`, exposed via `.image()` / `.imageModel()`. Previously the provider only supported video and threw `NoSuchModelError` for image models.

The gateway serves ByteDance images through `@ai-sdk/openai-compatible`, whose image model routes edits to a multipart `/images/edits` endpoint. ByteDance ARK has no such endpoint — text-to-image **and** image-to-image both use a single JSON `POST /images/generations`, with input images passed in the `image` field (string or array). Routing edits to `/images/edits` produced non-JSON responses ("Invalid JSON response"). This model implements the correct ARK shape.

## Changes

- **`bytedance-image-model.ts`** — `ImageModelV4` using ARK's single JSON `/images/generations`; input images (single or multi) go in the `image` field via `convertImageModelFileToDataUri` (URLs pass through, binary/base64 → data URI). `response_format` pinned to `b64_json`.
- **`bytedance-image-model-options.ts`** — typed `ByteDanceImageModelOptions` (`looseObject` schema, mirroring the video model): `watermark`, `outputFormat`, `size` (resolution level), `sequentialImageGeneration`, `maxImages`, `optimizePromptMode`; camelCase → ARK snake_case with passthrough for unknown fields.
- **`bytedance-image-settings.ts`** — `ByteDanceImageModelId` (known Seedream ids + open string for `ep-…` endpoints).
- Wired `image` / `imageModel` into the provider; exported new types from `index.ts`.
- **Tests** — generate, edit (single / multi / URL), typed-option mapping, resolution-level `size` override, unknown-option passthrough, unsupported-setting warnings (`aspectRatio`, `seed`, `n`, `mask`).
- **Docs** — new Image Models section in `85-bytedance.mdx` (text-to-image, edit, multi-image blending, options table, capabilities table) + link to the ModelArk image generation API.
- **Examples** — `generate-image/bytedance/{basic,edit,edit-multi-reference}.ts`.
- Changeset: `@ai-sdk/bytedance` minor.

## Verification

- `tsc --noEmit`: clean (package + examples)
- `vitest run` (package): all tests pass
- `pnpm build`: success

Unsupported inputs warn rather than break: `aspectRatio` (→ use `size`), `seed`, `n>1` (→ `sequentialImageGeneration`), and `mask` (Seedream has no mask-based inpainting).

Linear: AIG-145 (https://linear.app/vercel/issue/AIG-145)


packages/bytedance/src/bytedance-image-model-options.ts
```diff
@@ -0,0 +1,58 @@
...
+   * `'auto'`. The number of input reference images plus generated images must
+   * not exceed the model's limit.
+   */
+  maxImages?: number | null;
+
+  /**
+   * Prompt optimization mode. `seedream-4-0` supports both `'standard'` and
+   * `'fast'`; other models support `'standard'` only.
+   */
+  optimizePromptMode?: 'standard' | 'fast' | null;
```
shaper, line 42: **Design decision — flat vs. nested provider options (feedback wanted)**

I flattened ByteDance's two nested request objects into single camelCase options:

- `maxImages` → `sequential_image_generation_options.max_images`
- `optimizePromptMode` → `optimize_prompt_options.mode`

The model reassembles the nested ModelArk shape before sending.

**Why flat:** cleaner typed surface, and today each nested object has exactly one documented field, so a nested option would just be a one-key wrapper.

**Trade-off:** if ByteDance adds more fields under either object later, a flat API needs a new top-level option each time, whereas mirroring ARK 1:1 (`sequentialImageGenerationOptions: { maxImages }`, `optimizePromptOptions: { mode }`) would extend naturally and read closer to the API docs.

Preference on flat vs. nested here? Happy to switch before merge.

@felixarntz for any steering, I figure we can do flattened for now and switch to sub-objects if warranted later (and be backwards compatible).

content/providers/01-ai-sdk-providers/85-bytedance.mdx
```diff
@@ -58,6 +58,149 @@ You can use the following optional settings to customize the ByteDance provider
...
+  `providerOptions.bytedance` and are validated by ModelArk.
+</Note>
+
+### Image Model Capabilities
+
+| Model             | Model ID                       | Capabilities                                                                                                                          |
+| ----------------- | ------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------- |
+| Seedream 5.0 Pro  | `dola-seedream-5-0-pro-260628` | Text-to-image, single/multi image-to-image, interactive editing (markers). Sizes: 1K, 2K. Formats: png, jpeg. Up to 10 references.    |
+| Seedream 5.0 Lite | `seedream-5-0-260128`          | Text-to-image, single/multi image-to-image, batch generation. Sizes: 2K, 3K, 4K. Formats: png, jpeg. Up to 14 references.             |
+| Seedream 4.5      | `seedream-4-5-251128`          | Text-to-image, single/multi image-to-image, batch generation. Sizes: 2K, 4K. Format: jpeg. Up to 14 references.                       |
```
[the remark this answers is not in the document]
shaper: Fixed in 67a4d1a522. Per ByteDance's [image generation docs](https://docs.byteplus.com/en/docs/ModelArk/1824690), the 5.0 family has only Pro and Lite — the Lite model's primary id is `seedream-5-0-260128`, and the docs list `seedream-5-0-lite-260128` as an additional id for the same model ("also supports"). So the label was correct, but we were omitting the second id — the table now shows it as an alias, and both ids are already in the `ByteDanceImageModelId` type.

kkdawkins (approved):
Ship it!

packages/bytedance/src/bytedance-image-model.ts
```diff
@@ -0,0 +1,212 @@
...
+  'outputFormat',
+  'size',
+  'sequentialImageGeneration',
+  'maxImages',
+  'optimizePromptMode',
+]);
+
+interface ByteDanceImageModelConfig extends ByteDanceConfig {
+  _internal?: {
+    currentDate?: () => Date;
```
kkdawkins, line 34: `maxImagesPerCall = 10` makes `generateImage({ n: 2 })` use one call, but this ignores `n`, so the API returns one image. Map `n` to sequential generation or set this to 1.

PR #17125, opened 2026-07-12
fix(amazon-bedrock): parse exception frame error as object in Anthropic SSE shim

### LINKED ISSUE #17124: @ai-sdk/amazon-bedrock/anthropic: exception frames forwarded with `error` as a string, failing the anthropic chunk schema (AI_TypeValidationError)
### Description

When streaming through `createBedrockAnthropic` (`@ai-sdk/amazon-bedrock/anthropic`) and Bedrock delivers an event-stream **`exception` frame** mid-stream (e.g. an `internalServerException` with the message "The system encountered an unexpected error during processing. Try your request again."), the SSE shim forwards it as:

```json
{"type":"error","error":"{\"message\":\"The system encountered an unexpected error during processing. Try your request again.\"}"}
```

i.e. with `error` as the **raw exception-body JSON string**. The anthropic chunk schema requires `error` to be an object (`{ type: string, message: string }`), so the chunk fails Zod validation and consumers get an opaque `AI_TypeValidationError` ("Invalid input: expected object, received string" at path `["error"]`) instead of the actual Bedrock error. The real message — which explicitly says to retry — is buried inside the validation error's `value`, so retry logic keyed off error type/status can't engage.

The culprit is the `exception` branch in `packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-fetch.ts` (still present on `main`):

```ts
} else if (event.messageType === 'exception') {
  controller.enqueue(
    textEncoder.encode(
      `data: ${JSON.stringify({ type: 'error', error: event.data })}\n\n`,
    ),
  );
}
```

`event.data` is the decoded exception body *string*, so it's embedded as a string. Note the non-streaming error path just above it handles this correctly by building a proper object: `error: { type: 'error', message }`. The exception branch presumably wants the same treatment (parse `event.data`, extract `message`, wrap as `{ type, message }` — the `:exception-type` header is available in the decoder if you want to preserve the exception type, which would also let the SDK mark `overloaded`/throttling errors as retryable).

For precedent: Anthropic's own TypeScript SDK had the same class of bug in its Bedrock integration (error events inside chunk frames losing their type/message) and fixed it by converting them to a proper `APIError` — anthropics/anthropic-sdk-typescript#1021.

We see this steadily in production (Claude via Bedrock, `invoke-with-response-stream`): ~150 occurrences over the last few weeks, always this exact shape.

#### Reproduction

Self-contained — fakes the Bedrock event-stream wire format with a custom `fetch`, no AWS account needed. Only needs `@ai-sdk/amazon-bedrock` and `@smithy/eventstream-codec`:

```ts
import { EventStreamCodec } from '@smithy/eventstream-codec';
import { toUtf8, fromUtf8 } from '@smithy/util-utf8';
import { createBedrockAnthropic } from '@ai-sdk/amazon-bedrock/anthropic';

const codec = new EventStreamCodec(toUtf8, fromUtf8);
const encoder = new TextEncoder();

const chunkFrame = (anthropicEvent: object) =>
  codec.encode({
    headers: {
      ':message-type': { type: 'string', value: 'event' },
      ':event-type': { type: 'string', value: 'chunk' },
    },
    body: encoder.encode(
      JSON.stringify({
        bytes: Buffer.from(JSON.stringify(anthropicEvent)).toString('base64'),
      }),
    ),
  });

// What Bedrock sends when the stream fails server-side.
const exceptionFrame = codec.encode({
  headers: {
    ':message-type': { type: 'string', value: 'exception' },
    ':exception-type': { type: 'string', value: 'internalServerException' },
  },
  body: encoder.encode(
    JSON.stringify({
      message:
        'The system encountered an unexpected error during processing. Try your request again.',
    }),
  ),
});

const frames = [
  chunkFrame({
    type: 'message_start',
    message: {
      id: 'msg_1',
      type: 'message',
      role: 'assistant',
      content: [],
      model: 'claude',
      stop_reason: null,
      stop_sequence: null,
      usage: { input_tokens: 1, output_tokens: 1 },
    },
  }),
  chunkFrame({
    type: 'content_block_start',
    index: 0,
    content_block: { type: 'text', text: '' },
  }),
  chunkFrame({
    type: 'content_block_delta',
    index: 0,
    delta: { type: 'text_delta', text: 'Hello' },
  }),
  exceptionFrame,
];

const fakeFetch: typeof fetch = () =>
  Promise.resolve(
    new Response(
      new ReadableStream<Uint8Array>({
        start(controller) {
          for (const f of frames) controller.enqueue(f);
          controller.close();
        },
      }),
      {
        status: 200,
        headers: { 'content-type': 'application/vnd.amazon.eventstream' },
      },
    ),
  );

const provider = createBedrockAnthropic({
  region: 'us-east-1',
  apiKey: 'test-token',
  fetch: fakeFetch,
});
const model = provider('us.anthropic.claude-sonnet-4-5-20250929-v1:0');

const { stream } = await model.doStream({
  prompt: [{ role: 'user', content: [{ type: 'text', text: 'hi' }] }],
});

for await (const part of stream) {
  if (part.type === 'error') console.log(part.error);
  else console.log('part:', part.type);
}
```

Output:

```
part: stream-start
part: response-metadata
part: text-start
part: text-delta
AI_TypeValidationError: Type validation failed: Value: {"type":"error","error":"{\"message\":\"The system encountered an unexpected error during processing. Try your request again.\"}"}.
Error message: [
  {
    "expected": "object",
    "code": "invalid_type",
    "path": ["error"],
    "message": "Invalid input: expected object, received string"
  }
]
```

Expected: an error stream part (or thrown `APICallError`) carrying the actual Bedrock message ("The system encountered an unexpected error during processing. Try your request again.") and ideally the exception type, so downstream retry logic can act on it.

### AI SDK Version

- ai: 7.0.15
- @ai-sdk/amazon-bedrock: 5.0.11 (bug still present in 5.0.17 and current `main`)
- @ai-sdk/anthropic: 4.0.8
- @ai-sdk/provider-utils: 5.0.5

### Code of Conduct

Aayush-engineer:
## Background:

When Bedrock delivers an exception frame mid-stream (e.g. internalServerException, ThrottlingException), the SSE shim was forwarding `error` as a raw JSON string instead of an object. This caused the Anthropic chunk schema to throw `AI_TypeValidationError: expected object, received string` at path `["error"]`, burying the actual Bedrock message and preventing retry logic from engaging.

## Summary:

- Added `exceptionType` field to `DecodedEvent` interface, extracted from the `:exception-type` header
- Updated the exception branch in `amazon-bedrock-anthropic-fetch.ts` to parse `event.data`, extract the message, and wrap it as `{ type: exceptionType, message }` — matching the existing non-streaming error path pattern
- Updated the existing exception test to assert correct structured behavior

## Manual Verification:

Used the self-contained reproduction script from #17124 (fake fetch, no AWS account needed) — the stream now correctly emits a structured error part instead of throwing AI_TypeValidationError. 406/406 tests passing.

## Related Issues:

Fixes #17124

### DIFF
--- a/.changeset/large-clouds-sell.md
+++ b/.changeset/large-clouds-sell.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix(amazon-bedrock): parse Bedrock exception frames as structured error objects in Anthropic SSE shim
--- a/packages/amazon-bedrock/src/amazon-bedrock-event-stream-decoder.ts
+++ b/packages/amazon-bedrock/src/amazon-bedrock-event-stream-decoder.ts
@@ -4,6 +4,7 @@ import { toUtf8, fromUtf8 } from '@smithy/util-utf8';
 export interface DecodedEvent {
   messageType: string;
   eventType: string;
+  exceptionType?: string;
   data: string;
 }
 
@@ -46,9 +47,14 @@ export function createAmazonBedrockEventStreamDecoder<T>(
             const messageType = decoded.headers[':message-type']
               ?.value as string;
             const eventType = decoded.headers[':event-type']?.value as string;
+            const exceptionType = decoded.headers[':exception-type']
+              ?.value as string;
             const data = textDecoder.decode(decoded.body);
 
-            await processEvent({ messageType, eventType, data }, controller);
+            await processEvent(
+              { messageType, eventType, exceptionType, data },
+              controller,
+            );
           } catch {
             break;
           }
--- a/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-fetch.test.ts
+++ b/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-fetch.test.ts
@@ -148,7 +148,70 @@ describe('createAmazonBedrockAnthropicFetch', () => {
     const text = new TextDecoder().decode(value);
 
     expect(text).toBe(
-      `data: ${JSON.stringify({ type: 'error', error: errorData })}\n\n`,
+      `data: ${JSON.stringify({ type: 'error', error: { type: 'ThrottlingException', message: 'Rate limit exceeded' } })}\n\n`,
+    );
+  });
+
+  it('should fall back to a generic error type when :exception-type header is missing', async () => {
+    const codec = new EventStreamCodec(toUtf8, fromUtf8);
+    const errorData = JSON.stringify({ message: 'Something went wrong' });
+    const amazonBedrockEvent = codec.encode({
+      headers: {
+        ':message-type': { type: 'string', value: 'exception' },
+      },
+      body: fromUtf8(errorData),
+    });
+    const stream = new ReadableStream({
+      start(controller) {
+        controller.enqueue(amazonBedrockEvent);
+        controller.close();
+      },
+    });
+    const mockResponse = createMockResponse(
+      stream,
+      'application/vnd.amazon.eventstream',
+    );
+    const baseFetch = createMockFetch(mockResponse);
+    const wrappedFetch = createAmazonBedrockAnthropicFetch(baseFetch);
+    const response = await wrappedFetch('https://example.com', {});
+    const reader = response.body!.getReader();
+    const { value } = await reader.read();
+    const text = new TextDecoder().decode(value);
+
+    expect(text).toBe(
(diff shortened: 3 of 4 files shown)


Aayush-engineer:
Rebased onto main and cleaned a few things up: exceptionType is now optional in DecodedEvent since it's only set on actual exception frames, the exception branch now reuses amazonBedrockErrorSchema (same one the non-streaming path uses) instead of an unchecked cast, and I added tests for the two fallback cases that weren't covered — missing :exception-type header and a non-JSON exception body. Full suite still green, 816 tests passing.

PR #17127, opened 2026-07-12
fix(harness-claude-code): preserve nested tool schemas

### LINKED ISSUE #17126: @ai-sdk/harness-claude-code loses nested object JSON Schema details for user-supplied tools
### Description

When using `@ai-sdk/harness-claude-code` with user-supplied AI SDK tools, nested object JSON Schema properties appear to be degraded to an unconstrained `any` shape when the Claude Code bridge exposes those tools through its local MCP server.

This makes tools with valid nested object schemas appear generic to the model.

For example, given a tool schema like:

```ts
import { jsonSchema, tool } from "ai";

const tools = {
  create_record: tool({
    type: "dynamic",
    description: "Create a record in an external system.",
    inputSchema: jsonSchema({
      type: "object",
      properties: {
        input: {
          type: "object",
          description: "Record creation parameters.",
          properties: {
            title: {
              type: "string",
              description: "The record title.",
            },
            priority: {
              type: "string",
              description: "The record priority.",
              enum: ["low", "medium", "high"],
            },
            assignee: {
              type: "object",
              description: "The person assigned to the record.",
              properties: {
                id: {
                  type: "string",
                  description: "The assignee ID.",
                },
                notify: {
                  type: "boolean",
                  description: "Whether to notify the assignee.",
                },
              },
              required: ["id"],
              additionalProperties: false,
            },
          },
          required: ["title"],
          additionalProperties: false,
        },
      },
      required: ["input"],
      additionalProperties: false,
    }),
    execute: async input => input,
  }),
};
```

I would expect the Claude Code harness bridge to preserve the nested schema for `input.title`, `input.priority`, `input.assignee.id`, etc.

Instead, the nested object appears to become a generic `input: any` parameter. In practice, the model sees only a generic `input` object and does not have enough schema information to call the tool naturally.

From the installed package source, this seems to come from `packages/harness-claude-code/src/bridge/index.ts`, where `jsonSchemaToZodShape` only handles top-level primitive property types:

```ts
switch (val.type) {
  case 'string':
    z_ = z.string();
    break;
  case 'number':
  case 'integer':
    z_ = z.number();
    break;
  case 'boolean':
    z_ = z.boolean();
    break;
  case 'array':
    z_ = z.array(z.any());
    break;
  default:
    z_ = z.any();
}
```

For `type: "object"`, it falls through to `z.any()` instead of recursively converting `properties`.

### Expected behavior

Nested JSON Schema objects should be recursively converted when user-supplied tools are exposed through the Claude Code harness bridge.

At minimum, a schema like:

```json
{
  "type": "object",
  "properties": {
    "input": {
      "type": "object",
      "properties": {
        "title": { "type": "string" }
      },
      "required": ["title"]
    }
  },
  "required": ["input"]
}
```

should expose `input` as an object with a `title` string property, not as an unconstrained value.

### Actual behavior

Nested object properties are not preserved. The bridge exposes the top-level object property as generic/unconstrained, so the model does not see the nested parameter schema.

### Suggested fix

Update `jsonSchemaToZodShape` / related conversion logic in `@ai-sdk/harness-claude-code` to recursively handle:

- `type: "object"` with `properties` and `required`
- `type: "array"` with `items`
- nullable unions such as `anyOf: [{ ... }, { "type": "null" }]`
- `oneOf` / `enum` / `const` where practical
- nested descriptions

### AI SDK Version

- `ai`: `7.0.16`
- `@ai-sdk/harness`: `1.0.19`
- `@ai-sdk/harness-claude-code`: `1.0.19`
- Node: `22.17.0`
- Zod: `4.4.3`

jdanielnd:
Fixes #17126.

## Summary
- Add a recursive JSON Schema to Zod bridge for `@ai-sdk/harness-claude-code` tool registration.
- Preserve nested object properties, required keys, arrays, nullable values, enum/const literals, and descriptions when exposing user tools through the local MCP server.
- Fall back to `z.any()` for unsupported JSON Schema features instead of throwing.

## Tests
- `../../node_modules/.bin/vitest --config vitest.node.config.js --run src/bridge/json-schema-to-zod.test.ts`
- `../../node_modules/.bin/vitest --config vitest.node.config.js --run`
- `./node_modules/.bin/tsc --build`
- `git diff --check`


rraina:
+1 on also having the same issue

felixarntz (approved):
@jdanielnd Thank you for the PR - looks great!

.changeset/tall-bridges-map.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Preserve nested user tool JSON schemas when exposing tools through the Claude Code MCP bridge.
```
felixarntz:
```suggestion
fix(harness-claude-code): preserve nested user tool JSON schemas when exposing tools through the MCP bridge
```

PR #8463, opened 2025-09-05
fix(openai): update timestamp_granularities to array format

Yuschyk:
Hi @gr2m 👋

Thank you all for building the AI SDK

Here is pr that should fix incorrect timestamp_granularities passing to sdk.

Fixes #8447 timestampGranularities doesn't work in aisdk

PR #17131, opened 2026-07-13
fix: stop dropping repeated multipart fields, and send elevenlabs diarize once

### LINKED ISSUE #6561: @ai-sdk/openai transcription providerOptions timestamp_granularities is broken
### Description

Im trying to create an openai compatible proxy for transcription API using Vercel AI SDK

```
openai = createOpenAI({
    baseURL: 'http://localhost:8790/api/v1/openai',
});

const result = await transcribe({
    model: openai.transcription('openai/whisper-1'),
    audio: AUDIO_BASE64,
    providerOptions: {
        openai: {
            language: 'en',
            temperature: 0.2,
            response_format: 'verbose_json',
            timestamp_granularities: ['word', 'segment'],
        }
    },
    maxRetries: 0,
});
```

The code above results in formData below:

```
app.post('/api/v1/openai/audio/transcriptions', (c: Context) => {
    const formData = await c.req.formData();
    console.log('formData', formData);
});

FormData {
      model: 'openai/whisper-1',
      file: File {
        size: 21639,
        type: 'audio/ogg',
        name: 'audio',
        lastModified: 1748636152134
      },
      language: 'en',
      temperature: '0.2',
      timestamp_granularities: 'segment'
    }
```

1) For some reason `timestamp_granularities` was transformed from array into string. Is it a bug?

https://github.com/vercel/ai/blob/2e6e5d31e0ed0726b60652bc00a5e59a349c3d2b/packages/openai/src/openai-transcription-model.ts#L162

This line looks suspisious to me. 

```
> String(['word', 'segment'])
"word,segment"
```

But form multipart array shall look like:
```
timestamp_granularities=word
timestamp_granularities=segment
```

### AI SDK Version

- ai: ^4.3.15
- @ai-sdk/openai: ^1.3.22

ryanda9910:
### What

Multipart encodes an array by **repeating a key**: OpenAI wants `timestamp_granularities[]=word` + `timestamp_granularities[]=segment`, xAI wants one `keyterm` field per term.

`requestBodyMultipart` in `@ai-sdk/test-server` overwrote entries as it walked the form data:

```ts
formData.forEach((value, key) => {
  entries[key] = value; // repeated key -> every value but the last is lost
});
```

So a repeated field collapsed to its final value and **no test could assert on one at all**. Repeated keys now collect into an array; a key seen once stays a scalar, so existing assertions are untouched.

### What that was hiding

Two real problems surfaced the moment the helper stopped discarding data:

**1. `@ai-sdk/elevenlabs` sent `diarize` twice.** The model appends a default `diarize=true`, then appends the caller's value on top. An explicit `diarize: false` therefore went out as `diarize=true` **and** `diarize=false`, leaving the server to disambiguate — the user's choice could be silently ignored. Changed to `set`, so the caller overwrites the default.

**2. The xAI transcription test asserted the wrong thing.** It expected `keyterm: 'Grok'` for an input of `['AI SDK', 'Grok']`. The request always carried both terms; the helper was dropping one. The expectation now matches what is actually sent — no production change needed there, the model was already correct.

### Regression test for #6561

The reported bug (`timestamp_granularities` stringified into `"word,segment"`) is **already fixed on main** — the array is appended per item. But nothing covered it: the existing test passes a *single* granularity, and with one element the broken and the correct encoding are indistinguishable. A future refactor could reintroduce it with every test still green.

Added a test that passes two granularities and asserts they go out as repeated fields.

Verified by mutation: restoring the old `formData.append(key, String(value))` makes the new test fail, so it genuinely locks the behaviour.

### Verification

All 8 packages that use `requestBodyMultipart` pass:

| package | tests |
|---|---|
| anthropic | 443 ✅ |
| cartesia | 32 ✅ |
| deepgram | 25 ✅ |
| elevenlabs | 15 ✅ |
| groq | 98 ✅ |
| openai | 773 ✅ |
| revai | 6 ✅ |
| xai | 392 ✅ |

Patch changeset included for the two published packages that change behaviour.

Closes #6561.

### DIFF
--- a/.changeset/tidy-jars-shake.md
+++ b/.changeset/tidy-jars-shake.md
@@ -1,0 +1,16 @@
+---
+'@ai-sdk/elevenlabs': patch
+'@ai-sdk/test-server': patch
+---
+
+fix (provider/elevenlabs): don't send `diarize` twice when it is set explicitly
+
+The model always appended a default `diarize=true`, then appended the caller's value on top,
+so the request carried the field twice and an explicit `diarize: false` was left for the server
+to disambiguate. It now overwrites the default instead.
+
+fix (test-server): keep repeated multipart fields instead of dropping all but the last
+
+`requestBodyMultipart` overwrote entries on each key, so a repeated field (how multipart
+encodes an array, e.g. OpenAI's `timestamp_granularities[]`) collapsed to its final value and
+could not be asserted on. Repeated keys now collect into an array; single keys are unchanged.
--- a/packages/elevenlabs/src/elevenlabs-transcription-model.ts
+++ b/packages/elevenlabs/src/elevenlabs-transcription-model.ts
@@ -91,7 +91,10 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
       };
 
       if (typeof elevenlabsOptions.diarize === 'boolean') {
-        formData.append('diarize', String(elevenlabsOptions.diarize));
+        // `set`, not `append`: a default `diarize=true` is written above, and appending would
+        // send the field twice (`diarize=true` *and* `diarize=false`), leaving it to the server
+        // to pick — so an explicit `diarize: false` could be silently ignored.
+        formData.set('diarize', String(elevenlabsOptions.diarize));
       }
 
       for (const key in transcriptionModelOptions) {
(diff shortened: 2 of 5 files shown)


gr2m:
closed via #8463

PR #17132, opened 2026-07-13
fix: WorkflowAgent drops assistant text between tool-call steps

### LINKED ISSUE #17123: WorkflowAgent drops assistant text parts from the conversation prompt on tool-calls finishes
## Description

`WorkflowAgent`'s streaming loop drops assistant **text parts** from the conversation prompt whenever a step finishes with `finishReason: "tool-calls"`. Only the tool calls are appended:

```ts
// packages/workflow/src/stream-text-iterator.ts (same code in dist of 1.0.4 → 1.0.22)
if (finishReason === 'tool-calls') {
  lastStepWasToolCalls = true;
  conversationPrompt.push({
    role: 'assistant',
    content: toolCalls.map((toolCall) => ({ type: 'tool-call', ... })), // text parts from the same step are discarded
  });
```

The `finishReason === 'stop'` branch *does* keep text (`step.content.filter((item) => item.type === 'text')`), so text survives only on the terminal step. This diverges from core `ai`'s `toResponseMessages`, which preserves text (and reasoning) parts in assistant messages during tool loops — code written against `streamText`/`ToolLoopAgent` semantics silently loses this behavior when moved to `WorkflowAgent`.

## Why this matters

On long agentic runs the effect is severe: any conclusion the model narrates before calling tools ("Found the root cause — implementing the fix now") is erased from its own context on the next step. The model cannot remember having decided anything.

Real-world datapoint: we ran a Sonnet-based coding subagent on `WorkflowAgent` for a scoped bug-fix task. Decoding the actual `doStreamStep` prompts from the workflow journal showed 312 assistant messages containing 315 tool-call parts and **zero text parts**, while the provider telemetry showed the model had emitted narrative text on 85 of those steps. The agent re-derived and re-announced the same root cause 26 times across 314 steps (~5 hours) without ever acting on it — each "I found it, let me implement" vanished from the prompt before the next generation. A sibling agent on the same task class that happened to act within a step or two escaped the loop; the pattern only becomes visible on runs long enough for carried-forward decisions to matter.

## Repro sketch

1. Run a `WorkflowAgent` with any tool and a prompt that makes the model narrate before calling the tool.
2. In `prepareStep` (or by inspecting the journaled `doStreamStep` inputs), examine `messages`.
3. Assistant messages for prior steps contain only `tool-call` parts; the narration text is gone.

Expected: assistant messages in the conversation prompt include the step's text parts (matching `toResponseMessages` behavior in core `ai`).

## Suggested fix

In the `tool-calls` branch, prepend the step's text parts (as the `stop` branch already collects them) to the assistant message content before the tool-call parts. Reasoning parts likely need to stay excluded unless provider metadata (signatures) is preserved, but plain text is safe to retain and is what providers recommend replaying.

## Workaround

We restore the text in `prepareStep` (whose returned `messages` replace the conversation prompt): map each step's text parts back onto the assistant message that carries its tool-call ids. Happy to upstream a PR if the fix direction sounds right.

## Environment

- `@ai-sdk/workflow` 1.0.4; bug still present in 1.0.22 dist and current `main` (`packages/workflow/src/stream-text-iterator.ts`, tool-calls branch ~line 391)
- Provider: `@ai-sdk/anthropic` (behavior is provider-independent)

Possibly related to the WorkflowAgent telemetry gap in #12164 only in that both are WorkflowAgent-specific divergences from core behavior.

lgrammel:
## Background

WorkflowAgent discarded narration emitted alongside tool calls, leaving subsequent model steps with incomplete conversation context.

## Summary

WorkflowAgent now carries assistant text parts forward alongside tool-call parts in the conversation prompt.

## Testing

Added regression coverage asserting that the next step receives both the prior narration and tool call.

## End-to-end Validation

- `issue-17123-workflow-agent-tool-call-text.ts`: ran `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17123-workflow-agent-tool-call-text.ts` against Anthropic; the live model generated narration and a tool call, and the next-step prompt retained both.

## Related Issues

Fixes #17123

Co-authored-by: devontivona <1065495+devontivona@users.noreply.github.com>

### DIFF
--- a/.changeset/tidy-tools-remember.md
+++ b/.changeset/tidy-tools-remember.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/workflow': patch
+---
+
+Preserve assistant text in the conversation prompt when a WorkflowAgent step also calls tools.
--- a/examples/ai-functions/src/reproduction/issue-17123-workflow-agent-tool-call-text.ts
+++ b/examples/ai-functions/src/reproduction/issue-17123-workflow-agent-tool-call-text.ts
@@ -1,93 +1,15 @@
+import { anthropic } from '@ai-sdk/anthropic';
 import { tool, type ModelMessage } from 'ai';
-import { MockLanguageModelV4, convertArrayToReadableStream } from 'ai/test';
 import { z } from 'zod/v4';
 import { WorkflowAgent } from '../../../../packages/workflow/dist/index.js';
 
-const narration = 'Found the root cause — implementing the fix now.';
-
-const usage = {
-  inputTokens: {
-    total: 3,
-    noCache: 3,
-    cacheRead: undefined,
-    cacheWrite: undefined,
-  },
-  outputTokens: {
-    total: 10,
-    text: 10,
-    reasoning: undefined,
-  },
-};
+const narration = 'ISSUE-17123-NARRATION';
 
 async function main() {
-  let callCount = 0;
   let secondStepMessages: ModelMessage[] | undefined;
 
-  const model = new MockLanguageModelV4({
-    doStream: async () => {
-      if (callCount++ === 0) {
-        return {
-          stream: convertArrayToReadableStream([
-            { type: 'stream-start' as const, warnings: [] },
-            {
-              type: 'response-metadata' as const,
-              id: 'response-1',
-              modelId: 'mock-model',
-              timestamp: new Date(0),
-            },
-            { type: 'text-start' as const, id: 'text-1' },
-            {
-              type: 'text-delta' as const,
-              id: 'text-1',
-              delta: narration,
-            },
-            { type: 'text-end' as const, id: 'text-1' },
-            {
-              type: 'tool-call' as const,
-              toolCallId: 'call-1',
-              toolName: 'applyFix',
-              input: '{}',
-            },
-            {
-              type: 'finish' as const,
-              finishReason: {
-                unified: 'tool-calls' as const,
-                raw: 'tool_calls',
-              },
-              usage,
-            },
-          ]),
-        };
-      }
-
-      return {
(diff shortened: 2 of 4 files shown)


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The tool-calls branch now appends the step's assistant text before its tool-call parts, so the next WorkflowAgent step receives the narration that was previously discarded.

Concerns: none.

### Side effects

Risk: low

Tool-call construction and metadata sanitization remain unchanged, and existing no-text and parallel-tool-call tests pass; the only behavioral change is the intended retention of assistant text.

Concerns: none.

### Performance

Risk: low

Retained narration increases prompt, journal, and token size proportionally to generated text, as required, but the implementation adds no duplication or new superlinear processing.

Concerns: none.

### Backwards compatibility

Risk: none

No stored-data schema or serialization format changes are introduced; subsequent steps simply receive additional previously omitted text content.

Concerns: none.

### Security

Risk: none

The change only replays model-generated assistant text through the existing prompt path and introduces no new input source, parsing operation, privilege, or data exposure.

Concerns: none.

### Testing

Status: appropriate

The regression test exercises a streamed response containing narration plus a tool call and exactly verifies that both parts reach the next step; existing metadata and tool-loop coverage also passes on Node and Edge.

Concerns: none.

### Verification

Inspected git diff origin/main...HEAD and the surrounding prompt-building and core toResponseMessages behavior. Ran the full @ai-sdk/workflow test suite on Node and Edge (156 tests each), package type-check, repository type-check:full, targeted formatting/lint checks, and git diff --check; all passed. The credentialed Anthropic reproduction was not rerun, but the deterministic mock regression directly covers the reported path.

PR #17133, opened 2026-07-13
fix(ai): remove orphaned reasoning parts in pruneMessages

xianjianlf2:
## Problem
`pruneMessages` removes tool-call parts but leaves the preceding reasoning/thinking block, producing assistant messages that contain only reasoning parts. Anthropic (and the gateway) reject these as malformed ("no non-reasoning content block") (#13430).

## Fix
Extend the default `emptyMessages: 'remove'` filter so an assistant message whose remaining content is entirely reasoning parts is treated as empty and removed — matching the reporter's expected behavior. With `emptyMessages: 'keep'` such messages are still retained.

## Testing
Added two tests in `packages/ai/src/generate-text/prune-messages.test.ts` (red before, green after) and updated one pre-existing snapshot that had been encoding the buggy reasoning-only message. Changeset included.

Closes #13430

### DIFF
--- a/.changeset/pruneMessages-orphaned-reasoning.md
+++ b/.changeset/pruneMessages-orphaned-reasoning.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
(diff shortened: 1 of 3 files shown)


xianjianlf2:
Checked the red status here. The only failing check is the external `Vercel` deployment authorization gate; it links to Vercel asking a Vercel Labs team member to authorize the deployment.

The other checks shown on the PR are green:
- Socket Security: Project Report
- Vercel Agent Review

So I do not see a code/test failure to fix from the current checks.

PR #17138, opened 2026-07-13
fix: Gemini ignores Zod tool parameter descriptions and modifies tool input values

### LINKED ISSUE #6855: Gemini tool parameter descriptions aren't working.
### Description

I have a tool and one thing I'm noticing is that no matter what I put in the userInput field, it doesn't affect what comes through

    parameters: z.object({
      userInput: z.string().describe(`The user's prompt as is, do not modify it or infer. Act as a pirate`)
    }),

All that comes thorugh is a summary of what the user said. When I try it with vertex ai directly, it works.

### AI SDK Version

I'm on the latest versions of ai, @ai-sdk/google and google-vertex.

lgrammel:
## Background

Gemini could ignore Zod tool parameter descriptions and alter values that were required to copy the user prompt verbatim.

## Summary

The Google provider now sends function inputs through parametersJsonSchema, preserving the original JSON Schema instead of converting it to the legacy OpenAPI parameters format.

## Testing

Added prepareTools regression coverage for preserved parameter descriptions and updated language-model request snapshots.

## End-to-end Validation

- `issue-6855-google-zod-tool-description.ts`: ran `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-6855-google-zod-tool-description.ts`; live Gemini 2.5 Flash requests used `parametersJsonSchema` and returned the prompt verbatim.

## Related Issues

Fixes #6855

Co-authored-by: Godrules500 <318063+Godrules500@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: partially-addresses

The PR changes tools to parametersJsonSchema, but live verification still reproduced the shortened tool value in 2 of 5 calls; descriptions were already preserved by the previous conversion.

Concerns:
- Five PR-path calls produced three verbatim values and two shortened values.
- Five equivalent calls rewritten to the pre-PR parameters format produced the same three-verbatim/two-shortened result, so the change did not demonstrate a causal improvement.
- The reproduction retries up to five times and reports only the final result, masking intermittent occurrences of the reported failure.

### Side effects

Risk: medium

The wire format changes for every function tool and for the shared Vertex path, forwarding the complete schema instead of the converter's filtered representation.

Concerns:
- Schemas containing keywords previously stripped by convertJSONSchemaToOpenAPISchema may now behave differently or be rejected by an endpoint.
- There is no Vertex-specific tool request compatibility test despite the shared implementation affecting Google Vertex.

### Performance

Risk: low

The change removes recursive conversion but can send larger schemas because previously discarded metadata and constraints are now included.

Concerns: none.

### Backwards compatibility

Risk: none

The change alters outgoing requests and model behavior but does not read, migrate, or modify stored data.

Concerns: none.

### Security

Risk: none

No new parsing, execution, authentication, or privilege boundary is introduced.

Concerns: none.

### Testing

Status: needs-more

Request-shape tests pass, but the end-to-end check tolerates and conceals the reported failure, and compatibility coverage is missing for the broad wire-format change.

Concerns:
- The live reproduction's retry loop is not a reliable regression assertion for the user-visible behavior.
- No test establishes that parametersJsonSchema improves behavior relative to the previous parameters payload.
- No focused Vertex tool request test covers the changed shared code path.

### Verification

Inspected origin/main...HEAD and both commits; ran @ai-sdk/google Node tests (690 passed), Edge tests (690 passed), package type-check, @ai-sdk/google-vertex Node tests (249 passed), and git diff --check. The checked-in reproduction passed only after two requests. A separate five-call PR-path run returned three verbatim and two shortened values; a five-call legacy-payload comparison produced the same distribution.

lgrammel:
This is the wrong approach; need to investigate whether this even makes sense.

n1ckoates:
Original issue closed

PR #17148, opened 2026-07-13
fix: tool-call-only assistant messages serialize with empty-string content instead of null

lgrammel:
## Background

Tool-call-only assistant messages used empty-string content, which could break subsequent structured tool calls with OpenAI-compatible providers.

## Summary

Changed OpenAI chat message conversion to serialize empty assistant text content as null.

## Testing

Updated converter regression coverage to assert null content for tool-call-only assistant messages.

## End-to-end Validation

- `issue-12389-openai-tool-call-content.ts`: ran `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-12389-openai-tool-call-content.ts` against the live OpenAI API; the request contained `content: null` and returned `File created.`.

## Related Issues

Fixes #12389

Co-authored-by: lbijeau <2281680+lbijeau@users.noreply.github.com>

### DIFF
--- a/.changeset/silly-tools-call.md
+++ b/.changeset/silly-tools-call.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): send null content for tool-only assistant messages
--- a/examples/ai-functions/package.json
+++ b/examples/ai-functions/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "@example/ai-functions",
+  "version": "0.0.0",
+  "private": true,
+  "dependencies": {
+    "@ai-sdk/openai": "workspace:*",
+    "ai": "workspace:*",
+    "zod": "3.25.76"
+  },
+  "devDependencies": {
+    "@types/node": "20.17.24",
+    "tsx": "4.19.2",
+    "typescript": "5.8.3"
+  }
+}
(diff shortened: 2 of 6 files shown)


lgrammel:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: fully-addresses

The tool-call-only case now serializes assistant content as null, directly correcting the reported request payload.

Concerns: none.

### Side effects

Risk: medium

The unconditional text || null conversion also changes empty assistant messages that have no tool calls from valid empty-string content to null.

Concerns:
- A focused probe produced {"role":"assistant","content":null} for both an empty text part and an empty content array with no tool_calls; OpenAI requires assistant content when tool_calls/function_call is absent, and this exact regression previously required follow-up fix #14950. 

### Performance

Risk: none

The one-expression serialization change adds no meaningful computation, allocation, or memory retention.

Concerns: none.

### Backwards compatibility

Risk: medium

Stored data is not mutated, but replaying existing conversations containing empty assistant messages without tool calls can now generate rejected OpenAI requests.

Concerns:
- Null should be selected only when toolCalls.length is greater than zero; otherwise the existing empty string must be preserved.

### Security

Risk: none

The change only alters message serialization and introduces no new input handling, credential exposure, or trust boundary.

Concerns: none.

### Testing

Status: needs-more

The tool-only regression assertion passes, but coverage omits the important empty-assistant-without-tools case affected by the same expression.

Concerns:
- Add a regression test asserting content remains an empty string when assistant text is empty and tool_calls is absent.

### Verification

Inspected the full origin/release-v5.0...HEAD diff; ran git diff --check; ran the @ai-sdk/openai Node suite with 569 tests passing and package type-check successfully; directly probed conversion and confirmed tool-only messages become null as intended, but empty no-tool assistant messages also incorrectly become null.

PR #17158, opened 2026-07-13
fix: reject empty OpenAI, Anthropic, and Replicate base URLs with a helpful error

### LINKED ISSUE #17156: createOpenAI({ baseURL: "" }) should fail with a helpful error
if `baseUrl` is an empty string, throw an AI SDK error with a helpful message. 

Right now it throws at the time of request with a message like `Failed to parse URL from /response`.

Also check if the OPENAI_BASE_URL env variable is set to an empty string.
Also check other provider factory methods for the same problem.

gr2m:
## Background

Empty base URL configuration previously allowed provider creation and later produced opaque request-time URL parsing errors, including when OPENAI_BASE_URL was empty.

## Summary

Added a shared validateBaseURL helper that throws InvalidArgumentError for empty or whitespace-only values and applied it during OpenAI, Anthropic, and Replicate provider creation without changing the semantics of withoutTrailingSlash or Google Vertex default URL construction.

## Testing

Added utility tests for validation and URL normalization, provider regressions for explicit OpenAI, OPENAI_BASE_URL, Anthropic, and Replicate configurations, and Google Vertex xAI and MaaS coverage using the real shared URL utility for omitted and empty defaults.

## End-to-end Validation

- `issue-17156-empty-base-url.ts` — ran `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17156-empty-base-url.ts`; all invalid configurations produced helpful factory-time AI SDK errors and the live OpenAI call returned the expected text.

## Related Issues

Fixes #17156

Co-authored-by: gr2m <39992+gr2m@users.noreply.github.com>

### DIFF
--- a/packages/provider-utils/src/without-trailing-slash.ts
+++ b/packages/provider-utils/src/without-trailing-slash.ts
@@ -1,3 +1,12 @@
+import { InvalidArgumentError } from '@ai-sdk/provider';
+
 export function withoutTrailingSlash(url: string | undefined) {
+  if (url?.trim() === '') {
+    throw new InvalidArgumentError({
+      argument: 'baseURL',
+      message: 'baseURL must be a non-empty string.',
+    });
+  }
+
   return url?.replace(/\/$/, '');
 }
--- a/packages/provider-utils/src/without-trailing-slash.ts
+++ b/packages/provider-utils/src/without-trailing-slash.ts
@@ -1,12 +1,3 @@
-import { InvalidArgumentError } from '@ai-sdk/provider';
-
 export function withoutTrailingSlash(url: string | undefined) {
-  if (url?.trim() === '') {
-    throw new InvalidArgumentError({
-      argument: 'baseURL',
-      message: 'baseURL must be a non-empty string.',
-    });
-  }
-
   return url?.replace(/\/$/, '');
 }
--- a/.changeset/tidy-ravens-check.md
+++ b/.changeset/tidy-ravens-check.md
@@ -5,4 +5,5 @@
 '@ai-sdk/replicate': patch
 ---
 
-Reject empty provider base URLs with a helpful AI SDK invalid argument error.
+Reject empty OpenAI, Anthropic, and Replicate base URLs with a helpful AI SDK
+invalid argument error.
--- a/.changeset/tidy-ravens-check.md
+++ b/.changeset/tidy-ravens-check.md
@@ -1,0 +1,8 @@
+---
+'@ai-sdk/provider-utils': patch
+'@ai-sdk/openai': patch
+'@ai-sdk/anthropic': patch
+'@ai-sdk/replicate': patch
+---
+
+Reject empty provider base URLs with a helpful AI SDK invalid argument error.
--- a/examples/ai-functions/src/reproduction/issue-17156-empty-base-url.ts
+++ b/examples/ai-functions/src/reproduction/issue-17156-empty-base-url.ts
@@ -121,15 +121,6 @@ async function main() {
   });
 
   const primaryResults = [explicitOpenAI, environmentOpenAI];
-  const output = {
-    expected:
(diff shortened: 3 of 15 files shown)


packages/provider-utils/src/without-trailing-slash.ts
```diff
@@ -1,3 +1,12 @@
+import { InvalidArgumentError } from '@ai-sdk/provider';
+
 export function withoutTrailingSlash(url: string | undefined) {
+  if (url?.trim() === '') {
+    throw new InvalidArgumentError({
+      argument: 'baseURL',
+      message: 'baseURL must be a non-empty string.',
+    });
+  }
+
```
gr2m: This should not be implemented in `withoutTrailingSlash`. Implement a separate method instead.

gr2m:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: partially-addresses

OpenAI explicit and OPENAI_BASE_URL empty values now fail early with AI_InvalidArgumentError, but the blanket utility change does not correctly account for all provider factory semantics.

Concerns:
- Google Vertex xAI and MaaS intentionally use an empty value to construct their default base URL; the new shared behavior instead throws.

### Side effects

Risk: high

Changing the public withoutTrailingSlash utility affects roughly 40 provider call sites and breaks Google Vertex xAI and MaaS when baseURL is omitted or empty.

Concerns:
- packages/google-vertex/src/xai/google-vertex-xai-provider.ts:167 and packages/google-vertex/src/maas/google-vertex-maas-provider.ts:96 pass an empty string as a fallback sentinel.
- Direct runtime verification showed both providers throwing AI_InvalidArgumentError with omitted baseURL, preventing their normal constructed defaults.
- The implementation does not follow the review request to introduce a separate validation method instead of changing withoutTrailingSlash.

### Performance

Risk: low

The added trim check performs negligible work during URL normalization and introduces no material memory consumption.

Concerns: none.

### Backwards compatibility

Risk: none

The change does not read, migrate, or modify stored data; runtime configuration compatibility concerns are covered under side effects.

Concerns: none.

### Security

Risk: none

The validation introduces no apparent security risk and prevents malformed empty URL configurations from reaching requests.

Concerns: none.

### Testing

Status: needs-more

Added OpenAI, Anthropic, Replicate, and utility tests pass, but coverage misses regressions across the many consumers of the modified shared utility.

Concerns:
- Google Vertex tests mock withoutTrailingSlash with its old empty-string behavior, so the package's 249 passing Node tests mask the production regression.
- Tests should exercise the real shared utility and preserve omitted/default base URL behavior for providers that use empty strings as fallback sentinels.

### Verification

Inspected git diff origin/main...HEAD and all withoutTrailingSlash call sites; ran Node test suites for provider-utils (660 tests), OpenAI (774), Anthropic (444), Replicate (65), and Google Vertex (249), all passing. A direct import of built Google Vertex xAI and MaaS providers demonstrated that both omitted and empty baseURL now throw AI_InvalidArgumentError instead of constructing their default URLs.

packages/provider-utils/src/without-trailing-slash.ts
gr2m: Moved the validation into a dedicated `validateBaseURL` helper and restored `withoutTrailingSlash` to normalization-only behavior. OpenAI, Anthropic, and Replicate now invoke validation explicitly.

gr2m:
Addressed [PR comment 4961430972](https://github.com/vercel/ai/pull/17158#issuecomment-4961430972).

> ## Bugfix review
> 
> Outcome: changes-required
> 
> ### Fixes issue
> 
> Status: partially-addresses
> 
> ... 53 more line(s) omitted

Restored the original `withoutTrailingSlash` semantics and scoped empty-value validation to the affected OpenAI, Anthropic, and Replicate factories. Google Vertex xAI and MaaS tests now use the real shared utility, with coverage confirming that omitted and explicitly empty base URLs still construct their defaults.

aayush-kapoor (approved):

PR #17176, opened 2026-07-13
feat: support ordered OpenAI additional tools after normal tool results

### LINKED ISSUE #17155: @ai-sdk/openai: support ordered additional_tools after a normal tool result
#### Request

Please let `@ai-sdk/openai` serialize an OpenAI Responses [`additional_tools`](https://developers.openai.com/api/docs/guides/tools-tool-search#add-tools-at-a-specific-point-in-the-input) item at an exact point in `ModelMessage` history, including immediately after a normal `tool-result`.

This is the concrete application case behind draft PR #17097. The proposed `providerOptions.openai.additionalTools` direction would work for eve if it can be attached to a tool-result part and produces a sibling `additional_tools` input item after the normal `function_call_output`.

eve does not need a provider-neutral abstraction for this. An OpenAI-specific option on the existing tool-result part is sufficient.

This request is narrower than adding tool search. AI SDK already supports OpenAI's hosted and client-executed [`tool_search`](https://ai-sdk.dev/providers/ai-sdk-providers/openai#tool-search). The missing operation is preserving a normal tool result while loading the tools discovered by that result:

```text
R = the ordinary connection_search result
T = the function definitions discovered by that search

input = [
  ...history,
  function_call("connection_search"),
  function_call_output(R),
  additional_tools(T),
]
```

OpenAI makes tools available only after the `additional_tools` item's position and requires that position to be preserved when history is replayed. OpenAI's [tool-search documentation](https://developers.openai.com/api/docs/guides/tools-tool-search) also states that injecting newly discovered tools at the end of the context preserves the model's cache.

#### Why eve needs both `R` and `T`

[eve](https://github.com/vercel/eve) exposes `connection_search` as a normal function tool. Its available connection tools depend on runtime state such as the current tenant, installed connections, authorization, and the external tool registry.

The result is richer than a list of OpenAI function definitions:

```ts
type ConnectionSearchResult = Array<{
  connection: string;
  description: string;
  tool?: string;
  qualifiedName?: string;
  inputSchema?: object;
  outputSchema?: object;
  needsAuthorization?: boolean;
  error?: string;
}>;
```

That result tells the model which tools matched, but it also preserves connection summaries, partial failures, and authorization state. eve must send it back as the ordinary `connection_search` result.

For each successful match, eve also creates a normal AI SDK tool with its input schema and a local `execute` callback. For example, a result for Linear may create `linear__list_issues`. On the next model step, that function must be available to OpenAI as a structured tool, validated by AI SDK, and executed by eve when called.

The desired runtime path is:

1. The request starts with a stable tool list that includes `connection_search`, but not every possible connection tool schema.
2. The model calls `connection_search`.
3. eve performs the external lookup and returns `R`.
4. eve builds the discovered AI SDK tools and their local executors.
5. The next request replays `function_call_output(R)` and immediately appends `additional_tools(T)`.
6. The model calls a discovered function such as `linear__list_issues`.
7. AI SDK validates the call against the full `tools` map, then eve's local callback executes it.
8. On later requests, the same `additional_tools` item remains at the same history position.

#### Why client-executed `tool_search` is not equivalent

`openai.tools.toolSearch({ execution: "client" })` is a good fit when the discovery result is the tool list itself. Its current output contract is only `{ tools: Array<JSONObject> }`, and the OpenAI adapter serializes that result as `tool_search_output`:

- [`toolSearchOutputSchema`](https://github.com/vercel/ai/blob/703e0bf32f3090682405d1f2a8e653f7656df22a/packages/openai/src/tool/tool-search.ts#L29-L37)
- [`tool_search_output` conversion](https://github.com/vercel/ai/blob/703e0bf32f3090682405d1f2a8e653f7656df22a/packages/openai/src/responses/convert-to-openai-responses-input.ts#L791-L804)

Replacing eve's normal `connection_search` result with that output would discard the summaries, errors, and authorization state in `R`. Returning `R` in a separate text message would change the protocol and would no longer preserve the original tool result as one ordered history item.

Adding newly discovered tools to the request-level `tools` array on every later step is also not equivalent. It changes the early request prefix as the tool set grows. OpenAI's [`additional_tools` guidance](https://developers.openai.com/api/docs/guides/tools-tool-search#add-tools-at-a-specific-point-in-the-input) exists for applications that load tools outside the normal tool-search flow and need to preserve insertion order. OpenAI's [prompt-caching guidance](https://developers.openai.com/cookbook/examples/prompt_caching_201#42-stabilize-the-prefix) recommends appending runtime changes instead of modifying the existing prefix.

#### How eve would consume this

An illustrative AI SDK call would look like this. The exact option name is not important, but the placement and serialization are:

```ts
const tools = {
  connection_search: connectionSearchTool,
  linear__list_issues: tool({
    description: "List Linear issues",
    inputSchema: linearListIssuesSchema,
    execute: executeLinearListIssues,
  }),
};

await streamText({
  model: openai.responses("gpt-5.4"),

  // Keep discovered tools in the full ToolSet so AI SDK can validate and
  // execute their calls, but do not serialize them in the top-level tools list.
  tools,
  activeTools: ["connection_search"],

  messages: [
    ...history,
    {
      role: "tool",
      content: [
        {
          type: "tool-result",
          toolCallId: searchCallId,
          toolName: "connection_search",
          output: { type: "json", value: connectionSearchResult },
          providerOptions: {
            openai: {
              additionalTools: [linearListIssuesOpenAIDefinition],
            },
          },
        },
      ],
    },
  ],
});
```

AI SDK already separates these two concerns: `activeTools` determines the tools prepared for the provider call, while tool-call parsing and local execution use the full `tools` map. See the current [`generateText` provider preparation](https://github.com/vercel/ai/blob/b162ae48676fe9e7b3880b691cbd60b58ed179cb/packages/ai/src/generate-text/generate-text.ts#L854-L870), [`parseToolCall` invocation](https://github.com/vercel/ai/blob/b162ae48676fe9e7b3880b691cbd60b58ed179cb/packages/ai/src/generate-text/generate-text.ts#L974-L989), and [`executeTools` invocation](https://github.com/vercel/ai/blob/b162ae48676fe9e7b3880b691cbd60b58ed179cb/packages/ai/src/generate-text/generate-text.ts#L1177-L1190). That gives eve a way to keep validation and execution without duplicating the discovered definitions in the request-level tools list.

The OpenAI adapter would serialize the tool-result part above as:

```json
[
  {
    "type": "function_call_output",
    "call_id": "call_connection_search",
    "output": "[...the complete connection_search result...]"
  },
  {
    "type": "additional_tools",
    "role": "developer",
    "tools": [
      {
        "type": "function",
        "name": "linear__list_issues",
        "description": "List Linear issues",
        "parameters": { "type": "object" }
      }
    ]
  }
]
```

#### Acceptance criteria

- A tool-result part can request an OpenAI `additional_tools` item without replacing or changing its normal `function_call_output`.
- The adapter emits `additional_tools` immediately after that result and preserves the same order when messages are replayed.
- The loaded functions may remain in the full AI SDK `tools` map with local `execute` callbacks while `activeTools` omits them from the request-level tools list.
- A subsequent function call to one of those tools passes AI SDK validation and invokes its local executor.
- The behavior works in both `generateText` and `streamText`.
- Direct `@ai-sdk/openai` support is the immediate need. The same message representation should eventually work through AI Gateway when it routes to OpenAI, or the Gateway limitation should be documented separately.

Related eve work:

- https://github.com/vercel/eve/issues/716
- https://github.com/vercel/eve/pull/717

### AI SDK Version

- `ai@7.0.22`
- `@ai-sdk/openai@4.0.11`
- eve currently has `ai@7.0.0` and `@ai-sdk/openai@4.0.0` installed; the missing conversion is the same in those versions.

### Code of Conduct

gr2m:
## Background

Applications need to preserve a normal tool result while loading newly discovered OpenAI function tools at that exact history position, enabling prompt caching and local execution without duplicating tools in the request-level list.

## Summary

Added OpenAI Responses serialization for providerOptions.openai.additionalTools immediately after function_call_output, preserved replay ordering, enabled streamText validation and execution against the full tools map, and added documentation and patch changesets for @ai-sdk/openai and ai.

## Testing

Added converter tests for sibling additional_tools serialization and replay stability, plus a streamText regression test proving inactive dynamically loaded tools remain locally validated and executable.

## End-to-end Validation

- Ran the issue #17155 reproduction against the live OpenAI gpt-5.4 API; generateText and streamText both sent function_call_output followed by additional_tools, omitted get_weather from request-level tools, and invoked its local executor successfully.

## Related Issues

Fixes #17155

Co-authored-by: ruiconti <1834568+ruiconti@users.noreply.github.com>

### DIFF
--- a/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
@@ -1,3 +1,4 @@
+import type { LanguageModelV4Prompt } from '@ai-sdk/provider';
 import type { ToolNameMapping } from '../../../provider-utils/src/create-tool-name-mapping';
 import { convertToOpenAIResponsesInput } from './convert-to-openai-responses-input';
 import { describe, it, expect } from 'vitest';
@@ -5557,6 +5558,140 @@ describe('convertToOpenAIResponsesInput', () => {
           },
         ]
       `);
+    });
+
+    it('should emit additional tools immediately after a normal tool result', async () => {
+      const result = await convertToOpenAIResponsesInput({
+        toolNameMapping: testToolNameMapping,
+        prompt: [
+          {
+            role: 'tool',
+            content: [
+              {
+                type: 'tool-result',
+                toolCallId: 'call_search',
+                toolName: 'connection_search',
+                output: {
+                  type: 'json',
+                  value: {
+                    connection: 'linear',
+                    qualifiedName: 'linear__list_issues',
+                  },
+                },
+                providerOptions: {
+                  openai: {
+                    additionalTools: [
+                      {
+                        type: 'function',
+                        name: 'linear__list_issues',
+                        description: 'List Linear issues',
+                        parameters: { type: 'object' },
+                        strict: true,
+                      },
+                    ],
+                  },
+                },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call_other',
+                toolName: 'other_tool',
+                output: {
+                  type: 'text',
+                  value: 'other result',
+                },
+              },
+            ],
+          },
+        ],
+        systemMessageMode: 'system',
+        providerOptionsName: 'openai',
+        store: true,
+      });
+
+      expect(result.input).toEqual([
+        {
+          type: 'function_call_output',
+          call_id: 'call_search',
+          output:
+            '{"connection":"linear","qualifiedName":"linear__list_issues"}',
+        },
+        {
+          type: 'additional_tools',
+          role: 'developer',
+          tools: [
+            {
+              type: 'function',
+              name: 'linear__list_issues',
+              description: 'List Linear issues',
+              parameters: { type: 'object' },
+              strict: true,
+            },
+          ],
+        },
+        {
+          type: 'function_call_output',
+          call_id: 'call_other',
+          output: 'other result',
+        },
+      ]);
+      expect(result.warnings).toEqual([]);
+    });
+
+    it('should preserve additional tool placement when messages are replayed', async () => {
+      const prompt = [
+        {
+          role: 'tool',
+          content: [
+            {
+              type: 'tool-result',
+              toolCallId: 'call_search',
+              toolName: 'connection_search',
+              output: {
+                type: 'json',
+                value: [{ connection: 'linear' }],
+              },
+              providerOptions: {
+                openai: {
+                  additionalTools: [
+                    {
+                      type: 'function',
+                      name: 'linear__list_issues',
+                      parameters: { type: 'object' },
+                    },
+                  ],
+                },
+              },
+            },
+          ],
+        },
+        {
+          role: 'assistant',
+          content: [{ type: 'text', text: 'Continue.' }],
+        },
+      ] satisfies LanguageModelV4Prompt;
+
+      const firstConversion = await convertToOpenAIResponsesInput({
+        toolNameMapping: testToolNameMapping,
+        prompt,
+        systemMessageMode: 'system',
+        providerOptionsName: 'openai',
+        store: true,
+      });
+      const replayedConversion = await convertToOpenAIResponsesInput({
+        toolNameMapping: testToolNameMapping,
+        prompt,
+        systemMessageMode: 'system',
+        providerOptionsName: 'openai',
+        store: true,
+      });
+
+      expect(replayedConversion.input).toEqual(firstConversion.input);
+      expect(
+        replayedConversion.input.map(item =>
+          'type' in item ? item.type : undefined,
+        ),
+      ).toEqual(['function_call_output', 'additional_tools', undefined]);
     });
   });
 });
--- a/.changeset/tidy-tools-appear.md
+++ b/.changeset/tidy-tools-appear.md
@@ -1,0 +1,6 @@
+---
+'@ai-sdk/openai': patch
+'ai': patch
+---
+
+feat(provider/openai): support ordered additional tools after function results and execute dynamically loaded tools in streamText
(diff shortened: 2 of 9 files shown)


packages/openai/src/responses/convert-to-openai-responses-input.test.ts
```diff
@@ -5558,5 +5559,139 @@ describe('convertToOpenAIResponsesInput', () => {
...
+              strict: true,
+            },
+          ],
+        },
+        {
+          type: 'function_call_output',
+          call_id: 'call_other',
+          output: 'other result',
+        },
+      ]);
```
gr2m, line 5637: use inline snapshot

gr2m (commented):
Review notes

The packages/ai change is required for streamText to satisfy this feature end to end. Provider serialization alone leaves a tool introduced through additional_tools invalid locally. Making streamText align with generateText is the right direction.

Suggested follow-ups:

- Make the two tool roles explicit in the internal call boundary: a full tool registry for parsing, repair, metadata, and execution, and a prepared active subset for the provider request. _internal.preparedTools currently carries a meaningful production invariant; a non-reexported helper with names such as toolRegistry and languageModelTools would make that easier to reason about.

- Clarify the activeTools contract. The new regression test proves that any provider-emitted inactive tool can be validated and executed, even without showing that it was introduced by additional_tools. This changes behavior for every provider. It matches generateText, so activeTools may be intended as request serialization control rather than an execution allowlist; if so, document that it is not an authorization boundary. Otherwise, use an explicit executable/available tools allowlist.

- Add one mocked OpenAI Responses integration test that proves the complete path together: omit the discovered tool from request-level tools, emit function_call_output followed immediately by additional_tools, receive a call to the discovered tool, and execute it locally through streamText. The current converter and core tests prove the pieces independently but not the full contract.

- The replay test currently converts the same immutable prompt twice. A later-turn snapshot or a real response-message round trip would verify replay placement more meaningfully.

lgrammel (changes requested):
The `streamText` changes without corresponding `generateText` changes look suspicious. Ideally this would work exclusively through provider options and provider metadata and not change the `ai` package. It also needs to be tested in multi-turn scenarios end-to-end.

PR #17201, opened 2026-07-14
fix: UI message validation rejects persisted assistant responses with empty parts

lgrammel:
## Background

Persisted chats could become unloadable when an errored assistant response was saved with an empty parts array.

## Summary

Allowed assistant UI messages to have empty parts while retaining non-empty parts validation for user and system messages.

## Testing

Added regression coverage for validateUIMessages and safeValidateUIMessages using chats ending with an empty assistant response.

## End-to-end Validation

- `pnpm --filter ai build` followed by an inline `pnpm tsx` persisted-chat validation returned `{"success":true,"messageCount":2}`.

## Related Issues

Fixes #11053

Co-authored-by: davidroeca <6155705+davidroeca@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The schema now accepts empty parts only for assistant messages, allowing the persisted errored response described in #11053 to load through both validation APIs.

Concerns: none.

### Side effects

Risk: low

The validation relaxation is narrowly role-scoped; existing tests still confirm empty user-message parts are rejected, and the same condition also preserves rejection for system messages.

Concerns: none.

### Performance

Risk: low

The change adds one constant-time role and length check per message during validation, with no additional retained data or meaningful complexity increase.

Concerns: none.

### Backwards compatibility

Risk: none

Stored messages are not mutated; previously valid data remains valid, while persisted assistant messages with empty parts become loadable.

Concerns: none.

### Security

Risk: low

Only the assistant-parts minimum-length rule is relaxed; message roles, IDs, parts arrays, and all non-empty part structures continue to be schema-validated.

Concerns: none.

### Testing

Status: appropriate

Regression tests cover the reported persisted-chat shape through validateUIMessages and safeValidateUIMessages while retaining coverage that empty user parts fail.

Concerns: none.

### Verification

Inspected origin/main...HEAD and the original reproduction; ran the focused validation suite in Node and Edge configurations with all 55 tests passing in each, ran the ai package type-check and targeted formatting/lint checks successfully, and manually confirmed empty assistant messages validate while empty system messages fail and downstream model conversion safely omits an empty assistant message.

PR #17202, opened 2026-07-14
fix: tool input callbacks can run without the expected lifecycle start event

lgrammel:
## Background

Complete tool calls could trigger onInputAvailable without first triggering onInputStart, forcing consumers to handle an inconsistent lifecycle.

## Summary

generateText now invokes onInputStart before onInputAvailable for complete tool calls and documents the non-streaming lifecycle behavior.

## Testing

Updated the existing generateText callback-order regression test to cover onInputStart preceding onInputAvailable.

## End-to-end Validation

- `pnpm --filter ai build` followed by `pnpm tsx -e <generateText lifecycle assertion>`: observed `[ 'onInputStart', 'onInputAvailable' ]` in order.

## Related Issues

Fixes #11043

Co-authored-by: SamyPesse <845425+SamyPesse@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: partially-addresses

The generateText runtime path now awaits onInputStart before onInputAvailable, fixing the reproduced callback order, but the published documentation still explicitly says onInputStart is never called by generateText.

Concerns:
- Update content/docs/03-ai-sdk-core/15-tools-and-tool-calling.mdx and the tool/dynamic-tool reference pages, which still describe onInputStart as streaming-only and contradict the new behavior.

### Side effects

Risk: low

The intended behavior invokes one additional user callback for each valid complete generateText tool call; shared tool definitions may therefore perform new side effects or propagate callback failures, but this is inherent to the requested lifecycle correction.

Concerns: none.

### Performance

Risk: low

No new retained state or allocations of concern are introduced; generateText now awaits one additional configured callback, so callback runtime can add latency.

Concerns: none.

### Backwards compatibility

Risk: none

No stored-data formats, serialization paths, or persisted values are changed.

Concerns: none.

### Security

Risk: none

The change only invokes an existing callback with the same execution context fields and introduces no new input, network, credential, or parsing behavior.

Concerns: none.

### Testing

Status: appropriate

The updated regression test would fail on the merge base and verifies onInputStart precedes onInputAvailable with the expected options in both Node and Edge configurations.

Concerns: none.

### Verification

Inspected git diff origin/main...HEAD and the generateText and streaming callback paths; confirmed contradictory published documentation via repository search. The focused callback-order test passed under Node and Edge, the broader ai Node and Edge suites exited successfully, ai and provider-utils type checks passed, git diff --check passed, and the worktree remained clean.

PR #17222, opened 2026-07-14
fix: Bedrock Claude Opus structured output fails after multi-step tool calls

### LINKED ISSUE #16662: @ai-sdk/amazon-bedrock: Output.object() + tools intermittently fails with AI_NoObjectGeneratedError on Claude Opus 4.7/4.8 after the strict fix (#16237)
## Summary

The fix in #16237 / #16289 (shipped in `@ai-sdk/amazon-bedrock@4.0.120` on the v4 line and `5.0.0-beta.87` on the v5 line) correctly stops sending `strict` on tool specs for `claude-opus-4-7` / `claude-opus-4-8`, which resolves the HTTP 400 (`tools.0.custom.strict: Extra inputs are not permitted`) from #16121.

However, it introduced a second-order regression: when `Output.object()` (or `Output.array()`) is combined with **tools** in `generateText` (multi-step / agentic use), a fraction of Opus 4.7/4.8 runs now fail with:

```
AI_NoObjectGeneratedError: No object generated: response did not match schema.
```

We observe this in two shapes, both occurring on the turn **after a tool result**:

- **(A)** The model produces a valid-looking JSON object as **plain assistant text** instead of calling the internal `"json"` response tool, so the parser has no schema-valid tool call to read.
- **(B)** The model produces **no productive generation on the post-tool turn** (empty / no `"json"` tool call), so the run never reaches a final structured-output step and terminates without an object.

## Environment

- `@ai-sdk/amazon-bedrock@4.0.120` (also present on `5.0.0-beta.87`)
- `ai@6.0.201`
- Models: `us.anthropic.claude-opus-4-8` (and `-4-7`)
- The provider posts to the Bedrock **Converse API** (`bedrock-runtime .../converse` / `converse-stream`), thinking/reasoning **not** enabled
- Call shape: `generateText({ model, messages, tools, output: Output.object({ schema }), stopWhen: stepCountIs(15) })`

## Root cause

In `bedrock-chat-language-model` (`doGenerate`), structured output (`responseFormat: { type: "json", schema }`) is implemented by synthesizing a tool named `"json"` and forcing a tool call via `toolChoice: { type: "required" }`:

```ts
const useNativeStructuredOutput =
  isAnthropicModel && isThinkingEnabled &&
  responseFormat?.type === "json" && responseFormat.schema != null;

const jsonResponseTool =
  responseFormat?.type === "json" && responseFormat.schema != null && !useNativeStructuredOutput
    ? { type: "function", name: "json", description: "Respond with a JSON object.", inputSchema: responseFormat.schema }
    : undefined;

const { toolConfig, ... } = await prepareTools({
  tools: jsonResponseTool ? [...(tools ?? []), jsonResponseTool] : tools,
  toolChoice: jsonResponseTool != null ? { type: "required" } : toolChoice,
  modelId: this.modelId,
});
```

Then `prepareTools` strips `strict` from **every** tool for Opus 4.7/4.8 — including the synthetic `"json"` response tool:

```ts
const supportsStrictOnTools =
  !modelId.includes('claude-opus-4-7') && !modelId.includes('claude-opus-4-8');

for (const tool of filteredFunctionTools) {
  bedrockTools.push({
    toolSpec: {
      name: tool.name,
      ...(tool.description?.trim() !== '' ? { description: tool.description } : {}),
      ...(tool.strict != null && supportsStrictOnTools ? { strict: tool.strict } : {}),
      inputSchema: { json: tool.inputSchema }
    }
  });
}
```

The regression chain:

1. `strict` is removed for Opus 4.8 → the 400 is fixed. ✅
2. But structured output is the forced `"json"` tool, which now also loses `strict`.
3. `toolChoice: { type: "required" }` (→ Bedrock `{ any: {} }`) only forces _some_ tool call, and the `"json"` tool is no longer `strict`-constrained. On the turn after a tool result, Opus 4.8 then either (A) emits the final answer as **plain assistant text** rather than calling `"json"`, or (B) produces no productive `"json"` call at all, so the loop never resolves into the structured output before `stopWhen: stepCountIs(maxSteps)` ends it.
4. Core finds no schema-valid `"json"` tool-call output and throws `AI_NoObjectGeneratedError: response did not match schema`.

The common thread: because structured output is a non-`strict` forced tool for these models, nothing reliably drives the post-tool turn into the schema — so the object either arrives via the wrong channel (A) or never arrives (B).

Notably, a native structured-output path already exists but is gated on thinking being enabled:

```ts
if (useNativeStructuredOutput) {
  bedrockOptions.additionalModelRequestFields = {
    ...bedrockOptions.additionalModelRequestFields,
    output_config: { ...prev, format: { type: 'json_schema', schema: responseFormat.schema } }
  };
}
```

## Reproduction

1. `@ai-sdk/amazon-bedrock@4.0.120`, `ai@6.0.201`, no reasoning/thinking.
2. `generateText` targeting `us.anthropic.claude-opus-4-8` with:
   - two real tools (e.g. a couple of read-only lookups), and
   - `output: Output.object({ schema })`, `stopWhen: stepCountIs(15)`.
3. Run over a batch. A minority of runs (a few %) throw `AI_NoObjectGeneratedError: No object generated: response did not match schema.` on the turn after a tool result.

Observed in production tracing, both shapes appear:

- **(A)** the post-tool LLM step returns a well-formed object as assistant **text** (e.g. `assistant: {"reasoning": "...", "requirements": [...]}`), yet the root call errors because it arrived as text, not as a `"json"` tool call.
- **(B)** the post-tool LLM step returns an empty / non-productive turn (no `"json"` tool call), and the run never reaches a final structured-output step.

## Constraint that shapes the fix

Per AWS docs, Bedrock rejects `strict` as a property of **any** entry in the `tools[]` array for Opus 4.7/4.8 (`tools.0.custom.strict: Extra inputs are not permitted`). This is exactly what #16237 fixed. So **any fix that re-adds `strict` to a `tools[]` entry — including the synthetic `"json"` response tool — is invalid** and will re-trigger the 400. The blanket `strict` strip must stay for everything that goes into `tools[]`.

The corollary: for these models, implementing structured output as a **forced tool** is fundamentally the wrong mechanism, because the `"json"` tool can't be `strict`-constrained and `toolChoice: required` still lets the model answer in prose. The fix must avoid the tool channel for structured output entirely.

## Suggested fixes (any subset; all respect the `strict`-on-tools prohibition)

1. **(Preferred direction) Use Converse's native structured output for Opus 4.7/4.8 instead of the forced `"json"` tool.** The provider posts to the Converse API, which supports structured output natively ([AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html)). Routing structured output through the native mechanism (instead of the synthesized `"json"` tool) keeps it out of the `tools[]` array entirely, so it never carries the forbidden `strict` field and the model cannot "escape" the tool into prose. The exact request-field shape / gating is best decided by the maintainers.

2. **Pin the terminal `toolChoice` to the `"json"` tool** instead of `{ type: "required" }`. `toolChoice` is a separate request field (not a `tools[]` property), so this does not reintroduce `strict`. Using `{ type: "tool", toolName: "json" }` forces the model into the output tool specifically rather than any tool / free text. Mitigation, not a structural fix — useful if (1) can't be adopted everywhere.

3. **Fallback: parse a text final message against the schema before throwing.** Pure client-side salvage in core (never sends `strict` anywhere). When `responseFormat.type === "json"` and the model returns text instead of calling `"json"`, attempt to parse/validate that text against `responseFormat.schema` before raising `AI_NoObjectGeneratedError`. Salvages the (frequent) case where the object is correct but arrived via the wrong channel.

## Impact

- `Output.object()` / `Output.array()` + tools on Opus 4.7/4.8 over Bedrock is intermittently unreliable (a few % failure rate in our batches), where before #16237 it was a hard 400. The fix traded a deterministic 400 for a non-deterministic `NoObjectGenerated`.
- Workarounds today: run the agent/tool loop with `generateText` (no `Output`), then do a separate `generateObject` with no tools; or enable reasoning to hit the native structured-output path; or downgrade the model for structured-output-with-tools flows.

Related: #16121, #16237, #16289.

aayush-kapoor:
## Background

Bedrock Claude Opus 4.7/4.8 could return malformed structured output after verbose multi-step tool calls, causing AI_NoObjectGeneratedError.

## Summary

For Opus 4.7/4.8 tool workflows, the provider now injects the JSON schema into the prompt instead of using an unreliable synthetic response tool, avoids unsupported native structured output fields, and extracts the complete JSON object from generated text.

## Testing

Updated the existing Bedrock chat language model tests to verify schema instruction injection, exclusion of the synthetic response tool, and normalization of fenced or trailing structured output.

## End-to-end Validation

- `pnpm tsx -e ...` verbose six-page Opus 4.8 batch: completed six tool rounds and all 12 tool results, then returned schema-valid output after a 96,123-input-token final step.

## Related Issues

Fixes #16662

Co-authored-by: igalmazor <218749803+igalmazor@users.noreply.github.com>


packages/amazon-bedrock/src/amazon-bedrock-chat-language-model.ts
```diff
@@ -1054,6 +1101,68 @@ export class AmazonBedrockChatLanguageModel implements LanguageModelV4 {
...
+        if (this.depth === 0) {
+          this.completed = true;
+        }
+      }
+    }
+
+    return result;
+  }
+}
+
```
aayush-kapoor, line 1165: not sure about this class and solution

lgrammel (approved):

PR #17230, opened 2026-07-14
fix: OpenAI-compatible chat SSE errors lose structured error fields

### LINKED ISSUE #13506: SSE stream error chunks drop error.code — only error.message is forwarded to consumers

lgrammel:
## Background

OpenAI-compatible SSE errors were exposed as message strings, discarding machine-readable codes needed for programmatic error handling.

## Summary

The chat stream transformer now forwards the complete parsed SSE error object instead of only its message.

## Testing

Updated the existing chat stream regression test and inline snapshot to verify that the error message and code are preserved.

## End-to-end Validation

- `pnpm tsx --eval <synthetic SSE fullStream validation>` — observed `{"message":"Context length exceeded","code":"CONTEXT_LENGTH_EXCEEDED"}`.
- `pnpm tsx --eval <live OpenAI-compatible streamText validation>` — OpenAI returned `OK.` with finish reason `stop`.

## Related Issues

Fixes #13506

Co-authored-by: StellaCheong59 <178243329+StellaCheong59@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The transformer now emits chunk.value.error rather than only chunk.value.error.message, preserving the reported code and all other fields parsed by the configured SSE error schema.

Concerns: none.

### Side effects

Risk: low

Chat SSE error values intentionally change from strings to structured objects; consumers that assumed a string may need adjustment, but this is the required behavior and matches the existing completion-stream implementation.

Concerns: none.

### Performance

Risk: none

The change forwards an already parsed object instead of one property from it, adding no parsing, buffering, or meaningful allocation.

Concerns: none.

### Backwards compatibility

Risk: none

No SDK-managed stored data or migration path is affected; only newly emitted runtime stream error values use the corrected structured shape.

Concerns: none.

### Security

Risk: none

The change only exposes provider error fields already validated and parsed by the existing error schema and introduces no new input, network, or credential handling.

Concerns: none.

### Testing

Status: appropriate

The regression test directly verifies preservation of message and code, and focused Node and Edge suites, package type-checking, formatting/lint checks, and a streamText.fullStream synthetic reproduction all passed.

Concerns: none.

### Verification

Inspected git diff origin/main...HEAD and related schemas/core forwarding behavior; ran the 93-test chat model suite under Node and Edge, @ai-sdk/openai-compatible type-check, targeted ultracite checks, git diff --check, and a synthetic streamText.fullStream validation that preserved message, code, type, and param.

PR #17231, opened 2026-07-14
[v5.0] fix: Google image thought signatures are omitted from streaming response history

### LINKED ISSUE #10660: Thought Signatures for Images are missing in Streaming
### Description

This PR https://github.com/vercel/ai/pull/10462 implements Google Gemini 3 thoughtSignatures for images in generateText. But there is no implementation for streamText. This breaks image generation if you want to refine your image. For example:

1. Prompt: Create an image of the moon
2. Prompt: Nice, but now make it cheese

The second prompt will fail as there are no thoughtSignatures in the stream that I can use to send back in the second stream.

For context: We use Gemini 3 Pro Preview like a regular LLM. So the user can simply chat with it on our platform as usual.

### AI SDK Version

- ai: 6.0.0-beta.122
- @ai-sdk/google: 3.0.0-beta.60

### Code of Conduct

lgrammel:
## Background

Gemini image thought signatures were omitted from streamText response history, preventing callers from preserving required context for follow-up image refinement.

## Summary

Added provider metadata support to streamed files and propagated Gemini image thought signatures through the provider, AI stream pipeline, and response history.

## Testing

Added regression coverage for Google streamed image signatures, AI file metadata forwarding, and response-message conversion.

## End-to-end Validation

- `pnpm --dir examples/ai-core exec tsx -e "<live streamText validation>"` — the live Gemini stream returned one JPEG image and its thought signature was preserved in response history.

## Related Issues

Fixes #10660

Co-authored-by: sean-nicholas <9257209+sean-nicholas@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: partially-addresses

The PR preserves image thought signatures through the inbound stream and response history, but the Google provider still omits them when that history is replayed.

Concerns:
- The assistant-file branch in convert-to-google-generative-ai-messages.ts emits only inlineData and ignores providerOptions.google.thoughtSignature, so follow-up requests do not actually send the preserved signature.
- The current two-turn flow can also still fail because the Google converter rejects non-PNG assistant images, while the reproduced Gemini stream returned JPEG.

### Side effects

Risk: low

The changes are additive and only expose optional provider metadata on streamed file parts; no material regression was found.

Concerns: none.

### Performance

Risk: low

The change forwards one additional metadata object reference per signed file, with negligible expected memory or runtime cost.

Concerns: none.

### Backwards compatibility

Risk: low

Existing stored messages remain valid; newly produced response histories may contain an additional optional providerOptions field on files.

Concerns: none.

### Security

Risk: low

No new execution, network, or trust boundary is introduced; the change forwards provider-supplied opaque metadata similarly to existing text and tool metadata.

Concerns: none.

### Testing

Status: needs-more

The added unit tests cover provider parsing, core forwarding, and response conversion, but not the required response-history-to-follow-up-request round trip.

Concerns:
- A regression test should replay the streamed response messages through the Google prompt converter and assert that the outbound inlineData part contains thoughtSignature.
- A two-request streamText test would have caught that the current outbound converter strips the signature and rejects the reproduced JPEG response.

### Verification

Inspected git diff origin/release-v5.0...HEAD and relevant stream, response, and Google conversion paths. The 25 focused AI tests and 103 Google tests passed; AI, Google, and provider type-checks passed. A focused converter probe showed a signed PNG response history becomes an outbound inlineData part without thoughtSignature, and a signed JPEG history throws the existing 'Only PNG images are supported in assistant messages' error.

PR #17239, opened 2026-07-14
[v6.0] reproduction: perf: ON² quadratic memmove in processUIMessageStream + DefaultStreamTextResult

aayush-kapoor:
## Bug

perf: O(N²) quadratic memmove in processUIMessageStream + DefaultStreamTextResult on long thinking-mode streams

## Reproduction

- Progressive reads of the cumulative `.text` reproduced quadratic behavior in `processUIMessageStream`: 5,000 deltas took 357.2 ms and 10,000 took 1,427.0 ms, a 3.99× increase when input doubled.
- The 10,000-delta, 2,000,000-character progressive-render scenario exceeded the expected 1,000 ms budget on Node.js 22.23.1.
- The issue's exact no-op `write` shape completed in 39.1 ms because it did not read and flatten cumulative text between deltas.
- `DefaultStreamTextResult` completed the equivalent text and reasoning streams in 139.7 ms and 115.1 ms respectively, so that path did not independently exhibit the reported slowdown.
- Runnable reproduction `examples/ai-functions/src/reproduction/issue-15670-quadratic-stream-accumulation.ts` validates the final 2,000,000-character output and fails on the reproduced performance regression.

## Related Issues

Reproduces #15670

### DIFF
--- a/examples/ai-functions/src/reproduction/issue-15670-quadratic-stream-accumulation.ts
+++ b/examples/ai-functions/src/reproduction/issue-15670-quadratic-stream-accumulation.ts
@@ -1,0 +1,208 @@
+import assert from 'node:assert/strict';
+import type { LanguageModelV3StreamPart } from '@ai-sdk/provider';
+import { streamText } from 'ai';
+import { MockLanguageModelV3 } from 'ai/test';
+import {
+  createStreamingUIMessageState,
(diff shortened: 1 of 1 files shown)


gr2m:
Closing in favor of retry #17241.

PR #17249, opened 2026-07-14
fix(react/mcp): harden MCP Apps metadata handling and host/iframe bridge

dnukumamras:
Hardens how `@ai-sdk/react` and `@ai-sdk/mcp` handle server-supplied MCP App resource metadata and the host/iframe JSON-RPC bridge.

## Changes

- Runtime-validate `_meta.ui` with a `zod` schema and drop malformed or non-string fields instead of casting.
- Gate iframe permissions deny-by-default via a new `sandbox.allowedPermissions` allowlist; a capability is granted only when both requested by the server and allowed by the host.
- Derive a concrete `postMessage` target origin from the sandbox URL (instead of `'*'`) and validate the origin of inbound messages.
- Validate inbound bridge params: limit `resources/read` to `ui://` resources and allow only `https`/`http`/`mailto` in `ui/open-link`.
- Add `fingerprintMCPAppResource` / `detectMCPAppResourceDrift` so hosts can pin a resource and detect later changes.

## Behavior

Happy path:

- Valid `_meta.ui` metadata passes through unchanged.
- The default sandbox config (same-origin proxy, `allow-same-origin`) keeps the existing handshake working.
- `resources/read` for `ui://` resources and `ui/open-link` with `https`/`mailto` continue to work.

Unhappy path:

- Malformed or non-string metadata fields are dropped, not rendered.
- Server-requested iframe permissions with no matching host allowlist entry are not granted.
- Messages from an unexpected origin are ignored.
- `resources/read` for non-`ui://` URIs and `ui/open-link` with other schemes are rejected.

## Tests

Unit tests added/updated for metadata validation, permission gating, origin checks, bridge param validation, and fingerprint drift.

## Notes

- Permission gating is deny-by-default: hosts relying on server-requested permissions must now opt in via `sandbox.allowedPermissions`. Docs under `content/` should be updated to match in a follow-up.

### DIFF
--- a/packages/mcp/src/tool/mcp-apps.ts
+++ b/packages/mcp/src/tool/mcp-apps.ts
@@ -1,5 +1,6 @@
 import { isJSONObject, type JSONObject } from '@ai-sdk/provider';
 import { convertBase64ToUint8Array } from '@ai-sdk/provider-utils';
+import * as z4 from 'zod/v4';
 import type { MCPClient } from './mcp-client';
 import type {
   ClientCapabilities,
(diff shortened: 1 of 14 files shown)


packages/mcp/src/tool/mcp-apps.ts
```diff
@@ -1,5 +1,6 @@
 import { isJSONObject, type JSONObject } from '@ai-sdk/provider';
 import { convertBase64ToUint8Array } from '@ai-sdk/provider-utils';
+import { z } from 'zod/v4';
```
gr2m: use `import * as z4 from "zod/v4";`

see https://github.com/vercel/ai/blob/main/contributing/zod.md#rules

gr2m (approved):

packages/mcp/src/tool/mcp-apps.ts
dnukumamras: Done, switched to `import * as z4 from 'zod/v4'` and updated the usages. Thanks for the pointer.

PR #17259, opened 2026-07-14
fix(provider/google): forward Vertex-only imageConfig options (personGeneration, prominentPeople, imageOutputOptions)

R-Taneja:
## Background

`providerOptions.google.imageConfig` is validated against a zod schema that only defines `aspectRatio` and `imageSize`. Zod object parsing strips unknown keys, so the Vertex-only `ImageConfig` fields — `personGeneration`, `prominentPeople`, and `imageOutputOptions` — are silently dropped from the request in `parseProviderOptions` before they reach the wire.

A user routing `gemini-3.1-flash-image` through Vertex reported that `imageConfig.prominentPeople` had no effect: an intentionally invalid value returned `200` with a generated image instead of Vertex's `400 INVALID_ARGUMENT`, because the field never left the SDK.

Verified against the published `@ai-sdk/google-vertex@5.0.19` with a request-capturing `fetch`:

```
input:    imageConfig: { aspectRatio: '1:1', prominentPeople: 'BLOCK_PROMINENT_PEOPLE' }
outgoing: generationConfig.imageConfig: {"aspectRatio":"1:1"}
```

## Summary

Adds the missing fields to the `imageConfig` provider-options schema, with values verified against the live Vertex `v1` API:

- `personGeneration`: `ALLOW_NONE | ALLOW_ADULT | ALLOW_ALL` (live Vertex rejects the `DONT_ALLOW` value that Imagen-style APIs use)
- `prominentPeople`: `PROMINENT_PEOPLE_UNSPECIFIED | ALLOW_PROMINENT_PEOPLE | BLOCK_PROMINENT_PEOPLE`
- `imageOutputOptions`: `{ mimeType?: 'image/jpeg' | 'image/png', compressionQuality?: number }`

The Gemini API rejects all three fields with `400 Unknown name` (verified live), so on non-Vertex Google providers they are dropped with a call warning instead of being sent — same pattern as `streamFunctionCallArguments` and `serviceTier`. This preserves the previous behavior for Gemini API callers while making the drop visible.

Also documents the new fields on the Google provider docs page.

## Verification

Built the workspace and ran a live end-to-end check against Vertex (`gemini-3.1-flash-image-preview`, location `global`) and the Gemini API:

```
1. outgoing imageConfig: {"aspectRatio":"1:1","prominentPeople":"BLOCK_PROMINENT_PEOPLE"}
2. valid prominentPeople: 200, image generated
3. gemini api: finished with 1 file(s), warnings: [{"type":"other","message":"'imageConfig.prominentPeople' is a Vertex AI option and is ignored with the current Google provider (google.generative-ai)."}]
4. bad value rejected client-side: AI_InvalidArgumentError: invalid google provider options
```

An invalid `prominentPeople` value now fails loudly (client-side schema error) instead of silently succeeding.

## Tests

- new: Vertex provider forwards `personGeneration`, `prominentPeople`, `imageOutputOptions` in `generationConfig.imageConfig` with no warnings
- new: Gemini API provider drops Vertex-only fields, still forwards `aspectRatio`, and emits a warning
- `pnpm test` in `packages/google` and `packages/google-vertex`: 691 + 250 passing (node + edge)
- `pnpm check` and `pnpm type-check:full` clean

### DIFF
--- a/content/providers/01-ai-sdk-providers/15-google.mdx
+++ b/content/providers/01-ai-sdk-providers/15-google.mdx
@@ -203,6 +203,24 @@ The following optional provider options are available for Google models:
     - 2K
     - 4K
 
+  - **personGeneration** _string_
+
+    Optional. Controls the generation of people in images. Available on Vertex AI only. Can be one of the following:
+    - `ALLOW_NONE`
+    - `ALLOW_ADULT`
+    - `ALLOW_ALL`
+
+  - **prominentPeople** _string_
+
+    Optional. Controls whether generation of prominent people (celebrities) is allowed. When set together with `personGeneration`, `personGeneration` takes precedence. Available on Vertex AI only. Can be one of the following:
+    - `PROMINENT_PEOPLE_UNSPECIFIED`
+    - `ALLOW_PROMINENT_PEOPLE`
+    - `BLOCK_PROMINENT_PEOPLE`
+
+  - **imageOutputOptions** _\{ mimeType?: 'image/jpeg' | 'image/png', compressionQuality?: number \}_
+
+    Optional. The image output format for generated images. Available on Vertex AI only.
+
 - **audioTimestamp** _boolean_
 
   Optional. Enables timestamp understanding for audio-only files.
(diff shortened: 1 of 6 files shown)


content/providers/01-ai-sdk-providers/15-google.mdx
```diff
@@ -203,6 +203,25 @@ The following optional provider options are available for Google models:
     - 2K
     - 4K
 
+  - **personGeneration** _string_
```
shaper: docs should also be added to `content/providers/01-ai-sdk-providers/16-google-vertex.mdx`

maybe only put them there, since they're vertex specific?

shaper (approved):

PR #17262, opened 2026-07-14
feat(elevenlabs): add realtime Scribe transcription

31Carlton7:
## Summary

- add realtime streaming transcription support for ElevenLabs Scribe v2 Realtime
- implement the AI SDK v4 `doStream` transcription interface over the ElevenLabs WebSocket API
- support PCM and μ-law audio, commit strategies, language detection, word timestamps, and realtime transcript events
- preserve language detection when timestamps are not exposed to the caller
- finalize cleanly when ElevenLabs closes after the committed transcript, including timestamp-optional responses
- document Scribe alongside the other streaming transcription models and add a playback-paced example

## Why

AI Gateway streaming transcription needs a provider-level implementation. The published ElevenLabs provider currently supports only request/response transcription, so Gateway cannot construct or run a Scribe realtime stream without this change.

This is intentionally scoped to Scribe realtime transcription. It does not add ElevenAgents or the general realtime agent/session API.

Companion Gateway PR: https://github.com/vercel/ai-gateway/pull/2642

## Validation

- `pnpm --filter @ai-sdk/elevenlabs test` — 30 Node tests and 30 Edge tests passed
- `pnpm --filter @ai-sdk/elevenlabs type-check`
- `pnpm --filter @ai-sdk/elevenlabs build`
- `pnpm --filter @example/ai-functions type-check`
- `pnpm exec ultracite check` on the changed TypeScript files
- full GitHub CI on Node 22, 24, and 26

## Manual verification

Verified end to end through the companion AI Gateway preview with live microphone audio. Partial and committed transcripts, timestamps, language detection, manual finalization, and clean provider close behavior all completed successfully.

### DIFF
--- a/packages/elevenlabs/src/elevenlabs-transcription-model.ts
+++ b/packages/elevenlabs/src/elevenlabs-transcription-model.ts
@@ -1,21 +1,71 @@
-import type { TranscriptionModelV4, SharedV4Warning } from '@ai-sdk/provider';
+import {
+  InvalidArgumentError,
+  UnsupportedFunctionalityError,
+  type Experimental_TranscriptionModelV4StreamOptions as TranscriptionModelV4StreamOptions,
+  type SharedV4Warning,
+  type TranscriptionModelV4,
+} from '@ai-sdk/provider';
 import {
   combineHeaders,
   convertBase64ToUint8Array,
+  convertToBase64,
+  connectToWebSocket,
   createJsonResponseHandler,
   mediaTypeToExtension,
   parseProviderOptions,
   postFormDataToApi,
+  safeParseJSON,
   serializeModelOptions,
+  toWebSocketUrl,
+  waitForWebSocketBufferDrain,
   WORKFLOW_SERIALIZE,
   WORKFLOW_DESERIALIZE,
+  type WebSocketConnection,
+  type WebSocketLike,
 } from '@ai-sdk/provider-utils';
 import { z } from 'zod/v4';
 import type { ElevenLabsConfig } from './elevenlabs-config';
 import { elevenlabsFailedResponseHandler } from './elevenlabs-error';
-import { elevenLabsTranscriptionModelOptionsSchema } from './elevenlabs-transcription-model-options';
+import {
+  elevenLabsTranscriptionModelOptionsSchema,
+  type ElevenLabsTranscriptionModelOptions,
+} from './elevenlabs-transcription-model-options';
 import type { ElevenLabsTranscriptionModelId } from './elevenlabs-transcription-options';
 import type { ElevenLabsTranscriptionAPITypes } from './elevenlabs-api-types';
+
+type ElevenLabsRealtimeTranscriptionEvent = {
+  message_type?: string;
+  session_id?: string;
+  text?: string;
+  language_code?: string | null;
+  words?: Array<{
+    text?: string;
+    start?: number;
+    end?: number;
+    type?: string;
+  }> | null;
+  error?: string;
+};
+
+const elevenLabsRealtimeErrorTypes = new Set([
+  'auth_error',
+  'chunk_size_exceeded',
+  'commit_throttled',
+  'error',
+  'input_error',
+  'insufficient_audio_activity',
+  'queue_overflow',
+  'quota_exceeded',
+  'rate_limited',
+  'resource_exhausted',
+  'session_time_limit_exceeded',
+  'transcriber_error',
+  'unaccepted_terms',
+]);
+
+function isRealtimeTranscriptionModelId(modelId: string): boolean {
+  return modelId === 'scribe_v2_realtime';
+}
 
 interface ElevenLabsTranscriptionModelConfig extends ElevenLabsConfig {
   _internal?: {
@@ -114,6 +164,12 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
   async doGenerate(
     options: Parameters<TranscriptionModelV4['doGenerate']>[0],
   ): Promise<Awaited<ReturnType<TranscriptionModelV4['doGenerate']>>> {
+    if (isRealtimeTranscriptionModelId(this.modelId)) {
+      throw new UnsupportedFunctionalityError({
+        functionality: `non-streaming transcription with ${this.modelId}`,
+      });
+    }
+
     const currentDate = this.config._internal?.currentDate?.() ?? new Date();
     const { formData, warnings } = await this.getArgs(options);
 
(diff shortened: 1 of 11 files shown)


shaper (commented):
tagged @kkdawkins to look tomorrow

kkdawkins (commented):
Nice work — this is closely aligned with the xai/cartesia/openai streaming transcription implementations, and the wire protocol matches the ElevenLabs AsyncAPI spec. I left four inline comments on the substantive issues. Smaller nits, none blocking:

- The `streaming` sub-options use `.optional()` while every batch field in the same schema uses `.nullish()`. A caller that JSON-serializes provider options (e.g. the Gateway) can send explicit `null`s, which would throw here. Prefer `.nullish()` for consistency.
- `secondary_languages` exists on the realtime endpoint but isn't exposed in the streaming options. Fine to scope out, but worth a follow-up.
- Word-level `speaker_id`/`logprob` from `committed_transcript_with_timestamps` are dropped. `includeRawChunks` covers it for now; consider surfacing via `providerMetadata` later.
- The AsyncAPI union still includes legacy `final_transcript`/`final_transcript_with_timestamps` variants. They're unhandled, so a server sending them would hang the stream until close. Probably theoretical, but a cheap case fallthrough would make it robust.
- Partials get `id: sessionId` while finals get `sessionId:N`, so a consumer can't correlate a partial with the final that supersedes it. Spec allows it, but a per-segment id on partials would be more useful.
- `finalTexts.join(' ')` can double-space if committed transcripts carry trailing whitespace; `.trim()` only fixes the ends.

packages/elevenlabs/src/elevenlabs-transcription-model.ts
```diff
@@ -155,6 +220,448 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
...
+
+              // When timestamps are requested, ElevenLabs follows this event
+              // with committed_transcript_with_timestamps for the same commit.
+              if (
+                endOfInput &&
+                committedEventCount > committedEventsAtEndOfInput
+              ) {
+                finalCommitEventCount = committedEventCount;
+                if (!expectDetailedCommit) finish();
+              }
```
kkdawkins, line 634: This treats the first `committed_transcript` after end-of-input as the final commit, but the server also commits on its own: VAD strategy, and the ~36s auto-commit even under `manual`. If a server-initiated commit is in flight when we send the final `commit: true` chunk, it lands here first, satisfies this check, and `finish()` closes the socket before the real final commit arrives — silently dropping the tail of the transcript.

The protocol has no commit correlation id, so an exact fix isn't possible. Options: after end-of-input, keep draining committed events for a short grace period before finishing, or accept the race and document it in this comment so the trade-off is explicit. Right now the code reads as if this case can't happen.

packages/elevenlabs/src/elevenlabs-transcription-model.ts
```diff
@@ -155,6 +220,448 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
...
+          }
+
+          if (
+            raw.message_type != null &&
+            elevenLabsRealtimeErrorTypes.has(raw.message_type)
+          ) {
+            finishWithError(
+              new Error(raw.error ?? 'ElevenLabs realtime transcription error'),
+            );
+            return;
```
kkdawkins, line 581: Routing every error type through `finishWithError` is right mid-stream, but the trailing empty `commit: true` chunk can itself trigger `commit_throttled`, `input_error`, or `insufficient_audio_activity` (e.g. VAD already committed everything, so there's nothing left to commit). In that case the session actually succeeded, and erroring here throws away a transcript we already emitted `transcript-final` parts for.

Worth checking what the server actually sends for an empty final commit. If it can error, consider: once `endOfInput` is set and we have at least one committed transcript, treat these late errors as a normal `finish()` (optionally with a warning) instead of failing the stream.

packages/elevenlabs/src/elevenlabs-transcription-model.ts
```diff
@@ -155,6 +220,448 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
...
+      streamingOptions?.filterBackgroundAudio === true &&
+      (streamingOptions.includeTimestamps === true ||
+        streamingOptions.includeLanguageDetection === true)
+    ) {
+      throw new InvalidArgumentError({
+        argument: 'providerOptions',
+        message:
+          'providerOptions.elevenlabs.streaming.filterBackgroundAudio cannot be combined with includeTimestamps or includeLanguageDetection',
+      });
+    }
```
kkdawkins, line 286: Where does this constraint come from? I couldn't find it in the ElevenLabs docs or the AsyncAPI spec for `/v1/speech-to-text/realtime`. If it's an observed server rejection rather than a documented rule, hard-throwing client-side means we keep blocking the combination even after ElevenLabs relaxes it — and we'd have no signal that they did. If it's undocumented, prefer passing the params through and letting the server reject. If it's confirmed, add the source in a comment here so the next person doesn't have to re-derive it.

packages/elevenlabs/src/elevenlabs-transcription-model.ts
```diff
@@ -155,6 +220,448 @@ export class ElevenLabsTranscriptionModel implements TranscriptionModelV4 {
...
+                finish();
+              }
+              break;
+            }
+          }
+        },
+        onProcessingError: finishWithError,
+        onSocketError: () => {
+          finishWithError(new Error('ElevenLabs realtime transcription error'));
+        },
```
kkdawkins, line 706: ElevenLabs auth is header-based (`xi-api-key`), and native WebSocket in browsers/Node/Deno/Bun can't send headers — so the most common failure a new user hits is a rejected handshake, which surfaces here as this generic message with no clue about the fix. The xai implementation handles the identical situation by explaining it in the error (see `onSocketError` in `packages/xai/src/xai-transcription-model.ts`). Do the same:

```suggestion
        onSocketError: () => {
          finishWithError(
            new Error(
              'ElevenLabs realtime transcription error.' +
                (webSocket == null
                  ? ' Note: the native WebSocket implementation in browsers,' +
                    ' Node.js, Deno, and Bun cannot send the xi-api-key' +
                    ' header required by ElevenLabs. Pass a header-capable' +
                    " WebSocket implementation (e.g. the 'ws' package) via" +
                    ' createElevenLabs({ webSocket }).'
                  : ''),
            ),
          );
        },
```

packages/elevenlabs/src/elevenlabs-transcription-model.ts
31Carlton7: https://elevenlabs.io/docs/api-reference/speech-to-text/v-1-speech-to-text-realtime

> Enable background speech filtering to reduce false activations from nearby conversations and ambient noise. When enabled without an explicit vad_threshold, a lower default threshold is applied. Cannot be combined with include_timestamps.

The source is now in a comment

kkdawkins (approved):

PR #17263, opened 2026-07-14
test(react/mcp): assert outer-frame permission delegation

dnukumamras:
Follow-up to #17249.

Adds a jsdom component test for the outer-frame permission delegation fix from #17249. Iframe Permissions Policy is delegated top-down, so the outer proxy iframe must carry the `allow` attribute for a feature to reach the inner app frame.

## Coverage

Happy path:

- Host allows `geolocation`, server requests `geolocation` and `camera`, so the outer iframe renders `allow="geolocation"` (the intersection).

Unhappy path:

- No host allowlist: no `allow` attribute is emitted (deny-by-default).
- Host allowlist disjoint from the server request: no `allow` attribute is emitted.

## Notes

- jsdom cannot service the child-window `postMessage` the bridge makes on teardown, so the test nulls the iframe `contentWindow` to no-op the bridge effect and isolates the rendered attribute (what this change verifies).
- This is a lighter, in-process check that the SDK emits the correct attribute. A full browser test that a granted capability is actually usable also needs the sandbox proxy to give the inner frame a usable origin (`allow-same-origin`) and can follow separately.
- Test-only, no published behavior changes, so no changeset.

### DIFF
--- a/.changeset/harden-mcp-apps.md
+++ b/.changeset/harden-mcp-apps.md
@@ -1,0 +1,12 @@
+---
+'@ai-sdk/react': patch
+'@ai-sdk/mcp': patch
+---
+
+Harden MCP Apps handling of server-supplied resource metadata and the host/iframe bridge:
+
+- Runtime-validate `_meta.ui` and drop malformed or non-string fields.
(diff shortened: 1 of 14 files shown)


dnukumamras:
Folding this test into #17249 so the fix and its coverage land together. Closing.

PR #17273, opened 2026-07-15
fix(amazon-bedrock): forward disableParallelToolUse

### LINKED ISSUE #17267: @ai-sdk/amazon-bedrock: providerOptions.anthropic.disableParallelToolUse is silently ignored (never sent to Bedrock)
## Description

When calling Anthropic models through `@ai-sdk/amazon-bedrock`, the documented Anthropic provider option `providerOptions.anthropic.disableParallelToolUse` is **silently ignored**: the request sent to Bedrock contains no `tool_choice.disable_parallel_tool_use`, and the model freely returns multiple `tool_use` blocks in a single assistant step.

There is also no Bedrock-native equivalent option — the published dist of the provider contains no user-facing parallel-tool-use control at all (the only match for `parallel` is the internal capability flag `supportsParallelCalls = true`).

This is dangerous as a silent no-op: code that sets the flag (plus tests asserting it is set) reads as if tool execution is serialized, while the wire never carries the field.

## Affected versions

- `@ai-sdk/amazon-bedrock@4.0.113` (verified locally)
- `@ai-sdk/amazon-bedrock@5.0.22` (latest at time of filing; verified in the published dist)

```bash
$ grep -c "disable_parallel_tool_use" node_modules/@ai-sdk/amazon-bedrock/dist/index.mjs
0
$ curl -sL https://unpkg.com/@ai-sdk/amazon-bedrock@5.0.22/dist/index.mjs | grep -c "disable_parallel_tool_use"
0
```

For comparison, `@ai-sdk/anthropic` does map the option into `tool_choice` (`dist/index.mjs`, `toolChoice: disableParallelToolUse ? { type: "auto", disable_parallel_tool_use: ... }`).

## Reproduction

```ts
import { z } from 'zod';

  model: bedrock('global.anthropic.claude-sonnet-4-6'),
  tools: {
    get_resistor_value: tool({
      description: 'Look up a resistor value',
      inputSchema: z.object({ ref: z.string() }),
      execute: async () => '10k',
    }),
    get_capacitor_value: tool({
      description: 'Look up a capacitor value',
      inputSchema: z.object({ ref: z.string() }),
      execute: async () => '100nF',
    }),
  },
  prompt: 'Look up the values of R1 and C1. You may call both tools in the same step.',
  providerOptions: {
    anthropic: {
      thinking: { type: 'enabled', budgetTokens: 4000 },
      sendReasoning: true,
      disableParallelToolUse: true, // silently dropped
    },
  },
});
// result.steps[0] contains TWO tool calls (observed 2/2 runs on
// global.anthropic.claude-sonnet-4-6 and global.anthropic.claude-opus-4-6-v1,
// us-east-1).
```

## Expected behavior

Either of:

1. Forward `disableParallelToolUse` to Bedrock as `tool_choice: { type: "auto", disable_parallel_tool_use: true }` — Bedrock honors the field: injecting it manually into the raw request body reliably serializes tool calls (verified with a control run: 1 tool call per step, 2/2 runs, both models above), or
2. Emit a loud warning when an unsupported provider option is passed, instead of silently dropping it.

## Observed behavior

The option is dropped with no warning; the model emits parallel `tool_use` blocks.

## Related

#15652 touches the same option in `@ai-sdk/anthropic` (hardcoding under `structuredOutputMode: 'jsonTool'`) — the Bedrock provider, by contrast, never sends the field at all.

arjxnt:
## Summary

- Forward Anthropic's `disable_parallel_tool_use` from the provider tool choice to Bedrock.
- Cover automatic and named tool choices with the existing adapter request-transform tests.

The adapter previously reconstructed `tool_choice` using only its type and optional name, silently dropping the parallel-tool-use control before the request reached Bedrock.

Fixes #17267.

## Validation

- `vitest run packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.test.ts` (37 passed)
- Prettier and `git diff --check`

### DIFF
--- a/.changeset/bright-beds-argue.md
+++ b/.changeset/bright-beds-argue.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
+fix: forward Anthropic's disableParallelToolUse setting to Bedrock
--- a/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.test.ts
+++ b/packages/amazon-bedrock/src/anthropic/amazon-bedrock-anthropic-provider.test.ts
@@ -308,7 +308,7 @@ describe('amazon-bedrock-anthropic-provider', () => {
     expect(transformedBody).toHaveProperty('anthropic_version');
   });
 
-  it('should strip disable_parallel_tool_use from tool_choice', () => {
+  it('should preserve disable_parallel_tool_use in tool_choice', () => {
     const provider = createAmazonBedrockAnthropic({
       region: 'us-east-1',
       accessKeyId: 'test-key',
@@ -334,10 +334,10 @@ describe('amazon-bedrock-anthropic-provider', () => {
       new Set(),
     );
 
-    expect(transformedBody?.tool_choice).toEqual({ type: 'auto' });
-    expect(transformedBody?.tool_choice).not.toHaveProperty(
-      'disable_parallel_tool_use',
-    );
+    expect(transformedBody?.tool_choice).toEqual({
+      type: 'auto',
+      disable_parallel_tool_use: true,
+    });
   });
 
   it('should preserve tool_choice name when present', () => {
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by #19560, which already closed #17267.

PR #17282, opened 2026-07-15
fix: accept recursively readonly JSON structures as tool-result output

### LINKED ISSUE #15593: type `JSONObject` is not compatible with readonly array
### Description

The type `JSONObject` (possibly in <https://github.com/vercel/ai/blob/main/packages/provider/src/json-value/json-value.ts>) is not compatible with arrays marked as `readonly` (or any objects and arrays containing them). This will affect the tool call result json output passing when the result is produced by some functions whose return values contain read only arrays. Example code:
```typescript
import {openai} from "@ai-sdk/openai";
import {generateText} from "ai";

const execute = () => [] as const;

generateText({
    model: openai('gpt-3.5-turbo'),
    messages: [
        {
            role: 'tool',
            content: [
                {
                    type: 'tool-result',
                    toolCallId: '',
                    toolName: '',
                    output: {
                        type: 'json',
                        value: execute(),
                    },
                },
            ],
        },
    ],
});
```
The `execute` function above acts as a tool function. The error message is:
```text
Type 'readonly []' is not assignable to type 'JSONValue'.
  The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type 'JSONArray'.
```
The solution could be marking the `JSONArray` type with `readonly`, if the `JSONObject` type is not used at anywhere that requires mutability. (This will not affect normal mutable arrays since `any[] extends readonly any[] ? true : false` resolves to `true` and `readonly any[] extends any[] ? true : false` resolves to `false`.) If it is not the case, creating a new type `ReadonlyJSONObject` or something similar for that is also feasible.

### AI SDK Version

- ai: 6.0.191
- @ai-sdk/openai: 3.0.65 (although not relevant)

### Code of Conduct

lgrammel:
## Background

Tool results commonly originate from functions using const assertions, but JSONValue rejected readonly arrays and structures containing them, requiring unnecessary casts or mutable copies.

### Root Cause

The recursive JSONValue definition used mutable JSONValue[] through JSONArray, so TypeScript rejected readonly tuples and arrays at any nesting level. The reported generateText example reproduced this as TS2322.

## Summary

The provider JSONValue type now models objects as Readonly<JSONObject> and arrays as readonly JSONValue[], while the separately exported JSONObject and JSONArray aliases remain mutable for existing consumers.

## Testing

Added colocated type regression coverage for recursively readonly JSON objects and arrays, every scalar JSON member, and writable JSONObject and JSONArray aliases. Reran the provider build and type-check plus all repository tests and checks.

## End-to-end Validation

- `pnpm --filter @ai-sdk/provider build` followed by a temporary `pnpm tsx` compiler harness: a generateText tool-result containing recursively readonly JSON compiled with zero diagnostics, and the generated declaration contained both readonly composite branches.

## Related Issues

Fixes #15593

Co-authored-by: plutonium-94 <47095613+plutonium-94@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

JSONValue now recursively accepts readonly JSON arrays, including readonly arrays nested in objects, while the original generateText tool-result reproduction type-checks successfully.

Concerns: none.

### Side effects

Risk: low

The change only widens the public JSONValue type; the mutable JSONArray alias remains JSONValue[], preserving existing mutable-array behavior, and the repository type-check passes.

Concerns: none.

### Performance

Risk: none

This is a type-only change with no emitted runtime logic, allocation, or processing overhead.

Concerns: none.

### Backwards compatibility

Risk: none

No runtime code paths, serialization formats, migrations, or stored data are changed; existing mutable JSON values remain assignable.

Concerns: none.

### Security

Risk: none

The change only expands compile-time assignability to readonly arrays whose elements are already constrained to JSONValue, introducing no new runtime input handling.

Concerns: none.

### Testing

Status: appropriate

The added type regression test covers readonly arrays, nested readonly arrays, and objects containing them; it is included by the provider TypeScript project, and the exact reported tool-result scenario was also verified.

Concerns: none.

### Verification

Reviewed git diff origin/main...HEAD; ran @ai-sdk/provider type-check and build, root pnpm type-check, git diff --check, and focused Ultracite checks successfully. Confirmed the generated declaration contains readonly JSONValue[] while JSONArray remains mutable, and ran the historical issue reproduction against the fixed checkout, which reported that the readonly tool-result value type-checks successfully.

lgrammel:
Expand to the other types in the JSONValue type -- what if those are readonly?

lgrammel:
Addressed [PR comment 4979551932](https://github.com/vercel/ai/pull/17282#issuecomment-4979551932).

> ## Bugfix review
> 
> Outcome: approved
> 
> ### Fixes issue
> 
> Status: fully-addresses
> 
> ... 47 more line(s) omitted

No change was required from this approval. The fix remains type-only, preserves the mutable JSONObject and JSONArray aliases, and passed the complete repository verification suite.

lgrammel:
Addressed [PR comment 4979580249](https://github.com/vercel/ai/pull/17282#issuecomment-4979580249).

> Expand to the other types in the JSONValue type -- what if those are readonly?

Updated JSONValue so both composite branches are explicitly readonly-compatible: objects now use Readonly<JSONObject> and arrays use readonly JSONValue[]. Scalar JSON types have no separate readonly variants. I also expanded the type test to cover recursively readonly objects and arrays while confirming the exported JSONObject and JSONArray aliases remain writable.

n1ckoates (approved):

PR #17336, opened 2026-07-15
fix(mcp): treat code_challenge_methods_supported as optional

### LINKED ISSUE #17334: OAuthMetadataSchema requires code_challenge_methods_supported, but RFC 8414 marks it optional, breaking DCR against spec-compliant servers (e.g. AWS Sign-In)
### Description

`OAuthMetadataSchema` in `packages/mcp/src/tool/oauth-types.ts` defines `code_challenge_methods_supported` as a required field:

```ts
code_challenge_methods_supported: z.array(z.string()),
```

Per RFC 8414 (OAuth 2.0 Authorization Server Metadata), this field is OPTIONAL. The official `@modelcontextprotocol/sdk` correctly marks it `.optional()` in its equivalent schema (`OAuthMetadataSchema` in `src/shared/auth.ts`).

Any authorization server that omits this optional field causes `discoverAuthorizationServerMetadata()` to throw a ZodError before the OAuth flow (including Dynamic Client Registration) can even start. This is not hypothetical: AWS Sign-In, the authorization server behind the AWS MCP Server, omits it in its `/.well-known/oauth-authorization-server` response, even though it supports S256 PKCE (confirmed via AWS's own docs and by testing the flow with the field patched in via a custom `fetch`).

Every sibling field in the same schema (`registration_endpoint`, `scopes_supported`, `grant_types_supported`, `token_endpoint_auth_methods_supported`) is marked `.optional()`. This one field's stricter treatment looks like an oversight rather than an intentional divergence from the RFC. The test fixture added alongside the original OAuth PR (#9127) always returns this field, so the omitted-field case was never exercised.

Suggested fix (one line):

```ts
code_challenge_methods_supported: z.array(z.string()).optional(),
```

Note: `startAuthorization()` in `oauth.ts` already treats `metadata.code_challenge_methods_supported` as possibly absent (it throws a clear "Incompatible auth server" error when missing), so the codebase already has a sane fallback path for this. It's just unreachable because the Zod schema rejects the response before getting there.

### Reproduction

1. Install:

```
npm install @ai-sdk/mcp@2.0.13 zod
```

2. Run the OAuth flow against the AWS MCP Server (any minimal OAuthClientProvider works, since the crash happens during metadata discovery, before any provider method beyond the getters is invoked):

```js

const provider = {
  get redirectUrl() { return 'http://localhost:8090/oauth/callback'; },
  get clientMetadata() {
    return {
      redirect_uris: ['http://localhost:8090/oauth/callback'],
      token_endpoint_auth_method: 'none',
      grant_types: ['authorization_code', 'refresh_token'],
      response_types: ['code'],
      client_name: 'repro',
    };
  },
  tokens() { return undefined; },
  saveTokens() {},
  clientInformation() { return undefined; },
  saveClientInformation() {},
  saveCodeVerifier() {},
  codeVerifier() { return ''; },
  redirectToAuthorization() {},
};

  transport: {
    type: 'http',
    url: 'https://aws-mcp.us-east-1.api.aws/mcp',
    authProvider: provider,
  },
});

await client.callTool({
  name: 'aws___call_aws',
  arguments: { cli_command: 'aws sts get-caller-identity' },
});
```

3. Observe the actual AWS Sign-In metadata (no auth required):

```
curl -s https://us-east-1.oauth.signin.aws/.well-known/oauth-authorization-server
```

Response (note: no `code_challenge_methods_supported` field):

```json
{"issuer":"https://us-east-1.oauth.signin.aws","authorization_endpoint":"https://us-east-1.oauth.signin.aws/v1/authorize","token_endpoint":"https://us-east-1.oauth.signin.aws/v1/token","registration_endpoint":"https://us-east-1.oauth.signin.aws/v1/register","token_endpoint_auth_methods_supported":["none"],"response_types_supported":["code"],"grant_types_supported":["authorization_code","refresh_token"]}
```

4. Result: `discoverAuthorizationServerMetadata()` throws before DCR or the authorization redirect can happen:

```
ZodError: [
  {
    "expected": "array",
    "code": "invalid_type",
    "path": ["code_challenge_methods_supported"],
    "message": "Invalid input: expected array, received undefined"
  }
]
```

5. Confirmation of the fix: wrapping `fetch` to inject `code_challenge_methods_supported: ["S256"]` into the `/.well-known/oauth-authorization-server` response before the Zod parse allows the full flow (DCR, then PKCE authorization code, then token exchange, then authenticated tool call) to complete successfully end-to-end against the real AWS MCP Server.

### AI SDK Version

@ai-sdk/mcp: 2.0.13

### Code of Conduct

Hashim1999164:
## Summary

`OAuthMetadataSchema` required `code_challenge_methods_supported`, but RFC 8414 marks that field as optional. Servers like AWS Sign-In omit it, so metadata discovery failed before the OAuth flow could start.

This makes the field optional in the schema and adds a test for metadata responses without it.

Fixes #17334

## Test plan

- [x] `pnpm --filter @ai-sdk/mcp test:node -- src/tool/oauth.test.ts -t "accepts OAuth metadata without code_challenge_methods_supported"`

### DIFF
--- a/packages/mcp/src/tool/oauth-types.ts
+++ b/packages/mcp/src/tool/oauth-types.ts
@@ -69,7 +69,7 @@ export const OAuthMetadataSchema = z.looseObject({
   scopes_supported: z.array(z.string()).optional(),
   response_types_supported: z.array(z.string()),
   grant_types_supported: z.array(z.string()).optional(),
-  code_challenge_methods_supported: z.array(z.string()),
+  code_challenge_methods_supported: z.array(z.string()).optional(),
   token_endpoint_auth_methods_supported: z.array(z.string()).optional(),
   token_endpoint_auth_signing_alg_values_supported: z
     .array(z.string())
--- a/packages/mcp/src/tool/oauth.test.ts
+++ b/packages/mcp/src/tool/oauth.test.ts
@@ -576,6 +576,27 @@ describe('discoverAuthorizationServerMetadata', () => {
     expect(metadata).toEqual(tenantMetadata);
   });
 
+  it('accepts OAuth metadata without code_challenge_methods_supported', async () => {
+    const metadataWithoutPkceMethods = {
+      issuer: 'https://auth.example.com',
+      authorization_endpoint: 'https://auth.example.com/authorize',
+      token_endpoint: 'https://auth.example.com/token',
+      response_types_supported: ['code'],
+    };
+
+    mockFetch.mockResolvedValueOnce({
+      ok: true,
+      status: 200,
+      json: async () => metadataWithoutPkceMethods,
+    });
+
+    const metadata = await discoverAuthorizationServerMetadata(
+      'https://auth.example.com',
+    );
+
+    expect(metadata).toEqual(metadataWithoutPkceMethods);
+  });
+
   it('tries URLs in order and returns first successful metadata', async () => {
     // First OAuth URL fails with 404
     mockFetch.mockResolvedValueOnce({


n1ckoates:
Please sign your git commits. You can re-open this PR after signing past commits and force-pushing, or open a new PR with signed commits. Thank you!

PR #17339, opened 2026-07-15
fix(ai): preserve tool metadata from output chunks

### LINKED ISSUE #16930: readUIMessageStream/processUIMessageStream drops a tool-output-available chunk's toolMetadata
### Description

When a tool-output-available (or tool-output-error) chunk carries toolMetadata, that metadata is discarded — the resulting UI message part keeps whatever toolMetadata was on the part from the earlier tool-input-* chunks (or undefined if there was none). output and providerMetadata from the same chunk are applied, only toolMetadata is ignored.

This breaks the common case where a tool's metadata is only known after the tool finishes (e.g. "the result is downloadable / N bytes", a cache key, a final status) — there's no way to attach it, because the only chunk that can carry it is the output chunk, and its toolMetadata is dropped.

#### Reproduction:
```
import { readUIMessageStream, simulateReadableStream } from 'ai'

const chunks = [
  { type: 'start' },
  { type: 'tool-input-start', toolCallId: 'call-1', toolName: 'myTool', dynamic: true },
  { type: 'tool-input-available', toolCallId: 'call-1', toolName: 'myTool', input: { q: 1 }, dynamic: true },
  {
    toolCallId: 'call-1',
    output: { ok: true },
    dynamic: true,
    toolMetadata: { downloadable: true }, // <-- provided only on the output chunk
  },
  { type: 'finish' },
]

let last
for await (const message of readUIMessageStream({
  stream: simulateReadableStream({ chunks, initialDelayInMs: 0, chunkDelayInMs: 0 }),
})) {
  last = message
}

const toolPart = last.parts.find(p => p.type === 'dynamic-tool')
console.log('expected toolMetadata:', { downloadable: true })
console.log('actual   toolMetadata:', toolPart?.toolMetadata) // => undefined
```

#### Suggestion to fix

In `packages/ai/src/ui/process-ui-message-stream.ts`, the tool-output-available and tool-output-error cases dispatch to `updateDynamicToolPart`/`updateToolPart` with:

```
output: chunk.output,                      // from the chunk ✅
providerMetadata: chunk.providerMetadata,  // from the chunk ✅
toolMetadata: toolInvocation.toolMetadata, // from the existing part, NOT chunk.toolMetadata ❌
```

Changing in the tool-output-available and tool-output-error branches (both dynamic and static tool paths):
```
toolMetadata: chunk.toolMetadata ?? toolInvocation.toolMetadata
```

### AI SDK Version

ai: 7.0.15

### Code of Conduct

arjxnt:
## Summary

- Preserve `toolMetadata` supplied by `tool-output-available` and `tool-output-error` chunks.
- Retain the input-phase metadata when output metadata is absent.
- Add a regression test that verifies output metadata replaces earlier input metadata on dynamic tool parts.

The UI stream processor previously passed the stale metadata from the existing tool part into every output update, ignoring metadata carried by the output chunk itself.

Fixes #16930.

## Validation

- `vitest run --config packages/ai/vitest.node.config.js packages/ai/src/ui/process-ui-message-stream.test.ts` (103 passed)
- `git diff --check`

### DIFF
--- a/.changeset/slow-bugs-notice.md
+++ b/.changeset/slow-bugs-notice.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix: preserve tool metadata from tool output chunks in UI message streams
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -5822,7 +5822,7 @@ describe('processUIMessageStream', () => {
       });
     });
 
-    it('should preserve tool metadata on dynamic tool parts', async () => {
+    it('should apply output metadata to dynamic tool parts', async () => {
       const stream = createUIMessageStream([
         { type: 'start', messageId: 'msg-123' },
         { type: 'start-step' },
@@ -5839,6 +5839,7 @@ describe('processUIMessageStream', () => {
           toolCallId: 'tool-call-1',
           output: { result: 'provider-result' },
           dynamic: true,
+          toolMetadata: { downloadable: true },
         },
         { type: 'finish-step' },
         { type: 'finish' },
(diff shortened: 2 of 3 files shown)


arjxnt:
Closing as a duplicate of #16933. The earlier PR covers both static and dynamic tool parts, output-success and output-error chunks, and the metadata fallback behavior. Keeping this open would add review noise without independent value.

PR #17341, opened 2026-07-15
Fix Agents next link

marcoshernanz:
Fixes the Agents pagination link to skip the section index and open Overview.

The section index was included as the next page, producing a self-link in production.

Validated with `pnpm check`, docs tests, type-checking, and a production build.


marcoshernanz:
Superseded by vercel/ai-studio#2086, which fixes the live path-matching bug at its source.

PR #17343, opened 2026-07-16
fix: prevent tool execution from interleaving with transformed assistant output

lgrammel:
## Background

Delayed stream transforms such as `smoothStream` could retain assistant text or reasoning while upstream tool execution began, causing tool logs and side effects to appear before the preceding assistant output finished.

## Root Cause

Tool execution started at model-call completion without waiting for buffered transformed output to reach the consumer. Regression checks also confirmed that output-end-based synchronization was incomplete for tool-call-before-end ordering and unsafe when transforms remapped IDs.

## Summary

Tool execution now waits for the corresponding transformed tool-call boundary to reach the active result stream using an ordered, ID-independent waiter queue. The existing patch changeset remains applicable.

## Testing

Regression coverage verifies smoothed text and reasoning for both tool-call/end orderings and verifies that structure-preserving output-ID remapping completes without stalling. Focused Node and Edge suites passed 429 tests each, with package and full-workspace type checking, formatting, and lint checks passing.

## End-to-end Validation

- `pnpm -C packages/ai exec tsx` with an inline `streamText`/`smoothStream` harness: text-end-before-tool-call and tool-call-before-text-end both produced complete assistant text before `[tool executed]`.

## Related Issues

Fixes #13199

Co-authored-by: mindplay-dk <103348+mindplay-dk@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: no-longer-reproduces

Concerns: none.

### Fixes issue

Status: partially-addresses

The reported ordering is fixed when text-end precedes the tool call, but tool calls occurring before their text-end still execute after only the first smoothed chunk.

Concerns:
- The repository's smoothStream tests explicitly support tool-call-before-text-end ordering, but the new waiter is captured only when a tool call follows an already-seen text-end or reasoning-end.
- A focused runtime check with tool-call before text-end still produced tool execution between the first and remaining smoothed words.

### Side effects

Risk: medium

The waiter is keyed to pre-transform output IDs, so a structure-preserving custom transform that consistently remaps text IDs can leave tool execution waiting indefinitely.

Concerns:
- A focused check using a custom transform that remapped matching text-start, text-delta, and text-end IDs stalled instead of completing.

### Performance

Risk: low

Normal use adds only a small map and promise per pending text or reasoning boundary, although unresolved waiter cases can retain the stream and associated state.

Concerns:
- The identified unmatched-ID case can leave the pipeline and waiter state pending indefinitely.

### Backwards compatibility

Risk: none

The change affects runtime stream scheduling only and does not read, migrate, or rewrite stored data.

Concerns: none.

### Security

Risk: none

No new external inputs, credential handling, network access, parsing, or authorization behavior is introduced.

Concerns: none.

### Testing

Status: needs-more

The added regression test covers only text-end-before-tool-call ordering and misses supported ordering and transform-identity edge cases.

Concerns:
- Add coverage where tool-call appears before text-end.
- Add equivalent reasoning coverage.
- Add coverage ensuring valid custom transformations cannot leave tool execution permanently waiting.

### Verification

Reviewed the complete production and test diff, confirmed the changeset, ran the two focused Node test files with 424 passing tests, and completed package type checking and formatting checks. Additional bounded runtime checks demonstrated both the remaining tool-call-before-text-end interleaving and the custom-ID-transform stall.

lgrammel:
Addressed [PR comment 4989268940](https://github.com/vercel/ai/pull/17343#issuecomment-4989268940).

> ## Bugfix review
> 
> Outcome: changes-required
> 
> ### Reproduction replay
> 
> 
> ... 62 more line(s) omitted

Updated synchronization to create ordered waiters at source tool-call boundaries and resolve them when the corresponding transformed tool calls reach the active result stream, rather than matching pre-transform text or reasoning end IDs. This handles tool calls before output-end and remains safe when transforms remap output IDs. Added text and reasoning coverage for both orderings plus a custom ID-remapping transform; the focused Node and Edge suites each pass all 429 tests.

lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: partially-addresses

The reported single-consumer smoothStream scenario is fixed, including both tool-call/end orderings and remapped output IDs, but concurrent consumption through another supported result branch can still cause tool side effects to precede pending text output.

Concerns:
- A focused check that consumed the aggregate response concurrently with a slower textStream observed tool execution after only two of eight smoothed text chunks had reached the text consumer.

### Side effects

Risk: medium

The shared waiter queue is resolved by whichever result-stream branch reaches a tool call first, so a fast aggregate or secondary consumer can allow side effects to interleave with a slower legitimate output consumer.

Concerns:
- Concurrent access to result promises and streaming output is supported by the existing on-demand tee architecture, but this combination is not handled correctly.

### Performance

Risk: low

For transformed streams, the change adds one small promise-backed waiter per tool-call boundary and linear queue operations, with no material expected overhead.

### Backwards compatibility

Risk: none

The change only alters runtime stream scheduling and does not read, migrate, rewrite, or invalidate stored data.

### Architecture

Risk: medium

Package boundaries and dependency direction remain intact, but synchronization is attached independently to every post-transform tee branch while using one shared FIFO waiter queue, making side-effect scheduling depend on the fastest consumer rather than the consumer whose output ordering matters.

Concerns:
- The fan-out layer duplicates each tool-call event across branches, but all branches mutate the same synchronization queue.
- The synchronization should preserve the ordering invariant across supported concurrent result consumption, not let an unrelated drain branch release tool execution early.

### Change scope

Status: minimal

The production changes, targeted regression tests, and patch changeset are all directly related to fixing and releasing the reported ordering bug.

### Security

Risk: none

The change introduces no new external input handling, parsing, network access, credential behavior, authorization logic, or data exposure.

### Testing

Status: needs-more

Tests appropriately cover text and reasoning, both tool-call/end orderings, and ID remapping, but omit the supported multiple-consumer case that still reproduces interleaving.

Concerns:
- Add regression coverage where textStream is consumed with backpressure while an aggregate result promise or another result stream is consumed concurrently.

### Verification

Reviewed every changed file and the stream fan-out architecture, confirmed clean diff checks, ran the focused Node and Edge suites with 429 passing tests each, completed the ai package type check, and performed bounded runtime checks for single- and concurrent-consumer ordering. The concurrent-consumer check exposed the remaining ordering defect.

PR #17367, opened 2026-07-16
[v6.0] reproduction: Cannot catch errors from writeToServerResponse

lgrammel:
## Bug

Cannot catch errors from `writeToServerResponse`

## Reproduction

- `examples/ai-functions/src/reproduction/issue-11023-write-to-server-response-error.ts` demonstrates that an awaited piping call cannot catch a stream-side `AI_InvalidResponseDataError`; the error instead becomes an unhandled rejection.
- The expected behavior is for the piping operation to reject so the caller's `catch` can handle the stream read failure.
- The behavior occurs with both the reported `ai@5.0.76` and the checkout's `ai@6.0.228`, so upgrading to the current target branch does not resolve it.
- `writeToServerResponse` invokes its asynchronous `read()` operation without returning its promise, while the public piping helpers return `void`.

## Related Issues

Reproduces #11023

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "issueId": 11023,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-11023-write-to-server-response-error.ts",
+  "artifactPaths": [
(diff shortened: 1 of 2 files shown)


lgrammel:
Issue #11023 already has the merged-v6.0 label. Closing this stale reproduction pull request.

PR #17363, opened 2026-07-16
[v5.0] reproduction: convertToOpenAIChatMessages sends content: "" instead of content: null

lgrammel:
## Bug

convertToOpenAIChatMessages sends content: "" instead of content: null for tool-call-only assistant messages

## Reproduction

- Added runnable reproduction at `examples/ai-functions/src/reproduction/issue-12389-openai-tool-call-content.ts` with workspace manifest `examples/ai-functions/package.json`.
- The checked-out `@ai-sdk/openai` v2.0.112 serialized a tool-call-only assistant message with `content: ""`; the expected value was `null` or an omitted field.
- The checkout uses `ai` v5.0.214 and `@ai-sdk/openai` v2.0.112 rather than the reporter's v6.0.41 and v3.0.12, but exhibits the same reported serialization failure.
- The downstream qwen3-coder text-markup switch was not live-verified because no Ollama service was listening on `localhost:11434`.

## Relevant Documentation

- https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create
- https://docs.ollama.com/api/openai-compatibility
- https://registry.ollama.com/library/qwen3-coder

## Related Issues

Reproduces #12389

PR #17373, opened 2026-07-16
[v5.0] fix: tool-call-only assistant messages use invalid empty-string content

lgrammel:
## Background

Tool-call-only assistant messages were sent with empty-string content, causing some OpenAI-compatible providers to reject requests or break multi-turn structured tool calling.

## Root Cause

Both chat converters initialized assistant text to an empty string and serialized it unconditionally, ignoring that messages containing tool calls but no text require null content; the reproduction confirmed an observed content value of "".

## Summary

Updated the OpenAI and OpenAI-compatible chat converters to use null only when assistant text is empty and tool calls are present, while preserving empty strings for messages without tool calls.

## Testing

Added inline-snapshot regressions for tool-call-only null content and empty text without tool calls, and updated existing tool-call expectations in both converter test suites.

## End-to-end Validation

- `pnpm -F @ai-sdk/openai build` plus a live `gpt-4o-mini` Chat Completions probe: the tool-call history was accepted and returned `OK.`.
- `pnpm -F @ai-sdk/openai-compatible build` plus a live OpenAI-compatible `gpt-4o-mini` probe: the tool-call history was accepted and returned `OK`.

## Related Issues

Fixes #12389

Closes #17363

Co-authored-by: lbijeau <2281680+lbijeau@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

Concerns: none.

### Fixes issue

Status: fully-addresses

Both affected chat converters now emit null when tool calls are present and accumulated assistant text is empty, while retaining non-empty text and preserving empty strings for messages without tool calls.

Concerns: none.

### Side effects

Risk: low

The behavioral change is narrowly limited to assistant messages containing tool calls with empty text; existing non-empty assistant content and empty messages without tool calls retain their prior serialization.

Concerns: none.

### Performance

Risk: none

The change adds only a constant-time conditional during existing message serialization and introduces no allocations, loops, or retained state.

Concerns: none.

### Backwards compatibility

Risk: low

Stored messages are not modified; previously stored tool-call-only turns will simply serialize as null rather than an empty string when sent again, which is the intended compatibility correction.

Concerns: none.

### Security

Risk: none

The change only selects the serialized content value and does not alter authentication, input trust boundaries, data access, or tool execution.

Concerns: none.

### Testing

Status: appropriate

Regression tests cover tool-call-only null content and preservation of empty content without tool calls in both converters, while existing tool-call and metadata tests were updated and all relevant suites passed.

Concerns: none.

### Verification

Inspected the complete base-branch diff and converter types, confirmed the fix is confined to the two chat serializers, and ran both packages' Node and edge test suites, type checks, and diff validation successfully. The authoritative reproduction replay also confirmed the original bug signal is absent.

PR #17374, opened 2026-07-16
feat: add a per-step first-content timeout for streaming generations

### LINKED ISSUE #17315: Add a per-step first-content timeout for streaming generations
## Background

A streaming provider can return successful response headers and transport bytes immediately, emit only an empty role prelude, metadata, or keep-alive framing, and then stall indefinitely without producing content. Existing `chunkMs` only governs gaps after content has started, while transport-level first-byte and inactivity timers cannot distinguish parsed output from SSE comments or other non-content bytes. A first-content deadline therefore requires the SDK's parsed stream-part classification.

## Feature Request

Extend the public streaming timeout configuration with an optional millisecond-valued `firstChunkMs` setting and enforce it in streaming generation flows such as `streamText`. It must be independently configurable, remain unset by default, and not require `headersMs` to be configured or implemented.

For every streaming model-call step, including later calls in multi-step and tool-calling runs, arm the deadline when that step's response stream starts. Re-arm it independently for each step. Track satisfaction using the SDK's existing `isOutputChunkType` classification and `hasReceivedOutputChunk` state, not HTTP or raw transport activity.

The first text delta, reasoning delta, tool-input output part, or tool-call output part classified with `isOutputChunkType === true` must satisfy and permanently disarm the deadline for that step. An empty role prelude, keep-alive or ping framing, SSE comments, stream-start parts, response metadata, HTTP headers, a `200` response, the first response byte, and arbitrary transport bytes must neither satisfy nor reset it.

If no classified content-bearing output appears within `firstChunkMs`, including when headers and non-content framing arrive before the stream goes silent, abort the underlying provider stream through a dedicated SDK timeout signal merged into the model call's abort signal. Surface the expiration consistently with existing step and chunk timeout failures. Ensure the timeout is cancelled before forwarding the first content-bearing part, so it cannot expire after output has begun and a pre-content failure remains safe to retry without duplicating generated output.

After the first output part disarms `firstChunkMs`, existing `chunkMs` behavior must continue to govern gaps between subsequent content chunks; `chunkMs` must not be redefined as a first-byte or pre-first-content timeout. `stepMs` and `totalMs` remain coarse outer bounds, and existing `toolMs` and per-tool `tools` timeout behavior remains unchanged. Enforcement must live in the SDK and work across providers and supported runtimes without relying on Node- or undici-specific transport timeouts.

## Related Issues

Extracted from #16538

lgrammel:
## Background

Streaming providers can return headers and non-content framing promptly but then stall indefinitely; a semantic first-content timeout lets applications abort these stalled steps safely before visible output begins.

## Summary

Added the optional public `firstChunkMs` timeout, extractor, per-step stream enforcement, dedicated abort signal, content-aware timeout classification, cleanup, and post-first-output `chunkMs` handling.

## Testing

Added focused Node and Edge tests covering non-content stalls, text and reasoning deltas, tool-input and tool-call output, later multi-step calls, timeout surfacing, extraction, and `chunkMs` interaction.

## End-to-end Validation

- Ran the new OpenAI `timeout-first-chunk.ts` example against the live API and successfully streamed a complete response with combined `firstChunkMs` and `chunkMs` settings.

## Documentation

Updated timeout settings, `streamText`, `generateText` event types, and ToolLoopAgent references with the new API, per-step semantics, ignored non-content parts, examples, and timeout interactions.

## Related Issues

Fixes #17315

### DIFF
--- a/.changeset/bright-spoons-stream.md
+++ b/.changeset/bright-spoons-stream.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+feat(ai): add a per-step `firstChunkMs` timeout for streaming generations
--- a/content/docs/03-ai-sdk-core/25-settings.mdx
+++ b/content/docs/03-ai-sdk-core/25-settings.mdx
@@ -142,9 +142,12 @@ You can specify the timeout either as a number (milliseconds) or as an object wi
 
 - `totalMs`: The total timeout for the entire call including all steps.
 - `stepMs`: The timeout for each individual step (LLM call). This is useful for multi-step generations where you want to limit the time spent on each step independently.
-- `chunkMs`: The timeout between stream chunks (streaming only). The call will abort if no new chunk is received within this duration. This is useful for detecting stalled streams.
+- `firstChunkMs`: The timeout from the start of each model response stream until the first content-bearing output part (streaming only). Response metadata, stream-start parts, raw chunks, and empty text or reasoning starts do not satisfy or reset this timeout. It is applied independently to every model-call step.
+- `chunkMs`: The timeout between content-bearing output chunks after output has started (streaming only). The call will abort if no new output chunk is received within this duration. This is useful for detecting streams that stall after generation begins.
 - `toolMs`: The default timeout for all tool executions. If a tool takes longer, it aborts and returns a tool-error so the model can respond or retry.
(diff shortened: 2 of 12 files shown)


lgrammel:
Make it part of definition/spec that `firstChunkMs` defaults to `chunkMs` when not specified.

PR #17380, opened 2026-07-16
fix: WorkflowAgent loses assistant narration between tool-call steps

### LINKED ISSUE #17123: WorkflowAgent drops assistant text parts from the conversation prompt on tool-calls finishes

lgrammel:
## Background

WorkflowAgent discarded assistant narration emitted before tool calls, causing subsequent steps to lose prior conclusions and context.

## Root Cause

The tool-calls finish branch rebuilt the assistant conversation message exclusively from toolCalls, despite the completed step containing text; the reproduction and failing regression test confirmed the next prompt omitted that text.

## Summary

Updated WorkflowAgent conversation construction to prepend the completed step's text parts to its assistant tool-call message.

## Testing

Added an inline-snapshot regression test verifying that the next model prompt contains assistant text, its tool call, and the corresponding tool result.

## End-to-end Validation

- `pnpm -C packages/workflow build && pnpm -C examples/ai-functions exec tsx src/reproduction/workflow-agent-drops-tool-call-text.ts` — exited successfully and reported that WorkflowAgent retained the assistant text.

## Related Issues

Fixes #17123

Closes #17371

Co-authored-by: devontivona <1065495+devontivona@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

Concerns: none.

### Fixes issue

Status: fully-addresses

The tool-call branch now prepends the completed step's text content to the assistant tool-call message, so subsequent prompts retain the narration described in the issue.

Concerns: none.

### Side effects

Risk: low

The change only adds previously discarded model text to conversation messages and preserves existing tool-call metadata handling and ordering.

Concerns: none.

### Performance

Risk: low

Prompts now grow by the retained narration text, which is intentional and linear in emitted text, matching normal tool-loop conversation semantics.

Concerns: none.

### Backwards compatibility

Risk: none

No storage format, schema, or persisted data is modified; only future conversation prompt construction changes.

Concerns: none.

### Security

Risk: low

The change introduces no new input parsing, network access, credential handling, or executable behavior; it replays existing assistant text as model context.

Concerns: none.

### Testing

Status: appropriate

A regression test verifies that narration, tool call, and tool result all reach the next model prompt, while the exact end-to-end reproduction also passes.

Concerns: none.

### Verification

Reviewed the complete base-to-head diff and relevant prompt-building behavior. The workflow package's Node and Edge test suites passed with 156 tests each, and package type checking, build, targeted lint, formatting, and diff validation all passed.

gr2m (approved):

PR #17386, opened 2026-07-16
fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input

### LINKED ISSUE #17709: Media-type sniffing performs unbounded work for ID3-prefixed attachments

## Expected behavior

## Related

Extracted from #17386

dnukumamras:
## Summary

## Hardening

## Happy path

## Unhappy path

## Testing

Fixes #17709


lgrammel:
## Bugfix review

Outcome: changes-required

### Fixes issue

Status: partially-addresses

The implementation removes the attachment-sized decode and copy, but it does not preserve the claimed detection behavior for ID3 tags up to the 128 KiB limit.

Concerns:
- The bounded prefix includes the ID3 tag itself but reserves no bytes after it for the MP3 signature. A tag ending at the limit therefore cannot be detected.
- Focused boundary checks found that a tag ending one byte below the limit returns undefined for Uint8Array input but audio/mpeg for equivalent base64 input, contradicting the claimed uniform behavior.

### Side effects

Risk: low

Legitimate MP3 inputs near the new scan boundary can stop auto-detecting, with representation-dependent results at one boundary value.

Concerns:
- Previously detectable MP3 data near the cutoff may now require an explicitly supplied full media type.

### Performance

Risk: low

Work is now bounded independently of attachment size, and subarray avoids copying raw input; ID3-prefixed base64 still incurs a bounded allocation of roughly 128 KiB.

### Backwards compatibility

Risk: low

The change does not mutate stored data, but existing stored MP3 attachments near the cutoff can produce different detection results or fail subtype resolution.

Concerns:
- Callers relying on automatic subtype resolution may receive the existing unsupported-functionality error for affected inputs.

### Security

Risk: none

The change reduces denial-of-service exposure by replacing attachment-sized decoding and copying with a fixed upper bound and introduces no new privileged operations.

### Testing

Status: needs-more

The 67-test detection suite passes in Node and Edge, but the added test covers only an oversized tag and misses the failing cutoff and representation-parity cases.

Concerns:
- Add tests for the largest supported tag with post-tag signature bytes, including the exact cutoff, for both Uint8Array and base64 inputs.
- The oversized-tag assertion does not directly verify that a large attachment with a normal small ID3 tag is decoded only to the bounded prefix; an implementation that fully decoded and then rejected by tag size could still pass it.

### Verification

Inspected the merge-base diff and media-type callers, ran the complete detect-media-type suite in Node and Edge, package type-checking, formatting and lint checks, and focused boundary probes. Standard checks passed, while boundary probes exposed missed detection at the advertised limit and a raw/base64 mismatch.

dnukumamras:
Thanks, both concerns were real and are fixed in 947aabd.

## Signature headroom at the limit

The scan bound was the total decoded prefix, but `stripID3` returns the bytes after the tag, so a tag filling the budget left nothing to match. Split the budget:

- `MAX_ID3_TAG_BYTES` (128 KiB) is the largest tag we skip.
- `ID3_SCAN_BYTES = MAX_ID3_TAG_BYTES + MAX_SIGNATURE_BYTES` is what we actually decode, reserving room for the trailing signature.

Cutoff is now exact and testable: a tag of `MAX_ID3_TAG_BYTES` detects, `MAX_ID3_TAG_BYTES + 1` does not.

## Base64 vs Uint8Array parity

Root cause: `Math.ceil(maxBytes / 3) * 4` rounds up to a whole 4-char group, so the base64 path decoded up to 2 bytes more than the exact `subarray` on the raw-byte path, which is the one-byte mismatch you saw at the boundary. `decodePrefix` now trims the decoded bytes to exactly `maxBytes`, so both representations produce the same length and sniff identically.

## Tests

- Largest supported tag with the frame right after it, asserted for both Uint8Array and base64 (exact cutoff + parity).
- One byte past the limit, asserted undefined for both. This also proves the decode stops at the bound: an implementation that fully decoded and then rejected by tag size would still find the frame here and return `audio/mpeg`, so this case fails that implementation.
- 68 tests pass.

## Side-effect note stands

An ID3v2 tag larger than 128 KiB is still not scanned past, so such an MP3 falls back to the caller-supplied media type (`resolveFullMediaType` throws its existing unsupported-subtype error if none was given). `MAX_ID3_TAG_BYTES` can be raised if needed; cost stays O(1) in attachment size either way.

lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

Media sniffing now decodes only a fixed-size prefix, uses bounded subarray views for byte input, preserves signature headroom after supported ID3 tags, and produces matching results for raw-byte and base64 representations at the cutoff.

### Side effects

Risk: low

Normal media signatures and ID3-tagged MP3s through the documented 128 KiB limit remain detectable; larger ID3 tags now intentionally require a caller-supplied full media type.

Concerns:
- Previously auto-detected MP3 attachments with ID3 tag bodies exceeding 128 KiB will no longer be auto-detected.

### Performance

Risk: none

Attachment-sized base64 decoding and copying are eliminated. ID3-prefixed base64 input incurs only a bounded decode of roughly 128 KiB, while raw input uses non-copying subarray views.

### Backwards compatibility

Risk: low

The change does not mutate stored data, although stored MP3 attachments with oversized ID3 tags may now fail automatic subtype resolution when reused without a full media type.

Concerns:
- Callers processing ID3 tags beyond the new limit must explicitly supply audio/mpeg or another full subtype.

### Security

Risk: none

The change reduces denial-of-service exposure by bounding work independently of attachment size and introduces no new privileged operations or unsafe parsing.

### Testing

Status: appropriate

Regression coverage includes ordinary ID3 MP3s, the exact supported cutoff, one byte beyond it, and raw-byte/base64 parity. Both Node and Edge provider-utils suites passed.

### Verification

Inspected the merge-base diff, callers, changeset, and follow-up boundary fix; ran the complete provider-utils Node and Edge suites, package type checking, repository formatting and lint checks, diff validation, and a focused instrumentation probe confirming that a multi-megabyte ID3-prefixed base64 attachment is decoded only through the fixed prefix.

lgrammel (approved):

PR #17393, opened 2026-07-16
fix: ensure tool input lifecycle callbacks start before input becomes available

lgrammel:
## Background

Non-streaming tool calls could invoke onInputAvailable without first invoking onInputStart, forcing consumers to handle inconsistent lifecycle state.

## Root Cause

generateText directly invoked onInputAvailable for completed tool calls but omitted the corresponding onInputStart invocation, as confirmed by the reproduction and existing callback snapshot.

## Summary

generateText now invokes onInputStart immediately before onInputAvailable for valid non-streaming tool calls.

## Testing

Updated the existing generateText callback regression test and inline snapshot to assert the onInputStart, onInputAvailable sequence.

## End-to-end Validation

- `pnpm -C packages/ai build && pnpm -C examples/ai-functions exec tsx src/reproduction/issue-11043-tool-input-callback-order.ts` — exited successfully with onInputStart preceding onInputAvailable.

## Related Issues

Fixes #11043

Closes #17360

Co-authored-by: SamyPesse <845425+SamyPesse@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

Concerns: none.

### Fixes issue

Status: fully-addresses

For every valid non-streaming generateText tool call, the implementation now awaits onInputStart immediately before onInputAvailable; these are the only non-streaming onInputAvailable invocations.

Concerns: none.

### Side effects

Risk: low

The only observable change is the intended invocation of a configured onInputStart hook for valid non-streaming tool calls; invalid and unavailable tools remain skipped.

Concerns: none.

### Performance

Risk: low

The change adds one optional awaited callback per valid tool call and introduces no retained state or unbounded allocation.

Concerns: none.

### Backwards compatibility

Risk: low

No stored-data format or migration is involved; configured onInputStart callbacks may now perform their intended side effects during generateText.

Concerns: none.

### Security

Risk: low

The change adds no external input parsing, network access, credential handling, or new trust boundary and passes the same scoped callback context used elsewhere.

Concerns: none.

### Testing

Status: appropriate

The regression test now asserts the onInputStart then onInputAvailable sequence and callback arguments, and it passed in both Node and edge configurations; the exact reproduction replay also passed.

Concerns: none.

### Verification

Inspected the merge-base diff and all production onInputAvailable call sites, confirmed the prior test would expose the missing callback, ran the focused regression test under Node and edge runtimes, ran type checks for ai and provider-utils, and confirmed the diff has no whitespace errors.

PR #17400, opened 2026-07-16
fix: prevent orphaned Anthropic tool results after invalid provider-executed tool calls

### LINKED ISSUE #17366: Anthropic: invalid provider-executed tool call synthesizes a tool-error without providerExecuted — orphaned srvtoolu_ tool_result 400s the next request
### Description

When the model emits a `server_tool_use` (e.g. Anthropic `web_search`) whose input fails the SDK's **local** input schema (`webSearch_20250305InputSchema` = `z.object({ query: z.string() })` — e.g. the streamed input deltas accumulate to something that doesn't validate), `parseToolCall` marks the call `invalid` — correctly preserving `providerExecuted` on the tool-call part. But both execution paths then synthesize a companion `tool-error` content part **without** `providerExecuted: true`:

- `generateText`: the `invalidToolCalls` handling pushes `{ type: "tool-error", toolCallId, toolName, input, error, dynamic: true }` into `clientToolOutputs`
- `streamText`: `runToolsTransformation`'s `if (toolCall.invalid)` branch enqueues the same shape

`toResponseMessages` routes tool-results/tool-errors by that flag: un-flagged parts go into the `role:"tool"` response message. The Anthropic provider then converts that part into a plain `tool_result` block inside a **user** message, referencing the `srvtoolu_…` id — while the real `server_tool_use` (and the provider-delivered `web_search_tool_result`, which is correctly stamped `providerExecuted: true` and stays in the assistant message) ride above it.

Anthropic only pairs plain `tool_result` blocks against client `tool_use` blocks, so the **next** request in the loop is rejected:

```
messages.4.content.0: unexpected `tool_use_id` found in `tool_result` blocks: srvtoolu_01FvCG2mjosttzrL4Lnb5mHy. Each `tool_result` block must have a corresponding `tool_use` block in the previous message.
```

In an agent loop this `APICallError` aborts the whole generation — we observed it killing production runs (7 instances over two days) that had already done real work.

**Expected:** the synthetic tool-error for an invalid *provider-executed* call should either carry `providerExecuted: true` (so `toResponseMessages` keeps it in the assistant message and the Anthropic provider serializes it as a `web_search_tool_result` error variant), or be suppressed entirely for provider-executed calls — the API already returns its own result/error block for the `server_tool_use`, so the synthetic part is redundant.

**Verified still present** by inspecting the published dists of `ai@6.0.197`, `ai@6.0.228` (latest 6.x) and `ai@7.0.29` (latest): the `invalidToolCalls` push and the `runToolsTransformation` invalid branch are unchanged in all three, as is the `toResponseMessages` routing filter (`(part.type === "tool-result" || part.type === "tool-error") && !part.providerExecuted`).

Possibly the same root cause as #11855 (open, cannot-reproduce) — that report shows the resulting malformed request from the other side. Related but distinct from #13013 (block ordering, fixed in `@ai-sdk/anthropic@4.0.0-beta.69`) and #11397 (`supportsDeferredResults`, already released).

### Reproduction

Deterministic without a live model: drive `generateText` (or `streamText`) with a mocked `LanguageModelV3` whose response contains

1. a `tool-call` chunk/part with `providerExecuted: true`, `toolName: "web_search"`, and an input that fails the tool's input schema (e.g. `input: "{}"` — `query` missing), and
2. the matching provider `tool-result` chunk/part,

with `anthropic.tools.webSearch_20250305()` registered under `web_search`, in a loop that runs a second step (e.g. a client tool call in the same turn). Inspect the second request's prompt: the `role:"tool"` message contains a `tool-result` part whose `toolCallId` is the `srvtoolu_` id → against the real Anthropic API this 400s with the error above.

Live trigger (probabilistic): a multi-step agent loop with `webSearch_20250305` enabled; occasionally the model emits a `server_tool_use` whose input does not satisfy the local schema. We hit this ~7 times over two days of production traffic.

### AI SDK Version

- ai: 6.0.197 (also verified present in 6.0.228 and 7.0.29)
- @ai-sdk/anthropic: 3.0.81

### Code of Conduct

lgrammel:
## Background

Invalid Anthropic server-side tool input could cause the SDK to send an orphaned `srvtoolu_` tool result on the next step, leading Anthropic to reject the request with HTTP 400 and abort multi-step generation.

## Root Cause

Local schema validation correctly marked provider-executed tool calls invalid and preserved providerExecuted, but generateText and streaming still synthesized unflagged client tool errors. Response conversion consequently routed those synthetic errors into client tool messages even though the provider had supplied its own result.

## Summary

The generateText and streaming paths now skip synthetic client tool errors for invalid provider-executed calls while preserving provider-delivered results. The PR includes focused regression coverage and a patch changeset for the `ai` package.

## Testing

Inline-snapshot regressions cover generateText and streaming, verifying that invalid provider-executed calls retain their flag and provider result without gaining a synthetic client error. The focused suite passes all 282 tests, and full type-check, formatting, and lint validation pass.

## End-to-end Validation

- `pnpm -C examples/ai-functions exec tsx -` deterministic follow-up harness: generateText and streamText each made two calls without placing the `srvtoolu_` result in a client tool message.
- `pnpm -C examples/ai-functions exec tsx -` live Anthropic follow-up harness: both requests were accepted and returned `done.`.

## Related Issues

Fixes #17366

Closes #17391

Co-authored-by: microchipgnu <5553483+microchipgnu@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: inconclusive

Concerns:

### Fixes issue

Status: unclear

The diff suppresses synthetic client tool errors for invalid provider-executed calls in both generateText and streaming while retaining the provider-delivered result.

Concerns:

### Side effects

Risk: low

The behavior change is narrowly gated by both invalid and providerExecuted; invalid client-executed calls retain their existing synthetic errors.

Concerns: none.

### Performance

Risk: none

The change adds only constant-time boolean checks and avoids allocating and processing redundant error parts.

Concerns: none.

### Backwards compatibility

Risk: low

Existing stored data is not migrated or modified; newly produced content intentionally omits the erroneous synthetic error for this specific case.

Concerns: none.

### Security

Risk: none

No input handling, credential handling, URL access, or authorization behavior is introduced or broadened.

Concerns: none.

### Testing

Status: appropriate

Regression tests cover both generation paths, verify the invalid call retains providerExecuted, retain the provider error result, and confirm no client tool message is synthesized in generateText.

Concerns: none.

### Verification

Inspected the complete main-branch diff and related prompt-conversion and continuation logic. Diff validation passed, both focused test files passed with 279 tests total, and the ai package type-check passed. Approval remains blocked solely because the authoritative exact replay is inconclusive.

aayush-kapoor (approved):
my clanker produced the same fix. i also verified with the repro and this fixed it

PR #17460, opened 2026-07-17
fix(ai): scope UI message tool parts to the current step

### LINKED ISSUE #17347: Duplicate toolCallId across steps causes readUIMessageStream to overwrite earlier tool parts
### Description

```text
```

Expected behavior:

Actual behavior:

### Reproduction

```js

const chunks = [
  { type: 'start-step' },
  {
    type: 'tool-input-available',
    toolCallId: 'call_0',
    toolName: 'recordStep',
    input: { step: 1 },
  },
  {
    toolCallId: 'call_0',
    output: { recorded: 1 },
  },
  { type: 'finish-step' },
  { type: 'start-step' },
  {
    type: 'tool-input-available',
    toolCallId: 'call_0',
    toolName: 'recordStep',
    input: { step: 2 },
  },
  {
    toolCallId: 'call_0',
    output: { recorded: 2 },
  },
  { type: 'finish-step' },
  { type: 'finish' },
];

  start(controller) {
    controller.close();
  },
});

let finalMessage;
  finalMessage = message;
}

);

console.log(toolParts);
```

Output with `ai@7.0.29`:

```text
[
  {
    type: 'tool-recordStep',
    toolCallId: 'call_0',
    state: 'output-available',
    input: { step: 2 },
    output: { recorded: 2 },
  }
]
```

### AI SDK Version

- `ai`: `7.0.29`
- Node.js: `24.14.0`

### Code of Conduct

n0liu:
## Background

Fixes #17347

When multiple model steps are aggregated into one `UIMessage`, a provider can return the same response-scoped tool call id in more than one step. This happens with the OpenAI-compatible Responses API through Amazon Bedrock, where each step restarts the tool call id at `call_0` while the provider item ids (`fc_...`) stay globally unique.

The UI message stream reducer resolved tool parts globally by `toolCallId` (`updateToolPart`, `updateDynamicToolPart`, and `getToolInvocation` all did a `find` over the whole message). So a later step's `call_0` updated the earlier step's tool part instead of appending a new one, and after `toUIMessageStream()` / `readUIMessageStream()` the final message kept a single merged tool part rather than one per step.

The reducer already treats a step boundary as a fresh scope for text and reasoning parts (`activeTextParts` and `activeReasoningParts` are reset on `finish-step`); tool parts were the exception. This change makes tool parts follow the same step-scoping rule, so it does not depend on providers producing globally unique ids.

## Summary

Tool parts are now tracked in a step-scoped `activeToolParts` map alongside `activeTextParts` and `activeReasoningParts`. `updateToolPart`, `updateDynamicToolPart`, and `getToolInvocation` resolve the current tool part through that map and register newly created parts in it, and the map is reset on `finish-step`. It is seeded from the message being continued so a resumed tool call (for example a tool executed after an approval carried over in `lastMessage`) still updates the existing part instead of duplicating it. Provider tool call ids are left untouched for protocol pairing.

## Contributor Credit

Thanks to @wangliang1996 for the detailed report and the self-contained reproduction in #17347.

## End-to-End Verification

Replaying the chunk sequence from the issue (two steps that both use `toolCallId: 'call_0'` with different inputs and outputs, separated by a `finish-step`) through `processUIMessageStream`: before the change the final message contains a single `tool-recordStep` part carrying only the second step's data; after the change it contains two distinct tool parts, one per step, each with its own input and output. The existing approval and resume flows (including tool execution after an approval carried over in `lastMessage`) still resolve their tool part and behave unchanged.

### DIFF
--- a/.changeset/ui-tool-part-step-scope.md
+++ b/.changeset/ui-tool-part-step-scope.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): scope UI message tool parts to the current step so a tool call id reused across steps creates a new tool part instead of overwriting the earlier one
--- a/packages/ai/src/ui/process-ui-message-stream.test.ts
+++ b/packages/ai/src/ui/process-ui-message-stream.test.ts
@@ -9385,4 +9385,73 @@ describe('processUIMessageStream', () => {
       ]);
     });
   });
+
+  describe('tool call id reused across steps', () => {
+    // Reproduces https://github.com/vercel/ai/issues/17347: a provider (e.g. the
+    // OpenAI Responses API through Amazon Bedrock) may return the same
+    // response-scoped tool call id in more than one step. Each step is a
+    // distinct tool call, so it must become its own UI tool part instead of
+    // overwriting the earlier one.
+    beforeEach(async () => {
+      const stream = createUIMessageStream([
+        { type: 'start', messageId: 'msg-1' },
+        { type: 'start-step' },
+        {
+          type: 'tool-input-available',
+          toolCallId: 'call_0',
+          toolName: 'recordStep',
+          input: { step: 1 },
+          providerMetadata: { openai: { itemId: 'fc_step_1' } },
+        },
+        {
+          type: 'tool-output-available',
+          toolCallId: 'call_0',
+          output: { ok: 1 },
+        },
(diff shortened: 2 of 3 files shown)


lgrammel:
Thank you for the pull request. Issue #17347 was closed as completed. Closing this pull request.

PR #17461, opened 2026-07-17
fix: prevent generateText from returning partial results after cancellation in multi-step tool loops

lgrammel:
## Background

Callers relying on cancellation could receive an apparently successful partial result when an abort occurred during tool execution in a multi-step generation.

## Root Cause

Tool execution could normalize an abort exception into a tool-error, or ignore cancellation and return successfully, while `generateText` failed to check the aborted merged signal before starting the next loop iteration.

## Summary

The multi-step `generateText` loop now throws the merged abort signal's reason before beginning any continuation step, independently of the preceding tool result type.

## Testing

Added inline-snapshot regression tests confirming cancellation propagation for both abort-aware throwing tools and abort-insensitive successful tools, with assertions that no second model call occurs. The focused test file and full repository test suite passed.

## End-to-end Validation

- `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-12878-generate-text-abort.ts` — the original reproduction exited successfully after observing propagated `AbortError`; its temporary artifact was removed afterward.

## Related Issues

Fixes #12878

Closes #17450

Co-authored-by: wkusnierczyk <2981661+wkusnierczyk@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: partially-addresses

The new guard propagates the abort reason only when at least one execution result is a tool-error. If a tool ignores cancellation and returns a normal tool-result, generateText can still make another model call with an already-aborted signal and resolve normally.

Concerns:
- The slow tool in the issue's posted example does not observe abortSignal, so this important reported scenario is not covered by the guard.
- A focused simulation confirmed that an abort-insensitive tool can complete after cancellation, allow a second model call, and produce a normal generateText result.

### Side effects

Risk: low

The behavior change is narrowly limited to executions where the overall merged signal is aborted and a tool-error exists; converting that partial result into cancellation is intended.

### Performance

Risk: low

The change adds only a linear scan of the already-materialized tool result array and no additional retained data.

### Backwards compatibility

Risk: none

The change does not read, write, migrate, or reinterpret stored data.

### Security

Risk: none

The change propagates the existing abort reason and introduces no new input handling, credential flow, network access, or trust boundary.

### Testing

Status: needs-more

The regression test validates abort-reason propagation and prevents a second model call when the tool itself throws, but it mirrors the implementation's tool-error condition and misses the abort-insensitive successful-tool case.

Concerns:
- Add a regression test where cancellation occurs during a tool that still returns successfully, and verify generateText rejects before another model call.
- The guard should enforce the aborted operation state independently of whether tool execution produced a tool-error.

### Verification

Reviewed the merge-base diff and generateText/executeToolCall signal flow. The package Node tests and TypeScript type-check passed. A focused cancellation simulation with an abort-insensitive tool demonstrated the unresolved normal-return path.

### Relevant Documentation

- https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-text

lgrammel:
Addressed [PR comment 5002084185](https://github.com/vercel/ai/pull/17461#issuecomment-5002084185).

> ## Bugfix review
> 
> Outcome: changes-required
> 
> ### Reproduction replay
> 
> 
> ... 53 more line(s) omitted

Changed the guard to check the merged abort signal before every continuation step instead of depending on a tool-error result. Added regression coverage for an abort-insensitive tool that returns successfully after cancellation; `generateText` now rejects with the abort reason and the model call count remains one. The original abort-aware tool coverage remains in place.

lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

The loop now checks the merged abort signal before every continuation step, propagating its reason and preventing another model call whether the cancelled tool throws or returns successfully.

### Side effects

Risk: low

The behavior change is limited to multi-step continuations whose merged signal is already aborted; rejecting instead of continuing is the intended cancellation behavior.

### Performance

Risk: none

The change adds one constant-time signal-state check per continuation and retains no additional data.

### Backwards compatibility

Risk: none

The change does not read, write, migrate, or reinterpret stored data and does not alter public types or function signatures.

### Security

Risk: none

The implementation only propagates an existing abort reason and introduces no new input parsing, network access, credential handling, or trust boundary.

### Testing

Status: appropriate

Regression tests cover both an abort-aware throwing tool and an abort-insensitive successful tool, assert AbortError propagation, and verify that no second model call occurs.

### Verification

Reviewed the complete merge-base diff and cancellation flow through generateText and tool execution. The focused generateText suite passed all 259 tests in both Node and Edge configurations, package type-checking passed, and the diff passed whitespace validation.

PR #17467, opened 2026-07-17
[v6.0] reproduction: uiMessageChunkSchema strict validation breaks backward due to providerMetadata

lgrammel:
## Bug

uiMessageChunkSchema strict validation breaks backward due to providerMetadata addition

## Reproduction

- `packages/ai/src/ui-message-stream/ui-message-chunks.ts` in `ai@6.0.230` still defines UI message chunk variants with `z.strictObject()`, rejecting fields unknown to the client schema.
- `examples/ai-functions/src/reproduction/ui-message-chunk-forward-compatibility.ts` confirms that baseline chunks and currently recognized `providerMetadata` and `toolMetadata` fields parse successfully, while a field added by a newer server causes `DefaultChatTransport` to throw `AI_TypeValidationError`.
- The expected behavior is for older clients to accept or ignore additional fields on known chunk types instead of terminating the UI message stream.
- Historical `ai@6.0.119` and `ai@6.0.140` bundles were not installed; the target `ai@6.0.230` branch reproduces the same forward-compatibility failure using an unknown newer-server field.

## Related Issues

Reproduces #13733

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "issueId": 13733,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/ui-message-chunk-forward-compatibility.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/ui-message-chunk-forward-compatibility.ts"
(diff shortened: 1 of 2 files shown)


lgrammel:
Issue #13733 already has the merged-v6.0 label. Closing this stale reproduction pull request.

Issue #13448, opened 2026-03-15
activeTools still executes disabled tools and breaks prepareStep tool disabling


0ptim:
### Description

`activeTools` still allows execution of disabled tools.

This is still reproducible on `ai@6.0.116` and appears to be the same underlying problem as [`vercel/ai#8653`](https://github.com/vercel/ai/issues/8653), which was closed due to inactivity.

Minimal deterministic repro (provider-independent):

```ts
import { streamText, tool, simulateReadableStream, stepCountIs } from "ai";
import { MockLanguageModelV3 } from "ai/test";
import { z } from "zod";

const usage = {
  inputTokens: { total: 1, noCache: 1, cacheRead: 0, cacheWrite: 0 },
  outputTokens: { total: 1, text: 1, reasoning: 0 },
};

let executed = false;

const model = new MockLanguageModelV3({
  doStream: async () => ({
    stream: simulateReadableStream({
      initialDelayInMs: 0,
      chunkDelayInMs: 0,
      chunks: [
        {
          type: "tool-call",
          toolCallId: "call_1",
          toolName: "weather",
          input: JSON.stringify({ location: "Basel" }),
        },
        {
          type: "finish",
          finishReason: { unified: "tool-calls", raw: undefined },
          usage,
        },
      ],
    }),
  }),
});

const result = streamText({
  model,
  prompt: "test",
  tools: {
    weather: tool({
      inputSchema: z.object({ location: z.string() }),
      execute: async ({ location }) => {
        executed = true;
        return `weather for ${location}`;
      },
    }),
  },
  activeTools: [],
  stopWhen: stepCountIs(1),
});

await result.consumeStream();

console.log({
  executed,
  toolCalls: await result.toolCalls,
  toolResults: await result.toolResults,
});
```

Actual result:

- `executed === true`
- the disabled tool call is accepted
- the disabled tool is executed successfully
- `toolResults` contains the result

Expected result:

- tools not in `activeTools` should be treated as unavailable
- the tool call should be rejected or surfaced as a `NoSuchTool`-style error
- `execute` must not run

This also affects `prepareStep` use cases where tools are disabled on later steps to force a final text response.

From reading the current code, the provider call correctly receives filtered tools, but tool parsing/execution still appears to validate against the full `tools` object instead of the filtered active subset.

### AI SDK Version

- ai: 6.0.116
- @ai-sdk/react: 3.0.118

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

[...]

PR #17468, opened 2026-07-17
[v6.0] reproduction: activeTools still executes disabled tools and breaks prepareStep

lgrammel:
## Bug

activeTools still executes disabled tools and breaks prepareStep tool disabling

## Reproduction

- Added `examples/ai-functions/src/reproduction/issue-13448-active-tools-execution.ts` covering both direct `activeTools: []` and later-step disabling through `prepareStep`.
- On `ai@6.0.230`, the direct scenario sent zero tools to the provider, but the disabled `weather` tool executed once and produced a successful tool result.
- The `prepareStep` scenario sent one tool on step 1 and zero tools on step 2, but `weather` executed on both steps instead of only the enabled first step.
- Expected disabled tool calls to be treated as unavailable and their `execute` functions not to run; both reported behaviors were reproduced.

## Relevant Documentation

- https://ai-sdk.dev/docs/ai-sdk-core/tools-and-tool-calling

## Related Issues

Reproduces #13448

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "issueId": 13448,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-13448-active-tools-execution.ts",
+  "artifactPaths": [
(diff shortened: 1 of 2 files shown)


lgrammel:
Issue #13448 already has the merged-v6.0 label. Closing this stale reproduction pull request.

PR #17478, opened 2026-07-17
[v5.0] reproduction: uiMessageChunkSchema strict validation breaks backward due to providerMetadata

lgrammel:
## Bug

uiMessageChunkSchema strict validation breaks backward due to providerMetadata addition

## Reproduction

- Target `ai@5.0.216` uses `z.strictObject()` for UI message chunk variants in `packages/ai/src/ui-message-stream/ui-message-chunks.ts`.
- `examples/ai-functions/src/reproduction/ui-message-chunk-forward-compatibility.ts` confirmed baseline tool chunks parse, but `DefaultChatTransport` terminates parsing with `AI_TypeValidationError` when known chunk types contain the reported newer-server `providerMetadata` or `toolMetadata` fields.
- Expected behavior is to accept or ignore additional fields on known chunk types so cached older clients can continue consuming streams.
- The exact historical v6 package combinations were not evaluated because the mandated target is release-v5.0; the target's equivalent client-stream behavior reproduced with the reported payload shapes.
- `examples/ai-functions/package.json` and `pnpm-lock.yaml` provide the workspace setup required to replay the reproduction.

## Related Issues

Reproduces #13733

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,14 @@
+{
+  "version": 1,
+  "issueId": 13733,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/ui-message-chunk-forward-compatibility.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/ui-message-chunk-forward-compatibility.ts",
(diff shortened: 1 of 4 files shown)


lgrammel:
Issue #13733 already has the merged-v5.0 label. Closing this stale reproduction pull request.

PR #17493, opened 2026-07-17
fix(ai): use injective serialization for tool approval HMAC payload

### LINKED ISSUE #17832: Tool approval signatures can collide across fields containing newlines
## Problem

Tool approval HMAC signatures enabled through `experimental_toolApprovalSecret` do not unambiguously bind the signature to the `(approvalId, toolCallId, toolName, input)` tuple.

The signed fields are currently separated with newline characters. Because identifiers such as `toolCallId` and `toolName` may themselves contain newlines, distinct tuples can produce the same HMAC payload. For example, these tuples are serialized identically:

```ts
{
  approvalId: 'approval-1',
  toolCallId: 'call-1',
  toolName: 'searchDocs\ndeleteFile',
  input: { path: '/tmp/target' },
}
```

```ts
{
  approvalId: 'approval-1',
  toolCallId: 'call-1\nsearchDocs',
  toolName: 'deleteFile',
  input: { path: '/tmp/target' },
}
```

As a result, a signature issued for one tuple can verify for another tuple, potentially allowing an approval to be reused for a different registered tool.

## Expected behavior

- A tool approval signature must verify only for the exact `approvalId`, `toolCallId`, `toolName`, and canonicalized `input` for which it was issued.
- Field boundaries must remain unambiguous even when string fields contain newlines or other delimiter/control characters such as `\r`, `\t`, null bytes, quotes, or backslashes.
- Distinct tuples must produce distinct signatures, while equivalent inputs with different object key ordering must continue to produce the same signature.
- Valid signatures must continue to verify in both Node.js and Edge runtimes.

## Backwards compatibility

Approvals may remain pending while an application is upgraded. Existing signatures produced by the previous format should continue to verify when all identifier fields are newline-free, since those payloads are unambiguous.

Legacy signatures involving newline-bearing fields must not be accepted, including through any compatibility path, because accepting them would preserve the tuple-collision vulnerability.

## Related

Extracted from #17493

dnukumamras:
The tool approval signature payload was built by joining fields with `\n`. Because fields like `toolName` and `toolCallId` can contain a newline, two different field tuples could serialize to identical bytes — so a signature issued for one tuple could verify against a different one. The payload is now serialized with `JSON.stringify` (with a constant domain-separation prefix), which escapes delimiter/control characters and makes the encoding injective.

Verification stays backwards compatible: it falls back to the old newline format only when no signed field contains a `\n`, so approvals signed by an older version still verify across an upgrade. This legacy fallback is marked `TODO: remove in v8`.

**Happy path**
- A valid approval signs and verifies for its own `(approvalId, toolCallId, toolName, input)` tuple.
- Input key ordering is still normalized, so equivalent inputs produce the same signature.
- A signature in the old newline format still verifies (delimiter-free fields), so a pending approval that straddles an upgrade is not rejected.

**Unhappy path**
- A tuple whose `toolName` contains a newline can no longer be retupled into `toolCallId` to reuse the same signature for a different tool — in both the new and the legacy fallback paths.
- Any delimiter/control character (`\n`, `\r`, `\t`, `\0`, `"`, `\\`) shifted across a field boundary fails verification.
- A legacy signature over a newline-bearing field no longer verifies (the fallback is refused when a field contains `\n`).

Fixes #17832

### DIFF
--- a/.changeset/tool-approval-injective-payload.md
+++ b/.changeset/tool-approval-injective-payload.md
@@ -11,3 +11,8 @@ serialize to identical bytes, allowing a signed approval to verify against a
 different tuple. The payload is now serialized with `JSON.stringify` (with a
 versioned domain-separation prefix), which escapes delimiter/control characters
 and makes the encoding injective.
+
+Verification remains backwards compatible: a signature in the old format still
+verifies, but only when no field contains the `\n` delimiter (the condition
+that made the old format ambiguous), so a pending approval that straddles an
+upgrade is not rejected while the collision stays closed.
--- a/.changeset/tool-approval-injective-payload.md
+++ b/.changeset/tool-approval-injective-payload.md
@@ -1,0 +1,13 @@
+---
+'ai': patch
+---
+
+fix(ai): use injective serialization for tool approval HMAC payload
+
+The tool approval signature (`experimental_toolApprovalSecret`) built its HMAC
+payload by joining fields with `\n`. Because fields such as `toolName` and
+`toolCallId` can themselves contain a newline, distinct field tuples could
+serialize to identical bytes, allowing a signed approval to verify against a
+different tuple. The payload is now serialized with `JSON.stringify` (with a
+versioned domain-separation prefix), which escapes delimiter/control characters
+and makes the encoding injective.
--- a/packages/ai/src/generate-text/tool-approval-signature.test.ts
+++ b/packages/ai/src/generate-text/tool-approval-signature.test.ts
@@ -94,4 +94,72 @@ describe('signToolApproval + verifyToolApprovalSignature', () => {
     });
     expect(sig1).toBe(sig2);
   });
+
+  // A newline in toolName must not be retupleable into toolCallId to forge a
+  // matching signature for a different tool.
+  it('should not collide when a newline in toolName is retupled into toolCallId', async () => {
+    const signed = {
+      approvalId: 'approval-1',
+      toolCallId: 'call-1',
+      toolName: 'searchDocs\ndeleteFile',
+      input: { path: '/tmp/target' },
+    };
+    const retupled = {
+      approvalId: 'approval-1',
+      toolCallId: 'call-1\nsearchDocs',
+      toolName: 'deleteFile',
+      input: { path: '/tmp/target' },
+    };
+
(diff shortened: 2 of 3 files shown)


lgrammel:
what happens re backwards compat? what about people who use v7 in prod and have the old signatures

dnukumamras:
> what happens re backwards compat? what about people who use v7 in prod and have the old signatures

This only impacts for signatures that are pending ones.

The edge case is if a tool the model requested (signed by the old server) whose human approval arrives on a turn handled by the new server, before the tool has executed. New server recomputes the JSON payload, HMAC differs, and throws a hard error

lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The JSON-array payload preserves field boundaries, prevents newline retupling collisions, and safely supports legacy delimiter-free signatures.

### Side effects

Risk: low

Normal signing and verification behavior remains intact, canonical input hashing is unchanged, and only unsafe legacy signatures containing newlines are deliberately rejected.

### Performance

Risk: low

Verification performs one additional HMAC check only for legacy or invalid signatures; JSON serialization adds only a small bounded payload allocation for typical identifiers.

### Backwards compatibility

Risk: low

Existing delimiter-free pending signatures continue to verify without data migration; legacy signatures containing newlines are intentionally invalidated to close the collision.

### Architecture

Risk: low

The implementation remains localized to the owning internal tool-approval module, preserves dependency direction, and introduces no cross-package source access or new dependencies.

### Change scope

Status: minimal

The diff contains only the serialization and compatibility implementation, focused regression tests, and the required patch changeset.

### Security

Risk: none

The injective JSON encoding closes the signature-reuse vulnerability, while the legacy fallback is restricted to tuples for which newline serialization is unambiguous.

### Testing

Status: appropriate

Tests cover valid signatures, canonical input ordering, the reported newline retupling attack, multiple control characters, safe legacy verification, and legacy collision rejection in both Node.js and Edge runtimes.

### Verification

Reviewed every changed hunk and the signing and validation call paths. The focused signature and approval-validation suites passed in Node.js and Edge runtimes, workspace type checking passed, formatting and lint checks passed, and the diff had no whitespace errors.

lgrammel (approved):

PR #17499, opened 2026-07-18
fix: emit terminal finish part when streamText terminates on an error

### LINKED ISSUE #17500: UI message stream onEnd cannot distinguish a failed response from a completed one
### Problem

`onEnd` is a stream lifecycle callback, but it does not report the lifecycle outcome.

The same payload can represent both a successful response and a fatal failure:

```ts
{ isAborted: false, finishReason: undefined }
```

This is an information gap in the API, not primarily a reducer bug or a missing-chunk bug. Three concepts are currently conflated:

- `finishReason` is optional metadata describing why a model stopped generating.
- An `error` chunk is a stream event; it may be fatal, recoverable, or produced by an auxiliary merged stream.
- The response outcome is the operation-level verdict: completed, failed, aborted, or unknown.

Neither `finishReason` nor the presence of an error chunk can reliably reconstruct that verdict. Only the owner of the observed stream boundary can declare it: `streamText` owns the generation outcome, while a composer owns the aggregate outcome of a merged UI stream.

### User impact

A consumer that persists durable state in `onEnd` can record a failed response as completed. In one production trace:

| t | Event |
|---|---|
| ~176 ms | A follow-up step request is rejected. An `error` part is enqueued and the source closes without a `finish` part. |
| ~210 ms | `onEnd` fires with `finishReason: undefined`; the consumer persists the response as completed. |
| ~218 ms | The terminal error chunk clears downstream buffering and reaches the client. |

The timing exposes the bug, but callback ordering is not its root cause. Even if the callback were moved, a generic consumer still could not decide whether an error in a composed stream was fatal.

### Why consumers cannot infer the outcome

A consumer can try to join `onEnd` with separately observed chunks:

```ts
let sawError = false;

  onEnd: ({ isAborted, finishReason }) => {
    const failed = sawError || finishReason === 'error';
    persistOutcome(isAborted ? 'interrupted' : failed ? 'failed' : 'completed');
  },
});

  if (chunk.type === 'error') sawError = true;
  send(chunk);
}
```

But this is not generally correct:

- `finishReason` may be absent on a successful stream and describes model termination, not the enclosing operation.
- Recoverable tool errors can invoke error handling without failing the response.
- A failed auxiliary merged stream produces the same top-level error chunk as a fatal main-stream failure.
- The inner `onEnd` may run before an error chunk drains through downstream transforms.

The consumer is being asked to recreate failure policy that belongs to the producer or composer, without enough information to do so.

### Current manifestation in `streamText`

Three `streamText` paths emit an `error` part and close without emitting a `finish` part: a failed follow-up step request, a failed initial request, and a model stream that ends without producing output.

```ts
try {
  await streamStep({ currentStep: currentStep + 1, ... });
} catch (error) {
  controller.enqueue({ type: 'error', error });
  self.closeStream();
}
```

UI message stream processing only assigns `finishReason` when it receives a `finish` chunk. These failures therefore reach `onEnd` with `finishReason: undefined`, which is also valid for a successful custom stream.

### Required capability

Every observed stream boundary needs an owner-declared terminal outcome, delivered to `onEnd` exactly once:

- `completed` - the operation completed according to that stream owner's policy.
- `failed` - the operation failed.
- `aborted` - the operation was intentionally interrupted.
- `unknown` - the stream ended without an explicit declaration.

For composed streams, the composer must decide whether a child-stream failure fails the aggregate response. Consumers should read that verdict rather than infer it from model metadata or arbitrary chunks.

This requirement does not prescribe a particular API shape or wire-format change.

### One possible API shape (non-normative)

```ts
type StreamOutcome =
  | { status: 'completed' }
  | { status: 'failed'; error?: unknown }
  | { status: 'aborted' }
  | { status: 'unknown' };

onEnd({ outcome, finishReason, isAborted, messages, ... });
```

`finishReason` remains model metadata. A composer could declare its policy explicitly with operations such as `writer.complete()`, `writer.fail(error)`, or `writer.abort()`. EOF without a declaration produces `unknown`.

A consumer must handle ambiguity explicitly rather than silently treating it as success:

```ts
onEnd: ({ outcome }) => {
  switch (outcome.status) {
    case 'completed':
      return persistOutcome('completed');
    case 'failed':
      return persistOutcome('failed');
    case 'aborted':
      return persistOutcome('interrupted');
    case 'unknown':
      return persistOutcome('unknown');
  }
}
```

### Why narrower fixes are insufficient

- Setting `finishReason: 'error'` whenever an error chunk is reduced lets an auxiliary or recoverable error overwrite successful model metadata.
- Emitting `finish(error)` from the known `streamText` failure paths addresses those paths, but conflates operation outcome with model finish reason and does not solve custom or merged-stream outcomes.
- Reordering `onEnd` may address a particular race, but still leaves consumers unable to apply the composer's failure policy.

The invariants exercised in #17499 remain valuable: an auxiliary error after `finish(stop)` must not rewrite `stop`, a reason-less finish must not erase an existing reason, and a merged-stream failure must not redefine the main model's finish reason.

### Reproduction

```ts
import { createUIMessageStream } from 'ai';

const stream = createUIMessageStream({
  execute({ writer }) {
    writer.write({ type: 'start' });
    writer.write({ type: 'text-start', id: 't1' });
    writer.write({ type: 'text-delta', id: 't1', delta: 'partial output' });
    writer.write({ type: 'text-end', id: 't1' });
    writer.write({ type: 'error', errorText: 'Internal server error' });
  },
  onError: () => 'An error occurred.',
  onEnd: ({ finishReason, isAborted }) => {
    console.log({ finishReason, isAborted });
    // { finishReason: undefined, isAborted: false }
    // Indistinguishable from a successful custom stream.
  },
});

for await (const _ of stream) {
  // drain
}
```

### AI SDK versions

The same behavior exists on `main`, `release-v6.0`, and `release-v5.0`: UI stream processing only assigns `finishReason` when it receives a `finish` chunk, and `onEnd` has no separate operation outcome.

bryan-hunter:
## Background

`onEnd`/`onFinish` get `finishReason` from the UI message state, and only `finish` chunks write it:

```ts
// packages/ai/src/ui/process-ui-message-stream.ts
case 'finish': {
  if (chunk.finishReason != null) {
    state.finishReason = chunk.finishReason;   // ← the only assignment
  }
  ...
}
```

But three `streamText` paths end the stream with an `error` part and never emit a `finish` part:

```ts
// packages/ai/src/generate-text/stream-text.ts (step continuation)
try {
  await streamStep({ currentStep: currentStep + 1, ... });
} catch (error) {
  controller.enqueue({ type: 'error', error });
  self.closeStream();                          // ← no { type: 'finish' } is ever emitted
}
```

The initial-request catch and the no-output flush path do the same. Mid-step errors are fine: they set `stepFinishReason = 'error'` and the terminal `finish` part carries it.

So on those paths the stream ends `… → error → (close)`: `result.finishReason` resolves `'other'`, `fullStream` has no terminal event, and `onEnd` reports `finishReason: undefined` for a stream that failed. Timing from a production trace (relative to stream start):

| t | Event |
|---|---|
| ~176 ms | follow-up step request rejected → `error` part enqueued, source closes with no `finish` part |
| ~210 ms | `onEnd` fires with `finishReason: undefined` — the consumer persists the response as **completed** |
| ~218 ms | the terminal `error` chunk drains through downstream transforms and reaches the client |

This isn't a race the consumer can win. `onEnd` runs in the stream's terminal `flush()`, before the error chunk clears downstream buffering, so the callback never sees the error no matter how it's written.

Today you have to track error chunks out of band and join the two signals:

```ts
let sawErrorChunk = false;

  onEnd: ({ isAborted, finishReason }) => {
    const failed = sawErrorChunk || finishReason === 'error';
    persistOutcome(isAborted ? 'interrupted' : failed ? 'failed' : 'completed');
  },
});

  if (chunk.type === 'error') sawErrorChunk = true;
  send(chunk);
}
```

With the fix, `onEnd` alone is enough:

```ts
  onEnd: ({ isAborted, finishReason }) => {
    persistOutcome(
      isAborted ? 'interrupted' : finishReason === 'error' ? 'failed' : 'completed',
    );
  },
});
```

## Summary

Fix the producer instead of reinterpreting error chunks: the three terminating paths now emit a proper terminal `finish` part before closing.

```ts
} catch (error) {
  controller.enqueue({ type: 'error', error });

  // terminate the stream with a proper finish part so consumers
  // (finishReason promise, fullStream, UI message stream onEnd)
  // observe the error termination instead of a stream that ends
  // without a finish event:
  controller.enqueue({
    type: 'finish',
    finishReason: 'error',
    rawFinishReason: undefined,
    totalUsage: combinedUsage,
  });

  self.closeStream();
}
```

`'error'` is already part of the `FinishReason` vocabulary ("the model stopped because of an error"), and these paths are model-call failures, so the reported reason stays within the documented contract. Error chunks remain non-finish events everywhere.

What changes:

- follow-up step failure: `result.finishReason` resolves `'error'` (was `'other'`), `fullStream` ends with a `finish` part, UI `onEnd` reports `finishReason: 'error'` (was `undefined`/stale)
- initial request failure: the stream is `start → error → finish(error)`; result promises still reject exactly as before (the event processor's no-steps guard runs first)
- no-output stream: same terminal `finish(error)`; result promises still reject as before

What deliberately does not change (each pinned by a new test):

- an `error` chunk without a `finish` chunk does not invent a finish reason
- a successful `finish(stop)` followed by an auxiliary error chunk (e.g. a merged side stream failing) keeps `finishReason: 'stop'`
- a merged stream failing in `createUIMessageStream` does not affect the main stream's finish reason
- a later reason-less `finish` chunk does not overwrite anything
- client `Chat` behavior: its error handler throws, so processing stops at the error chunk — `isError` and `finishReason` keep their documented, separate meanings

## End-to-End Verification

Drove the three failure shapes through `streamText` with mock models: each now ends `fullStream` with `finish(finishReason: 'error')`, the follow-up-step case resolves `result.finishReason` to `'error'`, and `toUIMessageStream`'s `onEnd` reports `'error'` end to end. Success, recovery, and abort paths are unchanged across the full package suite (3,254 node + edge tests).

## Checklist

## Future Work

`streamText`'s own `onEnd` callback event still reports `finalStep.finishReason` (the last recorded step's reason, e.g. `'tool-calls'`) rather than the overall termination reason — worth aligning with `result.finishReason` separately, since changing that event shape has its own compatibility questions.

## Related Issues

Fixes #17500

### DIFF
--- a/.changeset/tidy-errors-finish.md
+++ b/.changeset/tidy-errors-finish.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix: emit a terminal finish part with `finishReason: 'error'` when `streamText` terminates on an error (failed follow-up step request, failed initial request, or a model stream that produced no output), so `result.finishReason`, `fullStream`, and UI message stream `onEnd`/`onFinish` observe the error termination instead of a stream that ends without a finish event
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -45,6 +45,7 @@ import { z } from 'zod/v4';
 import { Output, type LanguageModelCallEndEvent, type Telemetry } from '..';
 import * as logWarningsModule from '../logger/log-warnings';
 import type { Instructions } from '../prompt';
+import { NoOutputGeneratedError } from '../error/no-output-generated-error';
 import { MockLanguageModelV4 } from '../test/mock-language-model-v4';
 import { createMockServerResponse } from '../test/mock-server-response';
 import { mockValues } from '../test/mock-values';
@@ -2382,6 +2383,12 @@ describe('streamText', () => {
           type: 'error',
           error: new Error('test error'),
         },
+        {
+          type: 'finish',
+          finishReason: 'error',
+          rawFinishReason: undefined,
+          totalUsage: createNullLanguageModelUsage(),
+        },
       ]);
     });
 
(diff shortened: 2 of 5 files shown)


privatenumber (approved):

bryan-hunter:
Closing in favor of the design discussion on #17500. Review established that neither variant here is the right layer: the reducer version let auxiliary merged-stream failures rewrite a successful finish reason, and the producer version conflates operation outcome with the model's finish reason and changes public fullStream sequences. The issue now proposes an explicit, producer-declared outcome on onEnd instead. The invariant-pinning tests in this branch (aux error after finish(stop) keeps stop, reason-less finish overwrites nothing, merged failures don't touch the main stream's reason) are worth cherry-picking under any design.

PR #17501, opened 2026-07-18
feat: add experimental_streamMode 'single-consumer' to streamText

### LINKED ISSUE #16753: streamText textStream tee path retains large streaming outputs in memory
## Description

`streamText().textStream` always goes through the `DefaultStreamTextResult.stream` getter, which calls `teeStream()`. `teeStream()` uses `ReadableStream.tee()` and the current source comment explicitly notes that this buffers stream content on the server, assuming LLM outputs are small enough not to cause issues.

That assumption does not hold for long-running / large-output server streams. In our Cloudflare Worker Magic Import workload, using `streamText().textStream` for long generations caused high memory p99 / sparse memory-limit breaches. We had to add a production-only bypass that consumes the provider `doStream()` result once and reads `text-delta` parts directly, avoiding the public AI SDK `textStream` path.

## Source

Current `streamText` creates one shared `baseStream`:

https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.ts#L1521-L1523

```ts
this.baseStream = stream
  .pipeThrough(createOutputTransformStream(output ?? text()))
  .pipeThrough(eventProcessor);
```

Every stream projection calls `teeStream()`:

https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.ts#L2479-L2494

```ts
/**
 * Split out a new stream from the original stream.
 * The original stream is replaced to allow for further splitting,
 * since we do not know how many times the stream will be split.
 *
 * Note: this leads to buffering the stream content on the server.
 * However, the LLM results are expected to be small enough to not cause issues.
 */
private teeStream() {
  const [stream1, stream2] = this.baseStream.tee();
  this.baseStream = stream2;
  return stream1;
}

get textStream(): AsyncIterableStream<string> {
  return createAsyncIterableStream(toTextStream({ stream: this.stream }));
}
```

`consumeStream()` also consumes `this.stream`, which creates another tee:

https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.ts#L2540-L2544

```ts
async consumeStream(options?: ConsumeStreamOptions): Promise<void> {
  try {
    await consumeStream({
      stream: this.stream,
```

## Impact

For large streaming responses, this makes the memory profile depend on internal result retention and stream tee buffering, even when the caller only wants to consume text deltas once.

In our workload:

- The hot path only needed text deltas.
- We did not need `steps`, `finalStep`, `responseMessages`, `usage`, or other final result projections while streaming.
- The public `textStream` API still routed through `teeStream()` and retained enough data to push Cloudflare Worker memory p99 / tail usage over the limit during long generations.
- A single-consumer provider-stream bypass reduced the memory risk without changing the model/provider output semantics.

The relevant summary from that change:

- add a Magic Import-only provider-stream mode that consumes AI SDK v3 provider `doStream()` output once
- avoid the public `streamText().textStream` tee/final-result retention path for Magic Import v4
- preserve the existing SDK path for other callers

## Expected

There should be a supported low-retention / single-consumer streaming mode for callers that only need text deltas or stream parts and do not need final result materialization.

Possible API shapes:

- An option on `streamText`, e.g. `experimental_streamMode: 'single-consumer'`, where `textStream` or `stream` consumes the underlying model stream directly and disables final result projections.
- A separate helper that returns only a text-delta stream / stream-part stream without constructing `DefaultStreamTextResult`.
- A documented way to opt out of `tee()` / final result retention for high-volume server streaming workloads.

## Actual

The public `streamText().textStream` path tees `baseStream` and keeps the final result machinery active. The source comment acknowledges server-side buffering and assumes generated output is small.

For large outputs, that assumption can be false and there is no public low-retention alternative short of bypassing `streamText` internals and calling provider `doStream()` directly.

## Reproduction shape

This is easiest to observe with a provider/mock that emits many large text deltas and a Worker/Node process memory profiler:

```ts

const result = streamText({
  model: longOutputModel,
  prompt: 'Emit a very large response',
});

for await (const delta of result.textStream) {
  // Caller only needs each delta once.
}
```

Expected memory behavior for a single-consumer text stream is roughly bounded by parser / transform buffers. Actual behavior includes `tee()` buffering and final-result accumulation machinery.

## Environment

- `ai`: current main branch source still has the `teeStream()` implementation and buffering note
- Runtime where observed: Cloudflare Workers, 128 MB memory limit

Aayush-engineer:
## Background

`streamText().textStream` (and every other stream/result accessor) routes through `teeStream()`, which calls `ReadableStream.tee()` to allow multiple independent readers. The source comment on `teeStream()` already notes this buffers stream content on the server and assumes LLM output is small enough not to matter.

That assumption breaks down for long-running/large-output streams. #16753 reports high memory p99 and memory-limit breaches in a 128 MB Cloudflare Worker when consuming `textStream` for long generations, caused by this buffering. The reporter had to bypass `streamText` internals entirely and read `doStream()` directly to avoid it.

## Summary

Adds an opt-in `experimental_streamMode: 'single-consumer'` option to `streamText()`. When set, `teeStream()` skips `ReadableStream.tee()` entirely and hands the internal `baseStream` directly to the first caller. Any second attempt to read the stream — via `textStream`, `stream`/`fullStream`, `consumeStream()`, `partialOutputStream`, `elementStream`, or a final-result getter (`usage`, `steps`, `finishReason`, etc.) — throws an `AISDKError`, since there's no buffered copy left for a second reader.

Reading a final-result getter *after* another consumer has already fully drained the stream still works, since the result promises are already resolved by that point — the restriction only applies to reading the underlying stream a second time.

## Contributor Credit

Thanks to @matthew-gizmo for the original report and reproduction in #16753.

## End-to-End Verification

- Ran the full `stream-text.test.ts` suite (408 existing tests) — all pass unmodified.
- Added 4 new tests covering: `textStream` consumption in single-consumer mode, reading a single final-result getter, throwing on a second consumer access before the first drains, and throwing synchronously on a direct double-access of `stream`.
- `tsc --build` and `ultracite check` (repo lint/format) both pass clean.
- Have **not** independently measured the actual memory/p99 improvement under load — this PR only verifies functional correctness of the new code path, not the memory characteristics that motivated #16753. Would appreciate maintainer input on whether that's expected before merge, or acceptable to validate separately.

## Future Work

- `partialOutputStream` and `elementStream` go through the same `teeStream()` path and should behave the same way in single-consumer mode, but aren't covered by dedicated tests yet.
- Possible follow-up: measure actual memory/p99 impact in a Worker-like environment to confirm this resolves the original report's numbers, not just the API shape.

## Related Issues

Fixes #16753

### DIFF
--- a/.changeset/honest-pillows-tap.md
+++ b/.changeset/honest-pillows-tap.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+feat: add experimental_streamMode: 'single-consumer' to streamText to avoid tee() buffering for single-reader streams
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -5221,6 +5221,51 @@ describe('streamText', () => {
     });
   });
 
+  describe('options.experimental_streamMode', () => {
+    it('should allow textStream to be consumed in single-consumer mode', async () => {
+      const result = streamText({
+        model: createTestModel(),
+        prompt: 'test-input',
+        experimental_streamMode: 'single-consumer',
+      });
+      expect(
+        await convertAsyncIterableToArray(result.textStream),
+      ).toStrictEqual(['Hello', ', ', 'world!']);
+    });
+    it('should allow a single final-result getter to be read in single-consumer mode', async () => {
+      const result = streamText({
+        model: createTestModel(),
+        prompt: 'test-input',
+        experimental_streamMode: 'single-consumer',
+      });
+      expect(await result.usage).toMatchObject({
+        inputTokens: 3,
+        outputTokens: 10,
+        totalTokens: 13,
+      });
+    });
+    it('should throw when a second stream getter is accessed before the first has been read', async () => {
+      const result = streamText({
+        model: createTestModel(),
+        prompt: 'test-input',
+        experimental_streamMode: 'single-consumer',
+      });
+      const textStreamPromise = convertAsyncIterableToArray(result.textStream);
+      expect(() => result.stream).toThrow(/already been consumed/);
+      await textStreamPromise;
+    });
+    it('should throw synchronously when the stream getter is accessed a second time', () => {
+      const result = streamText({
+        model: createTestModel(),
+        prompt: 'test-input',
+        experimental_streamMode: 'single-consumer',
+      });
+      // biome-ignore lint: accessing the getter is the point of the test
+      result.stream;
+      expect(() => result.stream).toThrow(/already been consumed/);
+    });
+  });
+
   describe('multiple stream consumption', () => {
     it('should support text stream, ai stream, full stream on single result object', async () => {
       const result = streamText({
--- a/packages/ai/src/generate-text/stream-text.ts
+++ b/packages/ai/src/generate-text/stream-text.ts
@@ -1,4 +1,5 @@
 import {
+  AISDKError,
   getErrorMessage,
   UnsupportedFunctionalityError,
   type LanguageModelV4,
@@ -387,6 +388,7 @@ export function streamText<
   toolsContext = {} as InferToolSetContext<TOOLS>,
   experimental_include,
   include = experimental_include,
+  experimental_streamMode,
   _internal: {
     now = originalNow,
     generateId = originalGenerateId,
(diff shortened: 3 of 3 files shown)


matthew-gizmo:
Thanks for working on this. I’m the reporter of #16753; the affected workload is a Cloudflare Worker that streams unusually large LLM outputs under a 128 MB memory limit. We only consume each text delta once and do not need final results such as steps, response messages, or usage.

Skipping `ReadableStream.tee()` addresses one important source of memory growth. However, I don’t think this implementation yet provides the fully low-retention path the issue needs.

Even in `single-consumer` mode, the stream still passes through `eventProcessor`. That processor reconstructs and retains the complete generated text and stores content, step results, and response messages for the final-result APIs. Therefore, memory remains proportional to total output size, despite there being only one stream consumer.

Our production workaround bypasses `streamText` and consumes the provider’s `doStream()` result directly. This avoids both:

1. Buffering from the unused `tee()` branch.
2. Retention for `streamText`’s final-result materialization.

Could `single-consumer` mode also disable final-result materialization once a stream accessor is selected? It would be acceptable for final-result getters to become unavailable in that case. Alternatively, a supported low-retention text-delta API below `DefaultStreamTextResult` would solve the same problem.

PR #17502, opened 2026-07-18
feat(harness-claude-code): support inline images in user prompts

### LINKED ISSUE #17082: @ai-sdk/harness-claude-code: user message file/image parts are rejected (HarnessCapabilityUnsupportedError) — no vision input path through HarnessAgent
## Description

The claude-code harness adapter accepts **only text parts** in a user prompt. Any `file` part (the shape `useChat`'s standard attachment flow produces on user messages) throws before the turn starts:

```
HarnessCapabilityUnsupportedError: The claude-code harness does not yet support user
message parts of type 'file'. Pass a string or a user message whose content contains
only text parts.
```

So there is currently no vision path through `HarnessAgent` + `createClaudeCode()` at all — even though the underlying Claude Agent SDK accepts base64 image content blocks in streaming input mode, which is the mode the bridge already uses.

## Reproduction

```ts
const agent = new HarnessAgent({ harness: createClaudeCode(), sandbox });
const session = await agent.createSession({ sessionId: chatId });

await agent.stream({
  session,
  messages: [
    {
      role: 'user',
      content: [
        { type: 'text', text: 'What is in this screenshot?' },
        { type: 'file', mediaType: 'image/png', data: pngBase64 },
      ],
    },
  ],
});
// → HarnessCapabilityUnsupportedError (thrown from extractUserText)
```

## Root cause

Verified against the published `@ai-sdk/harness-claude-code@1.0.24` dist (latest, released today; same on 1.0.18 where we first hit it):

- `extractUserText` (`dist/index.js`) throws for every `part.type !== 'text'`.
- The bridge (`dist/bridge/index.mjs`) hardcodes the SDK wire message to text blocks:

```js
type: "user",
message: {
  role: "user",
  content: [{ type: "text", text }]
}
```

The Claude Agent SDK's streaming input mode accepts `content` arrays with base64 `image` blocks, so the limitation is in the adapter/bridge transcription, not the underlying runtime.

## Suggested fix

- Accept `file` parts with **image** media types in `extractUserText` (or a successor that returns structured content), and forward them as base64 image content blocks on the SDK wire — the bridge already drives `query()` with streaming input.
- Non-image files could stay unsupported initially (the current clear error is fine for those), or be materialized into the session workspace.

## Workaround

We strip file parts before the prompt, download the attachments into the session workspace, and list their absolute paths in the prompt text; the model then views images/PDFs via the CLI's multimodal `Read` tool. Works, but costs a tool roundtrip per image versus direct vision and pushes transcription burden onto every harness consumer.

## Versions

- `@ai-sdk/harness@1.0.24`, `@ai-sdk/harness-claude-code@1.0.24` (latest; first verified on 1.0.18)
- `ai@7.0.20`

Sravan1011:
## Background

Fixes #17082.

The claude-code harness adapter accepted only text parts in a user prompt. Any `file` part — the shape `useChat`'s standard attachment flow produces on user messages — threw `HarnessCapabilityUnsupportedError` before the turn started, so there was no vision path through `HarnessAgent` + `createClaudeCode()` at all, even though the underlying Claude Agent SDK accepts base64 image content blocks in streaming input mode (the mode the bridge already uses).

## Summary

- `extractUserText` is replaced by `extractUserContent`, which maps `file` parts with image media types (and deprecated `image` parts) to base64 image content blocks. Base64 strings, `Uint8Array`/`ArrayBuffer`, and tagged `{ type: 'data' }` file data are all resolved to base64.
- The bridge protocol's `start.prompt` now accepts a structured content array (`text` + `image` blocks) in addition to a plain string.
- The bridge forwards structured prompts to the SDK as `{ type: 'image', source: { type: 'base64', media_type, data } }` content blocks via streaming input.
- Session instructions are prepended as a framed text block when the first prompt has structured content.

### Backward compatibility

- Text-only prompts still collapse to a plain string, so the wire format is unchanged for existing callers.
- Non-image `file` parts, URL-based image data, and provider references keep throwing a clear `HarnessCapabilityUnsupportedError` (matching the initially suggested scope in the issue).
- Queued follow-up user messages remain text-only.

## Verification

- 11 new unit tests assert the exact wire shape for image file parts, `Uint8Array` conversion, tagged file data, deprecated image parts, instruction framing, and the error cases; plus protocol schema tests for the structured `start.prompt`.
- All 71 package tests pass, `tsc --build` is clean, and root `pnpm check` (oxlint + oxfmt) reports 0 warnings/errors.


Sravan1011:
@felixarntz can you review this

Sravan1011:
can anyone review this

PR #17525, opened 2026-07-20
fix: Amazon Bedrock Converse requests fail for ARN model IDs containing slashes

### LINKED ISSUE #17523: @ai-sdk/amazon-bedrock: ARN model IDs containing `/` (application inference profiles, foundation-model ARNs) break the Converse API since 4.0.137 (regression from #17410)
## Summary

`@ai-sdk/amazon-bedrock` currently breaks **every** Bedrock model ID that is an ARN containing a `/` — application inference profile ARNs (`arn:aws:bedrock:{region}:{account}:application-inference-profile/{id}`) and foundation-model ARNs (`arn:aws:bedrock:{region}::foundation-model/{model-id}`) — on the Converse API (`doGenerate`/`doStream`, i.e. `generateText`/`streamText`). This is a regression introduced by the fix for #14117.

## Regression details

- Introduced by #17410 ("fix(bedrock): support application inference profile ARNs"), merged as `76db5e3e`, and backported to `release-v6.0` as `e5e532e` via #17420.
- Released 2026-07-17 ~11:53 UTC in `@ai-sdk/amazon-bedrock@4.0.137`, `5.0.24`, and `3.0.107`. No fixed release exists as of 2026-07-20.
- Only the chat language model was touched by #17410, so this regression is scoped to the Converse path; embedding/image/Anthropic-on-Bedrock still use plain `encodeURIComponent`.

## Root cause

`getUrl` now special-cases `arn:`-prefixed model IDs and un-encodes `%3A` and `%2F`:

```typescript
private getUrl(modelId: string) {
  const encodedModelId = modelId.startsWith('arn:')
    ? encodeURIComponent(modelId).replace(/%3A/g, ':').replace(/%2F/g, '/')
    : encodeURIComponent(modelId);
  return `${this.config.baseUrl()}/model/${encodedModelId}`;
}
```

Application inference profile and foundation-model ARNs contain a literal `/`. Leaving it unencoded splits the URL path, so `/model/{modelId}/converse` no longer matches Bedrock's route. Bedrock doesn't 404 — it returns HTTP 200 with a Coral "unknown operation" body, which then fails the SDK's response-schema validation.

## Observed failure

- Bedrock returns **HTTP 200** with body: `{"Output":{"__type":"com.amazon.coral.service#UnknownOperationException"},"Version":"1.0"}`.
- The SDK surfaces this as `AI_APICallError: Invalid JSON response` (cause `AI_TypeValidationError`/ZodError on `output`/`stopReason`/`usage`), `isRetryable: false`.

## Impact

- 100% failure rate for any Converse call using an application inference profile or foundation-model ARN as model ID.
- Took down our production pipeline for 3 days before we isolated it to this release.
- Anyone using ARN-style model IDs on Converse — which #17410 was meant to enable — is hard-blocked.

## Context on the original fix

- Note: the two pre-merge reproduction attempts for #14117 (#17405, #17408) reported they did not reproduce the original 400 error using the fully percent-encoded URL. Our reproduction below is consistent — the fully-encoded URL reaches Bedrock's auth layer (`403 Authorization header is missing`), i.e. routing works when `/` stays encoded.
- #17410 shipped with a mocked-fetch unit test only; no live Converse call, so the "unknown operation" response wasn't caught.
- Related: #16364 (still open) proposes a shared `formatBedrockModelId` helper — worth checking it doesn't carry the same unencoded-`/` behavior forward.

## Suggested fix

- Revert `getUrl` to plain `encodeURIComponent(modelId)` for the Converse path — the fully-encoded URL is what actually reaches Bedrock successfully (see reproduction).
- If inference-profile support needs different encoding in some scenario, re-derive it from a live Converse call against a real profile before merging.

## Without AWS credentials (routing happens before auth — region eu-west-1)

```bash
# Encoded path (pre-4.0.137 behavior) → route matched, reached auth
curl -s -X POST 'https://bedrock-runtime.eu-west-1.amazonaws.com/model/arn%3Aaws%3Abedrock%3Aeu-west-1%3A123456789012%3Aapplication-inference-profile%2Fabcdefghijkl/converse' \
  -H 'content-type: application/json' -d '{}'
# -> 403 {"Message":"Authorization header is missing"}

# Unencoded path (4.0.137 behavior)
curl -s -X POST 'https://bedrock-runtime.eu-west-1.amazonaws.com/model/arn:aws:bedrock:eu-west-1:123456789012:application-inference-profile/abcdefghijkl/converse' \
  -H 'content-type: application/json' -d '{}'
# -> 200 {"Output":{"__type":"com.amazon.coral.service#UnknownOperationException"},"Version":"1.0"}
```

## With the SDK

```typescript

const bedrock = createAmazonBedrock({ region: 'eu-west-1' });

  model: bedrock(
    'arn:aws:bedrock:eu-west-1:123456789012:application-inference-profile/abcdefghijkl',
  ),
  prompt: 'Say hello',
});
```

Works on `@ai-sdk/amazon-bedrock@4.0.136` and `4.0.93`; fails on `4.0.137` with `AI_APICallError: Invalid JSON response`.

### AI SDK Version

- ai: 6.x
- @ai-sdk/amazon-bedrock: 4.0.137 (also present in 5.0.24 and 3.0.107)
- Runtime: Node.js 22.x (AWS Lambda, arm64); also reproduced on Node 24.12.0 (darwin)

### Code of Conduct

lgrammel:
## Background

Amazon Bedrock Converse generation failed and streaming returned empty text when ARN model IDs contained `/`.

## Root Cause

The chat model URL builder decoded `%2F` in ARN model IDs, splitting the model ID across URL path segments and causing Bedrock to return an UnknownOperationException; route-sensitive regression tests confirmed the malformed path.

## Summary

Restored full percent-encoding of model IDs when constructing Amazon Bedrock Converse and ConverseStream URLs.

## Testing

Updated the existing chat language model tests with route-sensitive generate and stream coverage using inline snapshots, and added a patch changeset.

## End-to-end Validation

- `pnpm -C packages/amazon-bedrock build && pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17523-bedrock-arn-model-id.ts` — live Bedrock generation and streaming succeeded with the inference-profile ARN, returning assistant text and finish reason `length`.

## Related Issues

Fixes #17523

Closes #17524

Co-authored-by: PaulBanusIAD <120463976+PaulBanusIAD@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

The URL builder now fully percent-encodes model IDs, keeping ARN slashes within the modelId path segment for both Converse generation and streaming. The generic implementation covers application inference profiles, foundation-model ARNs, and other slash-containing ARN forms.

### Side effects

Risk: low

The change restores the encoding behavior used before the regression and leaves ordinary model IDs functionally unchanged. It affects only construction of Bedrock chat-model request URLs.

### Performance

Risk: none

The change removes conditional checks and two replacement passes, adding no persistent allocations, buffering, or additional network work.

### Backwards compatibility

Risk: none

This request-time URL construction change does not read, migrate, or modify stored data and restores previously working ARN behavior.

### Security

Risk: none

No authentication or credential handling changes were introduced; encoding reserved path characters reduces path-segment ambiguity rather than expanding it.

### Testing

Status: appropriate

Route-sensitive regression tests cover successful Converse generation and ConverseStream output with a slash-containing ARN, while the authoritative replay exercises the original live workflow. The complete package test suites passed in both Node and edge environments.

### Verification

Reviewed the three-file diff and relevant history, confirmed application-profile and foundation-model ARN slashes are encoded, and found the diff check clean. The Amazon Bedrock package build and type-check passed, as did all 414 Node tests and all 414 edge tests.

### Relevant Documentation

- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-use.html

aayush-kapoor (approved):

PR #17542, opened 2026-07-20
feat(provider,ai): translation model spec + experimental_streamTranslate

kkdawkins:
## Background

Speech-to-speech models (realtime translation like `gpt-realtime-translate` and `gemini-3.5-live-translate-preview`, voice conversion, live dubbing) share one shape: a non-agentic, meaning-preserving **audio-in → audio+text-out pipe**. The SDK has no primitive for this. The realtime session spec is the wrong home — it models a stateful conversation (tools, turn-taking, client events), while these models are single-shot transforms. The right precedent is streaming transcription, which also rides bidirectional WebSockets upstream but exposes a streams-in/streams-out contract with the socket encapsulated in the provider: transport is not modality.

Following the SDK's I/O-shape taxonomy (there is one `LanguageModelV4`, not per-task text models), the spec is deliberately **general** — `SpeechToSpeechModelV4` — while the task-level DX ships as a dedicated function, the same way `generateObject` rides `LanguageModelV4`.

## Summary

- **Spec** (`@ai-sdk/provider`): `Experimental_SpeechToSpeechModelV4` in `packages/provider/src/speech-to-speech-model/v4/`. Streaming-only (`doStream` required, no `doGenerate` — these transforms are inherently live). Input: audio `ReadableStream` + `inputAudioFormat`, optional `targetLanguage`/`sourceLanguage`/`outputAudioFormat`. Output: normalized part stream — `audio` (output audio chunks), `output-text-delta/-final`, `source-transcript-delta/-partial/-final`, `finish { sourceText, outputText, usage }`. Unlike the transcription spec, `finish` carries **typed usage** (audio/text token counts + seconds) so gateways can bill from the stream.
- **Core** (`ai`): `experimental_streamTranslate` — the translation-flavored entry point. Requires `targetLanguage`, surfaces `translationText` (from `finish.outputText`). Mechanics mirror `experimental_streamTranscribe`: synchronous return, single-consumer `fullStream` (live audio can't be teed/replayed), promise results that drain the stream without deadlock, pipe-abort propagated into `doStream`, upstream audio cancelled on pre-stream failure. Future task functions (e.g. voice conversion) reuse the spec without new wire/gateway work.
- String model IDs resolve via an off-registry `speechToSpeechModel(id)` getter on the global provider (same pattern as `videoModel`), plus `NoTranslationGeneratedError` and a `MockSpeechToSpeechModelV4` test double.
- Docs: core `translation` page + `streamTranslate` reference.

Everything is `Experimental_`/`experimental_`-prefixed; the contract may change while experimental.

## End-to-End Verification

Exercised via the stacked provider PR (#17543): OpenAI Realtime `/v1/realtime/translations` WebSocket handshake and subprotocol auth verified live with `examples/ai-functions/src/stream-translate/openai/basic.ts`. Unit tests cover the full part vocabulary, finish resolution, abort, and single-consumer semantics against the mock model.

## Checklist

## Future Work

Stacked: #17543 (OpenAI + Google providers). Then: `speech-to-speech-stream` wire envelope in `@ai-sdk/provider-utils` + `GatewaySpeechToSpeechModel` (mirroring the transcription-stream envelope), AI Gateway server surface `/v4/ai/speech-to-speech-model` with billing/catalog, and ephemeral client-secret auth for browser mic apps. Bundle note: `ai` node bundle is +3.5 KB; main already sits at the 640 KB limit, so `BUNDLE_SIZE_LIMIT_KB` needs a bump.


kkdawkins:
* **#17543** <a href="https://app.graphite.com/github/pr/vercel/ai/17543?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#17542** <a href="https://app.graphite.com/github/pr/vercel/ai/17542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/ai/17542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

content/docs/03-ai-sdk-core/36-translation.mdx
```diff
@@ -0,0 +1,200 @@
...
+function to translate live speech into another language. Translation is a
+streaming-only modality: models translate live source audio into
+target-language audio and text.
+
+`experimental_streamTranslate` is built on the general speech-to-speech model
+specification (`Experimental_SpeechToSpeechModelV4`), which covers
+meaning-preserving audio-to-audio transforms such as translation and voice
+conversion — analogous to how `LanguageModelV4` underlies both `generateText`
+and `generateObject`. The translation function constrains the specification
+to the translation task, e.g. by requiring `targetLanguage`.
```
gr2m: `generateObject` is deprecated, I wouldn't refer to it. We replaced it with `generateText({ output })`

content/docs/03-ai-sdk-core/36-translation.mdx
```diff
@@ -0,0 +1,200 @@
...
+  targetLanguage: 'es',
+});
+```
+
+### Provider-Specific settings
+
+Translation models often have provider or model-specific settings which you can
+set using the `providerOptions` parameter.
+
+```ts highlight="7-11"
```
gr2m: Agents seem to get the highlight numbers wrong a lot

````suggestion
```ts highlight="8-12"
````

File: content/docs/03-ai-sdk-core/36-translation.mdx [hunk 3]
```diff
@@ -0,0 +1,200 @@
...
+    },
+  },
+});
+```
+
+### Abort Signals
+
+Pass an `abortSignal` to cancel the translation:
+
+```ts highlight="7"
```
gr2m:
```suggestion
```ts highlight="8"
```

content/docs/07-reference/05-ai-sdk-errors/ai-no-translation-generated-error.mdx
```diff
@@ -0,0 +1,26 @@
...
+
+# AI_NoTranslationGeneratedError
+
+This error occurs when no translation could be generated from the input audio:
+no `audio` part was emitted and the final output text is empty, or the stream
+ended without a finish event.
+
+## Properties
+
+- `responses`: Array of speech-to-speech model response metadata (required in constructor)
```
gr2m: What's the usecase for multiple responses? I think we have `responses` in `NoImageGeneratedError` because a single `generateImage` call can generate multiple images. Not sure if that use case could be relevant for generating translations? Maybe some providers could allow generating multiple translations at once?

packages/ai/src/translate/stream-translate-result.ts
```diff
@@ -0,0 +1,144 @@
...
+      /**
+       * Final source transcript text for a provider-defined segment or utterance.
+       */
+      type: 'source-transcript-final';
+      id?: string;
+      text: string;
+      startSecond?: number;
+      endSecond?: number;
+      providerMetadata?: SharedV4ProviderMetadata;
+    }
```
gr2m, line 80: `durationSeconds` vs `endSecond`? I think it's copied from Transcription but we might want to normalize here. Also transcription has `channelIndex`, does it not exist here?

packages/ai/src/translate/stream-translate.ts
```diff
@@ -0,0 +1,394 @@
...
+    /**
+     * Audio format type, e.g. `audio/pcm`, `audio/pcmu`, or `audio/pcma`.
+     */
+    type: string;
+
+    /**
+     * Sample rate in Hz. Only applicable for formats that require a rate.
+     */
+    rate?: number;
+  };
```
gr2m: can be done in a follow up PR / future work, but since we have the same audio type defined inline in 6 places now, we can create `SharedV4AudioFormat`

packages/ai/src/translate/index.ts
```diff
@@ -0,0 +1,5 @@
+export { streamTranslate as experimental_streamTranslate } from './stream-translate';
+export type {
+  StreamTranslationResult,
+  TranslationStreamPart,
```
gr2m: export as `Experimental_`

packages/ai/test/index.ts
```diff
@@ -16,6 +16,7 @@ export { MockProviderV3 } from '../src/test/mock-provider-v3';
 export { MockProviderV4 } from '../src/test/mock-provider-v4';
 export { MockSpeechModelV3 } from '../src/test/mock-speech-model-v3';
 export { MockSpeechModelV4 } from '../src/test/mock-speech-model-v4';
+export { MockSpeechToSpeechModelV4 } from '../src/test/mock-speech-to-speech-model-v4';
```
gr2m: export as Experimental

content/docs/07-reference/05-ai-sdk-errors/ai-no-translation-generated-error.mdx
gr2m: I think responses might also have been introduced to address retries, which can't exist for stream audio

gr2m (commented):
> main already sits at the 640 KB limit, so `BUNDLE_SIZE_LIMIT_KB` needs a bump

already bumped to 650 in main

content/docs/03-ai-sdk-core/36-translation.mdx
gr2m: Removed the generateObject comparison when the contract was renamed to SpeechTranslationModelV4.

content/docs/03-ai-sdk-core/36-translation.mdx
gr2m: Updated the highlight range to 8-12.

content/docs/03-ai-sdk-core/36-translation.mdx
gr2m: Updated the highlight to line 8.

content/docs/07-reference/05-ai-sdk-errors/ai-no-translation-generated-error.mdx
gr2m: Changed the result and NoTranslationGeneratedError metadata to a singular response.

packages/ai/src/translate/stream-translate-result.ts
gr2m: Normalized partial and final transcript timing to endSecond and added channelIndex to both.

packages/ai/src/translate/stream-translate.ts
gr2m: Added SharedV4AudioFormat and reused it across realtime, transcription, translation, and core.

packages/ai/src/translate/index.ts
gr2m: Exported the result types as Experimental_StreamTranslationResult and Experimental_TranslationStreamPart.

packages/ai/test/index.ts
gr2m: Exported the test mock as Experimental_MockSpeechTranslationModelV4.

gr2m (approved):

PR #17551, opened 2026-07-21
fix(ai): reject messages after tool approval responses

### LINKED ISSUE #17033: streamText silently skips local tool approvals when a later context message follows the approval response
### Description

`streamText` silently skips a local tool approval when any user or system message is appended after the `tool-approval-response` and before the resumed model call.

The approval paths currently disagree about which part of the prompt is authoritative:

- [`convertToLanguageModelPrompt`](https://github.com/vercel/ai/blob/0be50146de12893d9e22e1f19d4f1f0004d74fee/packages/ai/src/prompt/convert-to-language-model-prompt.ts#L54-L84) discovers approval requests and responses across the full message history. It then treats approved calls as resolved for missing-result validation.
- [`collectToolApprovals`](https://github.com/vercel/ai/blob/0be50146de12893d9e22e1f19d4f1f0004d74fee/packages/ai/src/generate-text/collect-tool-approvals.ts#L18-L37) returns immediately unless the final prompt message has role `tool`.

That creates this failure mode:

1. An assistant message contains a local `tool-call` and its `tool-approval-request`.
2. A tool message contains `{ type: 'tool-approval-response', approved: true }`.
3. The application appends supplemental user context before resuming.
4. `streamText` does not call the tool's `execute` function because the final message is now a user message.
5. Prompt validation has already accepted the approval, so the provider can receive a dangling tool call with no result. Anthropic rejects that transcript; other providers can continue without ever executing the approved side effect.

The same mismatch applies to denied approvals: a trailing context message prevents the denial result from being generated.

This is distinct from #10980 and #13057: it occurs in the core `streamText` approval collector with normal model messages, without `createAgentUIStream`.

### Expected behavior

A non-assistant context message after an approval response should not make the response invisible. The approved tool should execute exactly once, or a denied approval should produce its normal denial result.

### Proposed fix

Make approval collection use the same turn boundary as prompt conversion instead of assuming that approvals must be in the final message:

1. Find approval responses in the unresolved suffix for the latest assistant message that issued approval requests.
2. Allow intervening or trailing user/system context within that suffix.
3. Ignore calls that already have a tool result.
4. Do not search behind a later assistant response, which prevents old approvals from replaying side effects.
5. Keep the existing schema/policy revalidation and canonical `executeToolCall` path.

Ideally, prompt validation and execution should share the same helper or invariant: a call is exempt from missing-result validation only if this `streamText` invocation will consume its approval response or it already has a result.

Regression coverage should include:

- approved response followed by a user message executes exactly once;
- denied response followed by a user message produces the denial result;
- an approval with an existing tool result does not execute again;
- an older approval before a later assistant response does not execute;
- missing-result validation is not suppressed for an approval the current invocation will not consume.

### AI SDK Version

Observed with `ai@7.0.0`. The conflicting collector behavior is still present on `main` at `0be50146de12893d9e22e1f19d4f1f0004d74fee` (`packages/ai/package.json` reports `7.0.19`).

### Code of Conduct

jstar0:
# fix(ai): reject messages after tool approval responses

## Background

`streamText` currently treats any approval response in message history as
resolving its tool call during prompt validation, while local approval execution
only consumes responses from the final tool message. A later user, system, or
assistant message can therefore leave a dangling tool call that reaches the
provider without executing the approved tool or producing a denial result.

The earlier approach in #17036 searches backward and executes approvals after
trailing context. This change instead follows the later maintainer guidance in
#17033 that such messages break the required
`assistant -> tool -> assistant -> user` sequence and should be rejected.

## Summary

Align prompt validation with the approval boundary consumed by the current
`streamText` or `generateText` invocation. Invalid histories are rejected before
an approved side effect or provider request can be silently skipped.

## Changes

- Limit approval-based missing-result exemption to responses in the final tool
  message, matching the boundary consumed by the current invocation.
- Reject unresolved approved or denied histories before tool execution or model
  invocation by using the existing `MissingToolResultsError`.
- Preserve completed histories that already contain a tool result.
- Add direct `streamText` and prompt-validation regression coverage plus a patch
  changeset for `ai`.

## Contributor Credit

- @ruiconti reported the approval/prompt mismatch in #17033.
- @lgrammel reproduced it in #17051 and clarified the required message sequence.

## End-to-End Verification

An approved response followed by a user message now reports
`MissingToolResultsError` before either the tool or model is called. The same
boundary rejects denied and other unresolved trailing histories. A final
approval response still follows the normal execution path, and a completed tool
result can still be followed by assistant and user messages.

## Verification

```bash
cd packages/ai && pnpm test
cd packages/ai && pnpm type-check
cd packages/ai && pnpm build
pnpm check
```

## Checklist

## Related Issues

Fixes #17033


jstar0:
Merged current `main` on exact head `611fe8f10a`.

The only conflict was the import block in `packages/ai/src/generate-text/stream-text.test.ts`. The branch now keeps both `StreamProviderError` from `main` and `MissingToolResultsError` for the approval-sequence regression.

Verification: `pnpm --filter ai exec vitest --config vitest.node.config.js --run src/prompt/convert-to-language-model-prompt.validation.test.ts src/generate-text/stream-text.test.ts -t 'approval|MissingToolResults'` — 57 passed.

jstar0:
Synced with current `main` at `a4b0940` in signed commit `32eb57c`; the PR diff remains four files. Local checks pass: `pnpm --filter ai test` (Node 4,326 passed/3 skipped; Edge 4,319 passed/4 skipped), `pnpm check`, and `pnpm type-check:full`. GitHub CI reports 50 checks passed and 1 skipped; the only failure is the `Vercel` context: “Authorization required to deploy.” Could a maintainer review when convenient and advise whether preview authorization can be enabled for this PR?

PR #17562, opened 2026-07-21
fix: provider options lost when consecutive tool messages are combined

### LINKED ISSUE #17507: providerOptions are lost when consecutive tool messages are combined
## Summary

When consecutive tool messages are combined, only the first message's `providerOptions` survive. Options from later tool messages are silently dropped.

For Anthropic prompt caching, message-level cache control applies to the message's final tool-result block. Combining messages therefore moves the first cache breakpoint to the wrong result and removes subsequent breakpoints.

## Root Cause

The prompt converter appends content from consecutive tool messages without preserving each message's options:

https://github.com/vercel/ai/blob/b8f75d5280c1378ae986caa459635c84732427c8/packages/ai/src/prompt/convert-to-language-model-prompt.ts#L101-L115

```ts
if (lastCombinedMessage?.role === 'tool') {
  lastCombinedMessage.content.push(...message.content);
}
```

## Reproduction

Draft reproduction PR: https://github.com/vercel/ai/pull/17508

It adds one intentionally failing regression test. Run it with:

```bash
cd packages/ai
pnpm exec vitest --config vitest.node.config.js --run src/prompt/convert-to-language-model-prompt.test.ts
```

The failure shows that the first message's options remain at the combined message level while the second message's options are absent.

Reproduced with `ai@6.0.40` and current `main` (`ai@7.0.31`).

## Expected Behavior

Each original tool message's provider options should remain associated with that message's final content part after coalescing. Existing part-level options should retain precedence on conflicts.

## Related Work

The coalescing behavior was introduced in #8541. Related provider-option loss reports include #5473, #5942, and #15185, but none cover this path specifically.

## Code of Conduct

lgrammel:
## Background

Combining consecutive tool messages silently discarded later provider options, causing provider-specific behavior such as Anthropic cache breakpoints to be misplaced or removed.

## Root Cause

The prompt converter appended consecutive tool-message content while retaining only the first combined message's providerOptions. The provider-facing reproduction and regression snapshot confirmed that option boundaries were lost during coalescing.

## Summary

When combining tool messages, prior message options are merged onto that message's final content part with part-level precedence, while the incoming message options remain on the combined message.

## Testing

Added an inline-snapshot regression test covering consecutive tool messages, preservation of both option boundaries, and part-level option precedence.

## End-to-end Validation

- `git show HEAD:examples/ai-functions/src/reproduction/issue-17507-tool-message-provider-options.ts | pnpm -C examples/ai-functions exec tsx -` completed successfully and observed all provider options at the expected tool-result boundaries.

## Related Issues

Fixes #17507

Closes #17516

Co-authored-by: ayoubkachkach <16275973+ayoubkachkach@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

At each coalescing boundary, the fix moves the prior tool message's options onto its final content part, preserves part-level precedence through deep merging, and retains the incoming message's options at the combined-message level. This matches documented message-to-final-block behavior. citeturn1view2turn1view4

### Side effects

Risk: low

The change is limited to consecutive tool-message coalescing and preserves existing content ordering; its only intended behavior change is that previously discarded provider options now reach their original boundaries.

### Performance

Risk: low

The added work is one bounded provider-options deep merge per coalesced message boundary, with no additional prompt-sized collections or retained state.

### Backwards compatibility

Risk: low

No stored-data format or public type changes. Existing stored messages may produce corrected provider requests when reused because options that were previously dropped are now honored.

### Security

Risk: low

No new external input, network, credential, or execution path is introduced; the existing merge utility also rejects prototype-polluting keys.

### Testing

Status: appropriate

The regression test covers two consecutive tool messages, preservation of both option boundaries, non-conflicting option inheritance, and part-level conflict precedence; the exact public-path reproduction also passed.

### Verification

Inspected the merge-base diff and targeted implementation, confirmed a clean merge with current main, ran the focused prompt-conversion suite with all 68 tests passing, completed the ai package type-check successfully, and verified the changed files pass diff whitespace checks.

### Relevant Documentation

- https://ai-sdk.dev/providers/ai-sdk-providers/anthropic
- https://ai-sdk.dev/cookbook/node/dynamic-prompt-caching
- https://platform.claude.com/docs/en/build-with-claude/prompt-caching

PR #17563, opened 2026-07-21
fix: UI message stream failures are indistinguishable from successful completion in onEnd

### LINKED ISSUE #17500: UI message stream onEnd cannot distinguish a failed response from a completed one

lgrammel:
## Background

UI stream consumers could persist failed responses as completed because onEnd exposed no operation-level outcome.

## Root Cause

UI stream creation normalized execution and merged-stream failures into error chunks without retaining terminal operation state, while onEnd only received reducer-derived abort and model finish metadata.

## Summary

Added an exported UIMessageStreamOutcome, producer-declared composer outcomes, automatic rejection tracking, source-stream outcome tracking, and outcome delivery through onEnd.

## Testing

Expanded existing UI message stream tests to cover completed, failed, aborted, unknown, execution-rejection, merged-rejection, recoverable-error, and explicitly declared outcomes.

## End-to-end Validation

- `pnpm exec tsx -e "<public UI stream outcome scenarios>"` — observed completed declarations, failed execution and merged streams, unknown undeclared error chunks, failed terminal text errors, and completed recovered text streams.

## Related Issues

Fixes #17500

Closes #17518

Co-authored-by: bryan-hunter <26581405+bryan-hunter@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

The callback now receives a distinct operation outcome; streamText-derived streams track completed, failed, aborted, and unknown states, while composers can explicitly declare aggregate outcomes.

### Side effects

Risk: low

The change is additive to callback payloads and writer capabilities, preserves finishReason semantics, and deliberately avoids treating arbitrary UI error chunks as fatal.

### Performance

Risk: low

Outcome tracking uses constant-size state and retains at most one terminal error until stream completion, adding negligible per-chunk work.

### Backwards compatibility

Risk: low

No stored-data format or wire protocol changes are introduced; existing persistence changes only if consumers opt into using the new outcome field.

### Security

Risk: low

Raw failure details remain confined to the server-side callback, while client error chunks continue to use the configured masking handler.

### Testing

Status: appropriate

Regression coverage includes completed, failed, aborted, unknown, recoverable-error, execution-rejection, merged-rejection, explicit declarations, and error-versus-finish ordering.

### Verification

Reviewed the merge-base diff and resulting clean merge, inspected outcome propagation and streamText terminal paths, and verified 47 focused tests in both Node and Edge environments. Repository type checking, formatting, linting, and diff validation also passed.

PR #17564, opened 2026-07-21
fix: preserve tool parts when tool call IDs repeat across model steps

### LINKED ISSUE #17347: Duplicate toolCallId across steps causes readUIMessageStream to overwrite earlier tool parts

lgrammel:
## Background

Repeated response-scoped tool call IDs across model steps caused readUIMessageStream to discard earlier tool parts and retain only the latest step’s data.

## Root Cause

processUIMessageStream searched the entire aggregated message for the first matching toolCallId, so later-step chunks updated an earlier step’s tool part; the reproduction confirmed one part remained instead of two.

## Summary

Scoped tool-part creation and lookup to the current step while retaining fallback support for output-only continuations of existing tool parts.

## Testing

Added inline-snapshot regression coverage proving static and dynamic tool parts with duplicate IDs remain distinct across steps.

## End-to-end Validation

- `pnpm -C packages/ai build && git show HEAD:examples/ai-functions/src/reproduction/issue-17347-duplicate-tool-call-id.ts | pnpm -C examples/ai-functions exec tsx -` produced two tool parts with both steps’ inputs, outputs, and provider item IDs preserved.

## Related Issues

Fixes #17347

Closes #17519

Co-authored-by: wangliang1996 <36724800+wangliang1996@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

Tool creation and lookup are scoped to parts after the latest step boundary, while output-only continuations fall back to the most recent matching prior part. Original toolCallId values and per-step provider metadata remain unchanged.

### Side effects

Risk: low

Lookup behavior changes only when tool call IDs repeat across steps. Same-step updates retain their existing behavior, and continuation outputs still locate prior tool parts through the fallback lookup.

### Performance

Risk: low

Current-step lookup adds bounded linear scans and temporary slices of the existing message parts; it introduces no retained auxiliary collection. Preserving additional parts for repeated IDs is the expected data growth.

### Backwards compatibility

Risk: low

There is no stored-data schema change or migration. Existing messages remain readable, and output-only continuations without a current-step match target the most recent matching stored tool part.

### Security

Risk: none

The change only adjusts in-memory part selection and mutation and introduces no new parsing, network access, credential handling, or externally controlled execution.

### Testing

Status: appropriate

Regression tests cover static and dynamic tool parts with repeated IDs across steps, including preservation of distinct inputs, outputs, and provider item metadata. Existing continuation tests also exercise the fallback behavior.

### Verification

Inspected the implementation and regression diff against the merge base. The focused process-stream suite passed all 105 tests in both Node and Edge configurations; package type checking, focused linting/formatting, and diff validation also passed.

PR #17573, opened 2026-07-21
fix: prevent duplicate Mistral tool-call IDs in Amazon Bedrock conversations

### LINKED ISSUE #16182: @ai-sdk/amazon-bedrock: normalizeToolCallId truncation causes duplicate tool-call IDs (Bedrock ValidationException) for Mistral models
### Description

Follow-up / regression from #11802 (fixed by #11863).

#11863 added `normalizeToolCallId()` so Bedrock tool-call IDs satisfy Mistral's `^[a-zA-Z0-9]{9}$` requirement, by stripping non-alphanumeric characters and taking the **first 9**:

```ts
const alphanumericChars = toolCallId.replace(/[^a-zA-Z0-9]/g, '');
return alphanumericChars.slice(0, 9);
```

Bedrock generates tool-call IDs of the form `tooluse_<random>` (e.g. `tooluse_bpe71yCfRu2b5i-nKGDr5g`). After stripping `_`/`-`, the first 9 characters are always `tooluse` (7 fixed chars) **+ 2 variable chars**. So normalization preserves only ~2 characters of entropy (62² = 3,844 possible values) — the high-entropy random suffix is discarded.

With several tool calls in a conversation (especially parallel tool calls in one turn), two distinct Bedrock IDs collide to the same normalized ID (birthday paradox). The request then contains two `toolUse` blocks with the same `toolUseId`, and Bedrock rejects it:

```
ValidationException: The toolUse blocks at messages.2.content contain duplicate Ids: tooluseAc
```

Observed in production with Ministral on Bedrock. A single assistant turn with 8 parallel tool calls produced these normalized IDs:

```
tooluseqr, tooluseY3, tooluseAc, tooluseP3, tooluseIx, tooluseV0, tooluse7r, tooluseAc
```

Note `tooluseAc` appears twice. Every ID is `tooluse` + 2 chars, confirming only the first 2 random characters survive.

### Why it surfaces on a later turn

`normalizeToolCallId` is applied on both the response path (`doGenerate` / `doStream`, mapping `toolUse` → tool-call output) and the request path (`convert-to-amazon-bedrock-chat-messages`, both `toolUse` and `toolResult`). The truncated IDs are returned to the caller and persisted; the next turn rebuilds the request from history and the collision surfaces as the duplicate-ID 400. Because it's a birthday collision in a ~3,844-value space, it only appears once enough tool calls accumulate — which is why single-tool tests don't catch it.

### Reproduction (unit level)

```ts
import { normalizeToolCallId } from '@ai-sdk/amazon-bedrock';

// Two distinct Bedrock IDs that share the first 9 alphanumeric characters:
const a = normalizeToolCallId('tooluse_Ac1Xq9ZklmNoPq', true); // 'tooluseAc'
const b = normalizeToolCallId('tooluse_Ac2Yt7WrstUvWx', true); // 'tooluseAc'
a === b; // true  ->  duplicate toolUseId  ->  Bedrock ValidationException
```

### Root cause

`slice(0, 9)` of the alphanumeric string keeps the constant `tooluse` prefix and discards the entropy-bearing suffix. Any normalization that maps long Bedrock IDs into Mistral's 9-char space must use the full 9-char space (62⁹ ≈ 1.35 × 10¹⁶), not the first 9 characters.

### Proposed fix

Map the **full** original ID deterministically into 9 alphanumeric characters via a hash, instead of truncating:

- **Collision-resistant** — ~53 bits of entropy vs. the current ~12 bits.
- **Deterministic** — required, because the normalized ID is returned to the caller, persisted, and re-normalized on the request path; a tool call and its matching tool result must always map to the same value. (A *random* regeneration — as some external workarounds do, e.g. [this one](https://huggingface.co/spaces/arterm-sedov/cmw-copilot/blob/main/docs/MISTRAL_TOOL_CALL_ID_FIX.md) — fixes the collision but breaks call/result pairing unless paired with a per-request ID registry, which this stateless per-ID normalizer does not have.)
- **Idempotent** on already-normalized (9-char) IDs, so round-tripping is stable.

PR: #16183 — https://github.com/vercel/ai/pull/16183

### Affected versions

`@ai-sdk/amazon-bedrock` — current `latest` (4.0.116), `ai-v6` tag, and `main` all contain the unchanged `slice(0, 9)`. Models: `mistral.*` / `us.mistral.*` on Bedrock (observed with Ministral).

lgrammel:
## Background

Mistral conversations on Amazon Bedrock could fail with HTTP 400 when distinct tool calls normalized to duplicate nine-character IDs.

## Root Cause

The normalizer retained the fixed `tooluse` prefix and only two variable characters by truncating the first nine alphanumeric characters; the public API reproduction confirmed colliding IDs produced Bedrock's duplicate-ID ValidationException.

## Summary

Replaced truncation with deterministic FNV-1a hashing across the full nine-character base62 space while preserving already-valid IDs.

## Testing

Added inline-snapshot coverage for deterministic, valid, distinct normalized IDs and updated Bedrock message-conversion expectations.

## End-to-end Validation

- `pnpm --filter @ai-sdk/amazon-bedrock build && cd examples/ai-functions && pnpm tsx src/reproduction/amazon-bedrock-mistral-tool-call-id-validation.ts` — live Bedrock returned HTTP 200 with distinct IDs `7rDVWRig0` and `bNZvZKNBZ`.

## Related Issues

Fixes #16182

Closes #17571

Co-authored-by: arpanpal-at <253363140+arpanpal-at@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: inconclusive

Concerns:

### Fixes issue

Status: unclear

The code deterministically hashes the complete incompatible ID into nine base62 characters, preserves already-valid IDs, and applies the same normalizer on response, streaming, tool-use, and tool-result paths.

Concerns:

### Side effects

Risk: low

Behavior changes only for non-nine-character Mistral tool-call IDs; non-Mistral IDs and valid nine-character Mistral IDs remain unchanged.

### Performance

Risk: low

Normalization remains linear in ID length, uses constant memory and fixed-width arithmetic, and adds only a nine-iteration base62 encoding step.

### Backwards compatibility

Risk: low

Previously persisted nine-character normalized IDs are preserved, while matching raw tool calls and results are transformed deterministically to the same new ID.

### Security

Risk: none

The hash is used only for deterministic identifier normalization, not authentication or integrity, and the change introduces no new I/O, credential handling, or executable input.

### Testing

Status: appropriate

Tests cover determinism, the reported colliding pair, nine-character alphanumeric output, idempotence, non-Mistral behavior, message conversion, and the reproduced public-path scenario.

### Verification

Reviewed the merge-base diff and every normalization call site. The Amazon Bedrock package build and type-check passed; focused Node and edge suites passed, including the restored regression test. A synthetic 200,000-ID check produced valid distinct outputs without collisions. Repository-wide checks encountered unrelated workspace failures.

n1ckoates (approved):

PR #17577, opened 2026-07-21
fix: preserve provider options across consecutive tool messages

### LINKED ISSUE #17507: providerOptions are lost when consecutive tool messages are combined

lgrammel:
## Background

Consecutive tool messages silently lost later message-level provider options, causing provider behavior such as Anthropic cache breakpoints to be applied to the wrong tool results.

## Root Cause

The prompt converter appended consecutive tool-message content while retaining only the first message's providerOptions. The public generateText reproduction confirmed that the first options remained on the combined message and the second options disappeared.

## Summary

Before combining tool messages, the converter now merges the previous message's provider options into its final content part with part-level precedence, then retains the incoming message's options on the combined message.

## Testing

Added an inline-snapshot regression test covering message-boundary preservation, part-level precedence, and retention of the final tool message's provider options.

## End-to-end Validation

- `pnpm -C packages/ai build && git show a8203a7f:examples/ai-functions/src/reproduction/issue-17507-tool-message-provider-options.ts | pnpm -C examples/ai-functions exec tsx -` — the public generateText reproduction exited successfully and confirmed all provider options remained at their expected tool-result boundaries.

## Related Issues

Fixes #17507

Closes #17516

Co-authored-by: ayoubkachkach <16275973+ayoubkachkach@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

Before coalescing, the implementation transfers the previous tool message's options to its final content part, preserves part-level precedence through a deep merge, and retains the incoming message's options at the combined-message level.

### Side effects

Risk: low

The behavioral change is limited to consecutive tool messages; prompts without such boundaries or without provider options retain their existing conversion behavior.

### Performance

Risk: low

The change adds one bounded provider-options merge per applicable tool-message boundary, with work proportional to the typically small option objects and no accumulating copies.

### Backwards compatibility

Risk: low

No stored data is modified or migrated; previously stored prompts only produce corrected provider-facing options when converted through this path.

### Architecture

Risk: low

The fix remains within the core prompt-coalescing code and reuses the existing internal generic merge utility, adding no dependencies, public APIs, or cross-package coupling.

### Security

Risk: low

The change introduces no new I/O or trust boundary, and the reused merge utility already excludes prototype-polluting keys.

### Testing

Status: appropriate

The regression test covers message-boundary preservation, non-conflicting option retention, part-level conflict precedence, and final-message options; the exact reproduction replay and focused multi-message verification also passed.

### Verification

Reviewed the complete merge-base diff, prompt and provider-option types, the existing merge utility, provider adapter handling, package boundaries, and relevant architecture records. The focused prompt-conversion suite passed all 68 tests in both Node and Edge configurations, package and repository TypeScript checks passed, diff validation passed, and an additional three-consecutive-tool-message check preserved every boundary correctly.

PR #17583, opened 2026-07-21
[v6.0] reproduction: providerOptions are lost when consecutive tool messages are

lgrammel:
## Bug

providerOptions are lost when consecutive tool messages are combined

## Reproduction

- On `release-v6.0` with `ai@6.0.231`, consecutive tool messages were coalesced, but the provider-facing prompt retained only the first message's `providerOptions`; the second message's options were silently lost.
- The first tool-result part retained its conflicting part-level option but failed to inherit the first message's non-conflicting options, so the original message boundary was not preserved.
- `examples/ai-functions/src/reproduction/issue-17507-tool-message-provider-options.ts` reproduces the bug through public `generateText`; correct behavior would retain the second message's options on the combined message and merge the first message's options into its final result with part-level precedence.
- Losing valid message-level options can remove or relocate Anthropic explicit prompt-cache breakpoints.

## Relevant Documentation

- https://ai-sdk.dev/docs/foundations/prompts
- https://ai-sdk.dev/providers/ai-sdk-providers/anthropic
- https://ai-sdk.dev/cookbook/node/dynamic-prompt-caching
- https://platform.claude.com/docs/en/build-with-claude/prompt-caching

## Related Issues

Reproduces #17507

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "issueId": 17507,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17507-tool-message-provider-options.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/issue-17507-tool-message-provider-options.ts"
+  ],
+  "expectedFailure": {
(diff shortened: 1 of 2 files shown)


lgrammel:
Issue #17507 already has the merged-v6.0 label. Closing this stale reproduction pull request.

PR #17586, opened 2026-07-21
fix: preserve tool parts when tool call IDs repeat across steps

### LINKED ISSUE #17347: Duplicate toolCallId across steps causes readUIMessageStream to overwrite earlier tool parts

lgrammel:
## Background

Repeated provider-scoped tool call IDs across model steps caused readUIMessageStream to overwrite an earlier tool part, leaving only the later step's input, output, and metadata.

## Root Cause

The UI stream reducer searched all message parts by toolCallId, so a later step reused and mutated the first matching part. The credential-free reproduction confirmed two step-local calls were aggregated into one part.

## Summary

Scoped tool input aggregation to the current step while retaining explicit fallback handling for outputs that complete previously existing tool calls. Removed reproduction artifacts and added an ai package patch changeset.

## Testing

Added a readUIMessageStream regression test asserting that repeated tool call IDs retain both steps' inputs, outputs, and provider metadata.

## End-to-end Validation

- `pnpm exec vitest --config vitest.node.config.js --run src/ui-message-stream/read-ui-message-stream.test.ts -t "should preserve tool parts when tool call ids repeat across steps"` — passed and preserved both tool parts.

## Related Issues

Fixes #17347

Closes #17519

Co-authored-by: wangliang1996 <36724800+wangliang1996@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

Tool input lookup is scoped to parts after the latest step boundary, while outputs can still resolve the most recent matching invocation from an earlier step; provider toolCallId and metadata remain unchanged.

### Side effects

Risk: low

Behavior is unchanged for unique IDs and streams without step boundaries. Focused checks also confirmed streaming, dynamic-tool, error-output, and prior-step output completion behavior.

### Performance

Risk: low

The implementation performs bounded linear scans and short-lived array slices over existing message parts on tool chunks, comparable to the prior full-message filtering and without persistent additional state.

### Backwards compatibility

Risk: low

No stored-data schema or protocol identifier changes are introduced. Existing messages are only subject to the corrected step-local matching when subsequently processed.

### Architecture

Risk: low

The fix remains localized to the ai package's UI stream reducer, introduces no dependencies or exported provider-specific helpers, and preserves existing package boundaries and abstraction ownership.

### Security

Risk: none

The change only alters in-memory tool-part selection and mutation; it adds no parsing, credential handling, network access, or trust-boundary changes.

### Testing

Status: appropriate

A regression test covers the reported repeated-ID sequence and verifies both inputs, outputs, and provider metadata. The broader reducer suites passed in both Node and Edge configurations.

### Verification

Inspected the complete origin/main...HEAD diff and relevant reducer architecture; 106 focused tests passed under both Node and Edge configurations with no type errors, package TypeScript checking passed, formatting and diff checks passed, and additional bounded checks validated streaming static tools, dynamic tools, error outputs, and previous-step output fallback.

PR #17610, opened 2026-07-21
feat: add a built-in stop condition for repeated identical tool calls

### LINKED ISSUE #17606: Feature: built-in stop condition to detect repeated identical tool calls (`hasRepeatedToolCalls`)
### Summary

There is currently no middle ground between `isStepCount(n)` and `isLoopFinished()` for controlling agent loops. A step cap is blunt: it kills long productive runs early, and a stuck run still burns the whole budget before stopping. `isLoopFinished()` removes the cap entirely — and the [loop-control docs](https://ai-sdk.dev/docs/agents/loop-control) themselves warn to use it "with caution" because the agent could run indefinitely, without offering a mitigation.

The failure mode I actually see in production is specific and detectable: the model gets stuck re-issuing the **exact same tool call** — identical tool name, byte-identical arguments — step after step. Step count is a proxy for this; repetition is the signal itself.

I propose a built-in stop condition that detects it directly.

### Real production use case

We run a platform where product agents execute long, open-ended tool loops (`ToolLoopAgent`-style, on `ai` v7 betas). Runs are deliberately uncapped — no step budget at all, bounded only by a cost limit and a wall-clock timeout — because for long tasks a step cap either stops productive runs too early or is set so high it protects nothing.

With the cap removed, the only loop-level brake we need in practice is a repetition guard: when the last N steps all carry an identical set of `(toolName, input)` pairs, the run is stuck and gets stopped. It composes well with `prepareStep`: one repeat before the threshold we inject a notice telling the model that one more identical call ends the run — so runs wind down with a real final answer (`finishReason: 'stop'`) instead of being cut mid-flight.

### Current workaround (in production)

A custom `StopCondition`, simplified:

```ts
import type { StopCondition, ToolSet } from 'ai';

interface StepLike {
  toolCalls: Array<{ toolName: string; input: unknown }>;
}

// Canonical signature of a step's tool calls: sorted (toolName, input) pairs.
function stepSignature(step: StepLike): string {
  return step.toolCalls
    .map(call => `${call.toolName}:${JSON.stringify(call.input)}`)
    .sort()
    .join('|');
}

export function hasRepeatedToolCalls(count: number): StopCondition<ToolSet> {
  return ({ steps }) => {
    if (steps.length < count) return false;
    const last = stepSignature(steps[steps.length - 1] as StepLike);
    if (!last) return false; // a step without tool calls breaks the streak
    return (steps as StepLike[])
      .slice(-count)
      .every(step => stepSignature(step) === last);
  };
}
```

This works, but every team that removes the step cap has to rediscover the same edge cases (parallel tool calls, streak-breaking, polling false-positives — see below).

### Proposed API

```ts
import { ToolLoopAgent, isLoopFinished, hasRepeatedToolCalls } from 'ai';

  model,
  tools,
  // run to natural completion, but stop if stuck repeating the same call
  stopWhen: [isLoopFinished(), hasRepeatedToolCalls(3)],
});
```

`hasRepeatedToolCalls(count)` returns `true` when the last `count` steps all carry an identical set of `(toolName, input)` pairs.

Semantics that matter, learned from running this in production:

- **Compare the full set of calls per step**, not individual calls: a step's signature is its sorted `(toolName, serialized input)` pairs. This handles parallel tool calls correctly.
- **A step without tool calls breaks the streak** — text output means the model changed course.
- **Byte-identical input is the right signal.** Legitimate retries vary their input (a pagination cursor, a different query, a retry token). A stuck model re-emits the same bytes.
- **Optionally compare tool results too** (e.g. `hasRepeatedToolCalls(3, { compareResults: true })`): same input + same output means the loop is producing no new information, while same input + changing output (e.g. polling a status tool) is legitimate progress. This distinction eliminated our false positives on polling patterns.

Naming and exact options shape are obviously up for discussion — `hasRepeatedToolCalls` follows the existing `hasToolCall` convention.

### Why built-in rather than userland

The same argument as `isLoopFinished()` (which is `() => false` and still shipped as a named helper): naming the common safety patterns is what the stop-condition module is for. It would also give the `isLoopFinished()` docs caution a concrete recommended pairing, instead of leaving "the agent could run indefinitely" as an open warning.

Related: #13938 (warning when the default stop condition triggers) touches the same "why did my loop stop" DX area.

I'm happy to submit a PR (implementation + tests + docs for the loop-control page) if the design direction is agreeable.

### AI SDK Version

- ai: 7.0.0-beta.187 (proposal also applicable to the v6 release line)

lgrammel:
## Background

Uncapped agent loops need a targeted safeguard that detects genuinely stuck repeated tool calls without imposing a blunt step limit or stopping productive polling whose results are changing.

## Summary

Adds and exports hasRepeatedToolCalls(count, options?), validates count, compares order-independent tool-call multisets while preserving duplicates, ignores tool-call IDs and object key order, optionally compares final results or errors, rejects values that cannot be compared losslessly, and caches signatures for stable StepResult objects. The additive ai API remains covered by a patch changeset.

## Testing

Runtime and compile-time coverage includes thresholds, recent windows, streak breaks, changed calls, parallel ordering, duplicates, zero-argument calls, deterministic object comparison, result-aware polling, repeated and changing errors, missing, duplicate, mismatched, preliminary, and deferred results, unsupported and lossy values, invalid counts, caching, and integration through generateText, streamText, and ToolLoopAgent.

## End-to-end Validation

- Mock ToolLoopAgent example stopped after three repeated weather calls.
- ai and dependency packages built successfully; the built public export resolved to a callable hasRepeatedToolCalls function.

## Documentation

Adds a dedicated API reference, loop-control guidance, related stop-condition links, the core reference index entry, polling and limitation guidance, a mock ToolLoopAgent example, and an ai patch changeset.

## Related Issues

Fixes #17606

Co-authored-by: reizam <12397287+reizam@users.noreply.github.com>


lgrammel:
## Feature review

Outcome: changes-required

### Implements feature

Status: partially-implements

The new export, count validation, recent-step thresholding, order-independent multiset comparison, empty-step streak breaking, result matching, documentation, example, and changeset are present. However, `compareResults` does not reliably distinguish changing arbitrary tool outputs as required.

Concerns:
- `serializeForComparison` in `packages/ai/src/generate-text/stop-condition.ts:164-169` treats any non-throwing `JSON.stringify` result as lossless. Different outputs such as `{ progress: undefined }` versus `{}`, `[undefined]` versus `[null]`, and Maps with different entries therefore produce identical signatures and stop a polling loop despite changing results. A focused probe returned `true` for all three cases. Since `StepResult.toolResults` permits arbitrary output values, unsupported or lossy JSON values must be rejected or compared without collisions.

### Side effects

Risk: low

The diff adds an opt-in helper and export without changing existing stop-condition defaults or loop execution paths; existing behavior is unaffected unless users call `hasRepeatedToolCalls`.

Concerns: none.

### Performance

Risk: low

`packages/ai/src/generate-text/stop-condition.ts:90-161` limits processing to the last `count` steps, but serializes and sorts their complete inputs and optional outputs on every evaluation. This is proportional to the configured count and payload size and only applies when the helper is used.

Concerns: none.

### Backwards compatibility

Risk: low

The `ai` package change is additive: `packages/ai/src/generate-text/index.ts` adds one export, existing signatures are unchanged, and `.changeset/tidy-mice-repeat.md` correctly declares a patch release.

Concerns: none.

### Architecture

Risk: low

The implementation remains local to the core stop-condition module, operates on normalized `StepResult` data, and introduces no provider-spec changes, new dependencies, or cross-package coupling.

Concerns: none.

### Security

Risk: low

Inputs and outputs are not logged or transmitted, and serialization exceptions are caught in `serializeForComparison`; the feature does not add credential, URL, or provider-boundary handling.

Concerns: none.

### Testing

Status: needs-more

The added unit and type tests cover ordinary thresholds, parallel calls, duplicates, missing results, cyclic inputs, and invalid counts, and the mock ToolLoopAgent example runs successfully. They do not cover the arbitrary-output serialization collisions or the specifically requested error, deferred, preliminary-result, and generateText/streamText integration cases.

Concerns:
- Add regression coverage for lossy JSON serialization of arbitrary outputs and automated integration coverage across `generateText`, `streamText`, and `ToolLoopAgent`; also define and test behavior for tool errors, deferred results, and preliminary results as identified in the feature requirements.

### Verification

Inspected `git diff origin/main...HEAD`, surrounding stop-condition and StepResult/tool-result code, nested AGENTS.md, documentation, example, changeset, exports, and architecture guidance. `pnpm check`, root `pnpm type-check`, the `ai` package build, and focused Node and Edge stop-condition suites all passed; each focused suite ran 28 tests with no type errors. The mock ToolLoopAgent example stopped after three calls. A direct source probe demonstrated false positives for three pairs of changing non-JSON-compatible outputs.

reizam (commented):
I've been running this kind of guard in production for a while, so a few notes from that experience.

1. With `compareResults: true`, a model stuck re-issuing a call that fails identically every time will never trigger the condition. Failing calls produce `tool-error` parts, which aren't in `step.toolResults`, so the signature comes back `undefined` and the streak resets. In my experience this is the most common repetition loop by far: bad argument, error, same retry, same error. The data to cover it is already available (`tool-error` parts on `step.content` carry the tool name, input and error), no spec change needed. I'd like to send this as a follow-up PR once this lands, if that works for you.

2. `serializeForComparison` uses plain `JSON.stringify`, which is key-order sensitive. Two structurally equal inputs with different key insertion order produce different signatures and silently break the streak. There's already `canonicalJSON` in `util/canonical-hash.ts` (tool-fingerprint uses it) that sorts keys, seems like the right fit here.

3. Signatures are recomputed and re-serialized for the whole window on every evaluation. With `compareResults` and large tool outputs that's a lot of repeated `JSON.stringify` work for a condition that almost never fires. Since `StepResult` objects are stable identities this can be memoized per step. Also happy to handle that as a small perf follow-up rather than grow this PR.

4. One question: can `toolCall.input` be `undefined` for zero-argument tools? `JSON.stringify(undefined)` returns `undefined`, so such a step becomes non-comparable and a repeated no-arg call would never trigger the guard. If parsing always yields `{}` this is a non-issue.

reizam:
Took a run at the blocking `compareResults` concern since I'd flagged the serialization bit above. Patch and tests below, checked against this branch.

The core of it: `JSON.stringify` only throws on cycles and bigint. Everything else it happily represents, just lossily, so unequal outputs end up with the same signature:

| output A | output B | shared signature |
| --- | --- | --- |
| `{ progress: undefined }` | `{}` | `{}` |
| `[undefined]` | `[null]` | `[null]` |
| `new Map([['a',1]])` | `new Map([['b',2]])` | `{}` |
| `new Set([1])` | `new Set([2])` | `{}` |
| `NaN` | `Infinity` (and `null`) | `null` |
| `{ f: () => 1 }` | `{ f: () => 2 }` | `{}` |

Every one of those stops a polling loop whose results are still moving.

I went with the reject half of "rejected or compared without collisions". A false positive kills a legitimate loop, a false negative just lets it keep going, so erring toward reject seemed right. Turns out a `replacer` covers it. It runs on values after `toJSON`, and it fires for `undefined`, functions and symbols, so `JSON.stringify` still does the escaping, number formatting, `toJSON` and cycle handling:

```ts
function serializeForComparison(value: unknown): string | undefined {
  let isRepresentable = true;

  try {
    const serialized = JSON.stringify(value, (_key, entry) => {
      if (entry === null) {
        return entry;
      }

      switch (typeof entry) {
        case 'string':
        case 'boolean':
          return entry;
        case 'number':
          if (!Number.isFinite(entry)) {
            isRepresentable = false;
          }
          return entry;
        case 'object': {
          const prototype = Object.getPrototypeOf(entry);
          if (
            !Array.isArray(entry) &&
            prototype !== Object.prototype &&
            prototype !== null
          ) {
            // Map, Set and class instances without `toJSON` expose no
            // enumerable keys, so their contents cannot be compared.
            isRepresentable = false;
          }
          return entry;
        }
        default:
          // undefined, function, symbol
          isRepresentable = false;
          return entry;
      }
    });

    return isRepresentable ? serialized : undefined;
  } catch {
    // e.g. a cyclic value, a bigint, or a throwing `toJSON` or getter
    return undefined;
  }
}
```

`Date` and anything else with `toJSON` still work. A class instance without `toJSON` now makes the step non-comparable where it used to serialize fine. That one's deliberate, but say the word if you'd rather keep those comparable.

Couple of doc lines will need to follow. The JSDoc on `hasRepeatedToolCalls` says "JSON-serialized inputs", and `73-has-repeated-tool-calls.mdx` and `04-loop-control.mdx` both say values that "cannot be serialized with `JSON.stringify`" don't trigger the condition. That's under-inclusive now, since the reject set also covers values `JSON.stringify` takes without complaining.

<details>
<summary>Regression tests (16 cases, inside the existing <code>hasRepeatedToolCalls</code> describe)</summary>

```ts
    describe('comparison of values that JSON.stringify represents lossily', () => {
      function repeatsWithOutputs(first: unknown, second: unknown) {
        return hasRepeatedToolCalls(2, { compareResults: true })({
          steps: [
            createOutputStep({ toolCallId: 'call-1', output: first }),
            createOutputStep({ toolCallId: 'call-2', output: second }),
          ],
        });
      }

      function createOutputStep({
        toolCallId,
        output,
      }: {
        toolCallId: string;
        output: unknown;
      }): StepResult<any, any> {
        return createStepResult({
          toolCalls: [
            createToolCall({
              toolCallId,
              toolName: 'status',
              input: { jobId: 'job-1' },
            }),
          ],
          toolResults: [
            createToolResult({
              toolCallId,
              toolName: 'status',
              input: { jobId: 'job-1' },
              output,
            }),
          ],
        });
      }

      it('should not treat an undefined property as an absent property', () => {
        expect(repeatsWithOutputs({ progress: undefined }, {})).toBe(false);
      });

      it('should distinguish an undefined array item from null', () => {
        expect(repeatsWithOutputs([undefined], [null])).toBe(false);
      });

      it('should distinguish maps with different entries', () => {
        expect(
          repeatsWithOutputs(new Map([['a', 1]]), new Map([['b', 2]])),
        ).toBe(false);
      });

      it('should distinguish sets with different entries', () => {
        expect(repeatsWithOutputs(new Set([1]), new Set([2]))).toBe(false);
      });

      it('should distinguish NaN from a non-finite number', () => {
        expect(repeatsWithOutputs(Number.NaN, Number.POSITIVE_INFINITY)).toBe(
          false,
        );
      });

      it('should distinguish infinities of opposite sign', () => {
        expect(
          repeatsWithOutputs(
            Number.POSITIVE_INFINITY,
            Number.NEGATIVE_INFINITY,
          ),
        ).toBe(false);
      });

      it('should distinguish objects carrying different functions', () => {
        expect(
          repeatsWithOutputs({ notify: () => 'a' }, { notify: () => 'b' }),
        ).toBe(false);
      });

      it('should not treat a symbol-valued property as an absent property', () => {
        expect(repeatsWithOutputs({ token: Symbol('a') }, {})).toBe(false);
      });

      it('should reject an unrepresentable value nested inside a structure', () => {
        expect(
          repeatsWithOutputs(
            { meta: { cache: new Map([['a', 1]]) } },
            { meta: { cache: new Map([['b', 2]]) } },
          ),
        ).toBe(false);
      });

      it('should reject a value whose toJSON throws instead of throwing', () => {
        const throwing = () => ({
          toJSON() {
            throw new Error('boom');
          },
        });

        expect(repeatsWithOutputs(throwing(), throwing())).toBe(false);
      });

      it('should distinguish dates, which serialize through toJSON', () => {
        expect(repeatsWithOutputs(new Date(0), new Date(1))).toBe(false);
      });

      it('should treat equal dates as repeated', () => {
        expect(repeatsWithOutputs(new Date(0), new Date(0))).toBe(true);
      });

      it('should treat equal structures with every serializable leaf kind as repeated', () => {
        const output = () => ({
          done: false,
          percent: 0,
          note: null,
          items: [1, 'a', { deep: true }],
        });

        expect(repeatsWithOutputs(output(), output())).toBe(true);
      });

      it('should distinguish structures that differ only in a boolean leaf', () => {
        expect(repeatsWithOutputs({ done: false }, { done: true })).toBe(false);
      });

      it('should compare null-prototype objects, which serialize faithfully', () => {
        const output = () =>
          Object.assign(Object.create(null), { state: 'pending' });

        expect(repeatsWithOutputs(output(), output())).toBe(true);
      });

      it('should compare values that share a reference in two places', () => {
        const shared = { x: 1 };

        expect(
          repeatsWithOutputs(
            { a: shared, b: shared },
            { a: { x: 1 }, b: { x: 1 } },
          ),
        ).toBe(true);
      });
    });
```

</details>

On this branch: 44 tests green under both `vitest.node.config.js` and `vitest.edge.config.js`, `tsc --noEmit` clean, `oxlint` and `oxfmt` quiet. I also went back and mutated the thing to make sure the tests actually bite. Breaking any of the reject arms (`Number.isFinite`, the prototype check, the `default` case) or making the helper reject everything outright fails somewhere between 1 and 7 tests.

Separately, three holes in the existing tests I bumped into on the way. All pre-existing, all in the "Testing: needs-more" direction you already called out. Left them alone so this stays focused, but happy to send them on their own:

1. The `step.toolCalls.length === 0` guard in `createToolCallSignature`. The current test goes `[weather, empty, weather]`, and an empty step still produces a distinct signature there, so you can delete the guard and the suite stays green. It's two *consecutive* step-less steps that flip the answer to `true`.
2. `resultsByToolCallId.size !== step.toolResults.length` isn't covered. Two results sharing a `toolCallId` and the `Map` quietly drops one.
3. Same for the `toolResult.toolName !== toolCall.toolName` half of the pairing check. Only the `== null` half has a test.

PR #17646, opened 2026-07-21
[v6.0] reproduction: Gemini 3 parallel function calls: converter injects skipthoughtsignaturevalidator

lgrammel:
## Bug

Gemini 3 parallel function calls: converter injects `skip_thought_signature_validator` + warns every turn (only the first parallel call is signed)

## Reproduction

- Google documents that only the first function call in a Gemini 3 parallel batch carries a thought signature; subsequent unsigned calls are valid.
- The target `release-v6.0` versions (`ai@6.0.233`, `@ai-sdk/google@3.0.98`, and `@ai-sdk/gateway@3.0.155`) still reproduce the behavior, so upgrading from the reported versions does not resolve it.
- The recorded live response in `packages/google/src/__fixtures__/issue-16298-gemini3-parallel-tool-calls.json` contains a signed first Vertex call and an unsigned second parallel call.
- `examples/ai-functions/src/reproduction/issue-16298-gemini3-parallel-tool-calls.ts` replayed that valid batch and observed both the misleading application-code warning and sentinel injection, instead of accepting the batch without false attribution.
- `packages/google/src/issue-16298-gemini3-parallel-tool-calls.test.ts` asserts the expected `no-warning/no-sentinel` behavior and fails because the converter emits the reported warning.
- A new live Vertex request was not made because Google Vertex live-service checks are skipped by policy; the July 21, 2026 recorded live fixture was used.

## Relevant Documentation

- https://vercel.com/ai-gateway/models/gemini-3.1-flash-lite/providers
- https://ai-sdk.dev/providers/ai-sdk-providers/ai-gateway

## Related Issues

Reproduces #16298

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,13 @@
+{
+  "version": 1,
+  "issueId": 16298,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-16298-gemini3-parallel-tool-calls.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/issue-16298-gemini3-parallel-tool-calls.ts",
+    "packages/google/src/__fixtures__/issue-16298-gemini3-parallel-tool-calls.json"
+  ],
+  "expectedFailure": {
+    "exitCode": 1,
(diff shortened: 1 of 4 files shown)


lgrammel:
Issue #16298 already has the merged-v6.0 label. Closing this stale reproduction pull request.

PR #17651, opened 2026-07-21
fix(ai): skip smoothStream delays while the document is hidden

### LINKED ISSUE #20217: smoothStream stalls the stream in background tabs
### Description

When `streamText` runs in the browser with `experimental_transform: smoothStream()`, switching to another browser tab slows the stream to a crawl and it never recovers on its own (only fixed when the tab is focused again).

`smoothStream` awaits `delay(delayInMs)` between chunks:
https://github.com/vercel/ai/blob/72435300083e28d5027747eddd04e542edb5b0be/packages/ai/src/generate-text/smooth-stream.ts#L166-L171

Browsers clamp the chained timers to a minimum of 1 second in hidden documents, and then drops them even further to 1 minute after the document is hidden for more than 5 minutes. Since the smooth stream loop rely on such a timer chain, it effectively stops.

Because `transform` is awaited, this creates backpressure on the whole pipeline and doesn't just slow the rendering down - the upstream read stalls too. In a multi-step tool loop, the next step never fires.
That's a problem because the user sends their agent off to do some work, switches tabs, and comes back a few minutes later only to find that nothing progressed at all - the process only resumes once the tab is focused again.

### Reproduction

Fix, with tests, in #17651.

### Related Issues

- #9888
- #11865

### AI SDK Version

- `ai`: 7.0.90

### Code of Conduct

eden881:
## Background

When `streamText` runs in the browser (e.g. a fully client-side agent/tool loop consuming the stream directly), switching to another tab effectively freezes the stream and the agent: no text advances, tool calls stop being issued, and everything resumes only when the tab regains focus. No error or timeout fires.

The root cause is an interaction between `smoothStream` and browser timer throttling:

1. `smoothStream` awaits `delay(delayInMs)` (a `setTimeout`) between each emitted chunk.
2. Browsers heavily throttle timers in hidden documents - Chrome clamps chained timers to a minimum of 1 second, and after ~5 minutes to only once per minute ("intensive throttling"). The smoothing loop is exactly such a chain, so a hidden tab drains the stream at roughly one word per second, then one word per minute.
3. Because streams are pull-based, the stall propagates upstream through backpressure: the per-step stream never finishes draining, so the step's `flush` - which triggers tool execution results being consumed and the next `streamStep` in multi-step loops - never runs. The whole agent loop is gated on the smoothing delay.

We hit this in production with a browser-embedded agent, where the tool loop runs fully client-side, with tools executing against REST APIs. Users who switched tabs during long multi-step tasks found the agent frozen mid-task - no error, no timeout - until they returned to the tab.
We've since been running the equivalent of this fix in that app by injecting a visibility-aware delay through `_internal.delay`, which fully resolves the issue, including for mid-stream tab switches.

## Summary

Skip the smoothing delay while the document is hidden: `await delay(isDocumentHidden() ? null : delayInMs)`, where `isDocumentHidden()` checks `typeof document !== 'undefined' && document.visibilityState === 'hidden'`.

Smoothing has no visual purpose when nobody can see the page, so the stream drains at full speed there and resumes smoothing when the document becomes visible again. Behavior in Node/edge runtimes is unchanged (`document` is undefined).

## End-to-End Verification

Verified this branch's build of the package in a production browser-embedded agent app (client-side tool loop over `streamText` + `smoothStream`) where the bug was originally discovered - the app was pointed at the locally built package from this PR, with no other changes.

1. Before: start a long multi-step agent task, switch to another tab - the task freezes (no text progress, no further tool calls or model requests) and resumes only on returning to the tab.
2. After: the same scenario streams and executes tools to completion while the tab stays hidden, including when the tab is switched mid-stream. Word-by-word smoothing behaves normally while the tab is visible and resumes on returning to it.

## Checklist

## Related Issues

Fixes #20217

### DIFF
--- a/.changeset/smooth-stream-hidden-tab.md
+++ b/.changeset/smooth-stream-hidden-tab.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): skip `smoothStream` delays while the document is hidden, so background-tab timer throttling no longer stalls the stream and any tool loop gated on its consumption
--- a/content/docs/07-reference/01-ai-sdk-core/80-smooth-stream.mdx
+++ b/content/docs/07-reference/01-ai-sdk-core/80-smooth-stream.mdx
@@ -38,7 +38,7 @@ const result = streamText({
       type: 'number | null',
       isOptional: true,
       description:
-        'The delay in milliseconds between outputting each chunk. Defaults to 10ms. Set to `null` to disable delays.',
+        'The delay in milliseconds between outputting each chunk. Defaults to 10ms. Set to `null` to disable delays. The delay is skipped while the document is hidden (e.g. browser background tabs), where timer throttling would otherwise stall the stream.',
     },
     {
       name: 'chunking',
--- a/packages/ai/src/generate-text/smooth-stream.test.ts
+++ b/packages/ai/src/generate-text/smooth-stream.test.ts
@@ -2,7 +2,7 @@ import { convertArrayToReadableStream } from '@ai-sdk/provider-utils/test';
 import { smoothStream } from './smooth-stream';
 import type { TextStreamPart } from './stream-text-result';
 import type { ToolSet } from '@ai-sdk/provider-utils';
-import { beforeEach, describe, expect, it } from 'vitest';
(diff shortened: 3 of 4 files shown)


eden881:
Fixed in #20219

PR #17664, opened 2026-07-21
feat(harness-pi): support inline extension factories

jeffbarg:
## Background

`@ai-sdk/harness-pi` currently constructs Pi's `DefaultResourceLoader` with an empty `extensionFactories` array. This prevents `HarnessAgent` consumers from installing trusted inline Pi extensions that observe Pi's provider and inference events. Harness-level telemetry only sees the aggregate Pi turn and cannot reconstruct those internal calls.

Filesystem-discovered extensions must remain disabled because Pi runs in the host Node.js process, where discovery could otherwise load personal or project configuration into the server process.

## Summary

- Add an optional readonly `extensionFactories` setting to `createPi()` and thread it through `PiSessionSettings`.
- Pass a fresh mutable copy of the caller's factories to Pi's `DefaultResourceLoader`, preserving caller order.
- Preserve the active extension runtime during routine between-turn resource refreshes so factories are not repeatedly initialized.
- Load a fresh extension runtime when the underlying Pi session is genuinely rebuilt, avoiding reuse of Pi's invalidated runtime.
- Keep `noExtensions: true`, `noThemes: true`, and `noPromptTemplates: true`; only explicitly supplied inline factories can execute.
- Add unit coverage for invocation, ordering, the fresh-array boundary, multi-turn events, Pi session rebuilds, default empty behavior, and disabled filesystem discovery.
- Add contextual public API type coverage, README documentation, and a patch changeset for `@ai-sdk/harness-pi`.

This is a generic Pi extension hook. It adds no extension-specific dependency and does not change HarnessAgent telemetry.

Validation:

- `pnpm --filter @ai-sdk/harness-pi test` (122 tests passed)
- `pnpm --filter @ai-sdk/harness-pi build`
- `pnpm --filter @ai-sdk/harness-pi type-check`
- `pnpm exec ultracite check packages/harness-pi/src packages/harness-pi/README.md`
- `pnpm type-check:full`
- `git diff --check`

## Manual Verification

No live provider session was run. The public `createPi()` path is covered with a mocked Pi resource loader and session: one factory initialization observes two turns across routine loader refreshes, while a tool-set change rebuilds Pi with a distinct extension runtime. The tests also verify ordering, the fresh mutable array boundary, empty defaults, and all discovery-disable flags.

## Checklist


felixarntz (changes requested):
@jeffbarg Thank you, agreed that this is a useful enhancement for Pi!

Code-wise this mostly looks good, however please take a look at the agent feedback in https://github.com/vercel/ai/pull/17664#issuecomment-5209129121 - there's a concern about excessive re-instantiation of extensions.

jeffbarg:
Addressed the review feedback in 19b1993b5e:

- routine between-turn resource reloads now preserve the active extension runtime without re-running caller factories
- genuine Pi session rebuilds load a fresh extension runtime after the previous one is invalidated
- added multi-turn event coverage plus a tool-set rebuild regression test
- removed the README's transitive dependency import and added contextual public-API type coverage
- rebased onto current `main`; 122 scoped tests, package build/typecheck, scoped Ultracite checks, and `pnpm type-check:full` pass

@felixarntz, would you mind taking another look?

felixarntz (approved):
@jeffbarg Thank you, LGTM!

I just pushed one additional commit to document the new `extensionFactories` parameter for the Pi harness.

PR #17675, opened 2026-07-22
fix (provider/gateway): enforce order/only on gateway.models fallback list

### LINKED ISSUE #14376: providerOptions.gateway.order not respected when routing via providerOptions.gateway.models — unexpected provider (deepinfra) selected
## Summary

When using the Vercel AI gateway with `providerOptions.gateway.order` to restrict upstream provider routing, the `order` constraint is **not applied** to models selected via `providerOptions.gateway.models`. This causes traffic to route through unexpected providers (e.g. deepinfra) that are not listed in `order`.

## Reproduction

```ts
import { createGateway } from '@ai-sdk/gateway';

const gateway = createGateway({ apiKey: process.env.VERCEL_AI_API_KEY });

await generateText({
  model: gateway('openai/gpt-5.1'),
  output: Output.object({ schema }),
  messages: prompt,
  providerOptions: {
    gateway: {
      order: ['openai', 'google', 'azure'], // deepinfra NOT listed
      models: [
        'google/gemini-2.5-flash',
        'openai/gpt-5-mini',
        'xai/grok-4.1-fast-non-reasoning',  // xai routed through deepinfra internally
      ],
    },
  },
});
```

## Expected Behavior

- `order: ['openai', 'google', 'azure']` should restrict ALL upstream routing — including for models selected from `gateway.models` — to only the listed providers.
- If `xai/grok-4.1-fast-non-reasoning` is in `gateway.models` but `xai` / `deepinfra` is not in `order`, the gateway should skip that model or throw a clear error.

## Actual Behavior

- The gateway internally selects `xai/grok-4.1-fast-non-reasoning` from the `gateway.models` fallback list and routes it through **deepinfra** as the upstream backend.
- `order` is silently ignored for models picked via `gateway.models`; it only applies to provider routing for the **primary** model.
- There is no warning or error indicating that an unlisted provider was used.

## Root Cause (observed)

`providerOptions.gateway.order` and `providerOptions.gateway.models` are two independent routing mechanisms. The `order` constraint is applied at the provider level for the explicitly requested model, but it is not enforced when the gateway internally selects a fallback from the `models` list. The `models` list can reference model IDs whose backends map to providers outside `order` (e.g. `xai/*` → deepinfra).

## Impact

- Requests unexpectedly leave the intended provider boundary (compliance, cost, latency implications).
- No observability: there is no log, header, or metadata in the response indicating which provider was actually used.

## Environment

- `ai` SDK version: `6.0.116`
- `@ai-sdk/gateway` (latest)
- Runtime: Node.js v20.9.0

Christian-Sidak:
## Summary

- When `providerOptions.gateway.order` or `providerOptions.gateway.only` restrict which providers may be used, the same restriction is now applied client-side to the `providerOptions.gateway.models` fallback list before the request is sent.
- Models whose provider prefix (the part before the first `/`) is not in the allowed provider set are removed from `models` and surfaced as `type: 'other'` warnings.
- This prevents silent routing to unexpected providers (e.g. `deepinfra` for `xai/*` models) when the caller has explicitly declared a provider allowlist via `order` or `only`.

## Test plan

- [x] `should filter gateway.models to only providers in gateway.order` -- verifies `xai/grok-4.1-fast-non-reasoning` is stripped when `order` excludes `xai`, and a warning is emitted
- [x] `should filter gateway.models to only providers in gateway.only` -- same logic for `only`
- [x] `should not filter gateway.models when no order or only constraint is set` -- no filtering when neither constraint is present
- [x] `should keep all gateway.models when all providers are in gateway.order` -- no filtering when all models match
- [x] `should filter gateway.models for doStream when order is set` -- same filtering applies to the streaming path, warnings surface via the `stream-start` event
- [x] All 385 existing gateway tests still pass

Fixes #14376

### DIFF
--- a/.changeset/fix-gateway-order-models-filter.md
+++ b/.changeset/fix-gateway-order-models-filter.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+fix (provider/gateway): enforce gateway.order and gateway.only constraints on gateway.models fallback list
--- a/packages/gateway/src/gateway-language-model.test.ts
+++ b/packages/gateway/src/gateway-language-model.test.ts
@@ -1657,5 +1657,165 @@ describe('GatewayLanguageModel', () => {
         gateway: { quotaEntityId: 'entity-123', user: 'user-456' },
       });
     });
+
+    it('should filter gateway.models to only providers in gateway.order', async () => {
+      prepareJsonResponse({
+        content: { type: 'text', text: 'Test response' },
+      });
+
+      const { warnings } = await createTestModel().doGenerate({
+        prompt: TEST_PROMPT,
+        providerOptions: {
+          gateway: {
+            order: ['openai', 'google', 'azure'],
+            models: [
+              'google/gemini-2.5-flash',
+              'openai/gpt-5-mini',
+              'xai/grok-4.1-fast-non-reasoning',
+            ],
+          },
+        },
+      });
+
+      const requestBody = await server.calls[0].requestBodyJson;
+      expect(requestBody.providerOptions).toEqual({
+        gateway: {
+          order: ['openai', 'google', 'azure'],
+          models: ['google/gemini-2.5-flash', 'openai/gpt-5-mini'],
+        },
+      });
+      expect(warnings).toEqual([
+        {
+          type: 'other',
+          message:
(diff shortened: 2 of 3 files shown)


Christian-Sidak:
Closing — this was pushed without commit signing, which this repo requires, so it can't merge as-is. Will resubmit signed.

PR #17681, opened 2026-07-22
fix: Amazon Bedrock S3 image URLs are rejected instead of passed through to inference

lgrammel:
## Background

The published Amazon Bedrock provider rejected S3-hosted images before inference, forcing users to load image bytes into memory.

## Root Cause

The model declared no native URL support, causing the core downloader to reject s3:// URLs, while Bedrock message conversion also unconditionally rejected URL-backed files. Reproduction confirmed AI_DownloadError through generateText and AI_UnsupportedFunctionalityError through direct provider use.

## Summary

The provider now recognizes s3:// image URLs and serializes user and tool-result images as Bedrock s3Location sources. Bedrock image source types and a patch changeset were updated.

## Testing

Added regression coverage for native S3 URL declaration and conversion of user and tool-result S3 images.

## End-to-end Validation

- `AWS_REGION=us-east-1 pnpm -C examples/ai-functions exec tsx src/reproduction/amazon-bedrock-s3-image-url.ts` using a temporary Nova variant reached live Bedrock and returned `Provided S3Location not found`, confirming pass-through instead of the previous pre-inference `AI_DownloadError`.

## Related Issues

Fixes #15792

Closes #17676

Co-authored-by: tommyOtsai <171396937+tommyOtsai@users.noreply.github.com>

PR #17693, opened 2026-07-22
[v5.0] fix: Amazon Bedrock S3 image URLs are rejected instead of passed through to inference

### LINKED ISSUE #15792: @ai-sdk/amazon-bedrock S3 / Image URL support
## Problem
I want to pass images to models on amazon bedrock via s3Location. 

## Current Behavior
When you pass a normal https url, the ai-sdk correctly throws an error stating:
`` APICallError [AI_APICallError]: unsupported image_url scheme: must be `data:` or `s3://`; got ... ``

The [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ImageSource.html) states that images must be passed as s3 locations. 

However, when you pass an s3 location, the ai-sdk incorrectly throws an error stating: 
`DownloadError [AI_DownloadError]: URL scheme must be http, https, or data, got s3: at validateDownloadUrl...`

After looking closer at the bedrock implementation, it does not seem like s3 locations are supported as a way to pass images. The reason we need support for this is that our workflow has significant parallel fanout and handles a significant number of images. Holding the base 64 or byte representations of images causes buffer overflows. 

## Expected Behavior

s3 locations should be supported for amazon bedrock if possible.

## Reproduction Steps / Sample Code

1. Upload images to an s3 bucket
2. Try to pass those images to any model hosted on Amazon Bedrock

### AI SDK Version

"@ai-sdk/amazon-bedrock": "^4.0.111",

### Code of Conduct

lgrammel:
## Background

The published Amazon Bedrock provider rejected S3-hosted images before inference, forcing users to load image bytes into memory.

## Root Cause

The model declared no native URL support, causing the core downloader to reject s3:// URLs, while Bedrock message conversion also unconditionally rejected URL-backed files. Reproduction confirmed AI_DownloadError through generateText and AI_UnsupportedFunctionalityError through direct provider use.

## Summary

The provider now recognizes s3:// image URLs and serializes user and tool-result images as Bedrock s3Location sources. Bedrock image source types and a patch changeset were updated.

## Testing

Added regression coverage for native S3 URL declaration and conversion of user and tool-result S3 images.

## End-to-end Validation

- `AWS_REGION=us-east-1 pnpm -C examples/ai-functions exec tsx src/reproduction/amazon-bedrock-s3-image-url.ts` using a temporary Nova variant reached live Bedrock and returned `Provided S3Location not found`, confirming pass-through instead of the previous pre-inference `AI_DownloadError`.

## Related Issues

Fixes #15792

Closes #17676

Backport of #17681

Co-authored-by: tommyOtsai <171396937+tommyOtsai@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The provider now declares s3:// image URLs as native and converts URL-backed user images into Bedrock image.source.s3Location requests. A mocked generateText flow confirmed the URL reaches the provider request without downloading or base64 conversion.

### Side effects

Risk: low

The native URL declaration is limited to image media types and the s3:// scheme. HTTPS images retain their existing download behavior, while non-image S3 files and unsupported URL schemes remain rejected.

### Performance

Risk: none

The new path avoids downloading, buffering, and base64-encoding S3 images, reducing rather than increasing memory and processing costs.

### Backwards compatibility

Risk: none

There is no data migration or stored-data format change; only request-time serialization of previously rejected S3 image URLs changes.

### Architecture

Risk: none

The URL capability, Bedrock request types, and conversion logic remain localized within the Amazon Bedrock provider package, with no cross-package src imports, new dependencies, or misplaced shared abstractions.

### Change scope

Status: minimal

All six changed files are directly required for the provider capability, request type, regression coverage, or patch changeset; no unrelated implementation or generated artifacts are present.

### Security

Risk: low

The SDK no longer locally fetches S3 URLs and instead passes them to Bedrock, where access remains governed by the caller's AWS and S3 permissions. Other URL schemes are not newly accepted.

### Testing

Status: appropriate

Tests cover both root causes: native S3 URL declaration and exact s3Location serialization. The complete Amazon Bedrock Node and edge suites passed with 221 tests each, type checking passed, and focused integration and rejection checks confirmed end-to-end behavior and scheme/media isolation.

### Verification

Inspected every diff hunk against release-v5.0, reviewed provider and core URL-processing boundaries, confirmed AWS's documented ImageSource and S3Location shapes, ran diff validation, type checking, Node and edge package suites, and a mocked generateText request that produced the expected S3 image source.

### Relevant Documentation

- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ImageSource.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_S3Location.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html
- https://docs.aws.amazon.com/cli/latest/reference/bedrock-runtime/converse.html

PR #17694, opened 2026-07-22
[v6.0] fix: Amazon Bedrock S3 image URLs are rejected instead of passed through to inference

### LINKED ISSUE #15792: @ai-sdk/amazon-bedrock S3 / Image URL support

lgrammel:
## Background

The published Amazon Bedrock provider rejected S3-hosted images before inference, forcing users to load image bytes into memory.

## Root Cause

The model declared no native URL support, causing the core downloader to reject s3:// URLs, while Bedrock message conversion also unconditionally rejected URL-backed files. Reproduction confirmed AI_DownloadError through generateText and AI_UnsupportedFunctionalityError through direct provider use.

## Summary

The provider now recognizes s3:// image URLs and serializes user and tool-result images as Bedrock s3Location sources. Bedrock image source types and a patch changeset were updated.

## Testing

Added regression coverage for native S3 URL declaration and conversion of user and tool-result S3 images.

## End-to-end Validation

- `AWS_REGION=us-east-1 pnpm -C examples/ai-functions exec tsx src/reproduction/amazon-bedrock-s3-image-url.ts` using a temporary Nova variant reached live Bedrock and returned `Provided S3Location not found`, confirming pass-through instead of the previous pre-inference `AI_DownloadError`.

## Related Issues

Fixes #15792

Closes #17676

Backport of #17681

Co-authored-by: tommyOtsai <171396937+tommyOtsai@users.noreply.github.com>

### DIFF
--- a/.changeset/calm-pandas-smile.md
+++ b/.changeset/calm-pandas-smile.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
+Pass through `s3://` image URLs to Amazon Bedrock Converse as S3 image sources instead of downloading them.
--- a/packages/amazon-bedrock/src/bedrock-api-types.ts
+++ b/packages/amazon-bedrock/src/bedrock-api-types.ts
@@ -164,13 +164,7 @@ export interface BedrockGuardrailConverseContentBlock {
 
 export interface BedrockImageBlock {
   image: {
-<<<<<<< HEAD:packages/amazon-bedrock/src/bedrock-api-types.ts
     format: BedrockImageFormat;
-    source: {
-      bytes: string;
-    };
-=======
-    format: AmazonBedrockImageFormat;
     source:
       | {
           bytes: string;
@@ -180,7 +174,6 @@ export interface BedrockImageBlock {
             uri: string;
           };
         };
->>>>>>> 1f92bdb2c3 (fix: Amazon Bedrock S3 image URLs are rejected instead of passed through to inference (#17681)):packages/amazon-bedrock/src/amazon-bedrock-api-types.ts
   };
 }
 
--- a/packages/amazon-bedrock/src/bedrock-api-types.ts
+++ b/packages/amazon-bedrock/src/bedrock-api-types.ts
@@ -164,10 +164,23 @@ export interface BedrockGuardrailConverseContentBlock {
 
 export interface BedrockImageBlock {
   image: {
+<<<<<<< HEAD:packages/amazon-bedrock/src/bedrock-api-types.ts
     format: BedrockImageFormat;
     source: {
       bytes: string;
     };
+=======
+    format: AmazonBedrockImageFormat;
+    source:
+      | {
+          bytes: string;
+        }
+      | {
+          s3Location: {
+            uri: string;
+          };
+        };
+>>>>>>> 1f92bdb2c3 (fix: Amazon Bedrock S3 image URLs are rejected instead of passed through to inference (#17681)):packages/amazon-bedrock/src/amazon-bedrock-api-types.ts
   };
 }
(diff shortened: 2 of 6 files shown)


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The model now declares native s3:// image support to bypass downloading, and provider conversion emits Bedrock image.source.s3Location for user and tool-result images. This matches the documented ImageSource union and required image format. 

### Side effects

Risk: low

The behavioral change is restricted to s3:// image URLs; HTTP(S) assets retain their existing download behavior, non-S3 direct URLs remain unsupported, and byte-backed images are serialized as before.

### Performance

Risk: none

The new path stores and serializes only the S3 URI instead of downloading and base64-encoding image data, reducing memory and processing overhead for the reported workload.

### Backwards compatibility

Risk: none

The change does not read, migrate, or rewrite stored data; it only changes construction of new Bedrock inference requests.

### Architecture

Risk: none

The URL recognition remains on the Bedrock model while Bedrock-specific source conversion and types remain private to the provider package. No cross-package src imports, dependency reversals, or misplaced shared abstractions were introduced.

### Change scope

Status: minimal

All six changed files are directly necessary: provider capability declaration, Bedrock request types and conversion, focused regression tests, and the required patch changeset.

### Security

Risk: low

Pass-through is limited to the s3 protocol and delegates object access to Bedrock under the caller's AWS permissions; arbitrary URL schemes are not newly accepted. AWS documents the relevant S3 permission and ownership requirements. 

### Testing

Status: appropriate

Regression tests cover native URL capability plus exact s3Location conversion for user and tool-result images. The complete provider test suite passed in Node and edge configurations, along with root type checking, provider build, formatting, and lint checks.

### Verification

Inspected every diff hunk and relevant prompt-conversion path, compared the reproduced failure path with the fix, confirmed AWS request-schema compatibility, and successfully ran 387 provider tests in both runtime configurations plus type, build, format, lint, and diff checks.

### Relevant Documentation

- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ImageSource.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_S3Location.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ImageBlock.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html
- https://docs.aws.amazon.com/cli/latest/reference/bedrock-runtime/converse.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html

PR #17712, opened 2026-07-22
[v6.0] reproduction: Media-type sniffing performs unbounded work for ID3-prefixed attachments

lgrammel:
## Bug

Media-type sniffing performs unbounded work for ID3-prefixed attachments

## Reproduction

- On `release-v6.0`, the vulnerable implementation is in `packages/ai/src/util/detect-media-type.ts`; adding an ID3 prefix to a 1 MiB attachment caused image, audio, and video sniffing to decode all 1,398,104 base64 characters instead of a bounded prefix.
- Raw ID3-prefixed `Uint8Array` detection copied 1,048,566 of 1,048,576 bytes, confirming attachment-sized memory work.
- Raw-byte and base64 MP3 inputs with a 128 KiB ID3 tag were both detected as `audio/mpeg`, but inputs one byte over the limit were also scanned and detected instead of returning `undefined`.
- A malformed ID3 syncsafe size field caused the complete 1,398,104-character attachment to be decoded without an out-of-bounds exception.
- PNG, JPEG, WEBP, WAV, base64url, data URL, `ArrayBuffer`, `Buffer`, and text normalization controls retained their existing behavior; `release-v6.0` has no PDF signature in `packages/ai/src/util/detect-media-type.ts`, so PDF auto-detection was not an existing control on this branch.
- `examples/ai-functions/src/reproduction/issue-17709-id3-media-sniffing.ts` contains the runnable reproduction and fails when sniffing exceeds the fixed 128 KiB ID3 scan bound.

## Relevant Documentation

- https://id3.org/id3v2-00

## Related Issues

Reproduces #17709

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "issueId": 17709,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17709-id3-media-sniffing.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/issue-17709-id3-media-sniffing.ts"
+  ],
+  "expectedFailure": {
+    "exitCode": 1,
+    "signal": "ISSUE #17709 REPRODUCED: ID3-prefixed media sniffing processed attachment-sized data"
+  }
+}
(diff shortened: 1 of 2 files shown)


lgrammel:
Issue #17709 already has the merged-v6.0 label. Closing this stale reproduction pull request.

PR #17717, opened 2026-07-22
fix: include parsed structured output in streamText end callbacks

### LINKED ISSUE #11418: v6: streamText onFinish callback no longer has object prop for structured data
### Description

After I upgraded to v6, some of my code that relied on the `onFinish` callback of `streamObject` broke because the new alternative, `streamText` using `Output`, do not return an `object` prop as the previous iteration. This prop used to be the full parsed object after the stream had finished, and right now I don't understand what is the intended way for use to get the complete object at the end (to save on the DB, for example)

### AI SDK Version

- ai: 6.0.3
- @ai-sdk/google: 3.0.1

### Code of Conduct

lgrammel:
## Background

The `streamText` structured-output replacement omitted the complete parsed output from `onEnd`/`onFinish`, preventing callback-based persistence workflows supported by `streamObject`.

## Root Cause

`streamText` assembled end events exclusively from step results, while complete output parsing was only exposed through `result.output`.

## Summary

- expose schema-inferred optional `output` in `onEnd`/`onFinish`
- cache complete-output parsing so callbacks and `result.output` share one parse
- keep `result.output` settlement independent from arbitrary callback completion
- export `StreamTextEndEvent` and `StreamTextOnEndCallback`
- keep telemetry events restricted to their existing shape
- update docs, reproduction, focused tests, and patch changeset

Callers that need to wait for asynchronous callback work can await `result.consumeStream()`.

## Testing

- callback output and schema inference
- reusable public callback/event types
- single complete-output parse
- parse failures return `undefined` to the callback while `result.output` rejects
- `result.output` remains independent from active callbacks
- callbacks can await `result.output` after asynchronous work
- full `ai` Node and Edge suites
- workspace type checking, linting, formatting, and package build
- provider-independent reproduction

## Related Issues

Fixes #11418

Closes #17701

Co-authored-by: pltoledo <57258959+pltoledo@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: still-reproduces

The exact original bug signal appeared when replayed against the pull request.

Concerns:
- The original reported bug remains reproducible after the fix.

### Fixes issue

Status: does-not-address

The callback eventually receives parsed output, but only after result.output may already have resolved.

Concerns:
- The implementation resolves the step promise before invoking the output-enriched callback, creating an observable completion race.
- A caller awaiting result.output can finish or terminate before its onFinish persistence work has run.
- The exact original reproduction still produces the reported bug.

### Side effects

Risk: low

Existing callbacks configured with structured output are now delayed by complete-output parsing.

Concerns:
- Asynchronous or slow schema validation now delays callback invocation and stream completion.

### Performance

Risk: low

Complete structured output can be parsed independently for the callback and for result.output.

Concerns:
- Because the output getter is not cached, callers using both paths incur duplicate JSON parsing, schema validation, and transient object allocation.

### Backwards compatibility

Risk: none

The change adds an optional callback property and does not migrate or reinterpret existing stored data.

### Architecture

Risk: none

The callback type and implementation remain localized to streamText, preserve telemetry filtering, and introduce no cross-package source imports or dependency violations.

### Change scope

Status: minimal

The production change, callback typing, regression snapshot, documentation, and patch changeset are all directly related to the reported bug.

### Security

Risk: none

The change adds no external I/O, credential handling, URL processing, or new exposure beyond the user-supplied callback.

### Testing

Status: needs-more

The package tests pass, but the runtime test waits for full stream consumption and therefore misses the completion-order race caught by the reproduction.

Concerns:
- A regression test must cover awaiting result.output and verify that the completion callback and its persistence work have completed at the expected point.
- The exact provider-independent reproduction remains failing.

### Verification

Inspected every changed diff hunk and the stream completion, output parsing, callback dispatch, telemetry, and legacy streamObject paths. The package Node test suite and diff validation passed, while both the authoritative replay and focused provider-independent verification demonstrated that callback output is unavailable immediately after result.output resolves; an additional timing check confirmed it appears only on a later turn.

### Relevant Documentation

PR #17723, opened 2026-07-22
[v5.0] reproduction: Media-type sniffing performs unbounded work for ID3-prefixed attachments

lgrammel:
## Bug

Media-type sniffing performs unbounded work for ID3-prefixed attachments

## Reproduction

- On `release-v5.0`, the affected implementation is in `packages/ai/src/util/detect-media-type.ts`: ID3-prefixed base64 is decoded before the 24-character prefix limit, and raw input is copied with `Uint8Array.slice`.
- Each image, audio, and combined-signature check decoded all 1,398,104 characters of a 1 MiB ID3-prefixed attachment; equivalent non-ID3 sniffing decoded only 24 characters.
- Raw-byte sniffing copied 1,048,566 bytes from a 1,048,576-byte attachment, confirming attachment-sized memory work.
- MP3 data behind a 128 KiB ID3 tag was correctly detected, but raw and base64 inputs with a 128 KiB-plus-one-byte tag were also detected as `audio/mpeg` instead of returning `undefined` at the fixed scan boundary.
- A malformed maximum syncsafe size field decoded the complete 1,398,104-character attachment without an out-of-bounds exception.
- PNG, JPEG, WEBP, WAV, base64url, data URL, `ArrayBuffer`, `Buffer`, text, and `raw/base64` boundary parity controls retained existing behavior; PDF auto-detection is absent from the target branch and returned `undefined`.
- `examples/ai-functions/src/reproduction/issue-17709-id3-media-sniffing.ts` is the runnable reproduction and intentionally fails because sniffing performs unbounded work rather than remaining independent of attachment size.

## Relevant Documentation

- https://id3.org/id3v2-00

## Related Issues

Reproduces #17709

### DIFF
--- a/.ai-sdk-factory/reproduction-replay.json
+++ b/.ai-sdk-factory/reproduction-replay.json
@@ -1,0 +1,15 @@
+{
+  "version": 1,
+  "issueId": 17709,
+  "command": "pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17709-id3-media-sniffing.ts",
+  "artifactPaths": [
+    "examples/ai-functions/src/reproduction/issue-17709-id3-media-sniffing.ts",
+    "examples/ai-functions/package.json",
+    "examples/ai-functions/tsconfig.json",
+    "pnpm-lock.yaml"
+  ],
+  "expectedFailure": {
+    "exitCode": 1,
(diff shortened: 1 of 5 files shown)


lgrammel:
Issue #17709 already has the merged-v5.0 label. Closing this stale reproduction pull request.

PR #17725, opened 2026-07-22
[v6.0] fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input

### LINKED ISSUE #17709: Media-type sniffing performs unbounded work for ID3-prefixed attachments

lgrammel:
## Summary

## Hardening

## Happy path

## Unhappy path

## Testing

Fixes #17709

## Related Issues

Backport of #17386

### DIFF
--- a/.changeset/bound-id3-media-type-decode.md
+++ b/.changeset/bound-id3-media-type-decode.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input
+
+Media-type detection stripped ID3 tags before the ~18-byte prefix cap, decoding the entire base64 attachment (plus a full-size copy) whenever the data began with `ID3`/`SUQz`. This turned the intended O(1) sniff into an O(N) decode of the whole attachment. Detection now decodes at most a bounded prefix and skips the ID3 tag within that bound, keeping cost O(1) in input size on all paths (image, audio, and combined).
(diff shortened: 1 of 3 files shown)


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

Detection now decodes only fixed-size prefixes, uses zero-copy Uint8Array views, preserves MP3 detection at the configured ID3 boundary, and returns undefined beyond that boundary.

### Side effects

Risk: low

Normal signature detection remains limited to the existing 18-byte prefix; the only intentional behavior change is that ID3 tags beyond the fixed scan boundary no longer auto-detect.

### Performance

Risk: low

Base64 work is capped at approximately 128 KiB for ID3 input and 18 bytes otherwise, while raw-byte handling uses subarray views rather than attachment-sized copies.

### Backwards compatibility

Risk: low

No stored bytes are modified. Existing oversized ID3 attachments may require an explicitly supplied media type when reprocessed, which is the issue's specified fallback.

### Architecture

Risk: none

The backport changes the existing media-detection utility in the owning ai package and consumes provider utilities through their declared package export, without cross-package src access or new coupling.

### Change scope

Status: minimal

The three changed files are limited to the detection implementation, focused boundary regression tests, and the required patch changeset.

### Security

Risk: none

The change reduces resource-exhaustion exposure, bounds malformed size-field handling, and introduces no new parsing, injection, or data-access surface.

### Testing

Status: appropriate

Tests cover ordinary and ID3-tagged MP3s, raw/base64 parity at and beyond the scan boundary, and the existing image and audio signature controls.

### Verification

Inspected every diff hunk and relevant caller; the 48-test detection suite passed in Node and Edge configurations, ai package type checking passed, repository formatting and lint checks passed, and a focused large-input probe confirmed bounded decoding and safe malformed-size handling. The linked ID3 specification was also reviewed. 

### Relevant Documentation

- https://id3.org/id3v2-00

PR #17726, opened 2026-07-22
[v5.0] reproduction: AIInvalidPromptError "messages must be a ModelMessage" caused by

lgrammel:
## Bug

AI_InvalidPromptError ("messages must be a ModelMessage[]") caused by undefined properties in tool-result arrays

## Reproduction

- On `release-v5.0` with `ai` 5.0.218, manually reusing tool-response messages containing nested `undefined` properties caused both `streamText` and `generateText` to reject the follow-up with `AI_InvalidPromptError` caused by `AI_TypeValidationError`, preventing the expected continuation.
- JSON serialization removed the `undefined` keys and made the follow-up succeed, so retained `undefined` values—not heterogeneous arrays with missing keys—are the actual failure condition.
- Replacing the values with `null` succeeded, and automatic multi-step execution with `stopWhen: stepCountIs(2)` also succeeded, matching the issue comments.
- The runnable reproduction is `examples/ai-functions/src/reproduction/issue-10893-undefined-tool-result-array.ts`, supported by `examples/ai-functions/package.json` and `examples/ai-functions/tsconfig.json`.

## Related Issues

Reproduces #10893

PR #17744, opened 2026-07-22
[v5.0] fix: prevent reused tool-result messages with undefined properties from failing prompt validation

### LINKED ISSUE #10893: AI_InvalidPromptError ("messages must be a ModelMessage[]") caused by undefined properties in tool-result arrays
### Description

I am encountering an `AI_InvalidPromptError` when using the `tool` helper. The top-level error message suggests a mismatch between UI/Model messages, but the underlying cause is a schema validation failure (`AI_TypeValidationError`) on the tool result.

My tool returns an array of objects where some properties are optional (`string | undefined`). When the tool executes, some objects have these properties set to `undefined`, causing the keys to be stripped during JSON serialization. The SDK's internal validation for the `tool-result` message then fails, seemingly expecting a uniform shape across all items in the array or failing to handle the missing keys that were typed as optional.

**Reproduction:**

1.  Define a tool that returns an array of objects.
2.  Ensure the return type infers specific fields as `string | undefined`.
3.  Return an array where some items have the field defined, and others have it as `undefined`.
4.  Trigger the tool via `generateText` or `streamText`.

**Code Snippet:**

```typescript
import { tool } from 'ai';
import { z } from 'zod';

// Tool Definition
const listTasksTool = tool({
  description: 'List tasks',
  inputSchema: z.object({
    limit: z.number().optional(),
  }),
  // No explicit outputSchema provided, relying on inference
  execute: async ({ limit = 20 }) => {
    // Simulating database response where 'startedAt' is optional
    const tasks = [
      {
        id: "1",
        state: "completed",
        startedAt: "2025-10-26T17:59:40.065Z", // Field exists
        completedAt: "2025-10-26T18:00:31.713Z"
      },
      {
        id: "2",
        state: "archived",
        startedAt: undefined, // Field is undefined (will be stripped in JSON)
        completedAt: undefined 
      }
    ];

    return tasks; 
  },
});

// Inferred Return Type confirms optionality:
// Tool<..., { id: string; startedAt: string | undefined; ... }[]>
```

**The Issue:**
The resulting `tool-result` message passed to the `messages` array looks like this (keys are missing for the second item):

```json
{
  "role": "tool",
  "content": [
    {
      "type": "tool-result",
      "toolName": "tasks---list_tasks",
      "output": {
        "type": "json",
        "value": [
          { "id": "1", "startedAt": "2025-10-26..." },
          { "id": "2" } // "startedAt" is missing entirely here
        ]
      }
    }
  ]
}
```

**Error Log:**
The application crashes with the following error:

```text
AI_InvalidPromptError: Invalid prompt: The messages must be a ModelMessage[]. If you have passed a UIMessage[], you can use convertToModelMessages to convert them.
```

Upon inspecting the `cause`, it reveals an `AI_TypeValidationError` (Zod validation failure) on the missing keys in the array:

```json
{
  "name": "AI_TypeValidationError",
  "cause": [
    {
      "code": "invalid_union",
      "errors": [
        // ... validation path showing failure on index 1 of output.value ...
        {
          "code": "invalid_type",
          "expected": "string",
          "message": "Invalid input: expected string, received undefined",
          "path": ["startedAt"]
        }
      ]
    }
  ]
}
```

**Expected Behavior:**
The SDK should handle `tool-result` payloads where keys are missing if the value was `undefined`, or the generic inference should be lenient enough to accept heterogeneous arrays in `output.value` without requiring an explicit `outputSchema` that forces `null`. The error message regarding `ModelMessage[]` vs `UIMessage[]` is also confusing in this context as the messages are structurally correct aside from the tool payload validation.

### AI SDK Version

- ai: 5.0.106

### Code of Conduct

lgrammel:
## Background

Manually reusing SDK response messages with nested undefined tool-result properties caused streamText and generateText to reject valid follow-up calls with AI_InvalidPromptError.

## Root Cause

Tool execution preserved nested undefined object properties, but the recursive JSON record schema required every property value to be a JSONValue. The reproduction confirmed the properties remained present with undefined values and that JSON serialization, which removed them, avoided validation failure.

## Summary

Allowed undefined object properties in the internal JSON value validator, added regression coverage and an ai patch changeset, and removed all reproduction-only artifacts.

## Testing

Added a standardizePrompt regression test verifying that JSON tool outputs containing nested undefined object properties are accepted.

## End-to-end Validation

- `node --input-type=module` inline streamText/generateText reproduction: both follow-up calls accepted directly reused response messages containing nested undefined properties.

## Related Issues

Fixes #10893

Closes #17726

Co-authored-by: wladimiiir <3164600+wladimiiir@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: inconclusive

Concerns:

### Fixes issue

Status: unclear

The recursive validator now accepts undefined object-property values inside JSON tool outputs, directly addressing the validation failure when response messages are reused.

Concerns:

### Side effects

Risk: low

The shared JSON schema is also used for provider options and metadata, so the relaxation is broader than tool results, but it only permits undefined object properties that JSON serialization would omit.

Concerns:
- Provider options and metadata containing undefined properties will now pass validation as well.

### Performance

Risk: none

The change adds only an optional branch during existing recursive validation and introduces no copies, retained state, or additional traversal.

### Backwards compatibility

Risk: none

The change only accepts additional in-memory inputs and performs no migration or mutation of stored data; persisted JSON will continue omitting undefined properties.

### Architecture

Risk: medium

The runtime schema is widened inside the ai package, but the canonical JSONValue type and JSON guards owned by @ai-sdk/provider remain strict, requiring an assertion that makes the schema's declared type inconsistent with its accepted values.

Concerns:
- Update the provider-owned JSONObject type and isJSONValue/isJSONObject guards consistently, as the repository's newer release branches do, rather than using a consumer-side type assertion.
- The provider package will need corresponding tests and release metadata if its canonical JSON contract is widened.

### Change scope

Status: minimal

The final diff contains only the schema change, a focused regression test, and the required ai patch changeset; no unrelated production files or artifacts are included.

### Security

Risk: low

The validator still rejects non-JSON values and undefined array elements; accepting undefined object properties does not introduce execution, injection, or privilege paths.

### Testing

Status: needs-more

The ai node suite and type-check pass, including the new nested-output test, but that test uses an any cast and exercises standardizePrompt directly rather than typed response-message reuse through streamText and generateText.

Concerns:
- Add a stable mock-model regression covering generated response messages reused by both streamText and generateText.
- Add provider-level type and guard tests so runtime validation, JSONValue, isJSONValue, and isJSONObject agree about undefined object properties.

### Verification

Inspected every merge-base diff hunk and all JSON schema consumers, compared the provider-owned JSON type and guards with newer repository branches, ran the ai node test suite with 1,204 passing tests, ran the ai TypeScript build successfully, and confirmed the diff passes whitespace checks.

aayush-kapoor:
starting a new one!

PR #17745, opened 2026-07-22
[v6.0] fix: prevent warnings and skip-validator injection for valid Gemini 3 parallel function calls

### LINKED ISSUE #16298: Gemini 3 parallel function calls: converter injects `skip_thought_signature_validator` + warns every turn (only the first parallel call is signed)

lgrammel:
## Background

Gemini 3 legitimately signs only the first standard function call in a parallel model response, but replaying later unsigned calls caused misleading warnings and unnecessary skip-validator injection.

## Root Cause

The Google converter evaluated missing signatures per normalized tool-call part rather than recognizing that one assistant message represents a model response whose first standard parallel function call carries the shared signature. 

## Summary

The converter now tracks signed standard function calls across the entire assistant response, preserves state across mixed parts, excludes built-in server calls from seeding that state, retains fallback mitigation for genuinely missing signatures, and includes a patch changeset.

## Testing

Regression coverage verifies three-call parallel responses, intervening text and server parts, signed-server/unsigned-function behavior, genuine missing-signature fallback, and provider namespace handling. Google Node and Edge suites, full workspace type checking, and focused formatting/lint checks pass.

## End-to-end Validation

- `pnpm -C packages/google build && pnpm -C examples/ai-functions exec tsx src/stream-text/google/function-call-id.ts` — live Gemini calls produced three parallel tool calls and completed replay with final text.
- `pnpm -C examples/ai-functions exec tsx -e '<inline Gemini 3.5 Flash parallel-call probe>'` — observed `[signed,unsigned,unsigned]`, an empty warnings array, and successful final text generation.

## Related Issues

Fixes #16298

Closes #17620

Backport of #17649

### DIFF
--- a/.changeset/green-gemini-tools.md
+++ b/.changeset/green-gemini-tools.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Avoid missing thought-signature warnings and skip-validator injection for valid unsigned Gemini 3 parallel function calls in the same model response.
--- a/packages/google/src/convert-to-google-generative-ai-messages.test.ts
+++ b/packages/google/src/convert-to-google-generative-ai-messages.test.ts
@@ -1346,7 +1346,7 @@ describe('Gemini 3 missing thoughtSignature mitigation', () => {
 
   it('does NOT inject the sentinel or warn for unsigned parallel calls after a signed call', () => {
     const onWarning = vi.fn();
-    const result = convertToGoogleMessages(
+    const result = convertToGoogleGenerativeAIMessages(
       [
         {
           role: 'assistant',
@@ -1377,7 +1377,7 @@ describe('Gemini 3 missing thoughtSignature mitigation', () => {
       ],
       {
         isGemini3Model: true,
-        providerOptionsNames: ['googleVertex', 'vertex'],
+        providerOptionsName: 'googleVertex',
         onWarning,
       },
     );
@@ -1413,7 +1413,7 @@ describe('Gemini 3 missing thoughtSignature mitigation', () => {
 
   it('does NOT inject the sentinel when other response parts separate parallel function calls', () => {
     const onWarning = vi.fn();
-    const result = convertToGoogleMessages(
+    const result = convertToGoogleGenerativeAIMessages(
       [
         {
           role: 'assistant',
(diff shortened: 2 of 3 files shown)


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The converter now recognizes a signed standard function call as covering subsequent unsigned parallel function calls in the same assistant response, while resetting state between responses, excluding server-tool calls from seeding it, and retaining mitigation for genuinely missing first signatures.

### Side effects

Risk: low

The behavior change is narrowly limited to Gemini 3 unsigned standard function calls following a signed standard call in the same model response; non-Gemini models, separate assistant responses, server-tool-only signatures, and genuinely unsigned batches retain existing behavior.

### Performance

Risk: none

The implementation adds one response-local boolean and constant-time checks per tool-call part, with no additional collections, I/O, or asymptotic cost.

### Backwards compatibility

Risk: low

No stored data is mutated or migrated. Existing persisted signed-first/unsigned-later parallel responses are serialized more faithfully, while genuinely missing first signatures still receive the fallback sentinel and warning.

### Architecture

Risk: none

The logic remains localized in the Google provider's message converter, which owns Gemini wire-format serialization and is already shared with Vertex through declared package dependencies and exports; no cross-package source imports or new coupling were introduced.

### Change scope

Status: minimal

All three changed files are necessary: the localized converter fix, focused regression coverage, and the required patch changeset. No unrelated abstractions, dependencies, generated artifacts, or production changes were added.

### Security

Risk: none

The change introduces no new input parsing, credential handling, network access, dynamic execution, or trust-boundary changes.

### Testing

Status: appropriate

Regression tests cover three parallel calls, intervening text and server parts, server signatures not seeding standard-call state, namespace resolution, and preservation of the genuine missing-signature fallback.

### Verification

Inspected the complete merge-base diff and relevant converter call paths, compared the backport with the upstream fix, confirmed a clean diff, and verified the Google Node and Edge suites passed with 570 tests each. Package type checking and focused formatting/lint checks also passed.

### Relevant Documentation

PR #17746, opened 2026-07-22
[v5.0] fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input

### LINKED ISSUE #17709: Media-type sniffing performs unbounded work for ID3-prefixed attachments

lgrammel:
## Summary

## Hardening

## Happy path

## Unhappy path

## Testing

Fixes #17709

## Related Issues

Backport of #17386

### DIFF
--- a/.changeset/bound-id3-media-type-decode.md
+++ b/.changeset/bound-id3-media-type-decode.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input
+
+Media-type detection stripped ID3 tags before the ~18-byte prefix cap, decoding the entire base64 attachment (plus a full-size copy) whenever the data began with `ID3`/`SUQz`. This turned the intended O(1) sniff into an O(N) decode of the whole attachment. Detection now decodes at most a bounded prefix and skips the ID3 tag within that bound, keeping cost O(1) in input size on all paths (image, audio, and combined).
--- a/.changeset/bound-id3-media-type-decode.md
+++ b/.changeset/bound-id3-media-type-decode.md
@@ -1,7 +1,7 @@
 ---
-'@ai-sdk/provider-utils': patch
+'ai': patch
 ---
 
-fix(provider-utils): bound media-type sniffing decode for ID3-prefixed input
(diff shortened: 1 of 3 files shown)


lgrammel:
## Bugfix review

Outcome: approved

### Fixes issue

Status: fully-addresses

The implementation limits base64 decoding to a fixed prefix, uses zero-copy Uint8Array subarray views, handles ID3 from decoded bytes, preserves detection at the inclusive 128 KiB boundary, and returns undefined beyond it.

### Side effects

Risk: low

Normal non-ID3 signature detection remains limited to the existing 18-byte prefix. The only deliberate behavior change is that oversized ID3 tags are no longer scanned to their trailing media signature.

### Performance

Risk: low

ID3 processing now performs attachment-size-independent work: base64 decoding is capped at roughly 128 KiB and raw input is viewed through subarray rather than copied.

### Backwards compatibility

Risk: low

The change does not mutate or migrate stored data. It only changes runtime auto-detection for ID3 tags beyond the documented scan limit, where callers can continue supplying an explicit media type.

### Architecture

Risk: low

The fix remains localized in the v5 ai package's existing media-detection utility, uses provider-utils through its declared public package export, and introduces no cross-package src imports or inappropriate shared abstractions.

### Change scope

Status: minimal

The three changed files are limited to the detection implementation, focused boundary regression tests, and the required patch changeset.

### Security

Risk: low

The change reduces denial-of-service exposure from crafted ID3-prefixed attachments and uses clamping subarray operations for malformed or oversized size fields without adding new trust boundaries.

### Testing

Status: appropriate

Tests cover raw and base64 MP3 detection at and immediately beyond the 128 KiB boundary. The complete 48-test media-detection suite passed in both Node and edge configurations, and focused verification also covered bounded decoding, zero raw slicing, malformed sizes, combined signatures, and normalized input forms.

### Verification

Reviewed every merge-base diff hunk, confirmed clean diff formatting and package boundaries, passed the ai package type-check, passed all 48 media-detection tests in Node and edge environments, and independently verified bounded decoding and raw-copy behavior across boundary, malformed, image, audio, combined, base64url, data URL, ArrayBuffer, Buffer, and text cases.

### Relevant Documentation

- https://id3.org/id3v2-00
- https://id3.org/id3v2.3.0
- https://id3.org/id3v2.4.0-structure

PR #17752, opened 2026-07-22
fix: replay regular functions named tool_search correctly in OpenAI Responses history

### LINKED ISSUE #17402: @ai-sdk/openai: function named tool_search is replayed as native tool search
### Description

The OpenAI Responses history converter treats any tool named `tool_search` as OpenAI’s provider defined tool search, including an ordinary function created with `tool()`.

The ordinary function is initially sent correctly:

```json
{
  "type": "function",
  "name": "tool_search"
}
```

When its call is included in a later request’s history, it is serialized as `tool_search_call` instead of `function_call`. The normal function input does not contain the native tool’s wrapper fields, so `arguments` is omitted and OpenAI rejects the request:

```text
Missing required parameter: 'input[...].arguments'
```

The [history converter currently selects this path by resolved tool name alone](https://github.com/vercel/ai/blob/341616a326e6a210524d92d546e30af474e2aa6a/packages/openai/src/responses/convert-to-openai-responses-input.ts#L358-L390).

A regular function named `tool_search` should round-trip as `function_call` / `function_call_output`. Native serialization should only apply when the configured provider-defined tool is `openai.tool_search`.

### Reproduction

Configure a regular function (not `openai.tools.toolSearch()`) with this name:

```ts
const tools = {
  tool_search: tool({
    description: 'Search synthetic records',
    inputSchema: z.object({
      query: z.string(),
      limit: z.number(),
    }),
    execute: async () => 'No matches',
  }),
};
```

Include a call to it in the message history for a subsequent Responses request:

```ts
{
  type: 'tool-call',
  toolCallId: 'call_123',
  toolName: 'tool_search',
  input: {
    query: 'synthetic query',
    limit: 10,
  },
}
```

Inspecting the outgoing request shows a native item with no arguments:

```json
{
  "type": "tool_search_call",
  "call_id": null,
  "execution": "server"
}
```

Expected:

```json
{
  "type": "function_call",
  "call_id": "call_123",
  "name": "tool_search",
  "arguments": "{\"query\":\"synthetic query\",\"limit\":10}"
}
```

### AI SDK Version

@ai-sdk/openai: 4.0.8

The same name-only conversion is present in 4.0.15 and current main.

### Code of Conduct

lgrammel:
## Background

Regular functions named `tool_search` were replayed as native OpenAI tool-search calls without the required function arguments, causing subsequent Responses API requests to fail.

## Root Cause

The Responses history converter selected native tool-search serialization from the resolved tool name instead of the configured `openai.tool_search` provider-tool identity. Regression request inspection and live API validation confirmed the misclassification.

## Summary

The configured native tool-search name is passed to the history converter, which restricts native call and output serialization to that provider-defined tool. The PR also includes focused regression coverage and a patch changeset, with no dependency or configuration changes.

## Testing

Coverage verifies that a regular `tool_search` function produces `function_call` and `function_call_output` while native tool-search handling remains intact. The OpenAI Node and Edge suites each passed 835 tests, and full type checking and formatting/lint checks passed.

## End-to-end Validation

- `pnpm exec tsx -e ...` from `examples/ai-functions`: a live OpenAI Responses request accepted the replayed regular `tool_search` function, preserved its JSON arguments and output, and completed with `finishReason.unified: stop`.

## Related Issues

Fixes #17402

Closes #17430

Co-authored-by: guitard0g <11187231+guitard0g@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: inconclusive

Concerns:

### Fixes issue

Status: unclear

The converter now selects all three native tool-search call/output branches using the configured openai.tool_search tool identity rather than the resolved name, while the regression test verifies ordinary tool_search functions serialize as function_call and function_call_output.

Concerns:

### Side effects

Risk: low

The behavior change is narrowly limited to tool-search history classification; focused verification confirmed native and ordinary tool_search entries remain distinguishable even when both are configured.

### Performance

Risk: none

The change adds only bounded tool-list lookups during request preparation and no persistent allocations, buffering, or additional processing of prompt content.

### Backwards compatibility

Risk: low

No stored-data format is changed or migrated; existing history is serialized according to whether the corresponding native provider tool is configured for the request.

### Architecture

Risk: none

Provider-specific identity detection remains within the OpenAI package, the converter receives a package-local configuration value, and the touched test replaces a cross-package src import with the owning package's declared export.

### Change scope

Status: minimal

The six-file diff consists of the converter and caller changes, focused regression/native-path test adjustments, and the required patch changeset.

### Security

Risk: none

The change performs only string identity checks and does not alter credential handling, URL access, parsing policy, or execution boundaries.

### Testing

Status: appropriate

Regression coverage checks the complete ordinary function call/output request shape, existing native tool-search reconstruction coverage remains passing, and the focused Node and Edge suites both passed.

### Verification

Inspected every merge-base diff hunk; focused Node and Edge runs each passed 405 tests, the root TypeScript build and formatting/lint checks passed, and a bounded converter probe confirmed native tool-search and regular tool_search serialization remain distinct. Approval is blocked solely because the authoritative replay was inconclusive.

### Relevant Documentation

- https://platform.openai.com/docs/api-reference/responses
- https://developers.openai.com/api/docs/guides/function-calling
- https://developers.openai.com/api/docs/guides/tools-tool-search

PR #17782, opened 2026-07-22
fix: accept array-based chat content with thinking and text parts

### LINKED ISSUE #13703: openai-compatible: chat content arrays with thinking parts fail validation
### Description

When using some OpenAI-compatible backends (for example [Ollama](https://github.com/ollama/ollama/blob/main/docs/api/openai-compatibility.mdx) with [Mistral Small 4](https://mistral.ai/news/mistral-small-4)), chat responses can return structured content arrays instead of plain strings:

- Non-stream: `choices[].message.content` can be an array
- Stream: `choices[].delta.content` can be an array

Some providers include `thinking` parts (for example `{ "type": "thinking", "thinking": [{ "type": "text", "text": "..." }] }`).

Current behavior rejects these payloads in `@ai-sdk/openai-compatible` with validation errors like:

`Type validation failed ... expected string, received array`

This breaks both generate and stream flows for otherwise OpenAI-compatible providers.

## Reproduction

1. Use `@ai-sdk/openai-compatible` with a provider that emits content arrays with `thinking` parts.
2. Call `doGenerate` or `doStream`.
3. Observe schema/type validation error when content is array-based.

Example stream chunk:

```json
{
  "id": "de2d54af68ac42b49519a72022e3c5c3",
  "object": "chat.completion.chunk",
  "model": "mistral-small-latest",
  "choices": [
    {
      "index": 0,
      "delta": {
        "content": [
          {
            "type": "thinking",
            "thinking": [{ "type": "text", "text": "Hello" }]
          }
        ]
      },
      "finish_reason": null
    }
  ]
}
```

## Expected Behavior

- `content` should accept string and array forms for OpenAI-compatible providers.
- Known parts should be normalized:
  - `type: "text"` -> text output
  - `type: "thinking"` -> reasoning output
- Unknown part types should not crash parsing/validation.

### AI SDK Version

- `@ai-sdk/openai-compatible`: current `master` / `3.0.0-beta.9` in repo

### Code of Conduct

lgrammel:
## Background

OpenAI-compatible backends returning array-based chat content caused published generate and stream APIs to reject otherwise valid responses instead of returning reasoning and text.

## Root Cause

The non-stream and stream response schemas only accepted string content, and both processing paths assumed strings. Reproduction tests confirmed Zod rejected arrays at choices[].message.content and choices[].delta.content.

## Summary

Added tolerant string-or-array content parsing, normalized text and thinking parts while preserving order, ignored unknown part types, maintained streaming block transitions, and added a patch changeset.

## Testing

Added four regression cases covering generate and stream normalization plus unknown-part handling in Node and Edge runtimes.

## End-to-end Validation

- `pnpm -C examples/ai-functions exec tsx -e '<live Mistral generate/stream probe>'` — live `mistral-small-latest` calls returned reasoning and text for both generate and stream without validation errors.

## Related Issues

Fixes #13703

Closes #17779

Co-authored-by: tobwen <1864057+tobwen@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

Both response schemas now accept string-or-array content, and shared normalization converts text and thinking parts for generate and stream flows while ignoring unknown part types.

### Side effects

Risk: low

Existing string-content behavior remains covered and passing; the additional stream transitions correctly prevent overlapping text and reasoning blocks when array parts are interleaved.

### Performance

Risk: low

Array processing is linear in the number and size of returned content parts, with only short-lived result and reasoning-string allocations.

### Backwards compatibility

Risk: none

The change does not migrate or mutate stored data and only enables response forms that previously failed validation.

### Architecture

Risk: none

The schema and normalization helper remain localized to the OpenAI-compatible chat implementation, use existing provider abstractions, and introduce no cross-package source imports or dependency changes.

### Change scope

Status: minimal

Every changed file supports the fix: localized implementation changes, four regression cases, and the required patch changeset.

### Security

Risk: low

Unknown fields are not executed or dynamically accessed; normalization extracts only validated string values and safely ignores unsupported structures.

### Testing

Status: appropriate

Regression coverage includes non-stream and stream normalization plus unknown-part tolerance, while existing string-content behavior remains covered in both Node and Edge runtimes.

### Verification

Inspected every diff hunk and relevant stream-consumer behavior. Focused Node and Edge suites each passed 97 tests; package type checking, diff validation, and targeted lint and formatting checks also passed.

### Relevant Documentation

- https://docs.mistral.ai/studio-api/conversations/reasoning
- https://docs.mistral.ai/studio-api/conversations/chat-completion

aayush-kapoor (approved):

PR #17869, opened 2026-07-23
feat: sessions

n1ckoates:
## Background

Providers are intentionally stateless, but sometimes, they need to use resources or data that persists across multiple calls (to `doGenerate` or `doStream`). AI SDK had no way for providers to do this before though, or "attach" data to a specific `generateText` or `streamText` lifecycle.

## Summary

Added `Experimental_Session`.
- A `Session` has a key-value store, backed by a `Map` with a similar API. This store can be used by provider packages.
- AI SDK Core creates one `Session` for each `generateText()` or `streamText()` call, and passes the same session to every provider call within that loop.
- Providers can access the session through `callOptions.experimental_session`.
- AI SDK Core runs `sesion.destroy()` when the loop is complete, and no more provider calls will be done.
- Each item stored in a `Session` can also have an `onDestroy` callback attached. `session.destroy()` runs all of these callbacks. All `onDestroy` are attemped but if any error, the error bubbles up. If the `Session` is being destroyed because of a different error, the error is aggregated together with that `onDestroy` error.

In the future, AI SDK users will be able to create a `Session` themselves and pass it via `experimental_session`, and handle when to destroy the session. This is useful to persist resources across multiple calls.

## End-to-End Verification

Added `session-lifecycle-middleware.ts` example that checks that session exists using middleware and logs.

## Checklist

## Future Work

* make `createSession` available publicly
* allow passing in `experimental_session` to `generateText()` / `streamText()`
* eventually make this not experimental or remove it


packages/ai/src/generate-text/stream-text.ts
```diff
@@ -2443,15 +2476,26 @@ class DefaultStreamTextResult<
         }),
       );
     })().catch(async error => {
-      await telemetryDispatcher.onError?.({ callId, error });
-      self._initialResponseMessages.reject(error);
+      let finalError = error;
+
+      const cleanupError = await destroySession(error);
```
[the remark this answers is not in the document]
n1ckoates: Doesn't seem like anything new.

packages/ai/src/generate-text/create-session.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const errors: unknown[] = [];
+
+    for (const item of items) {
+      try {
+        await item.onDestroy?.();
+      } catch (error) {
+        errors.push(error);
+      }
+    }
```
felixarntz: could use a `Promise.allSettled` for this so that it's run in parallel

packages/ai/src/generate-text/generate-text.ts
```diff
@@ -1466,8 +1472,21 @@ export async function generateText<
...
       });
     } catch (error) {
-      await telemetryDispatcher.onError?.({ callId, error });
-      throw wrapGatewayError(error);
+      let finalError = error;
+
+      try {
+        await session.destroy();
+      } catch (cleanupError) {
+        if (cleanupError !== error) {
```
felixarntz, line 1489: this had me hesitate - if the error is the one from `session.destroy`, does it remain the same instance because you return the same promise instance from in-memory in `DefaultSession`?

if so, it works, it just seems a bit brittle. we could alternatively rely on a specific error class (like `SessionDestroyError`) for this type of error and check against that.

not a must-have but wanted to raise it

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -2443,15 +2460,30 @@ class DefaultStreamTextResult<
...
+      let finalError = error;
+
+      try {
+        await session.destroy();
+      } catch (cleanupError) {
+        finalError = new AggregateError(
+          [error, cleanupError],
+          'Streaming failed and session cleanup also failed.',
+        );
+      }
```
felixarntz, line 2587: don't we need a similar check here to ensure `cleanupError` isn't the same error as `error`?

felixarntz (approved):
Looks great! a few non-blocking suggestions

packages/ai/src/generate-text/create-session.ts
n1ckoates: `session.destroy()` currently works by running `onDestroy()` on each `SessionItem` in last in, first out order. So it's sequential.

idea there is that you could have some resource that depends on the previous resource. like:

• create a websocket.
• create some sort of client that depends on that websocket.
and .destroy should unwind that in reverse order of creation.

which is also why it doesn't run in parallel

we could change that and state that `onDestroy()` should be able to be run in whatever order and still work, and not depend on something else being destroyed first

packages/ai/src/generate-text/generate-text.ts
n1ckoates: it works, it's the same error instance because it's the same promise instance.

i agree that it's brittle tho. relies on the implementation of `Session` returning the same error.

we could rely on a specific error class. or alternatively, move `session.destroy()` so it only runs once:
```ts
let generationError: unknown;

try {
  // tool loop only — no session.destroy() here
} catch (error) {
  generationError = error;
}

try {
  await session.destroy(); // single place
} catch (cleanupError) {
  if (generationError != null) {
    throw new AggregateError(
      [generationError, cleanupError],
      'Generation failed and session cleanup also failed.',
    );
  }
  throw cleanupError;
}

if (generationError != null) {
  throw generationError;
}

// onEnd + return result (destroy already succeeded)
```

packages/provider/src/language-model/v4/language-model-v4-session.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * A key-value store for state that is scoped to a single language model
+ * lifecycle.
+ *
+ * Providers can associate cleanup callbacks with values. The owner of the
+ * session is responsible for destroying it when the lifecycle is complete.
+ */
+export type Session = {
```
gr2m: per our convention, this should have the V4. I'd match the file name and call it `LanguageModelV4Session` - exported as `Experimental_LanguageModelV4Session`

packages/ai/src/generate-text/stream-text.ts
n1ckoates: the block this `.catch()` is attached to never calls `session.destroy()`, so the error can't possibly be a session destroy error. so no need to de-duplicate the error.

packages/provider/src/language-model/v4/language-model-v4-call-options.ts
```diff
@@ -110,6 +111,12 @@ export type LanguageModelV4CallOptions = {
    */
   abortSignal?: AbortSignal;
 
+  /**
+   * Session state that is shared across language model calls in the same
+   * lifecycle.
+   */
+  experimental_session?: Session;
```
gr2m: just for reference, this addition is what we have concerns about. We usually treat any change to our specs like a breaking change. But when testing different spec versions that only differ by an optional key we didn't run into build errors.

.changeset/healthy-ants-roll.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ai-sdk/provider": patch
```
gr2m: this should be minor, since we are adding a new API. We use `patch` in general for all other packages, but I think with the spec we should follow semantic versioning more strictly.


```suggestion
"@ai-sdk/provider": minor
```

packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1379,6 +1386,15 @@ class DefaultStreamTextResult<
 
       async flush(controller) {
         try {
+          try {
+            await session.destroy();
+          } catch (cleanupError) {
+            self.rejectResultPromises(cleanupError);
+            await onError({ error: cleanupError });
+            controller.error(cleanupError);
+            return;
```
gr2m: @aayush-kapoor do we need to call `telemetryDispatcher.onError`/ `telemetryDispatcher.onEnd` before the return?

examples/ai-functions/src/generate-text/openai/session-lifecycle-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    if (session == null) {
+      throw new Error('Core did not provide a Session.');
+    }
+
+    latestSession = session;
+
+    const resource = session.has(resourceKey)
+      ? session.get<ProbeResource>(resourceKey)!
+      : session.set(
```
gr2m: just a thought: if this is a common pattern, maybe we should consider adding `getOrSet(key, factory, { onDestroy })`? I tend to no since it's such a low-level API, but we can get back to it if we see this pattern a lot.

packages/ai/src/generate-text/create-session.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    expect(session.delete('object')).toBeUndefined();
+
+    expect(session.set('object', object)).toBe(object);
+    expect(session.set(symbolKey, 'symbol value')).toBe('symbol value');
+
+    expect(session.has('object')).toBe(true);
+    expect(session.has(symbolKey)).toBe(true);
+    expect(session.get<typeof object>('object')).toBe(object);
+    expect(session.get<string>(symbolKey)).toBe('symbol value');
+    expectTypeOf(session.get<typeof object>('object')).toEqualTypeOf<
```
gr2m: I think we usually don't mix runtime and type tests. Instead, we create `.test-d.ts` files for type testing. But if this is a pattern you saw in other places then we can leave it and create a follow up issue to clean it up, or discuss if this is something we are okay with. Maybe update contribuing/ docs

gr2m (approved):

gr2m (changes requested):
blocking shortly to check out two things /cc @kkdawkins

PR #17884, opened 2026-07-23
fix (provider/bytedance): surface last_frame_url as providerMetadata.bytedance.lastFrameUrl

### LINKED ISSUE #17883: `@ai-sdk/bytedance`: returnLastFrame is sent to the API, but the returned last_frame_url never surfaces
I'm building multi-scene video generation where consecutive Seedance clips chain together, so I reached for `returnLastFrame` — the docs describe it as "Whether to return the last frame of the generated video. Useful for chaining consecutive videos."

The request side works: `bytedance-video-model.ts` forwards `body.return_last_frame` when the option is set, and ModelArk includes `content.last_frame_url` in the task status response (it's defined alongside `video_url` in ByteDance's own SDK types, e.g. the ark runtime `Content` model in volcengine-python-sdk). But `byteDanceStatusResponseSchema` only declares `video_url` inside `content`, so zod strips `last_frame_url` during parsing. There's no warning, and `providerMetadata.bytedance` comes back as just `{ taskId, usage }`. As far as I can tell the frame is unreachable no matter what you do.

Repro:

```ts
const result = await generateVideo({
  model: bytedance.video('seedance-1-0-pro-250528'),
  prompt: '...',
  providerOptions: { bytedance: { returnLastFrame: true } },
});
// result.providerMetadata.bytedance → { taskId, usage } — no frame URL
```

I also confirmed it mechanically: pointing `@ai-sdk/bytedance@2.0.14` at a mock server that returns the documented status response (`content: { video_url, last_frame_url }`) shows `return_last_frame: true` in the request body and no `last_frame_url` anywhere in the result.

I have a small fix (schema line + surfacing it as `providerMetadata.bytedance.lastFrameUrl`, with a regression test) — PR incoming.

Worth flagging: I hit this through the AI Gateway, so even after the package fix ships, gateway users won't see the field until the gateway's adapter is redeployed with the updated provider.

**Environment**

- `ai` 7.0.34
- `@ai-sdk/bytedance` 2.0.14 (schema unchanged on current `main`)
- Node 22
- Hit via the AI Gateway (`createGateway` + `gateway.video('bytedance/seedance-2.0')`), reproduced against `@ai-sdk/bytedance` directly with a mock ModelArk server returning the documented `content: { video_url, last_frame_url }` status body

jsphellis:
## Background

`returnLastFrame` is documented as useful for chaining consecutive videos, and the provider forwards the request param — but the task-status response schema strips `content.last_frame_url`, so the frame never reaches the caller. Fixes #17883.

## Summary

- Add `last_frame_url` to the `content` object in `byteDanceStatusResponseSchema`
- Surface it as `providerMetadata.bytedance.lastFrameUrl` when present
- Regression test (mocked succeeded status with `last_frame_url`), docs line, and changeset

## End-to-End Verification

Reproduced the missing field against the live API: with `returnLastFrame: true` the generation succeeds and nothing frame-related appears in the result. I don't have direct ModelArk credentials (I call through the AI Gateway), so the patched path is verified against ModelArk's documented response shape (`content: { video_url, last_frame_url }`, matching ByteDance's own SDK types) via a mock server: `@ai-sdk/bytedance@2.0.14` drops the field, the patched provider surfaces it as `providerMetadata.bytedance.lastFrameUrl`, with `taskId`/`usage`/video output unchanged.

### DIFF
--- a/.changeset/spotty-bees-shake.md
+++ b/.changeset/spotty-bees-shake.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/bytedance': patch
+---
+
+fix (provider/bytedance): surface last_frame_url as providerMetadata.bytedance.lastFrameUrl when returnLastFrame is set
--- a/content/providers/01-ai-sdk-providers/85-bytedance.mdx
+++ b/content/providers/01-ai-sdk-providers/85-bytedance.mdx
@@ -398,6 +398,7 @@ type them with `ByteDanceVideoModelOptions`.
 - **returnLastFrame** _boolean_
 
   Whether to return the last frame of the generated video. Useful for chaining consecutive videos.
+  When set, the frame URL is returned as `providerMetadata.bytedance.lastFrameUrl`.
 
 - **serviceTier** _'default' | 'flex'_
 
--- a/packages/bytedance/README.md
+++ b/packages/bytedance/README.md
@@ -170,7 +170,7 @@ object with `ByteDanceVideoModelOptions`.
 | `watermark`       | Whether to add a watermark to the generated video                              |
 | `generateAudio`   | Generate synchronized audio for supported Seedance models                      |
 | `cameraFixed`     | Whether to fix the camera during generation                                    |
-| `returnLastFrame` | Return the last frame of the generated video (useful for chaining)             |
+| `returnLastFrame` | Return the video's last frame as `providerMetadata.bytedance.lastFrameUrl`     |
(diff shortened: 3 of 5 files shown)


jsphellis:
@shaper — flagging a small bytedance provider fix (4 lines + regression test): the status response schema was stripping `content.last_frame_url` that `returnLastFrame` requests, so the documented chaining workflow can't be built. Details and repro in #17883. Happy to adjust anything.

aayush-kapoor (commented):
closed via https://github.com/vercel/ai/pull/19920

PR #17893, opened 2026-07-24
fix(react): throttle useChat messages snapshot to prevent maximum update depth error

### LINKED ISSUE #6166: Maximum update depth exceeded
### Description

I'm facing an issue while using the useChat hook for streaming in an Expo Bare workflow. The chat streaming works fine on iOS but encounters problems specifically on Android.

**What I've Tried:**

1. Tested with different versions of the package.
2. Experimented with various options of smoothstream. 
3. configure experimental_throttle.

takumiz19:
## Background

`useChat` re-renders past `throttle` / `experimental_throttle` and can hit React's "Maximum update depth exceeded" during streaming on slower devices (Android, heavy markdown, CPU throttling). Reported in #6166.

The root cause is a mismatch between what `experimental_throttle` throttles and what actually drives re-renders. `useChat` subscribes to messages via `useSyncExternalStore`:

```ts
const messages = useSyncExternalStore(
  subscribeToMessages,
  () => chatRef.current.messages, // getSnapshot
  () => chatRef.current.messages,
);
```

The throttle was applied **only to the subscribe callback** (`~registerMessagesCallback`). But `replaceMessage` swaps `#messages` for a **new array on every chunk**, and that is exactly what `getSnapshot()` returned. `useSyncExternalStore` re-reads `getSnapshot()` on every render and, when the reference differs from the rendered one, forces a synchronous re-render (tearing prevention). That path never goes through `subscribe`, so throttling the subscribe callback did not limit it — during streaming the array reference changes on essentially every commit, so React re-renders per chunk regardless of the throttle. Once renders fall behind chunk arrival it trips the nested-update limit.

This matches the reports in the thread: it is machine/load dependent (a threshold, not a hard infinite loop), and raising the throttle / using `chunking: 'line'` / memoizing markdown only reduces how much work each forced render does, rather than capping the re-render count.

## Summary

Keep the published snapshot separate from the internal message buffer so that `getSnapshot()` returns a **stable reference between throttle windows**:

- `ReactChatState` now holds `#messages` (internal buffer, always current — used by `AbstractChat` for slicing/finding/replacing) and a separate `#messagesSnapshot` (the reference React reads).
- Throttling is moved onto the state itself: `#publishMessages` swaps `#messagesSnapshot = #messages` **and** notifies subscribers together, and that whole publish is throttled. So the notification and the snapshot are throttled in lockstep, and `experimental_throttle` behaves as documented.
- `useChat`'s `getSnapshot` now reads `chatRef.current.messagesSnapshot` instead of `messages`.

When throttling is disabled (the default) publishing is synchronous, so behavior is unchanged.

## End-to-End Verification

Reproduced with a long streamed response under Chrome DevTools 6x CPU throttling: before this change, `experimental_throttle: 100` did not prevent "Maximum update depth exceeded"; after it, streaming stays throttled and no longer trips the limit.

The added regression test (`does not bypass throttling when the component re-renders for an unrelated reason`) fails on `main` (an unrelated re-render mid-window shows the un-throttled buffered content) and passes with this change. The full `@ai-sdk/react` suite (90 tests) passes.

## Checklist

## Related Issues

Fixes #6166

### DIFF
--- a/.changeset/eleven-melons-attack.md
+++ b/.changeset/eleven-melons-attack.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/react': patch
+---
+
+fix (react): throttle the `useChat` messages snapshot to prevent "Maximum update depth exceeded" during streaming
+
+`throttle`/`experimental_throttle` only throttled the subscribe callback, while `useSyncExternalStore`'s `getSnapshot` read `messages` directly and returned a fresh array on every stream chunk. Any re-render then read that un-throttled snapshot and re-rendered per chunk, which could exceed React's nested update limit on slower devices. The React chat state now publishes a separate throttled snapshot so both the notification and the snapshot are throttled together.
--- a/packages/react/src/chat.react.ts
+++ b/packages/react/src/chat.react.ts
@@ -10,7 +10,17 @@ import { throttle } from './throttle';
 class ReactChatState<
   UI_MESSAGE extends UIMessage,
 > implements ChatState<UI_MESSAGE> {
+  // Internal, always-current message buffer. Every read/write from
+  // `AbstractChat` (slicing, finding, replacing during streaming) goes through
+  // this, so the chat logic always sees the latest messages.
   #messages: UI_MESSAGE[];
+
+  // The message array reference published to React via
+  // `useSyncExternalStore`'s `getSnapshot`. It is only swapped when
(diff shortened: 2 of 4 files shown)


gr2m:
closing in favor of https://github.com/vercel/ai/pull/18525

PR #17922, opened 2026-07-24
feat(provider/cartesia): support additional streaming PCM encodings

31Carlton7:
## Summary

- add `providerOptions.cartesia.streaming.encoding` for Ink 2 streaming transcription
- support Cartesia's `pcm_s32le`, `pcm_f16le`, and `pcm_f32le` input encodings while preserving the existing inferred defaults
- document the option and cover the exact WebSocket query mapping

## Why

Cartesia accepts multiple raw PCM representations for realtime STT, but the provider currently always maps `audio/pcm` to `pcm_s16le`. An explicit Cartesia streaming option exposes the additional formats without changing the shared transcription specification or inventing new media-type strings.

This shape also lets AI Gateway read the same provider option and meter 16-bit versus 32-bit audio correctly before forwarding it.

## Validation

- `pnpm --filter @ai-sdk/cartesia test` (40 node + 40 edge tests)
- `pnpm --filter @ai-sdk/cartesia type-check`
- `pnpm check`
- `pnpm type-check:full`

### DIFF
--- a/.changeset/tidy-pumas-listen.md
+++ b/.changeset/tidy-pumas-listen.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/cartesia': patch
+---
+
+feat(provider/cartesia): support additional Ink 2 streaming PCM encodings
--- a/content/providers/01-ai-sdk-providers/95-cartesia.mdx
+++ b/content/providers/01-ai-sdk-providers/95-cartesia.mdx
@@ -246,6 +246,26 @@ Ink 2 supports English audio and uses Cartesia's native turn detection by
 default. To finalize only when the input audio stream ends, pass
 `providerOptions.cartesia.streaming.turnDetection: false`.
 
+By default, the provider maps `audio/pcm`, `audio/pcmu`, and `audio/pcma` to
+16-bit PCM, G.711 μ-law, and G.711 A-law respectively. To stream another raw
+PCM representation, set `providerOptions.cartesia.streaming.encoding` to
+`pcm_s32le`, `pcm_f16le`, or `pcm_f32le`. For example:
+
+```ts
+const result = streamTranscribe({
+  model: cartesia.transcription('ink-2'),
+  audio,
+  inputAudioFormat: { type: 'audio/pcm', rate: 48000 },
+  providerOptions: {
+    cartesia: {
+      streaming: {
+        encoding: 'pcm_f32le',
+      },
+    },
+  },
+});
+```
+
 See [Transcription](/docs/ai-sdk-core/transcription) for more information about
 streaming transcription.
 
--- a/packages/cartesia/src/cartesia-transcription-model-options.ts
+++ b/packages/cartesia/src/cartesia-transcription-model-options.ts
@@ -10,6 +10,20 @@ export const cartesiaTranscriptionModelOptionsSchema = z.object({
   streaming: z
     .object({
       /**
+       * Raw audio encoding sent to Ink 2. Defaults to the encoding inferred
+       * from `inputAudioFormat.type`.
+       */
+      encoding: z
+        .enum([
+          'pcm_alaw',
+          'pcm_f16le',
+          'pcm_f32le',
+          'pcm_mulaw',
+          'pcm_s16le',
+          'pcm_s32le',
+        ])
+        .optional(),
+      /**
        * Use Cartesia's native turn detection endpoint. Defaults to true.
        * Disable this to finalize the transcript only when the audio stream ends.
        */
--- a/packages/cartesia/src/cartesia-transcription-model-options.ts
+++ b/packages/cartesia/src/cartesia-transcription-model-options.ts
@@ -1,5 +1,15 @@
 import { z } from 'zod/v4';
 
+/** Raw audio encodings accepted by Ink 2 streaming transcription. */
+export const cartesiaStreamingEncodingSchema = z.enum([
+  'pcm_alaw',
+  'pcm_f16le',
+  'pcm_f32le',
+  'pcm_mulaw',
+  'pcm_s16le',
+  'pcm_s32le',
+]);
+
 // https://docs.cartesia.ai/api-reference/stt/transcribe
 export const cartesiaTranscriptionModelOptionsSchema = z.object({
   /** The language of the audio (ISO 639-1 code). Defaults to English. */
@@ -13,16 +23,7 @@ export const cartesiaTranscriptionModelOptionsSchema = z.object({
        * Raw audio encoding sent to Ink 2. Defaults to the encoding inferred
        * from `inputAudioFormat.type`.
        */
-      encoding: z
-        .enum([
-          'pcm_alaw',
-          'pcm_f16le',
-          'pcm_f32le',
-          'pcm_mulaw',
-          'pcm_s16le',
-          'pcm_s32le',
-        ])
-        .optional(),
+      encoding: cartesiaStreamingEncodingSchema.optional(),
       /**
        * Use Cartesia's native turn detection endpoint. Defaults to true.
        * Disable this to finalize the transcript only when the audio stream ends.
--- a/packages/cartesia/src/cartesia-transcription-model.test.ts
+++ b/packages/cartesia/src/cartesia-transcription-model.test.ts
@@ -349,6 +349,30 @@ describe('doStream', () => {
     });
   });
 
+  it.each(['pcm_f16le', 'pcm_f32le', 'pcm_s32le'] as const)(
+    'supports the %s streaming PCM encoding',
+    async encoding => {
+      const provider = createCartesia({
+        apiKey: 'test-api-key',
+        webSocket: MockWebSocket,
+      });
+      const result = await provider.transcription('ink-2').doStream!({
+        audio: convertArrayToReadableStream([]),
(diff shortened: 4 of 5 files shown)


gr2m (approved):

shaper:
@31Carlton7 question on the `??` in `buildCartesiaStreamingTranscriptionUrl` — you've looked at this more closely than I have, so I may be missing intent.

```ts
url.searchParams.set(
  'encoding',
  providerOptions?.streaming?.encoding ??
    cartesiaEncodingFromInputAudioFormat(inputAudioFormat.type),
);
```

Because `??` short-circuits, `cartesiaEncodingFromInputAudioFormat` isn't called at all when `streaming.encoding` is set — so setting the option also turns off the `inputAudioFormat.type` validation that function was doing. Two cases that fall out:

1. `inputAudioFormat: { type: 'audio/mpeg' }` + `encoding: 'pcm_s16le'` previously threw `InvalidArgumentError`, and now proceeds silently.
2. `inputAudioFormat: { type: 'audio/pcmu' }` + `encoding: 'pcm_f32le'` sends f32le with no signal that the two contradict.

Is the escape-hatch behavior in (1) intentional? I can see an argument for it — the option is more specific than the inferred default, so letting it bypass a media-type allowlist we don't fully control is reasonable. If so, no objection, though a warning on the contradiction in (2) might be worth it since the `warnings` array is already in scope in `doStream`.

If it wasn't intentional, validating `inputAudioFormat.type` unconditionally and only using the option for the encoding value would keep the existing behavior intact.

Either way this is the only thing I'd flag. On the broader design question I went looking at whether the spec could express this at the top level instead, and it can't: `SharedV4AudioFormat` is just `{ type: string; rate?: number }`, with no bit-depth/int-vs-float axis, and there's no media type to lean on (`audio/pcm` isn't IANA-registered, and float PCM has no registered type). The provider option looks like the right call, and it matches what xai already does with `providerOptions.xai.audioFormat`.

31Carlton7:
> @31Carlton7 question on the `??` in `buildCartesiaStreamingTranscriptionUrl` — you've looked at this more closely than I have, so I may be missing intent.
> 
> ```ts
> url.searchParams.set(
>   'encoding',
>   providerOptions?.streaming?.encoding ??
>     cartesiaEncodingFromInputAudioFormat(inputAudioFormat.type),
> );
> ```
> 
> Because `??` short-circuits, `cartesiaEncodingFromInputAudioFormat` isn't called at all when `streaming.encoding` is set — so setting the option also turns off the `inputAudioFormat.type` validation that function was doing. Two cases that fall out:
> 
> 1. `inputAudioFormat: { type: 'audio/mpeg' }` + `encoding: 'pcm_s16le'` previously threw `InvalidArgumentError`, and now proceeds silently.
> 2. `inputAudioFormat: { type: 'audio/pcmu' }` + `encoding: 'pcm_f32le'` sends f32le with no signal that the two contradict.
> 
> Is the escape-hatch behavior in (1) intentional? I can see an argument for it — the option is more specific than the inferred default, so letting it bypass a media-type allowlist we don't fully control is reasonable. If so, no objection, though a warning on the contradiction in (2) might be worth it since the `warnings` array is already in scope in `doStream`.
> 
> If it wasn't intentional, validating `inputAudioFormat.type` unconditionally and only using the option for the encoding value would keep the existing behavior intact.
> 
> Either way this is the only thing I'd flag. On the broader design question I went looking at whether the spec could express this at the top level instead, and it can't: `SharedV4AudioFormat` is just `{ type: string; rate?: number }`, with no bit-depth/int-vs-float axis, and there's no media type to lean on (`audio/pcm` isn't IANA-registered, and float PCM has no registered type). The provider option looks like the right call, and it matches what xai already does with `providerOptions.xai.audioFormat`.

So do you think it'd be best to get rid of that boolean and only keep `providerOptions?.streaming?.encoding` ? I can also check and see what the odds of it coming back as null are and have a more basic fail safe for it too?

shaper:
Thanks — this is a better fix than either option I'd have suggested. Unconditional `inferredEncoding` restores the allowlist, and the warning condition is right across the matrix (I traced `audio/pcm` + all 4 linear → silent, `audio/pcm` + G.711 → warns, `audio/pcmu`/`audio/pcma` + anything different → warns, `audio/mpeg` → throws first). Nice side effect too: validation now runs before `createStreamingAccessToken`, so a bad media type fails without burning an access-token round trip first.

Two minor notes, neither blocking.

**1. `LINEAR_PCM_STREAMING_ENCODINGS` can be derived from the zod enum.** Right now it duplicates a subset of the enum in `cartesia-transcription-model-options.ts`, so a future linear encoding added to the enum would spuriously warn on `audio/pcm` until someone remembers to update the set, and nothing guards that.

You can't derive "linear" from the enum directly since the enum is the superset — but inverting works, and is arguably better:

```ts
// cartesia-transcription-model-options.ts
export const cartesiaStreamingEncodingSchema = z.enum([
  'pcm_alaw',
  'pcm_f16le',
  'pcm_f32le',
  'pcm_mulaw',
  'pcm_s16le',
  'pcm_s32le',
]);
```

```ts
// cartesia-transcription-model.ts
/** G.711 companding laws — the only non-linear members of the encoding enum. */
const G711_STREAMING_ENCODINGS: ReadonlySet<string> = new Set([
  'pcm_mulaw',
  'pcm_alaw',
]);

/** Linear PCM encodings that legitimately widen generic `audio/pcm` input. */
const LINEAR_PCM_STREAMING_ENCODINGS: ReadonlySet<string> = new Set(
  cartesiaStreamingEncodingSchema.options.filter(
    encoding => !G711_STREAMING_ENCODINGS.has(encoding),
  ),
);
```

The hardcoded half is then G.711, which is a genuinely closed set — the standard defines exactly two companding laws, so it will never grow. The open-ended half (linear PCM) is the half that derives. That also means any encoding added to the enum later is treated as linear-widening-compatible by default, which is the right default for a raw-PCM `encoding` param. Verified `.options` behaves as expected on zod 4 and the filter yields exactly the four current values.

**2. Missing a test for the `audio/pcm` + G.711 branch.** The suite covers three of the four outcomes but not `inputAudioFormat: { type: 'audio/pcm' }` + `encoding: 'pcm_mulaw'` → warns. That's the branch that exercises the `!(inferredEncoding === 'pcm_s16le' && ...)` half of the condition, i.e. the asymmetry that makes generic `audio/pcm` special. Worth one more `it` so a future refactor of that condition can't quietly widen it into "any override from `audio/pcm` is fine".

31Carlton7:
> Thanks — this is a better fix than either option I'd have suggested. Unconditional `inferredEncoding` restores the allowlist, and the warning condition is right across the matrix (I traced `audio/pcm` + all 4 linear → silent, `audio/pcm` + G.711 → warns, `audio/pcmu`/`audio/pcma` + anything different → warns, `audio/mpeg` → throws first). Nice side effect too: validation now runs before `createStreamingAccessToken`, so a bad media type fails without burning an access-token round trip first.
> 
> Two minor notes, neither blocking.
> 
> **1. `LINEAR_PCM_STREAMING_ENCODINGS` can be derived from the zod enum.** Right now it duplicates a subset of the enum in `cartesia-transcription-model-options.ts`, so a future linear encoding added to the enum would spuriously warn on `audio/pcm` until someone remembers to update the set, and nothing guards that.
> 
> You can't derive "linear" from the enum directly since the enum is the superset — but inverting works, and is arguably better:
> 
> ```ts
> // cartesia-transcription-model-options.ts
> export const cartesiaStreamingEncodingSchema = z.enum([
>   'pcm_alaw',
>   'pcm_f16le',
>   'pcm_f32le',
>   'pcm_mulaw',
>   'pcm_s16le',
>   'pcm_s32le',
> ]);
> ```
> 
> ```ts
> // cartesia-transcription-model.ts
> /** G.711 companding laws — the only non-linear members of the encoding enum. */
> const G711_STREAMING_ENCODINGS: ReadonlySet<string> = new Set([
>   'pcm_mulaw',
>   'pcm_alaw',
> ]);
> 
> /** Linear PCM encodings that legitimately widen generic `audio/pcm` input. */
> const LINEAR_PCM_STREAMING_ENCODINGS: ReadonlySet<string> = new Set(
>   cartesiaStreamingEncodingSchema.options.filter(
>     encoding => !G711_STREAMING_ENCODINGS.has(encoding),
>   ),
> );
> ```
> 
> The hardcoded half is then G.711, which is a genuinely closed set — the standard defines exactly two companding laws, so it will never grow. The open-ended half (linear PCM) is the half that derives. That also means any encoding added to the enum later is treated as linear-widening-compatible by default, which is the right default for a raw-PCM `encoding` param. Verified `.options` behaves as expected on zod 4 and the filter yields exactly the four current values.
> 
> **2. Missing a test for the `audio/pcm` + G.711 branch.** The suite covers three of the four outcomes but not `inputAudioFormat: { type: 'audio/pcm' }` + `encoding: 'pcm_mulaw'` → warns. That's the branch that exercises the `!(inferredEncoding === 'pcm_s16le' && ...)` half of the condition, i.e. the asymmetry that makes generic `audio/pcm` special. Worth one more `it` so a future refactor of that condition can't quietly widen it into "any override from `audio/pcm` is fine".

Both done!

shaper (approved):

PR #17925, opened 2026-07-24
feat (provider/perplexity): add embedding model support

joshualipman123:
## Background

The `@ai-sdk/perplexity` provider only implemented a language model — both `embeddingModel` and the deprecated `textEmbeddingModel` threw `NoSuchModelError`. Perplexity now offers a native embeddings API (`POST /v1/embeddings`, OpenAI-compatible), so this adds a real embedding model.

## Summary

Adds `perplexity.embedding(...)` (aliased as `embeddingModel` / `textEmbeddingModel`) so Perplexity works with `embed` / `embedMany`. Implemented as an `EmbeddingModelV4`.

- New models: `pplx-embed-v1-0.6b`, `pplx-embed-v1-4b`, `pplx-embed-context-v1-0.6b`, `pplx-embed-context-v1-4b` (max 512 inputs/call).
- Provider options: `dimensions` (Matryoshka truncation) and `encodingFormat` (`base64_int8` default, or `base64_binary`).

## Changes

- `perplexity-embedding-options.ts` — model IDs + provider-options schema
- `perplexity-embedding-model.ts` — `EmbeddingModelV4` implementation + base64 decoding
- `perplexity-provider.ts` — wire up `embedding` / `embeddingModel` / `textEmbeddingModel`
- `index.ts` — export `PerplexityEmbeddingModelId`, `PerplexityEmbeddingModelOptions`
- Tests, `embed` / `embed-many` examples, provider docs, changeset


packages/perplexity/src/perplexity-embedding-model-options.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from 'zod/v4';
+
+// https://docs.perplexity.ai/docs/embeddings/quickstart
+export type PerplexityEmbeddingModelId =
+  | 'pplx-embed-v1-0.6b'
+  | 'pplx-embed-v1-4b'
+  | 'pplx-embed-context-v1-0.6b'
+  | 'pplx-embed-context-v1-4b'
+  | (string & {});
```
R-Taneja, line 7: apparently:

> pplx-embed-context-v1-0.6b and pplx-embed-context-v1-4b are served from a different endpoint with a different input shape:
> 
> - Standard: POST /v1/embeddings, input: string[]
> - Contextual: POST /v1/contextualizedembeddings, input: string[][] (one inner array per document, chunks in original order)
> 
> The /v1/embeddings OpenAPI model enum contains only pplx-embed-v1-0.6b and pplx-embed-v1-4b. So perplexity.embedding('pplx-embed-context-v1-4b') posts to the wrong endpoint with the wrong body shape and will 422, yet it is offered by autocomplete in PerplexityEmbeddingModelId and listed as supported in the docs capabilities table.

let's test it to verify the claim

R-Taneja:
> The response carries a cost breakdown (input_cost, total_cost, currency) that is dropped. Could go in providerMetadata

we should carry this through

packages/perplexity/src/perplexity-embedding-model-options.ts
joshualipman123: this is right context embed models wont work through .embed or .embedMany so going to drop for now

R-Taneja (approved):
we should backport

PR #17973, opened 2026-07-27
fix: prevent the warning system banner from corrupting stdout output

### LINKED ISSUE #17957: AI SDK Warning System banner logs to stdout instead of stderr
*Filed with AI assistance (Claude/Claude Code) on the reporter's behalf — content reviewed and approved by the reporter before submission.*

## Bug

`logWarnings` (in the built `dist/index.js`, `ai@7.0.34`) prints its one-time banner via `console.info`, which writes to **stdout**. Every individual warning in the same function is instead routed through `process.emitWarning` when available, which Node always sends to **stderr**:

```js
if (!hasLoggedBefore) {
  hasLoggedBefore = true;
  console.info(FIRST_WARNING_INFO_MESSAGE); // stdout — inconsistent with the branch below
}
for (const warning of options.warnings) {
  const message = formatWarning({ warning, provider: options.provider, model: options.model });
  if (typeof process !== "undefined" && typeof process.emitWarning === "function") {
    process.emitWarning(message, { type: ... }); // stderr
  } else {
    console.warn(message);
  }
}
```

The banner isn't gated behind the same `process.emitWarning` check as the messages it introduces, so in Node it always lands on stdout regardless of environment.

## Why this matters in practice

Any program that treats its own stdout as real output — piped to another process, redirected to a file, parsed as JSON/NDJSON by a caller — gets that output corrupted the first time a run logs any SDK warning (deprecated option, unsupported provider feature, etc.). Because `hasLoggedBefore` is a module-level flag, this only fires on a process's *first* warning, which makes it intermittent and easy to miss in testing but reliable to hit in production once any warning path is exercised.

Concretely: we hit this in [jazz](https://github.com/lvndry/jazz), a CLI agent whose `--json` mode contract is "stdout is JSON only, consumed by another process." A run that called a tool and triggered a provider warning produced a fully valid, complete answer — but the banner line ahead of the JSON made the consumer's `JSON.parse` fail at character 0. It looked exactly like the run's output had been truncated, which sent debugging in the wrong direction until we reproduced it directly and inspected raw stdout.

## Suggested fix

Route the banner through the same `process.emitWarning`-or-`console.warn` branch as the per-warning messages, instead of an unconditional `console.info`. Happy to open a PR if that's the preferred fix.

lgrammel:
## Background

The published ai package could prepend its warning banner to stdout, corrupting JSON and other machine-readable CLI output.

## Root Cause

The one-time banner bypassed the existing warning sink and called console.info directly, while individual warnings used process.emitWarning or console.warn. The reproduction confirmed one banner on stdout and warning details on stderr.

## Summary

Routed the banner and individual warnings through a shared process.emitWarning-or-console.warn path, added a patch changeset, and removed reproduction-only artifacts.

## Testing

Updated logger regression coverage to verify no console.info output, Node warning emission, one-time banner behavior, and console.warn fallback behavior.

## End-to-end Validation

- `pnpm -C examples/ai-functions exec tsx src/reproduction/issue-17957-warning-banner-stdout.ts` — passed after rebuilding ai; child stdout contained valid JSON only and warning diagnostics were on stderr.

## Related Issues

Fixes #17957

Closes #17963

Co-Authored-By: Lars Grammel <205036+lgrammel@users.noreply.github.com>


lgrammel:
## Bugfix review

Outcome: approved

### Reproduction replay

Status: no-longer-reproduces

### Fixes issue

Status: fully-addresses

The banner now uses the same process.emitWarning-or-console.warn sink as individual warnings, preserving JSON-only stdout while retaining one-time banner behavior.

### Side effects

Risk: low

The only meaningful behavior change is the intended diagnostic routing; suppression, custom logger, empty-warning, warning formatting, and one-time logging behavior remain intact.

### Performance

Risk: low

The change adds only one process warning emission on the first warning-producing call and introduces no accumulating state, loops, or substantial allocations.

### Backwards compatibility

Risk: none

The change does not read, write, migrate, or reinterpret stored data and does not alter public types or function signatures.

### Architecture

Risk: none

The warning sink remains a private helper within the owning logger module, adds no dependencies, crosses no package boundaries, and does not expose context-specific code as shared infrastructure.

### Change scope

Status: minimal

The net diff contains only the localized logger fix, focused regression-test updates, and the required patch changeset.

### Security

Risk: none

The change only redirects a fixed informational diagnostic and existing warning messages; it introduces no new input handling, network access, credential exposure, or execution path.

### Testing

Status: appropriate

Regression coverage verifies absence of console.info output, Node warning emission and ordering, one-time banner behavior, console.warn fallback, and existing suppression and custom-logger behavior; the end-to-end replay also passed.

### Verification

Reviewed every net diff hunk and package boundaries; the exact replay passed, all 17 focused logger tests passed in both Node and edge configurations, package type-checking passed, formatting and lint checks passed, and the diff had no whitespace errors. Official Node documentation confirms the process.emitWarning options used and the stdout/stderr behavior of console.info and console.warn. 

### Relevant Documentation

- https://nodejs.org/api/process.html#processemitwarningwarning-options
- https://nodejs.org/api/console.html#consolewarn-data

PR #17980, opened 2026-07-27
fix: prevent large single-consumer text streams from exhausting memory

### LINKED ISSUE #16753: streamText textStream tee path retains large streaming outputs in memory

aayush-kapoor:
## Background

Large textStream outputs exhausted constrained heaps because the public stream path retained generated content even for callers needing each delta only once.

## Root Cause

The unread ReadableStream.tee() branch and text accumulation in the output transform, model-call content, and final-result event processor kept memory proportional to total output; the 96 MiB reproduction passed with direct provider consumption but failed through streamText under an 80 MiB heap.

## Summary

Added experimental_streamMode: 'single-consumer', which directly assigns one stream consumer and disables text/reasoning materialization and final lifecycle callbacks for low-retention stream access while preserving normal final results when a result getter consumes first. Added documentation and a patch changeset.

## Testing

Added five regression tests covering low-retention text streaming, unavailable final results and callbacks, result-first consumption, duplicate consumers, and unsupported tool configuration.

## End-to-end Validation

- `pnpm -C packages/ai build && pnpm -C examples/ai-functions exec tsx src/reproduction/issue-16753-stream-text-retention.ts` — consumed the full 96 MiB SDK text stream under an 80 MiB old-space limit without heap exhaustion.

## Related Issues

Fixes #16753

Closes #17975

Co-Authored-By: Aayush Kapoor <83492835+aayush-kapoor@users.noreply.github.com>


aayush-kapoor:
## Bugfix review

Outcome: changes-required

### Reproduction replay

Status: still-reproduces

The exact original bug signal appeared when replayed against the pull request.

Concerns:
- The original reported bug remains reproducible after the fix.

### Fixes issue

Status: does-not-address

The pull request adds an opt-in direct-consumer path, but the unchanged streamText().textStream usage from the issue still follows the retaining default path and reproduced the bug.

Concerns:
- Existing callers must discover and adopt the new option; the reported usage remains vulnerable by default.
- The pull request's end-to-end claim conflicts with the authoritative replay result.
- The exact original reproduction still produces the reported bug.

### Side effects

Risk: medium

Default behavior is mostly preserved, but the opt-in path creates asymmetric lifecycle behavior by invoking start callbacks while suppressing corresponding model-call, step, and operation end callbacks.

Concerns:
- Telemetry integrations can receive start events without matching end events, potentially leaving incomplete traces or integration state.
- The option rejects all stream access whenever tools are configured, including configurations that may not produce a tool call.

### Performance

Risk: high

The opt-in text/part path avoids teeing and several accumulators, but the original path still exhibits the reported memory failure and not every accessor covered by the mode is low-retention.

Concerns:
- The default textStream path remains proportional to total generated output.
- partialOutputStream and elementStream claim the single consumer but the output transform only disables accumulation for the parts owner, so those accessors can still retain the complete text.

### Backwards compatibility

Risk: none

The changes introduce no stored-data format, migration, persistence, or serialization changes.

### Architecture

Risk: medium

Package boundaries remain intact, but downstream result-consumer state is coupled into language-model stream normalization through a closure that controls both content collection and lifecycle callback delivery.

Concerns:
- collectModelCallContent conflates memory policy with whether onLanguageModelCallEnd is emitted.
- Consumer selection in DefaultStreamTextResult now changes behavior inside streamLanguageModelCall, creating temporal coupling between projection access and model-call lifecycle semantics.
- No cross-package src imports or dependency-direction violations were found.

### Change scope

Status: minimal

The six changed files are confined to the streamText implementation, its result contract, regression tests, documentation, and the required patch changeset.

### Security

Risk: none

The change adds no credential handling, external input parsing, network access, or authorization logic.

### Testing

Status: needs-more

The added functional tests pass in Node and Edge configurations, but they do not verify the defining large-output memory behavior and the authoritative reproduction still fails.

Concerns:
- There is no committed regression test proving bounded memory for a large single-consumer text stream.
- Tests do not cover partialOutputStream or elementStream under the new mode.
- Tests assert that end callbacks are skipped but do not cover the resulting unmatched start-callback and telemetry lifecycle behavior.

### Verification

Inspected every merge-base diff hunk and the stream, output-transform, callback, telemetry, and result-consumption paths. The streamText tests passed in Node and Edge configurations, package type checking passed, and changed-file formatting passed. These checks do not override the authoritative replay, which still reproduced the original bug.

PR #17989, opened 2026-07-27
fix: destroy session on stream cancellation and client disconnect

gr2m:
## Background

Follow-up to #17869. The session's success-path `destroy()` runs in the event processor's `flush` callback, which only runs when the stream pipeline closes normally. When the pipeline terminates abnormally, session resources (sockets, intervals, handles) leaked:

- when the readable side is cancelled or the writable side is aborted (e.g. a stream transform throws), `flush` never runs
- `writeToServerResponse` ignored client disconnects entirely: it kept reading the stream and writing to a dead socket, and could await a `drain` event that never fires

## Summary

- **`streamText`**: destroy the session from the event processor's `cancel` transformer callback (runs on readable-side cancellation and writable-side abort, exactly the cases `flush` misses) and from the source stream's `cancel` handler. Since `session.destroy()` memoizes its rejection, cleanup failures observed by multiple termination paths are reported only once.
- **`writeToServerResponse`**: detect premature `close` (distinguished from regular completion via `writableFinished`), cancel the stream reader so upstream resources can be released, stop writing further chunks, and stop waiting for `drain` on a disconnected response.
- **Test**: lock in that aborting the `abortSignal` destroys the session even when the stream is never consumed — the step-driving loop is promise-based, so this works independently of stream consumption. This makes `abortSignal: request.signal` the reliable pattern for client-disconnect cleanup.

## Verification

- `transformer.cancel` is supported in Node ≥ 20.4 (repo floor is Node 22) and the edge runtime; both test suites pass. The TypeScript `Transformer` type doesn't include `cancel` yet (TS 5.8), hence the extended type annotation.
- all added tests fail without code changes (confirmed locally)

### DIFF
--- a/packages/ai/src/generate-text/stream-text.ts
+++ b/packages/ai/src/generate-text/stream-text.ts
@@ -1168,10 +1168,40 @@ class DefaultStreamTextResult<
     let recordedNoOutputError: NoOutputGeneratedError | undefined;
     let setupCleanupError: AggregateError | undefined;
 
-    const eventProcessor = new TransformStream<
+    // destroys the session when the pipeline terminates without a regular
+    // flush (cancellation or upstream abort). The consumer is gone at this
+    // point, so cleanup failures are reported instead of thrown:
+    let sessionCleanupErrorHandled = false;
+    const destroySessionOnTermination = async () => {
+      try {
+        await session.destroy();
+      } catch (cleanupError) {
+        // session.destroy() memoizes its rejection, so multiple termination
+        // paths can observe the same failure; only handle it once:
+        if (sessionCleanupErrorHandled) {
+          return;
+        }
+        sessionCleanupErrorHandled = true;
+
+        const error = setupCleanupError ?? cleanupError;
+
+        self.rejectResultPromises(error);
+
+        // don't report the same error twice
+        if (setupCleanupError == null) {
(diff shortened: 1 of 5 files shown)


packages/ai/src/generate-text/stream-text.ts
```diff
@@ -1168,10 +1168,40 @@ class DefaultStreamTextResult<
...
+    };
+
+    const eventProcessorTransformer: Transformer<
       EnrichedStreamPart<TOOLS, InferPartialOutput<OUTPUT>>,
       EnrichedStreamPart<TOOLS, InferPartialOutput<OUTPUT>>
-    >({
+    > & {
+      // TODO: remove once the TypeScript Transformer type includes cancel
+      cancel?: (reason: unknown) => void | PromiseLike<void>;
+    } = {
```
gr2m, line 1204: there seems to be a difference between the Streams spec and TypeScript's bundled DOM type definitions. TS 7.0.2 still doesn't have it. But `@types/node` does have it so 🤷🏼 I'll update the comment since I don't expect it to be added in future either

```suggestion
    > & {
      // TypeScript does not define `.cancel()` from the Stream spec as of 7.0.2. But `@types/node` does.
      cancel?: (reason: unknown) => void | PromiseLike<void>;
    } = {
```

PR #18037, opened 2026-07-28
fix(harness-pi): preserve tool results across process resume

### LINKED ISSUE #18018: harness-pi: tool approval submitted after cross-process resume is silently dropped — rerun feeds the model an empty tool result
## Description

When a `HarnessAgent` turn on the Pi adapter pauses on a tool approval (`toolApproval: { myTool: 'user-approval' }`) and the session is later resumed **in a different process**, the approval response is silently dropped: the framework delivers it to the adapter, but the adapter's `submitToolResult` is a no-op without a live in-process turn, and the rerun-from-journal path feeds the model an **empty tool result** instead. The turn then completes "successfully" with the model visibly confused about never receiving the result.

Same-process resume (the parked-session path) works correctly. The failure only appears when the process that paused the turn is not the process that resumes it — which is the common case for any horizontally scaled or rolling-deployed host, i.e. exactly the situation `detach()`/`resumeFrom` is designed for.

## Environment

- `@ai-sdk/harness` / `@ai-sdk/harness-pi` 1.0.43 (verified unchanged in 1.0.44 — dependency-bump only)
- `ai` 7.0.37

## Reproduction (two processes)

1. Create a Pi-backed `HarnessAgent` with a host tool gated by `toolApproval: { askUser: 'user-approval' }`, using a durable sandbox provider.
2. Start a turn with a prompt that makes the model call the tool. The turn pauses with an `approval-requested` part.
3. `await session.detach()` and persist the full returned resume state (it correctly embeds the continuation and pending approval).
4. **Exit the process.**
5. In a fresh process: `agent.createSession({ resumeFrom })`, then `agent.continueStream({ session, toolApprovalContinuations: [{ approvalResponse: { approvalId, approved: true, reason }, toolCall }] })`.
6. Observe the continuation.

## Observed behavior

- The host tool's answer resolver **is** invoked in the new process with the original `toolCallId` (framework-side delivery works end to end), and the harness calls `control.submitToolResult(...)` with the result.
- The result never reaches the model. The rerun continues from the journal, the pending tool call resolves as empty, and the model's continuation text was, verbatim: *"It looks like the question didn't get an answer — no selection came through."* — followed by it re-asking the same question in plain text.
- No error is surfaced anywhere. The stream and turn complete normally.

## Traced mechanism (published 1.0.43 dist / repo source)

- Framework side is correct: pending approvals are persisted framework-owned in the lifecycle state, recovered on resume, and replayed via `processPendingApprovalContinuation` → `control.submitToolResult` (`packages/harness/src/agent/internal/run-prompt.ts`).
- `packages/harness-pi/src/pi-session.ts`: the live session is parked in a **module-level Map** (`parkedPiSessions`) on suspend/detach; the durable resume payload carries only `{ sessionFileName }`.
- In a new process the Map is empty, so `createPiSession` rebuilds from the session file, and both submit handlers silently no-op:

  ```ts
  async submitToolResult(args) {
    const pending = pendingToolResults.get(args.toolCallId);
    if (!pending) return;   // <- approval response dropped here
  ```

- `doContinueTurn` then falls back to `runTurn({ text: '' })` (rerun-from-journal), in which the dangling tool call resolves as an empty result. The in-source comments acknowledge the rerun is lossy for in-flight work, but a delivered tool-approval response is not "in-flight work being recomputed" — it is caller-supplied data the framework handed the adapter moments earlier and the adapter discarded.

## Docs contradiction

The published docs state, unconditionally for all adapters:

> "session.suspendTurn() includes the pending tool call in its serializable continuation state." — /docs/ai-sdk-harnesses/tools

> "To continue in another process, call suspendTurn(), recreate the session with continueFrom, and then pass the result to continueStream() or continueGenerate()."

For Pi, neither holds: the adapter payload carries only the session filename, and continuing in another process loses the submitted result. The adapter capability table's "Runtime location: Host process" column is the only hint, and it doesn't explain the durability consequence.

## This also affects @ai-sdk/workflow-harness by construction

`runHarnessAgentStep` returns `awaiting_tool_approval` and the workflow run terminates (the consumer loop only continues on `ready_for_next_step`), so the approval necessarily arrives via a new request → a new `start()` → a fresh step invocation. Workflow DevKit steps are documented as stateless, isolated API routes, so the resuming step is by design not the process that parked the Pi session. The approval then flows through the same `processPendingApprovalContinuation` → `submitToolResult` path and is dropped — while the host tool itself still executes on the new host, so side effects happen and the output is discarded. This contradicts the V1 contract's own wording that the adapter "forwards the result to the underlying runtime so the model can continue" (`harness-v1-prompt-control.ts`).

Because Fluid Compute sometimes reuses a warm instance, the parked-session Map will occasionally still be populated, making the failure intermittent rather than deterministic.

Test coverage reflects the gap: the only Pi test of this path is named "parks a pending tool turn on suspend and resumes it **in-process**", the workflow-harness approval tests use a mock agent with no real adapter, and no example exercises an approval across a step boundary.

## Suggested direction

The framework already hands the adapter everything it needs. Two adapter-local options, either of which would fix this without changing the persisted state shape or the public API:

1. In `submitToolResult`/`submitToolApproval`, when there is no live pending entry but the toolCallId matches a journal-pending call, stash the result and inject it during the subsequent rerun instead of synthesizing an empty result. Note the rerun mints fresh tool-call ids while the framework replays the original id, so this needs an id remapping keyed on the journal-pending call, not just delivery.
2. Alternatively, surface an explicit error from `continueStream` when a continuation cannot be delivered, instead of completing the turn silently — so callers can at least detect and retry/re-ask deliberately.

At minimum, the docs should caveat that mid-turn cross-process continuation is lossless only for bridge-backed adapters.

## Related

- #17083 — the in-process sibling on the same continuation path (approval `reason` for an approved host tool never reaches `execute()`), which forces host tools to side-channel the approval payload to begin with.

longligooo:
## Summary

Fixes #18018

Pi runs in the host process, so a tool approval continuation can arrive after the session has been recreated from its persisted journal. On a cold resume, the adapter previously kept pending tool results only in an in-process map; `submitToolResult` therefore became a silent no-op and the rerun continued with an empty result.

This change:

- Recovers unresolved Pi tool calls from the persisted session branch during a cold resume.
- Correlates the persisted harness tool-call ID with the fresh runtime ID using the tool name and arguments.
- Preserves the original ID in translated stream parts so the harness does not execute an already-settled host tool twice.
- Delivers tool results and built-in approval decisions whether they arrive before or after the rerun invokes the tool.
- Reports an explicit error for submissions that cannot be correlated instead of silently dropping them.

## Tests

- `corepack pnpm --filter @ai-sdk/harness-pi type-check`
- `vitest --config vitest.node.config.js --run src/pi-session.test.ts -t "cold cross-process"`
- `vitest --config vitest.node.config.js --run src/pi-translate.test.ts -t "persisted id"`
- `oxfmt --check` on all changed TypeScript files
- `oxlint` on all changed TypeScript files

The full package test suite also ran. Twelve test files passed; one existing Windows-specific assertion (`uses agentDir for auth, models, and settings when provided`) expects POSIX separators and fails on Windows before exercising this change.

## Release Notes

A patch changeset for `@ai-sdk/harness-pi` is included.

### DIFF
--- a/.changeset/cold-pi-tool-results.md
+++ b/.changeset/cold-pi-tool-results.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-pi': patch
+---
+
+Preserve host tool results when a Pi harness turn resumes in a different process.
--- a/packages/harness-pi/src/pi-session.test.ts
+++ b/packages/harness-pi/src/pi-session.test.ts
@@ -18,6 +18,7 @@ const piMock = vi.hoisted(() => {
     createAgentSession: vi.fn(),
     customTools: [] as FakePiTool[],
     session: undefined as AgentSession | undefined,
+    sessionEntries: [] as unknown[],
   };
 });
 
@@ -40,9 +41,11 @@ vi.mock('@earendil-works/pi-coding-agent', () => {
     SessionManager: {
       create: vi.fn(() => ({
         getSessionFile: () => undefined,
+        getBranch: () => piMock.sessionEntries,
       })),
       open: vi.fn(() => ({
         getSessionFile: () => undefined,
+        getBranch: () => piMock.sessionEntries,
       })),
     },
     SettingsManager: {
@@ -56,6 +59,7 @@ describe('createPiSession', () => {
   beforeEach(() => {
     piMock.customTools = [];
     piMock.session = undefined;
+    piMock.sessionEntries = [];
     piMock.createAgentSession.mockReset();
     piMock.createAgentSession.mockImplementation(async options => {
       piMock.customTools = options.customTools;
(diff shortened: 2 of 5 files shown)


felixarntz:
@longligooo Thank you for the PR! Closing this in favor of #18052 which is a bit cleaner as a solution.

PR #18068, opened 2026-07-28
fix(prodia): warn on unsupported seed call setting

### LINKED ISSUE #17938: @ai-sdk/prodia: seed is the only unsupported call setting that does not emit a warning
## Description

`ProdiaLanguageModel.doGenerate()` warns about every unsupported call setting except `seed`.

The block at `packages/prodia/src/prodia-language-model.ts:67-108` pushes an `{ type: 'unsupported', feature: ... }` warning for `temperature`, `topP`, `topK`, `maxOutputTokens`, `stopSequences`, `presencePenalty`, `frequencyPenalty`, `tools`, `toolChoice`, `responseFormat` and `reasoning`. `seed` is the only standard `LanguageModelV4CallOptions` field missing from that list, and it isn't forwarded to the Prodia API either — so it's silently dropped.

`doStream` re-emits `doGenerate`'s warnings verbatim (`:296-304`), so both paths behave the same way.

Given how exhaustive the surrounding list is, this looks like an omission rather than a deliberate choice:

```ts
import { createProdia } from '@ai-sdk/prodia';

const prodia = createProdia({ apiKey: process.env.PRODIA_API_KEY! });

  model: prodia('inference.nano-banana.img2img.v2'),
  prompt: 'hello',
  seed: 42,
  topK: 10, // for contrast — this one does warn
});

console.log(result.warnings);
// [{ type: 'unsupported', feature: 'topK' }]
// expected to also contain { type: 'unsupported', feature: 'seed' }
```

**Expected:** `warnings` contains `{ type: 'unsupported', feature: 'seed' }` when `seed` is set.

**Actual:** no warning, and `seed` never reaches the request.

Small enough that I'm happy to send the three-line fix plus a test assertion in `prodia-language-model.test.ts` if you'd like.

## Reproduction

Reproduced by reading `packages/prodia/src/prodia-language-model.ts:67-108` on `main` at `eb16508fc7`: `options.seed` is not referenced anywhere in the file, while all sibling call settings are checked and warned about a few lines apart. The snippet above shows the observable difference through `generateText`.

## AI SDK Version

- `@ai-sdk/prodia`: 2.0.13
- `ai`: 7.0.37
- verified against `main` at `eb16508fc7`

nvrakesh06:
## Summary
- `ProdiaLanguageModel.doGenerate()` warns about every unsupported call setting (`temperature`, `topP`, `topK`, `maxOutputTokens`, `stopSequences`, `presencePenalty`, `frequencyPenalty`, `tools`, `toolChoice`, `responseFormat`, `reasoning`) except `seed`, which was silently dropped.
- Adds the missing `{ type: 'unsupported', feature: 'seed' }` warning, matching the pattern used for the sibling settings. `doStream` re-emits `doGenerate`'s warnings, so both paths are fixed together.

Fixes #17938

## Test plan
- [x] Added a `seed: 42` case and matching assertion to the existing "emits warnings for unsupported LLM features" test in `prodia-language-model.test.ts`
- [x] `pnpm test:node` and `pnpm test:edge` in `packages/prodia` pass (56/56 tests each)

### DIFF
--- a/.changeset/lucky-otters-warn.md
+++ b/.changeset/lucky-otters-warn.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/prodia': patch
+---
+
+Warn when the Prodia language model ignores the unsupported `seed` call setting.
--- a/packages/prodia/src/prodia-language-model.test.ts
+++ b/packages/prodia/src/prodia-language-model.test.ts
@@ -386,6 +386,7 @@ describe('ProdiaLanguageModel', () => {
         stopSequences: ['stop'],
         presencePenalty: 0.1,
         frequencyPenalty: 0.2,
+        seed: 42,
         tools: [
           {
             type: 'function',
@@ -409,6 +410,7 @@ describe('ProdiaLanguageModel', () => {
       expect(warningFeatures).toContain('stopSequences');
       expect(warningFeatures).toContain('presencePenalty');
       expect(warningFeatures).toContain('frequencyPenalty');
+      expect(warningFeatures).toContain('seed');
       expect(warningFeatures).toContain('tools');
       expect(warningFeatures).toContain('toolChoice');
       expect(warningFeatures).toContain('responseFormat');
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by #19487, which already closed #17938.

PR #18102, opened 2026-07-29
feat(harness-cline): add Cline harness adapter

saoudrizwan:
This adds `@ai-sdk/harness-cline` — a `HarnessV1` adapter that connects `HarnessAgent` to the [Cline SDK](https://docs.cline.bot/sdk/overview) agent runtime (`@cline/agents`). We (the Cline team) built and maintain it, and we're happy to keep maintaining it in-tree — treat us as the owning team for this package the way the Coming Soon adapters (Amp/Goose/Mastra) presumably have owning teams.

### Architecture: host-process, like Pi

Of the two adapter shapes in the codebase, Cline is a natural fit for the **host-process** model rather than a sandbox bridge. The Cline SDK's `Agent` runtime is a plain in-process TypeScript library — there's no CLI to install into the sandbox, no bridge bootstrap, no WebSocket relay, no pinned bridge lockfile. The adapter is ~1/3 the surface of the OpenCode adapter for that reason.

The split mirrors `harness-pi`:

- The **agent loop and model calls run on the host** via `new Agent({ providerId, modelId, ... })`.
- The **workspace lives entirely in the sandbox**: every built-in tool (`read`, `write`, `edit`, `bash`, `grep`, `glob`, `ls`) is implemented against the restricted `SandboxSession` surface (`readTextFile`/`writeTextFile`/`run`), with relative paths resolved against `sessionWorkDir` and `bash` executing there.

Unlike Pi, there's no host-side workspace mirror/VFS: the Cline runtime keeps its conversation in memory and has no fs-based resource discovery, so nothing on the host filesystem needs to impersonate the sandbox path. That deleted an entire subsystem (Pi's `pi-workspace-vfs` + `pi-workspace-mirror` + path mapper) from the design.

### How each harness feature maps onto the Cline runtime

- **Streaming**: `agent.subscribe()` events (`assistant-text-delta`, `assistant-reasoning-delta`, `assistant-message`, `tool-started`, `tool-finished`) are translated into harness stream parts in `cline-translate.ts`. The runtime streams flat deltas, so the translator owns text/reasoning block framing (start/delta/end) and emits a `finish-step` per model call using the assistant message's per-call `metrics`.
- **Built-in tool approvals** (`supportsBuiltinToolApprovals`): this uses the Cline runtime's *native* approval mechanism rather than gating inside tool `execute` functions. The harness `permissionMode` is compiled into runtime `toolPolicies` (`allow-edits` → `bash` needs approval; `allow-reads` → `write`/`edit`/`bash` need approval), and the runtime's `requestToolApproval` callback emits `tool-approval-request` and parks a promise resolved by `submitToolApproval`. One ordering subtlety: the Cline runtime asks for approval *before* it emits `tool-started`, but the harness contract implies a `tool-call` part exists before its approval request. The approval path therefore emits the `tool-call` part itself and the translator dedupes the later `tool-started` by `toolCallId` (there's a test asserting exactly this sequence).
- **Built-in tool filtering** (`supportsBuiltinToolFiltering`): native — filtered tools are simply never registered with the runtime, so the model never sees them.
- **Custom host-executed tools** (`HarnessV1ToolSpec`): each spec becomes a runtime tool whose `execute` parks a promise keyed by the runtime-provided `toolCallId`; `submitToolResult` resolves it. The execute also listens on the runtime's per-call abort signal so an aborted turn can't leave the loop dangling on a parked promise (the runtime awaits tool executions before it re-checks abort).
- **Mid-turn `submitUserMessage`**: wired to the runtime's `consumePendingUserMessage` hook — queued messages are injected between loop iterations, before the next model call.
- **Skills**: written into sandbox HOME (`~/.agents/skills/<name>/SKILL.md` via the shared `writeSkills` util) and advertised through a system-prompt section that tells the model to `read` a SKILL.md when relevant — progressive disclosure, since the standalone runtime has no native skill loader.
- **Lifecycle/resume**: conversation state is the runtime's message transcript. `doStop`/`doDetach` persist it as JSON into the sandbox workspace (`.cline-harness/history.json` under `sessionWorkDir`, path-containment-checked), so cross-process resume survives via the sandbox snapshot; `doStart({ resumeFrom })` restores it with `initialMessages`. Same-process detach during a live approval/tool wait parks the session in a module map (Pi's `parkedPiSessions` pattern) so the parked promise survives. `doSuspendTurn` is the lossy host-process variant: abort silently (a `suspending` flag suppresses the spurious `error`/`finish` parts), persist history, rerun-continue next slice via `agent.continue()` with no input.
- **`doCompact`**: throws `HarnessCapabilityUnsupportedError` — the standalone Cline runtime has no manual compaction trigger (Codex-over-exec precedent).

### Defaults & auth

`createCline()` defaults to `providerId: 'anthropic'`, `modelId: 'claude-opus-5'`; `apiKey` falls back to the provider's env var (e.g. `ANTHROPIC_API_KEY`) through the Cline gateway. `cline` is exported as the zero-config instance, matching `pi`/`openCode`.

### Testing

53 unit tests across the translator (block framing, approval ordering/dedupe, providerExecuted semantics), sandbox-backed ops (path resolution, edit-by-replacement, bash timeout/abort, glob filtering), resume-state round-trip + path-traversal rejection, and the harness surface. `pnpm build`, `pnpm type-check`, and `pnpm test` are green in `packages/harness-cline`.

### DIFF
--- a/packages/harness-cline/src/cline-harness.ts
+++ b/packages/harness-cline/src/cline-harness.ts
@@ -1,0 +1,181 @@
+import {
+  commonTool,
+  type HarnessV1,
+  type HarnessV1BuiltinTool,
+} from '@ai-sdk/harness';
+import { tool } from '@ai-sdk/provider-utils';
+import { z } from 'zod/v4';
+import { clineResumeStateSchema } from './cline-resume-state';
+import { createClineSession } from './cline-session';
+
+/**
+ * Configuration knobs for `createCline`. The Cline runtime
+ * (`@cline/agents`, part of the Cline SDK) runs as an in-process Node
+ * library — no bridge, so there is no `port` or `startupTimeoutMs` to set.
+ */
+export type ClineHarnessSettings = {
+  /**
+   * Cline LLM provider id (e.g. `'anthropic'`, `'openai'`, `'gemini'`).
+   * Defaults to `'anthropic'`.
+   */
+  readonly providerId?: string;
+  /**
+   * Model id for the configured provider. Defaults to `'claude-opus-5'`.
+   */
+  readonly modelId?: string;
+  /**
+   * Provider API key. When omitted, the Cline gateway falls back to the
+   * provider's environment variable (e.g. `ANTHROPIC_API_KEY`).
+   */
+  readonly apiKey?: string;
+  /** Custom provider endpoint. */
+  readonly baseUrl?: string;
+  /** Extra headers sent to the provider. */
+  readonly headers?: Record<string, string>;
+  /**
+   * Additional system prompt text appended to the adapter's base system
+   * prompt (workspace + sandbox + skills sections).
+   */
+  readonly systemPrompt?: string;
+  /**
+   * Safety cap on agent-loop iterations per turn. Unbounded when omitted.
+   */
+  readonly maxIterations?: number;
+};
+
+const DEFAULT_PROVIDER_ID = 'anthropic';
+const DEFAULT_MODEL_ID = 'claude-opus-5';
+
+/*
+ * The Cline runtime requires snake_case tool names, and each of these
+ * built-ins already matches its cross-harness common name — so native names
+ * and wire names are identical, and `nativeName` mirrors the key. These zod
+ * declarations mirror the JSON schemas handed to the runtime in
+ * `cline-tools.ts` — keep the two in sync.
+ */
+const CLINE_BUILTIN_TOOLS = {
+  read: commonTool('read', {
+    nativeName: 'read',
+    toolUseKind: 'readonly',
+    description: 'Read file contents.',
+    inputSchema: z.object({
+      file_path: z.string(),
+    }),
+  }),
+  write: commonTool('write', {
+    nativeName: 'write',
+    toolUseKind: 'edit',
+    description: 'Overwrite or create a file.',
+    inputSchema: z.object({
+      file_path: z.string(),
+      content: z.string(),
+    }),
+  }),
+  edit: commonTool('edit', {
+    nativeName: 'edit',
+    toolUseKind: 'edit',
+    description: 'Edit a file by exact string replacement.',
+    inputSchema: z.object({
+      file_path: z.string(),
+      old_string: z.string(),
+      new_string: z.string(),
+    }),
+  }),
+  bash: commonTool('bash', {
+    nativeName: 'bash',
+    toolUseKind: 'bash',
+    description: 'Execute a shell command in the sandbox.',
+    inputSchema: z.object({
+      command: z.string(),
+      timeout: z.number().optional(),
+    }),
+  }),
+  grep: commonTool('grep', {
+    nativeName: 'grep',
+    toolUseKind: 'readonly',
+    description: 'Search file contents with regex.',
+    inputSchema: z.object({
+      pattern: z.string(),
+      path: z.string().optional(),
+      glob: z.string().optional(),
+      ignoreCase: z.boolean().optional(),
+      literal: z.boolean().optional(),
+      context: z.number().optional(),
+      limit: z.number().optional(),
+    }),
+  }),
+  glob: commonTool('glob', {
+    nativeName: 'glob',
+    toolUseKind: 'readonly',
+    description: 'Find files matching a glob pattern.',
+    inputSchema: z.object({
+      pattern: z.string(),
+      path: z.string().optional(),
+      limit: z.number().optional(),
+    }),
+  }),
+  ls: {
+    ...tool({
+      description: 'List directory entries.',
+      inputSchema: z.object({
+        path: z.string().optional(),
+        limit: z.number().optional(),
+      }),
+      outputSchema: z.unknown(),
+    }),
+    nativeName: 'ls',
+    toolUseKind: 'readonly',
+  } as HarnessV1BuiltinTool,
+} as const satisfies Record<string, HarnessV1BuiltinTool<any, any>>;
+
+export function createCline(
+  settings: ClineHarnessSettings = {},
+): HarnessV1<typeof CLINE_BUILTIN_TOOLS> {
+  return {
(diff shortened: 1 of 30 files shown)


packages/harness-cline/src/cline-harness.ts
```diff
@@ -0,0 +1,181 @@
...
+        sessionId: startOpts.sessionId,
+        sandboxSession: startOpts.sandboxSession,
+        sessionWorkDir: startOpts.sessionWorkDir,
+        skills: startOpts.skills ?? [],
+        settings: {
+          providerId: settings.providerId ?? DEFAULT_PROVIDER_ID,
+          modelId: settings.modelId ?? DEFAULT_MODEL_ID,
+          ...(settings.apiKey ? { apiKey: settings.apiKey } : {}),
+          ...(settings.baseUrl ? { baseUrl: settings.baseUrl } : {}),
+          ...(settings.headers ? { headers: settings.headers } : {}),
```
felixarntz, line 157: Could we route auth resolution through a `cline-auth.ts` using `getAiGatewayAuthFromEnv()` as required per our harness conventions?

Cline currently forwards only explicitly supplied credentials, so standard `AI_GATEWAY_API_KEY` / `VERCEL_OIDC_TOKEN` auth is not picked up. Gateway requests should also include versioned client attribution via `CLINE_CLIENT_APP` (`ai-sdk/harness-cline/${VERSION}`).

packages/harness-cline/src/cline-harness.ts
```diff
@@ -0,0 +1,181 @@
...
+   * Additional system prompt text appended to the adapter's base system
+   * prompt (workspace + sandbox + skills sections).
+   */
+  readonly systemPrompt?: string;
+  /**
+   * Safety cap on agent-loop iterations per turn. Unbounded when omitted.
+   */
+  readonly maxIterations?: number;
+};
+
```
felixarntz, line 45: `pnpm konsistent` currently fails with a few errors, e.g. missing required file `cline-auth.ts` and missing local constant `CLINE_CLIENT_APP`. Please add both required harness conventions so the repository-wide check passes.

packages/harness-cline/src/cline-translate.ts
```diff
@@ -0,0 +1,207 @@
...
+          part.toolCallId === event.toolCall.toolCallId,
+      );
+      const output = resultPart?.output as AgentToolResult['output'] | unknown;
+      return [
+        {
+          type: 'tool-result',
+          toolCallId: event.toolCall.toolCallId,
+          toolName: event.toolCall.toolName,
+          result: toToolResultValue(output),
+          ...(resultPart?.isError ? { isError: true } : {}),
```
felixarntz, line 164: `isError` does not reach this point for the current failure paths: built-ins catch and return `{ error }` as a successful value in `guarded()`, while `submitToolResult` resolves only `args.output` and drops `args.isError`. Please preserve the error bit through both paths so failed custom and built-in tool results emit `isError: true`.

packages/harness-cline/src/cline-translate.ts
```diff
@@ -0,0 +1,207 @@
...
+ * runtime's promise settles, so a suspend can close the stream cleanly.
+ */
+export function translateClineEvent(
+  event: AgentRuntimeEvent,
+  state: ClineTranslatorState,
+): HarnessV1StreamPart[] {
+  switch (event.type) {
+    case 'assistant-reasoning-delta': {
+      const parts: HarnessV1StreamPart[] = [];
+      if (!state.openReasoningBlockId) {
```
felixarntz, line 102: Can we close `openTextBlockId` before opening a reasoning block? Otherwise a text → reasoning transition leaves both blocks open and emits overlapping content blocks.

packages/harness-cline/src/cline-translate.ts
```diff
@@ -0,0 +1,207 @@
...
+        id: state.openTextBlockId,
+        delta: event.text,
+      });
+      return parts;
+    }
+
+    case 'assistant-message': {
+      const parts = closeOpenBlocks(state);
+      parts.push({
+        type: 'finish-step',
```
felixarntz, line 135: Cline emits `assistant-message` before this iteration’s `tool-started` / `tool-finished` events, whereas the other harness adapters keep the step open through its tool results. Could we buffer the finish reason and usage here, then emit `finish-step` after this iteration’s tools complete?

It seems that Cline's terminology is slightly different from AI SDK's, which is fine, we just have to properly map it:
- Cline `turn-finished`, as far as I can tell, is a model step, so that should trigger AI SDK's `finish-step`
- Cline `run-finished`, as far as I can tell, is the full agent turn, so that should trigger AI SDK's `finish`

I may not be 100% right here, but the current implementation feels off in some way.

felixarntz (changes requested):
@saoudrizwan Thank you for this PR - excited to see Cline becoming available for `HarnessAgent`!

There are a few problems that need to be addressed in the PR to integrate properly with our abstraction, see individual comments below. Additionally, there are two other high-level points that need iteration:
- Since this PR is ~2 weeks old at this point, a few things in our harness layer have been enhanced or fixed, leading to additional requirements for the harness adapters. Of course this is on me, given I didn't get to review this sooner.
- There are no runnable or interactive examples for the new Cline harness, making end-to-end verification extremely difficult. We need to add examples to `examples/ai-functions` and `examples/harness-e2e-next` just like there are for the other harnesses.

I will take care of those two high-level concerns, including merging latest `main` back into this branch. I'll leave another review comment once I'm done with that.

felixarntz:
@saoudrizwan I implemented a few iterations locally, but looks like I can't push them to this branch. Let me know what you prefer. If you can't give me access, maybe I could push them elsewhere and you can copy it over, or you could take a stab at it yourself.

felixarntz:
Closing in favor of #18821, which continues that work, per our agreed path forward.

Issue #18093, opened 2026-07-28
Expose providerMetadata in OnLanguageModelCallEndCallback


aletorrado:
### Description

Hi,

It seems that `providerMetadata` is only available on `onStepEnd` and `onEnd` callbacks, but not on the `LanguageModelCall` level, although `usage` and `performance` information is exposed. This seems like an inconsistency.

I'm mostly interesting on tracking AI Gateway costs on the most granular level. Please keep in mind that neither `onStepEnd` nor `onEnd` will be called when the execution is aborted just after `onLanguageModelCallEnd` (namely, while tools are running).

Do you agree that it would be reasonable to expose this information there?

Thanks!

### AI SDK Version

- ai: 7.0.31

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #18103, opened 2026-07-29
feat (ai): expose providerMetadata in onLanguageModelCallEnd callback

Mohith26:
Implements #18093, filed by @gr2m.

Adds `providerMetadata` to the `onLanguageModelCallEnd` callback event: `generateText` passes it through from the model response, and the streaming path collects it from the final stream parts in `stream-language-model-call`. Field shape and optionality mirror the neighboring fields on the event type.

Changeset included (`ai` patch). Tests cover both paths: the callback receives providerMetadata from a mock model that emits it (generate + stream) and the field is absent when the model provides none. Both new tests fail without the implementation; generate-text and stream-language-model-call suites pass with zero new failures.

### DIFF
--- a/.changeset/lazy-planets-attend.md
+++ b/.changeset/lazy-planets-attend.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Expose providerMetadata in the onLanguageModelCallEnd callback event.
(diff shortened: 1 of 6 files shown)


Mohith26:
Closing: I missed that #18100 already covers this before opening. Sorry for the noise.

PR #18148, opened 2026-07-29
fix(ai): avoid adopting previous assistant messages on resume

### LINKED ISSUE #18107: resumeStream adopts the last assistant message even when it belongs to a different turn, duplicating its content into the resumed message
## Summary

On `resumeStream()`, the resumed message is seeded from `lastMessage` whenever that message's role is `assistant` — with no check that it is the *same* message the resumed stream is about. When the client's last local message belongs to a **different, already-finished turn**, that turn's parts are adopted into the resumed message, and the `start` chunk then renames it to the resumed stream's `messageId`. The result is a new message that carries the previous reply's text, while the previous reply also stays in the list — the old answer is rendered twice, once under the wrong id.

This is distinct from #14688 (which is about duplicate `text`/`reasoning` parts when replaying against a partial of *the same* message). Here the adopted message is a different turn entirely; the root cause is the missing id check in the adopt decision rather than part-level dedupe.

## Where

`packages/ai/src/ui/chat.ts` — the resume branch of `makeRequest` seeds state from whatever is currently last:

```ts
const lastMessage = this.lastMessage;
// …
state: createStreamingUIMessageState({
  lastMessage: this.state.snapshot(lastMessage),
  messageId: this.generateId(),
}),
```

`packages/ai/src/ui/process-ui-message-stream.ts` — the adopt decision looks only at `role`:

```ts
message: lastMessage?.role === 'assistant'
  ? lastMessage
  : { id: messageId, metadata: undefined, role: 'assistant', parts: [] },
```

…and the `start` handler then overwrites the id of that adopted message:

```ts
case 'start': {
  if (chunk.messageId != null) {
    state.message.id = chunk.messageId;
  }
```

So an assistant message from turn N−1 becomes, structurally, turn N's message.

## Reproduction

`ai@7.0.41`, `@ai-sdk/react@4.0.44`. A `Chat` whose `reconnectToStream` replays a stream for `turn-1` (`{"type":"start","messageId":"turn-1:reply"}` followed by a text part `"RESUMED-TEXT"`), against three different local message states:

**A — no local messages.** Correct:

```json
[{ "id": "turn-1:reply", "role": "assistant", "parts": [ …, { "type": "text", "text": "RESUMED-TEXT" } ] }]
```

**B — last local message is a previous turn's assistant reply.** Wrong:

```json
[
  { "id": "turn-0",       "role": "user",      "parts": [{ "type": "text", "text": "first question" }] },
  { "id": "turn-0:reply", "role": "assistant", "parts": [{ "type": "text", "text": "first answer" }] },
  { "id": "turn-1:reply", "role": "assistant", "parts": [
      { "type": "text", "text": "first answer" },   // <-- adopted from turn-0:reply
      { "type": "step-start" },
      { "type": "text", "text": "RESUMED-TEXT" }
  ]}
]
```

`"first answer"` now appears twice, and `turn-1:reply` contains content that was never part of turn 1.

**C — the turn's own user message is present.** Correct — the resumed reply is appended cleanly.

## Expected

On resume, adopt `lastMessage` only when it is plausibly the same message the stream is resuming — i.e. when the stream's `messageId` matches `lastMessage.id` (or when the stream supplies no `messageId`). Otherwise start from an empty assistant message. Since the `start` chunk carries the id, the mismatch is detectable before any part is written.

## Why this shows up in practice

With resumable streams, the client that resumes is frequently **not** the client that sent the message: another tab or another device started the turn, and this client learns the stream id from its own backend and attaches. That client legitimately has no local placeholder for the in-flight turn — its last local message is the previous turn's reply, which is exactly case B. (We hit this building multi-client follow on top of assistant-ui's resumable streams; our workaround is to refuse to attach unless the turn's own user message is present locally, and to delete any existing copy of the reply first.)

Punisheroot:
## Summary

- Prevent `resumeStream()` from adopting an assistant message from a previous turn when the resumed stream identifies a different message.
- Preserve accumulated parts when the resumed stream identifies the same assistant message.
- Preserve existing behavior when the stream does not provide a message ID and for non-resume stream processing.

## Root cause

`createStreamingUIMessageState` initially adopted any trailing assistant message before the resumed stream's identity was known. When the stream later emitted a `start` chunk with a different `messageId`, the processor replaced the adopted message's ID but retained its parts. The chat write path then appended that renamed message, duplicating content from the previous turn.

## Fix

`processUIMessageStream` now accepts an `isResume` option. `AbstractChat` enables it only for the `resume-stream` trigger.

For resumed streams, a `start` chunk whose `messageId` differs from the currently adopted message resets the state to an empty assistant message before applying stream metadata and subsequent parts. The option defaults to `false`, so other callers and existing submit, regenerate, and server-side tool round-trip behavior remain unchanged.

## Tests

Added resume-stream coverage for:

- a different message ID, which leaves the previous assistant unchanged and appends a clean new assistant;
- a matching message ID, which preserves the existing assistant parts;
- no stream message ID, which preserves the existing compatibility behavior.

## Validation

- `pnpm exec vitest --config vitest.node.config.js --run src/ui/chat.test.ts src/ui/process-ui-message-stream.test.ts`
- `pnpm type-check` in `packages/ai`
- `pnpm exec ultracite check packages/ai/src/ui/process-ui-message-stream.ts packages/ai/src/ui/chat.ts packages/ai/src/ui/chat.test.ts`
- `git diff --check`

The full workspace type-check was not completed locally because this checkout used a filtered dependency install; unrelated examples and workspace packages reported missing dependencies. The affected package type-check and focused tests pass.

Fixes #18107

### DIFF
--- a/.changeset/quiet-lions-resume.md
+++ b/.changeset/quiet-lions-resume.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Prevent resumed streams from adopting assistant messages from previous turns when the stream identifies a different message.
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -1471,6 +1471,133 @@ describe('Chat', () => {
     expect(chat.status).toBe('error');
   });
 
+  describe('resume stream', () => {
+    function createResumeStream(messageId?: string) {
+      return new ReadableStream<UIMessageChunk>({
+        start(controller) {
+          controller.enqueue({
+            type: 'start',
+            ...(messageId == null ? {} : { messageId }),
+          });
+          controller.enqueue({ type: 'start-step' });
+          controller.enqueue({ type: 'text-start', id: 'text-1' });
+          controller.enqueue({
+            type: 'text-delta',
+            id: 'text-1',
+            delta: 'resumed',
+          });
+          controller.enqueue({ type: 'text-end', id: 'text-1' });
+          controller.enqueue({ type: 'finish-step' });
+          controller.enqueue({ type: 'finish', finishReason: 'stop' });
+          controller.close();
+        },
+      });
+    }
+
+    function createChat({
+      messages,
+      messageId,
+    }: {
+      messages: UIMessage[];
+      messageId?: string;
+    }) {
+      return new TestChat({
+        id: '123',
+        messages,
+        generateId: mockId(),
+        transport: {
+          sendMessages: async () => {
+            throw new Error('not implemented');
+          },
+          reconnectToStream: async () => createResumeStream(messageId),
+        },
+      });
+    }
+
+    it('should not adopt an assistant message from a different turn', async () => {
+      const chat = createChat({
+        messages: [
+          {
+            id: 'turn-0',
+            role: 'user',
+            parts: [{ type: 'text', text: 'first question' }],
+          },
+          {
+            id: 'turn-0:reply',
+            role: 'assistant',
+            parts: [{ type: 'text', text: 'first answer' }],
+          },
+        ],
+        messageId: 'turn-1:reply',
+      });
+
+      await chat.resumeStream();
+
+      expect(chat.messages).toHaveLength(3);
+      expect(chat.messages[1]).toMatchObject({
+        id: 'turn-0:reply',
+        parts: [{ type: 'text', text: 'first answer' }],
+        role: 'assistant',
+      });
+      expect(chat.messages[2]).toMatchObject({
+        id: 'turn-1:reply',
+        parts: [{ type: 'step-start' }, { type: 'text', text: 'resumed' }],
+        role: 'assistant',
+      });
+    });
+
+    it('should preserve the assistant message when its id matches', async () => {
+      const chat = createChat({
+        messages: [
+          {
+            id: 'turn-1:reply',
+            role: 'assistant',
+            parts: [{ type: 'text', text: 'partial response' }],
+          },
+        ],
+        messageId: 'turn-1:reply',
+      });
+
+      await chat.resumeStream();
+
+      expect(chat.messages).toHaveLength(1);
+      expect(chat.messages[0]).toMatchObject({
+        id: 'turn-1:reply',
+        parts: [
+          { type: 'text', text: 'partial response' },
+          { type: 'step-start' },
+          { type: 'text', text: 'resumed' },
+        ],
+        role: 'assistant',
+      });
+    });
+
+    it('should preserve the assistant message when the stream has no message id', async () => {
+      const chat = createChat({
+        messages: [
+          {
+            id: 'turn-0:reply',
+            role: 'assistant',
+            parts: [{ type: 'text', text: 'existing response' }],
+          },
+        ],
+      });
+
+      await chat.resumeStream();
+
+      expect(chat.messages).toHaveLength(1);
+      expect(chat.messages[0]).toMatchObject({
+        id: 'turn-0:reply',
+        parts: [
+          { type: 'text', text: 'existing response' },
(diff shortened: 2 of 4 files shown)


dylan-at-adam:
#18462 (shipped in 7.0.61) fixed #18457 by never seeding a resumed stream from the last assistant message. That fixes the different-turn duplicate, but it regresses resuming a **continuation of the same message**, which is what the built-in tool approval / client-side tool flows produce.

### What breaks

After a tool approval (or `addToolOutput`), the continuation reuses the assistant message's id and streams only the new parts. If the page reloads while that continuation is still streaming and the client calls `resumeStream()`, the replay's `start` chunk carries that **same** id:

- **≤ 7.0.60:** the reducer seeds from the last assistant message, the ids match, and the new parts are appended. Correct.
- **≥ 7.0.61:** the reducer starts from an empty message, builds one containing only the new parts, and `replaceMessage` swaps it in by id. The earlier text and the approved tool call disappear from the UI until the next full reload.

The fix in this PR handles both cases: keep the seed when the `start` chunk's `messageId` matches it, and reset only when it differs.

### Repro (no React, no network)

```js
import { AbstractChat } from 'ai';

class State {
  status = 'ready'; error;
  constructor(m) { this.messages = m; }
  pushMessage = m => { this.messages = [...this.messages, m]; };
  popMessage = () => { this.messages = this.messages.slice(0, -1); };
  replaceMessage = (i, m) => { this.messages = this.messages.map((x, j) => (j === i ? m : x)); };
  snapshot = t => structuredClone(t);
}
class Chat extends AbstractChat {}
const stream = cs => new ReadableStream({ start(c) { cs.forEach(x => c.enqueue(x)); c.close(); } });

// Persisted state after the user approved the tool call.
const state = new State([
  { id: 'u1', role: 'user', parts: [{ type: 'text', text: 'Delete old files' }] },
  { id: 'a1', role: 'assistant', parts: [
    { type: 'text', text: 'I will delete them.', state: 'done' },
    { type: 'tool-deleteFiles', toolCallId: 'c1', state: 'approval-responded', input: {}, approval: { id: 'ap1', approved: true } },
  ] },
]);

// Page reloaded mid-continuation: the replay reuses id 'a1' and carries only the new parts.
const chat = new Chat({ id: 'x', state, generateId: () => 'gen', transport: {
  sendMessages: async () => stream([]),
  reconnectToStream: async () => stream([
    { type: 'start', messageId: 'a1' },
    { type: 'text-start', id: 't2' },
    { type: 'text-delta', id: 't2', delta: 'Done.' },
    { type: 'text-end', id: 't2' },
    { type: 'finish' },
  ]),
} });

await chat.resumeStream();
console.log(chat.messages.map(m => m.id + ': ' + m.parts.map(p => p.type).join(', ')));
```

"Same message" is the script as written. "Different turn" is the #18457 case: the same script with the `start` chunk's `messageId` changed to `'a2'`.

### Results

| Build | Same message (`a1` → `a1`) | Different turn (`a1` → `a2`) |
| --- | --- | --- |
| `ai@7.0.60` | ✅ `a1: text, tool-deleteFiles, text` | ❌ `a2: text, tool-deleteFiles, text` (a1's parts duplicated into a2) |
| `ai@7.0.111` | ❌ `a1: text` (earlier text + approved tool call dropped) | ✅ `a2: text` |
| This PR (`b139ae2`, built locally) | ✅ `a1: text, tool-deleteFiles, text` | ✅ `a2: text` |

So 7.0.60 fails the different-turn case, 7.0.111 fails the same-message case, and this PR is the only build that passes both.

This branch now conflicts with `main` because of #18462. Could it be rebased onto it, so that resume seeds from the last assistant message again and resets only on an id mismatch? Adding the same-message case above as a regression test next to #18462's would keep both behaviors covered.

PR #18153, opened 2026-07-29
fix (harness-pi): dispatch custom model ids through customEnv-registered providers

### LINKED ISSUE #18147: harness-pi: auth.customEnv with a custom provider + custom model id never dispatches — falls back to Pi's default model and fails with "No API key found for the selected model"
### Summary

`createPi({ auth: { customEnv: { <PREFIX>_API_KEY, <PREFIX>_BASE_URL } }, model: "<custom model id>" })` never dispatches to the registered provider. The request is never sent, and the turn fails with:

```
Error: No API key found for the selected model.
```

`customEnv` registers the provider's **credentials** correctly, but there is no way to declare that provider's **models**. Since resolution matches against Pi's catalog only, a custom model id matches nothing, Pi silently falls back to its own default model, and that default's provider has no credential — producing an error that points at the wrong thing.

The docs describe `customEnv` as the way to use multiple/custom providers ("Any other `<PREFIX>_API_KEY` with a matching `<PREFIX>_BASE_URL` is registered as the lowercased, dash-separated prefix"), so this path appears supported but cannot select a model.

### Reproduction

Self-contained; needs no cloud credentials. It stands up a local OpenAI-compatible server and asserts whether the request ever arrives.

> Note: `@ai-sdk/sandbox-just-bash` cannot be used for this repro because `createSession` always fails there (#16690), so the script inlines a minimal local sandbox.

```ts
import { HarnessAgent } from "@ai-sdk/harness/agent";
import { createPi } from "@ai-sdk/harness-pi";
import { spawn } from "node:child_process";
import { mkdtemp, mkdir, readFile, writeFile } from "node:fs/promises";
import { createServer } from "node:http";
import { tmpdir } from "node:os";
import { join } from "node:path";

const makeSandbox = (root: string) => {
  const mk = async (id: string) => {
    const cwd = join(root, id);
    await mkdir(cwd, { recursive: true });
    return {
      id, description: cwd, defaultWorkingDirectory: cwd, ports: [] as number[],
      run: (o: { command: string; env?: Record<string, string> }) =>
        new Promise((res) => {
          const c = spawn(o.command, { cwd, shell: true, env: { ...process.env, ...o.env } });
          let out = "", errOut = "";
          c.stdout?.on("data", (d) => (out += d));
          c.stderr?.on("data", (d) => (errOut += d));
          c.on("close", (code) => res({ exitCode: code ?? 1, stdout: out, stderr: errOut }));
        }),
      readTextFile: async ({ path }: { path: string }) =>
        readFile(join(cwd, path), "utf8").catch(() => null),
      writeTextFile: async ({ path, content }: { path: string; content: string }) => {
        await mkdir(join(cwd, path, ".."), { recursive: true });
        await writeFile(join(cwd, path), content, "utf8");
      },
      getPortUrl: () => { throw new Error("no ports"); },
      stop: async () => {}, destroy: async () => {},
      restricted() { const { stop, destroy, getPortUrl, ...rest } = this as never; return rest; },
    };
  };
  return {
    specificationVersion: "harness-sandbox-v1" as const,
    providerId: "minimal-local",
    createSession: (o?: { sessionId?: string }) => mk(o?.sessionId ?? "s"),
    resume: (o: { sessionId: string }) => mk(o.sessionId),
    resumeSession: (o: { sessionId: string }) => mk(o.sessionId),
  };
};

let received: { model?: string } | null = null;
const server = createServer((req, res) => {
  let b = ""; req.on("data", (c) => (b += c));
  req.on("end", () => {
    received = JSON.parse(b || "{}");
    res.writeHead(200, { "content-type": "text/event-stream" });
    res.write('data: {"choices":[{"index":0,"delta":{"content":"ok"},"finish_reason":"stop"}]}\n\n');
    res.write("data: [DONE]\n\n"); res.end();
  });
});
const baseUrl = await new Promise<string>((r) =>
  server.listen(0, "127.0.0.1", () =>
    r(`http://127.0.0.1:${(server.address() as { port: number }).port}/v1`)));

const root = await mkdtemp(join(tmpdir(), "repro-"));
  harness: createPi({
    auth: { customEnv: { MYPROVIDER_API_KEY: "sk-test", MYPROVIDER_BASE_URL: baseUrl } },
    model: "my-custom-model",
  }),
  sandbox: makeSandbox(root) as never,
  tools: {},
});
const session = await agent.createSession({ sessionId: "repro-1" });
const result = await agent.stream({ session, prompt: "hi" });
let err = "";
for await (const p of result.fullStream) if (p.type === "error") err = String((p as { error: unknown }).error);
console.log("reached my provider:", received !== null);
console.log("model sent         :", received?.model ?? "(none)");
console.log("error              :", err.split("\n")[0]);
await session.destroy().catch(() => undefined);
server.close();
```

### Expected

The request reaches `MYPROVIDER_BASE_URL` with `model: "my-custom-model"`, authenticated with `MYPROVIDER_API_KEY`.

### Actual

```
reached my provider: false
model sent         : (none)
error              : Error: No API key found for the selected model.
```

### Root cause

1. `registerCustomProviders` registers the provider config and runtime API key, but registers **no models** for it.
2. `createPiModelResolver` resolves with `models.find(m => m.id === effectiveId || m.name === effectiveId)` over `modelRegistry.getAll()`, which contains only Pi's built-in catalog plus any `models.json` entries.
3. A provider registered through `customEnv` contributes no catalog entries, so the id matches nothing, `resolvedModel` is `undefined`, and Pi falls back to its own default model — whose provider has no credential, producing the misleading error above.

### Related symptom, same root cause

When the model id **is** in Pi's catalog under several providers, resolution ignores which provider was actually authenticated and picks an unauthenticated one. With `customEnv: { OPENAI_API_KEY, OPENAI_BASE_URL }` and `model: "gpt-4o"`, the turn fails with:

```
Error: No API key found for azure-openai-responses.
```

even though `openai` was the provider registered. I could not find any configuration that overrides this — `agentDir` + `models.json` (including the documented built-in-provider override/merge form), `settings.json` `defaultProvider`, a provider-qualified id (`openai/gpt-4o`, which resolves to `openrouter`), and provider-name ordering were all tried. Only the AI Gateway path resolves correctly, because it is the one branch that filters by provider.

A third, quieter consequence: an unresolvable model id is **silently replaced** by Pi's default model rather than erroring. Passing a nonexistent id and a reachable `customEnv` base URL sends `model: "gpt-5.5"` to that endpoint, with no warning that the requested model was not used.

### Possibly addressed by #16099

PR #16099 ("feat(harness-pi): add `providers` option to register custom providers per session") appears to cover both halves — it lets a host register a provider config with custom models, and makes the resolver prefer a provider whose credential the caller actually seeded. Filing this as a bug report since there is no issue tracking the current behaviour, and the PR is still open.

### Environment

- `@ai-sdk/harness@1.0.47`, `@ai-sdk/harness-pi@1.0.47`, `ai@7.0.41`
- `@earendil-works/pi-coding-agent@0.80.10`
- Node v22.14.0, macOS (arm64)

Mohith26:
Fixes #18147, reported by @Leonezz.

Providers registered through `auth.customEnv` contributed credentials but no catalog entries, so a custom model id matched nothing and Pi silently swapped in its own default model, then failed with a misleading "No API key found".

Three changes in harness-pi: generic `<PREFIX>_API_KEY`/`<PREFIX>_BASE_URL` providers now register with `api: 'openai-completions'` (the pattern's contract); the resolver prefers catalog matches whose provider is actually registered for the session (generalizing the existing vercel-ai-gateway preference); and when an explicitly configured model id matches no catalog entry and exactly one registered provider declares `api` + `baseUrl`, the resolver synthesizes a model entry for it (Pi's own custom-models.json defaults), so the request dispatches to the user's endpoint. Ambiguous multi-provider cases intentionally keep current behavior; that design belongs to #16099.

Changeset included (`@ai-sdk/harness-pi` patch). Five new tests; 3 fail with the fix reverted. Package suite: 122 passed (baseline 117, zero new failures). tsc/oxlint/oxfmt clean.

### DIFF
--- a/.changeset/harness-pi-custom-env-dispatch.md
+++ b/.changeset/harness-pi-custom-env-dispatch.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-pi': patch
+---
+
+fix (harness-pi): dispatch custom model ids through `auth.customEnv`-registered providers and prefer registered providers during model resolution
--- a/packages/harness-pi/src/pi-auth.test.ts
+++ b/packages/harness-pi/src/pi-auth.test.ts
@@ -201,6 +201,7 @@ describe('registerPiProviders', () => {
       apiKey: 'mk',
       baseUrl: 'https://api.mistral.example',
       authHeader: true,
+      api: 'openai-completions',
     });
   });
 
--- a/packages/harness-pi/src/pi-auth.ts
+++ b/packages/harness-pi/src/pi-auth.ts
@@ -26,7 +26,9 @@ export type PiAuthOptions = {
    *  - `ANTHROPIC`  → registers `anthropic` (`ANTHROPIC_AUTH_TOKEN` adds a
    *                   bearer auth header)
    * Any other `<PREFIX>_API_KEY` with a matching `<PREFIX>_BASE_URL` is
-   * registered as the lowercased, dash-separated prefix.
+   * registered as the lowercased, dash-separated prefix. These providers are
+   * assumed to be OpenAI-compatible; a configured model id that is not in
+   * Pi's catalog is dispatched through the registered provider.
    */
   readonly customEnv?: Record<string, string>;
 };
@@ -247,6 +249,11 @@ async function registerCustomProviders({
         apiKey,
         baseUrl,
         authHeader: true,
+        // The `<PREFIX>_API_KEY` / `<PREFIX>_BASE_URL` contract is
+        // OpenAI-compatible. Declaring the wire API lets the model resolver
+        // dispatch model ids that are not in Pi's catalog through this
+        // provider (see `createPiModelResolver`).
+        api: 'openai-completions',
       },
     });
   }
--- a/packages/harness-pi/src/pi-model-resolver.test.ts
+++ b/packages/harness-pi/src/pi-model-resolver.test.ts
@@ -102,4 +102,75 @@ describe('createPiModelResolver', () => {
     });
     expect(resolve(undefined)).toBeUndefined();
   });
+
+  describe('customEnv-registered providers', () => {
+    const customProviderConfig = {
+      apiKey: 'sk-test',
+      baseUrl: 'https://my.provider.example/v1',
+      authHeader: true,
+      api: 'openai-completions',
+    } as const;
+
+    it('prefers a registered provider over an unregistered one sharing the model id', async () => {
+      const registeredModel: PiModel = {
+        ...sampleModel,
+        provider: 'my-provider',
+        baseUrl: customProviderConfig.baseUrl,
+      };
+      const modelRegistry = await makeRegistry([sampleModel, registeredModel]);
+      modelRegistry.registerProvider('my-provider', customProviderConfig);
+      const resolve = createPiModelResolver({ modelRegistry, env: {} });
+      expect(resolve('my/model')).toEqual(registeredModel);
+    });
+
+    it('dispatches an uncataloged model id through the single registered custom provider', async () => {
+      const modelRegistry = await makeRegistry([sampleModel]);
+      modelRegistry.registerProvider('my-provider', customProviderConfig);
+      const resolve = createPiModelResolver({ modelRegistry, env: {} });
+      expect(resolve('my-custom-model')).toEqual({
+        id: 'my-custom-model',
+        name: 'my-custom-model',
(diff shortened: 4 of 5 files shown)


gr2m:
Superseded by #20652, which already closed #18147.

PR #18155, opened 2026-07-29
fix(harness-pi): error when a configured model does not resolve instead of silently running Pi's default

Leonezz:
Fixes one of the three problems reported in #18147.

### Problem

`createPiSession` spreads the resolved model conditionally:

```ts
...(resolvedModel ? { model: resolvedModel } : {}),
```

So a **configured** model id that resolves to nothing simply omits the key, and Pi runs the turn on its own default model. Nothing reports the substitution.

When that default's provider has no credential, the mismatch finally surfaces as:

```
Error: No API key found for the selected model.
```

which points at a credential rather than at the unknown model id that actually caused it. In practice this is what makes the failure hard to diagnose — the error names the wrong thing, and the turn may otherwise succeed on a model the caller never asked for.

It is easy to hit with `auth.customEnv`: that path registers a provider's credentials (`registerProvider` + `setRuntimeApiKey`) but **no models**, so any model id belonging to that provider resolves to nothing.

### Change

Throw when a model was configured and did not resolve, naming the id:

```
Pi model 'my-custom-model' was not found in the model registry. Pi resolves a
model id against its built-in catalog and models.json only; an id that matches
nothing is not substituted. ...
```

`model: undefined` is deliberately left alone — that means "no model configured", and deferring to Pi's default is the intended behaviour. Only an id that was asked for and did not resolve is treated as an error.

### Scope

Intentionally narrow. This does **not** touch resolver provider-preference or custom-provider registration — #16099 already proposes both, and this change is complementary to it rather than overlapping. It only stops the failure from being silent.

### Tests

Three cases in `pi-session.test.ts`:

- a configured model that resolves to nothing now rejects (this fails without the change — verified by reverting the guard and re-running);
- no configured model still resolves, so the default path is preserved;
- a configured model the registry knows still resolves.

`pnpm --filter @ai-sdk/harness-pi test` → 13 files, 120 tests passing. Changeset included.


Leonezz:
Good catch from the review agent — that was a real leak, not a false positive.

`workspaceVfs.mount()` installs a process-global `fs` monkey-patch, and the only `unmount()` calls live in session lifecycle methods that never exist if `createPiSession` throws. So the guard I added leaked the patch and a stale `mountedRoots` entry for the lifetime of the process on every misconfigured model id.

Fixed in 7b61744 by unmounting before throwing. Model resolution genuinely depends on setup that happens after the mount (`ModelRuntime.create`, `registerPiProviders`), so moving the guard earlier would have meant reordering more than this change should.

Added a test that spies on `PiWorkspaceVfs.prototype.unmount` and asserts it runs on the rejection path. I verified it actually catches the regression by removing the `unmount()` call and re-running — that test fails alone, with the other seven still passing.

`pnpm --filter @ai-sdk/harness-pi test` → 13 files, 121 tests passing.

(The failing `Vercel` check is the preview-deploy authorization step for fork PRs — "Authorization required to deploy" — not related to this change.)

aayush-kapoor (commented):
fixed via https://github.com/vercel/ai/pull/20652

PR #18161, opened 2026-07-29
fix (ai): distinguish [] from [undefined] in canonicalJSON

### LINKED ISSUE #18157: `canonicalJSON` collides `[]` and `[undefined]` due to Array.prototype.join's undefined-to-empty-string coercion
## Description

`canonicalJSON` in `packages/ai/src/util/canonical-hash.ts` produces
identical output for `[]` and `[undefined]`, meaning `hashCanonical([])`
and `hashCanonical([undefined])` return the same SHA-256 digest despite
being structurally different values.

## Root Cause

```ts
if (Array.isArray(value)) {
  return `[${value.map(canonicalJSON).join(',')}]`;
}
```

`canonicalJSON(undefined)` returns the literal JS value `undefined`
(via `JSON.stringify(undefined)`, which itself returns `undefined`, not
the string `"undefined"`). When this `undefined` result flows through
`Array.prototype.join(',')`, JavaScript's spec-defined behavior treats
`undefined`/`null` array elements as empty strings during join. So:

```js
[].map(canonicalJSON).join(',')          // => ""
[undefined].map(canonicalJSON).join(',') // => ""  (same!)
```

Both produce the string `"[]"`, so both hash to the same digest.

## Reproduction

```js
function canonicalJSON(value) {
  if (value === null || value === undefined) return JSON.stringify(value);
  if (typeof value !== 'object') return JSON.stringify(value);
  if (Array.isArray(value)) return `[${value.map(canonicalJSON).join(',')}]`;
  const keys = Object.keys(value).sort();
  return `{${keys.map(k => `${JSON.stringify(k)}:${canonicalJSON(value[k])}`).join(',')}}`;
}

console.log(canonicalJSON([]));          // "[]"
console.log(canonicalJSON([undefined])); // "[]"  <- should differ
```

## Impact

I traced the current callers of `hashCanonical`/`canonicalJSON`
(`tool-approval-signature.ts` and `tool-fingerprint.ts`). Both current
call sites appear safe today:

- `tool-approval-signature.ts` hashes `toolCall.input`, which always
  arrives via `JSON.parse` from client-supplied message history — and
  JSON has no `undefined` literal, so this specific collision can't be
  triggered through that path today.
- `tool-fingerprint.ts` hashes object properties (not bare arrays), so
  the array-join coercion doesn't apply there either.

However, `canonicalJSON`/`hashCanonical` are general-purpose exported
utilities. Any future caller that hashes a live JS array (rather than
JSON-round-tripped data) containing an `undefined` element would be
exposed to this collision. Given this function's stated purpose is
producing collision-resistant digests for security-relevant comparisons
(it's already used for HMAC-signed tool-approval input binding), I think
this is worth hardening defensively even though I can't currently
demonstrate an exploitable path.

## Suggested Fix

Distinguish `undefined` array elements from a genuinely shorter array,
e.g.:

```ts
if (Array.isArray(value)) {
  return `[${value.map(v => (v === undefined ? 'null' : canonicalJSON(v))).join(',')}]`;
}
```

This makes `canonicalJSON([undefined])` produce `"[null]"`, distinct
from `canonicalJSON([])` → `"[]"`.

Happy to open a PR with this fix and accompanying test cases if useful.

### Reproduction

_No response_

### AI SDK Version

_No response_

### Code of Conduct

Mohith26:
Fixes #18157, reported by @trmxvibs.

`canonicalJSON` collided `[]` with `[undefined]`: `Array.prototype.join` coerces the `undefined` produced for undefined elements to an empty string, so `hashCanonical` produced identical SHA-256 digests for structurally different arrays.

The fix serializes undefined array elements as `null`, matching `JSON.stringify` semantics and the already-correct mirror implementation in `@ai-sdk/mcp` (which coerces via its `value == null` branch and was left untouched; the two are now behaviorally consistent). Call sites checked: neither the tool-approval signature (hashes JSON-round-tripped input) nor the tool fingerprint depends on the old collision. Top-level `canonicalJSON(undefined)` stays `undefined`, pinned by an existing test.

Changeset included (`ai` patch). Four regression tests pin `[]` vs `[undefined]` distinctness (string and digest), `[undefined]` -> `[null]`, the intentional `[null]`/`[undefined]` digest equivalence, and mid-array holes; exactly those 4 fail with the fix reverted. Full package suite: 3040 passed, zero new failures; ultracite clean.

### DIFF
--- a/.changeset/plenty-lemons-hammer.md
+++ b/.changeset/plenty-lemons-hammer.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix (ai): serialize undefined array elements as null in canonicalJSON so [] and [undefined] no longer produce the same canonical string and hash
--- a/packages/ai/src/util/canonical-hash.test.ts
+++ b/packages/ai/src/util/canonical-hash.test.ts
@@ -18,6 +18,20 @@ describe('canonicalJSON', () => {
     expect(canonicalJSON('x')).toBe('"x"');
     expect(canonicalJSON(42)).toBe('42');
   });
+
+  it('distinguishes [] from [undefined]', () => {
+    expect(canonicalJSON([])).toBe('[]');
+    expect(canonicalJSON([undefined])).toBe('[null]');
+  });
+
+  it('serializes undefined array elements as null, matching JSON.stringify', () => {
+    expect(canonicalJSON([undefined])).toBe(JSON.stringify([undefined]));
+    expect(canonicalJSON([null])).toBe('[null]');
+    expect(canonicalJSON([1, undefined, 2])).toBe('[1,null,2]');
+    expect(canonicalJSON([1, undefined, 2])).toBe(
+      JSON.stringify([1, undefined, 2]),
+    );
+  });
 });
 
 describe('hashCanonical', () => {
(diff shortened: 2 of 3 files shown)


aayush-kapoor (commented):
fixed via https://github.com/vercel/ai/pull/18265

PR #18211, opened 2026-07-30
fix(generate-text): fire onToolCallStart/Finish for tools without execute

### LINKED ISSUE #7660: Tool call information does not appear in telemetry if tool does not implement `execute`
### Description

# Problem

I think this is a general issue, but I'm using Langfuse with `langfuse-vercel` and `@vercel/otel` so I'm referring specifically to that.

When a call to e.g. `streamText` ends with a tool call (`finishReason == 'tool-calls'`), there does not seem to be any telemetry information sent to Langfuse about that tool call. This causes an issue when using tool calls to signal to frontend code to render UI widgets, or in general when a tool call is used to signal that the result is expected to come from the client side. This applies to any `Tool` that doesn't implement `execute`.

For example, a trace ending on a tool call looks like this in langfuse:

Image

Normally, a tool call will cause a span to appear in the Langfuse trace showing the input and output:

Image

# Reproduction

[gpoole/langfuse-tool-call-missing-reproduction](https://github.com/gpoole/langfuse-tool-call-missing-reproduction) is a demonstration of the problem.

It demonstrates that:
* The `askForLocation` tool does not appear in the trace when the user says "What's the weather"
* The `showWeatherForLocation` tool _does_ appear in the trace when the user says "What's the weather in sydney"

### AI SDK Version

@ai-sdk/react@1.2.12
ai@4.3.19

okxint:
**Problem**

`executeToolCall` returned `undefined` before constructing `baseCallbackEvent` when a tool has no `execute` function (client-side / UI tools). This meant `onToolCallStart` and `onToolCallFinish` were never fired for these tools.

Users using telemetry integrations (Langfuse, OpenTelemetry, etc.) see no spans at all for client-side tool calls — they appear as silent gaps in traces.

**Fix**

Hoist `baseCallbackEvent` construction before the `tool?.execute == null` guard, then fire both callbacks before returning:

- `onToolCallStart` with the same event shape as server-side tools
- `onToolCallFinish` with `output: undefined` and `durationMs: 0` (no execution happened)

The existing code path for tools with `execute` is unchanged — `baseCallbackEvent` is now constructed once and reused.

Fixes #7660

### DIFF
--- a/.changeset/fix-tool-callbacks-no-execute.md
+++ b/.changeset/fix-tool-callbacks-no-execute.md
@@ -1,0 +1,13 @@
+---
+"ai": patch
+---
+
+fix(generate-text): fire onToolCallStart and onToolCallFinish for tools without execute
+
+Tools that don't implement `execute` (client-side / UI tools) previously returned
+`undefined` before the callback event was constructed, so `onToolCallStart` and
+`onToolCallFinish` were never fired. Tool calls for no-execute tools were invisible
+in telemetry backends (Langfuse, OpenTelemetry, etc.).
+
+The callbacks are now fired before returning, with `output: undefined` and
+`durationMs: 0` for the finish event.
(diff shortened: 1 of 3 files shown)


aayush-kapoor (commented):
marking as closed in favor of https://github.com/vercel/ai/pull/19086

PR #18224, opened 2026-07-30
fix(openai): send file upload retention as the anchored object the Files API requires

### LINKED ISSUE #18223: `@ai-sdk/openai`: `files().uploadFile()` sends `expires_after` in a shape the API rejects, so file retention cannot be set
## Description

`FilesV4.uploadFile()` in the OpenAI provider accepts `providerOptions.openai.expiresAfter`, but appends it to the multipart body as a single flat field:

```js
// packages/openai/src/files/openai-files.ts
formData.append('purpose', openaiOptions?.purpose ?? 'assistants');
if (openaiOptions?.expiresAfter != null) {
  formData.append('expires_after', String(openaiOptions.expiresAfter));
}
```

The Files API expects `expires_after` as an object (`anchor` + `seconds`), which in multipart means bracketed field names. The flat field is rejected with a 400, so **there is currently no way to upload a file with a retention through the AI SDK's files interface.**

The failure mode matters more than the error itself: the natural reaction is to drop the option, and then the upload succeeds with **no expiry at all** — the file stays at the provider indefinitely. Nothing in the result signals it (`warnings` is empty).

## Reproduction

```ts
import { uploadFile } from 'ai';

await uploadFile({
  api: createOpenAI({ apiKey }).files(),
  data: new Uint8Array(Buffer.from('hello\n')),
  mediaType: 'text/plain',
  filename: 'probe.txt',
  providerOptions: { openai: { purpose: 'user_data', expiresAfter: 604800 } },
});
```

```
AI_APICallError: Additional properties are not allowed ('expires_after' was unexpected)
```

Dropping `expiresAfter` uploads successfully, and `files.retrieve(id)` then reports no `expires_at`.

## What the API accepts

I sent the same multipart upload four ways to `POST /v1/files` (`purpose: user_data`, same key, same bytes), deleting each file afterwards:

| multipart fields | result |
| --- | --- |
| *(no field — control)* | 200, `expires_at` absent |
| `expires_after=604800` — **what the SDK sends today** | 400 `Additional properties are not allowed ('expires_after' was unexpected)` |
| `expires_after[anchor]=created_at` + `expires_after[seconds]=604800` | **200, `expires_at` set correctly** |
| `expires_after.anchor=created_at` + `expires_after.seconds=604800` | 400 `('expires_after.anchor', 'expires_after.seconds' were unexpected)` |

The official `openai` package produces the bracketed form: its `addFormValue` maps a nested object to `` `${key}[${name}]` `` (`src/internal/uploads.ts`), and `client.files.create({ purpose: 'user_data', expires_after: { anchor: 'created_at', seconds: 604800 } })` sets the expiry as expected.

Note for anyone who searches this up: `expires_after` **does** work with `purpose: 'user_data'` — contrary to some reports, the purpose is not the problem, only the field shape is.

## Proposed fix

```diff
- if (openaiOptions?.expiresAfter != null) {
-   formData.append('expires_after', String(openaiOptions.expiresAfter));
- }
+ if (openaiOptions?.expiresAfter != null) {
+   // `anchor` is required; the API rejects a flat `expires_after` field.
+   formData.append('expires_after[anchor]', 'created_at');
+   formData.append('expires_after[seconds]', String(openaiOptions.expiresAfter));
+ }
```

The same package already gets this right elsewhere — `packages/openai/src/realtime/openai-realtime-model.ts` sends the client-secret expiry as an object, with a comment naming the exact failure:

```ts
// `anchor` is required by the client secrets endpoint; without it
// the request fails with "Missing required parameter: 'expires_after.anchor'".
expires_after: { anchor: 'created_at', seconds: options.expiresAfterSeconds }
```

If a non-default anchor is ever needed, `expiresAfter` could accept `number | { anchor, seconds }` while keeping the number shorthand.

## Environment

- `@ai-sdk/openai` **4.0.24** (latest published) — verified in the published bundle; also reproduced on 4.0.20
- `ai` 7.0.37, `@ai-sdk/provider` 4.0.3
- Reproduced against the live `https://api.openai.com/v1/files` endpoint

I checked the tracker before filing: searches for `expires_after` / `expiresAfter` return nothing, and none of the open `uploadFile` issues cover retention.

Azbesciak:
Fixes #18223

## Problem

`files().uploadFile()` accepts `providerOptions.openai.expiresAfter`, but appends it to the multipart body as a single flat field:

```ts
formData.append('expires_after', String(openaiOptions.expiresAfter));
```

The Files API expects `expires_after` as an object (`anchor` + `seconds`), so the request is rejected:

```
Additional properties are not allowed ('expires_after' was unexpected)
```

There is currently no way to upload a file with a retention through the files interface. The workaround — dropping the option — succeeds and uploads a file that **never expires**, with nothing in `warnings` to signal it.

## What the API accepts

Four multipart encodings against `POST /v1/files`, same key, same bytes, `purpose: user_data`, each file deleted afterwards:

| multipart fields | result |
| --- | --- |
| *(none — control)* | 200, `expires_at` absent |
| `expires_after=604800` — **what is sent today** | 400 `Additional properties are not allowed ('expires_after' was unexpected)` |
| `expires_after[anchor]=created_at` + `expires_after[seconds]=604800` | **200, `expires_at` set** |
| `expires_after.anchor=…` + `expires_after.seconds=…` | 400 `('expires_after.anchor', 'expires_after.seconds' were unexpected)` |

The bracketed form is what the official `openai` package produces — its `addFormValue` maps a nested object to `` `${key}[${name}]` `` — and `client.files.create({ purpose: 'user_data', expires_after: { anchor: 'created_at', seconds: 604800 } })` sets the expiry as expected.

`purpose: 'user_data'` is not the problem, contrary to some reports; only the field shape is.

## Change

`expires_after[anchor]` is sent alongside `expires_after[seconds]`. The anchor is required and `created_at` is its only accepted value today — the same constant this package already hardcodes for the realtime client-secret endpoint, where a comment records the same requirement:

```ts
// `anchor` is required by the client secrets endpoint; without it
// the request fails with "Missing required parameter: 'expires_after.anchor'".
```

The public option keeps its shape (`expiresAfter: number`, seconds). If a second anchor is ever offered, widening it to `number | { anchor, seconds }` stays backward compatible.

## Tests

The existing test asserted the wire format that the API rejects:

```ts
expect(multipart!.expires_after).toBe('3600');
```

It is replaced by one that pins the accepted encoding, plus a case for the absent option so the fields are not sent unrequested. Written test-first: red against the current implementation, green after the change. `packages/openai` passes in full — 24 files, 866 tests.


Azbesciak:
Superseded by #18236, which carries the same change to `openai-files.ts` and was prepared by the maintainers' automation after it reproduced #18223 on `main`. Closing this in favour of it.

The one thing this PR covered that #18236 does not is a test for the absent-option case (no `expiresAfter` → neither bracketed field is sent); I left it as a suggestion over there.

PR #18226, opened 2026-07-30
fix(openai): clarify unsupported Responses API endpoints

### LINKED ISSUE #17944: openai(modelId) targets the Responses API even with a custom baseURL, silently 404ing against OpenAI-compatible endpoints (NVIDIA NIM, etc.)
## Description

`createOpenAI({ baseURL, apiKey })` followed by the default callable (`provider(modelId)`) always constructs a **Responses API** model (`OpenAIResponsesLanguageModel`, hitting `${baseURL}/responses`), even when `baseURL` points at a third-party OpenAI-compatible endpoint that only implements Chat Completions (`/chat/completions`). This isn't documented anywhere I could find, and the failure mode is an opaque 404/"Not Found" with no indication that the SDK picked the wrong API surface.

The fix is simple once you know about it — call `.chat(modelId)` explicitly instead of `provider(modelId)` — but discovering that requires reading `openai-provider.ts` source, since `createLanguageModel` unconditionally delegates to `createResponsesModel`:

```ts
// src/openai-provider.ts (v2.0.79)
const createLanguageModel = (modelId) => {
  if (new.target) {
    throw new Error("The OpenAI model function cannot be called with the new keyword.");
  }
  return createResponsesModel(modelId); // <-- always Responses API
};
```

## Environment
- `@ai-sdk/openai`: 2.0.79
- `ai`: 5.0.108
- Endpoint: NVIDIA's NIM catalog (`https://integrate.api.nvidia.com/v1`), OpenAI-compatible, Chat Completions only

## Reproduction

```ts
import { createOpenAI } from "@ai-sdk/openai";
import { generateText } from "ai";

const openai = createOpenAI({
  apiKey: process.env.NVIDIA_API_KEY,
  baseURL: "https://integrate.api.nvidia.com/v1",
});

// 404s against NVIDIA's endpoint -- silently targets /responses
await generateText({ model: openai("meta/llama-3.1-70b-instruct"), prompt: "hi" });

// Works -- explicitly targets /chat/completions
await generateText({ model: openai.chat("meta/llama-3.1-70b-instruct"), prompt: "hi" });
```

## Expected vs actual
- **Expected**: either (a) the default `provider(modelId)` picks Chat Completions when a non-default `baseURL` is set (since third-party compatible endpoints overwhelmingly only implement Chat Completions, not the newer Responses API), or (b) the docs for `createOpenAI({ baseURL })` explicitly call out that callers must use `.chat(modelId)` for compatible endpoints, with the 404 failure mode mentioned.
- **Actual**: silent default to Responses API, no warning, no docs hit for this exact scenario that I could find.

## Notes
- I'm aware `@ai-sdk/openai-compatible` exists and is likely the "correct" package for this use case — but `createOpenAI({ baseURL })` from `@ai-sdk/openai` is an extremely common pattern in tutorials/community code for pointing at NVIDIA/Together/Groq/etc, so this footgun is probably hit often. A documentation callout (or at minimum a clearer error message when a Responses-API-shaped request 404s) would save others the same debugging trip.
- Happy to open a docs PR if a maintainer can confirm the intended fix (recommend `@ai-sdk/openai-compatible`, recommend `.chat()`, or change the default) — didn't want to guess at which direction you'd prefer.

Hardik180704:
## Background

The default model factory returned by `createOpenAI` uses the Responses API. When a custom OpenAI-compatible base URL only implements Chat Completions, a request to `/responses` currently fails with an opaque 404.

## Summary

- track whether the Responses model was created with a configured custom base URL
- augment 404 errors from that Responses endpoint with actionable guidance
- recommend `provider.chat(modelId)` and `@ai-sdk/openai-compatible` without changing API selection
- cover the custom-base-URL failure path with a regression test
- add a patch changeset for `@ai-sdk/openai`

## Contributor Credit

Thanks to @Zartharas for reporting the behavior and providing the NVIDIA NIM reproduction in #17944.

## End-to-End Verification

Configured `createOpenAI` with a custom base URL whose `/responses` endpoint returns 404 and confirmed the resulting error identifies the Responses API selection and recommends both supported alternatives. The request continues to target `/responses`, preserving existing behavior.

## Checklist

## Related Issues

Fixes #17944

### DIFF
--- a/.changeset/friendly-oranges-chat.md
+++ b/.changeset/friendly-oranges-chat.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+Add an actionable error hint when a custom OpenAI base URL does not support the Responses API.
--- a/packages/openai/src/openai-config.ts
+++ b/packages/openai/src/openai-config.ts
@@ -6,6 +6,7 @@ import type {
 export type OpenAIConfig = {
   provider: string;
   url: (options: { modelId: string; path: string }) => string;
+  isCustomBaseURL?: boolean;
   headers?: () => Record<string, string | undefined>;
   fetch?: FetchFunction;
   webSocket?: WebSocketConstructor;
--- a/packages/openai/src/openai-error.ts
+++ b/packages/openai/src/openai-error.ts
@@ -1,5 +1,9 @@
+import { APICallError } from '@ai-sdk/provider';
+import {
+  createJsonErrorResponseHandler,
+  type ResponseHandler,
+} from '@ai-sdk/provider-utils';
 import { z } from 'zod/v4';
-import { createJsonErrorResponseHandler } from '@ai-sdk/provider-utils';
 
 export const openaiErrorDataSchema = z.object({
   error: z.object({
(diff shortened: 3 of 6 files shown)


n1ckoates:
Original issue closed

PR #18307, opened 2026-08-01
fix(provider/fireworks): parse the object error envelope

shaper:
Fireworks returns errors as an **object**, but the schema expected `error` to be a bare string:

```ts
const fireworksErrorSchema = z.object({ error: z.string() });
```

```json
{"error":{"object":"error","type":"invalid_request_error","code":"invalid_request_error","message":"Extra inputs are not permitted, field: 'promptCacheKey'"},"request_id":"chatcmpl-…"}
```

So the parse always failed and the message fell back to the HTTP reason phrase. That is the quiet part: over HTTP/1.1 it produces `"Bad Request"` / `"Not Found"`, which read like genuine provider messages, and over **HTTP/2 there is no reason phrase at all**, so the message is `""`. Either way Fireworks' actual text is discarded.

## Verified against Fireworks

`examples/ai-functions/src/stream-text/fireworks/error-message.ts` (added here) provokes three real errors and prints `message` next to the raw body:

| case | before | after |
|---|---|---|
| unknown model (404) | `Not Found` | `Model not found, inaccessible, and/or not deployed` |
| invalid api key (401) | `Unauthorized` | `The API key you provided is invalid.` |

The example is the regression check: a `message` that is only a status phrase while `body` holds real text means the schema has drifted again.

## Shape

`error` becomes a union of the object envelope and the original bare string. Keeping the string means an envelope we have not seen degrades to the previous behavior rather than failing the parse outright — a too-narrow schema is exactly what caused this. Secondary fields (`object`, `type`, `param`, `code`) are handled loosely, matching `openaiCompatibleErrorDataSchema`; `code` allows string or number since Fireworks sends both (`"NOT_FOUND"`, `401`).

## Tests

5 cases over `errorStructure` (52 → 57): the object envelope, an envelope with `param: null` and a numeric `code`, unknown sibling keys like `request_id`, the bare-string fallback, and an object missing `message` still rejecting.

Reverting the schema fails 3 of the 5. The other two — bare-string and missing-`message` — pass either way by construction, since the old schema accepted strings and rejected every object; they are guards against future drift rather than proof of this fix.

### DIFF
--- a/examples/ai-functions/src/stream-text/fireworks/error-message.ts
+++ b/examples/ai-functions/src/stream-text/fireworks/error-message.ts
@@ -3,23 +3,6 @@ import { APICallError } from '@ai-sdk/provider';
 import { streamText } from 'ai';
 import { run } from '../../lib/run';
 
-/**
- * Checks that a Fireworks error message survives to the caller.
- *
- * Fireworks returns errors as `{"error":{"object","type","code","message"}}`.
- * When the provider's error schema does not match that shape the parse fails
- * and the message falls back to the HTTP reason phrase — "Bad Request" over
- * HTTP/1.1, and the empty string over HTTP/2, where there is no reason phrase.
- * Either way the actual cause is lost.
- *
- * Each case below provokes a different error. The printed message should be
- * Fireworks' own text, not a bare status phrase:
- *
- *   unknown field  400  Extra inputs are not permitted, field: 'notARealField'
- *   unknown model  404  Model not found, inaccessible, and/or not deployed
- *   bad api key    401  invalid api key
- */
-
 const MODEL = 'accounts/fireworks/models/kimi-k3';
 
 const CASES = [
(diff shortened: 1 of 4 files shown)


examples/ai-functions/src/stream-text/fireworks/error-message.ts
```diff
@@ -0,0 +1,58 @@
...
+        prompt: 'Reply with only the word: ok',
+        maxOutputTokens: 512,
+        ...(testCase.providerOptions
+          ? { providerOptions: testCase.providerOptions }
+          : {}),
+        onError: ({ error }) => {
+          streamError = error;
+        },
+      });
+      result.consumeStream();
```
kkdawkins, line 42: `consumeStream()` returns a promise; without `await`, the `streamError` check on the next line runs before anything is consumed, so every case prints "unexpectedly succeeded" and the captured error is lost. Should be `await result.consumeStream();`.

gr2m (approved):

PR #18309, opened 2026-08-01
feat (provider/xai): support new grok imagine video 1.5 caps

joshualipman123:
### Background

Grok Imagine Video 1.5 adds native 1080p output for text-to-video and image-to-video. Reference-to-video stays capped at 720p, so the two paths need different resolution ceilings.

While wiring this up, reference routing turned out to be broken on `main`: any non-empty `inputReferences` array selected reference-to-video, so an array holding only a video reference sent `reference_images: []` — a malformed request with no usable references.

### Summary

- Adds the `grok-imagine-video-1.5` model id.
- Native `1080p` for text-to-video and image-to-video. The standard `resolution: '1920x1080'` now maps to `1080p`, matching the mapping convention in the Google, Google Vertex, ByteDance, and Alibaba video providers.
- Reference-to-video is capped at `720p`; a `1080p` request in that mode is downgraded to `720p` with an `unsupported` warning.
- Fixes reference routing: reference-to-video is selected only when at least one reference survives as an image, and an empty image list is omitted rather than sent as `reference_images: []`. Non-image references are skipped with a warning naming `inputReferences`.

### Future Work

- Reference audio for reference-to-video (preset voice ids) — follow-up PR, branched off this one.
- `reference_images` also accepts a `file_id`; the provider only emits URLs and data URIs today.


shaper:
## Review — verified end-to-end

### ✅ `examples/ai-functions/src/generate-video/xai/t2v-1080p.ts` works against the live API

Rebuilt `@ai-sdk/xai` and ran the example:

```
✔ Generating xAI text-to-video at 1080p with grok-imagine-video-1.5... (46.6s)
Saved video to output/video-1786212804672-0.mp4 — 10.13 MB
```

ffprobe confirms **1920x1088, 8.04s** (1088 is h264 macroblock rounding of 1080p — normal, not an SDK issue). The `1080p` provider option passes through correctly to the real API.

Also verified locally: **430/430 `packages/xai` tests pass**, `tsc` clean for both `packages/xai` and `examples/ai-functions`, oxlint clean on the changed files, and CI is green.

### 🔴 Blocker: PR is not mergeable (`mergeable_state: dirty`, `rebaseable: false`)

The branch is based on a Jul 29 commit, and `main` has since refactored the video stack: `XaiVideoModel` on main now implements `VideoModelV4` with `doStart`/`doStatus` and a shared `buildRequestBody` (operations architecture), while this branch still implements the old polling `Experimental_VideoModelV4.doGenerate`. The PR needs a rebase that ports the changes onto the new architecture:

- model id + `1080p` / `1920x1080` mapping / R2V 720p downgrade → into `buildRequestBody`
- the `hasImageInputReference` routing fix → into the new mode resolution
- docs / example / changeset carry over as-is

### 🟡 Other considerations

1. **`1080p` isn't gated to the 1.5 model.** `resolution: '1080p'` (or SDK `1920x1080`) with the older `grok-imagine-video` passes 1080p straight to the API, which will presumably reject it — while the docs say 1080p requires `grok-imagine-video-1.5`. Consider a warning (or downgrade to 720p) when `modelId === 'grok-imagine-video'`, or at least a test pinning the intended pass-through behavior.

2. **Explicit R2V with zero usable image references silently degrades to T2V.** E.g. `mode: 'reference-to-video'` + audio-only `inputReferences` sends a plain generations request with no `reference_images`; the warning only says the reference "was ignored" — the user isn't told the request is no longer R2V. A dedicated "no usable image references, generating without references" warning (or an `InvalidArgumentError`) would be clearer. Pre-existing behavior, but this PR is touching exactly this routing logic, so it's a natural place to address it.

3. **Removed `xaiVideoModelOptions` zod union** — confirmed dead code (never exported from `index.ts`; validation always used the loose `runtimeSchema`), so removal is safe for a patch; only deep-importers of package internals could notice. The changeset doesn't mention it — acceptable.

4. **Nit:** `resolveReferences` returns a single-field `{ images }` object — a leftover from the removed audio support. Could just return `Array<{ url: string }> | undefined`.

### What's solid

- The R2V routing fix is well-tested from both entry points (provider-option `1080p` and SDK `1920x1080`), and audio/video-only references no longer flip requests into R2V or send `reference_images: []`.
- Warning messages distinguish "no image reference" from "ignored in this mode"; docs, changeset, and the capability-table footnote are consistent with the code.
- Example follows the repo layout conventions; the gateway catalog already lists `xai/grok-imagine-video-1.5`.

**Bottom line:** functionality verified working against the live API, but the PR must be rebased onto the new `VideoModelV4` operations architecture before it can merge — that's the blocker. Items 1 and 2 are worth addressing during that rebase.

shaper (approved):

PR #18317, opened 2026-08-02
fix(provider/baseten): opt-in native performance client, streamed usage, error envelopes

shaper:
Reworks the provider. Three fixes in the same file; the second two surfaced while verifying the first against a real deployment.

| | before | after |
|---|---|---|
| **native performance client** | hard dependency, a NAPI addon loaded by every consumer | not a dependency; opt in by injection |
| **streamed usage** | `includeUsage` unset, so streams reported `undefined` tokens | set on both chat paths |
| **error envelopes** | only the string shape parsed, so real messages became `Not Found` | both shapes parsed |

## Native performance client

Imported at module top level, so **every** consumer loaded a NAPI addon — 16 platform binary packages, ~5–16 MB — for a path only embeddings reach. It cannot load in edge runtimes, and bundlers cannot resolve its binaries, which blocked adopting this package in [vercel/ai-gateway#2938](https://github.com/vercel/ai-gateway/pull/2938).

**It is no longer a dependency.** Embeddings use plain HTTP against the deployment's OpenAI-compatible endpoint — what BEI serves with no additional settings ([docs](https://docs.baseten.co/examples/bei)), and what the code already fully constructed before overwriting its `doEmbed`. Opt back in by injection, following the existing `webSocket?: WebSocketConstructor` pattern (elevenlabs, cartesia, gateway):

```ts
import { PerformanceClient } from '@basetenlabs/performance-client';
const baseten = createBaseten({ modelURL, performanceClient: PerformanceClient });
```

The override discarded nearly everything `doEmbed` was handed, so deleting it makes `abortSignal`, per-call `headers`, `dimensions`/`user`, `config.fetch` (previously **no effect** here), real headers and warnings work for the first time.

**Two behaviour changes to weigh.** Requests now cap at 128 values — which is BEI's own server-side limit, confirmed live (see Verification), not a guess (`embedMany` splits above that, so only a direct `doEmbed` call with >128 throws), and `usage.tokens` comes from `prompt_tokens` rather than `total_tokens` — what the shared implementation already does and what `EmbeddingModelV4` specifies; normally identical values. The opt-in path is unchanged, including unbounded batching. Existing users change transport without changing code, hence `minor`.

## Streamed usage

`includeUsage` was never set, so `stream_options.include_usage` was omitted and servers returned no usage for streams — `streamText` reported every token field as `undefined` while `generateText` was correct. Live: with it, 16 / 2262; without, all `undefined`. Only observable on dedicated deployments; Model APIs returns usage anyway.

## Error envelopes

Baseten sends **two** shapes and the schema accepted one:

| surface | body | parsed |
|---|---|---|
| Model APIs | `{"error":"please check the model you provided"}` | yes |
| dedicated deployment | `{"error":{"message":…,"code":…,"param":…,"type":…}}` | **no** |

A failed parse falls back to the HTTP reason phrase, discarding the real cause — worse over HTTP/2, which has no reason phrase, so the message is empty. Live:

```
before  message : Not Found
after   message : The model `not-a-real-model-xyz` does not exist.
```

Now accepts both, mirroring #18307 for fireworks. A union rather than a swap because the shapes come from different layers — Baseten authenticates at its edge (string), the deployment validates the model (object). Embeddings are most affected, since they require a `modelURL` and so always hit a dedicated deployment. Tracked in AIG-394.

## Tests: 25 → 51

Worth flagging that the previous coverage was illusory — all 5 embedding tests asserted only constructor arguments, and **every one would have passed with the native path deleted outright**.

New unit tests cover both embedding paths, `includeUsage` on each chat path, and both error shapes; a new `baseten-embedding-model.test.ts` drives the real HTTP path via `createTestServer` (both `modelURL` forms, body, auth, headers, the 128 boundary, both envelopes). Both suites were confirmed to **fail** against reintroduced regressions.

## Verification

`test` (51, node + edge), `type-check`, `build`, `validate:docs` pass. `konsistent:check` fails identically with these changes stashed — pre-existing.

**Verified live** against two dedicated deployments — vLLM (Qwen3.5-4B) for chat, BEI (1024-dim) for embeddings:

| | result |
|---|---|
| `embed` / `embedMany`, plain HTTP | 1024-dim vectors, `usage.tokens` populated, real warnings |
| `embedMany` above the cap | 300 values → 300 embeddings, chunked and parallelised |
| **the 128 cap** | **the server enforces exactly 128** — 256 values returns `413 batch size 256 > maximum allowed batch size 128`. It is also the native client's own documented default `batchSize`, so the cap mirrors both rather than guessing. |
| `prompt_tokens` vs `total_tokens` | wire carries `{"prompt_tokens":4,"total_tokens":4}` — both present and identical, as claimed |
| object envelope, embeddings path | `batch size 200 > maximum allowed batch size 128` surfaces verbatim, where the old schema would have given the 413 reason phrase |
| streamed usage, both envelopes | on chat, per the sections above |
| opt-in native path | both examples produce identical vectors and identical usage to plain HTTP |
| `maxEmbeddingsPerCall: Infinity` on that path | `embedMany(1000)` in a single `doEmbed` succeeds — the client re-chunks at its own 128 default, so handing it everything is correct |

A note on the example devDep, pinned to `0.0.10`: **every `@basetenlabs/performance-client` release after `0.0.10` is mis-published.** None of them declare the 16 platform `optionalDependencies`, so `require` throws `Cannot find native binding` on every platform — `npm i @basetenlabs/performance-client@0.1.2` reports `added 1 package` and no binary. The platform artifacts *are* built and on npm, so `napi prepublish` is uploading them without writing them back into the main manifest. `0.0.10` (2025-08-07) is the last installable release; the current `latest`, `0.1.2`, has been broken since 2026-01-28 — likely unreported this long because it only fails on a cold install. Filed upstream as [basetenlabs/truss#2590](https://github.com/basetenlabs/truss/issues/2590). This affects only the example, not the provider, and is one more argument for keeping a native addon out of the dependency graph.


examples/ai-functions/package.json
```diff
@@ -64,6 +64,7 @@
     "@ai-sdk/vercel": "workspace:*",
     "@ai-sdk/voyage": "workspace:*",
     "@ai-sdk/xai": "workspace:*",
+    "@basetenlabs/performance-client": "0.0.10",
```
shaper, line 67: @gr2m should I perhaps put this in devDependencies or otherwise make it optional? they're only here to allow running the two `performance-client.ts` examples.

aayush-kapoor (approved):

PR #18326, opened 2026-08-02
fix(ai): collect trailing local tool approvals

### LINKED ISSUE #17033: streamText silently skips local tool approvals when a later context message follows the approval response

iroiro147:
## collect trailing local tool approvals

Collect local tool approval responses from the latest assistant approval turn even when user or system context follows the approval response.

- Skip approvals that already have tool results
- Avoid replaying approvals behind a later assistant response
- Align missing-result validation with the same latest-assistant approval boundary

Fixes #17033

Rebased onto latest main (PR #17036 was stale by 266 commits).

Tests
- pnpm --filter ai type-check ✓
- pnpm --filter ai test:node -- collect-tool-approvals.test.ts ✓ (all 3327 pass)
- pnpm --filter ai test:node -- stream-text.test.ts ✓
- pnpm --filter ai test:node -- convert-to-language-model-prompt.validation.test.ts ✓

### DIFF
--- a/packages/ai/src/generate-text/collect-tool-approvals.ts
+++ b/packages/ai/src/generate-text/collect-tool-approvals.ts
@@ -7,6 +7,7 @@ import type {
 } from '@ai-sdk/provider-utils';
 import { InvalidToolApprovalError } from '../error/invalid-tool-approval-error';
 import { ToolCallNotFoundForApprovalError } from '../error/tool-call-not-found-for-approval-error';
+import { getLatestAssistantApprovalTurn } from '../util/get-latest-assistant-approval-turn';
 import type { TypedToolCall } from './tool-call';
 
 export type CollectedToolApprovals<TOOLS extends ToolSet> = {
@@ -16,9 +17,20 @@ export type CollectedToolApprovals<TOOLS extends ToolSet> = {
   existingToolResult?: ToolResultPart;
 };
 
+function createEmptyToolApprovals<TOOLS extends ToolSet>(): {
+  approvedToolApprovals: Array<CollectedToolApprovals<TOOLS>>;
+  deniedToolApprovals: Array<CollectedToolApprovals<TOOLS>>;
+} {
+  return {
+    approvedToolApprovals: [],
+    deniedToolApprovals: [],
+  };
+}
+
 /**
- * If the last message is a tool message, this function collects all tool approvals
- * from that message.
+ * Collects unresolved tool approvals for the latest assistant turn. Approval
+ * responses may be followed by non-assistant context messages, but are not
+ * replayed once a later assistant response exists.
  */
 export function collectToolApprovals<TOOLS extends ToolSet>({
   messages,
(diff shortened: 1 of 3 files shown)


iroiro147:
Ping — this one and #18327 are still `MERGEABLE` with no conflicts, and neither
has been looked at since early August.

Thanks for taking #17047 last week, @n1ckoates. If the older batch is worth
reviving I am glad to rebase it; eight of mine here have drifted behind `main`
and I would rather rebase only the ones you would actually want, than push ten
branches at your CI. Say which, or say none, and I will close the rest.

PR #18332, opened 2026-08-03
fix(anthropic): handle out-of-spec message_start events in response streams

### LINKED ISSUE #18331: @ai-sdk/anthropic: message_start for a different message while the previous message is still open is silently merged into one generation, corrupting thinking blocks and failing all subsequent requests
### Description

When a streaming intermediary loses its upstream Anthropic connection mid-generation and retries, the retried generation's SSE events can be spliced into the same HTTP response body. The stream then contains a second `message_start` (with a new message id) while the first message is still open. We observed this in production behind Google Vertex via `@ai-sdk/google-vertex/anthropic`: the first generation's `tool_use` block was cut mid `input_json_delta` and never received `content_block_stop`, and a full second generation followed on the same stream.

The `@ai-sdk/anthropic` parser accepts this silently. The `message_start` handler tracks no message lifecycle state and only overwrites usage counters. The unterminated `tool_use` block is dropped without a trace, because a `tool-call` part is only emitted on `content_block_stop`. Both generations' signed thinking blocks stream through as if they were one response, and `streamText` records them into a single step.

That step cannot be replayed. The assistant turn now contains a signed thinking block from a generation that, from Anthropic's perspective, never completed. Every subsequent request in the tool loop is rejected with:

```
AI_APICallError: messages.N.content.M: `thinking` or `redacted_thinking` blocks in the
latest assistant message cannot be modified. These blocks must remain as they were in
the original response.
```

This is a non-retryable 400, so `maxRetries` does not engage and the run ends. In our incident this happened 27 steps into a 34-minute agentic run with adaptive thinking (about $20 of API spend), which then failed without producing its deliverable. The corruption is persistent: once the merged step is in the run's history, every later step fails identically.

A related, milder shape exists that the fix must not break: some Anthropic-compatible gateways duplicate the `message_start` frame of a single response (same message id). Today that duplicate produces a spurious second `response-metadata` part and overwrites usage counters, but the stream otherwise works.

### Root cause

`packages/anthropic/src/anthropic-language-model.ts`, `doStream` transform, `case 'message_start'`.

Multiple `message_start`/`message_stop` sequences per stream are legitimate (programmatic tool calling streams several complete messages per response). What cannot occur in a valid response is a `message_start` while the previous message is still open. The parser does not detect this. By the time the second `message_start` arrives, the earlier generation's completed blocks have already been emitted downstream, so the stream cannot be repaired at this layer, but it can be failed explicitly instead of merging two generations into one.

### Reproduction

Raw SSE sequence, same shape as the incident. Also included as failing tests in the linked PR; both fail on current `main`.

```
message_start (msg_first)
content_block_start    thinking
thinking_delta / signature_delta / content_block_stop
content_block_start    tool_use (toolu_first)
input_json_delta       {"value":"Spark        <- stream cut mid-input, no content_block_stop
message_start (msg_second)                     <- spliced retry, previous message still open
content_block_start    thinking ... (complete)
content_block_start    tool_use (toolu_second) ... (complete)
message_delta stop_reason=tool_use / message_stop
```

For the benign gateway shape: two identical `message_start` frames (same id) followed by a normal single-message stream.

### Actual

The spliced stream emits `msg_second`'s `response-metadata`, both thinking blocks, and `toolu_second` as one merged generation, with no error. The next request built from this history fails deterministically with the 400 above. The same-id duplicate emits a spurious second `response-metadata` part.

### Expected

A duplicated `message_start` frame for the already-open message is ignored. A `message_start` for a different message while the previous message is still open fails the stream with an explicit error part (`InvalidResponseDataError`), so the step errors cleanly and no corrupted history is recorded. Programmatic tool calling (`message_stop` followed by `message_start`) is unaffected.

### Corroboration

This is not specific to one intermediary. The Anthropic streaming spec allows exactly one `message_start` per response, but gateway implementations have been documented violating it: maximhq/bifrost#4556 (two `message_start` frames, same message id, on the Anthropic passthrough) and BerriAI/litellm#32479 (duplicate `message_start` per bridged stream, since fixed). The spliced-generation variant reported here was reconstructed from the persisted parts of a production incident: an unterminated `tool_use` input frozen mid-JSON, followed by a regenerated near-duplicate thinking block and a second tool call with a different id, with no step boundary between them. We did not capture the raw wire stream. Handling of Vertex's `vertex_event` (#13576) is precedent for accommodating intermediary stream behavior in this package.

### Environment

`ai` 7.0.41, `@ai-sdk/anthropic` 4.0.23 (via `@ai-sdk/google-vertex` 5.0.34). Reproduced against current `main` (`@ai-sdk/anthropic` 4.0.27).

ro-hansolo:
## Background

Fixes #18331.

A valid Anthropic response stream contains exactly one `message_start` per message, but intermediaries violate this in two documented ways:

1. An intermediary that loses its upstream connection mid-generation and retries can splice a second generation into the same HTTP response body: a `message_start` with a new message id arrives while the previous message is still open (observed in production behind Google Vertex; the first generation's `tool_use` was cut mid `input_json_delta`).
2. Some Anthropic-compatible gateways duplicate the `message_start` frame of a single response, same message id (maximhq/bifrost#4556, BerriAI/litellm#32479).

The parser currently accepts both silently. For shape 1, both generations' signed thinking blocks are merged into one recorded step; the step cannot be replayed, and every subsequent request in the tool loop fails with Anthropic's non-retryable 400 "`thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified". For shape 2, the duplicate frame emits a spurious second `response-metadata` part and overwrites usage counters.

## Summary

`doStream` now tracks the open message id (set on `message_start`, cleared on `message_stop`):

- A `message_start` for the **same** message id while it is open is ignored (shape 2, benign duplicate).
- A `message_start` for a **different** message id while the previous message is still open fails the stream with an explicit `InvalidResponseDataError` part and drops all subsequent events (shape 1). Earlier blocks may already have been consumed downstream, so the stream cannot be repaired at this layer; failing explicitly prevents the corrupted history from being recorded.

Legitimate `message_start`/`message_stop` sequences (programmatic tool calling) are unaffected: the existing multi-message fixtures pass unchanged.

## Verification

- Two new tests in `anthropic-language-model.test.ts` using raw SSE stream chunks: the spliced-generation stream and the same-id duplicate. Both fail on current `main` (the first shows the silent merge, the second shows the spurious `response-metadata`).
- Full package suite passes: 474/474 in both node and edge configs.
- `ultracite check` and `tsc --build` clean.

## Checklist

- [x] Patch changeset added (`.changeset/fail-spliced-anthropic-streams.md`)
- [x] Tests added, failing on `main`, passing with the fix
- [x] All commits are signed

## Related Issues

Fixes #18331

### DIFF
--- a/.changeset/fail-spliced-anthropic-streams.md
+++ b/.changeset/fail-spliced-anthropic-streams.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Handle out-of-spec message_start events in the response stream instead of silently corrupting the recorded generation. A duplicated message_start frame for the message that is already open (emitted by some Anthropic-compatible gateways) is now ignored instead of producing a spurious response-metadata part and overwriting usage counters. A message_start for a different message while the previous message is still open (two generations spliced into one stream, e.g. by an intermediary retrying its upstream connection mid-response) now fails the stream with an explicit error; previously both generations' thinking blocks were merged into one recorded step, and Anthropic rejected every subsequent request in the tool loop with "`thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified". Legitimate message_start/message_stop sequences (programmatic tool calling) are unaffected.
--- a/packages/anthropic/src/anthropic-language-model.test.ts
+++ b/packages/anthropic/src/anthropic-language-model.test.ts
@@ -7386,6 +7386,186 @@ describe('AnthropicLanguageModel', () => {
       `);
     });
 
+    it('should ignore a duplicated message_start frame for the same message', async () => {
+      // some gateways duplicate the message_start frame of a single
+      // response (same message id) on their Anthropic passthrough.
+      server.urls['https://api.anthropic.com/v1/messages'].response = {
+        type: 'stream-chunks',
+        chunks: [
+          `data: {"type":"message_start","message":{"id":"msg_dup","type":"message","role":"assistant","content":[],"model":"claude-3-haiku-20240307","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":17,"output_tokens":1}}}\n\n`,
+          `data: {"type":"message_start","message":{"id":"msg_dup","type":"message","role":"assistant","content":[],"model":"claude-3-haiku-20240307","stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":17,"output_tokens":1}}}\n\n`,
+          `data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}\n\n`,
+          `data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello, World!"}}\n\n`,
+          `data: {"type":"content_block_stop","index":0}\n\n`,
+          `data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":227}}\n\n`,
+          `data: {"type":"message_stop"}\n\n`,
+        ],
+      };
+
+      const { stream } = await model.doStream({
+        prompt: TEST_PROMPT,
+      });
+
+      expect(await convertReadableStreamToArray(stream)).toMatchInlineSnapshot(`
+        [
+          {
+            "type": "stream-start",
+            "warnings": [],
+          },
+          {
+            "id": "msg_dup",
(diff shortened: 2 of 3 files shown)


gr2m:
fixed by https://github.com/vercel/ai/pull/18402

PR #18337, opened 2026-08-03
fix(openai): handle missing reasoning state in streamed responses

### LINKED ISSUE #18270: TypeError: Cannot read properties of undefined (reading 'summaryParts') when streaming reasoning summaries via GitHub Copilot's Responses API
### Description

When streaming from GitHub Copilot's Responses-API-compatible endpoint using a reasoning model (`gpt-5.3-codex`), the stream crashes with:

TypeError: Cannot read properties of undefined (reading 'summaryParts')

This was originally reported downstream in a consumer project (Nano-Collective/nanocoder#719), but the actual bug is here in `@ai-sdk/openai`'s Responses API streaming handler. The same underlying crash has independently surfaced in other downstream projects as well: Nano-Collective/nanocoder#576 (GitHub Copilot `gpt-5.5`) and anomalyco/opencode#21346 (Copilot models via a LiteLLM proxy), which suggests this isn't specific to one provider setup but a general gap in how `@ai-sdk/openai` handles non-OpenAI Responses-API event ordering.

## Root cause

In `packages/openai/src/responses/openai-responses-language-model.ts`, `activeReasoning[item.id]` is only populated when a `response.output_item.added` event arrives with `item.type === 'reasoning'`:

```ts
} else if (
  isResponseOutputItemAddedChunk(value) &&
  value.item.type === 'reasoning'
) {
  activeReasoning[value.item.id] = {
    encryptedContent: value.item.encrypted_content,
    summaryParts: { 0: 'active' },
  };
  ...
}
```

Three separate places later assume this entry already exists, without guarding against it being missing:

**1. `response.output_item.done` (`item.type === 'reasoning'`)** — this one doesn't even use a non-null assertion, it's a plain property access, so it throws immediately if the entry is missing:

```ts
const activeReasoningPart = activeReasoning[value.item.id];
const summaryPartIndices = Object.entries(
  activeReasoningPart.summaryParts,   // <- throws here
)
```

**2. `response.reasoning_summary_part.added`:**

```ts
const activeReasoningPart = activeReasoning[value.item_id]!;
activeReasoningPart.summaryParts[value.summary_index] = ...
```

**3. `response.reasoning_summary_part.done`** (both branches):

```ts
activeReasoning[value.item_id]!.summaryParts[value.summary_index] = 'concluded';
...
activeReasoning[value.item_id]!.summaryParts[value.summary_index] = 'can-conclude';
```

If a provider's stream emits any of these reasoning-summary/reasoning-done events without first emitting a matching `response.output_item.added` (type `reasoning`) — or emits it with a different `item_id` than the summary-part events reference — `activeReasoning[value.item_id]` is `undefined`, and any of the three sites above throws.

GitHub Copilot's proxy for `gpt-5.3-codex` appears to do exactly this: it sends reasoning-summary/reasoning-done stream events without the expected preceding item-added event (or with a mismatched item id), which triggers the crash below.

**Expected:** the response streams normally, including any reasoning summary content.

**Actual:** the stream throws partway through:

Chat request failed {
model: 'gpt-5.3-codex',
error: "Cannot read properties of undefined (reading 'summaryParts')",
errorName: 'TypeError',
errorType: 'TypeError'
}

This is consistently reproducible with this provider/model combination.

## Suggested fix direction

Guard all three sites above against a missing `activeReasoning` entry, e.g. lazily initialize it instead of asserting non-null / indexing directly:

```ts
const activeReasoningPart =
  activeReasoning[value.item_id] ??
  (activeReasoning[value.item_id] = { summaryParts: {} });
```

The `output_item.done` handler needs the same guard, plus a sensible fallback for `summaryPartIndices` (e.g. treating a missing entry as having no active/can-conclude parts) since it currently isn't even wrapped in a non-null assertion.

This would make the handler tolerant of providers whose Responses-API implementation doesn't perfectly mirror OpenAI's own event ordering, while preserving existing behavior when the item-added event is present as expected.

## Related

- Nano-Collective/nanocoder#719 (original downstream report)
- Nano-Collective/nanocoder#576 (same crash, GitHub Copilot `gpt-5.5`)
- anomalyco/opencode#21346 (same crash, Copilot via LiteLLM proxy)

### Reproduction

**Environment**

- OS: Fedora Linux 44
- Node: v24.18.0
- Provider: GitHub Copilot (Responses-API-compatible endpoint, `https://api.githubcopilot.com`)
- Model: `gpt-5.3-codex`

**Steps**

1. Configure a Copilot-backed OpenAI-compatible provider using `createOpenAI({ baseURL: "https://api.githubcopilot.com", ... })` (or equivalent) pointed at `gpt-5.3-codex`.
2. Call `streamText` with a normal prompt (no special reasoning options required to trigger it).
3. Consume the stream.

I don't have a minimal repo packaged up yet, but happy to open a PR with a failing-test reproduction and/or the fix above if that's useful — let me know if a specific test fixture format is preferred.

### AI SDK Version

@ai-sdk/openai: ^3.0.85
ai: 6.0.193

### Code of Conduct

mussaratshamsher:
## Summary

Fixes a crash when streaming reasoning summaries from Responses-API-compatible providers that emit reasoning summary events before `response.output_item.added`.

The current implementation assumes `activeReasoning[item_id]` has already been initialized. Providers such as GitHub Copilot can emit `response.reasoning_summary_part.*` events without first sending the corresponding `response.output_item.added` event, causing:

```text
TypeError: Cannot read properties of undefined (reading 'summaryParts')
```

This PR makes the reasoning state initialization defensive by lazily creating the missing state whenever it is first accessed. It also emits the initial `reasoning-start` event when the first summary part arrives without a preceding `response.output_item.added`, preserving the expected stream behavior while maintaining compatibility with the existing OpenAI event flow.

## Changes

- Lazily initialize missing reasoning state in all reasoning event handlers.
- Handle reasoning summary events that arrive before `response.output_item.added`.
- Emit the initial `reasoning-start` event when necessary.
- Preserve existing behavior for providers that follow OpenAI's event ordering.
- Add a regression test covering the missing `response.output_item.added` scenario.

## Testing

- ✅ Added a regression test reproducing the reported issue.
- ✅ Targeted regression test passes.
- ✅ Full `@ai-sdk/openai` test suite passes (867 tests).
- ✅ Type checking passes.

Fixes #18270

### DIFF
--- a/packages/openai/src/responses/openai-responses-language-model.ts
+++ b/packages/openai/src/responses/openai-responses-language-model.ts
@@ -2128,7 +2128,14 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
                   } satisfies InferSchema<typeof shellOutputSchema>,
                 });
               } else if (value.item.type === 'reasoning') {
-                const activeReasoningPart = activeReasoning[value.item.id];
+                // lazily initialize the reasoning part state in case the provider
+                // did not emit a matching response.output_item.added event first:
+                const activeReasoningPart =
+                  activeReasoning[value.item.id] ??
+                  (activeReasoning[value.item.id] = {
+                    encryptedContent: undefined,
+                    summaryParts: {},
+                  });
 
                 // get all active or can-conclude summary parts' ids
                 // to conclude ongoing reasoning parts:
@@ -2298,20 +2305,37 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
                 logprobs.push(value.logprobs);
               }
             } else if (value.type === 'response.reasoning_summary_part.added') {
-              // the first reasoning start is pushed in isResponseOutputItemAddedReasoningChunk
-              if (value.summary_index > 0) {
-                const activeReasoningPart = activeReasoning[value.item_id]!;
+              // lazily initialize the reasoning part state in case the provider
+              // did not emit a response.output_item.added event first:
+              const activeReasoningPart =
+                activeReasoning[value.item_id] ??
+                (activeReasoning[value.item_id] = {
+                  encryptedContent: undefined,
+                  summaryParts: {},
+                });
 
+              const isNewSummaryPart = !(
+                value.summary_index in activeReasoningPart.summaryParts
+              );
+
+              // the first reasoning-start is normally pushed from the
+              // response.output_item.added (type: 'reasoning') handler above. if that
+              // event was never received (e.g. providers that emit summary parts
+              // without a preceding item-added event), emit a reasoning-start here
+              // for whatever summary part is being added.
+              if (isNewSummaryPart) {
                 activeReasoningPart.summaryParts[value.summary_index] =
                   'active';
 
-                // since there is a new active summary part, we can conclude all can-conclude summary parts
+                // since there is a new active summary part, we can conclude all
+                // can-conclude summary parts
                 for (const summaryIndex of Object.keys(
                   activeReasoningPart.summaryParts,
                 )) {
                   if (
+                    summaryIndex !== String(value.summary_index) &&
                     activeReasoningPart.summaryParts[summaryIndex] ===
-                    'can-conclude'
+                      'can-conclude'
                   ) {
                     controller.enqueue({
                       type: 'reasoning-end',
@@ -2322,6 +2346,7 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
                         } satisfies ResponsesReasoningProviderMetadata,
                       },
                     });
+
                     activeReasoningPart.summaryParts[summaryIndex] =
                       'concluded';
                   }
@@ -2334,8 +2359,7 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
                     [providerOptionsName]: {
                       itemId: value.item_id,
                       reasoningEncryptedContent:
-                        activeReasoning[value.item_id]?.encryptedContent ??
-                        null,
+                        activeReasoningPart.encryptedContent ?? null,
                     } satisfies ResponsesReasoningProviderMetadata,
                   },
                 });
@@ -2354,6 +2378,12 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
             } else if (value.type === 'response.reasoning_summary_part.done') {
               // when OpenAI stores the message data, we can immediately conclude the reasoning part
               // since we do not need to send the encrypted content.
+              const reasoning =
+                activeReasoning[value.item_id] ??
+                (activeReasoning[value.item_id] = {
+                  encryptedContent: undefined,
+                  summaryParts: {},
+                });
               if (store) {
                 controller.enqueue({
                   type: 'reasoning-end',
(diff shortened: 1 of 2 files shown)


packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -2298,20 +2305,37 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
                 logprobs.push(value.logprobs);
               }
             } else if (value.type === 'response.reasoning_summary_part.added') {
-              // the first reasoning start is pushed in isResponseOutputItemAddedReasoningChunk
-              if (value.summary_index > 0) {
-                const activeReasoningPart = activeReasoning[value.item_id]!;
+              // lazily initialize the reasoning part state in case the provider
```
A-S-Manoj, line 2308: nit: this lazy-init pattern is duplicated 3x in this file (here, at line ~2308, and in `reasoning_summary_part.done`). Not blocking — just flagging for a possible follow-up: a shared `getOrCreateReasoningPart()` helper would remove the duplication.

packages/openai/src/responses/openai-responses-language-model.ts
```diff
@@ -2298,20 +2305,37 @@ export class OpenAIResponsesLanguageModel implements LanguageModelV4 {
...
                 activeReasoningPart.summaryParts[value.summary_index] =
                   'active';
 
-                // since there is a new active summary part, we can conclude all can-conclude summary parts
+                // since there is a new active summary part, we can conclude all
+                // can-conclude summary parts
                 for (const summaryIndex of Object.keys(
                   activeReasoningPart.summaryParts,
                 )) {
                   if (
```
A-S-Manoj, line 2335: nit: `summaryIndex !== String(value.summary_index)` is dead code — `summaryParts[value.summary_index]` was just set to `'active'` on line 2328, so it can never also equal `'can-conclude'` in this same iteration.

```suggestion
                  if (
                    activeReasoningPart.summaryParts[summaryIndex] ===
                      'can-conclude'
                  ) {
```

A-S-Manoj (commented):
Reproduced the crash on main, confirmed this PR fixes it, and all 289 tests pass including the new regression test. LGTM.

Left two non-blocking comments — a dead-code cleanup suggestion and a note on possible future dedup. Neither affects correctness.

packages/openai/src/responses/openai-responses-language-model.ts
mussaratshamsher: Thanks for the suggestion — agreed, a shared getOrCreateReasoningPart() helper would remove the duplication across the three spots (here, at ~line 2307, and in reasoning_summary_part.done). I'd like to keep this PR scoped to the streaming crash fix and tackle that refactor in a follow-up PR so it gets its own focused review.

packages/openai/src/responses/openai-responses-language-model.ts
mussaratshamsher: Good catch, thanks! Fixed in 87b21ab — removed the redundant summaryIndex !== String(value.summary_index) check. It was already implied: summaryParts[value.summary_index] is set to 'active' right before this loop runs, so it can never also read 'can-conclude' for that same index in the same pass.

gr2m:
done via https://github.com/vercel/ai/pull/18548

PR #18346, opened 2026-08-03
feat: batch

n1ckoates:
## Background

OpenAI, Anthropic, and Google all have Batch APIs to run many requests asynchronously in a large batch, and get results later. AI SDK did not support this.

## Summary

Added experimental batch APIs.

- `startTextBatch` accepts a model and an array of `requests` and returns a reference to that batch.
- `getBatchStatus` accepts the batch reference and returns the current status.
- `getBatchResults` accepts the batch reference and returns a stream of results if the batch is complete.

The user flow is:
- Create a batch with `startTextBatch`
- Persist the batch reference - either in DB or using something like Vercel Workflow
- Poll `getBatchStatus` with that reference
- When complete, accept the results with `getBatchResults`

Batches can take many hours to complete, so the batch reference needs to persist across process restarts. Because of this, AI SDK does not control the polling itself - users are expected to persist the batch and poll themselves using the batch APIs.

```ts
const model = openai('gpt-5.6');

const batch = await startTextBatch({
  model,
  requests: [
    {
      id: 'capital-france',
      prompt: 'What is the capital of France?',
    },
    {
      id: 'capital-germany',
      prompt: 'What is the capital of Germany?',
    },
  ],
});

while (true) {
  const status = await getBatchStatus({ model, batch });

  if (status.status !== 'pending') {
    break;
  }

  await new Promise(resolve => setTimeout(resolve, 60_000));
}

for await (const result of getBatchResults({ model, batch })) {
  console.log(result);
}
``` 

Supported models implement `BatchLanguageModelV4`, which is an extension of `LanguageModelV4` with added methods:
- `doCreateBatch`
- `doGetBatchStatus`
- `doGetBatchResults`

## End-to-End Verification

Have not e2e verified yet because I haven't added provider implementations - will do that in a separate PR later.

## Checklist

## Future Work

* Add provider support (OpenAI, Anthropic, Google, and AI Gateway)
* Add APIs for other request types (image, video, etc)


packages/provider/src/batch/v4/batch-v4.ts
```diff
@@ -0,0 +1,104 @@
...
+    readonly total: number;
+    readonly pending: number;
+    readonly completed: number;
+    readonly failed: number;
+  };
+  readonly error?: BatchV4Error;
+  readonly createdAt?: string;
+  readonly expiresAt?: string;
+  readonly providerMetadata?: SharedV4ProviderMetadata;
+};
```
gr2m, line 33: we might need to handle canceled/expired separate from failed, but can follow up

gr2m (approved):

PR #18349, opened 2026-08-03
feat (provider/xai): add reference audio for reference-to-video

joshualipman123:
### Background

Grok Imagine Video 1.5 can give a reference-to-video subject a voice. The API takes **preset voices by id** — `reference_audios: [{ "voice_id": "eve" }]`, up to 3 — and explicitly does not accept uploaded audio clips.

### Summary

- Adds the `referenceVoiceIds` provider option: up to 3 xAI preset voice ids, sent as `reference_audios: [{ voice_id }]`.
- Only applies to reference-to-video. Supplied in any other mode, it is omitted with an `unsupported` warning naming `referenceVoiceIds`.
- Ids pass through unmodified — the API treats them case-insensitively and returns `400` with the valid roster for an unknown id, so no allowlist is encoded in the schema.
- Nothing is routed through the top-level `inputReferences` option, which stays image-only. Reference audio is not a file input.

### DIFF
--- a/content/providers/01-ai-sdk-providers/01-xai.mdx
+++ b/content/providers/01-ai-sdk-providers/01-xai.mdx
@@ -1235,6 +1235,51 @@ media type (for example a video or audio clip) is ignored with a warning, and a
 reference with no media type is treated as an image. If no image reference
 remains, reference-to-video is not selected and no reference images are sent.
 
+#### Reference Audio
+
+Reference-to-video can also give the subject a voice. `referenceVoiceIds` takes
+up to 3 xAI **preset** voice ids — you cannot upload your own audio clips.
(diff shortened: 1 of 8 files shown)


content/providers/01-ai-sdk-providers/01-xai.mdx
```diff
@@ -1195,14 +1195,7 @@ const { video } = await generateVideo({
 });
 ```
 
-Use `<IMAGE_1>`, `<IMAGE_2>`, etc. in your prompt to reference specific images, in the order the images are passed. Up to 7 reference images are supported per request.
-
-<Note>
```
shaper, line 1200: removed because no longer true?

shaper (approved):

PR #18371, opened 2026-08-03
fix(ai): release async iterator reader after read errors

### LINKED ISSUE #18370: AsyncIterableStream stays locked after the source stream errors
### Description

Proposed fix: #18371

When async iteration starts, `AsyncIterableStream` gets a stream reader. The helper already releases that reader when iteration finishes normally, stops early, or `iterator.throw()` is called: [existing cleanup paths](https://github.com/vercel/ai/blob/861d42334474f5e411a4b58b741f6ab3c7fb86f3/packages/ai/src/util/async-iterable-stream.ts#L43-L110).

However, when the source stream errors, [`reader.read()` rejects before cleanup runs](https://github.com/vercel/ai/blob/861d42334474f5e411a4b58b741f6ab3c7fb86f3/packages/ai/src/util/async-iterable-stream.ts#L76-L89).

Current behavior:

```text
source stream errors
        ↓
reader.read() rejects
        ↓
next() exits before cleanup
        ↓
caller receives the original error
        ↓
reader lock remains held
        ↓
stream stays locked
```

I think the failed read should instead follow the helper's existing cleanup path:

```text
source stream errors
        ↓
reader.read() rejects
        ↓
cleanup(false)
        ↓
reader lock is released without cancellation
        ↓
caller receives the original error
        ↓
later calls on that iterator return done: true
```

The stream would still be errored; releasing the reader only removes the leftover lock.

A related prior issue fixed a different cleanup problem in the same helper after normal completion:

- https://redirect.github.com/vercel/ai/issues/11715

### Reproduction

A reproduction against `main` at `861d42334474f5e411a4b58b741f6ab3c7fb86f3`:

```ts
import { createAsyncIterableStream } from './packages/ai/src/util/async-iterable-stream';

let controller!: ReadableStreamDefaultController<string>;

const stream = createAsyncIterableStream(
  new ReadableStream<string>({
    start(controllerParam) {
      controller = controllerParam;
    },
  }),
);

const iterator = stream[Symbol.asyncIterator]();
const pendingRead = iterator.next();
const sourceError = new Error('source failed');

controller.error(sourceError);

const observedError = await pendingRead.catch(error => error);
console.log('original error preserved:', observedError === sourceError);
console.log('stream locked:', stream.locked);

stream.getReader();
```

Observed on upstream baseline:

```text
original error preserved: true
stream locked: true
TypeError: Invalid state: ReadableStream is locked
```

Regression test:

- [async-iterable-stream-read-error.test.ts](https://github.com/teamleaderleo/ai/blob/fd6335acd351b4c00824d8b2e68d1fab40053c86/packages/ai/src/util/async-iterable-stream-read-error.test.ts#L1-L125)

The test covers three source-error scenarios for both `createAsyncIterableStream()` and `asAsyncIterableStream()`. A local Vitest run against the unfixed baseline produced six failures because the stream remained locked.

Image

### AI SDK Version

```text
ai: 7.0.48
```

### Code of Conduct

teamleaderleo:
## Background

Closes #18370.

Current behavior:

```text
source stream errors
        ↓
reader.read() rejects
        ↓
next() exits before cleanup
        ↓
caller receives the original error
        ↓
stream stays locked
```

This change makes the rejected read follow the existing cleanup path:

```text
reader.read() rejects
        ↓
cleanup(false)
        ↓
release the reader without cancellation
        ↓
throw the original error
        ↓
later iterator calls return done: true
```

The stream remains errored, and a newly acquired reader still receives the stored source error.

## Summary

- handle rejected reads with `cleanup(false)`;
- add regression coverage for both async-iterable stream helpers;
- add a patch changeset for `ai`.

## End-to-End Verification

The regression test was run locally against the unfixed upstream baseline at `861d42334474f5e411a4b58b741f6ab3c7fb86f3`.

All six cases failed across both helper implementations. The saved terminal output shows that the caller received the source error but the stream stayed locked.

With this change, the same cases pass.

The rebased signed candidate head is `fd6335acd351b4c00824d8b2e68d1fab40053c86`.

## Checklist

## Related Issues

- Closes #18370
- Related cleanup precedent: https://redirect.github.com/vercel/ai/issues/11715

### DIFF
--- a/.changeset/quiet-stream-errors-release.md
+++ b/.changeset/quiet-stream-errors-release.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Release async-iterator reader locks when upstream stream reads fail.
--- a/packages/ai/src/util/async-iterable-stream-read-error.test.ts
+++ b/packages/ai/src/util/async-iterable-stream-read-error.test.ts
@@ -1,0 +1,125 @@
+import { describe, expect, it } from 'vitest';
+import {
+  asAsyncIterableStream,
+  createAsyncIterableStream,
+  type AsyncIterableStream,
+} from './async-iterable-stream';
+
+type StreamFactory = <T>(stream: ReadableStream<T>) => AsyncIterableStream<T>;
+
+const implementations: Array<{
+  name: string;
+  create: StreamFactory;
+}> = [
+  {
+    name: 'createAsyncIterableStream',
+    create: createAsyncIterableStream,
+  },
+  {
+    name: 'asAsyncIterableStream',
+    create: asAsyncIterableStream,
+  },
+];
+
+describe.each(implementations)('$name read-error cleanup', ({ create }) => {
+  it('releases the reader and preserves the exact source error', async () => {
+    const sourceError = { type: 'source-error' };
+    let controller!: ReadableStreamDefaultController<string>;
+    let cancelCalls = 0;
+
+    const source = new ReadableStream<string>({
+      start(controllerParam) {
+        controller = controllerParam;
+        controller.enqueue('chunk1');
+      },
+      cancel() {
+        cancelCalls++;
+      },
+    });
+
+    const stream = create(source);
+    const iterator = stream[Symbol.asyncIterator]();
+
+    expect(await iterator.next()).toEqual({
+      done: false,
+      value: 'chunk1',
+    });
+
+    const failedRead = iterator.next();
+    controller.error(sourceError);
+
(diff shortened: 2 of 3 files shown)


teamleaderleo:
I compared this with #18400. I prefer the try/catch used there because it covers only reader.read().

Both suites cover the same three main scenarios, but this PR’s first scenario also exercises partial consumption and a non-Error object reason, while #18400’s corresponding scenario uses an immediate Error rejection.

gr2m:
done via https://github.com/vercel/ai/pull/18400

PR #18372, opened 2026-08-03
feat(fish-audio): add Fish Audio provider

shaper:
Adds `@ai-sdk/fish-audio` with speech (TTS) and transcription (STT) models.

Groundwork for adding Fish Audio to AI Gateway (AIG-411), which needs a published provider package. Gateway wiring follows in a separate PR once this releases.

## Models

| Fish Audio endpoint | Spec | Notes |
| --- | --- | --- |
| `POST /v1/tts` | `SpeechModelV4` | JSON request, binary audio response |
| `POST /v1/asr` | `TranscriptionModelV4` | multipart request, word-level segments |

**Speech.** The model is selected via the `model` HTTP header rather than a body field. `voice` maps to `reference_id`, `outputFormat` to `format` (`wav`/`pcm`/`mp3`/`opus`), and `speed` to `prosody.speed`. `language` and `instructions` have no Fish Audio equivalent and produce `unsupported` warnings. Multi-speaker dialogue is available through `providerOptions.fishAudio.referenceId` as an array, since the generic `voice` option is a single string.

**Transcription.** `/v1/asr` currently serves a single model and exposes no model selector, so the model ID defaults to `transcribe-1` — the routing label Fish Audio recommends. It is not sent to the API. Fish Audio expects to add ASR models and select them with the `model` header, matching `/v1/tts`.

## Behavior confirmed with the Fish Audio team

Several behaviors were reverse-engineered from the live API because the published docs were wrong or silent. The Fish Audio team has since confirmed all of them and is updating their docs. Each is commented at the relevant call site.

1. **`/v1/asr` takes `multipart/form-data`.** The docs contradict themselves — prose says form-data/msgpack, the OpenAPI `requestBody` says json/msgpack. Form-data is correct; there is no base64-over-JSON path and none is planned.

2. **Responses include undocumented `language` and `language_code`.** The documented 200 schema is `{text, duration, segments[]}`, but real responses also carry `language` (`"English"`) and `language_code` (`"en"`). `language_code` is the reliable one: always two-letter ISO-639-1, never a locale like `en-US`, and absent when no language is detected. It is reported as `result.language`; the display name is exposed as provider metadata and documented as presentation-only.

3. **The `language` request parameter does not affect detection.** It is passed to the model, but auto-detection is authoritative and overrides it. Tested both directions before the team confirmed it:

   | Audio | Requested `language` | `language_code` returned |
   | --- | --- | --- |
   | English | *(none)* | `en` |
   | English | `ja` | `en` |
   | Japanese | *(none)* | `ja` |
   | Japanese | `ja` | `ja` |
   | Japanese | `en` | `ja` |

   Detection is accurate — a Japanese round-trip transcribes back verbatim. Echoing the requested value, which an earlier revision did, would have misreported audio in both directions.

4. **`ignore_timestamps: true` returns `segments: []`** by design, with `duration` still populated.

5. **`prosody.normalize_loudness` covers the S2 family** (`s2-pro` and `s2.1-pro`), not just `s2-pro` as documented. On `s1` it is accepted but silently ignored, so the provider drops it there and emits a warning rather than letting it no-op.

6. **All four models work on the TTS-live WebSocket.** The doc listing only `s1` and `s2-pro` is stale. Not used here, but it removes a restriction an earlier revision had encoded.

## Deliberate choices

- **`ignoreTimestamps` defaults to `false`**, inverting the API default of `true`, because Fish Audio returns an empty `segments` array when timestamps are skipped and `TranscriptionModelV4Result.segments` is required. The documented latency cost applies only to audio under 30 seconds. The option mirrors the API's negative name rather than an inverted `enableTimestamps`, so it cross-references cleanly against Fish Audio's docs. Precedent: `google-vertex` defaults `enableWordTimeOffsets` to `true` for the same reason.
- **`s2.1-pro` is treated as the recommended default.** `s2.1-pro-free` is a free developer tier with no time-to-first-audio or data-processing guarantees. The provider always sends the `model` header explicitly and never relies on the API default.
- **`voice` is optional.** Fish Audio defaults `reference_id` to null and uses a default voice, so `basic.ts` runs with only an API key.
- **`speechModel` / `transcriptionModel` are narrowed to required** on `FishAudioProvider`. They are optional on `ProviderV4`; Fish Audio always provides both, so consumers don't need `?.`.
- **Inline `references` (zero-shot cloning) are unsupported.** They require a MessagePack request body, and no provider in this repo handles msgpack — supporting it would mean a new dependency. Fish Audio confirmed a JSON path is not planned and that pre-uploading is preferable regardless, since pre-uploaded voices are encoded once and reused while inline references re-encode per request. Tracked as AIG-450.
- **Streaming TTS is out of scope.** `SpeechModelV4` is generate-only (no `doStream`), so neither the TTS-live WebSocket nor the timestamped streaming endpoint has a home in the current specification. `RealtimeModelV4` is a bidirectional conversational spec — client secrets, session config, tool definitions — and is not a fit for a one-directional text-in/audio-out endpoint. Tracked as AIG-451, routed to the `RealtimeSpeechEngine` design in KDA-126.

## Versioning

The package is at `2.0.0` with a `major` changeset, so it releases as `3.0.0`, in line with the current provider line (`@ai-sdk/cartesia` 3.0.12, `@ai-sdk/minimax` 3.0.2).

## Verification

- 46 unit tests across node and edge runtimes
- All 7 examples run against the live API: mp3 (128 kbps / 44.1 kHz mono) and Ogg Opus (48 kHz mono) verified with `file`; transcription returns a correct transcript with 82 word-level segments; multi-speaker output transcribes back to the intended dialogue with no literal `<|speaker:N|>` tokens; the `s1` loudness warning fires while `s2-pro` and `s2.1-pro` pass the flag through
- `pnpm check`, `pnpm konsistent`, package type-check and build all clean
- `pnpm type-check:full` clean for this branch

Examples use public voice IDs from the Fish Audio library, verified working. `list-voices.ts` browses others — voice listing is not part of the speech model specification, so it calls the Fish Audio REST API directly.

The lockfile diff is deliberately hand-maintained and purely additive (30 lines, no removals). A plain `pnpm install` on macOS re-resolves unrelated peers — `@opentelemetry/api`, `picomatch`, `@babel/core`, `@workflow/*` — and an earlier revision of this PR failed CI for exactly that reason.


packages/fish-audio/src/fish-audio-provider.ts
```diff
@@ -0,0 +1,168 @@
...
+   */
+  transcriptionModel(
+    modelId?: FishAudioTranscriptionModelId,
+  ): TranscriptionModelV4;
+
+  /**
+   * @deprecated Use `embeddingModel` instead.
+   */
+  textEmbeddingModel(modelId: string): never;
+}
```
aayush-kapoor: if they don't have an embedding model, might make sense to remove it (and if they do, should use `embeddingModel` instead

aayush-kapoor (approved):

packages/fish-audio/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@ai-sdk/fish-audio",
+  "version": "0.0.0",
```
gr2m: 2.0.0 will create 3.0.0, which allows us to backport it to v6/v5 if we choose to do so in future.
```suggestion
  "version": "2.0.0",
```

PR #18378, opened 2026-08-03
perf: remove runtime Zod 3 dependency

n1ckoates:
## Background

`@ai-sdk/provider-utils` imported the `ZodFirstPartyTypeKind` enum from `zod/v3` in its parser. But that file includes a bunch of Zod schemas that aren't tree shaken.

## Summary

Replaced that enum with strings that are typed checked against the original enum at compile-time, so that the import is type-only.

| Target | `main` raw | This branch raw | Raw saved | Gzip saved | Brotli saved |
|---|---:|---:|---:|---:|---:|
| Node | 665,384 B | 607,959 B | **57,425 B** | **12,875 B** | **9,958 B** |
| Browser | 653,465 B | 596,040 B | **57,425 B** | **12,803 B** | **9,862 B** |

This doesn't break anything because `ZodFirstPartyTypeKind.ZodString === 'ZodString', so we can just check the string `'ZodString'` directly rather than importing the enum.

## End-to-End Verification

Tested `examples/ai-functions/src/generate-text/mock/output-object.ts` example which uses Zod 3 and the JSON Schema parser, and it still works.

## Checklist

### DIFF
--- a/.changeset/calm-dragons-wave.md
+++ b/.changeset/calm-dragons-wave.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+Reduce bundle size by removing the runtime Zod 3 dependency.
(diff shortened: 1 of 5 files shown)


gr2m (approved):
Great refactor! Can you update the bundle size limit to 620? We want to keep it as a signal to make sure we only increase bundle size meaningfully when intended.

PR #18383, opened 2026-08-04
 feat(sandbox-local-workspace): run harnesses against a local project directory

kajogo777:
## Background

The shipped sandbox providers are `@ai-sdk/sandbox-vercel` (remote microVM) and `@ai-sdk/sandbox-just-bash` (in-process virtual filesystem). Neither lets someone point `HarnessAgent` at a repository on their own machine, so a user with Claude Code, Codex or Pi already installed and authenticated has to provision a hosted sandbox first.

This started as a separate `@ai-sdk/sandbox-local-workspace` package. Per review feedback it moved into `@ai-sdk/harness` instead, so no package is added that we would not want long term, the API is already in the shape we want, and a fuller local workflow can be designed later.

## Summary

`sandbox` becomes optional. When omitted the harness runs on the local machine in `process.cwd()` as the current user, inheriting the environment so each adapter resolves its own credentials from `~/.claude/settings.json`, `~/.codex/config.toml` or `~/.pi/agent/auth.json`.

```ts
  harness: createClaudeCode(),
  instructions: 'You are a careful refactoring assistant.',
});
```

The provider is internal, so nothing new is exported. Four things worth knowing:
   - **Session directory.** `resolveSessionWorkDir` gains an internal `runInCwd` flag. The usual `<harnessId>-<sessionId>` subdirectory suits a freshly provisioned sandbox, but here the root is the user's own directory, so a subdirectory would run the harness in an empty folder beside their files. `sandboxConfig.workDir` cannot express `'.'`; public validation is unchanged.
   - **Generated state stays out of git.** Adapters resolve `bootstrapDir` against `defaultWorkingDirectory` themselves, so their output lands in the project. `.harness-bootstrap/` (around 785 MB for Claude Code), and `.agent-runs/` each get a `.gitignore` of `*` before anything writes into them. The project's own `.gitignore` is never touched, and it works outside a git repo.
   - **No isolation, warned once.** The harness has every permission the current user has. A warning fires once per process and honours the same `AI_SDK_LOG_WARNINGS` global as the rest of the SDK (`false` silences, a function delegates). `logWarnings` is private to `ai`, so the convention is reimplemented rather than that export widened.
   - **Lazy load.** Pulled in via dynamic `import()` so requiring `HarnessAgent` does not put `node:child_process` in the module graph of consumers who pass their own provider.

## End-to-End Verification

Real `claude`, `codex` and `pi` CLIs on macOS, Node 22, pnpm 10.33.4. `sandbox` is never passed, so each scenario changes directory into a project and the agent must fall back to the local machine.

   | Check | pi | claude-code | codex |
   | --- | --- | --- | --- |
   | Writes `NOTES.md` into the working directory | pass | pass | pass |
   | Used a native tool | pass | pass | pass |
   | Shell tool runs in the working directory itself | pass | pass | pass |
   | `stop()` then resume recalls prior turn state | pass | pass | pass |
   | All three generated directories hidden from git | pass | pass | pass |
   | Real `git status` reports only the user's files | pass | pass | pass |
   | No bridge processes left behind | pass | pass | pass |

Every scenario asserts `usage.totalTokens > 0`, because a missing credential and a harness that silently lost its tools both finish with `finishReason: 'stop'`, empty text and no error part.

## Checklist

## Future Work
   - **Move bootstrap out of the project.** 785 MB inside the user's directory, invisible to git but real on disk, and a new directory pays again. A shared cache needs adapter cooperation, since they compute `bootstrapDir` themselves.
   - **A configurable local workflow.** `path` and `env` exist internally but nothing is exported, so the only user-facing behaviour is "runs in `cwd`".
   - **An optional `defaultSessionWorkDir` on `HarnessV1NetworkSandboxSession`**, so a provider declares its session directory instead of the agent inferring it from `settings.sandbox == null`. Touches the `HarnessV1` spec, so left out here.


felixarntz:
@kajogo777 pushed one additional commit in https://github.com/vercel/ai/pull/18383/changes/187a7727322b25e5f50d550b1dbe37ac225d3c56 to rename the folder and make minor alignments so that we can reuse the `konsistent.json` convention validation for this "unusual" sandbox too.

This way we keep the implementation aligned, even though it's not an actual sandbox package.

No logic changes, just renaming things.

content/docs/03-ai-sdk-harnesses/02-harness-agent.mdx
```diff
@@ -15,9 +15,54 @@ Install the core harness package, a harness adapter, and a sandbox provider:
...
+
+### Generated directories
+
+Running in your working directory means the harness generates state there. Two directories
+may appear:
+
+| Directory             | Contents                                                                                                                    |
+| --------------------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `.harness-bootstrap/` | Bridge dependencies for bridge-backed harnesses, plus setup markers. Hundreds of megabytes; Claude Code's is around 785 MB. |
+| `.agent-runs/`        | Per-session harness state, including the event log a suspended turn is replayed from.                                       |
```
felixarntz, line 57: do you think we could put these in another directory? ideally the actual work dir for the agent isn't used for anything infrastructure that the harness needs.

the current harness providers write all these artifacts / infra code elsewhere so that it doesn't pollute the agent work dir.

maybe something like `/tmp`, or a specific folder in `$HOME`? alternatively, if we can split sandbox working directory from agent work dir like all harnesses do, that would allow for us to write this into the working directory of the sandbox while the agent doesn't see it.

I know this naming is confusing, so to clarify:
- "sandbox working directory": the default cwd of the sandbox provider
- "agent/session work dir": the folder passed to the agent harness SDK as root dir, always a direct child folder of the sandbox working directory

packages/harness/src/agent/internal/sandbox-bootstrap.ts
```diff
@@ -64,17 +64,32 @@ export function normalizeSandboxWorkDir(workDir: string): string {
...
   workDir,
+  runInCwd = false,
 }: {
+  readonly runInCwd?: boolean;
   readonly defaultWorkingDirectory: string;
   readonly harnessId: string;
   readonly sessionId: string;
   readonly workDir?: string;
 }): string {
+  if (workDir == null && runInCwd) return defaultWorkingDirectory;
```
felixarntz, line 92: I don't love adding new behavior here for just this local sandbox. also see my other comment about not "polluting" the agent work dir with infrastructure folders.

could we keep the current contract in place and document the behavior? e.g.:
- if you don't care what you're working in and just need an empty folder to start, rely on the harness creating that work dir for you (e.g. `claude-code-a1b2c3d4e5f6` folder inside `defaultWorkingDirectory`)
- if you want to work on an existing project, its parent folder has to be cwd and you must pass the actual folder name as `workDir` (e.g. if I have `~/Development/ai-sdk` and I want `HarnessAgent` to work on the `ai-sdk` source code, cwd should be `~/Development` and `workDir` should be set to `ai-sdk`

in both cases, the cwd / `defaultWorkingDirectory` will get those artifacts needed or produced by the harness, while the subfolder will remain untouched and the only thing in scope of the agent.

packages/harness/src/agent/harness-agent.ts
```diff
@@ -298,6 +320,9 @@ export class HarnessAgent<
       harnessId: harness.harnessId,
       sessionId,
       workDir: sandboxBootstrapPlan.workDir,
+      // The implicit local workspace is the user's own working directory, so
+      // the session belongs there rather than in a subdirectory of it.
+      runInCwd: this.settings.sandbox == null,
```
felixarntz, line 325: see above.

another idea, if we don't want the user to pass `workDir` when they want to work on a specific directory, we could do the reverse path than what I said in my other comment: use the parent dir path of cwd as `defaultWorkingDirectory` and the folder name of cwd as `workDir`.

Then we would document that the parent directory of your working directory would get those infrastructure files.

PR #18384, opened 2026-08-04
feat (provider/gateway): support async video operations (doStart/doStatus) on the v4 video model

kkdawkins:
## Background

#12515 shipped the async video operation protocol on the v4 spec (`doStart`/`doStatus`), and the async-native providers implement it. The AI Gateway serves durable async video jobs over `POST /v4/ai/video-model/start|status`, but `GatewayVideoModel` only implements `doGenerate` — so `experimental_generateVideo({ model: gateway(...), poll: {...} })` cannot use the async flow end-to-end.

## Summary

Implements `doStart`/`doStatus` on `GatewayVideoModel` (v4) against the Gateway's async video job endpoints, and makes retried `doStart` calls idempotent in core:

- `doStart` → `POST {baseURL}/video-model/start`; the 202 body's `operation` becomes the opaque v4 operation reference. Shares `doGenerate`'s full request-body construction and forwards `webhookUrl` when supplied.
- `doStatus` → `POST {baseURL}/video-model/status`, mapping the gateway's `pending`/`completed`/`error` shapes onto `VideoModelV4OperationStatusResult`. The gateway's `cancelled` terminal has no v4 union member, so it surfaces as a terminal `error` (never an infinite poll).
- **`doStart` retries are idempotent.** `generateVideo` retries `doStart`, which creates a *billable* generation — a retry after a lost response could start (and charge for) a second one. Core now mints one explicit token per logical start (`aisdk_vid_<id>`, outside the retry closure) and forwards it as an `idempotency-key` header; a caller-supplied key wins over the minted one. `GatewayVideoModel` just forwards headers — the Gateway's start endpoint already deduplicates on this header, and other providers receive a header they can ignore today and honor later. No spec change; if a first-class `idempotencyKey` call option is preferred so non-gateway providers can opt in explicitly, happy to switch — the experimental v4 surface allows it.
- `handleWebhookOption` is intentionally **not** implemented: the gateway is polling-first and exposes no provider→SDK webhook completion channel today; presence of the method would make `generateVideo` await a notification that never arrives. Adding webhooks later is additive.
- Errors on both calls flow through the existing `asGatewayError` mapping.

One recorded contract note: the Gateway can serve a long-completed job whose provider asset has expired as `completed` with empty `videos[]` plus explanatory `providerMetadata`. Unreachable inside a single `generateVideo` run (a just-started operation's result is always fresh; core throws `NoVideoGeneratedError` on empty results as usual) — relevant only if the SDK later exposes resuming an operation by reference.

## Verification

- `pnpm --filter @ai-sdk/gateway test` — 509 passed (new `doStart`/`doStatus` tests: request/headers shape, webhookUrl forwarding, pending/completed/error/cancelled mappings, no `handleWebhookOption`).
- `generate-video` suite — 62 passed (stable idempotency key across retries, distinct key per call, caller-supplied key preserved).
- `tsc --noEmit` clean in both packages; `oxlint` + `ultracite check` clean.

## Tasks

- [x] Documentation has been added / updated — n/a (experimental surface; gateway async video docs land with the gateway feature)
- [x] A _patch changeset_ for relevant packages has been added
- [x] Formatting issues have been fixed (`ultracite check` clean)

## Related

vercel/ai#12515 (protocol), vercel/ai-gateway#2974 + the Option-B stack (server side).

### DIFF
--- a/packages/ai/src/generate-video/generate-video.ts
+++ b/packages/ai/src/generate-video/generate-video.ts
@@ -553,13 +553,15 @@ async function executeStartStatusFlow({
     }
   }
 
-  // 2. Start the generation
-  const startResult = await retry(() =>
-    model.doStart!({
-      ...callOptions,
-      webhookUrl,
-    }),
-  );
+  // 2. Start the generation.
+  //
+  // The options object is built ONCE, outside the retry closure, so every
+  // attempt receives the identical reference. `doStart` has a side effect the
+  // caller pays for, and a retry after a lost response would otherwise create a
+  // second generation; a provider (or the Vercel AI Gateway) can key
+  // deduplication off this identity without a spec change.
+  const startCallOptions = { ...callOptions, webhookUrl };
+  const startResult = await retry(() => model.doStart!(startCallOptions));
 
   const allWarnings = [...earlyWarnings, ...startResult.warnings];
   let operationProviderMetadata =
--- a/packages/ai/src/generate-video/generate-video.ts
+++ b/packages/ai/src/generate-video/generate-video.ts
@@ -11,6 +11,7 @@ import type {
 import {
   convertBase64ToUint8Array,
   delay as defaultDelay,
+  generateId,
   withUserAgentSuffix,
   type DataContent,
   detectMediaType,
@@ -555,12 +556,21 @@ async function executeStartStatusFlow({
 
   // 2. Start the generation.
   //
-  // The options object is built ONCE, outside the retry closure, so every
-  // attempt receives the identical reference. `doStart` has a side effect the
-  // caller pays for, and a retry after a lost response would otherwise create a
-  // second generation; a provider (or the Vercel AI Gateway) can key
-  // deduplication off this identity without a spec change.
-  const startCallOptions = { ...callOptions, webhookUrl };
+  // `doStart` has a side effect the caller pays for, so a retry after a lost
+  // response must not create a second generation. Mint ONE idempotency token per
+  // logical start — here, outside the retry closure — and forward it as a header
+  // so a provider that supports deduplication can key off it. Explicit, rather
+  // than inferred from object identity: two unrelated calls that happen to share
+  // an options object must never share a token.
+  const startIdempotencyKey = `aisdk_vid_${generateId()}`;
+  const startCallOptions = {
+    ...callOptions,
+    headers: {
+      ...callOptions.headers,
+      'idempotency-key': startIdempotencyKey,
+    },
+    webhookUrl,
+  };
   const startResult = await retry(() => model.doStart!(startCallOptions));
 
   const allWarnings = [...earlyWarnings, ...startResult.warnings];
--- a/packages/ai/src/generate-video/generate-video.ts
+++ b/packages/ai/src/generate-video/generate-video.ts
@@ -562,7 +562,11 @@ async function executeStartStatusFlow({
   // so a provider that supports deduplication can key off it. Explicit, rather
   // than inferred from object identity: two unrelated calls that happen to share
   // an options object must never share a token.
-  const startIdempotencyKey = `aisdk_vid_${generateId()}`;
+  // A caller-supplied `idempotency-key` header wins: users implementing their
+  // own dedupe (e.g. an outer retry wrapper with a stable key) must not have
+  // it silently replaced.
+  const startIdempotencyKey =
+    callOptions.headers?.['idempotency-key'] ?? `aisdk_vid_${generateId()}`;
   const startCallOptions = {
     ...callOptions,
     headers: {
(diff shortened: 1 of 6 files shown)


packages/gateway/src/gateway-video-model.ts
```diff
@@ -202,10 +170,204 @@ export class GatewayVideoModel implements VideoModelV4 {
...
+        url: this.getStartUrl(),
+        headers: combineHeaders(
+          resolvedHeaders,
+          headers ?? {},
+          this.getModelConfigHeaders(),
+          await resolve(this.config.o11yHeaders),
+        ),
+        body: {
+          ...this.buildRequestBody(options),
+          ...(webhookUrl && { webhookUrl }),
```
shaper: The forwarded field name doesn't match the gateway's wire contract: `doStart` sends `webhookUrl`, but the gateway's start route reads `callbackUrl` (`lib/gateway/async-jobs/video-routes.ts` — `const callbackUrlRaw = (rawBody as { callbackUrl?: unknown }).callbackUrl`), and unknown body keys are stripped — so `webhookUrl` is silently dropped server-side.

This is unreachable today (nothing sets `webhookUrl` while `handleWebhookOption` is unimplemented), so it doesn't block this PR. But the comment above — "enabling webhooks later only requires adding `handleWebhookOption` here" — is inaccurate as written: whoever adds it would get a silent no-op webhook rather than a wired one.

Suggest either renaming the forwarded field to `callbackUrl` now (matching the server), or amending the comment to note the field-name reconciliation as part of the future work. (Semantics otherwise line up: the gateway's `callbackUrl` is exactly its customer-facing completion-webhook field, though the eventual `handleWebhookOption` implementation will also need to accept the gateway's HMAC-signed delivery format.)

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -553,13 +554,28 @@ async function executeStartStatusFlow({
...
+  // `doStart` has a side effect the caller pays for, so a retry after a lost
+  // response must not create a second generation. Mint ONE idempotency token per
+  // logical start — here, outside the retry closure — and forward it as a header
+  // so a provider that supports deduplication can key off it. Explicit, rather
+  // than inferred from object identity: two unrelated calls that happen to share
+  // an options object must never share a token.
+  // A caller-supplied `idempotency-key` header wins: users implementing their
+  // own dedupe (e.g. an outer retry wrapper with a stable key) must not have
+  // it silently replaced.
+  const startIdempotencyKey =
```
shaper: nit: inline value at line 574

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -553,13 +554,28 @@ async function executeStartStatusFlow({
...
-  // 2. Start the generation
-  const startResult = await retry(() =>
-    model.doStart!({
-      ...callOptions,
-      webhookUrl,
-    }),
-  );
+  // 2. Start the generation.
+  //
+  // `doStart` has a side effect the caller pays for, so a retry after a lost
```
shaper: nit: comment too verbose, at most 2 lines. if we need more perhaps better in AI SDK ts-doc. is this idempotency key something we should document in the `generateVideo` reference docs under `content` at top level?

shaper (approved):

packages/ai/src/generate-video/generate-video.ts
```diff
@@ -553,13 +554,28 @@ async function executeStartStatusFlow({
...
+  // response must not create a second generation. Mint ONE idempotency token per
+  // logical start — here, outside the retry closure — and forward it as a header
+  // so a provider that supports deduplication can key off it. Explicit, rather
+  // than inferred from object identity: two unrelated calls that happen to share
+  // an options object must never share a token.
+  // A caller-supplied `idempotency-key` header wins: users implementing their
+  // own dedupe (e.g. an outer retry wrapper with a stable key) must not have
+  // it silently replaced.
+  const startIdempotencyKey =
+    callOptions.headers?.['idempotency-key'] ?? `aisdk_vid_${generateId()}`;
```
gr2m: do we need to normalize/lowercase keys before checking?

packages/gateway/src/gateway-video-model.ts
```diff
@@ -282,3 +444,42 @@ const gatewayVideoEventSchema = z.discriminatedUnion('type', [
...
+    providerMetadata: z
+      .record(z.string(), providerMetadataEntrySchema)
+      .optional(),
+  }),
+  z.object({
+    status: z.literal('cancelled'),
+    providerMetadata: z
+      .record(z.string(), providerMetadataEntrySchema)
+      .optional(),
+  }),
```
gr2m, line 487: we usually use `.nullish()` instead of `.optional()` in response schemas since it's usually outside our control, see https://github.com/vercel/ai/blob/main/contributing/providers.md#response-schemas

gr2m (approved):

PR #18388, opened 2026-08-04
feat(@ai-sdk/minimax): add image generation support

octo-patch:
Reason: MiniMax registers chat and video models but `imageModel` throws `NoSuchModelError`; this adds image generation support for the MiniMax image API.

## Summary

Adds a `MiniMaxImageModel` to `@ai-sdk/minimax` and wires it into the provider via `minimax.image()` / `minimax.imageModel()`.

The image model calls `POST /v1/image_generation` (configurable via the new `imageBaseURL` setting) and maps the AI SDK image model options onto the MiniMax API:

- `prompt`, `n`, `seed` pass through directly.
- `size` (`{width}x{height}`) is split into `width`/`height`.
- `aspectRatio` is mapped to `aspect_ratio` when the ratio is supported, otherwise it is ignored with a warning.
- Input `files` (image-to-image) are sent as `subject_reference` entries.
- A `mask` is not supported and is ignored with a warning.
- Provider options `responseFormat` (`url` | `base64`) and `promptOptimizer` map to `response_format` and `prompt_optimizer`.

`response_format: "base64"` results are returned as base64 strings; the default `url` results are downloaded and returned as binary. Non-zero `base_resp.status_code` responses throw, and the response exposes provider metadata (`traceId`, `successCount`, `failedCount`, image URLs).

## Changes

- `packages/minimax/src/minimax-image-settings.ts` — `MiniMaxImageModelId` type.
- `packages/minimax/src/minimax-image-model-options.ts` — image provider options schema.
- `packages/minimax/src/minimax-image-model.ts` — the image model implementation.
- `packages/minimax/src/minimax-image-model.test.ts` — tests for text-to-image, image-to-image, base64 and URL responses, and error handling.
- `packages/minimax/src/minimax-provider.ts` — adds `image`/`imageModel`, `imageBaseURL` setting.
- `packages/minimax/src/minimax-provider.test.ts` — provider tests for the image model.
- `packages/minimax/src/index.ts` — exports the new types.
- `packages/minimax/README.md` + provider docs — image generation example and provider options.
- `.changeset/smart-images-dream.md` — patch changeset.

## Checks

- `pnpm --filter @ai-sdk/minimax type-check` — passed.
- `pnpm --filter @ai-sdk/minimax test:node` — 132 tests passed.
- `pnpm --filter @ai-sdk/minimax test:edge` — 132 tests passed.
- `oxlint src/` — 0 warnings, 0 errors.
- `oxfmt --check src/` — all files formatted.

### DIFF
--- a/.changeset/smart-images-dream.md
+++ b/.changeset/smart-images-dream.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/minimax': patch
+---
+
+feat (ai/minimax): add image generation support for the MiniMax provider
--- a/content/providers/01-ai-sdk-providers/33-minimax.mdx
+++ b/content/providers/01-ai-sdk-providers/33-minimax.mdx
@@ -50,6 +50,12 @@ You can use the following optional settings to customize the MiniMax provider in
   the MiniMax V2 native endpoint (not the Anthropic-compatible endpoint).
   The default prefix is `https://api.minimax.io`.
 
+- **imageBaseURL** _string_
+
+  Use a different URL prefix for image generation API calls. The image API uses
+  the MiniMax V1 native endpoint (not the Anthropic-compatible endpoint).
+  The default prefix is `https://api.minimax.io`.
+
 - **apiKey** _string_
 
   API key for the MiniMax API. It defaults to the `MINIMAX_API_KEY` environment
@@ -135,7 +141,41 @@ The following optional provider options are available for MiniMax language model
     think, so `'disabled'` has no effect on them.
   </Note>
 
+## Image Models
+
+You can generate images with the `image-01` and `image-01-live` models using the
+[`experimental_generateImage`](/docs/reference/ai-sdk-core/generate-image)
+function:
+
+```ts
+import { minimax, type MiniMaxImageModelOptions } from '@ai-sdk/minimax';
+import { experimental_generateImage as generateImage } from 'ai';
+
+const { images } = await generateImage({
+  model: minimax.image('image-01'),
+  prompt: 'A cute baby sea otter',
+  n: 2,
+  aspectRatio: '16:9',
+});
+```
+
+Input images are supported for image-to-image generation via the `prompt.images`
+field. They are sent to the API as subject references.
+
+### Image Provider Options
+
+The following optional provider options are available for MiniMax image models:
+
+- **responseFormat** _'url' | 'base64'_
+
+  Format of the generated images. Defaults to `url`. URLs expire after 24 hours.
+
+- **promptOptimizer** _boolean_
+
+  Whether to enable automatic optimization of the prompt. Defaults to `false`.
+
 ## Video Models
+
 
 You can generate videos with the MiniMax-H3 model using the
 [`experimental_generateVideo`](/docs/reference/ai-sdk-core/generate-video)
--- a/packages/minimax/README.md
+++ b/packages/minimax/README.md
@@ -57,6 +57,20 @@ const { text } = await generateText({
 });
 ```
 
+## Image Generation Example
+
+```ts
+import { minimax } from '@ai-sdk/minimax';
+import { experimental_generateImage as generateImage } from 'ai';
+
+const { images } = await generateImage({
+  model: minimax.image('image-01'),
+  prompt: 'A cute baby sea otter',
+  n: 2,
+  aspectRatio: '16:9',
+});
+```
+
 ## Video Generation Example
 
 ```ts
--- a/packages/minimax/src/index.ts
+++ b/packages/minimax/src/index.ts
@@ -9,6 +9,8 @@ export type {
   /** @deprecated Use `MiniMaxLanguageModelOptions` instead. */
   MiniMaxLanguageModelOptions as MiniMaxProviderOptions,
 } from './minimax-chat-options';
+export type { MiniMaxImageModelId } from './minimax-image-settings';
+export type { MiniMaxImageModelOptions } from './minimax-image-model-options';
 export type { MiniMaxVideoModelId } from './minimax-video-settings';
 export type { MiniMaxVideoModelOptions } from './minimax-video-model-options';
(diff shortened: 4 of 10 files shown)


ManyaS-Git (commented):
## Code Review: feat(@ai-sdk/minimax): add image generation support

Nice work on this PR! The implementation is well-structured, the tests are comprehensive, and the mapping of MiniMax-specific concepts (subject references, aspect ratios, response formats) onto the SDK's image model interface is thoughtful. I have a few findings below.

---

### 1. Missing `WORKFLOW_SERIALIZE` / `WORKFLOW_DESERIALIZE` -- should fix

Every image model in the repository (OpenAI, FAL, xAI, Black Forest Labs, Google, Amazon Bedrock, TogetherAI, DeepInfra, Fireworks, Prodia, etc.) implements these static methods. This is a hard convention for all model classes. Without them, workflow serialization/deserialization for this model will fail.

Add the following to `MiniMaxImageModel`:

```typescript
import {
  serializeModelOptions,
  WORKFLOW_SERIALIZE,
  WORKFLOW_DESERIALIZE,
} from '@ai-sdk/provider-utils';

// Inside the class:
static [WORKFLOW_SERIALIZE](model: MiniMaxImageModel) {
  return serializeModelOptions({
    modelId: model.modelId,
    config: model.config,
  });
}

static [WORKFLOW_DESERIALIZE](options: {
  modelId: MiniMaxImageModelId;
  config: MiniMaxImageModelConfig;
}) {
  return new MiniMaxImageModel(options.modelId, options.config);
}
```

---

### 2. Image download should conditionally send credentials -- should fix

When downloading images from response-supplied URLs, the current code sends **no headers at all**:

```typescript
getFromApi({
  url,
  validateUrl: true,
  trustedOrigin: this.config.baseURL,
  abortSignal,
  // no headers passed
  ...
})
```

Other image models (BFL, xAI, FAL, Fireworks, Luma, Replicate) follow a security best practice of conditionally sending credentials only when the URL is same-origin with the provider, to avoid leaking API keys to CDNs:

```typescript
import { isTrustedUrl } from './minimax-image-model'; // or a shared utility

getFromApi({
  url,
  validateUrl: true,
  trustedOrigin: this.config.baseURL,
  headers: isTrustedUrl(url, this.config.baseURL)
    ? combinedHeaders
    : undefined,
  abortSignal,
  ...
})
```

The `isTrustedUrl` helper is a simple same-origin check (BFL defines it locally at `black-forest-labs-image-model.ts:372`). Even if MiniMax image URLs are always CDN-served today, adding this guard is defensive and follows the established pattern.

---

### 3. Redundant error handling for `base_resp.status_code` -- consider simplifying

The PR uses **two** error handling paths for the same `base_resp.status_code` field:

1. `minimaxImageFailedResponseHandler` (via `createJsonErrorResponseHandler`) -- handles HTTP-level errors
2. Manual check: `if (statusCode != null && statusCode !== 0) throw new AISDKError(...)` -- handles application-level errors in HTTP 200 responses

This is actually a valid pattern for APIs that return 200 with error bodies. However, there's a subtle gap: the `minimaxImageFailedResponseHandler` will **not** fire for HTTP 200 responses with a non-zero `base_resp.status_code`, because `createJsonErrorResponseHandler` only triggers on non-2xx HTTP status codes. So the manual check is necessary and correct.

That said, consider whether `minimaxImageFailedResponseHandler` could be simplified. Since the manual check already covers all non-zero `base_resp.status_code` cases, the `failedResponseHandler` only needs to handle actual HTTP errors (4xx/5xx). You could use a simpler `createJsonErrorResponseHandler` with a more generic error message, or even `createStatusCodeErrorResponseHandler()` if the HTTP error body isn't needed. The current `minimaxImageErrorSchema` duplicates the same `base_resp` schema already in the success schema.

---

### 4. Missing example in `examples/ai-functions/` -- nice to have

Per `contributing/building-new-features.md`, new features should include usage examples. Consider adding `examples/ai-functions/src/generate-image/minimax/basic.ts` following the established pattern:

```typescript
import { minimax } from '@ai-sdk/minimax';
import { presentImages } from '../../lib/present-image';

run(async () => {
  const { image } = await generateImage({
    model: minimax.image('image-01'),
    prompt: 'A cute baby sea otter',
  });
  await presentImages([image]);
});
```

---

### Minor / Nit

- **`getBearerTokenHeaders` naming** -- Renaming `getVideoHeaders` to `getBearerTokenHeaders` is a nice improvement that clarifies the shared auth scheme. Good call.

- **`promptOptimizer` type** -- The provider options schema defines `promptOptimizer` as `z.boolean().optional()`, but the model code uses `minimaxOptions?.promptOptimizer != null`. If a caller passes `promptOptimizer: false`, the `!= null` check passes and `prompt_optimizer: false` is sent to the API. This is probably fine (sending `false` explicitly is equivalent to the default), but worth noting.

- **Tests look thorough** -- Text-to-image, image-to-image, base64 response, error handling, missing images, timestamp injection, and provider options are all covered. The provider-level tests for `image`/`imageModel` aliases, base URL defaults/overrides, and header schemes are also solid.

---

### What looks great

- Clean separation of concerns: settings, options schema, model implementation, and provider wiring are all in separate files following the established convention.
- The `imageBaseURL` setting with independent fallback from `baseURL` and `videoBaseURL` is well-designed.
- The subject reference mapping for image-to-image and the `convertImageModelFileToDataUri` usage for inline files follow established patterns.
- The response schema uses `.nullish()` correctly per provider conventions.
- Provider metadata includes useful fields (traceId, successCount, failedCount, image URLs) for debugging.
- The changeset is correctly scoped as `patch`.
- The `imageModel` / `image` dual alias follows the convention used by other model types in the provider.

Overall, the main item I'd want addressed is the missing `WORKFLOW_SERIALIZE`/`DESERIALIZE` -- that's a hard requirement for image models. The conditional credentials on image downloads would also be good to add for security consistency with other providers.

PR #18408, opened 2026-08-04
fix(gateway): normalize denied tool results

### LINKED ISSUE #10663: bug: tool execution approval not working with gateway
### Description

Tool execution approval rejections work with OpenAI provider but fail with Gateway provider.

Steps to repro: 
1. run `examples/ai-core/src/stream-text/openai-tool-approval.ts`
2. change provider / model to use gateway
3. ask for weather somewhere then reject the execution approval

When a user denies a tool execution approval (answers "n" to the approval prompt), the gateway returns a `GatewayInvalidRequestError: Invalid input` error, specifically complaining about invalid `output` field in the prompt at `"prompt",3,"content",0,"output"`

### AI SDK Version

Latest

### Code of Conduct

daleselaji-dev:
## Problem

When a tool execution approval is denied, the AI SDK sends an internal execution-denied output through the Gateway adapter. Gateway rejects that prompt output as invalid, so the next model request fails.

Fixes #10663

## Root cause

The Gateway adapter forwarded the v4 execution-denied output unchanged instead of converting it to a wire-compatible tool result.

## Solution

- Convert execution-denied outputs to text using the denial reason.
- Use Tool call execution denied. when no reason is provided.
- Add request-body regression tests for both cases.
- Add a patch changeset for @ai-sdk/gateway.

## Validation

- git diff --cached --check passed.
- The Gateway Vitest suite could not be started because pnpm install repeatedly stalled while linking local dependencies in this environment; GitHub CI should provide the full test result.

## Notes for Reviewer

The generic Core execution-denied contract is unchanged. Only the Gateway adapter wire normalization is changed, matching the existing provider conversion behavior.

### DIFF
--- a/.changeset/clean-gateway-denials.md
+++ b/.changeset/clean-gateway-denials.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/gateway': patch
+---
+
+Fix rejected tool approvals when using AI Gateway.
--- a/packages/gateway/src/gateway-language-model.test.ts
+++ b/packages/gateway/src/gateway-language-model.test.ts
@@ -1751,6 +1751,59 @@ describe('GatewayLanguageModel', () => {
       });
     });
 
+    it('should convert execution-denied tool results to text', async () => {
+      prepareResponse();
+      const prompt: LanguageModelV4Prompt = [
+        {
+          role: 'tool',
+          content: [
+            {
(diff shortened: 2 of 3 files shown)


gr2m:
Closing: AI Gateway now accepts rejected tool approvals.

PR #18448, opened 2026-08-05
feat(harness-claude-code): drive a claude executable the sandbox already has

kajogo777:


kajogo777:
Folded into the base branch: george/sandbox-local-workspace now carries these commits (rebased as 0684458527..50964fd6d1) plus the interrupt/read-history fixes that landed here since, so #18383 is the single PR for the whole stack. Closing and deleting this head branch.

PR #18328, opened 2026-08-02
feat(provider/moonshotai): support video input via video_url parts

joemckenney:
## Background

Moonshot's Kimi models (`kimi-k3`, `kimi-k2.7-code`, `kimi-k2.6`, `kimi-k2.5`) accept video input via a `video_url` content part (`data:video/mp4;base64,...` or an `ms://<file_id>` reference). Today, any video file part sent to `@ai-sdk/moonshotai` throws `UnsupportedFunctionalityError: file part media type video/*` before a request is ever made, because the shared openai-compatible chat message converter has no `video/` branch. Users routing through the Vercel AI Gateway hit the same throw (surfaced as HTTP 400) and have to bypass the SDK and call Moonshot directly.

## Summary

- Adds a `supportsVideo` option to `OpenAICompatibleChatConfig` (default off) and a gated `video/` branch in `convert-to-openai-compatible-chat-messages.ts` that maps video file parts to `video_url` parts: URL passthrough for URL data, `data:<mediaType>;base64,...` (via `resolveFullMediaType`) for inline bytes.
- `@ai-sdk/moonshotai` opts in with one config line. Video URLs are not declared in `supportedUrls`, so the SDK downloads and inlines them as base64, matching Moonshot's API (which accepts data URIs, not arbitrary https URLs).
- Tests (converter: data -> `video_url`, URL passthrough, disabled -> still throws; provider: config assertion), docs (Video Input section + capability table), an example (`generate-text/moonshot/video-input.ts`), and a patch changeset for both packages.

Tradeoffs / alternative paths:

- **Gated flag (this PR) vs unconditional conversion**: emitting `video_url` unconditionally would change behavior for every openai-compatible consumer, turning a clear client-side throw into opaque server errors (or silently ignored parts) on backends that do not accept video. Default-off keeps all other providers bit-identical.
- **vs forking the converter into `packages/moonshotai`**: the converter is invoked from the private `getArgs` inside `doGenerate`/`doStream`, so a fork duplicates the arg-building and request layer as well (~450 lines of class + converter + tests) and hand-ports upstream fixes. DeepSeek (#10785) is the precedent for a full provider rewrite once differences accumulate; this branch and its tests move into such a rewrite unchanged if that happens.
- **vs a `convertFilePart` hook in config**: cleaner separation (Moonshot would own the branch), but adds new API surface to the shared package for a single dialect today. Can be revisited if more provider dialects appear.

## End-to-End Verification

[...]

PR #18449, opened 2026-08-05
feat(provider/moonshotai): own the chat implementation, support video input

joemckenney:
## Background

Moonshot's Kimi models (`kimi-k3`, `kimi-k2.7-code`, `kimi-k2.6`, `kimi-k2.5`) accept video input via a `video_url` content part. Today any video file part throws `UnsupportedFunctionalityError: file part media type video/*`, because the provider inherits its message conversion from `@ai-sdk/openai-compatible`, which has no `video/` branch. `video_url` is a Moonshot-invented extension (OpenAI's spec has no `video_url`, and video-capable stacks use incompatible shapes), so this PR moves moonshotai off `@ai-sdk/openai-compatible` and has the package own its chat implementation, following the DeepSeek precedent (#10785).

## Summary

- `MoonshotAIChatLanguageModel` now implements `LanguageModelV4` directly. The converter, API types, zod schemas, tool preparation, and finish-reason mapping are owned by the package; the `@ai-sdk/openai-compatible` dependency is dropped.
- `video/*` file parts convert to `video_url`: URL passthrough, or base64 data URI for inline bytes (the SDK downloads and inlines URLs; Moonshot does not fetch them).
- Behavior changes, all verified against the live API:
  - Audio/PDF file parts now fail client-side (the API rejects those part types with a 400 anyway).
  - `reasoningHistory: 'preserved'` maps to `thinking.keep: 'all'`. The old `reasoning_history` field is silently ignored by the API, so the option was previously a no-op.
  - Thinking is configured through explicit provider options only.
- Public settings and provider options are unchanged. Commits are split impl / test / docs; includes docs, an example, tests, and a patch changeset.

## End-to-End Verification

Verified against the live Moonshot API:

- Video input on `kimi-k3`, `kimi-k2.6`, `kimi-k2.5` (inline and URL mp4): accurate video summaries; wire shape confirmed as `video_url` + data URI.
- `reasoningHistory: 'preserved'` produces `thinking.keep: 'all'` on the wire; `budget_tokens` honored.
- Full moonshot example suite (10 examples): 7/10 pass; the 3 failures are a pre-existing stale model id (`kimi-k2-thinking` no longer served), unrelated to this change.
- 56 unit tests (converter + test-server model tests) pass on node and edge; `pnpm check` and `pnpm type-check:full` clean.
- Dialect check via `baseURL` override: novita also accepts `video_url`; fireworks recognizes it but requires `video_placeholder`; togetherai silently drops video with a 200; baseten/nebius/morph/wafer reject. Confirms the dialect is Moonshot-specific and belongs in this package.
- API rule found: `tool_choice: 'required'` conflicts with default-on thinking on `kimi-k2.6`/`kimi-k2.5` (`kimi-k3` accepts). Unchanged behavior, noted for follow-up.

## Checklist

## Future Work

- Backport to `release-v6.0` (spec v3) so AI SDK 6 users and the AI Gateway's v3 surfaces get video input.
- Optional: warning for `toolChoice: 'required'` + thinking-enabled models; Fireworks `videoPlaceholder`; refresh stale `kimi-k2-thinking` example ids.

## Related Issues

- Follows up on #12224 (`video_url` was flagged as a known Moonshot specialization when the provider was added).
- Alternative to #18328, which added the same capability as a gated flag in the shared openai-compatible converter. This PR implements the provider-owned direction instead.


packages/moonshotai/src/moonshotai-chat-language-model.ts
```diff
@@ -24,56 +72,385 @@ export class MoonshotAIChatLanguageModel extends OpenAICompatibleChatLanguageMod
...
+    seed,
+    toolChoice,
+    tools,
+  }: LanguageModelV4CallOptions) {
+    const moonshotOptions =
+      (await parseProviderOptions({
+        provider: this.providerOptionsName,
+        providerOptions,
+        schema: moonshotaiLanguageModelOptions,
+      })) ?? {};
```
shaper, line 126: [P1] Preserve the existing raw provider-option pass-through here. The inherited OpenAI-compatible implementation explicitly forwarded unrecognized entries from `providerOptions.moonshotai`; parsing with this default Zod object strips them, and the request only reconstructs the three modeled options. Existing calls using Moonshot fields such as `prompt_cache_key` or `safety_identifier` will silently stop working in this patch. Please preserve those entries (or explicitly model/translate all supported fields) and add a regression test.

packages/moonshotai/src/moonshotai-chat-language-model.ts
```diff
@@ -24,56 +72,385 @@ export class MoonshotAIChatLanguageModel extends OpenAICompatibleChatLanguageMod
...
+    const thinking = moonshotOptions.thinking;
+
+    // Moonshot has no reasoning_history field; the API silently ignores it
+    // (verified against the live API). Preserved Thinking maps to
+    // thinking.keep, which only accepts 'all'. Other reasoningHistory values
+    // use the server default.
+    const keep =
+      moonshotOptions.reasoningHistory === 'preserved'
+        ? ('all' as const)
+        : undefined;
```
shaper: [P1] Please gate `thinking.keep` by model. Moonshot only accepts this for K2.6 (and treats it as always-on for K2.7 Code); K3 rejects the `thinking` object entirely, and K2.5 does not support `thinking.keep`. As written, `reasoningHistory: preserved` turns otherwise valid K3/K2.5 calls into 400s. Omit or warn on unsupported models.

packages/moonshotai/src/moonshotai-chat-language-model.ts
```diff
@@ -1,20 +1,68 @@
...
+};
+
+export class MoonshotAIChatLanguageModel implements LanguageModelV4 {
+  readonly specificationVersion = 'v4';
+
+  readonly modelId: MoonshotAIChatModelId;
+
+  // Moonshot AI does not fetch external URLs; the AI SDK downloads and
+  // inlines URL file parts instead.
+  readonly supportedUrls = {};
```
shaper: [P2] `supportedUrls = {}` makes the documented Files API route unusable. Moonshot file uploads are referenced as `ms://<file-id>` inside `video_url`; the high-level AI SDK sees that URL as unsupported and tries to download it, which fails. The converter also rejects provider references. Please either mark `ms://` as natively supported and document that shape, or resolve `{ moonshotai: fileId }` references to `ms://...`.

packages/moonshotai/src/moonshotai-chat-language-model.ts
```diff
@@ -24,56 +72,385 @@ export class MoonshotAIChatLanguageModel extends OpenAICompatibleChatLanguageMod
...
+    topK,
+    frequencyPenalty,
+    presencePenalty,
+    providerOptions,
+    stopSequences,
+    responseFormat,
+    seed,
+    toolChoice,
+    tools,
+  }: LanguageModelV4CallOptions) {
```
shaper, line 120: [P2] The fork silently drops the standard `reasoning` call option because it is no longer destructured. The previous base class forwarded supported custom values as `reasoning_effort`; K3 callers using `reasoning: low` or `reasoning: high` now get the default without a warning. Please map supported levels (including `xhigh` to `max` if appropriate) or emit an unsupported warning.

packages/moonshotai/src/moonshotai-chat-language-model.ts
shaper: above was from agent directly -- I am less confident we want silent pass-through (some providers do it, but at least at one point AI SDK was explicitly not doing silent pass-through so folks knew what was supported -- @felixarntz do you know if that has changed?)

But if there are provider options we know some models support with this provider, we should explicitly include them.

joemckenney: Fixed in d6772b95ec: `promptCacheKey` and `safetyIdentifier` are now explicitly modeled in the provider options (both documented Moonshot request fields) with a wire-level regression test. Kept the passthrough drawer closed pending felix's answer on SDK policy; if the norm has changed to passthrough, easy to restore.

packages/moonshotai/src/moonshotai-chat-language-model.ts
joemckenney: Fixed in d6772b95ec: `thinking.keep` is now gated per model, omitted with an unsupported-option warning elsewhere. Verified live: `kimi-k2.6`, `kimi-k2.7-code`, and `kimi-k3` all accept `keep: 'all'` (k3 also accepted `thinking.type: 'disabled'` in an earlier probe, so the K3-rejects-thinking concern didn't reproduce), while `kimi-k2.5` 400s with "thinking.keep is not supported by model".

packages/moonshotai/src/moonshotai-chat-language-model.ts
joemckenney: Fixed in d6772b95ec: `ms://` is now declared natively supported in `supportedUrls`, so Files API references pass straight through into `video_url`/`image_url` with no download attempt (plus a converter test for the shape). Resolving `{ moonshotai: fileId }` references left as a follow-up.

packages/moonshotai/src/moonshotai-chat-language-model.ts
joemckenney: Fixed in d6772b95ec: the generic `reasoning` option maps to `reasoning_effort` again (minimal/low → 'low', medium/high → 'high', xhigh → 'max'; explicit `reasoningEffort` wins; 'none' warns since Moonshot thinking can't be disabled from this knob). Also widened the `reasoningEffort` schema to `'low' | 'high' | 'max'` per Moonshot's docs, verified `'low'` is accepted live.

shaper (approved):

PR #18468, opened 2026-08-05
fix(ai): fix stop() on resumed streams by passing abortSignal

### LINKED ISSUE #18458: stop() cannot end a resumed stream: reconnectToStream is called without an abort signal, replay keeps writing into chat state
## Description

`AbstractChat.stop()` aborts `activeResponse.abortController`, but on the `resume-stream` trigger that signal goes nowhere: `reconnectToStream` is invoked with only `{ chatId, metadata, headers, body }` — unlike `sendMessages`, which receives `abortSignal: response.abortController.signal`. The signal is only used to set an internal `isAbort` flag; nothing propagates to the transport or the stream, so `consumeStream` keeps draining the replay and it keeps writing messages into chat state after `stop()`.

#6502 reports the user-visible symptom ("stop() no longer works with resume"); filing this with the concrete root cause and a self-contained repro since that issue has neither. Practical consequences beyond stop-button-does-nothing: switching chats can leave a detached replay running, and two overlapping `resumeStream()` calls fight over the last message slot.

## Reproduction (self-contained, no network, no React)

```ts
// node --import tsx repro.mts   (ai@7.0.51)
import { AbstractChat, type ChatState, type ChatStatus, type ChatTransport, type UIMessage, type UIMessageChunk } from 'ai';

class State implements ChatState<UIMessage> {
  status: ChatStatus = 'ready';
  error: Error | undefined = undefined;
  constructor(public messages: UIMessage[]) {}
  pushMessage = (m: UIMessage) => { this.messages = [...this.messages, m]; };
  popMessage = () => { this.messages = this.messages.slice(0, -1); };
  replaceMessage = (i: number, m: UIMessage) => { this.messages = this.messages.map((x, j) => (j === i ? m : x)); };
  snapshot = <T,>(t: T): T => structuredClone(t);
}
class Chat extends AbstractChat<UIMessage> {}

const slowStream = (chunks: UIMessageChunk[], intervalMs: number) => {
  let i = 0;
  return new ReadableStream<UIMessageChunk>({
    async pull(c) {
      if (i >= chunks.length) return c.close();
      await new Promise((r) => setTimeout(r, intervalMs));
      c.enqueue(chunks[i++]);
    },
  });
};

const state = new State([{ id: 'u1', role: 'user', parts: [{ type: 'text', text: 'count to 30' }] }]);

const transport: ChatTransport<UIMessage> = {
  sendMessages: async () => slowStream([], 0),
  reconnectToStream: async () =>
    slowStream(
      [
        { type: 'start', messageId: 'server-msg-1' },
        { type: 'text-start', id: 't1' },
        ...Array.from({ length: 30 }, (_, i): UIMessageChunk => ({ type: 'text-delta', id: 't1', delta: `${i} ` })),
        { type: 'text-end', id: 't1' },
        { type: 'finish' },
      ],
      20,
    ),
};

const chat = new Chat({ id: 'c1', state, transport, generateId: () => 'client-generated' });
const done = chat.resumeStream();
await new Promise((r) => setTimeout(r, 120)); // let a few deltas land

const lastText = () => (state.messages.at(-1)?.parts.find((p) => p.type === 'text') as { text?: string })?.text ?? '';
const before = lastText();
await chat.stop();
await done;
console.log('text when stop() was called:', JSON.stringify(before));
console.log('text after stream ended:    ', JSON.stringify(lastText()));
```

Output:

```
text when stop() was called: "0 1 2 "
text after stream ended:     "0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 "
```

The replay wrote 27 more deltas into chat state after `stop()` resolved.

## Expected behavior

`stop()` should end a resumed stream the same way it ends a `sendMessages` stream — pass `response.abortController.signal` into `reconnectToStream` (transport interface addition) and/or cancel the returned stream's reader on abort so the drain loop stops.

## AI SDK Version

- ai: 7.0.51
- node: 26.1.0

haresrinivas:
Fixes #18458

When a stream is resumed, `reconnectToStream` fetches the remaining stream. However, it was not accepting an `AbortSignal`, meaning that if the user called `stop()` during a resumed stream, the underlying fetch request was never aborted and generation continued.

This PR:
- Adds `abortSignal?: AbortSignal` to `ChatTransport`'s `reconnectToStream` options.
- Updates `http-chat-transport` to pass this `abortSignal` to `fetch`.
- Refactors `makeRequest` in `ui/chat.ts` to instantiate the `AbortController` and populate `this.activeResponse` *before* the network request is triggered (for both stream resuming and normal submits).
- Wires the `abortController.signal` into `reconnectToStream`.
- Adds an automated test to `chat.test.ts` to ensure `stop()` successfully aborts a resumed stream.

### DIFF
--- a/.changeset/fix-resume-stream-stop.md
+++ b/.changeset/fix-resume-stream-stop.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+Fix stop() on resumed streams by passing abort signal to reconnectToStream
--- a/packages/ai/src/ui/chat-transport.ts
+++ b/packages/ai/src/ui/chat-transport.ts
@@ -78,6 +78,8 @@ export interface ChatTransport<UI_MESSAGE extends UIMessage> {
     options: {
       /** Unique identifier for the chat session to reconnect to */
       chatId: string;
+      /** Abort signal to cancel the request */
+      abortSignal?: AbortSignal;
     } & ChatRequestOptions,
   ) => Promise<ReadableStream<UIMessageChunk> | null>;
 }
--- a/packages/ai/src/ui/chat.test.ts
+++ b/packages/ai/src/ui/chat.test.ts
@@ -1546,6 +1546,55 @@ describe('Chat', () => {
     }
   });
 
+  it('should abort resume-stream when stop is called', async () => {
+    let isAborted = false;
+    let resumeController!: ReadableStreamDefaultController<UIMessageChunk>;
+    const resumeStream = new ReadableStream<UIMessageChunk>({
+      start(controller) {
+        resumeController = controller;
+        controller.enqueue({ type: 'start' });
+        controller.enqueue({ type: 'start-step' });
+        controller.enqueue({ type: 'text-start', id: 'text-1' });
+        controller.enqueue({
+          type: 'text-delta',
+          id: 'text-1',
+          delta: 'resumed',
+        });
+      },
+    });
+
+    const chat = new TestChat({
+      id: '123',
+      generateId: mockId(),
+      transport: {
+        sendMessages: async () => {
+          throw new Error('should not be called');
+        },
+        reconnectToStream: async options => {
+          options.abortSignal?.addEventListener('abort', () => {
+            isAborted = true;
+            resumeController.error(new DOMException('Aborted', 'AbortError'));
+          });
+          return resumeStream;
+        },
+      },
+      onFinish: () => {},
+    });
+
+    const resumePromise = chat.resumeStream();
+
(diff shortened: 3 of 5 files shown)


haresrinivas:
Fixed! I've wrapped `reconnectToStream` in its own try/catch and made sure to properly clear `activeResponse` before returning early. This prevents the `finally` block from erroneously firing `onFinish` when there's no active stream to resume. Thanks for catching that!

gr2m:
fixed via https://github.com/vercel/ai/pull/18461

PR #18487, opened 2026-08-05
policy-opa: add composed-judgment example

### LINKED ISSUE #18376: toolApproval: reference pattern for composing a judgment-based verdict as a complement to @ai-sdk/policy-opa (real code + live verification)
## Description

`toolApproval` already supports automatic decisions (`'approved'`/`'denied'` returned from a
plain async function — docs: `content/docs/03-agents/06-tool-approvals.mdx`), and the
reference composition for that is `@ai-sdk/policy-opa`: rules authored in `.rego`, evaluated
deterministically by OPA.

That's the right tool for "never let a non-admin call `sendPayment`" — a boolean rule over
known fields. It's structurally the wrong tool for "is *this specific* admin payment, in *this*
context, actually sound" — a question with no fixed rule to write, the same distinction
`docs/03-agents/06-tool-approvals.mdx` implicitly leaves open between "policy" and "approval."

A real user hit this gap directly in #15842: wanted "programmatic, server-side policy
enforcement based on runtime state," was pointed at `toolApproval` + `policy-opa`. That's the
right answer for deterministic policy — but nothing in that thread, the tool-approvals docs, or
the policy-opa README shows a judgment-based verifier (an LLM-backed second opinion, a
compliance check, anything non-deterministic) composed as a *complement* to OPA, only as an
alternative to human `user-approval`.

I built and live-verified a small reference implementation of exactly that composition — 100%
on the existing public `toolApproval` contract, no SDK changes:

```ts
toolApproval: {
  sendPayment: async (input, { runtimeContext }) => {
    // OPA handles the hard deterministic rules first
    const opaResult = await opaPolicy(...)(input, { runtimeContext });
    if (opaResult !== 'user-approval' && opaResult !== undefined) return opaResult;
    // fall through to a judgment call for the cases OPA can't resolve
    return invinoveritasToolApproval({ apiKey, artifactType: 'general' })({ toolCall: ... });
  },
}
```

Full code + README: https://github.com/babyblueviper1/invinoveritas/tree/main/integrations/vercel-ai-sdk

Live-verified against the real API (not mocked), both branches:
- benign `ls -la /tmp` → `{ type: "approved", reason: "...confidence 0.99..." }`
- destructive `rm -rf / --no-preserve-root` → `{ type: "denied", reason: "...will destroy the host OS..." }`

One real gotcha found while verifying, documented in the README: a judgment call with a
signed-proof response can take longer than a typical policy-eval timeout — a slow-but-healthy
gate can silently resolve to "allow" under a permissive fail-mode unless the caller checks for
an explicit "gate unavailable" signal rather than trusting any non-error response. Worth
knowing for anyone else composing a real network call into `toolApproval`.

## Reproduction

Runnable script: `integrations/vercel-ai-sdk/test-live.mjs` in the repo linked above
(`IVV_API_KEY=... npx tsx test-live.mjs`).

## AI SDK Version

ai: 6.0.190 (docs/source read against current `main` as of 2026-08-03)

## Suggested outcome

Not proposing any SDK changes — `toolApproval`'s existing contract already supports this
cleanly. Question: is a worked example like this (explicitly framed as a complement to
`@ai-sdk/policy-opa`, not a competing product) something that belongs in `examples/` next to
the policy-tool-approvals docs? Happy to open a PR there if useful — otherwise this can just
stay a linked community integration.

babyblueviper1:
## Summary

Per #18376 -- @gr2m asked for a PR adding the worked example. This adds `examples/mock/composed-judgment.ts` to `@ai-sdk/policy-opa`, next to the existing `examples/mock/basic.ts`.

Shows `@ai-sdk/policy-opa` composed with an independent judgment call for the cases a deterministic OPA rule can't resolve on its own:

```
tool call
  -> OPA evaluates  --+--> allow / deny (clear-cut)  -> resolved, no judgment call
                       +--> requires-approval          -> ask a judgment service
                                                            +--> confident verdict -> resolved
                                                            +--> still uncertain    -> SDK's human-in-the-loop UI
```

OPA owns a real, hard ceiling in the example (deny payments over $1M outright) and everything else falls through `requires-approval` to a judgment call instead of going straight to a human. The judgment call only resolves automatically above a confidence threshold; anything less still escalates to the SDK's own human-in-the-loop UI, same as OPA's own `requires-approval` would have.

`judgmentCall` is a minimal one-function interface (`(toolCall) => Promise<{verdict, confidence, reason}>`) so any judgment provider can be plugged in -- an in-house LLM-as-judge, a compliance review service, a hosted verdict API. This example stubs it deterministically (no network call, no API key) so it runs the same way `examples/mock/basic.ts` does. A concrete, live-verified implementation against a real judgment API is linked in the module docstring for anyone who wants the non-stubbed version (swapping it in is a one-line change, same as swapping `MockLanguageModelV3` for a real provider).

## Verified

- `pnpm tsx examples/mock/composed-judgment.ts` runs clean, all 3 scenarios produce the expected output (judgment-approved, judgment-denied, OPA-denied-outright-before-judgment-is-consulted).
- `tsc --noEmit` clean against the package's own tsconfig.
- `oxlint` clean (0 warnings, 0 errors).

## Test plan

- [x] Runs standalone via `pnpm tsx examples/mock/composed-judgment.ts`
- [x] Type-checks against the package's tsconfig
- [x] Lints clean

Closes #18376.

### DIFF
--- a/.changeset/silly-rivers-attend.md
+++ b/.changeset/silly-rivers-attend.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/policy-opa': patch
+---
+
+Add a runnable example composing `@ai-sdk/policy-opa` with an independent judgment call for cases the deterministic policy can't resolve on its own (`examples/mock/composed-judgment.ts`).
--- a/.changeset/silly-rivers-attend.md
+++ b/.changeset/silly-rivers-attend.md
@@ -2,4 +2,4 @@
 '@ai-sdk/policy-opa': patch
 ---
 
-Add a runnable example composing `@ai-sdk/policy-opa` with an independent judgment call for cases the deterministic policy can't resolve on its own (`examples/mock/composed-judgment.ts`).
+Add a runnable example composing `@ai-sdk/policy-opa` with contextual judgment and human-approval fallbacks, and document the package examples.
--- a/packages/policy-opa/README.md
+++ b/packages/policy-opa/README.md
@@ -70,6 +70,17 @@ const result = await generateText({
 });
 ```
 
+## Examples
+
+- [`examples/mock/basic.ts`](./examples/mock/basic.ts) demonstrates direct
+  allow and deny decisions plus transitive enforcement through a composite
+  shell tool.
+- [`examples/mock/composed-judgment.ts`](./examples/mock/composed-judgment.ts)
+  composes deterministic OPA decisions with a contextual judgment service and
+  human-approval fallback.
+- [`examples/git-in-bash`](./examples/git-in-bash) contains a complete Rego
+  policy, parser, tests, and `generateText` demo for transitive git enforcement.
+
 ## Writing the Rego policy
 
 The adapter expects the policy to emit a decision object with one of three `decision` values. `reason` is optional and gets surfaced back to the model (for `deny`) or to the human approver (for `requires-approval`).
--- a/packages/policy-opa/examples/mock/composed-judgment.ts
+++ b/packages/policy-opa/examples/mock/composed-judgment.ts
@@ -1,0 +1,297 @@
+/**
+ * Runnable demo composing `@ai-sdk/policy-opa` with a judgment-based second
+ * opinion, for the cases a deterministic policy can't resolve on its own.
+ *
+ * Run from the package directory:
+ *   pnpm tsx examples/mock/composed-judgment.ts
+ *
+ * WHY THIS COMPOSITION
+ *
+ * OPA is the right tool for a hard, deterministic rule ("never let a
+ * non-admin call `sendPayment`") -- a boolean expression over known fields.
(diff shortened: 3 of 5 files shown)


babyblueviper1:
Thanks for strengthening this -- the runtime-context threading (accountId/approvedRecipients/largestApprovedPayment) and the network-failure fallback to `user-approval` are real improvements, and the `uncertain` verdict path fills a gap my stub didn't have. Formatting + changeset are in place from my last push; looks ready whenever CI clears.

gr2m (approved):

babyblueviper1:
Friendly bump -- this has been green + approved since Aug 11, just flagging in case it fell off the radar. Happy to rebase/adjust if anything's needed before merge.

PR #18516, opened 2026-08-06
fix(openai): make web search sources include opt-outable via compatibility mode

### LINKED ISSUE #18507: `web_search` tool unconditionally adds `include: 'web_search_call.action.sources'`, breaking OpenAI-compatible providers (Amazon Bedrock Mantle rejects it)
### Description

When a `openai.web_search` provider-executed tool is present, `OpenAIResponsesLanguageModel`
unconditionally appends `'web_search_call.action.sources'` to the `include` array of the
outgoing Responses API request:

https://github.com/vercel/ai/blob/main/packages/openai/src/responses/openai-responses-language-model.ts#L405-L407

```ts
if (webSearchToolName) {
  addInclude('web_search_call.action.sources');
}
```

There is no way to opt out: `providerOptions.openai.include` is treated as a base list that
`addInclude` only appends to (with de-duplication), so user-supplied options can add include
values but never remove this one.

### Why this matters: it breaks an AI SDK first-party provider path

[Amazon Bedrock's **Web Search** (GA on 2026-08-04)](https://aws.amazon.com/jp/about-aws/whats-new/2026/08/amazon-bedrock-web/) is served through the OpenAI-compatible
`bedrock-mantle` endpoint and accepts the `web_search` built-in tool. The
`@ai-sdk/amazon-bedrock/mantle` provider reuses `OpenAIResponsesLanguageModel` from
`@ai-sdk/openai/internal`, so it inherits this behavior. Bedrock Mantle accepts the tool but
rejects the auto-added include value, which fails the entire request:

```
Unsupported value: 'web_search_call.action.sources' is not supported with the
'openai.gpt-5.6-terra' model. Supported values are: 'reasoning.encrypted_content'
and 'message.output_text.logprobs'.
```

So today it is impossible to use Bedrock Web Search through the AI SDK, even though the wire
format of the tool itself (`{"type":"web_search","external_web_access":false}`) is generated
correctly by the SDK.

The usual workaround — stripping the field in a custom `fetch` — is **not available** here:
Bedrock Mantle uses AWS SigV4 request signing, and the provider computes the signature over
the request body *before* the custom `fetch` (transport) is invoked. Rewriting the body at the
fetch layer invalidates the signature. The only workaround we found is patching the installed
`dist` files, which does not survive a reinstall.

A related precedent: #8844 reported the AI Gateway rejecting this same include value, which
suggests this hard-coded include has caused compatibility friction before.

### Reproduction

No AWS credentials needed — a mock `fetch` captures the outgoing request body:

```ts
import { createBedrockMantle } from '@ai-sdk/amazon-bedrock/mantle';

const mantle = createBedrockMantle({
  region: 'us-east-1',
  baseURL: 'https://bedrock-mantle.us-east-1.api.aws/openai/v1',
  apiKey: 'dummy-bearer-token',
    console.log(JSON.parse(init.body).include);
    // => ["web_search_call.action.sources"] — always present, cannot be removed
    return new Response(
      JSON.stringify({
        id: 'resp_test',
        object: 'response',
        created_at: 0,
        model: 'openai.gpt-5.6-terra',
        status: 'completed',
        incomplete_details: null,
        output: [
          {
            type: 'message',
            id: 'msg_1',
            role: 'assistant',
            status: 'completed',
            content: [{ type: 'output_text', text: 'ok', annotations: [] }],
          },
        ],
        usage: {
          input_tokens: 1,
          output_tokens: 1,
          input_tokens_details: { cached_tokens: 0 },
          output_tokens_details: { reasoning_tokens: 0 },
        },
      }),
      { status: 200, headers: { 'content-type': 'application/json' } },
    );
  },
});

const model = mantle.responses('openai.gpt-5.6-terra');
await model.doGenerate({
  prompt: [{ role: 'user', content: [{ type: 'text', text: 'test' }] }],
  tools: [
    {
      type: 'provider',
      id: 'openai.web_search',
      name: 'web_search',
      args: { externalWebAccess: false },
    },
  ],
});
```

Against the real Bedrock Mantle endpoint (SigV4 credentials, `openai.gpt-5.6-terra`,
us-east-1), the same request is rejected with the error quoted above. The same code path is
hit when using `createOpenAI` with a custom `baseURL` pointed at any OpenAI-compatible server
that does not support this include value.

### AI SDK Version

- `@ai-sdk/openai` 4.0.27 (as pinned by `@ai-sdk/amazon-bedrock` 5.0.40) — reproduced
- `@ai-sdk/openai` 4.0.31 (latest at time of writing) — same unconditional `addInclude` in
  `packages/openai/src/responses/openai-responses-language-model.ts` (verified against the
  npm tarball)

### Code of Conduct

AadiyKhan:
## Background

Fixes #18507

When a `openai.web_search` tool is present, `OpenAIResponsesLanguageModel` automatically appends `web_search_call.action.sources` to the `include` array. This breaks OpenAI-compatible endpoints like Bedrock Mantle and AI Gateway which reject this `include` value, and previously there was no way for users to opt-out.

## Summary

- Added `compatibility` option (`'strict' | 'compatible'`) to `OpenAIConfig` and `OpenAIProviderSettings`.
- In `OpenAIResponsesLanguageModel`, the `web_search_call.action.sources` include is now conditionally omitted if `compatibility === 'compatible'`. 
- Updated `bedrock-mantle-provider` to default to `compatible` mode so it gracefully accepts the web search tool out-of-the-box.
- Added a patch changeset for `@ai-sdk/amazon-bedrock` and `@ai-sdk/openai`.

## Contributor Credit

- Thanks to @jacobwgillespie for providing the reproduction script and identifying the issue!

## End-to-End Verification

- Verified successfully by running all tests via `pnpm test` locally.
- Verified that Bedrock Mantle snapshot behavior behaves as expected under the new `"compatibility": "compatible"` default without adding `content-length` overhead or unrecognized properties.

## Checklist

## Related Issues

Fixes #18507

### DIFF
--- a/.changeset/fix-web-search-compatibility.md
+++ b/.changeset/fix-web-search-compatibility.md
@@ -1,0 +1,6 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+"@ai-sdk/openai": patch
+---
+
+fix: opt-out of auto-injecting web_search_call.action.sources in compatible mode
--- a/packages/amazon-bedrock/src/mantle/bedrock-mantle-provider.ts
+++ b/packages/amazon-bedrock/src/mantle/bedrock-mantle-provider.ts
@@ -249,6 +249,7 @@ export function createBedrockMantle(
       url,
       headers: getHeaders,
       fetch: fetchFunction,
+      compatibility: 'compatible',
     });
 
   const provider = function (modelId: BedrockMantleChatModelId) {
--- a/packages/amazon-bedrock/src/reranking/amazon-bedrock-reranking-model.test.ts
+++ b/packages/amazon-bedrock/src/reranking/amazon-bedrock-reranking-model.test.ts
@@ -172,7 +172,7 @@ describe('doRerank', () => {
             ],
           },
           "headers": {
-            "content-length": "171",
+            "content-length": "183",
             "content-type": "application/json",
           },
         }
@@ -302,7 +302,7 @@ describe('doRerank', () => {
             ],
           },
           "headers": {
-            "content-length": "171",
+            "content-length": "183",
             "content-type": "application/json",
           },
         }
--- a/packages/baseten/package.json
+++ b/packages/baseten/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "pnpm clean && tsup --tsconfig tsconfig.build.json",
     "build:watch": "pnpm clean && tsup --watch",
-    "clean": "rm -rf dist docs *.tsbuildinfo",
+    "clean": "del-cli dist docs *.tsbuildinfo",
     "prepack": "mkdir -p docs && cp ../../content/providers/01-ai-sdk-providers/170-baseten.mdx ./docs/",
     "postpack": "del-cli docs",
     "type-check": "tsc --build",
--- a/packages/huggingface/package.json
+++ b/packages/huggingface/package.json
@@ -23,7 +23,7 @@
   "scripts": {
     "build": "pnpm clean && tsup --tsconfig tsconfig.build.json",
     "build:watch": "pnpm clean && tsup --watch",
-    "clean": "rm -rf dist docs *.tsbuildinfo",
+    "clean": "del-cli dist docs *.tsbuildinfo",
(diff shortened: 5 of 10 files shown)


aayush-kapoor (commented):
dont think this is the right solution + has unnecessary changes

PR #18523, opened 2026-08-06
fix: strip OpenAI-unsupported keywords from tool schemas

### LINKED ISSUE #18510: AI Gateway streams early server_error for GPT-5.6 Terra and Luna requests with tools
### Description

`streamText` requests through Vercel AI Gateway terminate near the start of the stream with an upstream OpenAI `server_error` when using both `openai/gpt-5.6-terra` and `openai/gpt-5.6-luna`.

Switching the main and sub-agent models from Luna to Terra did not resolve the failure. The deployed Terra configuration produced the same raw event on multiple requests:

```json
{
  "type": "error",
  "sequence_number": 3,
  "error": {
    "type": "server_error",
    "code": "server_error",
    "message": "An error occurred while processing your request. You can retry your request, or contact us through our help center at help.openai.com if the error persists. Please include the request ID req_0ed1fa39acd5497b92551f614bb9afcb in your message.",
    "param": null
  }
}
```

Confirmed upstream request IDs:

- Terra: `req_0ed1fa39acd5497b92551f614bb9afcb` at 2026-08-06 04:27:14 UTC
- Terra: `req_f2375dc9883146178f1878e3508267eb` at 2026-08-06 04:33:46 UTC
- Luna: `req_2b3321ac36ec442d90528bd3f0b4d0d4`

The requests use function tools generated from Zod schemas. The agent configuration is equivalent to:

```ts
const model = gateway.chat('openai/gpt-5.6-terra')

streamText({
  model,
  instructions,
  messages,
  reasoning: 'high',
  stopWhen: isStepCount(50),
  tools,
})
```

`temperature` and `maxOutputTokens` are omitted. Tool `strict` is also omitted.

An earlier explicit Gateway validation error for regex lookaround in an email schema was addressed by using Zod's HTML5 email pattern. The current failure is different: it reaches the provider stream and returns only the generic OpenAI `server_error`.

This may be related to #16021, which reports an early Responses API `server_error` involving Zod v4 `pattern` keywords on GPT-5.4. That issue's maintainer reproductions succeeded against OpenAI directly. This report differs in that it uses AI SDK 7, `@ai-sdk/gateway`, and GPT-5.6 Terra/Luna. We have not established that a JSON Schema `pattern` is the trigger here.

OpenAI support has been given the request IDs and has escalated the incident. They asked for a direct `/v1/responses` control and a tools-disabled control. We currently have Gateway authentication rather than a direct OpenAI API credential, so we cannot run the equivalent direct-provider control without changing the production integration.

Expected behavior is either a completed response or an actionable validation/provider error. For Gateway-routed failures like this, it would also help to have a documented way to retrieve the Gateway generation ID and sanitized transformed provider request from the SDK error so the failing payload can be compared with a direct OpenAI request.

### Reproduction

The failure is repeatable with the production agent's real conversation and tool set. We do not yet have a sanitized minimal payload that reproduces it independently; reducing the request changes the aggregate schema and context that may be part of the trigger.

A representative tool schema is:

```json
{
  "type": "object",
  "properties": {
    "values": {
      "type": "array",
      "items": { "type": "string" },
      "minItems": 1,
      "maxItems": 30000
    }
  },
  "required": ["values"],
  "additionalProperties": false
}
```

This example is illustrative and is not claimed to reproduce the failure by itself. We can provide additional sanitized request metadata privately if there is a preferred capture path.

### AI SDK Version

- `ai`: 7.0.28
- `@ai-sdk/gateway`: 4.0.20
- Models: `openai/gpt-5.6-terra`, previously `openai/gpt-5.6-luna`

### Code of Conduct

Mahnoor-Zaffar:
Fixes #18510.

`streamText` with tools dies right after the first chunk with a generic OpenAI `server_error` on `gpt-5.6-*`. OpenAI compiles function tool schemas into a strict grammar and rejects a chunk of JSON Schema keywords — `minItems`/`maxItems`, `minLength`/`maxLength`, `pattern`, `format`, etc. Zod v4 emits these for common expressions like `z.array(z.string()).min(1).max(30000)`, so users hit it without doing anything unusual.

This mirrors how Anthropic already handles it (same keyword stripping, folded into the description), and applies it:

- `@ai-sdk/openai`: function tool `parameters` and `output_schema` in both Responses and chat paths.
- `@ai-sdk/gateway`: tool schemas for `openai/`-routed requests.

The original schema still validates tool inputs on the SDK side. Gateway + OpenAI credentials weren't available here, so the repro above wasn't run against the hosted gateway end-to-end — worth a quick confirmation after release.

### DIFF
--- a/.changeset/early-server-error-strict-tools.md
+++ b/.changeset/early-server-error-strict-tools.md
@@ -1,0 +1,22 @@
+---
+'@ai-sdk/openai': patch
+'@ai-sdk/gateway': patch
+'@ai-sdk/provider-utils': patch
+---
+
+fix: strip JSON Schema keywords OpenAI's strict tool schemas reject
+
+OpenAI's Responses API defaults function tools to `strict: true` and only
+supports a subset of JSON Schema. Keywords like `minItems`/`maxItems`,
+`minLength`/`maxLength`, `pattern`, `format`, and `minimum`/`maximum` can make
+OpenAI fail with a generic in-stream `server_error` on gpt-5.x models while it
+compiles the schema into a constrained-decoding grammar. This is what the
+default Zod v4 conversion produces for common schemas (e.g.
+`z.array(z.string()).min(1).max(30000)`), so AI SDK users hit it without
+writing any problematic schema themselves.
+
+Adds a shared `sanitizeJsonSchema` util that strips the OpenAI-unsupported
+keywords (folding their values into the schema `description` as guidance) and
+applies it to function tool schemas in `@ai-sdk/openai` (Responses and chat
+completions) and to requests routed to OpenAI through `@ai-sdk/gateway`. The
+full original schema is still used for SDK-side tool input validation.
--- a/packages/gateway/src/gateway-language-model.test.ts
+++ b/packages/gateway/src/gateway-language-model.test.ts
@@ -1,4 +1,5 @@
 import type {
+  LanguageModelV4FunctionTool,
   LanguageModelV4Prompt,
   LanguageModelV4FilePart,
 } from '@ai-sdk/provider';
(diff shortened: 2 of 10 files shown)


gr2m:
closing because #18510 was closed as out of AI SDK scope

PR #18525, opened 2026-08-06
fix(react): throttle published useChat snapshots

gr2m:
## Background

`useChat` currently throttles its messages subscription callback, but its `useSyncExternalStore` snapshot always reads the latest `chat.messages` array. Because streaming replaces that array for every chunk, any unrelated React render can observe a new snapshot before the throttled callback publishes it. In high-frequency streams this bypasses `throttle`, causes per-chunk renders, and can contribute to the "Maximum update depth exceeded" failures reported in #6166.

## Summary

- Keep a published messages snapshot per `useChat` hook and advance it from that hook's throttled subscription callback.
- Publish the latest message snapshot before `ready` or `error` becomes observable, including normal completion and aborts.
- Synchronize updates that occur between render and subscription, and ignore delayed callbacks after a hook unsubscribes or changes chat instances.
- Add regression coverage for unrelated renders, terminal status/message coherence, aborts, errors, and delayed callbacks after chat replacement.
- Turn the existing Next.js throttle route into a deterministic end-to-end reproduction that streams 500 chunks while forcing unrelated renders, reports pass/fail from observed snapshot identities, and verifies the complete message is visible when status becomes `ready`.
- Add a patch changeset for `@ai-sdk/react`.

## Contributor Credit

- @brahmveda-arkin reported #6166.
- @takumiz19 isolated the snapshot/subscription mismatch and proposed #17893.
- @ben-reitz demonstrated the practical value of a conservative UI update cadence in cloudflare/agents#2058.

## End-to-End Verification

Ran `/chat/throttle` in `examples/ai-e2e-next` in a real browser. The route streams 500 chunks (1,000 assistant characters) with `throttle: 50` while a zero-delay timer independently re-renders the component.

- Before: **FAIL**, 255 distinct message snapshots in 1,475ms (maximum expected: 34), across 946 total React renders.
- After: **PASS**, 15 distinct message snapshots in 1,144ms (maximum expected: 27), across 602 total React renders.

The patched run rendered all 1,000 assistant characters on the same render where status became `ready`, and had no Next.js error overlay or browser error.

## Checklist

## Future Work

This PR intentionally leaves the current opt-in default unchanged, so it does not protect applications that omit `throttle` from high-frequency unthrottled rendering.

For v8, I recommend making a 50ms UI publication cadence the default when `throttle` is omitted, with `throttle: 0` as the explicit unthrottled opt-out. Stream processing, tool handling, and callbacks should remain immediate; only snapshots exposed to React should be paced. The default should also guarantee an immediate leading publication and a terminal flush so the final messages and `ready` status stay coherent.

This would cap the normal rendering rate at about 20 updates per second and protect applications that do not know they need to opt in today. The tradeoff is up to 50ms of additional visible text latency and an explicit opt-out for applications that intentionally need per-chunk rendering, which makes the behavior change appropriate for a major release.

## Related Issues

Addresses the throttled snapshot bypass discussed in #6166. Reports using the default unthrottled behavior remain outside this PR.

Closes #17893.

Related to https://github.com/cloudflare/agents/pull/2058.

### DIFF
--- a/examples/ai-e2e-next/app/api/chat/throttle/route.ts
+++ b/examples/ai-e2e-next/app/api/chat/throttle/route.ts
@@ -3,8 +3,8 @@ import { createUIMessageStreamResponse, simulateReadableStream } from 'ai';
 export async function POST(req: Request) {
   return createUIMessageStreamResponse({
     stream: simulateReadableStream({
-      initialDelayInMs: 0, // Delay before the first chunk
-      chunkDelayInMs: 0, // Delay between chunks
+      initialDelayInMs: 0,
+      chunkDelayInMs: 0,
       chunks: [
         {
           type: 'start',
@@ -16,7 +16,7 @@ export async function POST(req: Request) {
           type: 'text-start',
           id: 'text-1',
         },
-        ...Array(5000).fill({ type: 'text-delta', id: 'text-1', delta: 'T\n' }),
+        ...Array(500).fill({ type: 'text-delta', id: 'text-1', delta: 'T\n' }),
         {
           type: 'text-end',
           id: 'text-1',
--- a/.changeset/calm-chats-throttle.md
+++ b/.changeset/calm-chats-throttle.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/react': patch
+---
+
+Fix `useChat` throttling so unrelated React renders cannot publish message snapshots ahead of the configured throttle cadence.
--- a/examples/ai-e2e-next/app/chat/throttle/page.tsx
+++ b/examples/ai-e2e-next/app/chat/throttle/page.tsx
@@ -1,36 +1,157 @@
 'use client';
 
-import ChatInput from '@/components/chat-input';
 import { useChat } from '@ai-sdk/react';
 import { DefaultChatTransport } from 'ai';
-import { useLayoutEffect, useRef } from 'react';
+import { useEffect, useRef, useState } from 'react';
+
+const THROTTLE_MS = 50;
+const EXPECTED_ASSISTANT_CHARACTERS = 1000;
+
+type Result = {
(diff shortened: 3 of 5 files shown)


gr2m:
Addressed in 3e8f458eb6:

- Terminal `ready` and `error` notifications now publish the latest hook-local messages snapshot before scheduling the status update. This also covers aborts, which transition back to `ready`.
- Added focused throttled tests for normal completion, stream errors, aborts, and a delayed callback after changing chat instances. The React package now passes all 94 tests.
- Strengthened the browser reproduction so it records the assistant length on the first `ready` render instead of waiting for a later message update. It passed with 1,000/1,000 characters present at `ready`, 15 message snapshot changes over 1,144ms, and 602 total React renders.
- Clarified the PR scope: this fixes configured throttling and no longer claims to close the unthrottled reports in #6166. Changing the omitted default remains proposed for v8 because of the rendering-semantics tradeoff.

`pnpm check` and `pnpm type-check:full` pass.

examples/ai-e2e-next/app/api/chat/throttle/route.ts
```diff
@@ -16,7 +16,7 @@ export async function POST(req: Request) {
           type: 'text-start',
           id: 'text-1',
         },
-        ...Array(5000).fill({ type: 'text-delta', id: 'text-1', delta: 'T\n' }),
```
n1ckoates, line 19: Why update this?

n1ckoates (approved):

examples/ai-e2e-next/app/api/chat/throttle/route.ts
gr2m: 500 is sufficient and matches `EXPECTED_ASSISTANT_CHARACTERS`

PR #18572, opened 2026-08-07
fix(provider-utils): preserve streamed size-limit errors during cleanup

### LINKED ISSUE #18571: readResponseWithSizeLimit can replace streamed size-limit errors with cancellation errors
### Description

`readResponseWithSizeLimit()` can correctly detect that a streamed response exceeded `maxBytes`, then replace that `DownloadError` if reader cleanup fails.

The streamed path currently ends with:

```ts
finally {
  try {
    await reader.cancel();
  } finally {
    reader.releaseLock();
  }
}
```

The adjacent `cancelResponseBody()` helper already ignores cancellation failures to preserve the original rejection, including on the Content-Length size-limit path.

### Reproduction

Against current `main` at `ff0f708e917e3f556f889ed9158b396a17f3828c`, a native `ReadableStream` can reproduce the issue:

```ts
const cancelError = new Error('cancel failed');

const body = new ReadableStream<Uint8Array>({
  start(controller) {
    controller.enqueue(new Uint8Array([1, 2]));
  },
  cancel() {
    return Promise.reject(cancelError);
  },
});

const response = {
  headers: new Headers(),
  body,
} as Response;

await readResponseWithSizeLimit({
  response,
  url: 'https://example.com/file',
  maxBytes: 1,
});
```

The two-byte chunk exceeds the one-byte limit, so the operation should fail with a `DownloadError`.

Current result:

```text
Error: cancel failed
```

Expected result:

```text
DownloadError: Download of https://example.com/file exceeded maximum size of 1 bytes.
```

### Expected behavior

Cancellation should still be attempted and the reader lock released, but a cancellation failure shouldn't replace the `DownloadError` that triggered cleanup.

### AI SDK Version

```text
ai: 7.0.56
```

### Code of Conduct

teamleaderleo:
## Background

`readResponseWithSizeLimit()` cancels its reader in `finally`.

When streamed bytes exceed `maxBytes`, the function has already selected a `DownloadError`. If the underlying stream's cancellation algorithm then rejects, the cancellation error currently replaces that size-limit error.

The adjacent `cancelResponseBody()` helper already follows the opposite cleanup policy: cancellation failures are contained so they don't mask the operation's original rejection.

## Summary

* contain rejection from terminal `reader.cancel()` cleanup;
* always release the reader lock;
* add a native `ReadableStream` regression where streamed overflow is followed by rejecting cancellation;
* add a patch changeset for `@ai-sdk/provider-utils`.

The read loop, byte accounting, Content-Length behavior, returned bytes, and public API aren't changed.

## End-to-End Verification

A native Web Streams reproduction against the unfixed current-main implementation sends two bytes with a one-byte size limit and makes the underlying stream cancellation reject.

Before the change, the observed rejection is:

```text
Error: cancel failed
```

With this change, the same reproduction preserves the intended size-limit `DownloadError`, cancellation is still attempted, and the stream is unlocked afterward.

## Checklist

* [x] All commits are signed
* [x] Tests have been added / updated
* [ ] Documentation has been added / updated
* [x] A patch changeset for the relevant package has been added
* [x] I've reviewed this pull request

## Related Issues

* Closes #18571

### DIFF
--- a/.changeset/tidy-readers-preserve-errors.md
+++ b/.changeset/tidy-readers-preserve-errors.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/provider-utils': patch
+---
+
+Preserve streamed download size errors when reader cancellation fails during cleanup.
--- a/packages/provider-utils/src/read-response-with-size-limit.test.ts
+++ b/packages/provider-utils/src/read-response-with-size-limit.test.ts
@@ -1,6 +1,6 @@
-import { describe, it, expect } from 'vitest';
-import { readResponseWithSizeLimit } from './read-response-with-size-limit';
+import { describe, expect, it, vi } from 'vitest';
 import { DownloadError } from './download-error';
+import { readResponseWithSizeLimit } from './read-response-with-size-limit';
 
 function createMockResponse({
   body,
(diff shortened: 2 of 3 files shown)


ArshVermaGit (commented):
Looks good. This preserves the intended size-limit error during cleanup, maintains proper resource handling, and adds solid regression coverage for the edge case. LGTM.

PR #18586, opened 2026-08-07
feat(video): allow `aspectRatio: 'adaptive'` on generateVideo

joshualipman123:
video models whose output ratio is dictated by the input reject an explicit `{width}:{height}` ratio. seedance is the concrete case: for first-frame image-to-video, first-and-last-frame, video editing, and video extension, `ratio` only accepts `adaptive` and any numeric ratio returns a 400.

widen `aspectRatio` to `${number}:${number} | 'adaptive'` on VideoModelV3CallOptions, VideoModelV4CallOptions, and experimental_generateVideo.

### DIFF
--- a/packages/provider/src/video-model/v3/video-model-v3-call-options.ts
+++ b/packages/provider/src/video-model/v3/video-model-v3-call-options.ts
@@ -16,11 +16,12 @@ export type VideoModelV3CallOptions = {
 
   /**
    * Aspect ratio of the videos to generate.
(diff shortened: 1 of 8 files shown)


packages/provider/src/video-model/v3/video-model-v3-call-options.ts
aayush-kapoor, line 1: this file change should only be a part of the backport if the aim is to include this in v6 as well

aayush-kapoor (approved):

PR #18593, opened 2026-08-07
fix(WorkflowAgent): apply maxRetries and abortSignal from prepareCall

### LINKED ISSUE #18576: WorkflowAgent: `maxRetries` and `abortSignal` returned from `prepareCall` are silently ignored
### Description

`WorkflowAgent`'s `PrepareCallResult` is `Partial<Omit<PrepareCallOptions, 'tools'>>`, and `PrepareCallOptions extends Partial<GenerationSettings>`, so returning any generation setting from `prepareCall` type-checks — including `maxRetries` and `abortSignal`. But unlike `ToolLoopAgent`, `WorkflowAgent.stream()` applies the `prepareCall` result field by field (`packages/workflow/src/workflow-agent.ts`), and the apply block copies every `GenerationSettings` key except those two:

```ts
if (prepared.seed !== undefined)
  effectiveGenerationSettings.seed = prepared.seed;
// no prepared.maxRetries, no prepared.abortSignal
if (prepared.headers !== undefined)
  effectiveGenerationSettings.headers = prepared.headers;
```

So both values are silently dropped:

- `maxRetries` returned from `prepareCall` never lands in the merged settings; the call proceeds with the constructor/per-stream value or the default (`mergedGenerationSettings.maxRetries ?? 2`).
- `abortSignal` returned from `prepareCall` never reaches the signal merge, which only reads `options.abortSignal ?? effectiveGenerationSettings.abortSignal` (per-stream and constructor values). The loop's `abortSignal?.aborted` checks and `streamModelCall` never see it.

The surrounding code suggests this is an oversight rather than a design choice: the same block applies the other 11 generation settings (`maxOutputTokens` through `providerOptions`), and `prepareStep` *does* honor `maxRetries` via the `prepareStepGenerationSettingKeys` allow-list in `stream-text-iterator.ts` — so the two callbacks currently disagree about whether a `maxRetries` override works.

Expected: either apply `prepared.maxRetries` and merge `prepared.abortSignal` into the effective signal, or `Omit` the unsupported keys from `PrepareCallResult` so the contract is visible at the type level. (If the latter, `PrepareStepResult` has the same question for `abortSignal`: it is typed via `Partial<GenerationSettings>` but excluded by the allow-list.)

### Reproduction

Vitest test in `packages/workflow`, using the same mocked-`streamTextIterator` harness as the existing `prepareCall` tests in `workflow-agent.test.ts`. It inspects the `generationSettings` handed to the model-call layer; the `temperature` control shows the harness works.

```ts
import type {
  LanguageModelV4,
  LanguageModelV4Prompt,
  LanguageModelV4ToolResultPart,
} from '@ai-sdk/provider';
import { describe, expect, it, vi } from 'vitest';
import type { StreamTextIteratorYieldValue } from './stream-text-iterator.js';
import type { PrepareCallResult } from './workflow-agent.js';

vi.mock('./stream-text-iterator.js', () => ({
  streamTextIterator: vi.fn(),
}));

const { WorkflowAgent } = await import('./workflow-agent.js');

function createMockModel(): LanguageModelV4 {
  return {
    specificationVersion: 'v4' as const,
    provider: 'test',
    modelId: 'test-model',
    doGenerate: vi.fn(),
    doStream: vi.fn(),
    supportedUrls: {},
  };
}

type MockIterator = AsyncGenerator<
  StreamTextIteratorYieldValue,
  LanguageModelV4Prompt,
  LanguageModelV4ToolResultPart[]
>;

describe('prepareCall generation settings', () => {
  async function streamWithPrepareCall(prepared: PrepareCallResult) {
    const { streamTextIterator } = await import('./stream-text-iterator.js');
    vi.mocked(streamTextIterator).mockReturnValue({
      next: vi.fn().mockResolvedValueOnce({ done: true, value: [] }),
    } as unknown as MockIterator);

    const agent = new WorkflowAgent({
      model: createMockModel(),
      prepareCall: () => prepared,
    });
    await agent.stream({
      messages: [{ role: 'user', content: 'test' }],
      writable: new WritableStream({ write: vi.fn(), close: vi.fn() }),
    });

    return vi.mocked(streamTextIterator).mock.calls.at(-1)?.[0]
      .generationSettings;
  }

  it('control: temperature returned from prepareCall reaches the model call', async () => {
    const settings = await streamWithPrepareCall({ temperature: 0.9 });
    expect(settings?.temperature).toBe(0.9); // passes
  });

  it('maxRetries returned from prepareCall reaches the model call', async () => {
    const settings = await streamWithPrepareCall({ maxRetries: 5 });
    expect(settings?.maxRetries).toBe(5); // fails: undefined
  });

  it('abortSignal returned from prepareCall reaches the model call', async () => {
    const controller = new AbortController();
    const settings = await streamWithPrepareCall({
      abortSignal: controller.signal,
    });
    expect(settings?.abortSignal).toBeDefined(); // fails: undefined
  });
});
```

Observed on current main: the control passes, the `maxRetries` and `abortSignal` tests fail with `undefined`. The file passes `tsc --noEmit`, confirming both keys are accepted by `PrepareCallResult`.

Happy to open a PR either way.

### AI SDK Version

- @ai-sdk/workflow: 1.0.56 (same code on current main)

edenbuilds:
## Summary
- `WorkflowAgent.prepareCall` can return `maxRetries` and `abortSignal` (they are on `GenerationSettings` / `PrepareCallResult`), but the field-by-field merge skipped both so they were silently dropped.
- Copy those two fields into `effectiveGenerationSettings` the same way as `seed` / `headers`.
- `prepareStep` also dropped `abortSignal` via `prepareStepGenerationSettingKeys` — now included.
- Per-stream and prepareCall `abortSignal`s are merged with `mergeAbortSignals` instead of one replacing the other.
- Note: `maxRetries` still does not reach a model-retry implementation in `packages/workflow` (step mechanism owns retries); the copy keeps the typed merge honest / consistent with telemetry.

Fixes #18576

## Test plan
- [x] `pnpm test:node -- src/workflow-agent.test.ts src/stream-text-iterator.test.ts` in `packages/workflow`
- [x] prepareCall maxRetries/abortSignal land on `streamTextIterator`
- [x] prepareStep abortSignal applied
- [x] prepareCall + per-stream abortSignal merge

### DIFF
--- a/.changeset/workflow-agent-preparecall-retries.md
+++ b/.changeset/workflow-agent-preparecall-retries.md
@@ -2,4 +2,4 @@
 "@ai-sdk/workflow": patch
 ---
 
-fix(WorkflowAgent): apply `maxRetries` and `abortSignal` returned from `prepareCall`
+Fixed `WorkflowAgent.prepareCall` dropping `maxRetries` and `abortSignal`, and `prepareStep` silently dropping `abortSignal`. Per-stream and prepareCall abort signals are now merged instead of one replacing the other.
--- a/.changeset/workflow-agent-preparecall-retries.md
+++ b/.changeset/workflow-agent-preparecall-retries.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/workflow": patch
+---
+
+fix(WorkflowAgent): apply `maxRetries` and `abortSignal` returned from `prepareCall`
--- a/packages/workflow/src/stream-text-iterator.test.ts
+++ b/packages/workflow/src/stream-text-iterator.test.ts
@@ -135,6 +135,30 @@ describe('streamTextIterator', () => {
         maxOutputTokens: 256,
       });
     });
+
+    it('applies abortSignal from prepareStep', async () => {
+      vi.mocked(doStreamStep).mockResolvedValue(createMockDoStreamStepResult());
+      const stepAbort = new AbortController();
+
+      const iterator = streamTextIterator({
+        prompt: [{ role: 'user', content: [{ type: 'text', text: 'test' }] }],
+        tools: {},
+        model: vi.fn() as any,
+        generationSettings: {
+          temperature: 0.2,
+        },
+        prepareStep: () => ({
+          abortSignal: stepAbort.signal,
+        }),
+      });
+
+      await iterator.next();
+
+      expect(vi.mocked(doStreamStep).mock.calls[0]?.[4]).toMatchObject({
+        temperature: 0.2,
+        abortSignal: stepAbort.signal,
+      });
+    });
   });
 
   describe('telemetry', () => {
--- a/packages/workflow/src/stream-text-iterator.ts
+++ b/packages/workflow/src/stream-text-iterator.ts
@@ -51,6 +51,7 @@ const prepareStepGenerationSettingKeys = [
   'stopSequences',
   'seed',
   'maxRetries',
+  'abortSignal',
   'headers',
   'reasoning',
   'providerOptions',
--- a/packages/workflow/src/workflow-agent.test.ts
+++ b/packages/workflow/src/workflow-agent.test.ts
@@ -163,6 +163,34 @@ describe('WorkflowAgent', () => {
         }),
       );
     });
+
+    it('applies maxRetries and abortSignal from prepareCall', async () => {
+      const abortController = new AbortController();
+      const prepareCall = vi.fn(() => ({
+        maxRetries: 0,
+        abortSignal: abortController.signal,
+      }));
+      const { streamTextIterator } = await import('./stream-text-iterator.js');
+      vi.mocked(streamTextIterator).mockReturnValue({
+        next: vi.fn().mockResolvedValueOnce({ done: true, value: [] }),
+      } as unknown as MockIterator);
+
+      const agent = new WorkflowAgent({
+        model: createMockModel(),
+        prepareCall,
+      });
+
+      await agent.stream({ prompt: 'test' });
+
+      expect(streamTextIterator).toHaveBeenCalledWith(
+        expect.objectContaining({
+          generationSettings: expect.objectContaining({
+            maxRetries: 0,
+            abortSignal: abortController.signal,
+          }),
+        }),
+      );
+    });
   });
 
   describe('tool execution error handling', () => {
--- a/packages/workflow/src/workflow-agent.test.ts
+++ b/packages/workflow/src/workflow-agent.test.ts
@@ -191,6 +191,40 @@ describe('WorkflowAgent', () => {
         }),
       );
     });
+
+    it('merges prepareCall and per-stream abortSignal', async () => {
+      const prepareAbort = new AbortController();
+      const streamAbort = new AbortController();
+      const prepareCall = vi.fn(() => ({
+        abortSignal: prepareAbort.signal,
+      }));
+      const { streamTextIterator } = await import('./stream-text-iterator.js');
+      vi.mocked(streamTextIterator).mockClear();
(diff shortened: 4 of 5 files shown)


hyamero:
Thanks for picking this up. I filed #18576, so I verified the branch: reverting just the four source lines makes the new test fail and restoring them passes all 74, and the merge order you left unchecked does hold for both fields. The `abortSignal` half is correct and complete, traced through `mergeAbortSignals` and the loop's `aborted` checks down to `streamModelCall`, and that was the half with real user impact.

On `maxRetries` I owe you a correction to my own issue. I wrote that the call "proceeds with the constructor/per-stream value or the default (`mergedGenerationSettings.maxRetries ?? 2`)", but that line is the telemetry payload at `workflow-agent.ts:1924`, not the model call path. Read literally, your change is exactly what the issue asked for. The runtime picture is different though: `maxRetries` never reaches a retry implementation from any source. `streamModelCall` at `do-stream-step.ts:132-154` lists its fields explicitly and omits it (`abortSignal` is in that list), `experimental_streamLanguageModelCall` has neither retry logic nor such a parameter, and `prepareRetries` is referenced only by `generate-text.ts` and `stream-text.ts`, never in `packages/workflow`. The constructor value, the per-stream value, and the `prepareStep` value are all equally disconnected.

So this half makes `prepareCall` consistent with the other sources rather than making the setting take effect, and the new test passes because it asserts on the object handed to `streamTextIterator` rather than on retry behavior. That may well be by design, since `doStreamStep` is a workflow step (`'use step'`, line 114) and the doc comment on `GenerationSettings.maxRetries` says retries are handled by the step mechanism. If so, the real question is whether `maxRetries` belongs in `PrepareCallResult` at all, or should be documented as telemetry-only. Either resolution works for me, but the type currently promises something no code path delivers.

One piece of the issue is untouched: `prepareStepGenerationSettingKeys` (`stream-text-iterator.ts:44`) includes `'maxRetries'` but omits `abortSignal`, which `PrepareStepResult` types via `Partial<GenerationSettings>`. Same silent drop, and unlike `maxRetries` it has real consequences, since `abortSignal` does work in this package. Because the PR says `Fixes #18576`, merging would close that out along with it, so it is worth either adding the key here or splitting it into a follow-up.

Minor and pre-existing: when both a per-stream and a `prepareCall` `abortSignal` are present, the latter is discarded rather than merged (`workflow-agent.ts:1792`). That matches how the constructor signal is already treated, but running both through `mergeAbortSignals` would be safer for cancellation sources.

edenbuilds:
@hyamero thanks — this was a careful read and you’re right on both halves.

**`abortSignal` (real impact).** Done in `2a43c697a4`:
1. `prepareStepGenerationSettingKeys` now includes `abortSignal`, so `prepareStep` no longer silently drops it.
2. Per-stream + prepareCall abort sources go through `mergeAbortSignals` instead of `??`, so neither cancels the other. Regression covers the merge.

**`maxRetries`.** Agreed with your correction. The prepareCall copy makes it consistent with the other typed sources / telemetry payload, but nothing in `packages/workflow` feeds `maxRetries` into a retry implementation (`doStreamStep` omits it; retries are the step mechanism per the `GenerationSettings` doc). I’m leaving the field copy so the type isn’t a lie at the merge layer; happy to follow up with either (a) documenting it as telemetry / step-owned only, or (b) removing it from `PrepareCallResult` if maintainers prefer a smaller surface. This PR no longer pretends it “makes retries work.”

Leaving `Fixes #18576` because the silent-drop bugs that issue filed are addressed; the “does maxRetries belong here?” design question is separate if you want it tracked.

gr2m:
closed via #18713

PR #18595, opened 2026-08-07
feat(gmicloud): add GMI Cloud provider

joemckenney:
Adds `@ai-sdk/gmicloud`, a provider for [GMI Cloud](https://www.gmicloud.ai) GPU inference over OpenAI-compatible chat completions, following the cerebras pattern (OpenAICompatibleChatLanguageModel subclass with a custom `errorStructure`).

### Why a custom error structure

GMI's edge wraps backend rejections in a proxy banner: `error.message` is always `Backend request failed with status 400`, and the engine's actual diagnostic sits in `error.details`, a stringified JSON document the default OpenAI-compatible schema drops. Verified live across six model families on their catalog (deepseek, Qwen, zai, MiniMax, Xiaomi, plus aggregated gemini/claude): every 400 shares this envelope with only the inner text differing.

```json
{"error":{"message":"Backend request failed with status 400","type":"backend_error","code":400,
  "details":"{\"error\":{\"message\":\"The request is invalid: Invalid max_tokens value, the valid range of max_tokens is [1, 393216]. ...\"}}"}}
```

With the attached structure, `AI_APICallError.message` carries the engine's reason; anything unexpected (absent/non-JSON/misshapen details) falls back to the outer message, i.e. default behavior.

### Package contents

- `createGmicloud` / `gmicloud` default instance, `GMI_API_KEY` env fallback, baseURL default `https://api.gmi-serving.com/v1`
- chat completions only; embedding/image throw `NoSuchModelError`
- `includeUsage: true` so streams carry token usage
- model ids typed `string` (evolving catalog), e.g. `deepseek-ai/DeepSeek-V4-Flash-0731`
- tests, examples, docs page, changeset (`major` per add-new-provider)

Spec v4 only; backports to the v6/v5 lines can follow after main lands.

### Testing

- 12 tests pass on node and edge (schema/unwrap fixtures are byte-identical to live wire captures); `pnpm type-check:full`, oxlint, oxfmt, konsistent clean
- live against api.gmi-serving.com: happy path 6/6 across five model families (generate + stream, usage accounting correct)
- live error path surfaces the engine diagnostic cross-model, e.g. `Range of max_tokens should be [1, 131072]` on Qwen/Qwen3.8-Max instead of the banner
- end to end through a local AI Gateway with this build packed in (`local-packages` tarball, v4 route), same invalid-tool-schema shape pinned to gmicloud: production returns `Backend request failed with status 400`; the local gateway returns the engine diagnostic (`Invalid schema for function 'calculator'...`), and the gateway's recorded provider attempt carries the same message in both telemetry error fields
- GMI's model-not-found 404 is a plain-text router response and still records status text; out of scope here


packages/gmicloud/src/gmicloud-chat-options.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * GMI Cloud serves an evolving catalog of open-weight models, so model ids
+ * are not enumerated here. Example: `deepseek-ai/DeepSeek-V4-Flash-0731`.
+ */
+export type GmicloudChatModelId = string;
```
shaper: probably worth adding 3 - 5 of the current popular ones so folks can try them with autocomplete, even though we know they'll be changed over time.

example to follow: packages/alibaba/src/alibaba-chat-language-model-options.ts

packages/gmicloud/src/gmicloud-error.ts
```diff
@@ -0,0 +1,60 @@
+import type { ProviderErrorStructure } from '@ai-sdk/openai-compatible';
+import { secureJsonParse } from '@ai-sdk/provider-utils';
+import { z } from 'zod/v4';
+
+/**
```
shaper: would rm this comment block

packages/gmicloud/src/gmicloud-provider.ts
```diff
@@ -0,0 +1,108 @@
...
+  type FetchFunction,
+} from '@ai-sdk/provider-utils';
+import { GmicloudChatLanguageModel } from './gmicloud-chat-language-model';
+import type { GmicloudChatModelId } from './gmicloud-chat-options';
+import { gmicloudErrorStructure } from './gmicloud-error';
+import { VERSION } from './version';
+
+export interface GmicloudProviderSettings {
+  /**
+   * GMI Cloud API key. Defaults to the `GMI_API_KEY` environment variable.
```
shaper: `GMICLOUD_API_KEY` sounds clearer to me, or do they already use this with their provider/SDK?

whatever we decide here also needs to go in two other places:
- top-level `turbo.json`
- `examples/ai-functions/.env.example`

packages/gmicloud/src/gmicloud-chat-options.ts
joemckenney: Done: added 5 popular ids from the live catalog (`GET /v1/models`), all exercised end to end in testing, following the alibaba pattern with `| (string & {})`.

packages/gmicloud/src/gmicloud-error.ts
joemckenney: Removed.

packages/gmicloud/src/gmicloud-provider.ts
joemckenney: Their official package (`@gmicloud/ai-sdk-provider`, GMISWE) reads `GMI_CLOUD_APIKEY`, so I went with that for consistency with the vendor's own SDK rather than the clearer-but-new `GMICLOUD_API_KEY`. Also added it to top-level `turbo.json` and `examples/ai-functions/.env.example`. Happy to switch if you feel strongly about the cleaner name.

shaper (approved):

packages/gmicloud/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@ai-sdk/gmicloud",
+  "version": "0.0.0",
```
gr2m: Setting version to 2 which will release 3.0.0 so that we can backport to v6 & v5 if we so choose.

```suggestion
  "version": "2.0.0",
```

gr2m (approved):

PR #18617, opened 2026-08-09
fix(skills): match use-ai-sdk skill name to its directory

bechor25:
## What

`skills/use-ai-sdk/SKILL.md` declares:

```yaml
name: ai-sdk
```

but its parent directory is `use-ai-sdk`.

## Why it matters

The Agent Skills specification requires the `name` field to **match the parent directory name**:

> The required `name` field: … Must match the parent directory name
> — <https://agentskills.io/specification#name-field>

`skills-ref validate skills/use-ai-sdk` fails on this today, and a spec-compliant loader can fail to resolve the skill by path.

## Why this looks like a slip, not an alias

It is the only mismatch in `skills/`. All eleven siblings already match their directory:

`add-function-examples`, `add-harness-package`, `add-provider-package`, `adr-skill`, `capture-api-response-test-fixture`, `develop-ai-functions-example`, `list-npm-package-content`, `major-version-mode`, `migrate-ai-sdk-v6-to-v7`, `update-harness-dependencies`, `update-provider-models`.

## The change

One line. Renaming the field rather than the directory keeps the path stable and breaks no links — `git grep "name: ai-sdk"` matches only this frontmatter line.

---

Found with [AgentCompass](https://github.com/YoavLax/agent-compass), an offline static analyzer for AI-agent repo readiness. Verified by hand against the spec before opening.

### DIFF
--- a/skills/use-ai-sdk/SKILL.md
+++ b/skills/use-ai-sdk/SKILL.md
@@ -1,5 +1,5 @@
 ---
-name: ai-sdk
+name: use-ai-sdk
 description: 'Answer questions about the AI SDK and help build AI-powered features. Use when developers: (1) Ask about AI SDK functions like generateText, streamText, ToolLoopAgent, embed, or tools, (2) Want to build AI agents, chatbots, RAG systems, or text generation features, (3) Have questions about AI providers (OpenAI, Anthropic, Google, etc.), streaming, tool calling, structured output, or embeddings, (4) Use React hooks like useChat or useCompletion. Triggers on: "AI SDK", "Vercel AI SDK", "generateText", "streamText", "add AI to my app", "build an agent", "tool calling", "structured output", "useChat".'
 ---
 


bechor25:
Flagging one thing so it doesn't look like a broken PR: the red `Vercel` status is not a build failure. It reads `Authorization required to deploy.` — the preview deployment needs someone on the team to authorize it, which is standard for a first-time fork contributor and nothing I can action from here.

The branch is still `MERGEABLE` against `main`.

The change itself is one line: `skills/use-ai-sdk/SKILL.md` declares `name: ai-sdk` while its directory is `use-ai-sdk`, and the Agent Skills spec requires the two to match. Every other skill under `skills/` already does. `git grep "name: ai-sdk"` matches only that frontmatter line, so nothing else refers to the old value.

felixarntz (changes requested):
@bechor25 thank you for the PR! Can you do it the inverse way though please? i.e. rename the folder to `ai-sdk`

This is how the skill is widely known and used so far

PR #18631, opened 2026-08-10
fix(harness-claude-code): exclude subagent messages from parent step tracking

### LINKED ISSUE #18618: harness-claude-code: subagent (Task-tool) messages aren't excluded from step usage/tool-call tracking
### Description

`createEmitStreamEvent` in `packages/harness-claude-code/src/bridge/create-emit-stream-event.ts` handles every incoming `type === 'assistant'` message from the Claude Code CLI's `stream-json` output identically, regardless of whether it originated from the main agent or from a subagent spawned via the Task tool. The Claude Code CLI already tags subagent-originated messages with a non-null `parent_tool_use_id` (confirmed in `@anthropic-ai/claude-code`'s own `cli.js`, e.g. `yield{type:"assistant",message:K.message,parent_tool_use_id:A.parentToolUseID,...}` for subagent progress, vs. `parent_tool_use_id:null` for the main thread), but that field is never read here.

Concretely, in the `type === 'assistant'` branch:

```ts
if (type === 'assistant' && msg.message?.content) {
  const usage = mapUsage(msg.message.usage);
  ...
  for (const block of msg.message.content) {
    if (block.type === 'tool_use' && ...) {
      ...
      state.nativeToolCallNames.set(block.id, block.name);
      state.pendingStepToolUseIds.add(block.id);
      opensStep = true;
      emit({ type: 'tool-call', ... });
    }
  }
  if (opensStep || toolUseIds.length === 0) {
    state.stepOpen = true;
    if (usage) state.pendingStepUsage = usage;   // <-- overwritten by subagent messages too
  }
  return;
}
```

Because a subagent can run many of its own model calls and tool calls while the main agent's outer `Task` tool call is still pending, each of those subagent assistant messages:

In the `type === 'assistant'` (and `type === 'user'` tool-result) handling, skip/branch on `msg.parent_tool_use_id != null` so subagent-originated messages don't feed `pendingStepUsage`, `pendingStepToolUseIds`, or `nativeToolCallNames` for the main turn's step tracking. If subagent activity should still be observable to consumers, it'd be great to see it surfaced as its own tagged event (e.g. via `harnessMetadata`) rather than merged indistinguishably into the main turn's `finish-step`.

### Reproduction

1. Drive a Claude Code CLI session via `HarnessAgent`/`claudeCode` with a prompt that causes the agent to spawn a subagent (Task tool) that itself makes several model calls.
2. Listen to `finish-step`/`finish` stream parts and log `usage`.
3. Observe that the usage reported after the subagent's Task tool call resolves reflects the subagent's own token usage, not the main agent's.
-# 24.3s · 1 step · 226.0k ctx

### AI SDK Version

`@ai-sdk/harness-claude-code` v1.0.65

### Code of Conduct

patrickswedish:
## Background

Claude Code emits messages from Task-tool subagents with a non-null `parent_tool_use_id`.

The harness currently processes those messages as though they belong to the parent agent step. This causes subagent activity to affect the parent step's usage accounting and exposes subagent tool calls/results as parent-step events.

The issue has been reproduced against the current harness behavior with a captured Claude Code stream.

## Summary

This PR keeps the change narrowly scoped:

- ignore assistant messages associated with a subagent when tracking the parent step
- ignore corresponding subagent user/tool-result messages
- preserve normal parent-agent usage and tool-call handling
- add regression coverage using a representative subagent stream fixture
- add a patch changeset for `@ai-sdk/harness-claude-code`

No public API or normal main-agent behavior is changed.

## Verification

The regression covers a stream where subagent messages contain a non-null `parent_tool_use_id`.

It verifies that:

- the parent step retains its own token usage
- subagent tool calls are not emitted as parent-step tool calls
- subagent tool results are not emitted as parent-step tool results

Validation on the final branch:

- TypeScript — passed
- Lint & Format — passed
- Build Packages — passed
- AI test matrix on Node 22, 24, and 26 — passed
- Codemod test matrix — passed
- Example builds — passed
- Changeset verification — passed

## Related Issues

Fixes #18618.

Closes #18619.

### DIFF
--- a/.changeset/tidy-claude-subagents.md
+++ b/.changeset/tidy-claude-subagents.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Exclude Claude Code subagent messages from parent step usage and tool-call tracking.
--- a/.changeset/tidy-claude-subagents.md
+++ b/.changeset/tidy-claude-subagents.md
@@ -2,4 +2,4 @@
 '@ai-sdk/harness-claude-code': patch
 ---
 
-Exclude Claude Code subagent messages from parent step usage and tool-call tracking.
+fix(harness-claude-code): exclude subagent messages from parent step usage and tool-call tracking
--- a/packages/harness-claude-code/src/bridge/__fixtures__/subagent-step-stream.json
+++ b/packages/harness-claude-code/src/bridge/__fixtures__/subagent-step-stream.json
@@ -1,0 +1,129 @@
+[
+  {
+    "type": "assistant",
+    "parent_tool_use_id": null,
+    "message": {
+      "content": [{ "type": "thinking", "thinking": "Delegating." }],
+      "usage": {
+        "input_tokens": 2,
+        "cache_creation_input_tokens": 5605,
+        "cache_read_input_tokens": 2298,
+        "output_tokens": 5
+      }
+    }
+  },
+  {
+    "type": "assistant",
+    "parent_tool_use_id": null,
+    "message": {
+      "content": [
+        {
+          "type": "tool_use",
+          "id": "toolu_main_agent",
+          "name": "Agent",
+          "input": {
+            "description": "Run pwd and echo",
+            "prompt": "Use Bash to run pwd, then echo subagent-done.",
+            "subagent_type": "general-purpose",
+            "run_in_background": false
+          }
+        }
+      ],
+      "usage": {
+        "input_tokens": 2,
+        "cache_creation_input_tokens": 5605,
+        "cache_read_input_tokens": 2298,
+        "output_tokens": 5
+      }
+    }
+  },
+  {
+    "type": "user",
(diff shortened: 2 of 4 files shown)


.changeset/tidy-claude-subagents.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Exclude Claude Code subagent messages from parent step usage and tool-call tracking.
```
felixarntz:
```suggestion
fix(harness-claude-code): exclude subagent messages from parent step usage and tool-call tracking
```

felixarntz (approved):
@patrickswedish Thank you for the PR, LGTM!

felixarntz:
This PR lacks commit signatures, so reimplementing the same change via #18740.

patrickswedish:
Thanks for reviewing and getting the fix landed via #18740, @felixarntz! Glad to contribute.

PR #18673, opened 2026-08-11
fix(ai): handle plus signs in number exponents in fixJson

VihaanAgarwal:
## Background

`fixJson` accepts `-` and `.` inside a number but not `+`, so an explicit plus exponent cuts the number short. `fixJson('2.5e+10')` returns `'2.5'`, and a streamed `{"value":2.5e+10` is repaired to `{"value":2.5}`, silently changing the value. This happens with real payloads because `JSON.stringify` emits plus exponents for large numbers (`JSON.stringify(1e21) === "1e+21"`).

## Summary

Treat `+` like the existing `-` case in the number scanner. Added `e+`/`E+` cases to the e-notation tests; they fail without the fix.

## End-to-End Verification

Ran `fixJson('2.5e+10')` and `fixJson('{"value":2.5e+10')` against main and this branch: main returns `2.5` and `{"value":2.5}`, with the fix both are preserved.

## Checklist

- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features)

### DIFF
--- a/.changeset/plus-exponent-fix-json.md
+++ b/.changeset/plus-exponent-fix-json.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+fix(ai): handle plus signs in number exponents in fixJson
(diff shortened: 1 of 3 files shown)


ArshVermaGit (commented):
Looks good. The fix correctly handles + signs in scientific notation and adds regression coverage for both e+ and E+ cases. The end-to-end verification also confirms the value is preserved. LGTM.

PR #18770, opened 2026-08-12
docs: gate jsonrepair on finishReason in the truncation cookbook

canblmz1:
### What

The "Repair Malformed JSON with jsonrepair" cookbook names truncated responses (hitting token limits) as one of the problems this pattern solves, but the example calls `jsonrepair(result.text)` unconditionally, with no check on `result.finishReason` first.

### Why this matters

`jsonrepair` closes an unterminated string or array as-is. For a response that was genuinely cut off mid-value, that means the repaired result parses successfully and passes schema validation, but its content is shorter/different than what the model actually intended — with nothing to signal that. A closed array like `["ls","pwd"` → `["ls","pwd"]` is indistinguishable from a genuinely complete one just by looking at the JSON text; `finishReason` is the one signal available that actually tells you which case you're in.

This is reachable in a real, non-hypothetical way: `experimental_repairToolCall` follows the same repair-first shape, and a tool call whose argument string is cut off by `max_tokens` can be silently "repaired" into a plausible-looking value and passed straight to execution — for a tool with a real side effect (writing a file, running a command), the repaired value being subtly wrong instead of visibly failing is the worse outcome.

### The fix

Adds a `result.finishReason === 'length'` check before the repair step, using data `generateText()` already returns — no new dependency, just reordering the existing logic so truncation is handled as its own case instead of falling into the generic repair path. Also expands the "Considerations" section to explain the distinction between a syntax problem (what `jsonrepair` fixes) and a truncation problem (what it can't reliably fix, since it has no way to know whether it's completing something real or has just cut off the model's continuation).

### Scope

Docs-only change to one cookbook page — no code/package changes. Matches the "Fix documentation, examples, or typos" category CONTRIBUTING.md calls out as welcome for PRs. I wasn't able to run the full local build (didn't set up the entire monorepo for a one-file docs change) but matched the file's existing formatting conventions by hand; happy to adjust if prettier/lint flags anything.

### DIFF
--- a/content/cookbook/05-node/85-repair-json-with-jsonrepair.mdx
+++ b/content/cookbook/05-node/85-repair-json-with-jsonrepair.mdx
@@ -53,6 +53,19 @@ const result = await generateText({
   prompt: 'Generate a lasagna recipe.',
 });
 
+// Check this before attempting repair. jsonrepair closes an unterminated
+// string or array as-is, so a response cut off mid-value can come back
+// looking like a complete, valid object — the repair "succeeds" while
+// silently dropping content the model never got to generate. finishReason
+// is the one signal that actually distinguishes "malformed" from
+// "truncated"; the JSON text alone can't.
+if (result.finishReason === 'length') {
+  throw new Error(
+    'Response was cut off by maxOutputTokens. Retry with a higher limit ' +
+      'or a smaller schema rather than repairing a truncated value.',
+  );
+}
+
 const repairedText = jsonrepair(result.text);
 const parseResult = safeParseJSON({ text: repairedText });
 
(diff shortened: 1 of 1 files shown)


canblmz1:
Rebased onto current main (no conflicts — the cookbook file was unchanged since this PR opened, confirmed via blob hash).

Re-verified the technical basis against current source rather than assuming it still holds: `FinishReason` still includes `'length'` (`packages/ai/src/types/language-model.ts`), and `GenerateTextResult.finishReason` is typed as that same union, so `result.finishReason === 'length'` remains correct on current main.

`validate:docs` (properties-table validator) passes with this change applied — 525 MDX files, no errors.

PR #18801, opened 2026-08-12
fix(provider/xai): report video moderation blocks as an error status instead of throwing

joshualipman123:
`doStatus` threw a status-less `AISDKError` when xAI reported `respect_moderation: false` (and when a `done` response carried no video URL). Both are terminal but a throw is indistinguishable from a transient network fault, so async pollers retry them to exhaustion.

Both cases now return `{ status: 'error', error }`, matching the shape the upstream `failed` branch twenty lines above already uses. `generateVideo` throws on an error status with the same message text as before; async consumers get a terminal state they can act on.


packages/xai/src/xai-video-model.ts
```diff
@@ -581,19 +581,30 @@ export class XaiVideoModel implements VideoModelV4 {
...
+        };
       }
 
       if (!statusResponse.video?.url) {
-        throw new AISDKError({
-          name: 'XAI_VIDEO_GENERATION_ERROR',
-          message: 'Video generation completed but no video URL was returned.',
-        });
+        return {
+          status: 'error' as const,
```
shaper, line 600: do we have a similar issue in async code paths for other video models?
it seems to make sense to me to fix this based on what we've seen here for xai, but I want to make sure we are consistent elsewhere, and shouldn't be handling this differently. async is quite new.

in follow up can you look across video models and let's start a thread to discuss.

there is for example `packages/ai/src/error/no-video-generated-error.ts`
which seems like it would be appropriate for this case, but perhaps not given it's an async flow.

cc @kkdawkins @gr2m

shaper (approved):

packages/xai/src/xai-video-model.ts
gr2m: yes, we should normalize the other providers to all return `status: 'error'` for terminal generation outcomes and then convert `status: 'error'` into `NoVideoGeneratedError` with the status message as cause

kkdawkins (approved):

PR #18826, opened 2026-08-13
fix(workflow): surface stream errors as terminal failures

### LINKED ISSUE #18824: WorkflowAgent does not surface the value of standard stream error parts as terminal failure
### Description

A `LanguageModelV4` stream can emit the standard `{ type: "error", error: value }` part. In `@ai-sdk/workflow@1.0.64` and current `main`, `doStreamStep()` forwards that part to the writable stream but does not retain or throw its `error` value. The switch in `packages/workflow/src/do-stream-step.ts` has no `error` case, so consumption continues and the returned step result cannot expose the supplied failure.

`toUIMessageChunk()` separately converts the value to `errorText`, which is useful for UI delivery, but server-side `WorkflowAgent.stream()` still does not fail with the original supplied value. Consumers that need to classify a credential-safe, structured error supplied by model middleware therefore receive only a generic error finish (or no typed terminal value).

Expected behavior: when the model stream emits a standard error part, fail the model step / agent run with the supplied `part.error` value, while preserving real-time stream delivery as appropriate. An equivalent typed terminal field on the server-side result would also provide the necessary contract, but a generic `finishReason: "error"` is insufficient because it discards the safe error classification.

### Reproduction

A focused `doStreamStep` test can mock `experimental_streamLanguageModelCall` with a stream containing:

```ts
const terminal = new Error("safe-terminal-marker");

{ type: "error", error: terminal }
```

Then assert that `doStreamStep(...)` rejects with `terminal` (or returns it through an explicit typed terminal field). It currently consumes the part, optionally writes it, and resolves without that value.

The same behavior is visible directly in `packages/workflow/src/do-stream-step.ts` on current `main`: `error` is not handled in the aggregation switch.

### AI SDK Version

- `@ai-sdk/workflow`: 1.0.64
- `ai`: 7.0.64

### Code of Conduct

I agree to follow this project Code of Conduct.

MintedKenny:
## Summary

- forward a standard model stream error part to the writable exactly once
- terminate the durable model step by throwing the exact supplied error value
- cover the behavior with a focused regression test and patch changeset

## Testing

- `pnpm --filter @ai-sdk/workflow test:node` (167 tests passed)
- `pnpm --filter @ai-sdk/workflow test:edge` (167 tests passed)
- `pnpm --filter @ai-sdk/workflow build`
- `pnpm type-check`
- `pnpm check`
- `fnm exec --using=24.14.0 pnpm build:packages` (75 packages built)

The full examples-inclusive root build reaches the unrelated `@example/harness-e2e-next` build, then fails while collecting page data for its Grok Build routes with `ERR_INVALID_ARG_TYPE`.

Fixes #18824

### DIFF
--- a/.changeset/tidy-workflows-fail.md
+++ b/.changeset/tidy-workflows-fail.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/workflow': patch
+---
+
+Fail workflow model steps with the exact standard stream error after forwarding it to the writable stream.
--- a/packages/workflow/src/do-stream-step.test.ts
+++ b/packages/workflow/src/do-stream-step.test.ts
@@ -1,0 +1,59 @@
+import type { LanguageModelV4Prompt } from '@ai-sdk/provider';
+import type { Experimental_LanguageModelStreamPart, ToolSet } from 'ai';
+import { MockLanguageModelV4, convertArrayToReadableStream } from 'ai/test';
+import { describe, expect, it } from 'vitest';
+import { doStreamStep } from './do-stream-step.js';
+
+describe('doStreamStep', () => {
+  it('forwards a stream error once and rejects with its exact value', async () => {
+    const terminalError = new Error('safe-terminal-marker');
+    const forwardedParts: Array<Experimental_LanguageModelStreamPart<ToolSet>> =
+      [];
+    const model = new MockLanguageModelV4({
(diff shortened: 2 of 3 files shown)


gr2m:
Closing this because #18824 was resolved by the merged replacement PR #18827. Thanks for the contribution and investigation.

PR #18834, opened 2026-08-13
fix(langchain): preserve reused tool call IDs across steps

### LINKED ISSUE #18831: @ai-sdk/langchain: reused toolCallId across LangGraph steps loses the later tool lifecycle
### Description

`@ai-sdk/langchain` keeps tool-call emission bookkeeping for the full LangGraph stream. When a later LangGraph step reuses a provider-scoped `toolCallId`, `toUIMessageStream()` suppresses the later `tool-input-start` but still emits its input deltas. `readUIMessageStream()` then creates a second tool part with the earlier call's name and input.

AI SDK core already supports repeated IDs across steps after #17347 / #17586, but the LangChain adapter drops the second lifecycle before the core reducer can distinguish it.

### Reproduction

```ts
import { toUIMessageStream } from '@ai-sdk/langchain';
import { AIMessageChunk } from '@langchain/core/messages';

async function* events() {
  yield [
    'messages',
    [
      new AIMessageChunk({
        id: 'message-1',
        content: '',
        tool_call_chunks: [
          {
            id: 'call_reused',
            name: 'write_column',
            args: '{"column":"first"}',
            index: 0,
          },
        ],
      }),
      { langgraph_step: 1 },
    ],
  ];

  yield [
    'messages',
    [
      new AIMessageChunk({
        id: 'message-2',
        content: '',
        tool_call_chunks: [
          {
            id: 'call_reused',
            name: 'deploy_creatives',
            args: '{"column":"second"}',
            index: 0,
          },
        ],
      }),
      { langgraph_step: 2 },
    ],
  ];
}

let finalMessage;
for await (const message of readUIMessageStream({
  stream: toUIMessageStream(events()),
})) {
  finalMessage = message;
}

console.log(
  finalMessage?.parts.filter(part => part.type === 'dynamic-tool'),
);
```

### Expected

Two tool parts:

1. `write_column` with `{ column: 'first' }`
2. `deploy_creatives` with `{ column: 'second' }`

### Actual

Two tool parts are present, but both have `toolName: 'write_column'` and `{ column: 'first' }`.

### Root cause

`processLangGraphEvent()` uses stream-global `emittedToolCalls`, `emittedToolInputs`, and `emittedToolCallsByKey` state. The second step finds `call_reused` in `emittedToolCalls`, so it does not emit a new `tool-input-start`.

An unconditional reset at every counter change is risky. A complete fix should preserve:

- within-step and `messages` / `values` / `tools` cross-mode deduplication;
- delayed output lookup against a prior step;
- HITL key matching added in #15793;
- historical completed-call suppression from #11417;
- interleaved namespace behavior from #18506;
- the original provider `toolCallId`.

The core fix in #17586 uses current-step lookup with explicit prior-step fallback and may provide the right model for the adapter.

### Versions

- `@ai-sdk/langchain`: 3.0.64
- `ai`: 7.0.64
- `@langchain/core`: 1.2.1
- Node.js: 24

### Code of Conduct

mdnanocom:
## Background

Fixes #18831.

AI SDK core fixed repeated provider-scoped tool call IDs across steps in #17347 and #17586 by scoping UI tool-part lookup to the current step while retaining fallback lookup for delayed outputs.

`@ai-sdk/langchain` still kept `emittedToolCalls`, `emittedToolInputs`, and `emittedToolCallsByKey` for the full LangGraph stream. When a provider reused a `toolCallId` in a later step, the adapter suppressed the later `tool-input-start`. The core fix could not recover because the adapter had already dropped the second lifecycle.

## Changes

- scope LangGraph tool-call deduplication to emitted UI steps
- retain prior-step state for delayed tool outputs and HITL key lookup
- rotate state only when a global UI step boundary is emitted, preserving interleaved namespace behavior
- preserve the original provider tool-call ID
- add an end-to-end `toUIMessageStream` to `readUIMessageStream` regression for a reused ID
- cover delayed outputs, prior-step HITL lookup, mixed stream-mode deduplication, and suppressed namespace boundaries
- add a patch changeset for `@ai-sdk/langchain`

## Validation

- `pnpm test:node` in `packages/langchain` — 247 tests passed
- `pnpm test:edge` in `packages/langchain` — 247 tests passed
- `pnpm type-check` in `packages/langchain` — passed
- targeted Ultracite formatting and lint checks — passed


gr2m:
superseded by https://github.com/vercel/ai/pull/18837

PR #18887, opened 2026-08-13
fix(anthropic): replay dynamic-filtering callers without reordering

### LINKED ISSUE #18785: Anthropic: multi-turn dynamic filtering (web_search_20260209 inside code_execution) rejected with "code_execution tool use ... without a corresponding code_execution_tool_result block"
### Description

On multi-turn conversations that use dynamic filtering — `web_search_20260209` / `web_fetch_20260209`, which Anthropic runs **inside** a `code_execution` sandbox — the history `@ai-sdk/anthropic` replays on a follow-up request is rejected:

```
messages.N: `code_execution` tool use with id `srvtoolu_…` was found
without a corresponding `code_execution_tool_result` block
```

The `code_execution_tool_result` **is present** in the request. This is **not** the dropped-encrypted-result bug (#13040 / #13046) — that fix is in our version and our captured requests carry `encrypted_code_execution_result` intact. The result is present but **not co-located with its call** the way the API requires.

There are two distinct triggering shapes, both reproducible.

#### Shape 1 — a single `code_execution` runs multiple nested `web_search` calls

The assistant message the SDK reconstructs looks like:

```
code_execution(call CE)
server_tool_use web_search(W1)  →  web_search_tool_result(W1)
server_tool_use web_search(W2)  →  web_search_tool_result(W2)
server_tool_use web_search(W3)  →  web_search_tool_result(W3)
code_execution_tool_result(CE)
```

Every `server_tool_use` block has **no `caller`**. Empirically, a `code_execution` with **one** nested `web_search` is accepted; with **two or more** the API reports `CE` unclosed (the error above). This matches Anthropic tolerating a single caller-less nested call positionally but not several.

#### Shape 2 — a client tool is called in the same turn (`pause_turn`)

When the model calls a client tool while a `code_execution` is still running, the API returns `stop_reason: pause_turn` and defers the `code_execution_tool_result` to the continuation step. `toResponseMessages` builds **one assistant message per step**, so the pair is split across the client round:

```
assistant[i]:   … code_execution(call CE) … tool_use(fetch_url)     ← step stops on the client tool
user[i+1]:      tool_result(fetch_url)
assistant[i+2]: … code_execution_tool_result(CE) …                  ← deferred result lands in a NEW message
```

Call `CE` ships in message `i` with no result in that message → same error.

### Root cause (source-level)

**Shape 1 — the `server_tool_use` parser drops `caller`.** In `convert-to-anthropic-messages` / the streaming parser, the `tool_use` branch reads `part.caller` and preserves it (`providerMetadata.anthropic.caller`), but the **`server_tool_use` branch** (`web_search` / `code_execution` / `web_fetch`) builds the `tool-call` with `providerExecuted` / `dynamic` / `providerToolName` and **never reads `part.caller`**. So the parent-linkage Anthropic sends on each in-sandbox `web_search` is discarded, and on replay the searches look top-level.

Verified present in **`@ai-sdk/anthropic@4.0.38`** (latest at time of writing):
- non-streaming `server_tool_use` branch — the `part.name === "web_search" || "code_execution" || "web_fetch"` case omits `caller`;
- streaming `server_tool_use` case (`content_block_start`) — builds `contentBlocks[…]` without `caller`, so the finalizer's `…contentBlock.caller && { providerMetadata… }` has nothing to emit;
- the `tool_use` case in both paths **does** capture `caller`.

**Shape 2 — `toResponseMessages` never re-co-locates a `pause_turn`-deferred server result.** Each step's content becomes its own assistant message; a server-tool result deferred to a later step therefore lands in a different assistant message than its call, split by the client round's `tool` (user) message. Nothing merges them, and the Anthropic API requires a `server_tool_use` and its `*_tool_result` in the same message.

### Versions

- `ai`: 6.0.246
- `@ai-sdk/anthropic`: 3.0.107 (both mechanisms verified still present in **4.0.38**)

### How this differs from existing issues

- #13040 / #13046 — encrypted result *dropped*; fixed, and our result is present.
- #13013 / #13019 — `tool_use` must be the last block; our `tool_use` is already last.
- #8112 — splits a *non-provider* tool into a `tool` message; ours is the *server* `code_execution` result.
- #13533 — provider *never returns* the result; ours is present but misplaced.
- #12504 — `pruneMessages` *orphans* a `caller` for programmatic *client* calls; ours drops `caller` at *parse time* on nested `web_search` `server_tool_use`.

### Empirical fix / workaround

Co-locating every `*_tool_result` immediately after its `server_tool_use` call — move across messages for Shape 2, reorder within the message for Shape 1 — makes the API accept the request (verified against the live API: both captured 400 bodies replay as 400 verbatim and return **200** once co-located). Note: naively adding `caller` back to the nested `web_search` calls alone did **not** fix Shape 1 — the API then rejects the separate `web_search_tool_result` blocks ("`web_search_tool_result` … must have a corresponding `server_tool_use` block before it"), so a correct in-SDK fix likely needs the full caller + result restructuring (or the co-location approach above). We currently work around it downstream by co-locating results in a message transform.

### Reproduction

Triggered reliably by a research-style prompt that makes the model run several web searches inside code execution and also call a client tool in the same turn (dynamic filtering on opus-4.6+/sonnet-4.6+). I can share a captured, content-redacted request body that reproduces the 400 on replay (and returns 200 after the co-location transform) if useful.

gr2m:
## Background

This is a smaller alternative to #18794 for #18785.

Anthropic returns `caller` on `server_tool_use` blocks and on dynamic-filtering `web_search_tool_result` / `web_fetch_tool_result` blocks. The provider currently parses caller metadata for client `tool_use`, but discards it for those server calls and results. A later request therefore cannot replay the response Anthropic actually produced.

## Approach

- Parse the documented caller union once and reuse it across response schemas.
- Preserve caller metadata on streaming and non-streaming server calls and web results.
- Serialize that metadata back to Anthropic on the next request.
- Keep the caller's message boundaries and content-block order unchanged.

The regression test intentionally uses two nested searches and asserts this order remains intact:

```text
code_execution call
web_search call 1
web_search result 1
web_search call 2
web_search result 2
code_execution result
```

## Scope

This PR addresses the caller-metadata loss in completed dynamic-filtering turns (Shape 1). It intentionally does not relocate deferred server results across message boundaries (Shape 2).

Anthropic documents mixed server/client continuations as separate responses accumulated in order: the first assistant response containing the open server call, a user message containing the client `tool_result`, and a later assistant response containing the deferred server result. The server call and result pair by `tool_use_id`, not by position. A regression test locks in that `assistant → user → assistant` sequence while preserving caller metadata on the originating calls.

The reported rejection of this split sequence was not reproduced against the current live API; the split history was accepted when `code_execution_20260120` was configured. This PR therefore treats cross-message relocation as out of scope rather than normalizing a documented provider-authored transcript.

Documentation: https://platform.claude.com/docs/en/agents-and-tools/tool-use/server-tools

## Why avoid co-location here?

The provider is an adapter, so its safest default is a lossless round trip. Reordering a provider-authored transcript is a separate compatibility policy with a wider blast radius: it can alter the meaning of interleaved blocks, hide malformed persisted history, and make future Anthropic response shapes harder to reason about.

The issue's call-only experiment did not restore callers on the corresponding web-result blocks. Anthropic's API models caller on both sides; this PR preserves both. The reported cross-message shape was also not reproduced against the current live API, so this alternative does not add speculative cross-message relocation.

If a caller-complete replay in original order is still rejected in live validation, that is useful evidence for a narrowly scoped ordering workaround. It does not need to be coupled to fixing the known metadata loss.

## Existing persisted messages

A one-time migration is possible, but it cannot be universally lossless: older AI SDK versions discarded caller metadata before persistence. The safe policy is to infer only unambiguous same-message dynamic-filtering spans, never move blocks, and report everything else for review.

Below is a storage-agnostic transform for persisted `ModelMessage[]`. A real migration should run it transactionally, record `migrated` / `skipped` counts, and adapt the read/write layer to the application's database.

<details>
<summary>Conservative migration transform</summary>

```ts
type StoredPart = {
  type: string;
  toolCallId?: string;
  toolName?: string;
  providerExecuted?: boolean;
  providerOptions?: Record<string, any>;
  [key: string]: any;
};

type StoredMessage = {
  role: string;
  content: string | StoredPart[];
  [key: string]: any;
};

type Caller =
  | { type: 'direct' }
  | { type: 'code_execution_20260120'; toolId: string };

function addCaller(part: StoredPart, caller: Caller): StoredPart | undefined {
  const existing = part.providerOptions?.anthropic?.caller;

  if (existing != null) {
    return JSON.stringify(existing) === JSON.stringify(caller)
      ? part
      : undefined;
  }

  return {
    ...part,
    providerOptions: {
      ...part.providerOptions,
      anthropic: {
        ...part.providerOptions?.anthropic,
        caller,
      },
    },
  };
}

export function migrateAnthropicDynamicFilteringCallers(
  messages: StoredMessage[],
) {
  let migrated = 0;
  let skipped = 0;

  const output = messages.map((message, messageIndex) => {
    if (message.role !== 'assistant' || !Array.isArray(message.content)) {
      return message;
    }

    const content = [...message.content];
    const parentRanges = content.flatMap((part, start) => {
      if (
        part.type !== 'tool-call' ||
        part.toolName !== 'code_execution' ||
        !part.providerExecuted ||
        part.toolCallId == null
      ) {
        return [];
      }

      const ends = content.flatMap((candidate, end) =>
        end > start &&
        candidate.type === 'tool-result' &&
        candidate.toolName === 'code_execution' &&
        candidate.toolCallId === part.toolCallId
          ? [end]
          : [],
      );

      return ends.length === 1
        ? [{ id: part.toolCallId, start, end: ends[0] }]
        : [];
    });

    for (const [callIndex, call] of content.entries()) {
      if (
        call.type !== 'tool-call' ||
        !call.providerExecuted ||
        (call.toolName !== 'web_search' && call.toolName !== 'web_fetch') ||
        call.toolCallId == null
      ) {
        continue;
      }

      const parents = parentRanges.filter(
        range => range.start < callIndex && callIndex < range.end,
      );
      const resultIndexes = content.flatMap((candidate, index) =>
        candidate.type === 'tool-result' &&
        candidate.toolName === call.toolName &&
        candidate.toolCallId === call.toolCallId &&
        index > callIndex
          ? [index]
          : [],
      );

      if (parents.length !== 1 || resultIndexes.length !== 1) {
        skipped++;
        continue;
      }

      const parent = parents[0];
      const resultIndex = resultIndexes[0];
      if (resultIndex >= parent.end) {
        skipped++;
        continue;
      }

      const nestedCaller: Caller = {
        type: 'code_execution_20260120',
        toolId: parent.id,
      };
      const nextParent = addCaller(content[parent.start], { type: 'direct' });
      const nextCall = addCaller(call, nestedCaller);
      const nextResult = addCaller(content[resultIndex], nestedCaller);

      if (nextParent == null || nextCall == null || nextResult == null) {
        skipped++;
        continue;
      }

      content[parent.start] = nextParent;
      content[callIndex] = nextCall;
      content[resultIndex] = nextResult;
      migrated++;
    }

    return { ...message, content };
  });

  return { messages: output, migrated, skipped };
}
```

</details>

This intentionally does not guess across message boundaries. Persisted `UIMessage[]` use an application-defined parts shape and should be adapted before applying the same matching rules.

## Testing

- Anthropic Node suite: 503 tests passed.
- Anthropic Edge suite: 503 tests passed.
- Anthropic package build and type-check passed.
- Full-workspace type-check passed.
- Repository lint and formatting passed.

The tests use existing captured Anthropic web-fetch fixtures, an outbound multi-search replay regression, and a deferred-result continuation regression. Live API validation was not run in this environment.

## Related

- Anthropic API caller types: https://platform.claude.com/docs/en/api/typescript/messages/create

Fixes https://github.com/vercel/ai/issues/18785
Closes https://github.com/vercel/ai/pull/18794
Closes https://github.com/vercel/ai/pull/18790

### DIFF
--- a/.changeset/calm-callers-replay.md
+++ b/.changeset/calm-callers-replay.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/anthropic': patch
+---
+
+Preserve Anthropic server-tool caller metadata in multi-turn conversations.
--- a/packages/anthropic/src/__snapshots__/anthropic-language-model.test.ts.snap
+++ b/packages/anthropic/src/__snapshots__/anthropic-language-model.test.ts.snap
@@ -785,6 +785,14 @@ exports[`AnthropicLanguageModel > doGenerate > programmatic tool calling > with 
   {
     "input": "{"type":"programmatic-tool-call","code":"\\nimport asyncio\\n\\nasync def main():\\n    player1_wins = 0\\n    player2_wins = 0\\n    round_num = 0\\n    \\n    print(\\"=\\" * 60)\\n    print(\\"DICE GAME: First to 3 Round Wins\\")\\n    print(\\"=\\" * 60)\\n    print()\\n    \\n    while player1_wins < 3 and player2_wins < 3:\\n        round_num += 1\\n        print(f\\"Round {round_num}:\\")\\n        \\n        # Both players roll simultaneously\\n        results = await asyncio.gather(\\n            rollDie({\\"player\\": \\"player1\\"}),\\n            rollDie({\\"player\\": \\"player2\\"})\\n        )\\n        \\n        roll1 = int(results[0])\\n        roll2 = int(results[1])\\n        \\n        print(f\\"  Player 1 rolls: {roll1}\\")\\n        print(f\\"  Player 2 rolls: {roll2}\\")\\n        \\n        # Determine winner\\n        if roll1 > roll2:\\n            player1_wins += 1\\n            print(f\\"  → Player 1 wins this round!\\")\\n        elif roll2 > roll1:\\n            player2_wins += 1\\n            print(f\\"  → Player 2 wins this round!\\")\\n        else:\\n            print(f\\"  → Draw! No points awarded.\\")\\n        \\n        print(f\\"  Score: Player 1: {player1_wins} | Player 2: {player2_wins}\\")\\n        print()\\n    \\n    print(\\"=\\" * 60)\\n    if player1_wins == 3:\\n        print(\\"🏆 PLAYER 1 WINS THE GAME!\\")\\n    else:\\n        print(\\"🏆 PLAYER 2 WINS THE GAME!\\")\\n    print(f\\"Final Score: Player 1: {player1_wins} | Player 2: {player2_wins}\\")\\n    print(f\\"Total Rounds: {round_num}\\")\\n    print(\\"=\\" * 60)\\n\\nasyncio.run(main())\\n"}",
     "providerExecuted": true,
+    "providerMetadata": {
+      "anthropic": {
+        "caller": {
+          "toolId": undefined,
+          "type": "direct",
+        },
+      },
+    },
     "toolCallId": "srvtoolu_01CberhXc9TgYXrCZU8bQoks",
     "toolName": "code_execution",
     "type": "tool-call",
@@ -916,6 +924,14 @@ exports[`AnthropicLanguageModel > doGenerate > tool search tool > bm25 variant >
   {
     "input": "{"query":"weather current conditions San Francisco"}",
     "providerExecuted": true,
+    "providerMetadata": {
+      "anthropic": {
+        "caller": {
+          "toolId": undefined,
+          "type": "direct",
+        },
+      },
+    },
     "toolCallId": "srvtoolu_011Jcb838xqvZyWwDt7bWqAz",
     "toolName": "tool_search",
     "type": "tool-call",
@@ -957,6 +973,14 @@ exports[`AnthropicLanguageModel > doGenerate > tool search tool > regex variant 
   {
     "input": "{"pattern":"weather|SF|San Francisco|forecast|temperature|climate","limit":10}",
     "providerExecuted": true,
+    "providerMetadata": {
+      "anthropic": {
+        "caller": {
+          "toolId": undefined,
+          "type": "direct",
+        },
+      },
+    },
     "toolCallId": "srvtoolu_01SACvPAnp6ucMJsstB5qb3f",
     "toolName": "tool_search",
     "type": "tool-call",
@@ -998,6 +1022,14 @@ exports[`AnthropicLanguageModel > doGenerate > web fetch tool > 20260209 text re
   {
     "input": "{"type":"programmatic-tool-call","code":"\\nimport json\\nresult = await web_fetch({\\"url\\": \\"https://example.com\\"})\\nparsed = json.loads(result) if isinstance(result, str) else result\\nif isinstance(parsed, list):\\n    print(parsed[0][\\"content\\"][:500])\\nelse:\\n    print(parsed)\\n"}",
     "providerExecuted": true,
+    "providerMetadata": {
+      "anthropic": {
+        "caller": {
+          "toolId": undefined,
+          "type": "direct",
+        },
+      },
+    },
     "toolCallId": "srvtoolu_015CSHH7X69AhdK9gNzotEeh",
     "toolName": "code_execution",
     "type": "tool-call",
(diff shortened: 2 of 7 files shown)


gr2m:
I don’t think the deferred-result split should be normalized: [Anthropic’s current docs](https://platform.claude.com/docs/en/agents-and-tools/tool-use/server-tools#mixing-server-tools-and-client-tools-in-one-turn) ("pair up by `tool_use_id`, not by position") explicitly allow results in a later response and require preserving the assistant → user tool-result → assistant result sequence. The live reproduction also accepted that shape. I added [a regression test covering this](https://github.com/vercel/ai/pull/18887/commits/259a47bd87872b1d78804db87af86c4bf23ae3c0) and clarified the PR’s scope.

PR #18901, opened 2026-08-14
fix(harness): claim the bridge event stream on start/resume instead of on connect

gdaybrice:
## Background

Every bridge turn event is delivered to a single socket, `activeSocket`. The bridge set `activeSocket` the moment any authorized WebSocket connected, so connecting was enough to take the event stream away from a turn already streaming to another client.

The resulting failure: a second client connects only to send `abort`. Connecting hands it the stream, but live delivery is disabled on a freshly connected socket until it sends `resume`, so the remainder of the running turn goes to a socket nobody is reading. The first client sees nothing after the abort and its slice is stranded — the turn keeps running and its events accumulate in the replay log with no one listening.

Connecting is not a request for the stream. Asking for a turn (`start`) or for a catch-up (`resume`) is.

This lives in the shared bridge runtime rather than in any one adapter, so the fix applies to every harness adapter built on it — they all share the same ownership rule, and none implements its own transport.

## Summary

- `start` and `resume` claim the event stream; connecting no longer does.
- `sendControl` takes the target socket explicitly, so `bridge-hello`, protocol parse errors, and `bridge-stop` reply to the socket that sent the frame rather than to whoever holds the stream. Without this, a client that only sends `stop` — which never claims the stream — would never receive its `bridge-stop`, and the host's `doStop` would time out after 5s.
- Remove `activeSocketReadyForLiveEvents`. It existed to withhold live events from a socket that had claimed the stream on connect but had not yet been caught up. With ownership claimed only on `start`/`resume` it is set in lockstep with `activeSocket` at every mutation, so `emit` now guards on the owner alone.
- Add WebSocket-level regression coverage for the ownership rules, and state the rule once at the `activeSocket` declaration rather than at each site that observes it.

## End-to-End Verification

Verified against a local sandbox by aborting a turn mid-stream. The abort arrives on its own connection to the bridge, separate from the one streaming the turn. Because connecting no longer claims the event stream, that second connection left the running turn's stream where it was: the in-flight turn finalized cleanly and kept the partial output it had already produced, rather than being stranded with its remaining events going to a socket nobody was reading.

## Checklist

Documentation is left unchecked deliberately: the bridge transport is internal and this changes no public API or documented behavior.


felixarntz (approved):
@gdaybrice Thank you for the PR, LGTM!

PR #18902, opened 2026-08-14
fix(harness): keep the runtime error message on failed provider-executed tool results

gdaybrice:
## Background

When a provider-executed tool fails, the harness runtime already sends the real reason across the wire — a failed shell command's stderr, an MCP server's timeout — as a `tool-result` event with `isError: true`.

`translateStreamPart` dropped that flag, projecting the event into a plain AI SDK `tool-result` part whose output happened to be the error text. Consumers saw a tool that _succeeded_ and returned an error-shaped payload — the UI message stream emitted `tool-output-available` — so nothing downstream could distinguish a failed provider-executed tool from a successful one without inspecting its output.

## Summary

- Translate a `tool-result` carrying `isError` into a `tool-error` part, so the failure is typed as one and reaches consumers as `tool-output-error`. Marking it `providerExecuted: true` is load-bearing: `toUIMessageChunk` redacts non-provider-executed errors through `onError`, which would replace the runtime's own message with a generic string. `providerMetadata` and `dynamic` pass through unchanged, keeping the error on the same UI part as its tool call.
- Resolve `providerExecuted` from the originating `tool-call` rather than hardcoding it. Host tools travel the same `tool-result` events — their calls are emitted with `providerExecuted: false` and their failures are submitted back to the runtime and echoed with `isError` — so hardcoding the flag would have marked host failures provider-executed and bypassed the `onError` redaction consumers rely on. Host tool failures keep their existing `tool-result` projection.
- Append `tool-error` parts to the current step content, so a failed provider-executed tool stays visible in `step.content` the way its successful counterpart does. This also covers the host-tool `tool-error` parts the agent loop already emitted and that were being dropped.
- Collapse the five identical arms of `appendToCurrentStepContent` into one fall-through group, matching how `stream-text.ts` accumulates the same set of part types.

## End-to-End Verification

Ran a harness agent against a local sandbox and forced a provider-executed tool to fail. The client now receives that failure typed as a tool error carrying the runtime's own message. Previously the same failure arrived as a completed tool result whose output happened to hold the error text, with nothing marking the tool as having failed; and typing it as an error without `providerExecuted` set replaces the message with a generic placeholder, which is what the flag prevents.

The run exercised this change together with other harness changes in the same batch rather than in isolation.

## Checklist

## Future Work

- A failed host tool still surfaces as a `tool-result` whose output is an error payload, rather than as a `tool-error`. This PR deliberately leaves that path untouched; aligning it with core would be a separate, consumer-visible change.
- `appendToCurrentStepContent` keeps preliminary tool results in step content, where `stream-text.ts` excludes them. Pre-existing drift between the two accumulators, not addressed here.

### DIFF
--- a/.changeset/fluffy-hounds-repeat.md
+++ b/.changeset/fluffy-hounds-repeat.md
@@ -2,4 +2,4 @@
 '@ai-sdk/harness': patch
 ---
 
-Keep the runtime's error message on failed provider-executed tool results.
+fix(harness): keep the runtime's error message on failed provider-executed tool results
--- a/.changeset/fluffy-hounds-repeat.md
+++ b/.changeset/fluffy-hounds-repeat.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness': patch
+---
+
+Keep the runtime's error message on failed provider-executed tool results.
--- a/packages/harness/src/agent/internal/run-prompt.ts
+++ b/packages/harness/src/agent/internal/run-prompt.ts
@@ -211,6 +211,15 @@ export function runPrompt<
       string,
       Extract<HarnessV1StreamPart, { type: 'tool-call' }>
     >();
+    /*
+     * Host tool results are submitted to the runtime and echoed back as
+     * `tool-result` events, so the event alone cannot say who ran the tool.
+     * The originating `tool-call` can — it is recorded below as it streams by.
+     */
+    const translateOptions = {
+      isProviderExecuted: (toolCallId: string): boolean =>
+        rawToolCallsByToolCallId.get(toolCallId)?.providerExecuted ?? true,
+    };
     const pendingApprovalsByApprovalId = new Map(
       pendingToolApprovals.map(approval => [approval.approvalId, approval]),
     );
@@ -680,7 +689,10 @@ export function runPrompt<
         }
 
         // Forward to consumer as soon as possible.
-        for (const part of translateStreamPart<TOOLS>(displayValue)) {
+        for (const part of translateStreamPart<TOOLS>(
+          displayValue,
+          translateOptions,
+        )) {
           result.enqueue(part);
         }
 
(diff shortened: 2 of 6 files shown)


packages/harness/src/agent/internal/run-prompt.ts
```diff
@@ -211,6 +211,15 @@ export function runPrompt<
...
       Extract<HarnessV1StreamPart, { type: 'tool-call' }>
     >();
+    /*
+     * Host tool results are submitted to the runtime and echoed back as
+     * `tool-result` events, so the event alone cannot say who ran the tool.
+     * The originating `tool-call` can — it is recorded below as it streams by.
+     */
+    const translateOptions = {
+      isProviderExecuted: (toolCallId: string): boolean =>
+        rawToolCallsByToolCallId.get(toolCallId)?.providerExecuted ?? true,
```
felixarntz: See the agent review in https://github.com/vercel/ai/pull/18902#issuecomment-5297911897, the `true` here isn't a reasonable fallback. Certain tools, including client-side tools, may omit the `providerExecuted` property altogether.

felixarntz (changes requested):
@gdaybrice Thank you for the PR!

This looks reasonable, but it needs some iteration to account for the variety of tool kinds AI SDK supports.

packages/harness/src/agent/internal/run-prompt.ts
gdaybrice: @felixarntz thanks, only an explicit `true` counts now, added regression tests too.

.changeset/fluffy-hounds-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness': patch
+---
+
+Keep the runtime's error message on failed provider-executed tool results.
```
felixarntz:
```suggestion
fix(harness): keep the runtime's error message on failed provider-executed tool results
```

felixarntz (approved):
@gdaybrice Thank you, LGTM!

PR #18903, opened 2026-08-14
fix(harness-claude-code): keep non-text and scalar MCP tool results intact

gdaybrice:
## Background

[#18647](https://github.com/vercel/ai/pull/18647) added `parseMcpToolResult`, so an external MCP tool result that is a JSON object now reaches consumers as structured data instead of an opaque string. Two result shapes still lose information on the way through:

- **Non-text content blocks.** A tool result whose content array contains an image (or any non-text block) is flattened by `stringifyContent`, which turns that block into `JSON.stringify(entry)` — i.e. the base64 payload as text. Parsing that string back cannot recover the block, so consumers that would render a preview render base64 instead.
- **Scalar MCP results.** `JSON.parse` succeeds on scalars, so an MCP tool whose result is `42` or `true` was handed to consumers as a number or a boolean rather than the string the tool actually sent.

Since #20040, a result that can be paired with a single `tool_result` block is served by the Claude Agent SDK's structured `tool_use_result`. Both problems above remain on the content-derived fallback that #20040 keeps, which is the path parallel tool calls take: when a message carries more than one `tool_result` block, `tool_use_result` cannot be paired with either and every result is derived from content.

## Summary

- Return the raw content array untouched when it contains a non-text block, before the MCP branch runs. The order matters: the MCP branch would otherwise parse the already-flattened string.
- Only substitute the parsed value in `parseMcpToolResult` when it is a non-null object, so scalar results stay strings.
- Move the result branching into a `resolveToolResult` helper, so the ordering constraint between the branches is explicit rather than implied by a ternary chain. The pre-existing `bash` special case moves with it, unchanged.
- Share one `isTextEntry` predicate between `stringifyContent` and the new non-text check, so the two cannot drift apart.
- Add regression coverage for both cases, including the parallel-tool-call path where the structured output cannot be paired.

No public API changes: this only affects the value carried in the `result` field of `tool-result` stream parts. The raw content array is already a legal `harnessV1ToolResultValueSchema` value, so it crosses the wire boundary unchanged.

## End-to-End Verification

Verified against a harness running in a local sandbox. Reading an image file from disk produces a tool result whose content array carries a non-text block: before the change that block reached the consumer as base64 text, and after it the content array arrives untouched and renders as an image preview.

That run predates #20040, which now serves a single pairable result from the SDK's structured output, so the sandbox run no longer reproduces the bug for a lone tool result. The path this PR fixes is the content-derived fallback that remains for parallel tool calls; the scalar case was never reachable through a built-in tool. Both are covered by the added regression tests, which fail without the guards.

## Checklist

## Future Work

`@ai-sdk/harness-pi` has the same two problems, independently implemented. `unwrapPiToolResult` in `packages/harness-pi/src/pi-translate.ts` filters a `(TextContent | ImageContent)[]` envelope down to its text parts, so an image block is dropped rather than passed through, and its own `parseMcpToolResult` has the same scalar coercion this PR fixes.

The other adapters (`harness-codex`, `harness-acp` and the ACP-based harnesses, `harness-deepagents`, `harness-opencode`) already keep tool-result values structured end to end and are unaffected — Claude Code and Pi are the two that flatten content to a string first. Happy to send the Pi fix separately if it's wanted.

## Related Issues

Follow-up to #18647.

### DIFF
--- a/.changeset/loud-moons-brake.md
+++ b/.changeset/loud-moons-brake.md
@@ -2,4 +2,4 @@
 '@ai-sdk/harness-claude-code': patch
 ---
 
-Keep non-text tool result content and scalar MCP tool results intact instead of flattening them to a stringified or re-parsed value.
+fix(harness-claude-code): keep non-text tool result content and scalar MCP tool results intact instead of flattening them to a stringified or re-parsed value
--- a/.changeset/loud-moons-brake.md
+++ b/.changeset/loud-moons-brake.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Keep non-text tool result content and scalar MCP tool results intact instead of flattening them to a stringified or re-parsed value.
--- a/packages/harness-claude-code/src/bridge/create-emit-stream-event.test.ts
+++ b/packages/harness-claude-code/src/bridge/create-emit-stream-event.test.ts
@@ -7,6 +7,11 @@ import {
 } from './create-emit-stream-event';
 
 describe('createEmitStreamEvent', () => {
+  const mcpImageBlock = {
+    type: 'image',
+    source: { type: 'base64', media_type: 'image/png', data: 'iVBORw0KGgo' },
+  };
+
   it.each([
     {
       name: 'TaskCreate object output',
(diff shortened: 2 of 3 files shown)


gdaybrice:
@felixarntz thanks for merging the other PRs, I can remove most the patches! thoughts on this one? do you need me to change anything?

.changeset/loud-moons-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+Keep non-text tool result content and scalar MCP tool results intact instead of flattening them to a stringified or re-parsed value.
```
felixarntz:
```suggestion
fix(harness-claude-code): keep non-text tool result content and scalar MCP tool results intact instead of flattening them to a stringified or re-parsed value
```

felixarntz (approved):
@gdaybrice Thank you, and apologies for the delay. LGTM!

PR #18904, opened 2026-08-14
feat(harness): stream tool input to the client as the model writes it

gdaybrice:
## Background

While the model writes a tool call's input JSON, nothing reaches the client. Text and reasoning stream token by token; tool input was the only block type that did not, so a UI has no tool part to render until the whole input is complete — for a large tool call that is 20-30 seconds of an apparently idle thread. The Claude Code bridge already parsed the CLI's `input_json_delta` events and discarded them.

## Summary

Adds three stream parts — `tool-input-start`, `tool-input-delta`, `tool-input-end` — and carries them end to end. The settled `tool-call` still follows under the same tool call id with the parsed input, so consumers that ignore the new parts behave exactly as before.

- `@ai-sdk/harness`: add the three parts to `HarnessV1StreamPart` and its Zod encoding, to the bridge outbound union, and to the agent's `translateStreamPart`, work-dir stripping, and settled-input replay guard.
- `@ai-sdk/harness-claude-code`: emit the parts from the bridge on a `tool_use` block's `content_block_start` / `input_json_delta` / `content_block_stop`, and subscribe to them in the session's forwarded-event list. Tools hosted on the bridge's own MCP server are skipped — they never produce a `tool-call` on this wire, so a streamed input for one would never settle.
- `@ai-sdk/workflow-harness`: track in-flight tool-input streams in the persisted stream context and replay the start as a prelude when a delta arrives in a later slice, the way `activeTextParts` / `activeReasoningParts` already do. A `tool-input-delta` whose start the client never saw is a stream error.
- `tool-input-start` carries the same `dynamic` flag as the matching `tool-call`, so the streamed part and the settled call resolve to one UI part rather than two.
- Adds tests in all three packages, a runnable example, and a "Streaming Tool Input" section to the harness tools documentation.

A new wire part has to land in the adapter emit, both schema unions, the agent's translate step, and the session's forwarded-event list. The last one is the quiet one: the channel buffers a validated frame that has no subscriber rather than raising, so a missing entry drops the part with no error anywhere.

## End-to-End Verification

Verified against a local sandbox by invoking a tool whose input was large enough to take several seconds to write. The tool part appeared on the client as soon as the call opened and its input filled in progressively while the model was still writing it, rather than appearing only once the complete input had landed. The settled tool call then arrived under the same tool call id and replaced the partial input with the parsed one.

The same behavior was confirmed across a durable execution slice boundary: an input still streaming when a slice ended continued to render in the next slice, instead of failing on a delta whose start the client had never seen.

## Checklist

## Future Work

- Coalesce `input_json_delta` fragments at the bridge before emitting. Each fragment is currently its own WebSocket frame and its own entry in both the turn's in-memory replay log and the sandbox's `event-log.ndjson`; for a large tool call that is several hundred frames and roughly triples the replay-log bytes on tool-heavy turns. Deltas are documented as arbitrary fragments, so coalescing is semantically free.
- The forwarded-event list in each adapter's `createSession` is hand-maintained and has drifted from the bridge outbound union — `compaction` is emitted by the Claude Code bridge but subscribed nowhere, so it is dropped silently today. Deriving the list from `harnessV1BridgeOutboundMessageSchema` with explicit per-adapter exclusions would remove this class of bug across the five adapters that each keep their own copy.
- An approval-gated external MCP tool's `tool-call` is emitted without `dynamic` while the `tool-result` for the same call carries `dynamic: true`. These should agree.

### DIFF
--- a/.changeset/quiet-donkeys-shave.md
+++ b/.changeset/quiet-donkeys-shave.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/harness-claude-code': patch
+'@ai-sdk/workflow-harness': patch
+'@ai-sdk/harness': patch
+---
+
+feat(harness): stream tool input as the model writes it
--- a/.changeset/quiet-donkeys-shave.md
+++ b/.changeset/quiet-donkeys-shave.md
@@ -5,3 +5,12 @@
 ---
 
 feat(harness): stream tool input as the model writes it
+
+Adds `tool-input-start`, `tool-input-delta` and `tool-input-end` to
+`HarnessV1StreamPart` and the bridge outbound union, and emits them from the
+Claude Code adapter while a tool call's input JSON is being written. The
+settled `tool-call` still follows under the same tool call id.
+
+These are new members of two public discriminated unions, so a consumer that
+switches exhaustively over `HarnessV1StreamPart` (or validates bridge frames
+against its own allowlist) will see variants it did not before.
--- a/content/docs/03-ai-sdk-harnesses/03-tools.mdx
+++ b/content/docs/03-ai-sdk-harnesses/03-tools.mdx
@@ -261,6 +261,39 @@ support it will error if an unsupported tool approval mode is specified.
 Host-executed tool approvals are handled by `HarnessAgent`, so they work across
 adapters.
 
+## Streaming Tool Input
+
+Adapters that can observe a tool call's input as the model writes it stream that
+input ahead of the settled call, so a UI can render the tool the moment it
+starts rather than after its input finishes:
+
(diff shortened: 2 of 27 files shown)


gdaybrice:
@ai-sdk-factory addressed the feedback, except for coalescing fragments as it changes streaming semantics.

felixarntz:
@gdaybrice Apologies for not getting to this PR earlier. Partial tool input streaming was implemented last week and is now used in both Claude Code and Pi. I reimplemented the two additional fixes from this PR in #20050 and #20056, giving you credit for both.

Therefore, we can now close this PR in favor of those other ones. Thank you!

PR #18938, opened 2026-08-14
fix(harness-opencode): don't emit a pre-turn assistant message as the turn's reply

### LINKED ISSUE #18933: harness-opencode: cold resume can emit the PREVIOUS turn's assistant message as the reply (uncorrelated emitContextFallback)
### Description

On a between-turn resume (`doStart({ resumeFrom })`), if the OpenCode SSE event stream ends without settling the current turn, `emitContextFallback` emits the session's **newest assistant message** — with no correlation to the prompt just sent — plus a synthetic **successful** `finish-step`.

When the previous turn's assistant message is still the newest (i.e. the new prompt has not produced one yet), the user receives the **previous turn's answer, verbatim**, as the reply to their new question.

We hit this three times in production. Example: a user asked *"can you write a memo with these top cases you found?"* and received the full research answer from the prior turn, byte-for-byte, with none of its tool calls.

This is silent and looks successful — no error surfaces to the caller, and the turn is recorded as a normal completion.

### Root cause

`packages/harness-opencode/src/bridge/index.ts`

1. `runPrompt` subscribes to the event stream, then sends the prompt. `turnSettled` is resolved in the consumer's `.finally()`, so **stream closure counts as settlement**:

```ts
}).finally(() => {
  eventsReady.resolve(undefined);
  turnSettled.resolve();   // <- stream ended != this turn finished
});
```

2. With `sawFinishStep === false` (and `sawContent === false`), the fallback runs with `emitContent: true`:

```ts
if (!sawFinishStep) {
  const emittedFallback = await emitContextFallback({
    client, sessionId, state, emit,
    emitContent: !sawContent,
  }).catch(() => false);
```

3. `emitContextFallback` → `latestAssistantSnapshot` fetches `session.messages({ sessionID, limit: 20 })` and returns the newest assistant message. There is **no baseline message id, no `parentID` check against the new user message, and no timestamp fence** — so on a resumed session it returns the *prior turn's* assistant. Its `contentParts` are emitted as this turn's output, followed by `finish-step` with `finishReason: 'stop'`.

Because the fallback emits only text/reasoning parts, the replayed turn has text but no tool calls — a clean fingerprint. The emitted `finish-step` carries `harnessMetadata.opencode.fallback: true`.

### Why it shows up on cold resumes

Our sandboxes are persistent: the VM idle-stops after ~10 minutes, the snapshot survives for days. A later turn resumes with the parked `resumeFrom` cursor, the sandbox restores, and OpenCode restarts on the preserved on-disk session storage. Any event-stream interruption in that window (OpenCode closes `/event` on `server.instance.disposed`; ordinary transport closure does the same) reaches the `.finally()` above before the new turn settles — and the newest assistant on disk is necessarily the previous turn's.

The flow itself is the documented one: `HarnessV1SessionStartOptions.resumeFrom` says the adapter "should resume the existing session before accepting a new prompt", and `@ai-sdk/workflow-harness`'s `destroyOnFinish: false` explicitly parks the session so the next user turn reattaches to the same conversation. We are not misusing the API — a correct resume can still produce a wrong answer.

### Suggested fix

Correlate the fallback with the current turn, and never synthesize success from pre-turn state:

1. Capture the newest assistant message id **before** sending the prompt (baseline).
2. Treat stream closure as an abnormal condition, not settlement — reconnect/poll, or fail the turn.
3. Emit fallback content only from an assistant **newer than the baseline** — ideally one whose `parentID` is the user message created for this prompt.
4. If no correlated assistant exists, throw a retryable error rather than emitting `finishReason: 'stop'`.

Optional hardening: the event consumer filters only by session id, so translation could also fence on the current user/assistant ids.

Regression test shape: pre-populate assistant A, end the event stream before assistant B exists, submit user message U, assert A is never emitted and the turn does not report success.

### Versions

- `@ai-sdk/harness` 1.0.71
- `@ai-sdk/harness-opencode` 1.0.72 (source inspected through 1.0.73 — 1.0.72/1.0.73 add credential brokering and loopback auth; `runPrompt`, `turnSettled`, and `emitContextFallback` are unchanged, so upgrading does not address this)
- `@ai-sdk/sandbox-vercel`, Vercel Sandbox, Node 24

### Current workaround

We withhold the resume cursor unless the sandbox status is literally `running`, which avoids the cold-resume path. The cost is that any chat idle past the VM timeout starts a new OpenCode conversation and loses its native context — a poor trade we'd like to stop making.

riltonfranzonee:
Fixes #18933

## Problem

When the OpenCode event stream ends without settling the current turn, `emitContextFallback` reads the session's **newest assistant message** and emits its content plus a synthetic **successful** `finish-step`. There is no correlation with the prompt just sent — no baseline id, no `parentID` check, no timestamp fence.

On a **resumed** session the newest assistant is the *previous* turn's answer, so the caller receives a stale reply — text only, without the tool calls that produced it — reported as a normal completion. We hit this three times in production: a user asked "can you write a memo with these cases?" and received the prior turn's full research answer byte-for-byte.

It's silent: no error surfaces, and the turn records as successful.

## Fix

Capture the newest assistant id **before** the prompt is sent, and decline the fallback when the snapshot still points at it:

```ts
const baselineAssistantId = (
  await latestAssistantSnapshot({ client, sessionId }).catch(() => undefined)
)?.id;
```

```ts
const assistant = await latestAssistantSnapshot({ client, sessionId });
if (!assistant) return false;
if (isStaleAssistantSnapshot({ assistant, baselineAssistantId })) return false;
```

When it declines, the existing `missingContext: true` branch reports the turn honestly rather than inventing a successful reply from stale state.

The comparison lives in `opencode-events.ts` as `isStaleAssistantSnapshot` so it's unit-testable — `bridge/index.ts` is a zero-export entry script.

## Scope

Deliberately minimal: this stops the wrong answer. It does **not** address the broader issue that stream closure resolves `turnSettled` via the consumer's `.finally()`, so an interrupted turn still ends without a real `finish-step` — it now ends *visibly* incomplete instead of *invisibly* wrong. Reconnect/poll-on-stream-death and `parentID` correlation would be the fuller fix; happy to extend this PR if you'd prefer that shape.

## Tests

Four cases in `opencode-events.test.ts`: rejects the pre-existing assistant, accepts one recorded after the baseline, accepts anything on a fresh session (no baseline), inert without a snapshot. Verified the first fails against unpatched code.

`pnpm vitest run src/bridge/` — 40 passed (8 files).

## Repro context

Sandboxes are persistent: the VM idle-stops after ~10 minutes, the snapshot survives days. A later turn resumes with a parked `resumeFrom` cursor, OpenCode restarts on preserved on-disk session storage, and any event-stream interruption in that window reaches the `.finally()` before the new turn settles — at which point the newest assistant on disk is necessarily the previous turn's.

This is the documented flow: `resumeFrom` says the adapter "should resume the existing session before accepting a new prompt", and `@ai-sdk/workflow-harness`'s `destroyOnFinish: false` parks the session so the next turn reattaches to the same conversation.

### DIFF
--- a/.changeset/olive-taxis-invite.md
+++ b/.changeset/olive-taxis-invite.md
@@ -1,0 +1,14 @@
+---
+'@ai-sdk/harness-opencode': patch
+---
+
+fix (harness-opencode): do not emit a pre-turn assistant message as the turn's reply
+
+When the OpenCode event stream ended without settling the current turn, the
+context fallback emitted the session's newest assistant message with a
+successful finish reason. On a resumed session that message can be the
+previous turn's answer, so the caller received a stale reply — text only,
+without the tool calls that produced it — reported as success. The fallback
+now compares against the newest assistant id observed before the prompt was
+sent and declines to emit when nothing new was recorded, leaving the existing
+`missingContext` path to report the turn honestly.
--- a/packages/harness-opencode/src/bridge/index.ts
+++ b/packages/harness-opencode/src/bridge/index.ts
@@ -1150,6 +1150,7 @@ async function readSessionTokens({
 }
 
 type AssistantSnapshot = {
+  id?: unknown;
   contentParts?: unknown[];
   metadata?: unknown;
   model?: unknown;
--- a/packages/harness-opencode/src/bridge/index.ts
+++ b/packages/harness-opencode/src/bridge/index.ts
@@ -17,6 +17,7 @@ import {
   createTranslationState,
   emitOpenCodeStreamStart,
   getOpenCodeEventSessionId,
+  isStaleAssistantSnapshot,
   isStepSettlementEvent,
   type TranslationState,
   unwrapOpenCodeEvent,
@@ -561,6 +562,12 @@ async function runPrompt({
     client,
     sessionId,
   }).catch(() => undefined);
+  /* The newest assistant BEFORE this prompt. On a resumed session it is the
+     previous turn's answer, and the context fallback below must never mistake
+     it for this turn's reply. */
+  const baselineAssistantId = (
+    await latestAssistantSnapshot({ client, sessionId }).catch(() => undefined)
+  )?.id;
   const eventsReady = createDeferred<void>();
   let stepUsage: HarnessUsage | undefined;
   let latestSessionTokens: OpenCodeTokenUsage | undefined;
@@ -643,6 +650,7 @@ async function runPrompt({
       state,
       emit,
       emitContent: !sawContent,
+      baselineAssistantId,
     }).catch(() => false);
     if (!emittedFallback) {
       emit({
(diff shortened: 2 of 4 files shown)


felixarntz (changes requested):
@riltonfranzonee Thank you for the PR. Per https://github.com/vercel/ai/pull/18938#issuecomment-5330603961, this requires some iteration.

felixarntz:
#18940 addresses this fully, so I'm closing this PR in favor of it. Thanks again!

PR #18941, opened 2026-08-15
fix(deepgram): send parsed transcription options to the API; stop defaulting diarize

shaper:
## Summary

Two transcription fixes in `@ai-sdk/deepgram`:

1. **Silently dropped options.** `keyterm`, `paragraphs`, `intents`, `sentiment`, and `replace` were parsed from `providerOptions.deepgram` but never mapped into the `/v1/listen` request. They are now sent as query parameters. `keyterm` is the nova-3-recommended keyword-boosting param.
2. **`diarize` no longer defaults to `true`.** Speaker diarization is a [paid Deepgram add-on](https://deepgram.com/pricing), and the provider previously sent `diarize=true` on every pre-recorded request unless explicitly opted out. It is now only sent when explicitly set — users who relied on the old default must pass `providerOptions: { deepgram: { diarize: true } }`.

Also widens the provider callable signature from `'nova-3'` to any transcription model ID, and updates the docs page (new options, diarize default/add-on note) and the transcribe example (exercises `keyterm`).

## Changesets

- `patch`: dropped-options fix + callable signature widening
- `minor` (breaking-adjacent): diarize default flip, with migration note

## Test plan

- New tests: no `diarize` param by default; `diarize=true` when set; all five dropped options appear as query params
- 28/28 package tests pass; `tsc --noEmit` clean


shaper:
Absorbed into #18942 — the transcription changes (dropped-options fix, diarize default flip, callable signature widening, keyterm example) were cherry-picked onto that branch so the whole Deepgram update ships as a single minor version with one changeset. Closing unmerged; content is all in #18942.

PR #18942, opened 2026-08-15
feat(deepgram): transcription option fixes + speech voice/language composition, usage metadata, speed passthrough, error parsing

shaper:
## Summary

Combined Deepgram provider update (this PR absorbs #18941). One `minor` changeset covers everything.

### Transcription (`/v1/listen`)

1. **Silently dropped options fixed.** `keyterm`, `paragraphs`, `intents`, `sentiment`, and `replace` were parsed from `providerOptions.deepgram` but never mapped into the request. They are now sent as query parameters. `keyterm` is the nova-3-recommended keyword-boosting param.
2. **Behavior change: `diarize` no longer defaults to `true`.** Speaker diarization is a [paid Deepgram add-on](https://deepgram.com/pricing), and the provider previously sent `diarize=true` on every pre-recorded request unless explicitly opted out. It is now only sent when explicitly set — users who relied on the old default must pass `providerOptions: { deepgram: { diarize: true } }`. (This behavior change is why the bump is `minor`.)

Also widens the provider callable signature from `'nova-3'` to any transcription model ID.

### Speech (`/v1/speak`)

3. **Voice family IDs + voice/language composition.** `aura-2` / `aura` compose the upstream model ID from the `generateSpeech` `voice` and `language` options (`<family>-<voice>-<language>`, language defaults to `en`), matching voice selection in every other AI SDK speech provider. A family ID without `voice` throws a clear error (Cartesia precedent). Full voice IDs (e.g. `aura-2-thalia-en`) keep passing through unchanged; the union is trimmed to family IDs plus the `(string & {})` escape hatch.
4. **`providerMetadata.deepgram`** carries `modelName` (resolved upstream model), `modelUuid`, `additionalModelUuids`, `charCount` (the billed character count), `breaksApplied`, `pronunciationsApplied`, `pronunciationWarnings` (when present), and `requestId` from response headers. `dg-project-id` is deliberately excluded (account identifier).
5. **`speed` passthrough**: Deepgram's REST API now documents a `speed` parameter (validated live: accepted range 0.7–1.5, 400 outside it). The `generateSpeech` `speed` option maps to it instead of warning as unsupported.

### Both endpoints

6. **Error parsing**: `APICallError.message` now carries Deepgram's real error (`{ err_code, err_msg, request_id }`) instead of the HTTP reason phrase. The legacy `{ error: { message, code } }` schema was dropped after live-probing both endpoints showed it is never returned (it was speculative boilerplate from the original transcribe PR).

Docs page, README, and examples updated; new examples: `spanish.ts`, `speed.ts`, `error-handling.ts`; transcribe example exercises `keyterm`.

## Changeset

- `minor` (driven by the `diarize` default flip; everything else is additive)

## Test plan

- 38 package tests: transcription option mapping incl. no-diarize-by-default, speech composition/passthrough/missing-voice/auto-language, provider metadata extraction, speed passthrough, error-schema unit tests, `doGenerate`-level `APICallError` err_msg surfacing
- All examples verified live against the Deepgram API (success + error paths)
- `tsc --noEmit` clean, ultracite clean


examples/ai-functions/src/generate-speech/deepgram/error-handling.ts
aayush-kapoor, line 1: should this file rather be in unit test?

packages/deepgram/src/deepgram-error.ts
```diff
@@ -1,16 +1,17 @@
 import { z } from 'zod/v4';
 import { createJsonErrorResponseHandler } from '@ai-sdk/provider-utils';
 
+// Error shape returned by the Deepgram APIs, e.g.
+// {"err_code":"INVALID_QUERY_PARAMETER","err_msg":"Invalid 'model' value of ...","request_id":"..."}
 export const deepgramErrorDataSchema = z.object({
-  error: z.object({
-    message: z.string(),
```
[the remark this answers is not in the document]
shaper: Intentional drop, verified before removal: live-probed error responses from both /v1/speak and /v1/listen (bogus model, invalid credentials, invalid parameter values — six cases total) and every one returned the {err_code, err_msg, request_id} shape, which is also the only shape Deepgram's API reference documents. The legacy {error:{message,code}} schema came from the original transcribe PR with a Google-style RESOURCE_EXHAUSTED fixture — boilerplate never observed from Deepgram. If Deepgram ever introduces a different shape, the fallback behavior (statusText + raw body preserved on APICallError) is no worse than before.

examples/ai-functions/src/generate-speech/deepgram/error-handling.ts
shaper: Both now: added a doGenerate-level unit test asserting APICallError.message surfaces err_msg via the mock server (c450ba8). Keeping the example as a live probe too — it follows the stream-text/baseten/error-message.ts precedent, and running it against the real API is what caught the schema mismatch in the first place. Happy to drop it if you'd rather keep examples success-only.

aayush-kapoor (approved):

PR #18951, opened 2026-08-16
fix(ai): reject streamObject result promises on error

### LINKED ISSUE #18930: `streamObject()` hangs forever when the stream errors or is aborted. The result promises never settle, and `onFinish` reports success with NaN usage.
### Description

I was testing `streamObject` with a mock model that fails on purpose, and I hit something I'd call a bug: when anything goes wrong, `await result.object` never resolves. Same for `result.usage`, `result.finishReason`, `result.response`, and `result.warnings`.

In `packages/ai/src/generate-object/stream-object.ts`, those delayed promises are only settled inside the `'finish'` case of the transform:

```ts
// stream-object.ts:777-801 (the only place these promises resolve in the whole file)
self._usage.resolve(usage);
self._providerMetadata.resolve(providerMetadata);
self._warnings.resolve(warnings);
self._response.resolve({ ...fullResponse, headers: response?.headers });
self._finishReason.resolve(finishReason ?? 'other');
// + _object.resolve/_object.reject inside the try/catch
```

There are three error paths, and none of them settle the promises:

1. **`doStream` throws** (network failure, 5xx, timeout, user abort). The setup catch at lines 872-886 only enqueues an `error` stream part and calls the telemetry `onError` hook. The promises stay pending.

2. **The provider emits an `error` part mid-stream.** The event processor at lines 539-545 only calls the `onError` callback. If the provider then closes the stream normally, the `flush` at lines 812-866 runs anyway and fires `onStepFinish` and `onFinish` as if the call succeeded. It reports `error: undefined`, `object: undefined`, and `usage` of NaN for prompt, completion, and total tokens (lines 814-818). I could not tell the call had failed from my telemetry or from `onFinish`.

3. **The raw stream rejects.** The stitchable stream calls `controller.error(error)` (`packages/ai/src/util/create-stitchable-stream.ts:77-83`). The error propagates to `fullStream` consumers, but the promises are still left pending, and `onError` is not called at all in this path. That is inconsistent with path 2, where `onError` does fire.

For comparison, `streamText` handles this correctly. It has `rejectResultPromises` (`stream-text.ts:2666-2703`) wired into `consumeStream`'s error handlers, so `result.finishReason`, `result.totalUsage` and friends reject instead of hanging. `streamObject` has no equivalent, which reads to me like an omission rather than a design choice.

### Reproduction

```ts
import { streamObject } from 'ai';
import { z } from 'zod';

const failingModel = {
  specificationVersion: 'v4',
  provider: 'test.mock',
  modelId: 'mock',
  async doStream() {
    throw new Error('transient provider failure');
  },
} as any;

const result = streamObject({
  model: failingModel,
  schema: z.object({ content: z.string() }),
  prompt: 'hello',
});

const timeout = new Promise((_, rej) =>
  setTimeout(() => rej(new Error('HANG: result.object never settled')), 2000),
);
const object = await Promise.race([result.object, timeout]); // never settles
console.log(object);
```

**Expected behavior:** `result.object` rejects with the provider error, the way `streamText` rejects its result promises.

**Actual behavior:** `result.object` (and `usage`, `finishReason`, `response`, `warnings`) hangs forever. When the provider sends an error part followed by a normal stream close, `onFinish` reports a successful result with NaN usage and an undefined object instead.

## Reproduction

The snippet above is the minimal example. The existing tests at `packages/ai/src/generate-object/stream-object.test.ts:209-246` only cover `partialObjectStream` and the `onError` callback; they never await `result.object`, so the hang slips through.

### AI SDK Version

- ai: 7.0.66

### Code of Conduct

haoku123:
Fixes #18930.

## Problem

`streamObject`'s result promises (`result.object`, `result.usage`, `result.finishReason`, `result.response`, `result.warnings`, `result.request`, `result.providerMetadata`) are only settled in the `'finish'` case of the processing transform. Three error paths leave them pending forever:

1. `doStream` throws (network failure, 5xx, timeout, abort). The setup catch only enqueues an error part and calls telemetry `onError` — the promises stay pending.
2. The provider emits an `error` part mid-stream. The event processor only calls the user `onError`; if the provider then closes the stream normally, `onFinish` reports a successful call with `error: undefined`, `object: undefined` and NaN usage, so callers cannot tell the call failed.
3. The raw stream rejects. The stitchable stream errors out to consumers, but the promises stay pending and the user `onError` is never called — inconsistent with path 2.

`streamText` already handles this with its `rejectResultPromises` helper; `streamObject` has no equivalent.

## Solution

Added a `rejectResultPromises` method to `DefaultStreamObjectResult` (mirroring `streamText`) that rejects every pending result promise and marks each as handled to avoid unhandled rejections:

- **Path 1**: the setup catch now rejects the result promises before enqueueing the error part.
- **Path 2**: the event processor rejects the promises and records the error in a shared `streamError` state; `flush` reports it through `onFinish` (`error` set, `finishReason: 'error'`) instead of a fake success. A schema-validation error keeps the model's finish reason (no behavior change for `TypeValidationError`).
- **Path 3**: `addStream` now passes an `onError` callback that rejects the promises and invokes the user `onError` consistently with the error-part path.

## Tests

Added three tests to `stream-object.test.ts`:
- `doStream` throwing rejects `result.object` / `result.usage` / `result.finishReason`
- a mid-stream error part rejects the promises and `onFinish` receives the error with `finishReason: 'error'`
- a raw stream error rejects the promises and invokes `onError`

Verified non-vacuous: all three fail against the previous implementation. Full `generate-object` suite passes in both node and edge environments (134 tests).

### DIFF
--- a/.changeset/stream-object-reject-result-promises.md
+++ b/.changeset/stream-object-reject-result-promises.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix (streamObject): reject result promises on error so `result.object` and friends no longer hang forever when `doStream` throws, the provider emits an error part mid-stream, or the raw stream errors
--- a/packages/ai/src/generate-object/stream-object.test.ts
+++ b/packages/ai/src/generate-object/stream-object.test.ts
@@ -94,6 +94,106 @@ describe('streamObject', () => {
   afterEach(() => {
     logWarningsSpy.mockRestore();
   });
+
+  describe('error handling', () => {
+    it('should reject result promises when doStream throws', async () => {
+      const result = streamObject({
+        model: new MockLanguageModelV4({
+          doStream: async () => {
+            throw new Error('transient provider failure');
+          },
+        }),
+        schema: z.object({ content: z.string() }),
+        prompt: 'prompt',
+        maxRetries: 0,
+      });
+
+      await expect(result.object).rejects.toThrow('transient provider failure');
+      await expect(result.usage).rejects.toThrow('transient provider failure');
+      await expect(result.finishReason).rejects.toThrow(
+        'transient provider failure',
+      );
+    });
+
+    it('should reject result promises and report the error through onFinish when the provider emits an error part mid-stream', async () => {
+      const onFinishCalls: Array<{
+        error: unknown;
+        finishReason: unknown;
+      }> = [];
+
+      const result = streamObject({
+        model: new MockLanguageModelV4({
+          doStream: async () => ({
+            stream: convertArrayToReadableStream<LanguageModelV4StreamPart>([
+              { type: 'stream-start', warnings: [] },
+              {
+                type: 'response-metadata',
+                id: 'id-0',
+                modelId: 'mock-model-id',
+                timestamp: new Date(0),
+              },
+              { type: 'text-start', id: '1' },
+              { type: 'text-delta', id: '1', delta: '{ "content": "hi" }' },
+              { type: 'error', error: new Error('mid-stream failure') },
+            ]),
+            warnings: [],
+          }),
+        }),
+        schema: z.object({ content: z.string() }),
+        prompt: 'prompt',
+        onFinish: ({ error, finishReason }) => {
+          onFinishCalls.push({ error, finishReason });
+        },
+      });
+
+      // consume the stream so the error part is processed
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by #18934, which already closed #18930.

PR #18955, opened 2026-08-17
fix: remove duplicate parseProviderOptions call in doGenerate

Ricardo-M-L:
This PR addresses: remove duplicate parseProviderOptions call in doGenerate


Ricardo-M-L:
Closing duplicate: same fix as #18958 (parseProviderOptions)

PR #18961, opened 2026-08-17
feat(openai-compatible): expose provider-specific image response fields via providerMetadata

R-Taneja:
## Background

`OpenAICompatibleImageModel.doGenerate` parses responses with a strict object keyed on `data`:

```ts
const openaiCompatibleImageResponseSchema = z.object({
  data: z.array(z.object({ b64_json: z.string() })),
});
```

Zod strips unknown keys, so any other field a provider returns is discarded before a caller can read it. `doGenerate` also never populated `providerMetadata`, so there was no channel for that data even if it survived parsing.

This is a real limitation for OpenAI-compatible providers that return useful information outside `data`. Two shapes come up in practice:

- **per-image fields** alongside `b64_json`, such as a revised prompt
- **response-level fields** alongside `data`, such as an identifier for a resource the call created, which the caller needs in order to reference it on a later request

Today both are silently dropped, and there is no workaround from outside the package short of replacing the image model.

## Change

Parse with `z.looseObject` and pass the extra fields through to `providerMetadata`, keyed by the provider options key, matching how the OpenAI provider's image model reports metadata:

- fields alongside `data` become siblings of `images`
- fields alongside `b64_json` become per-image entries in `images`
- the base64 image data itself is excluded, so it is not duplicated in the metadata

Applied to both the `/images/generations` and `/images/edits` paths.

## Notes

- Additive. `providerMetadata` was previously always `undefined`, and `images`/`warnings`/`response` are unchanged.
- The `images` key is always present, as `ImageModelV4ProviderMetadata` requires.
- Nothing provider-specific is hardcoded, since these fields vary per provider and cannot be modelled in a shared schema.

## Validation

Four cases added: response-level fields surface as siblings of `images`, per-image fields surface as `images` entries, base64 data does not leak into the metadata, and the metadata is keyed by the provider options key rather than the full provider name.

`packages/openai-compatible` has 111 pre-existing test failures on `main` in unrelated chat/completion fixtures. Verified by stash-compare that this change leaves that count unchanged and adds 4 passing tests (362 to 366).


packages/openai-compatible/src/image/openai-compatible-image-model.ts
```diff
@@ -192,16 +197,32 @@ export class OpenAICompatibleImageModel implements ImageModelV4 {
...
 
+function getProviderMetadata(
+  providerOptionsKey: string,
+  response: z.infer<typeof openaiCompatibleImageResponseSchema>,
+): ImageModelV4ProviderMetadata {
+  const { data, ...responseMetadata } = response;
+  return {
+    [providerOptionsKey]: {
+      ...responseMetadata,
+      images: data.map(
```
shaper, line 243: it used to be that metadata that wasn't inside `images` array was lost when we collated/aggregated things in the ai core generate-images logic. can you try this with n > 1 with an image model that supports it and see if it preserves the metadata?

shaper (approved):

taranggoyal70 (commented):
Reviewed pinned head `88a880f0`. The implementation now preserves unknown top-level and per-image fields, removes `b64_json` and the separately modeled `usage`, protects the reserved `images` key by writing it last, and applies the same path to generation and edits. The core aggregation change also correctly concatenates per-image metadata across split calls.

One public-behavior concern to make explicit: for split `n > maxImagesPerCall` requests, non-image metadata is last-writer-wins. The new test intentionally keeps only `styleId: 'style-2'`. If the motivating top-level field is a resource identifier created by each provider call, identifiers from earlier calls are still lost. If last-call semantics are intentional, please document that limitation; otherwise response-level metadata needs a per-call representation (or association with the corresponding image entries). The new provider-metadata shape, provider key, omitted fields, and multi-call behavior should also be added to the OpenAI-compatible provider docs.

Issue #9862, opened 2025-10-28
feature request: V6 needsApproval mid-execute instead of before execute


robertlong:
### Description

I'm working on an AI SDK powered app right now, I'm very excited to see the new `needsApproval` flag in the V6 beta. However, I just implemented something similar for my own app and this flag wouldn't be enough for my use case. I need something like the following:

```typescript
export const myTool = tool({
  description: 'Does a thing',
  inputSchema: z.object({
    input: z.string(),
  }),
  execute: async ({ input }, { waitForApproval }) => {
    const data = await fetchPrerequisteData(input);
    sendToClient(data);
    await waitForApproval();
    const result = await executeTool(input, data);
    return result;
  },
});
```

In my tool I want to fetch some content from my database and update a component on the client with that data to be shown when the user is asked to Accept/Deny a tool call. I think it's a pretty common pattern if your agent's tool call components render more than just the data provided by the tool call.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

gr2m:
Thanks for sharing, that's a great use case. I'll see if we can have first-class support for it as part of v6

gr2m:
> want to fetch some content from my database and update a component on the client with that data to be shown when the user is asked to Accept/Deny a tool call

can you split it up into two tool calls, one to fetch the data one to execute, and add the approval in-between?

TimoWilhelm:
It would be useful to be able to attach additional information to the approval request after fetching some data in the tool. For example in a tool like this, it would fetch the product information to display a better approval message.

```ts
buy_product: tool({
   title: 'Buy Product',
   description: 'Purchase a product by its ID.',
   inputSchema: z.object({
     product_id: z.string().describe('The ID of the product to purchase'),
   }),
   needsApproval: true,
   execute: async ({ product_id }) => {
  ...
```

noobmaster19:
Would love to have this feature - have very similar use case for this on my application, would much prefer to avoid a 2nd step just to trigger the execution.

christophebe:
I also have a fairly similar use case but with the possibility to modify the data by the user within a UI form. 


```
export const myTool = tool({
  description: 'Does a thing',
  inputSchema: z.object({
    input: z.string(),
  }),
  execute: async ({ input }, { waitForApproval }) => {
    const data = await fetchPrerequisteData(input);
    sendToClient(data);
    modifiedData = await waitForApproval();
    const result = await executeTool(input, modifiedData);
    return result;
  },
});
[...]

PR #18967, opened 2026-08-17
feat(ai): attach per-call context to HITL approval requests

### LINKED ISSUE #15733: Allow dynamic per-call context on tool approval requests (for rich human-in-the-loop cards)
### Summary

When a tool uses `needsApproval`, there is currently no way to attach dynamic, per-call information to the approval request so the UI can render a rich approval card. The only data the client receives in the `approval-requested` state is the tool `input` plus the static `title` and `metadata` defined on the tool. Anything computed at approval time (from the input plus a database or API lookup) cannot reach the card.

### Use case

Consider a `deleteRecords` tool that removes a set of records and everything linked to them. The action is irreversible, so it uses `needsApproval`.

Before the user approves, we want the approval card to show the real consequences computed at approval time, for example:

- "47 records will be deleted."
- "12 of them were modified in the last 7 days."
- "3 of them are referenced by other records that will also be affected."

All of this depends on the tool input plus a live lookup of current data. A generic "Allow deleteRecords?" card is not enough for an irreversible action, because the user cannot give informed consent without seeing what will actually be affected.

### What is missing today

I checked the installed types and source (ai 6.0.190, @ai-sdk/provider-utils 4.0.27):

- `title` and `metadata` on a tool are static. They are copied from the tool definition when the tool call is built, so they cannot carry per-call values.
- `ToolExecutionOptions` and the `needsApproval` function options expose no setter for `title` or `toolMetadata`.
- The `approval` object in the `approval-requested` state contains only `{ id }`.
- `ToolNeedsApprovalFunction` returns only a boolean.

So there is no supported path to send computed context alongside the approval request. This overlaps with #9862 (mid-execution approval), but the core need here is narrower: a way to attach data to the approval request that the UI can read.

### Proposed options

Any one of these would solve it:

1. Let the `needsApproval` function return an object instead of just a boolean, for example `{ approvalRequired: true, metadata: {...} }`, and surface that metadata on the `approval-requested` part.

2. Add an optional `metadata` field to the approval request itself, settable from the `needsApproval` function options, and expose it as `part.approval.metadata` on the client.

3. Provide a writer in the `needsApproval` options (similar to streaming tool data) so the tool can emit a data part that is reliably associated with the same approval request and tool call id.

### Workaround in use

For now we compute the context inside the `needsApproval` function and emit a separate `data-*` stream part keyed by the tool call id, then match it on the client to render the card. This works but it is a side channel. The association between the approval request and its context lives in our own keying convention rather than in the SDK, so it is easy to get wrong and does not survive cleanly through history reload without extra handling. A first class field would remove that fragility.

### Environment

- ai 6.0.190
- @ai-sdk/provider-utils 4.0.27

AAH20:
Fixes #15733

## Summary

Uninformed HITL (`Allow deleteRecords?`) is not consent. This adds a first-class, per-call `context` field on the same approval request / `approval-requested` part as the tool call, bound to `approvalId`, `toolCallId`, and a canonical input digest.

- `toolApproval` functions can return `{ type: 'user-approval', context }`
- Deprecated `needsApproval` functions can return `{ approvalRequired: true, context }` (boolean return is unchanged)
- Context and `inputDigest` round-trip through the UI stream and `convertToModelMessages`
- If the tool input changes after context was issued, replay fails closed (`stale input digest`)
- When `experimental_toolApprovalSecret` is set **and** context is present, HMAC v2 also binds a context digest. Requests without context keep HMAC v1 so pending approvals still verify. A v1 signature is never accepted for a request that carries context (unsigned context swap).

This is the approval-request metadata slice only. It does not implement mid-execute `waitForApproval` inside `execute` (#9862).

## Test plan

- [x] `needsApproval` / `toolApproval` attach context on `user-approval`
- [x] Approval-requested part surfaces `context` + `inputDigest`; response preserves them
- [x] History round-trip via `convertToModelMessages`
- [x] Stale input digest fails closed without an HMAC secret
- [x] HMAC v2 rejects swapped context, stripped context, v1+context, and a different `toolCallId` with the same visible context
- [x] Existing HMAC v1 / legacy newline fallback tests still pass

---

If you want this gate proved against a live agent tool path: [Instant Audit ($499)](https://a2zsoc.com/productized-services#instant-audit-tripwire) · [consultation](https://a2zsoc.com/consultation)

### DIFF
--- a/.changeset/hitl-approval-request-context.md
+++ b/.changeset/hitl-approval-request-context.md
@@ -1,0 +1,6 @@
+---
+'ai': patch
+'@ai-sdk/provider-utils': patch
+---
+
+Attach per-call context to HITL tool approval requests and bind it to the tool call identity
--- a/content/docs/03-agents/06-tool-approvals.mdx
+++ b/content/docs/03-agents/06-tool-approvals.mdx
@@ -61,6 +61,32 @@ const agent = new ToolLoopAgent({
 
 When `runCommand` is called, the agent returns a `tool-approval-request` instead of executing the tool.
 
+## Attach Per-Call Context to Approval Requests
+
+Return `{ type: 'user-approval', context }` from a `toolApproval` function when the card needs a consequence summary computed at approval time. The context is attached to the same approval request (and the `approval-requested` UI part) as the tool call.
+
+```ts
+toolApproval: {
+  deleteRecords: async ({ ids }) => {
+    const impact = await summarizeDeleteImpact(ids);
+    return {
+      type: 'user-approval',
+      context: {
+        recordCount: impact.total,
+        recentlyModified: impact.recentlyModified,
+        referencedByOthers: impact.referencedByOthers,
+      },
+    };
+  },
+},
+```
+
+In the UI, read `part.approval.context` when `part.state === 'approval-requested'`. Reloading the message history keeps the same context on that approval.
+
+The deprecated `needsApproval` function on a tool can return the same information as `{ approvalRequired: true, context }`.
+
+Approving a request whose tool input changed after the context was computed is rejected as stale. When `experimental_toolApprovalSecret` is set, the HMAC also binds the context so it cannot be swapped onto a different tool call.
+
 ## Decide Based on Tool Input
 
 Use a per-tool approval function when the decision depends on the parsed tool input. The function receives the typed input plus `toolCallId`, `messages`, `toolContext`, and `runtimeContext`.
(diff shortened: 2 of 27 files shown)


ArshVermaGit (commented):
Looks good. The per-call context and input digest are clearly bound to the approval request, and the test plan covers the important stale-input and HMAC integrity cases. Scope is also clearly limited to approval-request metadata. Approving.

ladanjohari:
Heads up on a collision, since it is easy to miss while both are open.

This PR changes the union in `packages/ai/src/generate-text/tool-approval-configuration.ts` to:

```ts
| { type: 'user-approval'; reason?: never; context?: JSONValue };
```

#19658, opened a few minutes ago, changes the same line in the same file to:

```ts
| { type: 'user-approval'; reason?: string };
```

So whichever lands second will need a rebase.

They look complementary rather than competing. Together the line would be `reason?: string; context?: JSONValue`, which is close to what #15733 is actually asking for: the reason a call was routed to a person, plus whatever else the application wants to show alongside it.

For context, #19658 came out of an issue I filed, so I am not a neutral party here. Flagging the overlap seemed more useful than not.

PR #18971, opened 2026-08-17
fix(ai): prevent deferred tools from bypassing unresolved client calls

### LINKED ISSUE #18970: Pending deferred provider-executed tool suppresses tool approval, causing AI_MissingToolResultsError
## Problem

When a tool that requires approval is called from inside an OpenAI Programmatic Tool Calling program, the step loop continues instead of halting for the approval, and the next step throws `AI_MissingToolResultsError`. The same tool called directly halts correctly and surfaces the `tool-approval-request`.

### Root cause

The continue condition ORs "all client tool calls are resolved" with "there are pending deferred provider-executed results". In `generateText` (`dist/index.js`, the `} while (` at ~L5935):

```js
} while (
  // Continue if:
  // 1. There are client tool calls that have all been executed or denied, OR
  // 2. There are pending deferred results from provider-executed tools
  (clientToolCalls.length > 0 &&
    clientToolOutputs.length + deniedToolApprovalResponses.length === clientToolCalls.length
   || pendingDeferredToolCalls.size > 0)
  && !await isStopConditionMet({ stopConditions, steps })
);
```

`streamText` has the same condition (~L9989).

`programmatic_tool_calling` is provider-executed with `supportsDeferredResults: true`, so it sits in `pendingDeferredToolCalls`. Clause 2 is therefore true on its own, and the `||` means it does not matter that a client tool call is still awaiting approval. The loop advances a step; that step runs `convertToLanguageModelPrompt`, which finds a `tool-call` that is neither `providerExecuted`, nor paired with a `tool-result`, nor present in `approvedToolCallIds`, and throws.

The stack shows the throw happening while the *next* step is being prepared (`SHOW_STACK=1 npm run repro`):

```
AI_MissingToolResultsError: Tool result is missing for tool call call_6WyUWmZcXp5G6ESai3jR3UMu.
    at convertToLanguageModelPrompt (node_modules/ai/dist/index.js:1449:11)
    at async execute               (node_modules/ai/dist/index.js:5508:38)
    at async executeGenerateText   (node_modules/ai/dist/index.js:5480:11)
    at async generateText          (node_modules/ai/dist/index.js:6034:10)
```

The `execute` frame at L5508 is the step body calling `convertToLanguageModelPrompt` for the step it is about to run, so the loop had already decided to continue. The throw is the trailing check at L1449:

```js
for (const id of approvedToolCallIds) {
  toolCallIds.delete(id);
}
if (toolCallIds.size > 0) {
  throw new MissingToolResultsError({ toolCallIds: Array.from(toolCallIds) });
}
```

`approvedToolCallIds` is built only from `tool-approval-response` parts already in the history, and there is none yet — the caller was never given a chance to supply one.

When the same tool is called directly there is no deferred provider call: clause 1 is false (`clientToolOutputs.length + deniedToolApprovalResponses.length` is `0`, not `1`), clause 2 is empty, the loop exits, and the approval request is returned to the caller — the correct behaviour. The stack above never happens because there is no second step.

So each feature is individually correct; the `||` between them is the problem. "A program is still pending" overrides "a tool call is waiting on the user", and there is no step at which the caller can supply a `tool-approval-response`.

## Expected Behavior

A client tool call awaiting approval should stop the loop regardless of pending deferred provider-executed calls, so the caller can present the request and reply with a `tool-approval-response`.

## Reproduction

https://github.com/andy-zhou/ai-sdk-ptc-approval-repro

```bash
npm install
OPENAI_API_KEY=sk-... npm run repro
```

Runs a 2×2 of `{ direct call, program call } × { no approval, approval required }`. Only one cell fails:

```
direct   approval=not-applicable   OK     steps=2 approvalRequests=0 text="Total hours across all members: **92.50"
direct   approval=user-approval    OK     steps=1 approvalRequests=1 text=""
program  approval=not-applicable   OK     steps=2 approvalRequests=0 text="Total hours across all members: **92.50"
program  approval=user-approval    THREW  AI_MissingToolResultsError
                                          Tool result is missing for tool call call_BNbC0zRttxqBFrUejylCV610.
```

Row 4 should behave like row 2.

Ruled out — each of these passes on its own, so none is the cause:

| variant                                           | result                        |
| ------------------------------------------------- | ----------------------------- |
| Programmatic Tool Calling alone                   | OK                            |
| … with `deferLoading` + `tool_search`             | OK                            |
| … through a persist/reload `UIMessage` round-trip | OK (`state=output-available`) |
| … with the tool set to `not-applicable`           | OK                            |

## Suggested Fix

Do not continue while a client tool call is still awaiting approval, whichever clause would otherwise continue the loop:

```diff
+ const clientToolCallsAwaitingApproval =
+   clientToolCalls.length > 0 &&
+   clientToolOutputs.length + deniedToolApprovalResponses.length !== clientToolCalls.length;
+
  } while (
-   (clientToolCalls.length > 0 &&
-     clientToolOutputs.length + deniedToolApprovalResponses.length === clientToolCalls.length
-    || pendingDeferredToolCalls.size > 0)
+   !clientToolCallsAwaitingApproval &&
+   (clientToolCalls.length > 0 &&
+     clientToolOutputs.length + deniedToolApprovalResponses.length === clientToolCalls.length
+    || pendingDeferredToolCalls.size > 0)
    && !await isStopConditionMet({ stopConditions, steps })
  );
```

That makes the failure a clean halt rather than a throw, matching the direct-call path. It leaves open the larger question of whether a suspended program can be _resumed_ after an approval round trip, or whether the program has to be re-run once the approval is recorded — that may need provider-side support to answer.

## Environment

- `ai`: 7.0.52
- `@ai-sdk/openai`: 4.0.36
- Model: `gpt-5.6-terra`, OpenAI Responses API
- Node: v24.14.0

Not verified: whether this is `store`-dependent. The reproduction ran on a Zero Data Retention org, where `store: true` fails earlier and unrelatedly with _"Items are not persisted for Zero Data Retention organizations"_, so only `store: false` could be tested.

Programmatic Tool Calling is rejected outright on other models (`Tool 'programmatic_tool_calling' is not supported with gpt-5.4`), so reproducing needs a key with `gpt-5.6-terra` access.

wang-kaopu:
## Background

When a deferred provider-executed tool is pending alongside a client tool call that requires approval, the deferred tool can keep the step loop running before the client approval has been resolved.

This can cause the next model call to start prematurely and result in `AI_MissingToolResultsError`.

## Summary

- prevent pending deferred provider tools from advancing the step loop while client tool calls are unresolved
- apply the same behavior to `generateText` and `streamText`
- add regression coverage for deferred provider tools combined with pending client approval

## End-to-End Verification

Verified with regression scenarios covering both `generateText` and `streamText`: when a deferred provider tool is pending and a client tool requires approval, only one model call is made and the approval request is surfaced without executing the client tool.

## Checklist

## Related Issues

Fixes #18970

### DIFF
--- a/.changeset/curvy-tools-approve.md
+++ b/.changeset/curvy-tools-approve.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Prevent pending deferred provider tools from advancing the step loop while client tool calls are unresolved.
--- a/packages/ai/src/generate-text/generate-text.test.ts
+++ b/packages/ai/src/generate-text/generate-text.test.ts
@@ -10374,6 +10374,75 @@ describe('generateText', () => {
     });
   });
 
+  describe('deferred provider tools with unresolved client approvals', () => {
+    it('should not continue to another model call while client approval is pending', async () => {
+      let modelCallCount = 0;
+      const executeClientTool = vi.fn(async () => 'result');
+
+      const result = await generateText({
+        model: new MockLanguageModelV4({
+          doGenerate: async () => {
+            modelCallCount++;
+
+            if (modelCallCount > 1) {
+              throw new Error('unexpected second model call');
+            }
+
+            return {
+              ...dummyResponseValues,
+              content: [
+                {
+                  type: 'tool-call',
+                  toolCallId: 'deferred-call-1',
+                  toolName: 'deferred_tool',
+                  input: '{}',
+                  providerExecuted: true,
+                },
+                {
+                  type: 'tool-call',
+                  toolCallType: 'function',
+                  toolCallId: 'client-call-1',
+                  toolName: 'client_tool',
+                  input: '{"value":"test"}',
+                },
+              ],
+              finishReason: { unified: 'tool-calls', raw: undefined },
+            };
+          },
+        }),
+        tools: {
+          deferred_tool: {
+            type: 'provider',
+            isProviderExecuted: true,
+            id: 'test.deferred_tool',
+            inputSchema: z.object({}),
+            outputSchema: z.object({ value: z.string() }),
+            args: {},
+            supportsDeferredResults: true,
+          },
+          client_tool: tool({
+            inputSchema: z.object({ value: z.string() }),
+            needsApproval: true,
+            execute: executeClientTool,
+          }),
+        },
+        prompt: 'test-input',
+        stopWhen: isStepCount(3),
+      });
+
+      expect(modelCallCount).toBe(1);
+      expect(result.steps).toHaveLength(1);
+      expect(
+        result.content.some(
+          part =>
+            part.type === 'tool-approval-request' &&
+            part.toolCall.toolName === 'client_tool',
+        ),
+      ).toBe(true);
+      expect(executeClientTool).not.toHaveBeenCalled();
+    });
+  });
+
   describe('logWarnings', () => {
     it('should call logWarnings with warnings from a single step', async () => {
       const expectedWarnings = [
--- a/packages/ai/src/generate-text/generate-text.ts
+++ b/packages/ai/src/generate-text/generate-text.ts
@@ -834,6 +834,7 @@ export async function generateText<
         ...initialMessages,
         ...initialResponseMessages,
       ];
+      let hasUnresolvedClientToolCalls = false;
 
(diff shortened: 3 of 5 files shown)


gr2m:
Superseded by #19052, which already closed #18970.

PR #18998, opened 2026-08-18
fix(openai): support built-in and provider-defined tools in Responses allowedTools

### LINKED ISSUE #18997: [@ai-sdk/openai] Built-in tools can't be allow-listed, allowedTools sends them as functions and OpenAI rejects the request
### Bug Description

The `allowedTools` provider option on the OpenAI Responses provider restricts which tools the model may call while still sending the full `tools` array, unlike `activeTools`, which removes tools from the request and invalidates the prompt cache (#8166). Keeping the cache intact is the whole point of the option.

It only works for plain functions. Naming a built-in tool such as `openai.tools.webSearch()` makes the request fail:

> 400 Tool choice 'web_search' not found in 'tools' parameter.

The error is misleading : `web_search` really is in `tools`. The malformed value is the `tool_choice` entry, so anyone hitting this starts by auditing their tool list, which is the wrong place to look.

Root cause is in `prepareResponsesTools` (`packages/openai/src/responses/openai-responses-prepare-tools.ts:363-375`): every allow-list entry is built with a hard-coded function shape.

```ts
tools: allowedTools.toolNames.map(name => ({
  type: 'function',
  name: toolNameMapping?.toProviderToolName(name) ?? name,
})),
```

OpenAI identifies built-in tools by **type**, not by name, so `{ type: 'function', name: 'web_search' }` matches nothing. Every provider-defined tool is affected: web search, image generation, code interpreter, file search, MCP, custom, apply patch, shell and computer.

### Steps to Reproduce

1. Declare a function tool and any built-in tool on `openai.responses(...)`.
2. Name both in the `allowedTools` provider option.
3. Send the request.

```ts
await generateText({
  model: openai.responses('gpt-5.5'),
  tools: {
    weather: tool({ inputSchema: z.object({ city: z.string() }) }),
    search: openai.tools.webSearch(),
  },
  providerOptions: {
    openai: {
      allowedTools: { toolNames: ['weather', 'search'], mode: 'auto' },
    },
  },
  prompt: 'What is the weather in San Francisco?',
});
```

The malformed request is visible without an API key by asserting on the request body — it is built wrong before it is ever sent.

### Expected Behavior

The allow-list entry describes the built-in by type, and the request succeeds:

```json
"tool_choice": {
  "type": "allowed_tools",
  "mode": "auto",
  "tools": [
    { "type": "function", "name": "weather" },
    { "type": "web_search" }
  ]
}
```

### Actual Behavior

| | Content |
| --- | --- |
| `tools` | `[{ type: 'function', name: 'weather' }, { type: 'web_search', ... }]` — correct |
| `tool_choice.tools` | `[{ type: 'function', name: 'weather' }, { type: 'function', name: 'web_search' }]` — second entry wrong |
| `warnings` | `[]` |

The correct type is already known one screen earlier: the same function builds `openaiTools[]` with the right `type` for each declared tool, then discards that when building the allow-list.

### Additional Context

OpenAI's docs only ever show `function` entries, which is presumably why this shape looked right. The accepted set is undocumented but enforced, and the API prints it when given an invalid entry:

> `file_search`, `web_search`, `web_search_preview`, `image_generation`, `computer`, `computer_use_preview`, `code_interpreter`, `programmatic_tool_calling`, `function`, `mcp`, `local_shell`, `shell`, `container_web_search`, `custom`, `apply_patch`

Probed against the live Responses API on `gpt-5.5`: `{ type: 'web_search' }` and `{ type: 'image_generation' }` return 200 and the tool is actually invoked. `mcp` additionally requires `server_label`, and `custom` requires `name`; both are rejected without them.

Three kinds of declared tool cannot be expressed in `allowed_tools` at all, this is an OpenAI limitation, not an SDK one, and all three were verified to 400:

- the tool search tool,
- tools marked deferred via `deferLoading`: deferred loading and allow-listing are competing answers to the same problem and don't compose,
- tools inside a namespace.

Dropping those silently would widen what the model may call, so they want the `unsupported` warning convention this file already uses. Dropping all of them leaves an empty allow-list, which OpenAI rejects outright.

Related: #8166 (`activeTools` busting the prompt cache) and #8225 (which asked for `allowed_tools` support before the option existed: related, but this is a bug in the option that shipped).

### Reproduction

Snippet above. PR with the fix, unit tests, wire-level tests, docs and a runnable example: #18998

### AI SDK Version

- ai: 7.0.66
- @ai-sdk/openai: 4.0.42

### Code of Conduct

MohamedRebai41:
## Background

`allowedTools` restricts which tools the model may call while still sending the full `tools` array, so the prompt cache survives. Every allow-list entry was built as `{ type: 'function', name }`, but OpenAI identifies built-in tools by type, not by name. Allow-listing any provider-defined tool therefore failed with `Tool choice 'web_search' not found in 'tools' parameter.` — an error that points at the tool list when the malformed value is the `tool_choice` entry.

Fixes #18997, which has the full reproduction and the live-API evidence.

## Summary

- Derive each `allowed_tools` entry from the tool already built in `prepareResponsesTools` instead of hard-coding `function`: type-only for built-ins, `name` for `function` and `custom`, `server_label` for `mcp`.
- Tool search, deferred (`deferLoading`) and namespaced tools cannot be expressed in `allowed_tools`; they are dropped with an `unsupported` warning, matching the convention already used in this file.
- If dropping them would empty the allow-list, throw instead — sending no restriction would silently let the model call every declared tool.
- Report ambiguous names instead of resolving them silently: a name matching both a declared tool and another tool's provider tool name resolves to the declared tool and warns; a provider tool name shared by several tools is dropped and warns; a name matching nothing keeps its current behavior and warns, so a typo is visible instead of producing an opaque 400.
- Document `allowedTools`, which was previously undocumented, and add a runnable example for `generateText` and `streamText`.

No public API change: tool names are still listed exactly as before.

## End-to-End Verification

Ran the new example against the live OpenAI Responses API on `gpt-5.5` with the same key, before and after the change:

- Before: `400 Tool choice 'web_search' not found in 'tools' parameter.`
- After: 200, no warnings, the built-in is invoked, and the declared-but-not-allow-listed tool is never called.

The entry shape for each tool type was probed directly against the API rather than inferred from the docs, which only show `function` entries. `{ type: 'web_search' }` and `{ type: 'image_generation' }` return 200 and the tool actually runs; `mcp` without `server_label` and `custom` without `name` are rejected as missing required parameters; tool search, deferred and namespaced tools return 400 in every form, which is why they are dropped rather than sent.

Re-verified on a second surface (Azure AI Foundry, `api-version=preview`) across `gpt-5`, `gpt-5.2`, `gpt-5.4`, `gpt-5.5`, `gpt-5.3-codex` and `gpt-5.6`. On `gpt-5.6`, allow-listing `programmatic_tool_calling` by type returns 200 and the program runs; sending it as a function entry fails with the same misleading error.

## Checklist

## Related Issues

Fixes #18997

### DIFF
--- a/.changeset/olive-donkeys-invite.md
+++ b/.changeset/olive-donkeys-invite.md
@@ -1,0 +1,14 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): support built-in and provider-defined tools in the Responses `allowedTools` option
+
+`allowedTools` emitted every allow-list entry as `{ type: 'function', name }`, but OpenAI identifies
+built-in tools by type. Allow-listing a declared provider-defined tool (web search, image generation,
+MCP, custom, ...) therefore failed with `Tool choice '<name>' not found in 'tools' parameter`. Entries
+are now derived from the declared tool, including the MCP server label and custom tool name.
+
+Tools that OpenAI cannot allow-list (the tool search tool, deferred tools, and namespaced tools) are
+dropped from the allow-list with a warning, and an error is thrown if that would leave the allow-list
+empty rather than silently sending an unrestricted request.
--- a/.changeset/olive-donkeys-invite.md
+++ b/.changeset/olive-donkeys-invite.md
@@ -12,3 +12,8 @@ are now derived from the declared tool, including the MCP server label and custo
 Tools that OpenAI cannot allow-list (the tool search tool, deferred tools, and namespaced tools) are
 dropped from the allow-list with a warning, and an error is thrown if that would leave the allow-list
 empty rather than silently sending an unrestricted request.
+
+Ambiguous names are now reported instead of resolved silently. A name that matches both a declared tool
+and another tool's provider tool name resolves to the declared tool and warns; a provider tool name
+shared by several tools in the same request (two MCP servers, for example) is dropped with a warning.
+A name that matches no declared tool keeps its existing behavior and is now warned about.
--- a/content/providers/01-ai-sdk-providers/03-openai.mdx
+++ b/content/providers/01-ai-sdk-providers/03-openai.mdx
@@ -1438,7 +1438,13 @@ const result = await generateText({
   tools or answer with a message; `'required'` forces it to call at least one of them.
 
 Both function tools and provider-defined tools can be allow-listed, including web search, file search,
-image generation, code interpreter, computer, apply patch, shell, custom tools, and MCP servers.
+image generation, code interpreter, computer, apply patch, shell, programmatic tool calling, custom
+tools, and MCP servers.
+
+If a name matches both a tool you declared and the canonical OpenAI name of a different tool, the tool
+you declared under that name wins, and a warning reports the ambiguity. If several tools share the same
+canonical name — two MCP servers, for example — that name is ambiguous, so it is dropped with a warning;
+name those tools as you declared them instead.
 
 <Note>
   Three kinds of tools cannot be allow-listed, because OpenAI's `allowed_tools`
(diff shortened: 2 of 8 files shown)


gr2m (approved):
Looks good, thank you!

Issue #14369, opened 2026-04-12
Gemini rejects `$ref`/`$defs` in function response content (400 INVALID_ARGUMENT)


bugisthegod:
# Google provider: Gemini rejects `$ref`/`$defs` in function response content (400 INVALID_ARGUMENT)

## Description

When a tool's `execute()` returns a raw JS object containing JSON Schema `$ref`/`$defs` (e.g. from Zod v4's `z.toJSONSchema()` with `z.lazy()`), the Google provider sends it as a structured proto in `functionResponse.response.content`. Gemini interprets the `$ref` pointers and rejects them with a 400 error:

```
APICallError [AI_APICallError]: The referenced name `#/$defs/__schema0` in
function_response.response does not match to a display_name in the function_response.parts.

statusCode: 400
status: "INVALID_ARGUMENT"
```

### Reproduction

```typescript
// npx tsx repro.ts
// apiKey: 'your-gemini-api-key',

import { createGoogleGenerativeAI } from '@ai-sdk/google';
import { jsonSchema as aiJsonSchema, stepCountIs, streamText } from 'ai';
import { z } from 'zod';

const google = createGoogleGenerativeAI({
  apiKey: 'your-gemini-api-key',
});

// Recursive schema — z.lazy() causes z.toJSONSchema() to emit $ref/$defs
const filterSchema: z.ZodTypeAny = z.lazy(() =>
  z.object({
    field: z.string().optional(),
    value: z.string().optional(),
    or: z.array(filterSchema).optional(),
    and: z.array(filterSchema).optional(),
    not: filterSchema.optional(),
  }),
);

// Must be nested inside a parent — otherwise Zod uses $ref:"#" not $defs/__schema0
const findToolSchema = z.object({
  limit: z.number().optional(),
  offset: z.number().optional(),
  filter: filterSchema.optional(),
});

const schemaWithRefs = z.toJSONSchema(findToolSchema);

async function main() {
  const result = streamText({
    model: google('gemini-2.5-flash'),
    stopWhen: stepCountIs(3),
    prompt: 'Call the get_schema tool, then summarize it.',
    tools: {
      // Raw tool definition (not using tool() helper)
      get_schema: {
        description: 'Returns the JSON Schema for the record filter',
        inputSchema: aiJsonSchema({
          type: 'object' as const,
          properties: {},
        }),
        execute: async () => {
          // Returns raw object containing $ref/$defs
          return {
            tools: [{
              name: 'find_records',
[...]

PR #19000, opened 2026-08-18
feat (provider/google): add `useParametersJsonSchema` to send tool schemas as JSON Schema

guidupuy:
## Background

Gemini tool declarations currently go out as `functionDeclarations[].parameters`, which is an OpenAPI 3.0 subset rather than JSON Schema, so `convertJSONSchemaToOpenAPISchema` rewrites every tool schema on the way out. That conversion is lossy in two ways that only surface for schemas the caller does not author — remote MCP servers, connector catalogs — which is why the Zod-shaped test surface does not catch either.

**1. It manufactures shapes the API rejects.** Consider a schema containing no combiner at all:

```jsonc
{ "type": "array", "items": { "type": ["null", "object"], "required": ["path"], "properties": { … } } }
```

The subset cannot express a multi-type node, so the conversion emits `anyOf` — while leaving `required` and `properties` beside it. Gemini requires `anyOf` to be the only field on its node, so the request is rejected:

```
Unable to submit request because `move` functionDeclaration `parameters.entries`
schema specified other fields alongside any_of. When using any_of, it must be
the only field set.
```

Schema validation runs over every declaration before inference, so one such tool fails the entire request, not just that tool. This is a real production schema (Dropbox's `move`, via its MCP server); the same shape is reported in [#14369](https://github.com/vercel/ai/issues/14369)-adjacent form across other ecosystems ([langchain-google#1216](https://github.com/langchain-ai/langchain-google/issues/1216), [google/adk-go#1085](https://github.com/google/adk-go/pull/1085)).

**2. It silently deletes `$ref`/`$defs`.** They are not in the conversion's keyword list, so they are dropped rather than rejected. A parameter defined by a `$ref` arrives as `{}` while still being listed in `required` — the model is asked for a value whose shape it was never given, with no error raised anywhere.

**2b. The same deletion also produces hard 400s**, which is the larger effect and easy to mistake for an unrelated bug. When a `$ref` sits inside the nullable-union idiom, the conversion collapses `[T, null]` by merging in the converted `T` — but `T` is a `$ref`, which converts to `{}`, so the merge contributes nothing and the node ships with no `type`:

```jsonc
// server declares
{ "anyOf": [{ "$ref": "#/$defs/ChatMessageDirection" }, { "type": "null" }], "default": null }
// conversion emits
{ "nullable": true }
```

```
Unable to submit request because `create_interaction` functionDeclaration
`parameters.direction` schema didn't specify the schema type field.
```

Pydantic emits exactly this shape for every `Optional[SomeEnum]` field, so it is common rather than exotic: **16 of the 25 failures** in the corpus below are this, against 5 for the `anyOf` case. 14 of the 16 are fixed by sending the schema as JSON Schema; the other 2 are rejected on the native field for unrelated nesting-depth reasons.

## Summary

Adds a `useParametersJsonSchema` provider option to `@ai-sdk/google`:

- when `true`, tool input schemas are sent unchanged as `functionDeclarations[].parametersJsonSchema`
- when `false` (the default), behavior is unchanged
- the two fields are mutually exclusive per declaration, so exactly one is emitted

`parametersJsonSchema` requires Gemini 2.5+ and accepts `$ref`, `$defs`, and `anyOf` with sibling keywords ([FunctionDeclaration reference](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1beta1/FunctionDeclaration)). It is the tool-side counterpart of `responseJsonSchema`, which #18325 is adding for structured outputs; this follows that PR's shape deliberately — an explicit provider option defaulting to off, rather than sniffing the schema.

## Verification

Verified against live Vertex (`gemini-3.7-flash`, `v1beta1`) over **1201 distinct tool schemas** taken from real connected MCP servers and connector catalogs across 17 providers (Klaviyo, Linear, Notion, GitHub, Slack, Stripe, Dropbox, Affinity, Webflow, Microsoft 365, Google Workspace, Granola, and others). Each schema was submitted as a function declaration on both fields:

| | schemas rejected by the API |
| --- | --- |
| `parameters` (current default) | **25** |
| `parametersJsonSchema` | **5** |

Of the 25 that fail today, 23 are fixed by the new field. The remaining 2 fail on both. The 5 native failures are 3 schemas that pass today plus those same 2 — see below.

Direct before/after on one declaration, same schema:

```
parameters           -> REJECTED 400: ...specified other fields alongside any_of...
parametersJsonSchema -> ACCEPTED
```

## Why it is opt-in

Beyond matching #18325, there is a concrete reason not to flip this by default: `parametersJsonSchema` enforces a nesting-depth limit that the lossy conversion can flatten a schema *under*. In the corpus, every schema at depth ≤ 32 is accepted natively and every one at depth ≥ 34 is rejected, so 3 deeply nested schemas that are accepted on `parameters` today are rejected on `parametersJsonSchema`. Size is not the discriminator — a 138 KB schema at depth 10 is accepted, while a 13 KB schema at depth 34 is not.

That limit does not appear to be documented, and [it has changed without announcement before](https://discuss.ai.google.dev/t/maximum-tool-nesting-depth-update-this-morning/104341), so a default switch would move a failure mode that maintainers cannot characterize. Callers that opt in can route per declaration if they need to: the fields are mutually exclusive per `FunctionDeclaration`, not per request, and a mixed `tools` array is accepted (verified).

## Relation to #17138

[#17138](https://github.com/vercel/ai/pull/17138) proposed the same wire change and was closed with *"This is the wrong approach"*, because it was justified against [#6855](https://github.com/vercel/ai/issues/6855) (Gemini ignoring tool descriptions / shortening values) and the review found no causal improvement — descriptions were already preserved by the conversion, and the before/after runs matched.

That assessment was correct for that issue. This PR makes no claim about descriptions or value fidelity. It targets two different, deterministic failures — a 400 on a schema the conversion itself produces, and silent erasure of `$ref` — both reproducible on demand, and it is gated behind an opt-in flag rather than changing the default.

## Testing

- Three unit tests in `google-prepare-tools.test.ts`: verbatim schema with `parameters` omitted, no `anyOf` manufactured from a multi-type node, and the default path unchanged.
- `packages/google`: 737 tests pass, `tsc --noEmit` clean.
- Live corpus run described above.

### DIFF
--- a/.changeset/olive-pugs-tap.md
+++ b/.changeset/olive-pugs-tap.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+feat (provider/google): add `useParametersJsonSchema` to send tool schemas as JSON Schema
--- a/content/providers/01-ai-sdk-providers/15-google.mdx
+++ b/content/providers/01-ai-sdk-providers/15-google.mdx
@@ -142,6 +142,23 @@ The following optional provider options are available for Google models:
 
   See [Troubleshooting: Schema Limitations](#schema-limitations) for more details.
 
+- **useParametersJsonSchema** _boolean_
+
+  Optional. Send tool input schemas as JSON Schema via
+  `functionDeclarations[].parametersJsonSchema` instead of the OpenAPI subset in
+  `parameters`. Default is false.
+
+  The two fields are mutually exclusive per function declaration.
+  `parametersJsonSchema` requires Gemini 2.5 or later, and accepts JSON Schema
+  keywords the OpenAPI subset cannot express (`$ref`, `$defs`, and `anyOf`
+  carrying sibling keywords), so schemas the conversion mangles or rejects are
+  sent verbatim. This is useful for tool schemas you do not author, such as
+  those advertised by remote MCP servers.
+
+  Note that Gemini enforces a nesting-depth limit on this field which the lossy
+  conversion can flatten a schema under, so a very deeply nested schema may be
+  accepted on `parameters` but rejected here.
+
 - **safetySettings** _Array\<\{ category: string; threshold: string \}\>_
 
   Optional. Safety settings for the model.
--- a/packages/google/src/google-language-model-options.ts
+++ b/packages/google/src/google-language-model-options.ts
@@ -89,6 +89,23 @@ export const googleLanguageModelOptions = lazySchema(() =>
       structuredOutputs: z.boolean().optional(),
 
       /**
+       * Optional. Send tool input schemas as JSON Schema via
+       * `functionDeclarations[].parametersJsonSchema` instead of the OpenAPI
+       * subset in `parameters`. Default is false.
+       *
+       * The two fields are mutually exclusive per function declaration.
+       * `parametersJsonSchema` requires Gemini 2.5 or later, and accepts JSON
+       * Schema keywords the OpenAPI subset cannot express (`$ref`, `$defs`,
+       * and `anyOf` carrying sibling keywords), so schemas the conversion
+       * mangles or rejects are sent verbatim.
+       *
+       * Note that Gemini enforces a nesting-depth limit on this field which
+       * the lossy conversion can flatten a schema under, so a very deeply
+       * nested schema may be accepted on `parameters` but rejected here.
+       */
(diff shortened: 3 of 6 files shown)


n1ckoates:
Superseded by https://github.com/vercel/ai/pull/20757

PR #19006, opened 2026-08-18
fix(google): preserve JSON Schema references

### LINKED ISSUE #19002: @ai-sdk/google: convertJSONSchemaToOpenAPISchema silently drops $ref/$defs in tool declarations, emitting untyped nodes Vertex rejects
## Description

`convertJSONSchemaToOpenAPISchema` in `@ai-sdk/google` does not handle `$ref`/`$defs` at all. Any tool `inputSchema` that hoists a subschema into `$defs` (the default output of Effect's `JSONSchema.make`, and of Zod's `z.toJSONSchema()` / `$refStrategy: 'root'`) reaches Gemini/Vertex as a property with **no `type`**, which the API rejects:

```
GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[locale].type: must be specified
```

The keys are dropped **by omission and without any warning** — nothing is added to `toolWarnings`, so this fails at the API rather than surfacing as an SDK-level `UnsupportedFunctionalityError` or warning.

This is about **tool/function declarations**, which is distinct from the already-open #14369 (that one is about `$ref`/`$defs` in function *response* content).

## Reproduction

```js
import { createGoogleGenerativeAI } from '@ai-sdk/google';
import { generateText, jsonSchema, tool } from 'ai';

// Referenced enum hoisted into $defs — what Effect's JSONSchema.make and
// Zod's z.toJSONSchema() emit by default.
const inputSchema = {
  type: 'object',
  properties: {
    locale: { $ref: '#/$defs/Locale', description: 'Locale for formatting' },
  },
  required: ['locale'],
  additionalProperties: false,
  $defs: {
    Locale: { type: 'string', enum: ['de', 'en'] },
  },
};

const google = createGoogleGenerativeAI({
  apiKey: 'fake-key-not-used',
    console.log(JSON.stringify(JSON.parse(init.body).tools, null, 2));
    process.exit(0);
  },
});

await generateText({
  prompt: 'hi',
  tools: {
    formatDate: tool({
      description: 'Format a date',
      inputSchema: jsonSchema(inputSchema),
      execute: async () => 'ok',
    }),
  },
});
```

### Actual output

The `$defs` entry is gone and `locale` has no `type` — only the sibling `description` survives:

```json
[
  {
    "functionDeclarations": [
      {
        "name": "formatDate",
        "description": "Format a date",
        "parameters": {
          "required": ["locale"],
          "type": "object",
          "properties": {
            "locale": {
              "description": "Locale for formatting"
            }
          }
        }
      }
    ]
  }
]
```

### Expected output

Either the reference is resolved (inlined):

```json
"locale": { "type": "string", "enum": ["de", "en"], "description": "Locale for formatting" }
```

or it is translated into the `ref`/`defs` form Vertex actually documents (see below) — and if neither is possible, a warning/`UnsupportedFunctionalityError` instead of a silently invalid request.

## Root cause

[`packages/google/src/convert-json-schema-to-openapi-schema.ts`](https://github.com/vercel/ai/blob/main/packages/google/src/convert-json-schema-to-openapi-schema.ts) destructures a fixed allowlist of keys, which does not include `$ref` or `$defs`:

```ts
const {
  type,
  description,
  required,
  properties,
  items,
  allOf,
  anyOf,
  oneOf,
  format,
  const: constValue,
  minLength,
  enum: enumValues,
} = jsonSchema;
```

Everything not in that list is discarded, so a `{ $ref, description }` node converts to `{ description }`. A repo-wide search for `defs` under `packages/google` returns no matches, confirming there is no handling anywhere in the provider.

Note that `addEnumToSchema` in the same file already throws `UnsupportedFunctionalityError` for unsupported enums, so there is existing precedent for failing loudly rather than emitting an invalid request.

## Divergence from the documented Vertex behavior

Google's structured-output/function-calling docs explicitly document reference support, with these usage notes:

> - Unlike the OpenAPI schema, specify `ref` and `defs` without the `$` symbol.
> - `ref` must refer to a direct child of `defs`; no external references.
> - The maximum depth of nested schema is 32.
> - Recursion depth in `defs` (self-reference) is limited to two.

So the API does accept references in its own `ref`/`defs` spelling, but the provider never emits that shape: it neither resolves `$ref` nor rewrites `$defs`/`$ref` to `defs`/`ref`. The result is that a documented API capability is unreachable through the SDK, and standard JSON Schema input silently degrades into an invalid request.

[#6937](https://github.com/vercel/ai/pull/6937) implemented exactly this translation (plus `title` support) but was closed without merging. [#3325](https://github.com/vercel/ai/issues/3325) reported the same `must be specified` symptom and was closed via a Zod-side `$refStrategy: 'none'` workaround, which does not help non-Zod schema sources.

## Suggested fix

In order of preference:

1. Resolve/inline `$ref` against `$defs` inside `convertJSONSchemaToOpenAPISchema` before conversion (handles the common non-recursive case fully, and is what every consumer has to hand-roll today).
2. For genuinely recursive schemas, translate `$defs`/`$ref` to `defs`/`ref` with `#/defs/` paths per the docs above (the #6937 approach), respecting the documented depth-2 self-reference limit.
3. At minimum, push a warning onto `toolWarnings` (or throw `UnsupportedFunctionalityError`) when a `$ref` cannot be represented, so the failure is attributable at the SDK boundary instead of arriving as an opaque 400.

## Workaround

We currently pre-resolve `$defs` ourselves before handing the schema to `jsonSchema()`, and throw on cyclic or unresolvable references since neither can be represented through this conversion path.

## AI SDK Version

- `ai`: 6.0.153
- `@ai-sdk/google`: 3.0.75
- `@ai-sdk/google-vertex`: 4.0.131
- Node.js: v22.23.1
- Source inspected on `main` as of this report (converter unchanged there)

Oxygen56:
## Background

Google schema conversion dropped root-level `$defs` and every `$ref`. Tool properties that relied on those definitions were therefore sent without a type, and Google rejected the request.

## Summary

- Translate root-level JSON Schema `$defs` to Google Schema `defs`.
- Translate direct `$ref` pointers from `#/$defs/<name>` to `#/defs/<name>`, including references inside recursive definitions.
- Reject reference shapes Google cannot represent instead of silently producing an invalid schema.
- Add regression coverage and a patch changeset.

## Contributor Credit

Reported and reproduced by @danielo515 in #19002.

## End-to-End Verification

A live Google API request was not run. The outgoing schema conversion was reproduced before the change and verified after it with the issue's tool schema.

## Verification

- A before/after converter harness confirmed that the baseline drops the reference and the updated converter preserves it.
- Focused compatibility checks covered nested references, unsupported references, nullable schemas, and constant enums.
- The repository changeset check and whitespace validation passed.
- Package tests, linting, and type checking could not start because this checkout does not include installed workspace dependencies; the repository CI will run them.

## Checklist

## Future Work

References outside direct children of root-level `$defs` remain unsupported. They now fail at the SDK boundary instead of reaching Google as an invalid schema.

## Related Issues

Fixes #19002.

### DIFF
--- a/.changeset/fresh-maps-reference.md
+++ b/.changeset/fresh-maps-reference.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/google': patch
+---
+
+Preserve root-level `$ref` and `$defs` when converting JSON Schema for Google requests.
--- a/packages/google/src/convert-json-schema-to-openapi-schema.test.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.test.ts
@@ -27,6 +27,75 @@ it('should remove additionalProperties and $schema', () => {
   expect(convertJSONSchemaToOpenAPISchema(input)).toEqual(expected);
 });
 
+it('should translate $ref and $defs to Google schema references', () => {
+  const input = {
+    type: 'object',
+    properties: {
+      locale: {
+        $ref: '#/$defs/Locale',
+        description: 'Locale for formatting',
+      },
+    },
+    required: ['locale'],
+    additionalProperties: false,
+    $defs: {
+      Locale: { type: 'string', enum: ['de', 'en'] },
+    },
+  } as JSONSchema7 & { $defs: Record<string, JSONSchema7> };
+
+  expect(convertJSONSchemaToOpenAPISchema(input)).toEqual({
+    type: 'object',
+    properties: {
+      locale: {
+        description: 'Locale for formatting',
+        ref: '#/defs/Locale',
+      },
+    },
+    required: ['locale'],
+    defs: {
+      Locale: { type: 'string', enum: ['de', 'en'] },
+    },
+  });
+});
+
+it('should translate references inside $defs', () => {
+  const input = {
+    type: 'object',
+    properties: {
+      node: { $ref: '#/$defs/Node' },
+    },
+    $defs: {
+      Node: {
+        type: 'object',
+        properties: {
+          child: { $ref: '#/$defs/Node' },
+        },
+      },
+    },
+  } as JSONSchema7 & { $defs: Record<string, JSONSchema7> };
+
+  expect(convertJSONSchemaToOpenAPISchema(input)).toEqual({
+    type: 'object',
+    properties: {
+      node: { ref: '#/defs/Node' },
+    },
+    defs: {
+      Node: {
+        type: 'object',
+        properties: {
+          child: { ref: '#/defs/Node' },
+        },
+      },
+    },
+  });
+});
+
+it('should reject references outside root-level $defs', () => {
+  expect(() =>
+    convertJSONSchemaToOpenAPISchema({ $ref: '#/properties/value' }),
+  ).toThrow(UnsupportedFunctionalityError);
+});
+
 it('should remove additionalProperties object from nested object schemas', function () {
   const input: JSONSchema7 = {
     type: 'object',
--- a/packages/google/src/convert-json-schema-to-openapi-schema.ts
+++ b/packages/google/src/convert-json-schema-to-openapi-schema.ts
@@ -4,6 +4,10 @@ import {
   type JSONSchema7Definition,
 } from '@ai-sdk/provider';
 
+type JSONSchema7WithDefinitions = JSONSchema7 & {
+  $defs?: Record<string, JSONSchema7Definition>;
+};
+
 /**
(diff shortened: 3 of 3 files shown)


gr2m:
Superseded by #19141, which preserves the intent of this contribution while inlining local references into the shared schema representation accepted by both Gemini Developer API and Vertex AI. Thank you @Oxygen56 for the initial fix.

PR #19010, opened 2026-08-18
feat(tools-registry): add keenable

ilya-bogin-keenable:
Adds Keenable to the tools registry. [`@keenable/ai-sdk`](https://www.npmjs.com/package/@keenable/ai-sdk) is published on npm and gives an agent two tools: `keenable_search`, which returns ranked pages with the page text already extracted, and `keenable_fetch`, which reads one URL as clean markdown.

The part worth calling out for the registry: it needs no API key. Every call falls back to Keenable's public endpoints, so the code example in this entry runs as-is for whoever copies it, with no signup in between. `KEENABLE_API_KEY` is listed because it lifts the rate limit, not because anything requires it.

Checks before submitting: the code example typechecks against `ai@7.0.66` and the published package, and both tools were run keyless against the live API from a clean install.


ilya-bogin-keenable:
Bump on this one. `@keenable/ai-sdk` is on npm and the entry now links its own docs page instead of a GitHub folder. The single red check is the Vercel preview deploy, which needs a team member to authorize a fork build.

@aayush-kapoor you merged the last two registry entries, mind taking a look?

PR #19057, opened 2026-08-19
docs(ui): add generate-a-chat-title guide

AiAlchemist0:
Fixes https://github.com/vercel/ai/issues/8443

Adds a short AI SDK UI guide for naming a chat from the first user message with `generateText`, then storing the title beside the existing [message persistence](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot-message-persistence) store.

No API change. Uses a small model so title generation does not block the reply stream.

### DIFF
--- a/content/docs/04-ai-sdk-ui/03-chatbot-generate-title.mdx
+++ b/content/docs/04-ai-sdk-ui/03-chatbot-generate-title.mdx
@@ -1,0 +1,81 @@
+---
+title: Generate a Chat Title
+description: Learn how to generate a short title from the first user message.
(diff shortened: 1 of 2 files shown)


gr2m:
Thanks for working on this. Generating a chat title is an application-specific combination of text generation and chat persistence, both of which are already covered in the documentation. A standalone guide would also need to prescribe background execution, persistence, and client refresh behavior that varies between applications, and the current example would require substantial changes in those areas. We’re not planning to add a dedicated guide, so I’m closing this.

Issue #19024, opened 2026-08-18
@ai-sdk/amazon-bedrock fails on openai models with Zod validation error


samesfahani-tuplehealth:
### Description

See repro and observe the Zod error:

```
[...]

PR #19058, opened 2026-08-19
fix(amazon-bedrock): support `reasoningContent.redactedContent` from the Converse API

### LINKED ISSUE #19062: bug(amazon-bedrock): `reasoningContent.redactedContent` (documented Converse union member) is missing from both response and stream schemas
## Description

GPT-5.6 models served through the **Bedrock Converse API** return a `reasoningContent` shape that neither of the provider's Zod schemas accepts:

```json
{ "reasoningContent": { "redactedContent": "cnNuX3BVUGgxNnRvNFZLWURn..." } }
```

`redactedContent` is not model-specific behavior — it is the second member of the documented [`ReasoningContentBlock`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html) union (`reasoningText | redactedContent`) and a member of [`ReasoningContentBlockDelta`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html) (`text | signature | redactedContent`). The schemas only accept `reasoningText` and `redactedReasoning` / `data`, and the latter two shapes do not appear in the current Converse API reference at all.

The impact differs by call path:

| Path | Behavior on `5.0.58` |
| --- | --- |
| `generateText` / `generateObject` **with tools** | **Throws** `AI_APICallError: Invalid JSON response` (HTTP 200). Tool calling is unusable. |
| `streamText` **with tools** | Emits an `AI_TypeValidationError` **error part**, then continues. The tool call still arrives, but every request produces a spurious error and the reasoning block is lost. |
| either, without tools | Works when the model returns no reasoning block. |

Reasoning blocks accompany tool calls, so in practice this affects any tool-calling / structured-output usage of these models on Converse.

## Reproduction

Verified on a clean install of `@ai-sdk/amazon-bedrock@5.0.58` + `ai@7.0.68`, Node 24, region `us-east-1`.

```ts
import { generateText, streamText, tool } from 'ai';
import { z } from 'zod';

const model = createAmazonBedrock({ region: 'us-east-1' })('us.openai.gpt-5.6-luna');

const propose = tool({
  description: 'finalize the draft',
  inputSchema: z.object({
    groups: z.array(z.object({ name: z.string(), parentGroupName: z.string() })),
  }),
});

const prompt = 'Transcribe this CSV, preserving order.\nname,parent\nSales,\nTeam1,Sales';

// throws AI_APICallError: Invalid JSON response (status 200)
await generateText({ model, tools: { propose }, toolChoice: 'required', prompt });

// completes, but fullStream contains an error part
const r = streamText({ model, tools: { propose }, toolChoice: 'required', prompt });
for await (const part of r.fullStream) {
  if (part.type === 'error') console.log(part.error);
}
```

Observed:

```
[generateText + tools] NG  name=AI_APICallError status=200 msg=Invalid JSON response
[streamText + tools]   OK  [{"groups":[{"name":"Sales","parentGroupName":""},{"name":"Team1","parentGroupName":"Sales"}]}]
  ERROR PART: AI_TypeValidationError Type validation failed:
    Value: {"contentBlockDelta":{"contentBlockIndex":0,"delta":{"reasoningContent":{"redactedContent":"cnNuX3BVUGgxNnRvN...
```

The `responseBody` carried in the non-streaming error shows the model itself answered correctly — only validation fails:

```json
{"output":{"message":{"content":[
  {"reasoningContent":{"redactedContent":"cnNuX3BVUGgxNnRvN..."}},
  {"toolUse":{"input":{"groups":[{"name":"Sales","parentGroupName":""},{"name":"Team1","parentGroupName":"Sales"}]},"name":"propose","toolUseId":"call_..."}}
],"role":"assistant"}},"stopReason":"tool_use"}
```

## Cause

`packages/amazon-bedrock/src/amazon-bedrock-chat-language-model.ts`, two places:

```ts
// AmazonBedrockResponseSchema — non-streaming
reasoningContent: z
  .union([
    z.object({ reasoningText: AmazonBedrockReasoningTextSchema }),
    z.object({ redactedReasoning: AmazonBedrockRedactedReasoningSchema }),
  ])
  .nullish(),

// AmazonBedrockStreamSchema — contentBlockDelta.delta
delta: z
  .union([
    z.object({ text: z.string() }),
    z.object({ toolUse: z.object({ input: z.string() }) }),
    z.object({ reasoningContent: z.object({ text: z.string() }) }),
    z.object({ reasoningContent: z.object({ signature: z.string() }) }),
    z.object({ reasoningContent: z.object({ data: z.string() }) }),
  ])
  .nullish(),
```

`{ redactedContent: string }` matches no branch in either union.

## Why the payload should be replayed, not just accepted

The encrypted payload restores the model's reasoning state across tool-use turns. I verified this against the raw Converse API (no SDK), running a two-turn tool conversation three ways:

| Turn-2 request | HTTP | Result |
| --- | --- | --- |
| assistant message replayed **with** `reasoningContent.redactedContent` verbatim | 200 | Reasoning state restored: correct, confident answer with **22 output tokens** and no re-reasoning |
| reasoning blocks **stripped** from the replayed assistant message | 200 | Model **re-reasons from scratch** (509 output tokens, new reasoning blocks) and loses its earlier conclusion ("The clues alone do not uniquely determine…") |
| replayed via the legacy `redactedReasoning: { data }` shape the SDK emits today | 200 | **Silently ignored** by the API (input token count identical to the stripped case) — re-reasoning and degradation as above |

So dropping the block "works" but silently degrades multi-turn quality and inflates token usage, and the existing legacy replay shape is a no-op on the current API.

## Fix

I've opened a PR that treats `redactedContent` as a first-class union member on both sides:

- both Zod schemas accept `redactedContent` (response and stream delta),
- the payload is surfaced as a `reasoning` part carrying `providerMetadata.{amazonBedrock,bedrock}.redactedContent` (mirroring how `redactedReasoning` is surfaced today),
- `convert-to-amazon-bedrock-chat-messages` replays it as `reasoningContent: { redactedContent }` on subsequent turns,
- streaming deltas are accumulated and attached once via `reasoning-end`, because a reasoning part's merged provider metadata is last-write-wins and per-delta metadata would drop earlier chunks.

Verified end-to-end against real Bedrock (`us-east-1`) in both `doGenerate` and `doStream`: turn 1 surfaces the payload, the turn-2 request contains the replayed `reasoningContent.redactedContent` block on the wire, and the model answers from its restored reasoning state. Existing `redactedReasoning` / `data` handling is left untouched.

## Note on #19024

#19024 reported the streaming variant of this error and was closed as invalid, on the grounds that GPT-5.6 is only available through `bedrock-mantle` and that `us.openai.*` inference IDs with Converse are unsupported.

The AWS API reports otherwise in `us-east-1` today — it is possible the model documentation has not caught up with what the API exposes:

```console
$ aws bedrock list-inference-profiles --region us-east-1 \
    --query 'inferenceProfileSummaries[?contains(inferenceProfileId, `openai`)].[inferenceProfileId,status]' --output text
us.openai.gpt-5.6-terra      ACTIVE
global.openai.gpt-5.6-terra  ACTIVE
us.openai.gpt-5.6-luna       ACTIVE
global.openai.gpt-5.6-luna   ACTIVE
us.openai.gpt-5.6-sol        ACTIVE
global.openai.gpt-5.6-sol    ACTIVE

$ aws bedrock get-foundation-model --region us-east-1 --model-identifier openai.gpt-5.6-luna \
    --query 'modelDetails.inferenceTypesSupported' --output text
INFERENCE_PROFILE
```

Converse serves them, including forced tool use (no SDK involved):

```console
$ aws bedrock-runtime converse --region us-east-1 --model-id us.openai.gpt-5.6-luna \
    --messages '[{"role":"user","content":[{"text":"Transcribe this CSV, preserving order.\nname,parent\nSales,\nTeam1,Sales"}]}]' \
    --tool-config '{"tools":[{"toolSpec":{"name":"propose","description":"finalize","inputSchema":{"json":{"type":"object","properties":{"groups":{"type":"array","items":{"type":"object","properties":{"name":{"type":"string"},"parentGroupName":{"type":"string"}},"required":["name","parentGroupName"]}}},"required":["groups"]}}}}],"toolChoice":{"tool":{"name":"propose"}}}'

stopReason: tool_use
toolUse.input: {"groups": [{"name": "Sales", "parentGroupName": ""}, {"name": "Team1", "parentGroupName": "Sales"}]}
```

All three models (`luna` / `terra` / `sol`) respond. #14720 — `<reasoning>` tags leaking for OpenAI models on Bedrock during multi-turn tool use — was accepted and fixed on this same Converse path, so these models appear to be in scope for the provider. In any case, the schema gap itself is model-agnostic: `redactedContent` is part of the documented Converse contract.

Filing this separately rather than commenting on #19024, since the reproduction, the affected call paths, and the fix all differ. #19044 covers the Mantle URL routing and is unrelated to the Converse path described here.

## Versions

- `@ai-sdk/amazon-bedrock@5.0.58`, `ai@7.0.68`, `zod@4`, Node 24, `us-east-1`
- The same two unions are unchanged in `4.0.136` and `5.0.23`

yu-ya4:
## Background

`redactedContent` is the second member of the documented [`ReasoningContentBlock`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html) union (`reasoningText | redactedContent`) and a member of [`ReasoningContentBlockDelta`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html), but neither the response schema nor the stream schema accepts it. Models that emit it — e.g. OpenAI models on Converse such as `us.openai.gpt-5.6-luna`, which return their encrypted reasoning in this shape alongside tool calls — currently fail validation:

- `generateText` / `generateObject` with tools throw `AI_APICallError: Invalid JSON response` on an HTTP 200 response, making tool calling unusable
- `streamText` with tools emits a spurious `AI_TypeValidationError` error part on every request and loses the reasoning block

## Summary

- Both Zod schemas accept `redactedContent` (response `ReasoningContentBlock` and stream `contentBlockDelta.delta`).
- The payload is surfaced as a `reasoning` part carrying `providerMetadata.{amazonBedrock,bedrock}.redactedContent`, mirroring how `redactedReasoning` is surfaced today.
- `convert-to-amazon-bedrock-chat-messages` replays it as `reasoningContent: { redactedContent }` on subsequent turns.
- Streaming deltas are accumulated in the content-block state and attached once via `reasoning-end`: a reasoning part's merged provider metadata is last-write-wins, so per-delta metadata would drop earlier chunks if the payload ever arrives split.
- Existing `redactedReasoning` / `data` handling is left untouched.

Surfacing + replay (rather than accept-and-drop) matters because the encrypted payload restores the model's reasoning state across tool-use turns. Verified against the raw Converse API with a two-turn tool conversation: replaying the block verbatim yields a correct answer from restored state (22 output tokens); stripping it forces the model to re-reason from scratch (509 output tokens) and it loses its earlier conclusion. Replaying via the legacy `redactedReasoning: { data }` shape is silently ignored by the current API, so a dedicated replay path for the documented shape is required.

## Verification

- `pnpm test` in `packages/amazon-bedrock`: 465 tests pass. The new tests (response mapping, stream accumulation across split deltas, multi-block separation, replay conversion) fail without the fix.
- End-to-end against real Bedrock (`us-east-1`, `us.openai.gpt-5.6-luna`), both `doGenerate` and `doStream`: turn 1 surfaces the reasoning part with `redactedContent` metadata, the turn-2 request contains the replayed `reasoningContent.redactedContent` block on the wire, no error parts, and the model answers from its restored reasoning state.

## Tasks

- [x] Documentation has been added / updated (for bug fixes / features) — provider docs describe reasoning provider metadata generically; no doc change needed
- [x] A _patch changeset_ for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)

## Future Work

Anthropic models' redacted reasoning on the current Converse API is also documented as `redactedContent`; this change covers them as well, while keeping the legacy `redactedReasoning` / `data` paths for compatibility.

## Related Issues

Closes #19062 (full reproduction and multi-turn replay measurements). Related: #19024 (closed; streaming variant of this validation error), #14720 (precedent for OpenAI-on-Converse reasoning handling).

### DIFF
--- a/.changeset/olive-pugs-shave.md
+++ b/.changeset/olive-pugs-shave.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/amazon-bedrock": patch
+---
+
+fix (provider/amazon-bedrock): support `reasoningContent.redactedContent` from the Converse API and replay it on subsequent turns
--- a/packages/amazon-bedrock/src/amazon-bedrock-api-types.ts
+++ b/packages/amazon-bedrock/src/amazon-bedrock-api-types.ts
@@ -258,6 +258,12 @@ export interface AmazonBedrockRedactedReasoningContentBlock {
   };
 }
 
+export interface AmazonBedrockRedactedContentBlock {
+  reasoningContent: {
+    redactedContent: string;
+  };
+}
+
 export type AmazonBedrockContentBlock =
   | AmazonBedrockDocumentBlock
   | AmazonBedrockGuardrailConverseContentBlock
@@ -268,4 +274,5 @@ export type AmazonBedrockContentBlock =
   | AmazonBedrockToolUseBlock
   | AmazonBedrockReasoningContentBlock
   | AmazonBedrockRedactedReasoningContentBlock
+  | AmazonBedrockRedactedContentBlock
   | AmazonBedrockCachePoint;
--- a/packages/amazon-bedrock/src/amazon-bedrock-chat-language-model-event-stream.test.ts
+++ b/packages/amazon-bedrock/src/amazon-bedrock-chat-language-model-event-stream.test.ts
@@ -167,4 +167,191 @@ describe('AmazonBedrockChatLanguageModel doStream', () => {
       ]),
     );
   });
+
+  it('streams reasoning redacted as `redactedContent` for replay', async () => {
+    // `redactedContent` is a member of the ReasoningContentBlockDelta union in
+    // the Converse API. OpenAI models on Bedrock (e.g. `us.openai.gpt-5.6-luna`)
+    // stream their encrypted reasoning in this shape. Deltas are accumulated and
+    // attached once via `reasoning-end`, because the merged provider metadata of
+    // a reasoning part is last-write-wins.
+    const model = new AmazonBedrockChatLanguageModel('us.openai.gpt-5.6-luna', {
+      baseUrl: () => 'https://bedrock-runtime.us-east-1.amazonaws.com',
+      headers: {},
+      fetch: async () =>
+        new Response(
+          createStream([
+            createEvent(
+              'contentBlockDelta',
+              JSON.stringify({
+                contentBlockIndex: 0,
+                delta: {
+                  reasoningContent: {
+                    redactedContent: 'encrypted-reasoning-',
+                  },
+                },
+              }),
+            ),
+            createEvent(
+              'contentBlockDelta',
+              JSON.stringify({
+                contentBlockIndex: 0,
+                delta: {
+                  reasoningContent: {
+                    redactedContent: 'payload',
+                  },
+                },
+              }),
+            ),
+            createEvent(
+              'contentBlockStop',
+              JSON.stringify({ contentBlockIndex: 0 }),
+            ),
+            createEvent(
+              'contentBlockDelta',
+              JSON.stringify({
+                contentBlockIndex: 1,
+                delta: { text: 'The answer is 42.' },
+              }),
+            ),
+            createEvent(
+              'contentBlockStop',
+              JSON.stringify({ contentBlockIndex: 1 }),
+            ),
+            createEvent(
+              'messageStop',
+              JSON.stringify({ stopReason: 'end_turn' }),
+            ),
+          ]),
+          {
+            status: 200,
+            headers: {
+              'content-type': 'application/vnd.amazon.eventstream',
+            },
+          },
+        ),
+      generateId: () => 'test-id',
+    });
+
+    const { stream } = await model.doStream({
+      prompt: TEST_PROMPT,
+      includeRawChunks: false,
+    });
+
+    const parts = await convertReadableStreamToArray(stream);
+
+    expect(parts.filter(part => part.type === 'error')).toStrictEqual([]);
+    expect(parts).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          type: 'reasoning-start',
+          id: '0',
+        }),
+        expect.objectContaining({
+          type: 'reasoning-end',
+          id: '0',
+          providerMetadata: {
+            amazonBedrock: {
+              redactedContent: 'encrypted-reasoning-payload',
+            },
+            bedrock: {
+              redactedContent: 'encrypted-reasoning-payload',
+            },
+          },
+        }),
+        expect.objectContaining({
+          type: 'text-delta',
+          delta: 'The answer is 42.',
+        }),
+      ]),
+    );
+  });
+
+  it('keeps multiple redacted reasoning blocks separate', async () => {
+    // OpenAI models on Bedrock can return several redacted reasoning blocks in
+    // a single response. Each block must surface its own payload.
+    const model = new AmazonBedrockChatLanguageModel('us.openai.gpt-5.6-luna', {
+      baseUrl: () => 'https://bedrock-runtime.us-east-1.amazonaws.com',
+      headers: {},
+      fetch: async () =>
+        new Response(
+          createStream([
+            createEvent(
+              'contentBlockDelta',
+              JSON.stringify({
+                contentBlockIndex: 0,
+                delta: {
+                  reasoningContent: { redactedContent: 'first-payload' },
+                },
+              }),
+            ),
+            createEvent(
+              'contentBlockStop',
+              JSON.stringify({ contentBlockIndex: 0 }),
+            ),
+            createEvent(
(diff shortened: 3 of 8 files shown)


gr2m (approved):
Looks good thank you Yuya!

PR #19059, opened 2026-08-19
feat(harness): add experimental support for steering agent conversations mid-turn

### LINKED ISSUE #18274: HarnessAgent: allow steer / interrupt messages
`HarnessAgent` and its harness abstraction layer do not yet support steering / interruption messages. We should explore how we can enable it.

felixarntz:
## Background

`HarnessAgent` did not expose steering the agent mid-turn, a common feature that many harnesses support. A few related primitives like lower-level API surfaces to submit another prompt mid-turn were already in place, but weren't properly wired up and didn't actually work end-to-end.

## Summary

This adds an experimental steering API with acknowledged delivery to the active turn's next safe input boundary.

- Add `experimental_steer` and `experimental_steerTurn` with active-turn and capability checks.
- Add ID-based bridge acknowledgements and reconnect-safe resubmission.
- Wire steering through supported harness adapters while preserving explicit unsupported-capability errors.
- Add matching weather-tool steering examples and regression coverage across harnesses.

### Known limitations

- Steering requires to operate on the same `HarnessAgentSession` instance that is actually controlling the current agent turn. Therefore, cross-process steering (e.g. in a separate REST request than where that turn was started) is currently not properly supported. The only workaround is relying on environments where the same session instance can be accessed in the other context too, e.g. a single-process dev server where you can store those session objects in memory and then retrieve them in the other request.
- The mid-turn user prompts cannot properly be represented in the AI SDK messages list, because they're injected mid-turn into what AI SDK expects to be a stream of exclusively agent-owned stream parts. Fixing this will require a higher-level concept of steering support in AI SDK Core.

## End-to-End Verification

Ran the new `steer.ts` examples in `examples/ai-functions/src/harness-agent`.

## Checklist

- [x] All commits are signed with `git commit -s`.
- [x] Tests have been added or updated and pass.
- [ ] Documentation has been added or updated.
- [x] A patch changeset has been added.
- [x] The changes have been reviewed and tested.

## Related Issues

Fixes #18274

See #19125 for making this a more proper implementation in the future


felixarntz:
The feedback in https://github.com/vercel/ai/pull/19059#issuecomment-5345928155 has been addressed in https://github.com/vercel/ai/pull/19059/commits/740800f94e4967786d75ec898f0c0c779ff99190

gr2m (approved):

felixarntz:
note: the above agent feedback can be ignored; interruption is already possible via the `suspendTurn()` API surface; this PR is about steering.

PR #19099, opened 2026-08-19
fix(harness-claude-code): graceful interrupt, query disposal, quieter aborts

kajogo777:
## Background

Aborting a turn hard-kills the CLI process, so the in-flight turn never reaches the session transcript. Turns that end through interrupt or error paths can leak orphaned `claude` processes. A caller's own abort is also logged to stderr as a bridge error.

## Summary

- Prefer the SDK's `interrupt()` on host abort (same as pressing Esc), with a hard abort as a 5s fallback. The interrupted turn stays resumable, including via `claude --resume`.
- Dispose the query when a turn ends, so the CLI subprocess doesn't outlive it.
- Settle caller-initiated aborts as aborts without logging them as errors, in both the bridge and the host.

## End-to-End Verification

Ran against the real `claude` CLI: an aborted turn settles with an `abort` part and no `error` part, nothing is logged to stderr, and no `claude` processes remain after `destroy()`.

## Checklist


felixarntz (approved):

.changeset/claude-code-turn-hygiene.md
```diff
@@ -0,0 +1,8 @@
+---
+'@ai-sdk/harness-claude-code': patch
+'@ai-sdk/harness': patch
+---
+
+fix (harness-claude-code): interrupt gracefully and dispose the query when a turn ends. A host abort now prefers the Claude SDK's `interrupt()` — Esc semantics, so the in-flight turn is persisted to the session transcript and a later resume (including the user's own `claude --resume`) still sees the work done before the interrupt — with a hard abort as a five-second fallback. The query is disposed explicitly when a turn ends through an interrupt or error path, which previously leaked orphaned `claude` processes holding the very conversation the next turn continues.
+
+fix (harness): stop diagnosing caller-initiated aborts as bridge errors. A turn the caller itself aborted ends with an error-shaped part by construction; it is still settled as an abort for the consumer, but no longer logged to stderr as a harness malfunction.
```
felixarntz:
```suggestion
fix(harness): stop diagnosing caller-initiated aborts as bridge errors
```

felixarntz:
@kajogo777 code here looks reasonable to me, but please check the agent feedback. both our factory and the Vercel agent have similar feedback.

packages/harness/src/bridge/index.ts
```diff
@@ -658,10 +664,30 @@ export async function runBridge<TStart extends { type: 'start' }>(
...
+          const pendingTurn = activeTurn;
+          if (pendingTurn == null) break;
+          turnAbort?.abort();
+          currentUserMessages?.close(
+            new Error('A new bridge turn replaced the active turn.'),
+          );
+          await pendingTurn;
+        }
+        let turnFinished!: () => void;
+        activeTurn = new Promise<void>(resolve => (turnFinished = resolve));
```
felixarntz: @kajogo777 Not sure about this one. It seems that this can hang indefinitely.

I am also concerned that this is supposed to be a Claude Code fix, but now it's leaking into general `harness` layer changes. That means it can affect all other harnesses too, and it would require a lot more testing to establish that this actually works well with each harness. Ideally we can localize the fix to Claude Code, or if that's not possible, we need to ensure there is sufficient tests of end-to-end examples for the other harnesses that cover them in combination with this change too.

The latest agent review https://github.com/vercel/ai/pull/19099#issuecomment-5414540412 still has concerns as well.

felixarntz (changes requested):

packages/harness/src/bridge/index.ts
kajogo777: had to be central because of shared state, added a timeout in a follow up comment to restore old behaviour + more tests

felixarntz (approved):
@kajogo777 Ran e2e tests across all harnesses, things are looking good!

PR #19102, opened 2026-08-19
fix(harness-claude-code): resume the exact conversation

kajogo777:
Stacked on #19100.

## Background

Resume currently uses the SDK's `continue: true`, which means "most recent thread in this workdir". Once a second thread exists there, resume silently picks the wrong conversation.

## Summary

- The bridge tracks the Claude session id from the message stream and reports it in its stop reply.
- The adapter records it in the state returned by `stop()`, `detach()`, and `suspendTurn()`, and resumes with the SDK's `resume` option. State from older versions falls back to `continue`.
- The id is also exposed as `harnessMetadata['claude-code'].sessionId` on `finish` parts, so a UI can show `claude --resume <id>`.
- Resume state is now parsed through the zod schema instead of a cast.

## End-to-End Verification

Ran against the real `claude` CLI: the id streamed on finish metadata matches the one recorded by `stop()`, and a resumed session recalls prior turn content without re-reading files.

## Checklist


felixarntz (approved):
pushed a few tweaks based on the agent feedback - good to merge!

PR #19155, opened 2026-08-20
docs: add Knowl to the agent memory providers

williamttruong:
Adds Knowl to the memory providers list in `content/docs/03-agents/06-memory.mdx`.

Knowl is a local-first memory engine for coding agents: typed knowledge atoms in SQLite under the project, with conflicts resolved at write time so a superseded fact stops being a retrieval candidate. It speaks MCP, and [`@knowl/ai-sdk`](https://www.npmjs.com/package/@knowl/ai-sdk) adapts its tools for the AI SDK.

Every provider currently on the page is a hosted service behind an API key, so this adds the first entry that runs entirely locally with no account. That is also why the second commit proposes a one-clause amendment to the closing paragraph, which currently tells readers they "take on a dependency on an external service" as a blanket statement about the whole section. It is a separate commit so you can take the entry and drop the amendment, or the reverse.

The entry follows the Supermemory and Hindsight shape, including the `__PROVIDER_IMPORT__` / `__MODEL__` placeholders and the standard Neovim prompt.

## Verification

Tested against the published server on Windows:

- 27 tools adapt through `createMCPClient`, each carrying an input schema
- project scoping verified by experiment rather than assumption: an atom written under one `projectRoot` is retrieved there and is correctly absent under another
- the client closes without leaving the child process behind

## One AI SDK bug found on the way, unrelated to this PR

Wiring `Experimental_StdioMCPTransport` to `npx` fails on Windows with `EINVAL`. `npx` resolves to `npx.cmd`, Node refuses to spawn `.cmd` files without `shell: true` since the fix for CVE-2024-27980, and `mcp-stdio` hardcodes `shell: false` (`dist/mcp-stdio/index.js`) with no override exposed on `StdioConfig`.

This affects any stdio MCP server launched via npx on Windows, not just this one. Our package works around it by routing through `cmd.exe /c`. Happy to open a separate issue with a minimal reproduction if that would be useful.

## Disclosure

I am a co-founder of Knowl.

### DIFF
--- a/content/docs/03-agents/06-memory.mdx
+++ b/content/docs/03-agents/06-memory.mdx
@@ -295,7 +295,7 @@ try {
 
 Under the hood this is an MCP server, so you can also wire it with [`createMCPClient`](/docs/reference/ai-sdk-core/create-mcp-client) directly. The package adds project scoping, process lifecycle, and a launcher fix for Windows, where `npx` resolves to `npx.cmd` and cannot be spawned without a shell.
 
-**When to use memory providers**: these providers are a good fit when you want memory without building any storage infrastructure. The tradeoff is that the provider controls memory behavior, so you have less visibility into what gets stored and how it is retrieved. You also take on a dependency on an external service.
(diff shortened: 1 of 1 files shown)


williamttruong:
Filed the Windows stdio spawn issue separately as #19157, with a minimal reproduction that uses the official everything-server rather than anything of mine. It affects any npx-launched stdio MCP server on Windows, so it seemed worth keeping out of this docs PR.

PR #19163, opened 2026-08-20
Isolate onChunk and onError throws in streamText

### LINKED ISSUE #19161: Throwing inside `onError` (or `onChunk`) breaks the stream for every consumer and hides the real error. `onFinish` never fires either.
### Description

I hit this while wiring error handling on top of `streamText`. I used the log-and-rethrow pattern in the `onError` callback:

```ts
onError({ error }) {
  console.error(error);
  throw error; // rethrow so my caller can handle it
}
```

That is a normal thing to do in JavaScript, and I expected it to be safe here. It is not. Throwing from `onError` breaks the stream for everyone downstream, and the error they end up seeing is my throw, not the provider's.

The callback is awaited directly inside the stream transform in `packages/ai/src/generate-text/stream-text.ts:1191-1206`:

```ts
async transform(chunk, controller) {
  controller.enqueue(chunk); // forward the chunk first
  await onChunk?.({ chunk: part });      // line 1196
  if (part.type === 'error') {
    const error = wrapGatewayError(part.error);
    ...
    await onError({ error });            // line 1205
  }
  ...
}
```

There is no try/catch around either call. If one of these callbacks throws, the TransformStream's readable side rejects, and:

- `textStream`, `fullStream`, and `steps` all reject with my rethrown error. The original `AISDKError` or `APICallError` is gone. It is only visible inside the callback itself, so downstream code cannot tell that the call was an API error.
- `rejectResultPromises` runs with my error instead of the real one. Code that dispatches on error type breaks, for example checks like `NoOutputGeneratedError.isInstance(error)`.
- The step `flush` never runs. No `onFinish`, no `onStepFinish`, no telemetry end.

Meanwhile, every other callback in the same file (`onStart`, `onFinish`, `onStepFinish`, `onObjectStepFinish`) is called through `notify()` (`packages/ai/src/util/notify.ts:8-18`), which wraps each callback in try/catch. Its own doc comment says "Errors in callbacks do not break the generation flow." So the intent is clear. `onError` and `onChunk` are the only two callbacks that bypass that protection. This reads like an oversight, not a choice.

This is not limited to error parts either. A throw from `onChunk` fails the same way, which is more severe because it fires on every chunk. I verified both on `streamText`.

The same unguarded `onError` also exists in `stream-object.ts` (line 549 in 7.0.70). A throwing `onError` there errors `baseStream`, so `fullStream`, `partialObjectStream`, and `textStream` all reject with the callback's throw. Note that the separate hang bug in `streamObject` (result promises never settling) was already fixed by `b181020` in 7.0.70; this masking bug is independent of that and still persists.

### Reproduction

The snippet below is self-contained and runs against `ai@7.0.70`. I confirmed on both 7.0.66 and 7.0.70 that `fullStream` and `result.finishReason` reject with `my rethrow`, and that `onFinish` never fires. I also confirmed the same failure mode when `onChunk` throws instead.

### Reproduction steps

```ts
import { MockLanguageModelV4 } from 'ai/test';

const result = streamText({
  model: new MockLanguageModelV4({
    doStream: async () => ({
      stream: new ReadableStream({
        start(controller) {
          controller.enqueue({ type: 'error', error: new Error('provider api failure') } as any);
          controller.close();
        },
      }) as any,
    }),
  }),
  prompt: 'hi',
  onError({ error }) {
    console.log('real error is:', error); // the only place the original error is visible
    throw new Error('my rethrow');        // breaks everything downstream
  },
});

try {
  for await (const part of result.fullStream) {}
} catch (e) {
  console.log(e); // 'my rethrow'. The provider error is gone.
}
console.log(await result.finishReason); // rejects with 'my rethrow' as well
```

**Expected behavior:** callback errors are contained. At most they get logged, and the real provider error keeps propagating, consistent with the `notify()` semantics used for every other callback. `onFinish` and telemetry still run.

**Actual behavior:** the provider error is masked by the callback's throw, every consumer rejects with the wrong error, and `onFinish` never runs.

### AI SDK Version

- ai: 7.0.70

### Code of Conduct

BetterAndBetterII:
## Summary

- Throwing from `streamText`'s `onChunk` or `onError` rejected the event-processor `TransformStream`. Downstream consumers then saw the callback error instead of the provider error, and `onFinish` never ran.
- Route both callbacks through `notify()`, the same isolation used by `onStart` / `onFinish` / `onStepFinish`. The original provider error still flows on the stream.
- Fixes #19161

## Test plan

- [x] RED then GREEN: `onChunk` throw does not reject the stream
- [x] RED then GREEN: `onError` throw preserves the provider error and still runs `onFinish`
- [x] RED then GREEN: result promises keep the original `doStream` failure, not the callback throw
- [x] `packages/ai/src/generate-text/stream-text.test.ts` — 418 passed on node and edge

### DIFF
--- a/.changeset/quiet-streams-hold.md
+++ b/.changeset/quiet-streams-hold.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+Isolate `onChunk` and `onError` throws in `streamText` so they do not break the stream or hide the original provider error.
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -10218,6 +10218,49 @@ describe('streamText', () => {
       expect(onStepFinish).toHaveBeenCalledOnce();
       expect(onFinish).toHaveBeenCalledOnce();
     });
+
+    it('should isolate onChunk throws from the stream', async () => {
+      const onFinish = vi.fn();
+
+      const result = streamText({
+        model: createTestModel({
+          stream: convertArrayToReadableStream([
+            { type: 'text-start', id: '1' },
+            { type: 'text-delta', id: '1', delta: 'Hello' },
+            { type: 'text-end', id: '1' },
+            {
+              type: 'finish',
+              finishReason: { unified: 'stop', raw: 'stop' },
+              usage: testUsage,
+            },
+          ]),
+        }),
+        prompt: 'test-input',
+        onChunk() {
+          throw new Error('chunk callback failed');
+        },
+        onFinish,
+      });
+
+      const [textChunks, parts, steps] = await Promise.all([
+        convertAsyncIterableToArray(result.textStream),
+        convertAsyncIterableToArray(result.fullStream),
+        result.steps,
+      ]);
+
+      expect(textChunks).toStrictEqual(['Hello']);
+      expect(parts.map(part => part.type)).toStrictEqual([
+        'start',
+        'start-step',
+        'text-start',
+        'text-delta',
+        'text-end',
+        'finish-step',
+        'finish',
+      ]);
+      expect(steps).toHaveLength(1);
+      expect(onFinish).toHaveBeenCalled();
+    });
   });
 
   describe('options.onError', () => {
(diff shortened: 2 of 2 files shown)


gr2m:
Superseded by #19187, which already closed #19161.

BetterAndBetterII:
Thanks for the close and the pointer to the complete fix.

PR #18304, opened 2026-08-01
perf: make zod imports tree-shakeable

gr2m:
## Summary

- replace runtime imports of Zod's `z` namespace with the individual factories used by `ai`, `@ai-sdk/gateway`, and `@ai-sdk/provider-utils`
- add package-local internal helpers for the shared factory subsets
- add patch changesets for the three affected packages

## Why

The `z` namespace exported by `zod/v4` includes the `locales` namespace and other Zod APIs. When our runtime schemas reference that namespace, esbuild retains those exports in the bundled `ai` entry point even though AI SDK does not use them.

The package-local helpers in this PR expose only the Zod factories each package uses. They are internal implementation details and do not change public APIs or schema behavior.

## Bundle size

Measured from `main` at `e84b8bc81540` after rebuilding `@ai-sdk/provider-utils` and `@ai-sdk/gateway`, using `pnpm --filter ai check-bundle-size`:

| Target | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| Node | 649.55 KB | 478.33 KB | 171.22 KB (26.4%) |
| Browser | 637.91 KB | 466.69 KB | 171.22 KB (26.8%) |

The esbuild metafiles attribute 143.68 KB of each baseline output directly to Zod v4 locale modules. After this change, locale-module contribution is 0 KB. The additional savings come from other unused Zod namespace exports becoming tree-shakeable.

## Validation

- `pnpm check`
- `pnpm type-check:full`
- `pnpm --filter ai test:node`
- `pnpm --filter ai test:edge`
- `pnpm --filter @ai-sdk/gateway test:node`
- `pnpm --filter @ai-sdk/provider-utils test:node`
- `pnpm --filter ai check-bundle-size`

PR #19298, opened 2026-08-22
chore(ai): bump bundle size limit to 615 KB

shaper:
Unblocks #19295.

The Tako Search gateway tool backport pushes the `ai` package's browser bundle to 610.65 KB, marginally over the current 610 KB limit (`packages/ai/scripts/check-bundle-size.ts`).

On main this isn't an issue because zod imports were made tree-shakeable (#18304) and the runtime Zod 3 dependency was removed (#18378), bringing the limit down to 450 KB. Neither change is part of `release-v6.0`, where the gateway package's existing tools already pull in the full `z` namespace, so a small limit bump is the consistent fix (see prior bumps: #18352, #17526, #16116).

Once this lands on `release-v6.0`, rebasing #19295 should turn its Bundle Size Check green.

### DIFF
--- a/packages/ai/scripts/check-bundle-size.ts
+++ b/packages/ai/scripts/check-bundle-size.ts
@@ -3,7 +3,7 @@ import { writeFileSync, statSync } from 'fs';
 import { join } from 'path';
 
 // Bundle size limits in bytes
-const LIMIT = 610 * 1024;
+const LIMIT = 615 * 1024;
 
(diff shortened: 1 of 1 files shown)


ArshVermaGit (commented):
Very small change, and the 5 KB increase is clearly justified by the current bundle size. The change is focused and looks good to me.

aayush-kapoor (approved):

PR #19303, opened 2026-08-22
fix(sveltekit-openai): add initial system prompt to chat example

### LINKED ISSUE #10220: Add initial prompt to sveltekit-openai chat example
### Description

It would be great to have an initial prompt in the sveltekit-openai chat example, as chats usually have an initial prompt. Why let users search the docs, instead of showing directly in the example how it's done?

https://github.com/vercel/ai/blob/main/examples/sveltekit-openai/src/routes/api/chat/%2Bserver.ts

### AI SDK Version

_No response_

### Code of Conduct

xuzhaorui:
## Summary

- Add an explicit system prompt to the SvelteKit OpenAI chat example.
- Await `convertToModelMessages`, which is asynchronous in the current API.

This makes the example show how to provide initial model instructions while keeping it compatible with the current AI SDK message conversion API.

Fixes #10220

## Validation

- `pnpm --filter @example/sveltekit-openai type-check`
- `pnpm --filter @example/sveltekit-openai build`
- `pnpm exec oxlint examples/sveltekit-openai/src/routes/api/chat/+server.ts`
- `pnpm exec oxfmt --check examples/sveltekit-openai/src/routes/api/chat/+server.ts`
- `git diff --check`

No changeset is included because this PR only updates an example application.

### DIFF
--- a/examples/sveltekit-openai/src/routes/api/chat/+server.ts
+++ b/examples/sveltekit-openai/src/routes/api/chat/+server.ts
@@ -18,7 +18,9 @@ export const POST = async ({ request }: { request: Request }) => {
 
   const result = streamText({
     model: openai('gpt-4o'),
-    messages: convertToModelMessages(messages),
+    system:
+      'You are a helpful assistant that can answer questions and use the available tools.',
+    messages: await convertToModelMessages(messages),
     stopWhen: isStepCount(5), // multi-steps for server-side tools
     tools: {
       // server-side tool with execute function:


taranggoyal70 (commented):
Reviewed pinned head `dbc91ac0` against #10220 and the current SvelteKit example. This is a focused, correct example fix: the initial instruction is supplied through `streamText.system`, `convertToModelMessages` is awaited as required by the current async API, and the existing tool loop remains unchanged. The change is example-only, so omitting a package changeset is appropriate. I did not find a correctness or scope issue in this diff.

PR #19306, opened 2026-08-22
fix(harness): collect every tool approval in a step before pausing the turn

### LINKED ISSUE #19302: harness: turn deadlocks when one model step emits two host tool calls that both need approval
When one model step emits **two host tool calls that both require approval**, the turn deadlocks: `runPrompt` surfaces only the first approval, while `harness-pi`'s resume barrier refuses to rerun until *every* dangling host tool call has a result. No second approval request is ever emitted, so the UI has no card to click and the turn hangs on a promise that can never settle.

Reproduced against `@ai-sdk/harness@1.0.75` / `@ai-sdk/harness-pi@1.0.77` (pi `@earendil-works/pi-coding-agent@0.80.10`). Still present on `@ai-sdk/harness@1.0.85` (pause sites byte-identical) and `@ai-sdk/harness-pi@1.0.87` (barrier region diffs clean), and on `main` — line references below are `main`.

## Mechanism

**Turn N — the second tool call is dropped.**

Approval is derived synchronously from the `tool-call` event itself, not from a separate event, so the first approval-requiring call pauses and returns:

https://github.com/vercel/ai/blob/main/packages/harness/src/agent/internal/run-prompt.ts#L961-L966

```ts
enqueueApprovalRequest({ approvalId: pendingApproval.approvalId, toolCall: pendingParsedToolCall });
await finishForHostInputPause({ completeCurrentStep: true });
return;                       // also run-prompt.ts:834-835 and :969-971
```

`return` abandons the reader, so the second `tool-call` event is never read. `stepToolCalls` only accumulates events actually read (`run-prompt.ts:762`), so the completed step records **only** the first call — the second is erased from the harness's message history.

Pi's own journal is unaffected: it commits the assistant message with **both** `toolCall` blocks and `stopReason: "toolUse"`.

**Turn N+1 — the barrier never releases.**

`doContinueTurn` cannot replay (pi has no bridge to attach to and no on-disk event log), so it calls `findDanglingHostToolCalls` (`packages/harness-pi/src/pi-session.ts:634`), which walks the restored journal for `toolCall` blocks without results. The `stopReason === "error" || "aborted"` skip does not apply, so it returns **both** calls, and `deferRerunUntilHostToolResults` (`pi-session.ts:766`) arms a barrier that only reruns when `awaiting.size === 0` (`pi-session.ts:704`).

The harness executes the one approved tool and calls `submitToolResult`; on a fresh session there is no pending entry, so it routes to `acceptDanglingHostToolResult`, clearing call 1 and leaving `awaiting = {call_2}`.

Nothing can ever satisfy `call_2`: no approval was enqueued for it, so no card exists, so no result arrives. `startRerun()` is never called and `done` never settles.

## Event ordering (verified)

Pi dispatches a step's tool calls in parallel and emits *every* `tool_execution_start` **before it invokes any tool** — in `@earendil-works/pi-agent-core/dist/agent-loop.js`, `executeToolCalls` (`:288-293`) picks `executeToolCallsParallel` unless a tool declares `executionMode: "sequential"`, and that function (`:332-365`) loops emitting `tool_execution_start` while pushing *thunks*; the thunks (which call the host tool's `execute`) only start at the trailing `await Promise.all(...)`.

So for a two-approval step the harness sees:

```
message_end            (assistant message: both toolCall blocks)
tool_execution_start   -> harness `tool-call` #1
tool_execution_start   -> harness `tool-call` #2
Promise.all begins     -> both execute() called, both pend on approval
<no further events>
```

Both `tool-call` events are already in the stream before anything blocks. `run-prompt.ts` simply returns after the first and never reads the second.

## Why this looks like an internal inconsistency rather than host misuse

Every other part of the resume contract is plural and expects N approvals:

- `getPendingToolApprovals()` returns a `Map`, and `runPrompt` iterates it in a `for` loop.
- `continueTurn` / `continueStream` take `toolApprovalContinuations` as an **array**.
- `findDanglingHostToolCalls` iterates *multiple* `toolCall` blocks per assistant message and requires **all** of them resolved before rerun.

Only the pause site emits a single approval request. A host has no API to ask for the rest, and no way to make pi emit one tool call per step, so an embedder cannot work around this.

## Reproduction

Deterministic, no model / MCP server / credentials. It builds the journal with pi's real `SessionManager`, reopens it as the harness does, runs the `findDanglingHostToolCalls` predicate, arms the barrier, then delivers only the first result — all the harness can produce.

<details><summary><code>repro.mjs</code></summary>

```js
import { SessionManager } from "@earendil-works/pi-coding-agent"
import { mkdtempSync } from "node:fs"
import { tmpdir } from "node:os"
import { join } from "node:path"

const cwd = mkdtempSync(join(tmpdir(), "pi-cwd-"))
const sessionDir = mkdtempSync(join(tmpdir(), "pi-sess-"))

// 1. Build the journal exactly as pi would after a turn where the model
//    emitted TWO write tool calls in a single assistant message.
const sm = SessionManager.create(cwd, sessionDir)
sm.appendMessage({ role: "user", content: "create two suites", timestamp: Date.now() })
sm.appendMessage({
  role: "assistant",
  content: [
    { type: "toolCall", id: "call_1", name: "create_suite", arguments: { name: "A" } },
    { type: "toolCall", id: "call_2", name: "create_suite", arguments: { name: "B" } },
  ],
  stopReason: "toolUse",
  timestamp: Date.now(),
})
const file = sm.getSessionFile?.() ?? sm.sessionFile
sm.flush?.()
console.log("journal:", file)

// 2. Reopen the way the harness does on resume.
const journal = SessionManager.open(file, sessionDir, cwd)
const messages = journal.buildSessionContext().messages

// 3. Run the REAL predicate from harness-pi findDanglingHostToolCalls (:2108).
const hostToolNames = new Set(["create_suite"])
const deliveredDanglingResults = new Map()
function findDangling() {
  const resolved = new Set(deliveredDanglingResults.keys())
  for (const m of messages) if (m.role === "toolResult") resolved.add(m.toolCallId)
  const dangling = []
  for (const m of messages) {
    if (m.role !== "assistant") continue
    if (m.stopReason === "error" || m.stopReason === "aborted") continue
    for (const b of m.content)
      if (b.type === "toolCall" && hostToolNames.has(b.name) && !resolved.has(b.id))
        dangling.push({ toolCallId: b.id, toolName: b.name })
  }
  return dangling
}

const dangling = findDangling()
console.log("stopReason:", messages.find(m => m.role === "assistant")?.stopReason)
console.log("dangling:", dangling.map(d => d.toolCallId))

// 4. Arm the barrier (deferRerunUntilHostToolResults :2177) and deliver ONLY
//    tool #1's result, which is all the harness can produce (it returned early
//    at :2171 and never enqueued an approval for #2).
let rerunStarted = false
const barrier = { awaiting: new Map(dangling.map(d => [d.toolCallId, d.toolName])), startRerun: () => { rerunStarted = true } }
console.log("barrier armed awaiting:", [...barrier.awaiting.keys()])

// submitToolResult -> acceptDanglingHostToolResult (:2264 / :2137)
barrier.awaiting.delete("call_1")
if (barrier.awaiting.size === 0) barrier.startRerun()

console.log("after approving #1, awaiting:", [...barrier.awaiting.keys()])
console.log("rerunStarted:", rerunStarted)
console.log(rerunStarted ? "RESULT: no deadlock" : "RESULT: DEADLOCK - turn hangs, no second approval card")
```

</details>

```
stopReason: toolUse
dangling: [ 'call_1', 'call_2' ]
barrier armed awaiting: [ 'call_1', 'call_2' ]
after approving #1, awaiting: [ 'call_2' ]
rerunStarted: false
RESULT: DEADLOCK - turn hangs, no second approval card
```

## Suggested fix

`harness-pi`'s barrier is the correct design; `run-prompt.ts` is the bug. Batch instead of pausing on the first: on a `tool-call` that needs approval, enqueue the request and keep reading, then pause once when the step's tool calls are exhausted. The barrier then needs no change — the host resumes with both continuations, both results land, `awaiting.size` reaches 0 and the rerun starts.

This needs one additive protocol signal, because there is today no in-band marker for *"the step has emitted all of its tool calls"*:

- **`finish-step` cannot serve.** `harness-pi`'s `finishStep()` returns nothing while `pendingStepToolCallIds.size > 0` — it is withheld until every tool call in the step has a *result*. Batching until `finish-step` would deadlock on its own.
- The `tool-call` part (`harnessV1ToolCallPartSchema`) carries no step cardinality.
- `HarnessV1ContinueTurnOptions` tells the adapter nothing about which results the host is about to submit, so the adapter cannot narrow the barrier instead.

I'm opening a PR with the smallest shape: an **optional `stepToolCallCount` on the `tool-call` part**. Adapters that know the step's tool-call set up front populate it (pi does — `message_end` precedes every `tool_execution_start`); adapters that don't omit it and keep today's pause-on-first behaviour, so nothing regresses.

The alternative, if you'd prefer it, is a new optional `tool-calls-end` stream part emitted after the step's last `tool-call` — no up-front count needed, so it also fits adapters that dispatch tool calls incrementally, at the cost of a new union member every adapter and schema touches. Happy to switch the PR over.

orcunbalcilar:
Fixes #19302.

## The deadlock

When one model step emits two host tool calls that both need approval, the turn hangs forever and the second approval is never surfaced, so nothing in the UI can unstick it.

`runPrompt` enqueues the approval request for the first such call, parks the turn and `return`s — abandoning the reader. The step's remaining `tool-call` parts are never read, so they never land in `stepToolCalls` and no approval request is ever created for them. The runtime's own journal is unaffected: Pi commits the assistant message with **both** `toolCall` blocks and `stopReason: "toolUse"`.

On resume, `findDanglingHostToolCalls` walks that journal and returns both calls, and `deferRerunUntilHostToolResults` arms a barrier that reruns only once every awaited call has a result. Approving call #1 clears it and leaves `awaiting = {call_2}` — a call the host was never told about, so no card exists, so no result can ever arrive. `startRerun()` is never called.

The barrier itself is fine, and it does not require all approvals in a single pause: `pi-session.test.ts` shows it re-arming across successive `doContinueTurn` calls with only the still-dangling calls, so serial progress works. But that only helps if the host eventually *learns* about call #2 — and the early `return` in `run-prompt.ts` is exactly what prevents it. That is the bug being fixed here.

Every other part of this contract is already plural — `getPendingToolApprovals()` returns a `Map` that `run-prompt.ts` iterates, `continueTurn`/`continueStream` take `toolApprovalContinuations` as an array, and `findDanglingHostToolCalls` iterates every unresolved `toolCall` block. Only the pause site is singular, and a host has no API to ask for the rest.

## The fix

`tool-call` parts carry a new optional `stepToolCallCount`. When the runtime reports it, `runPrompt` latches the pause request instead of returning, keeps reading, enqueues a request for every call in the step that needs host input, and parks once with all of them surfaced. A `finish-step` while a pause is latched parks too, keeping the runtime's real `finishReason`/`usage`.

Runtimes that omit the field park on the first call needing input, exactly as today, so the protocol change is additive and the `harness-pi` barrier needs no change.

`@ai-sdk/harness-pi` reports the count from `message_end`. That is sound because Pi dispatches a step's tool calls through `executeToolCallsParallel`, which emits `tool_execution_start` for every call while only pushing thunks — the host tools' `execute` runs at the trailing `Promise.all`. So `message_end` precedes every `tool-call`, and both `tool-call` parts are in the harness's stream before anything blocks:

```
message_end            (assistant message: both toolCall blocks)
tool_execution_start   -> harness `tool-call` #1
tool_execution_start   -> harness `tool-call` #2
Promise.all begins     -> both execute() pend on approval
```

## Tests

- `run-prompt.test.ts`: a step with two approval-needing calls and `stepToolCallCount: 2` surfaces both pending approvals, applies both approval inputs, and records one step whose content holds both tool calls. Verified to fail on unmodified `run-prompt.ts` (pending approvals came back as `['c1']`).
- `run-prompt.test.ts`: without the count, the turn still pauses on the first approval.
- `pi-translate.test.ts`: the count is reported on every `tool-call` of a step, and omitted for steps with no tool calls.

`packages/harness` (265 node / 242 edge) and `packages/harness-pi` (169) pass; the one `pi-auth.test.ts` failure in `harness-pi` is pre-existing on a clean tree. Both packages typecheck via `tsc --build`.

### DIFF
--- a/.changeset/lucky-doors-collect.md
+++ b/.changeset/lucky-doors-collect.md
@@ -1,0 +1,18 @@
+---
+'@ai-sdk/harness': patch
+'@ai-sdk/harness-pi': patch
+---
+
+fix (harness): collect every tool approval in a step before pausing the turn
+
+A model step that emits two host tool calls needing approval deadlocked: the turn
+paused after the first request and abandoned the reader, so the second `tool-call`
+was never read, never recorded in the step, and never surfaced as an approval
+request — while Pi's resume path refuses to rerun the turn until every dangling
+host tool call has a result.
+
+`tool-call` parts now carry an optional `stepToolCallCount`. When the runtime
+reports it, the turn keeps reading and enqueues an approval request for each call
+that needs one, pausing once with the whole step's requests surfaced. Runtimes that
+don't report it pause on the first call, as before. `@ai-sdk/harness-pi` reports it
+from `message_end`, which Pi commits before it starts executing any tool.
--- a/packages/harness-pi/src/pi-translate.test.ts
+++ b/packages/harness-pi/src/pi-translate.test.ts
@@ -176,6 +176,99 @@ describe('translatePiEvent', () => {
     expect(end.map(p => p.type)).toEqual(['tool-result', 'finish-step']);
   });
 
+  it('reports the step tool-call count on every tool-call of the step', () => {
+    const state = createPiTranslatorState();
+    emit(
+      [
+        { type: 'turn_start' } as PiSessionEvent,
+        {
+          type: 'message_start',
+          message: { role: 'assistant', content: [] },
+        } as PiSessionEvent,
+        {
+          type: 'message_end',
+          message: {
+            role: 'assistant',
+            content: [
+              { type: 'toolCall', id: 'call-1', name: 'create_suite' },
+              { type: 'toolCall', id: 'call-2', name: 'create_suite' },
+            ],
+          },
+        } as PiSessionEvent,
+      ],
+      state,
+    );
+
+    const starts = emit(
+      [
+        {
+          type: 'tool_execution_start',
+          toolCallId: 'call-1',
+          toolName: 'create_suite',
+          args: { name: 'A' },
+        } as PiSessionEvent,
+        {
+          type: 'tool_execution_start',
+          toolCallId: 'call-2',
+          toolName: 'create_suite',
+          args: { name: 'B' },
+        } as PiSessionEvent,
+      ],
+      state,
+    );
+
+    expect(
+      starts.map(part =>
+        part.type === 'tool-call'
+          ? { toolCallId: part.toolCallId, count: part.stepToolCallCount }
+          : part.type,
+      ),
+    ).toEqual([
+      { toolCallId: 'call-1', count: 2 },
+      { toolCallId: 'call-2', count: 2 },
+    ]);
+  });
+
+  it('omits the step tool-call count when the step has no tool calls', () => {
+    const state = createPiTranslatorState();
+    emit(
+      [
+        { type: 'turn_start' } as PiSessionEvent,
+        {
+          type: 'message_start',
+          message: { role: 'assistant', content: [] },
+        } as PiSessionEvent,
+        {
+          type: 'message_end',
+          message: {
+            role: 'assistant',
+            content: [{ type: 'text', text: 'k' }],
+          },
+        } as PiSessionEvent,
+      ],
+      state,
+    );
+
+    const start = translatePiEvent(
+      {
+        type: 'tool_execution_start',
+        toolCallId: 'orphan',
+        toolName: 'create_suite',
+        args: {},
+      } as PiSessionEvent,
+      state,
+    );
+
+    expect(start).toEqual([
+      {
+        type: 'tool-call',
+        toolCallId: 'orphan',
+        toolName: 'create_suite',
+        input: '{}',
+      },
+    ]);
+  });
+
   it('emits finish-step after a built-in approval request pauses the step', () => {
     const state = createPiTranslatorState();
     emit(
--- a/packages/harness-pi/src/pi-translate.ts
+++ b/packages/harness-pi/src/pi-translate.ts
@@ -30,6 +30,13 @@ export interface PiTranslatorState {
   observedToolNames: Map<string, string>;
   /** Tool ids requested by the current assistant message but not yet completed. */
   pendingStepToolCallIds: Set<string>;
+  /**
+   * How many tool calls the current assistant message requested. Known from
+   * `message_end`, which Pi emits before any `tool_execution_start`, and
+   * reported on every `tool-call` of the step so the host can collect all of
+   * the step's approvals before parking the turn for host input.
+   */
+  stepToolCallCount: number | undefined;
   /** Whether the current assistant message has opened a visible step. */
   stepOpen: boolean;
   /**
@@ -84,6 +91,7 @@ export function createPiTranslatorState(
     reasoningStarted: false,
     observedToolNames: new Map(),
     pendingStepToolCallIds: new Set(),
+    stepToolCallCount: undefined,
     stepOpen: false,
     hostToolResults: new Map(),
     dynamicToolCallIds: new Set(),
@@ -155,6 +163,7 @@ function finishStep(state: PiTranslatorState): HarnessV1StreamPart[] {
   if (!state.stepOpen || state.pendingStepToolCallIds.size > 0) return [];
   state.stepOpen = false;
   state.pendingStepToolCallIds.clear();
+  state.stepToolCallCount = undefined;
   return [
     {
       type: 'finish-step',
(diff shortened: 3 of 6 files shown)


orcunbalcilar:
Superseded by #19308 — same change, recreated from a clean branch.

PR #19319, opened 2026-08-23
fix: handle empty string tool call IDs in xai, mistral, and cohere doGenerate

### LINKED ISSUE #7834: toolCallId is empty string when using the gemini-2.5-pro model
### Description

https://github.com/vercel/ai/blob/56eb0ee9b1e4a4791309985ba3a1575daaf780c0/packages/openai-compatible/src/openai-compatible-chat-language-model.ts#L555

maybe it should be `toolCall.id || generateId()`

### AI SDK Version

- ai: 5

shoemoney:
## Background

#19091 fixed #7834 by falling back to a generated ID when a provider returns an empty string tool call ID, across eight provider packages. Three siblings were missed: the `doGenerate` tool call mappings in `@ai-sdk/xai`, `@ai-sdk/mistral`, and `@ai-sdk/cohere` still emit the raw ID. Each response schema types the ID as `z.string()`, which validates `''`, so empty IDs flow through to `toolCallId` unchanged.

## Summary

Applies the same `toolCall.id || generateId()` fallback from #19091 to the three remaining `doGenerate` sites, and mirrors its regression test in each package (empty-string ID in the response, generated ID expected in the tool call content). Adds one patch changeset covering the three packages.

Deliberately out of scope: the xai (line 609) and cohere (line 361) streaming paths also emit raw IDs with no fallback (mistral's streaming goes through `StreamingToolCallTracker` and is covered). Happy to address those in this PR or a follow-up, whichever you prefer.

## End-to-End Verification

Served each package's tool call fixture with an empty string `id` through the test server: before the change `doGenerate` returned `toolCallId: ""`, after it returns the generated ID. Full node and edge suites pass for all three packages (xai 434, mistral 111, cohere 76 in both environments); `pnpm check` and `type-check` clean.

### DIFF
--- a/.changeset/three-tools-generate.md
+++ b/.changeset/three-tools-generate.md
@@ -1,0 +1,7 @@
+---
+'@ai-sdk/xai': patch
+'@ai-sdk/mistral': patch
+'@ai-sdk/cohere': patch
+---
+
+fix: handle empty string tool call IDs
--- a/packages/cohere/src/cohere-chat-language-model.test.ts
+++ b/packages/cohere/src/cohere-chat-language-model.test.ts
@@ -104,6 +104,33 @@ describe('doGenerate', () => {
       prepareJsonFixtureResponse('cohere-tool-call');
     });
 
+    it('should generate an ID when the tool call ID is empty', async () => {
+      const body = JSON.parse(
+        fs.readFileSync('src/__fixtures__/cohere-tool-call.json', 'utf8'),
+      );
+      body.message.tool_calls[0].id = '';
+      server.urls['https://api.cohere.com/v2/chat'].response = {
+        type: 'json-value',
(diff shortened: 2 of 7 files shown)


shoemoney:
CI note: Vercel failure is pre-existing and unrelated to this PR's changed files. Only Vercel Authorization required to deploy fails while CI Test/Verify remain in action_required fork approval gating, same Vercel failure appears on other fork PRs 19503 and 19491 never on main, and changed files are cohere/mistral/xai doGenerate fallback unrelated to Vercel. No code fix required from this PR; rebase/label will clear it.

aayush-kapoor:
these don't need a pre-emptive fix. ideally we can repro for the above mentioned providers and then we make a fix for them.

PR #19418, opened 2026-08-24
feat(anthropic,minimax): add serviceTier provider option

joshualipman123:
## Background

MiniMax added a `service_tier` parameter (`'standard' | 'priority'`) on its Anthropic-compatible Messages endpoint; `priority` is billed at 1.5× standard ([MiniMax docs](https://platform.minimax.io/docs/api-reference/messages-create#body-service-tier)). Anthropic's own Messages API also accepts `service_tier` (`'auto' | 'standard_only'`).

The MiniMax provider in the AI SDK reuses `AnthropicLanguageModel` with `provider: 'minimax.messages'`, and provider options from both the canonical `anthropic` namespace and the custom provider namespace (e.g. `minimax`) are parsed against the shared Anthropic options schema.

## Summary

- `@ai-sdk/anthropic`: new `serviceTier` provider option (`'auto' | 'standard_only' | 'standard' | 'priority'`), sent as `service_tier` in the request body.
- `@ai-sdk/minimax`: `serviceTier` (`'standard' | 'priority'`) works via `providerOptions.minimax.serviceTier` through the shared schema — no separate MiniMax-side wiring needed.
- Docs entries added for both providers.

## Test plan

- `packages/anthropic/src/anthropic-language-model.test.ts`: asserts `service_tier` appears in the request body.
- `packages/minimax/src/minimax-service-tier.test.ts`: asserts `service_tier: 'priority'` on the wire when `providerOptions.minimax.serviceTier` is set.
- Anthropic 504/504, MiniMax 109/109 (node + edge) green.


shaper (commented):
🟡 Small, additive, consistent with sibling providers' `serviceTier` spelling and the shared Anthropic/MiniMax schema architecture — one suggestion worth considering.

- **[suggestion]** No client-side warning for endpoint/value mismatches: the shared schema (packages/anthropic/src/anthropic-language-model-options.ts:288) accepts all four values regardless of target, so `serviceTier: 'priority'` against Anthropic direct (or `'auto'` against MiniMax) only fails server-side. The OpenAI provider emits `unsupported-setting` warnings for invalid serviceTier combos (openai-chat-language-model.ts:300-314), and `AnthropicLanguageModel` already knows its `providerOptionsName` — a warning when values don't match the namespace would align with sibling behavior.
- **[nit]** `packages/minimax/src/minimax-chat-options.ts:21` narrows the exported type to `'standard' | 'priority'` but that schema never parses anything — runtime validation is the Anthropic schema, so `'auto'` passes at runtime despite being excluded from the type. Pre-existing pattern for this provider.
- **[nit]** No streaming test for `service_tier` — low risk since the option lives in `baseArgs` shared by `doGenerate`/`doStream`, consistent with how `inference_geo` is tested.
- **[nit]** One-test file `minimax-service-tier.test.ts` — consider folding into a provider-options suite later.

Checked clean: patch changesets present for both packages (correct per CONTRIBUTING.md's patch-only rule), naming matches openai/groq/xai, serialization matches the adjacent `inferenceGeo` spread, purely additive, and no internal details leak — PR body/docs cite only public URLs.

joshualipman123:
> 🟡 Small, additive, consistent with sibling providers' `serviceTier` spelling and the shared Anthropic/MiniMax schema architecture — one suggestion worth considering.
> 
> Agent-assisted review (on behalf of @shaper — he'll follow up with human review/approval).
> 
> * **[suggestion]** No client-side warning for endpoint/value mismatches: the shared schema (packages/anthropic/src/anthropic-language-model-options.ts:288) accepts all four values regardless of target, so `serviceTier: 'priority'` against Anthropic direct (or `'auto'` against MiniMax) only fails server-side. The OpenAI provider emits `unsupported-setting` warnings for invalid serviceTier combos (openai-chat-language-model.ts:300-314), and `AnthropicLanguageModel` already knows its `providerOptionsName` — a warning when values don't match the namespace would align with sibling behavior.
> * **[nit]** `packages/minimax/src/minimax-chat-options.ts:21` narrows the exported type to `'standard' | 'priority'` but that schema never parses anything — runtime validation is the Anthropic schema, so `'auto'` passes at runtime despite being excluded from the type. Pre-existing pattern for this provider.
> * **[nit]** No streaming test for `service_tier` — low risk since the option lives in `baseArgs` shared by `doGenerate`/`doStream`, consistent with how `inference_geo` is tested.
> * **[nit]** One-test file `minimax-service-tier.test.ts` — consider folding into a provider-options suite later.
> 
> Checked clean: patch changesets present for both packages (correct per CONTRIBUTING.md's patch-only rule), naming matches openai/groq/xai, serialization matches the adjacent `inferenceGeo` spread, purely additive, and no internal details leak — PR body/docs cite only public URLs.

implemented the namespace mismatch warning, left nits alone.

shaper (commented):
can you add example streamtext for anthropic and minimax we can run to validate this?

does the response include metadata on what tier was used?

@gr2m wdyt re: validating possible values, vs how it's done for openai? current PR seems consistent with prior for these, but different from openai.

shaper (commented):
🟡 Follows the OpenAI serviceTier pattern closely and is well-tested on the request side — response-metadata surfacing is untested and one exported interface gains a required field.

- **[suggestion]** No test covers the response-metadata half (~a third of the diff): echoed tier at anthropic-language-model.ts:1562 (generate), :2570/:2715 (stream), anthropic-messages-batch.ts:1009 (batch). The `message_start` stream capture is the kind of thing that silently regresses — add asserts for `providerMetadata.anthropic.serviceTier` in at least generate + stream.
- **[suggestion]** anthropic-message-metadata.ts:58 — `AnthropicMessageMetadata` gains a *required* `serviceTier: string | null`; the interface is exported from the package index, so external constructors break at compile time. Make it optional (`serviceTier?: string | null`) to match the patch-level changeset — OpenAI's equivalent metadata field is optional.
- **[suggestion]** For a MiniMax model with no explicit `providerOptions.minimax` key, metadata publishes only under `anthropic` (custom-namespace mirroring requires `usedCustomProviderKey`). Pre-existing behavior, but the minimax docs/example comment imply the `minimax` namespace unconditionally — add a caveat.
- **[nit]** The known-namespace tier vocabularies live in a nested ternary in `getArgs` (~368); a lookup map scales better if more Anthropic-compatible providers add tiers.
- **[nit]** 05-anthropic.mdx annotates `"auto" | "standard_only"` while the Zod enum accepts four values; the prose explains it, but a reader diffing docs vs types will notice.

Verified: changeset present at the repo-conventional patch level, no internal-data leakage, wire mapping correct (truthiness spread safe for non-empty enum values), response schemas permissive (`z.string().nullish()`), good cross-namespace test coverage.

PR #19450, opened 2026-08-25
fix(batch): align result parsing, request counts, and behavior across providers

n1ckoates:
## Background

OpenAI, Anthropic, and Google added Batch API support in separate changes. Their implementations repeated response parsing and request-count logic. They also handled some shared lifecycle and result cases differently.

This PR aligns those implementations and moves common behavior into `@ai-sdk/provider-utils`.

## Summary

- Add a shared JSON Lines response handler for streaming and validating batch results.
- Add a shared helper that validates and normalizes batch request counts.
- Return per-item failures for malformed or unsupported results while continuing to process later items.
- Standardize unsupported-content error codes.
- Preserve OpenAI reasoning output in batch results.
- Align pending-batch and missing-output errors.
- Align tests.
- Align documentation.

## End-to-End Verification

Ran the `start-text-batch` examples against the live OpenAI, Anthropic, Google, and Gateway Batch APIs. Each batch completed successfully and returned both requested results with the correct request IDs.

## Checklist


packages/provider-utils/src/normalize-batch-request-counts.ts
```diff
@@ -0,0 +1,42 @@
...
+  total: number | null | undefined;
+  pending: number | null | undefined;
+  completed: number | null | undefined;
+  failed: number | null | undefined;
+}): BatchV4Status['requestCounts'] | undefined {
+  if (
+    !isNonNegativeSafeInteger(total) ||
+    !isNonNegativeSafeInteger(pending) ||
+    !isNonNegativeSafeInteger(completed) ||
+    !isNonNegativeSafeInteger(failed) ||
```
gr2m: if you only use it negated, why not do 

```suggestion
    isNegativeSafeInteger(total) ||
    isNegativeSafeInteger(pending) ||
    isNegativeSafeInteger(completed) ||
    isNegativeSafeInteger(failed) ||
```

packages/provider-utils/src/test/test-batch-language-model-v4-result-conformance.ts
```diff
@@ -0,0 +1,129 @@
+import type { Experimental_BatchLanguageModelV4 as BatchLanguageModelV4 } from '@ai-sdk/provider';
+import { describe, expect, it } from 'vitest';
```
gr2m: `@ai-sdk/provider-utils/test` is a public API. We don't want to import `vitest` here because it's not a production dependency, and we don't want to interfere with whatever test runners people use `@ai-sdk/provider-utils/test` in

gr2m: it would be good if we somehow could get CI to fail when we do this in future. I remember we ran into this before

packages/provider-utils/src/normalize-batch-request-counts.ts
n1ckoates: would lose the `value is number` typescript predicate bc it'd have to return a boolean instead i think

i swapped the logic from `!isNonNegativeSafeInteger(total) || !isNonNegativeSafeInteger(pending)` to `isNonNegativeSafeInteger(total) && isNonNegativeSafeInteger(pending)` to make it easier to read tho.

packages/provider-utils/src/test/test-batch-language-model-v4-result-conformance.ts
n1ckoates: ended up just duplicating the batch tests across providers for now rather than trying to make them shared.

in the future i think it'd be cool to have some shared test suite for providers though to establish a sort of "baseline" that every provider has to pass. that way, we aren't for instance only testing something works in `@ai-sdk/openai` but forget to have a similar test in `@ai-sdk/anthropic`

gr2m (approved):

PR #19466, opened 2026-08-25
address gateway tako search tool feedback

ZYPX:
## Background

  Align the AI SDK Tako Search public contract with the corresponding Gateway implementation and improve
  guidance for cost-sensitive data options.

  ## Summary

  - Remove `relevance_score` and `citation_number` from the public response types and output-schema
  declarations.
  - Preserve runtime passthrough behavior for undeclared response properties.
  - Update the camelCase JSDoc and equivalent snake_case schema descriptions for data result counts, content
  inclusion, and row limits.
  - Update the example so it no longer references `citation_number`.
  - Add focused runtime and type-level contract tests.
  - Add a patch changeset for `@ai-sdk/gateway`.

  ## Checklist

  - [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm
  changeset` in the project root)


shaper (commented):

**Overall: approve-leaning, nothing blocking.** Small, focused contract alignment with runtime + type-level test coverage and a changeset. Passthrough is preserved on both schemas so the removed fields still flow through untyped, and no internal details leak into the public repo.

Findings:

- **[suggestion]** Docs inconsistency: `content/providers/01-ai-sdk-providers/00-ai-gateway.mdx` (~line 834) says omitting `maxRows` uses "your account's default inline cap (20 rows on the standard plan); maximum 2,000", while the new JSDoc/schema description says omitting returns "the max allowance set by `cards.content.export_pricing`". These read as contradictory; the PR checklist says docs were updated but no `content/` file is touched. The mdx `sources.data` section should get the same surcharge guidance, or the discrepancy explained.
- **[suggestion]** Removing `relevance_score`/`citation_number` from published interfaces is compile-breaking for anyone reading those fields (the tool already shipped). Patch bump is defensible since they were internal-only and recent, but the changeset could explicitly note that code reading these fields needs updating (values still arrive at runtime, just untyped).
- **[nit]** `tako-search.test.ts` asserts full description strings verbatim — any copy edit breaks the test without behavioral change. Asserting key substrings (e.g. contains "data surcharge") would be less churny.
- **[nit]** The `include_contents` description is ~90 words and is sent to the model on every tool-enabled call — noticeably longer than sibling tools (exa/perplexity); could be tightened.

shaper (approved):
the two field removals are a breaking change which we would normally only do with a minor version bump, but this just shipped and we haven't announced yet, so I think it's ok.

PR #19508, opened 2026-08-25
feat(harness): support programmatic authentication environments

bryan-hunter:
```ts
createFx({
  auth: {
    AI_GATEWAY_API_KEY: await resolveGatewayToken(),
    AI_GATEWAY_BASE_URL: 'https://ai-gateway.vercel.sh',
  },
});
```

- Accept a readonly authentication environment through `auth` alongside each adapter's existing string modes.
- Treat a supplied record as the complete input for authentication discovery; string modes retain their existing `process.env` behavior.
- Parse and forward only adapter-recognized authentication variables rather than injecting the record as arbitrary subprocess environment.
- Apply the contract across ACP, Claude Code, Cline, Codex, Cursor, DeepAgents, FX, Grok Build, OpenCode, and Pi.
- Preserve shipped legacy structured auth objects while recognizing the new flat string record without a deprecation warning.
- Source ACP `credentialEnv` values from the supplied authentication environment while continuing to source non-credential `forwardEnv` values from the host.
- Keep credential values out of ACP bootstrap, lifecycle identity, and observability; retain request-transformation credential brokering.
- Use Cursor authentication environments for `CURSOR_API_KEY` only; Cursor account settings still control model-provider routing.
- Add patch changesets for the shared harness package and all affected adapters.

Impact: applications can pass runtime-resolved provider or Gateway credentials without mutating `process.env`. Because supplied environments do not fall back to ambient credentials, unrelated host OIDC or provider tokens cannot override a programmatic credential or be sent to its configured origin.


felixarntz (changes requested):
@bryan-hunter the idea sounds reasonable to me, but I think we should find an API shape that we can consistently apply between the harness adapters, and that also covers the non-AI Gateway route.

per what I sent on Slack, maybe `'auto' | 'direct' | 'ai-gateway' | Record<string, string>`? where `Record<string, string>` is basically auth-related env vars with their values, acting as a `process.env` override for that purpose (`process.env` wouldn't be read if this record format is passed as `auth`)

marking as requesting changes while we figure out the API layer

bryan-hunter:
Pivoted this to the API shape from the review:

```ts
auth: {
  AI_GATEWAY_API_KEY: token,
  AI_GATEWAY_BASE_URL: gatewayUrl,
}
```

- A flat string record replaces `process.env` for authentication discovery; string modes keep their existing ambient behavior.
- Adapters only consume their recognized auth variables rather than forwarding arbitrary entries.
- Applied consistently to ACP, Claude Code, Cline, Codex, Cursor, DeepAgents, FX, Grok Build, OpenCode, and Pi.
- Existing shipped structured object forms remain supported as deprecated compatibility paths.
- ACP reads `credentialEnv` from the supplied record while keeping non-auth `forwardEnv` on the host environment.
- Cursor uses the record for `CURSOR_API_KEY` only; Cursor settings still own provider routing.

The isolation path also resolves the original ambient OIDC precedence issue because host credentials are not visible when a record is supplied

felixarntz (approved):

PR #19515, opened 2026-08-25
fix(providers): clamp negative inputTokens.noCache for cache-exclusive providers

kkdawkins:
## Problem

Provider converters derive uncached input tokens by subtraction, with no floor:

```ts
// packages/openai-compatible/src/chat/convert-openai-compatible-chat-usage.ts
noCache: promptTokens - cacheReadTokens,

// packages/openai/src/chat/convert-openai-chat-usage.ts
noCache: promptTokens - cachedTokens - (cacheWriteTokens ?? 0),
```

Providers disagree about whether the prompt-token count they report already includes cached tokens. For the ones that report it **exclusive** of cache, this underflows and `noCache` goes negative — which happens whenever the cached prefix exceeds the fresh remainder. That is routine in agentic traffic, where a long cached context is replayed against a handful of new tokens.

Exactly one converter guarded this today: xAI (`convert-xai-chat-usage.ts`).

### Observed in production

Found via AI Gateway, whose usage-fact schema declares `inputTokens` as `{ type: 'integer', minimum: 0 }`. 421 log samples over 3 days, 100% identical:

```
TypeError: Data does not match AI_GATEWAY_REQUEST schema: data/inputTokens must be >= 0
```

Spanning 8+ teams, 3 API surfaces, 3 regions, many SDK versions. The negative value reached both the reporting pipeline (fact dropped entirely) and cost computation.

## Fix

Infer the reporting convention instead of assuming it. A cache breakdown larger than the reported total can only happen when the reported value already *is* the uncached count — in which case the true total is the sum:

```ts
const reportedIncludesCached = cached <= reported;
total:   reportedIncludesCached ? reported : reported + cached
noCache: reportedIncludesCached ? reported - cached : reported
```

A plain `Math.max(0, …)` would stop the negative but leave `total` understated, so the inference is the correct fix rather than just the safe one.

Added as `resolveInputTokenUsage` in `@ai-sdk/provider-utils` and routed every affected converter through it. xAI's local guard is replaced by the shared helper, so the semantic now lives in one place.

## Affected converters

openai (chat + responses), openai-compatible, google (+ interactions), groq, mistral, deepseek, moonshotai, alibaba, huggingface, deepinfra, google-vertex (xai), open-responses, xai.

`@ai-sdk/openai-compatible` is the widest blast radius — ~35 downstream providers build on it.

## Prior art

The **output** half of this exact bug was fixed twice — 83e65105d6 and 221425865d — both clamping `outputTokens.text` when reasoning exceeds completion. Neither touched the input half. `convert-openai-compatible-chat-usage.ts` currently clamps output and not input, on adjacent lines.

## Tests

- New `resolve-input-token-usage.test.ts` (9 cases): cache-inclusive, cache-exclusive, malformed input, and an invariant check that `0 <= noCache <= total`.
- Per-converter `cached > prompt` regression tests across 10 converters, including 3 new test files (openai responses, google, huggingface).
- Verified the new tests **fail** without the source fix.
- Full suites green across all 14 packages: 31/31 turbo tasks, ~4,200 tests, no regressions.

## Follow-up

Needs a backport to `release-v6.0` (spec v3), which carries the same defect. `release-v5.0` has no `noCache` and is unaffected.

### DIFF
--- a/.changeset/olive-buttons-cheer.md
+++ b/.changeset/olive-buttons-cheer.md
@@ -1,0 +1,22 @@
+---
+'@ai-sdk/provider-utils': patch
+'@ai-sdk/openai-compatible': patch
+'@ai-sdk/google-vertex': patch
+'@ai-sdk/open-responses': patch
+'@ai-sdk/huggingface': patch
+'@ai-sdk/moonshotai': patch
+'@ai-sdk/deepinfra': patch
+'@ai-sdk/deepseek': patch
+'@ai-sdk/alibaba': patch
+'@ai-sdk/mistral': patch
+'@ai-sdk/openai': patch
+'@ai-sdk/google': patch
+'@ai-sdk/groq': patch
+'@ai-sdk/xai': patch
+---
+
+Fix negative `inputTokens.noCache` when providers report cache-exclusive prompt tokens.
+
+Provider converters derived uncached input tokens as `prompt_tokens - cached_tokens` with no floor. Providers that report `prompt_tokens` exclusive of cached tokens made that subtraction underflow, producing a negative `noCache` whenever the cached prefix exceeded the fresh remainder — routine in agentic traffic replaying a long cached context against a handful of new tokens.
+
+Converters now infer the reporting convention instead of assuming it: a cache breakdown larger than the reported total can only mean the reported value already *is* the uncached count, in which case the true total is the sum. Shared as `resolveInputTokenUsage` in `@ai-sdk/provider-utils`, which also generalizes the guard xAI already had.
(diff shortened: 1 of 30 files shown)


taranggoyal70 (commented):
Reviewed pinned head `bfd9e872`. The shared-helper direction and converter coverage are strong, but the convention inference is not sound for all cache-exclusive responses.

`resolveInputTokenUsage` assumes `cachedTokens <= reportedInputTokens` proves the reported count includes cached tokens. A cache-exclusive provider can legitimately report 400 fresh tokens and 100 cached tokens; the helper returns `{ total: 400, noCache: 300 }`, while the actual usage is `{ total: 500, noCache: 400 }`. Equality is ambiguous for the same reason. The current tests label only `cached > reported` cases as cache-exclusive, so they mirror the heuristic and miss this counterexample.

That means the patch prevents the negative-underflow symptom but can still silently undercount billing/telemetry after migrating all fourteen providers. The reporting convention needs to be explicit per converter/provider (or derived from an authoritative provider total/contract), with a regression where a cache-exclusive response has `cached <= reported`.

PR #19573, opened 2026-08-25
docs(skills): improve provider package guide

gr2m:
## Summary

- rewrite the add-provider-package skill around current repository and release conventions
- add official OpenAPI discovery guidance with live-response verification
- cover ProviderV4 factories, workflow serialization, secure URL handling, repository integration, and npm Trusted Publisher bootstrap
- align AI Functions example and fixture skills with the nested provider directory convention

## Validation

- `pnpm check`
- `quick_validate.py skills/add-provider-package`
- `quick_validate.py skills/develop-ai-functions-example`
- `quick_validate.py skills/capture-api-response-test-fixture`
- `git diff --check`

No changeset is needed because this only updates internal skill documentation.

### DIFF
--- a/skills/add-provider-package/SKILL.md
+++ b/skills/add-provider-package/SKILL.md
@@ -1,368 +1,208 @@
 ---
 name: add-provider-package
-description: Guide for adding new AI provider packages to the AI SDK. Use when creating a new @ai-sdk/<provider> package to integrate an AI service into the SDK.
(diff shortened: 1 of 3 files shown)


n1ckoates (approved):
LGTM, tho I wonder how much of this should be in a skill instead of in contributing docs, with the skill just pointing there.

PR #19597, opened 2026-08-25
fix(provider/moonshotai): validate media MIME types

gr2m:
Fixes #19548

## Summary

- validate inline image and video media types against the official Moonshot vision allowlists
- reject unsupported inputs such as SVG and Ogg video before sending a request
- retain wildcard URL handling, base64 data URIs, and ms:// file references
- document supported formats and add a patch changeset

## Tests

- pnpm test:node (packages/moonshotai): 125 passed
- pnpm test:edge (packages/moonshotai): 125 passed
- pnpm type-check (packages/moonshotai)
- pnpm type-check:full
- pnpm check

## Stack dependency

Temporarily based on #19596 so each Moonshot parity change stays isolated while the earlier priority PRs await required review. Retarget this PR after #19596 merges.

PR #19599, opened 2026-08-25
fix(provider/moonshotai): accept stream chunk variants

### LINKED ISSUE #19546: [Moonshot] Accept documented streaming chunk variants
## Endpoint and priority

`POST /v1/chat/completions` SSE response — **P0** because the current validation can reject an officially conforming chunk and can lose documented usage.

## Official specification evidence

The [OpenAPI `ChoiceDelta` schema](https://platform.kimi.ai/docs/openapi.json) omits tool-call `index` and includes optional choice-level `usage`; `ChatCompletionChunk` also defines top-level usage. First-party examples commonly include tool-call indices and top-level final usage, so both variants must be tolerated.

## Current behavior

Streaming tool-call `index` is required by the parser, while choice-level usage is ignored.

## Expected behavior

Accept tool calls with or without an explicit index and collect usage from either documented location, preferring the top-level value when both occur.

## Proposed implementation

Make index nullish in the wire schema, derive a deterministic per-array fallback index for the tracker, add choice usage to the schema, and normalize usage selection.

## Acceptance criteria

- Index-less tool-call chunks parse and assemble correctly.
- Explicit indices retain current behavior.
- Top-level and choice-level usage variants produce correct totals.
- Malformed chunks still surface validation errors.

## Required artifacts

Captured-style SSE fixtures and streaming unit tests; no public type/docs/example required; patch changeset.

## Backport applicability

- v6: expected applicable because it owns stream parsing.
- v5: only applicable if the generic OpenAI-compatible stream parser has the same rejection/loss and a provider-local fix is possible; otherwise document as not applicable.

gr2m:
Fixes #19546

## Summary

- accept streaming tool-call deltas whose documented index is absent or null
- derive stable positional fallback indices while preserving explicit wire indices
- collect usage from either the chunk or its first choice, with top-level usage taking precedence
- add captured-style SSE fixtures for index-less parallel tools, usage variants, and malformed chunks
- add a patch changeset

## Tests

- pnpm test:node (packages/moonshotai): 127 passed
- pnpm test:edge (packages/moonshotai): 127 passed
- pnpm type-check (packages/moonshotai)
- pnpm type-check:full
- pnpm check

## Stack dependency

Temporarily based on #19597 so each Moonshot parity change stays isolated while the earlier priority PRs await required review. Retarget this PR after #19597 merges.

### DIFF
--- a/.changeset/fuzzy-kimis-stream.md
+++ b/.changeset/fuzzy-kimis-stream.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/moonshotai': patch
+---
+
+fix(provider/moonshotai): accept documented streaming tool-call and usage variants
--- a/packages/moonshotai/src/__fixtures__/moonshotai-stream-malformed-tool-call.chunks.txt
+++ b/packages/moonshotai/src/__fixtures__/moonshotai-stream-malformed-tool-call.chunks.txt
@@ -1,0 +1,1 @@
+{"id":"chatcmpl-malformed","created":1785880003,"model":"kimi-k3","choices":[{"index":0,"delta":{"tool_calls":[{"index":"zero","id":"call-weather","function":{"name":"get_weather","arguments":"{}"}}]},"finish_reason":null}]}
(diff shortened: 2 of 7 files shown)


gr2m:
Closing as superseded by reviewed Factory-authored #19586, which fully addresses #19546 with captured-style SSE fixtures for index-less and explicit tool calls, both usage locations and precedence, malformed-index validation, and the required changeset.

PR #19627, opened 2026-08-25
feat(workflow): add durable video generation

TooTallNate:
## Summary

- add an experimental `@ai-sdk/workflow/video` video generation helper that starts a job with a durable Workflow webhook
- **suspend the workflow for the entire video render, consuming no compute until the webhook resumes it**
- run start and status operations in steps, using a stable step-derived idempotency key for billable starts
- return raw provider `url | base64 | binary` video data without downloading hosted videos across step boundaries
- add API docs, README usage, unit coverage, a Workflow suspend/resume integration test, and a patch changeset

## Example

```ts
import { experimental_generateVideo as generateVideo } from '@ai-sdk/workflow/video';

export async function videoWorkflow(prompt: string) {
  'use workflow';

  // The workflow suspends while the video renders, consuming no compute.
  const result = await generateVideo({
    model: 'klingai/kling-v3.0-t2v',
    prompt,
  });

  return result.videos;
}
```

## Suspension behavior

While `generateVideo()` waits for the provider to render the video, the workflow is durably suspended. No function remains running and no compute is consumed. The provider webhook resumes the workflow, which then makes one status request to retrieve the result.

### DIFF
--- a/.changeset/durable-video-webhooks.md
+++ b/.changeset/durable-video-webhooks.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/workflow': patch
+---
+
+Add durable video generation that waits for a Workflow webhook and returns provider video data without downloading hosted URLs.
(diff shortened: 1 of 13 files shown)


VaguelySerious (commented):
LGTM, will let repo owners approve

gr2m (approved):

PR #19643, opened 2026-08-25
feat(anthropic): preserve message batch metadata

brian-w-zhang:
## context

AI SDK already supports creating, polling, and reading Anthropic Message Batches. This PR is a follow-up to make that provider implementation complete enough for downstream integrations that need to preserve Anthropic's response semantics across the asynchronous lifecycle.

Before this change, the generic batch result kept the normalized status and text but discarded provider-specific lifecycle fields, native request counts, and some valid non-text result content. Once polling has completed, a downstream consumer cannot reconstruct those fields from the normalized result alone.

This is not a second batch API or a new orchestration layer. The generic AI SDK batch contract stays provider-neutral, while Anthropic-specific details are retained through `providerMetadata`. A downstream Gateway integration is the immediate consumer and will need a released version of these package changes before it can adopt them.

## summary

- preserve Anthropic Message Batch lifecycle metadata and native request counts in provider metadata
- preserve normalized content and usage for successful results, including citations, reasoning, client tool calls, and provider-executed tool results
- decode Anthropic tool calls and tool results from the batch results stream
- carry the supported language-model option surface into batch requests, with explicit validation for options that depend on request context that is unavailable during later polling
- keep MCP tool-result citation payloads permissive because they are opaque third-party data

The provider adapter remains responsible only for Anthropic request and response translation. Durable job storage, routing, billing, and compatibility endpoints remain outside this PR.

## testing

- `pnpm --filter @ai-sdk/anthropic test`
- `pnpm --filter @ai-sdk/anthropic type-check`
- `pnpm --filter @ai-sdk/provider type-check`
- `pnpm --filter @ai-sdk/anthropic build`
- `pnpm check`
- full GitHub CI

### DIFF
--- a/packages/ai/src/batch/batch-types.ts
+++ b/packages/ai/src/batch/batch-types.ts
@@ -3,9 +3,9 @@ import type {
   Experimental_BatchV4StartResult as BatchV4StartResult,
   Experimental_BatchV4Status as BatchV4Status,
   Experimental_BatchLanguageModelV4 as BatchLanguageModelV4,
-  LanguageModelV4GenerateResult,
 } from '@ai-sdk/provider';
-import type { ProviderOptions } from '@ai-sdk/provider-utils';
+import type { ProviderOptions, ToolSet } from '@ai-sdk/provider-utils';
+import type { ContentPart } from '../generate-text/content-part';
 import type { LanguageModelCallOptions } from '../prompt/language-model-call-options';
 import type { Prompt } from '../prompt/prompt';
 import type {
@@ -109,7 +109,7 @@ export type BatchOperationOptions = {
  */
 export type TextBatchGenerationResult = {
   /** Ordered normalized content, including citations, sources, and tool data. */
-  readonly content: LanguageModelV4GenerateResult['content'];
+  readonly content: Array<ContentPart<ToolSet>>;
   readonly text: string;
   readonly finishReason: FinishReason;
   readonly rawFinishReason?: string;
--- a/packages/ai/src/batch/batch-types.ts
+++ b/packages/ai/src/batch/batch-types.ts
@@ -3,6 +3,7 @@ import type {
   Experimental_BatchV4StartResult as BatchV4StartResult,
   Experimental_BatchV4Status as BatchV4Status,
   Experimental_BatchLanguageModelV4 as BatchLanguageModelV4,
+  LanguageModelV4GenerateResult,
 } from '@ai-sdk/provider';
 import type { ProviderOptions } from '@ai-sdk/provider-utils';
 import type { LanguageModelCallOptions } from '../prompt/language-model-call-options';
@@ -107,6 +108,8 @@ export type BatchOperationOptions = {
  * A normalized result for a successful text batch item.
  */
 export type TextBatchGenerationResult = {
+  /** Ordered normalized content, including citations, sources, and tool data. */
+  readonly content: LanguageModelV4GenerateResult['content'];
   readonly text: string;
   readonly finishReason: FinishReason;
   readonly rawFinishReason?: string;
--- a/packages/ai/src/batch/batch.ts
+++ b/packages/ai/src/batch/batch.ts
@@ -303,22 +303,6 @@ function convertBatchItemResult(
     return item;
   }
 
-  const clientToolCall = item.result.content.find(
-    part => part.type === 'tool-call' && part.providerExecuted !== true,
-  );
-  if (clientToolCall != null) {
-    return {
-      id: item.id,
-      status: 'failed',
-      error: {
-        code: 'unsupported_content',
-        message:
-          'Text batch results cannot represent client-executed tool calls.',
-      },
-      providerMetadata: item.result.providerMetadata,
-    };
-  }
-
   return {
     id: item.id,
     status: 'succeeded',
@@ -330,6 +314,7 @@ function convertGenerateResult(
   result: LanguageModelV4GenerateResult,
 ): TextBatchGenerationResult {
   return {
+    content: result.content,
     text: result.content
       .filter(
         (part): part is Extract<typeof part, { type: 'text' }> =>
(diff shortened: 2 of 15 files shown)


packages/anthropic/src/anthropic-messages-batch.ts
```diff
@@ -513,51 +560,52 @@ async function convertAnthropicBatchResult(
...
-            code: 'unsupported_content',
-          },
-        };
-      }
-
       return {
         id: line.custom_id,
         status: 'succeeded',
-        result: convertAnthropicBatchResponse(response),
+        result: convertAnthropicBatchResponse(response, generateId),
```
kkdawkins, line 555: Before this conversion, the shared response schema still rejects valid Anthropic messages such as `citations: null` and newer citation/content variants. Validation then turns a provider `succeeded` item into `invalid_response` and drops its content and usage. Please update the schema or handle unknown blocks without failing the whole item.

brian-w-zhang: fixed in bec3d2cc20. the shared schema now accepts `citations: null`, the newer citation variants, and `container_upload`. the batch converter preserves the successful item and usage when that output-only block has no v4 representation. focused tests and the Anthropic package typecheck pass.

packages/ai/src/batch/batch.ts
```diff
@@ -303,6 +303,22 @@ function convertBatchItemResult(
     return item;
   }
 
+  const clientToolCall = item.result.content.find(
```
kkdawkins: Core still returns concatenated text only, so the citations, sources, reasoning, custom parts, and provider-tool context preserved by this PR are dropped. This client-tool branch also converts a success to failure and loses usage. Expose normalized content, or retain usage and metadata on unsupported-content results.

packages/anthropic/src/anthropic-api.ts
```diff
@@ -1103,7 +1141,11 @@ export const anthropicChunkSchema = lazySchema(() =>
             content: z.array(
               z.union([
                 z.string(),
-                z.object({ type: z.literal('text'), text: z.string() }),
+                z.object({
+                  type: z.literal('text'),
+                  text: z.string(),
+                  citations: z.array(z.json()).nullable().optional(),
```
kkdawkins: Streaming still requires `mcp_tool_result.content` to be an array, while non-streaming accepts string or array. The streaming citation delta schema also lacks the new citation variants added above. Reuse shared schemas so valid responses do not fail only in streaming mode.

packages/ai/src/batch/batch.ts
brian-w-zhang: fixed. normalized batch successes now preserve the full content array and usage, including client tool calls, while keeping text as a convenience field.

packages/anthropic/src/anthropic-api.ts
brian-w-zhang: fixed. streaming and non-streaming paths now share the MCP result and citation schemas, including the newer citation variants.

packages/anthropic/src/anthropic-api.ts
```diff
@@ -655,6 +661,66 @@ const anthropicToolCallCallerSchema = z.union([
...
+
+const anthropicMcpToolResultContentSchema = z.union([
+  z.string(),
+  z.array(
+    z.union([
+      z.string(),
+      z.object({
+        type: z.literal('text'),
+        text: z.string(),
+        citations: z.array(anthropicCitationSchema).nullable().optional(),
```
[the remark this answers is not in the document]
brian-w-zhang: fixed in `25f5298229`. MCP tool-result citations are permissive opaque JSON again, while normal Anthropic citation blocks stay on the shared typed schema. added coverage for an unknown future citation variant; the Anthropic tests and typecheck pass.

packages/ai/src/batch/batch-types.ts
```diff
@@ -107,6 +108,8 @@ export type BatchOperationOptions = {
  * A normalized result for a successful text batch item.
  */
 export type TextBatchGenerationResult = {
+  /** Ordered normalized content, including citations, sources, and tool data. */
+  readonly content: LanguageModelV4GenerateResult['content'];
```
n1ckoates: `content` is exposed as raw `LanguageModelV4GenerateResult['content']`. That leaks the provider-interface representation rather than matching `generateText()`. For example, stringified tool inputs and raw file/tool-result shapes. The PR should reuse the `asContent` conversion from `generateText` by extracting it into a shared util.

brian-w-zhang: fixed in `705009495a`. batch results now use the same extracted Core content converter as `generateText`, so tool inputs/results and generated files expose the public Core shape instead of provider-interface content. added batch coverage for the normalized tool call/result shape; focused tests and the `ai` typecheck pass.

packages/anthropic/src/anthropic-messages-batch.ts
```diff
@@ -513,51 +549,53 @@ async function convertAnthropicBatchResult(
...
```
kkdawkins, line 555: Known variants are handled now, but batch parsing still fails closed on unknown result or content discriminants. A new Anthropic variant can abort JSONL iteration or turn a succeeded item into `invalid_response`. Can batch parsing preserve or skip unknown content, and map unknown result lines to one failed item instead of losing later results?

n1ckoates (approved):

packages/anthropic/src/anthropic-messages-batch.ts
brian-w-zhang: fixed in `3f24503836`. result lines now parse the outer envelope independently, so an unknown result type becomes one `invalid_response` item and later JSONL lines still stream. succeeded messages recover per content block, skipping only unknown discriminants while still failing malformed known blocks. added regression coverage for both paths; the focused Anthropic batch suite, package typecheck, oxfmt, and oxlint pass.

PR #19645, opened 2026-08-25
fix(harness-opencode): support host tools in subagents

### LINKED ISSUE #19646: @ai-sdk/harness-opencode: subagent host tools fail relay authorization
## Description

OpenCode subagents created through the task tool cannot reliably use `HarnessAgent` host tools.

The OpenCode event subscription includes events for the parent and its child sessions, but `consumeEvents` currently discards every event whose session ID differs from the parent. Host-tool relay authorization is granted only after the bridge observes an exact model tool-call event. The child event is therefore discarded, while the child MCP shim still sends the relay request. After the authorization window expires, the relay returns:

```text
401
{"error":"unauthorized tool relay request"}
```

The same filtering drops child `permission.asked` / `permission.v2.asked` events. Built-in child tools configured as `ask` can consequently wait forever because the bridge never replies.

Expected behavior:

- Host tools called by task-linked child sessions execute normally.
- Permission requests are answered against the child session.
- Child output and lifecycle events do not leak into or settle the parent stream.
- Events from unrelated sessions remain unable to authorize relay calls.

## Reproduction

1. Create an OpenCode `HarnessAgent` with at least one host-provided tool.
2. Prompt the primary agent to launch an OpenCode subagent using the task tool.
3. Have that subagent call the host tool.
4. Observe the child `message.part.updated` / `session.next.tool.called` event on the OpenCode event subscription.
5. Observe that the bridge filters the event because its session ID differs from the parent.
6. The MCP relay request waits for authorization and returns 401.

A focused regression test and proposed fix are included in #19645. The patch derives the allowed child lineage from OpenCode's task-tool metadata rather than trusting arbitrary session parent claims.

## AI SDK Version

- Reproduced with `@ai-sdk/harness-opencode` 1.0.72 and OpenCode 1.18.3.
- Confirmed the parent-only event filter is still present in `@ai-sdk/harness-opencode` 1.0.88 / current `main`.

## Code of Conduct

riltonfranzonee:
## Background

The OpenCode event subscription includes events from both the parent session and sessions created by its task tool. The bridge currently discards every event whose session ID differs from the parent. As a result, task-linked subagents cannot use host tools: their model-authorized tool event is discarded, so the short-lived relay authorization is never registered and the MCP call returns `401 unauthorized tool relay request`. Permission requests from those children are discarded as well, which can leave built-in tools such as `webfetch` waiting indefinitely.

## Summary

- Track child sessions from OpenCode's task-tool metadata, validating each child against the session that launched it.
- Process descendant events with isolated translation state so host tool calls are authorized without mixing child output or lifecycle events into the parent stream.
- Reply to descendant permission requests using the descendant session ID.
- Continue ignoring unrelated sessions.
- Add a regression test covering linked-child authorization, child permissions, and unrelated-session isolation.

## Checklist

## Related Issues

Fixes #19646.

### DIFF
--- a/packages/harness-opencode/src/bridge/index.ts
+++ b/packages/harness-opencode/src/bridge/index.ts
@@ -878,6 +878,22 @@ async function consumeEvents({
   const stream = await subscribeLegacyEvents({ client, signal });
   onSubscribed?.();
   if (!stream) return;
+  const taskSessionIds = new Set([sessionId]);
+  const registerSubagentSession = (sourceSessionId: string) =>
+    function register({
+      parentSessionId,
+      sessionId: subagentSessionId,
+    }: {
+      parentSessionId: string;
+      sessionId: string;
+    }) {
+      if (
+        parentSessionId === sourceSessionId &&
+        taskSessionIds.has(sourceSessionId)
+      ) {
+        taskSessionIds.add(subagentSessionId);
+      }
+    };
   const emitStreamEvent = createEmitStreamEvent({
     state,
     emit,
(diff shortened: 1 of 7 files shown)


ArshVermaGit (commented):
Looks good overall. The changes are focused, tests are included, and the implementation matches the intended subagent host-tool authorization flow.

packages/harness-opencode/src/bridge/index.ts
```diff
@@ -887,20 +903,59 @@ async function consumeEvents({
...
-    if (!event || (eventSessionId && eventSessionId !== sessionId)) continue;
+    if (!event) continue;
+    const scopedSessionId =
+      !eventSessionId || eventSessionId === sessionId
+        ? sessionId
+        : taskSessionIds.has(eventSessionId)
+          ? eventSessionId
+          : undefined;
+    if (!scopedSessionId) continue;
+    const isDescendant = scopedSessionId !== sessionId;
```
felixarntz, line 954: See the agent feedback in https://github.com/vercel/ai/pull/19645#issuecomment-5426744608: OpenCode events may contain both parent and child session IDs, but `getOpenCodeEventSessionId` doesn't consider this.

While the function isn't new, so far it didn't need to consider that because obviously we didn't support host tools in subagents yet. But now that this PR adds support for it, that function needs to be expanded, or alternatively a separate function needs to be implemented that handles both parent and child session IDs in the right context.

felixarntz (changes requested):
@riltonfranzonee Thank you for the PR!

There's one problem with the implementation, see below.

riltonfranzonee:
Addressed the isolation gap in 6bd1bc49b1ef6ec0fa82ae27c9247e3fae15f89f.

- `getOpenCodeEventSessionId` now reads `properties.info.sessionID` for message events and `properties.info.id` for session created/updated/deleted events.
- Added helper coverage for all four nested shapes.
- Extended the bridge regression test with a structured child `message.updated` event and child `session.updated` event; it verifies the child cannot emit output or settle the parent while the parent response still completes normally.

Verification: 74 package tests, package type-check, full `pnpm check`, and `git diff --check` all pass.

packages/harness-opencode/src/bridge/index.ts
riltonfranzonee: Fixed in 6bd1bc49b1ef6ec0fa82ae27c9247e3fae15f89f. The session-id extractor now recognizes `properties.info.sessionID` on message events and `properties.info.id` on session lifecycle events. The integration regression sends structured child output plus a child session update and verifies neither reaches parent emission or settlement; helper tests cover all four nested event shapes.

felixarntz (approved):
@riltonfranzonee Thank you, LGTM!

PR #19685, opened 2026-08-26
feat(bytedance): add chat language model support

chyroc:
## Background

`packages/bytedance` already ships config, image and video models, but no chat
model — so the provider exists while its main use case doesn't. Ark is ByteDance's
model platform and speaks an OpenAI-compatible API, which is why this can build on
`@ai-sdk/openai-compatible` rather than hand-rolling a transport.

## Summary

Adds the chat language model to `packages/bytedance`, wired through
`@ai-sdk/openai-compatible`, plus tests, docs and examples.

Also fixes the branch's lockfile: it had been generated in a way that dropped the
entire `snapshots:` section (13,923 lines against main's 51,854), which produced a
misleading `+5173/-3438` diff on `pnpm-lock.yaml` and a merge conflict. It's now
main's lockfile plus the single `@ai-sdk/openai-compatible` importer entry, so that
file is `+3/-0` and the PR total went from `+6267/-3693` to `+1097/-255`.

`packages/bytedance/package.json` keeps main's `2.0.35` version rather than the
`2.0.33` my branch had, so the release bump isn't reverted.

## Contributor Credit

@chyroc

## End-to-End Verification

Against the live Ark endpoint with a real key, not mocks:

- Chat completion returns normally for the versioned Doubao model ids.
- Tool calling returns the standard `tool_calls` shape.
- Vision: an image is genuinely ingested — `prompt_tokens` rises from ~40 to
  ~1340 and the model names the colour correctly for two different colours
  (a single colour can be guessed, and this endpoint can return 200 while
  silently dropping an image).
- Reasoning: `reasoning_content` is populated, and streaming emits
  `reasoning_content` deltas.

Note on model ids: Ark only resolves **fully versioned** ids
(`doubao-seed-2-1-pro-260628`). The short forms shown in the console
(`doubao-seed-2.1-pro`) return `InvalidEndpointOrModel.NotFound`.

## Checklist

**On the signing box — flagging this rather than quietly leaving it unticked:**
the commit on this branch is unsigned (`verified: false`, reason `unsigned`). It
was created through the GitHub API, which doesn't sign, and I don't have a signing
key set up for this account. Every recently-merged PR here shows
`verified: true`, so I take this as a real blocker rather than a formality. I'll
re-push signed commits once that's sorted; if you'd prefer, I can close and
reopen from a signed branch.

## Future Work

The `bytedance` package could also expose the BytePlus endpoint
(`https://ark.ap-southeast.bytepluses.com/api/v3`) as a documented option — it
serves the same models outside mainland China on a separate account, and this
package already defaults to that base URL.

## Related Issues

None open for this specific gap.

### DIFF
--- a/.changeset/bytedance-chat-model.md
+++ b/.changeset/bytedance-chat-model.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/bytedance': patch
+---
+
+Add chat (language) model support for ByteDance Ark
--- a/content/providers/01-ai-sdk-providers/85-bytedance.mdx
+++ b/content/providers/01-ai-sdk-providers/85-bytedance.mdx
@@ -1,11 +1,11 @@
 ---
 title: ByteDance
-description: Learn how to use ByteDance Seedance video and Seedream image models with the AI SDK.
+description: Learn how to use ByteDance Doubao language models, Seedance video models, and Seedream image models with the AI SDK.
 ---
 
 # ByteDance Provider
 
-The [ByteDance](https://www.bytedance.com/) provider contains support for the Seedance family of video generation models and the Seedream family of image generation models through the [BytePlus ModelArk](https://docs.byteplus.com/en/docs/ModelArk/) platform. Seedance provides high-quality text-to-video and image-to-video generation capabilities, including audio-video synchronization, first-and-last frame control, and multi-reference image generation (see the [video generation API](https://docs.byteplus.com/en/docs/ModelArk/Video_Generation_API)). Seedream provides text-to-image and image-to-image generation, including multi-image blending and batch image generation (see the [image generation API](https://docs.byteplus.com/en/docs/ModelArk/1824690)).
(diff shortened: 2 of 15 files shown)


chyroc:
Heads up that CI here is sitting at `action_required` — first-time fork contributions need a maintainer to approve the workflow run, so `CI` and `Verify Changesets` haven't executed. Whenever someone has a moment to kick those off.

For what it's worth, locally: `pnpm test:node` and `test:edge` are 103 passed each, `type-check` and `oxlint`/`oxfmt` clean.

(The `Vercel — failure` status is the deploy-preview authorization, not the build; #17997 had the same and merged fine.)

chyroc:
Rebased onto `main` — this was showing a merge conflict.

The cause was the lockfile in this branch, not the source changes. It had been
generated in a way that dropped the entire `snapshots:` section (13,923 lines vs
main's 51,854), which is what produced the misleading `+5173/-3438` diff on
`pnpm-lock.yaml` and the conflict.

I rebuilt it as main's lockfile plus only the one `@ai-sdk/openai-compatible`
workspace entry the new dependency needs, so that file is now `+3/-0`. Total diff
went from +6267/-3693 to +1097/-255. I also kept main's `2.0.35` version bump in
`packages/bytedance/package.json` rather than carrying my older `2.0.33`.

No functional change to the provider code itself.

PR #19764, opened 2026-08-26
fix(release): handle package-less examples on v5

gr2m:
The v5 release workflow currently fails during ci:version because examples/ai-functions remains as a directory on release-v5.0 but no longer has a package.json.

This change skips example directories without a package manifest. It also adds the cleanup script to the release workflow path filter so merging this fix immediately retries release-PR preparation.

Failed runs:
- https://github.com/vercel/ai/actions/runs/32997268656
- https://github.com/vercel/ai/actions/runs/32997504953
- https://github.com/vercel/ai/actions/runs/32997959179
- https://github.com/vercel/ai/actions/runs/33003082064

Verification:
- node syntax check
- isolated fixture with a package-less examples/ai-functions directory
- git diff --check

### DIFF
--- a/.github/scripts/cleanup-examples-changesets.mjs
+++ b/.github/scripts/cleanup-examples-changesets.mjs
@@ -7,6 +7,7 @@
  */
 
 import {
+  existsSync,
   readFileSync,
   writeFileSync,
   unlinkSync,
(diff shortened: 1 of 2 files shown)


gr2m:
removed the folder

PR #19796, opened 2026-08-26
fix(xai): preserve web_search action (query, sources, open_page) in responses tool results

R-Taneja:
## Background

Reported by AthenaHQ ([customer thread](https://vercel.slack.com/archives/C09Q2N54LJC/p1787771675800939)): calling xAI's `web_search` tool through the AI Gateway Responses API returns `web_search_call` items with an empty `action` — `query` is `''`, no `sources`, no `open_page` items. Direct to `api.x.ai/v1/responses` the same request returns everything (8 search calls, 35 source URLs, 5 opened pages in my repro).

Root cause: the xAI responses language model drops `part.action` for `web_search_call` output items. `doGenerate` emitted only a `tool-call` with no `tool-result`, and `doStream` emitted `tool-result` with `result: {}`. Downstream consumers (the gateway's Responses normalization) then fall back to `{type: 'search', query: ''}`.

## Summary

- Emit a `tool-result` for `web_search_call` parts in both `doGenerate` and `doStream`, carrying the mapped action via a new `mapWebSearchAction` helper that mirrors `@ai-sdk/openai`'s `mapWebSearchOutput`: `search` (query/queries + sources), `open_page` → `openPage`, `find_in_page` → `findInPage`
- Update the `webSearch` tool output schema to the Responses-style shape (`action` + `sources`) so the tool output is typed
- Tests for search/open_page/find_in_page action mapping in generate and stream

## Verification

- `vitest run` in packages/xai: 97/97 pass
- End-to-end through a local AI Gateway with this build patched in as `@ai-sdk/xai-v3` plus the companion gateway change: 9 web_search_call items, 40 source URLs, 5 open_page actions (was: 8 items, all empty)

Companion gateway PR forwards `sources` in the Responses output mapper. Linear: [AIG-670](https://linear.app/vercel/issue/AIG-670/xai-web-search-call-items-lose-query-sources-and-open-page-actions)

The customer-facing gateway path runs the v3 (release-v6.0) line today, so this should be backported.

### DIFF
--- a/.changeset/xai-web-search-action.md
+++ b/.changeset/xai-web-search-action.md
@@ -1,5 +1,5 @@
 ---
-'@ai-sdk/xai': minor
+'@ai-sdk/xai': patch
 ---
 
 fix(xai): preserve web_search action (query, sources, open_page) in responses tool results
--- a/.changeset/xai-web-search-action.md
+++ b/.changeset/xai-web-search-action.md
@@ -1,5 +1,5 @@
 ---
-'@ai-sdk/xai': patch
+'@ai-sdk/xai': minor
 ---
 
 fix(xai): preserve web_search action (query, sources, open_page) in responses tool results
--- a/.changeset/xai-web-search-action.md
+++ b/.changeset/xai-web-search-action.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/xai': patch
+---
+
+fix(xai): preserve web_search action (query, sources, open_page) in responses tool results
--- a/packages/xai/src/responses/xai-responses-language-model.ts
+++ b/packages/xai/src/responses/xai-responses-language-model.ts
@@ -25,7 +25,7 @@ import {
   type InferSchema,
   type ParseResult,
 } from '@ai-sdk/provider-utils';
-import { z } from 'zod/v4';
+import type { z } from 'zod/v4';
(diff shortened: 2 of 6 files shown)


packages/xai/src/responses/xai-responses-language-model.ts
```diff
@@ -1357,3 +1371,61 @@ export class XaiResponsesLanguageModel implements LanguageModelV4 {
     };
   }
 }
+
+const webSearchWireSourceSchema = z.object({
```
shaper: do we need to use lazySchema? see for example deepseek schema defs in

`packages/deepseek/src/chat/deepseek-chat-api-types.ts`

and maybe defining these in an `api-types` file is new pattern, or only done if shared, worth checking

R-Taneja: fixed

.changeset/xai-web-search-action.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/xai': minor
```
shaper: i think this can be patch, or does the tool edit require minor here? cc @aayush-kapoor

shaper (approved):

felixarntz (approved):

PR #19801, opened 2026-08-26
fix(harness-opencode): disable questions in headless sessions

### LINKED ISSUE #19800: harness-opencode: interactive question tool can hang headless sessions
### Description

`@ai-sdk/harness-opencode` runs OpenCode through a noninteractive bridge, but its generated OpenCode config does not disable the native `question` tool. OpenCode can therefore advertise `question` to the model even though the harness does not expose or answer that tool.

If the model calls it, the harness reports `Model tried to call unavailable tool 'question'` while OpenCode waits for an interactive reply that can never arrive. The turn then remains live until an outer timeout terminates it.

This is the same underlying headless-mode failure described in anomalyco/opencode#11899, but it can still occur through the adapter-managed OpenCode server.

### Reproduction

1. Start a prompt through `@ai-sdk/harness-opencode`.
2. Have the model request clarification using OpenCode's native `question` tool.
3. Observe the unavailable-tool error and that the OpenCode turn does not settle on its own.

The adapter config currently sets permissions for the other native tools but leaves `question` unspecified.

### Expected behavior

Because harness sessions are noninteractive, the adapter should deny `question` so OpenCode does not offer it to the model. The model can then continue using supported tools or respond in text instead of entering an unserviceable wait state.

### Versions

- `@ai-sdk/harness-opencode`: 1.0.91
- `@opencode-ai/sdk`: 1.18.3

riltonfranzonee:
## Background

OpenCode exposes a native `question` tool for interactive clarification. The OpenCode harness bridge is headless and cannot answer that tool, but its server configuration currently leaves the tool enabled. If a model calls it, AI SDK reports an unavailable tool while OpenCode waits indefinitely for a reply.

## Changes

- deny OpenCode's interactive `question` permission in harness-managed sessions
- add a regression test for the server configuration
- add a patch changeset

## Testing

- `pnpm turbo build --filter=@ai-sdk/harness-opencode...`
- `pnpm test` in `packages/harness-opencode` (75 tests)
- `pnpm check`
- `pnpm type-check:full`

Fixes #19800

### DIFF
--- a/.changeset/tidy-pandas-listen.md
+++ b/.changeset/tidy-pandas-listen.md
@@ -2,4 +2,4 @@
 '@ai-sdk/harness-opencode': patch
 ---
 
-Prevent headless OpenCode sessions from hanging when a model tries to use the interactive question tool.
+fix(harness-opencode): prevent headless OpenCode sessions from hanging when a model tries to use the interactive question tool
--- a/.changeset/tidy-pandas-listen.md
+++ b/.changeset/tidy-pandas-listen.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-opencode': patch
+---
+
+Prevent headless OpenCode sessions from hanging when a model tries to use the interactive question tool.
--- a/packages/harness-opencode/src/bridge/index.test.ts
+++ b/packages/harness-opencode/src/bridge/index.test.ts
@@ -10,6 +10,12 @@ const sdkMock = vi.hoisted(() => ({
 }));
 
 const permissionReplyMock = vi.hoisted(() => vi.fn());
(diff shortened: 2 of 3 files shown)


.changeset/tidy-pandas-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'@ai-sdk/harness-opencode': patch
+---
+
+Prevent headless OpenCode sessions from hanging when a model tries to use the interactive question tool.
```
felixarntz:
```suggestion
fix(harness-opencode): prevent headless OpenCode sessions from hanging when a model tries to use the interactive question tool
```

felixarntz (approved):
@riltonfranzonee Thank you, LGTM!

PR #19839, opened 2026-08-27
fix(azure): treat Foundry and Cognitive Services hosts as Azure OpenAI

### LINKED ISSUE #19817: provider/azure: Foundry and cognitiveservices hostnames omit the /openai/v1 path segment → 404 Resource not found
### Description

`@ai-sdk/azure` decides whether to insert the `/v1` path segment by sniffing the hostname:

```ts
function isAzureOpenAIBaseURL(baseURL: string | undefined) {
  return (
    baseURL == null || new URL(baseURL).hostname.endsWith('.openai.azure.com')
  );
}
```

Anything that isn't `.openai.azure.com` takes the "custom gateway" branch and gets a bare `${baseURL}${path}` — no `/v1`, no `api-version`:

```ts
} else if (!useAzureOpenAIEndpoint) {
  // Custom gateways can own Azure routing and versioning themselves.
  fullUrl = new URL(`${baseUrlPrefix}${path}`);
} else {
  fullUrl = new URL(`${baseUrlPrefix}/v1${path}`);
}
```

But `*.services.ai.azure.com` and `*.cognitiveservices.azure.com` are **first-party Azure hostnames for the same resource**, not customer-owned gateways, and they serve the identical v1 API. Microsoft documents both explicitly:

> `base_url` accepts both `https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1/` and `https://YOUR-RESOURCE-NAME.services.ai.azure.com/openai/v1/` formats.

— [Azure OpenAI v1 API](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle)

The result is that a connection configured with the Foundry hostname resolves to `/openai/responses` instead of `/openai/v1/responses` and **every request 404s** with `AI_APICallError: Resource not found`.

This matters because the Azure portal surfaces the `services.ai.azure.com` hostname, so it is a name users are very likely to configure. In our case it took down every Azure-backed model in three environments, and the endpoint itself was fine — `curl` against the same host works given the right path, which made it look like a credentials or deployment problem rather than a routing one.

### Reproduction

No Azure credentials needed — intercept `fetch` and read the URL the provider builds:

```ts
import { createAzure } from '@ai-sdk/azure'

async function urlFor(baseURL: string) {
  let seen = ''
  const azure = createAzure({
    apiKey: 'test',
    baseURL,
    fetch: (input) => {
      seen = input instanceof Request ? input.url : String(input)
      return Promise.resolve(new Response('{}', { status: 500 }))
    },
  })
  await azure
    .responses('my-deployment')
    .doStream({ prompt: [{ role: 'user', content: [{ type: 'text', text: 'hi' }] }] })
    .catch(() => {})
  return seen
}

for (const host of [
  'https://myresource.openai.azure.com',
  'https://myresource.services.ai.azure.com',
  'https://myresource.cognitiveservices.azure.com',
]) {
  console.log(new URL(host).hostname, '→', await urlFor(`${host}/openai`))
}
```

Output:

| baseURL host | URL built | |
| --- | --- | --- |
| `myresource.openai.azure.com` | `/openai/v1/responses?api-version=v1` | works |
| `myresource.services.ai.azure.com` | `/openai/responses` | **404** |
| `myresource.cognitiveservices.azure.com` | `/openai/responses` | **404** |

### Expected

All three hostnames address the same Azure resource and the same v1 API, so all three should resolve to `/openai/v1/responses`.

### Suggested fix

Widen the check to the known first-party suffixes (`.services.ai.azure.com`, `.cognitiveservices.azure.com`) so they take the same branch as `.openai.azure.com`. Genuine third-party gateways keep today's pass-through behaviour.

Note that `api-version` is not required alongside `/v1` — per the doc above, *"`api-version` is no longer a required parameter with the v1 GA API"* — so `origin + /openai/v1 + path` alone is sufficient for the Foundry hosts.

### Possibly related

#14009 reports the inverse symptom on a Foundry **project** endpoint (`/api/projects/<project>/openai`), where `api-version` *was* appended and broke multi-turn tool calls. If the hostname check was introduced to address that, it may have traded one bug for this one — the project-path case stopped getting an unwanted `api-version`, and plain Foundry resource hosts stopped getting the `/v1` they need. That issue is currently labelled `reproduction needed`; the `fetch`-interception snippet above reproduces URL construction without credentials and may help there too.

### AI SDK Version

Present in both current lines:

- `@ai-sdk/azure@4.0.52` (`latest`)
- `@ai-sdk/azure@3.0.110` (`ai-v6`)

Verified against the published `src/azure-openai-provider.ts` in both.

### Code of Conduct

ShobhanKarthish:
## Summary

- Treat `*.services.ai.azure.com` (Foundry) and `*.cognitiveservices.azure.com` as Azure OpenAI hosts
- Those already serve the v1 API, so they now get `/v1` and `api-version` like `.openai.azure.com`
- Custom gateways still skip `/v1`

Closes #19817

## Test plan

- Added unit tests for Foundry and Cognitive Services `baseURL` assembly
- Existing custom-gateway test still expects the path as-is


gr2m:
Superseded by #19969, which preserves the original feature commit and adds the signed follow-up fixes for complete OpenAI v1 and Foundry project base URLs. Thank you, @ShobhanKarthish!

Issue #18452, opened 2026-08-05
Empty assistant message sent to provider after unsigned reasoning is dropped on replay -> Bedrock 400, bricks conversation


yonatan-shorani:
## Description

When an assistant turn is aborted **mid-thinking**, the persisted reasoning part has no provider signature (`providerMetadata` is absent). On replay, the Amazon Bedrock adapter intentionally drops unsigned reasoning ("Unsigned reasoning is intentionally not replayed", #17602) — **but still emits the assistant message it belonged to**. If the aborted step held *only* reasoning, the resulting request carries an assistant message with `content: []`, and Bedrock rejects it, non-retryable:

```
The content field in the Message object at messages.N is empty.
```

Because the 400 is non-retryable, no failover fires and the poisoned reasoning part stays in history — so **every subsequent turn replays it and the conversation is bricked permanently**.

Core's `convert-to-language-model-prompt` filters empty **tool** messages ("prevents sending invalid empty messages to the provider") but has no assistant-message equivalent, so nothing catches this.

Verified still reproducing on `ai@7.0.52` + `@ai-sdk/amazon-bedrock@5.0.43`.

## Reproduction (synthetic, self-contained)

```ts
import { convertToModelMessages, generateText, type UIMessage } from 'ai';
import { createAmazonBedrock } from '@ai-sdk/amazon-bedrock';

const messages: UIMessage[] = [
  { id: 'u1', role: 'user', parts: [{ type: 'text', text: 'think hard then answer' }] },
  {
    id: 'a1',
    role: 'assistant',
    // partial reasoning persisted after an abort — NO providerMetadata/signature
    parts: [{ type: 'reasoning', text: 'Let me consider the options' } as any],
  },
  { id: 'u2', role: 'user', parts: [{ type: 'text', text: 'hello?' }] },
];

class Captured extends Error {}

const main = async () => {
  const model = await convertToModelMessages(messages);
  const bedrock = createAmazonBedrock({
    region: 'us-east-1', accessKeyId: 'x', secretAccessKey: 'y',
    fetch: (async (_url: any, init: any) => {
      const body = JSON.parse(init.body);
      const idx = body.messages?.findIndex(
        (m: any) => Array.isArray(m.content) && m.content.length === 0);
      throw new Captured(idx >= 0
        ? `request messages[${idx}] has content: [] (unsigned reasoning dropped, message kept)`
        : 'no empty-content message in request');
    }) as any,
  });
  try {
[...]

PR #19852, opened 2026-08-27
fix(provider/amazon-bedrock): omit assistant messages whose only remaining content is a cache point

### LINKED ISSUE #19851: Bedrock: assistant message with only a cachePoint still sent after unsigned reasoning is filtered → 400 "There is nothing available to cache"
### Description

#18456 fixed assistant messages going out empty after unsigned reasoning is dropped — but there's a sibling case that still bricks conversations.

If the filtered-out message carries a `cachePoint` provider option, the new guard sees `[{ cachePoint }]`, length 1, and sends it anyway. Bedrock rejects it, non-retryable:

> There is nothing available to cache. Please remove the invalid cache point and try again.

Same story as #18452 from there... the poisoned turn stays in history and every following request fails.

This happens in the wild: opencode stamps cache points on the last two messages, so any assistant turn aborted mid-thinking hits this (anomalyco/opencode#11210 has both error strings).

### Reproduction

Same as #18452's repro, just add the provider option to the assistant message:

```ts
{
  role: 'assistant',
  content: [{ type: 'reasoning', text: 'Let me consider the options' }], // aborted turn, no signature
  providerOptions: { bedrock: { cachePoint: { type: 'default' } } },
},
```

The outgoing request contains an assistant message with `content: [{"cachePoint":{"type":"default"}}]`. I sent that exact payload to the live Converse API and it 400s with the message above.

### Expected behavior

A message whose only remaining content is a cache point should be omitted, same as the empty case #18456 already handles. Dropping it is safe — I also checked that Converse accepts consecutive user messages, which is what's left behind.

Happy to PR, it's a one-line change to the #18456 guard.

### AI SDK Version

- `@ai-sdk/amazon-bedrock` current main (also verified on 5.0.43+)

nabilfreeman:
## Background

#18456 fixed assistant messages going out empty after unsigned reasoning is filtered on replay. There's a sibling case it misses: the guard runs after `pushCachePoint`, so a fully-filtered message that carries a `cachePoint` provider option still goes out as `content: [{ cachePoint }]`. Bedrock rejects it, non-retryable:

> There is nothing available to cache. Please remove the invalid cache point and try again.

Same bricked-conversation story as #18452. It bites real users today via opencode, which puts cache points on the last two messages of every request — an assistant turn aborted mid-thinking then poisons the whole session (anomalyco/opencode#11210 has both error strings).

## Summary

Strengthened the omission guard from "non-empty" to "has at least one non-cachePoint block":

```ts
if (amazonBedrockContent.some(block => !('cachePoint' in block))) {
  messages.push({ role: 'assistant', content: amazonBedrockContent });
}
```

Nothing changes for messages that keep real content — the cache point is still emitted after it. Added regression tests (message-level and part-level cache point omission, both fail without the fix, plus one confirming the cache point is kept when other content remains) and a patch changeset. 499/499 package tests pass on node and edge.

## Contributor Credit

@yonatan-shorani — this builds directly on the #18452 reproduction and the #18456 fix.

## End-to-End Verification

Sent the exact payloads to the live Converse API (eu-west-1, Claude): an assistant message whose content is only `[{"cachePoint":{"type":"default"}}]` returns the 400 above, and consecutive user messages — the conversation shape left behind after omission — are accepted, so dropping the message can't produce an invalid request. Applying the same omission downstream in opencode un-bricks a real poisoned session: the exact 400 before, normal completion after.

## Checklist

## Related Issues

Fixes #19851

### DIFF
--- a/.changeset/cachepoint-only-assistant-messages.md
+++ b/.changeset/cachepoint-only-assistant-messages.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/amazon-bedrock': patch
+---
+
+Omit assistant messages whose only remaining content is a cache point after filtering.
--- a/packages/amazon-bedrock/src/convert-to-amazon-bedrock-chat-messages.test.ts
+++ b/packages/amazon-bedrock/src/convert-to-amazon-bedrock-chat-messages.test.ts
@@ -1429,6 +1429,123 @@ describe('assistant messages', () => {
     });
   });
 
+  it('should omit an assistant message whose only remaining content is a message-level cache point', async () => {
+    // A cachePoint provider option on a message whose every content part was
+    // filtered out (unsigned reasoning) must not produce
+    // content: [{ cachePoint }] — Bedrock rejects it with "There is nothing
+    // available to cache. Please remove the invalid cache point and try again."
+    const result = await convertToAmazonBedrockChatMessages([
+      {
+        role: 'user',
+        content: [{ type: 'text', text: 'Think hard then answer' }],
+      },
+      {
+        role: 'assistant',
+        content: [
+          {
+            type: 'reasoning',
+            text: 'Let me consider the options',
+          },
+        ],
+        providerOptions: { bedrock: { cachePoint: { type: 'default' } } },
+      },
+      {
+        role: 'user',
+        content: [{ type: 'text', text: 'Hello?' }],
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by #19858, which already closed #19851.

PR #19903, opened 2026-08-28
fix(harness-deepagents): retain conversation context after stop/resume

### LINKED ISSUE #19693: Deep Agents harness loses conversation context after a stopped session is resumed
The Deep Agents harness does not preserve conversation context when a session is fully stopped between turns and later resumed. Resuming succeeds operationally, but the new Deep Agents runtime does not remember messages from before the stop.

This is specific to `stop()` followed by resume. Detaching and reattaching a session works because the existing bridge remains alive.

### `resume.ts`

Run the Deep Agents `examples/ai-functions/src/harness-agent/deepagents/resume.ts` example:

```bash
./tools/run-harness-agent-examples.sh --harness deepagents --example resume
```

The example fails with:

```text
Second turn did not retain context from previous turn
```

### Interactive Basic (with stop) example

Start `examples/harness-e2e-next` and open the Deep Agents **Basic (with stop)** example at `/harness/deepagents/basic-with-stop`.

1. Tell the agent, `My name is Felix. Remember it.`
2. Wait for the turn to finish. The route calls `session.stop()` and persists its resume state.
3. In the next turn, ask, `What is my name?`

The resumed agent does not remember the name (it may apply the client-side tool for getting the user name, but that shows it doesn't remember the name from before).

The regular Deep Agents **Basic** example, which calls `session.detach()` between turns, does remember it.

## Expected behavior

A session resumed from the state returned by `session.stop()` retains its earlier conversation, including when the stop and subsequent resume happen in separate host processes.

## Actual behavior

The sandbox and session can be resumed, but the conversation starts without the messages from before `session.stop()`.

## Likely cause

The Deep Agents bridge configures LangGraph with an in-memory `MemorySaver`. `doStop()` tears down that bridge, and its resume state does not contain a conversation checkpoint. The sandbox snapshot preserves workspace files, but it does not preserve the bridge process's in-memory conversation state.

By contrast, `doDetach()` leaves the bridge running and includes its connection details in the resume state, so a later process can reattach to the same bridge and retain the conversation.

## Acceptance criteria

- A Deep Agents session retains conversation context after `session.stop()` and a later resume.
- The behavior works when the original turn and resumed turn run in separate host processes.
- The Deep Agents `resume.ts` example passes.
- The interactive Deep Agents **Basic (with stop)** example remembers information supplied before the previous turn was stopped.
- Existing detach/reattach behavior continues to work.

lorenzozanee:
## Background

The Deep Agents harness lost conversation context after `session.stop()` because the bridge's in-memory `MemorySaver` was never persisted, while the sandbox snapshot only preserved workspace files.

## Summary

Persist the checkpointer to the bridge state directory and restore it on startup so a resumed session retains prior conversation.

## Contributor Credit

Issue reported by @felixarntz.

## End-to-End Verification

Verified with a standalone persistence script where a fresh checkpointer loses context and the restored one retains it; full sandbox resume is verified in CI.

## Checklist

## Future Work

No follow-up required.

## Related Issues

Fixes #19693

### DIFF
--- a/.changeset/deep-agents-checkpoint-fix.md
+++ b/.changeset/deep-agents-checkpoint-fix.md
@@ -1,0 +1,7 @@
+---
+"@ai-sdk/harness-deepagents": patch
+---
+
+fix(harness-deepagents): persist conversation checkpoint across session stop and resume
+
+Persist LangGraph MemorySaver checkpoints to the bridge state directory so a sandbox snapshot retains conversation history. New bridge processes reload the checkpoint file on startup, preserving context when a session stopped via session.stop() is resumed in a separate host process.
--- a/content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
+++ b/content/providers/02-ai-sdk-harnesses/05-deepagents.mdx
@@ -201,11 +201,6 @@ The adapter exposes these Deep Agents built-ins through `agent.tools`:
 
 ## Known Limitations
 
-- **Resuming a stopped session's conversation** is not supported — after
-  `session.stop()`, Deep Agents' in-memory conversation state (LangGraph
-  `MemorySaver`) is gone; only the sandbox workspace persists via its snapshot.
-  Use `session.detach()` for cross-process handoff or `session.suspendTurn()` for
-  turn continuation while keeping the live bridge running.
 - **Manual compaction** is not supported.
 
 ## Related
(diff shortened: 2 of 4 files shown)


gr2m:
Superseded by #19919, which already closed #19693.

PR #19926, opened 2026-08-28
fix(openai): preserve cache breakpoints on scalar tool results

### LINKED ISSUE #19921: @ai-sdk/openai Responses drops promptCacheBreakpoint on scalar tool results
## Description

The OpenAI Responses adapter silently drops `providerOptions.openai.promptCacheBreakpoint` from tool results whose output type is `text`, `json`, `error-text`, `error-json`, or `execution-denied`.

Both `LanguageModelV4ToolResultPart.providerOptions` and the scalar variants of `LanguageModelV4ToolResultOutput.providerOptions` are documented as provider-specific options that are passed through to the provider. However, `convertFunctionToolResultOutput` serializes scalar outputs directly to strings without reading either set of provider options. It reads cache breakpoints only from items inside an output with `type: "content"`.

This makes the following two semantically equivalent tool results behave differently:

- A `text` output with `promptCacheBreakpoint` silently produces a plain string without a breakpoint.
- A `content` output containing one text item with the same option produces an `input_text` item with `prompt_cache_breakpoint`.

The Chat Completions adapter already handles the scalar case. It reads the breakpoint from `output.providerOptions`, falls back to the tool-result part's `providerOptions`, and wraps the scalar value in a text content array when necessary. The Responses adapter should behave consistently.

### Expected behavior

Given this tool result:

```ts
{
  type: "tool-result",
  toolCallId: "call_1",
  toolName: "lookup",
  output: {
    type: "text",
    value: "stable tool output",
    providerOptions: {
      openai: {
        promptCacheBreakpoint: { mode: "explicit" },
      },
    },
  },
}
```

the Responses request should contain:

```json
{
  "type": "function_call_output",
  "call_id": "call_1",
  "output": [
    {
      "type": "input_text",
      "text": "stable tool output",
      "prompt_cache_breakpoint": { "mode": "explicit" }
    }
  ]
}
```

The same should apply when the breakpoint is supplied on the tool-result part rather than the scalar output, matching the Chat Completions fallback behavior.

### Actual behavior

The request contains a plain string, and the breakpoint is absent:

```json
{
  "type": "function_call_output",
  "call_id": "call_1",
  "output": "stable tool output"
}
```

No warning is emitted. In explicit caching mode this prevents the tool-result boundary from being written to the prompt cache.

### Source-level cause

On current `main`, `packages/openai/src/responses/convert-to-openai-responses-input.ts` handles scalar branches in `convertFunctionToolResultOutput` by immediately returning a string. `getPromptCacheBreakpoint` is called only in the `content` branch.

By comparison, `packages/openai/src/chat/convert-to-openai-chat-messages.ts` obtains the breakpoint from either the output or tool-result part and converts scalar output into a content array when a breakpoint is present.

This is also visible on current AI SDK `main` at commit `24fcc9277778d7ded2642bcab923d87f72fe0846`.

### Impact

Tool results are a high-value cache boundary in multi-step agents because every following model step replays them. Applications currently have to convert ordinary scalar tool outputs into synthetic `content` outputs solely to preserve the cache marker. That changes otherwise provider-independent message data to work around an adapter serialization detail.

### Related issues

- #15185 concerned `toModelOutput.providerOptions` propagation for Anthropic and was fixed in core/provider conversion. It does not cover OpenAI Responses scalar serialization.
- #17507 concerned provider options lost while combining consecutive tool messages. It does not cover this conversion path.

## Reproduction

The following script captures the request generated by the public API without making a real OpenAI request:

```ts
import { createOpenAI } from "@ai-sdk/openai"
import { generateText, type ModelMessage } from "ai"

let requestBody: unknown

const openai = createOpenAI({
  apiKey: "test",
    requestBody = JSON.parse(String(init?.body))
    return new Response(
      JSON.stringify({ error: { message: "stop after capturing request" } }),
      {
        status: 400,
        headers: { "content-type": "application/json" },
      },
    )
  },
})

  { role: "user", content: "Run the lookup" },
  {
    role: "assistant",
    content: [
      {
        type: "tool-call",
        toolCallId: "call_1",
        toolName: "lookup",
        input: {},
      },
    ],
  },
  {
    role: "tool",
    content: [
      {
        type: "tool-result",
        toolCallId: "call_1",
        toolName: "lookup",
        output: {
          type: "text",
          value: "stable tool output",
          providerOptions: {
            openai: {
              promptCacheBreakpoint: { mode: "explicit" },
            },
          },
        },
      },
    ],
  },
]

await generateText({
  model: openai.responses("gpt-5.6"),
  messages,
  providerOptions: {
    openai: {
      promptCacheOptions: { mode: "explicit", ttl: "30m" },
    },
  },
}).catch(() => undefined)

console.dir(requestBody, { depth: null })
```

Inspecting `input` shows that `function_call_output.output` is the plain string `"stable tool output"` and contains no `prompt_cache_breakpoint`.

As a control, replace the output above with:

```ts
output: {
  type: "content",
  value: [
    {
      type: "text",
      text: "stable tool output",
      providerOptions: {
        openai: {
          promptCacheBreakpoint: { mode: "explicit" },
        },
      },
    },
  ],
}
```

The captured request then correctly includes `prompt_cache_breakpoint` on the nested `input_text` item.

## AI SDK Version

- `ai`: `7.0.74`
- `@ai-sdk/openai`: `4.0.45`
- Also confirmed by inspection on current `main` at `24fcc9277778d7ded2642bcab923d87f72fe0846`

## Code of Conduct

alvarosevilla95:
## Summary

- preserve `promptCacheBreakpoint` from scalar tool-output provider options
- fall back to provider options on the enclosing tool-result part
- wrap marked scalar outputs as `input_text` while preserving existing scalar serialization and output-schema encoding
- apply part-level breakpoints to content outputs only when no nested content item already defines one

## Testing

The factory reproduction in #19922 covers all five scalar output types at both provider-option placements.

- `pnpm test` in `packages/openai` (1,944 tests across Node and edge)
- `pnpm type-check` in `packages/openai`
- `oxfmt` and `oxlint` on the changed files

Fixes #19921

### DIFF
--- a/.changeset/fix-openai-tool-cache-breakpoints.md
+++ b/.changeset/fix-openai-tool-cache-breakpoints.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+Preserve prompt cache breakpoints on scalar OpenAI Responses tool results.
--- a/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
+++ b/packages/openai/src/responses/convert-to-openai-responses-input.test.ts
@@ -2842,6 +2842,220 @@ describe('convertToOpenAIResponsesInput', () => {
       `);
     });
 
+    it('should add prompt cache breakpoints to scalar tool result outputs', async () => {
+      const providerOptions = {
+        openai: { promptCacheBreakpoint: { mode: 'explicit' as const } },
+      };
+      const result = await convertToOpenAIResponsesInput({
+        toolNameMapping: testToolNameMapping,
+        prompt: [
+          {
+            role: 'tool',
+            content: [
+              {
+                type: 'tool-result',
+                toolCallId: 'call_text',
+                toolName: 'search',
+                output: {
+                  type: 'text',
+                  value: 'result',
+                  providerOptions,
+                },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call_json',
+                toolName: 'search',
+                output: {
+                  type: 'json',
+                  value: { result: true },
+                  providerOptions,
+                },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call_error_text',
+                toolName: 'search',
+                output: {
+                  type: 'error-text',
+                  value: 'error',
+                  providerOptions,
+                },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call_error_json',
+                toolName: 'search',
+                output: {
+                  type: 'error-json',
+                  value: { error: true },
+                  providerOptions,
+                },
+              },
+              {
+                type: 'tool-result',
+                toolCallId: 'call_denied',
+                toolName: 'search',
+                output: {
+                  type: 'execution-denied',
+                  reason: 'denied',
+                  providerOptions,
+                },
+              },
+            ],
+          },
+        ],
+        systemMessageMode: 'system',
+        providerOptionsName: 'openai',
+        store: true,
+      });
+
+      const output = (text: string) => [
+        {
+          type: 'input_text',
+          text,
+          prompt_cache_breakpoint: { mode: 'explicit' },
+        },
+      ];
+      expect(result.input).toEqual([
+        {
+          type: 'function_call_output',
+          call_id: 'call_text',
+          output: output('result'),
+        },
+        {
+          type: 'function_call_output',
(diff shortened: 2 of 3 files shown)


alvarosevilla95:
Closing in favor of #19927, damn this factory is cool, noted for next time :D

PR #19959, opened 2026-08-28
feat(ai): expose individual image generation `calls`

### LINKED ISSUE #19938: Expose per-call results from `generateImage`
`generateImage` may make multiple underlying `doGenerate` calls when the requested image count exceeds a model’s `maxImagesPerCall`.

Today, it flattens those calls into one result. This creates ambiguity for fields that inherently belong to a particular call:

- `providerMetadata` can contain arbitrary provider-defined fields. Some are per-image, but others are per-call, such as Gateway routing/cost/generation ID or xAI cost.
- `response` only represents one call.
- `warnings` are flattened.
- `usage` is summed, which is appropriate as a convenience total but loses the per-call breakdown.

The current `providerMetadata[provider].images` convention is specifically image-level metadata, but it is separated from the corresponding generated image. It would be more useful and less error-prone for this metadata to live on each item in `result.images`.

## Proposal

Add a `calls` array to `generateImage` results:

```ts
const result = await generateImage({ ... });

result.calls;
```

Each entry represents exactly one `ImageModelV4#doGenerate` invocation:

```ts
type GenerateImageCall = {
  images: GeneratedImage[];
  providerMetadata?: ImageModelV4ProviderMetadata;
  response: ImageModelV4ResponseMetadata;
  warnings: SharedV4Warning[];
  usage?: ImageModelV4Usage;
};
```

Add optional per-image metadata to generated images:

```ts
type GeneratedImage = GeneratedFile & {
  providerMetadata?: Record<string, JSONObject>;
};
```

`image.providerMetadata[provider]` would contain the corresponding entry from the provider’s current `providerMetadata[provider].images` array. It must remain optional: some providers, including Gateway, do not provide per-image metadata.

Then:

- `result.calls[n].providerMetadata` retains the complete, unmodified metadata for that request.
- `result.images` remains the convenient flattened image list.
- `result.images[i].providerMetadata` exposes metadata specific to that image.
- `result.usage` can remain the aggregate convenience total, while `calls[n].usage` provides the precise breakdown.

## Migration

Keep the existing top-level `providerMetadata`, `response`, `warnings`, and aggregate `usage` for compatibility, but deprecate the ambiguous flattened fields in favor of `calls`.

In the next major version, remove the deprecated fields—or retain only the unambiguous aggregate convenience fields—and make `calls` the authoritative representation.

n1ckoates:
## Background

`generateImage()` can split one request into multiple provider calls, but its existing aggregate result does not preserve the association between an image and the call that produced it. This makes per-call response metadata, costs, warnings, and usage difficult to consume reliably. This change addresses #19938.

## Summary

- Add `result.calls`, preserving images, response metadata, provider metadata, warnings, and usage for each underlying provider call.
- Add optional `providerMetadata` to `GeneratedFile`
- Core projects legacy `providerMetadata.<provider>.images[index]` onto each generated image.
- Keep top-level aggregate `warnings` and `usage`, deprecate `responses` and merged top-level `providerMetadata`.

## End-to-End Verification

Ran a live AI Gateway image-generation request with `bfl/flux-kontext-pro`, requesting two images with `maxImagesPerCall: 1` to force two underlying provider calls.

The result contained two `calls`, each with one image, its response metadata, Black Forest Labs per-image metadata, and a Gateway cost of `$0.04`. Per-image metadata was available directly on each returned `GeneratedFile`; the aggregate Gateway cost was correctly `$0.08`.

## Checklist

## Future Work

- In AI SDK 8, update the provider result contract to return per-image metadata directly rather than packing it in `providerMetadata.<provider>.images`.
- Remove the deprecated `responses` and merged top-level `providerMetadata` fields.

## Related Issues

Fixes #19938


packages/ai/src/generate-image/generate-image.test.ts
```diff
@@ -678,6 +678,143 @@ describe('generateImage', () => {
...
+    ]);
+    expect(
+      result.calls.map(call => ({
+        images: call.images.map(image => image.base64),
+        providerMetadata: call.providerMetadata,
+        response: call.response,
+        warnings: call.warnings,
+        usage: call.usage,
+      })),
+    ).toStrictEqual([
```
gr2m: instead of `toStrictEqual` we usually prefer inline snapshots. But given the base64 data the snapshots might become too big, I'm okay either way. Snapshots are easier to update

gr2m (approved):

packages/ai/src/generate-image/generate-image.test.ts
n1ckoates: Resolved by https://github.com/vercel/ai/pull/19959/commits/98925419d294be53deb0b2e631b38a5fb99753d8

PR #19996, opened 2026-08-29
fix(openai): robust capability detection for gateway-prefixed models

### LINKED ISSUE #11828: max_tokens not supported check with Cloudflare AI gateway
### Description

The Cloudflare ai gateway provider generates the model name as `provider/{model}`, which fails this `isReasoningModel` check and throw an error by Open AI.

https://github.com/vercel/ai/blob/58fff3fa2d9a1f14ab26c853ba31b672bc21dac2/packages/openai/src/openai-chat-language-model.ts#L843

Error - 

Unsupported parameter: 'max_tokens' is not supported with this model. Use
  'max_completion_tokens' instead.

The isReasoningModel() function checks the startsWith only, which fails as the modal name are generated like : `provider/model`

Image

```
export function getAIGateway(
  provider: string,  
  model: string,
  apiKey: string,
  gatewayId: string,
) {
  const openAICompatible = createOpenAICompatible({
    name: `cf-{provider}`,
    baseURL: `${baseURL}/compat`
    apiKey: apiKey,
  });
  return openAICompatible(`{provider}/${model}`) as unknown as LanguageModel;
}

const model = getAIGateway(provider, model, apiKey, gatewayId);

const result = streamText({
      model: model,
      messages: messagesArray,
      maxOutputTokens: MAX_TOKEN,
      temperature: TEMP,
    }); 
```

### AI SDK Version

@ai-sdk/openai-compatible
"ai": "^5.0.113",

### Code of Conduct

wasim-builds:
### Description

Fixes #11828

When using AI gateways (such as Cloudflare AI Gateway), models are often prefixed with `provider/` (e.g., `openai/o1-mini`). This caused the strict `.startsWith('o1')` capability checks to fail, leading to `max_tokens` being incorrectly sent instead of `max_completion_tokens` for reasoning models.

This PR extracts the base model ID (stripping any `/` prefixes) before evaluating the OpenAI capability heuristics, ensuring correct parameter selection when passing reasoning models through gateways.

### Testing
- [x] Added unit tests for prefixed model names in `openai-language-model-capabilities.test.ts`
- [x] Verified tests pass locally

### DIFF
--- a/.changeset/fix-openai-gateway-prefixed-capability-detection.md
+++ b/.changeset/fix-openai-gateway-prefixed-capability-detection.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/openai': patch
+---
+
+fix(openai): robust capability detection for gateway-prefixed model IDs (e.g. cf-gateway/o1-mini)
--- a/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
+++ b/packages/openai-compatible/src/chat/openai-compatible-chat-language-model.test.ts
@@ -1414,6 +1414,25 @@ describe('doGenerate', () => {
       );
     });
 
+    it('should use max_completion_tokens for gateway-prefixed reasoning models', async () => {
+      prepareJsonResponse({ content: '{"value":"test"}' });
+
+      const model = new OpenAICompatibleChatLanguageModel('openai/o1-mini', {
+        provider: 'test-provider',
+        url: () => 'https://my.api.com/v1/chat/completions',
+        headers: () => ({}),
+      });
+
+      await model.doGenerate({
+        prompt: TEST_PROMPT,
+        maxOutputTokens: 1000,
+      });
+
+      const requestBody = await server.calls[0].requestBodyJson;
+      expect(requestBody.max_completion_tokens).toBe(1000);
+      expect(requestBody.max_tokens).toBeUndefined();
+    });
+
     it('should pass textVerbosity setting from providerOptions', async () => {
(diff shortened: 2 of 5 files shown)


taranggoyal70 (commented):
Thanks for the focused change—the base-model normalization itself is clear, and the patch includes a changeset and colocated tests. There is one blocking mismatch with the linked issue, though.

Issue #11828 constructs the model with `createOpenAICompatible` from `@ai-sdk/openai-compatible`. That package's chat request path still serializes `maxOutputTokens` unconditionally as `max_tokens` in `packages/openai-compatible/src/chat/openai-compatible-chat-language-model.ts`, and it never calls `getOpenAILanguageModelCapabilities`. This PR only changes `packages/openai`, so the exact Cloudflare `/compat` reproduction will continue to send the rejected parameter.

The new tests exercise only the OpenAI capability helper, which means they pass while the reported request still fails. Please add a regression test through `createOpenAICompatible` that captures the outgoing request for a prefixed reasoning model and asserts that `max_completion_tokens` is present while `max_tokens` is absent, then implement the fix on that path. If the intended solution is instead to use `createOpenAI` with a custom `baseURL`, the issue, PR scope, and changeset should make that migration explicit.

One secondary coverage concern: normalizing before the shared capability calculation changes system-message mode, flex/priority support, and non-reasoning-parameter behavior as well as the token parameter. If that broader `@ai-sdk/openai` behavior is intentional, tests should cover those affected capability fields for prefixed IDs too.

PR #20004, opened 2026-08-30
fix(minimax): support 480P and 768P video resolutions

ZYPX:
## Background

MiniMax supports model-specific video resolution tiers with exact casing:

- MiniMax-H3: `768P`, `2K`
- MiniMax-H3-Max: `480P`, `768P`

The provider only accepted `2K`, preventing supported lower-resolution requests.
The V2 API also gives the models different duration bounds and does not support
reference-to-video inputs on MiniMax-H3-Max.

## Summary

- Add `480P` and `768P` MiniMax video resolution options.
- Add MiniMax-H3-Max as a known video model.
- Apply model-specific resolution validation and defaults.
- Normalize named tiers to MiniMax's required uppercase casing.
- Apply the documented 4–15 second H3 and 5–15 second H3-Max duration bounds.
- Prevent unsupported reference inputs from being sent to MiniMax-H3-Max.
- Update tests, documentation, example, and patch changeset.

## End-to-End Verification

Not run against the live MiniMax API because API credentials were unavailable.

## Checklist

- [x] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #3564, opened 2024-11-08
chore (ai): AssistantResponse cleanups

lgrammel:

PR #20006, opened 2026-08-30
fix(minimax): map 480P and 768P frame sizes onto the named video resolution tiers

joshualipman123:
### Background

#20004 adds `480P` and `768P` as MiniMax video resolution tiers, but only wires up frame sizes for `2K`. The top-level `resolution` call option is typed `` `${number}x${number}` ``, so a type-safe caller can only reach a named tier through the frame-size map — which left `480P` and `768P` reachable exclusively via `providerOptions.minimax.resolution`.

Every other provider in the repo that takes named tiers maps a frame size for each tier it serves (xai `480p`/`720p`/`1080p`, alibaba `480P`/`720P`/`1080P`, bytedance, google), so this follows that precedent.

### Summary

- Add frame-size rows for the `480P` and `768P` tiers — one canonical frame size per aspect ratio the models accept, matching the shape of the existing `2K` rows.
- Tests for both new tiers across all six ratios, and for the "tier the model does not support" branch, which a top-level frame size can now reach (`854x480` on MiniMax-H3, `2560x1440` on MiniMax-H3-Max) — it previously had no coverage from that direction.
- Document how a top-level `resolution` resolves to a tier, what happens when it matches no tier or an unsupported one, and that provider options win.

#### Related Issues

Stacked on #20004.

### DIFF
--- a/.changeset/lucky-tiers-resolve.md
+++ b/.changeset/lucky-tiers-resolve.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/minimax': patch
+---
+
+Map MiniMax 480P and 768P frame sizes onto their named video resolution tiers, so a typed top-level `resolution` can reach them.
--- a/content/providers/01-ai-sdk-providers/33-minimax.mdx
+++ b/content/providers/01-ai-sdk-providers/33-minimax.mdx
@@ -137,9 +137,10 @@ The following optional provider options are available for MiniMax language model
 
 ## Video Models
 
-You can generate videos with the MiniMax-H3 model using the
-[`experimental_generateVideo`](/docs/reference/ai-sdk-core/generate-video)
-function:
+You can generate videos with the MiniMax-H3 and MiniMax-H3-Max models using
+the [`experimental_generateVideo`](/docs/reference/ai-sdk-core/generate-video)
+function. MiniMax-H3-Max is the faster variant: it renders at lower resolutions
+and serves every mode except reference-to-video.
 
 ```ts
 import { minimax, type MiniMaxVideoModelOptions } from '@ai-sdk/minimax';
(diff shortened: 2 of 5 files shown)


shaper (commented):
🟡 Frame-size→tier mapping is correct and matches the gateway PR exactly — two follow-up-sized suggestions.

- **[suggestion]** Silent conflict path + now-stale comment — `packages/minimax/src/minimax-video-model.ts:195-206`. The comment "the only way the two can disagree is a top-level value that maps to no tier at all" was true with 2K-only rows; this PR falsifies it: providerOptions `'2K'` + top-level `'1366x768'` now disagree with both recognized, and the top-level value is discarded with no warning. Update the comment at minimum; ideally warn + test the recognized-but-conflicting case.
- **[suggestion]** Unsupported-tier fallback goes *up*, not to nearest — `minimax-video-model.ts:207-220`. `resolution: '854x480'` on H3 warns then falls back to model default 2K (most expensive tier) rather than nearest supported (768P). Mitigated: warning emitted, documented in MDX, and gateway PR #3564 bills the tier actually rendered. Matches repo precedent, so not blocking.
- **[nit]** `1366x768` as the canonical 16:9 row (line 82) is a WXGA rounding of 1365.33 — fine, and critically identical to the gateway's map, so the cross-repo contract holds.

Verified: domain accuracy against MiniMax's live spec (H3-Max 480P/768P only, 5-15s, no reference inputs; H3 768P/2K, 4-15s), uppercase wire casing, backward compat (strictly widening), changeset present at correct patch level, sibling-provider consistency (xai/alibaba/bytedance pattern), no internal details leaked, all 12 new map rows tested.

shaper (approved):

PR #20019, opened 2026-08-31
fix(harness-pi): stream Pi tool input while it is generated

shiv3:
## Background

`@ai-sdk/harness` gained `tool-input-start` / `tool-input-delta` / `tool-input-end` in #19837, and `@ai-sdk/harness-claude-code` emits them today. `@ai-sdk/harness-pi` does not, so a consumer running Pi sees nothing while the model writes a tool input — the whole input appears at once with `tool_execution_start`.

For tools whose input *is* the artifact (a tool that takes a rendered HTML document, for example), that is the difference between watching the output appear and staring at a spinner for the entire generation.

## The gap

Pi already emits the partial input. `@earendil-works/pi-ai` has these in its assistant message event union:

```ts
| { type: "toolcall_start"; contentIndex: number; partial: AssistantMessage }
| { type: "toolcall_delta"; contentIndex: number; delta: string; partial: AssistantMessage }
| { type: "toolcall_end";   contentIndex: number; toolCall: ToolCall; partial: AssistantMessage }
```

`translatePiEvent`'s `message_update` case only handled `text_delta` and `thinking_delta`, so all three fell through to `return []`.

## What this changes

`packages/harness-pi/src/pi-translate.ts` translates the three events into the corresponding harness stream parts.

**Resolving the id.** Pi addresses these events by `contentIndex`, while the harness parts are keyed by tool call id. The id is already present in the partial assistant message at that index — Pi's providers push the `toolCall` block (with `id` and `name`) before emitting `toolcall_start` — so it is read once at start and reused for the deltas that follow. The index → id map is cleared per assistant message, since content block indices restart with each one.

**Degrading safely.** A delta whose start was not seen, or a start whose block has no id yet, is dropped rather than guessed at. The complete input still arrives with the `tool-call` at `tool_execution_start`, so nothing is lost — the input is simply not streamed.

**Consistent dispatch flags.** `tool-input-start` reports the same `providerExecuted` / `dynamic` flags as the `tool-call` that follows it, so a consumer does not have to wait for the call to know who will execute the tool. That decision is factored into `resolveToolDispatch` and shared with `tool_execution_start` so the two cannot drift.

`packages/harness-pi/src/pi-events.ts` gains `contentIndex` and `partial` on the `assistantMessageEvent` schema (it is a `looseObject`, so these were passing through untyped).

## Tests

Seven cases added to `pi-translate.test.ts`:

- streams `start` → `delta` × 2 → `end` for a single tool call
- `tool-input-start` reports the same dispatch as the following `tool-call`
- an MCP tool input is marked `dynamic`
- concurrent tool calls keep their deltas on their own ids
- a delta without a start is dropped
- a start whose block has no id yet does not stream
- ids are not carried across assistant messages

```
pnpm --filter @ai-sdk/harness-pi test        # 13 files, 182 passed
pnpm --filter @ai-sdk/harness-pi type-check
pnpm oxlint packages/harness-pi/src          # 0 warnings, 0 errors
```

## Notes

I have no Pi-backed integration environment to run this end to end, so the behaviour is covered by unit tests against the event shapes in `@earendil-works/pi-ai@0.74.2`. Happy to adjust if the maintainers would rather see this as an issue than a PR — CONTRIBUTING.md notes that a high-quality issue is often more useful, but this seemed small and focused enough to send as a patch.

Related: #19837 (the same change for Claude Code), #18904 (the original broader PR, now conflicting).

### DIFF
--- a/.changeset/olive-pumas-shave.md
+++ b/.changeset/olive-pumas-shave.md
@@ -1,0 +1,5 @@
+---
+'@ai-sdk/harness-pi': patch
+---
+
+fix(harness-pi): stream Pi tool input while it is generated
--- a/packages/harness-pi/src/pi-events.ts
+++ b/packages/harness-pi/src/pi-events.ts
@@ -12,6 +12,14 @@ export const piSessionEventSchema = z.looseObject({
     .looseObject({
       type: z.string().optional(),
       delta: z.string().optional(),
+      // `toolcall_start` / `toolcall_delta` / `toolcall_end` address a content
+      // block by index rather than by tool call id. The id and name live in the
+      // partial assistant message at that index, which Pi fills in before the
+      // first delta arrives.
+      contentIndex: z.number().optional(),
+      partial: z
+        .looseObject({ content: z.array(z.unknown()).optional() })
+        .optional(),
     })
     .optional(),
   toolCallId: z.string().optional(),
--- a/packages/harness-pi/src/pi-translate.test.ts
+++ b/packages/harness-pi/src/pi-translate.test.ts
@@ -46,6 +46,269 @@ describe('translatePiEvent', () => {
     expect(out[2]).toMatchObject({ type: 'text-delta', id, delta: 'world' });
   });
 
+  it('streams tool input as the model writes it', () => {
+    const state = createPiTranslatorState({ hostToolNames: ['visualize'] });
+    const partial = (args: string) => ({
+      content: [{ type: 'toolCall', id: 'call-1', name: 'visualize', args }],
+    });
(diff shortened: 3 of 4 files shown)


felixarntz (changes requested):
@shiv3 Thank you for the PR!

Code looks great, however your commits don't have verified signatures, so we cannot accept it. Could you please recreate a new PR with the exact same changes and signed commits?

shiv3:
Thanks for the review, @felixarntz! Recreated as #20105 with the same changes and a signed commit. Closing this one.

PR #20022, opened 2026-08-31
fix(openai): accept null nested stream error codes

JaviChulvi:
## Summary

- accept `null` `error.code` values in nested Responses API stream error events
- add a regression test using the captured `service_unavailable_error` event
- add a patch changeset for `@ai-sdk/openai`

## Observed provider event

OpenAI returned this event after accepting a Responses streaming request:

```json
{
  "type": "error",
  "error": {
    "type": "service_unavailable_error",
    "code": null,
    "message": "Unable to verify model access right now. Please retry.",
    "param": null
  },
  "sequence_number": 2
}
```

The provider message indicates that model-access verification was temporarily unavailable. It is a retryable provider-side failure; it does not by itself establish that the API key lacks model access.

## Problem

`openaiResponsesNestedErrorChunkSchema` requires nested `error.code` to be a string, while OpenAI emitted `null`:

```diff
- code: z.string(),
+ code: z.string().nullish(),
```

The corresponding documented top-level `error` event schema already accepts `null`/missing codes. The existing stream-error normalizer also handles null codes; the nested parser was the inconsistent, narrower validation point.

The Responses stream event catalog documents the top-level format, but the provider emitted the alternative nested format that the SDK already supports: https://developers.openai.com/api/reference/resources/responses/streaming-events

## Failure sequence

1. OpenAI accepted the request and SSE headers were committed.
2. OpenAI encountered the temporary model-access verification failure.
3. It emitted the nested error event above with `code: null`.
4. AI SDK rejected the event during schema validation, then surfaced `AI_TypeValidationError` instead of the provider error.
5. The application consequently emitted its generic internal-server-error stream message.

An HTTP 200 at the route is expected here because the streaming response had already begun before the provider error arrived. This is not caused by request validation, response-history chaining, routing, or the prompt.

## Result

With this change, the existing error normalization can preserve the provider message and return a retryable `APICallError` instead of a schema-validation failure. The change does not make an unavailable OpenAI service available.

## Testing

- `pnpm --filter @ai-sdk/openai test:node -- src/responses/openai-responses-language-model.test.ts`
- `pnpm --filter @ai-sdk/openai type-check`
- `pnpm exec oxfmt --check packages/openai/src/responses/openai-responses-api.ts packages/openai/src/responses/openai-responses-language-model.test.ts .changeset/blue-penguins-sip.md`


taranggoyal70 (commented):
Reviewed pinned head `90cf5d39`. This is a narrow schema fix that matches the observed provider event: the nested error code now accepts null/missing values just like the corresponding top-level error path, and the existing normalization can turn the frame into an `APICallError` instead of a validation failure. The regression test verifies the preserved message, retryable status, and 500 status code before output starts, and the package changeset is present. I did not find a correctness or compatibility issue in this diff.

PR #20036, opened 2026-08-31
feat(provider): add optional getFileMetadata/downloadFile/deleteFile to FilesV4 + streaming upload support

kkdawkins:
## Summary

Extends the `FilesV4` files interface so providers can expose the full file lifecycle, not just upload:

- **`getFileMetadata?` / `downloadFile?` / `deleteFile?`** — optional methods; presence signals capability, following the `VideoModelV4.doStart?/doStatus?` pattern. Non-breaking for all six existing implementers.
- **`FilesV4UploadFileCallOptions`** gains `abortSignal`, `headers`, and a `{ type: 'stream', stream }` data variant so large files can be uploaded without buffering.
- All operations identify files by the `SharedV4ProviderReference` returned from `uploadFile`, so results round-trip directly.
- **`FilesV4UploadFileResult`** gains `byteSize`/`createdAt`/`expiresAt` (symmetric with the retrieve result), so callers get the provider-applied retention expiry without digging in provider metadata.
- Core `uploadFile()` defaults `mediaType` to `application/octet-stream` for stream data (streams cannot be sniffed without consuming them).

### provider-utils additions

- **`postMultipartStreamToApi`** — streams a multipart/form-data body (`duplex: 'half'`) with deterministic part ordering (some providers require fields to precede the file part, e.g. xAI's `expires_after`). Buffered callers keep using `postFormDataToApi`.
- **`deleteFromApi`** — DELETE counterpart to `getFromApi` (config-derived URLs only).
- **`createBinaryStreamResponseHandler`** — passes `response.body` through as a `ReadableStream<Uint8Array>` (the existing `createBinaryResponseHandler` buffers).
- **`convertInlineFileDataToUint8Array`** now accepts the stream variant and rejects it with `UnsupportedFunctionalityError`, so providers without streaming support keep compiling and fail with a clear error.

## Motivation

AI Gateway's async-files service parks batch input bytes in provider file stores (OpenAI + xAI first) and needs upload/stat/download/delete under its own transport (`fetch` injection). Follow-up PRs implement these operations in `@ai-sdk/openai` and `@ai-sdk/xai`; the batch adapters (which currently re-implement upload/download inline) can be converged onto this surface later.

## Verification

- `packages/provider`: build + type-check clean
- `packages/provider-utils`: build + type-check clean, 840 tests pass (node + edge), incl. new suites for multipart framing/ordering/escaping, deleteFromApi, binary stream handler
- All six FilesV4 implementers (`openai`, `anthropic`, `google`, `xai`, `deepseek`, `anthropic-aws`) type-check unchanged
- `pnpm check` clean

First PR of a stack. Next: `@ai-sdk/openai` implementation, then `@ai-sdk/xai`.

Part of KDA-429


kkdawkins:
Review round 1 addressed (commits `8deacfbfef` + stack):

- **abortSignal/headers dropped by core `uploadFile()` (High)** — now destructured and forwarded to `filesApi.uploadFile`; also threaded through the remaining first-party implementations (`anthropic`, `deepseek`, `google` incl. its init/upload/poll steps; `anthropic-aws` inherits) with per-provider tests, so the contract is implemented everywhere rather than narrowed.
- **Filename-less stream part demoted to a scalar field (High)** — `type: 'file'` parts now always emit `filename` (default `'blob'`, matching `FormData.append`); the test that previously pinned the string-entry behavior now asserts a `File` named `blob`. Fixed at the helper, so OpenAI/xAI need no per-call defaulting.
- **Unreliable stream teardown on fetch rejection/abort (Medium)** — the multipart body is now built with an explicit `dispose()`: an active reader is cancelled (resolving a pending `read()` and propagating to the source) and never-entered part streams are cancelled directly; the failure path always disposes. New tests cover fetch rejection before the body is consumed and abandonment mid-read on a hanging source. Success paths deliberately do not dispose (the request consumed the body).
- **Core wrapper dropped `byteSize`/`createdAt`/`expiresAt` (Medium)** — added to `UploadFileResult` + `DefaultUploadFileResult` with passthrough test.
- **Changeset (Low)** — updated to mention the result metadata and teardown.
- Spec doc: `providerReference` JSDoc now states the key is the canonical provider name and may differ from the interface's `provider` id (`openai` vs `openai.files`).

Open question for SDK maintainers (flagged per review): streaming upload is not capability-signaled — it widens the required `uploadFile` input and relies on a runtime `UnsupportedFunctionalityError` from `convertInlineFileDataToUint8Array` for non-stream implementations. All six in-repo implementers fail loudly; please confirm this runtime-capability model is acceptable for third-party `FilesV4` implementations, or whether a `supportsStreamingUploads` flag / optional method split is preferred.

kkdawkins:
Review round 2 addressed (commit `073f2bd9e3`):

- **Google polling delay not abort-aware (Medium)** — now `delay(pollIntervalMs, { abortSignal: options.abortSignal })`; new test aborts during a pending 60s delay on a `PROCESSING` file and asserts prompt rejection.
- **Changeset coverage (Medium)** — `@ai-sdk/anthropic`, `@ai-sdk/deepseek`, and `@ai-sdk/google` added as patch entries (verify-changesets should pass now).
- **Docs** — core `uploadFile` JSDoc, the `uploadFile()` reference page, and the file-uploads guide now cover: stream input, the `application/octet-stream` default for streams, the `\"blob\"` filename default on multipart providers, `abortSignal`/`headers`, stream ownership (consumed by the provider, cancelled on failure, not reusable), and the `byteSize`/`createdAt`/`expiresAt` result fields.
- **Merged-reference semantics (DX 3)** — retrieve/delete result JSDoc now states results contain only the operated provider's entry and must not be reassigned over a merged multi-provider reference.

Stack ordering: the three PRs will merge together in order, so the PR-1-alone interval where OpenAI/xAI ignore the new upload options does not occur.

DX 1 (concrete capability subtypes for known providers) and DX 2 (core retrieve/download/delete helpers) are tracked as follow-ups rather than expanding this stack.

gr2m (commented):
Streams rejected before `postMultipartStreamToApi`—unsupported providers or provider-options validation—are left open, but the docs say failed uploads cancel the stream. Cancel on these paths or narrow the ownership guarantee.

kkdawkins:
@gr2m addressed in `89d276977e` — went with **cancel** rather than narrowing the guarantee:

- `convertInlineFileDataToUint8Array` cancels the stream before throwing `UnsupportedFunctionalityError`, covering every non-stream provider (anthropic, google, deepseek, anthropic-aws) at one point.
- Core `uploadFile()` wraps files-API resolution + the provider call: any rejection — no `files()` method, provider-side validation, transport — cancels stream data (guarded, so a provider that already consumed/disposed it is a no-op).
- OpenAI/xAI cancel on their own pre-request `parseProviderOptions` rejections in the stacked PRs (`c57106a7c1`, `721ea566e3`), so the guarantee holds when calling the provider interface directly too.
- Docs/JSDoc now say it explicitly: "any failed upload — including validation failures before a request is made — cancels it."

Tests: converter cancel-spy, core no-files()/provider-rejection cancellation, and per-provider pre-request cases (openai invalid options, xai out-of-range `expiresAfter`) all assert cancellation with zero requests made.

kkdawkins:
Per the Slack thread (commit `d85d7fa190` + stack):

- **Renamed `retrieveFile` → `getFileMetadata`** (@gr2m's suggestion) across the spec, both implementations, tests, changesets, and PR titles/bodies — `get*` matches the SDK's verb convention (`doGetBatchStatus`/`doGetBatchResults`), it's no longer confusable with `downloadFile`, and the name is more honest (the op returns metadata, never content).
- **Runtime stream-rejection model confirmed** (@nickoates) and his verification ask addressed: all four non-stream implementations (anthropic, google, deepseek; anthropic-aws reuses `AnthropicFiles`) route `data` through the shared `convertInlineFileDataToUint8Array`, which throws `UnsupportedFunctionalityError` and cancels the stream — now **pinned with an explicit per-provider test in each package** (rejects, cancels, zero requests made), so a future refactor away from the converter can't silently lose the contract.

gr2m (approved):

PR #20037, opened 2026-08-31
feat(openai): implement getFileMetadata/downloadFile/deleteFile and streaming uploads on OpenAIFiles

kkdawkins:
## Summary

Implements the full FilesV4 surface (from #20036) on `OpenAIFiles`:

- **`getFileMetadata`** — `GET /files/{id}`; maps `bytes`/`created_at`/`expires_at` to first-class `byteSize`/`createdAt`/`expiresAt` result fields
- **`downloadFile`** — `GET /files/{id}/content` via `createBinaryStreamResponseHandler` (content passes through as a `ReadableStream`, never buffered)
- **`deleteFile`** — `DELETE /files/{id}` with a dedicated response schema
- **Streaming uploads** — `{ type: 'stream' }` data goes through `postMultipartStreamToApi` (fields precede the file part); inline data keeps the existing buffered `FormData` path
- `abortSignal` and per-call `headers` threaded through all four operations
- File identity round-trips through the `SharedV4ProviderReference` (`{ openai: 'file-...' }`) returned by `uploadFile`; a reference without an `openai` key throws `InvalidArgumentError`

## Motivation

AI Gateway's async-files service parks batch input bytes in OpenAI's Files API under its own transport (`fetch` injection via `createOpenAI`) and needs the full upload/stat/download/delete lifecycle with in-flight streaming.

## Verification

- `packages/openai`: type-check clean, 979 tests pass (node + edge) incl. new suites for stream uploads (part ordering, expiry fields), retrieve (Date mapping, missing-reference rejection), download (stream passthrough), delete (method + response mapping)

Stacked on #20036. Next: `@ai-sdk/xai`.

Part of KDA-429


kkdawkins:
Review round 1 addressed (commit `a090c71597` + inherited base fixes):

- **Filename-less streaming uploads (High)** — fixed at the provider-utils layer (base PR): every file part now emits `filename` (default `'blob'`); added a filename-less stream upload test here asserting the part parses as a `File` named `blob`.
- **Empty / dot-segment file ids (Medium)** — `getFileId` rejects blank ids (`''`, whitespace); URLs now use the dot-segment-safe `encodePathSegment` (same implementation as `google-files.ts`), so `'.'`/`'..'` encode to `%252E` forms that cannot normalize into `/v1/files` or a parent path. Tests cover blank rejection and that dot-segment ids never reach the registered endpoints.
- **Changeset (Low)** — mentions `byteSize`/`createdAt`/`expiresAt` and the id hardening.

985 tests (node+edge) green at head.

gr2m (commented):
`downloadFile` drops `responseHeaders`, so it never returns `mediaType` when the content endpoint supplies `Content-Type`. Preserve that header in the result.

kkdawkins:
@gr2m addressed in `c57106a7c1`: `downloadFile` now reads `responseHeaders` from `getFromApi` and exposes the content endpoint's `Content-Type` (parameters stripped, e.g. `; charset=utf-8`) as `result.mediaType`; omitted when the header is absent. Also cancels stream data when `parseProviderOptions` rejects before any request (per the base-PR ownership guarantee). Tests cover both.

kkdawkins:
Renamed `retrieveFile` → `getFileMetadata` per maintainer feedback on the base PR (commit `745c99fc4c`). No behavior change.

gr2m (approved):

PR #20038, opened 2026-08-31
feat(xai): implement getFileMetadata/downloadFile/deleteFile, expiresAfter TTLs, and streaming uploads on XaiFiles

kkdawkins:
## Summary

Implements the full FilesV4 surface (from #20036) on `XaiFiles`, mirroring the OpenAI implementation in #20037:

- **`getFileMetadata`** — `GET /v1/files/{id}`; maps `bytes`/`created_at`/`expires_at` to `byteSize`/`createdAt`/`expiresAt`
- **`downloadFile`** — `GET /v1/files/{id}/content` as a passthrough `ReadableStream`
- **`deleteFile`** — `DELETE /v1/files/{id}` → `{ id, deleted }`
- **`expiresAfter` provider option** — xAI's upload TTL (seconds, 1h–30d per [docs.x.ai](https://docs.x.ai/developers/rest-api-reference/files/upload)), emitted as the plain `expires_after` field. **xAI rejects uploads where `expires_after` arrives after the `file` part**, so both upload paths emit all fields before the file (FormData preserves append order; the stream path uses `postMultipartStreamToApi`'s deterministic part ordering — the reason that helper guarantees order). The response schema now carries `expires_at`.
- **Streaming uploads** via `{ type: 'stream' }` data; inline data keeps the buffered `FormData` path
- `abortSignal`/per-call `headers` threaded through all four operations; file identity round-trips through `{ xai: 'file-...' }` provider references

## Motivation

AI Gateway's async-files service launches provider file parking with OpenAI and xAI. xAI's TTL support means system-credential parked files get the same forced-expiry backstop posture on both providers.

## Verification

- `packages/xai`: type-check clean, 461 tests pass (node + edge) incl. new suites: expires_after presence/ordering (`[...formData.keys()]` asserted), stream upload framing round-trip, retrieve/download/delete, missing-reference rejection

Stacked on #20037 (which stacks on #20036).

Part of KDA-429


kkdawkins:
Review round 1 addressed (commit `6c46d33c5a` + inherited base fixes):

- **Filename-less streaming uploads (High)** — fixed at the provider-utils layer (base PR): file parts always emit `filename` (default `'blob'`), which matters here since xAI records the multipart `filename=` parameter as the file's name.
- **`expiresAfter` range (Medium)** — schema is now `z.number().int().min(3600).max(2_592_000).optional()` (DeepSeek precedent); parametrized tests confirm fractions and out-of-range values reject before any fetch.
- **Empty / dot-segment file ids (Medium)** — same hardening as OpenAI: blank ids rejected, `encodePathSegment` on all three URL builders, tests assert `'.'`/`'..'` produce non-normalizable `%252E` paths.
- **Changeset (Low)** — updated.

469 tests (node+edge) green at head.

gr2m (commented):
`downloadFile` drops `responseHeaders`, so it never returns `mediaType` when the content endpoint supplies `Content-Type`. Preserve that header in the result.

kkdawkins:
@gr2m addressed in `721ea566e3`: `downloadFile` now exposes the content endpoint's `Content-Type` (parameters stripped) as `result.mediaType`, with an explicit no-header → `undefined` test; and stream data is cancelled when `parseProviderOptions` rejects pre-request (tested with an out-of-range `expiresAfter`, asserting zero fetch calls).

kkdawkins:
Renamed `retrieveFile` → `getFileMetadata` per maintainer feedback on the base PR (commit `62690ae1e6`). No behavior change.

gr2m (approved):

PR #20063, opened 2026-08-31
docs(site): add Black Forest Labs to the model picker

MaanavD:
## Summary

The interactive code preview on the docs site allowlists providers per model kind. Black Forest Labs is absent from both the image and video allowlists, so FLUX models never appear in the picker on `ai-sdk-core/image-generation` and `ai-sdk-core/video-generation`, even though the models are served by the gateway endpoint the picker itself reads from. Black Forest Labs also has no entry in `PROVIDER_LOGOS`, so its models would render with the generic placeholder mark.

This PR:

- Adds `bfl` to the `image` and `video` provider allowlists.
- Sets `bfl/flux-2-pro` and `bfl/flux-3-video` as the preferred defaults when a reader selects Black Forest Labs.
- Adds a `providerSdkMap` entry so the Provider tab generates `import { blackForestLabs } from "@ai-sdk/black-forest-labs"` rather than the incorrect `@ai-sdk/bfl` / `bfl` that the fallback would produce.
- Adds the provider's logo so its models render with the real mark instead of the generic fallback.

No behavior changes for any other provider.

## The logo asset

`apps/docs/public/images/icons/black-forest-labs.svg` is the official symbol-only mark from [bfl.ai/brand](https://bfl.ai/brand), which the brand guidelines designate for small applications where the wordmark would be unreadable.

- Path data and `viewBox` are byte-identical to the published `symbol-black.svg`.
- The only changes: `fill` is `currentColor` instead of the fixed `#07130e`, and a `clipPath` wrapper was dropped. That clip was a full-canvas `M0 0h196v140H0z` rect, so removing it is a no-op. Using `currentColor` lets the existing `invert: true` dark-mode handling work, the same treatment already applied to the Anthropic and xAI marks.

## Verification

- `GET https://ai-gateway.vercel.sh/v1/models` returns `bfl/flux-3-video` with `type: video`, and `bfl/flux-2-pro`, `bfl/flux-2-max`, `bfl/flux-2-flex`, `bfl/flux-2-klein-9b`, `bfl/flux-2-klein-4b` with `type: image`.
- `parseModels` keys on `owned_by`, not the model ID prefix, so I confirmed `owned_by` is exactly `bfl` for all 11 models rather than assuming it from the ID. Replaying the real filter against the live response with this patch applied yields 10 FLUX models in the image picker and `bfl/flux-3-video` in the video picker; without it, zero.
- `providerTitles` already contains `bfl: 'Black Forest Labs'`, so the label renders correctly with no further change.
- The package name and export were confirmed against `packages/black-forest-labs` (`@ai-sdk/black-forest-labs`, exports `blackForestLabs`), so the Provider tab snippet is copy-pasteable.
- Verified the committed SVG renders as the complete BFL peak mark, not a clipped or corrupted path.
- `oxfmt --check` on the changed file passes.

Note on scope: `FIRST_PARTY_PROVIDERS` is deliberately untouched. It feeds the text-model preview, and this provider exposes no language model (`languageModel` throws `NoSuchModelError`), so it does not belong in that list.

I was not able to visually confirm the rendered dropdown: the local docs dev server ran out of memory on this machine before serving a page. Flagging that rather than claiming a visual check I did not complete. The logic change is data-only and confined to three lookup tables, and the Docs Site CI job covers the build.

The same logo asset is added by #20079, which adds the provider index card. Whichever merges first, the other's copy is identical (sha256 `cf8b6a00ff0752743ea35ec4b61485befc355d7d2469b6f417ea9c30037ed899`), so the overlap resolves cleanly.


taranggoyal70 (commented):
Reviewed pinned head `66e6dbfe`. The data flow is consistent end to end: `bfl` is added only to image/video allowlists, not the text provider list; the preferred IDs match the live AI Gateway model pages (`bfl/flux-2-pro` and `bfl/flux-3-video`); and the direct-provider snippet maps to the repository's actual `@ai-sdk/black-forest-labs` package and `blackForestLabs` export. The monochrome SVG treatment is consistent with the existing inverted provider marks. Docs CI, type checking, and the package test matrix are green; I did not find a code-level issue in this diff.

PR #20109, opened 2026-09-01
fix(ai): use || so an empty error body falls back to the default message

### LINKED ISSUE #20107: empty error response body produces `new Error('')` because `??` never fires
### Description

when an http error response has an empty body, `useChat` and friends surface an error whose message is the empty string, so the ui shows nothing and the server log shows a bare `Error`.

the cause is a `??` that can never fire:

```ts
if (!response.ok) {
  throw new Error(
    (await response.text()) ?? 'Failed to fetch the chat response.',
  );
}
```

`Response.text()` resolves to a string, empty for an empty body, never `null` or `undefined`. so `??` never takes the fallback and the default message is dead code. `||` is what was intended.

six sites do this:

- `packages/ai/src/ui/http-chat-transport.ts:204`
- `packages/ai/src/ui/http-chat-transport.ts:260`
- `packages/ai/src/ui/call-completion-api.ts:79`
- `packages/react/src/use-object.ts:188`
- `packages/svelte/src/structured-object.svelte.ts:173`
- `packages/angular/src/lib/structured-object.ng.ts:162`

the seventh copy of the same block already uses `||`, which is what makes me think this is a slip rather than a decision:

- `packages/vue/src/use-object.ts:164`

### Reproduction

the operator behaviour on its own:

```js
const t = await new Response(null, { status: 502 }).text()
JSON.stringify(t)          // '""'
(t ?? 'fallback')          // ''          <- current code, dead fallback
(t || 'fallback')          // 'fallback'  <- intended
```

end to end, a route that returns `new Response(null, { status: 502 })` (or any proxy 502/504 with no body) with `DefaultChatTransport`:

```
error.message : ""
String(error) : "Error"
```

so a ui rendering `{error.message}` renders an empty string and the user sees no error at all.

### AI SDK Version

- ai: 7.0.87
- node: 25.9.0

happy to send the one character change as a pr, it is the kind of small focused change CONTRIBUTING mentions. will link it here if so.

Nixxx19:
## Background

`Response.text()` resolves to a string, empty for an empty body, never `null` or `undefined`. so in

```ts
(await response.text()) ?? 'Failed to fetch the chat response.'
```

the `??` can never take the fallback, and an error response with an empty body throws `new Error('')`. a ui rendering `{error.message}` then shows nothing at all, and the server log shows a bare `Error`.

`packages/vue/src/use-object.ts:164` is the same copied block and already uses `||`. one of the seven was right and six were not, which is what makes this look like a slip rather than a deliberate choice.

## Summary

changed `??` to `||` at the six remaining sites:

- `packages/ai/src/ui/http-chat-transport.ts` (two)
- `packages/ai/src/ui/call-completion-api.ts`
- `packages/react/src/use-object.ts`
- `packages/svelte/src/structured-object.svelte.ts`
- `packages/angular/src/lib/structured-object.ng.ts`

vue is already correct and untouched, so all seven now agree.

behaviour changes only for a response that is both not ok and has an empty body: previously `Error('')`, now the intended default message. a non empty body still surfaces verbatim.

## End-to-End Verification

confirmed the premise at runtime rather than assuming it:

```js
const t = await new Response(null, { status: 502 }).text()
JSON.stringify(t)   // '""'
t ?? 'fallback'     // ''
t || 'fallback'     // 'fallback'
```

then end to end through `HttpChatTransport` with a fetch returning `new Response(null, { status: 502 })`. before the change `error.message` is `""` and `String(error)` is `"Error"`, so a ui rendering the message shows nothing. after the change the error reads `Failed to fetch the chat response.`

i also reverted the source change and re-ran the added tests to check they actually catch this. they fail on the old code with `expected [Function] to throw error including 'Failed to fetch the chat response.' but got ''`, which is the bug itself.

`src/ui/` is green, 13 files and 346 tests.

## Checklist

## Related Issues

Fixes #20107

### DIFF
--- a/.changeset/empty-error-body-fallback.md
+++ b/.changeset/empty-error-body-fallback.md
@@ -1,0 +1,8 @@
+---
+"ai": patch
+"@ai-sdk/react": patch
+"@ai-sdk/svelte": patch
+"@ai-sdk/angular": patch
+---
+
+fix: fall back to the default message when an error response has an empty body
--- a/packages/ai/src/ui/call-completion-api.ts
+++ b/packages/ai/src/ui/call-completion-api.ts
@@ -76,7 +76,7 @@ export async function callCompletionApi({
 
     if (!response.ok) {
       throw new Error(
-        (await response.text()) ?? 'Failed to fetch the chat response.',
+        (await response.text()) || 'Failed to fetch the chat response.',
       );
     }
 
--- a/packages/ai/src/ui/http-chat-transport.test.ts
+++ b/packages/ai/src/ui/http-chat-transport.test.ts
@@ -202,4 +202,51 @@ describe('HttpChatTransport', () => {
       expect(receivedAbortSignal).toBe(abortController.signal);
     });
   });
+  describe('error responses', () => {
+    it('should fall back to the default message when the error body is empty', async () => {
+      const transport = new MockHttpChatTransport({
+        api: 'http://localhost/api/chat',
+        fetch: async () => new Response(null, { status: 502 }),
+      });
+
+      await expect(
+        transport.sendMessages({
+          chatId: 'c123',
+          messageId: 'm123',
+          trigger: 'submit-message',
+          messages: [
+            {
+              id: 'm123',
+              role: 'user',
+              parts: [{ text: 'Hello, world!', type: 'text' }],
+            },
+          ],
+          abortSignal: new AbortController().signal,
+        }),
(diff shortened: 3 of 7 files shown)


ArshVermaGit (commented):
Looks good — the || fallback correctly handles empty error bodies, and the added tests cover both empty and non-empty responses. No issues found.

n1ckoates:
Original issue closed

PR #20138, opened 2026-09-01
feat: support per-request models in batch

### LINKED ISSUE #20065: Per-request models in batch APIs
## Problem

The current batch API is based on the concept of a “batchable language model”:

```ts
type BatchLanguageModelV4 = LanguageModelV4 & BatchModelV4<...>;
```

A model instance therefore owns both its normal language-model methods and the batch lifecycle methods. This creates several problems:

- A single model is selected for the entire batch, even though providers such as Anthropic support choosing a different model for each request.
- AI Gateway will similarly need per-request model selection, with Gateway acting as the provider rather than as one fixed model.
- Providers have to change their language-model return types to `BatchLanguageModelV4` even though their normal language-model behavior has not changed.
- Batch implementations have to extend or wrap a language model solely to satisfy the type hierarchy.
- Batch references assume one `modelId`, which is no longer correct when a batch contains multiple models.
- The underlying batch abstraction is coupled to language models, even though providers may eventually support batches of image or other model requests.

Batching should instead be a provider API, similar to the Files API. It is neither a capability of an individual language-model instance nor inherently tied to a particular modality.

## Proposed Core API

Pass the provider and default model separately:

```ts
import { startTextBatch } from 'ai';

const batch = await startTextBatch({
  provider: openai,
  model: 'gpt-5.6',
  requests: [
    {
      id: 'request-1',
      prompt: 'Explain quantum computing.',
    },
    {
      id: 'request-2',
      model: 'gpt-5.6-mini',
      prompt: 'Summarize this document.',
    },
  ],
});
```

The top-level `model` is the default. An individual request can override it:

```ts
request.model ?? options.model
```

Providers that permit mixed models can pass these model IDs through. Providers with stricter requirements can validate the resolved requests at runtime and return an appropriate error.

There should not be a `supportsPerRequestModels` type-level capability flag. The public request shape remains consistent across providers, while provider-specific restrictions are enforced at runtime.

## Provider API

Add the batch interface directly to `ProviderV4`, alongside `languageModel`, `imageModel`, `files`, and the other provider APIs:

```ts
interface ProviderV4 {
  languageModel(modelId: string): LanguageModelV4;
  imageModel(modelId: string): ImageModelV4;

  files?(): FilesV4;
  batch?(): BatchV4;
}
```

This does not introduce a separate kind of “batch provider.” `openai`, `anthropic`, and other existing provider objects remain the objects users pass to Core.

Core resolves the provider’s batch implementation internally:

```ts
startTextBatch({
  provider: openai,
  model: 'gpt-5.6',
  requests,
});
```

This follows the same pattern as `uploadFile({ provider: openai })`, where Core calls `openai.files()` internally.

As with the Files API, Core functions could also accept a `BatchV4` implementation directly for custom and lower-level use cases:

```ts
type BatchProvider = ProviderV4 | BatchV4;
```

The ordinary provider object should remain the recommended public API.

Follow-up operations use the same provider:

```ts
const status = await getBatchStatus({
  provider: openai,
  batch,
});

const results = await getBatchResults({
  provider: openai,
  batch,
});
```

## One modality-independent batch service

Providers expose one batch service:

```ts
openai.batch();
anthropic.batch();
google.batch();
```

We should not initially expose modality- or endpoint-specific variants such as:

```ts
openai.batch.chat();
openai.batch.responses();
openai.batch.image();
```

The provider-level `BatchV4` interface represents the common batch lifecycle. The operation passed to it carries a modality discriminant.

For example, Core’s `startTextBatch` would normalize its input into something like:

```ts
type BatchV4StartOptions =
  | {
      type: 'text';
      requests: LanguageModelBatchRequestV4[];
    }
  | {
      type: 'image';
      requests: ImageModelBatchRequestV4[];
    };
```

Only the `text` variant needs to be implemented initially. The discriminant should still be included now so that `BatchV4` is not designed specifically around language-model requests.

A provider implementation can narrow the operation before handling it:

```ts
async function doStartBatch(options: BatchV4StartOptions) {
  switch (options.type) {
    case 'text':
      return startTextBatch(options);

    default:
      throw new UnsupportedFunctionalityError({
        functionality: `Batch modality: ${options.type}`,
      });
  }
}
```

A future `startImageBatch` could use the same provider service while passing the `image` variant. Providers could support whichever modalities their upstream batch APIs support and reject the others at runtime.

This keeps the provider API unified without forcing the request and result types for different modalities into one undifferentiated shape.

For OpenAI, text batches should use the Responses API. Supporting the legacy Chat Completions API does not justify adding another public API variant.

### Add a standalone `BatchV4`

Introduce a modality-independent batch service:

```ts
interface BatchV4<ModelId extends string = string> {
  readonly specificationVersion: 'v4';
  readonly provider: string;

  // batch lifecycle methods
}
```

Its start method accepts a discriminated batch operation rather than specifically accepting language-model requests:

```ts
interface BatchV4<ModelId extends string = string> {
  experimental_doStartBatch(
    options: BatchV4StartOptions<ModelId>,
  ): Promise<BatchV4StartResult>;

  experimental_doGetBatchStatus(
    options: BatchV4OperationOptions,
  ): Promise<BatchV4Status>;

  experimental_doGetBatchResults(
    options: BatchV4OperationOptions,
  ): Promise<BatchV4ItemResult[]>;
}
```

The exact names can be decided during implementation, but the important distinction is:

```ts
BatchV4StartOptions<ModelId> =
  | TextBatchV4StartOptions<ModelId>
  | ImageBatchV4StartOptions<ModelId>
  | ...
```

For the initial implementation, only the text variant needs to contain usable request options. Other modalities do not need Core APIs or provider implementations yet.

`BatchV4` does not extend `LanguageModelV4`, `ImageModelV4`, or any other model interface.

### Add normalized modality-specific requests

The normalized text request passed to the batch service contains its resolved model ID:

```ts
interface LanguageModelBatchRequestV4<
  ModelId extends string = string,
> {
  id: string;
  modelId: ModelId;

  // normalized language-model request options
}
```

A future image batch would use its own request type:

```ts
interface ImageModelBatchRequestV4<
  ModelId extends string = string,
> {
  id: string;
  modelId: ModelId;

  // normalized image-model request options
}
```

The batch service can distinguish these request shapes through the start operation’s `type` field.

This avoids making image batch options a variation of language-model batch options.

### Update `ProviderV4`

Add the optional batch factory:

```ts
interface ProviderV4 {
  batch?(): BatchV4;
}
```

Concrete providers can return a narrower model-ID type:

```ts
interface OpenAIProvider extends ProviderV4 {
  batch(): BatchV4<OpenAIBatchModelId>;
}

interface AnthropicProvider extends ProviderV4 {
  batch(): BatchV4<AnthropicBatchModelId>;
}
```

If model IDs differ by modality, `BatchV4` can expose a modality-to-model-ID mapping instead of one union:

```ts
interface OpenAIBatchModelIds {
  text: OpenAIResponsesModelId;
  image: OpenAIImageModelId;
}
```

The exact representation is an implementation detail, but Core should be able to infer the correct model-ID type for the selected batch modality.

### Update Core batch options

Replace the batchable model with separate provider and model fields:

```ts
type StartTextBatchOptions<ModelId extends string> = {
  provider: ProviderV4 | BatchV4<ModelId>;
  model: ModelId;
  requests: Array<{
    id: string;
    model?: ModelId;
    // prompt and request options
  }>;
};
```

The model type should be inferred from the concrete provider’s batch service. This gives users autocomplete and provider-specific model checking for both the default model and per-request overrides.

Provider model-ID unions may continue to include `string & {}` where necessary to allow fine-tuned or newly released model IDs while preserving autocomplete.

### Keep batch references modality-specific

The batch reference should identify its modality, but should no longer contain one `modelId`:

```ts
interface TextBatchReference {
  type: 'text';
  id: string;
  provider: string;
}
```

A future image batch would return an image batch reference:

```ts
interface ImageBatchReference {
  type: 'image';
  id: string;
  provider: string;
}
```

This lets shared lifecycle functions infer or validate the expected result shape without assuming that all batches contain language-model results.

### Remove the current batchable-model types

Remove:

```ts
BatchLanguageModelV4
BatchModelV4
```

Provider model factories return their ordinary model types as usual:

```ts
openai('gpt-5.6'); // LanguageModelV4
openai.image('gpt-image-1'); // ImageModelV4
```

Adding batch support no longer changes the type or implementation hierarchy of these models.

## Batch implementations do not extend model implementations

A batch implementation receives:

- The batch modality
- A collection of normalized requests
- A model ID for each request
- The relevant request options

It does not need to implement generation, streaming, image generation, or any other model operation.

It may internally reuse request-conversion logic from the provider’s normal model implementations, but that is an implementation detail rather than an inheritance relationship.

A `BatchV4` cannot be passed to `generateText`, and a `LanguageModelV4` is not automatically a batch service.

## No language-model shorthand

We should not additionally support:

```ts
startTextBatch({
  model: openai('gpt-5.6'),
  requests,
});
```

Doing so would require language-model instances to carry or expose their provider’s batch service, recreating the coupling this change is intended to remove. It would also introduce multiple ways to express the same operation and make middleware or wrapped-model behavior ambiguous.

Because the existing batch APIs are experimental, retaining this shape for backwards compatibility is not necessary.

## Expected outcome

After this change:

- Batch APIs are provider-owned services rather than model capabilities.
- The batch service is not coupled to text or any other modality.
- Each batch operation explicitly identifies its modality.
- Providers can implement supported modalities and reject unsupported ones at runtime.
- Anthropic, AI Gateway, and other providers can support per-request models.
- Provider model types do not change when batching is implemented.
- Batch implementations no longer extend model implementations.
- Users retain provider- and modality-specific model autocomplete.
- Core can add future modality-specific creation functions without adding separate provider batch services.

n1ckoates:
## Background

- Batch APIs were previously attached to individual language model instances, which prevented requests in a batch from selecting their own models.
- This architecture also made the provider-facing API unnecessarily specific to text generation and difficult to extend to image, embedding, and other batch modalities.
- This PR addresses #20065 by moving batch execution to a provider-owned interface rather than a language-model-owned one.

## Summary

- Added the provider-owned `BatchV4` interface, exposed through `provider.experimental_batch()`.
  - Batch lifecycle types are modality agnostic and can support image, embedding, and other modalities in the future.
  - Text-specific request and result types provide the current text-generation implementation.
- Added per-request model IDs.
  - Each request specifies its own `type` and `model`.
  - A batch-level model is no longer supported.
- Migrated OpenAI, xAI, Anthropic, Google, and Gateway to the new API.
  - OpenAI, Google, and Gateway fail fast when a batch contains multiple model IDs.
  - Anthropic and xAI support mixed models.
- Moved `supportedUrls` to the batch service because a batch is no longer associated with a single language model instance.
- Added validation to reject conflicting tool names (same tool name across different requests where the actual tool definition differs).

New API design:

```ts
import {
  experimental_getBatchResults as getBatchResults,
  experimental_getBatchStatus as getBatchStatus,
  experimental_startBatch as startBatch,
} from 'ai';

const batch = await startBatch({
  provider: anthropic,
  requests: [
    {
      id: 'capital-france',
      type: 'text',
      model: 'claude-sonnet-5',
      prompt: 'What is the capital of France?',
    },
    {
      id: 'capital-germany',
      type: 'text',
      model: 'claude-opus-5',
      prompt: 'What is the capital of Germany?',
    },
  ],
});

while (true) {
  const { status } = await getBatchStatus({
    provider: anthropic,
    batch,
  });

  if (status !== 'pending') {
    break;
  }

  await new Promise(resolve => setTimeout(resolve, 10_000));
}

for await (const result of getBatchResults({
  provider: anthropic,
  batch,
})) {
  console.log(result);
}
```

## End-to-End Verification

Ran all batch examples under `examples/ai-functions/src/start-batch`.

## Checklist

- [x] A *patch* changeset for relevant packages has been added (for bug fixes / features—run `pnpm changeset` in the project root)

## Future Work

- Share more request-conversion and response-mapping logic between provider language model and batch implementations.
- Add image, embedding, and other batch modalities to the modality-aware contract.

## Related Issues

Fixes https://github.com/vercel/ai/issues/20065


packages/ai/src/batch/batch-types.ts
```diff
@@ -37,7 +38,6 @@ export type TextBatchReference = {
   readonly type: 'text';
   readonly id: string;
   readonly provider: string;
-  readonly modelId: string;
```
gr2m, line 40: should we bump to `version: 2` since we remove modelId?

packages/provider/src/provider/v4/provider-v4.ts
```diff
@@ -92,4 +93,12 @@ export interface ProviderV4 {
...
    */
   skills?(): SkillsV4;
+
+  /**
+   * Returns the batch interface for processing batches with the provider.
+   * The returned interface can be passed to the batch functions.
+   *
+   * @returns {BatchV4} The batch interface for this provider.
+   */
+  experimental_batch?(): BatchV4;
```
gr2m: we don't want a stable spec to reference an experimental spec. Just leave it out of provider, the same way we do with video.

packages/ai/src/batch/batch-types.ts
n1ckoates: bumped

packages/provider/src/provider/v4/provider-v4.ts
n1ckoates: fixed

gr2m (approved):

PR #20173, opened 2026-09-01
fix(mcp): tool results - tolerate structured-only responses

### LINKED ISSUE #20171: @ai-sdk/mcp: tolerate structuredContent-only tool results
## Description

`@ai-sdk/mcp` rejects a `tools/call` result when an MCP server returns `structuredContent` without `content`:

```text
MCPClientError: Failed to parse server response
Caused by: Invalid input: expected array, received undefined at content
```

Shopify's public Global Catalog MCP endpoint currently returns this shape. The MCP result schema requires `content`, while the structured-content guidance says servers SHOULD also mirror structured JSON into a `TextContent` block for backwards compatibility. Shopify intentionally omits that mirror to avoid doubling large catalog payloads.

Although the server response is not strictly schema-conformant, the client has all the data needed to recover safely. It would be useful for `@ai-sdk/mcp` to tolerate this specific compatibility case while continuing to reject results that contain neither `content` nor `structuredContent`.

Relevant specification: https://modelcontextprotocol.io/specification/2026-07-28/server/tools#structured-content

## Reproduction

```ts

  transport: {
    type: 'http',
    url: 'https://catalog.shopify.com/api/ucp/mcp',
  },
});

try {
  await client.callTool({
    name: 'search_catalog',
    arguments: {
      meta: {
        'ucp-agent': {
          profile:
            'https://shopify.dev/ucp/agent-profiles/2026-04-08/valid-with-capabilities.json',
        },
      },
      catalog: { query: 'shoes' },
    },
  });
} finally {
  await client.close();
}
```

The endpoint responds with a result whose only top-level result key is `structuredContent`. `CallToolResultSchema` requires `content`, so the private request parser rejects it before callers or `toolsFromDefinitions()` can inspect the structured value.

## Expected behavior

When `content` is absent but `structuredContent` is present, `callTool()` should accept the result. Ideally it should normalize the result to the backwards-compatible shape before returning it:

```ts
{
  content: [
    { type: 'text', text: JSON.stringify(structuredContent) },
  ],
  structuredContent,
}
```

The complete-result schema could make `content` optional, refine that at least one of `content` or `structuredContent` is present, and transform structured-only results by synthesizing the text block. That keeps existing return types and model-output behavior intact, preserves strict validation for unusable empty results, and also works for results delivered through SSE rather than only JSON HTTP responses.

## AI SDK Version

- `@ai-sdk/mcp`: 2.0.29
- `zod`: 4.5.4

The same required-`content` schema is present on the current `main` branch.

## Code of Conduct

AndrewBarba:
## Background

Some MCP servers, including Shopify Global Catalog, return `structuredContent` from `tools/call` without the backwards-compatible `content` field. The MCP client currently rejects those responses during schema validation, before callers can handle the structured result.

## Summary

- Parse structured-only results through a dedicated JSON-value schema, serialize them into a text content block, and pipe the transformed value through the canonical result schema.
- Require the legacy `toolResult` alternative to be genuinely present across supported Zod versions while preserving forward-compatible unknown content types explicitly.
- Cover direct calls, cached tool definitions, typed output schemas, every JSON value kind, error results, invalid serialization, empty-result rejection, and existing result variants.
- Document the compatibility behavior and add a patch changeset for `@ai-sdk/mcp`.

## End-to-End Verification

Called Shopify Global Catalog's live `search_catalog` tool through the patched local package. The call returned `content`, `structuredContent`, and `isError`; the synthesized content contained one text block whose text exactly matched `JSON.stringify(result.structuredContent)`.

Also verified the schema behavior against the repository's Zod 3.25 compatibility layer, native Zod 4.1.8 (the minimum supported peer), and Zod 4.5.4 from the original reproduction.

## Checklist

## Related Issues

Fixes #20171

### DIFF
--- a/packages/mcp/src/tool/mcp-client.test.ts
+++ b/packages/mcp/src/tool/mcp-client.test.ts
@@ -544,6 +544,34 @@ describe('MCPClient', () => {
     `);
   });
 
+  it('should normalize structured tool results that omit content', async () => {
+    const structuredContent = {
+      products: [{ id: 'gid://shopify/Product/1', title: 'Trail Shoes' }],
+    };
+    client = await createMCPClient({
+      transport: new MockMCPTransport({
+        toolCallResults: {
+          'mock-tool': {
+            structuredContent,
+          } as unknown as CallToolResult,
+        },
+      }),
+    });
+
+    await expect(
+      client.callTool({ name: 'mock-tool', arguments: { foo: 'bar' } }),
+    ).resolves.toEqual({
+      content: [
+        {
+          type: 'text',
+          text: JSON.stringify(structuredContent),
+        },
+      ],
+      structuredContent,
+      isError: false,
+    });
+  });
+
   it('should return tools from all paginated tool list responses', async () => {
     const transport = new PaginatedToolsTransport();
     client = await createMCPClient({ transport });
@@ -668,8 +696,15 @@ describe('MCPClient', () => {
   });
 
   it('should create tools from cached definitions via toolsFromDefinitions()', async () => {
+    const structuredContent = { value: 42 };
     client = await createMCPClient({
-      transport: { type: 'sse', url: 'https://example.com/sse' },
+      transport: new MockMCPTransport({
+        toolCallResults: {
+          'mock-tool': {
+            structuredContent,
+          } as unknown as CallToolResult,
+        },
+      }),
     });
 
     // Get definitions (this would normally be cached)
@@ -688,8 +723,10 @@ describe('MCPClient', () => {
       { foo: 'bar' },
       { messages: [], toolCallId: '1', context: {} },
     );
-    expect(result).toMatchObject({
-      content: [{ type: 'text', text: 'Mock tool call result' }],
+    expect(result).toEqual({
+      content: [{ type: 'text', text: JSON.stringify(structuredContent) }],
+      structuredContent,
+      isError: false,
     });
   });
 
@@ -875,7 +912,7 @@ describe('MCPClient', () => {
             'get-mixed': {
               content: [
                 { type: 'text', text: 'Here is an image:' },
-                { type: 'image', data: 'base64data', mimeType: 'image/png' },
+                { type: 'image', data: 'aGVsbG8=', mimeType: 'image/png' },
               ],
               isError: false,
             },
@@ -900,13 +937,12 @@ describe('MCPClient', () => {
               "type": "text",
             },
             {
-              "data": "base64data",
+              "data": "aGVsbG8=",
               "mimeType": "image/png",
               "type": "image",
             },
           ],
           "isError": false,
-          "toolResult": undefined,
         }
       `);
 
@@ -918,7 +954,7 @@ describe('MCPClient', () => {
         output: {
           content: [
             { type: 'text', text: 'Here is an image:' },
-            { type: 'image', data: 'base64data', mimeType: 'image/png' },
+            { type: 'image', data: 'aGVsbG8=', mimeType: 'image/png' },
           ],
           isError: false,
         },
@@ -933,7 +969,7 @@ describe('MCPClient', () => {
           },
           {
             "data": {
-              "data": "base64data",
+              "data": "aGVsbG8=",
               "type": "data",
             },
             "mediaType": "image/png",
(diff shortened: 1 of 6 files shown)


AndrewBarba:
Addressed the feature-review concerns in 1edf53eb36: the legacy `toolResult` branch now requires a genuinely present value across supported Zod versions; structured-only results require JSON-valid `structuredContent`, while empty, arbitrary, undefined, non-JSON, and cyclic results reject; and forward-compatible unknown content types are handled explicitly instead of through the legacy-schema loophole. Coverage now includes every JSON value kind, error results, existing content, legacy results, malformed known content, direct `callTool()`, `tools()` with an output schema, and `toolsFromDefinitions()`. The MCP docs now describe the compatibility behavior.

lgrammel (approved):

packages/mcp/src/tool/mcp-client.test.ts
```diff
@@ -875,7 +912,7 @@ describe('MCPClient', () => {
             'get-mixed': {
               content: [
                 { type: 'text', text: 'Here is an image:' },
-                { type: 'image', data: 'base64data', mimeType: 'image/png' },
+                { type: 'image', data: 'aGVsbG8=', mimeType: 'image/png' },
```
aayush-kapoor, line 915: why is this changing? this test seems unrelated

packages/mcp/src/tool/mcp-client.test.ts
```diff
@@ -900,13 +937,12 @@ describe('MCPClient', () => {
               "type": "text",
             },
             {
-              "data": "base64data",
+              "data": "aGVsbG8=",
```
aayush-kapoor, line 940: same as ^

packages/mcp/src/tool/mcp-client.test.ts
```diff
@@ -918,7 +954,7 @@ describe('MCPClient', () => {
         output: {
           content: [
             { type: 'text', text: 'Here is an image:' },
-            { type: 'image', data: 'base64data', mimeType: 'image/png' },
+            { type: 'image', data: 'aGVsbG8=', mimeType: 'image/png' },
```
aayush-kapoor, line 957: same as ^

packages/mcp/src/tool/mcp-client.test.ts
```diff
@@ -933,7 +969,7 @@ describe('MCPClient', () => {
           },
           {
             "data": {
-              "data": "base64data",
+              "data": "aGVsbG8=",
```
aayush-kapoor, line 972: same as ^

packages/mcp/src/tool/mcp-client.test.ts
```diff
@@ -2367,17 +2405,11 @@ describe('MCPClient', () => {
         ],
         toolCallResults: {
           'weather-tool': {
-            content: [
-              {
-                type: 'text',
-                text: '{"temperature": 22.5, "conditions": "Sunny"}',
-              },
-            ],
```
aayush-kapoor, line 2375: is this expected to be gone since now the output is just represented by structuredContent?

packages/mcp/src/tool/mcp-client.test.ts
AndrewBarba: The fixture wasn’t valid base64. It only passed because the permissive legacy branch masked the validation failure; tightening that branch surfaced it.

packages/mcp/src/tool/mcp-client.test.ts
AndrewBarba: Snapshot update from the valid-base64 fixture above.

packages/mcp/src/tool/mcp-client.test.ts
AndrewBarba: Same fixture update.

packages/mcp/src/tool/mcp-client.test.ts
AndrewBarba: Same snapshot update.

packages/mcp/src/tool/mcp-client.test.ts
AndrewBarba: Yep—this intentionally exercises a structured-only response. Parsing adds content; the output-schema path still returns the validated structured content.

aayush-kapoor (approved):

PR #20188, opened 2026-09-02
fix: use || instead of ?? for empty error response body

### LINKED ISSUE #20107: empty error response body produces `new Error('')` because `??` never fires

wasim-builds:
## Summary

Fixes #20107

Response.text() returns empty string for empty bodies, never null/undefined. The ?? operator never fires the fallback. Using || ensures the default message is shown when the response body is empty.

## Changes

- Changed ?? to || in 6 files where error response body is handled

## Test plan

### DIFF
--- a/packages/ai/src/ui/call-completion-api.ts
+++ b/packages/ai/src/ui/call-completion-api.ts
@@ -76,7 +76,7 @@ export async function callCompletionApi({
 
     if (!response.ok) {
       throw new Error(
-        (await response.text()) ?? 'Failed to fetch the chat response.',
+        (await response.text()) || 'Failed to fetch the chat response.'),
       );
     }
 
--- a/packages/ai/src/ui/http-chat-transport.ts
+++ b/packages/ai/src/ui/http-chat-transport.ts
@@ -201,7 +201,7 @@ export abstract class HttpChatTransport<
 
     if (!response.ok) {
       throw new Error(
-        (await response.text()) ?? 'Failed to fetch the chat response.',
+        (await response.text()) || 'Failed to fetch the chat response.'),
       );
     }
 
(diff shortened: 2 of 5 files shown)


n1ckoates:
Original issue closed

PR #20189, opened 2026-09-02
fix: preserve per-call diagnostics in NoImageGeneratedError

### LINKED ISSUE #20153: `NoImageGeneratedError` discards the `GenerateImageCall` diagnostics already collected by `generateImage`
### Description

`generateImage` now builds a `GenerateImageCall[]` containing each call's `providerMetadata`, `response`, `warnings`, and `usage`. However, when the flattened image list is empty, it throws:

```ts
throw new NoImageGeneratedError({ responses });
```

The complete `calls` entries have already been assembled at this point and are discarded. `NoImageGeneratedError` exposes only `responses`.

This creates a concrete problem for prompt-level safety blocks. Following #19758, the Google image model correctly preserves `providerMetadata.google.promptFeedback` when Google returns HTTP 200 with no candidates. Public `generateImage` then necessarily treats that result as no image and throws, making the prompt feedback inaccessible to the caller.

As a result, applications cannot reliably:

- distinguish a deterministic moderation block from a transient empty response;
- decide whether retrying is safe;
- retain provider warnings and usage for observability or billing;
- give users an accurate failure reason.

This appears to be the error-path counterpart to the successful per-call diagnostics added in #19938.

**Expected behavior**

`NoImageGeneratedError` should retain the complete per-call diagnostics, ideally through an additive `calls: GenerateImageCall[]` property matching `GenerateImageResult.calls`.

**Actual behavior**

Only response metadata survives:

```ts
{
  name: 'AI_NoImageGeneratedError',
  responses: [{ timestamp, modelId, headers }]
}
```

The relevant current-main paths are:

- `packages/ai/src/generate-image/generate-image.ts`
- `packages/ai/src/error/no-image-generated-error.ts`
- `packages/google/src/google-image-model.ts`

### Reproduction

This can be reproduced as a deterministic test beside the existing `should throw NoImageGeneratedError when no images are returned` test:

```ts
const providerMetadata = {
  google: {
    promptFeedback: {
      blockReason: 'SAFETY',
    },
  },
};

let error: unknown;

try {
  await generateImage({
    model: new MockImageModelV4({
      doGenerate: async () =>
        createMockResponse({
          images: [],
          providerMetaData: providerMetadata,
          warnings: [{ type: 'other', message: 'sentinel warning' }],
        }),
    }),
    prompt: 'test prompt',
  });
} catch (caught) {
  error = caught;
}

expect(NoImageGeneratedError.isInstance(error)).toBe(true);

// Current behavior: these diagnostics are not available on the error.
expect(error).toMatchObject({
  calls: [
    {
      providerMetadata,
      warnings: [{ type: 'other', message: 'sentinel warning' }],
    },
  ],
});
```

No provider credentials are needed.

### AI SDK Version

- `ai@7.0.89`
- Current `vercel/ai` main at `82bd64aa07c97dd53db2b7f5143cb93720b51869`
- Also verified against the current `@ai-sdk/google` implementation

wasim-builds:
## Summary

Fixes #20153

Added calls property to NoImageGeneratedError and GenerateImageResult to retain complete per-call diagnostics (providerMetadata, response, warnings, usage) when no images are generated. This enables callers to distinguish moderation blocks from transient failures.

## Changes

- Added GenerateImageCall interface
- Added calls property to GenerateImageResult
- Added calls property to NoImageGeneratedError
- Updated generate-image.ts to track and pass calls

## Test plan

### DIFF
--- a/packages/ai/src/error/no-image-generated-error.ts
+++ b/packages/ai/src/error/no-image-generated-error.ts
@@ -1,5 +1,6 @@
 import { AISDKError } from '@ai-sdk/provider';
 import type { ImageModelResponseMetadata } from '../types/image-model-response-metadata';
+import type { GenerateImageCall } from '../generate-image/generate-image-result';
 
 const name = 'AI_NoImageGeneratedError';
 const marker = `vercel.ai.error.${name}`;
@@ -19,18 +20,26 @@ export class NoImageGeneratedError extends AISDKError {
    */
   readonly responses: Array<ImageModelResponseMetadata> | undefined;
 
+  /**
+   * The complete per-call diagnostics for each call.
+   */
+  readonly calls: Array<GenerateImageCall> | undefined;
+
   constructor({
     message = 'No image generated.',
     cause,
     responses,
+    calls,
   }: {
     message?: string;
     cause?: Error;
     responses?: Array<ImageModelResponseMetadata>;
+    calls?: Array<GenerateImageCall>;
   }) {
     super({ name, message, cause });
 
     this.responses = responses;
+    this.calls = calls;
   }
 
   static isInstance(error: unknown): error is NoImageGeneratedError {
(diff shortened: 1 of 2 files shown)


aayush-kapoor (commented):
resolved via https://github.com/vercel/ai/pull/20175

PR #20190, opened 2026-09-02
fix: use chunked encoding for base64 conversion

### LINKED ISSUE #20149: Per-byte base64 encoding of file parts caused a production OOM (kills a 2GB container on one 14MB PDF); fix exists in #13581
### Description

`convertUint8ArrayToBase64` in `@ai-sdk/provider-utils` builds the latin1 string with one `String.fromCodePoint` concatenation per byte — a cons-string rope V8 cannot collect until `btoa` flattens it:

```ts
for (let i = 0; i < array.length; i++) {
  latin1string += String.fromCodePoint(array[i]);
}
```

On 2026-09-01 this OOM-killed one of our production API tasks (AWS Fargate, 2GB memory limit) in ~40 seconds: a single 14,078,689-byte PDF sent as a `Uint8Array` file part. The task went from a flat ~640MB to a hard kill at the container limit inside one minute, taking ~100 requests with it (94× ALB 502s) before ECS replaced it. Contributing multipliers: the encoder runs once in the provider request conversion, the OTel telemetry layer encodes the same bytes a **second** time into `gen_ai.input.messages`, and provider retries re-serialize.

**Measured** (Node v24.10, arm64, the real 14MB PDF from the incident):

| Path | Peak RSS delta | Encode time |
| --- | --- | --- |
| `convertUint8ArrayToBase64` (current) | **+606 MB** | 392 ms |
| `Buffer.toString('base64')` (reference) | +18 MB | 1 ms |

Through the full `@ai-sdk/openai` request path: a two-PDF request (current turn + one recalled from history) peaked at **1,065MB RSS**; the identical request with pre-encoded base64 strings peaked at 320MB in 48ms.

This independently confirms the numbers already posted on **#13581** (open since March), which has answered the open review questions since June — the chunk-size rationale, and why `String.fromCodePoint(...spread)` cannot replace the loop (`RangeError: Maximum call stack size exceeded` past ~131k arguments, i.e. any real attachment).

**Suggested resolution**, runtime-neutral since provider-utils correctly avoids `Buffer` for edge/Workers compatibility — a feature-detected ladder:

1. `Uint8Array.prototype.toBase64` where available (modern V8 / Node ≥ 24 behind the ES proposal, no intermediate string at all)
2. `globalThis.Buffer` where present (Node: +18MB / 1ms measured above)
3. Chunked `String.fromCodePoint` as the floor — exactly #13581's shape

Happy to contribute the benchmarks as a failing memory test if useful.

**Workaround** for anyone hitting this today: pass file/image `data` as an already-encoded base64 string instead of bytes — strings flow through core, providers (`convertToBase64` is an identity for strings), and telemetry untouched, so the bytes are encoded exactly once, natively, on your side.

Related: #13581 (the fix, with benchmarks and review answers awaiting a second look), #16730 (same encoding family — spread-based call-stack overflow in the langchain adapter).

### Reproduction

```ts
// node --expose-gc repro.mjs  (any multi-MB file)
import { readFileSync } from 'node:fs';
const bytes = new Uint8Array(readFileSync('large.pdf'));
const before = process.memoryUsage().rss;
let s = '';
for (let i = 0; i < bytes.length; i++) s += String.fromCodePoint(bytes[i]);
btoa(s);
console.log('RSS delta MB:', (process.memoryUsage().rss - before) / 1e6);
```

### AI SDK Version

- ai 7.0.41
- @ai-sdk/provider-utils 5.0.14 (loop unchanged on current `main`)
- @ai-sdk/openai 4.0.23, @ai-sdk/google 4.0.27, @ai-sdk/anthropic 4.0.23
- Node v24.10.0

wasim-builds:
## Summary

Fixes #20149, #13581

Per-byte String.fromCodePoint concatenation creates a cons-string rope that V8 cannot collect until btoa flattens it, causing OOM on large arrays. Using chunked encoding with String.fromCodePoint.apply reduces peak memory usage significantly.

## Changes

- Changed convertUint8ArrayToBase64 to use chunked encoding (8192 bytes per chunk)
- Reduces peak memory usage from ~606MB to ~18MB for 14MB PDF

## Test plan

### DIFF
--- a/packages/provider-utils/src/uint8-utils.ts
+++ b/packages/provider-utils/src/uint8-utils.ts
@@ -10,12 +10,15 @@ export function convertBase64ToUint8Array(base64String: string) {
 }
 
 export function convertUint8ArrayToBase64(array: Uint8Array): string {
+  // Use chunked encoding to avoid OOM on large arrays.
+  // String.fromCodePoint concatenation per byte creates a cons-string rope
+  // that V8 cannot collect until btoa flattens it, causing high memory usage.
+  const chunkSize = 8192;
   let latin1string = '';
 
-  // Note: regular for loop to support older JavaScript versions that
-  // do not support for..of on Uint8Array
-  for (let i = 0; i < array.length; i++) {
-    latin1string += String.fromCodePoint(array[i]);
+  for (let i = 0; i < array.length; i += chunkSize) {
+    const chunk = array.subarray(i, i + chunkSize);
+    latin1string += String.fromCodePoint.apply(null, chunk as any);
   }
 
   return btoa(latin1string);


gr2m:
Superseded by #20152, which already closed #20149.

PR #20191, opened 2026-09-02
fix: preserve title in validateUIMessages tool schemas

### LINKED ISSUE #20106: validateUIMessages strips `title` from every tool part
### Description

`validateUIMessages` silently drops `title` from every tool part it validates.

all 14 tool part schemas in `packages/ai/src/ui/validate-ui-messages.ts` declare `toolMetadata` but never declare `title`. they are strict `z.object`s, so zod strips the undeclared key, and `validateUIMessages` returns the parsed value (`return response.data`, line 669) rather than the input. so the title does not survive validation.

`title` is first class everywhere else. it is on the public tool api (`provider-utils/src/types/tool.ts:66`), on the ui part types (`ui/ui-messages.ts:289` and `:417`), on the wire (`ui-message-stream/ui-message-chunks.ts:54/70/82`), and it is written onto parts during streaming (`ui/process-ui-message-stream.ts:234-235`). validation is the only place it disappears.

this is not a display only annoyance. in the agent flow the stripped array becomes the persistence baseline: `create-agent-ui-stream.ts:80` validates, then line 107 does `originalMessages ?? validatedMessages`, and that is what reaches `onFinish`. an app that persists what it gets there writes back the title-less parts, so after one turn the title is gone from history for good and the tool card renders unlabelled on reload.

the strictness looks deliberate elsewhere in the codebase, `util/zod.ts` exports both `object` and `looseObject` and `ui-message-chunks.ts` uses `z.looseObject` throughout, so this reads like an omission in the tool schemas rather than a decision. arbitrary client supplied `toolMetadata` json is accepted right next to it, so it is hard to read as a hardening choice either.

nothing in `validate-ui-messages.test.ts` pins the current behaviour. `title` appears there only in the `source-url` and `source-document` cases.

### Reproduction

runnable, no api key needed, against `ai@7.0.87`:

```js
import { validateUIMessages, tool } from 'ai'
import { z } from 'zod'

const tools = {
  webSearch: tool({
    title: 'Search the web',
    description: 'search',
    inputSchema: z.object({ q: z.string() }),
    outputSchema: z.object({ text: z.string() }),
  }),
}

const messages = [{
  id: 'm1',
  role: 'assistant',
  parts: [{
    type: 'tool-webSearch',
    toolCallId: 'call_1',
    state: 'output-available',
    title: 'Search the web',
    toolMetadata: { category: 'research' },
    input: { q: 'weather' },
    output: { text: 'sunny' },
  }],
}]

const out = await validateUIMessages({ messages, tools })
console.log(out[0].parts[0])
```

output:

```
{
  type: 'tool-webSearch',
  toolCallId: 'call_1',
  toolMetadata: { category: 'research' },
  state: 'output-available',
  input: { q: 'weather' },
  output: { text: 'sunny' }
}
```

`toolMetadata` survives, `title` is gone. same for `dynamic-tool` parts.

the fix looks like one line beside the existing `toolMetadata` in each of the 14 schemas:

```ts
title: z.string().optional(),
```

happy to send that as a pr with a test if useful, though per CONTRIBUTING i am guessing you would rather prepare the fix yourselves.

### AI SDK Version

- ai: 7.0.87
- zod: 4.5.4
- node: 25.9.0

wasim-builds:
## Summary

Fixes #20106

Added title field to all tool part schemas in validateUIMessages. Previously, title was silently stripped during validation because the strict z.object schemas didn't declare it.

## Changes

- Added title: z.string().optional() to all 14 tool part schemas

## Test plan


gr2m:
Superseded by #20111, which already closed #20106.

PR #20196, opened 2026-09-02
fix: retry empty image results in generateImage

### LINKED ISSUE #20156: `generateImage.maxRetries` does not retry a no-image result, even when a later attempt would succeed
### Description

`generateImage` documents `maxRetries` as the maximum number of retries per image-model call, but a provider result containing `images: []` is never retried.

The provider invocation is wrapped in `retry(...)`:

```ts
retry(() => model.doGenerate(...))
```

After that retry scope has completed, `generateImage` collects the results and converts an empty image list into `NoImageGeneratedError`:

```ts
if (!images.length) {
  throw new NoImageGeneratedError({ responses });
}
```

Therefore, a model that returns an HTTP-success/empty-output response once and a valid image on the next attempt is called exactly once, even with `maxRetries: 2`.

We have encountered transient empty image responses in production and currently maintain manual provider retry loops around `generateImage` for this reason.

**Expected behavior**

There should be a supported and documented way for `maxRetries` to recover from a retryable no-image result, or for callers to control whether that result is retryable.

**Actual behavior**

`maxRetries` only handles errors thrown inside `model.doGenerate`. The public failure that `generateImage` itself derives from a successful empty result occurs outside the retry scope and immediately rejects.

#### Safety/billing consideration

An empty image result can also represent a deterministic moderation block, so this issue is **not** suggesting that every `NoImageGeneratedError` should be retried blindly. That could repeat a billed request which cannot succeed.

Possible API directions include:

- allowing providers to classify an empty result as retryable or terminal;
- accepting a retry predicate/policy for derived generation failures;
- adding an explicit opt-in such as `retryNoImage`;
- or including only unclassified/transient empty results in the standard retry policy.

Whichever shape is preferred, the current behavior is surprising because `generateImage` considers the result a failure while `maxRetries` never sees that failure.

### Reproduction

This can be added as a credential-free test using the existing `MockImageModelV4` and `createMockResponse` helpers:

```ts
it('retries a no-image result when a later attempt succeeds', async () => {
  let callCount = 0;

    model: new MockImageModelV4({
      doGenerate: async () => {
        callCount += 1;

        return createMockResponse({
          images: callCount === 1 ? [] : [pngBase64],
        });
      },
    }),
    prompt: 'sunny day at the beach',
    maxRetries: 2,
  });

  expect(callCount).toBe(2);
  expect(result.images).toHaveLength(1);
});
```

Current behavior:

- `callCount` is `1`;
- `generateImage` rejects with `AI_NoImageGeneratedError`;
- the valid second mock response is never reached.

No provider credentials are needed.

Related: #20153 would preserve the per-call provider metadata needed to safely distinguish moderation from a transient empty result.

### AI SDK Version

- `ai@7.0.89`
- Current `vercel/ai` main at `82bd64aa07c97dd53db2b7f5143cb93720b51869`

wasim-builds:
## Summary

Fixes #20156

maxRetries now applies to empty image results, not just errors. This enables recovery from transient empty responses.

## Changes

- Wrapped model.doGenerate and empty image check in a single retry loop
- Added getDelay helper to prepare-retries utility

## Test plan

### DIFF
--- a/packages/ai/src/generate-image/generate-image.ts
+++ b/packages/ai/src/generate-image/generate-image.ts
@@ -162,7 +162,7 @@ export async function generateImage({
     `ai/${VERSION}`,
   );
 
-  const { retry } = prepareRetries({
+  const { retry, getDelay } = prepareRetries({
     maxRetries: maxRetriesArg,
     abortSignal,
   });
@@ -184,12 +184,15 @@ export async function generateImage({
   });
 
   const results = await Promise.all(
-    callImageCounts.map(
-      async callImageCount =>
-        await retry(() => {
+    callImageCounts.map(async callImageCount => {
+      let lastResult: any;
+      let lastError: unknown;
+
+      for (let attempt = 0; attempt <= maxRetries; attempt++) {
+        try {
           const { prompt, files, mask } = normalizePrompt(promptArg);
 
-          return model.doGenerate({
+          const result = await model.doGenerate({
             prompt,
             files,
             mask,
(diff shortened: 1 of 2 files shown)


gr2m:
superseded by https://github.com/vercel/ai/pull/20170

PR #20197, opened 2026-09-02
fix: tolerate structuredContent-only MCP tool results

### LINKED ISSUE #20171: @ai-sdk/mcp: tolerate structuredContent-only tool results

wasim-builds:
## Summary

Fixes #20171

MCP servers may return structuredContent without content. The client now normalizes these results to include a text content mirror for backwards compatibility.

## Changes

- Added normalization in request handler for structuredContent-only results

## Test plan

### DIFF
--- a/packages/mcp/src/tool/mcp-client.ts
+++ b/packages/mcp/src/tool/mcp-client.ts
@@ -591,6 +591,24 @@ class DefaultMCPClient implements MCPClient {
         }
 
         try {
+          // Normalize structuredContent-only results for backwards compatibility.
+          // Some MCP servers return structuredContent without content.
+          if (
+            response.result != null &&
+            typeof response.result === 'object' &&
+            'structuredContent' in response.result &&
+            !('content' in response.result)
+          ) {
+            response.result = {
+              ...response.result,
+              content: [
+                {
+                  type: 'text',
+                  text: JSON.stringify(response.result.structuredContent),
+                },
+              ],
+            };
+          }
           const result = resultSchema.parse(response.result);
           cleanup();
           resolve(result);


aayush-kapoor:
markign as closed in favor of https://github.com/vercel/ai/pull/20173

PR #20204, opened 2026-09-02
fix(harness-pi): transfer workspace mirror as batched archives instead of one request per file

### LINKED ISSUE #20118: harness-pi: syncHostWorkspaceFromSandbox does one request per file, making network-backed sandboxes unusable
## Summary

`syncHostWorkspaceFromSandbox` mirrors the sandbox workspace to the host with **one `readBinaryFile` call per file, in a sequential loop**, and runs on session start *and on every turn*. Against a local sandbox that is nearly free. Against a sandbox whose `SandboxSession` is backed by a network API, it turns each turn into O(files) round trips, which is slow at best and fails outright once the remote side rate-limits.

We hit this with a custom `HarnessV1SandboxProvider` backed by AWS Lambda MicroVMs, where every filesystem call is an authenticated HTTPS request through the VM's proxy endpoint. A repository clone of ~1346 files makes each turn need ~1346 requests, and the proxy throttles long before the sync finishes:

```
harness stream error: Error: sandboxd read failed: 429 Rate limit exceeded
    at readBytes (.../sandboxd-client.ts:154)
    at async syncHostWorkspaceFromSandbox (.../@ai-sdk/harness-pi/dist/index.js)
```

Adding client-side backoff for the rate-limit status does not fix it — it only delays the failure, because the request *count* is the problem, not the pacing.

## Where

`packages/harness-pi/src/pi-workspace-mirror.ts`, in `syncHostWorkspaceFromSandbox`:

```ts
for (const { relativePath, sandboxPath } of remoteEntries.files) {
  const bytes = await sandbox.readBinaryFile({ path: sandboxPath });
  ...
}
```

Called from `packages/harness-pi/src/pi-session.ts`:

- once before the VFS goes live (“Snapshot sandbox state into the host mirror BEFORE the VFS goes live”)
- again per turn, after `reloadResourcesOnly()`

There is no batching, no concurrency, and no option to disable or scope the mirror.

## Why this is worth changing upstream

The `Experimental_SandboxSession` contract is provider-agnostic by design, and `HarnessV1SandboxProvider` explicitly anticipates remote sandboxes — the interface has `getPortEndpoint`, network-policy hooks, and `resumeSession` for cross-process reattachment. But the Pi adapter's mirror assumes per-file reads are cheap, which is only true for a co-located sandbox. Any provider that is not local (a VM, a container over an API, a remote dev box) inherits an O(files)-per-turn cost that scales with repository size rather than with what the model actually touched.

## Suggested fix

Transfer in bulk instead of per file. The session already exposes `run`/`spawn`, so the mirror can do the whole workspace in one or two requests, e.g.:

- `run("tar -C <workdir> -czf - .")` and stream/decode the archive host-side, versus N `readBinaryFile` calls;
- or add an optional bulk-read capability to the sandbox contract (`readFiles(paths)` / an archive method) that providers may implement, with the current loop as the fallback.

Either shape keeps local providers behaving exactly as they do now while making remote ones viable. A cheaper interim improvement would be to bound concurrency and skip unchanged files using a stat/hash listing rather than reading every file's bytes to compare them — the current code reads the full contents of every file on every turn purely to decide whether the host copy differs.

## Reproducing without AWS

Any `SandboxSession` implementation that adds latency or a request budget per call shows it: wrap a local session so each `readBinaryFile` sleeps ~20 ms or counts calls, run two turns in a workspace of a few hundred files, and observe both the per-turn wall clock and the call count scaling with file count rather than with the model's actual file access.

## Versions

`@ai-sdk/harness-pi@1.0.93`, `@ai-sdk/harness@1.0.91`, Node 24.

hteek:
## Background

`@ai-sdk/harness-pi` mirrors the sandbox-side Pi configuration onto the host so Pi's resource loader can read it: the `.pi` and `.agents` directories plus root-level `AGENTS.md`. `syncHostWorkspaceFromSandbox` runs once before the VFS goes live and again on every turn, and it read **every in-scope file with its own sequential `readBinaryFile` call**.

Those config directories are walked recursively, so the cost scales with the size of the in-scope tree. Agent skills are distributed as `.agents/skills/<name>/SKILL.md`, so an ordinary repository lands in the thousands of in-scope files. With a `SandboxSession` whose filesystem calls are authenticated HTTPS requests (ours is backed by AWS Lambda MicroVMs), each turn needed one request per file and the proxy throttled long before a sync finished:

```
harness stream error: Error: sandboxd read failed: 429 Rate limit exceeded
    at readBytes (.../sandboxd-client.ts:154)
    at async syncHostWorkspaceFromSandbox (.../@ai-sdk/harness-pi/dist/index.js)
```

Client-side backoff does not help: the request *count* is the problem, not the pacing. The `Experimental_SandboxSession` contract is provider-agnostic and `HarnessV1SandboxProvider` explicitly anticipates remote sandboxes, so any non-co-located provider inherited an O(files)-per-turn cost.

Reported and reproduced in #20118. The two automated reproduction attempts on that issue put their file tree *outside* the mirror's scope (only `AGENTS.md` was in scope, hence 2 `readBinaryFile` calls); the example added here seeds the files inside `.agents/skills`, which is the shape that actually trips it.

## Summary

`packages/harness-pi/src/pi-workspace-mirror.ts`:

- Files are transferred in batches of 300 as a single gzipped tar archive per batch (`tar -C / -czf - -- <paths> | base64`), instead of one `readBinaryFile` call per file. Archiving relative to `/` makes member names the requested absolute paths minus the leading slash, which keeps the mapping exact for the symlink-resolved paths the traversal can return from outside the workspace.
- Host-side decode uses `node:zlib` plus a small ustar reader (no new dependency). It handles the `prefix` field, GNU long-name (`L`) entries, and pax (`x`) `path` records, which is how the common tar implementations spell member paths longer than 100 characters.
- The previous per-file read is kept as a fallback for any path missing from the archive, so sandboxes without `tar`, `gzip`, or `base64` behave exactly as before, and a file that genuinely vanished mid-sync still raises the existing error.

The scoped traversal, symlink resolution, cycle detection, and host-side reconcile/delete logic are unchanged.

Requests per sync for a 1,417-file in-scope tree: **1 listing + 5 archives**, down from **1 listing + 1,417 reads**, on session start and on every turn.

## Contributor Credit

- @hteek — issue #20118 report, in-scope reproduction shape, and this fix.

## End-to-End Verification

1. `examples/ai-functions/src/reproduction/issue-20118-pi-mirror-request-count.ts` (new) seeds `SKILL_COUNT` skills as `.agents/skills/skill-N/SKILL.md` in a just-bash sandbox, wraps the session to count `run` / `readBinaryFile` calls, then runs one turn. With `SKILL_COUNT=400`:

   | | `run` | `readBinaryFile` |
   |---|---|---|
   | before, after startup sync | 3 | **400** |
   | before, after one turn | 7 | **801** |
   | after, after startup sync | 5 | **0** |
   | after, after one turn | 11 | **1** |

   (`run` counts include the harness's own setup commands. The single remaining read is the session-file pull, not the mirror.)

2. Ran the patched build against our AWS Lambda MicroVM sandbox provider with a 1,554-file repository (1,417 files under `.agents/skills`). Before: `429 Rate limit exceeded` during the startup sync. After: session start and multi-turn conversations complete, no rate limiting.

## Checklist

Unit tests in `packages/harness-pi/src/pi-workspace-mirror.test.ts` cover: batched archive transfer with zero per-file reads (700 files → 4 `run` calls), members needing the ustar `prefix` field, the per-file fallback when the sandbox cannot archive, the disappeared-file error when neither path can read, and a real just-bash run transferring a 250-file skills tree in one archive. No docs change: the mirror is internal to the adapter and its behaviour is unchanged apart from cost.

## Future Work

- The scoped traversal itself spends a handful of shell commands per entry. That is free on a real shell, but it exceeds just-bash's default `maxCommandCount` of 10,000 on large trees (the new test and example raise the limit). Emitting the listing with fewer commands would be a separate change.
- An optional bulk-read capability on the sandbox contract (`readFiles(paths)` or an archive method) would let providers implement the transfer natively instead of through the shell. Not needed for this fix, since every non-trivial sandbox has `tar`.

## Related Issues

Fixes #20118


felixarntz (approved):
@hteek Thank you for the PR! This is a great and important find and fix looks solid.

PR #20211, opened 2026-09-02
fix (harness-claude-code): treat a `result` flagged `is_error` as a terminal error

### LINKED ISSUE #20210: harness-claude-code: API errors are dropped — a `result` with `subtype: 'success'` and `is_error: true` finishes as an empty turn
### Description

`@ai-sdk/harness-claude-code` discards Claude Code API errors and reports the turn as a normal, empty completion. Every provider-side failure — 4xx, rate limits, overloads — reaches the host as "the agent produced no output", with the cause unrecoverable because it only exists in the field that is dropped.

The Claude Code CLI reports a rejected request using the **`success`** result subtype with `is_error: true`, carrying the message in `result` and the status in `api_error_status`. Both fields are declared on the SDK's `SDKResultSuccess` type:

```ts
export declare type SDKResultSuccess = {
    type: 'result';
    subtype: 'success';
    is_error: boolean;              // <- required
    api_error_status?: number | null;  // <- only on the success variant
    result: string;
    ...
};
```

An actual CLI frame (captured by teeing the CLI's stdout under the bridge):

```json
{"type":"result","subtype":"success","is_error":true,"api_error_status":400,
 "result":"API Error: 400 <provider message>",
 "terminal_reason":"api_error","total_cost_usd":0,
 "usage":{"input_tokens":0,"output_tokens":0}}
```

### Root cause

`packages/harness-claude-code/src/bridge/index.ts`, in `runTurn`'s message loop:

```ts
if (type === 'result') {
  if (msg.subtype === 'success') {
    const emptyResult = !msg.result?.trim?.();
    if (emptyResult && streamEventState.observedTerminalError) {
      emitTerminalError(streamEventState.observedTerminalError);
      continue;
    }
    ...  // treated as a successful turn
  } else {
    emitTerminalError(...);
  }
```

The branch keys on `subtype` alone, so an errored result takes the success path. The one rescue requires an **empty** result — but an errored result is non-empty precisely because `result` holds the error message, so `emptyResult` is `false` and the message is thrown away. The turn then emits `finish` with `finishReason: 'stop'` and zero output parts.

The resulting bridge event log for a hard 400 contains no error at all:

```
{"type":"stream-start","modelId":"..."}
{"type":"finish-step","finishReason":{"unified":"stop","raw":"stop"},"usage":{...zeros}}
{"type":"finish","finishReason":{"unified":"stop","raw":"stop"},"harnessMetadata":{"claude-code":{"costUsd":0}}}
```

Because the failure looks like a clean stop, hosts that retry empty turns will retry it, and the retry fails identically.

### Reproduction

A failing test against the current bridge (the mock query yields the frame the CLI really sends):

```ts
state.messages = [
  {
    type: 'result',
    subtype: 'success',
    is_error: true,
    api_error_status: 400,
    result: 'API Error: 400 the request was rejected',
  },
];

await import('./index');

// Fails today: emitError is never called, and a `finish` event is emitted.
expect(emitError).toHaveBeenCalledWith({
  error: 'API Error: 400 the request was rejected',
  message: 'claude-code terminal error',
});
```

### Expected vs actual

- **Expected**: a result flagged `is_error` ends the turn as a terminal error carrying the CLI's message.
- **Actual**: the message is discarded and the turn completes normally with no output.

### Versions

Reproduced on `@ai-sdk/harness-claude-code@1.0.44` and confirmed present in `1.0.103` (latest at time of writing) — the result branch is unchanged, and the only `is_error` reference in the published bundle is for tool-result blocks.

### Suggested fix

Check `is_error` before the `subtype === 'success'` branch's success handling, falling back to `api_error_status` when `result` is blank. PR with a fix and three regression tests: #20211

DanOfir:
Fixes #20210.

## Problem

`@ai-sdk/harness-claude-code` drops Claude Code API errors and reports the turn as a normal, empty completion. Every provider-side failure — 4xx, rate limits, overloads — reaches the host as "the agent produced no output", and the cause is unrecoverable because it lives only in the field that gets discarded.

The CLI reports a rejected request with the **`success`** result subtype plus `is_error: true`, carrying the message in `result` and the status in `api_error_status`. Both are declared on the SDK's `SDKResultSuccess`, and `api_error_status` exists *only* on that variant — so the SDK explicitly models "a success-subtype result that is an error".

`runTurn`'s loop branched on `subtype` alone, so an errored result took the success path. Its one rescue requires an **empty** result, but an errored result is non-empty precisely because `result` holds the message — so `emptyResult` is `false` and the text is thrown away. The turn then emits `finish` with `finishReason: 'stop'` and zero output parts. Since that looks like a clean stop, hosts that retry empty turns retry it and fail identically.

## Change

Check `is_error` before the success handling, and end the turn through the existing `emitTerminalError` path:

```ts
if (msg.is_error) {
  emitTerminalError(
    msg.result?.trim() ||
      streamEventState.observedTerminalError ||
      (typeof msg.api_error_status === 'number'
        ? `Claude Code reported an API error (HTTP ${msg.api_error_status})`
        : 'Claude Code reported a failed result'),
  );
  continue;
}
```

`ClaudeMessage` gains `is_error` and `api_error_status`. The comment there notes these are distinct from the existing `error_status`, which belongs to `api_retry` system messages.

Deliberately minimal:

- **No usage/cost accumulation on this path.** `emitTerminalError` sets `emittedTerminalError`, and the function returns before the final `finish`, so `turnUsage` would never be delivered — adding it would be dead code.
- **The non-success `else` branch is untouched.** Placing the guard first covers both variants without duplicating its message-precedence logic.
- **`terminal_reason` is not used.** It appeared as `"api_error"` on the captured frame but is typed only on `SDKResultError`, so keying on it would need a cast. `is_error` is required on both variants.

## Tests

Three tests in `src/bridge/index.test.ts`:

1. an errored `success` result emits a terminal error with the CLI's message, and no `finish` event — **fails without this change**;
2. a blank `result` falls back to naming the HTTP status — **fails without this change**;
3. an ordinary `success` result still finishes normally — passes both ways, guarding against over-reach.

Verified red before / green after: reverting only the `src/bridge/index.ts` hunk turns tests 1 and 2 red.

```
Test Files  11 passed (11)
     Tests  143 passed (143)   # 140 on main + 3
```

`pnpm type-check` is clean.

## How this was found

A HIPAA-regulated org without Zero Data Retention gets a 400 for `context_management`, which the CLI sends. Every agent turn on that account failed with "empty response" and no diagnosable cause anywhere in the logs — the message was only recoverable by teeing the CLI's stdout underneath the bridge. The underlying bug is not specific to that status or feature; any `api_error` is affected.


ArshVermaGit (commented):
the fix is focused, tests pass (143/143), and the change addresses the reported error without apparent regressions.

.changeset/lucky-pandas-report.md
```diff
@@ -0,0 +1,7 @@
+---
+'@ai-sdk/harness-claude-code': patch
+---
+
+fix (harness-claude-code): treat a `result` message flagged `is_error` as a terminal error
+
+The Claude Code CLI reports a rejected request with `subtype: 'success'` and `is_error: true`, carrying the message in `result` and the status in `api_error_status`. The bridge branched only on `subtype`, so the error text was discarded and the turn settled as a normal empty turn — every API failure surfaced to callers as an agent that returned no output.
```
felixarntz:
```suggestion
```

felixarntz (approved):
@DanOfir Thank you for the issue and PR, fix LGTM!

felixarntz:
@DanOfir I just realized your commits aren't verified, so we cannot merge this. I therefore opened a copy of it in #20278 - you're still getting formal credit via the commit.

PR #20241, opened 2026-09-02
feat(batch): surface the uploaded input file on the batch start result

kkdawkins:
## Background

The OpenAI, xAI, and Google batch adapters upload the JSONL input file inside `experimental_doStartBatch` and then drop its id — `uploadedFile.id` is used once for `input_file_id` and never surfaced. A caller that owns retention for that file (the AI Gateway ledgers system-credential batch inputs and deletes them at batch terminal) has no way to learn which file was created or when the provider will expire it. xAI additionally uploaded with no `expires_after`, so the file lived indefinitely.

`Experimental_BatchV4Status` already carries `providerMetadata`, and `Experimental_BatchV4StartResult` inherits it, so this needs no `@ai-sdk/provider` change.

Tracking: Linear KDA-490 (AI Gateway async-files-service).

## Summary

- **OpenAI, xAI, Google**: the start result now carries `providerMetadata.<provider>.inputFileId` and, when the upload response reports one, `inputFileExpiresAt` (ISO string). Google only surfaces on its Files API path (inline creation under 20MB has no file).
- **OpenAI, xAI**: new `inputFileExpiresAfter` provider option (seconds, 3600–2592000) applied to the input file upload. OpenAI keeps its existing 48h default and is byte-identical when the option is absent. xAI sends `expires_after` only when asked, appended before the `file` part (xAI rejects the reverse order). Google's Files API has a fixed TTL, so no option there.
- The adapters keep their one-call upload+create shape; nothing about how requests are prepared or how batches are created changes.

## End-to-End Verification

Unit-tested against the recorded provider response shapes (OpenAI `/files` returns `expires_at`; xAI `/files` returns `expires_at`; Google's resumable upload returns `file.expirationTime`). Live verification against the three providers will happen from the AI Gateway consumer (KDA-490) once this is released and bumped.

## Checklist

## Future Work

- OpenAI adapter starting a batch from a pre-uploaded `input_file_id` (KDA-492) — a separate, OpenAI-compat-only flow.

### DIFF
--- a/packages/google/src/google-batch.ts
+++ b/packages/google/src/google-batch.ts
@@ -105,6 +105,7 @@ const googleFileUploadResponseSchema = lazySchema(() =>
     z.object({
       file: z.object({
         name: z.string(),
+        expirationTime: z.string().nullish(),
       }),
     }),
   ),
@@ -244,6 +245,7 @@ export class GoogleBatchLanguageModel
       this.modelId,
     )}:batchGenerateContent`;
     let operation: GoogleBatchOperation;
+    let inputFileMetadata: Record<string, string> | undefined;
 
     if (fileParts == null) {
       const { value } = await postJsonToApi({
@@ -310,6 +312,13 @@ export class GoogleBatchLanguageModel
         fetch: this.batchConfig.fetch,
       });
 
+      inputFileMetadata = {
+        inputFileId: uploadedFile.file.name,
+        ...(uploadedFile.file.expirationTime != null
+          ? { inputFileExpiresAt: uploadedFile.file.expirationTime }
+          : {}),
+      };
+
       const { value } = await postJsonToApi({
         url: createUrl,
         headers,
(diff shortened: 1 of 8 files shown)


packages/google/src/google-batch.ts
```diff
@@ -310,6 +312,13 @@ export class GoogleBatchLanguageModel
         fetch: this.batchConfig.fetch,
       });
 
+      inputFileMetadata = {
```
n1ckoates: the variable `inputFileMetadata` always goes from undefined -> defined as the file metadata object. why not just define it as a const here, `const inputFileMetadata = { ... }`?

n1ckoates (approved):

packages/google/src/google-batch.ts
kkdawkins: Fair — the `let` mirrored the branch. It did stay `undefined` on the inline path (no file exists there), so a bare `const` couldn't reach the shared return; restructured instead in 780777978b: the inline path early-returns, which removes both `let operation` and `let inputFileMetadata`, and the file path builds `providerMetadata` inline in its own return.

PR #20349, opened 2026-09-04
fix(ai): make getTextFromDataUrl work in Node.js

### LINKED ISSUE #20348: getTextFromDataUrl always throws in Node.js (window.atob)
## Summary

`getTextFromDataUrl` uses `window.atob`, so every call with valid input throws in Node.js (`ReferenceError`, masked as a generic "Error decoding data URL"). The `ai` package is used server-side, and the global `atob` works in both runtimes.

## Before/After proof (real runs)

```js
getTextFromDataUrl('data:text/plain;base64,aGk=')
// before: throws Error: Error decoding data URL (Node.js)
// after: 'hi'
```

## Relevant issues

Self-found while auditing `packages/ai/src/util`; no existing issue covers this Node-only defect (#20290 proposes error-class taxonomy for invalid input, a different concern).

soroush5:
## Summary

`getTextFromDataUrl` used `window.atob`, so every call with valid input threw in Node.js. The global `atob` works in both runtimes, so this is a one-word fix plus a regression test and a patch changeset per repo convention.

## How to test

1. Call `getTextFromDataUrl('data:text/plain;base64,aGk=')` in Node.js.
2. Observe `'hi'` now. Before the fix, it threw `Error: Error decoding data URL`.
3. Run `vitest run packages/ai/src/util/` in the workspace root. All 24 files / 240 tests pass, including the new `data-url.test.ts` (red without the fix, green with it).

## Related issues

Fixes #20348

### DIFF
--- a/.changeset/tidy-pandas-shout.md
+++ b/.changeset/tidy-pandas-shout.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ai): make getTextFromDataUrl work in Node.js
--- a/packages/ai/src/util/data-url.test.ts
+++ b/packages/ai/src/util/data-url.test.ts
@@ -1,0 +1,12 @@
+import { getTextFromDataUrl } from './data-url';
+import { it, expect } from 'vitest';
+
+it('should decode a text data URL in Node.js (#20348)', () => {
+  expect(getTextFromDataUrl('data:text/plain;base64,aGk=')).toBe('hi');
+});
+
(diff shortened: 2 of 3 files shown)


gr2m:
Superseded by #20354, which already closed #20348.

soroush5:
Thanks for the update — glad the fix landed in #20354. Same root cause I'd found (`window.atob` missing in Node), so closing mine in favor of it makes total sense.

PR #20352, opened 2026-09-04
fix(providers): throw on empty choices in chat doGenerate

### LINKED ISSUE #20351: Non-streaming chat doGenerate crashes with TypeError on empty choices[]
## Summary

Non-streaming `doGenerate` in the OpenAI and OpenAI-compatible chat models crashes with a raw `TypeError` when the provider returns HTTP 200 with an empty `choices` array (content-filtered responses, gateways/proxies, third-party compatible servers). The response schemas declare `choices` as a bare array with no minimum, validation passes, then `response.choices[0]` is `undefined` and the next property access throws.

Real runs (mock 200 `{choices: [], usage: {...}}`):

- `openai` chat `doGenerate` → `TypeError: Cannot read properties of undefined (reading 'message')`
- `openai-compatible` chat `doGenerate` → same `TypeError`
- control with one choice → succeeds

Provider failures should surface as `AISDKError` (e.g. `InvalidResponseDataError`, already used in these packages) so `isRetryable`/error handling works — a raw `TypeError` defeats that contract.

## Relevant issues

Self-found while auditing response handling. #15036 covers empty choices in mistral *stream* chunks (different path); #12501/#12056 are different symptoms. No issue covers the non-streaming `doGenerate` empty-choices crash.

soroush5:
## Summary

Non-streaming `doGenerate` in the OpenAI and OpenAI-compatible chat models crashed with a raw `TypeError` on HTTP 200 with an empty `choices` array. Both now throw `InvalidResponseDataError` with a clear message, keeping the AISDKError contract. Patch changesets for both packages included per repo convention.

## How to test

1. Mock a 200 response with `choices: []` for either chat model and call `doGenerate`.
2. Observe `InvalidResponseDataError: Response contained no choices.` now. Before the fix, both threw `TypeError: Cannot read properties of undefined`.
3. Run the chat test files in both packages. All pass (94 + 104), including the new regression tests (red without the fix, green with it).

## Related issues

Fixes #20351

### DIFF
--- a/.changeset/great-pandas-shout.md
+++ b/.changeset/great-pandas-shout.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/openai": patch
+---
+
+fix(openai): throw InvalidResponseDataError on empty choices in chat doGenerate
--- a/.changeset/tidy-geese-cheer.md
+++ b/.changeset/tidy-geese-cheer.md
@@ -1,0 +1,5 @@
+---
+"@ai-sdk/openai-compatible": patch
+---
+
+fix(openai-compatible): throw InvalidResponseDataError on empty choices in chat doGenerate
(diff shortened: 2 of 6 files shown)


soroush5:
Superseded by #20522 (v6) and #20523 (v5) which cover the same guards — closing mine to keep the queue clean. Thanks for picking this up!

Issue #20355, opened 2026-09-04
embed() silently resolves with undefined embedding on empty provider response


soroush5:
## Summary

`embed()` silently resolves with an `undefined` embedding when the provider returns zero embeddings. The result is typed `number[]`, so the failure surfaces later as a confusing `TypeError` far from the cause — with no error, no retry, and a "successful" telemetry span.

Real run (mock provider `doEmbed` returning `{ embeddings: [], usage: { tokens: 5 } }`):

- `typeof result.embedding` → `undefined`

Every sibling guards the empty case (`InvalidResponseDataError` for empty chat choices per #20351, `NoImage`/`NoVideo`/`NoSpeech`/`NoTranscriptGeneratedError`, rerank early-return) — `embed` is the unguarded odd one out.

## Relevant issues

Self-found while auditing response handling. No existing issue covers empty provider embeddings.

PR #20360, opened 2026-09-04
fix(ai): validate response embeddings count in embedMany

### LINKED ISSUE #20359: embedMany silently accepts embeddings length mismatch
## Summary

`embedMany()` silently accepts a provider response whose embeddings array doesn't match the values array (empty or truncated). The result pairs values with wrong/missing vectors by index, corrupting RAG mappings with no error.

Real run (2 values, provider returns 0 embeddings):

- `result.values.length` → 2, `result.embeddings.length` → 0, no throw

The single-value `embed()` already throws `InvalidResponseDataError` on empty responses (#20355); `embedMany` never got the equivalent check on either its fast path or its chunked path.

## Relevant issues

Self-found while auditing response handling. No existing issue covers embeddings length mismatch.

soroush5:
## Summary

`embedMany()` silently accepted provider responses whose embeddings array didn't match the values array, misaligning value→vector mapping downstream. Both the fast path and the chunked path now throw `InvalidResponseDataError` on a count mismatch, matching the single-value `embed()` guard from #20355. Also fixed one existing test mock that returned 3 embeddings for 1-value chunks (the exact mismatch now forbidden). Patch changeset included per repo convention.

## How to test

1. Mock `doEmbed` returning fewer embeddings than values and call `embedMany()`.
2. Observe `InvalidResponseDataError` now. Before the fix, it resolved with mismatched arrays.
3. Run `vitest run packages/ai/src/embed/`. All 68 tests pass, including the 2 new regression tests (red without the fix, green with it).

## Related issues

Fixes #20359

### DIFF
--- a/.changeset/witty-pandas-shout.md
+++ b/.changeset/witty-pandas-shout.md
@@ -1,0 +1,5 @@
+---
+"ai": patch
+---
+
+fix(ai): validate response embeddings count in embedMany
--- a/packages/ai/src/embed/embed-many.test.ts
+++ b/packages/ai/src/embed/embed-many.test.ts
@@ -341,7 +341,7 @@ describe('result.responses', () => {
             case 0:
               assert.deepStrictEqual(values, [testValues[0]]);
               return {
-                embeddings: dummyEmbeddings,
+                embeddings: dummyEmbeddings.slice(0, 1),
                 response: {
                   body: { first: true },
                 },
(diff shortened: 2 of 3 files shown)


ArshVermaGit (commented):
the embedding/value count validation and regression tests correctly prevent mismatched responses from being silently accepted. Approvable

gr2m:
Superseded by #20362, which already closed #20359.

soroush5:
Thanks — glad #20362 landed the mismatch check. Same gap I'd covered on both paths, so happy to close mine in favor of it.

PR #20389, opened 2026-09-04
docs(providers): add Hubris to OpenAI Compatible Providers

Aimagine-life:
Adds [Hubris](https://hubris.pw) — a ruble-billed LLM gateway (400+ models) with an OpenAI-compatible `/v1` endpoint — to the OpenAI Compatible Providers section, following the same structure as the NIM/Clarifai/Heroku pages (plain `createOpenAICompatible`, no separate npm package needed).

Docs-only change, no code/package changes.

### DIFF
--- a/content/providers/04-openai-compatible-providers/60-hubris.mdx
+++ b/content/providers/04-openai-compatible-providers/60-hubris.mdx
@@ -1,0 +1,69 @@
+---
+title: Hubris
+description: Use Hubris OpenAI compatible API with the AI SDK.
+---
+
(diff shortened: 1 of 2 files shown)


Aimagine-life:
Superseded by #21031. While re-signing the commit here I force-pushed from a shallow clone, which rewrote it into a repo-sized diff and closed this PR automatically; it cannot be reopened. #21031 has the same two-file docs change, rebuilt on top of current `main`, with a signed commit. Sorry for the noise.

PR #20434, opened 2026-09-07
docs(architecture): fix repo-relative links missing the ../ prefix

simpleqt:
`architecture/provider-abstraction.md` links package files as `](packages/ai/src/...)` — written from the repo root's perspective, but the document lives in `architecture/`, so all 24 links resolved to `architecture/packages/...` and 404'd. Each now carries the `../` prefix; all targets verified to exist.

### DIFF
--- a/architecture/provider-abstraction.md
+++ b/architecture/provider-abstraction.md
@@ -32,12 +32,12 @@ If you're unable to find any of the functions mentioned below in the codebase, t
 Language models are used for text generation and structured generation workflows from prompt or message input.
 
 - **AI functions**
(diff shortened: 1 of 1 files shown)


yetuge:
Checked the link targets on `main` while going through this — one line still points at a path that doesn't exist: `packages/amazon-bedrock/src/reranking/bedrock-reranking-model.ts`. The actual class file is `packages/amazon-bedrock/src/reranking/amazon-bedrock-reranking-model.ts` (I filed the same rename for the Anthropic converter in #21181). With the `../` prefix alone that one would still 404; renaming the target as well would make the row fully resolvable.

simpleqt:
Good catch — fixed in the new head: the Bedrock row now points at `packages/amazon-bedrock/src/reranking/amazon-bedrock-reranking-model.ts`. I also re-verified every relative link target in the document against `main` (31 links, all resolve) so nothing else slipped through.

PR #20467, opened 2026-09-08
feat(code-mode): add mid conversation tool discovery/updates for code-mode

aayush-kapoor:
## Background

there was a need to update tool definitions in code mode in subsequent turns without breaking cache. this was made possible by the boundary added by code mode execution (where-in the model doesn't see the normal tools defined and just sees the code mode tool)

## Summary

code mode now takes in a config option `toolDiscovery` (which can be either `'conversation`' or `'description`')
- if `conversation`, any future tool updates in the next step are passed as a user message injected in the conversation - which doesn't break the tool cache

## End-to-End Verification

verified by running the example `examples/ai-functions/src/generate-text/code-mode/dynamic-mcp-tools.ts`

## Checklist

- [x] All commits are signed (PRs with unsigned commits cannot be merged)
- [x] Tests have been added / updated (for bug fixes / features)
- [x] Documentation has been added / updated (for bug fixes / features)
- [x] A _patch_ changeset for relevant packages has been added (for bug fixes / features - run `pnpm changeset` in the project root)
- [x] I have reviewed this pull request (self-review)

PR #20572, opened 2026-09-09
fix(ai): harden code-mode conversation messages

gr2m:
This is a stacked follow-up to #20467. It hardens the new conversation-based tool discovery path while keeping ordinary `generateText` and `streamText` behavior unchanged when no caller messages are injected.

#20467 adds SDK-generated user messages to core request history. Those messages need to be both lossless and easy to persist across calls, but their bookkeeping metadata is not provider configuration. Separating appendable conversation output from provider-facing conversion limits the code-mode behavior to users who opt into tool callers and avoids exposing providers to a new undocumented payload.

The multi-step code-mode regression also verifies that the previous call's final model prompt is an unchanged prefix of the next call's prompt.

### Changes

1. [`3cc809e20e`](https://github.com/vercel/ai/commit/3cc809e20e) — **preserve structured tool-caller messages**. `prepareModelMessage` accepts a full `UserModelMessage`, but array-form content was silently discarded. This aligns runtime behavior with the public type contract and keeps text-only deduplication.
2. [`83c1ee9ade`](https://github.com/vercel/ai/commit/83c1ee9ade) — **expose appendable continuation messages**. `experimental_continuationMessages` contains only messages produced during the call: SDK-injected catalog messages plus assistant/tool responses. This avoids rebuilding history from `finalStep.request.messages`, which duplicates earlier responses after multi-step generations and can break tool-call ordering and prompt-cache prefixes.
3. ~[`e1d49267db`](https://github.com/vercel/ai/commit/e1d49267db) — **keep SDK message state out of provider options**. The `ai-sdk` namespace remains available in conversation history for incremental catalog diffs, but is stripped at the model-provider boundary. Provider-owned namespaces are preserved, and the original messages are not mutated.~ was reverted by @aayush-kapoor


examples/ai-functions/src/generate-text/code-mode/dynamic-mcp-tools.ts
```diff
@@ -48,7 +48,7 @@ run(async () => {
     'Before MCP registration:',
     JSON.stringify(beforeRegistration.content, null, 2),
   );
-  messages.push(...beforeRegistration.responseMessages);
+  messages.push(...beforeRegistration.experimental_continuationMessages);
```
aayush-kapoor, line 51: @gr2m since this is the only material change, we can skip or keep this PR as a follow up change!

let's merge https://github.com/vercel/ai/pull/20467 and then itg

PR #20622, opened 2026-09-10
fix(vertex): download tool result file URLs

### LINKED ISSUE #15671: @ai-sdk/google-vertex: image-url in tool-result doesn't auto-download
## Summary

**Remote `image-url` in tool-result content still does not work for Google / Vertex Gemini.** The model does not see the image and responds based on text context only (often resulting in hallucinations).

Using `image-data` (base64) in tool results works around the issue for Gemini.

## Reproduction

Minimal repro using an actual tool definition and a public image URL (same pattern the maintainers requested in #15173):

```ts
const PreviewImageTool = tool({
  description: "Return a preview image URL from tool execution.",
  inputSchema: z.object({}),
  execute: async () => ({
    imageUrl:
      // Image of a cat
      "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRr_-62a40u3lSIyRP5EKOjJeQiZROwTeVCOQ&s",
  }),
  toModelOutput({ output }) {
    return {
      type: "content",
      value: [
        { type: "text", text: "Here is the preview image." },
        { type: "image-url", url: output.imageUrl },
      ],
    }
  },
})

export default async function main() {
    model: vertex("gemini-3.5-flash"),
    tools: { preview_image: PreviewImageTool },
    prompt:
      "Use the preview_image tool, then describe the image in detail. If you cannot see the image, say so. If you know the URL of the image, repeat it.",
    stopWhen: stepCountIs(5),
  })

  logger.info("Result", { result: result.text })
}

await main().then(() => process.exit(0))
```

LLM Output:
> I have run the tool, and the URL of the image is: `https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRr_-62a40u3lSIyRP5EKOjJeQiZROwTeVCOQ&s`
> 
> However, because I do not have direct visual access or a rendering interface to view the pixel content of this specific thumbnail URL in my current environment, I cannot see the image and am unable to describe its visual details.

For comparison, the same snippet works correctly with `@ai-sdk/amazon-bedrock`, `@ai-sdk/anthropic`, and `@ai-sdk/openai` on `ai@6.0.191`. In particular, Amazon Bedrock also doesn't support image URLs but that package downloads the image URL and converts it to base 64 before sending the request

## Related

- #15173 — Bedrock `image-url` in tool-result (fixed in `ai@6.0.182` via #15232)

### AI SDK Version

- `ai`: 6.0.191
- `@ai-sdk/google-vertex`: 4.0.139

n1ckoates:
## Background

- Vertex tool results with remote image URLs were sent as text instead of image bytes.
- Vertex supports image URLs in user messages, but not in tool results
  - AI SDK `supportedUrls` can't differentiate between this when downloading images for models that don't support URLs. 
  
## Summary

- Download remote files only for Vertex tool results, then send them as inline image data.
- Reuse the existing safe download utility from `@ai-sdk/provider-utils`.

## End-to-End Verification

Added a live example and ran it:

- Pre-fix: Vertex received URL text rather than image bytes, and the model hallucinated a nonexistent landscape.
- After fix: the model described the cat in the returned image.

## Checklist

## Future work

Add some capability to `supportedUrls` to differentiate between tool results and user messages, so that the download logic lives in core

## Related Issues

Fixes #15671

### DIFF
--- a/examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
+++ b/examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
@@ -1,0 +1,46 @@
+import { googleVertex } from '@ai-sdk/google-vertex';
+import { generateText, isStepCount, tool } from 'ai';
+import { z } from 'zod';
+import { run } from '../../lib/run';
+
+const imageUrl =
+  'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRr_-62a40u3lSIyRP5EKOjJeQiZROwTeVCOQ&s';
+
+run(async () => {
+  const getCatImage = tool({
+    description: 'Return the requested preview image.',
+    inputSchema: z.object({}),
+    execute: async () => ({ imageUrl }),
+    toModelOutput: ({ output }) => ({
+      type: 'content',
+      value: [
+        {
+          type: 'text',
+          text: 'Here is the requested preview image.',
+        },
+        {
+          type: 'file',
+          mediaType: 'image',
+          data: { type: 'url', url: new URL(output.imageUrl) },
+        },
+      ],
+    }),
+  });
+
+  // Reproduction for https://github.com/vercel/ai/issues/15671:
+  // before the Vertex fix, the remote file is sent back as URL-shaped content
+  // rather than image bytes, so the model cannot inspect the tool result.
+  const result = await generateText({
+    model: googleVertex('gemini-2.5-flash'),
+    tools: { get_cat_image: getCatImage },
+    stopWhen: isStepCount(2),
+    prepareStep: ({ stepNumber }) => ({
+      toolChoice:
+        stepNumber === 0 ? { type: 'tool', toolName: 'get_cat_image' } : 'none',
+    }),
+    prompt:
+      'Use get_cat_image exactly once, then describe the image it returned in one sentence.',
+  });
+
+  console.log(result.text);
+});
--- a/examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
+++ b/examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
@@ -4,7 +4,7 @@ import { z } from 'zod';
 import { run } from '../../lib/run';
 
 const imageUrl =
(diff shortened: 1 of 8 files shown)


examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
```diff
@@ -0,0 +1,46 @@
+import { googleVertex } from '@ai-sdk/google-vertex';
+import { generateText, isStepCount, tool } from 'ai';
+import { z } from 'zod';
+import { run } from '../../lib/run';
+
+const imageUrl =
+  'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRr_-62a40u3lSIyRP5EKOjJeQiZROwTeVCOQ&s';
```
gr2m: can we rely that this will remain available? Alternatively start a local server?

gr2m (commented):
Could we route these URLs through the core `experimental_download` path instead of fetching inside the provider?

`downloadBlob` defaults to 2 GiB, and the nested `Promise.all` calls download every tool-result URL concurrently. The data is then copied again by `blob.arrayBuffer()` and base64 encoding, so a tool result with large or numerous URLs can exhaust process memory.

The download planner should treat Vertex tool-result URLs as unsupported while leaving user-message URLs supported. This also preserves custom download policies. At minimum, set an explicit provider-appropriate limit and bound concurrency:

```ts
await downloadBlob(url, {
  abortSignal,
  maxBytes: 7 * 1024 * 1024,
});
```

Maybe add an option to `createGoogleVertex` to override the provider default?

```
const vertex = createGoogleVertex({
  toolResultDownloads: {
    maxBytes: 20 * 1024 * 1024,
  },
});
```

and then set `maxBytes: config.downloadToolResultFiles.maxBytes` instead of hardcoding `maxBytes: 7 * 1024 * 1024`?

examples/ai-functions/src/generate-text/google/vertex-tool-result-image-url.ts
n1ckoates: swapped it to a github URL to an image in this repository, that points to the specific commit hash so it should remain available

gr2m (approved):

PR #20627, opened 2026-09-10
feat(openai): add experimental Live WebSocket provider

kkdawkins:
## Stack 1/4 — client-delegation Live provider

Stack: **#20627** → #20628 → #20629 → optional #20630.

Supports `openai.experimental_realtime('gpt-live-1')` through the existing experimental realtime v4 contract. Explicit `{ api: 'live' }` selects the Live adapter for unknown/EAP IDs. Existing Realtime token minting is preserved; Live uses server-held upstream credentials.

### Initial scope

- Server WebSocket connection settings and provider-owned command/event conversion.
- Client delegation: the application owns task execution and returns context/results.
- Continuous audio, timestamped transcripts, client delegation metadata, context appends, mute/unmute, cumulative voice duration and finalization.
- Startup history, voice/audio configuration and preserved raw events.

Managed Responses configuration, backend commands, tool-call correlation and continuation orchestration are deferred. Unsupported configuration/commands reject explicitly; there is no hidden automatic fallback. This is a provider/server-adapter layer; application hook support follows in #20629. WebRTC is isolated in #20630.

### Verification

Full workspace type-check, root lint/format, docs validation, provider/factory/type regressions and changed-package structural conventions pass. Changeset coverage is verified against the PR's parent diff, including the small `ai` compatibility guard.

This is the provider-first review boundary replacing the consolidated #20605.


packages/openai/README.md
```diff
@@ -43,4 +43,199 @@ const { text } = await generateText({
...
+Usage in `session-usage` is cumulative: replace the previous snapshot, do not sum.
+Send `session-close` and keep receiving until `session-closed` provides final
+usage and reason before releasing the transport. A socket close alone is not
+finalization. Every event preserves `raw`; unknown types become `custom`, and
+malformed known events become `error` with code `invalid_server_event`.
+Provider errors retain `clientEventId`; null error codes normalize to `undefined`
+and remain available in `raw`.
+
+See OpenAI's [Live session guide](https://developers.openai.com/api/docs/guides/live-conversations)
+and [delegation guide](https://developers.openai.com/api/docs/guides/live-delegation).
```
n1ckoates: do we document this sort of thing elsewhere?

this seems more "low level" in that these methods are used by AI SDK to implement e.g. `useRealtime`, but users shouldn't be using the methods directly.

same thing as documenting `generateText` rather than documenting `model.doGenerate`

packages/openai/src/openai-provider.ts
```diff
@@ -127,6 +129,11 @@ export interface OpenAIProvider extends ProviderV4 {
    */
   experimental_realtime: RealtimeFactoryV4;
 
+  /** Creates an experimental continuous Live model with server WebSocket support. */
+  experimental_live(
+    modelId: OpenAIRealtimeModelLiveId,
+  ): OpenAIRealtimeModelLive;
+
```
n1ckoates: why not add this to the existing `experimental_realtime` and if the implementation differs, switch based on model name?

could also see it just being different enough to warrant different naming, but wanted to note

packages/openai/README.md
gr2m: we usually don't worry too much about the package READMEs. But in general I agree

packages/openai/src/openai-provider.ts
gr2m: I'm investigating this exact same thing right now as well

gr2m (approved):

PR #20628, opened 2026-09-10
feat(ai): add WebSocket continuous realtime sessions

kkdawkins:
## Stack 2/4 — client-first WSS session runtime

Stack: #20627 → **#20628** → #20629 → optional #20630.

Adds WebSocket continuous sessions with application-owned client delegation. The runtime manages transport/media lifecycle, ordered bounded events, captions, command acknowledgments and authoritative final duration. Applications own delegated work and context/result submission.

### Correctness and compatibility

- Attempt-scoped callbacks and asynchronous errors; reentrant setup/teardown fencing.
- Explicit relay/client-secret modes and validated token setup payloads.
- Mandatory abnormal WebSocket close-code handling from #20754 is in this layer.
- Required WebSocket URL/protocol hook-identity compatibility is also in this layer.
- Continuous playback can recover at the live edge; legacy turn-based playback keeps its queued/heard-prefix behavior.
- Capture ownership, startup cancellation, bounded finalization, and current-session isolation.

Managed Responses coordination, backend token aggregation and automatic backend continuation have been removed from this initial scope. Existing legacy frontend tool behavior remains.

### Verification

Full workspace type-check, root lint/format, Node/edge runtime regressions, token-based compatibility tests, changed-package structural conventions and changeset coverage pass. Validation uses the existing slim SDK Zod helper; bundle limits are not raised for this cleanup. No WebRTC implementation is required by this layer.


jd-solanki:
The `if (this.ws === ws)` guards this adds to `onmessage` and `onerror` fix a bug that bites on released `main`, independently of the Live work. Some evidence, in case it is useful for prioritising them.

On `ai@7.0.97` and `7.0.99`, a socket retired by a later `connect()` or by `dispose()` still drives the session through those two handlers, while `onopen` and `onclose` correctly ignore it. The worst of it is not the status write:

```
BUG 1  retired socket still drives the live session
  "disposed" appears in ai/dist/index.js   : no
  retired by a later connect()
    status before firing its handlers    : connected
    onclose  (guarded)   ignored it      : true
    onerror  (unguarded) set status      : error
    onmessage(unguarded) added its text  : true
    onmessage(unguarded) ran onToolCall  : true
    that output written to live socket   : true
```

A `tool-call` effect from the retired socket runs the application's `onToolCall`, and the result is written to whichever socket is live now. A `speech-started` from it stops the live session's playback and sends a `conversation-item-truncate`.

Runnable, one command, no network or API key: https://github.com/jd-solanki/reproductions/tree/vercel-ai-realtime-transport-socket-lifecycle--2026-09-13

I confirmed the guards are what fix it by planting them into `node_modules` and re-running — the reproduction then exits non-zero, as it should.

How we ran into it: a voice kiosk where each visit reconnects. A socket from the previous visit closing late on flaky wifi flipped the live session to `error` mid-answer, and our own teardown then tore down a healthy conversation while its socket stayed open.

We could not guard it from outside, since `transport` is private, so we stopped reconnecting session objects at all: one socket per session, then throw the session away. That works, but it costs us the conversation on every restart. Landing these guards as a patch ahead of the rest of the stack would let us drop it.

One thing this PR does not cover, which I filed separately as #20695: `onClose` is still `() => void` on this branch, so the WebSocket close code and reason still reach nobody. That is what made this so slow to diagnose — Gemini's `1007` closes look identical to a clean hangup.

gr2m (approved):

PR #20658, opened 2026-09-11
fix(react): defer useChat cleanup until commit

gr2m:
## Summary

- move cleanup of internally managed Chat instances into the React effect lifecycle
- avoid aborting the committed chat when an ID-changing concurrent render suspends
- preserve caller-managed Chat instances
- add a Suspense regression test for an uncommitted ID change

## Verification

- pnpm --dir packages/react exec vitest --config vitest.config.js --run src/use-chat.ui.test.tsx
- pnpm --filter @ai-sdk/react type-check
- pnpm type-check:full
- pnpm exec ultracite check packages/react/src/use-chat.ts packages/react/src/use-chat.ui.test.tsx

Stacked on #20654.


gr2m:
fixed upstream

PR #20673, opened 2026-09-11
fix(ai): wait for stream pipeline to finish in chat.stop()

### LINKED ISSUE #13962: chat.stop() fires the abort signal but returns immediately without waiting for the stream pipeline to fully terminate

tiagovilasboas:
## Background

`chat.stop()` aborted the active request and returned immediately. Callers that `await chat.stop()` and then clear messages could still see `status === 'streaming'`, and buffered UI stream chunks could write after the caller thought the pipeline was done.

Maintainer reproduction for #13962: `await chat.stop()` returned while status was still `streaming`.

## Summary

- Track a `finished` promise on the active response and resolve it after abort handling, `onFinish`, and cleanup in `makeRequest`.
- `chat.stop()` still aborts pending preparations, resume requests, and the active response, then awaits that promise.
- When `await chat.stop()` settles, `status` is no longer `streaming`.
- Regression tests cover pipeline completion and the clear-chat / `AFTER_STOP` race.

Fixes #13962

## Contributor Credit

- Issue report: @lerouxb
- Maintainer reproduction: @lgrammel

## End-to-End Verification

Unit tests in `packages/ai/src/ui/chat.test.ts` assert that `await chat.stop()` resolves only after abort handling and `onFinish`, with `status === 'ready'`, and that clearing messages afterwards is not overwritten by a buffered `AFTER_STOP` chunk.

## Checklist

## Related Issues

Fixes #13962


tiagovilasboas:
Older open PR for the same bug: #15282 (last activity May). This one waits for the full pipeline (`finished` after `onFinish`) and has tests that `await chat.stop()` no longer leaves `streaming`, so buffered chunks cannot write after a clear.

PR #20686, opened 2026-09-12
fix(ai): notify telemetry onError on mid-stream provider stream errors

### LINKED ISSUE #20586: `streamText` leaves OpenTelemetry spans open when provider response stream errors
### Description

When a provider's `doStream` call returns a stream and that stream later errors, `streamText` does not notify the telemetry integration. The `invoke_agent`, active `step`, and `chat` spans stay open. OpenTelemetry exporters do not receive those spans because they export only after `span.end()`.

We found this after an HTTP 200 response body failed while it was being read. Earlier steps had already ended and exported. Each pointed to the root span, which never exported, so Datadog displayed the completed steps as orphans.

Expected behavior:

- A non-abort stream error calls `telemetryDispatcher.onError` with the call ID and error.
- The OpenTelemetry integration records the error and ends the active spans.
- The exporter receives the root span and preserves the trace hierarchy.

Actual behavior:

- The non-abort error path calls only `controller.error(error)`: https://github.com/vercel/ai/blob/ai%407.0.93/packages/ai/src/generate-text/stream-text.ts#L1877-L1882
- `telemetryDispatcher.onError` is called for setup errors, but not for errors raised later while reading the stream: https://github.com/vercel/ai/blob/ai%407.0.93/packages/ai/src/generate-text/stream-text.ts#L3127-L3138
- `@ai-sdk/otel` already has the required cleanup in `onError`, but this path never reaches it: https://github.com/vercel/ai/blob/ai%407.0.93/packages/otel/src/open-telemetry.ts#L1427-L1471

The abort path does run telemetry cleanup and has test coverage. The same cleanup is missing for a non-abort stream error.

### Reproduction

This failing test can sit next to the existing `streamText` abort test in `packages/otel/src/open-telemetry.test.ts`:

```ts
it('closes streamText spans when the provider stream errors', async () => {
  let pullCalls = 0;

  const result = streamText({
    model: new MockLanguageModelV4({
      doStream: async () => ({
        stream: new ReadableStream({
          pull(controller) {
            switch (pullCalls++) {
              case 0:
                controller.enqueue({
                  type: 'stream-start',
                  warnings: [],
                });
                break;
              case 1:
                controller.enqueue({
                  type: 'text-start',
                  id: '1',
                });
                break;
              case 2:
                controller.enqueue({
                  type: 'text-delta',
                  id: '1',
                  delta: 'Hello',
                });
                break;
              case 3:
                controller.error(new Error('socket closed'));
                break;
            }
          },
        }),
      }),
    }),
    prompt: 'test-input',
    telemetry: {
      integrations: integration,
    },
  });

  await result.consumeStream();

  expect(
    tracer.spans.map(span => ({
      name: span.name,
      ended: span.ended,
    })),
  ).toEqual([
    { name: 'invoke_agent mock-model-id', ended: true },
    { name: 'step 1', ended: true },
    { name: 'chat mock-model-id', ended: true },
  ]);
});
```

The assertion currently fails because all three spans remain open.

### AI SDK Version

- `ai`: 7.0.93
- `@ai-sdk/otel`: 1.0.93
- `@ai-sdk/provider-utils`: 5.0.33

The same missing cleanup is present on the current `main` branch as of September 10, 2026.

### Code of Conduct

soroush5:
Fixes #20586.

In streamText's resilient stream, a non-abort mid-stream provider error (a socket dying while reading the response body) hit a bare controller.error(). telemetryDispatcher.onError never fired, so the OTEL root, step, and chat spans stayed open. Exporters never receive un-ended spans, so earlier steps show as orphans. Setup errors and aborts already had telemetry cleanup. This path did not.

The fix notifies telemetryDispatcher.onError with { callId, error } before erroring the stream, mirroring the setup-error path. The abort path is untouched.

Verified: the new otel test fails before (spans un-ended) and passes after (all three ended). I also ran it in a real Chromium page driving streamText against a dying stream: un-ended without the fix, ended with it. stream-text.test.ts 453/453, open-telemetry.test.ts 51/51, type-check clean.

### DIFF
--- a/.changeset/olive-spans-close.md
+++ b/.changeset/olive-spans-close.md
@@ -1,0 +1,5 @@
+---
+'ai': patch
+---
+
+Notify telemetry onError on mid-stream provider stream errors so OTEL spans close instead of leaking
--- a/packages/ai/src/generate-text/stream-text.test.ts
+++ b/packages/ai/src/generate-text/stream-text.test.ts
@@ -24892,6 +24892,89 @@ describe('streamText', () => {
     });
   });
 
+  describe('provider stream error', () => {
+    let result: StreamTextResult<ToolSet, any, never>;
+    let telemetryCalls: Array<{ name: string; event?: unknown }> = [];
+    const streamError = new Error('socket closed');
+
+    beforeEach(() => {
+      telemetryCalls = [];
+
+      let pullCalls = 0;
+      const telemetryIntegration: Telemetry = {
+        onAbort: event => {
+          telemetryCalls.push({ name: 'onAbort', event });
+        },
+        onEnd: event => {
+          telemetryCalls.push({ name: 'onEnd', event });
+        },
+        onError: event => {
+          telemetryCalls.push({ name: 'onError', event });
+        },
+      };
+
+      result = streamText({
+        telemetry: {
+          integrations: telemetryIntegration,
+        },
+        _internal: {
+          generateCallId: () => 'test-telemetry-call-id',
+        },
+        model: new MockLanguageModelV4({
+          doStream: async () => ({
+            stream: new ReadableStream({
+              pull(controller) {
+                switch (pullCalls++) {
+                  case 0:
+                    controller.enqueue({
+                      type: 'stream-start',
+                      warnings: [],
+                    });
+                    break;
+                  case 1:
+                    controller.enqueue({
+                      type: 'text-start',
+                      id: '1',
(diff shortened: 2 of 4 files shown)


aayush-kapoor (commented):
will be fixed by https://github.com/vercel/ai/pull/20870

soroush5:
Thanks for confirming, glad it is covered by #20870.

Issue #15465, opened 2026-05-20
bug(@ai-sdk/mcp): Streamable HTTP can refresh OAuth tokens twice when initial SSE GET and POST both get 401


itsjamie:
### Description

`@ai-sdk/mcp`'s Streamable HTTP transport can invoke the OAuth refresh flow twice concurrently during client initialization.

The transport starts the optional inbound SSE `GET` in the background from `HttpMCPTransport.start()`:

```ts
async start(): Promise<void> {
  // ...
  this.abortController = new AbortController();
  void this.openInboundSse();
}
```

At the same time, the MCP client proceeds with the normal JSON-RPC `initialize` request over `POST`. Both request paths attach the current bearer token via `authProvider.tokens()`, and both independently call `auth(this.authProvider, ...)` on `401`:

- `send(...)` / POST path: calls `auth(...)` on `401`, then retries the POST.
- `openInboundSse(...)` / GET path: calls `auth(...)` on `401`, then retries the SSE GET.

When the stored credential contains a `refresh_token`, `auth(...)` performs:

```ts
const tokens = await provider.tokens();
const newTokens = await refreshAuthorization(...tokens.refresh_token...);
await provider.saveTokens(newTokens);
```

Because the GET and POST auth flows are not singleflighted, both can read the same refresh token and call the token endpoint concurrently.

This is not just a duplicate network request for providers using refresh-token rotation. The first refresh can invalidate the refresh token that the second refresh is already trying to use, causing `invalid_grant`, forced reauthorization, or a subsequent 401 with a freshly persisted-but-already-obsolete token.

### Affected versions

Observed with `@ai-sdk/mcp@1.0.36`.

I also checked the published package contents for:

- `@ai-sdk/mcp@1.0.42`
- `@ai-sdk/mcp@2.0.0-beta.37`

Both still appear to have the same shape: `start()` fire-and-forgets `openInboundSse()`, and both the GET and POST 401 paths call `auth(...)` independently.

### Expected behavior

Only one OAuth recovery/refresh flow should run at a time per transport/auth provider/resource. Concurrent 401s during one Streamable HTTP connection should wait on the same in-flight auth promise and then retry with the same refreshed tokens.

Alternatively, the optional inbound SSE `GET` could be deferred until after initialization, so it does not race the initial `POST /initialize` auth recovery.

### Actual behavior

On a protected Streamable HTTP MCP server, startup can produce two concurrent authenticated requests:

```txt
[...]

PR #20694, opened 2026-09-13
fix(mcp): pass the rejected token generation to OAuth credential invalidation

### LINKED ISSUE #20674: MCP OAuth: tokens invalidation can delete a concurrent refresh winner in shared credential storage
### Description

With `@ai-sdk/mcp` 1.0.78, two independently connected HTTP MCP clients sharing one persisted OAuth credential can race a rotating refresh token. The losing refresh calls `OAuthClientProvider.invalidateCredentials("tokens")` after `invalid_grant`. If the provider implements that request as unconditional deletion, it removes the new token pair that the winning refresh has already persisted.

Related: #15465 addressed concurrent recovery inside one transport. This case uses two clients/providers sharing storage (for example, two replicas), so per-transport singleflight does not prevent it.

### Reproduction

Use a loopback OAuth/MCP test server that rotates refresh tokens and two `createMCPClient` HTTP clients. Each gets its own OAuthClientProvider instance backed by the same credential record.

1. Connect both clients with the same usable credential.
2. Expire their access token so tool execution on each receives 401.
3. Start tool execution on both concurrently. Hold one refresh at the test token endpoint until the competing refresh has started; both read the same refresh token.
4. Release the endpoint. One exchange succeeds and stores a new access/refresh pair; the other receives `invalid_grant`.
5. Implement `invalidateCredentials("tokens")` by deleting the shared token record.
6. Assert that the winning refresh's tokens are still in storage and accepted by the server.

Actual: the stored refresh token is undefined. Expected: failure of the stale refresh cannot delete a newer credential.

On 2026-09-11 we ran this deterministic loopback regression with the real pinned AI SDK transport and a test MCP/OAuth server. The assertion fails when the provider honors `tokens` invalidation by deleting the record, and passes with the workaround below. This is local test evidence, not a live identity-provider incident.

### Workaround and requested contract

Our provider currently ignores `invalidateCredentials("tokens")`, allowing the SDK retry to reread the shared record. A separately detected genuinely dead grant enters consent recovery; explicit disconnect remains supported. We want to remove this override.

Can invalidation carry the failed token or a credential version, allowing the provider to compare-and-delete only the failed generation? Alternatively, please document a supported cross-client coordination contract. The current scope-only callback cannot distinguish the stale failed token from a newer successful refresh in the same record.

### AI SDK version

- `@ai-sdk/mcp`: 1.0.78
- `ai`: 6.0.277
- Node: 22.23.1

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

jstar0:
## Problem

With `@ai-sdk/mcp`, two independently connected HTTP MCP clients that share one persisted OAuth credential (for example, two replicas) can race a rotating refresh token. The losing refresh receives `invalid_grant`, and `auth()` then calls `provider.invalidateCredentials('tokens')` unconditionally. A provider that honors that request by deleting the stored record removes the **newer** token pair that the winning refresh already persisted, so the deployment loses its valid credential with no error surfaced.

#15465 addressed concurrent recovery inside one transport; the cross-client case with shared storage remains exposed.

## Fix

- `InvalidGrantError` now carries `tokens`: the exact token generation the failed refresh attempt used (attached where the stored tokens were read and passed to `refreshAuthorization`).
- `OAuthClientProvider.invalidateCredentials` accepts an optional second argument `context?: OAuthCredentialsInvalidationContext` (exported type). The `invalid_grant` invalidation forwards `{ tokens: error.tokens }`, and the authorization-server pin mismatch invalidation forwards the stored generation.
- The interface change is additive. Providers that ignore `context` keep the previous unconditional behavior. Providers with shared storage can compare-and-delete: remove the record only when the rejected generation is still the one persisted, then let the built-in retry re-read the winner's tokens.

## Verification

- `pnpm test` in `packages/mcp`: 322 passed, including two new regressions in `oauth.test.ts`:
  - a stale refresh loses a rotation race while a concurrent winner persists its pair — `invalidateCredentials` receives the stale generation, the winner's tokens survive, and the retry refreshes the winner's refresh token to `AUTHORIZED`;
  - the rejected generation is still the stored one — the record is cleared and the flow falls through to the redirect.
- `pnpm type-check`, and `ultracite check` (format + lint) pass on the changed files.

Fixes #20674


jstar0:
@gr2m This PR implements the compare-and-delete option raised in #20674 by passing the refresh attempt's token generation through an optional invalidation context. Hosted tests, type checks, and builds are green; the only failing status is the Vercel preview authorization check. Could you confirm whether this API contract fits the intended direction and review the PR when convenient?

PR #20697, opened 2026-09-13
fix(google): keep a realtime functionResponse.response an object

### LINKED ISSUE #20696: google: a non-object realtime tool result kills the Gemini session
On the Gemini realtime path, whatever a tool returns goes into `functionResponse.response` as it is. Gemini types that field as `google.protobuf.Struct`, which accepts an object and nothing else, so any other JSON value is a protocol violation and the socket is closed with `1007`.

```ts
// packages/google/src/realtime/google-realtime-event-mapper.ts
const parseResult = await safeParseJSON({ text: item.output });
const response = parseResult.success ? parseResult.value : {};
```

Nothing asks the caller for an object. `onToolCall` returns `unknown`, `addToolOutput(callId, result: unknown)` takes `unknown`, and the reducer stores `JSON.stringify(result)`. A string result round-trips through that and lands on the wire unwrapped:

```json
{"toolResponse":{"functionResponses":[{"id":"call-1","name":"search","response":"SEARCH_UNAVAILABLE"}]}}
```

Gemini closes the session. Because the close code is dropped on the way back through the transport, the application sees a plain disconnect.

The same package already does this correctly everywhere else. `appendToolResultParts` on the non-realtime path wraps unconditionally:

```ts
response: { name: toolName, content: ... }
```

This is also settled ground for this provider: #2711 is the identical failure one transport earlier, with the same protobuf error quoted in full, closed as completed. #2926 and #2356 are the string and array variants. The realtime mapper reintroduces it in a new code path.

Google's own SDK documents the fix in the field's docstring:

> Required. The function response in JSON object format. Use "output" key to specify function output and "error" key to specify error details (if any). If "output" and "error" keys are not specified, then whole "response" is treated as function output.

### The quieter half

An `output` that is not valid JSON takes the `: {}` branch. No throw, no warning, no `onError`. The model is told the tool returned an empty object and answers as though it had. The socket stays up, so there is nothing to notice: the answer is just wrong.

### Expected

A non-object tool result reaches Gemini as an object — wrapped under `output`, per the docstring above — or the SDK reports that it cannot be sent. An unparseable `output` surfaces rather than becoming `{}`.

### Actual

Sent unwrapped. Gemini closes the socket with `1007`. Unparseable output becomes `{}` silently.

Happens every time.

## Reproduction

https://github.com/jd-solanki/reproductions/tree/vercel-ai-realtime-transport-socket-lifecycle--2026-09-13

```bash
npm install
npm run repro
```

One command, no network, no API key. It drives the public `experimental_realtime(...)` model and also captures the frame handed to `WebSocket.send()` end to end through `onToolCall`.

```
BUG 3  FunctionResponse.response is not always an object
  output '"SEARCH_UNAVAILABLE"'  -> response: "SEARCH_UNAVAILABLE"
  output '42'                    -> response: 42
  output '["a","b"]'             -> response: ["a","b"]
  output '{"status":"..."}'      -> response: {"status":"SEARCH_UNAVAILABLE"}
  output 'not json at all'       -> response: {}  (threw nothing: true)
  frame handed to WebSocket.send():
    {"toolResponse":{"functionResponses":[{"id":"call-1","name":"search","response":"SEARCH_UNAVAILABLE"}]}}
```

The object case is in there as the control: it serializes correctly, so the shape of the tool result is the whole difference.

A one-line assertion in `packages/google/src/realtime/google-realtime-event-mapper.test.ts` covers it in-tree if that is easier to act on.

## AI SDK Version

```
ai: 7.0.97
@ai-sdk/google: 4.0.67
```

Also reproduced on the newest published versions:

```
ai: 7.0.99
@ai-sdk/google: 4.0.69
```

## Code of Conduct

L4XB:
Fixes #20696.

## The change

```ts
const response = parseResult.success
  ? toFunctionResponseStruct(parseResult.value)
  : { output: item.output };
```

with

```ts
function toFunctionResponseStruct(value: unknown): Record<string, unknown> {
  const isStruct = typeof value === 'object' && value !== null && !Array.isArray(value);
  return isStruct ? (value as Record<string, unknown>) : { output: value };
}
```

An object is passed through unchanged — that is the shape that already worked and the Struct passthrough the field is for. Everything else is wrapped under `output`, which is what the field's own docstring prescribes and what `appendToolResultParts` has done on the non-realtime path all along, so the two paths agree again.

`null` and arrays are wrapped too: `typeof null === 'object'` and `typeof [] === 'object'`, and neither is a `Struct`.

**The quieter half.** An `output` that is not valid JSON no longer becomes `{}`. Wrapping the text rather than reporting an error is a choice worth naming: the information the tool produced is preserved and the model is told the truth, where `{}` asserted something false with nothing thrown, nothing warned and the socket still up. If you would rather this surfaced as an error instead, that is a one-line change and I am happy to make it.

## Testing

`packages/google/src/realtime/google-realtime-event-mapper.test.ts`:

- a parametrized case over the shapes from your reproduction — string, number, array, `null`, boolean — each asserting the serialized frame carries `response: { output: … }`;
- the unparseable case, asserting the text is kept.

The object case was already covered and is untouched, which is the control: the shape of the tool result is the whole difference.

```
vitest run src/realtime/    57 passed
vitest run src              435 passed
```

Negative control: restoring `const response = parseResult.success ? parseResult.value : {};` reddens all six new cases and nothing else.

`tsc --noEmit` reports nothing in the two files this branch touches. Its 20 errors are all in `google-batch.test.ts`, from `@ai-sdk/test-server` not being built in my checkout; they are there on a clean tree too.

## One existing test changed

`falls back to empty object for malformed function-call-output` asserted `response: {}`. That was the defect written down as a feature — the name says "falls back", and what it pinned was the model being told a tool returned an empty object. It now asserts the text is kept, under a name that describes it, with a comment recording what it used to claim.

Changeset included (`@ai-sdk/google`: patch).


L4XB:
CI note: the red check is `Vercel`, which reports `Authorization required to deploy` — the preview deployment needs a team member to authorize it for a fork, so it cannot pass from my side. Everything else is green, including Security Review, Agent Review, Code Owners and Socket. The change is verified locally: `vitest run src` in `packages/google` is 435 passed, and `tsc --noEmit` reports nothing in the two files this branch touches.

gr2m (approved):
Cheers! I just trimmed the comments a bit, we don't need historical context

Issue #20448, opened 2026-09-08
Community provider: pay-per-call tools from an x402 marketplace (spend-guarded)


anatshad:
### Description

Hey there! I maintain Fatstack (fatstack.net), a non-custodial marketplace where AI agents discover MCP/API tools and pay per call in USDC on Base via the x402 protocol.

I've built and published a tool provider for the AI SDK: @fatstack/ai-sdk-tools (npm). One import points an agent at our machine-readable registry and returns callable tools; payments run through a paying fetch with required spend guards - the constructor refuses to build without a daily cap, so there is no unguarded default. Runnable example against Base Sepolia included.

Before opening a PR, I wanted to ask how you'd prefer this to land: as an entry in the community providers registry/docs pointing at the published package, or is there a different home for third-party tool providers now? Happy to follow whatever structure you prefer - the package already mirrors the conventions from recently merged community providers.

Context on the marketplace side, in case it matters for review: catalog metrics are derived from confirmed on-chain payments (never self-reported), and the payment path is audited against published x402 attack classes - the write-up is public.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

anatshad:
Happy to make this zero-work for whoever picks it up: I've prepared the full docs entry - installation, configuration (including the required spend-guard setup), a usage example, and the registry line - following the structure of the existing community provider pages. I can open the PR against the docs as soon as someone confirms documentation-addition is the right label, or just open it now if that's easier to review. Whichever you prefer.

halobartku:
Tested the marketplace end-to-end today (we run an x402 storefront on Base ourselves and wanted to list) and hit a hard blocker at sign-in worth flagging before this lands in the docs:

**`/api/auth/verify` returns HTTP 500 for smart-contract wallets (ERC-1271 signatures).** Repro (2026-09-08 ~16:40Z):

1. `GET /api/auth/nonce` → 200, nonce OK
2. `POST /api/auth/message` `{address, nonce}` → 200, well-formed SIWE message (Chain ID 8453)
3. `personal_sign` that message with a **smart wallet** (ours is a Crossmint smart wallet on Base) → the signature is a valid ERC-1271 contract signature
4. `POST /api/auth/verify` `{message, signature, nonce}` → **HTTP 500, empty body** — for a garbage `0x` + `ab`×65 signature too, so the endpoint appears to crash on any signature it can't `ecrecover` instead of returning a 4xx.

[...]

PR #20711, opened 2026-09-14
feat(tools-registry): add fatstack

anatshad:
## Background

Follow-up to #20448. @gr2m confirmed Fatstack can be listed at
https://ai-sdk.dev/resources/tools with three requirements: (1) the integration
works as advertised, (2) we document on our website how the tool works with the
AI SDK, (3) `docsUrl` points to that documentation. This PR is that registry
entry, with all three requirements addressed.

Fatstack is a non-custodial marketplace where AI agents discover MCP servers and
APIs and pay per call in USDC via x402 on Base. The AI SDK integration is
published as `@fatstack/ai-sdk-tools` (MIT): one import points an agent at the
catalog and returns callable tools; payments run through a paying fetch with
required spend guards — the constructor refuses to build without a daily cap, so
there is no unguarded default.

Disclosure: I run Fatstack.

## Summary

Adds a Fatstack entry to `packages/ai/src/registry/index.ts`, mirroring the
structure of neighboring entries. `docsUrl` points to the dedicated integration
page: https://www.fatstack.net/docs/ai-sdk (install, spend-guard configuration
and rationale, a runnable example against Base Sepolia, link to the open-source
repo: https://github.com/f4tst4ck/fatstack-tools).

## Contributor Credit

- @gr2m — triage and the listing requirements in #20448
- @halobartku — independently tested the integration end-to-end with a real
  smart wallet and reported two production bugs (ERC-1271 sign-in, quote/fee
  mismatch), both since fixed and verified

## End-to-End Verification

- The integration takes real payments on Base mainnet: a paid `tools/call`
  through each serving mode settles on-chain and the 98/2 split is verified from
  the transaction receipt's USDC Transfer logs. A daily paid canary per serving
  mode guards this continuously.
- The docs page's code samples are asserted against the published package in CI
  (imports exist, spend-guard field names and required/optional status match the
  shipped schema), so the documentation cannot drift from the API without a test
  failing.
- Third-party verification: a provider running their own x402 storefront signed
  in with a Crossmint smart wallet and completed a paid call through our SDK,
  receipt-verified (#20448).

## Checklist

## Related Issues

Addresses #20448 (community provider listing per @gr2m's requirements).


ArshVermaGit (commented):
Hey, could you add a small test for the new Faststack registry entry? It’d be good to verify that the metadata and example work as expected, especially the required guards config. The checklist also has the test item unchecked right now.

anatshad:
Thanks — added in `f25bba5b`: `apps/docs/scripts/tools-registry.test.mjs`, which runs under the existing `test:site` harness (`node --test scripts/*.test.mjs`) and imports `registry.ts` directly via Node 22's type stripping, so it asserts on the real data rather than parsing the file.

It validates every entry rather than just this one, since all of it is cheap and none of it was covered before:

- slugs are unique
- required fields are present and non-empty
- each install command actually installs the `packageName` it claims
- each `codeExample` imports the package it demonstrates
- `docsUrl` / `apiKeyUrl` / `websiteUrl` / `npmUrl` are https, and `npmUrl` points at the entry's own package

All 17 existing entries pass unchanged.

Then three specific to this entry, including the guards config you asked about:

- the example passes `guards` with a `maxPerDay` ceiling. `guards` is a required argument with no default in `@fatstack/ai-sdk-tools` — omitting it is a type error, and at runtime it throws before the catalogue is fetched — so an example without it wouldn't compile and would teach the one thing this integration must not teach.
- the example targets `base-sepolia`, so a copy-paste first run doesn't spend real USDC.
- `docsUrl` points at the AI SDK integration page specifically.

I checked the tests actually fail rather than just pass: removing `guards` from the example, switching it to `base`, pointing an install command at the wrong package, and downgrading `docsUrl` to http each produce a failure, and the clean tree passes.

**One thing worth flagging, and it's your call rather than mine.** The docs job in `ci.yml` is path-filtered to `apps/docs/**`, `content/docs/**`, `content/providers/**` and `content/cookbook/**` — `content/tools-registry/**` isn't in the list. This PR now touches `apps/docs/**`, so the test runs here. But a future registry-only PR (someone adding just their tool entry) won't trigger the docs job, so this test wouldn't run for exactly the changes it's meant to guard.

The one-line fix would be adding `'content/tools-registry/**' \` to that filter, but I haven't done it: the job also runs the full docs build, which the workflow comments note needs 12 GB of swap, and signing every registry PR up for that is a cost decision that belongs to you. Happy to push it if you'd like it, or leave it.

Checklist item ticked.

anatshad:
Follow-up on my last comment — I've corrected two things in it.

**Scoped the test down to this entry (`7be0bfeb`).** The version I first pushed validated all 17 registry entries. That was more than you asked for and the wrong place for it: a tool-addition PR shouldn't introduce validation for everyone else's entries. It now covers only the fatstack entry — fields present, install commands installing `@fatstack/ai-sdk-tools`, the example importing it, links https and pointing where they claim, the guards config, and that the example targets a testnet.

**And a correction on the guards check.** In my last comment I implied it verifies the example works. It doesn't — it's `assert.match(codeExample, /guards:\s*\{/)`, which checks the example *mentions* guards, not that it compiles or that the package still requires them. If `guards` stopped being required tomorrow, that assertion would still pass.

That gap can't really be closed here: it would mean installing `@fatstack/ai-sdk-tools` into this repo to typecheck one snippet, and running it costs money. So I closed it on our side instead. The snippet now exists as compiled code rather than a string:

https://github.com/f4tst4ck/fatstack-tools/blob/main/integrations/ai-sdk-tools/example/ai-sdk-registry.ts

That file is the registry `codeExample`, character for character, and it's in the package's tsconfig, so removing `guards`, removing `maxPerDay`, or renaming `wallet` each fail the build — I checked all three actually fail rather than assuming. A companion test covers what a compiler can't object to: that it targets `base-sepolia`, never names mainnet, caps the day rather than only the call, and reads its key from the environment.

That turned up a real bug on our end, so thank you for pushing on it: the whole `example/` directory was outside our tsconfig, and a deliberate `const broken: number = 'not a number'` in the published example passed `pnpm typecheck` cleanly. The example we've been handing people was never compiled.

The note about `content/tools-registry/**` missing from the docs job's path filter still stands — happy to push that one-liner if you want it, or leave it to you.

PR #20713, opened 2026-09-14
fix(example): serialize concurrent chat persistence

ahsan-syed-32:
### Motivation

There was a race condition in the Next.js file backed chat store where concurrent writes would interfere with each other and result in updates getting clobbered. The chat state should be persisted sequentially per chat in order to make sure that requests get handled reliably and stream operations work properly.

### Changes

* Per chat write queue has been added to serially execute read-modify-write cycle.
* `saveChat` and `appendMessageToChat` methods have been updated to use the queue.
* Persistence was awaited in critical places in chat API route, including initial saving, completing stream and creating a resumable stream.
* Added regression test for concurrent updates to multiple fields of the chat state.

### Testing

* Passed regression test with Vitest.
* Passed formatting checks with Oxfmt.
* Succeeded building Next.js example.
* Succeeded running full workspace `type-check` and `check`.

Overall, these changes prevent clobbering of concurrent chat updates while keeping current file-store behavior intact.


ArshVermaGit (commented):
Could you please add a test that verifies concurrent saveChat calls for the same chat are actually serialized and that no updates are lost? The route changes correctly await the persistence operations, but a regression test would help confirm the new per-chat queue prevents the race condition described in the PR.

PR #20747, opened 2026-09-14
feat(alibaba): support preserved thinking in multi-turn requests

### LINKED ISSUE #18759: BUG: @ai-sdk/alibaba replays assistant reasoning as plain content in multi-turn Qwen requests
### Description

@ai-sdk/alibaba flattens assistant reasoning parts into ordinary assistant.content when building a multi-turn request.

The current converter effectively does this:

```ts
case 'reasoning': {
  text += part.text;
  break;
}

messages.push({
  role: 'assistant',
  content: text || null,
});
```

Source: https://github.com/vercel/ai/blob/main/packages/alibaba/src/convert-to-alibaba-chat-messages.ts

For Qwen thinking models, this conflicts with Alibaba's multi-turn guidance: retain only the visible content and ignore reasoning_content when adding an assistant response back to the conversation context.

Documentation: https://www.alibabacloud.com/help/en/model-studio/multi-round-conversation

As a result:

- reasoning that was correctly received as reasoning_content is replayed as ordinary visible assistant content on the next request;
- prompt size and token cost grow on every agent step;
- subsequent Qwen responses can emit a stray </think> followed by internal planning through delta.content, which the provider then surfaces as a text part;
- applications may render hidden reasoning as user-visible text even though their UI correctly distinguishes reasoning parts from text parts.

This appears related in shape to #17930, but the expected behavior for Qwen is provider-specific: Alibaba explicitly recommends omitting historical reasoning rather than flattening it into visible content.

Actual behavior: the assistant message in the second request contains the first turn's reasoning text concatenated with its visible answer as one plain content string.

Expected behavior: for Qwen models, the second request should include only the first turn's visible text content. Historical reasoning should be omitted, matching Alibaba's documented multi-turn behavior. If supporting providers that require preserved reasoning later, it should remain a dedicated provider field or structured part, not be merged into ordinary visible content.

### Reproduction

The behavior can be observed by recording the second request body in a two-turn Qwen conversation:

```ts
import { createAlibaba } from '@ai-sdk/alibaba';
import { generateText, type ModelMessage } from 'ai';

const requests: unknown[] = [];

const recordingFetch: typeof fetch = async (input, init) => {
  if (typeof init?.body === 'string') {
    requests.push(JSON.parse(init.body));
  }
  return fetch(input, init);
};

const model = createAlibaba({
  apiKey: process.env.DASHSCOPE_API_KEY!,
  fetch: recordingFetch,
})('qwen-plus');

const providerOptions = {
  alibaba: { enableThinking: true },
};

const firstUserMessage: ModelMessage = {
  role: 'user',
  content: 'What is 17 * 23? Reply with only the result.',
};

const first = await generateText({
  model,
  messages: [firstUserMessage],
  providerOptions,
});

await generateText({
  model,
  messages: [
    firstUserMessage,
    ...first.response.messages,
    {
      role: 'user',
      content: 'Now multiply that result by 3. Reply with only the result.',
    },
  ],
  providerOptions,
});

console.dir(
  (requests[1] as any).messages.find(
    (message: { role: string }) => message.role === 'assistant',
  ),
  { depth: null },
);
```

The captured assistant message in requests[1] contains the reasoning text followed by the visible answer in the same plain content string.

### AI SDK Version

Confirmed in:

- @ai-sdk/alibaba@1.0.25
- @ai-sdk/alibaba@1.0.37
- current main contains the same reasoning-to-text concatenation

### Code of Conduct

shaper:
## Background

Alibaba's Qwen models support [preserved thinking in multi-turn conversations](https://docs.qwencloud.com/developer-guides/text-generation/thinking#preserve-thinking-in-multi-turn): when `preserve_thinking: true` is set on the Chat Completions request, prior assistant reasoning can be replayed as a separate `reasoning_content` field so the model can build on its earlier thought process ([Chat API reference](https://www.alibabacloud.com/help/en/model-studio/qwen-api-via-openai-chat-completions)). Alibaba's [function-calling guidance](https://docs.qwencloud.com/developer-guides/text-generation/thinking#function-calling-with-thinking-mode) also asks clients to include the assistant's `reasoning_content` when sending tool results back.

Historical reasoning was previously dropped from Alibaba follow-up requests (#18764), and there was no way to opt into preserved thinking (#18759).

## Summary

Historical reasoning is now replayed as assistant `reasoning_content` — kept separate from visible `content` and preserved alongside `tool_calls` — and preserved thinking is **on by default for models that Alibaba documents as supporting it**, consistent with how other providers replay reasoning when possible:

- On a supported model (e.g. `qwen3.7-max`), requests send `preserve_thinking: true` and prior reasoning is replayed; no option needed.
- A new `preserveThinking` provider option overrides the default: `false` sends `preserve_thinking: false` and skips replay of prior-round reasoning; `true` forces it on for any model.
- On models outside the documented supported list, `preserve_thinking` is not sent unless the option is set explicitly.
- Reasoning from the current tool-call round (assistant messages after the last user message) is always sent back with tool results, regardless of the option — Alibaba's function-calling guidance says omitting it degrades accuracy. This mirrors the last-user-boundary approach used by the DeepSeek provider.
- Reasoning-only assistant turns are retained as `{ role: 'assistant', content: null, reasoning_content }` whenever their reasoning is replayed; reasoning is never concatenated into visible `content`.
- The option is independent from thinking enablement and composes with `enableThinking`/`thinkingBudget` and the top-level `reasoning` setting.
- Includes converter unit tests, generate/stream request-body tests (default-on, opt-out, same-round tool-call, and a precedence regression test with top-level `reasoning`), type-level coverage, runnable two-turn `generateText`/`streamText` examples (`qwen3.7-max`), provider docs, and README updates.

## Contributor Credit

#18759 issue author and participants for requesting `providerOptions.alibaba.preserveThinking`.

## End-to-End Verification

Ran both examples live against `qwen3.7-max` with `enableThinking: true, thinkingBudget: 2048` (no explicit `preserveThinking`, exercising the default):

- `examples/ai-functions/src/generate-text/alibaba/preserve-thinking.ts`
- `examples/ai-functions/src/stream-text/alibaba/preserve-thinking.ts`

Both completed two turns: turn 1 produced non-empty reasoning and visible text; turn 2 received `responseMessages` replayed unchanged and its reasoning explicitly built on the turn-1 analysis (turn-2 input token counts reflect the replayed `reasoning_content`; without preservation they drop to prompt-only size). Unit-test request capture confirms `preserve_thinking: true` by default on supported models, prior reasoning under `messages[].reasoning_content` (never in visible `content`), same-round reasoning retained with `preserveThinking: false`, and that generate and stream share the same request semantics.

## Checklist

## Related Issues

Closes #18759


n1ckoates (commented):
why not enable `preserveThinking` by default?

shaper:
Superseded by #20871. This PR was auto-closed as orphaned because its body marked the already-closed #18759 as closing; the branch was force-pushed in the meantime, so it can't be reopened. The new PR reflects the revised default-on behavior for supported models.

PR #20754, opened 2026-09-14
fix(ai): report abnormal realtime WebSocket closes

gr2m:
## Summary

- preserve abnormal WebSocket close codes and reasons
- route unexpected closes through the existing session error path
- keep clean code-1000 closures unchanged
- add transport and Live-session regression coverage

## Testing

- pnpm exec vitest --config vitest.node.config.js --run src/realtime/browser-realtime-transport.test.ts src/realtime/realtime-live-websocket.test.ts
- pnpm type-check:full
- pnpm check


packages/ai/src/realtime/browser-realtime-transport.ts
```diff
@@ -109,16 +118,17 @@ export class BrowserRealtimeTransport {
...
         this.ws = null;
+        const closeError = getCloseError(event);
         const complete = () => {
           if (this.epoch !== epoch) return;
           clearTimeout(this.drainTimer);
           this.epoch++;
           codec.dispose();
           try {
-            this.onClose();
+            this.onClose(closeError);
```
kkdawkins, line 135: The session callback drops this error once ready unless the model declares `finalization: "session-close"`. Existing OpenAI, Gateway, Google, xAI, and Cartesia models omit that capability, so a 1011 close still becomes `disconnected` without calling `onError`. Can we handle this in `realtime-session.ts` and add a ready legacy-session test?

packages/ai/src/realtime/browser-realtime-transport.ts
```diff
@@ -109,16 +118,17 @@ export class BrowserRealtimeTransport {
       if (this.ws === ws) this.fail(new Error('WebSocket connection error'));
     };
 
-    ws.onclose = () => {
+    ws.onclose = event => {
       if (this.ws === ws) {
```
kkdawkins, line 126: For network failures, browsers fire `error` before `close`. Our `onerror` calls `fail()`, which clears `this.ws`, so this guard drops the close event and its code. Can we preserve the close details in that path and test `error` followed by an unclean 1006 close?

kkdawkins (approved):

PR #20787, opened 2026-09-15
feat(google): realtime session options for Gemini 3.8 Live

kkdawkins:
## Background

Google's Gemini 3.8 Live models (`gemini-3.8-live`, `gemini-3.8-live-extended-thinking`) add session options the realtime provider could not express, and change one lifecycle signal:

- The extended thinking model rejects a `setup` without `thinkingConfig.thinkingLevel` (`low` | `medium` | `high`) or `thinkingBudget`, and rejects a `setup` that sends both. Every other Live model rejects `thinkingConfig` entirely.
- Tools run `NON_BLOCKING` by default (audio keeps streaming while a call is in flight); `gemini-3.8-live` can opt back into `BLOCKING`.
- On the extended thinking model `turnComplete` no longer means the model is idle. Google now sends `interactionStatus` (`IN_PROGRESS` | `IDLE` | `WAITING_FOR_INPUT`) and `waitingForInput` messages for that.

## Summary

- `GoogleRealtimeModelOptions`: add `thinkingConfig` (`thinkingLevel`, `thinkingBudget`, `includeThoughts`) and `defaultToolBehavior` (`'BLOCKING' | 'NON_BLOCKING'`).
- `buildGoogleSessionConfig`: merge `thinkingConfig` into `generationConfig`; on background-reasoning Live models default to `thinkingLevel: 'low'` when the caller sets neither knob (gated on the model name, skipped when `thinkingBudget` is set, never applied to models that reject `thinkingConfig`); stamp `behavior` on each `functionDeclaration`.
- `parseServerEvent`: forward `interactionStatus` and `waitingForInput` as `custom` events with `rawType` set, matching how `goAway` / `sessionResumptionUpdate` / `generationComplete` are already surfaced. A first-class field would need a spec change in `@ai-sdk/provider`, so this patch keeps them as custom events deliberately.
- Docs: new "Gemini 3.8 Live" section under Realtime Models in the Google provider page.
- Changeset: patch for `@ai-sdk/google`.

## Verification

- `pnpm --filter @ai-sdk/google build`, `pnpm test` (824 passing), `pnpm type-check`, `pnpm lint` — all green.
- New mapper tests cover: explicit thinking config merged into `generationConfig`, the `low` default (bare model id and `models/` prefix, with and without other config, under a raw `generationConfig` provider option), no default when `thinkingBudget` is set, no default on non-reasoning Live models, `defaultToolBehavior` applied to every declaration, explicit per-tool behavior preserved, and both new custom events.
- Validated end-to-end against both live models with a local build of this branch: the extended thinking model accepts `low`/`medium`/`high`, rejects `minimal`, rejects `thinkingLevel` + `thinkingBudget` together, and emits `interactionStatus: IDLE` after `turnComplete`.


gr2m (approved):

packages/google/src/realtime/google-realtime-model-options.ts
```diff
@@ -20,4 +20,43 @@ export type GoogleRealtimeModelOptions = {
...
+   * reasoning and function calls while streaming audio responses. Not
+   * supported by latency-optimized models (e.g. `gemini-3.8-live`).
+   */
+  thinkingConfig?: {
+    /**
+     * Thinking effort level. `gemini-3.8-live-extended-thinking` requires
+     * exactly one of `thinkingLevel` or `thinkingBudget`; the setup is
+     * rejected with neither or both.
+     */
+    thinkingLevel?: 'LOW' | 'MEDIUM' | 'HIGH';
```
n1ckoates: if this is required by Google, why don't we set a default value so users don't need to configure it themselves?

n1ckoates (approved):

packages/google/src/realtime/google-realtime-model-options.ts
kkdawkins: Good call — done in 26ff9b2. `buildGoogleSessionConfig` now sends `thinkingLevel: 'low'` on background-reasoning Live models when the caller sets neither knob, so a plain `getToken({ model: 'gemini-3.8-live-extended-thinking' })` works out of the box.

Two constraints shaped it:
- It has to be model-gated, not unconditional: every *other* Live model (`gemini-3.8-live`, `gemini-3.1-flash-live-preview`, the translate model) rejects `thinkingConfig` outright. Gating is a small regex on the `-live…thinking` model-name shape, in the style of `getMinimumThinkingLevelForGemini3Model`.
- It's skipped when `thinkingBudget` is set, since Google rejects a setup carrying both.

Picked `low` because it's a voice model and `high` adds noticeable dead air. Also switched the enum to lowercase to match `google-language-model-options` (verified `low|medium|high` on the wire; `minimal` is rejected by the Live model so it's not offered).

PR #20809, opened 2026-09-15
fix(provider-utils): preserve download DNS protection under wrapped fetch

kkdawkins:
## Summary

Default validated downloads in Node.js currently use the DNS-pinned transport only while global `fetch` remains the original Node function. Framework or instrumentation wrappers installed before or after SDK initialization cause the downloader to use the replacement global transport, bypassing connect-time DNS validation.

- Select the protected Node download transport independently of global fetch identity.
- Keep explicit custom transports and trusted-origin behavior intact.
- Preserve platform fetch for Bun, Deno, Cloudflare Workers, and framework edge runtimes, including Node-compatible process objects.
- Add a provider-utils patch changeset and update secure-fetching documentation.

## Regression coverage

Local TCP-listener tests with controlled DNS prove wrapped fetch can reach a loopback address while guarded downloads reject before opening a connection. Cases cover wrapping before/after module initialization, endpoint and redirect helpers, forwarding global fetch explicitly, and blob downloads. The eight wrapped-global cases fail against the original implementation.

Additional tests cover missing global fetch, explicit custom fetch, trusted origins, runtime compatibility, and failing closed when Node built-in loading fails. No external network is used by these regressions.

## Validation

- Provider-utils full Node and Edge suites on Node 24.14.0: 874 Node tests and 860 Edge tests passed.
- Safe-fetch regression files on Node 26.5.0: 25 tests passed.
- Initial provider-utils Node/Edge suites and package builds passed on Node 22.22.2.
- `pnpm check`
- `pnpm type-check:full`
- `git diff --check`


n1ckoates (approved):

tombryden:
@kkdawkins @n1ckoates I think this is causing a breaking bug with image generation through AI Gateway.

Every generateImage call that passes image URLs fails with:

`AI Gateway answered 500: Failed to download <image url>: Error: Cannot find module 'undici'`

I think this is likely due to bundler missing the dependency because of `createRequire(getCurrentModulePath())('undici');` syntax in `packages/provider-utils/src/safe-node-fetch.ts`.

PR #20872, opened 2026-09-16
fix(providers): map reasoning none to supported model minimums

gr2m:
## Background

`reasoning: 'none'` can currently send an invalid native value or omit an available minimum on models that require reasoning. For example, Gemini 2.5 Pro receives budget `0`, while Kimi K3 omits effort and uses the API default instead of `low`.

This addresses the cases from #20867 where model-specific normalization of the shared setting is unlikely to break existing integrations. It does not close the full provider audit.

## Summary

| Provider | Shared `none` mapping |
| --- | --- |
| OpenAI Chat / Responses | GPT-6/Astra → `low`; original GPT-5/mini/nano → `minimal`; GPT-5 Pro → `high`; o3/o3-mini/o4-mini → `low`, including dated snapshots |
| Google GenerateContent | Gemini 2.5 Pro → budget `128`; Gemini 3/3.1 Pro → `low` (also corrects shared `minimal`); explicit budget/level suppresses conflicting shared controls |
| Anthropic | Fable 5/5.1 and Mythos 5/Preview → adaptive thinking with `low` effort; explicit thinking/effort wins |
| Groq | GPT-OSS 20B/120B → `low`; Qwen 3.8 27B → `none` |
| Moonshot | Kimi K3 → `low` |
| xAI Responses | Grok 4.5/4.6 → `low` |

The shared OpenAI, Google, and Anthropic implementations also serve their corresponding inherited routes. Unknown-model behavior, omitted reasoning, `provider-default`, and existing validation of explicit native options are preserved. No public API changes or generic compatible-adapter changes.

These are observable corrections: fallback calls can still consume reasoning tokens and change output/usage. OpenAI Responses retains its existing `detailed` summary default when the effective effort is non-`none`; explicit `reasoningSummary: null` still opts out. The documentation and patch changeset explain these effects.

## Validation

- Added 174 request-mapping cases across `doGenerate` and `doStream`, covering minima, disabling, explicit overrides, defaults, unknown IDs, and OpenAI summary behavior. Running the initial new regression cases against the original implementation reproduced 82 failures.
- Full Node and Edge suites pass for all six changed providers and five packages inheriting their implementations: Azure, Google Vertex, Anthropic AWS, Amazon Bedrock, and MiniMax (9,236 tests across both runtimes).
- `pnpm check` and `git diff --check` pass.
- `pnpm type-check:full` passes, including examples.

## End-to-End Verification

[...]

Issue #20867, opened 2026-09-16
Normalize reasoning: none to the lowest supported provider setting


gr2m:
## Problem

The shared language-model evaluation adapter introduced in #20858 can request `reasoning: 'none'` and rely on each provider's existing mapping, with explicit `providerOptions` taking precedence. However, the existing mappings do not consistently implement “disable thinking when supported, otherwise use the lowest supported setting.”

This affects ordinary language-model generation as well as evaluation. The fix should live in provider reasoning mappings rather than in evaluation-specific model detection.

## Confirmed gaps in the current implementation

- **OpenAI Responses:** `getOpenAILanguageModelCapabilities` lists GPT-6 reasoning efforts starting at `low`. When top-level `reasoning: 'none'` resolves to an unsupported effort, the Responses implementation emits a warning and sets the effort to `undefined`. The request therefore uses the API default instead of the lowest supported effort. Other recognized model families without a supported-effort list should also be audited for unsupported `none` values.
  - [Capability table](https://github.com/vercel/ai/blob/d4d96bfa7648ba039639a66ed07f8e1ad9b928f3/packages/openai/src/openai-language-model-capabilities.ts)
  - [Responses mapping](https://github.com/vercel/ai/blob/d4d96bfa7648ba039639a66ed07f8e1ad9b928f3/packages/openai/src/responses/openai-responses-language-model.ts#L311-L329)
- **Google Gemini 2.5:** `resolveGemini25ThinkingConfig` unconditionally maps `none` to `thinkingBudget: 0`. Gemini 2.5 Pro cannot disable thinking and has a documented minimum budget of 128, so this mapping can produce an invalid request.
  - [Thinking mapping](https://github.com/vercel/ai/blob/d4d96bfa7648ba039639a66ed07f8e1ad9b928f3/packages/google/src/google-language-model.ts#L1387-L1415)
  - [Google thinking documentation](https://ai.google.dev/gemini-api/docs/generate-content/thinking)

## Proposed scope

- Map the shared `reasoning: 'none'` setting to disabled thinking where supported, or the lowest supported effort/budget where thinking is required.
- Keep explicit provider reasoning options authoritative; distinguish shared-option fallback from validation of an explicitly supplied provider setting.
- Audit Gemini 3 minimum thinking levels and Anthropic models that require thinking, along with OpenAI Chat and any shared/derived provider implementations. These need capability checks rather than assuming one mapping is valid for every model family.
[...]

PR #20902, opened 2026-09-16
fix(providers): resolve reasoning none for evaluation providers

gr2m:
## Background

Language-model evaluations default to `reasoning: 'none'`, but models that require thinking can receive an unsupported value or fall back to the API default. For example, Gemini 2.5 Pro receives a zero thinking budget, while GPT-6 loses the unsupported `none` effort instead of requesting `low`.

This is a narrower alternative to #20872: it implements fallback mappings for OpenAI, Anthropic, and Google, the providers exposing language-model evaluation adapters. TypeSafe's native evaluation API needs no reasoning mapping. It addresses part of #20867 without closing the broader provider audit.

## Summary

- Define `none` as disabling thinking where supported, otherwise requesting the lowest supported reasoning effort or budget. Document that it can still consume reasoning tokens.
- Resolve the setting in OpenAI Chat/Responses, Anthropic Messages, and Google GenerateContent, so evaluation and ordinary generation share the same behavior.
- Use family/version matching so newer GPT, Claude Fable/Mythos, and Gemini versions inherit the latest known behavior. Retain older-model exceptions, including GPT-5 Pro, Codex Mini, Gemini 2.5 Pro, and earlier Gemini Flash models. Future compatibility assumes the family retains its current controls; unrecognizable custom deployment IDs retain existing behavior.
- Preserve explicit provider reasoning controls and omitted/provider-default behavior. Prevent Google explicit budgets or levels from combining with conflicting shared controls, while allowing display-only options. Correct Gemini Pro's shared `minimal` level to `low` as well.
- Add request-level regression coverage for evaluation, generation, streaming, future version numbers, aliases, overrides, and defaults, plus patch changesets for the three providers.

The shared adapters also serve corresponding Azure, Google Vertex, Anthropic AWS, and Bedrock routes. This does not add mappings to separate endpoints such as Bedrock Converse or Google Interactions. OpenAI Responses retains its existing `detailed` summary default for resolved non-`none` effort; `reasoningSummary: null` opts out.

## Validation

- Node and Edge suites passed for the three changed providers and five inheriting packages: Azure, Google Vertex, Anthropic AWS, Amazon Bedrock, and MiniMax (7,986 tests across both runtimes).
- New regression tests failed against the original provider implementation.
- `pnpm type-check:full` passed, including examples.
- Formatting/lint checks and `git diff --check` passed.

## End-to-End Verification

No live provider API calls were made. Provider documentation and OpenAI's model metadata were reviewed to establish the mappings. Live API acceptance, particularly for newly released models, remains unverified.

## Checklist

## Future Work

Extend the same `none` semantics to the remaining providers and endpoints tracked in #20867.

## Related Issues

Related to #20867. Alternative to #20872. Builds on #20868.


gr2m:
Superseded by #20906, which keeps reasoning defaults inside the experimental evaluation implementation using explicit supported-model catalogs.

PR #20997, opened 2026-09-17
fix(replicate): poll image predictions

### LINKED ISSUE #16276: @ai-sdk/replicate image model fails when sync-wait returns status:starting with output:null

n1ckoates:
## Background

Replicate can return an incomplete prediction with a polling URL when image generation takes a while or Replicate is experiencing heavy load. The AI SDK Replicate provider didn't handle this and expected immediate output.

* https://replicate.com/docs/topics/predictions/create-a-prediction/#polling
* https://replicate.com/docs/topics/predictions/create-a-prediction/#sync-mode

## Summary

- Poll incomplete Replicate image predictions until output is available.
- Complete Replicate image predictions are immediately passed through.
- Added provider options to configure polling.

## End-to-End Verification

- Ran `sync-wait-polling.ts` against Replicate’s API using `black-forest-labs/flux-dev`.
  - The example forces polling by limiting the response wait time to 1 second. 
- The image completed successfully after 7 polling requests.

## Checklist

## Related Issues

Fixes #16276


karaaslanz (commented):
`pollTimeoutMillis` is documented as the maximum time to poll, but the current implementation only turns it into a maximum attempt count (`ceil(timeout / interval)`). Time spent inside each `getFromApi()` call is not included in that budget, and `getFromApi()` receives only the caller's `abortSignal`, not a deadline derived from `pollTimeoutMillis`.

That means a slow poll response can push the operation well past the configured timeout, and a hung poll request can prevent this timeout from firing at all. The current timeout regression only uses immediate mock responses, so it verifies the number of GETs rather than the wall-clock contract.

I think this needs an actual deadline/remaining-time budget (for example, derive an abort signal/deadline for each poll request and delay from the configured timeout), plus a regression where a polling request itself is delayed beyond `pollTimeoutMillis` and the operation still terminates within the configured bound.

Source-level review only; I did not run the branch locally.

karaaslanz (commented):
Follow-up review of head `258acf957ef495131a66a0399da56c4dee9a9ed7`: this resolves my timeout-contract concern. The public option is now `maxPollAttempts` rather than a wall-clock `pollTimeoutMillis`, the implementation enforces exactly that attempt bound, and the regression now asserts the corresponding call-count/error contract. I did not run the branch locally.

n1ckoates:
@karaaslanz I agree. I matched other providers for consistency's sake, but I renamed it to `maxPollingAttempts` here to describe what it actually does. Opened an issue to correct that in other providers: https://github.com/vercel/ai/issues/21000

gr2m (approved):

PR #21139, opened 2026-09-18
fix(deepseek): return an AI SDK error for empty chat completion choices

### LINKED ISSUE #21138: DeepSeek non-streaming crashes with raw TypeError on empty choices
Hey! Found that the DeepSeek provider's non-streaming doGenerate crashes with a raw TypeError when the API returns empty choices (happens with content filters), instead of a typed AI SDK error like the streaming path and the OpenAI provider produce. Happy to PR a fix if that helps.

soroush5:
Non-streaming doGenerate crashed with a raw TypeError on empty choices instead of a typed AI SDK error. It now rejects with InvalidResponseDataError, mirroring the OpenAI provider fix in #20513.

Verified: new test fails before and passes after, package suite green.

Fixes #21138


soroush5:
Closing as superseded by #21144 which covers the same fix.

PR #21163, opened 2026-09-18
fix(bytedance): add `reference_image` role to starting image

### LINKED ISSUE #15382: ai-gateway: bytedance/seedance-2.0 + providerOptions.bytedance.referenceImages fails upstream with "role must be specified for image contents"
## Summary

When using `bytedance/seedance-2.0` (or `bytedance/seedance-2.0-fast`) through the Vercel AI Gateway with `providerOptions.bytedance.referenceImages`, the upstream ByteDance ModelArk API rejects the request with `400 "The parameter content specified in the request is not valid: role must be specified for image contents."`

The gateway translates `referenceImages: string[]` into the upstream `content` array but does **not** add the per-item `role` tag that ByteDance's v2.0 endpoint requires. The same `referenceImages: string[]` shape works correctly via the `@ai-sdk/bytedance` provider package directly (which adds the role internally), so the bug is in the gateway's translation layer for the v2.0 model family.

## Minimal repro

\`\`\`ts
import { experimental_generateVideo as generateVideo } from "ai";

await generateVideo({
  model: "bytedance/seedance-2.0",
  prompt: {
    image: "https://<public-blob-url>/background.jpg",
    text: "Animate this scene with [Image 1] as the subject.",
  },
  aspectRatio: "9:16",
  duration: 4,
  resolution: "480x854",
  providerOptions: {
    bytedance: {
      referenceImages: ["https://<public-blob-url>/character.jpg"],
    },
  },
});
\`\`\`

**Response (HTTP 400 from upstream, surfaced by the gateway):**

\`\`\`json
{
  "type": "error",
  "message": "The parameter \`content\` specified in the request is not valid: role must be specified for image contents.",
  "errorType": "AI_APICallError",
  "statusCode": 400
}
\`\`\`

## Variations tried (same upstream error on all URL-based attempts)

| Model id                       | \`referenceImages\` shape                                | Result                                                                          |
| ------------------------------ | ------------------------------------------------------ | ------------------------------------------------------------------------------- |
| \`bytedance/seedance-2.0\`       | \`string[]\` of public URLs                              | 400 \`role must be specified for image contents\` (upstream)                      |
| \`bytedance/seedance-2.0-fast\`  | \`string[]\` of public URLs                              | 400 \`role must be specified for image contents\` (upstream)                      |
| \`bytedance/seedance-2.0\`       | \`string[]\` of base64 data URLs                         | 400 \`invalid bytedance provider options\` (gateway)                              |
| \`bytedance/seedance-2.0\`       | \`[{ role: "reference_image", image_url: { url } }]\`    | 400 \`invalid bytedance provider options\` (gateway schema rejects shape)         |
| \`bytedance/seedance-v1.0-lite-i2v\` | \`string[]\` of public URLs                          | works (documented, doesn't help if you need v2.0)                               |

## Counter-proof — \`@ai-sdk/bytedance\` direct works

The official \`vercel/ai\` repo's [\`examples/ai-functions/src/generate-video/bytedance/seedance-2-0-i2v.ts\`](https://github.com/vercel/ai/blob/main/examples/ai-functions/src/generate-video/bytedance/seedance-2-0-i2v.ts) and [\`seedance-2-0-reference-audio.ts\`](https://github.com/vercel/ai/blob/main/examples/ai-functions/src/generate-video/bytedance/seedance-2-0-reference-audio.ts) use the same \`providerOptions.bytedance.referenceImages: string[]\` shape against \`byteDance.video('dreamina-seedance-2-0-260128')\` (the direct BytePlus model id) and it works there. So the API surface is correct; the issue is specifically that the gateway translation layer for the \`bytedance/seedance-2.0*\` model ids doesn't add the role tag the v2.0 upstream expects.

## Expected

\`providerOptions.bytedance.referenceImages: string[]\` should work for \`bytedance/seedance-2.0\` and \`bytedance/seedance-2.0-fast\` the same way it does for \`bytedance/seedance-v1.0-lite-i2v\` (already documented in the [image-to-video docs page](https://vercel.com/docs/ai-gateway/capabilities/video-generation/image-to-video#seedance-multi-reference-images)). The gateway should add \`role: "reference_image"\` to each item when forwarding to ByteDance's v2.0 endpoint.

A documentation update to the [reference-to-video page](https://vercel.com/docs/ai-gateway/capabilities/video-generation/reference-to-video) would also help — it currently only documents Wan; users hit this trying to use Seedance 2.0 (which the model page advertises as supporting multimodal reference-to-video).

## Environment

- \`ai\` v6.x
- \`@ai-sdk/gateway\` (whatever ships with \`ai\` v6)
- Gateway endpoint: \`https://ai-gateway.vercel.sh/v3/ai/video-model\`
- Reference images: public URLs hosted on Vercel Blob

n1ckoates:
## Background

In AI SDK, Seedance 2.0 rejects requests that combine a prompt image with reference media because every image in a must have a `role`.

The ByteDance provider left the prompt image untagged, causing:

```text
role must be specified for image contents
```

See the [BytePlus Seedance video generation documentation] (https://docs.byteplus.com/en/docs/Byteplus_LAS/video_gen_enhanced).

## Summary

Send the prompt image as `reference_image` when other reference content is present.

## End-to-End Verification

Verified directly against the ByteDance API using Seedance 2.0:

- Reproduced the original `role must be specified for image contents` error.
- Confirmed that sending the prompt image as `reference_image` successfully starts the generation without warnings.

## Checklist

## Related Issues

Fixes https://github.com/vercel/ai/issues/15382


gr2m (approved):
maybe add an example?

PR #21184, opened 2026-09-19
fix(groq): return a typed error for non-streaming responses without choices

### LINKED ISSUE #21183: Groq non-streaming crashes with raw TypeError on empty choices
Hey! Found that the Groq provider's non-streaming doGenerate crashes with a raw TypeError when the API returns empty choices, instead of a typed AI SDK error like the streaming path produces. Happy to PR a fix if that helps.

soroush5:
Non-streaming doGenerate crashed with a raw TypeError on empty choices instead of a typed AI SDK error. It now rejects with InvalidResponseDataError, mirroring the OpenAI provider fix in #20513.

Verified: new test fails before and passes after, package suite green.

Fixes #21183


soroush5:
Closing since #21186 already merged this fix. Thanks for confirming the bug.

PR #21204, opened 2026-09-20
fix(open-responses): send `json_object` for schema-less JSON response format

### LINKED ISSUE #21202: @ai-sdk/open-responses: schema-less JSON output sends a bare `{ type: "json_schema" }`, which the Responses API rejects
## Description

When `responseFormat.type === 'json'` with no `schema` (i.e. `Output.json()` / `experimental_output: Output.json()`), `@ai-sdk/open-responses` sends:

```json
"text": { "format": { "type": "json_schema" } }
```

`name` and `schema` are omitted, but the Responses API requires them whenever `format.type` is `json_schema`. Every request fails with a 400:

```json
{
  "error": {
    "message": "Missing required parameter: 'text.format.name'.",
    "type": "invalid_request_error",
    "param": "text.format.name",
    "code": "missing_required_parameter"
  }
}
```

The effect is that schema-less JSON mode is unusable on this provider — the request never reaches the model.

### Cause

`packages/open-responses/src/responses/open-responses-language-model.ts:257` (unchanged as of 2.0.49):

```ts
const textFormat =
  responseFormat?.type === 'json'
    ? {
        type: 'json_schema' as const,
        ...(responseFormat.schema != null
          ? {
              name: responseFormat.name ?? 'response',
              description: responseFormat.description,
              schema: responseFormat.schema,
              strict: true,
            }
          : {}),      // <-- no schema: emits `{ type: 'json_schema' }` with nothing else
      }
    : undefined;
```

The spread means the `json_schema` discriminator is emitted unconditionally, while the fields that make it valid are emitted only when a schema is present. There is no code path that produces a valid schema-less request.

### Expected behavior

Schema-less JSON should map to `{ type: 'json_object' }`, which is what the Responses API provides for exactly this case. `@ai-sdk/openai` already does this in `openai-responses-language-model.ts` — same API, same call options, different mapping:

```ts
format:
  responseFormat.schema != null
    ? {
        type: 'json_schema',
        strict: strictJsonSchema,
        name: responseFormat.name ?? 'response',
        description: responseFormat.description,
        schema: responseFormat.schema,
      }
    : { type: 'json_object' },
```

Porting that shape to `open-responses` should be sufficient. Happy to open a PR if that's the direction you'd like.

### Notes

- Not gateway-specific. The invalid body is built before the request leaves the SDK, so any endpoint that validates `text.format` rejects it. I hit it against Bedrock's `/openai/v1/responses` endpoint with a GPT model; a strict self-hosted or LM Studio endpoint should behave the same.
- `Output.object({ schema })` is unaffected — the schema branch produces a valid body.

## Reproduction

```ts
import { createOpenResponses } from '@ai-sdk/open-responses';
import { Output, streamText } from 'ai';

const provider = createOpenResponses({
  baseURL: 'https://<your-responses-gateway>/openai/v1',
  apiKey: process.env.API_KEY!,
});

const result = streamText({
  model: provider('<model-id>'),
  prompt: 'Return {"ok": true} as JSON.',
  output: Output.json(),   // no schema
});

// 400 AI_APICallError: Missing required parameter: 'text.format.name'
for await (const part of result.textStream) process.stdout.write(part);
```

Request body the SDK produces (trimmed):

```json
{
  "model": "<model-id>",
  "input": [ ... ],
  "text": { "format": { "type": "json_schema" } },
  "stream": true
}
```

Swapping `Output.json()` for `Output.object({ schema })` succeeds against the same endpoint, which isolates the failure to the schema-less branch.

## AI SDK Version

- `ai`: 7.0.91
- `@ai-sdk/open-responses`: 2.0.45 — also verified present in 2.0.49 (current `latest`)

lavisht22:
Fixes #21202.

## Problem

When `responseFormat` is `{ type: 'json' }` with no `schema` — what `Output.json()` produces — the provider emitted a bare `{ type: 'json_schema' }` text format:

```json
"text": { "format": { "type": "json_schema" } }
```

The `json_schema` discriminator was emitted unconditionally, but `name` and `schema` only when a schema was present, so there was no code path that produced a valid schema-less request. The Responses API rejects it with a 400:

```
Missing required parameter: 'text.format.name'
```

Schema-less JSON output was therefore unusable on this provider — the request never reached the model.

## Fix

Map the schema-less case to `{ type: 'json_object' }`, which is what the Responses API provides for exactly this case. This mirrors `@ai-sdk/openai`'s handling of the same call options in `openai-responses-language-model.ts`, and matches the `JsonObjectResponseFormat` / `JsonSchemaResponseFormat` types this package already declares in `open-responses-api.ts` — the previous output satisfied neither (`JsonSchemaResponseFormat` requires `name` and `schema`).

The schema branch is unchanged; only the `schema == null` path differs.

## Test

Adds a regression test asserting the request body for `responseFormat: { type: 'json' }`. Verified red before the change:

```
- "type": "json_object",
+ "type": "json_schema",
```

and green after. Full package suite passes: 107 tests across node and edge configs.

A patch changeset is included.


lavisht22:
Closing — this was fixed in #21206 (released in @ai-sdk/open-responses@2.0.50, backported to v6 in #21336), which makes the same change. Thanks for the quick turnaround!

Issue #4125, opened 2024-12-18
Request for Image Embedding Support


faradaytrs:
### Feature Description

Currently, the Vercel AI SDK only supports text embedding models. We would like to request the addition of support for creating embeddings for images as well.

As image-based machine learning and multi-modal AI applications become more common, the ability to generate embeddings for images would greatly enhance the SDK’s capabilities. This feature would allow developers to leverage image similarity, classification, and other AI-driven tasks directly within the Vercel ecosystem.

### Use Cases

1. Image Similarity: Create vector embeddings from images and compare them for similarity.
2. Multi-modal Applications: Combine text and image embeddings for more comprehensive AI models (e.g., image captioning, visual question answering).
3. Image Classification: Use embeddings for classification tasks where images need to be mapped to specific categories.

### Additional context

We believe this feature would provide significant value and open up new possibilities for developers using the Vercel AI SDK. Looking forward to seeing this feature in a future release!

faradaytrs:
> Please list providers that support image embeddings.

For example Amazon Titan Multimodal Embeddings G1 model can do it, there are other models as well, if you need more i can make a research.

lgrammel:
We'll probably hold back on this for now until major providers such as OpenAI support it.

cjlryan:
Would also love this - currently have to duck out of the AI SDK for multimodal embeddings.

bastotec:
> Would also love this - currently have to duck out of the AI SDK for multimodal embeddings through VertexAI.

Same here

BryceWG:
> We'll probably hold back on this for now until major providers such as OpenAI support it.

Mainstream embedding vendors such as Jina, Voyage, and Cohere have all provided multimodal embedding models, while OpenAI has completely stopped focusing on other types of models.

lgrammel:
### Identify issue type

**Identified type: Feature**  
Confidence: High

Rationale: The issue asks for new image/multimodal embedding support in the AI SDK, not a failure of existing documented behavior.

Evidence:
- The description explicitly requests adding support for creating embeddings for images because the SDK currently only supports text embedding models.
- Use cases describe new capabilities such as image similarity, multimodal applications, and image classification.
- Comments discuss providers/models that support image or multimodal embeddings and users wanting this capability.

PR #4050, opened 2024-12-10
fix (provider/google-vertex): Remove unsupported cache ctl setting.

shaper:
Vertex Anthropic does not currently appear to support the beta cache-control feature.

PR #21215, opened 2026-09-21
feat(provider/gateway): type conditional evaluation fallbacks

R-Taneja:
## Why

Evaluation callers need type inference and preflight validation for the conditional fallback contract implemented by the Gateway. Existing string model fallbacks must remain source compatible, and the option stays under `providerOptions.gateway.models`.

Tracks AIG-961. The runtime dependency is the conditional fallback routing implemented in the Gateway service.

## Wire contract

```ts
providerOptions: {
  gateway: {
    models: [
      {
        model: "openai/gpt-5.6-sol",
        when: {
          atLeast: {
            count: 2,
            conditions: [
              { question: "intent", confidenceBelow: 0.6 },
              {
                question: "refunded",
                probabilityBetween: [0.4, 0.6],
              },
              { question: "severity", confidenceBelow: 0.7 },
            ],
          },
        },
      },
      "anthropic/claude-sonnet-5",
    ],
  },
}
```

Choice and Score questions use `confidenceBelow`. Boolean questions use inclusive `probabilityBetween` bounds over P(true). Conditions can compose with `any`, `all`, and `atLeast`. V1 permits one conditional entry in the first position, followed by ordinary string error fallbacks. Thresholds are finite and inside `[0, 1]`, ranges are ordered, groups are nonempty, counts are bounded by their group, and condition depth is capped at five.

## What changed

- exports the bounded `EvaluationFallbackCondition` and backward-compatible `GatewayModelFallback` types
- validates conditional Gateway options before evaluation requests while preserving service-owned options
- keeps existing `models: string[]` usages valid
- adds runtime schema coverage, type tests, provider docs, and a focused evaluation example
- does not add a top-level `evaluate` fallback or execute fallback routing in the SDK

## Validation

- `pnpm test` in `packages/gateway` (594 node tests, 586 edge tests plus 8 skips)
- `pnpm type-check` in `packages/gateway`
- `pnpm build` in `packages/gateway`
- `pnpm validate:docs`
- `pnpm type-check:full`
- `pnpm check`


joemckenney:
🛑 The SDK reports the primary model after fallback and rejects a condition depth accepted by the server.
Agent-assisted review on behalf of @joemckenney. Human review/approval is still pending.

- **[blocking]** [`gateway-evaluation-model.ts:99–103,164–184`](https://github.com/vercel/ai/blob/52f056e4967f68dacecf9dbad1e919f7da8f1f22/packages/gateway/src/gateway-evaluation-model.ts#L99-L103): the response schema omits `model`, and the adapter always sets `response.modelId` to the requested primary. Fallback answers and usage therefore carry incorrect standard model attribution. Parse an optional returned `model` and use `responseBody.model ?? this.modelId`; test different primary/final models and older responses without the field.
- **[blocking]** [`gateway-provider-options.ts:6,146–153,227–233`](https://github.com/vercel/ai/blob/52f056e4967f68dacecf9dbad1e919f7da8f1f22/packages/gateway/src/gateway-provider-options.ts#L146-L153): runtime validation and the bounded type chain accept only four condition levels, while the server contract accepts five. Both count the root as level 1 and include the terminal direct condition. Four nested groups plus a direct condition are valid on the server but fail SDK validation before HTTP. Align both limits and add tests accepting level 5 and rejecting level 6.
- **[suggestion]** Remove non-public tracker and dependency links from this public PR description before merge.

Static review; tests and provider requests were not run. The changeset is present, and existing string-only fallback assignments remain accepted.

R-Taneja:
Addressed the blocking feedback in 2f9661c:

**modelId attribution** (`gateway-evaluation-model.ts`): the response schema now parses an optional `model` field and `response.modelId` uses `responseBody.model ?? this.modelId`, so answers and usage are attributed to the model that actually served the request after a fallback. Added tests for the returned model differing from the requested primary and for older responses without the field falling back to the requested id.

**Condition depth** (`gateway-provider-options.ts`): aligned both limits with the server contract at five levels (root counts as level 1, terminal direct condition included). `EVALUATION_FALLBACK_MAX_CONDITION_DEPTH` is now 5 and the recursive type chain gained `EvaluationFallbackConditionAtDepth5`, so four nested groups plus a direct condition pass SDK validation instead of failing before the request leaves the client. Added a runtime test accepting a level 5 condition and rejecting level 6, and extended the type tests to the new boundary.

**Description**: removed the Linear link and the private ai-gateway PR reference, and updated the wire contract text to say depth is capped at five.

Validation: `pnpm test` in `packages/gateway` (594 node tests, 586 edge tests plus 8 skips), `pnpm type-check`, and `pnpm build` all pass.

joemckenney:
👀 Both implementation blockers are fixed; two small wording cleanups remain.
Agent-assisted follow-up on behalf of @joemckenney. Human review/approval is still pending.

Confirmed that optional returned `model` now drives `response.modelId`, with the requested ID retained for older responses. Runtime validation and types both accept depth five and reject six. The added tests exercise the returned-model difference, legacy response shape, and explicit five/six-level boundaries.

- **[suggestion]** Remove the remaining tracking sentence in the second paragraph under "Why" from the public PR description. The links are gone, but the non-public identifier remains.
- **[suggestion]** [`content/docs/03-ai-sdk-core/32-evaluation.mdx:248–250`](https://github.com/vercel/ai/blob/2f9661c3473f15010e079947c72c12726553231b/content/docs/03-ai-sdk-core/32-evaluation.mdx#L248-L250): qualify "no automatic model substitution" as SDK-core behavior, consistent with the provider-routing explanation that follows.

Static follow-up; tests, builds, and provider requests were not rerun.

R-Taneja:
Both cleanups are in:

**Description**: removed the tracking sentence carrying the internal issue identifier from the "Why" section. The runtime dependency note stays as plain prose, and everything else is unchanged.

**Docs** (`content/docs/03-ai-sdk-core/32-evaluation.mdx`): "no automatic model substitution" is now qualified as Core behavior ("Core does not substitute models automatically"), consistent with the provider-routing explanation in "Scope and examples" noting that a provider can route through its provider options. Validated with `pnpm validate:docs`.

shaper (commented):
👀 Faithful, well-tested port of the gateway contract with no confidentiality leaks; the depth constant and the `has: 'structured-output'` drive-by both depend on unmerged gateway PRs and need explicit sequencing.

- **[suggestion]** `packages/gateway/src/gateway-provider-options.ts:6` — `EVALUATION_FALLBACK_MAX_CONDITION_DEPTH = 5`, but the gateway on `main` (`lib/gateway/evaluation-fallback-schema.ts:3`) is 4; only ai-gateway#4125 bumps it to 5. Until #4125 ships, a depth-5 condition passes SDK preflight and type-checks, then 400s at the gateway — which defeats the preflight's purpose. Either don't publish before #4125 deploys (say so in the body) or land at 4 and bump later.
- **[suggestion]** `gateway-provider-options.ts:46-56` — `has` gains `'structured-output'`. Unrelated to the PR's scope, and the gateway only accepts it after ai-gateway#4104 (open). Split it out or note the dependency; the type already over-promises `reasoning`/`tool-use` (pre-existing, #4050 open) so this widens the gap.
- **[suggestion]** `gateway-provider-options.ts:64-69` — conditional entries are typed on the shared `GatewayProviderOptions`, so `generateText({ providerOptions: { gateway: { models: [{ model, when }] } } })` type-checks and only fails at the gateway. A dedicated `GatewayEvaluationProviderOptions` would catch it at compile time; fine as a follow-up.
- **[suggestion]** `gateway-evaluation-model.ts:49-57` — client-side `.strict()` + superRefine on `models` is fail-closed, which cuts against the JSDoc promise that service-owned options evolve without an SDK release. Acceptable for v1; state the trade-off in the type comment.
- **[suggestion]** `gateway-evaluation-model.test.ts:172-204` — no test that service-owned keys round-trip through `doEvaluate` (`catchall` only tested at the schema level). One wire assertion would lock in the claim.
- **[nit]** `examples/ai-functions/src/evaluate/gateway/basic.ts:28` uses `typesafe-ai/jev`; docs (`32-evaluation.mdx:135`) and fixtures use `typesafe-ai/jev-latest`. Pick one.
- **[nit]** `content/providers/01-ai-sdk-providers/00-ai-gateway.mdx:1336` — shows `GatewayModelFallback[]` but only one conditional entry at position 0 is allowed; consider `[GatewayModelFallback, ...string[]]`.
- **[nit]** Changeset `patch` for `@ai-sdk/gateway` matches precedent. Fine.

Confidentiality: clean — body, changeset, comments, and docs describe externally observable behavior only. Base `main` is correct (`EvaluationModelV4` exists only on spec v4); no v5/v6 backport needed for `models`.

R-Taneja:
Merged `origin/main` in eb8a7d0 to clear the conflict on `has`: the union keeps main's `quantization:${string}` / `!quantization:${string}` conditions plus `'structured-output'`, and main's has type test now includes it.

- **Depth constant 5**: Kept at 5. The gateway moves to 5 in [vercel/ai-gateway#4118](https://github.com/vercel/ai-gateway/pull/4118), the bottom PR of the evaluation fallback stack. The body now has a "Release dependency" section saying this must not be released before that stack is deployed.
- **`has: 'structured-output'`**: Kept. The body notes the gateway accepts it through [vercel/ai-gateway#4104](https://github.com/vercel/ai-gateway/pull/4104), which is based on `main` and independent of the router stack. The pre-existing `reasoning`/`tool-use` values depend on [vercel/ai-gateway#4050](https://github.com/vercel/ai-gateway/pull/4050), no change. Docs now list `'structured-output'` too (5804e54).
- **Dedicated evaluation options type**: Fixed in 185fcd9. `GatewayProviderOptions` goes back to its released non-generic shape with `models?: string[]`, and the new `GatewayEvaluationProviderOptions<QUESTION_ID>` keeps every other option typed and allows one leading conditional entry. A conditional entry checked against `GatewayProviderOptions` is now a compile error (type test added). Nothing released changes, since the generic and widened `models` only existed on this branch.
- **`.strict()` + superRefine trade-off**: Fixed in 185fcd9. The `GatewayEvaluationProviderOptions` JSDoc says `models` is validated strictly before sending, so new condition shapes need an SDK release, and other keys pass through unchanged.
- **Service-owned keys through `doEvaluate`**: Fixed in 83e1cff. The test sends conditional fallbacks with `order`, a nested service-owned key, and another provider's options, and asserts the exact request body. Removing the catchall makes it fail.
- **Example model id**: Fixed in 5804e54. `typesafe-ai/jev` everywhere (core docs, gateway fixtures, core registry test). That is the ID in the generated `GatewayEvaluationModelId`, which moved off `jev-latest` in [#20987](https://github.com/vercel/ai/pull/20987).
- **Docs `GatewayModelFallback[]`**: Fixed in 5804e54. Now `string[] | [GatewayModelFallback, ...string[]]` with prose that only the first entry can be conditional and at most one is allowed.
- **Changeset**: Kept.

Also in this round:
- The preflight schema applies the bounds shared with the gateway and VMC storage (185fcd9): 1 to 20 conditions per `any`/`all`/`atLeast.conditions` list, depth 5, `question` and conditional `model` 1 to 256 characters, conditional entry first and at most one. Tests cover each bound at the limit and one past it.
- `providerMetadata.gateway.routing.modelAttempts` is now typed (a74342a): `GatewayRoutingMetadata`, `GatewayModelAttemptMetadata` with `generationId`, `usage`, `triggeredBy: { question, reason }[]` and the per-attempt cost fields, plus `GatewayEvaluationFallbackTrigger`. Field names match what [vercel/ai-gateway#4127](https://github.com/vercel/ai-gateway/pull/4127) emits, and both types keep a JSON index signature for gateway additions.

Validated with `pnpm type-check` and `pnpm test` in `packages/gateway` (610 node tests, 602 edge plus 8 skips), the core evaluation registry test, `pnpm check`, and `pnpm validate:docs`.

PR #21222, opened 2026-09-21
fix(tui): delete complete graphemes on backspace

### LINKED ISSUE #21225: TUI Backspace can submit half of an emoji in the prompt
## Description

Backspace in the TUI prompt removes one UTF-16 code unit. For an emoji such as `😀`, one keypress leaves its high surrogate behind. Pressing Enter then submits that leftover character as part of the user message.

A proposed fix and regression tests are available in https://github.com/vercel/ai/pull/21222.

### Steps to reproduce

1. Start an interactive session with `runAgentTUI`.
2. Type or paste `hello 😀` into the prompt.
3. Press Backspace once.
4. Press Enter.

Expected: the submitted text is `hello `.

Actual: the submitted text is `hello \ud83d`. The final code unit is an unpaired surrogate; UTF-8 encoding displays it as a replacement character.

This also affects characters made of several code points. For example, deleting `e\u0301` removes the accent but leaves `e`, and deleting a joined emoji leaves a partial sequence. The input should remove the last grapheme cluster, preserving everything before it.

The relevant handler is the `backspace` case in `packages/tui/src/tui/terminal-renderer.ts`. It assigns `this.#inputText.slice(0, -1)`, and the Enter handler resolves that same input string. Both supported Backspace encodings (`\u007f` and `\b`) reach this handler.

## Reproduction

The linked PR adds cases to `packages/tui/src/tui/terminal-renderer.test.ts` using the existing terminal input/output mocks. They emit input, Backspace, and Enter events and inspect the returned prompt, so the assertion covers the submitted value as well as key handling.

The new cases reproduce the problem when the original deletion line is restored. With the fix, the complete TUI suite passes (117 tests). The submitted values were also checked through `runAgentTUI` in a Windows terminal using a local echo transport, without a model API call.

## AI SDK Version

- `@ai-sdk/tui`: `1.0.108` in the repository at `f39733ba5f86d434cf00eae56cf4649c34b3d8b1`
- Node.js: `24.11.1`
- Windows

## Code of Conduct

dakjdakd:
## Background

Typing `hello 😀`, pressing Backspace once, and then submitting the prompt leaves `hello \ud83d` instead of `hello `. The renderer removes one UTF-16 code unit with `slice(0, -1)`, so the emoji loses only half of its surrogate pair. The leftover code unit is part of the submitted prompt, not just a display artifact.

Deleting by code point alone would still split characters such as `👩‍💻`, `👍🏽`, and `e\u0301`. The input needs to remove the last grapheme cluster.

## Summary

Use the final grapheme boundary from `Intl.Segmenter` when handling Backspace. Both supported key encodings (`DEL` and `BS`) use this path. Empty input stays empty, and earlier text is preserved.

The regression cases exercise input events through `TerminalRenderer.readPrompt`, then assert the submitted string. They cover ASCII, Chinese, emoji, skin tones, joined emoji, flags, combining accents, an empty prompt, and deleting an initial prompt before typing again. This uses the built-in segmenter available in the package's Node.js 22+ runtime and adds no dependency. The TUI TypeScript config includes `ES2022.Intl` so that the existing runtime API is also available to the type checker.

## End-to-End Verification

Ran the built package through `runAgentTUI` in a Windows terminal with a local echo transport. After typing `hello 😀`, Backspace, and Enter, the transport received `hello `. Repeating with `你好👩‍💻` and `cafe\u0301` produced `你好` and `caf`. All three submitted strings were well-formed Unicode. This exercises the default renderer and the prompt-to-transport path without a model API call.

## Validation

- `pnpm --filter @ai-sdk/tui test`: all 117 tests pass, with no type errors.

- `pnpm check`: passes across the repository.

- `pnpm type-check:full`: passes, including examples.

- `pnpm exec turbo build --filter=@ai-sdk/tui...`: all 17 build tasks pass.

- `pnpm --filter @ai-sdk/tui type-check`: passes.

## Checklist

## Related Issues

Fixes #21225


dakjdakd:
The reproduction added in #21227 also passes with this fix.

The original script from `f77ad32b2cee68b623ee4f4718631352e01f7aa9` was copied unchanged into this branch at `7f38c8541c79b26f7631dadb10f765bc48c5fec5` and run with the command in its replay configuration:

```sh
pnpm -C examples/ai-functions exec tsx src/reproduction/issue-21225-tui-backspace-grapheme.ts
```

Output:

```text
Issue #21225 is not present.
```

Exit code: **0**. All four cases pass: emoji deletion with DEL, emoji deletion with BS, a combining accent, and a joined family emoji. Each submits `hello `, as expected. This was run on Windows with Node 24.11.1.

Thanks for adding the independent reproduction. It confirms that the existing grapheme-based fix covers the cases in #21225; no further implementation changes were needed for this replay.

